PR #4924, opened 2024-08-11
[2.6] MOD-7339: Fix field statistics counting bug (#4919)

raz-mon:
Backport of #4919 to 2.6


oshadmi (approved):
Assuming test with replicaof is in another issue

PR #4928, opened 2024-08-14
MOD-7453: Fix tag wildcard

raz-mon:

Fixes a bug in the tag wildcard matcher in which we could have reached an endless loop.
See tests for an example of such a scenario.

We opt to enter the inner loop that handles the wildcard (i.e., `*`) instead of skipping it as if it was a regular character. This is due to the current behavior of the wildcard matcher, in which there are no escaped characters (see [Windows wildcard matching](https://en.wikipedia.org/wiki/Matching_wildcards#:~:text=No%20escape%20characters%20are%20defined)).

**Mark if applicable**

### DIFF
--- a/tests/pytests/test_wildcard.py
+++ b/tests/pytests/test_wildcard.py
@@ -387,3 +387,26 @@ def testSuffixCleanup(env):
   conn.execute_command('DEL', 'doc')
 
   forceInvokeGC(env, 'idx')
+
+def testMOD7453():
+  """Tests that we don't enter an infinite loop when we match a wildcard to a
+    wildcard in the matched term"""
+
+  env = DialectEnv()
+  conn = getConnectionByEnv(env)
+
+  # Create an index with a TEXT and TAG field
+  env.cmd('FT.CREATE', 'idx', 'SCHEMA', 'tag', 'TAG')
+
+  # Populate the db
+  conn.execute_command('HSET', 'doc1', 'tag', 'ba*cl')
+
+  # Search via "problematic" wildcard
+  MAX_DIALECT = set_max_dialect(env)
+  for dialect in range(2, MAX_DIALECT + 1):
+    env.set_dialect(dialect)
+    res = env.cmd('FT.SEARCH', 'idx', "@tag:{w'*a*'}")
+    env.assertEqual(res, [1, 'doc1', ['tag', 'ba*cl']])
+
+    res = env.cmd('FT.SEARCH', 'idx', "@tag:{w'*a*?'}")
+    env.assertEqual(res, [1, 'doc1', ['tag', 'ba*cl']])
--- a/deps/wildcard/wildcard.c
+++ b/deps/wildcard/wildcard.c
@@ -16,7 +16,7 @@ match_t Wildcard_MatchChar(const char *pattern, size_t p_len, const char *str, s
     //printf("%d ", i);
     if (pattern_end != pattern_itr) {
       const char c = *pattern_itr;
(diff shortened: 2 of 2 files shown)


tests/pytests/test_wildcard.py
```diff
@@ -387,3 +387,26 @@ def testSuffixCleanup(env):
...
+
+def testMOD7453():
+  """Tests that we don't enter an infinite loop when we match a wildcard to a
+    wildcard in the matched term"""
+
+  env = DialectEnv()
+  conn = getConnectionByEnv(env)
+
+  # Create an index with a TEXT and TAG field
+  env.cmd('FT.CREATE', 'idx', 'SCHEMA', 'tag', 'TAG')
```
oshadmi, line 399: Also test with `TEXT`?

oshadmi: Also add to [unit test](https://github.com/RediSearch/RediSearch/blob/cd4469dcf1ad9832cdca24368248033cfe44c58c/tests/ctests/test_wildcard.c#L174C5-L174C15)?

raz-mon: I tested locally and there were no problems
I can add some here as well

src/wildcard.c
```diff
@@ -13,34 +13,46 @@ match_t Wildcard_MatchChar(const char *pattern, size_t p_len, const char *str, s
...
-    if (pattern_end != pattern_itr) {
+    if (pattern_end > pattern_itr) {
       const char c = *pattern_itr;
-      if ((str_end != str_itr) && (c == *str_itr || c == '?')) {
+      if ((str_end > str_itr) && (c == *str_itr || c == '?') && (c != '*')) {
+        // Equal characters or '?' match --> advance both pointers
         ++str_itr;
         ++pattern_itr;
         continue;
       } else if (c == '*') {
```
kei-nan, line 23: optional.
since you are checking c != '*' at the start maybe the if which checks c == '*' can be at the start?

src/wildcard.c
```diff
@@ -13,34 +13,46 @@ match_t Wildcard_MatchChar(const char *pattern, size_t p_len, const char *str, s
...
-        while ((pattern_end != pattern_itr) && (*pattern_itr == '*')) {
+        while ((pattern_end > pattern_itr) && (*pattern_itr == '*')) {
+          // Multiple '*' are equivalent to a single '*' --> skip them
           ++pattern_itr;
         }
         const char d = *pattern_itr;
-        while ((str_end != str_itr) && !(d == *str_itr || d == '?')) {
-          ++str_itr;
+        if (d != '?') {
```
kei-nan, line 29: maybe also check that the pattern iterator didn't reach the end?

src/wildcard.c
```diff
@@ -13,34 +13,46 @@ match_t Wildcard_MatchChar(const char *pattern, size_t p_len, const char *str, s
...
+      // are added to the string - partial match
       return PARTIAL_MATCH;
     } else if (ns_itr == NULL) {
+      // Pattern depleted but string not, and no '*' was found -> no match
       return NO_MATCH;
     }
+    // Backtrack
     pattern_itr = np_itr;
     str_itr = ns_itr;
   }
```
kei-nan, line 58: tried commenting on the assert line but it didn't let me...
is the assert(0) needed here?

src/wildcard.c
raz-mon: Right 👍

src/wildcard.c
raz-mon: Why? `\0` is a character as well..

kei-nan (approved):

PR #4933, opened 2024-08-18
MOD-7588: Add warning for reaching `MAXPREFIXEXPANSIONS`

raz-mon:

This PR adds a warning for reaching the `MAXPREFIXEXPANSIONS` in a query to the RESP3 `warning` section.

1. `QueryError` was added the `reached_maxprefixexpansions;` field, indicating that `maxprefixexpansions` was reached.

**Mark if applicable**


src/query_error.h
```diff
@@ -74,6 +74,9 @@ typedef enum {
 typedef struct QueryError {
   QueryErrorCode code;
   char *detail;
+
+  // warnings
+  bool reached_maxprefixexpansions;
```
kei-nan, line 79: This seems to break our naming convention, no?
Do we want underscores or camel case?

src/aggregate/aggregate_exec.c
```diff
@@ -619,6 +619,8 @@ static void sendChunk_Resp3(AREQ *req, RedisModule_Reply *reply, size_t limit,
     } else if (rc == RS_RESULT_ERROR) {
       // Non-fatal error
       RedisModule_Reply_SimpleString(reply, QueryError_GetError(req->qiter.err));
+    } else if (req->qiter.err) {
+      RedisModule_Reply_SimpleString(reply, "Max prefix expansions reached");
```
kei-nan, line 623: Maybe this should be a separate element in the array?
Can't timeout happen and also max prefix expansions reached happen at the same time?

raz-mon: I agree in general that an array makes sense here, as they indeed can both occur for the same execution.
The thing is, that this change would break the response, such that we won't be able to release this in a patch.

Moreover, they have the same effect on the results - truncating them, and you don't know which results didn't come from which error as well. I think that timeout has higher precedence thus it should be reported first, and we should extend this to be an array once we release a breaking change to our response.

src/aggregate/aggregate_exec.c
```diff
@@ -619,6 +619,8 @@ static void sendChunk_Resp3(AREQ *req, RedisModule_Reply *reply, size_t limit,
     } else if (rc == RS_RESULT_ERROR) {
       // Non-fatal error
       RedisModule_Reply_SimpleString(reply, QueryError_GetError(req->qiter.err));
+    } else if (req->qiter.err->reached_maxprefixexpansions) {
+      RedisModule_Reply_SimpleString(reply, "Max prefix expansions limit was reached");
```
oshadmi, line 623: I suggest to clearly add the actual value of the limit (make sure the expected limit is indeed in place)

raz-mon: Since we assume this value is **known**, since the user either explicitly configures it or can get it via the config parameters' values, I think it's quite redundant..

src/index.c
```diff
@@ -1606,7 +1606,7 @@ PRINT_PROFILE_FUNC(printUnionIt) {
   // if MAXPREFIXEXPANSIONS reached
   if (ui->norig == config->iteratorsConfig->maxPrefixExpansions) {
     RedisModule_Reply_SimpleString(reply, "Warning");
-    RedisModule_Reply_SimpleString(reply, "Max prefix expansion reached");
+    RedisModule_Reply_SimpleString(reply, "Max prefix expansions limit was reached");
```
oshadmi, line 1609: Agreeing that rephrasing the error message is not considered a breaking change since it is in `FT.PROFILE`

src/aggregate/aggregate_exec.c
oshadmi: This could reduce ping-pong with users, asking for more info.

src/aggregate/aggregate_exec.c
kei-nan: Looking at the code I think it is already an array, no?
```    RedisModule_Reply_ArrayEnd(reply); // >warnings```

Comment (raz-mon) [reply src/aggregate/aggregate_exec.c]: It is, but populating it with more warnings may be more breaking in a way. 
We plan to have a continuation PR to this PR which will refactor the `QueryError` to be `QueryStatus` and contain a similar API for the warnings as there currently is for the errors. We will also make the warnings array actually able to reply with multiple warnings if breakage will allow.
If this won't be an urgent PR we may also do this here, but seems that we want to have a quick fix that will later be improved.

**Review by kei-nan — APPROVED** [review]

File: src/query.c [hunk 5]
```diff
@@ -1218,8 +1233,12 @@ static IndexIterator *Query_EvalTagWildcardNode(QueryEvalCtx *q, TagIndex *idx,
     void *ptr;
 
     // Find all completions of the prefix
-    while (TrieMapIterator_NextWildcard(it, &s, &sl, &ptr) &&
-          (itsSz < q->config->maxPrefixExpansions)) {
+    while (TrieMapIterator_NextWildcard(it, &s, &sl, &ptr)) {
+      if (itsSz >= q->config->maxPrefixExpansions) {
```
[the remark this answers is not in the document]
raz-mon: Same semantics, but I'm okay with having it outside of the loop.

PR #4935, opened 2024-08-18
Single build path [MOD-7480]

alonre24:

Complete the unified build for the module with the coordinator's component which is now an integral part of the module. This includes:
* Removing the option to build the module without a coordinator (COORD=0), and removing the `RS_COORDINTOR` build flag (which is now always true).
* As for the `LITE` flavor - it is still possible to build an artifact under the name `searchlight` for backward compatibility, **but the behavior will be the same as the non-light module** (that is, the coordinator will take place and start lazily in a clustered environment).  
* Move `coord/module.c` file into the module's `module.c` (module's entry point where `RedisModule_OnLoad` function should be) and resolve dependencies (link the entire module against hiredis/sds.h rather than rmutil/sds.h, and let the CPP tests use it but wrapping it with `extern C`).  
* Let the `cmakelists.txt` file in the project root dir be the single entry point from which the modules (and unit test) artifacts are built (having the `cmakelists.txt` under `coord` a subdirectory only, not an entry point as it was).
* Move coordinator unit tests into the ctests and cpptests dirs and remove test files that are outdated and look more like POCs than tests
* Move ramp files into `pack` dir and rename them according to their purpose (ce, lite, enterprise) 
* Remove redundant `RSBuildType` enum and `build-info` dir, and use `IsEnterprise()` call in runtime to determine the cluster build type instead. 
* Bump libuv version to the latest - 1.48.0 (this also removes warnings in the code build...)

**Mark if applicable**


CMakeLists.txt
```diff
@@ -143,29 +126,56 @@ else() # OSS RediSearch
...
+
+    add_library(testdeps STATIC ${FINAL_OBJECTS} $<TARGET_OBJECTS:redismock>)
+
+    target_link_libraries(testdeps
+            ${LIBUV_LIBS}
+            redisearch-geometry
+            redismock
+            VectorSimilarity
+            ${HIREDIS_LIBS}
+            ${SSL_LIBS})
```
DvirDukhan, line 170: can you explain why we need to link VSS, libuv, hiredis, libssl to the unit tests?

Makefile
```diff
@@ -17,7 +17,7 @@ make setup         # install prerequisited (CAUTION: THIS WILL MODIFY YOUR SYSTE
 make fetch         # download and prepare dependant modules
 
 make build          # compile and link
-  COORD=0|1|oss|rlec  # build coordinator (1|oss: Open Source, rlec: Enterprise) default: oss
+  COORD=1|oss|rlec    # build coordinator (1|oss: Open Source, rlec: Enterprise) default: oss
```
DvirDukhan, line 20:
```suggestion
  COORD=oss|rlec    # build coordinator (1|oss: Open Source, rlec: Enterprise) default: oss
```

Makefile
```diff
@@ -246,8 +224,7 @@ CMAKE_FILES+= \
 	tests/ctests/CMakeLists.txt \
 	tests/cpptests/CMakeLists.txt \
 	tests/cpptests/redismock/CMakeLists.txt \
-	tests/pytests/CMakeLists.txt \
-	tests/c_utils/CMakeLists.txt
+	tests/pytests/CMakeLists.txt
```
DvirDukhan, line 227: not your doing, but do we really need pytsts/CMakeLists.txt?

DvirDukhan (commented):
Nice!
More questions than comments

sbin/unit-tests
```diff
@@ -65,7 +65,7 @@ OSNICK=$($READIES/bin/platform --osnick)
 
 #---------------------------------------------------------------------------------------- Setup
 
-export EXT_TEST_PATH=${BINROOT}/search/example_extension/libexample_extension.so
+export EXT_TEST_PATH=${BINROOT}/search-ce/example_extension/libexample_extension.so
```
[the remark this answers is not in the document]
alonre24: Sort of, it is what we used to call "coord-oss" build (with the new build process changes introduced in this PR)

src/coord/rmr/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 
-get_filename_component(root ../../.. ABSOLUTE)
+get_filename_component(root ${CMAKE_CURRENT_LIST_DIR}/../../.. ABSOLUTE)
```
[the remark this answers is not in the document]
alonre24: I think it is for getting the root dir absolute path and using it later on. I can check if there are other CMAKE conventions to do so

src/coord/CMakeLists.txt
```diff
@@ -42,109 +31,26 @@ add_compile_definitions(
...
-add_compile_options("-fvisibility=hidden")
-
-#set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fvisibility=hidden")
-#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden")
-
 #----------------------------------------------------------------------------------------------
 
 include_directories(
-	${root}/coord
+	${root}/src/coord
```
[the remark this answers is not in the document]
alonre24: Paths under `include_directories` are "directories to those the compiler uses to search for include files. Relative paths are interpreted as relative to the current source directory." - so we are not really including the entire source tree here

CMakeLists.txt
alonre24: Good point, I checked and actually we don't. I removed `testdeps` completely and used only `redismock` instead

Makefile
alonre24: Actually, it seems like none of this is necessary (some leftovers probably...) - removing it all

GuyAv46 (commented):
🧼 🧽

Makefile
```diff
@@ -227,29 +205,6 @@ ifeq ($(STATIC),1)
...
-ifneq ($(NO_TESTS),1)
-CMAKE_FILES+= \
-	deps/googletest/CMakeLists.txt \
-	deps/googletest/googlemock/CMakeLists.txt \
-	deps/googletest/googletest/CMakeLists.txt \
-	tests/ctests/CMakeLists.txt \
-	tests/cpptests/CMakeLists.txt \
-	tests/cpptests/redismock/CMakeLists.txt \
-	tests/pytests/CMakeLists.txt \
-	tests/c_utils/CMakeLists.txt
```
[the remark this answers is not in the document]
alonre24: Removed and all worked smoothly, so I guess it is a leftover...

CMakeLists.txt
```diff
@@ -41,20 +28,36 @@ add_compile_definitions(
     "GIT_VERSPEC=\"${GIT_VERSPEC}\""
     "GIT_SHA=\"${GIT_SHA}\""
     REDISMODULE_SDK_RLEC
-    _GNU_SOURCE)
+    _GNU_SOURCE
+    RS_COORDINATOR)
```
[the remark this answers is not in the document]
alonre24: I guess you are right, I removed it from the code, please review carefully and see that our assumptions are correct (that is, we never need to use the code under #ifndef RS_COORDINATOR even in SA scenario)

pack/ramp-ce.yml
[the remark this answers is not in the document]
alonre24: We can go with community :)

src/module.c
```diff
@@ -997,16 +997,9 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx, RedisModuleString **argv,
 
   RM_TRY(NumericIndexType_Register, ctx);
 
-#ifndef RS_COORDINATOR
-// on a none coordinator version (for RS light/lite) we want to raise cross slot if
-// the index and the document do not go to the same shard
-#define INDEX_ONLY_CMD_ARGS 1, 1, 1
-#define INDEX_DOC_CMD_ARGS 1, 2, 1
```
GuyAv46, line 1004: I think this is still needed for `LITE`, without it commands won't get redirected to the correct shard

src/module.c
```diff
@@ -1081,12 +1069,8 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx, RedisModuleString **argv,
 
   RM_TRY(RedisModule_CreateCommand, ctx, RS_SUGGET_CMD, RSSuggestGetCommand, "readonly", 1, 1, 1);
 
-#ifndef RS_COORDINATOR
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_CURSOR_CMD, RSCursorCommand, "readonly", 2, 2, 1);
```
GuyAv46, line 1085: Same here for `LITE`

src/module.c
```diff
@@ -1117,18 +1101,7 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx, RedisModuleString **argv,
...
-#ifndef RS_COORDINATOR
-  // we are running in a normal mode so we should raise cross slot error on alias commands
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_ALIASADD, AliasAddCommand, "readonly", 1, 2, 1);
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_ALIASADD_IF_NX, AliasAddCommandIfNX, "readonly", 1, 2,
-         1);
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_ALIASUPDATE, AliasUpdateCommand, "readonly", 1, 2, 1);
-
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_ALIASDEL, AliasDelCommand, "readonly", 1, 1, 1);
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_ALIASDEL_IF_EX, AliasDelIfExCommand, "readonly", 1, 1,
-         1);
```
GuyAv46, line 1130: Same here for `LITE`

CMakeLists.txt
```diff
@@ -143,29 +126,46 @@ else() # OSS RediSearch
...
+            ${CMAKE_LD_LIBS})
 
     extract_debug_symbols(redisearch)
 
     add_dependencies(redisearch VectorSimilarity)
+
+endif()
+
+if (${COORD_TYPE} STREQUAL "oss")
+    target_compile_definitions(redisearch-coord PRIVATE RS_CLUSTER_OSS)
```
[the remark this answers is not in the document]
GuyAv46: We can. It can wait for when we merge the `module.c` files (or at leat the `OnLoad` function)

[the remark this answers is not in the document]
GuyAv46: But the one we are using is still split into 2 parts

DvirDukhan (approved):

GuyAv46 (approved):

tests/ctests/coord_tests/CMakeLists.txt
```diff
@@ -1,13 +1,15 @@
 
-get_filename_component(root ../../../.. ABSOLUTE)
-message("# coord/rmr/test: root: " ${root})
+if (NOT TEST_MODULE)
+    set(TEST_MODULE redisearch)
+endif()
 
 include_directories(${root}/src/coord/rmr)
+add_compile_definitions(RMUTIL_NO_SDS)
```
GuyAv46, line 7: still needed?

alonre24: no, removing

sbin/unit-tests
```diff
@@ -142,22 +142,22 @@ if [[ $CPP_TESTS == 1 ]]; then
...
-			TEST_NAME="$(basename $test)" sanitizer_defs
-			{ $test; (( E |= $? )); } || true
-		done
-	else
-		find $TESTS_DIR -name "test_*" -type f -executable -print
-	fi
-else
-	$OP $GDB_CMD $TESTS_DIR/$TEST
-fi
+for TESTS_DIR in "$(cd $BINROOT/search-community/tests/ctests/coord_tests; pwd)" "$(cd $BINROOT/search-community/tests/cpptests; pwd)"; do
```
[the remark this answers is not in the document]
alonre24: The logic above is a bit different, running `rstest` only and not all `test_*` files

GuyAv46 (approved):

GuyAv46 (approved):

PR #4937, opened 2024-08-19
MOD-7596: Add mirroring support

kei-nan:
We would like to conditionally mirror the repository to a different repository

1. Current state: no such support
2. Change: Add a CI step to conditionally mirror the branch
3. Outcome: Conditionally mirror certain branches

1. MOD-7596

1. CI Flow

**Mark if applicable**

### DIFF
--- a/.github/workflows/event-mirror.yml
+++ b/.github/workflows/event-mirror.yml
@@ -1,0 +1,21 @@
+name: Release a Version
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  mirror:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/event-mirror.yml
```diff
@@ -0,0 +1,21 @@
+name: Release a Version
```
raz-mon, line 1: Better name?

raz-mon (commented):
Nice
1. Consider adding an action that does this for all tags, not only releases (if we want to have full sync in the tags).
2. Consider adding a validation that this is a tag (i.e., name starts with a `v..`), so that branches don't get overriden (currently pushed with `-f` as well, so this can be aggressive)

raz-mon (approved):

PR #4950, opened 2024-08-21
Fix numeric union reopen on profile [MOD-7454]

GuyAv46:

Fixing a crash caused by a wrong cast of iterators when in PROFILE mode

1. `UI_Foreach`

**Mark if applicable**

### DIFF
--- a/tests/pytests/test_issues.py
+++ b/tests/pytests/test_issues.py
@@ -97,6 +97,31 @@ def test_1667(env):
   env.expect('ft.search idx a').equal([0])
   env.expect('ft.search idx b').equal([1, 'doc_b', ['text', 'b']])
 
+@skip(cluster=False)
+def test_MOD_7454(env: Env):
+  env.cmd('FT.CREATE', 'idx', 'SCHEMA', 'n', 'NUMERIC')
+  n_docs = 1100 * env.shardsCount # We need more than 1000 docs in each shard
+  with env.getClusterConnectionIfNeeded() as conn:
+    for i in range(n_docs):
+      conn.execute_command('HSET', f'doc{i}', 'n', i)
+
+  # We have more than 1000 docs in each shard, and the query should return all of them.
+  # In cluster mode, FT.AGGREGATE (and FT.PROFILE AGGREGATE) uses cursors in each shard, reading
+  # 1000 at a time and then aggregate the replies.
+  # The second batch read from each shard will require to "reopen" the numeric index (on each shard).
+  # We have a large numeric range so we expect to have a union iterator of multiple numeric iterators.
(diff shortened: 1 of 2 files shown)


raz-mon (commented):
Nice catch! 🔥 
Two comments/questions:
1. I see that the wrapping of the union iterators does not change `ui->num`, which may be confusing (related to this bug as well..) - this is a good opportunity to check out if we have any other places that we confuse these iterators somehow.
2. We have a similar mechanism working for the intersect iterator (w.r.t profiling) - can something like this happen there as well?

tests/pytests/test_issues.py
```diff
@@ -97,6 +97,31 @@ def test_1667(env):
   env.expect('ft.search idx a').equal([0])
   env.expect('ft.search idx b').equal([1, 'doc_b', ['text', 'b']])
 
+@skip(cluster=False)
+def test_MOD_7454(env: Env):
+  env.cmd('FT.CREATE', 'idx', 'SCHEMA', 'n', 'NUMERIC')
+  n_docs = 1100 * env.shardsCount # We need more than 1000 docs in each shard
```
raz-mon, line 103: Better to take this value from the config rather than hardcode the current value, such that if its default is ever changed this test will stay relevant

GuyAv46:
2. Other iterators that register an "onReopen" function (with `ConcurrentSearch_AddKey`) are passing the iterator directly, so when we add profile iterators in between the iterators it doesn't change the context that was passed to the concurrent handler.
Numeric unions are the only exception today, where the union iterator of the numeric ranges is passed as the context, and then each child is wrapped with a profile iterator and requires the additional dereference.
the `UI_Foreach` function assumes it gets a union of `IndexReader` iterators. This is also not correcting the general case but is correct for numeric ranges union (beside the profile case).

tests/pytests/test_issues.py
raz-mon: Wasn't there a config-param for it? 🤔

GuyAv46:
1. The `UI_Foreach` function is the only one we assume the children type is `IndexReader` necessarily. It is only used by the numeric union range reopen flow. Assuming the children should all be `IndexReader`, the only exception is when we profile the sub-tree. I can add an assertion on the type before calling the callback

tests/pytests/test_issues.py
GuyAv46: There is a parameter in the config structure, but no dedicated function to access it 😢

raz-mon: Nope, than nothing to do here

raz-mon (approved):
🔥

PR #4726, opened 2024-06-09
MOD-7543 purge readies from sanitizer flow

DvirDukhan:
**Describe the changes in the pull request**
This PR purges Readies from the sanitizer flow, allowing for an easier, less intrusive workflow by using the standard docker and dependency toolchain.
This PR unifies the flows for all test flows (test, coverage, sanitizer) to use `task-test.yml`.
This PR moves the llvm and rust installations to our common `.install` folder, where they should be, and removes them from `build-rejson` script 

**As a temporary workaround, this PR also disables sanitizer with JSON tests, until a fix is submitted to RedisJSON repo**

Also, we link statically to libstdc++ when we run sanitizer, since otherwise we crash in geometry tests otherwise. This is fine as long as we test sanitizer against redis 7.4. When we move to test against redis 8, we should link libstdc++ dynamically (as we do from now on). 

**Which issues this PR fixes**
MOD-7453

**Main objects this PR modified**
task-test.yml
build-rejson

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #4957, opened 2024-08-25
MOD-7465 MOD-7466: Add ACL support for RediSearch, phase 1

raz-mon:

This PR adds support for the Redis ACL feature in RediSearch, by creating a new `search` ACL command category, and registering the module commands to their appropriate command categories.

This PR also bumps the versions we use with Valgrind and Sanitizer to `7.4`.

**Mark if applicable**


raz-mon:
FYI the new tests are failing since they depend on the [fix](https://github.com/redis/redis/commit/3b1b1d14869c88d45f43fd4c333b9d1f54295fa9) we merged to Redis, which was not released yet (pass against `unstable`).

src/coord/module.c
```diff
@@ -2132,68 +2132,70 @@ RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT.ALIASDEL", SafeCmd(MastersFanoutCommandHandler), "readonly", 0, 0, -1));
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT._ALIASDELIFX", SafeCmd(MastersFanoutCommandHandler), "readonly", 0, 0, -1));
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT.ALIASUPDATE", SafeCmd(MastersFanoutCommandHandler), "readonly", 0, 0, -1));
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT.SYNUPDATE", SafeCmd(MastersFanoutCommandHandler),"readonly", 0, 0, -1));
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT.SYNFORCEUPDATE", SafeCmd(MastersFanoutCommandHandler),"readonly", 0, 0, -1));
+    RM_CREATE_COMMAND("search", ctx, "FT.CREATE", SafeCmd(MastersFanoutCommandHandler), "readonly", 0, 0, -1);
+    RM_CREATE_COMMAND("search", ctx, "FT._CREATEIFNX", SafeCmd(MastersFanoutCommandHandler), "readonly", 0, 0, -1);
+    RM_CREATE_COMMAND("search", ctx, "FT.ALTER", SafeCmd(MastersFanoutCommandHandler), "readonly", 0, 0, -1);
+    RM_CREATE_COMMAND("search", ctx, "FT._ALTERIFNX", SafeCmd(MastersFanoutCommandHandler), "readonly", 0, 0, -1);
+    RM_CREATE_COMMAND(NULL, ctx, "FT.DROPINDEX", SafeCmd(MastersFanoutCommandHandler), "readonly",0, 0, -1);
```
kei-nan, line 2172: why is this NULL?
same for next command

src/module.c
```diff
@@ -1008,135 +1008,160 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx, RedisModuleString **argv,
 #define INDEX_DOC_CMD_ARGS 2, 2, 1
 #endif
 
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_INDEX_LIST_CMD, IndexList, "readonly", 0, 0, 0);
+  // Create the `search` ACL command category
+  RedisModuleString* cat_name = RedisModule_CreateString(ctx, "search", strlen("search"));
```
kei-nan, line 1012: Maybe wrap this in a function that accepts a `const char*`?

src/coord/module.c
raz-mon: Will be removed in `v3.0` which is the first version this will be released in

src/module.h
```diff
@@ -43,6 +43,34 @@ do {                                            \
...
+    if (aclCategories != NULL) {                                               \
+      result = RedisModule_SetCommandACLCategories(command, aclCategories);    \
+      if (result == REDISMODULE_ERR) {                                         \
+        RedisModule_Log(ctx, "warning",                                        \
+          "Failed to set ACL categories for command " STRINGIFY(SECOND_ARG(__VA_ARGS__)) ". Got error code: %d", errno); \
+        return REDISMODULE_ERR;                                                \
+      }                                                                        \
+    }                                                                          \
+  }
+    // if (aclCategories != NULL &&                                               \
```
kei-nan, line 71: comment part can be removed.

src/module.c
raz-mon: What do you mean by "this"? The registration of the new category?

src/module.h
```diff
@@ -43,6 +43,34 @@ do {                                            \
 #define RedisModule_ReplyWithLiteral(ctx, literal) \
   RedisModule_ReplyWithStringBuffer(ctx, literal, sizeof(literal) - 1)
 
+// A macro that extracts the first argument from a variadic macro
+#define SECOND_ARG_HELPER(first, second, ...) second
+#define SECOND_ARG(...) SECOND_ARG_HELPER(__VA_ARGS__)
+
+#define RM_CREATE_COMMAND(aclCategories, ...)                                  \
```
kei-nan, line 50: instead of a macro can this be a function?
Then the acl param won't have to be first.

src/module.c
```diff
@@ -1008,135 +1008,160 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx, RedisModuleString **argv,
...
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_SYNDUMP_CMD, SynDumpCommand, "readonly",
+  RM_CREATE_COMMAND("search", ctx, RS_SYNDUMP_CMD, SynDumpCommand, "readonly",
          INDEX_ONLY_CMD_ARGS);
 
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_ALTER_CMD, AlterIndexCommand, "write",
+  RM_CREATE_COMMAND("search", ctx, RS_ALTER_CMD, AlterIndexCommand, "write",
          INDEX_ONLY_CMD_ARGS);
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_ALTER_IF_NX_CMD, AlterIndexIfNXCommand, "write",
+
+  // TODO: Verify
```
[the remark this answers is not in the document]
raz-mon: No, tnx

src/module.h
raz-mon: Thought I removed it, tnx

src/module.h
raz-mon: Why does it bother that it's first?
It obviously can be a function, but I don't see a huge benefit here

src/coord/module.c
```diff
@@ -2132,68 +2132,70 @@ RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT.AGGREGATE", SafeCmd(DistAggregateCommand), "readonly", 0, 1, -2));
+    RM_CREATE_COMMAND("read search", ctx, "FT.AGGREGATE", SafeCmd(DistAggregateCommand), "readonly", 0, 1, -2);
   } else {
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT.AGGREGATE", SafeCmd(DistAggregateCommand), "readonly", 0, 0, -1));
+    RM_CREATE_COMMAND("read search", ctx, "FT.AGGREGATE", SafeCmd(DistAggregateCommand), "readonly", 0, 0, -1);
   }
-  RM_TRY(RedisModule_CreateCommand(ctx, "FT.INFO", SafeCmd(InfoCommandHandler), "readonly", 0, 0, -1));
-  RM_TRY(RedisModule_CreateCommand(ctx, "FT.SEARCH", SafeCmd(DistSearchCommand), "readonly", 0, 0, -1));
-  RM_TRY(RedisModule_CreateCommand(ctx, "FT.PROFILE", SafeCmd(ProfileCommandHandler), "readonly", 0, 0, -1));
+  RM_CREATE_COMMAND("search", ctx, "FT.INFO", SafeCmd(InfoCommandHandler), "readonly", 0, 0, -1);
```
kei-nan, line 2143: Maybe create constants for the strings so it will be easier to maintain.

raz-mon: Yeah.. They're constants for the SA.. Since we use it only here and we have the other constants I'll leave it for now

tests/pytests/test_acl.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # Check that one of our commands is listed in a non-search category
+    res = env.cmd('ACL', 'CAT', 'read')
+    env.assertTrue('FT.SEARCH' in res)
+
+def test_acl_non_default_user(env):
+    """Tests that a user with a non-default ACL can't access the search
+    category"""
+    # Create a user with no command permissions
+    env.expect('ACL', 'SETUSER', 'test', 'on', '>123', '~*', '&*').ok()
```
kei-nan, line 43: Consider changing "test" -> "test-user"
Also if the `>123`, `-*` and `&*` have meaning then move them to be variables so they will have some context and it will be easier to maintain.

tests/pytests/test_acl.py
```diff
@@ -0,0 +1,77 @@
...
+        '_FT.CURSOR', 'FT.INFO', 'FT.SUGDEL', '_FT.INFO', '_FT.SUGDEL',
+        '_FT.ALTER', '_FT.DICTDEL', '_FT.SYNUPDATE', 'FT.DICTDUMP',
+        'FT.EXPLAIN', '_FT.SPELLCHECK', 'FT.AGGREGATE', 'FT.SUGLEN',
+        '_FT.SUGLEN', 'FT.PROFILE', 'FT.ALTER', 'FT.SUGGET', '_FT.CREATE',
+        'FT.DICTDEL', 'FT.CURSOR', 'FT.ALIASDEL', 'FT.SUGADD', '_FT.DICTADD',
+        'FT.SYNDUMP', 'FT.CREATE', '_FT.PROFILE', '_FT.SEARCH', 'FT.DICTADD',
+        'FT.SYNFORCEUPDATE', 'FT._ALIASDELIFX', 'FT._CREATEIFNX',
+        '_FT.DEBUG', '_FT.CONFIG', '_FT.TAGVALS', '_FT.SUGGET',
+        'search.CLUSTERREFRESH', 'FT._ALIASADDIFNX', '_FT._ALTERIFNX',
+        '_FT.SUGADD', 'FT._ALTERIFNX', '_FT._ALIASDELIFX', 'search.CLUSTERSET',
```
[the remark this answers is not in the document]
raz-mon: I don't know, but I didn't touch the ramp file in the PR..

tests/pytests/test_acl.py
raz-mon: `123` is the password, and the other indicators indicate the key, pubsub and command permissions for the new ACL user. Since these are dynamic and change from user to user (i.e., not reused mostly, and specifically in this test), I think they should stay as they are.

src/module.h
kei-nan: Mainly maintainability.
It is much easier to maintain a function than a macro.
Also consider you stopped using RM_TRY which means we won't get verbose prints if the function succeeds.

For example if someone introduces a code which returns early from the macro then it would prevent all the following functions from being called.

tests/pytests/test_acl.py
kei-nan: If you don't want to store the values in variables them please add a comment explaining them.
Especially the pubsub and command permissions.

src/coord/module.c
kei-nan: Maybe we want to explicitly differentiate the two cases?
e.g make one overload of registration for acl commands and another for deprecated commands?
That way we can be sure no one will register a command that needs to be with an acl and will just pass NULL.

tests/pytests/test_acl.py
raz-mon: There is already an explanation above. That's the only one relevant for the test.

raz-mon: I can add more if it's very important to you, but I think this is enough.

src/module.h
raz-mon: This would be a big change that I don't think is necessary. I went with the currently implemented macro approach that overall makes sense here and is comfortable. If someone would need to return early, he would need it to do it appropriately, sure.
Regarding the log, you practically don't see it today anyways, and there is no need for this positive log IMO.

raz-mon: If there was a failure - you know about it since the server fails. Otherwise - all is good. This is the way we log today and are supposed to.
Why don't we also report more successes in our code-base? It garbages the log for no reason

src/module.c
raz-mon: Why?
We do this only once, here, and under no circumstances will this code be called elsewhere - that is not aloud.
If it's to save 4 lines of code from this function I prefer not to.

tests/pytests/test_acl.py
raz-mon: Added

src/module.h
raz-mon: Will consider this since it's more comfortable at the end of the day..

tests/cpptests/redismock/redismock.cpp
```diff
@@ -772,6 +772,17 @@ static int RMCK_KillForkChild(int child_pid) {
   return waitpid(child_pid, NULL, 0);
 }
 
+static int RMCK_AddACLCategory(RedisModuleCtx *ctx, const char *category) {
+  // Nothing for the mock.
+  return REDISMODULE_OK;
+}
+
+static int RMCK_SetCommandACLCategories(RedisModuleCommand *cmd,
+                                                    const char *categories) {
```
kei-nan, line 781: if you get a chance, put the second argument in the same line as the first.

src/coord/module.c
```diff
@@ -2130,66 +2123,66 @@ RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
   }
 
   if (RSBuildType_g == RSBuildType_OSS) {
     RedisModule_Log(ctx, "notice", "Register write commands");
     // suggestion commands
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT.SUGADD", SafeCmd(SingleShardCommandHandler), "readonly", 0, 0, -1));
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT.SUGGET", SafeCmd(SingleShardCommandHandler), "readonly", 0, 0, -1));
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT.SUGDEL", SafeCmd(SingleShardCommandHandler), "readonly", 0, 0, -1));
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT.SUGLEN", SafeCmd(SingleShardCommandHandler), "readonly", 0, 0, -1));
+    RM_TRY(RMCreateCommand(ctx, "FT.SUGADD", SafeCmd(SingleShardCommandHandler), "readonly", 0, 0, -1, "write search"))
```
oshadmi, line 2150: `readonly` flag and `write` permission are contradicting - should remove `readonly`?

src/coord/module.c
```diff
@@ -2130,66 +2123,66 @@ RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+    RM_TRY(RMCreateCommand(ctx, "FT.SUGGET", SafeCmd(SingleShardCommandHandler), "readonly", 0, 0, -1, "read search"))
+    RM_TRY(RMCreateCommand(ctx, "FT.SUGDEL", SafeCmd(SingleShardCommandHandler), "readonly", 0, 0, -1, "write search"))
```
oshadmi, line 2152: `readonly` flag and `write` permission are contradicting - should remove `readonly`?

src/coord/module.c
raz-mon: Could be breaking. We should check this separately

src/coord/module.c
raz-mon: Could be breaking. We should check this separately

src/module.c
```diff
@@ -1008,136 +1036,150 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx, RedisModuleString **argv,
 #define INDEX_DOC_CMD_ARGS 2, 2, 1
 #endif
 
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_INDEX_LIST_CMD, IndexList, "readonly", 0, 0, 0);
+  // Create the `search` ACL command category
+  if (RedisModule_AddACLCategory(ctx, "search") == REDISMODULE_ERR) {
```
oshadmi, line 1040: Consider using a variable/macro

src/module.c
```diff
@@ -1008,136 +1036,150 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx, RedisModuleString **argv,
...
-         0);
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_ALIASUPDATE, AliasUpdateCommand, "readonly", 0, 0, 0);
-
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_ALIASDEL, AliasDelCommand, "readonly", 0, 0, 0);
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_ALIASDEL_IF_EX, AliasDelIfExCommand, "readonly", 0, 0,
-         0);
+  RM_TRY(RMCreateCommand(ctx, RS_ALIASADD, AliasAddCommand, "readonly", 0, 0, 0, "search"))
+  RM_TRY(RMCreateCommand(ctx, RS_ALIASADD_IF_NX, AliasAddCommandIfNX, "readonly", 0, 0,
+         0, "search"))
+  RM_TRY(RMCreateCommand(ctx, RS_ALIASUPDATE, AliasUpdateCommand, "readonly", 0, 0, 0, "write search"))
```
oshadmi, line 1178: `write` is not needed

src/module.c
```diff
@@ -1008,136 +1036,150 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx, RedisModuleString **argv,
 #define INDEX_DOC_CMD_ARGS 2, 2, 1
 #endif
 
-  RM_TRY(RedisModule_CreateCommand, ctx, RS_INDEX_LIST_CMD, IndexList, "readonly", 0, 0, 0);
+  // Create the `search` ACL command category
+  if (RedisModule_AddACLCategory(ctx, "search") == REDISMODULE_ERR) {
+      RedisModule_Log(ctx, "warning", "Could not add `search` ACL category, errno: %d\n", errno);
```
oshadmi, line 1041: Should not be a warning if this will prevent the module from being loaded and can also prevent a server from starting,

raz-mon: Why not?

oshadmi: I thought it should be an error but I see the vast majority of what we currently use is "warning"

raz-mon:
Sanitizer is failing since currently (due to `readies` issues) we run the sanitizer using Redis `7.2.0` which does not have the ACL for modules functionality this PR uses.
This is a problem, since we already oblige to use `7.4` as the minimum redis version and should be fully compatible with `7.4`.
This is handled in #4726 - once it will be merged sanitizer will pass.

FYI @DvirDukhan @oshadmi

src/coord/module.c
```diff
@@ -2130,66 +2123,68 @@ RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT.SYNUPDATE", SafeCmd(MastersFanoutCommandHandler),"readonly", 0, 0, -1));
-    RM_TRY(RedisModule_CreateCommand(ctx, "FT.SYNFORCEUPDATE", SafeCmd(MastersFanoutCommandHandler),"readonly", 0, 0, -1));
+    RM_TRY(RMCreateCommand(ctx, "FT.CREATE", SafeCmd(MastersFanoutCommandHandler), "readonly", 0, 0, -1, SEARCH_ACL_CATEGORY))
+    RM_TRY(RMCreateCommand(ctx, "FT._CREATEIFNX", SafeCmd(MastersFanoutCommandHandler), "readonly", 0, 0, -1, SEARCH_ACL_CATEGORY))
+    RM_TRY(RMCreateCommand(ctx, "FT.ALTER", SafeCmd(MastersFanoutCommandHandler), "readonly", 0, 0, -1, SEARCH_ACL_CATEGORY))
+    RM_TRY(RMCreateCommand(ctx, "FT._ALTERIFNX", SafeCmd(MastersFanoutCommandHandler), "readonly", 0, 0, -1, SEARCH_ACL_CATEGORY))
+    //  TODO: Fix both below! Not deprecated
+    RM_TRY(RMCreateCommand(ctx, "FT.DROPINDEX", SafeCmd(MastersFanoutCommandHandler), "readonly",0, 0, -1, "write slow dangerous " SEARCH_ACL_CATEGORY))
+    RM_TRY(RMCreateCommand(ctx, "FT._DROPINDEXIFX", SafeCmd(MastersFanoutCommandHandler), "readonly",0, 0, -1, "write slow dangerous " SEARCH_ACL_CATEGORY))
```
kei-nan, line 2161: On the one hand this is marked as "readonly" and on the other we ask for "write" acl.
Seems a bit contradicting

kei-nan (approved):

src/module.c
raz-mon: Yes, anyways the server will fail to load.

tests/pytests/test_acl.py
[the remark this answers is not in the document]
oshadmi: Should be done in another PR.

tests/pytests/test_acl.py
```diff
@@ -0,0 +1,83 @@
...
+    load"""
+    res = env.cmd('ACL', 'CAT')
+    env.assertTrue('search' in res)
+
+@skip(redis_less_than="7.4.1")
+def test_acl_search_commands(env):
+    """Tests that the RediSearch commands are registered to the `search`
+    ACL category"""
+    res = env.cmd('ACL', 'CAT', 'search')
+    # TODO: Do we want to hide the internal commands (e.g., "_FT.*")?
```
oshadmi, line 14: Remove this comment? (different PR)

tests/pytests/test_acl.py
```diff
@@ -0,0 +1,82 @@
...
+        "User test has no permissions to run the 'FT.CREATE' command")
+    env.expect('FT.SEARCH', 'idx', 'hello').error().contains(
+        "no such index")
+
+    # Add `test` `search` permissions
+    env.expect('AUTH', 'default', '').true()
+    env.expect('ACL', 'SETUSER', 'test', '+@search').ok()
+    env.expect('AUTH', 'test', '123').true()
+
+    # `test` should now be able to run `write` commands like `FT.CREATE`
```
oshadmi, line 72: `FT.CREATE` is not in the `write` category (can run it because of `search` which is tested below)
Can try `FT.SUGADD` which has `write`?

tests/pytests/test_acl.py
```diff
@@ -0,0 +1,82 @@
...
+        "User test has no permissions to run the 'FT.CREATE' command")
+
+    env.expect('AUTH', 'default', '').true()
+    # Add `test` read permissions
+    env.expect('ACL', 'SETUSER', 'test', '+@read').ok()
+    env.expect('AUTH', 'test', '123').true()
+
+    # `test` should now be able to run `read` commands like `FT.SEARCH', but not
+    # `search` commands like `FT.CREATE`
+    env.expect('FT.CREATE', 'idx', 'SCHEMA', 'txt', 'TEXT').error().contains(
```
oshadmi, line 62: Better to test it on all new `read` commands

tests/pytests/test_acl.py
oshadmi: And other newly added `write` commands

tests/pytests/test_acl.py
raz-mon: Why all commands?

oshadmi: I mean something like putting the commands in some list and do a for-each (for the positive check that command can run and for the negative check that command cannot run)

oshadmi (approved):

nafraf (approved):

PR #4964, opened 2024-08-28
[WIP] HNSW vector index - handle inplace deletion after async [MOD-7643]

alonre24:

1. ...
2. ...

**Mark if applicable**

### DIFF
--- a/deps/thpool/thpool.c
+++ b/deps/thpool/thpool.c
@@ -664,8 +664,6 @@ static void *thread_do(redisearch_thpool_t *thpool_p) {
   threadCtx thread_ctx = {.thread_state = THREAD_RUNNING};
 
   while (true) {
-    LOG_IF_EXISTS("debug", "Thread %s is running iteration", thread_name)
-
     /** Read job from queue and execute it.
      * @note At this point the thread state can be either RUNNING or TERMINATE_WHEN_EMPTY which
      * are the only valid indices of pull_and_execute_ht. */
--- a/tests/pytests/test_vecsim.py
+++ b/tests/pytests/test_vecsim.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import random
+import time
 
 from RLTest import Env
 from common import *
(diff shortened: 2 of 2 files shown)


tests/pytests/test_vecsim.py
```diff
@@ -2380,3 +2381,44 @@ def get_debug_info():
...
+    conn.execute_command(debug_cmd(), 'WORKERS', 'DRAIN')
+
+    bg_indexing = 0
+    for index_i in range(n_indexes):
+        index_prefix = f'idx_{index_i}z'
+        bg_indexing += index_info(env, f'index:{index_prefix}')['indexing']
+        vector_index_info = get_vecsim_debug_dict(env, f'index:{index_prefix}', 'v')
+        env.assertEqual(to_dict(vector_index_info['BACKEND_INDEX'])['INDEX_LABEL_COUNT'], n_vectors // 2,
+                        message=(index_prefix, vector_index_info))
+    print(f'{bg_indexing}/{n_indexes} indexes were still reindexing after switching back to in-place mode')
```
GuyAv46, line 2424: Leftover?

deps/thpool/thpool.c
```diff
@@ -664,8 +664,6 @@ static void *thread_do(redisearch_thpool_t *thpool_p) {
   threadCtx thread_ctx = {.thread_state = THREAD_RUNNING};
 
   while (true) {
-    LOG_IF_EXISTS("debug", "Thread %s is running iteration", thread_name)
```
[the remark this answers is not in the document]
alonre24: It was littering the log, and I'm not sure if its really needed

tests/pytests/test_vecsim.py
alonre24: Actually no, I did it to avoid flakiness but still indicate if there are no bg indexing. Maybe I can use print only if bg_indexing = 0

deps/thpool/thpool.c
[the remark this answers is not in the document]
GuyAv46: Well it’s only on debug logging level, but OK

tests/pytests/test_vecsim.py
GuyAv46: I think this is better.
I would suggest:

1. Use `env.debugPrint`
2. Check if we are in GitHub CI, and if so, add the prefix to yield a warning annotation

I think we should also have a way to print warnings with RLTest

PR #4975, opened 2024-08-29
MOD-7695: Decide Ubuntu Runner Using Variable

kei-nan:
Decide which ubuntu runner to use based on a github variable

1. We always use `ubuntu-latest`
2. Use the runner from the `RUNS_ON_UBUNTU_LATEST` variable
3. Allow to choose which ubuntu latest runner without modifying our code

2. MOD-7695

1. Workflow Files

**Mark if applicable**

### DIFF
--- a/.github/workflows/benchmark-flow.yml
+++ b/.github/workflows/benchmark-flow.yml
@@ -37,7 +37,7 @@ env:
 
 jobs:
   benchmark-steps:
-    runs-on: ubuntu-latest
+    runs-on: ${{ vars.RUNS_ON_UBUNTU_LATEST }}
     container: ${{ inputs.container }}
     steps:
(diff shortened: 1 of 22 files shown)


.github/workflows/benchmark-flow.yml
```diff
@@ -37,7 +37,7 @@ env:
 
 jobs:
   benchmark-steps:
-    runs-on: ubuntu-latest
+    runs-on: ${{ vars.RUNS_ON_UBUNTU_LATEST }}
```
raz-mon, line 40: This sounds like a boolean variable, yet I assume this variable will get the Ubuntu version we will run on, correct?
If so, rename to something more explicit like `UBUNTU_VERSION`

PR #4985, opened 2024-09-04
Fail to load a schema rule from RDB without asserting - [MOD-7587]

GuyAv46:

In case we fail to load a schema rule (for example, when a JSON index is loaded from the RDB but the JSON module is not available), return an error instead of asserting

**Mark if applicable**

### DIFF
--- a/tests/pytests/test_json.py
+++ b/tests/pytests/test_json.py
@@ -612,12 +612,12 @@ def testDemo(env):
 @skip(cluster=True, no_json=True)
 def test_JSON_RDB_load_fail_without_JSON_module(env: Env):
     env.expect('FT.CREATE', 'idx', 'ON', 'JSON', 'SCHEMA', '$.t', 'TEXT').ok()
-    env.stop()
+    env.stop() # Save state to RDB
     env.assertEqual(len(env.envRunner.modulePath), len(env.envRunner.moduleArgs))
     env.envRunner.modulePath.pop() # Assumes Search module is the first and JSON module is the second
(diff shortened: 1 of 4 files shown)


alonre24 (approved):

tests/pytests/test_json.py
```diff
@@ -609,6 +609,17 @@ def testDemo(env):
     res =env.cmd('FT.AGGREGATE', 'airports', 'sfo', 'SORTBY', '1', '@iata', 'LOAD', '1', '$')
     env.assertEqual(toSortedFlatList(res), toSortedFlatList(expected_res))
 
+@skip(cluster=True, no_json=True, asan=True)
+def test_JSON_RDB_load_fail_without_JSON_module(env: Env):
+    env.expect('FT.CREATE', 'idx', 'ON', 'JSON', 'SCHEMA', '$.t', 'TEXT').ok()
```
oshadmi, line 614: How can you create the index `ON JSON` when `no_json=True`
(it should fail with `Invalid rule type: JSON`)

GuyAv46: I skip the test if we don't have JSON at all. I first create an index with the module, and then reload the RDB after relaunching redis without the module. The test has to be skipped if the module is not available for the first step

oshadmi: Consider to also create an index `on HASH`, in addition to the index `ON JSON`, so it would be successfully loaded before, and then see we clean it up as well.

PR #5071, opened 2024-10-01
Fix flaky test_async_updates_sanity - avoid timeout in force invoke GC

alonre24:

When we run `forceInvokeGC`, there is a default timeout in the debug command that blocks the client, triggering forkGC in the background and then replying. However, we saw that this timeout frequently reached in sanitizer tests, in which case we kill the child fork GC process upon shutting down the environment and the thread running the GC is stuck waiting for the child process pipe.
Hence, we move the force GC in the test to run in the background while we update vectors and wait for the fork GC without timeout when finishing the test.

**Mark if applicable**


tests/pytests/test_multithread.py
```diff
@@ -308,8 +308,9 @@ def test_async_updates_sanity():
                       query_before_update.tobytes())
         env.assertGreater(float(res[2][1]), float(0))
 
-        # Invoke GC, so we clean zombies for which all their repair jobs are done.
-        forceInvokeGC(env)
+        # Invoke GC, so we clean zombies for which all their repair jobs are done. We run in background
+        # so in case child process is not receiving cpu time, we do not hang the gc thread in the parent process.
+        forceInvokeGC(env, idx='idx', timeout=None, run_in_background=True)
```
GuyAv46, line 313: No need to specify idx and timeout in this case

tests/pytests/test_multithread.py
```diff
@@ -323,7 +324,7 @@ def test_async_updates_sanity():
     env.assertEqual(run_command_on_all_shards(env, *[debug_cmd(), 'WORKERS', 'RESUME']), ['OK']*n_shards)
     env.assertEqual(run_command_on_all_shards(env, *[debug_cmd(), 'WORKERS', 'DRAIN']), ['OK']*n_shards)
 
-    forceInvokeGC(env)
+    forceInvokeGC(env, idx='idx', timeout=0)
```
GuyAv46, line 327: no need to specify idx here

tests/pytests/common.py
```diff
@@ -320,14 +320,17 @@ def get_vecsim_debug_dict(env, index_name, vector_field):
     return to_dict(env.cmd(debug_cmd(), "VECSIM_INFO", index_name, vector_field))
 
 
-def forceInvokeGC(env, idx = 'idx', timeout = None):
+def forceInvokeGC(env, idx='idx', timeout=None, run_in_background=False):
```
meiravgri, line 323: Does the scenario you describe, where the child process is starved, might occur in all the tests, particularly when we run in sanitizer mode?
If so consider setting `run_in_background`default value to `True`

tests/pytests/test_multithread.py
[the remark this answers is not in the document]
alonre24: It doens't, but I sugest to see what will happen with timeout of 5 minutes and see if we must do additional change

tests/pytests/common.py
alonre24: But other tests do rely on running the GC *not* in the background

GuyAv46 (dismissed):

tests/pytests/common.py
```diff
@@ -320,14 +320,17 @@ def get_vecsim_debug_dict(env, index_name, vector_field):
...
-            env.debugPrint("forceInvokeGC: note timeout is infinite, consider using a big timeout instead.", force=True)
+    if run_in_background:
+        env.cmd(debug_cmd(), 'GC_FORCEBGINVOKE', idx)
+    elif timeout is not None:
+        # Note: timeout==0 means infinite (no timeout)
         env.cmd(debug_cmd(), 'GC_FORCEINVOKE', idx, timeout)
     else:
         env.cmd(debug_cmd(), 'GC_FORCEINVOKE', idx)
+
+
```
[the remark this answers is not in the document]
meiravgri: you marked this as resolved but didn't remove the redundant space :D

tests/pytests/common.py
```diff
@@ -320,14 +320,17 @@ def get_vecsim_debug_dict(env, index_name, vector_field):
...
-    if timeout is not None:
-        if timeout == 0:
-            env.debugPrint("forceInvokeGC: note timeout is infinite, consider using a big timeout instead.", force=True)
+    if run_in_background:
+        env.cmd(debug_cmd(), 'GC_FORCEBGINVOKE', idx)
+    elif timeout is not None:
+        # Note: timeout==0 means infinite (no timeout)
         env.cmd(debug_cmd(), 'GC_FORCEINVOKE', idx, timeout)
     else:
         env.cmd(debug_cmd(), 'GC_FORCEINVOKE', idx)
```
meiravgri, line 331: so essintialy we can't pass both `timeout` and `run_in_background=False`
I suggest splitting this into 2 functions for each use case.

meiravgri (approved):

PR #5122, opened 2024-10-19
MOD-7915: Implement aux_save2

nafraf:

Module implements `aux_save2` which checks whether there are any aux data to be saved, and only if there is, call `aux_save`

This PR requires https://github.com/RediSearch/RediSearch/pull/5231

1. MOD-7915

1. ...
2. ...

**Mark if applicable**


src/dictionary.c
```diff
@@ -198,6 +198,15 @@ static void SpellCheckDictAuxSave(RedisModuleIO *rdb, int when) {
   dictReleaseIterator(iter);
 }
 
+static void SpellCheckDictAuxSave2(RedisModuleIO *rdb, int when) {
+  if (when == REDISMODULE_AUX_BEFORE_RDB) {
+    return;
+  }
+  if (dictSize(spellCheckDicts)) {
+    SpellCheckDictAuxSave(rdb, when);
+  }
```
oshadmi, line 207: @nafraf What about adding 2 tests with `FT.DICTADD` and `FT.DICTDEL`  similar to the tests added with `FT.CREATE` and `FT.DROPINDEX`?

nafraf: @oshadmi  Thanks. Tests added.
`SpellCheckDictAuxSave2()` was modified:  if at least one dictionary has terms, all the dictionaries are saved.

src/dictionary.c
```diff
@@ -198,6 +198,27 @@ static void SpellCheckDictAuxSave(RedisModuleIO *rdb, int when) {
   dictReleaseIterator(iter);
 }
 
+static void SpellCheckDictAuxSave2(RedisModuleIO *rdb, int when) {
+  if (when == REDISMODULE_AUX_BEFORE_RDB) {
+    return;
+  }
+  if (dictSize(spellCheckDicts)) {
```
oshadmi, line 205: We need to decide how to handle empty suggestion dictionary.

if we consider an empty suggestion dictionary as non-existing (there is no command to delete it, just to delete its entries) and we avoid serializing it to rdb (need to change `aux_save`) could it be considered as non-breaking?
And if only empty suggestion dictionaries exist, rdb will not have any related data, and the search module would not be needed to load such an rdb.

It seems the change in behavior is is that we will create the Redis key not when loading the rdb (since it will not be there), but only when calling `FT.SUGADD` on such empty dictionaries.

Before the change, we would create the empty Redis key already when loading from rdb.

If we keep writing empty suggestion dictionaries, we would still need the search module to load rdbs which has such dictionaries, which is not benefiting from the intention of `aux_save2`, so maybe an alternative could be to skip writing empty suggestion dictionaries if all of them are empty, and to write them if at least one of them is non-empty. Then the search module would be required to load an rdb only if is contains actual related data (meaning that suggestion dictionary names on their own are insignificant)

src/dictionary.c
```diff
@@ -191,6 +191,15 @@ static void SpellCheckDictAuxSave(RedisModuleIO *rdb, int when) {
   dictReleaseIterator(iter);
 }
 
+static void SpellCheckDictAuxSave2(RedisModuleIO *rdb, int when) {
+  if (when == REDISMODULE_AUX_BEFORE_RDB) {
```
oshadmi, line 195: Consider removing (checked in SpellCheckDictAuxSave)

nafraf: OK. Done.

DvirDukhan (commented):
Thanks
Let's add some tests with older RDB to see that the logic of save/load is valid

src/dictionary.c
```diff
@@ -53,14 +51,20 @@ int Dictionary_Del(RedisModuleCtx *ctx, const char *dictName, RedisModuleString
...
   return valuesDeleted;
 }
 
-int Dictionary_Dump(RedisModuleCtx *ctx, const char *dictName, char **err) {
+void Dictionary_Dump(RedisModuleCtx *ctx, const char *dictName) {
   Trie *t = SpellCheck_OpenDict(ctx, dictName, REDISMODULE_READ);
   if (t == NULL) {
-    *err = "could not open dict key";
-    return -1;
+    RedisModule_ReplyWithSetOrArray(ctx, 0);
```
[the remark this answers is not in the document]
nafraf: It is returning an empty array.
We did that to align the behavior of `FT.DICTDUMP unexistent_dict` to `HGETALL` behavior 
```sh
127.0.0.1:6379> HGETALL unexistent_key
(empty array)
```

src/dictionary.c
nafraf: Update, some changes related to spell check dictionaries:
* We are not loading empty dictionaries, we skip them.
* We are deleting the dictionary when it becomes empty.
* Due to the previous changes, there are no empty dictionaries when we are saving the rdb.

src/dictionary.c
```diff
@@ -191,13 +200,25 @@ static void SpellCheckDictAuxSave(RedisModuleIO *rdb, int when) {
   dictEntry *entry;
   while ((entry = dictNext(iter))) {
     const char *key = dictGetKey(entry);
-    RedisModule_SaveStringBuffer(rdb, key, strlen(key) + 1 /* we save the /0*/);
     Trie *val = dictGetVal(entry);
```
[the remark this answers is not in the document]
nafraf: Now we are skipping the empty dictionaries on loading.

src/suggest.c
```diff
@@ -177,6 +177,10 @@ int RSSuggestDelCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
   const char *str = RedisModule_StringPtrLen(argv[2], &len);
   RedisModule_ReplyWithLongLong(ctx, Trie_Delete(tree, str, len));
 
+  if (tree->size == 0) {
```
oshadmi, line 180: Consider moving to a different PR (changing the suggestion is more "breaking" since it is a redis key)

nafraf: OK. The changes to suggestion dictionaries are part of a differente PR: https://github.com/RediSearch/RediSearch/pull/5320

src/dictionary.c
```diff
@@ -38,13 +38,11 @@ int Dictionary_Add(RedisModuleCtx *ctx, const char *dictName, RedisModuleString
   return valuesAdded;
 }
 
-int Dictionary_Del(RedisModuleCtx *ctx, const char *dictName, RedisModuleString **values, int len,
-                   char **err) {
```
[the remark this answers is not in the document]
nafraf: No. We don't use it. It was removed in https://github.com/RediSearch/RediSearch/pull/5231
Now, in case we can't open the dictionary, we don't return an error, we return 0.

src/dictionary.c
```diff
@@ -106,13 +104,8 @@ int DictDelCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
   const char *dictName = RedisModule_StringPtrLen(argv[1], NULL);
 
-  char *error;
-  int retVal = Dictionary_Del(ctx, dictName, argv + 2, argc - 2, &error);
-  if (retVal < 0) {
-    RedisModule_ReplyWithError(ctx, error);
-  } else {
-    RedisModule_ReplyWithLongLong(ctx, retVal);
-  }
+  int retVal = Dictionary_Del(ctx, dictName, argv + 2, argc - 2);
```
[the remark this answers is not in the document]
nafraf: Yes. `Dictionary_Del()` does not return error, it was returning error in case it could not open the dict in WRITE mode, but now we are trying to open it in READ mode and returning 0 if it does not exist.

tests/pytests/test_aux_save2.py
```diff
@@ -0,0 +1,278 @@
...
+    # Save state to RDB
+    env.stop()
+    # Restart without modules
+    env.assertEqual(len(env.envRunner.modulePath), 2)
+    env.assertEqual(len(env.envRunner.moduleArgs), 2)
+    env.envRunner.modulePath.pop()
+    env.envRunner.moduleArgs.pop()
+    env.envRunner.modulePath.pop()
+    env.envRunner.moduleArgs.pop()
+    env.envRunner.masterCmdArgs = env.envRunner.createCmdArgs('master')
```
oshadmi, line 45: Consider moving to a helper function (reduce repetition)

tests/pytests/test_aux_save2.py
```diff
@@ -0,0 +1,278 @@
...
+    # Attempt to load RDB fails because the RDB contains an empty
```
oshadmi, line 208: If we change the suggestion dictionary to be deleted when its last entry is removed - this rdb could be reloaded

tests/pytests/test_aux_save2.py
nafraf: OK. Helper function created, but maybe we should add that functionality to RLTest because we are going to need it for CONFIG params tests also.

oshadmi (approved):

PR #5123, opened 2024-10-20
MOD-7872: Add Types To Hide User Input

kei-nan:
Add types that can be used to hide user input.
We want to avoid outputting user data outside the module.

1. Before: there are no types to hide user data
2. Change: Add types and functions to hide user data
3. After: Having types that can hide user data

1. MOD-7872

1. Obfuscation

**Mark if applicable**


src/obfuscation/hidden.c
```diff
@@ -0,0 +1,204 @@
...
+  } else {
+    UserString* d = (UserString*)*dst;
+    if (s->length > d->length) {
+      d->user = rm_realloc((void*)d->user, s->length);
+      d->length = s->length;
+    }
+    strncpy((void*)d->user, s->user, s->length);
+    if (d->length > s->length) {
+      memset((void*)d->user + s->length, 0, d->length - s->length);
+    }
```
nafraf, line 171: Shouldn't we to reallocate `d->user` and update `d->length` also in this case?

nafraf (commented):
1. In the PR description, I know that you are using the template, but maybe we should show that 1. is the current state, 2. is the change done, 3. is the outcome. WDYT?
2. This is part of a bigger feature, then tests will be added later, right?

src/obfuscation/hidden.h
```diff
@@ -0,0 +1,53 @@
...
+struct HiddenStringStruct;
+struct HiddenSizeStruct;
+struct HiddenNameStruct;
+typedef struct HiddenStringStruct HiddenString;
+typedef struct HiddenSizeStruct HiddenSize;
+typedef struct HiddenNameStruct HiddenName;
+
+// Hides the string and obfustaces it
+HiddenString *HideAndObfuscateString(const char *str, uint64_t length, bool takeOwnership);
+// Hides the size and obfustaces it
```
nafraf, line 19:
```suggestion
// Hides the size and obfuscates it
```

src/obfuscation/hidden.c
kei-nan: In the end did add an update for the length.

tests/cpptests/redismock/internal.h
```diff
@@ -236,11 +236,15 @@ struct KVDB {
...
-  void erase(const std::string &key) {
+  bool erase(const std::string &key) {
     auto e = db.find(key);
+    if (e == db.end()) {
+      return false;
+    }
     Value *v = e->second;
     db.erase(e);
     v->decref();
+   	return true;
```
nafraf, line 247: tab? In the local editor I see it aligned, but here it is misaligned.

nafraf (dismissed):

nafraf (approved):

src/obfuscation/hidden.c
```diff
@@ -0,0 +1,123 @@
+#include "hidden.h"
+#include "rmalloc.h"
+#include "util/minmax.h"
+#include "redis_index.h"
+#include "query_node.h"
+#include "reply_macros.h"
+
+typedef struct {
+  const char *user;
+  uint64_t length;
```
oshadmi, line 10: Why not use `size_t` instead of `uint64_t`?
(`size_t` is portable, sized accordingly, and is also used by other library functions)

src/obfuscation/hidden.c
```diff
@@ -0,0 +1,123 @@
...
+} UserString;
+
+HiddenName *NewHiddenName(const char* name, uint64_t length, bool takeOwnership) {
+  UserString* value = rm_malloc(sizeof(*value));
+  value->user = name;
+  if (takeOwnership) {
+    value->user = rm_strndup(name, length);
+  }
```
oshadmi, line 18: Consider avoiding the write-after-write? (use an else clause?)
Do we know which branch would be taken more often? (use `__builtin_expect`?)

src/obfuscation/hidden.c
```diff
@@ -0,0 +1,123 @@
...
+  rm_free(value);
+};
+
+static inline int Compare(const char *left, size_t left_length, const char *right, size_t right_length) {
+  int result = strncmp(left, right, MIN(left_length, right_length));
+  if (result != 0 || left_length == right_length) {
+    return result;
+  } else {
+    return left_length < right_length ? -1 : 1;
```
oshadmi, line 36: Consider subtracting

src/obfuscation/hidden.c
```diff
@@ -0,0 +1,123 @@
...
+  }
+}
+
+static inline int CaseSensitiveCompare(const char *left, size_t left_length, const char *right, size_t right_length) {
+  int result = strncasecmp(left, right, MIN(left_length, right_length));
+  if (result != 0 || left_length == right_length) {
+    return result;
+  } else {
+    return left_length < right_length ? -1 : 1;
```
oshadmi, line 45: Consider subtracting

src/obfuscation/hidden.c
kei-nan: Worried about casting size_t to a signed type.

nafraf (approved):

PR #5128, opened 2024-10-21
MOD-7979: Check RS_FIELDMASK_ALL Mask Edge Case

kei-nan:
Check against mask edge case of being RS_FIELDMASK_ALL.

1. When translating the field mask we don't check this edge case, can lead to all sort of problems.
2. Check this edge case, return every fieldIndex

1. MOD-7979

1. Mask Translation

**Mark if applicable**

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -1195,6 +1195,11 @@ static inline uint16_t TranslateMask(uint64_t maskPart, t_fieldIndex *translatio
 uint16_t IndexSpec_TranslateMaskToFieldIndices(const IndexSpec *sp, t_fieldMask mask, t_fieldIndex *out) {
   uint16_t count = 0;
   const uint8_t LOW_OFFSET = 0;
+  if (mask == RS_FIELDMASK_ALL) {
+    count = array_len(sp->fieldIdToIndex);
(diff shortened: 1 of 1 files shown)


src/spec.c
```diff
@@ -1195,6 +1195,11 @@ static inline uint16_t TranslateMask(uint64_t maskPart, t_fieldIndex *translatio
 uint16_t IndexSpec_TranslateMaskToFieldIndices(const IndexSpec *sp, t_fieldMask mask, t_fieldIndex *out) {
   uint16_t count = 0;
   const uint8_t LOW_OFFSET = 0;
+  if (mask == RS_FIELDMASK_ALL) {
```
nafraf, line 1198: Is it possible to add a test to cover this case?

kei-nan: We have a test that causes mask to be `RS_FIELDMASK_ALL`.
Problem is the side affects of the test didn't cause a crash so we didn't have a way to detect them.

Looking again at the code, maybe a better solution would be to check the size of the translation table and not exceed it regardless of the mask we get.

nafraf (approved):

PR #5129, opened 2024-10-24
MOD-7612: config params

nafraf:
Create CONFIG parameter using the [Module Configurations API](https://redis.io/docs/latest/develop/reference/modules/modules-api-ref/#module-configurations-api) for each of the existing module configuration parameters that today are managed using FT.CONFIG command.

1. The current state briefly
2. What is the change

1. MOD-7612

1. ...
2. ...

**Mark if applicable**


src/config.c
```diff
@@ -1147,3 +1340,269 @@ RSTimeoutPolicy TimeoutPolicy_Parse(const char *s, size_t n) {
...
+        set_numeric_config, NULL,
+        &(RSGlobalConfig.numericTreeMaxDepthRange)) == REDISMODULE_ERR) {
+    return REDISMODULE_ERR;
+  }
+
+  if (RedisModule_RegisterNumericConfig(
+      ctx, "search-bg-index-sleep-gap", DEFAULT_BG_INDEX_SLEEP_GAP,
+      REDISMODULE_CONFIG_IMMUTABLE | REDISMODULE_CONFIG_UNPREFIXED, 1,
+      UINT32_MAX, get_bg_index_sleep_gap,
+      set_bg_index_sleep_gap, NULL, NULL) == REDISMODULE_ERR) {
```
oshadmi, line 1360: Can also use `RSGlobalConfig.numBGIndexingIterationsBeforeSleep` as privdata

src/config.c
```diff
@@ -20,28 +20,51 @@
...
     return NULL;
   }
 }
 
+// ext-load
+CONFIG_API_STRING_SETTER(set_ext_load);
+
+RedisModuleString* get_ext_load(const char *name, void *privdata) {
+  char *str = *(char **)privdata;
+  if (str) {
```
oshadmi, line 58: Can avoid this check? (`privdata` is initialized properly)

nafraf: No, we can't. `RSGlobalConfig.extLoad = NULL` by default.
CONFIG GET returns `""` when the variable is `NULL`

```sh
127.0.0.1:6379> ft.config get extload
1) 1) EXTLOAD
   2) (nil)
127.0.0.1:6379> config get search-ext-load
1) "search-ext-load"
2) ""
```

src/config.c
nafraf: OK. Done

.github/workflows/task-build-artifacts.yml
```diff
@@ -117,8 +101,25 @@ jobs:
                   ;;
               esac
 
-              # Perform checkout to the REF
-              perform_checkout_ref
+              # Configure the safe directory
```
oshadmi, line 104: Not part of this PR, right?

.github/workflows/event-merge-to-queue.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     uses: ./.github/workflows/task-get-latest-tag.yml
     with:
       repo: redis/redis
-      prefix: '7.4'
+      prefix: '8.0'
```
oshadmi, line 23: Needed for new API/flag, right?

src/config.h
```diff
@@ -10,13 +10,23 @@
...
 #include "reply.h"
+#include "util/config_macros.h"
 
 typedef enum {
   TimeoutPolicy_Return,       // Return what we have on timeout
   TimeoutPolicy_Fail,         // Just fail without returning anything
   TimeoutPolicy_Invalid       // Not a real value
 } RSTimeoutPolicy;
 
+static const int int_on_timeout_vals[2] = {
```
oshadmi, line 21: Consider rename?
Maybe `on_timeout_enums`?

src/config.h
```diff
@@ -10,13 +10,23 @@
...
+  TimeoutPolicy_Return,
+  TimeoutPolicy_Fail
+};
+static const char *on_timeout_vals[2] = {
```
oshadmi, line 25: Can be used in `TimeoutPolicy_ToString`

.github/workflows/task-build-artifacts.yml
nafraf: Workflows were not modified by this PR. You saw it during in the review because I merged master branch in this branch

.github/workflows/event-merge-to-queue.yml
nafraf: Yes. It is needed for the new flag. But this PR did not modified it, that change is part of #5318 which was merged to master branch.

src/config.c
```diff
@@ -20,28 +20,107 @@
...
+                           void *privdata, RedisModuleString **err) {
+  REDISMODULE_NOT_USED(name);
+  REDISMODULE_NOT_USED(err);
+  *(unsigned int *)privdata = (unsigned int) val;
+  return REDISMODULE_OK;
+}
+
+long long get_uint_numeric_config(const char *name, void *privdata) {
+  REDISMODULE_NOT_USED(name);
+  return (long long)(*(unsigned int *)privdata);
```
kei-nan, line 52: Why the double cast?
Can't you cast it to a long long pointer and dereference it?

src/config.c
```diff
@@ -20,28 +20,107 @@
...
+  REDISMODULE_NOT_USED(err);
+  *(int *)privdata = val;
+  return REDISMODULE_OK;
+}
+
+int set_inverted_bool_config(const char *name, int val, void *privdata,
+                             RedisModuleString **err) {
+  REDISMODULE_NOT_USED(name);
+  REDISMODULE_NOT_USED(err);
+  *(int *)privdata = (val == 0);
```
kei-nan, line 67: what happens if type bool is passed?
in some compilers the size of bool is 1 and it doesn't have to be the same size as int.

src/config.c
```diff
@@ -20,28 +20,107 @@
...
+
+int set_immutable_string_config(const char *name, RedisModuleString *val, void *privdata,
+                      RedisModuleString **err) {
+  REDISMODULE_NOT_USED(name);
+  REDISMODULE_NOT_USED(err);
+  char **ptr = (char **)privdata;
+  if (val) {
+    size_t len;
+    const char *ret = RedisModule_StringPtrLen(val, &len);
+    if (len > 0) {
```
kei-nan, line 89: Talked about this line with Omer, this flow prevents the user from setting an empty string as a value.
I think that should be allowed in general.

src/config.c
```diff
@@ -20,28 +20,107 @@
...
+
 // EXTLOAD
 CONFIG_SETTER(setExtLoad) {
   int acrc = AC_GetString(ac, &config->extLoad, NULL, 0);
   RETURN_STATUS(acrc);
 }
 
 CONFIG_GETTER(getExtLoad) {
-  if (config->extLoad) {
+  if (config->extLoad && strlen(config->extLoad) > 0) {
```
kei-nan, line 103: maybe you want to use sdslen instead?

src/config.c
```diff
@@ -229,6 +327,23 @@ CONFIG_GETTER(getMinOperationWorkers) {
   return sdscatprintf(ss, "%lu", config->minOperationWorkers);
 }
 
+// min-operation-workers
+int set_min_operation_workers(const char *name,
```
kei-nan, line 331: can't you use one of the generic type setters at the start of this file?
Is this custom function needed?
Doesn't seem like it does any validation.

src/config.c
```diff
@@ -344,7 +459,24 @@ CONFIG_SETTER(setFrisoINI) {
   RETURN_STATUS(acrc);
 }
 CONFIG_GETTER(getFrisoINI) {
-  return config->frisoIni ? sdsnew(config->frisoIni) : NULL;
+  if (config->frisoIni && strlen(config->frisoIni) > 0) {
```
kei-nan, line 462: Doesn't this change the behaviour?
We no longer return empty strings?
Are we sure that is what we want?

If this is supposed to be backported then maybe we shouldn't break existing behaviour in that regard.

src/config.c
```diff
@@ -682,14 +840,32 @@ int ReadConfig(RedisModuleString **argv, int argc, char **err) {
...
+      // If we don't have a coordinator or this configuration has no trigger, this value
+      // is meaningless and should be ignored
+      if (curVar->setValue(&RSGlobalConfig, &ac, curVar->triggerId, &status) != REDISMODULE_OK) {
+        *err = rm_strdup(QueryError_GetError(&status));
+        QueryError_ClearError(&status);
+        return REDISMODULE_ERR;
+      }
+    } else {
+      // Consume the value
+      RedisModule_Log(NULL, "notice", "Can't set %s, because it was already set by CONFIG", curVar->name);
```
kei-nan, line 862: Don't we want to return an error and fill err with this string instead?
This logic seems a bit off.

src/config.c
```diff
@@ -682,14 +840,32 @@ int ReadConfig(RedisModuleString **argv, int argc, char **err) {
...
+    if ((currValue != NULL && currValueDefault!= NULL
+          && sdscmp(currValue, currValueDefault) == 0 ) ||
+          (currValue == NULL && currValueDefault == NULL)) {
+      // `triggerId` is set by the coordinator when it registers a trigger for a configuration.
+      // If we don't have a coordinator or this configuration has no trigger, this value
+      // is meaningless and should be ignored
+      if (curVar->setValue(&RSGlobalConfig, &ac, curVar->triggerId, &status) != REDISMODULE_OK) {
+        *err = rm_strdup(QueryError_GetError(&status));
+        QueryError_ClearError(&status);
+        return REDISMODULE_ERR;
```
kei-nan, line 858: shouldn't you also free the sds strings here if you are returning early?

src/config.c
```diff
@@ -344,7 +459,24 @@ CONFIG_SETTER(setFrisoINI) {
...
+  } else {
+    return NULL;
+  }
+}
+
+// friso-ini
+RedisModuleString * get_friso_ini(const char *name, void *privdata) {
+  char *str = *(char **)privdata;
+  if (str) {
+    if (config_friso_ini) {
```
kei-nan, line 473: Why do you need `config_friso_ini` as a global variable?
Can't you just return a redis module string and have redis free it?

src/config.c
```diff
@@ -366,6 +498,20 @@ CONFIG_GETTER(getOnTimeout) {
   return sdsnew(TimeoutPolicy_ToString(config->requestConfigParams.timeoutPolicy));
 }
 
+// on-timeout
+int set_on_timeout(const char *name, int val, void *privdata,
```
kei-nan, line 502: maybe have a generic set for the specific type instead of having it specific to `on_timeout`?

src/config.c
```diff
@@ -682,14 +840,32 @@ int ReadConfig(RedisModuleString **argv, int argc, char **err) {
...
-      *err = rm_strdup(QueryError_GetError(&status));
-      QueryError_ClearError(&status);
-      return REDISMODULE_ERR;
+    // if the value is different from default, we can't update it, because the
+    // CONFIG SET has higher priority than the FT.CONFIG SET
+    RSConfig RSDefaultConfig = RS_DEFAULT_CONFIG;
+    sds currValue = curVar->getValue(&RSGlobalConfig);
+    sds currValueDefault = curVar->getValue(&RSDefaultConfig);
+
+    if ((currValue != NULL && currValueDefault!= NULL
```
kei-nan, line 849: can you store the condition in a boolean variable name called `equals`?
IMHO it would simplify the reading of the code.

kei-nan: Maybe even put the error case(inequality) at the start of the function to simplify the code.

src/config.c
```diff
@@ -1147,3 +1324,351 @@ RSTimeoutPolicy TimeoutPolicy_Parse(const char *s, size_t n) {
 void iteratorsConfig_init(IteratorsConfig *config) {
   *config = RSGlobalConfig.iteratorsConfigParams;
 }
+
+
+int RegisterModuleConfig(RedisModuleCtx *ctx) {
```
kei-nan, line 1329: Suggestion:
Have structs to represent config information for numeric, boolean, string and so on and have array for each type.
Then you can have a loop to register each config option instead of the blob of RM_TRY calls.

Its optional, you can resolve this thread if you feel it is too much work.

src/coord/config.c
```diff
@@ -12,6 +12,7 @@
 #include "rmutil/util.h"
 #include "rmutil/strings.h"
 #include "hiredis/hiredis.h"
+#include "../module.h"
```
kei-nan, line 15: This sort of include can be a bit problematic, are we sure it is needed?
If so then you can mark this as resolved.

src/module.c
```diff
@@ -3482,15 +3489,35 @@ RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
   RSConfigOptions_AddConfigs(&RSGlobalConfigOptions, GetClusterConfigOptions());
   ClusterConfig_RegisterTriggers();
 
+  // Register the module configuration parameters
+  GetRedisVersion(ctx);
+  Version unstableRedis = {8, 1, 0};
```
kei-nan, line 3494:
```suggestion
const  Version unstableRedis = {8, 1, 0};
```

src/module.c
```diff
@@ -3482,15 +3489,35 @@ RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+  bool unprefixedConfigSupported = (CompareVersions(redisVersion, unstableRedis) >= 0) ? true : false;
```
kei-nan, line 3495:
```suggestion
  const bool unprefixedConfigSupported = (CompareVersions(redisVersion, unstableRedis) >= 0) ? true : false;
```

src/module.c
```diff
@@ -3482,15 +3489,35 @@ RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+  if (unprefixedConfigSupported) {
+    if (RegisterModuleConfig(ctx) == REDISMODULE_ERR) {
+      RedisModule_Log(ctx, "warning", "Error registering module configuration");
+      return REDISMODULE_ERR;
+    }
+  }
+
+  // Check if we are actually in cluster mode
+  bool isClusterEnabled = checkClusterEnabled(ctx);
```
kei-nan, line 3504:
```suggestion
  const bool isClusterEnabled = checkClusterEnabled(ctx);
```

tests/pytests/test_config.py
```diff
@@ -416,3 +416,849 @@ def testSetACLUsername():
             env.assertTrue(False)
     except Exception as e:
         env.assertEqual(str(e), 'Timeout: operation timeout exceeded')
+
+################################################################################
+# Test CONFIG SET/GET numeric parameters
+################################################################################
+
+def _removeModuleArgs(env: Env):
```
kei-nan, line 424: Maybe add a comment for this function to explain why it is needed?

tests/pytests/test_config.py
```diff
@@ -416,3 +416,849 @@ def testSetACLUsername():
...
+                              'search-topology-validation-timeout']:
+                continue
+
+            f.write(f'{configName} {minValue}\n')
+
+    # Start the server using the conf file and check each value
+    env = Env(noDefaultModuleArgs=True, redisConfigFile=redisConfigFile)
+    for configName, argName, default, minValue, maxValue, immutable in numericConfigs:
+        # Skip cluster parameters
+        if configName in ['search-threads',
```
kei-nan, line 642: maybe add in the dict another boolean to say if it is a cluster variable?

src/config.c
oshadmi: `long long` and `unsigned int` may have different sizes, so dereferencing as `unsigned int` and then converting to long long seems safer.

src/config.c
oshadmi: Seems that the RHS is a Boolean and it can be safely converted to an int to match the LHS

src/config.c
oshadmi: Specifically with ACL password and ACL user, an empty string should be supported.
What about friso ini and extension load - seems they support an empty string with 2.10.5.
Let's keep it.

src/config.c
oshadmi: Isn't `extLoad` a plain char*, and not an sds string?

src/config.c
```diff
@@ -50,6 +129,7 @@ CONFIG_SETTER(setNoGc) {
 
 CONFIG_BOOLEAN_GETTER(getNoGc, gcConfigParams.enableGC, 1)
 
+// NO_MEM_POOLS
```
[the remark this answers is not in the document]
oshadmi: The name of the related config parameter

src/config.c
```diff
@@ -85,6 +165,7 @@ CONFIG_GETTER(getMinStemLen) {
   return sdscatprintf(ss, "%u", config->iteratorsConfigParams.minStemLength);
 }
 
+// FORKGC_SLEEP_BEFORE_EXIT
```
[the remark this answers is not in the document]
oshadmi: The name of the related config parameter (for which the following setter function is defined)

src/config.c
oshadmi: Needs to also call `workersThreadPool_SetNumWorkers`

src/config.c
oshadmi: Size of `*privdata` could be different (it is an enum config)

src/config.h
nafraf: Done.

src/config.h
nafraf: Done.

src/config.c
nafraf: We removed those validations. Now the FT.CONFIG param will have precedence over CONFIG params.

src/config.c
oshadmi: Nice suggestion. We'll try it on the next PR in this area.

src/config.c
```diff
@@ -26,6 +26,65 @@
 #define RS_MAX_CONFIG_TRIGGERS 1 // Increase this if you need more triggers
 RSConfigExternalTrigger RSGlobalConfigTriggers[RS_MAX_CONFIG_TRIGGERS];
 
+typedef struct {
+  const char *FTConfigName;
+  const char *ConfigName;
+} configPair_t;
+
+configPair_t __configPairs[] = {
```
oshadmi, line 34: On next PR in this area, we can wrap the registration functions (for Bool, Enum, Numeric and String) with a wrapper that will also receive the "old" name and add to an array of __configPairs - this way all would be defined in one place.

src/config.c
oshadmi: `FT.CONFIG GET` would keep returning NULL for empty value as before.

src/config.c
oshadmi: OK (then we can use the same getter for all string configuration parameters)

tests/pytests/test_config.py
```diff
@@ -677,29 +713,104 @@ def testConfigFileAndArgsNumericParams():
...
-                          'search-topology-validation-timeout']:
+        if clusterConfig:
             continue
 
         res = env.cmd('CONFIG', 'GET', configName)
         env.assertEqual(res, [configName, str(minValue)])
         res = env.cmd(config_cmd(), 'GET', argName)
         env.assertEqual(res, [[argName, str(minValue)]])
 
+@skip(cluster=True, redis_less_than='8.0')
```
oshadmi, line 737: We can use '7.99' to also work with RC1

oshadmi (dismissed):

oshadmi (approved):

oshadmi (approved):

PR #5137, opened 2024-10-30
MOD-8009: Allow Users To Configure Cursor Index Limitation Through Global Config

kei-nan:
* move cursor limitation per index to be a configuration parameter
* solve a bug where the coordinator would stop reading from internal cursors

1. The 128 cursor limitation is hard coded, coordinator relies on result length to determine if it is done
2. Move the limitation to the configuration, coordinator should only rely on the cursor id being non zero
3. User could update the cursor limitation, coordinator should continue reading from the internal cursors until the cursor is depleted.

1. MOD-8009

1. Cursors

**Mark if applicable**

- [X] This PR introduces serialization changes - changed the error message a bit


src/config.c
```diff
@@ -636,6 +636,23 @@ CONFIG_GETTER(getBGIndexSleepGap) {
 CONFIG_BOOLEAN_SETTER(set_PrioritizeIntersectUnionChildren, prioritizeIntersectUnionChildren)
 CONFIG_BOOLEAN_GETTER(get_PrioritizeIntersectUnionChildren, prioritizeIntersectUnionChildren, 0)
 
+CONFIG_SETTER(setIndexCursorLimit) {
```
GuyAv46, line 639: Add this new config to `test_configs`

src/config.c
```diff
@@ -840,6 +857,10 @@ RSConfigOptions RSGlobalConfigOptions = {
                      "high memory consumption.",
          .setValue = setCursorMaxIdle,
          .getValue = getCursorMaxIdle},
+        {.name = "INDEX_CURSOR_LIMIT",
+         .helpText = "max number of cursors that can be opened in the context of a single index",
```
GuyAv46, line 861: per shard, in the case of cluster mode, correct? consider mentioning that.

src/coord/dist_aggregate.c
```diff
@@ -123,7 +123,8 @@ static int netCursorCallback(MRIteratorCallbackCtx *ctx, MRReply *rep) {
 
   // rewrite and resend the cursor command if needed
   int rc = REDIS_OK;
-  bool done = !getCursorCommand(rep, cmd, MRIteratorCallback_GetCtx(ctx));
+  // done should only be determined based on the cursor and not on the set of results we get
+  const bool done = !getCursorCommand(rep, cmd, MRIteratorCallback_GetCtx(ctx));
```
GuyAv46, line 129: consider moving the call to this function to where we check `done`

src/cursor.c
```diff
@@ -206,16 +206,14 @@ Cursor *Cursors_Reserve(CursorList *cl, StrongRef global_spec_ref, unsigned inte
...
-
   // If we are in a coordinator ctx, the spec is NULL
-  if (spec && (spec->activeCursors >= spec->cursorsCap)) {
+  if (spec && spec->activeCursors >= RSGlobalConfig.indexCursorLimit) {
     /** Collect idle cursors now */
     Cursors_GCInternal(cl, 0);
-    if (spec->activeCursors >= spec->cursorsCap) {
-      QueryError_SetError(status, QUERY_ELIMIT, "Too many cursors allocated for index");
+    if (spec->activeCursors >= RSGlobalConfig.indexCursorLimit) {
+      QueryError_SetErrorFmt(status, QUERY_ELIMIT, "CURSOR_LIMIT was reached for index, limit of %lld reached", RSGlobalConfig.indexCursorLimit);
```
GuyAv46, line 214: consider improving the grammar of this error

kei-nan:
```suggestion
      QueryError_SetErrorFmt(status, QUERY_ELIMIT, "CURSOR_LIMIT of %lld was reached for an index", RSGlobalConfig.indexCursorLimit);
```
?

src/coord/dist_aggregate.c
kei-nan: getCursorCommand modifies cmd contents internally
So currently it has to happen before the if block.(since the if block accesses cmd)

src/config.c
kei-nan: Yep, it is just that up until now this was implicit in all of our configuration parameters.
e.g `CURSOR_MAX_IDLE` or any other shard configuration.
Don't think we should deviate from the convention.

src/coord/dist_aggregate.c
GuyAv46: But notice that none of the changes are relevant to what we checked there, And also if it returns false (`done` is true) we don't change it

src/util/misc.c
```diff
@@ -22,3 +23,9 @@ char *strtolower(char *str) {
   }
   return str;
 }
+
+int GetRedisErrorCodeLength(const char* error)
+{
```
GuyAv46, line 28:
```suggestion
int GetRedisErrorCodeLength(const char* error) {
```

src/util/misc.c
```diff
@@ -22,3 +23,9 @@ char *strtolower(char *str) {
...
+  const char* errorSpace = strchr(error, ' ');
+  return errorSpace ? errorSpace - error : 0;
+}
```
GuyAv46, line 31: missing new line

src/util/misc.h
```diff
@@ -16,4 +16,6 @@ void GenericAofRewrite_DisabledHandler(RedisModuleIO *aof, RedisModuleString *ke
 
 char *strtolower(char *str);
 
+int GetRedisErrorCodeLength(const char* error);
+
 #endif
```
GuyAv46, line 21: consider adding a new line

tests/ctests/test_error_parsing.c
```diff
@@ -0,0 +1,33 @@
...
+  testErrorCodeFormat("ERR", "");
+  testErrorCodeFormat(" ERR", "");
+  testErrorCodeFormat("", "");
+  return 0;
+}
+
+TEST_MAIN({
+  TESTFUNC(testErrorCodeLengthExtraction);
+  TESTFUNC(testErrorCodeFormatting);
+})
```
GuyAv46, line 33: missing new line

src/cursor.c
```diff
@@ -206,16 +206,14 @@ Cursor *Cursors_Reserve(CursorList *cl, StrongRef global_spec_ref, unsigned inte
...
-
   // If we are in a coordinator ctx, the spec is NULL
-  if (spec && (spec->activeCursors >= spec->cursorsCap)) {
+  if (spec && spec->activeCursors >= RSGlobalConfig.indexCursorLimit) {
     /** Collect idle cursors now */
     Cursors_GCInternal(cl, 0);
-    if (spec->activeCursors >= spec->cursorsCap) {
-      QueryError_SetError(status, QUERY_ELIMIT, "Too many cursors allocated for index");
+    if (spec->activeCursors >= RSGlobalConfig.indexCursorLimit) {
+      QueryError_SetErrorFmt(status, QUERY_ELIMIT, "CURSOR_LIMIT of %lld has been reached for an index", RSGlobalConfig.indexCursorLimit);
```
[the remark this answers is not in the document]
kei-nan: Redis takes the first word of the error message and shows it as an error code.
So in the error stats the user will see `INDEX_CURSOR_LIMIT ` as an error.
Wondering if that will be too confusing.
I can probably add it at the end of the message.
WDYT?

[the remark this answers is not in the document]
oshadmi: From user perspective, seeing an error stat with `INDEX_CURSOR_LIMIT` could be better than `CURSOR_LIMIT`

src/config.c
```diff
@@ -636,6 +636,19 @@ CONFIG_GETTER(getBGIndexSleepGap) {
 CONFIG_BOOLEAN_SETTER(set_PrioritizeIntersectUnionChildren, prioritizeIntersectUnionChildren)
 CONFIG_BOOLEAN_GETTER(get_PrioritizeIntersectUnionChildren, prioritizeIntersectUnionChildren, 0)
 
+CONFIG_SETTER(setIndexCursorLimit) {
+  long long limit;
+  int acrc = AC_GetLongLong(ac, &limit, AC_F_GE0);
+  CHECK_RETURN_PARSE_ERROR(acrc);
+  config->indexCursorLimit = limit;
+  RETURN_STATUS(acrc);
```
GuyAv46, line 644:
```suggestion
  int acrc = AC_GetLongLong(ac, &config->indexCursorLimit, AC_F_GE0);
  RETURN_STATUS(acrc);
```

src/config.c
```diff
@@ -636,6 +636,19 @@ CONFIG_GETTER(getBGIndexSleepGap) {
 CONFIG_BOOLEAN_SETTER(set_PrioritizeIntersectUnionChildren, prioritizeIntersectUnionChildren)
 CONFIG_BOOLEAN_GETTER(get_PrioritizeIntersectUnionChildren, prioritizeIntersectUnionChildren, 0)
 
+CONFIG_SETTER(setIndexCursorLimit) {
+  long long limit;
+  int acrc = AC_GetLongLong(ac, &limit, AC_F_GE0);
```
oshadmi, line 641: @kei-nan If not using `AC_F_GE1` does zero have any special meaning? Currently it seems not.
We don't intend to make it unlimited, right?

kei-nan: No intention to make it unlimited, 0 currently means we won'tt allow cursors to be opened.

src/coord/rmr/rmr.c
```diff
@@ -529,7 +529,11 @@ MRIteratorCtx *MRIteratorCallback_GetCtx(MRIteratorCallbackCtx *ctx) {
 }
 
 void MRIteratorCallback_AddReply(MRIteratorCallbackCtx *ctx, MRReply *rep) {
-  MRChannel_Push(ctx->ic->chan, rep);
+  if (MRChannel_Push(ctx->ic->chan, rep) & CHANNEL_CLOSED) {
+    // We are using a verbose log level to prevent flodding the logs with this message in production
+    RedisModule_Log(RSDummyContext, "verbose", "Tried pushing a reply to a closed channel");
+    MRReply_Free(rep);
+  }
```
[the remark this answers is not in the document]
kei-nan: Well, in broadcast failure the reply got successfully inserted into the channel so we shouldn't release it.

I can add some verbose logging to cover that case.

src/coord/rmr/chan.c
```diff
@@ -73,12 +73,12 @@ size_t MRChannel_Size(MRChannel *chan) {
   return ret;
 }
 
-int MRChannel_Push(MRChannel *chan, void *ptr) {
+PushErrorMask MRChannel_Push(MRChannel *chan, void *ptr) {
 
   pthread_mutex_lock(&chan->lock);
-  int rc = 1;
+  int rc = 0;
```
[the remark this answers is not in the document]
kei-nan: Yes, I think the commit I added this morning should fix that.

src/coord/dist_aggregate.c
```diff
@@ -137,11 +136,7 @@ static int netCursorCallback(MRIteratorCallbackCtx *ctx, MRReply *rep) {
         MRIteratorCallback_AddReply(ctx, rep); // to be picked up by getNextReply
         // User code now owns the reply, so we can't free it here ourselves!
         rep = NULL;
-      } else {
-        done = true;
       }
-    } else {
-      done = true;
```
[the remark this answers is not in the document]
kei-nan: Not sure I understand.
Do you mean add an assertion that results exist and is of type map?

GuyAv46 (approved):

src/coord/rmr/rmr.c
```diff
@@ -529,7 +529,15 @@ MRIteratorCtx *MRIteratorCallback_GetCtx(MRIteratorCallbackCtx *ctx) {
...
 void MRIteratorCallback_AddReply(MRIteratorCallbackCtx *ctx, MRReply *rep) {
-  MRChannel_Push(ctx->ic->chan, rep);
+  PushErrorMask mask = MRChannel_Push(ctx->ic->chan, rep);
+  if (mask & CHANNEL_CLOSED) {
+    // We are using a verbose log level to prevent flodding the logs with this message in production
+    RedisModule_Log(RSDummyContext, "verbose", "Tried pushing a reply to a closed channel");
+    MRReply_Free(rep);
+  }
+  if (mask & BROADCAST_FAILURE) {
+    RedisModule_Log(RSDummyContext, "warning", "Failed broadcasting the reply in the channel");
```
oshadmi, line 539: @kei-nan Since every `MRChannel_Push` may also fail with flag `BROADCAST_FAILURE`, `warning` level could bloat log. Why not use `verbose` level here as well?

kei-nan: That is what I tended to do at the start.

Why I chose to use a warning level is because it is more tied to an OS system call failure than a logic flow failure.

Broadcast failure means `pthread_cond_broadcast` failed which is a very fundamental system call which we don't expect to fail very often.
On the other hand if we have a logic flow somewhere which is consistent then the channel being closed can happen quite a bit.

src/cursor.c
```diff
@@ -206,16 +206,14 @@ Cursor *Cursors_Reserve(CursorList *cl, StrongRef global_spec_ref, unsigned inte
...
-
   // If we are in a coordinator ctx, the spec is NULL
-  if (spec && (spec->activeCursors >= spec->cursorsCap)) {
+  if (spec && spec->activeCursors >= RSGlobalConfig.indexCursorLimit) {
     /** Collect idle cursors now */
     Cursors_GCInternal(cl, 0);
-    if (spec->activeCursors >= spec->cursorsCap) {
-      QueryError_SetError(status, QUERY_ELIMIT, "Too many cursors allocated for index");
+    if (spec->activeCursors >= RSGlobalConfig.indexCursorLimit) {
+      QueryError_SetErrorFmt(status, QUERY_ELIMIT, "INDEX_CURSOR_LIMIT of %lld has been reached for an index", RSGlobalConfig.indexCursorLimit);
```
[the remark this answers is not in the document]
kei-nan: I think the receiver of the error will know which index it is about since he issues the query request.
If the receiver of the error chooses to output the index name on any error it gets it can cause redundant information to be shown to the user.

src/coord/rmr/rmr.c
oshadmi: So ideally we prefer to keep the `warning` level, but let's reduce the risk of spamming the log so it is  suitable for a patch release.

tests/pytests/test_cursors.py
```diff
@@ -249,7 +249,7 @@ def exceedCursorCapacity(env):
         env.cmd('FT.AGGREGATE', 'idx', '*', 'WITHCURSOR', 'COUNT', 1)
 
     # Trying to create another cursor should fail
-    env.expect('FT.AGGREGATE', 'idx', '*', 'WITHCURSOR', 'COUNT', 1).error().contains('Too many cursors allocated for index')
+    env.expect('FT.AGGREGATE', 'idx', '*', 'WITHCURSOR', 'COUNT', 1).error().contains('CURSOR_LIMIT')
```
oshadmi, line 252:
```suggestion
    env.expect('FT.AGGREGATE', 'idx', '*', 'WITHCURSOR', 'COUNT', 1).error().contains('INDEX_CURSOR_LIMIT')
```

oshadmi (approved):
Nice💪🏼

src/coord/dist_aggregate.c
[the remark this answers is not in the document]
raz-mon: Yes an assertion that the response is in the format that we want it to be in.
We actually have already many such assertions so it's probably not needed, but just from interest, are these lines covered?

raz-mon (approved):

src/coord/dist_aggregate.c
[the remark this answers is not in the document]
kei-nan: I don't think we read the response anywhere else.
So I don't think so.
For that reason would prefer to not introduce any new assertions in this branch which is meant to be more of a patch.

PR #5138, opened 2024-10-30
Remove an unnecessary assertion - [MOD-8008]

GuyAv46:

We removed an unnecessary assertion in the UV thread initialization that verifies the shard has topology information waiting before the first distributed command is sent.
The assertion is unnecessary because we have a mechanism that sets aside jobs until we can send them to all the relevant shards.

Note that still, no command will be sent until the topology is set.

**Mark if applicable**

### DIFF
--- a/src/coord/rmr/rq.c
+++ b/src/coord/rmr/rq.c
@@ -111,7 +111,6 @@ static void topologyAsyncCB(uv_async_t *async) {
 /* start the event loop side thread */
 static void sideThread(void *arg) {
   REDISMODULE_NOT_USED(arg);
-  RedisModule_Assert(__atomic_load_n(&pendingTopo, __ATOMIC_ACQUIRE));
   // Mark the event loop thread as running before triggering the topology check.
   loop_th_running = true;
   uv_async_send(&topologyAsync); // start the topology check
@@ -124,7 +123,9 @@ static void verify_uv_thread() {
     uv_timer_init(uv_default_loop(), &topologyFailureTimer);
     uv_async_init(uv_default_loop(), &topologyAsync, topologyAsyncCB);
     // Verify that we are running on the event loop thread
-    RedisModule_Assert(uv_thread_create(&loop_th, sideThread, NULL) == 0);
+    int uv_thread_create_status = uv_thread_create(&loop_th, sideThread, NULL);
+    RedisModule_Assert(uv_thread_create_status == 0);
+    REDISMODULE_NOT_USED(uv_thread_create_status);
     RedisModule_Log(RSDummyContext, "verbose", "Created event loop thread");
   }
 }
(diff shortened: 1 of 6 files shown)


src/coord/rmr/rq.c
```diff
@@ -111,7 +111,6 @@ static void topologyAsyncCB(uv_async_t *async) {
 /* start the event loop side thread */
 static void sideThread(void *arg) {
   REDISMODULE_NOT_USED(arg);
-  RedisModule_Assert(__atomic_load_n(&pendingTopo, __ATOMIC_ACQUIRE));
```
GuyAv46, line 114: The redundant assertion

src/coord/rmr/rq.c
```diff
@@ -124,7 +123,9 @@ static void verify_uv_thread() {
     uv_timer_init(uv_default_loop(), &topologyFailureTimer);
     uv_async_init(uv_default_loop(), &topologyAsync, topologyAsyncCB);
     // Verify that we are running on the event loop thread
-    RedisModule_Assert(uv_thread_create(&loop_th, sideThread, NULL) == 0);
+    int uv_thread_create_status = uv_thread_create(&loop_th, sideThread, NULL);
+    RedisModule_Assert(uv_thread_create_status == 0);
+    REDISMODULE_NOT_USED(uv_thread_create_status);
```
alonre24, line 128: Was that required? `RedisModule_Assert` is running on release build right?

tests/pytests/test.py
```diff
@@ -4022,6 +4022,28 @@ def prepare_env(env):
...
+        with TimeLimit(2, 'Failed waiting (SUCCESS!)'):
+            con.execute_command(debug_cmd(), 'SHARD_CONNECTION_STATES')
+            env.assertTrue(False, message='Expected to fail')
+    except Exception as e:
+        env.assertContains('Failed waiting (SUCCESS!)', str(e))
+
+    # Now re-set the topology and call the debug command again
+    env.expect('SEARCH.CLUSTERREFRESH').ok()
+    # We should also see the effect of setting the number of workers
+    env.expect(debug_cmd(), 'SHARD_CONNECTION_STATES').equal([ANY, [ANY] * (workers + 1)] * env.shardsCount)
```
alonre24, line 4044: Nice test.
Had we wanted to test the actual scenario (updating `WORKERS` with replica shard) - would that mimic the buggy behavior in OSS? Is the `SEARCH.CLUSTERREFRESH` command sets the topology in replicas?

alonre24 (commented):
:+1:

src/coord/rmr/rq.c
GuyAv46: Correct. But this is a better practice

tests/pytests/test.py
GuyAv46: The entire test is the closest we can get to the faulty behavior. We make sure we don’t know the topology and then push a first job to the UV thread. The refresh command is the OSS version of `clusterSet`. There is no real reason to reproduce it with replicas beside this is how we discovered it

alonre24 (approved):

PR #5139, opened 2024-10-30
[MOD-8034] expose indexing failures to INFO modules

meiravgri:
This PR exposes indexing errors counter to `INFO MODULES`

## `search_errors_*` section
added fields:
`errors_indexing_failures` - reflects the total number of indexing errors in all living indices. One document may cause failure in multiple indices.
`errors_indexing_failures_max`
## `searc_field_FieldType`

- Introduce a global counter for indexing failures for each field type.
- introduce `FieldSpec_AddError` that adds an error to a specific field object, and also updates the global per-field-type error counter.

The global counter is updated at runtime, once the error occurred. (as opposed to saving a counter per index and aggregating it per `INFO` call)
The counter should be decreased whenever an index is dropped.

We expose the global counter value in INFO modules, along with each field type stats.

**Mark if applicable**


GuyAv46 (dismissed):

src/redisearch_api.c
```diff
@@ -952,6 +951,14 @@ TotalSpecsInfo RediSearch_TotalInfo(void) {
       info.gc_stats.totalCycles += gcStats.numCycles;
       info.gc_stats.totalTime += gcStats.totalMSRun;
     }
+
+    // Index errors metrics
+    size_t index_error_count = IndexSpec_GetIndexErrorCount(sp);
+    info.indexing_errors.indexing_failures += index_error_count;
+    if (info.indexing_errors.max_indexing_failures < index_error_count) {
+      info.indexing_errors.max_indexing_failures = index_error_count;
+    }
```
alonre24, line 960: Shouldn't we initialize `max_indexing_failures` before the loop?

tests/pytests/test_index_error.py
```diff
@@ -240,6 +293,50 @@ def test_partial_doc_index_failures(env):
...
+        ]
+
+        expected_no_error_field_stats = [
+            'identifier', index_to_ok_field[idx], 'attribute', index_to_ok_field[idx], 'Index Errors',
+            ['indexing failures', 0, 'last indexing error', 'N/A', 'last indexing error key', 'N/A']
+        ]
+
+        env.assertEqual(info['num_docs'], 0)
+        env.assertEqual(info['field statistics'][0], expected_failed_field_stats)
+        env.assertEqual(info['field statistics'][1], expected_no_error_field_stats)
```
alonre24, line 337: I understand why that is correct, but looking at it now, it might be misleading... Perhaps it would be better if we at least increase the `indexing failures` on each field in the spec anyway once we encounter an error in one of the fields, so it won't look like a partial indexing was made. WDYT?

alonre24 (commented):
Nice! Few comments

src/redisearch_api.c
meiravgri: initialized to 0

meiravgri: maybe you're right because indexing_errors is a nested struct

tests/pytests/test_index_error.py
meiravgri: I think it makes the per-field indexing error counter redundant, as a we would like to identify field-types with higher errors amount

meiravgri: i don't see a reason why it may look like partial indexing since the error is counted in the index level as well.

alonre24: I see what you mean. How about trying to index each field instead of failing fast and setting the error counter for the bad fields? It would increase latency but in error code path we should care less. How big of a change is that?

src/spec.c
```diff
@@ -1569,7 +1573,9 @@ void IndexSpec_RemoveFromGlobals(StrongRef spec_ref) {
 
   // Remove spec's fields from global statistics
   for (size_t i = 0; i < spec->numFields; i++) {
-    FieldsGlobalStats_UpdateStats(spec->fields + i, -1);
+    FieldSpec *field = spec->fields + i;
+    FieldsGlobalStats_UpdateStats(field, -1);
+    FieldsGlobalStats_UpdateIndexError(field->types, -FieldSpec_GetIndexErrorCount(field));
```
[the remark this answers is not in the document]
alonre24: Nevermind. I'm convinced - let's go with this approach :)

alonre24 (approved):

GuyAv46 (approved):

PR #5150, opened 2024-11-03
[MOD-8035] Add field index error for JSON field indexing failures

meiravgri:
Introduce tracking field-level indexing error in JSON.


src/global_stats.h
```diff
@@ -55,7 +55,13 @@ extern GlobalStats RSGlobalStats;
 void FieldsGlobalStats_UpdateStats(FieldSpec *fs, int toAdd);
 
 /**
- * Exposing all the fields that > 0 to INFO command.
+ * Add or increase `toAdd` number of errors to the global index errors counter of field_type.
+ * `toAdd` can be negative to decrease the counter.
```
DvirDukhan, line 59: in which case?

src/global_stats.c
```diff
@@ -53,6 +54,10 @@ void FieldsGlobalStats_UpdateStats(FieldSpec *fs, int toAdd) {
   }
 }
 
+void FieldsGlobalStats_UpdateIndexError(FieldType field_type, int toAdd) {
+  FieldIndexErrorCounter[INDEXTYPE_TO_POS(field_type)] += toAdd;
```
DvirDukhan, line 58: adding int to size_t

src/module-init/module-init.c
```diff
@@ -135,6 +136,14 @@ void RS_moduleInfoFunc(RedisModuleInfoCtx *ctx, int for_crash_report) {
   RedisModule_InfoAddFieldDouble(ctx, "total_cycles", stats.totalCycles);
   RedisModule_InfoAddFieldDouble(ctx, "total_ms_run", stats.totalTime);
 
+  // Errors statistics
+  RedisModule_InfoAddSection(ctx, "errors");
+  IndexingErrorsStats indexing_errors = total_info.indexing_errors;
+    // indexing errors
+  RedisModule_InfoAddFieldDouble(ctx, "errors_indexing_failures", indexing_errors.indexing_failures);
+  // highest number of failures out of all specs
+  RedisModule_InfoAddFieldDouble(ctx, "errors_indexing_failures_max", indexing_errors.max_indexing_failures);
```
DvirDukhan, line 145: what is the real world purpose for this field? also consider rename to `max_indexing_errors` or something similar

tests/pytests/test_index_error.py
```diff
@@ -255,23 +354,57 @@ def test_vector_indexing_with_json(env):
...
+    env.assertEqual(field_error_dict, expected_error_dict)
 
     error_dict = to_dict(info["Index Errors"])
-    expected_error_dict[indexing_failures_str] = 1
-    expected_error_dict[last_indexing_error_key_str] = 'doc{1}'
-    expected_error_dict[last_indexing_error_str] = 'Invalid vector length. Expected 2, got 3'
-
     env.assertEqual(error_dict, expected_error_dict)
+
+def test_multiple_index_failures_json(env):
```
DvirDukhan, line 369: skip(no_json=True)

tests/pytests/test_info_modules.py
```diff
@@ -17,6 +17,28 @@ def info_modules_to_dict(conn):
...
+
+def get_search_tag_field_info(count:int,
+                              sortable_count: int = 0,
+                              no_index_count:int = 0,
+                              case_sensitive_count:int = 0,
+                              index_errors:int = 0):
+  return f'Tag={count}' \
+         f'{",Sortable=" + str(sortable_count) if sortable_count else ""}' \
+         f'{",NoIndex=" + str(no_index_count) if no_index_count else ""}' \
+         f'{",CaseSensitive=" + str(case_sensitive_count) if case_sensitive_count else ""}' \
```
DvirDukhan, line 34: I know it is not a part of this PR but do we have a tests that checks what happens when the values are 0?

src/global_stats.c
meiravgri: we are doing it to allow negative `toAdd `

tests/pytests/test_info_modules.py
meiravgri: yes 
fields options like sortable etc, are only printed when set (you can see here that the string is empty)
`IndexErrors` is always printed and is 0 zero in some of the tests

DvirDukhan (approved):

PR #5171, opened 2024-11-06
MOD-8050 MOD-8052: Add ACL username config param

raz-mon:

We add a new load-time configuration parameter `OSS_ACL_USERNAME`, which is the username with which the coordinator authenticates against the shard connections it holds.
This, together with the existing `OSS_GLOBAL_PASSWORD` will allow the user to configure an ACL user that will be able to execute RediSearch internal commands.

* Notice: Since we are talking about load-time config-params, once a user connects with a bad username\password, this will trigger an authentication loop in our uv-thread, that will end only when the config-params' values are set to the correct values, and the shards are reset (i.e., shutdown and started back). This is the current state we have with the existing password config-param as well. A change of that mechanism can be discussed outside the scope of this PR.

**Mark if applicable**

### DIFF
--- a/src/coord/config.c
+++ b/src/coord/config.c
@@ -121,6 +121,18 @@ CONFIG_GETTER(getTopologyValidationTimeout) {
   return sdsfromlonglong(realConfig->topologyValidationTimeoutMS);
 }
 
+// ACL_USERNAME
+CONFIG_GETTER(getOSSACLUsername) {
+  SearchClusterConfig *realConfig = getOrCreateRealConfig((RSConfig *)config);
+  return sdsnew(realConfig->aclUsername);
+}
+
+CONFIG_SETTER(setOSSACLUsername) {
+  SearchClusterConfig *realConfig = getOrCreateRealConfig((RSConfig *)config);
+  int acrc = AC_GetString(ac, &realConfig->aclUsername, NULL, 0);
+  RETURN_STATUS(acrc);
+}
+
 static RSConfigOptions clusterOptions_g = {
     .vars =
         {
@@ -136,7 +148,8 @@ static RSConfigOptions clusterOptions_g = {
             {.name = "OSS_GLOBAL_PASSWORD",
              .helpText = "Global oss cluster password that will be used to connect to other shards",
              .setValue = setGlobalPass,
-             .getValue = getGlobalPass},
+             .getValue = getGlobalPass,
+             .flags = RSCONFIGVAR_F_IMMUTABLE},
             {.name = "CONN_PER_SHARD",
              .helpText = "Number of connections to each shard in the cluster. Default to 0. "
                          "If 0, the number of connections is set to `WORKERS` + 1.",
@@ -157,6 +170,11 @@ static RSConfigOptions clusterOptions_g = {
                          "Default is 30000 (30 seconds). 0 means no timeout.",
              .setValue = setTopologyValidationTimeout,
              .getValue = getTopologyValidationTimeout,},
+             {.name = "OSS_ACL_USERNAME",
+             .helpText = "Set the username for the ACL user used by the coordinator to connect to the shards on OSS cluster.",
+             .setValue = setOSSACLUsername,
+             .getValue = getOSSACLUsername,
+             .flags = RSCONFIGVAR_F_IMMUTABLE},
             {.name = NULL}
             // fin
         }
--- a/src/coord/rmr/conn.c
+++ b/src/coord/rmr/conn.c
@@ -627,8 +627,7 @@ static void MRConn_ConnectCallback(const redisAsyncContext *c, int status) {
     if (ssl_context) SSL_CTX_free(ssl_context);
   }
 
-  // If this is an authenticated connection, we need to atu
-
+  // If this is an authenticated connection, we need to authenticate
   if (conn->ep.password) {
     if (MRConn_SendAuth(conn) != REDIS_OK) {
       detachFromConn(conn, 1);
--- a/src/coord/rmr/conn.c
+++ b/src/coord/rmr/conn.c
@@ -6,6 +6,8 @@
 
 #include "conn.h"
 #include "reply.h"
+#include "../config.h"
+#include "module.h"
 #include "hiredis/adapters/libuv.h"
 
 #include <uv.h>
@@ -423,7 +425,7 @@ static int MRConn_SendAuth(MRConn *conn) {
   CONN_LOG(conn, "Authenticating...");
 
   // if we failed to send the auth command, start a reconnect loop
-  if (redisAsyncCommand(conn->conn, MRConn_AuthCallback, conn, "AUTH %s", conn->ep.auth) ==
+  if (redisAsyncCommand(conn->conn, MRConn_AuthCallback, conn, "AUTH %s %s", clusterConfig.aclUsername, conn->ep.password) ==
       REDIS_ERR) {
     MRConn_SwitchState(conn, MRConn_ReAuth);
     return REDIS_ERR;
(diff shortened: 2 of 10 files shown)


src/coord/rmr/conn.c
```diff
@@ -6,6 +6,8 @@
 
 #include "conn.h"
 #include "reply.h"
+#include "../config.h"
```
kei-nan, line 9: can you fix this include line?
I don't think we generally use relative path includes in our code.

src/coord/rmr/conn.c
```diff
@@ -423,7 +425,7 @@ static int MRConn_SendAuth(MRConn *conn) {
   CONN_LOG(conn, "Authenticating...");
 
   // if we failed to send the auth command, start a reconnect loop
-  if (redisAsyncCommand(conn->conn, MRConn_AuthCallback, conn, "AUTH %s", conn->ep.auth) ==
+  if (redisAsyncCommand(conn->conn, MRConn_AuthCallback, conn, "AUTH %s %s", clusterConfig.aclUsername, conn->ep.password) ==
       REDIS_ERR) {
     MRConn_SwitchState(conn, MRConn_ReAuth);
```
kei-nan, line 430: I don't get it, looked at all the `MRConn_SendAuth` calls.
It seems like if it returns failure then we change the state to connecting, which is fine.
But internally in between we change the state to Reauth, which seems to start a timer.
Isn't there a better way of doing this?
Seems redundant to move to reauth state just to immediately move to connecting state.

tests/pytests/test_config.py
```diff
@@ -1,9 +1,32 @@
 from RLTest import Env
 from includes import *
-from common import skip, config_cmd
+from common import *
 
 not_modifiable = 'Not modifiable at runtime'
 
+def _test_arg_str(arg_name, arg_value, ret_value=None):
```
kei-nan, line 7:
```suggestion
def _test_config_str(arg_name, arg_value, ret_value=None):
```

tests/pytests/test_config.py
```diff
@@ -1,9 +1,32 @@
...
+    if ret_value == None:
+        ret_value = arg_value
+    env = Env(moduleArgs=arg_name + ' ' + arg_value, noDefaultModuleArgs=True)
+    if env.env == 'existing-env':
+        env.skip()
+    env.expect(config_cmd(), 'get', arg_name).equal([[arg_name, ret_value]])
+    env.stop()
+
+def _test_arg_num(arg_name, arg_value):
```
kei-nan, line 16:
```suggestion
def _test_config_num(arg_name, arg_value):
```

tests/pytests/test_config.py
```diff
@@ -1,9 +1,32 @@
...
+    env = Env(moduleArgs=f'{arg_name} {arg_value}', noDefaultModuleArgs=True)
+    if env.env == 'existing-env':
+        env.skip()
+    env.expect(config_cmd(), 'get', arg_name).equal([[arg_name, str(arg_value)]])
+    env.stop()
+
+def _test_arg_true_false(arg_name, res):
```
kei-nan, line 23:
```suggestion
def _test_config_boolean(arg_name, res):
```

tests/pytests/test_config.py
```diff
@@ -370,14 +371,45 @@ def testAllConfigCoord(env):
...
+
+    # Setting the `OSS_ACL_USERNAME` configuration without the `OSS_GLOBAL_PASSWORD`
+    # the configuration should not do anything since we don't try to authenticate.
+    _test_arg_str('OSS_ACL_USERNAME', 'test')
+
+    # Set both the username and password. This should fail since we have no such
+    # user.
+    env = Env(moduleArgs='OSS_ACL_USERNAME test_user OSS_GLOBAL_PASSWORD 123456', noDefaultModuleArgs=True)
+
+    timeout = 3 # 5 seconds, more than enough for the an env to be up
```
kei-nan, line 409:
```suggestion
    timeout = 3 # 3 seconds, more than enough for the an env to be up
```

tests/pytests/test_config.py
```diff
@@ -370,14 +371,45 @@ def testAllConfigCoord(env):
...
+    try:
+        with TimeLimit(timeout):
```
kei-nan, line 411: I  think the behavior should be that the user gets a proper error.
I don't think we should be in an infinite loop.

tests/pytests/test_config.py
```diff
@@ -370,14 +371,45 @@ def testAllConfigCoord(env):
...
+            env.cmd('FT.SEARCH', 'idx', '*')
+            env.assertTrue(False)
+    except Exception as e:
+        env.assertEqual(str(e), 'Timeout: operation timeout exceeded')
```
kei-nan, line 415: I think tests that rely on timeouts tend to eventually become flaky.

kei-nan (changes requested):
Left a few comments, 
Mainly I think the infinite loop scenario is not ideal and I suggest considering returning a proper error to the end user.

src/coord/rmr/conn.c
raz-mon: Where do we change to connecting?
I think we stay on `ReAuth`.

tests/pytests/test_config.py
raz-mon: I agree in general, but this is an interesting state cluster-wise as well, thus should be thought about separately since this is an existing issue.
Currently the cluster will need a restart for the new load-time config-param values anyways so I'm not sure this is a big problem.
Bottom line - this is the current behavior, and if we want to fix it we need to think about it more and do it in a more complete manner separately.

tests/pytests/test_config.py
raz-mon: I agree it's not ideal, yet a hanging client is the current behavior of this situation, so until we don't return a proper error for this situation I have no other way to test this.
I can comment this part out for now if we don't want it to be part of our CI (that I can understand and agree with)

tests/pytests/test_config.py
kei-nan: I would accept the existing behavior argument if the test didn't exist.
The fact you created a test to verify this behavior seems to imply this behavior is part of the code you added.
Otherwise what exactly did you add behavior wise that you want to test.

src/coord/rmr/conn.c
kei-nan: See all the places `MRConn_SendAuth` gets called and what happens when it returns an error code to the caller.
The calling code proceeds to then change the start to connecting.
Unless I'm missing something.

tests/pytests/test_config.py
```diff
@@ -148,79 +173,55 @@ def testAllConfig(env):
     env.assertEqual(res_dict['UNION_ITERATOR_HEAP'][0], '20')
     env.assertEqual(res_dict['INDEX_CURSOR_LIMIT'][0], '128')
 
-@skip(cluster=True)
+# @skip(cluster=True)
```
oshadmi, line 176: Remove line?

src/coord/rmr/conn.c
kei-nan: After discussing, we decided it is out of scope.

kei-nan (approved):

src/coord/config.c
```diff
@@ -136,7 +148,8 @@ static RSConfigOptions clusterOptions_g = {
             {.name = "OSS_GLOBAL_PASSWORD",
              .helpText = "Global oss cluster password that will be used to connect to other shards",
              .setValue = setGlobalPass,
-             .getValue = getGlobalPass},
+             .getValue = getGlobalPass,
+             .flags = RSCONFIGVAR_F_IMMUTABLE},
```
oshadmi, line 152: Is this a breaking change?
Is it only breaking the ability to set the password, but actually it was only considered in module init?

src/coord/config.c
```diff
@@ -157,6 +170,11 @@ static RSConfigOptions clusterOptions_g = {
                          "Default is 30000 (30 seconds). 0 means no timeout.",
              .setValue = setTopologyValidationTimeout,
              .getValue = getTopologyValidationTimeout,},
+             {.name = "OSS_ACL_USERNAME",
```
oshadmi, line 173: Can we do it only in OSS? (use `#ifdef RS_CLUSTER_ENTERPRISE`)

src/coord/config.c
raz-mon: Effectively not, since we rely on load-time setting of this config param only

oshadmi (approved):

src/coord/config.c
raz-mon: We didn't do this until now, we would need more approvals to start doing it

PR #5183, opened 2024-11-08
Add global marked deleted to stats [MOD-6053]

alonre24:

Add the total logically deleted documents (i.e., documents that were deleted and have not yet been handled by the GC) among all indexes to the global shard info report. 

**Mark if applicable**

### DIFF
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -1299,9 +1299,6 @@ static int periodicCb(void *privdata) {
 
   // We need to acquire the GIL to use the fork api
   RedisModule_ThreadSafeContextLock(ctx);
-  // Now that we hold the GIL, we can cache this value knowing it won't change by the main thread
-  // upon deleting a docuemnt (this is the actual number of documents to be cleaned by the fork).
-  size_t num_docs_to_clean = gc->deletedDocsFromLastRun;
 
   gc->execState = FGC_STATE_SCANNING;
 
@@ -1320,6 +1317,9 @@ static int periodicCb(void *privdata) {
     return 1;
   }
 
+  // Now that we hold the GIL, we can cache this value knowing it won't change by the main thread
+  // upon deleting a docuemnt (this is the actual number of documents to be cleaned by the fork).
+  size_t num_docs_to_clean = gc->deletedDocsFromLastRun;
   gc->deletedDocsFromLastRun = 0;
 
   gc->retryInterval.tv_sec = RSGlobalConfig.gcConfigParams.forkGc.forkGcRunIntervalSec;
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -25,6 +25,7 @@
 #include "rmutil/rm_assert.h"
 #include "suffix.h"
 #include "resp3.h"
+#include "global_stats.h"
 
 #define GC_WRITERFD 1
 #define GC_READERFD 0
@@ -1302,7 +1303,7 @@ static int periodicCb(void *privdata) {
     StrongRef_Release(early_check);
     return 1;
   }
-
+  size_t num_docs_to_clean = gc->deletedDocsFromLastRun;
   int gcrv = 1;
   pid_t cpid;
   TimeSample ts;
@@ -1395,6 +1396,7 @@ static int periodicCb(void *privdata) {
   gc->stats.numCycles++;
   gc->stats.totalMSRun += msRun;
   gc->stats.lastRunTimeMs = msRun;
+  IndexsGlobalStats_UpdateLogicallyDeleted(-num_docs_to_clean);
 
   return gcrv;
 }
@@ -1437,6 +1439,7 @@ void FGC_Apply(ForkGC *gc) NO_TSAN_CHECK {
 
 static void onTerminateCb(void *privdata) {
   ForkGC *gc = privdata;
+  IndexsGlobalStats_UpdateLogicallyDeleted(-gc->deletedDocsFromLastRun);
   WeakRef_Release(gc->index);
   RedisModule_FreeThreadSafeContext(gc->ctx);
   rm_free(gc);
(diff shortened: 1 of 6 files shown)


src/global_stats.c
```diff
@@ -162,3 +162,12 @@ void DialectsGlobalStats_AddToInfo(RedisModuleInfoCtx *ctx) {
...
   }
 }
+
+void IndexsGlobalStats_UpdateLogicallyDeleted(int toAdd) {
+    INCR_BY(RSGlobalStats.totalStats.logically_deleted, toAdd);
+}
+
+void IndexesGlobalStats_AddToInfo(RedisModuleInfoCtx *ctx) {
+    RedisModule_InfoAddSection(ctx, "indexes_statistics");
+    RedisModule_InfoAddFieldULongLong(ctx, "total_logically_deleted_docs", READ(RSGlobalStats.totalStats.logically_deleted));
```
GuyAv46, line 172: Don’t we have some section this metric could fit in? Are we planning to add more fields to this section?
Consider adding it to an existing section (maybe the GC section?)

GuyAv46 (commented):
Nice
A few comments
Note that you can hold the GIL in most if not all of the places you used the new counter API. Consider ensuring that and remove the need for atomic counting and reading

src/global_stats.c
alonre24: I guess we can move it to GC, WDYT @meiravgri ?

src/fork_gc.c
```diff
@@ -1302,7 +1303,7 @@ static int periodicCb(void *privdata) {
     StrongRef_Release(early_check);
     return 1;
   }
-
+  size_t num_docs_to_clean = gc->deletedDocsFromLastRun;
```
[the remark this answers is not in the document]
alonre24: I don't think that we need the GIL here if we know that `gc` is the owner of this struct, but I'll move anyway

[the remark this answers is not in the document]
GuyAv46: The periodic callback is called from the GC thread while the counter is increased from the `onDelete` callback that is called from the main thread

src/global_stats.c
meiravgri: We already have index statistics section called "index"
I'd put it there along with `number_of_indexes` and `active_*` statistics

src/fork_gc.c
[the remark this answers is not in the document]
meiravgri: we rely here on the assumption that the child process will clean `gc->deletedDocsFromLastRun` docs, so we need to make sure it happens atomically with the fork.

tests/pytests/test_info_modules.py
```diff
@@ -454,3 +454,49 @@ def test_redis_info_modules_vecsim():
...
+  num_fields = 3
+  env.expect('FT.CREATE', 'idx1', 'SCHEMA', 'text', 'TEXT', 'tag', 'TAG', 'num', 'NUMERIC').ok()
+  env.expect(debug_cmd(), 'GC_STOP_SCHEDULE', 'idx1').ok()  # Stop GC for this index to keep the deleted docs
+  set_doc(f'doc:1').equal(num_fields)
+  env.assertEqual(get_logically_deleted_docs(), 0)
+  env.expect('DEL', 'doc:1').equal(1)
+  env.assertEqual(get_logically_deleted_docs(), 1)
+
+  # Create another index, expect that the deleted document will not be indexed.
+  env.expect('FT.CREATE', 'idx2', 'SCHEMA', 'text', 'TEXT', 'tag', 'TAG', 'num', 'NUMERIC').ok()
```
meiravgri, line 477: lets add here fields that are not collected by the gc, vector and geometry, and validate they are not counted

tests/pytests/test_info_modules.py
```diff
@@ -454,3 +454,49 @@ def test_redis_info_modules_vecsim():
...
+  env.expect(debug_cmd(), 'GC_STOP_SCHEDULE', 'idx2').ok()  # Stop GC for this index to keep the deleted docs
```
meiravgri, line 478: you could set `FORK_GC_RUN_INTERVAL` config to a very high number and run with force invoke to avoid all this wajars

src/fork_gc.c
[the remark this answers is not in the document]
meiravgri: consider adding a cpp test to simulate the scenario where this variable is not guarded and the counter is out of balance

src/fork_gc.c
```diff
@@ -1437,6 +1439,7 @@ void FGC_Apply(ForkGC *gc) NO_TSAN_CHECK {
 
 static void onTerminateCb(void *privdata) {
   ForkGC *gc = privdata;
+  IndexsGlobalStats_UpdateLogicallyDeleted(-gc->deletedDocsFromLastRun);
```
meiravgri, line 1442: is there a scenario where the gc is manually stopped and this callback will not be called?

src/global_stats.c
meiravgri: Move to this wonderful function the other fields of the section

src/global_stats.h
```diff
@@ -39,6 +39,7 @@ typedef struct {
   size_t total_query_commands;        // Number of successful query commands, including `FT.CURSOR READ`
   clock_t total_query_execution_time; // Total time spent on queries (in clock ticks)
   uint_least8_t used_dialects;        // bitarray of dialects used by all indices
+  size_t logically_deleted;           // Number of logically deleted documents in all indices
```
meiravgri, line 42: please add that `logically_deleted` means documents that were marked wit `DELETED` flag but whose memory was not yet cleaned by the GC

meiravgri (commented):
nice and clean (tartei)

tests/pytests/test_info_modules.py
alonre24: What do you mean? We count on a document level, we don't distinguish between documents that have fields collected by the GC and ones that don't.

src/fork_gc.c
alonre24: Shouldn't be. If so, and the spec is dropped, then no one frees the GC and we get a leak

GuyAv46 (approved):

tests/pytests/test_info_modules.py
meiravgri: right

meiravgri (approved):

PR #5184, opened 2024-11-08
MOD-8080, MOD-8055: Fix crash on error handling, Add OSS ACL validation for FT.SUG* commands

raz-mon:

This PR does the following:
1. Fix a bug we had in which we did not handle a possible error in redis's side upon calling the `RedisModule_Call()` API function causing a NULL response that we did not take into account.
2. Adds the key indexes to the command registration of the `FT.SUG*` commands so that Redis performs the ACL key validations properly.
3. Removes the extra redirection done by the coordinator on OSS, since we now handle the command in the correct shard with the relevant key.
4. Tests that we correctly validate user ACL permissions in our `FT.SUG*` commands.

**Mark if applicable**


src/module.c
```diff
@@ -2725,12 +2725,13 @@ static int genericCallUnderscoreVariant(RedisModuleCtx *ctx, RedisModuleString *
...
   /*
    * v - argv input array of RedisModuleString
+   * E - return errors as RedisModuleCallReply object (instead of NULL)
    * C - same client
    * M - respect OOM
    * 0 - same RESP protocol
    * ! - replicate the command if needed (allows for replication)
    */
-  RedisModuleCallReply *r = RedisModule_Call(ctx, localCmd, "vCM0!", argv + 1, argc - 1);
+  RedisModuleCallReply *r = RedisModule_Call(ctx, localCmd, "vECM0!", argv + 1, argc - 1);
```
oshadmi, line 2734: Should this be in a separate PR so it could be cherry-picked to other versions?

oshadmi:
@raz-mon Where is this change? (BTW it could be considered a breaking change)
> Fixes the key-position flags of the registration of the FT.SUG* commands

raz-mon:
@oshadmi Comment updated, I removed this from this PR.
I do think we should consider adding it, but it is not necessary for this PR.

src/module.c
raz-mon: Actually I can cherry-pick this to all the version branches, since I do not test the new `search` category we added to `8.0` only.
Added the backports.

GuyAv46 (dismissed):

GuyAv46 (approved):

PR #5186, opened 2024-11-09
Performance improvement in indexBulkFields - [MOD-8093]

GuyAv46:
## **Describe the changes in the pull request**

We are removing a redundant caching mechanism in `indexBulkFields` that is no longer in use and takes a long time to initialize.

The caching mechanism helped with field keys and inverted index caching. The first is no longer relevant, and the second only helps when we use bulk indexing (which we don’t do anymore).

The redundant caching mechanism requires a large memory initialization, which takes a large percentage of the indexing time for no reason.

### Note:
This is a performance fix, the bug does not affect the correctness of any search

### Future work:
Removing the need for "formatted keys", stop using the spec's dictionary for fields, and store the relevant data in the field spec instead (today we usually extract the field data by first obtaining the field spec by its name, then getting the formatted redis string from it, and lastly performing a dictionary lookup by the formatted key name)

**Mark if applicable**

### DIFF
--- a/src/geometry_index.c
+++ b/src/geometry_index.c
@@ -33,7 +33,7 @@ static GeometryIndex *openGeometryKeysDict(const IndexSpec *spec, RedisModuleStr
   if (!write) {
     return NULL;
   }
-  
+
   GeometryIndex *idx = GeometryIndexFactory(fs->geometryOpts.geometryCoords);
   const GeometryApi *api = GeometryApi_Get(idx);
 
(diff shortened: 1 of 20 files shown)


alonre24 (approved):

alonre24 (approved):

src/geometry_index.c
```diff
@@ -46,43 +46,23 @@ static GeometryIndex *openGeometryKeysDict(const IndexSpec *spec, RedisModuleStr
...
-  }
-  *idxKey = RedisModule_OpenKey(redisCtx, keyName, REDISMODULE_READ | REDISMODULE_WRITE);
-
-  /* Create an empty value object if the key is currently empty. */
-  if (RedisModule_KeyType(*idxKey) == REDISMODULE_KEYTYPE_EMPTY) {
-    GeometryIndex *idx = GeometryIndexFactory(fs->geometryOpts.geometryCoords);
-    RedisModule_ModuleTypeSetValue(*idxKey, GeometryIndexType, idx);
-    return idx;
-  } 
-  if (RedisModule_ModuleTypeGetType(*idxKey) == GeometryIndexType) {
```
oshadmi, line 71: @GuyAv46 We can also remove the global `GeometryIndexType`?

PR #5187, opened 2024-11-10
MOD-8058: Avoid error with _NUM_SSTRING

oshadmi:

When calling the shard with an the argument `_NUM_SSTRING` it is returning numeric values as errors. Need to avoid this error as it unneededly increases the error count and messes up `info errorstats`.

Before fix:
```
                |       Standalone                              |           Called by Coordinator              |
                |-----------------------------------------------|----------------------------------------------|
                |   Without _NUM_SSTRING |  With _NUM_SSTRING   |   Without _NUM_SSTRING |  With _NUM_SSTRING  |
----------------|------------------------|----------------------|------------------------|---------------------|
STRING format   |   string               |  ERROR (N/A)         |   string (N/A)         |  ERROR              |
                |------------------------|----------------------|------------------------|---------------------|
EXPAND format   |                                           numeric                                            |
                |------------------------|----------------------|------------------------|---------------------|
```

After fix:

For RESP3 we are replying double numbers.
For RESP2 we are replying with simple string from the shards to the coordinator, and the reply is converted back to number by `MRReply_ToValue()`. 
```

                |       Standalone                              |           Called by Coordinator              |
                |-----------------------------------------------|----------------------------------------------|
                |   Without _NUM_SSTRING |  With _NUM_SSTRING   |   Without _NUM_SSTRING |  With _NUM_SSTRING  |
----------------|------------------------|----------------------|------------------------|---------------------|
STRING format   |   string               |  numeric (N/A)       |   string (N/A)         |  numeric            |
                |------------------------|----------------------|------------------------|---------------------|
EXPAND format   |                                           numeric                                            |
                |------------------------|----------------------|------------------------|---------------------|



1. MOD-8058


1. ...
2. ...

**Mark if applicable**

### CODE REVIEW

File: src/value.c [hunk 1]
```diff
@@ -675,7 +675,11 @@ int RSValue_SendReply(RedisModule_Reply *reply, const RSValue *v, SendReplyFlags
         size_t len = RSValue_NumToString(v->numval, buf);
 
         if (flags & SENDREPLY_FLAG_TYPED) {
-          return RedisModule_Reply_Error(reply, buf);
+          if (reply->resp3) {
+            return RedisModule_Reply_Double(reply, v->numval);
+          } else { // RESP2
+            return RedisModule_Reply_Error(reply, buf);
```
raz-mon, line 681: Add a comment as to why we do this ugly workaround

raz-mon (commented):
Looking good 👍 
A couple of notes:
1. Please add documentation in both sides of this workaround as to why we do it, i.e., in both the shard side and coordinator side.
2. In order to fully mitigate this bad behavior (in the error-stats only), we can try to use RESP3 only in our coordinator-shard communication. This is not an effort for this PR, but let's at least assess that effort and create a ticket for a continuation of this semi-remedy to fully solve the issue.
3. We can also investigate the option of using the `STATUS` reply type. It's a bit risky and error-prone and smells a lot like the old solution, so I'm not a fan, but it will at least not result in bad error-stats that can be highly confusing. This can be part of the continuation ticket to fully solve this, as this requires some digging in the code to make sure that it is fine (since we already reply with simple strings, which result in status reply types).
4. Please update the PR comment 🙂.

MeirShpilraien:
Lets specify in the top comment that it only fixes the problem if resp3 is used.

src/coord/dist_aggregate.c
```diff
@@ -166,7 +166,14 @@ RSValue *MRReply_ToValue(MRReply *r) {
   if (!r) return RS_NullVal();
   RSValue *v = NULL;
   switch (MRReply_Type(r)) {
-    case MR_REPLY_STATUS:
+    case MR_REPLY_STATUS: {
+      // In RESP2, RSValue_SendReply() send double values as simple strings, 
+      // so we need to convert them back to doubles here
+      double d = 0;
+      RS_LOG_ASSERT(MRReply_ToDouble(r, &d), "non-numeric status reply");
```
raz-mon, line 173:
```suggestion
      RS_LOG_ASSERT(MRReply_ToDouble(r, &d), "Unexpected non-numeric status reply");
```

raz-mon (commented):
Nice.
Notice that this is an error-prone fix (just as the original one was), thus I think we should think of this as a dangerous patch fix and nothing more, and move on to use RESP3 in our shard-coordinator communication ASAP (CC @oshadmi  @DvirDukhan).

src/coord/dist_aggregate.c
```diff
@@ -166,7 +166,14 @@ RSValue *MRReply_ToValue(MRReply *r) {
...
+      v = RS_NumVal(d);
+      break;
+    }
     case MR_REPLY_STRING: {
       size_t l;
       const char *s = MRReply_String(r, &l);
```
kei-nan, line 179: Consider changing the code in `MR_REPLY_ERROR` case.
Maybe add an assertion there as well?

raz-mon (approved):

kei-nan (approved):

PR #5204, opened 2024-11-12
MOD-8097, MOD-8114 Fix memory counting

meiravgri:
This PR fixes bugs in statistics updates during gc.

## wrong update of `range->invertedIndexSize` when returns empty from the fork process
In the previous implementation, when the inverted index of a numeric range returned with no docs from the fork process, `range->invertedIndexSize` stat was set to 0.
This is wrong since an empty range consumes memory of inverted index metadata struct size + block struct size + initial block capacity size.
In this PR `range->invertedIndexSize` is decreased by `nbytesCollected` and increased by `nbytesAdded`.

In addition, once the gc parent finishes iterating all the nodes received from the fork, it starts cleaning empty nodes. 
`NumericRangeTree_TrimEmptyLeaves` returns the amount of memory released due to empty nodes removal.
This is where we update the stats of a completely removed range (inverted index).

## fix overflow in spec->stats.invertedSize during `FGC_updateStats`

before the fix: 
```
sctx->spec->stats.invertedSize += bytesAdded - bytesCollected;
```
Since all the variables are of type size_t, and `bytesAdded` might be smaller than `bytesCollected`, sctx->spec->stats.invertedSize might get overflowed.

the fix is to split this into two lines.
```
sctx->spec->stats.invertedSize += bytesAdded;
sctx->spec->stats.invertedSize -= bytesCollected;
```

## Fix bytes collected by the gc statistics

Currently, the number of bytes collected by the gc inclused `bytesAdded`, but it doesn’t include `bytesAdded` during the cleaning process.

This might result in counting bytes added by the parent in `bytesCollected`, when releasing an entire inverted index.

After the fix, we will include also `bytesAdded` in the gc stats
```
  gc->stats.totalCollected += bytesCollected;
  gc->stats.totalCollected -= bytesAdded;
```

# Additional changes in this PR
- Add total `invertedIndexSize` to the numeric tree struct
- Since opening an index in WRITE mode, might result in creating it, debug commands open inverted index for READ. If the invertedindex was not initialized yet, the result will be in the format of an empty index.
- add numeric tree struct overhead to the memory count in `IndexSpec_TotalMemUsage`

# Additional bug fixes
- Fix unsafe read of a numeric tree without holding a valid reference to its spec in the GC thread - MOD-8114

### DIFF
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -1008,7 +1008,7 @@ static FGCError FGC_parentHandleNumeric(ForkGC *gc) {
 
   rm_free(fieldName);
 
-  if (status == FGC_DONE && rt && gc->cleanNumericEmptyNodes) {
+  if (status == FGC_COLLECTED && rt && gc->cleanNumericEmptyNodes) {
     StrongRef spec_ref = WeakRef_Promote(gc->index);
     IndexSpec *sp = StrongRef_Get(spec_ref);
     if (!sp) return FGC_SPEC_DELETED;
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -1008,18 +1008,17 @@ static FGCError FGC_parentHandleNumeric(ForkGC *gc) {
 
   rm_free(fieldName);
 
-  if (gc->cleanNumericEmptyNodes && rt && rt->emptyLeaves >= rt->numRanges / 2) {
+  if (status == FGC_DONE && rt && gc->cleanNumericEmptyNodes) {
     StrongRef spec_ref = WeakRef_Promote(gc->index);
     IndexSpec *sp = StrongRef_Get(spec_ref);
-    if (!sp) {
-      return FGC_SPEC_DELETED;
-    }
+    if (!sp) return FGC_SPEC_DELETED;
     RedisSearchCtx sctx = SEARCH_CTX_STATIC(gc->ctx, sp);
     RedisSearchCtx_LockSpecWrite(&sctx);
-
-    NRN_AddRv rv = NumericRangeTree_TrimEmptyLeaves(rt);
-    // rv.sz is the number of bytes added. Since we are cleaning empty leaves, it should be negative
-    FGC_updateStats(gc, &sctx, 0, -rv.sz, 0);
+    if (rt->emptyLeaves >= rt->numRanges / 2) { // TODO: count `numLeaves` in the tree and use it here
+      NRN_AddRv rv = NumericRangeTree_TrimEmptyLeaves(rt);
+      // rv.sz is the number of bytes added. Since we are cleaning empty leaves, it should be negative
+      FGC_updateStats(gc, &sctx, 0, -rv.sz, 0);
+    }
     RedisSearchCtx_UnlockSpec(&sctx);
     StrongRef_Release(spec_ref);
   }
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -421,7 +421,12 @@ static void FGC_childCollectNumeric(ForkGC *gc, RedisSearchCtx *sctx) {
 
   for (int i = 0; i < array_len(numericFields); ++i) {
     RedisModuleString *keyName = IndexSpec_GetFormattedKey(sctx->spec, numericFields[i], INDEXFLD_T_NUMERIC);
-    NumericRangeTree *rt = OpenNumericIndex(sctx, keyName);
+    NumericRangeTree *rt = openNumericKeysDict(sctx->spec, keyName, OPEN_INDEX_READ);
+
+    // No entries were added to the numeric field, hence the tree was not initialized
+    if (!rt) {
+      continue;
+    }
 
     NumericRangeTreeIterator *gcIterator = NumericRangeTreeIterator_New(rt);
 
@@ -958,8 +963,7 @@ static FGCError FGC_parentHandleNumeric(ForkGC *gc) {
 
     RedisSearchCtx_LockSpecWrite(sctx);
 
-    RedisModuleString *keyName = IndexSpec_GetFormattedKeyByName(sctx->spec, fieldName, INDEXFLD_T_NUMERIC);
-    rt = OpenNumericIndex(sctx, keyName);
+      rt = openNumericKeysDict(sctx->spec, keyName, OPEN_INDEX_READ);
 
     if (rt->uniqueId != rtUniqueId) {
       status = FGC_PARENT_ERROR;
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -934,9 +934,12 @@ cleanup:
 static FGCError FGC_parentHandleNumeric(ForkGC *gc) {
   size_t fieldNameLen;
   char *fieldName = NULL;
+  const FieldSpec *fs = NULL;
+  RedisModuleString *keyName = NULL;
   uint64_t rtUniqueId;
   NumericRangeTree *rt = NULL;
   FGCError status = recvNumericTagHeader(gc, &fieldName, &fieldNameLen, &rtUniqueId);
+  bool initialized = false;
   if (status == FGC_DONE) {
     return FGC_DONE;
   }
@@ -963,7 +966,12 @@ static FGCError FGC_parentHandleNumeric(ForkGC *gc) {
 
     RedisSearchCtx_LockSpecWrite(sctx);
 
+    if (!initialized) {
+      fs = IndexSpec_GetField(sctx->spec, fieldName, strlen(fieldName));
+      keyName = IndexSpec_GetFormattedKey(sctx->spec, fs, fs->types);
+      fieldStats = FGC_getFieldStatsObject(gc, fs->types);
       rt = openNumericKeysDict(sctx->spec, keyName, OPEN_INDEX_READ);
+    }
 
     if (rt->uniqueId != rtUniqueId) {
       status = FGC_PARENT_ERROR;
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -53,6 +53,7 @@ static void FGC_updateStats(ForkGC *gc, RedisSearchCtx *sctx,
   gc->stats.totalCollected -= bytesAdded;
 }
 
+// Buff shouldn't be NULL.
 static void FGC_sendFixed(ForkGC *fgc, const void *buff, size_t len) {
   RS_LOG_ASSERT(len > 0, "buffer length cannot be 0");
   ssize_t size = write(fgc->pipefd[GC_WRITERFD], buff, len);
@@ -68,7 +69,7 @@ static void FGC_sendFixed(ForkGC *fgc, const void *buff, size_t len) {
 
 static void FGC_sendBuffer(ForkGC *fgc, const void *buff, size_t len) {
   FGC_SEND_VAR(fgc, len);
-  if (buff && len > 0) {
+  if (len > 0) {
     FGC_sendFixed(fgc, buff, len);
   }
 }
@@ -256,7 +257,8 @@ static bool FGC_childRepairInvidx(ForkGC *gc, RedisSearchCtx *sctx, InvertedInde
   if (array_len(blocklist) == idx->size) {
     // no empty block, there is no need to send the blocks array. Don't send
     // any new blocks.
-    FGC_sendBuffer(gc, NULL, 0);
+    size_t len = 0;
+    FGC_SEND_VAR(gc, len);
   } else {
     FGC_sendBuffer(gc, blocklist, array_len(blocklist) * sizeof(*blocklist));
   }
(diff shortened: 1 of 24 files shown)


src/fork_gc.c
```diff
@@ -66,7 +68,7 @@ static void FGC_sendFixed(ForkGC *fgc, const void *buff, size_t len) {
 
 static void FGC_sendBuffer(ForkGC *fgc, const void *buff, size_t len) {
   FGC_SEND_VAR(fgc, len);
-  if (len > 0) {
+  if (buff && len > 0) {
```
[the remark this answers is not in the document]
meiravgri: not that im aware of

src/fork_gc.c
```diff
@@ -959,8 +963,11 @@ static FGCError FGC_parentHandleNumeric(ForkGC *gc) {
 
     RedisSearchCtx_LockSpecWrite(sctx);
 
-    RedisModuleString *keyName = IndexSpec_GetFormattedKeyByName(sctx->spec, fieldName, INDEXFLD_T_NUMERIC);
-    rt = OpenNumericIndex(sctx, keyName);
+    if (!initialized) {
+      fs = IndexSpec_GetField(sctx->spec, fieldName, strlen(fieldName));
+      keyName = IndexSpec_GetFormattedKey(sctx->spec, fs, fs->types);
+      rt = openNumericKeysDict(sctx->spec, keyName, OPEN_INDEX_READ);
+    }
```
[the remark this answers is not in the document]
meiravgri: 1. oops.
2. right. We need the spec to get the these objects
3. leftovers from future PR 😅

src/fork_gc.c
```diff
@@ -993,16 +1002,15 @@ static FGCError FGC_parentHandleNumeric(ForkGC *gc) {
...
     StrongRef spec_ref = WeakRef_Promote(gc->index);
     IndexSpec *sp = StrongRef_Get(spec_ref);
-    if (!sp) {
-      return FGC_SPEC_DELETED;
-    }
+    if (!sp) return FGC_SPEC_DELETED;
     RedisSearchCtx sctx = SEARCH_CTX_STATIC(gc->ctx, sp);
     RedisSearchCtx_LockSpecWrite(&sctx);
-    if (gc->cleanNumericEmptyNodes) {
+    if (rt->emptyLeaves >= rt->numRanges / 2) { // TODO: count `numLeaves` in the tree and use it here
```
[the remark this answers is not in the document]
meiravgri: amazing

src/spec.c
```diff
@@ -337,6 +337,25 @@ double IndexesScanner_IndexedPercent(IndexesScanner *scanner, IndexSpec *sp) {
   }
 }
 
+size_t IndexSpec_collect_numeric_overhead(IndexSpec *sp) {
+  // Traverse the fields and calculates the overhead of the tags
```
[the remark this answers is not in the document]
meiravgri: which tags?

src/spec.c
```diff
@@ -369,6 +388,7 @@ size_t IndexSpec_TotalMemUsage(IndexSpec *sp, size_t doctable_tm_size, size_t ta
   res += doctable_tm_size ? doctable_tm_size : TrieMap_MemUsage(sp->docs.dim.tm);
   res += text_overhead ? text_overhead :  IndexSpec_collect_text_overhead(sp);
   res += tags_overhead ? tags_overhead : IndexSpec_collect_tags_overhead(sp);
+  res += IndexSpec_collect_numeric_overhead(sp);
```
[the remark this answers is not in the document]
meiravgri: This is an optimization used by ft.info command, to avoid calculating the overhead twice.
Since the numeric tree overhead is not exposed we don't need it.

GuyAv46 (approved):

src/numeric_index.h
```diff
@@ -137,7 +139,7 @@ void NumericRangeTree_Free(NumericRangeTree *t);
 
 extern RedisModuleType *NumericIndexType;
 
-NumericRangeTree *OpenNumericIndex(const RedisSearchCtx *ctx, RedisModuleString *keyName);
+NumericRangeTree *openNumericKeysDict(IndexSpec* spec, RedisModuleString *keyName, int write);
```
DvirDukhan, line 142:
```suggestion
NumericRangeTree *openNumericKeysDict(IndexSpec* spec, RedisModuleString *keyName, int readWriteMode);
```

tests/cpptests/index_utils.cpp
```diff
@@ -36,3 +48,80 @@ InvertedIndex *createIndex(int size, int idStep, int start_with) {
...
+        continue;
+    }
+    size_t curr_node_memory = NumericRangeGetMemory(currNode);
+
+    // Ensure stats are correct
+    if (curr_node_memory != currNode->range->invertedIndexSize) {
+        *failed_range = currNode;
+    }
+
+    total_tree_mem += curr_node_memory;
```
DvirDukhan, line 122: indent

tests/cpptests/index_utils.cpp
```diff
@@ -36,3 +48,80 @@ InvertedIndex *createIndex(int size, int idStep, int start_with) {
...
+
+    while ((currNode = NumericRangeTreeIterator_Next(Iterator))) {
+    if (!currNode->range) {
+        continue;
+    }
+    size_t curr_node_memory = NumericRangeGetMemory(currNode);
+
+    // Ensure stats are correct
+    if (curr_node_memory != currNode->range->invertedIndexSize) {
+        *failed_range = currNode;
```
DvirDukhan, line 119: always the last one?

tests/pytests/common.py
```diff
@@ -668,3 +668,35 @@ def verify_shard_init(shard):
...
     env.assertEqual(db_res, res, message=message)
+
+# fields should be in capital letters
+def getInvertedIndexInitialSize_MB(fields: list[str]) -> float:
+    total_size = 0
+    for field in fields:
+        if field in ['GEO', 'NUMERIC']:
+            block_size = 48
+            initial_block_cap = 6
+            inverted_index_meta_data = 48
```
DvirDukhan, line 679: can it be useful somewhere else, and avoid "magic numbers" in function?

tests/pytests/common.py
```diff
@@ -668,3 +668,35 @@ def verify_shard_init(shard):
...
+            total_size += (block_size + initial_block_cap + inverted_index_meta_data) / float(1024 * 1024)
+        elif field not in ['TEXT', 'TAG', 'GEOMETRY', 'VECTOR']:
+            raise ValueError(f'Field {field} is not supported')
+
+    return total_size
+
+def check_index_info(env, idx, exp_num_records, exp_inv_idx_size, msg=""):
```
DvirDukhan, line 686: why do we need this one if we have `compare_index_info_dict`?
maybe we can expend `compare_index_info_dict` with `keys_to_compare`?

tests/cpptests/test_cpp_range.cpp
```diff
@@ -218,6 +222,148 @@ TEST_F(RangeTest, testRangeIteratorMulti) {
...
+  }
+
+  // delete some docs
+  size_t deleted_docs = num_docs / 4;
+
+  // Add random numbers if needed
+  while (generated_numbers.size() < deleted_docs) {
+      size_t random_val = dis(gen);
+      generated_numbers.insert(random_val);
+  }
```
DvirDukhan, line 307: when it will happen?

tests/cpptests/test_cpp_range.cpp
```diff
@@ -218,6 +222,148 @@ TEST_F(RangeTest, testRangeIteratorMulti) {
...
+
+  // check memory
+  expected_mem = get_spec(index)->stats.invertedSize;
+  numeric_tree_mem = CalculateNumericInvertedIndexMemory(rt, &failed_range);
+  if (failed_range) {
+    print_failure();
+    FAIL();
+  }
+  ASSERT_EQ(rt->invertedIndexesSize, numeric_tree_mem);
+  ASSERT_EQ(rt->invertedIndexesSize, expected_mem);
```
DvirDukhan, line 328: do you want to establish a ground truth on empty index (similar to the python tests)?

tests/cpptests/test_cpp_forkgc.cpp
```diff
@@ -136,20 +109,81 @@ static InvertedIndex *getTagInvidx(RedisSearchCtx *sctx, const char *field,
...
+protected:
+  const char *numeric_field_name = "n";
+
+  void SetUp() override {
+    FGCTest::SetUp();
+    RediSearch_CreateNumericField(this->ism, numeric_field_name);
+  }
+};
+
+TEST_F(FGCTestNumeric, testNumeric) {
```
DvirDukhan, line 132: this one seems very similar to test_cpp_range.cpp::RangeIndexTest::testNumericTreeMemory
anything I miss?

tests/cpptests/test_cpp_range.cpp
```diff
@@ -218,6 +222,148 @@ TEST_F(RangeTest, testRangeIteratorMulti) {
...
+
+}
+
+TEST_F(RangeIndexTest, testNumericTreeOverhead) {
```
DvirDukhan, line 332: can/should we merge both tests?

DvirDukhan (commented):
Thanks
added comments

src/fork_gc.c
```diff
@@ -254,7 +257,8 @@ static bool FGC_childRepairInvidx(ForkGC *gc, RedisSearchCtx *sctx, InvertedInde
   if (array_len(blocklist) == idx->size) {
     // no empty block, there is no need to send the blocks array. Don't send
     // any new blocks.
-    FGC_sendBuffer(gc, NULL, 0);
+    size_t len = 0;
+    FGC_SEND_VAR(gc, len);
```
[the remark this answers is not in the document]
meiravgri: `FGC_sendBuffer` calls `FGC_sendFixed` that calls `write`. Calling it with NULL causes the compiler to raise `-Wnonnull` warning

src/fork_gc.c
```diff
@@ -959,8 +965,12 @@ static FGCError FGC_parentHandleNumeric(ForkGC *gc) {
 
     RedisSearchCtx_LockSpecWrite(sctx);
 
-    RedisModuleString *keyName = IndexSpec_GetFormattedKeyByName(sctx->spec, fieldName, INDEXFLD_T_NUMERIC);
-    rt = OpenNumericIndex(sctx, keyName);
+    if (!initialized) {
+      fs = IndexSpec_GetField(sctx->spec, fieldName, strlen(fieldName));
+      keyName = IndexSpec_GetFormattedKey(sctx->spec, fs, fs->types);
+      rt = openNumericKeysDict(sctx->spec, keyName, OPEN_INDEX_READ);
```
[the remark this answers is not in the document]
meiravgri: `openNumericKeysDict` only returns NULL when the tree was not initialized yet, which is not possible if we get here

tests/pytests/common.py
meiravgri: where is somewhere else? It's in a common file available for any test in this directory
And those are the structs size in our code, how can i get them other than hard coding?
I wrapped it in a function so it could be easily changed in the future if needed

tests/pytests/common.py
meiravgri: I didn't want to make major changes in tests that were not relevant to this pr.
I moved these functions here from the test files because I noticed they were duplicated in multiple files.

tests/cpptests/test_cpp_range.cpp
meiravgri: If a random number was generated twice during insertion

tests/cpptests/test_cpp_range.cpp
meiravgri: I don't get the question

tests/cpptests/test_cpp_forkgc.cpp
meiravgri: They are similar, but `testNumericTreeMemory` purpose is to test that the new member of the tree (`invertedIndexesSize`) correctly captures the sum of all its ranges memory, which is updated both in insertions and in deletions(gc), by summing all ranges in the tree and comparing the sum to `invertedIndexesSize`.
`testNumeric` tests that the gc correctly updates the memory of the ranges. It relies on the correctness of `invertedIndexesSize` so it can avoid iterating the tree in its validations

tests/cpptests/test_cpp_range.cpp
meiravgri: the gc test also checks the spec's inverted indexes stats, which is updated according to the ranges stats.

GuyAv46 (dismissed):

DvirDukhan (dismissed):

GuyAv46 (dismissed):

tests/pytests/common.py
```diff
@@ -668,3 +668,35 @@ def verify_shard_init(shard):
 def cmd_assert(env, cmd, res, message=None):
     db_res = env.cmd(*cmd)
     env.assertEqual(db_res, res, message=message)
+
+# fields should be in capital letters
+def getInvertedIndexInitialSize_MB(fields: list[str]) -> float:
```
GuyAv46, line 673: No need to fix it here as well?

GuyAv46 (dismissed):

GuyAv46 (approved):

meiravgri:
Won't backport to 2.6 as #4218, introducing improvements of inverted indexes memory tracking, essential for this pr, wasn't merged into 2.6

In addition, to avoid potential breaking changes, opening the numeric index for read won't be merged into 2.6 as well, since the mechanism of opening the numeric index is a bit different.

PR #5216, opened 2024-11-14
Improve query parsing - [MOD-7794, MOD-7764]

GuyAv46:

Reducing the cases where the query quietly returns 0 results with no explanation.

Including:
1. Validating field types while parsing the query
2. Validating the minimum value in a numeric range is smaller than the maximal value

Some fixes are only available with `DIALECT` > 1.

We also drop the support of the legacy `FILTER` and `GEOFILTER` options of `FT.SEARCH`

**Mark if applicable**


raz-mon:
@GuyAv46 Why are we dropping `FILTER` and `GEOFILTER`?

src/optimizer_reader.c
```diff
@@ -224,16 +224,16 @@ IndexIterator *NewOptimizerIterator(QOptimizer *qOpt, IndexIterator *root, Itera
   oi->numDocs = qOpt->sctx->spec->docs.size;
   oi->childEstimate = root->NumEstimated(root->ctx);
 
+  const FieldSpec *field = IndexSpec_GetField(qOpt->sctx->spec, qOpt->fieldName);
```
raz-mon, line 227: `NULL` case handling missing

raz-mon (commented):
Partial review (23/46)

src/query_parser/v2/parser.y
```diff
@@ -445,78 +443,35 @@ text_union(A) ::= text_union(B) OR text_expr(C). [OR] {
...
+    array_free(B);
+    A = NULL;
+  } else {
+    t_fieldMask mask = 0;
+    if (ctx->sctx->spec) {
+      for (int i = 0; i < array_len(B); i++) {
+        mask |= FIELD_BIT(B[i].fs);
+      }
     }
+    array_free(B);
```
raz-mon, line 471: Is this needed given the defined destructor?

src/query_parser/v2/parser.y
```diff
@@ -641,74 +594,72 @@ termlist(A) ::= param_term(B) param_term(C). [TERMLIST]  {
 }
 
 termlist(A) ::= termlist(B) param_term(C) . [TERMLIST] {
-    A = B;
-    if (!(C.type == QT_TERM && StopWordList_Contains(ctx->opts->stopwords, C.s, C.len))) {
```
raz-mon, line 645: Where did the stop-word check go?

src/query_parser/v2/parser.y
```diff
@@ -278,16 +281,11 @@ static inline char *toksep2(char **s, size_t *tokLen) {
   });
 }
 
-%type modifierlist { Vector* }
+%type modifier { FieldName }
```
raz-mon, line 284: Is this the largest field in the union? If so, consider changing to `FieldName *` and heap allocating.
See [link](https://sqlite.org/src/doc/trunk/doc/lemon.html#ptype:~:text=But%20the%20grammar%20designer,just%20need%20to%20know) for motivation.

src/query_parser/v2/parser.y
```diff
@@ -733,67 +684,91 @@ text_expr(A) ::= PERCENT PERCENT PERCENT param_term(B) PERCENT PERCENT PERCENT.
...
-    A = NewVector(char *, 2);
-    char *s = rm_strndup(B.s, B.len);
-    Vector_Push(A, s);
-    s = rm_strndup(C.s, C.len);
-    Vector_Push(A, s);
+  if (ctx->sctx->spec) {
+    if (!FIELD_IS(B.fs, INDEXFLD_T_FULLTEXT)) {
+      REPORT_WRONG_FIELD_TYPE(B, SPEC_TEXT_STR);
+      A = NULL;
+    } else {
```
raz-mon, line 701: We couldn't really search on tags with this modifier list before right?
If so, great! (breaking, but great!)

src/query_parser/v2/parser.y
```diff
@@ -733,67 +684,91 @@ text_expr(A) ::= PERCENT PERCENT PERCENT param_term(B) PERCENT PERCENT PERCENT.
...
+      A = B;
+      array_append(A, second);
     }
-    A = NULL;
+  } else {
+    A = B;
+    FieldName second = { .tok = C };
+    array_append(A, second);
```
raz-mon, line 742: When do we get here?

src/query_parser/v1/parser.y
```diff
@@ -486,16 +486,23 @@ expr(A) ::= modifier(B) COLON tag_list(C) . {
...
-
-        A = NewTagNode(s, slen);
+        A = NewTagNode(NULL);
         QueryNode_AddChildren(A, C->children, QueryNode_NumChildren(C));
 
         // Set the children count on C to 0 so they won't get recursively free'd
         QueryNode_ClearChildren(C, 0);
         QueryNode_Free(C);
+
+        if (ctx->sctx->spec) {
```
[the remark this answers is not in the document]
GuyAv46: Not sure what was the original scenario, but today we parse the query in the coordinator in some cases to prepare for specific query types (like `KNN`). When parsing from the coordinator we don't have a spec context, but we also don't use the parsed `AST` to generate actual iterators.
So:
- If we don't have a spec context, we expect the caller won't use the `AST` to iterate over a spec, and we don't check for the field type nor existence (we can't) and just set the `FieldSpec` to `NULL`.
- If we have a spec context, we validate that the requested fields exist and match the sub-query type

src/query_node.h
```diff
@@ -115,7 +117,7 @@ typedef struct {
 } QueryNumericNode;
 
 typedef struct {
-  const struct GeoFilter *gf;
+  struct GeoFilter *gf;
```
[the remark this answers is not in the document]
GuyAv46: maybe only after casting the `const` away?
Anyway this was the only type to be const, now we are aligned

src/query_param.c
```diff
@@ -55,6 +55,7 @@ QueryParam *NewNumericFilterQueryParam_WithParams(struct QueryParseCtx *q, Query
 }
 
 void QueryParam_Free(QueryParam *p) {
+  if (!p) return;
```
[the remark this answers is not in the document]
GuyAv46: Both :)
I started calling it after some checks without checking if it was `NULL` for a simple and unified flow, but I also think that in some cases it could've happened.
Maybe I should try to cause it and see if it's an actual bug

src/query_param.c
```diff
@@ -213,9 +214,8 @@ int QueryParam_Resolve(Param *param, dict *params, QueryError *status) {
     case PARAM_NUMERIC_MIN_RANGE:
     case PARAM_NUMERIC_MAX_RANGE:
     {
-      int inclusive = 1;
       int isMin = param->type == PARAM_NUMERIC_MIN_RANGE ? 1 : 0;
-      if (parseDoubleRange(val, &inclusive, (double*)param->target, isMin, param->sign, status) == REDISMODULE_OK)
+      if (parseDoubleRange(val, (double*)param->target, isMin, param->sign, status) == REDISMODULE_OK)
```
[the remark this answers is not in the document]
GuyAv46: Correct, not with this function. Only here and it was ignored.

tests/cpptests/test_cpp_query.cpp
```diff
@@ -108,11 +108,11 @@ TEST_F(QueryTest, testParser_delta) {
...
   // difference between `expr` and `text_expr` were introduced in version 2
-  assertValidQuery_v(1, "@title:@num:[0 10]");
-  assertValidQuery_v(1, "@title:(@num:[0 10])");
+  assertValidQuery_v(1, "@title:@bar:[0 10]");
+  assertValidQuery_v(1, "@title:(@bar:[0 10])");
   assertValidQuery_v(1, "@t1:@t2:@t3:hello");
-  assertInvalidQuery_v(2, "@title:@num:[0 10]");
-  assertInvalidQuery_v(2, "@title:(@num:[0 10])");
+  assertInvalidQuery_v(2, "@title:@bar:[0 10]");
+  assertInvalidQuery_v(2, "@title:(@bar:[0 10])");
```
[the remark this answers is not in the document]
GuyAv46: So tedious...

src/query_parser/v2/parser.y
GuyAv46: Yes. See the [destructor](https://sqlite.org/src/doc/trunk/doc/lemon.html#destructor) section. Mainly:

> It is important to note that the value of a non-terminal is passed to the destructor whenever the non-terminal is removed from the stack, unless the non-terminal is used in a C-code action. If the non-terminal is used by C-code, then it is assumed that the C-code will take care of destroying it. More commonly, the value is used to build some larger structure, and we don't want to destroy it, which is why the destructor is not called in this circumstance.

src/query_parser/v2/parser.y
GuyAv46: This might be an oversigt, but we only check this here (`termlist(A) ::= termlist(B) param_term(C)`) but not in the initial rule of the term list (`termlist(A) ::= param_term(B) param_term(C)`).
Do we need to filter stop-words in a tag search? If so, we should check it for the first two terms as well. I was under the assumption we don't need to filter stop-words for tag searches

src/query_parser/v2/parser.y
GuyAv46: I deduced from the rule (singular) that uses the `modifierlist` type that we only allow it for `TEXT` fields. Were you able to run a `@t1|t2:{tag}` query with dialect > 1?

src/query_parser/v2/parser.y
GuyAv46: Only possible when we don't have a spec context, which should only happen if we parse the query from the coordinator (not for the iterator's purpose)

src/optimizer_reader.c
GuyAv46: It was missing before, so we have the same behavior.
I assume we make the validation before, so now we know it will succeed.
We might need to investigate this further

src/query_parser/v2/parser.y
GuyAv46: This is OK today, as this type is `QueryToken` + a pointer, and we already use larger structs in the union

src/query_param.c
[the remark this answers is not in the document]
GuyAv46: Verified. This couldn't happen until now

src/query_parser/v2/parser.y
GuyAv46: OK, this is a bug. We should validate that the term is not a stop word, and we miss this filtering in the base rule.
I'll revert this change, and we should fix the bug in another PR.

raz-mon (approved):

alonre24 (commented):
Woohoo

tests/pytests/test.py
```diff
@@ -3657,8 +3626,10 @@ def testSchemaWithAs(env):
   conn.execute_command('HSET', 'b', 'foo', 'world')
 
   for _ in env.reloadingIterator():
-    env.expect('ft.search idx @txt:hello').equal([0])
-    env.expect('ft.search idx @txt:world').equal([0])
+    env.expect('ft.search idx @txt:hello DIALECT 1').equal([0]).noError()
+    env.expect('ft.search idx @txt:hello DIALECT 2').error().contains('Unknown field')
+    env.expect('ft.search idx @txt:world DIALECT 1').equal([0]).noError()
+    env.expect('ft.search idx @txt:world DIALECT 2').error().contains('Unknown field')
```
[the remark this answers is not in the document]
GuyAv46: Opened MOD-8145

alonre24 (approved):

PR #5223, opened 2024-11-18
MOD-8062: Hide internal commands in new ACL category for RediSearch internal commands

raz-mon:
In this PR we introduce the new `_search_internal` ACL command category, to which we register all of our internal commands, i.e., all the `_FT.*` commands. We register these commands ONLY to this command category, such that only the coordinator will be eligible to execute them via the permissions given to its user-defined ACL user.
A correct configuration of the ACL permissions, in which all other users are disallowed to execute RediSearch internal commands by appending the `-@_search_internal` command category to their command permissions.
Notice: These commands change between our OSS and enterprise builds. On enterprise this command category should not be used.

**Mark if applicable**

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -988,10 +988,12 @@ int RMCreateSearchCommand(RedisModuleCtx *ctx, const char *name,
 
   int rc = REDISMODULE_OK;
   char *categories;
-  if (!strcmp(aclCategories, "")) {
-    categories = SEARCH_ACL_CATEGORY;
+
+  if (strncmp(name, "_", 1) == 0) {
+    // Internal command, register to the internal ACL category
+    rm_asprintf(&categories, strcmp(aclCategories, "") != 0 ? "%s %s" : "%.0s%s", aclCategories, SEARCH_ACL_INTERNAL_CATEGORY);
   } else {
-    rm_asprintf(&categories, "%s %s", aclCategories, SEARCH_ACL_CATEGORY);
+    rm_asprintf(&categories, strcmp(aclCategories, "") != 0 ? "%s %s" : "%.0s%s", aclCategories, SEARCH_ACL_CATEGORY);
   }
 
   if (RedisModule_SetCommandACLCategories(command, categories) == REDISMODULE_ERR) {
@@ -999,9 +1001,7 @@ int RMCreateSearchCommand(RedisModuleCtx *ctx, const char *name,
     rc = REDISMODULE_ERR;
   }
 
-  if (strlen(categories) != strlen(SEARCH_ACL_CATEGORY)) {
-    rm_free(categories);
-  }
+  rm_free(categories);
 
   return rc;
 }
@@ -1066,6 +1066,12 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx, RedisModuleString **argv,
   // Create the `search` ACL command category
   if (RedisModule_AddACLCategory(ctx, SEARCH_ACL_CATEGORY) == REDISMODULE_ERR) {
       RedisModule_Log(ctx, "warning", "Could not add " SEARCH_ACL_CATEGORY " ACL category, errno: %d\n", errno);
+      return REDISMODULE_ERR;
+  }
+
+  // Create the `search_internal` ACL command category
+  if (RedisModule_AddACLCategory(ctx, SEARCH_ACL_INTERNAL_CATEGORY) == REDISMODULE_ERR) {
+      RedisModule_Log(ctx, "warning", "Could not add " SEARCH_ACL_INTERNAL_CATEGORY " ACL category, errno: %d\n", errno);
       return REDISMODULE_ERR;
   }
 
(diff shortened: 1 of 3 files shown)


src/module.h
```diff
@@ -57,6 +57,7 @@ do {                                            \
   RedisModule_ReplyWithStringBuffer(ctx, literal, sizeof(literal) - 1)
 
 #define SEARCH_ACL_CATEGORY "search"
+#define SEARCH_ACL_INTERNAL_CATEGORY "search_internal"
```
oshadmi, line 60: Can we name the internal category with a leading underscore? (as internal commands)

src/module.c
```diff
@@ -988,20 +988,20 @@ int RMCreateSearchCommand(RedisModuleCtx *ctx, const char *name,
 
   int rc = REDISMODULE_OK;
   char *categories;
-  if (!strcmp(aclCategories, "")) {
-    categories = SEARCH_ACL_CATEGORY;
+
+  if (strncmp(name, "_", 1) == 0) {
```
oshadmi, line 992: Consider a function/macro to detect internal commands

src/module.c
```diff
@@ -988,20 +988,20 @@ int RMCreateSearchCommand(RedisModuleCtx *ctx, const char *name,
...
+    // Internal command, register to the internal ACL category
+    rm_asprintf(&categories, strcmp(aclCategories, "") != 0 ? "%s %s" : "%.0s%s", aclCategories, SEARCH_ACL_INTERNAL_CATEGORY);
```
oshadmi, line 994: Should this be done only with oss cluster?

src/module.h
raz-mon: Yes that makes sense 👌

src/module.c
raz-mon: `strncmp` is the detection function 🤔 
A macro is possible but not sure I see a good reason to use one here

src/module.c
```diff
@@ -1069,6 +1069,12 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx, RedisModuleString **argv,
       return REDISMODULE_ERR;
   }
 
+  // Create the `search_internal` ACL command category
+  if (RedisModule_AddACLCategory(ctx, SEARCH_ACL_INTERNAL_CATEGORY) == REDISMODULE_ERR) {
+      RedisModule_Log(ctx, "warning", "Could not add " SEARCH_ACL_INTERNAL_CATEGORY " ACL category, errno: %d\n", errno);
```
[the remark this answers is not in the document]
raz-mon: Both.
We log this warning, but also fail loading the module (error)

tests/pytests/test_acl.py
```diff
@@ -8,88 +8,114 @@ def test_acl_category(env):
...
+        # Let's add `write` permissions to `test`
+        conn.execute_command('AUTH', 'default', '')
+        conn.execute_command('ACL', 'SETUSER', 'test', '+@write')
+        conn.execute_command('AUTH', 'test', '123')
+
+        # `test` should now be able to run `write` commands
+        for cmd in WRITE_SEARCH_COMMANDS:
+            try:
+                res = conn.execute_command(cmd)
+                env.assertTrue(False)
```
oshadmi, line 105: If should be able to run `write` commands - should reach here? (without an exception)

oshadmi: Oh, there's no `search` category yet?

tests/pytests/test_acl.py
```diff
@@ -8,88 +8,114 @@ def test_acl_category(env):
...
+            except Exception as e:
+                env.assertNotContains("User test has no permissions", str(e))
+
+        # Add `test` search permissions
+        conn.execute_command('AUTH', 'default', '')
+        conn.execute_command('ACL', 'SETUSER', 'test', '+@search', '+@_search_internal')
```
oshadmi, line 111: What about failing to run internal commands without `_search_internal` category?

tests/pytests/test_acl.py
raz-mon: No, it fails due to an imcomplete command, non-related to permissions.
Added a comment for that.

src/module.c
```diff
@@ -988,18 +989,22 @@ int RMCreateSearchCommand(RedisModuleCtx *ctx, const char *name,
 
   int rc = REDISMODULE_OK;
   char *categories;
-  if (!strcmp(aclCategories, "")) {
-    categories = SEARCH_ACL_CATEGORY;
+  bool is_internal = false;
+
+  if (IS_INTERNAL_COMMAND(name)) {
+    // Internal command, register to the internal ACL category ONLY
+    categories = SEARCH_ACL_INTERNAL_CATEGORY;
```
oshadmi, line 996: Need further discussion - whether _internal_search category is always separated from other categories or should we also register to additional categories?
* If only _internal_search - it needs to be removed explicitly, (usually admin can define `+@all -@slow` etc. which would not remove _internal_search
* If _internal_search is combined with others, `+@slow` could give users permission to search internal commands without intending to

@MeirShpilraien @LiorKogan

MeirShpilraien (approved):

oshadmi (approved):

PR #5229, opened 2024-11-19
[MOD-8125] Create inverted index in write operations only

meiravgri:
This PR ensures we create an inverted index only in write operations (when possible. see exceptions below)
## Read operations
- ft.info
- INFO
- queries
- GC

####  `FT.DEBUG SPEC_INVIDXES_INFO INDEX_NAME`
Returnes
- `inverted_indexes_dict_size`
- `inverted_indexes_memory`

## Exceptions
Some read operations still create the index to avoid breaking changes, or where a fundamental change was required:
### debug commands
- `VecsimInfo` and `dumpHNSWData` we need an index to create an info iterator.
- `DumpGeometryIndex` calls abstract `api->dump(idx, ctx)`. To allow avoiding from creating the index, we need a dump function that can imitate an empty index for `idx == NULL`.
### Queries
- `Query_EvalGeometryNode`: since the query string is validated **after** we are trying to open the geometry index, in case of a wrong query string and uninitialized index, we will return results as if **the index was empty, instead of raising an error.** However, If we create the index (as we are doing today), we can proceed to query validation. 
So until we change the order of verification, we open the index with `CREATE` option.


src/vector_index.c
```diff
@@ -11,13 +11,14 @@
 #include "rdb.h"
 #include "util/workers_pool.h"
 #include "util/threadpool_api.h"
+#include "redis_index.h"
 
-static VecSimIndex *openVectorKeysDict(IndexSpec *spec, RedisModuleString *keyName, int write) {
+VecSimIndex *openVectorKeysDict(IndexSpec *spec, RedisModuleString *keyName, bool create_if_index) {
```
GuyAv46, line 16:
```suggestion
VecSimIndex *OpenVectorIndex(IndexSpec *spec, RedisModuleString *keyName, bool create_if_index) {
```

tests/pytests/common.py
```diff
@@ -670,31 +670,52 @@ def cmd_assert(env, cmd, res, message=None):
...
+        try:
+            res = float(d2[key]) == float(value)
+            if _assert:
+                env.assertTrue(res, message=msg + " value is different in key: " + key)
+            else:
+                if res == False:
+                    return False
+        except KeyError:
+            if _assert:
+                env.assertTrue(False, message=msg + f" key {key} exists in {d1_name} but doesn't exist in {d2_name}")
```
GuyAv46, line 711:
```suggestion
                env.assertContains(key, d2, message=msg + f" key {key} exists in {d1_name} but doesn't exist in {d2_name}")
```

tests/pytests/test_common.py
```diff
@@ -39,3 +39,21 @@ def test_compare_lists_with_delta(env):
...
+    env.assertTrue(compare_numeric_dicts(env, d1, d2, _assert=False), message="comparing identical dicts")
+
+    # compare dicts with different values
+    d2["fish"] = "0.2"
+    env.assertFalse(compare_numeric_dicts(env, d1, d2, _assert=False), message="comparing dicts with different values")
+    # compare dicts with different keys
+    d3 = d1.copy()
+    d1["rabbit"] = 5
+    try:
+        compare_numeric_dicts(env, d1, d3, _assert=False)
```
GuyAv46, line 56: Is `_assert=False` causing an exception aligned with the `RLTest` behavior?

tests/pytests/common.py
```diff
@@ -670,31 +670,52 @@ def cmd_assert(env, cmd, res, message=None):
...
 
+# Iterates items in d1 and compare their keys[value] with d2
+# asserts when a key is missing in d2
+# For simplicity, all values are compared as floats
+def compare_numeric_dicts(env, d1, d2, d1_name="d1", d2_name="d2", msg="", _assert=True):
+    for key, value in d1.items():
+        try:
+            res = float(d2[key]) == float(value)
+            if _assert:
+                env.assertTrue(res, message=msg + " value is different in key: " + key)
```
GuyAv46, line 705:
```suggestion
                env.assertTrue(res, message=msg + " value is different in key: " + key, depth=depth+1)
```

tests/pytests/common.py
GuyAv46, line 711:
```suggestion
                env.assertTrue(False, message=msg + f" key {key} exists in {d1_name} but doesn't exist in {d2_name}", depth=depth+1)
```

tests/pytests/common.py
```diff
@@ -670,31 +670,52 @@ def cmd_assert(env, cmd, res, message=None):
...
     d = index_info(env, idx)
     env.assertEqual(float(d['num_records']), exp_num_records, message=msg)
 
     if(exp_inv_idx_size != None):
         env.assertEqual(float(d['inverted_sz_mb']), exp_inv_idx_size, message=msg)
 
+# Iterates items in d1 and compare their keys[value] with d2
+# asserts when a key is missing in d2
+# For simplicity, all values are compared as floats
+def compare_numeric_dicts(env, d1, d2, d1_name="d1", d2_name="d2", msg="", _assert=True):
```
GuyAv46, line 700:
```suggestion
def compare_numeric_dicts(env, d1, d2, d1_name="d1", d2_name="d2", msg="", _assert=True, depth=0):
```

tests/pytests/common.py
```diff
@@ -670,31 +670,52 @@ def cmd_assert(env, cmd, res, message=None):
...
+            else:
+                raise KeyError
+    return True
+
 def compare_index_info_dict(env, idx, expected_info_dict, msg=""):
     d = index_info(env, idx)
-    for key, value in expected_info_dict.items():
-        env.assertEqual(float(d[key]), value, message=msg + " failed for info key: " + key)
-
+    compare_numeric_dicts(env, expected_info_dict, d, "expected_info_dict", "index_info", msg)
```
GuyAv46, line 718:
```suggestion
    compare_numeric_dicts(env, expected_info_dict, d, "expected_info_dict", "index_info", msg, depth=1)
```

tests/pytests/test_index.py
```diff
@@ -0,0 +1,107 @@
...
+        "vec", "VECTOR",
+            "FLAT", "6", "TYPE", "FLOAT32", "DIM", "2", "DISTANCE_METRIC", "L2",
+    ).ok()
+
+    # Sanity check - empty spec
+    check_dict_size(env, "after creation")
+
+    # call ft.info
+    env.cmd("FT.INFO", "idx")
+    check_dict_size(env, "calling ft.info")
```
alonre24, line 31: What about testing `info everything`?

src/spec.c
```diff
@@ -1022,7 +1025,10 @@ VectorIndexStats IndexSpec_GetVectorIndexStats(IndexSpec *sp) {
     const FieldSpec *fs = sp->fields + i;
     if (FIELD_IS(fs, INDEXFLD_T_VECTOR)) {
       RedisModuleString *vecsim_name = IndexSpec_GetFormattedKey(sp, fs, INDEXFLD_T_VECTOR);
-      VecSimIndex *vecsim = OpenVectorIndex(sp, vecsim_name);
+      VecSimIndex *vecsim = openVectorKeysDict(sp, vecsim_name, DONT_CREATE_INDEX);
+      if (!vecsim) {
+        continue;
```
alonre24, line 1030: :muscle:

alonre24 (commented):
Good catch! So except for vector and geometry, were there any other places where we created an index unnecessarily (except for the tag in the GC)?

src/vector_index.c
meiravgri: ok

tests/pytests/test_common.py
meiravgri: It is used by the function as a hint to avoid assertions (to test expected failures)

src/debug_commands.c
```diff
@@ -450,9 +467,11 @@ DEBUG_COMMAND(DumpTagIndex) {
     RedisModule_ReplyWithError(sctx->redisCtx, "Could not find given field in index spec");
     goto end;
   }
-  TagIndex *tagIndex = TagIndex_Open(sctx, keyName, false);
+  const TagIndex *tagIndex = TagIndex_Open(sctx, keyName, DONT_CREATE_INDEX);
+
+  // Field was not initialized yet
   if (!tagIndex) {
-    RedisModule_ReplyWithError(sctx->redisCtx, "can not open tag field");
+    RedisModule_ReplyWithEmptyArray(sctx->redisCtx);
```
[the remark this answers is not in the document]
meiravgri: we are, 2 lines above

tests/pytests/test_index.py
```diff
@@ -0,0 +1,107 @@
+from common import *
+
+def check_dict_size(env, msg, inverted_indexes_dict_size = 0, inverted_indexes_memory = 0):
```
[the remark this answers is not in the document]
meiravgri: The debug command returns data on the inverted indexes in the spec, not on a specific one

src/tag_index.c
```diff
@@ -285,12 +285,12 @@ RedisModuleString *TagIndex_FormatName(RedisSearchCtx *sctx, const char *field)
 }
 
 /* Open the tag index */
-TagIndex *TagIndex_Open(const RedisSearchCtx *ctx, RedisModuleString *key, int openWrite) {
+TagIndex *TagIndex_Open(const RedisSearchCtx *ctx, RedisModuleString *key, bool create_if_index) {
```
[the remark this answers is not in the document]
meiravgri: going with create_if_missing to align with other field types

tests/pytests/test_index.py
meiravgri: will add

src/fork_gc.c
```diff
@@ -879,7 +879,7 @@ static FGCError FGC_parentHandleTerms(ForkGC *gc) {
 
   RedisSearchCtx_LockSpecWrite(sctx);
 
-  InvertedIndex *idx = Redis_OpenInvertedIndex(sctx, term, len, 1, NULL);
+  InvertedIndex *idx = Redis_OpenInvertedIndex(sctx, term, len, DONT_CREATE_INDEX, NULL);
```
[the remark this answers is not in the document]
meiravgri: No. But if we got here the term existed before we forked
Regarding gc - don't we have unit tests for text?

tests/pytests/common.py
meiravgri: assertContains doesn't support message :(

GuyAv46: We should fix it

tests/pytests/test_common.py
```diff
@@ -39,3 +39,21 @@ def test_compare_lists_with_delta(env):
...
+        env.assertTrue(False, message="Comparing dicts with different keys")
+    except KeyError:
+        env.assertTrue(True)
```
GuyAv46, line 59: We can also just pass…

src/query.c
```diff
@@ -895,7 +895,11 @@ static IndexIterator *Query_EvalGeometryNode(QueryEvalCtx *q, QueryNode *node) {
   RS_LOG_ASSERT(node->type == QN_GEOMETRY, "query node type should be geometry");
 
   const FieldSpec *fs = node->gmn.geomq->fs;
-  const GeometryIndex *index = OpenGeometryIndex(q->sctx->spec, fs);
+
+  // TODO: open with DONT_CREATE_INDEX once the query string is validated before we get here.
+  // Currently, if  we use DONT_CREATE_INDEX, and the index was not initalized yet, and the query is invalid,
+  // we return results as if the index was empty, instead of raising an error.
```
GuyAv46, line 901: What should be done here? Parse the query before getting here? Fix something in our parser? Can we do something in this PR?

GuyAv46 (dismissed):

src/query.c
meiravgri: Yes. not in the scope of this PR
I'm opening a ticket

GuyAv46 (approved):

alonre24 (approved):

meiravgri:
won't be cherry-picked to 2.6 due to discrepancy in open index functions signature

PR #5239, opened 2024-11-21
MOD-8054: Validate index prefixes

raz-mon:
This PR adds the validation of index prefixes for the `FT.AGGREGATE`, `FT.SEARCH`, `FT.PROFILE` commands between the coordinator and shards.
This fixes a potential in which the shard-index may change to a different index with the same name while a query while already dispatched from the coordinator is awaiting the shard execution. This situation may cause an ACL breach, or just yield irrelevant results.

**Mark if applicable**


src/coord/dist_aggregate.c
```diff
@@ -538,6 +540,17 @@ static void buildMRCommand(RedisModuleString **argv, int argc, int profileArgs,
...
   array_append(tmparr, "_NUM_SSTRING");
 
+  // Add the index prefixes to the command, for validation in the shard
+  array_append(tmparr, "_INDEX_PREFIXES");
+  arrayof(sds) prefixes = sp->rule->prefixes;
+  char *n_prefixes;
+  rm_asprintf(&n_prefixes, "%u", array_len(prefixes));
+  array_append(tmparr, n_prefixes);
+  for (uint i = 0; i < array_len(prefixes); i++) {
+    // MRCommand_Append(xcmd, prefixes[i], sdslen(prefixes[i]));
```
oshadmi, line 550: Remove line?

tests/pytests/common.py
```diff
@@ -657,13 +657,8 @@ def get_passphrase():
...
 def verify_shard_init(shard):
     with TimeLimit(5, 'Failed to verify shard initialization'):
-        while True:
-            try:
-                shard.execute_command('FT.SEARCH', 'non-existing', '*')
-                raise Exception('Expected FT.SEARCH to fail')
-            except redis_exceptions.ResponseError as e:
-                if 'no such index' in str(e):
-                    break
+        shard.execute_command('FT.CREATE', 'idx' ,'SCHEMA', 't', 'TEXT')
```
GuyAv46, line 660: Consider giving it a less common name?
Also, doesn't it fail before the shard is initialized with "cannot distribute command"? That's why it was wrapped with a `try..except` before

tests/pytests/test.py
```diff
@@ -1230,7 +1230,7 @@ def testInKeys(env):
     with env.assertResponseError():
         env.cmd('ft.search', 'idx', 'hello', 'INKEYS', -1)
     with env.assertResponseError():
-        env.cmd('ft.search', 'idx', 'hello', 'inkeys', 4, 'foo')
+        env.cmd('ft.search', 'idx', 'hello', 'inkeys', 5, 'foo')
```
GuyAv46, line 1233: why is this change needed?

tests/pytests/test.py
```diff
@@ -3822,7 +3822,7 @@ def testUsesCounter(env):
     env.cmd('ft.info', 'idx')
     env.cmd('ft.search', 'idx', '*')
 
-    assertInfoField(env, 'idx', 'number_of_uses', 3)
+    assertInfoField(env, 'idx', 'number_of_uses', 3 if not env.isCluster() else 4)
```
GuyAv46, line 3825: because of the coordinator's usage?

tests/pytests/test.py
```diff
@@ -4305,3 +4305,53 @@ def test_notIterTimeout(env):
...
     ).error().contains('Timeout limit was reached')
+
+@skip(cluster=False, min_shards=2)
+def test_incompatibleIndex(env):
+    """Tests that we get an error if we try to query an index with a different
+    schema than the one used in the query"""
+
+    # Connect to two shards
+    first_conn = env.getConnection(0)
+    second_conn = env.getConnection()
```
GuyAv46, line 4316: isn't that the same shard? can we do
```suggestion
    first_conn = env.getConnection(0)
    second_conn = env.getConnection(1)
```
to be more explicit?

tests/pytests/test_search_params.py
```diff
@@ -58,8 +58,6 @@ def test_param_errors(env):
     env.assertEqual(conn.execute_command('HSET', 'key1', 'foo', 'PARAMS', 'bar', 'PARAMS'), 2)
 
     # Test errors in PARAMS definition: duplicated param, missing param value, wrong count
-    env.expect('FT.SEARCH', 'idx', '*', 'NOCONTENT', 'PARAMS', '4', 'param1', 'val1').error().contains('Bad arguments for PARAMS: Expected an argument')
-    env.expect('FT.SEARCH', 'idx', '*', 'NOCONTENT', 'PARAMS', '2', 'param1').error().contains('Bad arguments for PARAMS: Expected an argument')
```
GuyAv46, line 62: why is it removed?

GuyAv46 (commented):
Very nice!
A few comments and suggestions

src/aggregate/aggregate_request.c
```diff
@@ -329,6 +329,13 @@ static int handleCommonArgs(AREQ *req, ArgsCursor *ac, QueryError *status, int a
     if (parseValueFormat(&req->reqflags, ac, status) != REDISMODULE_OK) {
       return ARG_ERROR;
     }
+  } else if (AC_AdvanceIfMatch(ac, "_INDEX_PREFIXES")) {
+    // Set the offset of the prefixes in the query, for further processing later
+    req->prefixesOffset = ac->offset - 1;
+
+    // Advance by the number of prefixes + 1 (for the n_prefixes argument)
+    long advance_by = strtol((const char *)AC_CURRENT(ac), NULL, 10);
+    AC_AdvanceBy(ac, advance_by + 1);
```
oshadmi, line 338: Since we are constructing the command, we know the args exist? (and no need to check return code)

tests/pytests/common.py
```diff
@@ -657,13 +657,8 @@ def get_passphrase():
 # Use FT.* command to make sure that the module is loaded and initialized
```
oshadmi, line 657: Update comment?

tests/pytests/test.py
oshadmi, line 3825: Consider avoiding this

tests/pytests/test_search_params.py
oshadmi, line 62: No need to keep this coverage?

src/aggregate/aggregate_request.c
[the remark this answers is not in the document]
raz-mon: Shouldn't happen, can add an assertion for that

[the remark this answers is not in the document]
raz-mon: This will basically lead to an early error. I'll try having this error be the one we want it to be, although stuff here are a bit intertwained

src/coord/dist_aggregate.c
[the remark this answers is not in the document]
raz-mon: Yes

src/coord/dist_aggregate.c
```diff
@@ -747,13 +769,17 @@ void RSExecDistAggregate(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
...
+  StrongRef_Release(strong_ref);
   RedisModule_EndReply(reply);
   return;
 
 // See if we can distribute the plan...
 err:
   assert(QueryError_HasError(&status));
   QueryError_ReplyAndClear(ctx, &status);
+  WeakRef_Release(ConcurrentCmdCtx_GetWeakRef(cmdCtx));
+  StrongRef_Release(strong_ref);
```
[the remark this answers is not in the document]
raz-mon: Yup, added 👍

tests/pytests/test.py
raz-mon: We have a not-ideal validation mechanism, that does not know to distinguish between the user and coordinator arguments in a shard command.
That yields this weird situation that we have today, and should be solved by performing the validations once in the coordinator I think.
Till then, we have a mixed state and this is just a workaround for that

tests/pytests/test.py
raz-mon: Yes, I think I'll make the cluster reference invisible to this.. Although then we can miss some references. Still seems clearer.

tests/pytests/test_search_params.py
raz-mon: Validations that depend on argument count on the shard are not all good currently due to the arguments we add in the coordinator. This as well will be solved when we do the validations on the coordinator.

src/aggregate/aggregate_request.c
raz-mon: Yes

tests/pytests/test.py
raz-mon: Yeah.. I can do that.

tests/pytests/test_search_params.py
raz-mon: I'll bring it back for non-cluster case at this point

raz-mon: Added the non-cluster case 👍

tests/pytests/test.py
raz-mon: This is 100% what is was supposed to be 🤔

tests/pytests/common.py
raz-mon: It's called before any indexes are created, but switched just in case.
Regarding the failure - doesn't seem to fail. Isn't the idea here to make sure that you don't get that error and the cluster is initialized appropriately?
The `try..except` before was catching the error that you expected upon indexing a non-existing index.

oshadmi (approved):

tests/pytests/test.py
GuyAv46: I didn't understand. If we search for `'ft.search', 'idx', 'hello', 'inkeys', 4, 'foo'` we don't fail?

src/aggregate/aggregate_request.c
```diff
@@ -334,8 +334,13 @@ static int handleCommonArgs(AREQ *req, ArgsCursor *ac, QueryError *status, int a
     req->prefixesOffset = ac->offset - 1;
 
     // Advance by the number of prefixes + 1 (for the n_prefixes argument)
-    long advance_by = strtol((const char *)AC_CURRENT(ac), NULL, 10);
-    AC_AdvanceBy(ac, advance_by + 1);
+    uint32_t advance_by;
+    if (AC_GetUnsigned(ac, &advance_by, AC_F_GE1) != AC_OK) {
```
GuyAv46, line 338: why not call `AC_GetVarArgs` to both advance and validate the number + that we have enough arguments to advance by?

src/coord/dist_aggregate.c
```diff
@@ -676,6 +690,14 @@ void RSExecDistAggregate(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
   QueryError status = {0};
   specialCaseCtx *knnCtx = NULL;
 
+  // Check if the index still exists, and promote the ref accordingly
+  StrongRef strong_ref = WeakRef_Promote(ConcurrentCmdCtx_GetWeakRef(cmdCtx));
+  IndexSpec *sp = StrongRef_Get(strong_ref);
+  if (!sp) {
+    QueryError_SetError(&status, QUERY_EDROPPEDBACKGROUND, NULL);
```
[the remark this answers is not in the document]
GuyAv46: wasn't resolved

tests/pytests/test.py
raz-mon: No. Since we add arguments to the shard command, from the coordinator.

raz-mon: I'll add it sooner so we won't have this issue on `FT.SEARCH`.

src/aggregate/aggregate_request.c
raz-mon: I don't need the copying mechanism to another ArgsCursor.
Added another assertion, as this case should never happen.

tests/pytests/test_search_params.py
raz-mon: Returned since I moved the new arguments closer to the query

src/aggregate/aggregate_request.c
raz-mon: I don't need the copying mechanism, and I also want to have assertions for this, as this should never happen

tests/pytests/test.py
GuyAv46: This sounds like a bug, no? What do we get if not an error? Does it consume the prefixes as the in-keys? Shouldn’t we fail in this case?

tests/pytests/common.py
GuyAv46: there is also the `ERRCLUSTER Uninitialized cluster state, could not perform command` that may be returned if we call a search command before any `SEARCH.CLUSTERSET` or the periodic `SEARCH.CLUSTERREFRESH` commands were executed (and we don't know the topology yet)

GuyAv46 (approved):

tests/pytests/common.py
raz-mon: Added the wrapping `try..except` to catch this flakiness

GuyAv46 (approved):
💯

PR #4858, opened 2024-07-11
MOD-7329: Determine if fields and documents are expired before returning them as valid results

kei-nan:
1. Add an internal ttl table to contain document and field expiration information
    - there will be a thread local timestamp that we check the expiration times against
    - it will be updated when a batch of results is obtained from the pipeline
2. In our field related iterators check if a field is expired before returning the document as a possible result
    - In query and hybrid iterators it required adding a loop, we might need to consider handling timeout 
    - note it will require a lot of fields being expired at roughly the same time
3. In our result processor check if the document itself was expired
4. This set of changes helps in the case of lazy expiration to some degree.
    - if the document/field expires between us retrieving it and opening it and retrieving its fields then the problematic edge case remains.
    - We will need a change at the redis level to be able to solve the problem 100%.

A clear and concise description of what the PR is solving, including:
1. Documents are expired during the query in a lazy fashion which causes empty documents to be returned to the user
    - For fields the issue is even trickier since we can return what the user perceive as incorrect results.
3. Try and minimize the edge case by checking for the expiration before returning a result up the chain
6. Less empty documents will be returned to the user.

**Which issues this PR fixes**
1. MOD-7329

**Main objects this PR modified**
1. Iterators relating to fields
2. Document Table

**Mark if applicable**

- [ ] This PR introduces API changes
- [X] This PR introduces serialization changes

PR #5241, opened 2024-11-21
[2.10] MOD-8097, MOD-8114 Fix memory counting (#5204)

meiravgri:
backport #5204

## conflicts:
- skip changes from #4858 that weren't backported 
- no WithMissingInfo test in test_cpp_llapi

## Backport bug fix
- https://github.com/RediSearch/RediSearch/pull/5008 contained, among other changes, a bug fix in FGC_parentHandleMissingDocs, where we update gc stats before counting the bytes collected by removing an empty index. However, **this pr was reverted in 2.10, including the bug fix.**
This backport re-includes this bug fix.

### DIFF
--- a/tests/pytests/common.py
+++ b/tests/pytests/common.py
@@ -669,3 +669,35 @@ def verify_shard_init(shard):
 def cmd_assert(env, cmd, res, message=None):
     db_res = env.cmd(*cmd)
     env.assertEqual(db_res, res, message=message)
+
+# fields should be in capital letters
+def getInvertedIndexInitialSize_MB(fields) -> float:
+    total_size = 0
+    for field in fields:
+        if field in ['GEO', 'NUMERIC']:
+            block_size = 48
+            initial_block_cap = 6
+            inverted_index_meta_data = 48
+            total_size += (block_size + initial_block_cap + inverted_index_meta_data) / float(1024 * 1024)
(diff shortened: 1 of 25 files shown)


tests/pytests/common.py
```diff
@@ -669,3 +669,35 @@ def verify_shard_init(shard):
...
+            inverted_index_meta_data = 48
+            total_size += (block_size + initial_block_cap + inverted_index_meta_data) / float(1024 * 1024)
+        elif field not in ['TEXT', 'TAG', 'GEOMETRY', 'VECTOR']:
+            raise ValueError(f'Field {field} is not supported')
+
+    return total_size
+
+def check_index_info(env, idx, exp_num_records, exp_inv_idx_size, msg=""):
+    d = index_info(env, idx)
+    env.assertEqual(float(d['num_records']), exp_num_records, message=msg)
```
GuyAv46, line 689: We should address the depth in all the assertions here. Without them we may miss the actual line in the test that caused the issue

GuyAv46 (approved):

tests/pytests/common.py
meiravgri: This is the motivation behind passing `msg`
i added a message describing the location in the tests that call this function
Did you notice tests where it's missing or not informative enough?

PR #5274, opened 2024-11-29
[MOD-7949] Avoid lazy expire upon scan keys in background

alonre24:

Avoid lazy expired during background scan of keys upon creating the index to avoid potential propagation of multiple DEL commands within a transaction (that might cause cross slot violation down the road).

We disabled the monitoring of has field expiration by default since otherwise, we call HPEXPIRETIME on every indexing
of a document, which causes lazy expiration even if not intended (this should be changed anyway since it was causing performance regressions)

**Mark if applicable**


src/document_basic.c
```diff
@@ -175,9 +175,10 @@ int Document_LoadSchemaFieldHash(Document *doc, RedisSearchCtx *sctx, QueryError
   if (spec->monitorFieldExpiration) {
     doc->fieldExpirations = getHashFieldExpirationTime(spec, sctx->redisCtx, doc->docKey);
   }
-  RedisModuleKey *k = RedisModule_OpenKey(sctx->redisCtx, doc->docKey, REDISMODULE_READ);
+  RedisModuleKey *k = RedisModule_OpenKey(sctx->redisCtx, doc->docKey,
+    REDISMODULE_READ | REDISMODULE_OPEN_KEY_NOEFFECTS);
   int rv = REDISMODULE_ERR;
-  // DvirDu: Is this even possible?
+  // This is possible if the key has expired for example.
```
kei-nan, line 181:
```suggestion
  // This is possible if the key has expired for example in previous redis API calls in this notification flow.
```

src/document_basic.c
```diff
@@ -243,7 +244,7 @@ int Document_LoadSchemaFieldJson(Document *doc, RedisSearchCtx *sctx, QueryError
   size_t nitems = sctx->spec->numFields;
   JSONResultsIterator jsonIter = NULL;
 
-  RedisModuleKey *k = RedisModule_OpenKey(sctx->redisCtx, doc->docKey, REDISMODULE_READ);
+  RedisModuleKey *k = RedisModule_OpenKey(sctx->redisCtx, doc->docKey, REDISMODULE_READ | REDISMODULE_OPEN_KEY_NOEFFECTS);
```
kei-nan, line 247: Maybe we also want the noexpire flag here?
WDYT?

DvirDukhan: isn't it part of no effect?

DvirDukhan (dismissed):

src/document_basic.c
alonre24: I don't think so because then we would avoid testing the expiration time, and we would index the expired doc in vain.

kei-nan (approved):

tests/pytests/test_expire.py
```diff
@@ -503,3 +509,47 @@ def testDocWithLongExpiration(env):
...
     env.expect('FT.SEARCH', 'idx', 'hello', 'NOCONTENT').apply(sort_document_names).equal([2, 'doc:1', 'doc:2'])
+
+
+# Verify that background indexing does not cause lazy expiration of expired documents.
+@skip(cluster=True)
+def test_background_index_no_lazy_expiration(env):
+    env.cmd('DEBUG', 'SET-ACTIVE-EXPIRE', '0')
+    env.expect('HSET', 'doc:1', 't', 'bar').equal(1)
+    env.expect('HSET', 'doc:2', 't', 'arr').equal(1)
+    env.expect('EXPIRE', 'doc:1', '1').equal(1)
```
MeirShpilraien, line 520: You can use hexpire to make the test faster.

PR #5295, opened 2024-12-02
Fix long PREFIX/SUFFIX/INFIX/WILDCARD queries - [MOD-7882]

GuyAv46:

Validate we managed to create a rune string from the input string and return an error if not.

1. MOD-7882
2. MOD-5291
3. MOD-7895

**Mark if applicable**

### DIFF
--- a/src/query.c
+++ b/src/query.c
@@ -35,6 +35,11 @@
 #include "wildcard.h"
 #include "geometry/geometry_api.h"
 
+#ifndef STRINGIFY
+#define __STRINGIFY(x) #x
+#define STRINGIFY(x) __STRINGIFY(x)
+#endif
+
 #define EFFECTIVE_FIELDMASK(q_, qn_) ((qn_)->opts.fieldMask & (q)->opts->fieldmask)
 
 static void QueryTokenNode_Free(QueryTokenNode *tn) {
@@ -561,6 +566,8 @@ static const char *PrefixNode_GetTypeString(const QueryPrefixNode *pfx) {
   }
 }
 
+#define TRIE_STR_TOO_LONG_MSG "query string is too long. Maximum allowed length is " STRINGIFY(MAX_RUNESTR_LEN)
+
 /* Evaluate a prefix node by expanding all its possible matches and creating one big UNION on all
  * of them.
  * Used for Prefix, Contains and suffix nodes.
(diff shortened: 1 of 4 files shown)


tests/pytests/test_issues.py
```diff
@@ -1248,3 +1248,18 @@ def test_mod_8142(env:Env):
   res2 = env.cmd('FT.SEARCH', 'idx', 'city=>[KNN 10 @v $BLOB]', 'WITHSCORES', 'RETURN', '1', 't', 'DIALECT', '2',
                                                                 'PARAMS', 2, 'BLOB', np.array([1, 0], dtype=np.float32).tobytes())
   env.assertEqual(res1, res2)
+
+@skip(cluster=True)
+def test_mod_7882(env:Env):
+  """
+  We currently not supporting searching for strings that are longer than 1024 characters in the Trie.
```
oshadmi, line 1255: Consider rephrase (`are not supporting` or `do not support`)

src/query.c
```diff
@@ -571,10 +581,11 @@ static IndexIterator *Query_EvalPrefixNode(QueryEvalCtx *q, QueryNode *qn) {
     return NULL;
   }
 
-  rune *str = NULL;
   size_t nstr;
-  if (qn->pfx.tok.str) {
-    str = strToFoldedRunes(qn->pfx.tok.str, &nstr);
+  rune *str = strToFoldedRunes(qn->pfx.tok.str, &nstr);
+  if (!str) {
+    QueryError_SetErrorFmt(q->status, QUERY_ELIMIT, "%s query string is too long", PrefixNode_GetTypeString(&qn->pfx));
```
oshadmi, line 587: Can we also include the limit in the message?  (value of `MAX_RUNESTR_LEN`)
(although we would also document it)
(also relevant for the other message below)

src/query.c
```diff
@@ -571,10 +581,11 @@ static IndexIterator *Query_EvalPrefixNode(QueryEvalCtx *q, QueryNode *qn) {
     return NULL;
   }
 
-  rune *str = NULL;
   size_t nstr;
-  if (qn->pfx.tok.str) {
-    str = strToFoldedRunes(qn->pfx.tok.str, &nstr);
+  rune *str = strToFoldedRunes(qn->pfx.tok.str, &nstr);
```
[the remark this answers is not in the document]
GuyAv46: I don't think it is possible. We don't parse a prefix node with no string, and we also checked that the `len >= minTermPrefix` (and `minTermPrefix >= 1`).
But we can go on the safe side

oshadmi (approved):

PR #5297, opened 2024-12-03
MOD-6058: Circular Buffer

raz-mon:
Adds a Circular Buffer to RediSearch, see the header file for more specification of the exposed API.

**Mark if applicable**


src/util/circular_buffer.c
```diff
@@ -0,0 +1,289 @@
...
+	_Atomic uint64_t write;       // write offset into data
+	size_t item_size;             // item size in bytes
+	_Atomic uint64_t item_count;  // current number of items in buffer
+	uint64_t item_cap;            // max number of items held by buffer
+	char *end_marker;             // marks the end of the buffer
+	char data[];                  // data
+};
+
+// Creates a new circular buffer, with `cap` items of size `item_size`
+CircularBuffer CircularBuffer_New
```
kei-nan, line 29: The function signature convention seems off.
Consider aligning it to the rest of our codebase.

src/util/circular_buffer.c
```diff
@@ -0,0 +1,289 @@
...
+
+	return cb->item_cap;
+}
+
+// Returns the size of each item in the buffer
+uint CircularBuffer_ItemSize
+(
+	const CircularBuffer cb  // buffer
+) {
+	return cb->item_size;
```
kei-nan, line 71: Can you add coverage for this function?

src/util/circular_buffer.c
```diff
@@ -0,0 +1,289 @@
...
+	CircularBuffer cb,   // buffer to populate
+	void *item           // item to add
+) {
+	RedisModule_Assert(cb   != NULL);
+	RedisModule_Assert(item != NULL);
+
+	// atomic update buffer item count
+	// do not add item if buffer is full
+	uint64_t item_count = atomic_fetch_add(&cb->item_count, 1);
+	if(unlikely(item_count >= cb->item_cap)) {
```
kei-nan, line 157:
```suggestion
	if (unlikely(item_count >= cb->item_cap)) {
```

src/util/circular_buffer.c
```diff
@@ -0,0 +1,289 @@
...
+		cb->item_count = cb->item_cap;
+		return 0;
+	}
+
+	// determine current and next write position
+	uint64_t offset = atomic_fetch_add(&cb->write, cb->item_size);
+
+	// check for buffer overflow
+	if(unlikely(cb->data + offset >= cb->end_marker)) {
```
kei-nan, line 166:
```suggestion
	if (unlikely(cb->data + offset >= cb->end_marker)) {
```

src/util/circular_buffer.c
```diff
@@ -0,0 +1,289 @@
...
+void *CircularBuffer_Reserve
+(
+	CircularBuffer cb  // buffer to populate
+) {
+	RedisModule_Assert(cb != NULL);
+
+	// atomic update buffer item count
+	// an item will be overwritten if buffer is full
+	uint64_t item_count = atomic_fetch_add(&cb->item_count, 1);
+	if(unlikely(item_count >= cb->item_cap)) {
```
kei-nan, line 209:
```suggestion
	if (unlikely(item_count >= cb->item_cap)) {
```

src/util/circular_buffer.c
```diff
@@ -0,0 +1,289 @@
...
+		cb->item_count = cb->item_cap;
+	}
+
+	// determine current and next write position
+	uint64_t offset = atomic_fetch_add(&cb->write, cb->item_size);
+
+	// check for buffer overflow
+	if(unlikely(cb->data + offset >= cb->end_marker)) {
```
kei-nan, line 217:
```suggestion
	if (unlikely(cb->data + offset >= cb->end_marker)) {
```

src/util/circular_buffer.c
```diff
@@ -0,0 +1,289 @@
...
+// This function pops the oldest item from the buffer.
+void *CircularBuffer_Read
+(
+	CircularBuffer cb,  // buffer to read item from
+	void *item          // [optional] pointer populated with removed item
+) {
+	RedisModule_Assert(cb != NULL);
+
+	// make sure there's data to return
+	if(unlikely(CircularBuffer_Empty(cb))) {
```
kei-nan, line 256:
```suggestion
	if (unlikely(CircularBuffer_Empty(cb))) {
```

src/util/circular_buffer.c
```diff
@@ -0,0 +1,289 @@
...
+
+	// copy item from buffer to output
+	if(item != NULL) {
+		memcpy(item, cb->read, cb->item_size);
+	}
+
+	// advance read position
+	// circle back if read reached the end of the buffer
+	cb->read += cb->item_size;
+	if(unlikely(cb->read >= cb->end_marker)) {
```
kei-nan, line 273:
```suggestion
	if (unlikely(cb->read >= cb->end_marker)) {
```

src/util/circular_buffer.c
```diff
@@ -0,0 +1,289 @@
...
+		return NULL;
+	}
+
+	void *read = cb->read;
+
+	// update buffer item count
+	cb->item_count--;
+
+	// copy item from buffer to output
+	if(item != NULL) {
```
kei-nan, line 266:
```suggestion
	if (item != NULL) {
```

src/util/circular_buffer.h
```diff
@@ -0,0 +1,97 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// forward declaration
+typedef struct _CircularBuffer _CircularBuffer;
+typedef _CircularBuffer* CircularBuffer;
+
+// Creates a new circular buffer, with `cap` items of size `item_size`
+CircularBuffer CircularBuffer_New
```
kei-nan, line 21: All these function signatures seem to deviate from our code convention.
Consider aligning them.

tests/cpptests/test_cpp_circularBuffer.cpp
```diff
@@ -0,0 +1,166 @@
...
+	int cap = 16;
+	CircularBuffer buff = CircularBuffer_New(sizeof(int), cap);
+
+	// remove item from an empty buffer should report failure
+	ASSERT_TRUE(CircularBuffer_Read(buff, &n) == NULL);
+
+	//--------------------------------------------------------------------------
+	// fill buffer
+	//--------------------------------------------------------------------------
+	for(int i = 0; i < cap; i++) {
```
kei-nan, line 33:
```suggestion
	for (int i = 0; i < cap; i++) {
```

tests/cpptests/test_cpp_circularBuffer.cpp
```diff
@@ -0,0 +1,166 @@
...
+		// make sure item was added
+		ASSERT_EQ(CircularBuffer_Add(buff, &i), 1);
+		// validate buffer's item count
+		ASSERT_EQ(CircularBuffer_ItemCount(buff), i+1);
+	}
+	ASSERT_EQ(CircularBuffer_Full(buff), true);
+
+	// forcefully try to overflow buffer
+	for(int i = 0; i < 10; i++) {
```
kei-nan, line 42:
```suggestion
	for (int i = 0; i < 10; i++) {
```

tests/cpptests/test_cpp_circularBuffer.cpp
```diff
@@ -0,0 +1,166 @@
...
+		ASSERT_EQ(CircularBuffer_Add(buff, &n), 0);
+	}
+
+	//--------------------------------------------------------------------------
+	// empty buffer
+	//--------------------------------------------------------------------------
+	for(int i = 0; i < cap; i++) {
```
kei-nan, line 49:
```suggestion
	for (int i = 0; i < cap; i++) {
```

tests/cpptests/test_cpp_circularBuffer.cpp
```diff
@@ -0,0 +1,166 @@
...
+		// get item from buffer
+		ASSERT_TRUE(CircularBuffer_Read(buff, &n) != NULL);
+
+		// validate item's value
+		ASSERT_EQ(n, i);
+	}
+	ASSERT_EQ(CircularBuffer_Empty(buff), true);
+
+	// forcefully try to read an item from an empty buffer
+	for(int i = 0; i < 10; i++) {
```
kei-nan, line 59:
```suggestion
	for (int i = 0; i < 10; i++) {
```

tests/cpptests/test_cpp_circularBuffer.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+TEST_F(CircularBufferTest, test_CircularBuffer_Circularity) {
+	int n;
+	int cap = 16;
+	CircularBuffer buff = CircularBuffer_New(sizeof(int), cap);
+
+	//--------------------------------------------------------------------------
+	// fill buffer
+	//--------------------------------------------------------------------------
+	for(int i = 0; i < cap; i++) {
```
kei-nan, line 75:
```suggestion
	for (int i = 0; i < cap; i++) {
```

tests/cpptests/test_cpp_circularBuffer.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+	// removing an item should make space in the buffer
+	ASSERT_TRUE(CircularBuffer_Read(buff, &n) != NULL);
+	ASSERT_EQ(CircularBuffer_Add(buff, &n), 1);
+
+	//--------------------------------------------------------------------------
+	// clear buffer
+	//--------------------------------------------------------------------------
+
+	while(CircularBuffer_Empty(buff) == false) {
```
kei-nan, line 92:
```suggestion
	while (CircularBuffer_Empty(buff) == false) {
```

tests/cpptests/test_cpp_circularBuffer.cpp
```diff
@@ -0,0 +1,166 @@
...
+		CircularBuffer_Read(buff, &n);
+	}
+
+	// add/remove elements cycling through the buffer multiple times
+	for(int i = 0; i < cap * 4; i++) {
```
kei-nan, line 97:
```suggestion
	for (int i = 0; i < cap * 4; i++) {
```

tests/cpptests/test_cpp_circularBuffer.cpp
```diff
@@ -0,0 +1,166 @@
...
+	for(int i = 0; i < cap; i++) {
+		int64_t *j = (int64_t *)malloc(sizeof(int64_t));
+		CircularBuffer_Add(buff, (void*)&j);
+	}
+
+	//--------------------------------------------------------------------------
+	// free the buffer
+	//--------------------------------------------------------------------------
+
+	for(int i = 0; i < cap; i++) {
```
kei-nan, line 124:
```suggestion
	for (int i = 0; i < cap; i++) {
```

tests/cpptests/test_cpp_circularBuffer.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+TEST_F(CircularBufferTest, test_CircularBuffer_Reserve) {
+
+	// -------------------------------------------------------------------------
+	// fill a buffer of size 16 with 32 integers
+	// -------------------------------------------------------------------------
+
+	uint cap = 16;
+	CircularBuffer buff = CircularBuffer_New(sizeof(int), cap);
+	for(int i = 0; i < 2 * cap; i++) {
```
kei-nan, line 141:
```suggestion
	for (int i = 0; i < 2 * cap; i++) {
```

tests/cpptests/test_cpp_circularBuffer.cpp
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	// make sure item count did not exceeded buffer cap
+	ASSERT_EQ(CircularBuffer_ItemCount(buff), CircularBuffer_Cap(buff));
+
+	// -------------------------------------------------------------------------
+	// assert override correctness
+	// -------------------------------------------------------------------------
+
+	for(uint i = 0; i < 16; i++) {
```
kei-nan, line 153:
```suggestion
	for (uint i = 0; i < 16; i++) {
```

src/util/circular_buffer.c
```diff
@@ -0,0 +1,289 @@
...
+	// offset is 1 - 3 = -2
+	//
+	// [C, ., ., ., ., ., ., A, B]
+	//     ^                 ^
+	//     W                 R
+
+	int offset = idx - cb->item_count;
+	offset *= cb->item_size;
+
+	if(offset >= 0) {
```
kei-nan, line 134:
```suggestion
	if (offset >= 0) {
```

src/util/circular_buffer.c
```diff
@@ -0,0 +1,289 @@
...
+// [., ., ., A, B, C, ., ., .]
+//           ^        ^
+//           R        W
+//
+void CircularBuffer_ResetReader
+(
+	CircularBuffer cb  // circular buffer
+) {
+	// compensate for circularity
+	uint64_t write = cb->write;
```
kei-nan, line 112: please add test coverage for this function

kei-nan:
@raz-mon since this should be a thread safe structure, consider adding a test with multiple threads.

src/util/circular_buffer.c
raz-mon: Looks quite the same as `IndexSpec_LoadUnsafeEx()`, which convention are you talking about?
Once we have the formatter we can be more aligned on this

src/util/circular_buffer.h
raz-mon: Not true. See IndexSpec.c, references.c etc.
Once we will have a formatter we will be more aligned on this. What format are you referring to?

raz-mon:
@kei-nan multi-threading tests will be added when we actually use this object inside some wrapper object that will implement our wanted functionality.

kei-nan:
> @kei-nan multi-threading tests will be added when we actually use this object inside some wrapper object that will implement our wanted functionality.

What kind of wrapper object do you think is required here?
It seems pretty standalone and self-contained container.

raz-mon:
@kei-nan A wrapper struct that will lock for read\write with a RW_Lock in case it's needed (it probably is - use-case dependent), and to manage the buffer in general to perform operations that involve several API calls.

src/module.h
```diff
@@ -24,6 +24,18 @@ extern "C" {
 
 #define PROXY_FILTERED "_proxy-filtered"
 
+// use likely and unlikely to provide the compiler with branch prediction information
+// for example:
```
kei-nan, line 28: Maybe move this to a more general header?

src/util/circular_buffer.c
```diff
@@ -0,0 +1,289 @@
...
+	char *end_marker;             // marks the end of the buffer
+	char data[];                  // data
+};
+
+// Creates a new circular buffer, with `cap` items of size `item_size`
+CircularBuffer CircularBuffer_New
+(
+	size_t item_size,  // size of item in bytes
+	uint cap           // max number of items in buffer
+) {
```
kei-nan, line 33: e.g:
```suggestion
CircularBuffer CircularBuffer_New(size_t item_size,  uint cap) {
```

src/module.h
raz-mon: Like what?

raz-mon: We can put it in its own file for explicitness

src/util/circular_buffer.c
raz-mon: Ah this.. I actually like it, but it is indeed a different format from what we currently have.
I think we should use this, but I can align until we apply the formatter.

src/util/circular_buffer.c
```diff
@@ -0,0 +1,253 @@
...
+
+// Circular buffer structure.
+// The buffer is of fixed size.
+// Items are removed by order of insertion, similar to a queue.
+struct _CircularBuffer {
+	char *read;                   // read data from here
+	_Atomic uint64_t write;       // write offset into data
+	size_t item_size;             // item size in bytes
+	_Atomic uint64_t item_count;  // current number of items in buffer
+	uint64_t item_cap;            // max number of items held by buffer
```
kei-nan, line 23: The comment here with variable name is a bit misleading.
I thought it was the max capacity of the container.
Consider changing variable name or comment.

src/util/circular_buffer.c
```diff
@@ -0,0 +1,253 @@
...
+
+// Returns buffer capacity
+uint64_t CircularBuffer_Cap(CircularBuffer cb) {
+	RedisModule_Assert(cb != NULL);
+
+	return cb->item_cap;
+}
+
+// Returns the size of each item in the buffer
+uint CircularBuffer_ItemSize(const CircularBuffer cb) {
```
kei-nan, line 57: Shouldn't the caller already knows the item size?
How can this container be used if the type stored is not known to the users?
I think this function can be removed.

src/util/circular_buffer.c
```diff
@@ -0,0 +1,253 @@
...
+// This function is thread-safe and lock-free
+int CircularBuffer_Add(CircularBuffer cb, void *item) {
+	RedisModule_Assert(cb   != NULL);
+	RedisModule_Assert(item != NULL);
+
+	// atomic update buffer item count
+	// do not add item if buffer is full
+	uint64_t item_count = atomic_fetch_add(&cb->item_count, 1);
+	if (unlikely(item_count >= cb->item_cap)) {
+		cb->item_count = cb->item_cap;
```
kei-nan, line 86: This assignment is not atomic, you aren't taking into account other threads that might be adding or removing items.
You should atomic decrease one from the counter.

src/util/circular_buffer.c
```diff
@@ -0,0 +1,253 @@
...
+		return 0;
+	}
+
+	// determine current and next write position
+	uint64_t offset = atomic_fetch_add(&cb->write, cb->item_size);
```
kei-nan, line 91: Consider ensuring item_size is aligned to a power of two in the constructor.
Also consider having sentinel values in between items to ensure one item doesn't overwrite another.

src/util/circular_buffer.c
```diff
@@ -0,0 +1,253 @@
...
+
+	// check for buffer overflow
+	if (unlikely(cb->data + offset >= cb->end_marker)) {
```
kei-nan, line 94: How can this happen if you already previously checked the item_count?

src/util/circular_buffer.c
```diff
@@ -0,0 +1,253 @@
...
+
+		// update write position
+		// multiple threads "competing" to update write position
+		// only the thread with the largest offset will succeed
+		// for the above example, W1 will succeed
+		//
+		// [., ., ., ., ., ., A, B, C]
+		//        ^
+		//        W
+		atomic_compare_exchange_strong(&cb->write, &overflow, offset + cb->item_size);
```
kei-nan, line 115: How does this handle multiple writers all detecting an overflow?
If there are limitations on the container then please mention them in the documentation.

src/util/circular_buffer.c
```diff
@@ -0,0 +1,253 @@
...
+// Returns a pointer to a 'item size' slot within the buffer.
+// This function is thread-safe and lock-free.
+void *CircularBuffer_Reserve(CircularBuffer cb) {
+	RedisModule_Assert(cb != NULL);
+
+	// atomic update buffer item count
+	// an item will be overwritten if buffer is full
+	uint64_t item_count = atomic_fetch_add(&cb->item_count, 1);
+	if (unlikely(item_count >= cb->item_cap)) {
+		cb->item_count = cb->item_cap;
```
kei-nan, line 135: Same thing here, you are assigning in a non atomic way.

src/util/circular_buffer.c
```diff
@@ -0,0 +1,253 @@
...
+	// determine current and next write position
+	uint64_t offset = atomic_fetch_add(&cb->write, cb->item_size);
+
+	// check for buffer overflow
+	if (unlikely(cb->data + offset >= cb->end_marker)) {
+		// write need to circle back
+		// [., ., ., ., ., ., A, B, C]
+		//                           ^  ^
+		//                           W0 W1
+		uint64_t overflow = offset + cb->item_size;
```
kei-nan, line 99: Why can't this function just call `CircularBuffer_Reserve`?

src/util/circular_buffer.c
```diff
@@ -0,0 +1,253 @@
...
+	memcpy(cb->data + offset, item, cb->item_size);
+
+	// report success
+	return 1;
+}
+
+// Reserve a slot within buffer.
+// Returns a pointer to a 'item size' slot within the buffer.
+// This function is thread-safe and lock-free.
+void *CircularBuffer_Reserve(CircularBuffer cb) {
```
kei-nan, line 128:
```suggestion
void *CircularBuffer_ReserveSlot(CircularBuffer cb) {
```

src/util/circular_buffer.c
```diff
@@ -0,0 +1,253 @@
...
+
+	// make sure there's data to return
+	if (unlikely(CircularBuffer_Empty(cb))) {
+		return NULL;
+	}
+
+	void *read = cb->read;
+
+	// update buffer item count
+	cb->item_count--;
```
kei-nan, line 184: shouldn't this be done in an atomic way?

src/util/circular_buffer.c
```diff
@@ -0,0 +1,253 @@
...
+//
+// CircularBuffer_ResetReader will set 'read' to A
+//
+// [., ., ., A, B, C, ., ., .]
+//           ^        ^
+//           R        W
+//
+void CircularBuffer_ResetReader(CircularBuffer cb) {
+	// compensate for circularity
+	uint64_t write = cb->write;
```
kei-nan, line 217: shouldn't you do an atomic load operation on write?

src/util/circular_buffer.c
```diff
@@ -0,0 +1,253 @@
...
+	int offset = idx - cb->item_count;
+	offset *= cb->item_size;
+
+	if (offset >= 0) {
+		// offset is positive, read from beginning of buffer
+		cb->read = cb->data + offset;
+	} else {
+		// offset is negative, read from end of buffer
+		cb->read = cb->end_marker + offset;
+	}
```
kei-nan, line 245: worth adding an assert that the read offset did not exceed the buffer boundaries.

kei-nan (commented):
Left some comments,
Looking at the changes more closely I think adding tests with multiple threads is something that we need to ensure the correctness of the container.

src/util/circular_buffer.h
```diff
@@ -0,0 +1,61 @@
...
+// Returns true if buffer is empty
+bool CircularBuffer_Empty(const CircularBuffer cb);
+
+// Returns true if buffer is full
+bool CircularBuffer_Full(const CircularBuffer cb);
+
+// Adds an item to buffer.
+// Returns 1 on success, 0 otherwise
+// This function is thread-safe and lock-free
+int CircularBuffer_Add(CircularBuffer cb, void *item);
```
kei-nan, line 41: The purpose of this container is eventually to support holding queries, or in general objects that were allocated on the heap.

It seems like writing can override existing items without destroying the "object".
Don't want the container to have to deal with object destruction.
But maybe provide an overload where the caller can provide a callback to release the existing slot before you write to it.
This also requires you to know if the slot read/write state.

tests/cpptests/test_cpp_circularBuffer.cpp
```diff
@@ -0,0 +1,205 @@
+#include "gtest/gtest.h"
+#include "src/util/circular_buffer.h"
+
+class CircularBufferTest : public ::testing::Test {};
+
+TEST_F(CircularBufferTest, testEmpty) {
+    CircularBuffer buff = CircularBuffer_New(sizeof(int), 16);
```
kei-nan, line 7: Indentation in this file seems off.

src/util/circular_buffer.c
raz-mon: Not sure I follow - the `item_cap` is indeed the max capacity of the buffer.
Please be more specific.

src/util/circular_buffer.c
raz-mon: It's not a must I agree, but doesn't hurt, go figure what use-case may need it some day.

src/util/circular_buffer.c
raz-mon: I think you're right. Will make this assignment atomic.

src/util/circular_buffer.c
raz-mon: Why enforce have item_size a power of two?
And why do we need the sentinal values? The writer should only write one type of object to the buffer.

src/util/circular_buffer.c
raz-mon: You were pointing at the last slot in the buffer

src/util/circular_buffer.c
raz-mon: I think you're right and this should be atomic as well - will change

src/util/circular_buffer.c
raz-mon: Two increments to `item_count`.
We can add an argument to prevent that, not sure I want that. will consider.

src/util/circular_buffer.c
raz-mon: Only one reader. See the comment above the function.

src/util/circular_buffer.c
kei-nan: In that case here is the suggested change:
```suggestion
	uint64_t item_cap;            // max number of items that can be held in buffer
```

src/util/circular_buffer.c
raz-mon: How can this happen?
The `item_count` is aligned with `write` pointer, and this function should be called under a lock, such that no writes are happening concurrently.

src/util/circular_buffer.h
raz-mon: Right.
We may add this functionality later on if needed.

tests/cpptests/test_cpp_circularBuffer.cpp
raz-mon: Which part of it?

src/util/circular_buffer.c
kei-nan: If you don't align the addresses it affects the retrieval time for a given address.
The CPU likes addresses to be aligned to the power of two.

By allowing Reserve function to exist you expose the underlying memory and allow external access to it.
That means someone else can mess with your memory.
So if we want to keep the function it would help to have sentinal values.

tests/cpptests/test_cpp_circularBuffer.cpp
raz-mon: Interesting the default tab operation is different than in other files (in other cpp test files as well btw).
Will fix

src/util/circular_buffer.c
kei-nan: but the addition happens elsewhere.
the two operations of adding + exchanging are not done together.
So one thread could have `offset` x and another x + y.
Both detected an overflow, both try to perform a compare exchange.
One will succeed while the other will not.

Lets say offset x was written into cb->write.

Both assume they were successful and continue doing a memcpy.

Now a new thread comes along and will override the item that was written in x+y.

This seems like a bug.

src/util/circular_buffer.c
kei-nan: it doesn't matter, you are competing with writer threads that are incrementing the same variable.
The decrement should be done atomically.

src/util/circular_buffer.c
raz-mon: Regarding the alignment, it seems like an overkill for this struct.
Regarding exposing the buffer - it's the idea here, as it allows the user to free the component that may have been already in that component. Moreover, I wish to maximize flexibility here, as this struct will be a component of a wrapping struct which will do such validations if necessary, exactly as done for instance with the array API we use (see `src/util/arr.h/c`).

src/util/circular_buffer.c
raz-mon: So this is not a problem.
Why? Since the `offset` is updated atomically, and then the thread with the highest value is promised to succeed in the update, as should be.

src/util/circular_buffer.c
raz-mon: See `pthread_rwlock` - no reader works while the lock is locked for write, which is the lock the reader takes here (since it is opposite) --> No need to be atomic.
I agree that it's a bit counter-inuitive that if you actually want to use multi-threading with this struct you need to wrap it with a struct and a lock for most cases, but it allows the user to also have a lock-less implementation.

src/util/circular_buffer.c
raz-mon: This is done under write lock, so no need for atomicity (not thread safe).
If we will want to later introduce a lock to the Circular Buffer we may do that, but let's first understand that we want\need that. I don't think so, as we will have a lock in the wrapping struct.

raz-mon:
@kei-nan Added some basic multi-threading tests for the `Add` and `Reserve` functions, which are the only thread-safe functions of the API.

src/util/circular_buffer.c
kei-nan: Wait, should reserve try to increment the `item_count`?
From the use cases you pointed out a valid item might already be in the slot.

raz-mon: Yes, it overwrites.

src/util/circular_buffer.c
raz-mon: Yes

kei-nan (approved):

src/util/circular_buffer.c
```diff
@@ -0,0 +1,254 @@
...
+  char data[];                  // data
+};
+
+// Creates a new circular buffer, with `cap` items of size `item_size`
+CircularBuffer CircularBuffer_New(size_t item_size, uint cap) {
+  CircularBuffer cb = rm_calloc(1, sizeof(_CircularBuffer) + item_size * cap);
+
+  cb->read       = cb->data;                      // initial read position
+  cb->write      = ATOMIC_VAR_INIT(0);            // write offset into data
+  cb->item_cap   = cap;                           // buffer capacity
```
oshadmi, line 33: Edge case: need to make sure it is not zero?
(may get to a wrong `write` offset in `CircularBuffer_Reserve`)

src/util/circular_buffer.c
```diff
@@ -0,0 +1,254 @@
...
+  memcpy(cb->data + offset, item, cb->item_size);
+
+  // report success
+  return 1;
+}
+
+// Reserve a slot within buffer.
+// Returns a pointer to a 'item size' slot within the buffer.
+// This function is thread-safe and lock-free.
+void *CircularBuffer_Reserve(CircularBuffer cb) {
```
oshadmi, line 129: Consider allowing the caller to know whether an overwrite occurred or not (and maybe get the overwritten data, so it could be cleaned up by the caller)

src/util/circular_buffer.c
raz-mon: Can you ellaborate?

oshadmi: I see you handled zero cap or item_size.

src/util/circular_buffer.c
oshadmi: Right, so the caller gets the pointer to the data, but still doesn't know whether it is about to be overwritten or not. We can let the caller handle that, but we can also reflect that if we like, e.g., with a boolean out parameter.

raz-mon: I agree it's a nice logic to have (not a must), I can add it

src/util/circular_buffer.c
```diff
@@ -0,0 +1,259 @@
...
+void *CircularBuffer_Reserve(CircularBuffer cb, bool *wasFull) {
+  RedisModule_Assert(cb != NULL);
+
+  // atomic update buffer item count
+  // an item will be overwritten if buffer is full
+  uint64_t item_count = atomic_fetch_add(&cb->item_count, 1);
+  if (unlikely(item_count >= cb->item_cap)) {
+    atomic_fetch_sub(&cb->item_count, 1);
+    if (wasFull != NULL) {
+      *wasFull = true;
```
oshadmi, line 140: Not setting `wasFull` to false when wasn't full?
(so the caller needs to clear `wasFull` before calling)

raz-mon: Right

oshadmi: So need to comment this in the header declaration?

raz-mon: Added the setting of it to false as well.

oshadmi (approved):

PR #4675, opened 2024-05-26
remove centos7 and ubuntu 18.04

DvirDukhan:
**Describe the changes in the pull request**

A clear and concise description of what the PR is solving, including:
remove support for centos7 and ubuntu 18/04

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #5310, opened 2024-12-04
MOD-8172: Fix CI for platforms that don't support Node.js > 16

raz-mon:
GitHub runners don't support Node.js 16 anymore + RediSearch still supports platforms that do not support any newer versions of Node.js = We can't use the `@v3` versions of the `checkout` and `configure-aws-credential` actions, which until now allowed to use Node.js 16.
--> This PR replaces the use of these actions with bash scripts that will run on the corresponding platform (container), such that we do not need to execute any TS code (via the Node.js environment).

### DIFF
--- a/.github/workflows/task-build-artifacts.yml
+++ b/.github/workflows/task-build-artifacts.yml
@@ -170,13 +170,6 @@ jobs:
           path: redis
       - name: Get Redis (node20 not supported)
         if: steps.node20.outputs.supported == 'false'
-        uses: actions/checkout@v3
-        with:
-          repository: redis/redis
-          ref: ${{ inputs.redis-ref }}
-          path: redis
-      - name: Get Redis (node20 not supported)
-        if: steps.node20.outputs.supported == 'false'
         run: |
             REPO_URL="https://github.com/redis/redis.git"
             DEST_DIR="redis"  # Directory to clone into
@@ -214,11 +207,6 @@ jobs:
           role-to-assume: ${{ vars.ARTIFACT_UPLOAD_AWS_ROLE_TO_ASSUME }}
           aws-region: ${{ vars.ARTIFACT_UPLOAD_AWS_REGION }}
       - name: Configure AWS Credentials Using Role (node20 not supported)
-        # if: vars.USE_AWS_ROLE == 'true' && steps.node20.outputs.supported == 'false'
-        # uses: aws-actions/configure-aws-credentials@v3
-        # with:
-        #   role-to-assume: ${{ vars.ARTIFACT_UPLOAD_AWS_ROLE_TO_ASSUME }}
-        #   aws-region: ${{ vars.ARTIFACT_UPLOAD_AWS_REGION }}
         run: |
           # Variables from the workflow
           ROLE_ARN="${{ vars.ARTIFACT_UPLOAD_AWS_ROLE_TO_ASSUME }}"
--- a/.github/workflows/task-build-artifacts.yml
+++ b/.github/workflows/task-build-artifacts.yml
@@ -175,7 +175,7 @@ jobs:
             DEST_DIR="redis"  # Directory to clone into
 
             # Clone the repository (shallow clone without tags)
-            git clone --depth=1 $REPO_URL $DEST_DIR
+            git clone $REPO_URL $DEST_DIR
             cd $DEST_DIR
 
             # Checkout the REF
--- a/.github/workflows/task-build-artifacts.yml
+++ b/.github/workflows/task-build-artifacts.yml
@@ -179,7 +179,8 @@ jobs:
             cd $DEST_DIR
 
             # Checkout the REF
-            git fetch origin ${{ inputs.redis-ref }}
+            git remote update
+            git fetch
             git checkout origin/${{ inputs.redis-ref }}
       - name: Build Redis
         working-directory: redis
--- a/.github/workflows/task-build-artifacts.yml
+++ b/.github/workflows/task-build-artifacts.yml
@@ -56,14 +56,9 @@ jobs:
       - name: Check if node20 is Supported
         id: node20 # TODO: Remove this when node20 is supported on all platforms, or when we drop support for theses platforms
         run: |
-          for platform in ubuntu:bionic centos:7 amazonlinux:2 alpine:3; do
+          for platform in ubuntu:bionic amazonlinux:2 alpine:3; do
             if [[ "${{ inputs.container }}" == "$platform" ]]; then
               echo "supported=false" >> $GITHUB_OUTPUT
-              # https://github.com/actions/checkout/issues/1809
-              echo "ACTIONS_RUNNER_FORCED_INTERNAL_NODE_VERSION=node16" >> $GITHUB_ENV
-              echo "ACTIONS_RUNNER_FORCE_ACTIONS_NODE_VERSION=node16" >> $GITHUB_ENV
-              # https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/
-              echo "ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true" >> $GITHUB_ENV
               exit 0
             fi
           done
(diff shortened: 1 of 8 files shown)


.github/workflows/task-build-artifacts.yml
```diff
@@ -80,20 +78,73 @@ jobs:
...
-          ref: ${{ env.REF }}
-          sparse-checkout-cone-mode: false
-          sparse-checkout: |
-            .install
-            tests/pytests/requirements.txt
-      - name: Setup specific
+        run: |
+          # Checkout script logic
+          perform_checkout_ref() {
+            REPO_URL="https://github.com/RediSearch/RediSearch.git"
```
oshadmi, line 84: Can we use something generic like
```
REPO_URL="https://github.com/${{ github.repository }}"
```

.github/workflows/task-build-artifacts.yml
```diff
@@ -80,20 +78,73 @@ jobs:
...
+          # Execute the logic based on the detected platform
+          echo "Detected platform: ${{ inputs.container }}"
+          case "${{ inputs.container }}" in
+            ubuntu:bionic | amazonlinux:2 | alpine:3)
+              echo "Installing prerequisites for ${{ inputs.container }}..."
+
+              # Install git (platform-specific package management)
+              case "${{ inputs.container }}" in
+                ubuntu:bionic)
+                  echo "Updating package lists..."
```
oshadmi, line 107: Echo messages also in other cases?

.github/workflows/task-build-artifacts.yml
```diff
@@ -158,10 +212,20 @@ jobs:
...
-          role-to-assume: ${{ vars.ARTIFACT_UPLOAD_AWS_ROLE_TO_ASSUME }}
-          aws-region: ${{ vars.ARTIFACT_UPLOAD_AWS_REGION }}
+        run: |
+          # Variables from the workflow
+          ROLE_ARN="${{ vars.ARTIFACT_UPLOAD_AWS_ROLE_TO_ASSUME }}"
+          AWS_REGION="${{ vars.ARTIFACT_UPLOAD_AWS_REGION }}"
+          SESSION_NAME="github-actions-session"
+
+          # Assume the AWS role and configure temporary credentials
+          echo "Assuming AWS Role: $ROLE_ARN"
```
oshadmi, line 222: Do we need such echoing?

.github/workflows/task-build-artifacts.yml
```diff
@@ -171,10 +235,26 @@ jobs:
...
+
+          # Configure AWS CLI with provided credentials and region
+          echo "Configuring AWS CLI with access keys..."
+          aws configure set aws_access_key_id "$AWS_ACCESS_KEY_ID"
+          aws configure set aws_secret_access_key "$AWS_SECRET_ACCESS_KEY"
+          aws configure set region "$AWS_REGION"
+
+          # Verify configuration
+          echo "Verifying AWS CLI configuration..."
+          aws sts get-caller-identity
```
oshadmi, line 258: Do we need such output?

.github/workflows/task-build-artifacts.yml
raz-mon: One print before is good enough, removing this one.

.github/workflows/task-test.yml
```diff
@@ -80,26 +75,61 @@ jobs:
...
-        uses: actions/checkout@v3
-        with:
-          path: setup
-          sparse-checkout-cone-mode: false
-          sparse-checkout: |
-            .install
-            tests/pytests/requirements.*
-      - name: Setup specific
+        run: |
+          # Execute the logic based on the detected platform
```
oshadmi, line 79: If this snippet is repeated, consider putting it in a bash script and reuse it?

.github/workflows/task-build-artifacts.yml
raz-mon: No, removed

.github/workflows/task-build-artifacts.yml
raz-mon: The print?
Don't see how it hurts

.github/workflows/task-build-artifacts.yml
```diff
@@ -56,14 +59,9 @@ jobs:
       - name: Check if node20 is Supported
         id: node20 # TODO: Remove this when node20 is supported on all platforms, or when we drop support for theses platforms
         run: |
-          for platform in ubuntu:bionic centos:7 amazonlinux:2 alpine:3; do
```
kei-nan, line 59: Why was centos:7 removed?

.github/workflows/task-test.yml
kei-nan: You could use a separate repo to house your custom actions and expose them as a github action.

.github/workflows/task-build-artifacts.yml
raz-mon: Was removed from our CI [here](https://github.com/RediSearch/RediSearch/pull/4675) (in master - if it still exists in the version branches it will not be removed there).

.github/workflows/task-test.yml
```diff
@@ -172,11 +201,7 @@ jobs:
         if: >
           steps.node20.outputs.supported == 'false' &&
           (steps.unit_tests.outcome == 'failure' || steps.standalone_tests.outcome == 'failure' || steps.coordinator_tests.outcome == 'failure')
-        uses: actions/upload-artifact@v3
-        with:
-          name: Test Logs ${{ steps.artifact-names.outputs.name }}
-          path: tests/**/logs/*.log*
-          if-no-files-found: ignore
+        run: echo "Currently not available..."
```
oshadmi, line 204: Can do it in a separate PR (maybe even just a `grep -C` for some `[FAIL]` etc. tokens

oshadmi (dismissed):

.github/workflows/task-build-artifacts.yml
raz-mon: I see it is not supported on any of the version branches as well (which makes sense..)

.github/workflows/task-test.yml
```diff
@@ -80,26 +75,59 @@ jobs:
...
-        with:
-          path: setup
-          sparse-checkout-cone-mode: false
-          sparse-checkout: |
-            .install
-            tests/pytests/requirements.*
-      - name: Setup specific
+        run: |
+          # Execute the logic based on the detected platform
+          echo "Detected platform: ${{ inputs.container }}"
```
kei-nan, line 80: Can all of these bash scripts move to separate yml files?
It might also allow some code reuse.

.github/workflows/task-build-artifacts.yml
GuyAv46: it should be supported on 2.6, but it was listed here just because there was no good reason to remove it (has no effect)

.github/workflows/task-test.yml
raz-mon: This is before we have our repo cloned, so I don't think so. At this point the machine only has this flow in hand.

raz-mon: Maybe we can hold scripts in the repo and use them like variables

.github/workflows/task-test.yml
```diff
@@ -172,11 +201,7 @@ jobs:
         if: >
```
GuyAv46, line 201: consider updating the comment here

.github/workflows/task-test.yml
```diff
@@ -80,26 +75,59 @@ jobs:
...
         if: steps.node20.outputs.supported == 'true'
         uses: actions/checkout@v4
         with:
           submodules: recursive
       - name: Full checkout (node20 not supported)
         if: steps.node20.outputs.supported == 'false'
-        uses: actions/checkout@v3
-        with:
-          submodules: recursive
+        run: echo "Done already earlier."
```
GuyAv46, line 130: Consider grouping all the node-20 steps together and all the non-node-20 together. I think it will make more sense than having these dummy steps

.github/workflows/task-build-artifacts.yml
raz-mon: Right, 2.6 as well.
It's confusing w.r.t to whether it is supported on latest or not, I prefer it not to appear where it is not supported

.github/workflows/task-test.yml
raz-mon: You mean to remove the failure condition?

GuyAv46 (approved):

.github/workflows/task-test.yml
kei-nan: How does this script get called?
Isn't it used like this?
```
```

raz-mon: Now I see your meaning - sounds good will try to add it.

PR #5319, opened 2024-12-05
MOD-8053: Validate ACL key permissions for RediSearch commands

raz-mon:
In this PR we add the keys ACL validation to all of our command handlers.
Specifically, we validate that a user can only access an index if his ACL key permissions are a superset of the prefixes indexes by the index it wishes to access.

Notice: After this PR we cannot query an added shard on OSS that does not know the index, since it will now return an `no such index` as part of the ACL validation we now do in the coordinator. This is a very unwanted situation that should not happen (shard that is not aware of the indexes in the DB), and can cause wrong results to be returned to the user! After this PR, we at least know that we will not return wrong results, since we do not allow such a query to "fly" without all the shards of the cluster knowing about all the existing indexes.

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -3548,7 +3548,6 @@ RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
     RM_TRY(RMCreateSearchCommand(ctx, "FT._ALIASDELIFX", SafeCmd(FanoutCommandHandlerIndexless), "readonly", 0, 0, -1, ""))
     RM_TRY(RMCreateSearchCommand(ctx, "FT.ALIASUPDATE", SafeCmd(FanoutCommandHandlerWithIndexAtFirstArg), "readonly", 0, 0, -1, ""))
     RM_TRY(RMCreateSearchCommand(ctx, "FT.SYNUPDATE", SafeCmd(FanoutCommandHandlerWithIndexAtFirstArg),"readonly", 0, 0, -1, ""))
-    RM_TRY(RMCreateSearchCommand(ctx, "FT.SYNFORCEUPDATE", SafeCmd(FanoutCommandHandlerWithIndexAtFirstArg),"readonly", 0, 0, -1, ""))
 
     // Deprecated OSS commands
     RM_TRY(RMCreateSearchCommand(ctx, "FT.GET", SafeCmd(SingleShardCommandHandlerWithIndexAtFirstArg), "readonly", 0, 0, -1, "read admin"))
--- a/src/module.c
+++ b/src/module.c
@@ -85,37 +85,46 @@ typedef struct {
 // Array containing commands names and the index of the index name in the
 // command arguments.
 static CommandIndexNamePos commandIndexPositions[] = {
-  {"FT.CURSOR",      1},
-  {"FT.SEARCH",      1},
-  {"FT.AGGREGATE",   1},
-  {"FT.INFO",        1},
-  {"FT.SPELLCHECK",  1},
-  {"FT.ALIASADD",    2},
-  {"FT.ALIASUPDATE", 2},
-  {"FT.PROFILE",     1},
-  {"FT.SYNUPDATE",   1},
-  {"FT.SYNDUMP",     1},
-  {"FT.ALTER",       1},
-  {"FT.DROPINDEX",   1},
-  {"FT.EXPLAIN",     1},
-  {"FT.EXPLAINCLI",  1},
-  {"FT.TAGVALS",     1},
-  {"FT.ADD",         1},
-  {"FT.GET",         1},
-  {"FT.DEL",         1},
-  {"FT.DROP",        1},
-  {"FT.TAGVALS",     1},
-  {"FT.MGET",        1},
-  {"FT.CREATE",     -1},  // Since this index does not exist.
-  {"FT.ALIASDEL",   -1},
-  {"FT.CONFIG",     -1},
-  {"FT.DICTADD",    -1},
-  {"FT.DICTDEL",    -1},
-  {"FT.DICTDUMP",   -1},
-  {"FT.SUGADD",     -1},
-  {"FT.SUGDEL",     -1},
-  {"FT.SUGGET",     -1},
-  {"FT.SUGLEN",     -1},
+  {"FT.CURSOR",       1},
+  {"FT.SEARCH",       1},
+  {"FT.AGGREGATE",    1},
+  {"FT.INFO",         1},
+  {"FT.SPELLCHECK",   1},
+  {"FT.ALIASADD",     2},
+  {"FT.ALIASUPDATE",  2},
+  {"FT.PROFILE",      1},
+  {"FT.SYNUPDATE",    1},
+  {"FT.SYNDUMP",      1},
+  {"FT.ALTER",        1},
+  {"FT.DROPINDEX",    1},
+  {"FT.EXPLAIN",      1},
+  {"FT.EXPLAINCLI",   1},
+  {"FT.TAGVALS",      1},
+  {"FT.ADD",          1},
+  {"FT.GET",          1},
+  {"FT.DEL",          1},
+  {"FT.DROP",         1},
+  {"FT.TAGVALS",      1},
+  {"FT.MGET",         1},
+  {"FT.CREATE",      -1},  // Since this index does not exist.
+  {"FT.ALIASDEL",    -1},
+  {"FT.CONFIG",      -1},
+  {"FT.DICTADD",     -1},
+  {"FT.DICTDEL",     -1},
+  {"FT.DICTDUMP",    -1},
+  {"FT.SUGADD",      -1},
+  {"FT.SUGDEL",      -1},
+  {"FT.SUGGET",      -1},
+  {"FT.SUGLEN",      -1},
+  // Do not validate replication commands
+  {"FT._ALTERIFNX",    -1},
+  {"FT._ALIASADDIFNX",    -1},
+  {"FT._ALIASDELIFX",    -1},
+  {"FT._DROPIFX",    -1},
+  {"FT._DROPINDEXIFX", -1},
+  {"FT._CREATEIFNX",   -1},
+  // What is this?
+  {"FT.SYNFORCEUPDATE", -1},
 };
 
 extern RSConfig RSGlobalConfig;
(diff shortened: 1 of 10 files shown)


src/module.c
```diff
@@ -2843,22 +2909,39 @@ int SpellCheckCommandHandler(RedisModuleCtx *ctx, RedisModuleString **argv, int
...
 
+static int CommandIndexPos(RedisModuleString *cmd) {
+  const char *cmdStr = RedisModule_StringPtrLen(cmd, NULL);
+  size_t n_commands = sizeof(commandIndexPositions) / sizeof(commandIndexPositions[0]);
+  for (size_t i = 0; i < n_commands; i++) {
+    if (strcasecmp(cmdStr, commandIndexPositions[i].cmdName) == 0) {
+      return commandIndexPositions[i].indexNamePos;
+    }
+  }
+  RS_LOG_ASSERT(false, "Command not found in commandIndexPositions");   // TODO: To be removed - for testing only
```
oshadmi, line 2920: Remove TODO

src/module.c
```diff
@@ -2843,22 +2909,39 @@ int SpellCheckCommandHandler(RedisModuleCtx *ctx, RedisModuleString **argv, int
...
+
+  if (NumShards == 1) {
     // There is only one shard in the cluster. We can handle the command locally.
     return genericCallUnderscoreVariant(ctx, argv, argc);
-  }
-  if (cannotBlockCtx(ctx)) {
+  } else if (cannotBlockCtx(ctx)) {
     return ReplyBlockDeny(ctx, argv[0]);
   }
-  RS_AutoMemory(ctx);
```
oshadmi, line 2861: Consider enabling auto memory after checks (as before this PR)

src/module.c
```diff
@@ -2870,24 +2953,50 @@ static int MastersFanoutCommandHandler(RedisModuleCtx *ctx, RedisModuleString **
...
+// Supports FT.ADD, FT.DEL, FT.GET, FT.SUGADD, FT.SUGGET, FT.SUGDEL, FT.SUGLEN.
+// If needed for more commands, make sure `MRCommand_GetShardingKey` is implemented for them.
+// Notice that only OSS cluster should deal with such redirections.
+int SingleShardCommandHandler(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+  if (argc < 2) {
+    return RedisModule_WrongArity(ctx);
+  } else if (!SearchCluster_Ready()) {
+    return RedisModule_ReplyWithError(ctx, CLUSTERDOWN_ERR);
+  }
+  RS_AutoMemory(ctx);
```
oshadmi, line 2965: Consider moving to after the checks

tests/pytests/test_acl.py
```diff
@@ -204,3 +204,72 @@ def test_internal_commands(env):
...
+def test_acl_key_permissions_validation(env):
+    """Tests that the key permission validation works properly"""
+
+    # Create an ACL user with partial key-space permissions
+    env.expect('ACL', 'SETUSER', 'test', 'on', '>123', '~h:*', '&*', '+@all').ok()
+
+    # Create an index on the key the user does not have permissions to access
+    # All prefixes will do (default)
+    idx_name = 'idx'
+    env.expect('FT.CREATE', idx_name, 'SCHEMA', 'n', 'NUMERIC')
```
oshadmi, line 218: Consider testing another index that has prefix `h:a` that could be used
And another index that has prefix `k:` that cannot be used

tests/pytests/test_cursors.py
```diff
@@ -232,7 +232,7 @@ def testIndexDropWhileIdle(env: Env):
         res, cursor = env.cmd(f'FT.CURSOR READ idx {cursor} COUNT 1') # read the last result
         env.assertEqual(res[1:], [[]] , message=f'res == {res}')
     else:
-        env.expect(f'FT.CURSOR READ idx {cursor}').error().contains('The index was dropped while the cursor was idle')
+        env.expect(f'FT.CURSOR READ idx {cursor}').error().contains('no such index')
```
oshadmi, line 235: Previous error message was helpful. Can we keep it?

tests/pytests/test_profile.py
```diff
@@ -15,7 +15,7 @@ def testProfileSearch(env):
   conn.execute_command('hset', '1', 't', 'hello')
   conn.execute_command('hset', '2', 't', 'world')
 
-  env.expect('ft.profile', 'profile', 'idx', '*', 'nocontent').error().contains('No `SEARCH` or `AGGREGATE` provided')
+  env.expect('ft.profile', 'profile', 'idx', '*', 'nocontent').error().contains('no such index')
```
oshadmi, line 18: Previous error message was helpful. Can we keep it?

src/module.c
```diff
@@ -3150,16 +3281,9 @@ int RSSearchCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc);
 static int DistSearchCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
   if (NumShards == 0) {
```
oshadmi, line 3282: Not necessarily in this PR - consider using `!SearchCluster_Ready`

oshadmi: (also in 2 more places)

oshadmi (commented):
💪🏼 Few comments

src/module.c
raz-mon: Agree, let's do it separately

tests/pytests/test_cursors.py
raz-mon: Unfortunately I don't think so

tests/pytests/test_profile.py
raz-mon: I don't think it was any better to be honest
I see how it is more "basic", but at the end you will get it once you query an existing index so it wasn't removed..
I can add test for it

oshadmi (approved):

src/module.c
```diff
@@ -2843,22 +2911,41 @@ int SpellCheckCommandHandler(RedisModuleCtx *ctx, RedisModuleString **argv, int
   return REDISMODULE_OK;
 }
 
+static int CommandIndexPos(RedisModuleString *cmd) {
```
oshadmi, line 2914: Need to make sure the overhead is negligible in benchmarks

oshadmi (dismissed):

src/module.c
raz-mon: You think?
I can run them no problem, but I doubt this will be noticed

src/module.c
```diff
@@ -59,8 +59,72 @@
...
+  } while(0);
+
+typedef struct {
+  const char *cmdName;
+  int indexNamePos;
+} CommandIndexNamePos;
+
+// Array containing commands names and the index of the index name in the
+// command arguments.
+static CommandIndexNamePos commandIndexPositions[] = {
```
kei-nan, line 87: This requires extra maintenance if we move arguments around or add new commands and seems a bit tricky to maintain.

An easier alternative is to add to the command function an additional argument which will denote the position of the index name, or just the index name.

This will require adding three command handler but will make this table redundant.
It will also remove the need to loop inside the command to find the right context to run with.

kei-nan (commented):
First round of comments.

src/module.c
```diff
@@ -74,6 +138,30 @@ static inline bool SearchCluster_Ready() {
   return NumShards != 0;
 }
 
+static bool ACLUserMayAccessIndex(RedisModuleCtx *ctx, IndexSpec *sp) {
+  if (RedisModule_ACLCheckKeyPrefixPermissions == NULL) {
+    // Running against a Redis version that does not support module ACL protection
+    RedisModule_Log(ctx, "warning", "Redis version does not support ACL API necessary for index protection");
```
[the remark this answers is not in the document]
raz-mon: This could actually be an assertion as it shouldn't happen, but we can be gentle here as well.
It can be in the initialization as well yes

src/module.c
```diff
@@ -74,6 +138,30 @@ static inline bool SearchCluster_Ready() {
...
+    return true;
+  }
+  RedisModuleString *user_name = RedisModule_GetCurrentUserName(ctx);
+  RedisModuleUser *user = RedisModule_GetModuleUserFromUserName(user_name);
+
+  if (!user) {
+    RedisModule_Log(ctx, "warning", "No user found");
```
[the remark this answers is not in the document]
raz-mon: It shouldn't happen as well, and notice that we don't allow the command to be executed in this case, so I think it's quite critical.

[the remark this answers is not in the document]
kei-nan: This goes to the log, if this went back as a reply then fine.
Maybe at least log the command that the user tried running?

In the documentation it says it can happen if the user is disabled.
Isn't it the same as to say the user doesn't have permissions?

Btw, aren't you missing a free call in the case of return false?

src/module.c
```diff
@@ -74,6 +138,30 @@ static inline bool SearchCluster_Ready() {
...
+    return false;
+  }
+
+  sds *prefixes = sp->rule->prefixes;
+  for (uint i = 0; i < array_len(prefixes); i++) {
+    if (RedisModule_ACLCheckKeyPrefixPermissions(user, RedisModule_CreateString(ctx, (const char *)prefixes[i], strlen(prefixes[i])), REDISMODULE_CMD_KEY_ACCESS) != REDISMODULE_OK) {
```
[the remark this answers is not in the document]
raz-mon: Done automatically by redis, but we can free it here as well sure

src/module.c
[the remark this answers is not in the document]
kei-nan: Maybe get the `RedisModuleUser*` as an argument?.
Then you can return an error in the macro.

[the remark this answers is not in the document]
raz-mon: Changed to `verbose`

src/module.c
raz-mon: Well, this introduces some nasty naming.. But is more maintainable. Went for it.

kei-nan (approved):
Overall looks okay,  like we discussed the isProfile seems a bit off.
It is possible to use additional functions to clean up that part of the code.
It would have been nicer if redis could allow us to register our own context along with the command.

Will leave it to your discretion.

kei-nan (approved):

oshadmi (approved):
Great 👍🏼
Need to label as breaking?

PR #5320, opened 2024-12-07
MOD-8223: Delete empty suggest dict

nafraf:
* When `FT.SUGDEL` is removing the last entry in a suggestion dictionary - the suggestion dictionary key is deleted , instead of keeping an empty dictionary ( (as Redis behaves, e.g., with Hashes)

* `FT.SUGGET` on a missing suggestion dictionary returns an empty array instead of `nil`  (as Redis behaves, e.g., with `HGETALL`)

2. Change: What is the change

1. MOD-8223
2. MOD-7915

1. ...

#### Mark if applicable


src/suggest.c
```diff
@@ -300,7 +305,7 @@ int RSSuggestGetCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
 
   Trie *tree = RedisModule_ModuleTypeGetValue(key);
   if (tree == NULL) {
-    RedisModule_ReplyWithNull(ctx);
+    RedisModule_ReplyWithSetOrArray(ctx, 0);
```
oshadmi, line 308: Will `RedisModule_ReplyWithArray` suffice?

nafraf: OK, thanks. Now I'm using `RedisModule_ReplyWithArray()` for empty and non-empty results.
But  the spell check dictionary is using `RedisModule_ReplyWithSetOrArray()`, what should we use to align both commands?

oshadmi (approved):

PR #5324, opened 2024-12-09
[MOD-8115] Free spec resources in the main thread

meiravgri:
This PR fixes flaky class FGCTest cpp tests.
In these tests we create a spec and add documents, potentially causing the global `TotalIIBlocks` number (global counter of the inverted indexes blocks across the model) to increase.
This number is decreased upon spec deletion in `IndexSpec_FreeUnlinkedData`
`IndexSpec_FreeUnlinkedData` is sent to a background thread if `RSGlobalConfig.freeResourcesThread` is set to its default value - `true`. In this case tests might start running while the clean-up process of previous tests is still active in the background and we can't rely on the global counters to not include changes made in previous tests, such as `TotalIIBlocks` number.
To fics that, this PR sets `RSGlobalConfig.freeResourcesThread` to false in all cpp tests, i.e disables the free resources thread in cpp-tests to ensure all the spec resources, including global objects such as the total number of blocks (`TotalIIBlocks`), are cleaned once the test has finished.

#### Better clean-up of `class FGCTest` test suite:
- use `pthread_join` instead of `pthread_detach` to wait for the gc `cbWrapper`
- add `run` flag to hint the gc wrapper thread to exit
- move `args_t args` to `class FGCTest` test suite members, instead of dynamic allocation.
 #### Allow multiple gc runs in each test 
(taken from https://github.com/RediSearch/RediSearch/pull/5049)

### DIFF
--- a/tests/cpptests/test_cpp_forkgc.cpp
+++ b/tests/cpptests/test_cpp_forkgc.cpp
@@ -47,22 +47,26 @@ static timespec getTimespecCb(void *) {
 typedef struct {
   void *fgc;
   RefManager *ism;
+  volatile bool runGc;
 } args_t;
-
-static pthread_t thread;
 
 void *cbWrapper(void *args) {
   args_t *fgcArgs = (args_t *)args;
-  ForkGC *fgc = reinterpret_cast<ForkGC *>(get_spec(fgcArgs->ism)->gc->gcCtx);
+  GCContext *gc = get_spec(fgcArgs->ism)->gc;
+  ForkGC *fgc = reinterpret_cast<ForkGC *>(gc->gcCtx);
 
-  // sync thread
-  while (fgc->pauseState != FGC_PAUSED_CHILD) {
-    usleep(500);
+  while (true) {
+    // sync thread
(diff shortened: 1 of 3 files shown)


tests/cpptests/test_cpp_forkgc.cpp
```diff
@@ -82,17 +88,16 @@ class FGCTest : public ::testing::Test {
   void runGcThread() {
     fgc = reinterpret_cast<ForkGC *>(get_spec(ism)->gc->gcCtx);
     thread = {0};
```
GuyAv46, line 90: Is this still needed?

GuyAv46 (approved):

tests/cpptests/test_cpp_forkgc.cpp
```diff
@@ -82,17 +88,16 @@ class FGCTest : public ::testing::Test {
   void runGcThread() {
     fgc = reinterpret_cast<ForkGC *>(get_spec(ism)->gc->gcCtx);
```
GuyAv46, line 89: Can we move this line to the setup function? IMO it doesn't make much sense to have it here

PR #5337, opened 2024-12-11
MOD-8238: Remove unhandled command

raz-mon:
This PR removes the `FT.SYNFORCEUPDATE` command, that we currently register to redis with no internal handler, such that it **always** returns `(error) ERR unknown command '_FT.SYNFORCEUPDATE`.


raz-mon:
@adrianoamaral FYI

PR #5347, opened 2024-12-12
Fixes for inverted indexes encoding - [MOD-8248]

GuyAv46:

While it was assumed the numeric encoder can encode any `doc_id` diff (while other encoders cannot encode delta > `UINT32_MAX`), all the encoders got a `uint32_t` delta, so the numeric encoder will not encode the right delta, causing later unrelated results to be returned.

The encoder also cannot encode any delta, but only deltas that fit in 7 bytes, due to the numeric header byte count limitation (counting 0-7 with 3 bits). We now also validate that before attempting to encode the delta, and creating a new block if needed.

1. Fix assertion condition
2. Consume less memory with numeric inverted indexes
3. Less pointer casting for safer code
4. Remove dead code

#### Mark if applicable


tests/pytests/test_stats.py
```diff
@@ -40,13 +40,13 @@ def runTestWithSeed(env, s=None):
 
     # Expected inverted index size total: 606 bytes
```
oshadmi, line 41: Update comment to 590

src/fork_gc.c
```diff
@@ -204,11 +204,8 @@ static bool FGC_childRepairInvidx(ForkGC *gc, RedisSearchCtx *sctx, InvertedInde
     // Capture the pointer address before the block is cleared; otherwise
     // the pointer might be freed! (IndexBlock_Repair rewrites blk->buf if there were repairs)
     void *bufptr = blk->buf.data;
-    int nrepaired = IndexBlock_Repair(blk, &sctx->spec->docs, idx->flags, params);
-    // We couldn't repair the block - return 0
-    if (nrepaired == -1) {
-      goto done;
-    } else if (nrepaired == 0) {
+    size_t nrepaired = IndexBlock_Repair(blk, &sctx->spec->docs, idx->flags, params);
+    if (nrepaired == 0) {
```
alonre24, line 208: not covered because it's from child?

src/inverted_index.c
```diff
@@ -1299,9 +1253,9 @@ IndexIterator *NewReadIterator(IndexReader *ir) {
 /* Repair an index block by removing garbage - records pointing at deleted documents,
  * and write valid entries in their place.
  * Returns the number of docs collected, and puts the number of bytes collected in the given
- * pointer. If an error occurred - returns -1
+ * pointer.
  */
-int IndexBlock_Repair(IndexBlock *blk, DocTable *dt, IndexFlags flags, IndexRepairParams *params) {
+size_t IndexBlock_Repair(IndexBlock *blk, DocTable *dt, IndexFlags flags, IndexRepairParams *params) {
```
alonre24, line 1258: Covered somewhere?

src/inverted_index.c
```diff
@@ -765,48 +709,61 @@ DECODER(readNumeric) {
...
   res->term.offsets = (RSOffsetVector){.data = BufferReader_Current(br), .len = res->offsetsSz};
   Buffer_Skip(br, res->offsetsSz);
-  CHECK_FLAGS(ctx, res);
+  return mask & ctx->mask;
 }
 
 DECODER(readFlagsOffsetsWide) {
-
-  qint_decode2(br, (uint32_t *)&res->docId, &res->offsetsSz);
+  uint32_t delta;
```
alonre24, line 744: Covered somewhere?

src/inverted_index.c
GuyAv46: Yes, by any GC test (we just don't see it today)

src/fork_gc.c
GuyAv46: Correct, it is only used by the fork GC child, for which we don't collect coverage reports of today

alonre24 (approved):

alonre24 (approved):

PR #5348, opened 2024-12-12
Improve text indexing performance - [MOD-8249]

GuyAv46:

Removing redundant calls to `strlen` that may slow text field indexing.

This PR doesn't have any new tests, as it doesn't introduce new logic. The fact that all the current tests pass is all the validation we need.

#### Mark if applicable


alonre24 (approved):
Nice catch, @fcostaoliveira how hard will it be to add a new dataset to our benchmarks that shows the performance difference in this use case (load long text docs)?

PR #5355, opened 2024-12-16
MOD-8271: Use TTL Information In IR_SkipTo

kei-nan:
Add field ttl verification when skipping to a document in IR_SkipTo.

1. Current: The skip to function doesn't verify ttl information on the document it skipped to and reads it when using a seeker.
2. Change: Add a check when using a seeker that the doc we seeked to is valid(field ttl wise)
3. Outcome: More correct behaviour

1. IndexReader

#### Mark if applicable


src/inverted_index.c
```diff
@@ -968,6 +968,32 @@ size_t IR_NumEstimated(void *ctx) {
...
+  t_fieldIndex fieldIndicesArray[FIELD_MASK_BIT_COUNT];
+  t_fieldIndex* sortedFieldIndices = fieldIndicesArray;
+  if (ir->filterCtx.field.isFieldMask) {
+    numFieldIndices = IndexSpec_TranslateMaskToFieldIndices(ir->sctx->spec,
+                                                            ir->filterCtx.field.value.mask,
+                                                            fieldIndicesArray);
+  } else if (ir->filterCtx.field.value.index != RS_INVALID_FIELD_INDEX) {
+    sortedFieldIndices = &ir->filterCtx.field.value.index;
+    ++numFieldIndices;
+  }
```
GuyAv46, line 991: Why do we need to perform this on every read? it seems that all this info is constant in the context of the running query. We should do it once during the iterator creation.

src/inverted_index.c
```diff
@@ -1043,6 +1056,19 @@ int IR_Read(void *ctx, RSIndexResult **e) {
 
 #define BLOCK_MATCHES(blk, docId) ((blk).firstId <= docId && docId <= (blk).lastId)
 
+static int IndexReader_ReadWithSeeker(IndexReader *ir, t_docId* inOutDocId) {
```
GuyAv46, line 1059: I don't like that we change the input soc id for the skip to function. This will cause future bugs

GuyAv46 (changes requested):

src/inverted_index.c
```diff
@@ -968,6 +968,32 @@ size_t IR_NumEstimated(void *ctx) {
 
 #define FIELD_MASK_BIT_COUNT (sizeof(t_fieldMask) * 8)
 
+// Used to detemrmine if the field mask for the given doc id are valid based on their ttl:
+// ir->filterCtx.predicate
+static inline bool VerifyFieldMaskExpirationForDocId(IndexReader *ir, t_docId docId) {
+  // If there isn't a ttl information then the doc fields are valid
+  if (!ir->sctx || !ir->sctx->spec || !DocTable_HasExpiration(&ir->sctx->spec->docs, docId)) {
```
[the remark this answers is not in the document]
kei-nan: From what I remember during unit tests we don't provide an index spec.

src/inverted_index.c
kei-nan: Expiration for fields is set per document.
You can only check if the fields expired when you reach the specific document.
Not sure I understand what you are suggesting.

src/inverted_index.c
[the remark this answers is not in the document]
GuyAv46: so let's fix that, instead of slowing down the production code

src/inverted_index.c
GuyAv46: all the marked lines are not related to the current document and should be done once at the iterator constructor.

src/inverted_index.c
```diff
@@ -1043,6 +1056,19 @@ int IR_Read(void *ctx, RSIndexResult **e) {
 
 #define BLOCK_MATCHES(blk, docId) ((blk).firstId <= docId && docId <= (blk).lastId)
 
+static int IndexReader_ReadWithSeeker(IndexReader *ir, t_docId* inOutDocId) {
+  if (!ir->decoders.seeker(&ir->br, &ir->decoderCtx, ir, *inOutDocId, ir->record)) {
```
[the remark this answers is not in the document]
kei-nan: Like we talked about yesterday, before starting the binary search we check if the current doc is good enough and if so we return it.

Also I think it is better to implement the check in a single place and not have every seeker function reimplement the call.

src/inverted_index.c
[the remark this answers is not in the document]
kei-nan: That seems out of scope for this task.
I can create a jira task.
WDYT?

GuyAv46 (approved):

raz-mon (dismissed):
LGTM

src/inverted_index.c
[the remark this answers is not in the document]
kei-nan: Created task: MOD-8795

src/inverted_index.c
```diff
@@ -893,6 +893,33 @@ size_t IR_NumEstimated(void *ctx) {
 
 #define FIELD_MASK_BIT_COUNT (sizeof(t_fieldMask) * 8)
 
+// Used to detemrmine if the field mask for the given doc id are valid based on their ttl:
```
DvirDukhan, line 896: typo

```suggestion
// Used to determine if the field mask for the given doc id are valid based on their ttl:
```

src/inverted_index.c
```diff
@@ -893,6 +893,33 @@ size_t IR_NumEstimated(void *ctx) {
...
+// ir->filterCtx.predicate
+static inline bool VerifyFieldMaskExpirationForDocId(IndexReader *ir, t_docId docId, t_fieldMask docFieldMask) {
+  // If there isn't a ttl information then the doc fields are valid
+  if (!ir->sctx || !ir->sctx->spec || !DocTable_HasExpiration(&ir->sctx->spec->docs, docId)) {
```
DvirDukhan, line 900: debug assert? why if we get a null context or spec it is returning true?

src/inverted_index.c
```diff
@@ -893,6 +893,33 @@ size_t IR_NumEstimated(void *ctx) {
...
+    return true;
+  }
+
+  // doc has expiration information, create a field id array to check for expiration predicate
+  size_t numFieldIndices = 0;
+  // Use a stack allocated array for the field indices, if the field mask is not a single field
+  t_fieldIndex fieldIndicesArray[FIELD_MASK_BIT_COUNT];
```
DvirDukhan, line 907: uninitialized?

src/inverted_index.c
```diff
@@ -893,6 +893,33 @@ size_t IR_NumEstimated(void *ctx) {
...
+  t_fieldIndex* sortedFieldIndices = fieldIndicesArray;
+  if (ir->filterCtx.field.isFieldMask) {
+    const t_fieldMask relevantMask = docFieldMask & ir->filterCtx.field.value.mask;
+    numFieldIndices = IndexSpec_TranslateMaskToFieldIndices(ir->sctx->spec,
+                                                            relevantMask,
+                                                            fieldIndicesArray);
+  } else if (ir->filterCtx.field.value.index != RS_INVALID_FIELD_INDEX) {
+    sortedFieldIndices = &ir->filterCtx.field.value.index;
+    ++numFieldIndices;
```
DvirDukhan, line 916: can we explicitly set it to `1` without all the C programming show offs?

src/inverted_index.c
```diff
@@ -962,6 +976,30 @@ int IR_Read(void *ctx, RSIndexResult **e) {
...
 
+static bool IndexReader_ReadWithSeeker(IndexReader *ir, t_docId docId) {
+  bool found = false;
+  while (!found) {
+    // try and find nextDocId using seeker
+    found = ir->decoders.seeker(&ir->br, &ir->decoderCtx, ir, docId, ir->record);
+    // ensure the entry is valid
+    if (found && !VerifyFieldMaskExpirationForDocId(ir, ir->record->docId, ir->record->fieldMask)) {
+      found = false;
+      docId = ir->record->docId + 1;
```
DvirDukhan, line 987: comment everything: why if you find the id and it is not expire you reset the search?
do not assume anything is readable and understandable in this code

src/inverted_index.c
```diff
@@ -962,6 +976,30 @@ int IR_Read(void *ctx, RSIndexResult **e) {
...
+
+    if (!found) {
+      if (BufferReader_AtEnd(&ir->br)) {
+        if (ir->currentBlock < ir->idx->size - 1) {
+          IndexReader_AdvanceBlock(ir);
+        } else {
+          break;
+        }
+      }
+    }
```
DvirDukhan, line 998: explain the logic here
also, can we move backwards, if not, why? if yes, why?
comment everything

src/inverted_index.c
```diff
@@ -893,6 +893,33 @@ size_t IR_NumEstimated(void *ctx) {
 
 #define FIELD_MASK_BIT_COUNT (sizeof(t_fieldMask) * 8)
 
+// Used to detemrmine if the field mask for the given doc id are valid based on their ttl:
+// ir->filterCtx.predicate
+static inline bool VerifyFieldMaskExpirationForDocId(IndexReader *ir, t_docId docId, t_fieldMask docFieldMask) {
```
DvirDukhan, line 898: returns true if.... returns false if...
I think you could perhaps rename it better. What's the verification here?

src/inverted_index.c
```diff
@@ -962,6 +976,30 @@ int IR_Read(void *ctx, RSIndexResult **e) {
 
 #define BLOCK_MATCHES(blk, docId) ((blk).firstId <= docId && docId <= (blk).lastId)
 
+static bool IndexReader_ReadWithSeeker(IndexReader *ir, t_docId docId) {
+  bool found = false;
+  while (!found) {
+    // try and find nextDocId using seeker
+    found = ir->decoders.seeker(&ir->br, &ir->decoderCtx, ir, docId, ir->record);
+    // ensure the entry is valid
+    if (found && !VerifyFieldMaskExpirationForDocId(ir, ir->record->docId, ir->record->fieldMask)) {
```
DvirDukhan, line 985: can you change to 
```
if (found) {
    if (!VerifyFieldMaskExpirationForDocId(ir, ir->record->docId, ir->record->fieldMask)) {
      found = false;
      docId = ir->record->docId + 1;
    }
}
else {
    if (BufferReader_AtEnd(&ir->br)) {
        if (ir->currentBlock < ir->idx->size - 1) {
          IndexReader_AdvanceBlock(ir);
        } else {
          break;
        }
      }
}
```
for readability

tests/pytests/test_expire.py
```diff
@@ -499,6 +499,36 @@ def testDocWithLongExpiration(env):
...
+    conn.execute_command('DEBUG', 'SET-ACTIVE-EXPIRE', '0')
+    conn.execute_command('FT.CREATE', 'idx', 'SCHEMA', 'x', 'TEXT', 'y', 'TEXT')
+    conn.execute_command('HSET', 'doc:0', 'x', 'hello', 'y', 'foo') # doc:expire internal id is 1001
+    # inverted index state
+    # 'hello': [1]
+    # 'foo': [1]
+    for i in range(1, 1001):
+        conn.execute_command('HSET', f'doc:{i}', 'x', 'hello', 'y', 'world')
+        # important we expire now since that assigns a new doc id for doc:expire - 1002
+        conn.execute_command('HPEXPIRE', f'doc:{i}', '1', 'FIELDS', '1', 'y') # doc:expire internal id is 5
```
DvirDukhan, line 514: I do not really understand those comments. you are digesting 1000 documents, which one of them is doc:expire and why it gets id 5?

tests/pytests/test_expire.py
```diff
@@ -499,6 +499,36 @@ def testDocWithLongExpiration(env):
...
+    # inverted index state
+    # 'hello': ['doc:0', 'doc:1', , ..., 'doc:1000', 'doc:1001']
+    # 'world': ['doc:1', , ..., 'doc:1000', 'doc:1001']
+    # 'foo': ['doc:0']
+
+    # expected flow
+    # - hello reader starts with doc:0
+    # - world reader starts with doc:1
+    # - intersect iterator reads doc:0 and tries to skip to it in world reader
+    # - world reader should skip to doc:1001 since all the other docs will be expired
```
DvirDukhan, line 525: isn't `hello` reader should skip to `1001` since `world` reader has been called with IR_SkipTo with `1`, uses `IndexReader_ReadWithSeeker` and returns `1001`?

tests/pytests/test_expire.py
```diff
@@ -499,6 +499,36 @@ def testDocWithLongExpiration(env):
     conn.execute_command('EXPIRE', 'doc:1', '30000')
     env.expect('FT.SEARCH', 'idx', 'hello', 'NOCONTENT').apply(sort_document_names).equal([2, 'doc:1', 'doc:2'])
 
+def testSeekToExpirationChecks(env):
```
DvirDukhan, line 502: can't we gtest?

src/inverted_index.c
kei-nan: Guy already commented on this, see:
https://github.com/RediSearch/RediSearch/pull/5355/files#r1886779547
Basically our unit tests can initialize a search ctx without a spec.

src/inverted_index.c
kei-nan: But then if we set found to false  in the first if we won't advance the block in cases we might need to advance it.
e.g the last id in the block is expired.

tests/pytests/test_expire.py
kei-nan: Don't we prefer flow tests over unit tests?
Also we want to merge this hopefully today.
We need this for the HFE completion.

src/inverted_index.c
kei-nan: yep, but `numFieldIndices ` should indicate in the end the number of valid entries.
I can memset each byte to be 0xff if you prefer.

src/inverted_index.c
kei-nan: I can document that this IR_SkipTo does not skip backward even if called with a lower number.
The can part depends on the caller or callee perspective.
Meaning we can get be asked to skip backward but we won't effectively do it.

I can only reliably document what this code does and not how it will be used.

tests/pytests/test_expire.py
kei-nan: missed that comment, remnants of old test.
Will update the comment.

src/inverted_index.c
kei-nan: Updated the comment, the problem with the naming of the function is that we have two predicates that mean different things.
So verify is the best name in that case IMHO.

src/inverted_index.c
kei-nan: Added a bunch of comment to this function.

src/inverted_index.c
```diff
@@ -962,6 +980,49 @@ int IR_Read(void *ctx, RSIndexResult **e) {
...
+      // the doc id is not valid, filter out the doc id and continue scanning
+      // we set docId to be the next doc id to search for to avoid infinite loop
+      // we rely on the doc id ordering inside the inverted index
+      // IMPORTANT:
+      // we still perform the AtEnd check to avoid the case the non valid doc id was at the end of the block
+      // block: [1, 4, 7, ..., 564]
+      //                        ^-- we are here, and 564 is not valid
+      found = false;
+      docId = ir->record->docId + 1;
```
raz-mon, line 1002: Given the transition to the next block in line 1007:1011, do we need this?

raz-mon (commented):
Looks good, small comments

src/inverted_index.c
```diff
@@ -893,6 +893,37 @@ size_t IR_NumEstimated(void *ctx) {
...
+  size_t numFieldIndices = 0;
+  // Use a stack allocated array for the field indices, if the field mask is not a single field
+  t_fieldIndex fieldIndicesArray[FIELD_MASK_BIT_COUNT];
+  t_fieldIndex* sortedFieldIndices = fieldIndicesArray;
+  if (ir->filterCtx.field.isFieldMask) {
+    const t_fieldMask relevantMask = docFieldMask & ir->filterCtx.field.value.mask;
+    numFieldIndices = IndexSpec_TranslateMaskToFieldIndices(ir->sctx->spec,
+                                                            relevantMask,
+                                                            fieldIndicesArray);
+  } else if (ir->filterCtx.field.value.index != RS_INVALID_FIELD_INDEX) {
```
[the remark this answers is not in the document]
kei-nan: I think there are cases where the mask is -1.
I think it was brought up in a previous PR, we looked in the code and discovered it can be RS_INVALID_FIELD_INDEX in some cases.

src/inverted_index.c
kei-nan: yes, the transition is optional
it only happens if we are at the end of the current block
if we are not then we could remain in the same block.

raz-mon: In that case, doesn't the `seeker` call already advance us to the id greater than the one we requested (if it did not find the one we requested)?

src/inverted_index.c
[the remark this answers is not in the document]
raz-mon: I see it is used in the wildcard iterator, debug command, minimalNumericReader, TermIndexReader 👍

tests/cpptests/test_cpp_forkgc.cpp
DvirDukhan, line 1: can you explain the changes in this file?

kei-nan: `TotalIIBlocks` is a global variable.
I assume it got affected by the newly added test which indexes documents and causes more blocks to exist.

tests/cpptests/test_cpp_expire.cpp
```diff
@@ -0,0 +1,80 @@
...
+  // should skip to last document
+  for (t_docId doc = 2; doc <= maxDocId; doc += 2) {
+    RSIndexResult *result = NULL;
+    // we index in hset and in hexpire
+    // for hset there won't be an expiration
+    // for hexpire there will be
+    // if we skip to an odd doc number we should get the requested doc id
+    // if we skip to an even doc number we should not get it since it will be expired
+    it->SkipTo(it->ctx, doc, &result);
+    ASSERT_EQ(result->docId, doc + 1);
```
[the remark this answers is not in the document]
kei-nan: thanks :)

src/inverted_index.c
kei-nan: The seeker only operates in the context of the current block.
It doesn't know how to advance to the next block.

DvirDukhan (dismissed):

oshadmi (dismissed):

raz-mon (dismissed):

raz-mon (approved):

DvirDukhan (dismissed):

tests/cpptests/test_cpp_expire.cpp
```diff
@@ -63,12 +63,12 @@ TEST_F(ExpireTest, testSkipTo) {
...
   ASSERT_EQ(it->LastDocId(it->ctx), 1);
   // should skip to last document, we index every doc twice so we should have 2 * maxDocId entries in the inverted index
-  for (t_docId doc = 2; doc <= (2 * maxDocId); doc += 2) {
+  for (t_docId doc = 2; doc < (2 * maxDocId); doc += 2) {
     RSIndexResult *result = NULL;
     // we index in hset and in hexpire
     // for hset there won't be an expiration
     // for hexpire there will be
-    // if we skip to an odd doc number we should get the requested doc id
+    // if we skip to an odd doc number we should gfet the requested doc id
```
oshadmi: Typo

tests/cpptests/test_cpp_expire.cpp
```diff
@@ -63,12 +63,12 @@ TEST_F(ExpireTest, testSkipTo) {
   IndexIterator *it = TagIndex_OpenReader(idx, sctx, "one", strlen("one"), 1.0, 0);
   ASSERT_EQ(it->LastDocId(it->ctx), 1);
   // should skip to last document, we index every doc twice so we should have 2 * maxDocId entries in the inverted index
-  for (t_docId doc = 2; doc <= (2 * maxDocId); doc += 2) {
+  for (t_docId doc = 2; doc < (2 * maxDocId); doc += 2) {
```
oshadmi, line 66: Should we not crash if a non-existing id is given to `SkipTo` ?

raz-mon (approved):

raz-mon (approved):

oshadmi (approved):

PR #4869, opened 2024-07-17
MOD-7253: Fix wildcard latency

raz-mon:
**Describe the changes in the pull request**

This PR aspires to fix the latency issue we have with our wildcard and NOT iterators, in which the doc-id increment is done one-by-one, resulting in a lot of time wasted on looking for un-existing documents.
The problem is encountered when there are many writes and deletions, such that there are big "holes" in the docIds of the existing docs in the database. This caused us to waste time incrementing the docIds and checking whether their corresponding documents exist.
We opt to fix this by adding an "existing-docs" inverted-index, holding the doc-ids of all the currently existing docs (up to GC update), which the wildcard and NOT iterators can utilize to efficiently jump between large doc-id deltas instead of the single-increment jumps used currently.
We use the existing inverted-index API so no new API is introduced.

This inverted index is cleaned by the GC, just like most of the other inverted indexes we hold.

**Mark if applicable**

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #5386, opened 2024-12-19
MOD-8192: Optimize OPTIONAL iterator with existing-index

raz-mon:
This PR incorporates the existing-index optimization done in #4869 for the OPTIONAL iterator, which is activated by passing  the `INDEXALL` argument in the index creation command.
It replaces the doc-id incremental counter with an iterator over the existing documents at the time of creation of the iterator.

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -1427,12 +1427,12 @@ static int OI_SkipTo_O(void *ctx, t_docId docId, RSIndexResult **hit) {
   RSIndexResult *wcii_res = NULL;
   if (docId > nc->wcii->LastDocId(nc->wcii->ctx)) {
     rc = nc->wcii->SkipTo(nc->wcii->ctx, docId, &wcii_res);
-    if (rc == INDEXREAD_EOF) {
-      IITER_SET_EOF(&nc->base);
-      return INDEXREAD_EOF;
-    } else if (rc == INDEXREAD_NOTFOUND) {
-      // This doc-id was deleted
-      return INDEXREAD_NOTFOUND;
+    if (rc != INDEXREAD_OK) {
+      if (rc != INDEXREAD_NOTFOUND) {
+        // EOF or timeout, set invalid
+        IITER_SET_EOF(&nc->base);
+      }
+      return rc;
     }
   }
 
@@ -1440,7 +1440,7 @@ static int OI_SkipTo_O(void *ctx, t_docId docId, RSIndexResult **hit) {
     // Has a real hit on the child iterator
     nc->base.current->weight = nc->weight;
   } else {
-    nc->virt->docId = nc->lastDocId = nc->wcii->LastDocId(nc->wcii->ctx);
+    nc->virt->docId = nc->lastDocId = docId;
     nc->virt->weight = 0;
     nc->base.current = nc->virt;
   }
@@ -1454,8 +1454,7 @@ static size_t OI_NumEstimated(void *ctx) {
   return nc->maxDocId;
 }
 
-/* Read has no meaning in the sense of an OPTIONAL iterator, so we just read the next record from
- * our child */
+// Read from an OPTIONAL iterator - Non-Optimized version.
 static int OI_ReadSorted_NO(void *ctx, RSIndexResult **hit) {
   OptionalMatchContext *nc = ctx;
   if (nc->lastDocId >= nc->maxDocId) {
@@ -1487,6 +1486,8 @@ static int OI_ReadSorted_NO(void *ctx, RSIndexResult **hit) {
   return INDEXREAD_OK;
 }
 
+// Read from optional iterator - Optimized version, utilizing the `existing docs`
+// inverted index.
 static int OI_ReadSorted_O(void *ctx, RSIndexResult **hit) {
   OptionalMatchContext *nc = ctx;
   if (nc->lastDocId >= nc->maxDocId) {
@@ -1496,9 +1497,10 @@ static int OI_ReadSorted_O(void *ctx, RSIndexResult **hit) {
   // Get the next docId
   RSIndexResult *wcii_res = NULL;
   int wcii_rc = nc->wcii->Read(nc->wcii->ctx, &wcii_res);
-  if (wcii_rc == INDEXREAD_EOF) {
+  if (wcii_rc != INDEXREAD_OK) {
+    // EOF, set invalid
     IITER_SET_EOF(&nc->base);
-    return INDEXREAD_EOF;
+    return wcii_rc;
   }
 
   int rc;
@@ -1507,7 +1509,6 @@ static int OI_ReadSorted_O(void *ctx, RSIndexResult **hit) {
     rc = nc->child->Read(nc->child->ctx, &nc->base.current);
     if (rc == INDEXREAD_EOF) {
       nc->nextRealId = nc->maxDocId + 1;
-      break;
     } else {
       nc->nextRealId = nc->base.current->docId;
     }
--- a/src/index.c
+++ b/src/index.c
@@ -1468,6 +1468,8 @@ static int OI_ReadSorted_NO(void *ctx, RSIndexResult **hit) {
     int rc = nc->child->Read(nc->child->ctx, &nc->base.current);
     if (rc == INDEXREAD_EOF) {
       nc->nextRealId = nc->maxDocId + 1;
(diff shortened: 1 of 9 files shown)


src/index.c
```diff
@@ -1308,60 +1305,82 @@ typedef struct {
...
   if (nc->lastDocId > nc->maxDocId) {
+    IITER_SET_EOF(&nc->base);
     return INDEXREAD_EOF;
   }
 
-  if (!nc->child) {
-    nc->virt->docId = docId;
-    nc->base.current = nc->virt;
-    return INDEXREAD_OK;
-  }
```
GuyAv46, line 1338: Dead code?
I guess we evaluate the iterator to be a simple wildcard in this case?

src/index.c
```diff
@@ -1372,14 +1391,72 @@ static int OI_SkipTo(void *ctx, t_docId docId, RSIndexResult **hit) {
...
+  RSIndexResult *wcii_res = NULL;
+  if (docId > nc->wcii->LastDocId(nc->wcii->ctx)) {
+    rc = nc->wcii->SkipTo(nc->wcii->ctx, docId, &wcii_res);
+    if (rc == INDEXREAD_EOF) {
+      IITER_SET_EOF(&nc->base);
+      return INDEXREAD_EOF;
+    } else if (rc == INDEXREAD_NOTFOUND) {
+      // This doc-id was deleted
+      return INDEXREAD_NOTFOUND;
+    }
```
GuyAv46, line 1436: Can any of these cases happen when we previously set `found = true`?

tests/cpptests/test_cpp_index.cpp
```diff
@@ -447,39 +447,39 @@ TEST_F(IndexTest, testPureNot) {
...
-    // printf("%d, ", h.docId);
-  }
-
-  ui->Free(ui);
-  // IndexResult_Free(&h);
-  InvertedIndex_Free(w);
-  InvertedIndex_Free(w2);
-}
+// // Note -- in test_index.c, this test was never actually run!
+// TEST_F(IndexTest, DISABLED_testOptional) {
```
GuyAv46, line 451: Noticed this one lately... Any reason it's disabled?
Any reason to keep it commented out and not remove it?

tests/pytests/test_issues.py
```diff
@@ -1209,6 +1209,7 @@ def test_mod_7495(env: Env):
 @skip(cluster=True)
 def test_mod_8142(env:Env):
   env.expect('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT').ok()
+  env.expect('FT.CREATE', 'idxOptimizedOptional', 'INDEXALL', 'ENABLE', 'SCHEMA', 't', 'TEXT').ok()
```
GuyAv46, line 1212: We also test the `NOT` iterator here, which also benefits from the `INDEXALL` optimization. Consider renaming the index:
```suggestion
  env.expect('FT.CREATE', 'idxOptimized', 'INDEXALL', 'ENABLE', 'SCHEMA', 't', 'TEXT').ok()
```
```suggestion
  env.expect('FT.CREATE', 'idxIndexAll', 'INDEXALL', 'ENABLE', 'SCHEMA', 't', 'TEXT').ok()
```

tests/pytests/test_scorers.py
```diff
@@ -293,29 +300,37 @@ def testScoreError(env):
...
+        env.expect('FT.SEARCH', 'idx', '~hello', 'ADDSCORES').error().equal('ADDSCORES is not supported on FT.SEARCH')
 
-    env.expect('FT.SEARCH', 'idx', '~hello', 'ADDSCORES').error().equal('ADDSCORES is not supported on FT.SEARCH')
+    for idx_def in [
+        ['FT.CREATE', 'idx', 'SCHEMA', 'title', 'TEXT'],
+        ['FT.CREATE', 'idx', 'INDEXALL', 'ENABLE', 'SCHEMA', 'title', 'TEXT']
+    ]:
+        env.expect(*idx_def).ok()
+        test_score(env, conn)
+        env.flush()
```
GuyAv46, line 336: Consider actually splitting the test into 2 tests, so we know which version has failed in case of a failure

GuyAv46 (commented):
Most optimal
A few comments

src/index.c
raz-mon: This is how we create a new iterator:
`nc->child = it ? it : NewEmptyIterator();`
So yes, dead code, and yes, it results in a simple non-optimized wildcard iterator

src/index.c
raz-mon: Yes, if the found doc was deleted, and the GC did not update the inverted index yet

tests/cpptests/test_cpp_index.cpp
raz-mon: Well, I think it can be removed since it was disabled for so much time (2018).
Will remove

src/index.c
raz-mon: I'll add a test with deleted docs (wanted to do it anyway..)

tests/pytests/test_scorers.py
raz-mon: Yeah thought about it, but figured a simple removal of the problematic index will clarify so it's not an issue.
The problem is flaky tests. I'll just create another index and run all queries against both, as I did in `test_issues`

tests/pytests/test_scorers.py
```diff
@@ -292,30 +299,36 @@ def testScoreError(env):
     env.expect('ft.add idx doc1 0.01 fields title hello').ok()
     env.expect('ft.search idx hello EXPLAINSCORE').error().contains('EXPLAINSCORE must be accompanied with WITHSCORES')
 
-def testExposeScore(env: Env):
-    env.expect('FT.CREATE idx ON HASH SCHEMA title TEXT').ok()
-    with env.getClusterConnectionIfNeeded() as conn:
-        conn.execute_command('HSET', 'doc1', 'title', 'hello')
+def _test_score(env, idx):
```
GuyAv46, line 302:
```suggestion
def _test_expose_score(env: Env, idx):
```

tests/pytests/test_existing.py
```diff
@@ -59,3 +59,55 @@ def test_existing_GC():
...
+    # Remove some docs
+    for i in range(1, n_docs + 1):       # Doc{i} matches doc with id = i
+        if i % 2 == 0:
+            env.cmd('DEL', f'doc{i}')
+
+    # Sanity check
+    env.assertEqual(env.cmd(debug_cmd(), 'DUMP_INVIDX', 'idx', 'hello'), [i for i in range(1, 11)])
+
+    # Apply the GC, to clean the deleted docs from the inverted indexes
+    env.expect(debug_cmd(), 'GC_FORCEINVOKE', 'idx').equal('DONE')
```
GuyAv46, line 90: Consider running the search query before the GC cleanup as well

raz-mon: Why?
Doesn't seem relevant to the test. I mainly wanted to make sure we have jumps in the doc-ids

GuyAv46: Is there a way to simulate the scenario of the wildcard iterator being up-to-date while the child iterator is not?

raz-mon: I don't think we can do it since it requires a specific timing that is hard to control:
* DocId `10` is in the last block of the index of `hello`, and we write a new entry to this index **WHILE** the GC is working (i.e., after the fork).
* DocId `10` is **NOT** in the last block of the existing-docs index, such that the additional document we added while the GC is working doesn't cause `10` to remain in this index as well.
* Optional\NOT query.

GuyAv46 (approved):

src/index.c
```diff
@@ -1410,17 +1486,53 @@ static int OI_ReadSorted(void *ctx, RSIndexResult **hit) {
...
+    // EOF, set invalid
+    IITER_SET_EOF(&nc->base);
+    return wcii_rc;
+  }
+
+  int rc;
+  // We loop over this condition, since it reflects that the index is not up to date.
+  while (wcii_res->docId > nc->nextRealId) {
+    rc = nc->child->Read(nc->child->ctx, &nc->base.current);
+    if (rc == INDEXREAD_EOF) {
```
kei-nan, line 1510: Consider handling the case of a timeout being returned from the child iterator.

kei-nan (approved):

PR #4733, opened 2024-06-10
MOD-7240: Detect escape state when tokenizing

oshadmi:
**Describe the changes in the pull request**
Fix #4732 

**Mark if applicable**

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #5391, opened 2024-12-21
MOD-7944: Support multi-byte char terms

### LINKED ISSUE #2015: Documents having uppercase fields aren't returned when using lower case letters for TEXT field search using FT.search in Redisearch
I am using Redisearch module  FT.search, trying to search for records querying TEXT field using below query for value like 'test' or 'Test'. It returns all the documents with lower case values like 'test or Test' and no data which has all the upper case letters 'TEST' are being returned.

FT.search MyIndex @MyField:"test" or FT.search MyIndex @MyField:"Test"

Am I missing anything here?

### LINKED ISSUE #4886: [BUG] FT.SEARCH word prefix is not working with cyrillic texts - [MOD-6407]
**Describe the bug**
When working with the cyrillic based texts, I noticed that the word prefix queries does not search any records for me. Another issue, a provided search term must be case-sensitive. On the other hand, word suffix works well but it is case-sensitive too.

**To Reproduce**
Setup a hash:
```
hset test:1 Ref "09add9ee-2544-11ee-a361-48df37b92096" Description "Абрикосове"
hset test:2 Ref "1b009444-4e4a-11ed-a361-48df37b92096" Description "Абранка"
hset test:3 Ref "6dbe932e-1aad-11ea-8c15-0025b502a06e" Description "Абрикосівка"
```

Create an index:
```
ft.create idx on hash prefix 1 test: schema Ref as ref tag Description as desc text
```

Try to search records with word prefixes:
```
FT.SEARCH idx "@desc:Абр*"
FT.SEARCH idx "@desc:Абрикос*"
```
Case-insensitive search is also does not give any result:
```
FT.SEARCH idx "@desc:абранка"
```
All above commands are ended with `No results found.` message.
Only a full case-sensitive search is working: `FT.SEARCH idx "@desc:Абранка"`.

**Expected behavior**
Found the records with word prefix and case-insensitive search.

**Environment (please complete the following information):**
 - `redis/redis-stack:latest` (7.2.5) docker image
 - redis search: 2.8.14

nafraf:

1. Current:
   `DefaultNormalize()` and `rm_strdupcase()` don't support multi-byte characters, and the conversion to lowercase of multi-byte characters are generating different terms for the same word with if written with different case.
   
2. Change:
    We implemented unicode_tolower() which converts multi-byte char strings to lowercase case, it uses the nunicode library that is already used in rune_util.

Function rm_strdupcase() was renamed to rm_normalize(), the conversion to lowercase is done using unicode_tolower().

To avoid breaking changes related to the suggestion dictionary, the old function strToFoldedRunes() was renamed to strToSingleCodepointFoldedRunes() and it is used only for the suggestions.
   

3. Outcome:
   Fix queries for prefix/contains/suffix using multi-byte characters.
   Multi-byte characters are supported in TEXT and TAG fields.
   Multi-byte char synonyms are supported.
   
   This PR does not include:
   - Multi-byte char stopwords
       - See `testStopWords()` in `tests/pytests/test_multibyte_char_terms.py` 
   - Diacritics removing during normalization
      - See `testDiacriticLimitation()` in `tests/pytests/test_multibyte_char_terms.py` 
   - Changes to multi-byte char suggestions
     - See `testSuggestions()` in `tests/pytests/test_multibyte_char_terms.py` 
   

1. MOD-7944
2. Fixes #2015 
3. Fixes #4886 

1. ...

#### Mark if applicable


src/tokenize.c
```diff
@@ -72,6 +69,73 @@ static char *DefaultNormalize(char *s, char *dst, size_t *len) {
   return dst;
 }
 
+static char *DefaultNormalize_utf8(char *s, char *dst, size_t *len) {
+  // TODO: set locale depending on the index language?
```
raz-mon, line 73: Can we address this `TODO`?

src/tokenize.c
```diff
@@ -72,6 +69,73 @@ static char *DefaultNormalize(char *s, char *dst, size_t *len) {
...
+  int escaped = 0;
+  mbstate_t state;
+  memset(&state, 0, sizeof(state));
+  wchar_t wc;
+  size_t len_wc;
+
+  for (size_t ii = 0; ii < origLen;) {
+    len_wc = mbrtowc(&wc, &s[ii], MB_CUR_MAX, &state);
+    if (len_wc == (size_t)-1 || len_wc == (size_t)-2) {
+      // Handle invalid multi-byte sequence
```
raz-mon, line 95: Is there a simple example for this that we can test (for coverage), or is this a highly edge-case?

src/tokenize.c
```diff
@@ -72,6 +69,73 @@ static char *DefaultNormalize(char *s, char *dst, size_t *len) {
...
+  *len = dstLen;
+  return dst;
+}
+
+/**
+ * Normalizes text.
+ * - s contains the raw token
+ * - dst is the destination buffer which contains the normalized text
+ * - len on input contains the length of the raw token. on output contains the
+ * on output contains the length of the normalized token
```
raz-mon, line 129:
```suggestion
 * - len on input contains the length of the raw token. On output contains the length of the normalized token
```

src/tokenize.c
```diff
@@ -72,6 +69,73 @@ static char *DefaultNormalize(char *s, char *dst, size_t *len) {
...
+ */
+static char *DefaultNormalize(char *s, char *dst, size_t *len) {
+  if (setlocale(LC_ALL, "en_US.UTF-8") != NULL) {
```
raz-mon, line 132: Why would this not succeed (according to the docs)?
I think we should raise an error in that case, in order to have consistent behavior and not switch behaviors under the hood without the user knowing why (and us probably, as there is no log as well)

src/tokenize.c
```diff
@@ -72,6 +69,73 @@ static char *DefaultNormalize(char *s, char *dst, size_t *len) {
   return dst;
 }
 
+static char *DefaultNormalize_utf8(char *s, char *dst, size_t *len) {
```
raz-mon, line 72: It would be really nice to have some clear documentation in this function!

src/util/strconv.h
```diff
@@ -106,8 +109,7 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
   return ret;
 }
 
-// strndup + lowercase in one pass!
-static char *rm_strdupcase(const char *s, size_t len) {
+static char *rm_strdupcase_singlechar(const char *s, size_t len) {
```
raz-mon, line 112:
```suggestion
static char *rm_strdupcase_singleBytechars(const char *s, size_t len) {
```

src/util/strconv.h
```diff
@@ -127,4 +129,61 @@ static char *rm_strdupcase(const char *s, size_t len) {
...
+  memset(&state, 0, sizeof(state));
+  wchar_t wc;
+  size_t len_wc;
+
+  while (*src) {
+    len_wc = mbrtowc(&wc, src, MB_CUR_MAX, &state);
+    if (len_wc == (size_t)-1 || len_wc == (size_t)-2) {
+      // Handle invalid multi-byte sequence
+      src++;
+      continue;
```
raz-mon, line 154: What's invalid here?
It would be great to have coverage for this if it is easy to produce. If not - add some logging at least?
Let's understand when this will create odd behavior, and treat it accordingly (i.e., if this will never happen - ok, otherwise this needs better handling)

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -0,0 +1,227 @@
...
+
+def is_locale_available(locale_name):
+    try:
+        locale.setlocale(locale.LC_ALL, locale_name)
+        return True
+    except locale.Error:
+        return False
+
+def testMultibyteChars(env):
+    if not is_locale_available('en_US.UTF-8'):
```
raz-mon, line 16: I see you add a lot of conditions for the availability of `locale` - why should it not be available?

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -0,0 +1,227 @@
...
+        res = conn.execute_command(
+            'FT.SEARCH', 'idx', '@t:БЪлга123', 'NOCONTENT')
+        env.assertEqual(res, expected)
+
+        # Search with prefix
+        res = conn.execute_command(
+            'FT.SEARCH', 'idx', '@t:БЪЛГА*', 'NOCONTENT')
+        env.assertEqual(res, expected)
+
+def testRussianAlphabet(env):
```
raz-mon, line 67: 😮

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -0,0 +1,227 @@
...
+    conn.execute_command('HSET', 'test:17', 't', 'æ')
+    conn.execute_command('HSET', 'test:18', 't', 'Æ')
+
+    if not env.isCluster():
+        res = env.cmd(debug_cmd(), 'DUMP_TERMS', 'idx')
+        # only 9 terms are indexed, the lowercase representation of the terms
+        # with diacritics, but the diacritis are not removed.
+        env.assertEqual(len(res), 9)
+    
+def testDiacriticLimitation(env):
```
raz-mon, line 155: Please add documentation as to what the test is testing (for the other tests as well). It's a nice TL;DR and in general can explain some "weird" stuff that happens in the test

raz-mon (commented):
Very nice 💪 
See some comments and suggestions

tests/pytests/test_multibyte_char_terms.py
nafraf: Some docker images don't include it. i.e.: In the CI, the `ubuntu:jammy` image, used for the sanitizer, don't include it.

src/tokenize.c
nafraf: It will fail if the selected locale is not found. The OS could not include it.
I'm not sure if we should raise an error or maybe add a configuration param to let the user to decide to use multi-byte or not.

src/query_parser/v1/parser.y
```diff
@@ -49,27 +49,6 @@
 #include "rmutil/vector.h"
 #include "query_node.h"
 
-// strndup + lowercase in one pass!
-static char *strdupcase(const char *s, size_t len) {
```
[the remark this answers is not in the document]
nafraf: Yes. It was duplicated.

src/stopwords.c
```diff
@@ -79,7 +79,7 @@ StopWordList *NewStopWordListCStr(const char **strs, size_t len) {
     // lowercase the letters
     for (size_t pos = 0; pos < tlen; pos++) {
       if (isalpha(t[pos])) {
-        t[pos] = tolower(t[pos]);
+        t[pos] = tolower(t[pos]);  // TODO: Support multi-byte UTF-8 characters?
```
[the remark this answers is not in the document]
nafraf: OK. Ticket: MOD-8443

src/tokenize.c
```diff
@@ -33,14 +37,7 @@ static void simpleTokenizer_Start(RSTokenizer *base, char *text, size_t len, uin
...
 
-/**
- * Normalizes text.
- * - s contains the raw token
- * - dst is the destination buffer which contains the normalized text
- * - len on input contains the length of the raw token. on output contains the
- * on output contains the length of the normalized token
- */
-static char *DefaultNormalize(char *s, char *dst, size_t *len) {
+static char *DefaultNormalize_singlechar(char *s, char *dst, size_t *len) {
```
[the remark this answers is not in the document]
nafraf: OK. Function renamed.

src/util/strconv.h
nafraf: OK. Function renamed as `rm_strdupcase_singleByteChars`

src/tokenize.c
nafraf: I added tests with invalid sequences, but those cases are not seen there. What if we add an assertion?

src/util/strconv.h
nafraf: I added a new tests, but the search for invalid sequences are reported as invalid syntax. What if we add an assert here also?

src/tokenize.c
raz-mon: Do you have any idea when this can happen?

src/tokenize.c
raz-mon: I think consistency is key here. We can't produce different results depending on the OS

src/util/strconv.h
raz-mon: An assertion sounds good to me, but let's try to understand when this should happen. If it shouldn't, an assertion is good.

tests/pytests/test_multibyte_char_terms.py
raz-mon: That's a problem. We have to be consistent in our responses, OS-independent.
Can we add it in our installation flow?

src/tokenize.c
```diff
@@ -54,24 +59,109 @@ static char *DefaultNormalize(char *s, char *dst, size_t *len) {
...
+ * Normalize a UTF-8 string, converting it to lower case and removing control 
+ * characters.
+ * - s contains the raw token
+ * - dst is the destination buffer which contains the normalized text
+ * - len on input contains the length of the raw token. On output contains the
+ *   length of the normalized token
+ */
+static char *DefaultNormalize_utf8(char *s, char *dst, size_t *len) {
+  // TODO: set locale depending on the index language?
+  setlocale(LC_ALL, "en_US.UTF-8");
```
kei-nan, line 96: should we reset the local upon exiting the function?
This is going to affect the entire process, no?
Might also propagate to redis and other modules.

src/tokenize.c
```diff
@@ -54,24 +59,109 @@ static char *DefaultNormalize(char *s, char *dst, size_t *len) {
...
+    // Reset the escape flag
     escaped = 0;
   }
 
   *len = dstLen;
   return dst;
 }
 
+static char *DefaultNormalize(char *s, char *dst, size_t *len) {
+  if (setlocale(LC_ALL, "en_US.UTF-8") != NULL) {
```
kei-nan, line 158: Also worried about this affecting the rest of the process.

src/util/strconv.h
```diff
@@ -127,4 +129,61 @@ static char *rm_strdupcase(const char *s, size_t len) {
...
+  }
+
+  *dst = '\0';
+
+  return ret;
+}
+
+// strndup + lowercase in one pass!
+static char *rm_strdupcase(const char *s, size_t len) {
+  if (setlocale(LC_ALL, "en_US.UTF-8") != NULL) {
```
kei-nan, line 182: maybe add a config command to set the locale?
Don't think this should keep overriding the existing local.
It might cause modules to fight on the locale the process will use.

Also in the documentation it says modifying the local in one thread while another is using the locale is undefined behaviour.
Not sure if we should use a global function to determine what a local function in the context of a thread will choose to do.

kei-nan (changes requested):
Added some comments, mostly on the use of setlocale

src/util/strconv.h
kei-nan: Consider using `newlocale` instead.

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -0,0 +1,265 @@
...
+def is_locale_available(locale_name):
+    try:
+        locale.setlocale(locale.LC_ALL, locale_name)
+        return True
+    except locale.Error:
+        return False
+
+def testMultibyteChars(env):
+    if not is_locale_available('en_US.UTF-8'):
+        env.skip()
```
DvirDukhan, line 17: I don't think we can allow ourselves to skip

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -0,0 +1,265 @@
...
+    conn.execute_command('HSET', 'test:1', 't', 'abcabc')
+    conn.execute_command('HSET', 'test:2', 't', 'ABCABC')
+    conn.execute_command('HSET', 'test:upper', 't', 'БЪЛГА123') # uppercase
+    conn.execute_command('HSET', 'test:lower', 't', 'бълга123') # lowercase
+    conn.execute_command('HSET', 'test:mixed', 't', 'БЪлга123') # mixed case
+
+    if not env.isCluster():
+        # only 2 terms are indexed, the lowercase representation of the terms
+        res = env.cmd(debug_cmd(), 'DUMP_TERMS', 'idx')
+        env.assertEqual(len(res), 2)
```
DvirDukhan, line 32: please also validate the terms

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -0,0 +1,265 @@
...
+
+    # test with russian uppercase stopwords.
+    env.cmd('FT.CREATE', 'idx2', 'ON', 'HASH', 'LANGUAGE', 'RUSSIAN',
+            'STOPWORDS', 3, 'ОТ', 'И', 'НЕ',
+            'SCHEMA', 't', 'TEXT', 'NOSTEM')
+    waitForIndex(env, 'idx2')
+    # This fails, there are 9 terms because the stopwords are not converted
+    # to lowercase correctly
+    # Ticket created to fix this: MOD-8443
+    res = env.cmd(debug_cmd(), 'DUMP_TERMS', 'idx2')
```
DvirDukhan, line 232: I think you need to validate terms and also query

tests/pytests/test_multibyte_char_terms.py
DvirDukhan, line 1: nice!
do we need to do it also for tags?

.github/workflows/event-pull_request.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       test-config: QUICK=1
       san: address
       env: ubuntu-latest
-      container: ubuntu:jammy
+      container: ubuntu:latest
```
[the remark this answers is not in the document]
nafraf: `ubuntu:jammy` does not include `en_US.utf8` locale by default. I changed only to check memory sanitizer with the new utf functions, but I'll revert that change.

```c
Ubuntu:latest - available locales:
C
C.utf8
POSIX
en_US.utf8


Ubuntu:jammy - available locales:
C
C.utf8
POSIX
```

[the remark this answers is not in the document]
nafraf: Change reverted.

tests/pytests/test_multibyte_char_terms.py
nafraf: OK. Done.

tests/pytests/test_multibyte_char_terms.py
nafraf: OK. Test modified.

src/util/strconv.h
nafraf: The code was modified, now we are using nunicode library.

src/tokenize.c
nafraf: Done. Now we don't depend on locale.

src/tokenize.c
nafraf: We don't need it now. The code was changed to use nunicode library.

tests/pytests/test_multibyte_char_terms.py
nafraf: Now we don't need to install them.

tests/pytests/test_multibyte_char_terms.py
nafraf: OK. Now we don't skip tests.

src/tokenize.c
```diff
@@ -34,16 +38,17 @@ static void simpleTokenizer_Start(RSTokenizer *base, char *text, size_t len, uin
...
- * - len on input contains the length of the raw token. on output contains the
- * on output contains the length of the normalized token
+ * - len on input contains the length of the raw token. On output contains the 
+ *   length of the normalized token
  */
-static char *DefaultNormalize(char *s, char *dst, size_t *len) {
-  size_t origLen = *len;
-  char *realDest = s;
-  size_t dstLen = 0;
+static char *DefaultNormalize_singleByteChars(char *s, char *dst, size_t *len) {
```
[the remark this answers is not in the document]
nafraf: This function was removed

src/util/strconv.h
nafraf: Thanks. We opted for a different solution which does not required to change the locale.

tests/pytests/test_multibyte_char_terms.py
nafraf: Yes. I added changes to support tags also.

src/tokenize.c
nafraf: The code was modified, now we don't change the locale

src/tokenize.c
nafraf: Thanks. I removed those changes.

src/query_parser/v1/parser.y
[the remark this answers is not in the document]
nafraf: Now v2 parser is part of the PR. Tests are done using the 4 dialects.

src/tokenize.c
nafraf: OK. Fixed. Thanks.

src/tokenize.c
```diff
@@ -34,16 +38,17 @@ static void simpleTokenizer_Start(RSTokenizer *base, char *text, size_t len, uin
...
+  size_t origLen = *len;  // original length of the token
+  char *realDest = s;     // pointer to the current destination buffer
+  size_t dstLen = 0;      // length of the destination buffer
```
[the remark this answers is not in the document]
nafraf: With the new approach using unicode, all the functions are covered.

src/util/strconv.h
```diff
@@ -106,8 +109,7 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
   return ret;
 }
 
-// strndup + lowercase in one pass!
-static char *rm_strdupcase(const char *s, size_t len) {
+static char *rm_strdupcase_singleByteChars(const char *s, size_t len) {
```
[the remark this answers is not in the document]
nafraf: Solved. We are using a single function.

src/util/strconv.h
```diff
@@ -127,4 +129,61 @@ static char *rm_strdupcase(const char *s, size_t len) {
   return ret;
 }
 
-#endif
\ No newline at end of file
+static char *rm_strdupcase_utf8(const char *s, size_t len) {
+  // TODO: set locale depending on the index language?
+  setlocale(LC_ALL, "en_US.UTF-8");
```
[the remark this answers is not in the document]
nafraf: Removed. We don't change the locale.

src/util/strconv.h
```diff
@@ -127,4 +129,61 @@ static char *rm_strdupcase(const char *s, size_t len) {
...
+
+  // Allocate memory for the destination string
+  char *ret = rm_strndup(s, len);
+  if (!ret) {
+    return NULL;
+  }
+
+  char *dst = ret;
+  char *src = dst;
```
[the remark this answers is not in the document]
nafraf: OK. Done.

src/util/strconv.h
```diff
@@ -127,4 +129,61 @@ static char *rm_strdupcase(const char *s, size_t len) {
...
+  mbstate_t state;
+  memset(&state, 0, sizeof(state));
+  wchar_t wc;
+  size_t len_wc;
+
+  while (*src) {
+    len_wc = mbrtowc(&wc, src, MB_CUR_MAX, &state);
+    if (len_wc == (size_t)-1 || len_wc == (size_t)-2) {
```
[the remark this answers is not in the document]
nafraf: Removed. We don't use this approach.

src/util/strconv.h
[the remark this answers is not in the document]
nafraf: Yes. Removed.

src/tokenize.c
```diff
@@ -54,24 +59,109 @@ static char *DefaultNormalize(char *s, char *dst, size_t *len) {
...
+    ii += len_wc;
+    // Reset the escape flag
     escaped = 0;
   }
 
   *len = dstLen;
   return dst;
 }
 
+static char *DefaultNormalize(char *s, char *dst, size_t *len) {
```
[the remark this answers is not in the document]
nafraf: There are some differences:

`DefaultNormalize()`:
- ignores control characters
- does not allocate memory for the destination

`rm_strdupcase()`:
- does not handle control characters, because the lexer filter them.
- allocate memory for the result

```sh
127.0.0.1:6379> FT.CREATE idx schema t text
OK
# This result in a term "hiworld"
127.0.0.1:6379> hset doc1  t "hi\nworld" 
(integer) 1
# You need to remove the control char to find it
127.0.0.1:6379> FT.SEARCH idx "hiworld"
1) (integer) 1
2) "doc1"
3) 1) "t"
   2) "hi\nworld"
# if you try to search with a control character, 
# the string will be split in two terms
127.0.0.1:6379> FT.SEARCH idx "hi\nworld"
1) (integer) 0
127.0.0.1:6379> FT.explaincli idx "hi\nworld"
 1) INTERSECT {
 2)   UNION {
 3)     hi
 4)     +hi(expanded)
 5)   }
 6)   UNION {
 7)     world
 8)     +world(expanded)
 9)   }
10) }
```

src/util/strconv.h
```diff
@@ -106,11 +110,97 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
   return ret;
 }
 
+// transform utf8 string to lowercase using nunicode library
+// encoded: the utf8 string to transform
+// in_len: the length of the utf8 string
+// dst: the destination buffer to store the transformed string
+// returns the bytes written to encoded, or 0 if the length of the transformed
+// string is greater than in_len and no transformation was done
+static size_t nunicode_tolower(const char *encoded, size_t in_len, char *dst) {
```
oshadmi, line 119: Nitpick: Consider renaming the function to not mention the specific library name, e.g., `unicode_tolower`

src/util/strconv.h
```diff
@@ -20,6 +21,9 @@
 // Case sensitive string equal
 #define STR_EQ(str, len, other) (len == strlen(other) && !strncmp(str, other, len))
 
+// Threshold for Small String Optimization (SSO)
+#define SSO_MAX_LENGTH 16
```
oshadmi, line 25: Any reason for this threshold? Can we increase it? e.g., what about 32, 64, or 128?

src/util/strconv.h
```diff
@@ -106,11 +110,97 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
...
+  return reencoded_len;
+}
+
+
 // strndup + lowercase in one pass!
 static char *rm_strdupcase(const char *s, size_t len) {
   char *ret = rm_strndup(s, len);
   char *dst = ret;
-  char *src = dst;
+  char *src = ret;
```
oshadmi, line 203: Not necessarily in this PR but do we need to maintain the escape state? For example see PR #4733

src/util/strconv.h
```diff
@@ -106,11 +110,97 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
...
+  uint32_t u_stack_buffer[SSO_MAX_LENGTH * 2];
+  uint32_t *u_buffer = u_stack_buffer;
+
+  if (in_len == 0) {
+    return 0;
+  }
+
+  // create a zero-terminated copy of the encoded string
+  if (in_len >= SSO_MAX_LENGTH - 1) {
+    encoded_copy = rm_strndup(encoded, in_len);
```
oshadmi, line 131: Can we avoid duplicating the string pointed by `encoded` and work on it directly? (since we are always using it as destination as well and we are already "destructive" and its size can only decrease from unescaping)

src/util/strconv.h
```diff
@@ -106,11 +110,97 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
...
+    }
+
+    // Break if the end of the string is reached
+    if (codepoint == 0) {
+      break;
+    }
+  }
+
+  // Encode Unicode codepoints back to utf8 string
+  ssize_t reencoded_len = nu_bytelen(u_buffer, nu_utf8_write);
```
oshadmi, line 180: In sample (https://bitbucket.org/alekseyt/nunicode/src/master/samples/folding.c)
it is not done (using `nu_utf8_write`)

tests/pytests/test_cn.py
```diff
@@ -49,7 +49,7 @@ def testMixedHighlight(env):
     txt = r"""
 Redis支持主从同步。数据可以从主服务器向任意数量的从服务器上同步，从服务器可以是关联其他从服务器的主服务器。这使得Redis可执行单层树复制。从盘可以有意无意的对数据进行写操作。由于完全实现了发布/订阅机制，使得从数据库在任何地方同步树时，可订阅一个频道并接收主服务器完整的消息发布记录。同步对读取操作的可扩展性和数据冗余很有帮助。[8]
 """
-    env.cmd('ft.create', 'idx', 'ON', 'HASH', 'LANGUAGE_FIELD', 'chinese', 'schema', 'txt', 'text')
```
oshadmi, line 52: Can we also keep this?

src/util/strconv.h
nafraf: Yes. Done in [b71f865](https://github.com/RediSearch/RediSearch/pull/5391/commits/b71f8656fa443d46a814923f21c05adf009a0ecb)

src/util/strconv.h
nafraf: OK. Done in [b71f865](https://github.com/RediSearch/RediSearch/pull/5391/commits/b71f8656fa443d46a814923f21c05adf009a0ecb)

tests/pytests/test_cn.py
nafraf: OK. Done.

src/util/strconv.h
```diff
@@ -106,11 +110,97 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
...
+    // do not use heap allocation for short strings
+    memcpy(encoded_stack_buffer, encoded, in_len);
+    encoded_stack_buffer[in_len] = '\0';
+  }
+
+  const char *encoded_char = encoded_copy;
+  ssize_t u_len = nu_strtransformnlen(encoded_copy, in_len, nu_utf8_read,
+                                              nu_tolower, nu_casemap_read);
+
+  if (u_len >= (SSO_MAX_LENGTH * 2) - 1) {
```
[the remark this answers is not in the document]
nafraf: I fixed it. We don't need to multiply, we know the unicode length, and we are allocating `uint32_t` instead of bytes.

src/util/strconv.h
nafraf: Because the average length of an English word is less than 16. But tokens can be larger than words, so I changed it now to 128.

src/util/strconv.h
nafraf: OK. Done.

src/util/strconv.h
nafraf: I think we need to write to the intermediate unicode buffer `u_buffer` because we need to check that we don't exceed the length of the original `utf8` string

src/util/strconv.h
```diff
@@ -169,25 +158,17 @@ static size_t nunicode_tolower(const char *encoded, size_t in_len, char *dst) {
...
-    // Break if the end of the string is reached
-    if (codepoint == 0) {
-      break;
-    }
   }
 
   // Encode Unicode codepoints back to utf8 string
-  ssize_t reencoded_len = nu_bytelen(u_buffer, nu_utf8_write);
+  ssize_t reencoded_len = nu_bytenlen(u_buffer, u_len, nu_utf8_write);
   if (reencoded_len > 0 && reencoded_len <= in_len) {
```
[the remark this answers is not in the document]
oshadmi: Confirmed this cannot happen (programatically checked all Unicode code points)

src/util/strconv.h
```diff
@@ -106,25 +110,98 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
...
+  if (u_buffer != u_stack_buffer) {
+    rm_free(u_buffer);
+  }
+
+  return reencoded_len;
+}
+
+
 // strndup + lowercase in one pass!
 static char *rm_strdupcase(const char *s, size_t len) {
```
oshadmi, line 181: Seems we should rename this function since it is also unescaping and lowering case

src/query.c
```diff
@@ -1309,6 +1310,7 @@ static IndexIterator *query_EvalSingleTagNode(QueryEvalCtx *q, TagIndex *idx, Qu
 
   switch (n->type) {
     case QN_TOKEN: {
+      n->tn.len = strlen(n->tn.str);
```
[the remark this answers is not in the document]
nafraf: It was moved inside each case, modifying the string/token according to the type.

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -0,0 +1,1171 @@
...
+            'FT.SEARCH', 'idx', "@t:(%%CRÜßET%%)", 'NOCONTENT', 'SORTBY', 't')
+        env.assertEqual(res, expected)
+
+        res = conn.execute_command(
+            'FT.SEARCH', 'idx', "@t:(%%crüßet%%)", 'NOCONTENT', 'SORTBY', 't')
+        env.assertEqual(res, expected)
+
+        # No results, the distance is > 1 because a multibyte char was replaced
+        res = conn.execute_command(
+            'FT.SEARCH', 'idx', "@t:(%grüXen%)", 'NOCONTENT', 'SORTBY', 't')
```
oshadmi, line 146: Need to understand if folding `ẞ` to `(s)(s)` requires support in the fuzzy logic

src/util/strconv.h
nafraf: Function renamed to `rm_normalize()`

tests/pytests/test_multibyte_char_terms.py
nafraf: I think it is a bug. The distance between `grüXen` and `grüßen` is 1.
We can modify the normalization: remove the transformation to folding and transform to lowercase only.
The drawback is that we'll go back to have to different terms for words written with `ß` and `ss`

oshadmi: So let's revert using folding and go back to transforming to lower - until we fix the fuzzy logic to handle it properly (and maybe also offsets data?)

src/query.c
```diff
@@ -1277,54 +1277,67 @@ static IndexIterator *Query_EvalTagWildcardNode(QueryEvalCtx *q, TagIndex *idx,
...
   switch (n->type) {
     case QN_TOKEN: {
+      tag_strtofold(n->tn.str, &n->tn.len, caseSensitive);
+      n->tn.len = strlen(n->tn.str);
       ret = TagIndex_OpenReader(idx, q->sctx, n->tn.str, n->tn.len, weight, fs->index);
       break;
     }
     case QN_PREFIX:
+      tag_strtofold(n->pfx.tok.str, &n->pfx.tok.len, caseSensitive);
+      n->pfx.tok.len = strlen(n->pfx.tok.str);
```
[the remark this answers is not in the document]
nafraf: OK. Done.

src/query.c
```diff
@@ -1057,15 +1057,46 @@ static IndexIterator *Query_EvalUnionNode(QueryEvalCtx *q, QueryNode *qn) {
 
 typedef IndexIterator **IndexIteratorArray;
 
+static void tag_strtofold(char *str, size_t *len, int caseSensitive) {
```
oshadmi, line 1060: Update function name since it is doing tolower now
```suggestion
static void tag_strtolower(char *str, size_t *len, int caseSensitive) {
```

src/util/strconv.h
```diff
@@ -106,25 +110,98 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
...
+
+  const char *encoded_char = encoded;
+  ssize_t u_len = nu_strtransformnlen(encoded, in_len, nu_utf8_read,
+                                              nu_tolower, nu_casemap_read);
+
+  if (u_len >= (SSO_MAX_LENGTH - 1)) {
+    u_buffer = (uint32_t *)rm_malloc(sizeof(*u_buffer) * (u_len + 1));
+  }
+
+  // Decode utf8 string into Unicode codepoints and transform to fold
```
oshadmi, line 135: Doing tolower now instead of fold
```suggestion
  // Decode utf8 string into Unicode codepoints and transform to lower
```

src/util/strconv.h
```diff
@@ -106,25 +110,98 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
...
+  uint32_t codepoint;
+  unsigned i = 0;
+  for (ssize_t j = 0; j < u_len; j++) {
+    // Read unicode codepoint from utf8 string
+    encoded_char = nu_utf8_read(encoded_char, &codepoint);
+    // Transform unicode codepoint to case fold
```
oshadmi, line 141: Doing tolower now instead of fold
```suggestion
    // Transform unicode codepoint to lower case
```

src/query.c
```diff
@@ -1057,15 +1057,46 @@ static IndexIterator *Query_EvalUnionNode(QueryEvalCtx *q, QueryNode *qn) {
...
+      --*len;
+    }
+    *str++ = *p++;
+  }
+  *str = '\0';
+
+  if (!caseSensitive) {
+    size_t newLen = unicode_tolower(origStr, origLen);
+    if (newLen && newLen < origLen) {
+      origStr[newLen] = '\0';
```
oshadmi, line 1077: Can also update `len` here and avoid calling `strlen` by the caller?

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -0,0 +1,1435 @@
...
+        env.assertEqual(res, [1, 'doc:eszett_3'], message=f'Dialect: {dialect}')
+
+        # Test phrase search, replacing ẞ by S.
+        # 0 results because ß was transformed to lowercase, not to S
+        res = conn.execute_command(
+            'FT.SEARCH', 'idx', "@t:(FUSBALL STRAsE)", 'NOCONTENT', 'SORTBY', 't')
+        env.assertEqual(res, [0], message=f'Dialect: {dialect}')
+
+        # Test phrase search, replacing ẞ by SS
+        # 0 results because ß is folded as a single S
```
oshadmi, line 340: Is it folded or transformed to lower?

nafraf: transformed to lower, comments updated.

src/query.c
nafraf: Yes. Done.

oshadmi (dismissed):

src/tokenize.c
oshadmi: Not setting locale anymore

src/stopwords.c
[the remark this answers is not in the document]
oshadmi: Related #5521

oshadmi (dismissed):

raz-mon (commented):
CRAZY good 💯 
See some comments

src/util/strconv.h
```diff
@@ -106,25 +110,98 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
...
+                                              nu_tolower, nu_casemap_read);
+
+  if (u_len >= (SSO_MAX_LENGTH - 1)) {
+    u_buffer = (uint32_t *)rm_malloc(sizeof(*u_buffer) * (u_len + 1));
+  }
+
+  // Decode utf8 string into Unicode codepoints and transform to lower
+  uint32_t codepoint;
+  unsigned i = 0;
+  for (ssize_t j = 0; j < u_len; j++) {
```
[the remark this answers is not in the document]
nafraf: We are using `ssize_t` to be aligned with the type of `u_len`. That's the type returned by nu_strtransformnlen

src/tag_index.c
```diff
@@ -110,9 +113,12 @@ static int tokenizeTagString(const char *str, const FieldSpec *fs, char ***resAr
     tok = TagIndex_SepString(sep, &p, &toklen, indexEmpty);
 
     if (tok) {
-      // lowercase the string (TODO: non latin lowercase)
+      // normalize the string
       if (!(flags & TagField_CaseSensitive)) { // check case sensitive
-        tok = strtolower(tok);
+        size_t newLen = unicode_tolower(tok, strlen(tok));
```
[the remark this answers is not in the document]
nafraf: No. We don't need to unescape at this point.

src/query.c
```diff
@@ -1057,15 +1057,47 @@ static IndexIterator *Query_EvalUnionNode(QueryEvalCtx *q, QueryNode *qn) {
 
 typedef IndexIterator **IndexIteratorArray;
 
+static void tag_strtolower(char *str, size_t *len, int caseSensitive) {
```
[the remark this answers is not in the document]
nafraf: Done.

src/trie/rune_util.c
```diff
@@ -45,9 +60,59 @@ char *runesToStr(const rune *in, size_t len, size_t *utflen) {
...
- * strToRunes except for line where __fold is called */
-rune *strToFoldedRunes(const char *str, size_t *len) {
+/* convert string to runes, lower them and return the lowereded runes */
+rune *strToLowerRunes(const char *str, size_t *len) {
+
+  // determine the length of the folded string
+  ssize_t rlen = nu_strtransformlen(str, nu_utf8_read,
+                                     nu_tolower, nu_casemap_read);
+  if (rlen > MAX_RUNESTR_LEN) {
+    if (len) *len = 0;
```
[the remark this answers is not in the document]
oshadmi: Handled by #5295 - query will error

raz-mon (dismissed):

oshadmi (dismissed):
💪🏼

CMakeLists.txt
```diff
@@ -41,6 +41,9 @@ if (SAN)
     endif()
 endif()
 
+# Needed to handle multi-byte characters architectural independently
```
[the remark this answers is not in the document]
nafraf: Done.

oshadmi (approved):

PR #5394, opened 2024-12-22
MOD-7875: Change AGGPLN To Not Directly Hold User Data

kei-nan:
Use the `HiddenString` type to add an indirection layer for the `AGGPLN` type to avoid exposing user data

1. Current: AGGPLN types hold const char* which directly points to user data
2. Change: Replace the const char* with `HiddenString` and use `HiddenString_GetUnsafe` where needed.
3. Outcome: Make it harder exposing user data

1. AGGPLN

#### Mark if applicable


sbin/unit-tests
ikalchev, line 182: Isn't this a typo (newline breaking `STATFILE`)

PR #5396, opened 2024-12-23
MOD-8393: Refactor python test to use fstring

Itzikvaknin:
Refactor all existing Python test cases under tests/pytests/ to use f-strings for string formatting. Ensure all instances of older string formatting methods (e.g., % operator, str.format() ) are updated to use f-strings. This will enhance code readability.

The refactoring was done using flynt tests/pytests.

### DIFF
--- a/tests/pytests/test_fuzzy.py
+++ b/tests/pytests/test_fuzzy.py
@@ -74,7 +74,7 @@ def testFuzzySyntaxError(env):
     for ch in unallowChars:
         error = None
         try:
-            env.cmd('ft.search', 'idx', '%%wor%sd%%' % ch)
+            env.cmd('ft.search', 'idx', f'%wor{ch}d%')
         except Exception as e:
             error = str(e)
         env.assertTrue('Syntax error' in error)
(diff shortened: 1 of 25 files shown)


tests/pytests/test_fuzzy.py
```diff
@@ -74,7 +74,7 @@ def testFuzzySyntaxError(env):
     for ch in unallowChars:
         error = None
         try:
-            env.cmd('ft.search', 'idx', '%%wor%sd%%' % ch)
+            env.cmd('ft.search', 'idx', f'%wor{ch}d%')
```
raz-mon, line 77: This is fuzzy!

raz-mon: Make sure this is the same

raz-mon (commented):
Great 💯 
See small comment, and make sure we don't have any more confusions with fuzzy search that also uses `%`..

tests/pytests/test_fuzzy.py
Itzikvaknin: The extra % is used to escape the % character itself.

oshadmi (approved):
Cool 👍🏼 😎

raz-mon (approved):

PR #5398, opened 2024-12-23
MOD-8053: Fix leak

raz-mon:
Fixes a leak in which we don't free the RedisModule_String of the ACL `user_name` .


DvirDukhan:
@raz-mon MOD?

raz-mon:
@DvirDukhan I'll connect it to the original PR (fixes a leak from it..)

PR #5403, opened 2024-12-24
MOD-8391: Report active threads-indexes upon crash

raz-mon:

This PR adds a global container holding a mapping between the active queries and the indexes they are working on, such that we can later add information to the logs in case of a crash (more use-cases may follow).

We add the thread to the global container for the following flows:
1. BG thread processing an `FT.SEARCH/AGGREGATE` command.
2. BG thread processing an `FT.CURSOR READ` command.
3. GC operation.
4. Index creation thread (indexer).
5. Main thread, when operating on an index (mainly in the shard, see comment below).

Notice:
1. We add the active thread to the global container only when it holds a StrongRef to it, since we will need to access Index data upon crashing, thus we must ensure that it exists (can't promote a WeakRef in the signal handler that is invoked after a crash).
2. Our main coverage is for the shard commands, as the coordinator commands are mostly stateless w.r.t the index on which they operate.
3. In the coordinator, we do not add the thread to the active-threads container in places that are not aware of the index the query is working on, such as the reducers of the `FT.INFO`, `FT.SEARCH` commands. The thread running the `FT.CURSOR` command is registered only for the `FT.CURSOR READ` subcommand.

#### Mark if applicable


src/spec.c
```diff
@@ -1593,7 +1594,7 @@ void IndexSpec_Free(IndexSpec *spec) {
 // Also assumes that the spec is existing in the global dictionary, so
 // we use the global reference as our guard and access the spec dierctly.
 // This function consumes the Strong reference it gets
-void IndexSpec_RemoveFromGlobals(StrongRef spec_ref) {
+void IndexSpec_RemoveFromGlobals(StrongRef spec_ref, bool removeActive) {
```
kei-nan, line 1597: not sure about tying thread stuff with index object.
Feels like we shouldn't tie them together

kei-nan (changes requested):

src/activeThreads.c
```diff
@@ -0,0 +1,117 @@
...
+ */
+
+#ifdef __linux__
+#include <syscall.h>
+#endif
+
+#include "activeThreads.h"
+#include "rmutil/rm_assert.h"
+
+ActiveThreads *activeThreads = NULL;
```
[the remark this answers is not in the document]
raz-mon: I prefer not having this kind of naming. Today it is easy to see that this variable is in global scope, so I prefer not overloading its name. Furthermore, if we add indicators for every variable characteristic they will become unreadable.

src/activeThreads.c
```diff
@@ -0,0 +1,117 @@
...
+ * This function allocates memory for the `activeThreads` structure and
+ * initializes the doubly-linked list for storing `ActiveThread` objects. It
+ * also initializes the mutex used for thread-safe operations on the
+ * `activeThreads` structure.
+ */
+void activeThreads_Init() {
+  activeThreads = rm_calloc(1, sizeof(ActiveThreads));
+  dllist_init(&activeThreads->list);
+  pthread_mutex_init(&activeThreads->lock, NULL);
+  pthread_key_create(&_activeThreadKey, NULL);
```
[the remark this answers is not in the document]
raz-mon: Can you elaborate?
I want the scope of the link to be the same as the scope of the use of the StrongRef of the index by the thread.

src/spec.c
raz-mon: We release the strong ref here, which is also the latest time I want to be in the active-threads container.
If we don't do this, we miss some important coverage.

src/activeThreads.c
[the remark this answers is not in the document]
kei-nan: I believe I commented on the variable names, here we are talking about file name.
I don't think we have such files in our repo, even if we did the overwhelming majority is using underscore opposed to camelCase.

src/spec.c
kei-nan: Remove from globals also gets called when the ref count reaches 0.
So it is possible that the same thread will call activeThreads_RemoveCurrentThread twice.

I think being strict here and knowing when we release will benefit us in the long run.

src/info/info_redis.c
```diff
@@ -270,3 +277,18 @@ void AddToInfo_RSConfig(RedisModuleInfoCtx *ctx) {
...
+
+void AddToInfo_ActiveThreads(RedisModuleInfoCtx *ctx) {
+  RedisModule_InfoAddSection(ctx, "active_threads");
+
+  // Assumes no other thread is currently accessing the active-threads container
+  DLLIST_FOREACH(node, &(activeThreads->list)) {
+    ActiveThread *at = DLLIST_ITEM(node, ActiveThread, llnode);
+    IndexSpec *spec = StrongRef_Get(at->spec_ref);
+    if (at->Ltid) {
+      RedisModule_InfoAddFieldULongLong(ctx, (const char *)spec->name, (unsigned long long)at->Ltid);
```
kei-nan, line 289: Consider outputting it in the same format redis outputs the thread ids in the crash log, would make our lives a bit easier.
e.g if they output it as hex then maybe output the id as a hex string.

src/activeThreads.c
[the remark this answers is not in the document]
raz-mon: Notice the need for the StrongRef to be valid, which may not be the case after it is released in the thread.
I will add the assertion, seems the safest and most reasonable option.

src/spec.c
raz-mon: The trade-off of adding the argument for more coverage seems good to me, since in the release process many things can go wrong.
It's not that an argument is such a hustle or logic changer.
Don't you think the trade-off is worth it?

src/info/info_redis.c
raz-mon: Yep, exactly why I added the linux `pid_t` thread-id in the Linux case (and printed in the same format).

src/active_threads.c
```diff
@@ -0,0 +1,125 @@
...
+ * This function allocates memory for the `activeThreads` structure and
+ * initializes the doubly-linked list for storing `ActiveThread` objects. It
+ * also initializes the mutex used for thread-safe operations on the
+ * `activeThreads` structure.
+ */
+void activeThreads_Init() {
+  activeThreads = rm_calloc(1, sizeof(ActiveThreads));
+  dllist_init(&activeThreads->list);
+  pthread_mutex_init(&activeThreads->lock, NULL);
+  pthread_key_create(&_activeThreadKey, NULL);
```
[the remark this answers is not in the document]
raz-mon: Will think about it, I'm not sure it will have added value.

src/spec.c
raz-mon: I think this argument makes sense, as we are looking at the releasing flow of an index.
The alternative yields uncovered flows, which is not a good tradeoff IMO.
How would you do this otherwise, without loosing this coverage?

src/active_queries/thread_info.c
```diff
@@ -0,0 +1,92 @@
...
+  // If watchdog kills the process it will notify the main thread which will use this list to output useful information
+  ActiveQueries *activeQueries = ActiveQueries_Init();
+  int rc = pthread_setspecific(activeQueriesKey, activeQueries);
+  initialized = rc == 0;
+  return rc;
+}
+
+void ThreadLocalStorage_Destroy() {
+  if (!initialized) {
+    return;
```
raz-mon, line 46: Not sure we need this

src/active_queries/thread_info.c
```diff
@@ -0,0 +1,92 @@
...
+      rm_free(info);
+      info = NULL;
+    }
+  }
+  return info;
+}
+
+void CurrentThread_SetIndexSpec(StrongRef specRef) {
+  SpecInfo *info = CurrentThread_GetSpecInfo();
+  assert(specRef.rm != NULL);
```
raz-mon, line 76: `assert` --> `RS_ASSERT`

src/active_queries/thread_info.c
```diff
@@ -0,0 +1,92 @@
...
+ActiveQueries *GetActiveQueries() {
+  return pthread_getspecific(activeQueriesKey);
+}
+
+SpecInfo *CurrentThread_TryGetSpecInfo() {
+  SpecInfo *info = pthread_getspecific(specInfoKey);
+  return info;
+}
+
+SpecInfo *CurrentThread_GetSpecInfo() {
```
raz-mon, line 61: Kind of a weird function.. Consider modifying it to only "get"

src/active_queries/thread_info.c
```diff
@@ -0,0 +1,92 @@
...
+    int rc = pthread_setspecific(specInfoKey, info);
+    if (rc) {
+      rm_free(info);
+      info = NULL;
+    }
+  }
+  return info;
+}
+
+void CurrentThread_SetIndexSpec(StrongRef specRef) {
```
raz-mon, line 74: Worth to mention that we require a StrongRef since we need to access the name of the index (currently that's the minimal information we need)

src/active_queries/thread_info.h
```diff
@@ -0,0 +1,48 @@
...
+SpecInfo* CurrentThread_GetSpecInfo();
+// Set the current spec the current thread is working on
+// If the thread will crash while pointing to this spec then the spec information will be outputted
+void CurrentThread_SetIndexSpec(StrongRef specRef);
+// Clear the current index spec the thread is working on
+void CurrentThread_ClearIndexSpec();
+
+#ifdef __cplusplus
+}
+#endif
```
raz-mon, line 48: Newline !!!!

src/info/info_redis.c
```diff
@@ -270,3 +280,44 @@ void AddToInfo_RSConfig(RedisModuleInfoCtx *ctx) {
...
+  }
+}
+
+void AddToInfo_ActiveQueries(RedisModuleInfoCtx *ctx) {
+  ActiveQueries *activeQueries = GetActiveQueries();
+  // If we are not the main thread then do not output the current queries
+  if (!activeQueries) {
+    return;
+  }
+  RedisModule_InfoAddSection(ctx, "active_queries");
```
raz-mon, line 308: I think we should have the same structure always, and have empty sections in case there is nothing to print for that section, such that we will have a consistent format.

src/module.c
```diff
@@ -1153,6 +1198,13 @@ static int RMCreateSearchCommand(RedisModuleCtx *ctx, const char *name,
 int RediSearch_InitModuleInternal(RedisModuleCtx *ctx) {
   GetRedisVersion(ctx);
 
+  // Prepare thread local storage for future threads
```
raz-mon, line 1201:
```suggestion
  // Prepare thread local storage for storing active queries/cursors
```

src/active_queries/active_queries.c
```diff
@@ -0,0 +1,74 @@
...
+ */
+ActiveQueries *ActiveQueries_Init() {
+  ActiveQueries* activeQueries = rm_calloc(1, sizeof(ActiveQueries));
+  dllist_init(&activeQueries->queries);
+  dllist_init(&activeQueries->cursors);
+  return activeQueries;
+}
+
+/**
+ * @brief Frees the active threads data structure.
```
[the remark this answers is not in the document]
kei-nan: Will keep the header comments, maintaining both is too error prone.

src/active_queries/thread_info.c
kei-nan: Added a comment in the header file.

src/active_queries/block_client.c
```diff
@@ -0,0 +1,56 @@
...
+  ActiveQueryNode *node = ActiveQueries_AddQuery(activeQueries, spec_ref, &req->ast);
+
+  // Prepare context for the worker thread
+  // Since we are still in the main thread, and we already validated the
+  // spec's existence, it is safe to directly get the strong reference from the spec
+  // found in buildRequest.
+  RedisModuleBlockedClient *blockedClient = RedisModule_BlockClient(ctx, NULL, NULL, FreeQueryNode, 0);
+  RedisModule_BlockClientSetPrivateData(blockedClient, node);
+  // report block client start time
+  RedisModule_BlockedClientMeasureTimeStart(blockedClient);
```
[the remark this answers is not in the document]
kei-nan: no changed needed in caller

src/active_queries/active_queries.c
```diff
@@ -0,0 +1,74 @@
...
+  RS_LOG_ASSERT((activeQueries->queries.prev == activeQueries->queries.next) &&
+                (activeQueries->queries.next == &activeQueries->queries),
+                "Active queries list is not empty");
+  RS_LOG_ASSERT((activeQueries->cursors.prev == activeQueries->cursors.next) &&
+              (activeQueries->cursors.next == &activeQueries->cursors),
+              "Active cursor list is not empty");
+  rm_free(activeQueries);
+}
+
+ActiveQueryNode* ActiveQueries_AddQuery(ActiveQueries* activeQueries, StrongRef spec, QueryAST* ast) {
```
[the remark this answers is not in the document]
kei-nan: I don't think we are running our tests with threads enabled.

src/active_queries/active_queries.c
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * This function allocates memory for the `ActiveQueries` structure and
+ * initializes the doubly-linked list for storing `ActiveThread` objects. It
+ * also initializes the mutex used for thread-safe operations on the
+ * `ActiveQueries` structure.
+ */
+ActiveQueries *ActiveQueries_Init() {
+  ActiveQueries* activeQueries = rm_calloc(1, sizeof(ActiveQueries));
+  dllist_init(&activeQueries->queries);
+  dllist_init(&activeQueries->cursors);
```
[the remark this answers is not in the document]
kei-nan: I think the separation could be a good thing, we can bring it up for a team discussion.

src/active_queries/active_queries.h
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * This structure is used to store information about an active query, including
+ * the query itself, and a strong reference to the `IndexSpec`
+ * associated with the query. Since we use the StrongRef, we know that we can
+ * safely access the `IndexSpec` upon crashing.
+ */
+typedef struct {
+  DLLIST_node llnode; // Node in the doubly-linked list
+  StrongRef spec;     // IndexSpec strong ref
+  QueryAST *ast;      // The query AST
```
raz-mon, line 30: Let's remove this, since it is unused.
We can add it when we'll need it.

src/active_queries/thread_info.c
```diff
@@ -0,0 +1,81 @@
...
+  rc = pthread_key_create(&specInfoKey, NULL);
+  assert(rc == 0);
+}
+
+static void __attribute__((destructor)) destroyKeys() {
+  pthread_key_delete(specInfoKey);
+  pthread_key_delete(activeQueriesKey);
+}
+
+int ThreadLocalStorage_Init() {
```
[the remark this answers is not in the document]
kei-nan: renamed the function

src/info/info_redis/info_redis.c
```diff
@@ -270,3 +281,71 @@ void AddToInfo_RSConfig(RedisModuleInfoCtx *ctx) {
...
+    // we are not the main thread, simply return
+    return;
+  }
+  DLLIST_FOREACH(node, &(activeQueries->cursors)) {
+    BlockedCursorNode *at = DLLIST_ITEM(node, BlockedCursorNode, llnode);
+    IndexSpec *spec = StrongRef_Get(at->spec);
+    char buffer[21]; // 20 is the max length of a uint64_t
+    sprintf(buffer, "%zu", at->cursorId);
+    RedisModule_InfoBeginDictField(ctx, buffer);
+    RedisModule_InfoAddFieldCString(ctx, "index", spec ? spec->name : "n/a");
```
raz-mon, line 334: Will need to be obfuscated when the obfuscation pr is merged

src/info/info_redis/info_redis.c
```diff
@@ -270,3 +281,71 @@ void AddToInfo_RSConfig(RedisModuleInfoCtx *ctx) {
...
+  SpecInfo *specInfo = CurrentThread_TryGetSpecInfo();
+  RedisModule_InfoAddSection(ctx, "current_thread");
+  if (!specInfo) {
+    return;
+  }
+  if (specInfo) {
+    StrongRef strong = WeakRef_Promote(specInfo->specRef);
+    IndexSpec *spec = StrongRef_Get(strong);
+    if (!spec) {
+      RedisModule_InfoAddFieldCString(ctx, "index", specInfo->specName ? specInfo->specName : "n/a");
```
raz-mon, line 296: Will need to be obfuscated soon (just to note)

src/info/info_redis/info_redis.c
```diff
@@ -270,3 +281,71 @@ void AddToInfo_RSConfig(RedisModuleInfoCtx *ctx) {
...
+void AddToInfo_CurrentThread(RedisModuleInfoCtx *ctx) {
+  SpecInfo *specInfo = CurrentThread_TryGetSpecInfo();
+  RedisModule_InfoAddSection(ctx, "current_thread");
+  if (!specInfo) {
+    return;
+  }
+  if (specInfo) {
+    StrongRef strong = WeakRef_Promote(specInfo->specRef);
+    IndexSpec *spec = StrongRef_Get(strong);
+    if (!spec) {
```
raz-mon, line 295: Would be nice to have some more documentation, for instance, when can this happen (`spec == NULL`)?

src/info/info_redis/info_redis.c
kei-nan: yep, but as discussed this is out of scope for this PR.

src/info/info_redis/info_redis.c
```diff
@@ -270,3 +281,71 @@ void AddToInfo_RSConfig(RedisModuleInfoCtx *ctx) {
   RedisModule_InfoAddFieldLongLong(ctx, "min_phonetic_term_length",
                                    RSGlobalConfig.minPhoneticTermLen);
 }
+
+// IF the crashing thread worked on a spec, output the spec name
+void AddToInfo_CurrentThread(RedisModuleInfoCtx *ctx) {
```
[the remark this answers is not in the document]
kei-nan: I think this information we are already getting from redis.
We know the thread that called this function is the one that crashed.
So if redis already marks the thread for us it seems pointless to print its thread id.

Anyhow this type of change is out of scope of this PR, IMHO.

src/info/info_redis/info_redis.c
```diff
@@ -270,3 +281,71 @@ void AddToInfo_RSConfig(RedisModuleInfoCtx *ctx) {
...
+  if (!activeQueries) {
+    // we are not the main thread, simply return
+    return;
+  }
+  DLLIST_FOREACH(node, &(activeQueries->cursors)) {
+    BlockedCursorNode *at = DLLIST_ITEM(node, BlockedCursorNode, llnode);
+    IndexSpec *spec = StrongRef_Get(at->spec);
+    char buffer[21]; // 20 is the max length of a uint64_t
+    sprintf(buffer, "%zu", at->cursorId);
+    RedisModule_InfoBeginDictField(ctx, buffer);
```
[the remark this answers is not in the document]
kei-nan: I don't think so, an index name is a constant thing which doesn't change from the creation point of the index.
If there are no data races on the specific variable then I think we don't need to lock at this present time.

src/active_queries/active_queries.c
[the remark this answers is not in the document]
kei-nan: for example you could have coordinator cursors.
The spec ref behaviour is a bit dfferent and the meaning of not having a spec ref means different things for cursors and queries.
That is just one difference, In general I think separating the stuff we need to analyse will provide more order in our analysis.

src/info/info_redis/info_redis.c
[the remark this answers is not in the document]
raz-mon: We have them printed only for Linux platforms.
I think it's nice data to have in case it's not printed, but can drop it for now if you don't want to overload the PR.

src/aggregate/aggregate_exec.c
```diff
@@ -868,10 +875,13 @@ static int buildRequest(RedisModuleCtx *ctx, RedisModuleString **argv, int argc,
     goto done;
   }
 
+  CurrentThread_SetIndexSpec(sctx->spec->own_ref);
+
   rc = AREQ_ApplyContext(*r, sctx, status);
   thctx = NULL;
   // ctx is always assigned after ApplyContext
   if (rc != REDISMODULE_OK) {
+    CurrentThread_ClearIndexSpec();
```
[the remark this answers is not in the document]
kei-nan: in this specific case by the caller, either due to a failure in building the request or when completing the request.

src/aggregate/aggregate_plan.h
```diff
@@ -63,6 +63,8 @@ typedef struct PLN_BaseStep {
 #define PLN_NEXT_STEP(step) DLLIST_ITEM((step)->llnodePln.next, PLN_BaseStep, llnodePln)
 #define PLN_PREV_STEP(step) DLLIST_ITEM((step)->llnodePln.prev, PLN_BaseStep, llnodePln)
 #define PLN_END_STEP(plan) DLLIST_ITEM(&(plan)->steps, PLN_BaseStep, llnodePln)
+// TODO: The above works by accident.
```
oshadmi, line 66: Please replace comment with a ticket

nafraf (dismissed):

nafraf (dismissed):

src/spec.c
```diff
@@ -3329,3 +3327,20 @@ static void DebugIndexes_ScanProc(RedisModuleCtx *ctx, RedisModuleString *keynam
 
   Indexes_ScanProc(ctx, keyname, key, &(dScanner->base));
 }
+
+StrongRef IndexSpecRef_Promote(WeakRef ref) {
```
raz-mon, line 3372: Let's add some good documentation here - not a trivial function to call

kei-nan: added comments in the header file.

kei-nan (approved):
Roles got switched during development.
Approving on behalf of Raz.

PR #5411, opened 2024-12-30
[MOD-8431] Improve RDB files download

meiravgri:
Enhanced the verbosity of the RDB file download process to better capture errors. 
Refactored the `downloadFiles` function for improved error handling and introduced a new `downloadFile` function for individual file downloads.

### DIFF
--- a/tests/pytests/common.py
+++ b/tests/pytests/common.py
@@ -336,7 +336,7 @@ def no_msan(f):
     def wrapper(env, *args, **kwargs):
         if SANITIZER == 'memory':
             fname = f.__name__
-            env.debugPrint("skipping {} due to memory sanitizer".format(fname), force=True)
+            env.debugPrint(f"skipping {fname} due to memory sanitizer", force=True)
             env.skip()
             return
         return f(env, *args, **kwargs)
@@ -347,7 +347,7 @@ def unstable(f):
     def wrapper(env, *args, **kwargs):
         if UNSTABLE == True:
             fname = f.__name__
-            env.debugPrint("skipping {} because it is unstable".format(fname), force=True)
+            env.debugPrint(f"skipping {fname} because it is unstable", force=True)
             env.skip()
             return
         return f(env, *args, **kwargs)
@@ -510,11 +510,11 @@ def compare_lists(env, list1, list2, delta=0.01, _assert=True):
     res = compare_lists_rec(list1, list2, delta + 0.000001)
     if res:
         if _assert:
-            env.assertTrue(True, message='%s ~ %s' % (str(list1), str(list2)))
+            env.assertTrue(True, message=f'{str(list1)} ~ {str(list2)}')
         return True
     else:
         if _assert:
-            env.assertTrue(False, message='%s ~ %s' % (str(list1), str(list2)))
+            env.assertTrue(False, message=f'{str(list1)} ~ {str(list2)}')
         return False
 
 class ConditionalExpected:
--- a/tests/pytests/common.py
+++ b/tests/pytests/common.py
@@ -1,4 +1,3 @@
-
 from includes import *
 try:
     from collections.abc import Iterable
@@ -24,6 +23,8 @@ from unittest.mock import ANY, _ANY
 from unittest import SkipTest
 import inspect
 import subprocess
+import requests
+import urllib3
 
 BASE_RDBS_URL = 'https://dev.cto.redis.s3.amazonaws.com/RediSearch/rdbs/'
 REDISEARCH_CACHE_DIR = '/tmp/redisearch-rdbs/'
@@ -744,15 +745,32 @@ def check_index_info_empty(env, idx, fields, msg="after delete all and gc", dept
     expected_size = getInvertedIndexInitialSize_MB(env, fields, depth=depth+1)
     check_index_info(env, idx, exp_num_records=0, exp_inv_idx_size=expected_size, msg=msg, depth=depth+1)
 
-def downloadFiles(rdbs = None):
+
+def downloadFiles(env, rdbs=None):
     if rdbs is None:
         return False
 
-    os.makedirs(REDISEARCH_CACHE_DIR, exist_ok=True) # create cache dir if not exists
+    os.makedirs(REDISEARCH_CACHE_DIR, exist_ok=True)  # create cache dir if not exists
+    urllib3.disable_warnings(
+        urllib3.exceptions.InsecureRequestWarning
+    )  # disable warning for downloading file without certificate
     for f in rdbs:
         path = os.path.join(REDISEARCH_CACHE_DIR, f)
         if not os.path.exists(path):
-            subprocess.run(["wget", "--no-check-certificate", BASE_RDBS_URL + f, "-O", path, "-q"])
+            response = requests.get(BASE_RDBS_URL + f, verify=False)
+            try:
+                response.raise_for_status()
+            except requests.exceptions.HTTPError as err:
+                env.assertTrue(
+                    False,
+                    message=f"Failed to download {f} from {BASE_RDBS_URL} with error: {err}",
+                    depth=1,
+                )
+                return False
+
+            with open(path, "wb") as file:
+                file.write(response.content)
+
         if not os.path.exists(path):
             return False
     return True
--- a/tests/pytests/common.py
+++ b/tests/pytests/common.py
@@ -776,6 +776,11 @@ def downloadFile(env, file_name, depth=0):
                 pass
             return False
     if not os.path.exists(path):
+        env.assertTrue(
+            False,
+            message=f"{path} does not exist after download",
+            depth=depth + 1,
+        )
         return False
     return True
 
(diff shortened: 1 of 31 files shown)


tests/pytests/test_aux_save2.py
```diff
@@ -138,12 +138,9 @@ def testLoadRdbWithEmptySpellcheckDict(env):
...
     rdbFilePath = os.path.join(dbDir, dbFileName)
-    if not downloadFiles(RDBS):
-        if CI:
-            env.assertTrue(False)  ## we could not download rdbs and we are running on CI, let fail the test
-        else:
-            env.skip()
-            return
+    res = downloadFiles(env, RDBS)
+    if not res:
+        return
```
GuyAv46, line 143:
```suggestion
    if not downloadFiles(env, RDBS):
        return
```

tests/pytests/test_aux_save2.py
```diff
@@ -138,12 +138,9 @@ def testLoadRdbWithEmptySpellcheckDict(env):
     dbFileName = env.cmd('config', 'get', 'dbfilename')[1]
```
GuyAv46, line 138: Can you also fix the test so we don't override the `env` right away? (remove the `env = Env()` line or the test `env` parameter

tests/pytests/test_rdb_compatibility.py
```diff
@@ -68,12 +65,9 @@ def testRDBCompatibility_vecsim():
...
     algorithms = ['FLAT', 'HNSW']
-    if not downloadFiles(rdbs):
-        if CI:
-            env.assertTrue(False)  ## we could not download rdbs and we are running on CI, let fail the test
-        else:
-            env.skip()
-            return
+    res = downloadFiles(env, rdbs)
+    if not res:
+        return
```
GuyAv46, line 70:
```suggestion
    if not downloadFiles(env, RDBS):
        return
```

tests/pytests/test_rdb_compatibility.py
```diff
@@ -26,12 +26,9 @@ def testRDBCompatibility(env):
...
     rdbFilePath = os.path.join(dbDir, dbFileName)
-    if not downloadFiles(RDBS):
-        if CI:
-            env.assertTrue(False)  ## we could not download rdbs and we are running on CI, let fail the test
-        else:
-            env.skip()
-            return
+    res = downloadFiles(env, RDBS)
+    if not res:
+        return
```
GuyAv46, line 31:
```suggestion
    if not downloadFiles(env, RDBS):
        return
```

tests/pytests/common.py
```diff
@@ -744,15 +745,31 @@ def check_index_info_empty(env, idx, fields, msg="after delete all and gc", dept
...
+            except requests.exceptions.HTTPError as err:
+                env.assertTrue(
+                    False,
+                    message=f"Failed to download {f} from {BASE_RDBS_URL} with error: {err}",
+                    depth=1,
+                )
+                return False
+
+            with open(path, "wb") as file:
+                file.write(response.content)
```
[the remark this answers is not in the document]
meiravgri: I didn't use the streaming option to have a minimal amount of calls to simplify the error handling mechanism
Anyway, the large files are all part of the same test, so the request object is hopefully released at the end of each loop iteration. The vecsim test files are relatively small.
So in this test file we have a max of 2 rdb files in the RAM at the same time

tests/pytests/common.py
```diff
@@ -744,15 +745,31 @@ def check_index_info_empty(env, idx, fields, msg="after delete all and gc", dept
...
+            response = requests.get(BASE_RDBS_URL + f, verify=False)
+            try:
+                response.raise_for_status()
+            except requests.exceptions.HTTPError as err:
+                env.assertTrue(
+                    False,
+                    message=f"Failed to download {f} from {BASE_RDBS_URL} with error: {err}",
+                    depth=1,
+                )
+                return False
```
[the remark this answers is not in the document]
meiravgri: request module doesn't raise an exception without calling this function

tests/pytests/test_rdb_compatibility.py
meiravgri: 😓 
Wanted to use it for logging or something

GuyAv46 (approved):

tests/pytests/test_short_read.py
```diff
@@ -622,7 +628,7 @@ def doTest(env: Env, test_name, rdb_name, expected_index):
         sendShortReads(env, fullPath, expected_index)
 
 # Dynamically create a test function for each rdb file
-@skip(cluster=True, redis_less_than='6.2.0', macos=True, asan=True, arch='aarch64')
+@skip(cluster=True, redis_less_than='6.2.0', macos=True, asan=True)
```
alonre24, line 631: Any reason we used to skip on arm? Also why skipping on macos?

tests/pytests/common.py
```diff
@@ -744,15 +743,39 @@ def check_index_info_empty(env, idx, fields, msg="after delete all and gc", dept
...
+                        "wget",
+                        "--no-check-certificate",
+                        BASE_RDBS_URL + f,
+                        "-O",
+                        path,
+                        "-q",
+                    ],
+                    check=True,
+                )
+            except subprocess.CalledProcessError as e:
```
alonre24, line 766: Exception is thrown only when passing `check=True`? Also what about other exception types, is that possible?

alonre24 (commented):
NICE!!

tests/pytests/test_short_read.py
```diff
@@ -597,21 +597,27 @@ def runShortRead(env, data, total_len, expected_index):
 seed = str(time.time())
 random.seed(seed)
 
-def downloadFile(file_name):
+def downloadFile(env, file_name):
     path = os.path.join(REDISEARCH_CACHE_DIR, file_name)
     path_dir = os.path.dirname(path)
     os.makedirs(path_dir, exist_ok=True)
     if not os.path.exists(path):
         subprocess.run(["wget", "--no-check-certificate", BASE_RDBS_URL + file_name, "-O", path, "-q"])
```
GuyAv46, line 605: we should add the check here as well as you did, or
```suggestion
        subprocess.run(["wget", "--no-check-certificate", BASE_RDBS_URL + file_name, "-O", path, "-q"]).check_returncode()
```

tests/pytests/test_short_read.py
```diff
@@ -597,21 +597,27 @@ def runShortRead(env, data, total_len, expected_index):
...
         if os.path.splitext(path)[-1] == '.zip':
-            return unzip(path, path_dir)
+            res = unzip(path, path_dir)
+            env.assertTrue(res, message='Failed to unzip ' + path)
+            return res
         else:
-            return os.path.exists(path) and os.path.getsize(path) > 0
+            exists = os.path.exists(path)
+            non_zero_size = os.path.getsize(path) > 0
```
GuyAv46, line 612: it is unsafe to split this condition, as you cannot call `os.path.getsize(path)` if the path does not exist

tests/pytests/test_short_read.py
meiravgri: IIRC, we coudnlt figure why it takes so long on MAC
I wanted to run it locally so i removed the restriction and it indeed got stuck (or i didn't wait long enough)
just forgot to revert 😅

tests/pytests/common.py
meiravgri: This is the exception thrown when check=True

tests/pytests/common.py
```diff
@@ -743,39 +743,47 @@ def check_index_info_empty(env, idx, fields, msg="after delete all and gc", dept
...
+            try:
+                os.remove(path)
+                env.debugPrint(f"Partially downloaded file {path}. Removing it.", force=True)
+            except OSError:
+                env.debugPrint(f"Failed to remove {path}", force=True)
+                pass
+            return False
+    if not os.path.exists(path):
+        return False
+    return True
```
GuyAv46, line 780: We can also just return `True` at this point no? No need to recheck if the file exists.
If I’m wrong, we can also just return the return value of `os.path.exists(path)`

tests/pytests/common.py
```diff
@@ -743,39 +743,47 @@ def check_index_info_empty(env, idx, fields, msg="after delete all and gc", dept
...
-                    os.remove(path)
-                    print(f"Partially downloaded file {f}. Removing it.")
-                except OSError:
-                    print(f"Failed to remove {f}")
-                    pass
-                return False
-        if not os.path.exists(path):
+        if not downloadFile(env, f, depth=depth + 1):
             return False
-    return True
```
GuyAv46, line 758: Not returning nothing at the end?

alonre24 (approved):

tests/pytests/common.py
meiravgri: I also want to believe that handeling errors from wget covers all potential failures, but since we didn't fully identify the reason for the missing/corrupted files, i chose to keep this sanity check.
Ill add a log here

PR #5415, opened 2024-12-31
Add an isAlphabetic check before stemming

Itzikvaknin:
- Added an isAlphabetic check before stemming to ensure only alphabetical tokens are processed.
- Added unittests for misc.h functions.
- Updated expected test_dialect results.

Stemming algorithms are defined for alphabetical tokens, and this update ensures non-alphabetical tokens are excluded from the stemming process.

### DIFF
--- a/src/util/misc.c
+++ b/src/util/misc.c
@@ -28,3 +28,14 @@ int GetRedisErrorCodeLength(const char* error) {
   const char* errorSpace = strchr(error, ' ');
   return errorSpace ? errorSpace - error : 0;
 }
+
+bool contains_non_alphabetic_char(char* str, size_t len) {
+    if (len == 0 || !str ) return false;
+    const char* non_alphabetic_chars = "0123456789!@#$%^&*()_+-=[]{}\\|;:'\",.<>/?`~§± ";
+    for (size_t i = 0; i < len; i++) {
+        if (strchr(non_alphabetic_chars, str[i])) {
+            return true;
+        }
+    }
+    return false;
+}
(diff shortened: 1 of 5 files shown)


src/util/misc.c
```diff
@@ -9,6 +9,8 @@
 #include <ctype.h>
 #include <string.h>
 
+#define NON_ALPHABETIC_CHARS "0123456789!@#$%^&*()_+-=[]{}\\|;:'\",.<>/?`~§± "
```
nafraf, line 12: How did we selected the characters in this list? Why do we need to include `§ (167)` and `± (177)` but not `© (169)`?

tests/pytests/test_dialect.py
```diff
@@ -159,7 +149,7 @@ def test_v1_vs_v2(env):
 
     # DIALECT 2 does not expand numbers
```
nafraf, line 150: Should we update this comment? Now, none of the dialects will expand numbers.

tests/pytests/test_dialect.py
```diff
@@ -159,7 +149,7 @@ def test_v1_vs_v2(env):
...
     res = env.cmd('FT.EXPLAINCLI', 'idx', '705', 'DIALECT', 1)
-    expected = ['UNION {', '  705', '  +705(expanded)', '}', '']
+    expected = ['705', '']
     env.assertEqual(res, expected)
     res = env.cmd('FT.EXPLAINCLI', 'idx', '705', 'DIALECT', 2)
     expected = ['705', '']
```
nafraf, line 155: We can remove this line. The expected result is the same for both dialects.

tests/pytests/test_dialect.py
```diff
@@ -187,7 +177,7 @@ def test_v1_vs_v2(env):
     env.assertEqual(res, expected)
 
     # terms wich contain numbers are expanded
```
nafraf, line 179: This comment can be updated.

PR #5418, opened 2025-01-01
MOD-6934: Do not create union for single children query node

Itzikvaknin:
We currently build a Union node even when we have a single child node iterator in several query_eval<type>node() functions. Changed it to return the single child iterator instead.

### DIFF
--- a/src/query.c
+++ b/src/query.c
@@ -650,10 +650,9 @@ static IndexIterator *Query_EvalPrefixNode(QueryEvalCtx *q, QueryNode *qn) {
   if (!ctx.its || ctx.nits == 0) {
     rm_free(ctx.its);
     return NULL;
-  // TODO: This should be a single iterator.
-  // } else if (ctx.nits == 1) {
-  //   // In case of a single iterator, we can just return it
-  //   return ctx.its[0];
+  } else if (ctx.nits == 1) {
+    // In case of a single iterator, we can just return it
+    return ctx.its[0];
   } else {
     return NewUnionIterator(ctx.its, ctx.nits, 1, qn->opts.weight,
                             QN_PREFIX, qn->pfx.tok.str, q->config);
@@ -834,6 +833,8 @@ static IndexIterator *Query_EvalLexRangeNode(QueryEvalCtx *q, QueryNode *lx) {
   if (!ctx.its || ctx.nits == 0) {
     rm_free(ctx.its);
     return NULL;
+  } else if (ctx.nits == 1) {
+    return ctx.its[0];
   } else {
     return NewUnionIterator(ctx.its, ctx.nits, 1, lx->opts.weight, QN_LEXRANGE, NULL, q->config);
   }
@@ -1384,6 +1385,12 @@ static IndexIterator *Query_EvalTagNode(QueryEvalCtx *q, QueryNode *qn) {
     } else {
       array_free(total_its);
     }
+  }
+
+  if (n == 1) {
+    ret = iters[0];
+    rm_free(iters);
+    goto done;
   }
 
   ret = NewUnionIterator(iters, n, 0, qn->opts.weight, QN_TAG, NULL, q->config);
--- a/src/query.c
+++ b/src/query.c
@@ -651,7 +651,9 @@ static IndexIterator *Query_EvalPrefixNode(QueryEvalCtx *q, QueryNode *qn) {
     rm_free(ctx.its);
     return NULL;
   } else if (ctx.nits == 1) {
-    return ctx.its[0];
+    IndexIterator *ret = ctx.its[0];
+    rm_free(ctx.its);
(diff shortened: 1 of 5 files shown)


src/query.c
```diff
@@ -650,10 +650,8 @@ static IndexIterator *Query_EvalPrefixNode(QueryEvalCtx *q, QueryNode *qn) {
   if (!ctx.its || ctx.nits == 0) {
     rm_free(ctx.its);
     return NULL;
-  // TODO: This should be a single iterator.
-  // } else if (ctx.nits == 1) {
-  //   // In case of a single iterator, we can just return it
-  //   return ctx.its[0];
+  } else if (ctx.nits == 1) {
```
kei-nan, line 653: Feels like there are more places where this is relevant, no?

Itzikvaknin: Changes applied to all Query_Eval functions except Query_EvalWildcardQueryNode(), as its union is overloaded with wildcard query data used by FT.PROFILE, which breaks without the union creation.

kei-nan (approved):
Looks good

src/query.c
```diff
@@ -834,6 +834,10 @@ static IndexIterator *Query_EvalLexRangeNode(QueryEvalCtx *q, QueryNode *lx) {
   if (!ctx.its || ctx.nits == 0) {
     rm_free(ctx.its);
     return NULL;
+  } else if (ctx.nits == 1) {
+    IndexIterator *ret = ctx.its[0];
+    rm_free(ctx.its);
+    return ret;
```
kei-nan, line 840: Maybe check why there isn't code coverage for this.
I think `testRangesOnTags` should cover this.

kei-nan (approved):

src/query.c
```diff
@@ -1065,6 +1069,10 @@ static IndexIterator *Query_EvalTagLexRangeNode(QueryEvalCtx *q, TagIndex *idx,
   if (ctx.nits == 0) {
     rm_free(ctx.its);
     return NULL;
+  } else if (ctx.nits == 1) {
+    IndexIterator *ret = ctx.its[0];
+    rm_free(ctx.its);
```
raz-mon, line 1074: Need to nullify `ctx.its` after freeing it (same for the rest of the changes)? Basically, is anyone asking `if (ctx.its)`?

tests/pytests/test_profile.py
```diff
@@ -58,8 +58,7 @@ def testProfileSearch(env):
 
   # test PREFIX
   actual_res = conn.execute_command('ft.profile', 'idx', 'search', 'query', 'hel*', 'nocontent')
-  expected_res = ['Type', 'UNION', 'Query type', 'PREFIX - hel', 'Counter', 1, 'Child iterators', [
-                    ['Type', 'TEXT', 'Term', 'hello', 'Counter', 1, 'Size', 1]]]
+  expected_res = ['Type', 'TEXT', 'Term', 'hello', 'Counter', 1, 'Size', 1]
```
raz-mon, line 61: Lovely as this is, I see we also loose important information here..
Maybe we should just change the print of these scenarios such that we get this information without the need for the `UNION` iterator wrapper.

tests/pytests/test_scorers.py
```diff
@@ -246,11 +246,9 @@ def testOptionalAndWildcardScoring(env):
...
     env.assertEqual(res, expected_res)
 
-    expected_res = [1, 'doc2', ['0.6489037427548099',
-                                ['Final BM25 : words BM25 0.65 * document score 1.00',
-                                 [['(Weight 1.00 * children BM25 0.65)',
-                                   ['words: (0.65 = IDF 0.69 * (F 1.00 * (k1 1.2 + 1)) '
-                                    '/ (F 1.00 + k1 1.2 * (1 - b 0.5 + b 0.5 * Doc Len 5 / Average Doc Len 4.00)))']]]]]]
+    expected_res = [1, 'doc2', ['0.6489037427548099', 
+                                ['Final BM25 : words BM25 0.65 * document score 1.00', 
+                                    ['words: (0.65 = IDF 0.69 * (F 1.00 * (k1 1.2 + 1)) / (F 1.00 + k1 1.2 * (1 - b 0.5 + b 0.5 * Doc Len 5 / Average Doc Len 4.00)))']]]]
```
raz-mon, line 251: No score changes right?

raz-mon (commented):
Very nice 👌 
See a few comments. Mainly - I think we need to add a mechanism to get the query type without needing to wrap the single iterator with a `UNION` iterator. Until we have that, we shouldn't loose information, such as the `PREFIX` case, and the other case we already left the `UNION` wrapper in.

tests/pytests/test_scorers.py
Itzikvaknin: Right :)

PR #5422, opened 2025-01-01
MOD-7570: Valid safe vector index pointer access

lerman25:
Validation on vector index pointer accesses to prevent null pointer access. 
Solving 2 bugs:
1. Creating index with schema containing vector index field with invalid parameters, resulting in crash when Inserting and then updating a doc that's in the index but without a vector field.
2. Debug command vecsim_info crashes when the vector index was created with invalid parameters.

### DIFF
--- a/src/debug_commands.c
+++ b/src/debug_commands.c
@@ -1152,12 +1152,6 @@ DEBUG_COMMAND(VecsimInfo) {
   // This call can't fail, since we already checked that the key exists
   // (or should exist, and this call will create it).
   VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName,CREATE_INDEX);
-  // @OMER - not checking ptr
-  // @OMER c1 - the comment above sais call can't fail
-  //         but what if the specs are "bad" so no creation-
-  //         where does the specs being checked?
-  // if null return error "cannot open vector index"
-
   if(!vecsimIndex)
   {
     SearchCtx_Free(sctx);
@@ -1229,9 +1223,6 @@ DEBUG_COMMAND(dumpHNSWData) {
   // This call can't fail, since we already checked that the key exists
   // (or should exist, and this call will create it).
   VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName, CREATE_INDEX);
-    // @OMER - not checking ptr
-    // @OMER - see comment c1
-  // same solution
   if(!vecsimIndex)
   {
     RedisModule_ReplyWithError(ctx, "Can't open vector index");
--- a/src/debug_commands.c
+++ b/src/debug_commands.c
@@ -1151,7 +1151,18 @@ DEBUG_COMMAND(VecsimInfo) {
   }
   // This call can't fail, since we already checked that the key exists
   // (or should exist, and this call will create it).
-  VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName, CREATE_INDEX);
+  VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName,CREATE_INDEX);
+  // @OMER - not checking ptr
+  // @OMER c1 - the comment above sais call can't fail
+  //         but what if the specs are "bad" so no creation-
+  //         where does the specs being checked?
+  // if null return error "cannot open vector index"
+
+  if(!vecsimIndex)
+  {
+    SearchCtx_Free(sctx);
+    return RedisModule_ReplyWithError(ctx, "Can't open vector index");
+  }
 
   VecSimInfoIterator *infoIter = VecSimIndex_InfoIterator(vecsimIndex);
   // Recursively reply with the info iterator
@@ -1218,6 +1229,16 @@ DEBUG_COMMAND(dumpHNSWData) {
   // This call can't fail, since we already checked that the key exists
   // (or should exist, and this call will create it).
   VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName, CREATE_INDEX);
+    // @OMER - not checking ptr
+    // @OMER - see comment c1
+  // same solution
+  if(!vecsimIndex)
+  {
+    RedisModule_ReplyWithError(ctx, "Can't open vector index");
+    goto cleanup;
+  }
+
+
   VecSimIndexBasicInfo info = VecSimIndex_BasicInfo(vecsimIndex);
   if (info.algo != VecSimAlgo_HNSWLIB) {
 	  RedisModule_ReplyWithError(ctx, "Vector index is not an HNSW index");
--- a/src/debug_commands.c
+++ b/src/debug_commands.c
@@ -1151,9 +1151,8 @@ DEBUG_COMMAND(VecsimInfo) {
   }
   // This call can't fail, since we already checked that the key exists
   // (or should exist, and this call will create it).
-  VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName,CREATE_INDEX);
-  if(!vecsimIndex)
-  {
+  VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName, CREATE_INDEX);
+  if(!vecsimIndex) {
     SearchCtx_Free(sctx);
     return RedisModule_ReplyWithError(ctx, "Can't open vector index");
   }
@@ -1223,8 +1222,7 @@ DEBUG_COMMAND(dumpHNSWData) {
   // This call can't fail, since we already checked that the key exists
   // (or should exist, and this call will create it).
   VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName, CREATE_INDEX);
-  if(!vecsimIndex)
-  {
+  if(!vecsimIndex) {
     RedisModule_ReplyWithError(ctx, "Can't open vector index");
     goto cleanup;
   }
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -138,18 +138,9 @@ static RSDocumentMetadata *makeDocumentId(RedisModuleCtx *ctx, RSAddDocumentCtx 
         for (int i = 0; i < spec->numFields; ++i) {
           if (spec->fields[i].types == INDEXFLD_T_VECTOR) {
             RedisModuleString *rmstr = IndexSpec_GetFormattedKey(spec, &spec->fields[i], INDEXFLD_T_VECTOR);
-            VecSimIndex *vecsim = openVectorIndex(spec, rmstr, CREATE_INDEX);
-            // ####################
+            VecSimIndex *vecsim = openVectorIndex(spec, rmstr, DONT_CREATE_INDEX);
             if(!vecsim)
-            {
-              const FieldSpec *fs = aCtx->fspecs + i;
-              // QueryError_SetError(status, QUERY_VECTORINDEX, "Could not open vector index");
-              IndexError_AddError(&aCtx->spec->stats.indexError, "Could not open vector index", doc->docKey);
-              FieldSpec_AddError(&spec->fields[i], "Could not open vector index", doc->docKey);
               continue;
-            }
-            // ####################
-            //@Omer - calling function checks for null, so return null
             VecSimIndex_DeleteVector(vecsim, dmd->id);
             // TODO: use VecSimReplace instead and if successful, do not insert and remove from doc
           }
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -145,7 +145,7 @@ static RSDocumentMetadata *makeDocumentId(RedisModuleCtx *ctx, RSAddDocumentCtx 
               const FieldSpec *fs = aCtx->fspecs + i;
               // QueryError_SetError(status, QUERY_VECTORINDEX, "Could not open vector index");
               IndexError_AddError(&aCtx->spec->stats.indexError, "Could not open vector index", doc->docKey);
-              FieldSpec_AddError(&aCtx->spec->fields[fs->index], "Could not open vector index", doc->docKey);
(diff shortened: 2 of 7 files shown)


tests/pytests/test_vecsim.py
```diff
@@ -40,6 +40,10 @@ def load_vectors_with_texts_into_redis(con, vector_field, dim, num_vectors, data
     p.execute()
     return id_vec_list
 
+def index_errors(env):
+    return to_dict(index_info(env)['Index Errors'])
+def field_errors(env,fld_index = 0):
+    return to_dict(to_dict(to_dict(index_info(env)['field statistics'][fld_index]))['Index Errors'])
```
alonre24, line 46: This can be moved to common in case tests in other files will use that

tests/pytests/test_vecsim.py
```diff
@@ -2394,3 +2394,30 @@ def test_switch_write_mode_multiple_indexes(env):
...
         prefix = "::warning title=Bad scenario in test_vecsim:test_switch_write_mode_multiple_indexes::" if GHA else ''
         print(f"{prefix}All vectors were done reindex before switching back to in-place mode")
+
+
+def test_vector_index_ptr_valid(env):
+    conn = getConnectionByEnv(env)
+    # Scenerio1: Vecsim Index scheme with numeric (or non-vector type) and vector type with "bad" parameter 
+    #            Insert partial doc - only numeric
+    #            Update Doc
+    # Note: Should be updated if "bad" parameters are updated
```
alonre24, line 2404: Let's just explain what is the "bad" parameter scenario that we are testing here

tests/pytests/test_vecsim.py
```diff
@@ -2394,3 +2394,30 @@ def test_switch_write_mode_multiple_indexes(env):
...
+    
+    UINT16_MAX = 2**16
+    M = UINT16_MAX + 1
+    dim = 2
+
+    conn.execute_command('FT.CREATE', 'idx','SCHEMA', 'n', 'NUMERIC',
+                    'v', 'VECTOR', 'HNSW', '8', 'TYPE', 'FLOAT16', 'DIM', dim, 'DISTANCE_METRIC', 'L2', 'M', M)   
+    conn.execute_command('HSET', 'doc', 'n', 0)
```
alonre24, line 2412: We can also test here that HSET returns 1

src/indexer.c
```diff
@@ -139,6 +139,17 @@ static RSDocumentMetadata *makeDocumentId(RedisModuleCtx *ctx, RSAddDocumentCtx
           if (spec->fields[i].types == INDEXFLD_T_VECTOR) {
             RedisModuleString *rmstr = IndexSpec_GetFormattedKey(spec, &spec->fields[i], INDEXFLD_T_VECTOR);
             VecSimIndex *vecsim = openVectorIndex(spec, rmstr, CREATE_INDEX);
+            // ####################
+            if(!vecsim)
```
alonre24, line 143: We should probably not return an error here, just continue

src/spec.c
```diff
@@ -2999,6 +2999,16 @@ void IndexSpec_DeleteDoc_Unsafe(IndexSpec *spec, RedisModuleCtx *ctx, RedisModul
           continue;
         }
         VecSimIndex *vecsim = kdv->p;
+        if(!vecsim)
```
alonre24, line 3002: We should use `openVectorIndex` API without lazy creation here

src/debug_commands.c
```diff
@@ -1151,7 +1151,18 @@ DEBUG_COMMAND(VecsimInfo) {
...
   // (or should exist, and this call will create it).
-  VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName, CREATE_INDEX);
+  VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName,CREATE_INDEX);
+  // @OMER - not checking ptr
+  // @OMER c1 - the comment above sais call can't fail
+  //         but what if the specs are "bad" so no creation-
+  //         where does the specs being checked?
+  // if null return error "cannot open vector index"
+
+  if(!vecsimIndex)
```
alonre24, line 1161: Let's add tests to this as well

tests/pytests/test_debug_commands.py
```diff
@@ -359,3 +359,17 @@ def testSpecIndexesInfo(env: Env):
     expected_reply["inverted_indexes_memory"] = getInvertedIndexInitialSize(env, ['NUMERIC'])
     debug_output = env.cmd(debug_cmd(), 'SPEC_INVIDXES_INFO', 'idx')
     env.assertEqual(to_dict(debug_output), expected_reply)
+
+# @skip(cluster=True)
+def testVecsimInfo(env: Env):
```
GuyAv46, line 364: should it run in cluster mode or not? Lets not keep the decorator commented out

tests/pytests/test_vecsim.py
```diff
@@ -2394,3 +2390,28 @@ def test_switch_write_mode_multiple_indexes(env):
...
+
+    res = conn.execute_command('FT.CREATE', 'idx','SCHEMA', 'n', 'NUMERIC',
+                    'v', 'VECTOR', 'HNSW', '8', 'TYPE', 'FLOAT16', 'DIM', dim, 'DISTANCE_METRIC', 'L2', 'M', M) 
+    env.assertEqual(res, 'OK')
+
+    res = conn.execute_command('HSET', 'doc', 'n', 0)
+    env.assertEqual(res, 1)
+    # Before bug fix, the following command would cause a server crash due to the null pointer accsess
+    res = conn.execute_command('HSET', 'doc', 'n', 1)
+    env.assertEqual(res, 0)
```
GuyAv46, line 2414: Is there anything we can observe about the field (from the info/a debug command) that we can add to the test?

tests/pytests/test_vecsim.py
```diff
@@ -2394,3 +2390,28 @@ def test_switch_write_mode_multiple_indexes(env):
...
+
+
+
```
GuyAv46, line 2417: Let's keep the file with a single empty line at the end.
(You can also configure VSCode to do it for you)

tests/pytests/test_vecsim.py
```diff
@@ -2394,3 +2390,28 @@ def test_switch_write_mode_multiple_indexes(env):
...
+    #            Update Doc
+    
+    # HNSW parameters the causes an execution throw (M > UINT16_MAX)
+    UINT16_MAX = 2**16
+    M = UINT16_MAX + 1
+    dim = 2
+
+    res = conn.execute_command('FT.CREATE', 'idx','SCHEMA', 'n', 'NUMERIC',
+                    'v', 'VECTOR', 'HNSW', '8', 'TYPE', 'FLOAT16', 'DIM', dim, 'DISTANCE_METRIC', 'L2', 'M', M) 
+    env.assertEqual(res, 'OK')
```
GuyAv46, line 2408:
```suggestion
    env.expect('FT.CREATE', 'idx','SCHEMA', 'n', 'NUMERIC',
                    'v', 'VECTOR', 'HNSW', '8', 'TYPE', 'FLOAT16', 'DIM', dim, 'DISTANCE_METRIC', 'L2', 'M', M).ok()
```

tests/pytests/test_vecsim.py
alonre24: Also, consider adding another command where we try to add a vector and validate the index error (to test the bad vector indexing creation on the way)

tests/pytests/test_debug_commands.py
alonre24, line 364:
```suggestion
def testVecsimInfo_badParams(env: Env):
```

src/debug_commands.c
```diff
@@ -1218,6 +1223,13 @@ DEBUG_COMMAND(dumpHNSWData) {
   // This call can't fail, since we already checked that the key exists
   // (or should exist, and this call will create it).
   VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName, CREATE_INDEX);
+  if(!vecsimIndex)
```
alonre24, line 1226: Let's add a test for this debug command as well

src/debug_commands.c
```diff
@@ -1151,7 +1151,12 @@ DEBUG_COMMAND(VecsimInfo) {
   }
   // This call can't fail, since we already checked that the key exists
   // (or should exist, and this call will create it).
-  VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName, CREATE_INDEX);
+  VecSimIndex *vecsimIndex = openVectorIndex(sctx->spec, keyName,CREATE_INDEX);
+  if(!vecsimIndex)
+  {
+    SearchCtx_Free(sctx);
+    return RedisModule_ReplyWithError(ctx, "Can't open vector index");
```
alonre24, line 1158: I wonder why codecov says that it isn't covered...

alonre24 (commented):
Nice! Few comments

src/indexer.c
```diff
@@ -138,7 +138,9 @@ static RSDocumentMetadata *makeDocumentId(RedisModuleCtx *ctx, RSAddDocumentCtx
         for (int i = 0; i < spec->numFields; ++i) {
           if (spec->fields[i].types == INDEXFLD_T_VECTOR) {
             RedisModuleString *rmstr = IndexSpec_GetFormattedKey(spec, &spec->fields[i], INDEXFLD_T_VECTOR);
-            VecSimIndex *vecsim = openVectorIndex(spec, rmstr, CREATE_INDEX);
+            VecSimIndex *vecsim = openVectorIndex(spec, rmstr, DONT_CREATE_INDEX);
+            if(!vecsim)
+              continue;
```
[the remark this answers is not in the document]
lerman25: Yes, because basically there is no error for the user to do something about, so why notify him.
It's best to report the error in creation of the index

tests/pytests/test_vecsim.py
lerman25: > Is there anything we can observe about the field (from the info/a debug command) that we can add to the test?
We decided not to raise an error in this scenario

src/vector_index.c
```diff
@@ -38,8 +38,9 @@ VecSimIndex *openVectorIndex(IndexSpec *spec, RedisModuleString *keyName, bool c
   // create new vector data structure
   kdv = rm_calloc(1, sizeof(*kdv));
   kdv->p = VecSimIndex_New(&fieldSpec->vectorOpts.vecSimParams);
-
-  dictAdd(spec->keysDict, keyName, kdv);
+  if(kdv->p){
+    dictAdd(spec->keysDict, keyName, kdv);  
+  }
```
GuyAv46, line 43: Why did we added this change? This is a leak of `kdv` first of all, and also we will try to create the index every time (and fail every time) this function is called with the create flag

lerman25: It's a solution for a bug,
In flushall, it calls valFreeCb (spec.c, 1882) that frees each key in kdv, which causes a crash if the index pointer is added to kdv even tough it's null.
I think it's ok, cause if kdv->p is null it means the index couldn't be opened so it's ok we try (and fail) again.

GuyAv46: OK, I agree it's fine to retry and fail to create each time in these rare cases.
But we do need to free the `kdv` allocation if we don't add it to the dictionary (or avoid allocating it)

GuyAv46 (approved):

tests/pytests/test_vecsim.py
```diff
@@ -2394,3 +2390,35 @@ def test_switch_write_mode_multiple_indexes(env):
...
+    UINT16_MAX = 2**16
+    M = UINT16_MAX + 1
+    dim = 4
+
+    env.expect('FT.CREATE', 'idx','SCHEMA', 'n', 'NUMERIC',
+                    'v', 'VECTOR', 'HNSW', '8', 'TYPE', 'FLOAT16', 'DIM', dim, 'DISTANCE_METRIC', 'L2', 'M', M).ok()
+
+    res = conn.execute_command('HSET', 'doc', 'n', 0)
+    env.assertEqual(res, 1)
+    # Before bug fix, the following command would cause a server crash due to the null pointer accsess
```
alonre24, line 2411:
```suggestion
    # Before bug fix, the following command would cause a server crash due to null pointer access to the vector index that filed to be created.
```

tests/pytests/test_vecsim.py
```diff
@@ -2394,3 +2390,35 @@ def test_switch_write_mode_multiple_indexes(env):
...
+    
+    # Sanity check - insert a vector, expect indexing faliure
+    res = conn.execute_command('HSET', 'doc1', 'v', create_np_array_typed([0]*dim,'FLOAT16').tobytes())
+    env.assertEqual(res, 1)
+
+    index_errors_dict = index_errors(env, 'idx')
+    env.assertEqual(index_errors_dict['last indexing error'], "Could not open vector for indexing")
+
+    # Check FlushAll OK - before bug fix, the following command would cause a server crash due to the null pointer accsess
+    env.expect('FLUSHALL').noError()
```
alonre24, line 2423: BTW, before the fix, we got OK from flushall, and then crashed in the background thread no?

alonre24 (commented):
Looks good! only two small comments

tests/pytests/test_vecsim.py
lerman25: Yes that's right, that means that the test will pass even if the bug is still there.
maybe I should a ping after flushall and expect a pong?

alonre24: Yes that's a good idea.
The test would have probably failed anyway since when the test ends there is another call to `FLUSHALL` by RLTest, but it is still better to rely on our check rather than the infra :)

alonre24 (approved):

PR #5439, opened 2025-01-05
MOD-5437: Report GIL time in FT.PROFILE

Itzikvaknin:
This PR introduces GIL lock time reporting to FT.PROFILE.

GIL holding time is measured and reported in two scopes:

- Query scope: This includes the main thread and safeloader result processor in the background configuration. In the foreground configuration, it is measured end-to-end.

- Safeloader result processor scope: This measures the time the safe loader result processor held the GIL.

GIL time is measured using the wall clock rather than the standard CPU time clock (the clock_t clock() function commonly used in the code).


src/profile.c
```diff
@@ -83,6 +82,11 @@ static double _recursiveProfilePrint(RedisModule_Reply *reply, ResultProcessor *
       case RP_MAX:
         RS_LOG_ASSERT(0, "RPType error");
         break;
+      case RP_SAFE_LOADER:
+        printProfileType(RPTypeToString(rp->type));
+        double rpGILTime = (double)RPProfile_GetGILTime(rp) / CLOCKS_PER_MILLISEC;
+        printProfileGILTime(rpGILTime);
+        break;
```
GuyAv46, line 89: 1. Let's put this case above the `RP_PROFILE` and `RP_MAX` handling
2. Why doesn't any test cover this case?

src/result_processor.c
```diff
@@ -1070,6 +1073,10 @@ clock_t RPProfile_GetClock(ResultProcessor *rp) {
   return self->profileTime;
 }
 
+clock_t RPProfile_GetGILTime(ResultProcessor *rp){
+  return rp->totalGILTime;
+}
```
GuyAv46, line 1078: uncovered?

src/result_processor.c
```diff
@@ -891,6 +892,8 @@ static int rpSafeLoaderNext_Accumulate(ResultProcessor *rp, SearchResult *res) {
   // Done loading. Unlock Redis
   RedisModule_ThreadSafeContextUnlock(sctx->redisCtx);
 
+  rp->totalGILTime += clock() - start;
```
GuyAv46, line 895: Is this used for statistics as well or only for profiling?
If only for profiling, consider calling `clock()` conditionally to avoid performance overhead

GuyAv46 (commented):
Nice and tidy

1. Please add a test to see the effect of this code change - you can add it to an existing profiling test if you want, we want to verify the reply structure matches the expectation (exact time value is not important or testable, of course. See how we use the `ANY` value in other tests)
2. Is this the only case in which we want to add the GIL holding time? Don't we also want to take the main-thread time into account?

src/aggregate/aggregate_exec.c
```diff
@@ -933,7 +938,7 @@ static int execCommandCommon(RedisModuleCtx *ctx, RedisModuleString **argv, int
     blockedClientReqCtx *BCRctx = blockedClientReqCtx_New(r, blockedClient, spec_ref);
     // Mark the request as thread safe, so that the pipeline will be built in a thread safe manner
     r->reqflags |= QEXEC_F_RUN_IN_BACKGROUND;
-
+    r->qiter.GILTime +=  IsProfile(r) ? (clock() - r->initClock) : 0;
```
GuyAv46, line 941: shouldn't we take the wall-time here as well?
```suggestion
    if (IsProfile(r)) r->qiter.GILTime += clock() - r->initClock;
```

src/result_processor.c
```diff
@@ -891,6 +898,15 @@ static int rpSafeLoaderNext_Accumulate(ResultProcessor *rp, SearchResult *res) {
...
   RedisModule_ThreadSafeContextUnlock(sctx->redisCtx);
 
+  if (isQueryProfile) {
+    clock_gettime(CLOCK_MONOTONIC, &rpEndTime);
+  } else {
+    rpEndTime.tv_nsec = 0;
+  }
+  double GILDuration =  (rpEndTime.tv_nsec - rpStartTime.tv_nsec)/1000000.0;
+  rp->parent->GILTime += GILDuration;
+  rp->GILTime += GILDuration;
```
GuyAv46, line 908: consider doing these operations only if `isQueryProfile`, and not always with 0 values

src/result_processor.h
```diff
@@ -75,6 +75,10 @@ typedef struct {
   // Contains our spec
   RedisSearchCtx *sctx;
 
+  bool isProfile;
```
GuyAv46, line 78: consider moving this member next to `timeoutPolicy`, so we have less padding in the struct

src/profile.c
```diff
@@ -125,6 +129,18 @@ void Profile_Print(RedisModule_Reply *reply, void *ctx) {
...
             (double)(req->pipelineBuildTime / CLOCKS_PER_MILLISEC));
+      //Print total GIL time
+        if (profile_verbose){
+          if (RunInThread()){
+            RedisModule_ReplyKV_Double(reply, "Total GIL time", req->qiter.GILTime);
+          } else {
+            struct timespec rpEndTime;
+            clock_gettime(CLOCK_MONOTONIC, &rpEndTime);
+            RedisModule_ReplyKV_Double(reply, "Total GIL time",
+              (double)(rpEndTime.tv_nsec - req->qiter.initTime.tv_nsec)/1000000.0);
```
GuyAv46, line 140: Isn't `tv_nsec` holding only up to a second, and are the whole seconds counted in the `tv_sec`?
I think we have to use a full diff function here. Also, what happens if the start time is "XX:XX:n.99" and the end is "XX:XX:(n+1).01"? Is the diff negative?

tests/pytests/test_profile.py
```diff
@@ -623,3 +625,19 @@ def test_shard_timers(env):
...
+  conn = getConnectionByEnv(env)
+
+  # Populate db
+  for i in range(100):
+    res = conn.execute_command('hset', 'doc%d' % i,
+                                  'f', 'hello world')
+    env.assertEqual(1, res) 
+  env.cmd('ft.create', 'idx', 'SCHEMA', 'f', 'TEXT')
+  res = env.cmd('FT.PROFILE', 'idx', 'AGGREGATE', 'query', 'hello', 'SORTBY', '1', '@f')
+  expected_result_processor_stats = ['Type', 'Threadsafe-Loader', 'GIL-Time', ANY , 'Time', '0', 'Counter', 100]
```
GuyAv46, line 640:
```suggestion
  expected_result_processor_stats = ['Type', 'Threadsafe-Loader', 'GIL-Time', ANY , 'Time', ANY, 'Counter', 100]
```

src/aggregate/aggregate_exec.c
```diff
@@ -908,7 +911,9 @@ static int execCommandCommon(RedisModuleCtx *ctx, RedisModuleString **argv, int
 
   if (!IsInternal(r) || IsProfile(r)) {
     // We currently don't need to measure the time for internal and non-profile commands
-    r->initClock = clock();
+    clock_t initClock = clock();
+    r->initClock = initClock;
```
raz-mon, line 915: Why not `r->initClock = clock()`?

src/result_processor.c
```diff
@@ -891,6 +898,15 @@ static int rpSafeLoaderNext_Accumulate(ResultProcessor *rp, SearchResult *res) {
   // Done loading. Unlock Redis
   RedisModule_ThreadSafeContextUnlock(sctx->redisCtx);
 
+  if (isQueryProfile) {
+    clock_gettime(CLOCK_MONOTONIC, &rpEndTime);
+  } else {
+    rpEndTime.tv_nsec = 0;
+  }
+  double GILDuration =  (rpEndTime.tv_nsec - rpStartTime.tv_nsec)/1000000.0;
```
raz-mon, line 906: Consider using `rs_timersub`, seems like what you want here.

src/profile.c
raz-mon: See `rs_timersub`

src/result_processor.c
```diff
@@ -883,6 +883,13 @@ static int rpSafeLoaderNext_Accumulate(ResultProcessor *rp, SearchResult *res) {
   // First, we verify that we unlocked the spec before we lock Redis.
   RedisSearchCtx_UnlockSpec(sctx);
 
+  struct timespec rpStartTime, rpEndTime;
+  bool isQueryProfile = rp->parent->isProfile;
+  if (isQueryProfile) {
+    clock_gettime(CLOCK_MONOTONIC, &rpStartTime);
+  } else {
+    rpStartTime.tv_nsec = 0;
```
raz-mon, line 891: `rpStartTime.tv_sec` should be zeroed as well?

tests/pytests/test_profile.py
```diff
@@ -623,3 +625,19 @@ def test_shard_timers(env):
...
   else:
     test_shard_timers(env)
+
+def testPofileGILTime():
+  env = Env(moduleArgs='WORKERS 1')
+  conn = getConnectionByEnv(env)
+
+  # Populate db
+  for i in range(100):
+    res = conn.execute_command('hset', 'doc%d' % i,
```
raz-mon, line 635: As the one that removed this from our python tests.. 🫨

raz-mon (commented):
Nice 👏 
Since we don't see that there is a non-zero value in the tests (can be flaky otherwise) - at least see locally that this is not always `0`

src/profile.h
```diff
@@ -12,7 +12,7 @@
 #define printProfileType(vtype) RedisModule_ReplyKV_SimpleString(reply, "Type", (vtype))
 #define printProfileTime(vtime) RedisModule_ReplyKV_Double(reply, "Time", (vtime))
 #define printProfileCounter(vcounter) RedisModule_ReplyKV_LongLong(reply, "Counter", (vcounter))
-#define printProfileGILTime(vtime) RedisModule_ReplyKV_Double(reply, "GIL-Time", (vtime))
+#define printProfileGILTime(vtime) RedisModule_ReplyKV_Double(reply, "GIL-Time", (vtime.tv_sec*1000.0 + (double)vtime.tv_nsec/1000000.0))
```
GuyAv46, line 15: consider implementing a `rs_timer_ms` helper for that

src/profile.c
```diff
@@ -129,15 +129,18 @@ void Profile_Print(RedisModule_Reply *reply, void *ctx) {
...
-            RedisModule_ReplyKV_Double(reply, "Total GIL time", req->qiter.GILTime);
+            RedisModule_ReplyKV_Double(reply, "Total GIL time", 
+              req->qiter.GILTime.tv_sec*1000.0 + (double)req->qiter.GILTime.tv_nsec/1000000.0);
           } else {
             struct timespec rpEndTime;
             clock_gettime(CLOCK_MONOTONIC, &rpEndTime);
+            rs_timersub(&rpEndTime, &req->qiter.initTime, &rpEndTime);
             RedisModule_ReplyKV_Double(reply, "Total GIL time",
-              (double)(rpEndTime.tv_nsec - req->qiter.initTime.tv_nsec)/1000000.0);
+              rpEndTime.tv_sec*1000 + (double)rpEndTime.tv_nsec/1000000);
```
GuyAv46, line 143: consider implementing a `rs_timer_ms` helper for that

src/profile.c
```diff
@@ -129,15 +129,18 @@ void Profile_Print(RedisModule_Reply *reply, void *ctx) {
         if (profile_verbose)
           RedisModule_ReplyKV_Double(reply, "Pipeline creation time",
             (double)(req->pipelineBuildTime / CLOCKS_PER_MILLISEC));
+
       //Print total GIL time
         if (profile_verbose){
           if (RunInThread()){
-            RedisModule_ReplyKV_Double(reply, "Total GIL time", req->qiter.GILTime);
+            RedisModule_ReplyKV_Double(reply, "Total GIL time", 
+              req->qiter.GILTime.tv_sec*1000.0 + (double)req->qiter.GILTime.tv_nsec/1000000.0);
```
GuyAv46, line 137: consider implementing a `rs_timer_ms` helper for that

src/result_processor.c
```diff
@@ -898,14 +894,11 @@ static int rpSafeLoaderNext_Accumulate(ResultProcessor *rp, SearchResult *res) {
   // Done loading. Unlock Redis
   RedisModule_ThreadSafeContextUnlock(sctx->redisCtx);
 
-  if (isQueryProfile) {
+  if (isQueryProfile){
```
GuyAv46, line 897: :(
```suggestion
  if (isQueryProfile) {
```

src/result_processor.c
```diff
@@ -883,13 +883,9 @@ static int rpSafeLoaderNext_Accumulate(ResultProcessor *rp, SearchResult *res) {
...
   RedisSearchCtx_UnlockSpec(sctx);
 
-  struct timespec rpStartTime, rpEndTime;
   bool isQueryProfile = rp->parent->isProfile;
-  if (isQueryProfile) {
-    clock_gettime(CLOCK_MONOTONIC, &rpStartTime);
-  } else {
-    rpStartTime.tv_nsec = 0;
-  }
+  struct timespec rpStartTime = {0}, rpEndTime = {0};
```
GuyAv46, line 887: has to be initialized?

src/result_processor.c
```diff
@@ -898,14 +894,11 @@ static int rpSafeLoaderNext_Accumulate(ResultProcessor *rp, SearchResult *res) {
   // Done loading. Unlock Redis
   RedisModule_ThreadSafeContextUnlock(sctx->redisCtx);
 
-  if (isQueryProfile) {
+  if (isQueryProfile){
     clock_gettime(CLOCK_MONOTONIC, &rpEndTime);
-  } else {
-    rpEndTime.tv_nsec = 0;
+    rs_timersub(&rpEndTime, &rpStartTime, &rp->GILTime);
```
GuyAv46, line 899: In some cases, we may use the `LOADER` multiple times. We should sum up the GIL time here as well, and not override it

tests/pytests/common.py
```diff
@@ -743,6 +743,21 @@ def check_index_info_empty(env, idx, fields, msg="after delete all and gc", dept
...
+                return [i] + sublist_index
+        elif element == target:
+            return [i]
+    return None  
+
+def access_nested_list(y: List[List[Any]], x: List[int]) -> Any:
+    result = y
+    for index in x:
+        result = result[index] 
+    return result
```
GuyAv46, line 760: Nice

tests/pytests/test_profile.py
```diff
@@ -631,13 +631,30 @@ def testPofileGILTime():
...
+  rp_GIL_time = rp_record[rp_record.index('GIL-Time') + 1]
+
+  # extract the total GIL time
+  total_GIL_index = recursive_index(res, 'Total GIL time')
+  total_GIL_index[-1] += 1
+  total_GIL_time = access_nested_list(res, total_GIL_index)
+
+  env.assertGreaterEqual(float(total_GIL_time), 0)
+  env.assertGreaterEqual(float(rp_GIL_time), 0)
+  env.assertGreaterEqual(float(total_GIL_time), float(rp_GIL_time))
```
GuyAv46, line 660: missing a new line

tests/pytests/test_profile.py
```diff
@@ -657,4 +657,5 @@ def testPofileGILTime():
 
   env.assertGreaterEqual(float(total_GIL_time), 0)
   env.assertGreaterEqual(float(rp_GIL_time), 0)
-  env.assertGreaterEqual(float(total_GIL_time), float(rp_GIL_time))
\ No newline at end of file
+  env.assertGreaterEqual(float(total_GIL_time), float(rp_GIL_time))
+
```
GuyAv46, line 661:
```suggestion

```

GuyAv46 (dismissed):
Nicky done

tests/pytests/test_profile.py
Itzikvaknin: Done 👍

GuyAv46 (dismissed):

src/profile.c
```diff
@@ -126,6 +130,19 @@ void Profile_Print(RedisModule_Reply *reply, void *ctx) {
           RedisModule_ReplyKV_Double(reply, "Pipeline creation time",
             (double)(req->pipelineBuildTime / CLOCKS_PER_MILLISEC));
 
+      //Print total GIL time
+        if (profile_verbose){
+          if (RunInThread()){
+            RedisModule_ReplyKV_Double(reply, "Total GIL time", 
+            rs_timer_ms(&req->qiter.initTime));
```
raz-mon, line 137: This is just the initial time, not the GIL time.
From a small test I did locally, I got:
```
7) Total GIL time
8) "1148342933.676"
```

IIUC, you already have the right time in `req->qiter.GILTime`

src/result_processor.c
```diff
@@ -891,6 +894,13 @@ static int rpSafeLoaderNext_Accumulate(ResultProcessor *rp, SearchResult *res) {
   // Done loading. Unlock Redis
   RedisModule_ThreadSafeContextUnlock(sctx->redisCtx);
 
+  if (isQueryProfile) {
+    clock_gettime(CLOCK_MONOTONIC, &rpEndTime);
+    rs_timersub(&rpEndTime, &rpStartTime, &rpEndTime);
+    rs_timeradd(&rpEndTime, &rp->GILTime, &rp->GILTime);
+    rs_timeradd(&rpEndTime, &rp->parent->GILTime, &rp->parent->GILTime);
```
raz-mon, line 901: Why do we need both increments? Seems redundant..
Are both printed\used? What's the difference between them?

src/result_processor.h
```diff
@@ -75,6 +75,9 @@ typedef struct {
   // Contains our spec
   RedisSearchCtx *sctx;
 
+  struct timespec initTime; //used with clock_gettime(CLOCK_MONOTONIC, ...)
+  struct timespec GILTime; // milliseconds
```
raz-mon, line 79:
```suggestion
  struct timespec initTime; //used with clock_gettime(CLOCK_MONOTONIC, ...)
  struct timespec GILTime;  // milliseconds
```

tests/pytests/common.py
raz-mon: Like

raz-mon (changes requested):
Nice!
See some comments. Mainly - An issue with the BG GIL time report

raz-mon (approved):
👌

PR #5441, opened 2025-01-05
Remove redundant implementation of strtolower

Itzikvaknin:
Remove redundant implementation of strtolower


src/util/misc.h
```diff
@@ -14,8 +14,10 @@
  */
 void GenericAofRewrite_DisabledHandler(RedisModuleIO *aof, RedisModuleString *key, void *value);
 
+//null-unsafe
 char *strtolower(char *str);
```
nafraf, line 18: Hi @Itzikvaknin
Can we remove this `strtolower()`? Because the same function is defined in `strconv.h` and I think we are not using this.

Itzikvaknin: Done

tests/ctests/test_misc.c
```diff
@@ -0,0 +1,29 @@
...
+    ASSERT_EQUAL(GetRedisErrorCodeLength("WRONGTYPE Operation"), 9);
+    
+    // Test multiple spaces
+    ASSERT_EQUAL(GetRedisErrorCodeLength("ERR multiple spaces here"), 3);
+    
+    return 0;
+}
+
+TEST_MAIN({
+  TESTFUNC(test_get_redis_error_code_length);
```
nafraf, line 28: I see that `GetRedisErrorCodeLength()` is also tested in `test_error_parsing.c`, should we merge the tests in a single file?

src/util/misc.h
```diff
@@ -14,8 +14,7 @@
  */
 void GenericAofRewrite_DisabledHandler(RedisModuleIO *aof, RedisModuleString *key, void *value);
 
-char *strtolower(char *str);
-
+//null-unsafe
```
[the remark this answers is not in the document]
nafraf: @Itzikvaknin We can remove this comment, right?

kei-nan (approved):

PR #5442, opened 2025-01-05
MOD-7595: Use Open Key Expiration Flags To Access Lazily Expired Keys/Fields

kei-nan:
We use the flags redis recently added in order to read lazily expired keys and fields.
This solves the issue where the user could receive None in the query results when a document/field was expired.

1. Current: User can receive None as a valid result in the query
2. Change: Use new redis flags to avoid the lazy expiration of the key/field
3. Outcome: The user will receive a more correct set of results.

1. Opening of keys

#### Mark if applicable


src/document.h
```diff
@@ -140,6 +140,13 @@ typedef struct {
   RedisModuleString *languageStr;  // Language string for HSET
 } AddDocumentOptions;
 
+// When indexing the document we are okay with it being expired, saves us the work of indexing it
+// If the fields lazily expire then we simply don't index them
+#define DOCUMENT_OPEN_KEY_INDEXING_FLAGS REDISMODULE_READ | REDISMODULE_OPEN_KEY_NOEFFECTS
+// When loading the document we are after the iterators phase, where we already verified the expiration time of the field and document
+// We don't allow any lazy expiration to happen here
+#define DOCUMENT_OPEN_KEY_QUERY_FLAGS REDISMODULE_READ | REDISMODULE_OPEN_KEY_NOEFFECTS | REDISMODULE_OPEN_KEY_NOEXPIRE | REDISMODULE_OPEN_KEY_ACCESS_EXPIRED
```
raz-mon, line 148: Seems that `REDISMODULE_OPEN_KEY_NOEFFECTS` ironically has no effect

tests/pytests/test_replicate.py
```diff
@@ -243,8 +243,8 @@ def testExpireDocsSortable():
                # the documents are ordered according to the sortkey values.
                # However, the loader fails to load doc1 and the result is marked as expired so
                # the value does not appear in the result.
-             [2, 'doc2', ['t', 'foo'], 'doc1', None],  # Without sortby - ordered by docid, notice doc1 was expired so the notification pushed it to the back of the line
-               [2, 'doc1', None, 'doc2', ['t', 'foo']])  # With sortby - ordered by the original value, bar > foo
+             [2, 'doc2', ['t', 'foo'], 'doc1', ['t', 'bar']],  # Without sortby - ordered by docid, notice doc1 was expired so the notification pushed it to the back of the line
+               [2, 'doc1', ['t', 'bar'], 'doc2', ['t', 'foo']])  # With sortby - ordered by the original value, bar > foo
```
raz-mon, line 247: Nice!

src/result_processor.c
```diff
@@ -128,7 +128,7 @@ static int rpidxNext(ResultProcessor *base, SearchResult *res) {
     } else {
       dmd = DocTable_Borrow(docs, r->docId);
     }
-    if (!dmd || (dmd->flags & Document_Deleted) || DocTable_IsDocExpired(docs, dmd, &sctx->time.current)) {
+    if (!dmd || (dmd->flags & Document_Deleted)) {
```
raz-mon, line 131: Why was this check removed?

src/rlookup.c
```diff
@@ -679,7 +679,7 @@ static int getKeyCommonJSON(const RLookupKey *kk, RLookupRow *dst, RLookupLoadOp
 
     if (japi_ver >= 5) {
       RedisModuleString* keyName = RedisModule_CreateString(ctx, keyPtr, strlen(keyPtr));
-      *keyobj = japi->openKeyWithFlags(ctx, keyName, REDISMODULE_OPEN_KEY_NOEFFECTS);
```
raz-mon, line 682: Was the read flag here just missed before?

src/rlookup.c
```diff
@@ -884,7 +884,7 @@ static int RLookup_JSON_GetAll(RLookup *it, RLookupRow *dst, RLookupLoadOptions
   RedisJSON jsonRoot;
   if (japi_ver >= 5) {
     RedisModuleString* keyName = RedisModule_CreateString(ctx, options->dmd->keyPtr, strlen(options->dmd->keyPtr));
-    jsonRoot = japi->openKeyWithFlags(ctx, keyName, REDISMODULE_OPEN_KEY_NOEFFECTS);
```
raz-mon, line 887: Here also

raz-mon (commented):
Nice!

src/result_processor.c
kei-nan: I guess it would be better to check.
If the document wasn't actively expired/deleted then we could theoretically still access it and open it.
If in the past we got null then we could get a valid document.
But per the rule of expiration we should probably check it against the query start time.

Will bring back old code.

src/rlookup.c
kei-nan: I guess so.
Not sure what the implications were.

src/rlookup.c
kei-nan: I guess so, maybe JSON adds it to the mask internally if it isn't set?

src/document.h
kei-nan: not sure what you mean.
Perhaps internally there is an overlap but API wise it affects the resulting bit mask.

raz-mon: Not here, in Redis.

src/rlookup.c
raz-mon: Let's give this a quick check before we change it. Probably either forgot or indeed added implicitly.

src/result_processor.c
raz-mon: Because it's OK semantically not to check this here. It's better to check, if the price is not too high.

src/document.h
kei-nan: Yes, that is what I meant by saying internally.
But I think we can only rely on the API they provide and not on any current internal implementation they may use.

raz-mon: If that API does nothing and is obsolete, we shouldn't use it.
This is not a change we need to do here, but we could have.
I'll take a deeper look later on.

kei-nan: I did not say the API is obsolete.

Redis defined an API for us to use.
We can't rely on any current implementation (which can change at any moment) they may have at this time as a replacement for using the defined API.

src/rlookup.c
kei-nan: Looked at redis code:
In RM_OpenKey they currently have:
```cpp
    if (mode & REDISMODULE_WRITE) {
        value = lookupKeyWriteWithFlags(ctx->client->db,keyname, flags);
    } else {
        value = lookupKeyReadWithFlags(ctx->client->db,keyname, flags);
        if (value == NULL) {
            return NULL;
        }
    }
```
So they only check for the wrtie bit.
Regardless it would be better we use the defined API.

src/rlookup.c
kei-nan: Looked at redis code:
In RM_OpenKey they have:
```cpp
    if (mode & REDISMODULE_WRITE) {
        value = lookupKeyWriteWithFlags(ctx->client->db,keyname, flags);
    } else {
        value = lookupKeyReadWithFlags(ctx->client->db,keyname, flags);
        if (value == NULL) {
            return NULL;
        }
    }
```
So they only check for the write bit

src/document.h
raz-mon: I said it, because that's the way it looks.
We should clarify this and make sure we're using the right flags, I'm not sure what part of that you don't agree with.
Anyways, as I said before, you can drop it as it's not an effort that has to be done in this PR.

src/result_processor.c
raz-mon: I see that we do 90% of the doc-expiration work in the field-expiration check that we do in the iterators level (the doc expiration test is done only in the `rpidxNext` result-processor.
I would at least check the possibility of adding the doc expiration check to the iterators, as it seems to be small relative to the field expiration work, and we already do most of the work.

Can and should also be done separately from this PR, since this was not changed here.

raz-mon (approved):

src/document.h
oshadmi: We should keep it since it has effect 😄

raz-mon: Yup, it does.
Misread some of the redis code.

src/rlookup.c
oshadmi: > maybe JSON adds it to the mask internally if it isn't set?

It doesn't

src/rlookup.c
oshadmi: And since it is now aligned with `getKeyCommonHash`, we can assume it is OK here as well (could fail if read from a read-only replica - but this is the behavior with HASH)

oshadmi (dismissed):

src/document.h
```diff
@@ -140,6 +140,13 @@ typedef struct {
   RedisModuleString *languageStr;  // Language string for HSET
 } AddDocumentOptions;
 
+// When indexing the document we are okay with it being expired, saves us the work of indexing it
```
raz-mon, line 143: Notice that the key isn't actually expired due to the `NOEFFECTS` flag. We just do not read it.

sbin/get-redisjson
```diff
@@ -91,18 +91,10 @@ else
...
-			STAGING=1
-		fi
 	elif [[ $os == macos ]]; then
 		# as we don't build on macOS for every platform, we converge to a least common denominator
 		if [[ $arch == x86_64 ]]; then
-			nick=catalina  # to be aligned with the rest of the modules in redis stack
-		else
-			[[ $nick == ventura ]] && nick=monterey
-			[[ $nick == sonoma ]] && nick=monterey
+			nick=ventura  # to be aligned with the rest of the modules in redis stack
```
raz-mon, line 97: I think we can leave Sonoma to use Sonoma no?

sbin/get-redisjson
```diff
@@ -91,18 +91,10 @@ else
 		fi
 	elif [[ $dist == mariner ]]; then
```
nafraf, line 92: I think we can remove this condition,  we set `OSS=1` in the `else`

sbin/get-redisjson
kei-nan: discussed offline, it should stay sonoma if arch is not x86_64.

oshadmi (dismissed):

raz-mon (dismissed):

oshadmi (dismissed):

raz-mon (dismissed):

oshadmi (approved):

raz-mon (approved):

PR #5460, opened 2025-01-06
Fix timeout with return partial results policy bug, implement debug timeout mock

meiravgri:
This PR addresses multiple bugs that occur when the timeout policy is return partial results.

## [MOD-8483] Bug 1: Cursor with `SORTBY` is never depleted
When a cursor read with a sorter encounters a timeout during the sorter’s accumulation phase, the sorter switches to yield mode and sets its `timedOut` flag.
After this, it begins emptying its heap, returning the results collected up to the timeout. 
During this process, each returned result has the status `RS_RESULT_OK.`
Once the heap is emptied, the sorter returns `RS_RESULT_TIMEDOUT`, and the current read command is done, with a valid cursor id. 
```
return self->timedOut ? RS_RESULT_TIMEDOUT : RS_RESULT_EOF;
```
At this point, every subsequent `FT.CURSOR READ` call receives `RS_RESULT_TIMEDOUT,` even though no more results are available. Since the cursor is only considered depleted when RS_RESULT_EOF is returned, it remains open indefinitely despite having no results left to provide.

### Fix:
To address this issue, the `timedOut` flag is now reset after the timeout status is returned. This ensures that subsequent `FT.CURSOR READ` commands return `RS_RESULT_EOF,` allowing the cursor to be properly marked as depleted once all results have been processed.

## [MOD-8482] Bug 2: Serializing empty result with resp3
in `sendChunk_Resp3`, During the initial execution of the query pipeline, if a timeout occurred before any results were successfully collected, the return code (rc) was incorrectly changed from `RS_RESULT_TIMEDOUT` to `RS_RESULT_OK` This caused an empty result to be serialized and incorrectly returned as a valid result.

Additionally, in standalone mode (SA), the `RP_INDEX` result processor checks for timeouts every 100 results. If a timeout occurred during the first pipeline run, the timeout counter of `RP_INDEX` was reset to 0, allowing up to 99 valid results to be returned before the next timeout check, thereby violating the timeout limit.

For subsequent pipeline runs, the rc for `RS_RESULT_TIMEDOUT` was handled correctly and not altered.

### Fix:
The fix ensures that the rc is not changed when a timeout occurs. 

## [MOD-8515] Bug 3: `FT.AGGREGATE WITHCURSOR` hangs when shards return 0 results, but cursor is not depleted
When shards return an empty results array (length 0), `MRIteratorCallback_AddReply` is not called to push the result to the channel, leaving the channel waiting indefinitely in `MRChannel_Pop`.

Additionally, since the cursor is not depleted `the MRIteratorCallback_Done` function—which decreases the pending count and calls `MRChannel_Unblock` to wake the waiting `MRChannel_Pop`—is skipped. 
Instead, `MRIteratorCallback_ProcessDone` is called, completing the results collection from the shard but leaving `MRChannel_Pop` in a hanging state.

### Fix:
Move `MRChannel_Unblock` call to `MRIteratorCallback_ProcessDone`, ensuring it is executed even when the cursor is not depleted. This change releases the conditional wait in `MRChannel_Pop`, preventing the hang.
Additionally, when no results are returned from the shards, we avoid returning EOF. Instead, a valid cursor ID is returned gracefully, allowing the next `FT.CURSOR READ` to attempt to collect results.

## [MOD-8606] Bug 4: `RP_Pager` incorrectly updates its counter when no results are received
When used with cursors, this behavior causes the counter to be incorrectly updated for each `FT.CURSOR READ` that times out. As a result, the total number of results returned by all cursor reads is less than the requested `LIMIT`, specifically reduced by the number of `FT.CURSOR READ` commands that encountered a timeout.
for example:

```
FT.AGGREGATE WITHCURSOR count 500000 LIMIT 0 500 TIMEOUT 1
- first read: 399 results returned (timedout)
Expected pager remaining: 101, but it is incorrectly set to 100.
- Next read: The pager counter reaches 0 after 100 results and returns EOF.
Cursor ID is set to 0 since EOF was reached. 
Total results: 499 instead of the requested 500.
```
### Fix:
Ensure that the counter is only updated when the upstream RPs return `RS_RESULT_OK`. This prevents erroneous decrements when no results are received, maintaining the correct count and ensuring the LIMIT is respected.

[MOD-8483]: https://redislabs.atlassian.net/browse/MOD-8483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MOD-8482]: https://redislabs.atlassian.net/browse/MOD-8482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MOD-8515]: https://redislabs.atlassian.net/browse/MOD-8515
[MOD-8606]:https://redislabs.atlassian.net/browse/MOD-8606


meiravgri:
split into
https://github.com/RediSearch/RediSearch/pull/5556 
and 
https://github.com/RediSearch/RediSearch/pull/5554

PR #5514, opened 2025-01-15
Fix test FT.PROFILE report GIL time

Itzikvaknin:
- Added unittests for utility functions in test_common.py

- Replaced env.assertContains() with env.assertTrue(recursive_contains(...)) due to unexpected behavior of assertContains when using contains ANY

- Minor change in qiter->isProfile boolean usage

### DIFF
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -912,6 +912,9 @@ static int execCommandCommon(RedisModuleCtx *ctx, RedisModuleString **argv, int 
   if (!IsInternal(r) || IsProfile(r)) {
     // We currently don't need to measure the time for internal and non-profile commands
     r->initClock = clock();
+  }
+
+  if (r->qiter.isProfile) {
     clock_gettime(CLOCK_MONOTONIC, &r->qiter.initTime);
   }
 
--- a/tests/pytests/common.py
+++ b/tests/pytests/common.py
@@ -751,10 +751,10 @@ def recursive_index(lst, target):
                 return [i] + sublist_index
         elif element == target:
             return [i]
-    return None
+    return -1
 
 def recursive_contains(lst, target):
-    return recursive_index(lst, target) is not None
+    return recursive_index(lst, target) != -1
 
 def access_nested_list(lst, index):
     result = lst
--- a/tests/pytests/common.py
+++ b/tests/pytests/common.py
@@ -751,12 +751,12 @@ def recursive_index(lst, target):
                 return [i] + sublist_index
         elif element == target:
             return [i]
-    return None  
+    return None
 
-def access_nested_list(y: List[List[Any]], x: List[int]) -> Any:
-    result = y
-    for index in x:
-        result = result[index] 
+def access_nested_list(lst, index):
+    result = lst
+    for entry in index:
+        result = result[entry]
     return result
 
 def downloadFile(env, file_name, depth=0):
--- a/tests/pytests/common.py
+++ b/tests/pytests/common.py
@@ -753,6 +753,9 @@ def recursive_index(lst, target):
             return [i]
     return None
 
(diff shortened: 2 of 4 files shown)


src/aggregate/aggregate_exec.c
```diff
@@ -912,6 +912,9 @@ static int execCommandCommon(RedisModuleCtx *ctx, RedisModuleString **argv, int
   if (!IsInternal(r) || IsProfile(r)) {
     // We currently don't need to measure the time for internal and non-profile commands
     r->initClock = clock();
+  }
+
+  if (r->qiter.isProfile) {
```
raz-mon, line 917: Have you considered consolidating all our clocks to use the `struct timespec` instead of the current redundancy? Obviously it is caused due to our mixed usage of the Wall-clock and the CPU time (via the `clock()` function)?
This will actually be great for us, since it will guarantee a minimal precision of nanoseconds, remove flakiness that we may receive from the current CPU time clock, and unify our clock mechanisms to the more precise option.

raz-mon: We don't have much of the CPU counter-clocks.

tests/pytests/common.py
```diff
@@ -751,12 +751,15 @@ def recursive_index(lst, target):
...
             return [i]
-    return None  
+    return None
 
-def access_nested_list(y: List[List[Any]], x: List[int]) -> Any:
-    result = y
-    for index in x:
-        result = result[index] 
+def recursive_contains(lst, target):
+    return recursive_index(lst, target) is not None
```
raz-mon, line 757: Maybe `-1` would be a more intuitive return value for the case the value is not found (in the `recursive_index` function)

tests/pytests/test_profile.py
```diff
@@ -631,19 +631,23 @@ def testPofileGILTime():
...
+                      'g', 'foo bar',
+                      'h', 'baz qux')
 
   env.cmd('ft.create', 'idx', 'SCHEMA', 'f', 'TEXT', 'g', 'TEXT', 'h', 'TEXT')
   res = env.cmd('FT.PROFILE', 'idx', 'AGGREGATE', 'query', 'hello' ,'SORTBY', '1', '@f')
-  expected_result_processor_stats = ['Type', 'Threadsafe-Loader', 'GIL-Time', ANY , 'Time', ANY, 'Counter', 100]
-  expected_total_GIL_time = ['Total GIL time', ANY]
-  env.assertContains(res, expected_result_processor_stats)
-  env.assertContains(res[1][1], expected_total_GIL_time)
+  env.assertNotEqual(res, None)
```
raz-mon, line 643: When would it be `None`? 🤔 
This check can be removed

tests/pytests/test_profile.py
```diff
@@ -631,19 +631,23 @@ def testPofileGILTime():
...
 
+  # Record structure:
+  # ['Type', 'Threadsafe-Loader', 'GIL-Time', ANY , 'Time', ANY, 'Counter', 100]
+  # ['Total GIL time', ANY]
+
+  env.assertTrue(recursive_contains(res, 'Threadsafe-Loader'))
+  env.assertTrue(recursive_contains(res, 'Total GIL time'))
```
raz-mon, line 650: Sweet

raz-mon (commented):
Cool.
See the comment about refactoring our CPU time clocks to Wall-time clocks. This doesn't have to be in this PR (maybe even shouldn't be), but it IS related, since the new `init_time` time that you added is redundant w.r.t the `initClock` that already existed.
It is the right solution taking into account our current use of the CPU time, but that should just be refactored away.

tests/pytests/common.py
Itzikvaknin: Agree 👍

raz-mon (approved):

PR #5520, opened 2025-01-17
Fix APPLY/FILTER parser - [MOD-7804]

GuyAv46:

Refactor parser to follow the expected order of operations.

This PR also enables some internal optimizations to the expression AST, such as:
1. perform in-place numeric calculations for literals, when possible
2. Double negation cancels each other and yields the inner node

1. MOD-6835

1. `APPLY`/`FILTER` parser

#### Mark if applicable


src/aggregate/expr/parser.y
```diff
@@ -57,40 +59,81 @@
...
+        RSExpr_Free(B);
+    } else {
+        A = RS_NewInverted(B);
+    }
+}
 
-expr(A) ::= STRING(B). { A =  RS_NewStringLiteral((char*)B.s, B.len); }
-expr(A) ::= number(B). { A = RS_NewNumberLiteral(B); }
+expr(A) ::= STRING(B). { A = RS_NewStringLiteral(B.s, B.len); }
+expr(A) ::= number(B). [LOWEST] { A = RS_NewNumberLiteral(B); }
```
kei-nan, line 114: Out of curiosity what does the [LOWEST] do?

src/aggregate/expr/parser.y
```diff
@@ -57,40 +59,81 @@
...
+        if (!cb) { // Function not found
+            rm_asprintf(&ctx->errorMsg, "Unknown function name '%.*s'", B.len, B.s);
+        } else { // Argument count mismatch
+            if (cb->minArgs == cb->maxArgs) {
+                rm_asprintf(&ctx->errorMsg, "Function '%.*s' expects %d arguments, but got %d", B.len, B.s, cb->minArgs, C->len);
+            } else {
+                rm_asprintf(&ctx->errorMsg, "Function '%.*s' expects between %d and %d arguments, but got %d",
+                                                B.len, B.s, cb->minArgs, cb->maxArgs, C->len);
+            }
+        }
```
kei-nan, line 133:
```suggestion
    RSFunctionInfo *cb = RSFunctionRegistry_Get(B.s, B.len);
    if (!cb) {
        rm_asprintf(&ctx->errorMsg, "Unknown function name '%.*s'", B.len, B.s);
    } else if (cb->minArgs <= C->len && C->len <= cb->maxArgs) {
        rm_asprintf(&ctx->errorMsg, "Function '%.*s' expects %d arguments, but got %d", B.len, B.s, cb->minArgs, C->len);
    } else {
        A = RS_NewFunc(cb, C);
    }
```

src/aggregate/functions/geo.c
```diff
@@ -37,27 +35,45 @@ static int parseField(RSValue *argv, double *geo, QueryError *status) {
...
+
+  case 4:
+    for (int i = 0, j = 0; i < 2; i++, j += 2) {
+      rv = parseLonLat(&argv[j], &argv[j + 1], geo[i]);
+      if (rv != REDISMODULE_OK) goto error;
+    }
+    break;
+
+  case 3:
+    if (RSValue_ToNumber(&argv[0], &dummy)) {
```
kei-nan, line 63: don't we prefer to have a stricter ordering on what comes first?

tests/cpptests/test_cpp_expr.cpp
```diff
@@ -93,14 +99,53 @@ TEST_F(ExprTest, testExpr) {
...
+    ASSERT_EQ(EXPR_EVAL_OK, ctx.eval());            \
+    auto res = RSValue_Dereference(&ctx.result());  \
+    ASSERT_EQ(RSValue_Number, res->t);              \
+    ASSERT_EQ(expected, res->numval);               \
+  }
+
+  TEST_ARITHMETIC("3 + 3", 6);
+  TEST_ARITHMETIC("3 - 3", 0);
+  TEST_ARITHMETIC("3 * 3", 9);
+  TEST_ARITHMETIC("3 / 3", 1);
```
kei-nan, line 117: A bit of an edge case, what should the result be if the user divides a number by 0?
Maybe add that as a test case?

kei-nan (changes requested):

src/aggregate/expr/parser.y
GuyAv46: I declared `LOWEST` as the lowest priority token, and I’m using it here to give this rule the lowest priority. This is to enforce the parser to prefer other rules for reducing the `number` token (for instance, the arithmetics rules), and only use this rule as a last resort.

src/aggregate/expr/parser.y
GuyAv46: I ordered this block the way I did so we have a single cleaning path in all the error cases.

src/aggregate/functions/geo.c
GuyAv46: I don’t think so

1. This function is symmetric, so I don’t see a reason to enforce order on text and numeric-pair inputs, as long as the text input is one coordinate and the numeric pair is another
2. All the allowed combinations are explicitly documented today

tests/cpptests/test_cpp_expr.cpp
GuyAv46: Good idea

src/aggregate/expr/parser.y
kei-nan: IMHO a cleaner option would be to store C->len before finding the function.
Then have a function which finds the function and assigns/frees the arguments.
e.g:
`
RSExpr *BindFunction(const char* name, size_t len, RSArgList* args) {
    RSFunctionInfo *cb = RSFunctionRegistry_Get(name, len);
    if (!cb) {
       RSArgList_Free(C);
       return NULL;
    }
    return RS_NewFunc(cb, C);
}
`

kei-nan: IMHO a cleaner option would be to store `C->len` before finding the function(for the length checks).
Then have a function which finds the function and assigns/frees the arguments.
e.g:
```cpp
RSExpr *BindFunction(RSFunctionInfo *cb, RSArgList* args) {
    if (!cb || args->len < cb->minArgs || cb->maxArgs < args->len) {
       RSArgList_Free(C);
       return NULL;
    }
    return RS_NewFunc(cb, args);
}
````

Comment (GuyAv46) [reply src/aggregate/expr/parser.y:133]: But we still need a syntax error message for each case

File: src/aggregate/expr/parser.y [hunk 5]
```diff
@@ -57,40 +59,88 @@
...
+expr(A) ::= expr(B) MINUS  expr(C). { A = RS_NewOp('-', B, C); }
+expr(A) ::= expr(B) POW    expr(C). { A = RS_NewOp('^', B, C); }
+expr(A) ::= expr(B) MOD    expr(C). { A = RS_NewOp('%', B, C); }
+// expr ::= number OP expr
+expr(A) ::= number(B) PLUS   expr(C). { A = RS_NewOp('+', RS_NewNumberLiteral(B), C); }
+expr(A) ::= number(B) DIVIDE expr(C). { A = RS_NewOp('/', RS_NewNumberLiteral(B), C); }
+expr(A) ::= number(B) TIMES  expr(C). { A = RS_NewOp('*', RS_NewNumberLiteral(B), C); }
+expr(A) ::= number(B) MINUS  expr(C). { A = RS_NewOp('-', RS_NewNumberLiteral(B), C); }
+expr(A) ::= number(B) POW    expr(C). { A = RS_NewOp('^', RS_NewNumberLiteral(B), C); }
+expr(A) ::= number(B) MOD    expr(C). { A = RS_NewOp('%', RS_NewNumberLiteral(B), C); }
```
raz-mon, line 77: Is there an added value for these rules other than saving us a reduce and a shift?

src/aggregate/expr/parser.y
```diff
@@ -57,40 +59,88 @@
...
+expr(A) ::= expr(B) MINUS  number(C). { A = RS_NewOp('-', B, RS_NewNumberLiteral(C)); }
+expr(A) ::= expr(B) POW    number(C). { A = RS_NewOp('^', B, RS_NewNumberLiteral(C)); }
+expr(A) ::= expr(B) MOD    number(C). { A = RS_NewOp('%', B, RS_NewNumberLiteral(C)); }
+// number := number OP number. In-place arithmetic, to optimize the AST
+number(A) ::= number(B) PLUS   number(C). { A = B + C; }
+number(A) ::= number(B) DIVIDE number(C). { A = B / C; }
+number(A) ::= number(B) TIMES  number(C). { A = B * C; }
+number(A) ::= number(B) MINUS  number(C). { A = B - C; }
+number(A) ::= number(B) POW    number(C). { A = pow(B, C); }
+number(A) ::= number(B) MOD    number(C). { A = fmod(B, C); }
```
raz-mon, line 91: Nice!

tests/pytests/test_aggregate.py
```diff
@@ -659,7 +659,7 @@ def testModulo(self):
         # With Floats
         res = self.env.cmd('ft.aggregate', 'games', '*',
                                        'APPLY', '547758.3 % 5.1')
-        self.env.assertEqual(res[1][1], '3')
+        self.env.assertEqual(res[1][1], '3.00000000008')
```
raz-mon, line 662: How did this change?

tests/pytests/test_issues.py
```diff
@@ -778,8 +778,8 @@ def test_as_startswith_as(env):
 def test_mod4296_badexpr(env):
   env.expect('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT').equal('OK')
   env.expect('HSET', 'doc', 't', 'foo').equal(1)
-  env.expect('FT.AGGREGATE', 'idx', 'foo', 'LOAD', 1, '@t', 'APPLY', '1%0', 'as', 'foo').equal([1, ['t', 'foo', 'foo', 'nan']])
-  env.expect('FT.AGGREGATE', 'idx', 'foo', 'LOAD', 1, '@t', 'APPLY', '1/0', 'as', 'foo').equal([1, ['t', 'foo', 'foo', 'nan']])
+  env.expect('FT.AGGREGATE', 'idx', 'foo', 'LOAD', 1, '@t', 'APPLY', '1%0', 'as', 'foo').apply(lambda x: str(float(x[-1][-1]))).equal('nan')
+  env.expect('FT.AGGREGATE', 'idx', 'foo', 'LOAD', 1, '@t', 'APPLY', '1/0', 'as', 'foo').equal([1, ['t', 'foo', 'foo', 'inf']])
```
raz-mon, line 782: Why not use the lambda here as well (or not use it above)?

tests/pytests/test_profile.py
```diff
@@ -147,6 +147,14 @@ def testProfileAggregate(env):
                 'apply', 'startswith(@t, "hel")', 'as', 'prefix')
   env.assertEqual(actual_res[1][1][0][5], expected_res)
 
+  expected_res = [['Type', 'Index', 'Counter', 2],
+                  ['Type', 'Loader', 'Counter', 2],
+                  ['Type', 'Projector - Literal banana', 'Counter', 2]]
```
raz-mon, line 152: Not returning error so it won't break?

raz-mon (commented):
Very nice!
Some small comments

src/aggregate/expr/parser.y
GuyAv46: YES

1. the `number OP expr` rule is required for the cases we shifted on the `number` before and we cannot reduce it now that we have advanced to the right `expr`
2. the `expr OP number` rule is required so the priority of the rule is of the **operation** and not the priority of the reducing of `numeric` to `expr`. Otherwise, queries like `@n * 2 + 1` may be interpreted as `@n * 3`

tests/pytests/test_aggregate.py
GuyAv46: previously, the modulo operation was performed on the rounded-down value of the numbers. We now use `fmod` to perform the floating point modulo operation, so we get a different value in this case.
The previous behavior can still be achieved with `floor(547758.3) % floor(5.1)`

tests/pytests/test_issues.py
GuyAv46: some implementations may return `-nan` instead of `nan`. by casting to Python float and back to string we get `nan` in both cases

src/aggregate/expr/parser.y
```diff
@@ -4,17 +4,19 @@
...
 
-%left PLUS MINUS.
-%left DIVIDE TIMES MOD POW.
+%left OR.
+%left AND.
+%right NOT.
 
-%right LP.
-%left RP.
+%nonassoc EQ NE LT LE GT GE.
```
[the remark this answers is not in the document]
GuyAv46: This case is given as an example in the lemon documentation. predicated should be `nonassoc` so expressions like `@n == @n2 == @n3` will cause a syntax error by the parser

src/aggregate/expr/expression.c
```diff
@@ -507,12 +491,13 @@ void RPEvaluator_Reply(RedisModule_Reply *reply, const char *title, const Result
...
 
   char buf[32];
+  size_t len;
   const char *literal;
   RPEvaluator *rpEval = (RPEvaluator *)rp;
   const RSExpr *expr = rpEval->eval.root;
   switch (expr->t) {
     case RSExpr_Literal:
-      literal = RSValue_ConvertStringPtrLen(&expr->literal, NULL, buf, sizeof(buf));
+      literal = RSValue_ConvertStringPtrLen(&expr->literal, &len, buf, sizeof(buf));
```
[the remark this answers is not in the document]
GuyAv46: the cases where the pointer is being set without validation were unreachable before this PR, so it couldn't crash before. This was the only place where the pointer was not set by the caller so I decided to change the call instead of the function

src/aggregate/functions/string.c
```diff
@@ -52,64 +45,46 @@ static int func_matchedTerms(ExprEval *ctx, RSValue *result, RSValue **argv, siz
...
-  }
-
-  size_t sz = 0;
-  char *p;
-
-  if (!(p = (char *)RSValue_StringPtrLen(argv[0], &sz))) {
-    RSValue_MakeReference(result, RS_NullVal());
-    return EXPR_EVAL_OK;
-  }
+#define stringfunc_to_generic(func)                                                           \
```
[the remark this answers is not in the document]
GuyAv46: Yes. I think it is better to have a dedicated implementation per callback

tests/pytests/test_profile.py
```diff
@@ -147,6 +147,14 @@ def testProfileAggregate(env):
...
+  actual_res = env.cmd('ft.profile', 'idx', 'aggregate', 'query', '*',
+                'load', 1, 't',
+                'apply', '"banana"', 'as', 'prefix')
+  env.assertEqual(actual_res[1][1][0][5], expected_res)
```
kei-nan, line 156: Optional: 
The [1][1][0][0] is chinese in my POV.
Maybe wrap the retreival in a function so there will be more context as to what you are trying to retreive.
Or don't use magic numbers and use constants instead.

kei-nan, line 156: Optional: 
The [1][1][0][0] is chinese in my POV.
Maybe wrap the retreival in a function so there will be more context as to what you are trying to retreive.
Or don't use magic numbers and use constants instead.

kei-nan (approved):
Added one optional comment, besides that looks good.

tests/pytests/test_profile.py
GuyAv46: replied below

GuyAv46: I agree, but this is the case in this entire test, and I rather keep the "style" consistent.
This case was added purely to improve overage (and make sure we don't leak/crash in the uncovered lines)

src/aggregate/expr/parser.y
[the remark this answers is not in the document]
raz-mon: Yeah I know about the docs' example, the question is if it actually introduces an issue, i.e., solves a bug that we had or we handled it as expected. Worth taking a look.

src/aggregate/expr/parser.y
raz-mon: Two things:
1. The first number that is shifted becomes an `expr` due to the rule `expr(A) ::= number(B)`.
Then, an `OP` arrives and is shifted.
Then, another number is shifted.
If no other rule matches the stack, it will become an `expr`, and then the `expr OP expr` will match.
**Then**, the next number (say `1`) will be shifted with the same logic.
2. How can the `number OP expr` happen? Wouldn't the number would become an `expr` upon the first shift-reduce?

tests/pytests/test_profile.py
raz-mon: Wasn't aware of this - is it documented?

raz-mon: Oh.. OK 🤷‍♂️

src/aggregate/expr/parser.y
raz-mon: OK, makes sense 👍 
Two notes:
1. Consider adding the computation of the `number OP number` in the `expr OP expr` rule code such that some rule overhead and precedence complexity is saved.
2. If we stay with this instead of incorporating (1), let's have some note here explaining why these rules are needed, or point to some test that explains it.

src/aggregate/expr/parser.y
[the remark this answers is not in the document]
GuyAv46: today such a query would've passed:

1. the expression will be interpret as `(@n == @n2) == @n3`
2. the boolean result of the left side will be interpreted as a numeric value of 0 or 1
3. lastly, we test `[0|1] == @n3`

This shouldn't be the expected behavior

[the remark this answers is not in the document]
raz-mon: Yeah, that's what I suspected.. We have the same on our query parser, but I think we can have only binary expressions there. Still worth changing, but it doesn't need to be on this PR.

raz-mon (approved):
💯

kei-nan (approved):

src/aggregate/expr/expression.c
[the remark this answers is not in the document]
meiravgri: it did 😄 
https://redislabs.atlassian.net/browse/MOD-9752

PR #5522, opened 2025-01-18
ACL modifications

raz-mon:
### PR Summary
This PR introduces the following changes related to ACL command categories, configuration parameters, and internal command registration:

#### Removed Features
- `_search_internal` ACL command category.
- New `OSS_ACL_USERNAME` configuration parameter.

#### Deprecated Features
- `OSS_GLOBAL_PASSWORD` configuration parameter.

#### New Functionality
- Internal commands are now registered with the new `internal` command flag **on OSS only**.
- Authentication for internal connections is performed using the new `INTERNALAUTH <secret>` API, of which the secret is retrieved via the new `RedisModule_GetInternalSecret` module API.
- Debug commands' success are aligned with redis's condition for allowing the execution of debug commands.

#### Additional
- A full alignment of internal commands on OSS and proxy-filtered/proxy-excluded commands on Enterprise, for all RediSearch commands.

#### Dependencies
- New internal commands and connections mechanism added to redis - https://github.com/redis/redis/pull/13740.
- New flag addition to the `RM_GetContextFlag()` module API depicting whether or not a context may execute internal commands - https://github.com/redis/redis/pull/13774.
- New module API in RedisModuleSDK submodule - https://github.com/RedisLabsModules/RedisModulesSDK/pull/71.
- New context flag in RedisModuleSDK submodule - https://github.com/RedisLabsModules/RedisModulesSDK/pull/72.


raz-mon:
Requires https://github.com/redis/redis/pull/13740 to be merged on redis, and https://github.com/RedisLabsModules/RedisModulesSDK/pull/71 to RedisModuleSDK (and bump submodule version to it) to pass CI.

pack/ramp-enterprise.yml
```diff
@@ -53,8 +53,6 @@ exclude_commands:
     - _FT.CURSOR
     - _FT.SPELLCHECK
     - _FT.TAGVALS
-    - _FT.DEL
-    - _FT.SAFEADD
```
raz-mon, line 57: Turns out we have a legacy `_FT.DEL` command that we (probably don't) need to still support on Enterprise.

kei-nan (changes requested):

src/coord/debug_commands.c
```diff
@@ -10,15 +10,22 @@
 #include "debug_command_names.h"
 #include "coord/rmr/redis_cluster.h"
 #include "module.h"
+#include "src/config.h"
 #include <assert.h>
 
 DEBUG_COMMAND(shardConnectionStates) {
+  if (!debugCommandsEnabled(ctx)) {
```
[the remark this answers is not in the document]
raz-mon: Right, good question.
See the redis conditions for enabling debug commands.
They have another option other than `yes` and `no` - `local` - which means that you can execute debug commands if you are a local connection.
In order to maintain this ability we need the dynamic check.

src/coord/rmr/conn.c
```diff
@@ -425,11 +426,34 @@ static int MRConn_SendAuth(MRConn *conn) {
   CONN_LOG(conn, "Authenticating...");
 
   // if we failed to send the auth command, start a reconnect loop
-  if (redisAsyncCommand(conn->conn, MRConn_AuthCallback, conn, "AUTH %s %s", clusterConfig.aclUsername, conn->ep.password) ==
-      REDIS_ERR) {
-    MRConn_SwitchState(conn, MRConn_ReAuth);
-    return REDIS_ERR;
+  size_t len = 0;
+
+  if (!IsEnterprise()) {
```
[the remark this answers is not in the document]
raz-mon: Right

src/module.h
```diff
@@ -22,12 +22,13 @@
 extern "C" {
 #endif
 
-#define PROXY_FILTERED "_proxy-filtered"
+#define CMD_PROXY_FILTERED "_proxy-filtered"
```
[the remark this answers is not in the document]
raz-mon: Sure, I'll add something concise. The main documentation is at Redis.

src/module.h
```diff
@@ -59,6 +56,10 @@ do {                                            \
 #define SEARCH_ACL_CATEGORY "search"
 #define SEARCH_ACL_INTERNAL_CATEGORY "_search_internal"
 
+#define NOPERM_ERR "-NOPERM User does not have the required permissions to query the index"
```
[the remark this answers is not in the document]
raz-mon: Should be removed, thanks.

tests/pytests/test_acl.py
```diff
@@ -3,22 +3,12 @@
...
 
 def test_acl_category(env):
     """Test that the `search` category was added appropriately in module
     load"""
     res = env.cmd('ACL', 'CAT')
-    for category in ['search', '_search_internal']:
-        env.assertContains(category, res)
+    env.assertContains('search', res)
 
 @skip(redis_less_than="8.0")
```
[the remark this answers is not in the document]
raz-mon: Sure.
Where do you see this version in Redis?
I'm seeing in the makefile `v7.99.3` currently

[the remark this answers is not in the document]
nafraf: We need to know the tag, to confirm the version where the feature is available.
You can find the number in [redis/src/version.h](https://github.com/redis/redis/blob/d88611d36f55299e83511b7fd41e27ab5811a237/src/version.h#L1) and [00-RELEASENOTES](https://github.com/redis/redis/blob/d88611d36f55299e83511b7fd41e27ab5811a237/00-RELEASENOTES#L5) (numbers of previous versions are also included)

src/coord/rmr/conn.c
```diff
@@ -627,16 +625,14 @@ static void MRConn_ConnectCallback(const redisAsyncContext *c, int status) {
...
-  // If this is an authenticated connection, we need to authenticate
-  if (conn->ep.password) {
+  // Authenticate on OSS always (as an internal connection), or on Enterprise if
+  // a password is set to the `default` ACL user.
+  if (!IsEnterprise() || conn->ep.password) {
     if (MRConn_SendAuth(conn) != REDIS_OK) {
       detachFromConn(conn, 1);
       MRConn_SwitchState(conn, MRConn_Connecting);
     }
-  } else {
```
kei-nan, line 636: Shouldn't we log an error in the else clause?
What state should the connection be at?

tests/pytests/test_acl.py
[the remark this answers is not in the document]
raz-mon: Than I will need the next one, `7.9.227` and on if I'm not mistaking.

src/coord/config.c
```diff
@@ -53,39 +53,20 @@ CONFIG_GETTER(getClusterTimeout) {
...
-// global-password
-int set_immutable_cluster_string_config(const char *name, RedisModuleString *val,
-                                      void *privdata, RedisModuleString **err) {
-  REDISMODULE_NOT_USED(name);
-  REDISMODULE_NOT_USED(err);
-  char **ptr = (char **)privdata;
-  size_t len;
-  const char *ret = RedisModule_StringPtrLen(val, &len);
-  *ptr = rm_strndup(ret, len);
+    "Notice: OSS_GLOBAL_PASSWORD is deprecated, inter-shard communication is now done via internal connections");
```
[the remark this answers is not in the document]
kei-nan: So maybe only issue the log once in the runtime of the process?
If a user/script gets/sets tries setting this it can overflow the logs.

src/coord/rmr/conn.c
raz-mon: Yes, we should thanks

src/coord/config.c
```diff
@@ -53,39 +53,20 @@ CONFIG_GETTER(getClusterTimeout) {
 }
 
 CONFIG_SETTER(setGlobalPass) {
-  SearchClusterConfig *realConfig = getOrCreateRealConfig(config);
-  if (realConfig->globalPass) {
-    rm_free((void *)realConfig->globalPass);
-  }
-  int acrc = AC_GetString(ac, &realConfig->globalPass, NULL, 0);
-  RETURN_STATUS(acrc);
-}
```
MeirShpilraien, line 62: We do not want to keep this configuration when we run on older Redis versions (that does not have internal secret)?

src/coord/rmr/conn.c
```diff
@@ -425,11 +426,31 @@ static int MRConn_SendAuth(MRConn *conn) {
...
-      REDIS_ERR) {
-    MRConn_SwitchState(conn, MRConn_ReAuth);
-    return REDIS_ERR;
+  size_t len = 0;
+
+  if (!IsEnterprise()) {
+    // Take the GIL before calling the internal function getter
+    RedisModule_ThreadSafeContextLock(RSDummyContext);
+    const char *internal_secret = RedisModule_GetInternalSecret(RSDummyContext, &len);
+    RedisModule_ThreadSafeContextUnlock(RSDummyContext);
```
MeirShpilraien, line 435: You should not release the GIL up until you finish using the `internal_secret`, otherwise that main thread might change it while you are using it.

src/coord/rmr/conn.c
```diff
@@ -425,11 +426,31 @@ static int MRConn_SendAuth(MRConn *conn) {
...
+
+    sds secret = sdsnewlen(internal_secret, len);
```
MeirShpilraien, line 437: I do not think you need to create an sds here, see how I did it here: https://github.com/RedisGears/LibMR/pull/56/files#diff-55c2de0fa49d05f6ed8f0c13cacedc85fba5d5739c8360567743f9f740df3179R596

src/module.c
```diff
@@ -181,6 +185,11 @@ int GetSingleDocumentCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int
     return RedisModule_ReplyWithError(ctx, "Unknown Index name");
   }
 
+  if (IsEnterprise() && !ACLUserMayAccessIndex(ctx, sctx->spec)) {
```
MeirShpilraien, line 188: Maybe better to unified this to a function, `checkEnterpriseACL`?

src/module.c
```diff
@@ -2925,17 +3000,17 @@ static int MastersFanoutCommandHandler(RedisModuleCtx *ctx,
 
 static int FanoutCommandHandlerWithIndexAtFirstArg(RedisModuleCtx *ctx,
   RedisModuleString **argv, int argc) {
-  MastersFanoutCommandHandler(ctx, argv, argc, 1);
+  return MastersFanoutCommandHandler(ctx, argv, argc, 1);
```
MeirShpilraien, line 3003: Was this a bug? Do we use the return value?

MeirShpilraien (commented):
Looks good,
Few small comments.

src/coord/config.c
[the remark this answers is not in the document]
kei-nan: The person who monitors the system and logs might not be the person who tried to use this deprecated API.
Maybe log the username who tried to do this modification?

This log feels lacking - either it is needed and, in that case, more useful information needs to be added.
Or if it is not needed then remove it.

src/coord/rmr/conn.c
raz-mon: Right, fixed 👍

src/coord/config.c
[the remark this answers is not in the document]
kei-nan: Maybe you could return this warning in the result to the user if it is resp3?
Then it will reach directly to the right person and not end up in our logs where it could be overlooked.

src/coord/rmr/conn.c
```diff
@@ -425,11 +426,31 @@ static int MRConn_SendAuth(MRConn *conn) {
...
-  if (redisAsyncCommand(conn->conn, MRConn_AuthCallback, conn, "AUTH %s %s", clusterConfig.aclUsername, conn->ep.password) ==
-      REDIS_ERR) {
-    MRConn_SwitchState(conn, MRConn_ReAuth);
-    return REDIS_ERR;
+  size_t len = 0;
+
+  if (!IsEnterprise()) {
+    // Take the GIL before calling the internal function getter
+    RedisModule_ThreadSafeContextLock(RSDummyContext);
+    const char *internal_secret = RedisModule_GetInternalSecret(RSDummyContext, &len);
```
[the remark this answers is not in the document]
raz-mon: As discussed, we will not run against older redis versions that do not support this API.
Nonetheless, we should do this in the module upload instead of crashing here.
Opened a ticket for doing that gracefully, and will remove the assertion.

src/module.c
raz-mon: Yeah sounds good

src/module.c
raz-mon: Just raised a compilation warning, I don't think this is used somewhere.

raz-mon: Checked, this value is actually ignored in redis. Command handler invoked at `cp->func(&ctx,(void**)c->argv,c->argc);`, in `RedisModuleCommandDispatcher()`.

src/coord/config.c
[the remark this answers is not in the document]
raz-mon: If anything, we should have added some warning upon calling any `_FT.CONFIG *` command, not in the single value setter/getter. We don't even do that currently.
We have many more deprecated config-params - a more global approach should be considered, but that's out of the scope of this PR.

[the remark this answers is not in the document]
kei-nan: Maybe we should bring this up to discussion with the rest of the team.
I don't see the benefit of this added log in its current state.

The warning alternative was just a suggestions.
The main point regarding the log still stands from my POV.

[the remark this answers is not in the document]
raz-mon: Which point, the differentiation?

kei-nan (approved):

src/module.c
```diff
@@ -123,6 +123,16 @@ bool ACLUserMayAccessIndex(RedisModuleCtx *ctx, IndexSpec *sp) {
...
 }
 
+// Validates ACL key-space permissions w.r.t the given index spec for Redis
+// Enterprise environments only.
+static inline bool checkEnterpriseACL(RedisModuleCtx *ctx, IndexSpec *sp) {
+  if (!IsEnterprise() || ACLUserMayAccessIndex(ctx, sp)) {
+    // Either OSS, or ACL check passed.
+    return true;
+  }
+  return false;
```
MeirShpilraien, line 133: Not critical but this can be changed to: `return !IsEnterprise() || ACLUserMayAccessIndex(ctx, sp)` no?

src/coord/rmr/conn.c
```diff
@@ -432,20 +432,17 @@ static int MRConn_SendAuth(MRConn *conn) {
...
+    int status = redisAsyncCommand(conn->conn, MRConn_AuthCallback, conn,
+        "AUTH %s %b", INTERNALAUTH_USERNAME, internal_secret, len);
+    if (status == REDIS_ERR) {
       MRConn_SwitchState(conn, MRConn_ReAuth);
-      return REDIS_ERR;
     }
-    sdsfree(secret);
-    return REDIS_OK;
+    RedisModule_ThreadSafeContextUnlock(RSDummyContext);
+    return status;
```
MeirShpilraien, line 442: We sure that the status return from `redisAsyncCommand` are the same as `REDIS_OK`/`REDIS_ERR`?

MeirShpilraien (dismissed):
Tiny comments, but nothing critical.
Great job :+1:

src/module.c
raz-mon: Sure, changing

src/coord/rmr/conn.c
raz-mon: Our whole connection mechanism uses them..
I agree it's not the best practice but let's change it later in a small refactor

src/config.c
```diff
@@ -1224,6 +1224,27 @@ void UpgradeDeprecatedMTConfigs() {
...
+    RedisModule_FreeCallReply(rep);
+    return NULL;
+  }
+  RedisModule_Assert(RedisModule_CallReplyLength(rep) == 2);
+  RedisModuleCallReply *valueRep = RedisModule_CallReplyArrayElement(rep, 1);
+  RedisModule_Assert(RedisModule_CallReplyType(valueRep) == REDISMODULE_REPLY_STRING);
+  size_t len;
+  const char* valueRepCStr = RedisModule_CallReplyStringPtr(valueRep, &len);
+
+  char* res = rm_calloc(1, len + 1);
```
kei-nan, line 1240: optional:
use `rm_strndup` instead

kei-nan (approved):

src/config.c
raz-mon: Agree, yet unrelated to the PR.

oshadmi (approved):

PR #5524, opened 2025-01-19
MOD-6786 Fix search on larger then 128 terms

lerman25:
Fix bug, terms is text longer then 128 (MAX_NORMALIZE_SIZE) cannot be found due to normalization inconsistency.
In indexing, the term was normalized only up to character 128, in search the entire query was normalized.
The limitation of the string normalization was moved to inside NO MODIFY flow, so now all of the term indexed is normalized.
Testes added.

### DIFF
--- a/src/tokenize.c
+++ b/src/tokenize.c
@@ -82,12 +82,12 @@ uint32_t simpleTokenizer_Next(RSTokenizer *base, Token *t) {
     char *tok = toksep(&self->pos, &origLen);
     // normalize the token
     size_t normLen = origLen;
-    if (normLen > MAX_NORMALIZE_SIZE) {
-      normLen = MAX_NORMALIZE_SIZE;
-    }
     char normalized_s[MAX_NORMALIZE_SIZE];
     char *normBuf;
     if (ctx->options & TOKENIZE_NOMODIFY) {
+      if (normLen > MAX_NORMALIZE_SIZE) {
+        normLen = MAX_NORMALIZE_SIZE;
+      }
       normBuf = normalized_s;
     } else {
       normBuf = tok;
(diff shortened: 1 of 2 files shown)


src/tokenize.c
```diff
@@ -82,12 +82,12 @@ uint32_t simpleTokenizer_Next(RSTokenizer *base, Token *t) {
     char *tok = toksep(&self->pos, &origLen);
     // normalize the token
     size_t normLen = origLen;
-    if (normLen > MAX_NORMALIZE_SIZE) {
-      normLen = MAX_NORMALIZE_SIZE;
-    }
     char normalized_s[MAX_NORMALIZE_SIZE];
     char *normBuf;
     if (ctx->options & TOKENIZE_NOMODIFY) {
+      if (normLen > MAX_NORMALIZE_SIZE) {
```
nafraf, line 88: Is this unreachable code?
It seems that we are creating the tokenizer with `TOKENIZE_NOMODIFY` but we are clearing it on `Start`

https://github.com/RediSearch/RediSearch/blob/987128879f49f766718f30665300d4f4bb7a7cd2/src/fragmenter.c#L121-L122

alonre24, line 88: Let's just comment and explain this change and the fact that it is dead

lerman25: Yes this is unreachable, TOKENIZE_NOMODIFY is not set.
Either way,  `FragmentList_FragmentizeBuffer` is only called from testing, so this part is a dead code.
I'm adding a comment to clarify.

alonre24 (approved):

nafraf (dismissed):
OK. Approved. Maybe we'll need another PR to remove the dead code.

nafraf (dismissed):

alonre24 (approved):

PR #5528, opened 2025-01-20
MOD-8561: Fix Inverted Index SeekTo Edge Case

kei-nan:

1. Current: Seeking to the last doc id in the last block when the reader is already pointing at it can crash.
2. Change: Avoid advancing to the next block if the id we need is in the end of the current block.
3. Outcome: Avoid crashing in this edge case.

We revert part of the changes from MOD-8255, PR #5350 

1. Inverted index

#### Mark if applicable

### DIFF
--- a/src/inverted_index.c
+++ b/src/inverted_index.c
@@ -903,6 +903,8 @@ int IR_Read(void *ctx, RSIndexResult **e) {
 
     // if needed - skip to the next block (skipping empty blocks that may appear here due to GC)
     while (BufferReader_AtEnd(&ir->br)) {
+      RS_LOG_ASSERT_FMT(ir->currentBlock < ir->idx->size, "Current block %d is out of bounds %d",
+                        ir->currentBlock, ir->idx->size);
       if (ir->currentBlock + 1 == ir->idx->size) {
         // We're at the end of the last block...
         goto eof;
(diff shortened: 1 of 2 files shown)


src/inverted_index.c
```diff
@@ -1014,7 +1016,9 @@ int IR_SkipTo(void *ctx, t_docId docId, RSIndexResult **hit) {
     goto eof;
   }
 
-  if (IR_CURRENT_BLOCK(ir).lastId < docId || BufferReader_AtEnd(&ir->br)) {
+  const t_docId lastIdOfCurrentBlock = IR_CURRENT_BLOCK(ir).lastId;
+  if (lastIdOfCurrentBlock < docId || (BufferReader_AtEnd(&ir->br) && lastIdOfCurrentBlock != docId)) {
+    // We need to skip to a next block since the requested doc id isn't in the current block
```
raz-mon, line 1021: Why not just remove the `|| BufferReader_AtEnd(&ir->br)` from the condition? Is there something special we should do in that case?

raz-mon (commented):
Nice catch

GuyAv46 (dismissed):

GuyAv46 (approved):

raz-mon (approved):

PR #5554, opened 2025-01-23
Introducing query debug mechanism :tada:

meiravgri:
This PR introduces debugging capabilities to the query execution process by introducing a new command to the debug commands set:

# New _FT.DEBUG command
## Syntax
```
_FT.DEBUG <QUERY> <DEBUG_QUERY_ARGS> DEBUG_PARAMS_COUNT <COUNT>
```

Parameters
* `<QUERY>`:
 Any valid `FT.SEARCH` or `FT.AGGREGATE` command. It supports both SA (standalone) and cluster mode.
* `<DEBUG_QUERY_ARGS>`:
 Currently, the following argument is supported:
    * `TIMEOUT_AFTER_N <N> [INTERNAL_ONLY]`: 
    simulates a timeout after processing `<N>` results. 
Internally, adds a result processor as the downstream result processor of the final execution step (e.g., `RP_INDEX` in SA or `RP_NETWORK` in coordinator).

     * `INTERNAL_ONLY` **(optional):** 
     In cluster mode, it specifies that the timeout only applies to **the internal query** command sent to the shards, without affecting the coordinator pipeline.

* `<DEBUG_PARAMS_COUNT>` should follow `<DEBUG_QUERY_ARGS>`. <COUNT> specifies the expected number of arguments in `<DEBUG_QUERY_ARGS>`.

For example, to simulate a timeout after 100 results, you can use the following debug query:
```
_FT.DEBUG FT.SEARCH  TIMEOUT_AFTER_N 100 DEBUG_PARAMS_COUNT 2
```

# Changes in query execution code

## Functions refactor
To facilitate the injection of debug logic without modifying core execution paths, several functions in the query command flow were broken down into smaller callbacks. This approach minimizes the impact of debug code on production performance.

1.  `execCommandCommon` was split into:
    *   `prepareRequest`
    *   `buildPipelineAndExecute`

2.  Changes in `execCommandCommon` arguments: 
  * The `withProfile` argument (previously constrained to a set of macros and a single value), was replaced with `execOptions`.
  * `execOptions` is a set of flags defined in the newly introduced `ExecOptions` enum, allowing multiple flags to be defined.

3. split `RSExecDistAggregate` into 
  * `prepareForExecution`
  * `executePlan`

## Additions to the production code

As mentioned, the goal was to introduce minimal changes to the production code so that debug mode wouldn't affect performance for regular users. However, some changes were inevitable:

1.  check the newly introduced `QEXEC_F_DEBUG` in `prepareExecutionPlan` (in the production code)
   * A new flag `QEXEC_F_DEBUG` was added to `QEFlags` enum.
   * If enabled, a callback parses debug parameters and adjusts the query execution plan accordingly.

# Debug command implmentation

1. `DEBUG_execCommandCommon` simulates `execCommandCommon`, by creating and initializing `AREQ_Debug` structure instead of `AREQ`.

2. `DEBUG_RSAggregateCommand` and `DEBUG_RSSearchCommand`:
  * Simulate the `FT.AGGREGATE` and `FT.SEARCH` commands, respectively. 
  * Internally call `DEBUG_execCommandCommon` with the respective command type flags (`COMMAND_AGGREGATE` or `COMMAND_SEARCH`).

### Entry point:
defined in debug_command.c.
  * **Standalone (SA) Mode:**
     * `_FT.DEBUG` followed by `FT.AGGREGATE` or `FT.SEARCH` invokes `DEBUG_RSAggregateCommand` or `DEBUG_RSSearchCommand` respectively. 
     * These commands are delegated to `DEBUG_execCommandCommon` with the appropriate command type hint.
  * **Cluster Mode:**
       * The coordinator receives `FT.AGGREGATE` or `FT.SEARCH` and calls `DistAggregateCommand` or `DistSearchCommand`.
       * Internally, the coordinator sends `_FT.DEBUG` followed by `_FT.AGGREGATE` or `_FT.SEARCH` command to all shards


tests/pytests/test_debug_commands.py
```diff
@@ -362,29 +366,285 @@ def testSpecIndexesInfo(env: Env):
...
+
+    def testAggregateDebug_MT(self):
+        self.env.expect(config_cmd(), 'SET', 'WORKERS', 4).ok()
+        self.AggregateDebug()
+        self.env.expect(config_cmd(), 'SET', 'WORKERS', 0).ok()
+
+    # compare results of regular query and debug query
+    def Sanity(self, cmd, query_params):
+        # avoid running this test in cluster mode, as it relies on the order of the results
+        skipTest(cluster=True)
```
alonre24, line 588: I think this sanity is important in cluster mode as well, can we test the results *set* is as expected in this case?

tests/pytests/test_debug_commands.py
```diff
@@ -362,29 +366,285 @@ def testSpecIndexesInfo(env: Env):
...
+        env = self.env
+        results_count = 200
+        timeout_res_count = results_count - 1
```
alonre24, line 591: why minus 1?

tests/pytests/test_debug_commands.py
```diff
@@ -362,29 +366,285 @@ def testSpecIndexesInfo(env: Env):
...
+
+        # expect that the first timeout_res_count of the regular query will be the same as the debug query
+        regular_res = env.cmd(*query)
+        debug_res = env.cmd(debug_cmd(), *query, *debug_params)
+        self.verifyResultsResp3(debug_res, timeout_res_count, f"{cmd} Sanity: compare regular and debug results", should_timeout=True)
+
+        for i in range(timeout_res_count):
+            env.assertEqual(regular_res["results"][i], debug_res["results"][i], message=f"Sanity: compare regular and debug results at index {i}")
+
+    def testSearchSanity(self):
```
alonre24, line 603: Do you think that sanity tests with ft.profile are also needed? Also, do you feel that the test are also covering the internal cursors mechanism?

tests/pytests/test_debug_commands.py
```diff
@@ -362,29 +366,285 @@ def testSpecIndexesInfo(env: Env):
...
+        should_timeout = True
+        check_res_count = True
+        while (cursor):
+            # The counter of timeout simulator rp is initialized to timeout_res_count and decreases with each result returned from the upstream rp.
+            # In cluster mode, RPNet_next (the upstream rp) checks for timeouts at the end/start of reading each shard's reply.
+            # i.e we will always consume the entire chunk of results from each shard before checking for timeout.
+            # If some shards returns fewer than timeout_res_count results (due to EOF), we will continue to read replies until
+            # the sum of their results exceeds timeout_res_count.
+            # if the sum of results from all shards exceeds timeout_res_count but *each* shard individually returns fewer:
+            # no timeout warning is issued, the cursor is depleted.
```
alonre24, line 545: In this case, are we stopping whenever we reach an accumulated number of results >= timeout_res_count (even if we did not collect results from some shards)?

tests/pytests/test_timeout.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # Populate the index
+    num_docs = 150
+    for i in range(num_docs):
+        conn.execute_command('HSET', f'doc{i}' ,'n', i)
+
+    # Before the bug fix, the first doc caused timeout and returned as an empty valid result. Since we reset the timeout counter of RP_INDEX,
+    # The next call to the query pipeline we will continue iterating over the results until EOF is reached or for another TIMEOUT_COUNTER_LIMIT reads.
+    # Now, upon timeout, the reply ends with no further calls to the query pipeline.
+    res = env.cmd('_ft.debug', 'FT.AGGREGATE', 'idx', '*', 'load', '1', '@n', 'LIMIT', 99, 110, 'TIMEOUT_AFTER_N', 99, 'DEBUG_PARAMS_COUNT', 2)
```
alonre24, line 24: Which bug fix is that? What do you mean by "the next call"? there is no cursor here. Not sure I'm following the scenario.

src/aggregate/aggregate_exec.c
```diff
@@ -1228,3 +1256,110 @@ int RSCursorCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
   RedisModule_EndReply(reply);
   return REDISMODULE_OK;
 }
+
+/********************************* DEBUG *********************************/
+/*  Using INTERNAL_ONLY with TIMEOUT_AFTER_N where N == 0 may result in an infinite loop in the coordinator.
+    Since shard replies are always empty, the coordinator might get stuck indefinitely waiting for results or a timeout.
```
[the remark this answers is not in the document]
meiravgri: The coordinator is not woken up when the shard returns an empty result (no docs). Also, reaching the timeout when the policy is RETURN doesn't mean EOF.
The coordinator will exit the loop in RPNetNext only when the timeout is reached (this is why we enforce a clock timeout here).

src/module.h
```diff
@@ -37,6 +37,8 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx);
 int IsMaster();
 bool IsEnterprise();
 
+size_t GetNumShards_UnSafe();
```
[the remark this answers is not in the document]
meiravgri: relied on comments like "// Here we have an unsafe read of `NumShards`. This is fine because its just a hint." and assumed that reading it without a guard is unsafe.
Also, until this PR, this variable was used only in this file so i wanted to emphasize that using this function should be done carefully.

src/result_processor.c
```diff
@@ -1140,3 +1144,93 @@ ResultProcessor *RPCounter_New() {
...
+    clock_gettime(CLOCK_MONOTONIC_RAW, &now);
+    rp_timeout->parent->sctx->time.timeout = now;
+
+    // search upstream for rpidxNext to set timeout limiter
+    ResultProcessor *cur = rp_timeout->upstream;
+    while (cur && cur->type != RP_INDEX) {
+        cur = cur->upstream;
+    }
+
+    if (cur) {
```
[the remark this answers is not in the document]
meiravgri: There is no RP_INDEX in the coordinator pipeline.
I don't think it's complicated to search for it on runtime as in most cases (actully all of them, except coordinator) this is the next RP

tests/pytests/test_debug_commands.py
```diff
@@ -362,29 +366,285 @@ def testSpecIndexesInfo(env: Env):
...
+        expected_results_count = self.env.shardsCount * timeout_res_count
+        # set LIMIT to be larger than the expected results count
+        limit = expected_results_count + 1
+        self.QueryWithLimit(basic_debug_query, timeout_res_count, limit, expected_res_count=expected_results_count, should_timeout=True, message="SearchDebug:")
+
+        # SEARCH always has a sorter
+        self.QueryWithSorter()
+
+        # with no sorter (dialect 4)
+        self.QueryWithLimit(basic_debug_query, timeout_res_count, limit, expected_res_count=expected_results_count, should_timeout=True, message="SearchDebug:")
```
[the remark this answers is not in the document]
meiravgri: It should have been used :sweat_smile:

tests/pytests/test_debug_commands.py
```diff
@@ -362,29 +366,285 @@ def testSpecIndexesInfo(env: Env):
...
+        env = self.env
+        self.setBasicDebugQuery("AGGREGATE")
+        basic_debug_query = self.basic_debug_query
+        self.QueryDebug()
+
+        # EOF will be reached before the timeout counter
+        limit = 2
+        res = self.QueryWithLimit(basic_debug_query, timeout_res_count=10, limit=limit, expected_res_count=limit, should_timeout=False)
+
+        self.QueryWithSorter(sortby_params=['sortby', 1, '@n', 'load', 1, '@n'])
```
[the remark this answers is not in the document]
meiravgri: not sure if this is a bug or a feature :)
but yeah it would give the same results without the load

tests/pytests/test_debug_commands.py
meiravgri: this mechanism doesn't support FT.PROFILE
Didn't understand the second question

tests/pytests/test_timeout.py
meiravgri: next call refers to crossing the pipeline (calling `rp->Next` from `sendChunk`)
in `sendChunk` we request for one result from the query pipeline, serialize it, and trigger the pipeline again
Also important to note here: RP_INDEX checks timeout only every 100 docs.
```
static void sendChunk_Resp3(AREQ *req, RedisModule_Reply *reply, size_t limit,
  cachedVars cv) {
    ....
    startPipeline(req, rp, &results, &r, &rc); -----> triggers pipeline, returns one result
    ....
    if (rp->parent->resultLimit && rc == RS_RESULT_OK) { -----> serialize the result
        nelem += serializeResult(req, reply, &r, &cv);
        SearchResult_Clear(&r);
    .....
        while (--rp->parent->resultLimit && (rc = rp->Next(rp, &r)) == RS_RESULT_OK) { ----> in each iteration trigger pipeline, get one result. stop iterating when EOF or TIMEOUT is reached
        nelem += serializeResult(req, reply, &r, &cv);
        SearchResult_Clear(&r);
        }
```

It simulates MOD-8482 Bug 2: Serializing empty result with resp3, 
LIMIT ensures we skip first 99 docs in the first pipeline cross, i.e check timeout in RP_INDEX (and also reset it)
and TIMEOUT_AFTER_N ensures we have 99 more iterations of the while loop (crossing the pipeline)
The bug was that we ignore the first timeout returned from the first cross (startPipeline) instead of returning immediately .
Instead, we continue to the while loop, where the next timeout check will happen only after 100 results. 
The final result will contain one empty doc, and 100 valid docs

After fixing the bug, the first timeout halts the query, as expected.

tests/pytests/test_debug_commands.py
meiravgri: To get timeout and not EOF

tests/pytests/test_debug_commands.py
meiravgri: Sorry, bad choice of words - i meant order of replies from the shards
Also, we actually "cheat" here in some sens, as a sorter query with timeout potentially returns different results compared to a query that is able to iterate the entire index (no timeout).
(in this case it works because the order of docs insertion was sorted so the docs we get are already sorted)
I prefer not to complicate this test. We have enough coverage for different scenarios in cluster mode.

src/coord/dist_aggregate.c
```diff
@@ -780,3 +801,74 @@ void RSExecDistAggregate(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
...
+  QueryError_ReplyAndClear(ctx, &status);
+  WeakRef_Release(ConcurrentCmdCtx_GetWeakRef(cmdCtx));
+  if (sp) {
+    StrongRef_Release(strong_ref);
+  }
+  SpecialCaseCtx_Free(knnCtx);
+  if (r) AREQ_Free(r);
+  RedisModule_EndReply(reply);
+  return;
+}
```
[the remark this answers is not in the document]
meiravgri: well i added if(r) here...


i'm not sure that passing all these params is nicer, and passing a pointer to a strong ref wouldn't pass in Rust just sayin :D 
but you are right that it's more maintainable

src/module.c
```diff
@@ -3049,13 +3049,27 @@ void RSExecDistAggregate(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
                          struct ConcurrentCmdCtx *cmdCtx);
 int RSAggregateCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc);
 
-static int DistAggregateCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+/** Debug */
```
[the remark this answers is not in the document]
meiravgri: i can't. They are using static function i don't feel like exposing.
I'll move declarations.

tests/pytests/test_debug_commands.py
meiravgri: didn't understand the question

src/module.c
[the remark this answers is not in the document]
meiravgri: moved most of it except the  the query imp that uses static functions

tests/pytests/test_debug_commands.py
meiravgri: Refining and moving this comment to the aggregate debug header as it is relevant to the general behaviour of FT.AGGREGATE and not specifically to cursors.

tests/pytests/test_timeout.py
meiravgri: In the first piepline cross (startPipeline) the pager will skip the first 99 results and we will get timeout because N = 99 = offset. 
In production, we would also get timeout from this cross (assuming that the elapsed time indicates timeout), because we crossed RP_INDEX TIMEOUT_COUNTER_LIMIT (99) times. at this point, the timeout limiter counter of RP_INDEX is reset and we need to cross it 99 more times to get timeout again.
Before the bug fix, despite the fact that the reply from startPipeline contains timeout, we continued to ask more results from the pipeline (the while loop).
In the the second pipeline cross, the pager doesn't skip docs anymore, so we only cross RP_INDEX one time at each while loop. i.e 99 more iteration before checking again timeout.
so we need to set N of the timeout RP to 99, to simulate the scenario of 99 iterations of the while loop.
After the bug fix the first cross (startPipeline) returns timeout and halts the query with 0 results.

alonre24 (dismissed):

src/result_processor.c
```diff
@@ -1134,3 +1138,96 @@ ResultProcessor *RPCounter_New() {
...
+    clock_gettime(CLOCK_MONOTONIC_RAW, &now);
+    rp_timeout->parent->sctx->time.timeout = now;
+
+    // search upstream for rpidxNext to set timeout limiter
+    ResultProcessor *cur = rp_timeout->upstream;
+    while (cur && cur->type != RP_INDEX) {
+        cur = cur->upstream;
+    }
+
+    if (cur) { // This is a coordinator pipeline
```
raz-mon, line 1193:
```suggestion
    if (cur) { // This is a shard pipeline
```
?

src/result_processor.c
```diff
@@ -1134,3 +1138,96 @@ ResultProcessor *RPCounter_New() {
...
+      RPTimeoutAfterCount->parent = &r->qiter;
+      // Insert the timeout processor between the last result processor and its downstream result processor
+      downstream->upstream = RPTimeoutAfterCount;
+      RPTimeoutAfterCount->upstream = cur;
+    }
+    downstream = cur;
+    cur = cur->upstream;
+  }
+  // Update the endProc to the new head in case it was changed
+  r->qiter.endProc = dummyHead.upstream;
```
raz-mon, line 1178: Consider replacing with an explicit condition:
```
if (r->qiter.endProc == r->qiter.rootProc) {..}
```
Just clearer when this happens

src/aggregate/aggregate_debug.c
```diff
@@ -0,0 +1,103 @@
...
+    const char *cmd = RedisModule_StringPtrLen(debug_argv[debug_argv_iter++], &n);
+    if (strncasecmp(cmd, "TIMEOUT_AFTER_N", n) == 0) {
+      unsigned long long results_count;
+      if (RedisModule_StringToULongLong(debug_argv[debug_argv_iter++], &results_count) != REDISMODULE_OK) {
+        QueryError_SetError(status, QUERY_EPARSEARGS, "Invalid TIMEOUT_AFTER_N count");
+        return REDISMODULE_ERR;
+      }
+      // Check for optional argument
+      if (debug_argv_iter != debug_params_count) {
+        // timeout should be applied only for shard commands
```
raz-mon, line 61: What do we actually do here for that to happen?
Seems from the new result processor that it is effective only for the shard pipeline anyways

raz-mon (commented):
Very nice!
See some comments
Not fully done reviewing everything (didn't go over the tests yet)

src/aggregate/aggregate_debug.h
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * **Standalone Mode:**
+ * - The timeout is applied after processing `N` results.
+ * - If the number of available documents matching the query is less than `N`, execution reaches EOF
+ *   instead of simulating a timeout.
+ *
+ * **Cluster Mode:**
+ *
+ * - **`FT.SEARCH`**
+ *   - The coordinator does not check for timeouts, and there is no query pipeline in `FT.SEARCH`.
```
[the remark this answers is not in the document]
meiravgri: I meant partial results timeout.
I'll clarify that

src/aggregate/aggregate_exec.c
```diff
@@ -1228,3 +1256,57 @@ int RSCursorCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
   RedisModule_EndReply(reply);
   return REDISMODULE_OK;
 }
+
+/* ======================= DEBUG ONLY ======================= */
```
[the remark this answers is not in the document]
meiravgri: It uses the status callbacks of the production code, which i didn't think we should expose.

src/aggregate/aggregate_debug.c
meiravgri: When reaching N results - the result processor sets the time to "now".
In FT.AGGREGATE, RPNetNext compares this time to the actual time elapsed, and returns timeout accordingly.
In FT.SEARCH and non strict timeout policy, only the shards determine timeout. 
In both cases the coordinator also checks whether the reply from the shard contains timeout warning

src/aggregate/aggregate_debug.c
```diff
@@ -0,0 +1,103 @@
...
+        cmd = RedisModule_StringPtrLen(debug_argv[debug_argv_iter++], &n);
+        if (isClusterCoord(debug_req) && (strncasecmp(cmd, "INTERNAL_ONLY", n) == 0)) {
```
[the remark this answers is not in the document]
meiravgri: Please explain "unexpected optional argument"

src/aggregate/aggregate_exec.c
```diff
@@ -1228,3 +1256,57 @@ int RSCursorCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+    goto error;
+  }
+
+  if (buildPipelineAndExecute(r, ctx, &status) != REDISMODULE_OK) {
+    goto error;
+  }
+
+  // We don't need debug args anymore
+  debug_params.debug_argv = NULL;
+  debug_params.debug_params_count = 0;
```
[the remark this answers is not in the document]
meiravgri: not really

src/result_processor.c
```diff
@@ -1134,3 +1138,96 @@ ResultProcessor *RPCounter_New() {
...
+ * Will add a result processor that will return timeout according to the results count specified.
+ * @param results_count: number of results to return. should be greater equal 0.
+ * The result processor will also change the query timing so further checks down the pipeline will also result in timeout.
+ */
+void PipelineAddTimeoutAfterCount(AREQ *r, size_t results_count) {
+  ResultProcessor *cur = r->qiter.endProc;
+  ResultProcessor dummyHead = { .upstream = cur };
+  ResultProcessor *downstream = &dummyHead;
+
+  // Search for the last result processor
```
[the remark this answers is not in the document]
meiravgri: Note that this is the same code as in `SetLoadersForBG`

src/result_processor.c
```diff
@@ -1134,3 +1138,96 @@ ResultProcessor *RPCounter_New() {
...
+  }
+  // Update the endProc to the new head in case it was changed
+  r->qiter.endProc = dummyHead.upstream;
+}
+
+static void RPTimeoutAfterCount_SimulateTimeout(ResultProcessor *rp_timeout) {
+    // set timeout to now for the RP up the chain to handle
+    static struct timespec now;
+    clock_gettime(CLOCK_MONOTONIC_RAW, &now);
+    rp_timeout->parent->sctx->time.timeout = now;
```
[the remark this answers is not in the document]
meiravgri: This mechanism exact purpose is to avoid solving timing issues by using "sleep" ;) 
how would it be flaky? we compare the clock with greater equal to the timeout time (refer to `TimedOut` in timeout.h )

src/aggregate/aggregate_exec.c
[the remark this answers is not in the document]
raz-mon: Makes more sense to me, your call

src/aggregate/aggregate_debug.c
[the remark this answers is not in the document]
raz-mon: For instance `TIMEOUT_AFTER_N 2 BANANA`
Not really important here since it's a debug command, but in general it's better to have an explicit check for the arguments we expect to get, and throw an error if we get something we didn't expect instead of still succeeding as if it wasn't given

src/result_processor.c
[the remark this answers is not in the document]
raz-mon: OK greater equal resolves flakiness, thought it may be just greater

src/aggregate/aggregate_debug.c
raz-mon: I was talking about this line (not so clear from the comment):
>// timeout should be applied only for shard commands

I don't understand what we are doing to avoid the coordinator timeout addition

src/aggregate/aggregate_debug.c
```diff
@@ -0,0 +1,103 @@
...
+    if (strncasecmp(cmd, "TIMEOUT_AFTER_N", n) == 0) {
+      unsigned long long results_count;
+      if (RedisModule_StringToULongLong(debug_argv[debug_argv_iter++], &results_count) != REDISMODULE_OK) {
+        QueryError_SetError(status, QUERY_EPARSEARGS, "Invalid TIMEOUT_AFTER_N count");
+        return REDISMODULE_ERR;
+      }
+      // Check for optional argument
+      if (debug_argv_iter != debug_params_count) {
+        // timeout should be applied only for shard commands
+        cmd = RedisModule_StringPtrLen(debug_argv[debug_argv_iter++], &n);
```
raz-mon, line 62: When we will have more options, won't this increment (`debug_argv_iter++`) be problematic, skipping the next debug option param?

src/aggregate/aggregate_debug.c
meiravgri: not adding the timeout result processor

src/aggregate/aggregate_debug.c
meiravgri: correct. will fix

src/result_processor.c
meiravgri: in a coordinator pipeline we don''t expect to find RP_INDEX

raz-mon: Exactly

meiravgri: right

tests/pytests/test_cursors.py
```diff
@@ -367,35 +367,102 @@ def next_cursor_command():
...
+    env.assertEqual(cursor, 0, message=f"expected cursor to be depleted after one FT.CURSOR READ.")
+    env.assertEqual(len(res['results']), 0, message=f"expected to receive 0 results after one FT.CURSOR READ. First query got {n_received} results, read results:{len(res['results'])}")
+
+    env.assertEqual(getCursorStats(env, 'idx')['index_total'], starting_cursor_count)
+
 def testCursorDepletionNonStrictTimeoutPolicy(env):
     """Tests that the cursor id is returned in case the timeout policy is
     non-strict (i.e., the `RETURN` timeout policy), even when a timeout is experienced"""
-
+    env = Env(protocol=3, moduleArgs='ON_TIMEOUT RETURN')
```
raz-mon, line 412: For coverage, since we still have separated response code-paths for RESP2/3, consider running this test in both protocols

tests/pytests/test_debug_commands.py
```diff
@@ -557,3 +561,252 @@ def testDebugScannerStatus(env: Env):
...
+
+        for i in range(timeout_res_count):
+            env.assertEqual(regular_res["results"][i], debug_res["results"][i], message=f"Sanity: compare regular and debug results at index {i}")
+
+    def testSearchSanity(self):
+        self.Sanity("SEARCH", ['SORTBY', 'n'])
+    def testAggSanity(self):
+        self.Sanity("AGGREGATE", ['LOAD', 1, '@n', 'SORTBY', 1, '@n'])
+
+    def testInternalOnly(self):
```
raz-mon, line 770: Consider adding a test with a "bad" optional argument, for coverage of that path as well (I would have that fail)

raz-mon (commented):
Minor comments on the tests, looks GREAT

src/aggregate/aggregate_debug.c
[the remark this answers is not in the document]
meiravgri: fixed

src/aggregate/aggregate_debug.c
```diff
@@ -0,0 +1,103 @@
...
+
+  // Parse the debug params
+  // For example debug_params = TIMEOUT_AFTER_N 2 [INTERNAL_ONLY]
+  size_t debug_argv_iter = 0;
+  while (debug_argv_iter < debug_params_count) {
+    size_t n;
+    const char *cmd = RedisModule_StringPtrLen(debug_argv[debug_argv_iter++], &n);
+    if (strncasecmp(cmd, "TIMEOUT_AFTER_N", n) == 0) {
+      unsigned long long results_count;
+      if (RedisModule_StringToULongLong(debug_argv[debug_argv_iter++], &results_count) != REDISMODULE_OK) {
```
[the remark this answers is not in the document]
meiravgri: done

tests/pytests/test_cursors.py
meiravgri: resp2 doesn't indicate timeout with non strict timeout policy...
I have a dedicated test for resp2 in test_debug_commands

tests/pytests/test_debug_commands.py
meiravgri: check InvalidParams. Does it covering your request?

meiravgri (commented):
// commented in the wrong PR

src/aggregate/aggregate_debug.c
```diff
@@ -0,0 +1,148 @@
...
+      // Getting TIMEOUT_AFTER_N as an array to use AC_IsInitialized API.
+      {.name = "TIMEOUT_AFTER_N",
+       .type = AC_ARGTYPE_SUBARGS_N,
+       .target = &timeoutArgs,
+       .slicelen = 1},
+      // optional arg for TIMEOUT_AFTER_N
+      {.name = "INTERNAL_ONLY", .type = AC_ARGTYPE_BOOLFLAG, .target = &internal_only},
+      // Used for testing purposes.
+      {.name = "DUMMY_DEBUG_OPTION", .type = AC_ARGTYPE_BOOLFLAG, .target = &dummy_debug_option},
+      {NULL}};
```
raz-mon, line 64: Is this intended to stay?

src/aggregate/aggregate_debug.c
```diff
@@ -0,0 +1,148 @@
...
+
+  ACArgSpec *errSpec = NULL;
+  int rv = AC_ParseArgSpec(&ac, debugArgsSpec, &errSpec);
```
raz-mon, line 67: An issue with doing it like this, is that the order of the debug arguments is not enforced. You can also use it for single advancements such that you just get the comfortable API but still enforce the order.
**Not critical** since it's a debug command.

tests/pytests/test_debug_commands.py
raz-mon: Yes!

raz-mon (dismissed):
Nice
Approving since the only comments are on the debug command, which does not have to be perfect. I think it would be better off without the dummy debug argument and with order enforcement on the arguments, but it's not a blocker for the PR AFAIC.

src/aggregate/aggregate_debug.c
meiravgri: I think this approach clearly defines the API for the currently available debug options while also providing flexibility for future extensions—such as supporting additional debug commands beyond `TIMEOUT_AFTER_N.`

I’m okay with `INTERNAL_ONLY` appearing anywhere within the debug parameters, as long as `TIMEOUT_AFTER_N` is present. However, N must always follow TIMEOUT_AFTER_N, which I believe maintains clarity and correctness

src/aggregate/aggregate_debug.c
meiravgri: It serves as an example of how this mechanism should behave when adding new capabilities, useful for both testing and future development. That said, it’s not essential.
I'll remove it

raz-mon (approved):
🔥

PR #5556, opened 2025-01-23
Bug fixes related to Timeout with return partial results policy

meiravgri:
This PR addresses multiple bugs that occur when the timeout policy is return partial results.

## [MOD-8483] Bug 1: Cursor with `SORTBY` is never depleted
When a cursor read with a sorter encounters a timeout during the sorter’s accumulation phase, the sorter switches to yield mode and sets its `timedOut` flag.
After this, it begins emptying its heap, returning the results collected up to the timeout. 
During this process, each returned result has the status `RS_RESULT_OK.`
Once the heap is emptied, the sorter returns `RS_RESULT_TIMEDOUT`, and the current read command is done, with a valid cursor id. 
```
return self->timedOut ? RS_RESULT_TIMEDOUT : RS_RESULT_EOF;
```
At this point, every subsequent `FT.CURSOR READ` call receives `RS_RESULT_TIMEDOUT,` even though no more results are available. Since the cursor is only considered depleted when RS_RESULT_EOF is returned, it remains open indefinitely despite having no results left to provide.

### Fix:
To address this issue, the `timedOut` flag is now reset after the timeout status is returned. This ensures that subsequent `FT.CURSOR READ` commands return `RS_RESULT_EOF,` allowing the cursor to be properly marked as depleted once all results have been processed.

## [MOD-8482] Bug 2: Serializing empty result with resp3
in `sendChunk_Resp3`, During the initial execution of the query pipeline, if a timeout occurred before any results were successfully collected, the return code (rc) was incorrectly changed from `RS_RESULT_TIMEDOUT` to `RS_RESULT_OK` This caused an empty result to be serialized and incorrectly returned as a valid result.

Additionally, in standalone mode (SA), the `RP_INDEX` result processor checks for timeouts every 100 results. If a timeout occurred during the first pipeline run, the timeout counter of `RP_INDEX` was reset to 0, allowing up to 99 valid results to be returned before the next timeout check, thereby violating the timeout limit.

For subsequent pipeline runs, the rc for `RS_RESULT_TIMEDOUT` was handled correctly and not altered.

### Fix:
The fix ensures that the rc is not changed when a timeout occurs. 

## [MOD-8515] Bug 3: `FT.AGGREGATE WITHCURSOR` hangs when shards return 0 results, but cursor is not depleted
When shards return an empty results array (length 0), `MRIteratorCallback_AddReply` is not called to push the result to the channel, leaving the channel waiting indefinitely in `MRChannel_Pop`.

Additionally, since the cursor is not depleted `the MRIteratorCallback_Done` function—which decreases the pending count and calls `MRChannel_Unblock` to wake the waiting `MRChannel_Pop`—is skipped. 
Instead, `MRIteratorCallback_ProcessDone` is called, completing the results collection from the shard but leaving `MRChannel_Pop` in a hanging state.

### Fix:
Move `MRChannel_Unblock` call to `MRIteratorCallback_ProcessDone`, ensuring it is executed even when the cursor is not depleted. This change releases the conditional wait in `MRChannel_Pop`, preventing the hang.
Additionally, when no results are returned from the shards, we avoid returning EOF. Instead, a valid cursor ID is returned gracefully, allowing the next `FT.CURSOR READ` to attempt to collect results.

## [MOD-8606] Bug 4: `RP_Pager` incorrectly updates its counter when no results are received
When used with cursors, this behavior causes the counter to be incorrectly updated for each `FT.CURSOR READ` that times out. As a result, the total number of results returned by all cursor reads is less than the requested `LIMIT`, specifically reduced by the number of `FT.CURSOR READ` commands that encountered a timeout.
for example:

```
FT.AGGREGATE WITHCURSOR count 500000 LIMIT 0 500 TIMEOUT 1
- first read: 399 results returned (timedout)
Expected pager remaining: 101, but it is incorrectly set to 100.
- Next read: The pager counter reaches 0 after 100 results and returns EOF.
Cursor ID is set to 0 since EOF was reached. 
Total results: 499 instead of the requested 500.
```
### Fix:
Ensure that the counter is only updated when the upstream RPs return `RS_RESULT_OK`. This prevents erroneous decrements when no results are received, maintaining the correct count and ensuring the LIMIT is respected.

##  [MOD-8794] Bug 5: freeFlag race condition
There is a race condition on `it->ctx.freeFlag` checked in `MRIterator_Release` to trigger an iterator release process in the following scenario:

**shard thread:** a shard decreases `inProcess` atomic counter. The counter reaches 0.

**coord thread:** now the coordinator (reader) reaches `MR_ManuallyTriggerNextIfNeeded`, since `inProcess == 0`  and it->ctx.pending > 0, is sets `it->ctx.freeFlag = false` and triggers another cursor commands to the shards 

**shard thread:** since `nProcess == 0`, it calls MRIterator_Release, sets `it->ctx.freeFlag = true` and returns immediately since `it->ctx.freeFlag` was false. 

now `it->ctx.freeFlag = true` so next shard to reach MRIterator_Release will release it prematurely.

### Fix:
move `it->ctx.freeFlag = false` from the coordinator (reader) thread to the uv thread, ensuring it happens sequentially after the shards (writers) are all done.

## [MOD-8792] (Not related to timeout) Wrong string length of command arg in `FlatSearchCommandHandler`

MRCommand_Insert expects to receive the length of the string written into n_prefixes.

instead, we return the size of the pointer variable.
```
  char *n_prefixes;
  rm_asprintf(&n_prefixes, "%u", array_len(prefixes));
  MRCommand_Insert(cmd, arg_pos++, n_prefixes, sizeof(n_prefixes) - 1);
  rm_free(n_prefixes);
```
### Fix:

rm_asprintf returns the number of bytes printed (excluding the null terminator)

use this to pass the length of the command string.
```
  char *n_prefixes;
  int cmd_str_len = rm_asprintf(&n_prefixes, "%u", array_len(prefixes));
  MRCommand_Insert(cmd, arg_pos++, n_prefixes, cmd_str_len);
  rm_free(n_prefixes);
```

[MOD-8483]: https://redislabs.atlassian.net/browse/MOD-8483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MOD-8482]: https://redislabs.atlassian.net/browse/MOD-8482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MOD-8515]: https://redislabs.atlassian.net/browse/MOD-8515
[MOD-8606]:https://redislabs.atlassian.net/browse/MOD-8606
[MOD-8794]:https://redislabs.atlassian.net/browse/MOD-8794
[MOD-8792]:https://redislabs.atlassian.net/browse/MOD-8792

### DIFF
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -1326,7 +1326,7 @@ static bool isClusterCoord(AREQ_Debug *debug_req) {
   return false;
 }
 
-int parseAndCompileDebug(AREQ_Debug *debug_req, const DebugQueryOptions *options, QueryError *status) {
+int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
   RedisModuleString **debug_argv = debug_req->debug_params.debug_argv;
   unsigned long long debug_params_count = debug_req->debug_params.debug_params_count;
 
@@ -1357,7 +1357,7 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, const DebugQueryOptions *options
       // Note, this will add a result processor as the downstream of the last result processor
       // (rpidnext for SA, or RPNext for cluster)
       // Take this into account when adding more debug types that are modifying the rp pipeline.
-      PipelineAddTimeoutAfterCount(&debug_req->r, results_count, options->timeout_cb);
+      PipelineAddTimeoutAfterCount(&debug_req->r, results_count);
     }
   }
 
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -581,7 +581,6 @@ static void sendChunk_Resp3(AREQ *req, RedisModule_Reply *reply, size_t limit,
       RedisModule_ReplyKV_LongLong(reply, "total_results", 0);
     } else if (rc == RS_RESULT_TIMEDOUT) {
       // Set rc to OK such that we will respond with the partial results
-      rc = RS_RESULT_OK;
       RedisModule_ReplyKV_LongLong(reply, "total_results", req->qiter.totalResults);
     } else {
       RedisModule_ReplyKV_LongLong(reply, "total_results", req->qiter.totalResults);
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -579,9 +579,6 @@ static void sendChunk_Resp3(AREQ *req, RedisModule_Reply *reply, size_t limit,
     // <total_results>
     if (ShouldReplyWithTimeoutError(rc, req)) {
       RedisModule_ReplyKV_LongLong(reply, "total_results", 0);
-    } else if (rc == RS_RESULT_TIMEDOUT) {
-      // Set rc to OK such that we will respond with the partial results
-      RedisModule_ReplyKV_LongLong(reply, "total_results", req->qiter.totalResults);
     } else {
       RedisModule_ReplyKV_LongLong(reply, "total_results", req->qiter.totalResults);
     }
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -1257,6 +1257,14 @@ int RSCursorCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
   return REDISMODULE_OK;
 }
 
+/********************************* DEBUG *********************************/
+/*  Using INTERNAL_ONLY with TIMEOUT_AFTER_N where N == 0 may result in an infinite loop in the coordinator.
+    Since shard replies are always empty, the coordinator might get stuck indefinitely waiting for results or a timeout.
+    If the query timeout is set to 0 (disabled), neither of these conditions is met.
+    To prevent this, if results_count == 0 and the query timeout is disabled, we enforce a forced timeout,
+    ideally large enough to break the infinite loop without impacting the requested flow */
+#define COORDINATOR_FORCED_TIMEOUT 1000
+
 // FT.DEBUG FT.AGGREGATE idx * <DEBUG_TYPE> <DEBUG_TYPE_ARGS> <DEBUG_TYPE> <DEBUG_TYPE_ARGS> ... DEBUG_PARAMS_COUNT 2
 // Example:
 // FT.AGGREGATE idx * TIMEOUT_AFTER_N 3 DEBUG_PARAMS_COUNT 2
@@ -1338,8 +1346,11 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
       if (debug_argv_iter != debug_params_count) {
         // timeout should be applied only for shard commands
         cmd = RedisModule_StringPtrLen(debug_argv[debug_argv_iter++], &n);
-        if (isClusterCoord(debug_req) && (strncasecmp(cmd, "INTERNAL_ONLY", n) == 0)) {
-          break;
+        if ((GetNumShards_UnSafe() > 1) && (strncasecmp(cmd, "INTERNAL_ONLY", n) == 0) && !(debug_req->r.reqflags & QEXEC_F_INTERNAL)) {
+          if (results_count == 0 && debug_req->r.reqConfig.queryTimeoutMS == 0) {
+            RedisModule_Log(RSDummyContext, "debug", "Forcing coordinator timeout for TIMEOUT_AFTER_N 0 and query timeout 0 to avoid infinite loop");
+            debug_req->r.reqConfig.queryTimeoutMS = COORDINATOR_FORCED_TIMEOUT;
+          }
         }
       }
 
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -1326,7 +1326,7 @@ static bool isClusterCoord(AREQ_Debug *debug_req) {
   return false;
 }
 
-int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
+int parseAndCompileDebug(AREQ_Debug *debug_req, const DebugQueryOptions *options, QueryError *status) {
   RedisModuleString **debug_argv = debug_req->debug_params.debug_argv;
   unsigned long long debug_params_count = debug_req->debug_params.debug_params_count;
 
@@ -1357,7 +1357,7 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
       // Note, this will add a result processor as the downstream of the last result processor
       // (rpidnext for SA, or RPNext for cluster)
       // Take this into account when adding more debug types that are modifying the rp pipeline.
-      PipelineAddTimeoutAfterCount(&debug_req->r, results_count);
+      PipelineAddTimeoutAfterCount(&debug_req->r, results_count, options->timeout_cb);
     }
   }
 
--- a/src/coord/rmr/chan.c
+++ b/src/coord/rmr/chan.c
@@ -88,6 +88,7 @@ void *MRChannel_Pop(MRChannel *chan) {
   pthread_mutex_lock(&chan->lock);
   while (!chan->size) {
     if (!chan->wait) {
+      chan->wait = true;  // reset the flag
       pthread_mutex_unlock(&chan->lock);
       return NULL;
     }
@@ -111,11 +112,5 @@ void MRChannel_Unblock(MRChannel *chan) {
   chan->wait = false;
   // unblock any waiting readers
   pthread_cond_broadcast(&chan->cond);
-  pthread_mutex_unlock(&chan->lock);
-}
-
-void MRChannel_Block(MRChannel *chan) {
-  pthread_mutex_lock(&chan->lock);
-  chan->wait = true;
   pthread_mutex_unlock(&chan->lock);
 }
--- a/src/coord/rmr/chan.c
+++ b/src/coord/rmr/chan.c
@@ -113,3 +113,9 @@ void MRChannel_Unblock(MRChannel *chan) {
   pthread_cond_broadcast(&chan->cond);
   pthread_mutex_unlock(&chan->lock);
 }
+
+void MRChannel_Block(MRChannel *chan) {
+  pthread_mutex_lock(&chan->lock);
+  chan->wait = true;
+  pthread_mutex_unlock(&chan->lock);
+}
--- a/src/coord/rmr/chan.c
+++ b/src/coord/rmr/chan.c
@@ -111,6 +111,6 @@ void MRChannel_Unblock(MRChannel *chan) {
   pthread_mutex_lock(&chan->lock);
   chan->wait = false;
   // unblock any waiting readers
-  pthread_cond_broadcast(&chan->cond);
+  pthread_cond_signal(&chan->cond);
   pthread_mutex_unlock(&chan->lock);
 }
--- a/src/coord/rmr/chan.h
+++ b/src/coord/rmr/chan.h
@@ -16,18 +16,15 @@ MRChannel *MR_NewChannel();
 void MRChannel_Push(MRChannel *chan, void *ptr);
 
 /* Pop an item, or wait until there is an item to pop or until the channel is closed.
- * Return NULL if the channel is closed*/
+ * Return NULL if the channel is empty and MRChannel_Unblock was called by another thread */
 void *MRChannel_Pop(MRChannel *chan);
 
 // Same as MRChannel_Pop, but does not lock the channel nor wait for results if it's empty.
 // This is unsafe, and should only be used when the caller is sure that the channel is not being used by other threads.
 void *MRChannel_UnsafeForcePop(MRChannel *chan);
 
-// Make channel unblocking. All subsequent calls to MRChannel_Pop will return NULL if the channel is empty.
+// Make channel unblocking.
 void MRChannel_Unblock(MRChannel *chan);
-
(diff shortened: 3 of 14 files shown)


src/coord/rmr/rmr.c
```diff
@@ -475,6 +475,8 @@ int MRIteratorCallback_ResendCommand(MRIteratorCallbackCtx *ctx) {
 void MRIteratorCallback_ProcessDone(MRIteratorCallbackCtx *ctx) {
   short inProcess = __atomic_sub_fetch(&ctx->it->ctx.inProcess, 1, __ATOMIC_RELEASE);
   if (!inProcess) {
+    // Unblock the channel before calling `ProcessDone`, as it may trigger the freeing of the iterator
+    MRChannel_Unblock(ctx->it->ctx.chan);
     MRIterator_Release(ctx->it);
```
GuyAv46, line 480: 1. Comment is incorrect
2. As discussed, this is a race condition and we should unblock the channel after attempting to release the iterator (setting the freeing flag). IMO the call to unblock should be part of the release flow, but we can also make the `MRIterator_Release` call return a value to indicate whether we freed the iterator or not, and call the unblock accordingly

raz-mon (commented):
GREAT work!
Partial review (since more changes are on the way)

src/aggregate/aggregate_exec.c
```diff
@@ -579,10 +579,6 @@ static void sendChunk_Resp3(AREQ *req, RedisModule_Reply *reply, size_t limit,
     // <total_results>
     if (ShouldReplyWithTimeoutError(rc, req)) {
       RedisModule_ReplyKV_LongLong(reply, "total_results", 0);
-    } else if (rc == RS_RESULT_TIMEDOUT) {
-      // Set rc to OK such that we will respond with the partial results
-      rc = RS_RESULT_OK;
-      RedisModule_ReplyKV_LongLong(reply, "total_results", req->qiter.totalResults);
     } else {
```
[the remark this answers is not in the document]
meiravgri: RS_RESULT_TIMEDOUT doesn't lead to error reply. This fix aligns the behaviour with resp2.
When RS_RESULT_TIMEDOUT is received upon first pipeline traverse, we return an empty result

src/aggregate/aggregate_exec.c
```diff
@@ -1350,7 +1354,10 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
         // timeout should be applied only for shard commands
         cmd = RedisModule_StringPtrLen(debug_argv[debug_argv_iter++], &n);
         if (isClusterCoord(debug_req) && (strncasecmp(cmd, "INTERNAL_ONLY", n) == 0)) {
-          break;
+          if (results_count == 0 && debug_req->r.reqConfig.queryTimeoutMS == 0) {
+            RedisModule_Log(RSDummyContext, "debug", "Forcing coordinator timeout for TIMEOUT_AFTER_N 0 and query timeout 0 to avoid infinite loop");
+            debug_req->r.reqConfig.queryTimeoutMS = COORDINATOR_FORCED_TIMEOUT;
```
GuyAv46, line 1359: Uncovered. Still needed?

src/coord/rmr/chan.h
```diff
@@ -16,14 +16,14 @@ MRChannel *MR_NewChannel();
...
- * Return NULL if the channel is closed*/
+ * Return NULL if the channel is empty and MRChannel_Unblock was called by another thread */
 void *MRChannel_Pop(MRChannel *chan);
 
 // Same as MRChannel_Pop, but does not lock the channel nor wait for results if it's empty.
 // This is unsafe, and should only be used when the caller is sure that the channel is not being used by other threads.
 void *MRChannel_UnsafeForcePop(MRChannel *chan);
 
-// Make channel unblocking. All subsequent calls to MRChannel_Pop will return NULL if the channel is empty.
+// Make channel unblocking.
```
GuyAv46, line 26:
```suggestion
// Make channel unblocking for a single call to `MRChannel_Pop`.
```

src/coord/rmr/chan.c
```diff
@@ -88,6 +88,7 @@ void *MRChannel_Pop(MRChannel *chan) {
   pthread_mutex_lock(&chan->lock);
   while (!chan->size) {
     if (!chan->wait) {
+      chan->wait = true;  // reset the flag
```
GuyAv46, line 91: Since we now necessarily only unblock a single reader, should we change the broadcast call to a signal in the unblock function?

src/coord/rmr/rmr.c
```diff
@@ -503,11 +510,7 @@ void MRIteratorCallback_Done(MRIteratorCallbackCtx *ctx, int error) {
   // Mark the command of the context as depleted (so we won't send another command to the shard)
   ctx->cmd.depleted = true;
   short pending = --ctx->it->ctx.pending; // Decrease `pending` before decreasing `inProcess`
-  if (pending <= 0) {
-    RS_LOG_ASSERT(pending >= 0, "Pending should not reach a negative value");
-    // Unblock the channel before calling `ProcessDone`, as it may trigger the freeing of the iterator
-    MRChannel_Unblock(ctx->it->ctx.chan);
-  }
+  RS_LOG_ASSERT(pending >= 0, "Pending should not reach a negative value");
```
GuyAv46, line 513: Can we replace this with a simple `assert` call?

src/module.c
```diff
@@ -3220,8 +3220,8 @@ static int prepareCommand(MRCommand *cmd, searchRequestCtx *req, RedisModuleBloc
   MRCommand_Insert(cmd, arg_pos++, "_INDEX_PREFIXES", sizeof("_INDEX_PREFIXES") - 1);
   arrayof(sds) prefixes = sp->rule->prefixes;
   char *n_prefixes;
-  rm_asprintf(&n_prefixes, "%u", array_len(prefixes));
-  MRCommand_Insert(cmd, arg_pos++, n_prefixes, sizeof(n_prefixes) - 1);
+  int string_len = rm_asprintf(&n_prefixes, "%u", array_len(prefixes));
+  MRCommand_Insert(cmd, arg_pos++, n_prefixes, string_len);
```
GuyAv46, line 3224: 🫨

tests/pytests/test_cursors.py
```diff
@@ -365,35 +365,113 @@ def next_cursor_command():
...
-        n_recieved += len(res) - 1
+        n_received += len(res["results"])
+        cursor_runs += 1
 
-    env.assertEqual(n_recieved, num_docs)
+    env.assertEqual(n_received, num_docs, message=f"unexpected results count after {cursor_runs} cursor runs (including the initial query)")
     # Ensure that the cursors we opened were closed properly
     env.assertEqual(getCursorStats(env, 'idx')['index_total'], starting_cursor_count)
 
+## TODO: Fix this test after we fix the bug.
```
GuyAv46, line 444: which bug?

tests/pytests/test_cursors.py
```diff
@@ -365,35 +365,113 @@ def next_cursor_command():
...
     """Tests that the cursor id is returned in case the timeout policy is
     non-strict (i.e., the `RETURN` timeout policy), even when a timeout is experienced"""
-
+    env = Env(protocol=3, moduleArgs='ON_TIMEOUT RETURN')
     conn = getConnectionByEnv(env)
-    run_command_on_all_shards(env, config_cmd(), 'SET', 'ON_TIMEOUT', 'RETURN')
 
     # Create the index
-    env.expect('FT.CREATE idx SCHEMA t text').ok()
+    env.expect('FT.CREATE idx SCHEMA t text sortable').ok()
```
GuyAv46, line 420: why sortable? missing `UNF`?

tests/pytests/test_cursors.py
```diff
@@ -365,35 +365,113 @@ def next_cursor_command():
...
+                          cursor_count], timeout_res_count)
+    VerifyTimeoutWarningResp3(env, res)
+
+    # Verify that the accumulated results (up to timeout_res_count) are returned after timeout
+    env.assertEqual(len(res['results']), timeout_res_count)
+    n_received = len(res['results'])
+
+    # Ensure the cursor is properly depleted after one FT.CURSOR READ
+    iter = 0
+    while (cursor):
```
GuyAv46, line 400: why use a while loop if we expect a single read?

GuyAv46 (commented):
Very nice

src/coord/rmr/rmr.c
meiravgri: what's the diffrence?

src/coord/rmr/chan.c
meiravgri: what are the risks? what is the benefit?

tests/pytests/test_cursors.py
meiravgri: THE bug.
I need to review this test

tests/pytests/test_cursors.py
meiravgri: To simulate the scenario we don't expect and have debugging info if it happens

src/aggregate/aggregate_exec.c
meiravgri: covered

tests/pytests/test_cursors.py
meiravgri: now its ready for review

tests/pytests/test_cursors.py
meiravgri: just left overs of local tests i did.
Not related to UNF(?)

meiravgri:
Runnning on all linux platform:
https://github.com/RediSearch/RediSearch/actions/runs/13267001852/job/37036891672 test_resp3:test_error_with_partial_results 
```
(FAIL):	0 > 0	test_resp3.py:1510 
```
as a result of fixing [MOD-8482](https://redislabs.atlassian.net/browse/MOD-8482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ) Bug 2: Serializing empty result with resp3.
```
(FAIL):	0 > 0	test_resp3.py:1510
```

```  
# `FT.AGGREGATE`
  res = conn.execute_command(
    'FT.AGGREGATE', 'idx', '*', 'TIMEOUT', '1'
  )

  # Assert that we got results
  env.assertGreater(len(res['results']), 0)
```
The test is flaky because the timeout is **not always triggered**- depends on the timing of the query execution.
When the timeout does occur (as expected), the result may return 0 results. Prior to fixing [MOD-8482](https://redislabs.atlassian.net/browse/MOD-8482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ), even if no documents were returned from the shards, **a non-valid result was still serialized**, making it seem like at least one result was returned.

Now, after the bug fix, when the timeout happens, no results are returned, causing the test to fail as it expects the results count to be greater than 0.

I'll fix the test to enforce timeout by calling the debug query mechanism.

src/coord/rmr/chan.c
GuyAv46: future-proofing, if we ever want to have multiple readers. We now unblock a single reader, so why wake everyone up?

src/coord/rmr/rmr.c
GuyAv46: Hopefully, soon, we will have dev-time and prod-time assertions, this one should be a dev-time assertion IMO, so no need for a redis module API assertion

tests/pytests/test_cursors.py
GuyAv46: so why not assert that we get cursor id 0 from the read, and add meaningful info if it fails?

src/coord/rmr/rmr.c
meiravgri: why do you think this is a dev time assertion? 
I suggest replacing it with a simple `assert` + log message so we won't crush (here) on production, WDYT?

tests/pytests/test_cursors.py
```diff
@@ -367,35 +367,108 @@ def next_cursor_command():
...
+    env.expect('FT.CREATE idx SCHEMA n numeric sortable').ok()
+
+    # Populate the index
+    num_docs = 150 * env.shardsCount
+    for i in range(num_docs):
+        conn.execute_command('HSET', f'doc{i}' ,'n', i)
+
+    # This simulates a scenario where shards return empty results due to timeout (so the cursor is still valid),
+    # but the coordinator managed to call 'getNextReply' and waits for replies in MRChannel_Pop, before it checked timeout.
+    # Note: An empty reply does not wake up the coordinator.
```
raz-mon, line 459: Now it does, use `did` for clarity

src/coord/rmr/chan.c
meiravgri: ok

tests/pytests/test_cursors.py
raz-mon: OK it's clear from the rest as well

raz-mon (approved):
Really nice!
Kudos for the super clear docs all around!
Bravo 👏

tests/pytests/test_cursors.py
meiravgri: It still doesn't. the channel will be unblocked because all the shards are done sending the current poll of results.

GuyAv46 (approved):

GuyAv46 (approved):

PR #5566, opened 2025-01-26
Fix Empty Numeric Value - [MOD-7244]

dor-forer:

* Added check for empty numeric string before legacy SEARCH FILTER (only if dialect version is greater than 2).
* Added check for an empty string when evaluating pramas (only if dialect version is greater than 2).

1. An empty string param doesn't evaluate as 0.

#### Mark if applicable

### DIFF
--- a/src/aggregate/aggregate_request.c
+++ b/src/aggregate/aggregate_request.c
@@ -1126,15 +1126,17 @@ int AREQ_ApplyContext(AREQ *req, RedisSearchCtx *sctx, QueryError *status) {
   SetSearchCtx(sctx, req);
   QueryAST *ast = &req->ast;
 
-  int rv = QAST_Parse(ast, sctx, opts, req->query, strlen(req->query), req->reqConfig.dialectVersion, status);
+  unsigned long dialectVersion = req->reqConfig.dialectVersion;
+
+  int rv = QAST_Parse(ast, sctx, opts, req->query, strlen(req->query), dialectVersion, status);
   if (rv != REDISMODULE_OK) {
     return REDISMODULE_ERR;
   }
 
-  if (QAST_EvalParams(ast, opts, status) != REDISMODULE_OK) {
+  if (QAST_EvalParams(ast, opts, status, dialectVersion) != REDISMODULE_OK) {
     return REDISMODULE_ERR;
   }
-  if (applyGlobalFilters(opts, ast, sctx, req->reqConfig.dialectVersion, status) != REDISMODULE_OK) {
+  if (applyGlobalFilters(opts, ast, sctx, dialectVersion, status) != REDISMODULE_OK) {
     return REDISMODULE_ERR;
   }
 
--- a/src/aggregate/aggregate_request.c
+++ b/src/aggregate/aggregate_request.c
@@ -454,18 +454,18 @@ err:
   return REDISMODULE_ERR;
 }
 
-static int parseQueryLegacyArgs(ArgsCursor *ac, RSSearchOptions *options, QueryError *status, bool *isEmptyFilterValue) {
+static int parseQueryLegacyArgs(ArgsCursor *ac, RSSearchOptions *options, bool *hasEmptyFilterValue, QueryError *status) {
   if (AC_AdvanceIfMatch(ac, "FILTER")) {
     // Numeric filter
     NumericFilter **curpp = array_ensure_tail(&options->legacy.filters, NumericFilter *);
-    *curpp = NumericFilter_LegacyParse(ac, status, isEmptyFilterValue);
+    *curpp = NumericFilter_LegacyParse(ac, hasEmptyFilterValue, status);
     if (!*curpp) {
       return ARG_ERROR;
     }
   } else if (AC_AdvanceIfMatch(ac, "GEOFILTER")) {
     GeoFilter *cur_gf = rm_new(*cur_gf);
     array_ensure_append_1(options->legacy.geo_filters, cur_gf);
-    if (GeoFilter_LegacyParse(cur_gf, ac, status) != REDISMODULE_OK) {
+    if (GeoFilter_LegacyParse(cur_gf, ac, hasEmptyFilterValue, status) != REDISMODULE_OK) {
       return ARG_ERROR;
     }
   } else {
@@ -509,7 +509,7 @@ static int parseQueryArgs(ArgsCursor *ac, AREQ *req, RSSearchOptions *searchOpts
 
   req->reqflags |= QEXEC_FORMAT_DEFAULT;
   bool optimization_specified = false;
-  bool isEmptyFilterValue = false;
+  bool hasEmptyFilterValue = false;
   while (!AC_IsAtEnd(ac)) {
     ACArgSpec *errSpec = NULL;
     int rv = AC_ParseArgSpec(ac, querySpecs, &errSpec);
@@ -547,7 +547,7 @@ static int parseQueryArgs(ArgsCursor *ac, AREQ *req, RSSearchOptions *searchOpts
       req->reqflags |= QEXEC_F_SEND_HIGHLIGHT;
 
     } else if ((req->reqflags & QEXEC_F_IS_SEARCH) &&
-               ((rv = parseQueryLegacyArgs(ac, searchOpts, status, &isEmptyFilterValue)) != ARG_UNKNOWN)) {
+               ((rv = parseQueryLegacyArgs(ac, searchOpts, &hasEmptyFilterValue, status)) != ARG_UNKNOWN)) {
       if (rv == ARG_ERROR) {
         return REDISMODULE_ERR;
       }
@@ -570,8 +570,8 @@ static int parseQueryArgs(ArgsCursor *ac, AREQ *req, RSSearchOptions *searchOpts
   }
 
   // In dialect 2, we require a non empty numeric filter
-  if (req->reqConfig.dialectVersion >= 2 && isEmptyFilterValue){
-      QERR_MKBADARGS_FMT(status, "Invalid numeric filter");
+  if (req->reqConfig.dialectVersion >= 2 && hasEmptyFilterValue){
+      QERR_MKBADARGS_FMT(status, "Invalid numeric/geo filter value/s");
       return REDISMODULE_ERR;
     }
 
@@ -1140,7 +1140,7 @@ int AREQ_ApplyContext(AREQ *req, RedisSearchCtx *sctx, QueryError *status) {
     return REDISMODULE_ERR;
   }
 
-  if (QAST_EvalParams(ast, opts, status, dialectVersion) != REDISMODULE_OK) {
+  if (QAST_EvalParams(ast, opts, dialectVersion, status) != REDISMODULE_OK) {
     return REDISMODULE_ERR;
   }
   if (applyGlobalFilters(opts, ast, sctx, dialectVersion, status) != REDISMODULE_OK) {
--- a/src/aggregate/aggregate_request.c
+++ b/src/aggregate/aggregate_request.c
@@ -571,9 +571,9 @@ static int parseQueryArgs(ArgsCursor *ac, AREQ *req, RSSearchOptions *searchOpts
 
   // In dialect 2, we require a non empty numeric filter
   if (req->reqConfig.dialectVersion >= 2 && hasEmptyFilterValue){
-      QERR_MKBADARGS_FMT(status, "Invalid numeric/geo filter value/s");
+      QERR_MKBADARGS_FMT(status, "Numeric/Geo filter value/s cannot be empty");
       return REDISMODULE_ERR;
-    }
+  }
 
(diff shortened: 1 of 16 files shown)


src/aggregate/aggregate_request.c
```diff
@@ -568,6 +569,12 @@ static int parseQueryArgs(ArgsCursor *ac, AREQ *req, RSSearchOptions *searchOpts
     }
   }
 
+  // In dialect 2, we require a non empty numeric filter
+  if (req->reqConfig.dialectVersion >= 2 && isEmptyFilterValue){
+      QERR_MKBADARGS_FMT(status, "Invalid numeric filter");
```
GuyAv46, line 574: might also be a `GEOFILTER`
```suggestion
      QERR_MKBADARGS_FMT(status, "Invalid numeric/geo filter value/s");
```

src/query.c
```diff
@@ -1568,19 +1568,19 @@ int QAST_Expand(QueryAST *q, const char *expander, RSSearchOptions *opts, RedisS
   return REDISMODULE_OK;
 }
 
-int QAST_EvalParams(QueryAST *q, RSSearchOptions *opts, QueryError *status) {
+int QAST_EvalParams(QueryAST *q, RSSearchOptions *opts, QueryError *status,  unsigned int dialectVersion) {
```
GuyAv46, line 1571:
```suggestion
int QAST_EvalParams(QueryAST *q, RSSearchOptions *opts, QueryError *status, unsigned int dialectVersion) {
```

src/query.c
```diff
@@ -1568,19 +1568,19 @@ int QAST_Expand(QueryAST *q, const char *expander, RSSearchOptions *opts, RedisS
...
   if (!q || !q->root || q->numParams == 0)
     return REDISMODULE_OK;
-  QueryNode_EvalParams(opts->params, q->root, status);
+  QueryNode_EvalParams(opts->params, q->root, status, dialectVersion);
   return REDISMODULE_OK;
 }
 
-int QueryNode_EvalParams(dict *params, QueryNode *n, QueryError *status) {
+int QueryNode_EvalParams(dict *params, QueryNode *n, QueryError *status, unsigned int dialectVersion) {
```
GuyAv46, line 1578: consider keeping the `status` parameter last (as a kind of output parameter)

src/query.h
```diff
@@ -124,8 +124,8 @@ IndexIterator *QAST_Iterate(QueryAST *ast, const RSSearchOptions *options,
 int QAST_Expand(QueryAST *q, const char *expander, RSSearchOptions *opts, RedisSearchCtx *sctx,
                 QueryError *status);
 
-int QAST_EvalParams(QueryAST *q, RSSearchOptions *opts, QueryError *status);
-int QueryNode_EvalParams(dict *params, QueryNode *node, QueryError *status);
+int QAST_EvalParams(QueryAST *q, RSSearchOptions *opts, QueryError *status, unsigned int dialectVersion);
+int QueryNode_EvalParams(dict *params, QueryNode *node, QueryError *status, unsigned int dialectVersion);
```
GuyAv46, line 128: consider keeping the `status` parameter last (as a kind of output parameter)

src/query_node.h
```diff
@@ -236,7 +236,7 @@ void QueryNode_ClearChildren(QueryNode *parent, int shouldFree);
  * Returns REDISMODULE_ERR
  * Otherwise, returns REDISMODULE_OK
  */
-int QueryNode_EvalParamsCommon(dict *params, QueryNode *node, QueryError *status);
+int QueryNode_EvalParamsCommon(dict *params, QueryNode *node, QueryError *status, unsigned int dialectVersion);
```
GuyAv46, line 239: consider keeping the `status` parameter last (as a kind of output parameter)

src/query_param.c
```diff
@@ -159,7 +159,15 @@ void QueryParam_InitParams(QueryParam *p, size_t num) {
...
 }
 
-int QueryParam_Resolve(Param *param, dict *params, QueryError *status) {
+int checkNumericValueValid(const char *val, unsigned int dialectVersion) {
+    if (dialectVersion >= 2 && (!*val)) {
+      return 0;
+    }
+
+  return 1;
+}
```
GuyAv46, line 168: This function is the center of this PR
1. Let's keep the indentation in the file coherent (double-space per indentation)
2. Comment on what we are validating, and why it's bound to the dialect of the query
3. Since it's the only validation here, consider having it inlined:
```suggestion
static inline bool checkNumericValueValid(const char *val, unsigned int dialectVersion) {
  return (dialectVersion < 2 || *val);
}
```

tests/pytests/test_empty.py
```diff
@@ -1316,6 +1310,27 @@ def testInvalidUseOfEmptyString():
...
+
+def testEmptyLegacyFilters():
+    """Tests empty values in legacy filters"""
+
+    env = DialectEnv()
+    conn = getConnectionByEnv(env)
+    env.expect('FT.CREATE', 'idx', 'SCHEMA', 'n', 'NUMERIC').ok()
+
+    conn.execute_command('HSET', 'doc1', 'n', 0)
+    res = conn.execute_command('FT.SEARCH', 'idx', '*', 'FILTER', 'n', '', '', 'DIALECT', 1)
```
GuyAv46, line 1325: Add a `GEOFILTER` case?

tests/pytests/test_empty.py
```diff
@@ -1316,6 +1310,27 @@ def testInvalidUseOfEmptyString():
         env.expect('FT.SEARCH', 'idx', '*=>[KNN "" @v $blob AS dist]').error().\
             contains('Syntax error')
 
+        
+        env.expect('FT.SEARCH', 'idx', '*', 'FILTER', 'n', '', '')
```
GuyAv46, line 1314: what is expected here?

tests/pytests/test_empty.py
```diff
@@ -1316,6 +1310,27 @@ def testInvalidUseOfEmptyString():
...
+    env.expect('FT.CREATE', 'idx', 'SCHEMA', 'n', 'NUMERIC').ok()
+
+    conn.execute_command('HSET', 'doc1', 'n', 0)
+    res = conn.execute_command('FT.SEARCH', 'idx', '*', 'FILTER', 'n', '', '', 'DIALECT', 1)
+    env.assertEqual(res, [1, 'doc1', ['n', '0']])
+    
+    MAX_DIALECT = set_max_dialect(env)
+    for dialect in range(2, MAX_DIALECT + 1):
+        env.set_dialect(dialect)
+        env.expect('FT.SEARCH', 'idx', '*', 'FILTER', 'n', '', '').error().contains('Invalid numeric filter')
```
GuyAv46, line 1331: Add a `GEOFILTER` case?

src/numeric_filter.c
```diff
@@ -74,11 +74,18 @@ NumericFilter *NumericFilter_LegacyParse(ArgsCursor *ac, QueryError *status) {
 
   // Parse the min range
   const char *s = AC_GetStringNC(ac, NULL);
+  if (!s || !*s) {
```
GuyAv46, line 77: `!s` is always false here, as we previously checked that `AC_NumRemaining(ac) >= 3`, so we know we have strings here.
Also, note that `parseDoubleRange` cannot handle `NULL`s so you can be sure you're not exposing the code to any new risks here

src/numeric_filter.c
```diff
@@ -74,11 +74,18 @@ NumericFilter *NumericFilter_LegacyParse(ArgsCursor *ac, QueryError *status) {
...
+    *hasEmptyFilterValue = true;
+  }
   if (parseDoubleRange(s, &nf->inclusiveMin, &nf->min, 1, 1, status) != REDISMODULE_OK) {
     NumericFilter_Free(nf);
     return NULL;
   }
+
   s = AC_GetStringNC(ac, NULL);
+  if (!s || !*s) {
```
GuyAv46, line 86: same as above

src/query_param.c
```diff
@@ -159,7 +159,11 @@ void QueryParam_InitParams(QueryParam *p, size_t num) {
   memset(p->params, 0, sizeof(*p->params) * num);
 }
 
-int QueryParam_Resolve(Param *param, dict *params, QueryError *status) {
+static inline bool checkNumericValueValid(const char *val, unsigned int dialectVersion) {
+  return (dialectVersion < 2 || *val);
```
GuyAv46, line 163: Add a comment on what and why we verify this

src/query_param.c
```diff
@@ -159,7 +159,15 @@ void QueryParam_InitParams(QueryParam *p, size_t num) {
   memset(p->params, 0, sizeof(*p->params) * num);
 }
 
-int QueryParam_Resolve(Param *param, dict *params, QueryError *status) {
+/**
+ * Checks if the given numeric value is not empty.
+ * If the dialect version is 2 or higher, the value must not be empty, otherwise it can be 0 for backward compatibility.
+ */
+static inline bool checkNumericValueValid(const char *val, unsigned int dialectVersion) {
```
raz-mon, line 166: Since this validates geo as well, consider expressing that in the function name.
Regardless, add it to the comment.

src/geo_index.c
```diff
@@ -32,16 +41,20 @@ int GeoFilter_LegacyParse(GeoFilter *gf, ArgsCursor *ac, QueryError *status) {
     QERR_MKBADARGS_AC(status, "<geo property>", rv);
     return REDISMODULE_ERR;
   }
+
+  CheckAndSetEmptyFilterValue(ac, hasEmptyFilterValue);
```
raz-mon, line 45: Consider putting these checks after the parsing (i.e., 4 lines below), and conditioning it on the value of the argument. I.e., if it is not `0`, we don't need this extra parsing.

src/module.c
```diff
@@ -1638,7 +1638,7 @@ void setKNNSpecialCase(searchRequestCtx *req, specialCaseCtx *knn_ctx) {
 // Prepare a TOPK special case, return a context with the required KNN fields if query is
 // valid and contains KNN section, NULL otherwise (and set proper error in *status* if error
 // was found).
-specialCaseCtx *prepareOptionalTopKCase(const char *query_string, RedisModuleString **argv, int argc,
+specialCaseCtx *prepareOptionalTopKCase(const char *query_string, RedisModuleString **argv, int argc, unsigned int dialectVersion,
```
raz-mon, line 1641:
```suggestion
specialCaseCtx *prepareOptionalTopKCase(const char *query_string, RedisModuleString **argv, int argc, uint dialectVersion,
```

raz-mon (commented):
Nicely done.
See a few comments

src/geo_index.c
dor-forer: The issue is that AC_GetDouble already advances the curser so the check is not relevant anymore.
Do you think I should pass the AC_F_NOADVANCE to AC_GetDouble, and advance the curser afterward?

src/aggregate/aggregate_request.c
```diff
@@ -568,6 +569,12 @@ static int parseQueryArgs(ArgsCursor *ac, AREQ *req, RSSearchOptions *searchOpts
     }
   }
 
+  // In dialect 2, we require a non empty numeric filter
+  if (req->reqConfig.dialectVersion >= 2 && hasEmptyFilterValue){
+      QERR_MKBADARGS_FMT(status, "Invalid numeric/geo filter value/s");
+      return REDISMODULE_ERR;
+    }
```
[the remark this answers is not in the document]
dor-forer: Where can it be?
Only retrieve the dialect version on: 
`int rv = handleCommonArgs(req, ac, status, 1);`
So it has to be after the while loop.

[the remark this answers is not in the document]
raz-mon: Can't it be deeper in the parsing paths, i.e., the specific parsing paths for numeric/geo, such that we can have a separate message for numeric and geo?

src/geo_index.c
raz-mon: Yeah that sounds good, since most of the times this won't be the case and we'll just advance the cursor by 1.

src/geo_index.c
```diff
@@ -32,20 +41,36 @@ int GeoFilter_LegacyParse(GeoFilter *gf, ArgsCursor *ac, QueryError *status) {
...
-  if ((rv = AC_GetDouble(ac, &gf->lon, 0) != AC_OK)) {
+
+  if ((rv = AC_GetDouble(ac, &gf->lon, AC_F_NOADVANCE) != AC_OK)) {
     QERR_MKBADARGS_AC(status, "<lon>", rv);
     return REDISMODULE_ERR;
   }
+  if (gf->lon == 0)
+  {
+    CheckAndSetEmptyFilterValue(ac, hasEmptyFilterValue);
+  }
```
raz-mon, line 52:
```suggestion
  if (gf->lon == 0) {
    CheckAndSetEmptyFilterValue(ac, hasEmptyFilterValue);
  }
```

raz-mon (commented):
LGTM 👍 
Small style comment

raz-mon (approved):

Issue #5532, opened 2025-01-21
[BUG] Redis 7.4.2 crashed by signal: 11, si_code: 128 - intermittently when running FT.SEARCH


erihu78:
**Describe the bug**
Today our production Redis-instances started crashing intermittently with the following error message: Redis 7.4.2 crashed by signal: 11, si_code: 128
We have found nothing that was changed or added at the point the crashes started.

Seems to be both Primary and Replica nodes that crash.
Every bug report seems to indicate the error occurred when running FT.SEARCH
But not all FT.SEARCH-statements of the same pattern cause crashes.

Please if anyone could assist in troubleshooting we would be very grateful

**To Reproduce**
Unknown at this point 

**Environment (please complete the following information):**
 - OS:ubuntu 22.04
 - CPU model: Intel
 - Redis Server version: 7.4.1
 - RediSearch version: 2.10.10 (apt package: redis-stack-server:7.4.0-v2)

```
=== REDIS BUG REPORT START: Cut & paste starting from here ===
93700:M 21 Jan 2025 13:52:35.865 # Redis 7.4.2 crashed by signal: 11, si_code: 1
93700:M 21 Jan 2025 13:52:35.865 # Accessing address: 0x1027ef9
93700:M 21 Jan 2025 13:52:35.865 # Crashed running the instruction at: 0x7ff82130af27

------ STACK TRACE ------
EIP:
/opt/redis-stack/lib/redisearch.so(+0x1d1f27)[0x7ff82130af27]

93700 redis-server *
/lib/x86_64-linux-gnu/libc.so.6(+0x42520)[0x7ff823a97520]
/opt/redis-stack/lib/redisearch.so(+0x1d1f27)[0x7ff82130af27]
/opt/redis-stack/lib/redisearch.so(IR_Read+0x4c)[0x7ff82130b63c]
/opt/redis-stack/lib/redisearch.so(IR_SkipTo+0x1d3)[0x7ff82130b8d3]
/opt/redis-stack/lib/redisearch.so(+0x1cb6a2)[0x7ff8213046a2]
/opt/redis-stack/lib/redisearch.so(+0x1c9c56)[0x7ff821302c56]
/opt/redis-stack/lib/redisearch.so(+0x1ca37f)[0x7ff82130337f]
/opt/redis-stack/lib/redisearch.so(+0x1cab77)[0x7ff821303b77]
/opt/redis-stack/lib/redisearch.so(+0x1c9c56)[0x7ff821302c56]
[...]

Issue #5517, opened 2025-01-16
[CRASH] RediSearch Crash after working for several hours, with Redis 7.2.0 crashed by signal: 11, si_code: 128 and Accessing address: (nil)


mjblackend:
**Describe the bug**
Am tried to use the redisearch module it works normally on my functional test but crash in load test, tested on multiple machines RED HAT and Linux

the whole node and its replica is down and then the cluster is down

=== REDIS BUG REPORT START: Cut & paste starting from here ===
1491639:M 15 Jan 2025 13:20:24.219 # Redis 7.2.0 crashed by signal: 11, si_code: 128
1491639:M 15 Jan 2025 13:20:24.219 # Accessing address: (nil)
1491639:M 15 Jan 2025 13:20:24.219 # Crashed running the instruction at: 0x7ff8453517a8

here is the crash report:
```

1491639:M 15 Jan 2025 13:20:24.083 . --- Processing packet of type ping, 2616 bytes
1491639:M 15 Jan 2025 13:20:24.083 . ping packet received: 9a16bdd59195961a231c528735b5202750da84f0
1491639:M 15 Jan 2025 13:20:24.083 . GOSSIP 031053b9e2e5e730788080a732624d26c3ce97d5 10.0.0.130:6379@16379 master
1491639:M 15 Jan 2025 13:20:24.083 . GOSSIP f8166b0f549bb584dfacdb0383a8fcf2873e04be 10.0.0.106:6379@16379 master
1491639:M 15 Jan 2025 13:20:24.083 . GOSSIP 4740f438b06a4734437782f940d07b3bfb877472 10.0.0.102:6379@16379 master
1491639:M 15 Jan 2025 13:20:24.087 . --- Processing packet of type ping, 2616 bytes
1491639:M 15 Jan 2025 13:20:24.087 . ping packet received: c44d5386b2c0eaa0f7c0d2dba4b04fd26ef5fe9b
1491639:M 15 Jan 2025 13:20:24.087 . GOSSIP f8166b0f549bb584dfacdb0383a8fcf2873e04be 10.0.0.106:6379@16379 master
1491639:M 15 Jan 2025 13:20:24.087 . GOSSIP 4740f438b06a4734437782f940d07b3bfb877472 10.0.0.102:6379@16379 master
1491639:M 15 Jan 2025 13:20:24.087 . GOSSIP 8e9d44176cb5157c49e7aa10ba89b0524f55315e 10.0.0.130:6380@16380 slave
[...]

PR #5591, opened 2025-02-02
Fix SkipTo edge case - [MOD-8561 MOD-8673]

GuyAv46:

Fix a bug in the GC, which should update the last ID of an inverted index when cleaning it, in case it was modified.

This last ID in the index is used in the inverted-index reader iterator to test if we can expect more results from this iterator quickly.

An edge case where the last id(s) were deleted and this value was not up-to-date could cause the index reader iterator to assume it has a block to skip to, while that's not the case, causing unsafe access to the end of the block array.

1. MOD-8673
2. #5517 
3. #5532

1. Add assertions to verify block index ranges
2. Fix the GC apply function

#### Mark if applicable

### DIFF
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -737,6 +737,7 @@ static void FGC_applyInvertedIndex(ForkGC *gc, InvIdxBuffers *idxData, MSG_Index
 
   idx->numDocs -= info->ndocsCollected;
   idx->gcMarker++;
+  idx->lastId = idx->blocks[idx->size - 1].lastId; // Update lastId
 }
 
 typedef struct {
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -737,6 +737,7 @@ static void FGC_applyInvertedIndex(ForkGC *gc, InvIdxBuffers *idxData, MSG_Index
 
   idx->numDocs -= info->ndocsCollected;
   idx->gcMarker++;
+  RS_LOG_ASSERT(idx->size, "Index should have at least one block");
   idx->lastId = idx->blocks[idx->size - 1].lastId; // Update lastId
 }
 
--- a/tests/pytests/test_issues.py
+++ b/tests/pytests/test_issues.py
@@ -1339,51 +1339,24 @@ def test_mod_6786(env:Env):
 @skip(cluster=True)
 def test_mod_8561(env:Env):
   env.expect(config_cmd(), 'SET', 'FORK_GC_CLEAN_THRESHOLD', '0').ok()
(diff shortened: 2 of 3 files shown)


src/fork_gc.c
```diff
@@ -737,6 +737,7 @@ static void FGC_applyInvertedIndex(ForkGC *gc, InvIdxBuffers *idxData, MSG_Index
 
   idx->numDocs -= info->ndocsCollected;
   idx->gcMarker++;
+  idx->lastId = idx->blocks[idx->size - 1].lastId; // Update lastId
```
kei-nan, line 740: isn't it possible idx->size is zero?

tests/pytests/test_issues.py
```diff
@@ -1335,3 +1335,28 @@ def test_mod_6786(env:Env):
   # Searching for the long term should return the document
   # Before fix, the long term was partialy normalized and the document was not found
   env.expect('FT.SEARCH', 'idx', long_term).equal([1, 'doc1', ['t', text_with_long_term]])
+
+@skip(cluster=True)
+def test_mod_xxxx(env:Env):
+  env.expect(config_cmd(), 'SET', 'FORK_GC_CLEAN_THRESHOLD', '0').ok()
+  env.expect('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT').ok()
+  n_docs_per_block = 100
+  n_blocks = 2.5
```
kei-nan, line 1344: what is the significance of the 2.5 blocks for the reproduction of the bug?

GuyAv46: It's not needed, but initially, I thought we must use the `IndexReader_SkipToBlock` binary search, and we have a quick look if the next block is the one we looking for so I needed at least 3 blocks, chose not to fill the last one completely to have some room for the last few docs that also has the other term

src/fork_gc.c
```diff
@@ -737,6 +737,8 @@ static void FGC_applyInvertedIndex(ForkGC *gc, InvIdxBuffers *idxData, MSG_Index
 
   idx->numDocs -= info->ndocsCollected;
   idx->gcMarker++;
+  RS_LOG_ASSERT(idx->size, "Index should have at least one block");
+  idx->lastId = idx->blocks[idx->size - 1].lastId; // Update lastId
```
raz-mon, line 741: 💌

src/fork_gc.c
raz-mon: I think let's have an assertion, max

raz-mon (approved):
AMAZING

raz-mon (approved):

kei-nan (approved):

PR #5597, opened 2025-02-04
[2.8] Fix inverted index last id update on GC

GuyAv46:
# Description
Backport of #5591 to `2.8`.

### DIFF
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -718,6 +718,8 @@ static void FGC_applyInvertedIndex(ForkGC *gc, InvIdxBuffers *idxData, MSG_Index
 
   idx->numDocs -= info->ndocsCollected;
   idx->gcMarker++;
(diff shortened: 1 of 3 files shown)


DvirDukhan (approved):

src/fork_gc.c
```diff
@@ -718,6 +718,8 @@ static void FGC_applyInvertedIndex(ForkGC *gc, InvIdxBuffers *idxData, MSG_Index
 
   idx->numDocs -= info->ndocsCollected;
   idx->gcMarker++;
+  RS_LOG_ASSERT(idx->size, "Index should have at least one block");
+  idx->lastId = idx->blocks[idx->size - 1].lastId; // Update lastId
```
DvirDukhan, line 722: I approve this but I think this should be a `getter` so we will have less static information we need to update

PR #5602, opened 2025-02-05
Update FT.AGGREGATE commands to match most recent in RedisInsight

pd-redis:
## Update commands.json to match most recent in RedisInsight

1. Current: commands.json in this repo differs from the latest in RedisInsight
2. Change: updated FT.AGGREGATE command in commands.json to match most recent in RedisInsight
3. Outcome: commands.json will be up to date

1. commands.json

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -1221,9 +1221,75 @@
             "multiple": true,
             "arguments": [
               {
+                "name": "reduce",
+                "token": "REDUCE",
+                "type": "pure-token"
+              },
+              {
                 "name": "function",
-                "type": "string",
-                "token": "REDUCE"
+                "type": "oneof",
+                "arguments": [
+                  {
+                    "name": "count",
+                    "type": "pure-token",
+                    "token": "COUNT"
+                  },
+                  {
+                    "name": "count_distinct",
+                    "type": "pure-token",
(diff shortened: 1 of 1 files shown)


commands.json
```diff
@@ -1332,6 +1741,7 @@
         "name": "filter",
         "type": "string",
         "optional": true,
+        "expression": true,
```
oshadmi, line 1744: Can also put all `APPLY` arguments here as well
[Docs](https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/aggregations/#filter-expressions) also say 
> Filter expressions follow the syntax of APPLY expressions, with the addition of the conditions


e.g.,
```FILTER "exists(@n) && ceil(@n)>5"```

oshadmi (approved):

commands.json
pd-redis: Do you mean simply copy `arguments` from `APPLY` to `FILTER`? It would still miss `the conditions ==, !=, <, <=, >, >=`!

oshadmi: Right, it is a complex expressions, something like
((!)?`argument`|`literal`) (`argument` `op` `argument`|`literal`))*
So let's not over complicate it now.

PR #5609, opened 2025-02-06
Run CI against redis unstable

raz-mon:
Modifies our pull-request and merge-queue CI flows to run against the redis `unstable` branch instead of the latest `8.0` tag.

### DIFF
--- a/.github/workflows/event-merge-to-queue.yml
+++ b/.github/workflows/event-merge-to-queue.yml
@@ -16,30 +16,24 @@ concurrency:
 # TODO: Use RedisJSON's `${{ vars.DEFAULT_REDISJSON_REF }}` branch when testing on nightly
 
 jobs:
-  get-latest-redis-tag:
-    uses: ./.github/workflows/task-get-latest-tag.yml
-    with:
-      repo: redis/redis
-      prefix: '8.0'
-
   docs-only: # Check whether the PR is only modifying docs
     uses: ./.github/workflows/task-check-docs.yml
 
   test-linux:
-    needs: [docs-only, get-latest-redis-tag]
+    needs: [docs-only]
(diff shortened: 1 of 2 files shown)


.github/workflows/event-merge-to-queue.yml
```diff
@@ -16,30 +16,24 @@ concurrency:
 # TODO: Use RedisJSON's `${{ vars.DEFAULT_REDISJSON_REF }}` branch when testing on nightly
 
 jobs:
-  get-latest-redis-tag:
-    uses: ./.github/workflows/task-get-latest-tag.yml
-    with:
-      repo: redis/redis
-      prefix: '8.0'
-
```
DvirDukhan, line 24: don't delete it, why not just change line 23  unstable?

.github/workflows/event-pull_request.yml
```diff
@@ -11,23 +11,17 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  get-latest-redis-tag:
```
DvirDukhan, line 14: same comment

.github/workflows/event-merge-to-queue.yml
raz-mon: There is no tag with the 'unstable' prefix

GuyAv46: When `8.0` is out we will want to remove the prefix input here, so reverting will be only the first step. Can we open a ticket with instructions + the PR number that needs to be reverted?

GuyAv46 (approved):

.github/workflows/event-merge-to-queue.yml
raz-mon: Yes

oshadmi (approved):

PR #5622, opened 2025-02-09
Simplify testing with JSON flow - purge readies [MOD-8156]

alonre24:

Move RedisJSON module to be a part of our test dependencies, and build the module from source always when running our Python flow tests.
* Set the branch in JSON to run against as part of our `make pytest`, and allow this parameter to be passed in CI flows (so we can change it according to the branch). 
* Build the RedisJSON module for benchmarks when running in CI (flow is still broken due to other stuff)
* Use a workaround for Alpine (as we had also before) 

#### Mark if applicable


.codespell/ignore_wordlist.txt
```diff
@@ -8,3 +8,4 @@ nd
 curEnt
 childRes
 specField
+runn
```
GuyAv46, line 11: is this needed? Isn't that a `readies` command? I don't see where you use it

.github/workflows/event-pull_request.yml
```diff
@@ -25,12 +25,15 @@ jobs:
...
       get-redis: unstable
+      rejson-branch: master
     secrets: inherit
 
   coverage:
     needs: docs-only
     if: ${{ vars.ENABLE_CODE_COVERAGE == 'true' && needs.docs-only.outputs.only-docs-changed == 'false' && !github.event.pull_request.draft }}
     uses: ./.github/workflows/flow-coverage.yml
+    with:
+      rejson-branch: master
```
GuyAv46, line 36: Missing sanitizer setup?

.github/workflows/flow-linux-platforms.yml
```diff
@@ -23,6 +23,14 @@ on:
         default: true
       redis-ref:
         type: string
+      rejson:
+        type: number
```
GuyAv46, line 27: why not make it boolean here as well?

.github/workflows/flow-linux-platforms.yml
```diff
@@ -92,6 +108,8 @@ jobs:
     secrets: inherit
     with:
       get-redis: ${{ inputs.redis-ref }}
+      rejson: ${{ (inputs.rejson == true || inputs.rejson == 1) && 1 || 0 }}
```
GuyAv46, line 111: Is this tested? (for resulting in `1` or `0` for any `inputs.rejson` value)

.install/alpine_linux_3.sh
```diff
@@ -6,6 +6,8 @@ $MODE apk update
 
 $MODE apk add --no-cache build-base gcc g++ make linux-headers openblas-dev \
     xsimd curl wget git python3 python3-dev py3-pip openssl openssl-dev \
-    tar xz which rsync bsd-compat-headers clang clang17-libclang curl
+    tar xz which rsync bsd-compat-headers curl \
+    bash cargo gcc autoconf automake py3-cryptography \
+    musl-dev libffi-dev openssh py-virtualenv clang18-libclang
```
GuyAv46, line 11: For JSON build? Does it affect CI duration?

tests/deps/setup_rejson.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+# Exit immediately with statys 1 (error) if a command exits with a non-zero status
+set -e
+
+# Set the default variables
+ROOT=${ROOT:=`pwd`}
```
GuyAv46, line 7: There are probably better ways to get the root directory of the project. I don't think this will work. Consider fetching the file location and going two steps back

tests/deps/setup_rejson.sh
```diff
@@ -0,0 +1,51 @@
...
+# See https://github.com/rust-lang/rust/pull/58575#issuecomment-496026747
+if [[ -f /etc/os-release ]]; then
+	OS_NAME=$(grep '^NAME=' /etc/os-release | sed 's/"//g')
+	OS_NAME=${OS_NAME#"NAME="}
+	if [[ $OS_NAME == "Alpine Linux" ]]; then
+		sed -i "s/^RUST_FLAGS=$/RUST_FLAGS=-C target-feature=-crt-static/g" Makefile
+	fi
+fi
+echo "Building RedisJSON module for branch $JSON_BRANCH..."
+BINROOT=${JSON_BIN_DIR} make SAN=$SAN > /dev/null  # print errors to console
```
GuyAv46, line 49: so no early exit if it is already built? Do we expect the build flow to return quickly if there is nothing to build?

tests/pytests/runtests.sh
```diff
@@ -43,10 +43,9 @@ help() {
 		UNSTABLE=1            Do not skip unstable tests (default: 0)
 		ONLY_STABLE=1         Skip unstable tests
 
-		REJSON=0|1|get|view     Also load RedisJSON module (get: force download from S3; view: from local view)
-		REJSON_BRANCH=branch    Use a snapshot of given branch name
-		REJSON_PATH=path|view   RedisJSON module path
-		REJSON_MODARGS=args     RedisJSON module arguments
+    REJSON=1|0            Also load RedisJSON module (default: 1)
+    REJSON_BRANCH=branch  Use RedisJSON module from branch (default: 'master')
+    REJSON_PATH=path      Use RedisJSON module at `path` (default: '' - build from source)
```
GuyAv46, line 48: Indentation breaks here. Consider replacing all tabs with spaces, or aligning with existing lines

tests/pytests/runtests.sh
```diff
@@ -468,6 +416,7 @@ run_tests() {
 
 	local E=0
 	if [[ $NOP != 1 ]]; then
+	  cd $HERE
```
GuyAv46, line 419: Why is this needed?

tests/deps/setup_rejson.sh
```diff
@@ -0,0 +1,51 @@
...
+if [ ! -d "${JSON_MODULE_DIR}" ]; then
+    echo "Cloning RedisJSON repository from ${JSON_REPO_URL} to ${JSON_MODULE_DIR}..."
+    git clone --quiet --recursive $JSON_REPO_URL $JSON_MODULE_DIR
+    echo "Done"
+else
+    echo "RedisJSON already exists in ${JSON_MODULE_DIR}"
+fi
+
+# Navigate to the module directory and checkout the specified branch and its submodules
+cd ${JSON_MODULE_DIR}
```
GuyAv46, line 34: Should also change the directory back when done (or is it happening by default?)

.codespell/ignore_wordlist.txt
alonre24: It is, still used in `runtests.sh`. Should be removed in next PRs as part of removing readies

.github/workflows/flow-linux-platforms.yml
alonre24: yes. I did it because I got an error before :smiling_face_with_tear:

.install/alpine_linux_3.sh
alonre24: I tried adding the installation of JSON module when the build failed, but these changes did not effect the JSON build for Alpine, so it can be reverted

tests/deps/setup_rejson.sh
alonre24: If it is already built the `make` command will be completed immediately. I didn't want to skip in case an so file exists since it might be not updated (when developing locally)

tests/deps/setup_rejson.sh
alonre24: We are passing ROOT to the script when we run tests, this default initialization is for benchmarks where this script is called directly from GitHub actions (and it works) - other ways just look more backward no?

tests/pytests/runtests.sh
alonre24: Probably due to the `cd` added in `setup_rejson`

tests/deps/setup_rejson.sh
alonre24: Looks like it isn't, I can change that

Makefile
```diff
@@ -352,10 +330,15 @@ else ifeq ($(SA),0)
...
+# RedisJSON defaults:
+REJSON ?= 1
+REJSON_BRANCH ?= master
+REJSON_PATH ?=
+
 run:
 ifeq ($(WITH_RLTEST),1)
-	$(SHOW)REJSON=$(REJSON) REJSON_PATH=$(REJSON_PATH) FORCE='' RLTEST= ENV_ONLY=1 LOG_LEVEL=$(LOG_LEVEL) \
-	MODULE=$(MODULE) REDIS_STANDALONE=$(REDIS_STANDALONE) SA=$(SA) \
+	$(SHOW)REJSON=$(REJSON) REJSON_PATH=$(REJSON_PATH) REJSON_BRANCH=$(REJSON_BRNACH) FORCE='' RLTEST= ENV_ONLY=1
```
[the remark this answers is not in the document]
alonre24: nice catch!

tests/pytests/runtests.sh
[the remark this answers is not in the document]
alonre24: Right. I will expose it here and in the makefile as well

tests/deps/setup_rejson.sh
```diff
@@ -0,0 +1,60 @@
...
+    return 0
+fi
+
+# Clone the RedisJSON repository if it doesn't exist
+if [ ! -d "${JSON_MODULE_DIR}" ]; then
+    echo "Cloning RedisJSON repository from ${JSON_REPO_URL} to ${JSON_MODULE_DIR}..."
+    run_command git clone --quiet --recursive $JSON_REPO_URL $JSON_MODULE_DIR
+    echo "Done"
+else
+    echo "RedisJSON already exists in ${JSON_MODULE_DIR}"
```
nafraf, line 38: Shouldn't we exit if the module was found?

alonre24: See https://github.com/RediSearch/RediSearch/pull/5622#discussion_r1959663019

nafraf (approved):

GuyAv46 (approved):

tests/pytests/runtests.sh
GuyAv46: So after fixing the `setup_rejson` to exit at the same location we entered, can we remove it?

tests/deps/setup_rejson.sh
GuyAv46: I think assuming the caller to the script is at the root directory is a bad practice in writing a script. IMO it is much more portable to assume the relative path of the file in the repo and use it to find the root directory, than enforcing the caller to be at the right directory when calling it, especially when the script is not at the root directory

tests/pytests/runtests.sh
alonre24: yes, I removed it

PR #5646, opened 2025-02-12
[2.10] MOD-7944: Support multi-byte char terms (#5391)

nafraf:
# Description
Manual backport of #5391 to `2.10`.
(cherry picked from commit 43701d8e8a46a2ecc03055d2328f2ef81677f064)

* Since Version 2.10 limits the normalization buffer to 128 bytes, `#define MAX_NORMALIZE_SIZE 128`, expected results in `testLongTerms()` were modified to respect that limitation.

* In `testMultibyteTag`, for tests using DIALECT 4, the number of results was changed to ANY, because in this version a wrong count is returned.

* In test_cpp_llapi.cpp, `RediSearch_CreateTagLexRangeNode()` receives only 5 arguments, in this version `const char* fieldName` is not part of the function signature


src/stemmer.c
```diff
@@ -1,9 +1,9 @@
...
- * Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
- * the Server Side Public License v1 (SSPLv1).
- */
-
+/*
+ * Copyright Redis Ltd. 2016 - present
+ * Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
+ * the Server Side Public License v1 (SSPLv1).
+ */
+
```
raz-mon, line 6: This has got to stop happening.. 🤦‍♂️

raz-mon (approved):

oshadmi (approved):

PR #5648, opened 2025-02-13
[2.6] MOD-7944: Support multi-byte char terms (#5391)

nafraf:
# Description
Manual backport of #5391 to `2.6`.
(cherry picked from commit 43701d8e8a46a2ecc03055d2328f2ef81677f064)

Changes:
* Since Version 2.6 limits the normalization buffer to 128 bytes, `#define MAX_NORMALIZE_SIZE 128`, expected results in `testLongTerms()` were modified to respect that limitation.

* For tests with indexes with LANGUAGE different from ENGLISH, the LANGUAGE was added to FT.SEARCH commands because this version does not take the language from index specification 

* Since dialect 4 is not available in this version, test are only done for dialects 1 - 3.

* In test_cpp_llapi.cpp, `RediSearch_CreateTagLexRangeNode()` receives only 5 arguments, in this version `const char* fieldName` is not part of the function signature

* common.py: 
   - Add `config_cmd()` and `run_command_on_all_shards()`
   - Replace `ftDebugCmdName()` by `debug_cmd()`


raz-mon (approved):

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -0,0 +1,1436 @@
...
+            'FT.SEARCH', 'idx', '@t:Etude', 'NOCONTENT', 'SORTBY', 't')
+        env.assertEqual(res, expected, message=f'Dialect: {dialect}')
+
+        # search term with diacritics
+        # the diacritics are not removed, so the terms WITHOUT diacritics are
+        # not found
+        expected = [2, 'mot:2', 'mot:4']
+        res = conn.execute_command(
+            'FT.SEARCH', 'idx', '@t:étude', 'NOCONTENT', 'SORTBY', 't',
+            'LANGUAGE', 'FRENCH')
```
oshadmi, line 478: `LANGUAGE` `FRENCH` is not part of the original PR #5391 but has no effect here since it is the same language as in the index definition, e.g.,

> res = conn.execute_command(
            'FT.SEARCH', 'idx', '@t:étude', 'NOCONTENT', 'SORTBY', 't')
```

oshadmi (approved):

PR #5667, opened 2025-02-17
MOD-8839: Remove assertions from production

raz-mon:
This PR removes our assertions from our production artifacts, by setting the `NDEBUG` flag to our compilation flags upon **not** building in debug mode.

Macros for assertions we wish to keep for production were added as well, see `RS_LOG_ASSERT_FMT_ALWAYS` and `RS_LOG_ASSERT_ALWAYS`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,6 +213,11 @@ endif
 CMAKE_COORD += -DCOORD_TYPE=$(COORD)
 _CMAKE_FLAGS += $(CMAKE_ARGS) $(CMAKE_STATIC) $(CMAKE_COORD) $(CMAKE_TEST) $(CMAKE_LITE)
 
+# If DEBUG is not set, set NDEBUG
+ifeq ($(DEBUG),)
+CC_FLAGS.defs += -DNDEBUG
+endif
+
 include $(MK)/defs
 
 MK_CUSTOM_CLEAN=1
--- a/deps/rmutil/rm_assert.h
+++ b/deps/rmutil/rm_assert.h
@@ -12,8 +12,8 @@
 
 #ifdef NDEBUG
 
-#define RS_LOG_ASSERT(ctx, condition, fmt, ...)    (__ASSERT_VOID_CAST (0))
-#define RS_LOG_ASSERT_STR(ctx, condition, str)     (__ASSERT_VOID_CAST (0))
+#define RS_LOG_ASSERT_FMT(condition, fmt, ...)
+#define RS_LOG_ASSERT(condition, str)
 
 #else
 
@@ -25,7 +25,16 @@
 
 #define RS_LOG_ASSERT(condition, str)  RS_LOG_ASSERT_FMT(condition, str "%s", "")
 
-#endif  //NDEBUG
+#endif  // NDEBUG
+
+// Assertions that we want to keep in production artifacts.
+#define RS_LOG_ASSERT_FMT_ALWAYS(condition, fmt, ...)                                   \
+    if (__builtin_expect(!(condition), 0)) {                                            \
+        RedisModule_Log(RSDummyContext, "warning", (fmt), __VA_ARGS__);                 \
+        RedisModule_Assert(condition); /* Crashes server and create a crash report*/    \
+    }
+
+#define RS_LOG_ASSERT_ALWAYS(condition, str)  RS_LOG_ASSERT_FMT_ALWAYS(condition, str "%s", "")
 
 #define RS_CHECK_FUNC(funcName, ...)                                          \
     if (funcName) {                                                           \
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,6 @@ endif()
 # To avoid this, we set the default signedness of char to unsigned.
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -funsigned-char")
 
-
 add_compile_definitions(
     "REDISEARCH_MODULE_NAME=\"${MODULE_NAME}\""
     "GIT_VERSPEC=\"${GIT_VERSPEC}\""


GuyAv46 (dismissed):
No assertion in the code is worth keeping at the moment, or are we planning to review the existing assertions in a follow-up PR?

Do we see a performance impact by removing the assertions?

deps/rmutil/rm_assert.h
```diff
@@ -1,35 +1,43 @@
...
 
 #define RS_LOG_ASSERT(condition, str)  RS_LOG_ASSERT_FMT(condition, str "%s", "")
 
-#endif  //NDEBUG
+#endif  // NDEBUG
+
+// Assertions that we want to keep in production artifacts.
+#define RS_LOG_ASSERT_FMT_ALWAYS(condition, fmt, ...) _RS_LOG_ASSERT(condition, fmt, __VA_ARGS__)
+
+#define RS_LOG_ASSERT_ALWAYS(condition, str)  RS_LOG_ASSERT_FMT_ALWAYS(condition, str "%s", "")
```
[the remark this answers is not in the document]
raz-mon: That's probably it, will take a look if it's needed.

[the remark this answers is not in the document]
raz-mon: It's needed for the `RedisModule_Log` call that it's using.

deps/rmutil/rm_assert.h
```diff
@@ -1,35 +1,43 @@
...
+        RedisModule_Log(RSDummyContext, "warning", (fmt), __VA_ARGS__);              \
+        RedisModule_Assert(condition); /* Crashes server and create a crash report*/ \
+    }
+
 #ifdef NDEBUG
 
-#define RS_LOG_ASSERT(ctx, condition, fmt, ...)    (__ASSERT_VOID_CAST (0))
-#define RS_LOG_ASSERT_STR(ctx, condition, str)     (__ASSERT_VOID_CAST (0))
+#define RS_LOG_ASSERT_FMT(condition, fmt, ...)
+#define RS_LOG_ASSERT(condition, str)
```
DvirDukhan, line 23: comment `NOP`

raz-mon:
@GuyAv46 Currently no assertions are staying. We can traverse the assertions and discuss if we find some that we may want to keep.
Regarding performance - haven't checked that, will run the benchmarks when done.
Turns out that there are issues with setting the `NDEBUG` flag from the Makefile and CMakeLists.txt file only, so we may need to merge https://github.com/RedisLabsModules/readies/pull/145 for this (cc @DvirDukhan).

src/coord/rmr/rq.c
```diff
@@ -124,7 +125,7 @@ static void verify_uv_thread() {
     uv_async_init(uv_default_loop(), &topologyAsync, topologyAsyncCB);
     // Verify that we are running on the event loop thread
     int uv_thread_create_status = uv_thread_create(&loop_th, sideThread, NULL);
-    RedisModule_Assert(uv_thread_create_status == 0);
+    RS_ASSERT(uv_thread_create_status == 0);
```
GuyAv46, line 128: I'm not sure if this is aligned in all versions, just so you are aware. In previous versions, the call to `uv_thread_create` may still be in the assertion expression, so we must move it out like here

src/rdb.h
```diff
@@ -21,7 +21,7 @@ do {                                                                \
   if (RedisModule_IsIOError(rdb)) {                                 \
     cleanup_exp;                                                    \
   }                                                                 \
-  RedisModule_Assert(oldbuf);                                       \
+  RS_ASSERT(oldbuf);                                       \
```
GuyAv46, line 24:
```suggestion
  RS_ASSERT(oldbuf);                                                \
```

src/redis_index.c
```diff
@@ -235,7 +235,7 @@ void RedisSearchCtx_UnlockSpec(RedisSearchCtx *sctx) {
   if (sctx->flags == RS_CTX_READONLY) {
     // We paused rehashing when we locked the spec for read. Now we can resume it.
     // Assert that it was actually previously paused
-    RedisModule_Assert(dictResumeRehashing(sctx->spec->keysDict));
+    RS_ASSERT(dictResumeRehashing(sctx->spec->keysDict));
```
GuyAv46, line 238: Move call out of assertion, or use `*ALWAYS` API

GuyAv46 (dismissed):
🦾

raz-mon:
Note that on sanitizer we ALWAYS have the `NDEBUG` flag **unset**, so that the assertions are active.

tests/cpptests/test_cpp_index.cpp
```diff
@@ -294,8 +294,7 @@ TEST_F(IndexTest, testGetEncoderAndDecoders) {
 
     // invalid flags combination
     default:
-      ASSERT_FALSE(InvertedIndex_GetDecoder(IndexFlags(curFlags)).decoder);
-      ASSERT_FALSE(InvertedIndex_GetEncoder(IndexFlags(curFlags)));
+      continue;
```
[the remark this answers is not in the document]
raz-mon: Due to this: https://github.com/RediSearch/RediSearch/pull/5667#issuecomment-2671409651
We have the assertions on always on our sanitizer build.

GuyAv46 (dismissed):

GuyAv46 (dismissed):

deps/rmutil/rm_assert.h
```diff
@@ -1,35 +1,48 @@
...
 
 #else
 
-#define RS_LOG_ASSERT_FMT(condition, fmt, ...)                                          \
-    if (__builtin_expect(!(condition), 0)) {                                            \
-        RedisModule_Log(RSDummyContext, "warning", (fmt), __VA_ARGS__);                 \
-        RedisModule_Assert(condition); /* Crashes server and create a crash report*/    \
-    } 
-
+#define RS_LOG_ASSERT_FMT(condition, fmt, ...) _RS_LOG_ASSERT_FMT(condition, fmt, __VA_ARGS__)
```
kei-nan, line 30: It seems like the code can be:
```suggestion
#ifdef NDEBUG
#define RS_LOG_DEBUG_ASSERT_FMT(condition, fmt, ...)
#else
#define RS_LOG_DEBUG_ASSERT_FMT(condition, fmt, ...) _RS_LOG_ASSERT_FMT(condition, fmt, __VA_ARGS__)
#endif

#define RS_LOG_ASSERT(condition, str)  RS_LOG_DEBUG_ASSERT_FMT(condition, str "%s", "")
#define RS_ASSERT(condition) RS_LOG_DEBUG_ASSERT_FMT(condition, "Assertion failed: %s", #condition)
#define RS_ABORT(str) RS_LOG_DEBUG_ASSERT_FMT(0, "Aborting: %s", str)
```

deps/rmutil/rm_assert.h
```diff
@@ -1,35 +1,48 @@
...
 #define RS_LOG_ASSERT(condition, str)  RS_LOG_ASSERT_FMT(condition, str "%s", "")
+#define RS_ASSERT(condition) _RS_LOG_ASSERT_FMT(condition, "Assertion failed: %s", #condition)
+#define RS_ABORT(str) _RS_LOG_ASSERT_FMT(0, "Aborting: %s", str)
+
+#endif  // NDEBUG
 
-#endif  //NDEBUG
```
kei-nan, line 28: It seems like the code can be:
```suggestion
#ifdef NDEBUG
#define RS_LOG_DEBUG_ASSERT_FMT(condition, fmt, ...)
#else
#define RS_LOG_DEBUG_ASSERT_FMT(condition, fmt, ...) _RS_LOG_ASSERT_FMT(condition, fmt, __VA_ARGS__)
#endif

#define RS_LOG_ASSERT(condition, str)  RS_LOG_DEBUG_ASSERT_FMT(condition, str "%s", "")
#define RS_ASSERT(condition) RS_LOG_DEBUG_ASSERT_FMT(condition, "Assertion failed: %s", #condition)
#define RS_ABORT(str) RS_LOG_DEBUG_ASSERT_FMT(0, "Aborting: %s", str)
```

kei-nan: Also, please add some comment for when developers should use this assert.

src/config.c
```diff
@@ -1235,14 +1235,14 @@ void UpgradeDeprecatedMTConfigs() {
 
 char *getRedisConfigValue(RedisModuleCtx *ctx, const char* confName) {
   RedisModuleCallReply *rep = RedisModule_Call(ctx, "config", "cc", "get", confName);
-  RedisModule_Assert(RedisModule_CallReplyType(rep) == REDISMODULE_REPLY_ARRAY);
+  RS_ASSERT(RedisModule_CallReplyType(rep) == REDISMODULE_REPLY_ARRAY);
```
kei-nan, line 1238: Do we want to continue parsing a reply if the type is not an array?
Why should this assert not exist in production?
Same goes for the other asserts in this function.

src/coord/info_command.c
```diff
@@ -164,8 +164,8 @@ static void convertField(InfoValue *dst, MRReply *src, InfoFieldType type) {
 // Extract an array of FieldSpecInfo from MRReply
 void handleFieldStatistics(InfoFields *fields, MRReply *src, QueryError *error) {
   // Input validations
-  RedisModule_Assert(src && fields);
-  RedisModule_Assert(MRReply_Type(src) == MR_REPLY_ARRAY);
+  RS_ASSERT(src && fields);
+  RS_ASSERT(MRReply_Type(src) == MR_REPLY_ARRAY);
```
kei-nan, line 168: I think the reply type is an important thing to assert on
especially when we don't cover that with an if in the code.

kei-nan (changes requested):
Looking at the code I am getting worried.
I have experience in the past with projects where most of the asserts didn't exist in production.

You try and analyze an issue and you see an assert in the code and it can be very easy to think it actually checks the condition in production.

This drops assurances under our feet and can lead to us taking a big step backward in our ability to analyse non-debug builds.

I personally think more asserts should be production asserts.

src/index.c
```diff
@@ -1848,7 +1848,7 @@ PRINT_PROFILE_FUNC(printUnionIt) {
   case QN_LEXRANGE : unionTypeStr = "LEXRANGE"; break;
   case QN_WILDCARD_QUERY : unionTypeStr = "WILDCARD"; break;
   default:
-    RS_LOG_ASSERT(0, "Invalid type for union");
+    RS_ABORT("Invalid type for union");
```
kei-nan, line 1851: If we end up in the default case then   char *unionTypeS will point at garbage.
We then try and send that grabage string in the reply.
I think this should go back to being a production assert.

Or you should nullify the string and handle the case where it remains null after the switch case.

raz-mon:
@kei-nan The idea is not to have assertions in production.
From two main reasons:
1. Don’t crash when you don’t need to.
2. Save the checks to enhance performance.
The extra reason:
3. Encourage the team to add MANY assertions to the code, to make sure all of our assumptions/constraints are strictly held.

src/index.c
raz-mon: We should fail on this in development time, not in production.
We may have places that need some better failure handling, this is not the PR for that, although we can add some handling for high-risk places, it will not be the whole assertion set.

src/coord/info_command.c
raz-mon: I think it is enough to do this in development and avoid this extra check in production.

src/config.c
raz-mon: The idea is that with good coverage and assertions in development time, we are sure enough that this will not happen in production, and avoid the checks to be as efficient as possible, and to avoid a crash if we can.

This is a matter of approach alignment, CC @DvirDukhan.

deps/rmutil/rm_assert.h
raz-mon: Sure, I can simplify it more to something like that.

src/aggregate/expr/expression.h
```diff
@@ -70,7 +70,7 @@ static const char *getRSConditionStrings(RSCondition type) {
   case RSCondition_And: return "&&";
   case RSCondition_Or: return "||";
   default:
-    RS_LOG_ASSERT(0, "oops");
+    RS_ABORT("oops");
```
kei-nan, line 73: not returning something in the default case

src/field_spec.c
```diff
@@ -44,7 +44,7 @@ const char *FieldSpec_GetTypeNames(int idx) {
   case IXFLDPOS_GEOMETRY: return SPEC_GEOMETRY_STR;
 
   default:
-    RS_LOG_ASSERT(0, "oops");
+    RS_ABORT("oops");
```
kei-nan, line 47: not returning anything in the default case.

src/info/field_spec_info.c
```diff
@@ -123,26 +123,26 @@ void FieldSpecInfo_Combine(FieldSpecInfo *info, const FieldSpecInfo *other) {
 // Deserializes a FieldSpecInfo from a MRReply.
 FieldSpecInfo FieldSpecInfo_Deserialize(const MRReply *reply) {
     FieldSpecInfo info = FieldSpecInfo_Init();
-    RedisModule_Assert(reply);
+    RS_ASSERT(reply);
     // Validate the reply type - array or map.
-    RedisModule_Assert(MRReply_Type(reply) == MR_REPLY_MAP || (MRReply_Type(reply) == MR_REPLY_ARRAY && MRReply_Length(reply) % 2 == 0));
+    RS_ASSERT(MRReply_Type(reply) == MR_REPLY_MAP || (MRReply_Type(reply) == MR_REPLY_ARRAY && MRReply_Length(reply) % 2 == 0));
```
kei-nan, line 128: reply type and length - seems important

src/info/index_error.c
```diff
@@ -143,32 +144,32 @@ IndexError IndexError_Deserialize(MRReply *reply) {
     IndexError error = IndexError_Init();
 
     // Validate the reply. It should be a map with 3 elements.
-    RedisModule_Assert(reply && (MRReply_Type(reply) == MR_REPLY_MAP || (MRReply_Type(reply) == MR_REPLY_ARRAY && MRReply_Length(reply) % 2 == 0)));
+    RS_ASSERT(reply && (MRReply_Type(reply) == MR_REPLY_MAP || (MRReply_Type(reply) == MR_REPLY_ARRAY && MRReply_Length(reply) % 2 == 0)));
```
kei-nan, line 147: reply type and length

src/json.c
```diff
@@ -607,7 +607,7 @@ int JSON_StoreInDocField(RedisJSON json, JSONType jsonType, FieldSpec *fs, struc
       rv = REDISMODULE_ERR;
       break;
     case JSONType__EOF:
-      RS_LOG_ASSERT(0, "Should not happen");
+      RS_ABORT("Should not happen");
```
kei-nan, line 610: I think rv should be set to REDISMODULE_ERR

src/module.c
```diff
@@ -2210,10 +2210,10 @@ static int cmp_scored_results(const void *p1, const void *p2, const void *udata)
 }
 
 static double parseNumeric(const char *str, const char *sortKey) {
-    RedisModule_Assert(str[0] == '#');
+    RS_ASSERT(str[0] == '#');
     char *eptr;
     double d = strtod(str + 1, &eptr);
-    RedisModule_Assert(eptr != sortKey + 1 && *eptr == 0);
+    RS_ASSERT(eptr != sortKey + 1 && *eptr == 0);
```
kei-nan, line 2216: this code looks very odd.
eptr will originate from str, 
doesn't seem like sort key is used properly in this function.

src/module.c
```diff
@@ -3537,7 +3537,7 @@ void Initialize_CoordKeyspaceNotifications(RedisModuleCtx *ctx) {
 
 static bool checkClusterEnabled(RedisModuleCtx *ctx) {
   RedisModuleCallReply *rep = RedisModule_Call(ctx, "CONFIG", "cc", "GET", "cluster-enabled");
-  RedisModule_Assert(rep && RedisModule_CallReplyType(rep) == REDISMODULE_REPLY_ARRAY &&
+  RS_ASSERT(rep && RedisModule_CallReplyType(rep) == REDISMODULE_REPLY_ARRAY &&
```
kei-nan, line 3540: IMHO important

src/notifications.c
```diff
@@ -407,7 +407,7 @@ int CheckVersionForShortRead() {
 void Initialize_RdbNotifications(RedisModuleCtx *ctx) {
   if (CheckVersionForShortRead() == REDISMODULE_OK) {
     int success = RedisModule_SubscribeToServerEvent(ctx, RedisModuleEvent_ReplBackup, ReplicaBackupCallback);
-    RedisModule_Assert(success != REDISMODULE_ERR); // should be supported in this redis version/release
+    RS_ASSERT(success != REDISMODULE_ERR); // should be supported in this redis version/release
```
kei-nan, line 410: here

src/redis_index.c
```diff
@@ -404,7 +404,7 @@ int Redis_DropScanHandler(RedisModuleCtx *ctx, RedisModuleString *kn, void *opaq
 
 int Redis_DeleteKey(RedisModuleCtx *ctx, RedisModuleString *s) {
   RedisModuleCallReply *rep = RedisModule_Call(ctx, "DEL", "s", s);
-  RedisModule_Assert(RedisModule_CallReplyType(rep) == REDISMODULE_REPLY_INTEGER);
+  RS_ASSERT(RedisModule_CallReplyType(rep) == REDISMODULE_REPLY_INTEGER);
```
kei-nan, line 407: here

src/redis_index.c
```diff
@@ -413,7 +413,7 @@ int Redis_DeleteKey(RedisModuleCtx *ctx, RedisModuleString *s) {
 int Redis_DeleteKeyC(RedisModuleCtx *ctx, char *cstr) {
   // Send command and args to replicas and AOF
   RedisModuleCallReply *rep = RedisModule_Call(ctx, "DEL", "c!", cstr);
-  RedisModule_Assert(RedisModule_CallReplyType(rep) == REDISMODULE_REPLY_INTEGER);
+  RS_ASSERT(RedisModule_CallReplyType(rep) == REDISMODULE_REPLY_INTEGER);
```
kei-nan, line 416: here

src/tag_index.c
```diff
@@ -164,7 +164,7 @@ int TagIndex_Preprocess(const FieldSpec *fs, const DocumentField *data, FieldInd
   case FLD_VAR_T_NUM:
   case FLD_VAR_T_BLOB_ARRAY:
   case FLD_VAR_T_GEOMETRY:
-    RS_LOG_ASSERT(0, "nope")
+    RS_ABORT("nope")
```
kei-nan, line 167: it seems like we leak memory in `FLD_VAR_T_NULL`
maybe we should also mark type as null in abort cases?

src/rlookup.c
```diff
@@ -443,7 +443,7 @@ static RSValue *jsonValToValue(RedisModuleCtx *ctx, RedisJSON json) {
     case JSONType__EOF:
       break;
   }
-  RS_LOG_ASSERT(0, "Cannot get here");
+  RS_ABORT("Cannot get here");
```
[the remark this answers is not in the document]
raz-mon: Added.

src/notifications.c
raz-mon: This can be kept AFAIC.

src/module.c
raz-mon: On module load.
Can be kept AFAIC.

src/module.c
raz-mon: Agree, this looks weird.
Let's open a ticket, this is not the place to fix it.

src/json.c
raz-mon: Agreed and added.

src/info/index_error.c
raz-mon: You think we should keep?
Why?

src/info/field_spec_info.c
raz-mon: I don't agree that this should be checked in production.
We cover this well enough in our development. This shouldn't change in production.

src/field_spec.c
raz-mon: Transformed to the `ALWAYS` version since we don't have a "good" value to return here.
@DvirDukhan LMK if you think differently.

src/aggregate/expr/expression.h
raz-mon: Changed to always - we should not get here.

tests/cpptests/index_utils.cpp
```diff
@@ -10,7 +10,8 @@ std::string numToDocStr(unsigned id) {
 
 size_t addDocumentWrapper(RedisModuleCtx *ctx, RSIndex *index, const char *docid, const char *field, const char *value) {
     size_t beforAddMem = get_spec(index)->stats.invertedSize;
-    assert(RS::addDocument(ctx, index, docid, field, value));
+    bool rv = RS::addDocument(ctx, index, docid, field, value);
+    assert(rv);
```
kei-nan, line 14: should this assert also change to RS_ASSERT?

src/tag_index.c
raz-mon: Why do we leak?
And not sure I understand your suggestion.

tests/cpptests/index_utils.cpp
raz-mon: No.
This assert goes away as well when `NDEBUG` is on.

src/tag_index.c
kei-nan: We don't free `arr` local variable.
it is initialized with ` arrayof(char*) arr = array_new(char *, 4);`
Regarding marking type as null see the code for the `FLD_VAR_T_NULL` case.

src/redis_index.c
kei-nan: Looks like the result value is ignored and is only used in legacy indexes.
I guess it is not needed.

src/info/index_error.c
kei-nan: Looks like `MRReply_MapElement` will return null if type is not map.
It also handles the case the element count is odd by returning null.
So I guess it is okay and assert is not needed in production.

src/coord/info_command.c
kei-nan: Looks like the MRReply functions are safe in this case.

src/index.c
kei-nan: Until we fix this code I think the assert should be a production assert.

src/info/field_spec_info.c
kei-nan: Took a closer look at the MR reply functions, they seem to handle the edge cases pretty well.

src/tag_index.c
raz-mon: You mean not freeing it before we crash? Otherwise we don't free it anyways until later.

Regarding NULL - that doesn't make sense to me, this is a dedicated type for the JSON NULL value, why would you mark other types as it?

src/index.c
raz-mon: I'm fine with that in this case, since we are not supposed to get here.

kei-nan (approved):

GuyAv46 (approved):

GuyAv46 (approved):

kei-nan (approved):

kei-nan (approved):

Makefile
```diff
@@ -215,8 +215,17 @@ endif
 CMAKE_COORD += -DCOORD_TYPE=$(COORD)
 _CMAKE_FLAGS += $(CMAKE_ARGS) $(CMAKE_STATIC) $(CMAKE_COORD) $(CMAKE_TEST) $(CMAKE_LITE)
 
+
 include $(MK)/defs
 
+# If DEBUG is not set, set NDEBUG. This will remove all assertions from the code.
```
oshadmi, line 221: Need to review whether all assertions are removed or some are still kept

.github/workflows/task-test.yml
```diff
@@ -202,26 +202,49 @@ jobs:
...
-        run: make pytest SA=0 LOG=1 CLEAR_LOGS=0 SAN=${{ inputs.san }} ${{ inputs.test-config }}
-          REJSON=${{ env.REJSON }} REJSON_BRANCH=${{ inputs.rejson-branch }}
+        env:
+          SAN: ${{ inputs.san }}
+          LOG: 1
+          CLEAR_LOGS: 0
+          SA: 0
+          REJSON: ${{ inputs.rejson }}
+          REJSON_BRANCH: ${{ inputs.rejson-branch }}
+          DEBUG: 1
```
oshadmi, line 246: Suggestion: Not real `DEBUG` but only to allow assertions. Consider something like `DIAG: 1` for diagnostics

src/aggregate/aggregate_exec.c
```diff
@@ -521,7 +521,7 @@ static void sendChunk_Resp2(AREQ *req, RedisModule_Reply *reply, size_t limit,
 
     if (resultsLen != REDISMODULE_POSTPONED_ARRAY_LEN && rc == RS_RESULT_OK && resultsLen != nelem) {
       RedisModule_Log(RSDummyContext, "warning", "Failed to predict the number of replied results. Prediction=%ld, actual_number=%ld.", resultsLen, nelem);
-      RS_LOG_ASSERT(0, "Precalculated number of replies must be equal to actual number");
+      RS_ABORT("Precalculated number of replies must be equal to actual number");
```
oshadmi, line 524: Could be dropped since it is next to a `RedisModule_Log`?

raz-mon: How does the log achieve the effect of an assertion?

.github/workflows/task-test.yml
raz-mon: See added `ENABLE_ASSERT` compilation flag

src/aggregate/aggregate_exec.c
raz-mon: Dropped

GuyAv46 (approved):

kei-nan (approved):

GuyAv46 (approved):

PR #5672, opened 2025-02-23
MOD-8862: Add DebugCTX mechanism and DebugIndexesScanner

lerman25:
This pull request adds several new ft.debug commands, DebugCTX global and a DebugIndexesScanner "derived" from IndexesScanner.

The new debug commands are:
1. SET_MAX_SCANNED_DOCS -  limit the amount of  docs that can be scanned in a background-scan. 
2. SET_PAUSE_ON_SCANNED_DOCS - pausing the scan after an amount of docs scanned.
3. SET_PAUSE_BEFORE_SCAN - pausing the scanning process before reading the first docm
4. SET_RESUME - resuming any paused state (such as 2,3).
5. GET_DEBUG_SCANNER_STATUS  -  getting the status of the (debug) scanner.

Changing 1,2,3 will put DebugCTX in debug mode, which is revertible by setting negative/zero values.

DebugCTX can be used for other debugging mechanism other then indexes scan.

Also remove whitespaces from a previous pytest


tests/pytests/common.py
```diff
@@ -820,3 +820,28 @@ def field_errors(env, idx = 'idx', fld_index = 0):
 def shardsConnections(env):
   for s in range(1, env.shardsCount + 1):
       yield env.getConnection(shardId=s)
+
+def waitForIndexFinishScan(env, idx = 'idx'):
+    scan_started = False
+    scan_ended = False
```
GuyAv46, line 826: Redundant
```suggestion
```

tests/pytests/common.py
```diff
@@ -820,3 +820,28 @@ def field_errors(env, idx = 'idx', fld_index = 0):
...
+def checkDebugScannerError(env, idx = 'idx', expected_error = ''):
+    env.expect(debug_cmd(), 'GET_DEBUG_SCANNER_STATUS', idx).error() \
+        .contains(expected_error)
+
+def waitForIndexPauseScan(env, idx = 'idx'):
+    while True:
+        res = getDebugScannerStatus(env, idx)
+        if res=='PAUSED':
+            break
+        time.sleep(0.1)
```
GuyAv46, line 847:
```suggestion
    while getDebugScannerStatus(env, idx) != 'PAUSED':
        time.sleep(0.1)
```

src/spec.c
```diff
@@ -2246,7 +2334,15 @@ static void IndexSpec_ScanAndReindexAsync(StrongRef spec_ref) {
...
   RedisModule_Log(RSDummyContext, "notice", "Register index %s for async scan", ((IndexSpec*)StrongRef_Get(spec_ref))->name);
 #endif
-  IndexesScanner *scanner = IndexesScanner_New(spec_ref);
+  IndexesScanner *scanner = NULL;
+  if (debugCtx.debugMode) {
+    scanner = (IndexesScanner*)DebugIndexesScanner_New(spec_ref);
+    IndexSpec *sp = StrongRef_Get(spec_ref);
+    sp->flags |= Index_DebugScanner;
+  }
+  else {
```
GuyAv46, line 2343:
```suggestion
  } else {
```

src/spec.c
```diff
@@ -2246,7 +2334,15 @@ static void IndexSpec_ScanAndReindexAsync(StrongRef spec_ref) {
 #ifdef _DEBUG
   RedisModule_Log(RSDummyContext, "notice", "Register index %s for async scan", ((IndexSpec*)StrongRef_Get(spec_ref))->name);
 #endif
-  IndexesScanner *scanner = IndexesScanner_New(spec_ref);
+  IndexesScanner *scanner = NULL;
```
GuyAv46, line 2337:
```suggestion
  IndexesScanner *scanner;
```

src/spec.c
```diff
@@ -2246,7 +2334,15 @@ static void IndexSpec_ScanAndReindexAsync(StrongRef spec_ref) {
...
+  if (debugCtx.debugMode) {
+    scanner = (IndexesScanner*)DebugIndexesScanner_New(spec_ref);
+    IndexSpec *sp = StrongRef_Get(spec_ref);
+    sp->flags |= Index_DebugScanner;
```
GuyAv46, line 2341: Does that mean the current scanner is a debug scanner?

1. What happens if the current scanner is being canceled?
2. When do we turn off this flag?

src/spec.c
```diff
@@ -2195,7 +2256,21 @@ static void Indexes_ScanAndReindexTask(IndexesScanner *scanner) {
...
+    {
+      debugCtx.pause = true;
+      RedisModule_ThreadSafeContextUnlock(ctx);
+      while (debugCtx.pause) // volatile variable
+      {
+        usleep(1000);
+      }
+      RedisModule_ThreadSafeContextLock(ctx);
+    }
+  }
```
GuyAv46, line 2272: Please fix code-blocks style

src/spec.c
```diff
@@ -2176,6 +2201,42 @@ static void Indexes_ScanProc(RedisModuleCtx *ctx, RedisModuleString *keyname, Re
...
+
+  if ((dScanner->maxDocsTBscanned > 0) && (scanner->scannedKeys >= dScanner->maxDocsTBscanned)) {
+    scanner->cancelled = true;
+    dScanner->status = DEBUG_INDEX_SCANNER_CODE_CANCELLED;
+  }
+
+  RedisModule_ThreadSafeContextUnlock(ctx);
+  while (debugCtx.pause) // volatile variable
+  {
```
GuyAv46, line 2226:
```suggestion
  while (debugCtx.pause) { // volatile variable
```

src/spec.c
```diff
@@ -2121,6 +2124,27 @@ void IndexesScanner_Free(IndexesScanner *scanner) {
...
+  dScanner->status = DEBUG_INDEX_SCANNER_CODE_NEW;
+
+  IndexSpec *spec = StrongRef_Get(global_ref);
+  spec->scanner = (IndexesScanner*)dScanner;
+
+  return dScanner;
+}
+
+void DebugIndexesScanner_Free(DebugIndexesScanner *dScanner)
+{
```
GuyAv46, line 2143:
```suggestion
void DebugIndexesScanner_Free(DebugIndexesScanner *dScanner) {
```

src/debug_commands.h
```diff
@@ -17,3 +18,14 @@ typedef struct DebugCommandType {
 } DebugCommandType;
 
 int RegisterDebugCommands(RedisModuleCommand *debugCommand);
+
+typedef struct DebugCTX {
+  bool debugMode;
+  int maxDocsTBscanned;
+  int maxDocsTBscannedPause;
+  bool pauseBeforeScan;
+  volatile bool pause;
```
GuyAv46, line 27: Consider renaming `pause` and `pauseBeforeScan` to indicate they are related to BG indexing, or wrap all the BG indexing attributes in a BG indexing dedicated struct (even though today it's the only set of attributes)

src/debug_commands.c
```diff
@@ -21,6 +21,34 @@
 #include "util/workers.h"
 #include "cursor.h"
 
+
+DebugCTX debugCtx = {0};
+
+void validateDebugMode(DebugCTX *debugCtx)
+{
```
GuyAv46, line 28: fix code-block style in this function

src/debug_commands.c
```diff
@@ -21,6 +21,34 @@
...
+  if (debugCtx->maxDocsTBscannedPause > 0)
+  {
+    debugCtx->debugMode = true;
+    return;
+  }
+  if (debugCtx->pauseBeforeScan)
+  {
+    debugCtx->debugMode = true;
+    return;
+  }
```
GuyAv46, line 46: Lots of code duplication, and setting to even more in the future.
Consider setting `debugCtx->debugMode` to the result of a complex condition, with each case on its own line and `||` between them

src/debug_commands.c
```diff
@@ -1402,6 +1430,123 @@ DEBUG_COMMAND(WorkerThreadsSwitch) {
...
+  if (argc != 3) {
+    return RedisModule_WrongArity(ctx);
+  }
+  long long pauseBeforeScan;
+  if (RedisModule_StringToLongLong(argv[2], &pauseBeforeScan) != REDISMODULE_OK) {
+    return RedisModule_ReplyWithError(ctx, "Invalid argument for 'SET_PAUSE_BEFORE_SCAN'");
+  }
+
+  debugCtx.pauseBeforeScan = !!pauseBeforeScan;
```
GuyAv46, line 1543: If this is a boolean option, consider allowing only `ON`/`OFF` or `TRUE`/`FALSE` (can be case insensitive) instead of numeric values

src/spec.h
```diff
@@ -122,6 +122,17 @@ struct DocumentIndexer;
...
+#define DEBUG_INDEX_SCANNER_CODE_NEW 0
+#define DEBUG_INDEX_SCANNER_CODE_RUNNING 1
+#define DEBUG_INDEX_SCANNER_CODE_DONE 2
+#define DEBUG_INDEX_SCANNER_CODE_CANCELLED 3
+#define DEBUG_INDEX_SCANNER_CODE_PAUSED 4
+#define DEBUG_INDEX_SCANNER_CODE_RESUMED 5
+
+#define DEBUG_INDEX_SCANNER_STATUS_STRS \
+  { "NEW", "SCANNING", "DONE", "CANCELLED", "PAUSED", "RESUMED" }
+
```
GuyAv46, line 135: 1. Consider using an `enum`
2. Consider moving all of these to `debug_commands.h` (status messages can maybe be on the c file only)
3. Consider adding a static assertion that the length of the status array is equal to the number of statuses (we have something similar for the coordinator's debug commands), so if someone adds/removes a status he will have to change the status messages. Alternatively, implement something similar to the query error messages array

GuyAv46 (commented):
Very nice implementation
Some styling comments and a few suggestions
Keep up the good work!

src/debug_commands.h
lerman25: I was thinking that `pause` can be used by other flows as well for any pause-resume need.
I agree about the BG indexing struct.

src/spec.c
lerman25: If we start in scan while in debugMode we assume we need a debug scanner.
Each Spec is assigned a  new scanner (once), if it's canceled then the scan is finished, and whatever not indexed so far will not be indexed. The scanner object is freed later on.
So the flag should always stay up, as we don't create a another scanner for that index.
Does this makes sense?

src/spec.h
lerman25: After we spoke, decided to keep it in spec.h
Changed to enum for cleaner and safer implementation.

src/spec.c
lerman25: Debug scanner flag was moved into IndexesScanner base class.

src/debug_commands.c
```diff
@@ -1433,6 +1562,12 @@ DebugCommandType commands[] = {{"DUMP_INVIDX", DumpInvertedIndex}, // Print all
                                {"DUMP_HNSW", dumpHNSWData},
                                {"SET_MONITOR_EXPIRATION", setMonitorExpiration},
                                {"WORKERS", WorkerThreadsSwitch},
+                               {"SET_MAX_SCANNED_DOCS", setMaxScannedDocs},
```
alonre24, line 1565: I would suggest putting all of this as subcommands under one debug command `BG_SCAN_CONTROLLER` or something like that (see how we do it in `WorkerThreadsSwitch`)

src/debug_commands.h
```diff
@@ -17,3 +19,20 @@ typedef struct DebugCommandType {
...
 
 int RegisterDebugCommands(RedisModuleCommand *debugCommand);
+
+// Struct used for debugging background indexing
+typedef struct BgIndexingDebugCtx {
+  int maxDocsTBscanned;
+  int maxDocsTBscannedPause;
+  bool pauseBeforeScan;
+  volatile atomic_bool pause;
+} BgIndexingDebugCtx;
```
alonre24, line 29: Let's document the purpose of each field here

src/spec.c
```diff
@@ -2176,6 +2213,42 @@ static void Indexes_ScanProc(RedisModuleCtx *ctx, RedisModuleString *keyname, Re
   ++scanner->scannedKeys;
 }
 
+
+static void DebugIndexes_ScanProc(RedisModuleCtx *ctx, RedisModuleString *keyname, RedisModuleKey *key,
```
alonre24, line 2217: Please move all the debug functions to a separate file, or at least to the end of this file with a clear separation

src/spec.c
```diff
@@ -2176,6 +2213,42 @@ static void Indexes_ScanProc(RedisModuleCtx *ctx, RedisModuleString *keyname, Re
...
+
+  RedisModule_ThreadSafeContextUnlock(ctx);
+  while (debugCtx.bgIndexing.pause) // volatile variable
+  {
+    dScanner->status = DEBUG_INDEX_SCANNER_CODE_PAUSED;
+    usleep(1000);
+  }
+  RedisModule_ThreadSafeContextLock(ctx);
+
+  if(dScanner->status == DEBUG_INDEX_SCANNER_CODE_PAUSED) {
```
alonre24, line 2244: nitpick: space after `if`

src/debug_commands.c
```diff
@@ -21,6 +21,20 @@
 #include "util/workers.h"
 #include "cursor.h"
 
+
+DebugCTX debugCtx = {0};
```
alonre24, line 25:
```suggestion
DebugCTX globalDebugCtx = {0};
```

src/spec.c
```diff
@@ -2195,7 +2268,16 @@ static void Indexes_ScanAndReindexTask(IndexesScanner *scanner) {
...
-  while (RedisModule_Scan(ctx, cursor, (RedisModuleScanCB)Indexes_ScanProc, scanner)) {
+  RedisModuleScanCB scanner_func = (RedisModuleScanCB)Indexes_ScanProc;
+  if (debugCtx.debugMode) {
+    scanner_func = (RedisModuleScanCB)DebugIndexes_ScanProc;
+    RedisModule_ThreadSafeContextUnlock(ctx);
+    while (debugCtx.bgIndexing.pause) { // volatile variable
+      usleep(1000);
+    }
+    RedisModule_ThreadSafeContextLock(ctx);
+  }
```
alonre24, line 2279: Please add documentation about what is going on here

alonre24 (approved):

src/spec.c
```diff
@@ -3217,3 +3269,62 @@ void Indexes_ReplaceMatchingWithSchemaRules(RedisModuleCtx *ctx, RedisModuleStri
...
+
+  return dScanner;
+}
+
+// This function is currently not called as the pointer is freed via the IndexSpec_Free function
+// It is kept here for completeness and in case it is needed in the future
+static void DebugIndexesScanner_Free(DebugIndexesScanner *dScanner)
+{
+  IndexesScanner_Free((IndexesScanner*)dScanner);
+}
```
GuyAv46, line 3296: Not covered - is it unused?
```suggestion
static void DebugIndexesScanner_Free(DebugIndexesScanner *dScanner) {
  IndexesScanner_Free((IndexesScanner*)dScanner);
}
```

src/spec.c
```diff
@@ -2215,6 +2249,13 @@ static void Indexes_ScanAndReindexTask(IndexesScanner *scanner) {
                   scanner->scannedKeys);
       goto end;
     }
+
+
```
GuyAv46, line 2253: 🧑‍🚀

src/spec.c
```diff
@@ -3217,3 +3269,62 @@ void Indexes_ReplaceMatchingWithSchemaRules(RedisModuleCtx *ctx, RedisModuleStri
...
+  }
+
+  if ((dScanner->maxDocsTBscanned > 0) && (scanner->scannedKeys >= dScanner->maxDocsTBscanned)) {
+    scanner->cancelled = true;
+    dScanner->status = DEBUG_INDEX_SCANNER_CODE_CANCELLED;
+  }
+
+  RedisModule_ThreadSafeContextUnlock(ctx);
+  while (globalDebugCtx.bgIndexing.pause) // volatile variable
+  {
```
GuyAv46, line 3319:
```suggestion
  while (globalDebugCtx.bgIndexing.pause) { // volatile variable
```

src/spec.h
```diff
@@ -122,6 +122,21 @@ struct DocumentIndexer;
...
+    DEBUG_INDEX_SCANNER_CODE_CANCELLED,
+    DEBUG_INDEX_SCANNER_CODE_PAUSED,
+    DEBUG_INDEX_SCANNER_CODE_RESUMED,
+
+    //Insert new codes here (before COUNT)
+    DEBUG_INDEX_SCANNER_CODE_COUNT  // Helps with array size checks
+    //Do not add new codes after COUNT
+} DebugIndexScannerCode;
+
+extern const char *DEBUG_INDEX_SCANNER_STATUS_STRS[DEBUG_INDEX_SCANNER_CODE_COUNT];
```
GuyAv46, line 138: 💯

src/spec.c
lerman25: Yes, unused, its in the comments.
Left for completion.

GuyAv46 (approved):
Very nice

alonre24 (approved):

lerman25:
\backport

PR #5677, opened 2025-02-24
fix(setup): Force wget to save the boost archive to the expected location

LukeMathWalker:
`wget`, by default, won't overwrite an existing file. It'll append a number suffix to the target path, e.g. `${BOOST_NAME}.tar.gz.2`.
This breaks the installation process if a previous download was interrupted halfway (e.g. via CTRL+C) leaving behind an incomplete archive.
By passing the -O flag, we opt out of this behaviour: the downloaded file will be at the expected path, overwriting what's already there if necessary.

### DIFF
--- a/.install/install_boost.sh
+++ b/.install/install_boost.sh
@@ -10,7 +10,7 @@ if [[ -d ${BOOST_DIR} ]]; then
     exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


kei-nan (approved):

Itzikvaknin (approved):

LukeMathWalker:
Closing in favour of https://github.com/RediSearch/RediSearch/pull/5692

PR #5693, opened 2025-02-26
feat: Add 'make compile_commands' to generate compile_commands.json file

LukeMathWalker:
Add a make target to generate the `compile_commands.json` file, which can then be provided as input to `clangd` to get LSP support for the project.
I marked `compile_commands.json` as `git`-ignored to avoid polluting the project by mistake.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,8 @@ make run           # run redis with RediSearch
   COORD=1|oss        # run cluster
   WITH_RLTEST=1      # run with RLTest wrapper
   GDB=1              # invoke using gdb
+make compile_commands # generate compile_commands.json for clangd
+
 
 make test          # run all tests
   REDIS_STANDALONE=1|0 # test with standalone/cluster Redis
@@ -355,6 +357,13 @@ endif
 endif
 
 .PHONY: run
+
+#----------------------------------------------------------------------------------------------
+compile_commands:
+		$(SHOW)$(MAKE) build CMAKE_ARGS="-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"
+		$(SHOW)ln -sf $(BINDIR)/compile_commands.json .
+
+.PHONY: compile_commands
 
 #----------------------------------------------------------------------------------------------
 
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ callgrind.out.*
 /deps/RedisJSON/
 /1/
 /.install/boost*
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -45,6 +45,8 @@ make run           # run redis with RediSearch
   COORD=1|oss        # run cluster
   WITH_RLTEST=1      # run with RLTest wrapper
   GDB=1              # invoke using gdb
+make compile_commands # generate compile_commands.json for clangd
```
kei-nan, line 48: What does LSP has to do with generating compilation commands?

Makefile
```diff
@@ -356,6 +358,13 @@ endif
 
 .PHONY: run
 
+#----------------------------------------------------------------------------------------------
+compile_commands:
+		$(SHOW)$(MAKE) build CMAKE_ARGS="-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"
+		$(SHOW)ln -sf $(BINDIR)/compile_commands.json .
```
oshadmi, line 364: Should we clean this link on `clean`?

Makefile
LukeMathWalker: `clangd` analyzes compilation commands to find where files are located, what compilation flags must be used, etc.
The relevant option in `clangd --help` is named `--compile-commands-dir`.

Makefile
LukeMathWalker: We should. Do you think it should be conditional to `ALL=1` having been provided or should it always be cleaned?

Makefile
LukeMathWalker: If you don't have `clangd` locally, you can look [here](https://clangd.llvm.org/config.html#compileflags).

LukeMathWalker:
An alternative approach we could take here: add a `COMPILE_COMMANDS` flag to `make build`, which (if set) will generate the JSON file into the artefact directory. 
No top-level symbolic link, it's up to the developer then to configure their `clangd` instance to look into `bin` for it.

Makefile
kei-nan: ALL==1 means the entire bindir is deleted - since it is generated there it will also get deleted.
Maybe it should be part of the CMakeLists.txt?
That way the clean command of the make cmake generated will clean it.

LukeMathWalker:
Closing in favour of https://github.com/RediSearch/RediSearch/pull/6099 after the build system restructuring.

PR #5694, opened 2025-02-26
build: Integrate Rust into the build system.

LukeMathWalker:
# Linkage

We start with an empty `trie_rs` crate, under `src/redisearch_rs`.
We build it as a static library and link the resulting *.a file into the final RediSearch shared object.
As the Rust workspace grows, we will create a unified entrypoint (a "facade" crate) for interfacing with the C code. It'll be compiled as a static library and linked against the C code, replacing the current `libtrie.a` entrypoint.

# Build artefacts

`bin/redisearch_rs` will be used as the Rust target directory for the project. It'll contain all intermediate and final build artefacts, for all build profiles (debug and optimized).

## Robustness

`src/redisearch_rs/.cargo/config` ensures that intermediate artefacts will always end up in `bin/redisearch_rs`, even if `cargo` commands are executed directly or by third-party tooling (e.g. language-server protocol implementations).

## Final artefacts

`bin/redisearch_rs` choice honors the existing convention of putting all build artefacts under `bin`. It doesn't follow, though, the existing `{platform}-{profile}/{dependency}` directory layout since it wouldn't be compatible with `cargo`'s expected target directory layout. 
We opt for a middleground: we copy to `{platform}-{profile}/redisearch_rs` the final artefacts (i.e. `libtrie.a`) which will be linked into the final shared object.


CMakeLists.txt
```diff
@@ -154,6 +155,9 @@ if (BUILD_STATIC)
     set_target_properties(redisearch PROPERTIES LINKER_LANGUAGE CXX)
     setup_shared_object_target(redisearch "")
     target_link_libraries(redisearch redisearch-geometry VectorSimilarity ${REDISEARCH_LIBS} ${CMAKE_LD_LIBS})
+   	target_link_libraries(redisearch
```
kei-nan, line 158: indentation

CMakeLists.txt
```diff
@@ -172,6 +176,9 @@ else() # OSS RediSearch
             ${SSL_LIBS}
             ${REDISEARCH_LIBS}
             ${CMAKE_LD_LIBS})
+   	target_link_libraries(redisearch
```
kei-nan, line 179: indentation

sbin/system-setup.py
```diff
@@ -18,6 +18,7 @@ def __init__(self, args):
 
     def common_first(self):
         self.install_downloaders()
+        self.run(f"{READIES}/bin/getrust")
```
kei-nan, line 21: I know we have a long-term plan on getting rid of readies, it might be better not to depend on it for rust.

src/redisearch_rs/.gitignore
```diff
@@ -0,0 +1,17 @@
+# Generated by Cargo
+# will have compiled files and executables
+/target/
```
kei-nan, line 3: do files get generated in the target folder?
Can we tell it to put build files in the bin folder?
That is where currently all our build artifacts go to.

kei-nan (changes requested):

Makefile
```diff
@@ -288,6 +297,26 @@ $(LIBUV):
...
+RUST_ARTIFACT_SUBDIR=debug
+else
+RUST_ARTIFACT_SUBDIR=release
+endif
+
+redisearch_rs:
+	@echo Building redisearch_rs..
+	$(SHOW)cd $(REDISEARCH_RS_DIR) && cargo build $(RUST_BUILD_MODE)
+	$(SHOW)mkdir -p $(REDISEARCH_RS_BINDIR)
+	$(SHOW)cp -r $(REDISEARCH_RS_DIR)/target/$(RUST_ARTIFACT_SUBDIR)/*.a $(REDISEARCH_RS_BINDIR)
```
MeirShpilraien, line 315: Maybe we can configure rust to build the artifacts directly to `REDISEARCH_RS_BINDIR`

src/redisearch_rs/.gitignore
```diff
@@ -0,0 +1,17 @@
...
+/redisgears_v8_plugin/target/
+/redisgears_core/target/
+/redisgears_plugin_api/target/
```
MeirShpilraien, line 6: Copy past I guess?

Makefile
```diff
@@ -253,8 +260,10 @@ include $(MK)/rules
 clean:
 ifeq ($(ALL),1)
 	$(SHOW)rm -rf $(BINROOT)
+	$(SHOW)cd $(REDISEARCH_RS_DIR) && cargo clean
```
MeirShpilraien, line 263: Maybe add a `redisearch_rs` clean rule? Which we can also call separately?

CMakeLists.txt
```diff
@@ -154,6 +155,9 @@ if (BUILD_STATIC)
     set_target_properties(redisearch PROPERTIES LINKER_LANGUAGE CXX)
     setup_shared_object_target(redisearch "")
     target_link_libraries(redisearch redisearch-geometry VectorSimilarity ${REDISEARCH_LIBS} ${CMAKE_LD_LIBS})
+   	target_link_libraries(redisearch
+  		debug ${root}/src/redisearch_rs/target/debug/libtrie_rs.a
+  		optimized ${root}/src/redisearch_rs/target/release/libtrie_rs.a)
```
MeirShpilraien, line 160: If you link to this path, why you also copy it to the main bin directory? Am I missing something?

MeirShpilraien (changes requested):
Maybe we can try to plan ahead a little.
For example:
1. Are we gone have a single static library or multiple? I guess that with a single static library we will need to change the cmake file once but then maybe we need to build differently on the rust side?
2. How does using redismodule-rs (in the future) is gone be added here (cause it is no longer gone be a static library).
3. Maybe we can refactor RediSearch build system to make it better than what it is currently, maybe there are things that we do not longer need (like build static)?
4. How do we plan to run the rust unit tests?

CMakeLists.txt
LukeMathWalker: Fixed ✅

CMakeLists.txt
LukeMathWalker: Fixed ✅

src/redisearch_rs/.gitignore
LukeMathWalker: Indeed, cleaned it up.

CMakeLists.txt
LukeMathWalker: I added the copy later on as a way to have all final artefacts in the "expected" location, but I forgot to change the linking directory.
I have amended it in [1068b0f](https://github.com/RediSearch/RediSearch/pull/5694/commits/1068b0f504dfa96408759c8e3ef239f77a5e85a4) to have all artefacts in the bin dir, and link from there.

Makefile
LukeMathWalker: Done in [1068b0f](https://github.com/RediSearch/RediSearch/pull/5694/commits/1068b0f504dfa96408759c8e3ef239f77a5e85a4)

src/redisearch_rs/.gitignore
LukeMathWalker: Done in [1068b0f](https://github.com/RediSearch/RediSearch/pull/5694/commits/1068b0f504dfa96408759c8e3ef239f77a5e85a4)

sbin/system-setup.py
LukeMathWalker: I see—what's the preferred way to replace readies at the moment?

LukeMathWalker:
On build artefacts: RediSearch's bin directory is currently structured as `{platform}`/`{dependency}`/`{profile}`.

`cargo` structures Rust's build directories a bit differently: `{platform}`/`{profile}`. The platform bit is optional if you're building for the same platform. The structure is detailed in [this reference](https://doc.rust-lang.org/cargo/reference/build-cache.html#build-cache).

This is not an issue with respect to  `make` integration (see https://github.com/RediSearch/RediSearch/pull/5694/commits/1068b0f504dfa96408759c8e3ef239f77a5e85a4), but it'll become an issue for IDE integration.
If you use an LSP for Rust (e.g. `rust-analyzer`), it'll look for a conventionally structured target directory. We can't point the LSP at the bin directory, since the structure is non-conformant.

I see two options:
1. Leave things as they are. When working on the Rust parts of the codebase, there'll be a duplicated target under `redisearch_rs`.
2. Change the way we organize Rust artefacts in the bin directory. Add a `redisearch_rs` folder directly under `bin`, holding all intermediate build artefacts for all profiles, and then copy just the final artefacts to the `{platform}`/`redisearch_rs`/`{profile}` folder, for linking.

With 2., we can then configure all Rust commands to use `bin/redisearch_rs` as their build directory, removing all duplication.

What do you think @MeirShpilraien @kei-nan?

sbin/system-setup.py
alonre24: We are already bringing rust for building JSON - see `.install/test_deps/install_rust.sh`

LukeMathWalker:
> 1. Are we gone have a single static library or multiple? I guess that with a single static library we will need to change the cmake file once but then maybe we need to build differently on the rust side?

Both options are viable.  
From the Rust perspective, it is preferrable to split the project into multiple crates—that's why `redisearch_rs` is already structured as a workspace, as in your original PR. We benefit, in particular, from better build performance since `cargo` uses the crate as its build unit.

Nonetheless, we can still expose a single library to the C side by having a facade Rust crate that re-exports the items we need for C from the other crates. We would compile that single Rust crate into a static library and the build process would pretty much remain as it currently stands in this PR (except for a target rename).

Alternatively, we could build multiple static libraries and link each of them independently. This would either translate into us having to touch the build system every time we add a new crate or integrating a more general solution that can do the heavy lifting for us (e.g. https://github.com/corrosion-rs/corrosion as @hdoordt suggested to me earlier today).

I prefer the first approach since it minimises build system complexity without any significant drawbacks. It may even end up to be faster, since we don't spend time producing N static libraries with significant overlaps (e.g. N standard libraries, etc.). 

> 2. How does using redismodule-rs (in the future) is gone be added here (cause it is no longer gone be a static library).

I planned to pull `redismodule-rs` in as a "regular" dependency of our Rust crates.
I'm not sure I follow the implications though—what would prevent us from building a static library once we pull in `redismodule-rs` as a dependency? It doesn't seem to force us to go down the dynamic library route, but I may be overlooking something.

> 3. Maybe we can refactor RediSearch build system to make it better than what it is currently, maybe there are things that we do not longer need (like build static)?

I don't think we have enough context at this stage to go after a deep refactor of the build system. If that's desirable, we can plan for it, but I wouldn't make it a blocker to basic Rust support.

> 4. How do we plan to run the rust unit tests?

I have added support for running Rust unit tests in https://github.com/RediSearch/RediSearch/pull/5694/commits/97e73588cd3c6e4fec88f33f74e39c65712621c7

src/redisearch_rs/.gitignore
kei-nan: If it goes now into the bin folder then can you remove /target/ from the git ignore file?

kei-nan:
> On build artefacts: RediSearch's bin directory is currently structured as `{platform}-{profile}/{dependency}`.
> 
> `cargo` structures Rust's build directories a bit differently: `target/{platform}/{profile}`. The platform bit is omitted if you're building for the host platform. The structure is detailed in [this reference](https://doc.rust-lang.org/cargo/reference/build-cache.html#build-cache). The `target` bit can be renamed, but it is expected that a single folder contains all artefacts.
> 
> This is not an issue with respect to `make` integration (see [1068b0f](https://github.com/RediSearch/RediSearch/commit/1068b0f504dfa96408759c8e3ef239f77a5e85a4)), but it'll become an issue for IDE integration. If you use an LSP for Rust (e.g. `rust-analyzer`), it'll look for a conventionally structured target directory. We can't point the LSP at the bin directory, since the structure is non-conformant.
> 
> I see two options:
> 
> 1. Leave things as they are. When working on the Rust parts of the codebase, there'll be a duplicated target under `redisearch_rs`.
> 2. Change the way we organize Rust artefacts in the bin directory. Add a `redisearch_rs` folder directly under `bin`, holding all intermediate build artefacts for all profiles, and then copy just the final artefacts to the `{platform}-{profile}/redisearch_rs` folder, for linking.
> 
> With 2., we can then configure all Rust commands to use `bin/redisearch_rs` as their build directory, removing all duplication.
> 
> What do you think @MeirShpilraien @kei-nan?

I would prefer option 2 out of the two, will defer to @MeirShpilraien

sbin/system-setup.py
LukeMathWalker: Should I invoke the script from `sbin/system-setup.py` or ensure it's invoked in each platform-specific installation script?
I'm not 100% sure about the relationship between these two installation "systems".

LukeMathWalker: All platform build is running [here](https://github.com/RediSearch/RediSearch/actions/runs/13634398389).

LukeMathWalker: Some issues with the `PATH` 🤔 Looking into it.

LukeMathWalker:
I've prototyped option 2. on a branch. You can see the diff [here](https://github.com/RediSearch/RediSearch/compare/the-return-of-the-trie...alternative-rust-artefact-layout?expand=1).

sbin/system-setup.py
LukeMathWalker: There seem to be some platform-specific quirks that must be sorted out.

DvirDukhan (approved):
We can proceed IMO
Following the long discussion, and for future sake, please add a short section explaining how the artifacts will be deployed and linked (e.g., folder structure, etc.) to the PR description.

LukeMathWalker:
I've updated the PR description and the commit message to reflect the end state of the PR.
In the end, I moved forward with approach 2. for the target directory cc @kei-nan.

Building succeeds on [all Linux platforms](https://github.com/RediSearch/RediSearch/actions/runs/13651690051) and [macOS](https://github.com/RediSearch/RediSearch/actions/runs/13651823123), but there are some issues with the [benchmark flow](https://github.com/RediSearch/RediSearch/actions/runs/13652016615). I'll troubleshoot the latter before asking for a final review ahead of merging.

LukeMathWalker:
The build step is now succeeding for benchmarks ([run](https://github.com/RediSearch/RediSearch/actions/runs/13653205913/job/38166320138)), but the flow itself fails with "connection refused" errors.
They don't seem to be related to this PR, since the same happens on other runs from the previous weeks (e.g. https://github.com/RediSearch/RediSearch/actions/runs/13284835295).

It'd be ideal if someone could confirm/review this further cc @alonre24 @kei-nan

kei-nan:
> The build step is now succeeding for benchmarks ([run](https://github.com/RediSearch/RediSearch/actions/runs/13653205913/job/38166320138)), but the flow itself fails with "connection refused" errors. They don't seem to be related to this PR, since the same happens on other runs from the previous weeks (e.g. https://github.com/RediSearch/RediSearch/actions/runs/13284835295).
> 
> It'd be ideal if someone could confirm/review this further cc @alonre24 @kei-nan

Yep, looks like benchmarks are broken.
Shouldn't block this PR
We will try and sort it out independent of this PR.

.install/test_deps/common_installations.sh
```diff
@@ -4,12 +4,17 @@ OS_TYPE=$(uname -s)
...
 
 activate_venv() {
-	echo "copy ativation script to shell config"
+	echo "copy activation script to shell config"
 	if [[ $OS_TYPE == Darwin ]]; then
 		echo "source venv/bin/activate" >> ~/.bashrc
 		echo "source venv/bin/activate" >> ~/.zshrc
 	else
 		echo "source $PWD/venv/bin/activate" >> ~/.bash_profile
+        # On platforms like Debian and Alpine this causes issues with
```
kei-nan, line 13: indentation.
Also not clear what "this" means in the sentance.
Do you mean appending the activation command into bash profile file?

LukeMathWalker: Fixed the indentation.
On the comment—yes, that's what I was referring to. I've reworded the comment to clarify it.

kei-nan (approved):
Looks good

kei-nan (approved):

PR #5714, opened 2025-03-02
MOD-8932: Add read all function to circular buffer

Itzikvaknin:
Add a ReadAll function to circularBuffer with the same concurrency restrictions as the standard read function.

### DIFF
--- a/src/util/circular_buffer.c
+++ b/src/util/circular_buffer.c
@@ -207,6 +207,46 @@ void *CircularBuffer_Read(CircularBuffer cb, void *item) {
   return read;
 }
 
+size_t CircularBuffer_ReadAll(CircularBuffer cb, void *dst, bool advance){
+  RedisModule_Assert(cb != NULL);
+
+  // make sure there's data to return
+  if (unlikely(CircularBuffer_Empty(cb))) {
+    return 0;
+  }
+
+  //calculate the offset_idx of the read pointer
+  size_t item_count = cb->item_count;
+  uint64_t write = cb->write;
+  uint idx = write / cb->item_size;
+  int read_idx = idx - item_count;
+
+  read_idx = (read_idx >= 0) ? read_idx : (cb->item_cap + read_idx);
+
+  size_t first_chunk = (read_idx + item_count <= cb->item_cap) ? item_count : (cb->item_cap - read_idx);
+  first_chunk *= cb->item_size;
+  size_t second_chunk = item_count - first_chunk;
+  second_chunk *= cb->item_size;
+
+  // copy item from buffer to output
+  void* read = cb->data + read_idx * cb->item_size;
+  if (dst != NULL) {
+    memcpy(dst, read, first_chunk);
+    if (second_chunk > 0) {
+      memcpy(dst + first_chunk, cb->data, second_chunk);
+    }
+  }
+
+  if (advance) {
+    // advance read position
+    // circle back if read reached the end of the buffer
+    cb->item_count = 0;
+    cb->read = cb->data + cb->write * cb->item_size;
+  }
+  // update buffer item count
+  return item_count;
+}
+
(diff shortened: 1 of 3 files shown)


src/util/circular_buffer.c
```diff
@@ -207,6 +207,49 @@ void *CircularBuffer_Read(CircularBuffer cb, void *item) {
...
+    if (second_chunk > 0) {
+      memcpy(dst + first_chunk, cb->data, second_chunk);
+    }
+  }
+
+  if (advance) {
+    // advance read position
+    // circle back if read reached the end of the buffer
+    cb->item_count = 0;
+    cb->read = cb->data + cb->write * cb->item_size;
```
benoitdion, line 247: is `cb->write` an index or an offset?

Itzikvaknin: offset

src/util/circular_buffer.c
```diff
@@ -207,6 +207,49 @@ void *CircularBuffer_Read(CircularBuffer cb, void *item) {
   return read;
 }
 
+// Read all items from buffer.
+// dst must be large enough to hold all items in the buffer.
+// This function is not thread-safe.
+// This function pops the oldest item from the buffer.
```
raz-mon, line 213: `advance` dependent
All the items, not just oldest

src/util/circular_buffer.c
```diff
@@ -207,6 +207,49 @@ void *CircularBuffer_Read(CircularBuffer cb, void *item) {
...
+
+  size_t first_chunk = (read_idx + item_count <= cb->item_cap) ? item_count : (cb->item_cap - read_idx);
+  size_t second_chunk = item_count - first_chunk;
+  first_chunk *= cb->item_size;
+  second_chunk *= cb->item_size;
+
+  // copy item from buffer to output
+  void* read = cb->data + read_idx * cb->item_size;
+  if (dst != NULL) {
+    memcpy(dst, read, first_chunk);
```
raz-mon, line 237: Note that this is a shallow copy - if we will have heap-allocated resources in our slow-log we will need either a deep copy or an ownership mechanism

src/util/circular_buffer.c
```diff
@@ -207,6 +207,49 @@ void *CircularBuffer_Read(CircularBuffer cb, void *item) {
...
+  read_idx = (read_idx >= 0) ? read_idx : (cb->item_cap + read_idx);
+
+  size_t first_chunk = (read_idx + item_count <= cb->item_cap) ? item_count : (cb->item_cap - read_idx);
+  size_t second_chunk = item_count - first_chunk;
+  first_chunk *= cb->item_size;
+  second_chunk *= cb->item_size;
+
+  // copy item from buffer to output
+  void* read = cb->data + read_idx * cb->item_size;
+  if (dst != NULL) {
```
raz-mon, line 236: Assert this in the beginning of the function

src/util/circular_buffer.c
```diff
@@ -207,6 +207,65 @@ void *CircularBuffer_Read(CircularBuffer cb, void *item) {
...
+size_t CircularBuffer_ReadAll(CircularBuffer cb, void *dst, bool advance){
+  RedisModule_Assert(cb != NULL);
+  RedisModule_Assert(dst != NULL);
+
+  if (unlikely(CircularBuffer_Empty(cb))) {
+    return 0;
+  }
+
+  // Calculate the starting position for reading
+  int64_t item_count = (int64_t)atomic_load(&cb->item_count);
```
kei-nan, line 231: `item_count` is size_t, consider changing its type to int64_t to properly show the limitations.
Otherwise the local item count you have here might be negative.

kei-nan: Looks like it was fixed, closing.

kei-nan (approved):

PR #5719, opened 2025-03-03
MOD-8993: Separate Index Error Messages Into Two Parts

kei-nan:
Part of the big obfuscation PR breakdown.
We would like to eventually output an obfuscated index info.
For that reason we need to separate index errors into two pars
first part is the general error message
second part is the more detailed error message that may contain user data
if we send the error to the user we would like to send out the more detailed error
if we output the error to the logs we might like to only output the general message part.

1. Current: The index error contain a single error message which might include user data
2. Change: Split the error into two, provide an API to either only fill out the first part or both.
3. Outcome: Allow us to conditionally determine at a later point what kind of error we would like to show.

1. Index Error

#### Mark if applicable


src/document_add.c
```diff
@@ -99,13 +99,12 @@ static int parseDocumentOptions(AddDocumentOptions *opts, ArgsCursor *ac, QueryE
...
       } else {
-        QueryError_SetErrorFmt(status, QUERY_EADDARGS, "Unknown keyword `%.*s` provided", (int)narg,
-                               s);
+        QueryError_SetWithUserDataFmt(status, QUERY_EADDARGS, "Unknown keyword", " `%.*s` provided", (int)narg, s);
         return REDISMODULE_ERR;
       }
       // Argument not found, that's ok. We'll handle it below
     } else {
-      QueryError_SetErrorFmt(status, QUERY_EADDARGS, "%s: %s", errArg->name, AC_Strerror(rv));
+      QueryError_SetWithUserDataFmt(status, QUERY_EADDARGS, errArg->name, ": %s", AC_Strerror(rv));
```
raz-mon, line 107: Not sure this error message is informative enough

src/info/index_error.c
```diff
@@ -54,22 +66,34 @@ void IndexError_AddError(IndexError *error, const char *error_message, RedisModu
 
 void IndexError_Clear(IndexError error) {
     if (!NA_rstr) initDefaultKey();
-    if (error.last_error != NA && error.last_error != NULL) {
-        rm_free(error.last_error);
-        error.last_error = NA;
+    if (error.last_error_without_user_data != NA && error.last_error_without_user_data != NULL) {
```
raz-mon, line 69: Are both possible here?
Watch out from hiding unreachable paths via multiple conditions

src/info/index_error.h
```diff
@@ -30,16 +35,27 @@ extern char* const IndexError_ObjectName;
 IndexError IndexError_Init();
 
 // Adds an error message to the IndexError. The error_count is incremented and the last_error is set to the error_message.
-void IndexError_AddError(IndexError *error, const char *error_message, RedisModuleString *key);
+void IndexError_AddError(IndexError *error, ConstErrorMessage withoutUserData, ConstErrorMessage withUserData, RedisModuleString *key);
+
+// Adds a query error to the index error using IndexError_AddError
+static inline void IndexError_AddQueryError(IndexError *error, const QueryError* queryError, RedisModuleString *key) {
```
raz-mon, line 41: Can you add a comment about the difference between this and `IndexError_AddError`?

src/query_error.c
```diff
@@ -77,6 +95,7 @@ void QueryError_SetErrorFmt(QueryError *status, QueryErrorCode code, const char
   rm_vasprintf(&status->detail, fmt, ap);
   va_end(ap);
   status->code = code;
+  status->message = status->detail;
```
raz-mon, line 98: Consider having one of the messages `NULL` in cases where they are the same.
This can also be used to understand if there was something to obfuscate if that will be interesting sometime..

raz-mon: Relevant also for `QueryError_SetError`

tests/cpptests/test_cpp_index_error.cpp
```diff
@@ -0,0 +1,30 @@
...
+  error.last_error_time = {0};
+  lastErrorKey = IndexError_LastErrorKeyObfuscated(&error);
+  text = RedisModule_StringPtrLen(lastErrorKey, NULL);
+  ASSERT_NE(key, lastErrorKey);
+  ASSERT_STREQ("Key@0", text);
+  RedisModule_FreeString(NULL, lastErrorKey);
+  IndexError_Clear(error);
+  RedisModule_FreeString(NULL, key);
+
+}
```
raz-mon, line 30: Newline

tests/pytests/test_debug_commands.py
```diff
@@ -392,7 +392,6 @@ def testHNSWdump_badParams(env: Env):
     # If index error is "Can't open vector index" then function tries to accsses null pointer
     env.expect(debug_cmd(), 'DUMP_HNSW', 'idx','v').error() \
         .contains("Can't open vector index")
-
```
raz-mon, line 395: Why?

src/info/index_error.c
kei-nan: I don't know, I'm maintaining the same logic as it was in the previous code.
Maybe when deserializing the info we would get null?
If you want, I can remove one of the conditions, but it feels a bit out of scope for this PR.

src/document_add.c
kei-nan: Noticed it doesn't even contain user data:
Changed to:
```cpp
QueryError_SetWithoutUserDataFmt(status, QUERY_EADDARGS, "Parsing error for document option %s: %s", errArg->name, AC_Strerror(rv));
```

oshadmi (commented):
Some initial comments

src/info/index_error.c
raz-mon: You're right, Forget about it.

src/query_error.c
kei-nan: It complicates the usage and makes things more implicit.
** Because something is null assume we can return the non-null field. **
I prefer this to be explicit as possible.

tests/pytests/test_debug_commands.py
kei-nan: will bring back the missing line - not sure why it was removed.

src/aggregate/aggregate_request.c
```diff
@@ -46,7 +45,7 @@ static bool ensureSimpleMode(AREQ *areq) {
  */
 static int ensureExtendedMode(AREQ *areq, const char *name, QueryError *status) {
   if (areq->reqflags & QEXEC_F_IS_SEARCH) {
-    QueryError_SetErrorFmt(status, QUERY_EINVAL,
+    QueryError_SetWithUserDataFmt(status, QUERY_EINVAL,
```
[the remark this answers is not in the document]
kei-nan: I guess it depends on who else will call `ensureExtendedMode`
As long as we don't get passed strings with user data then we should be okay.
Will change it to be without user data version

src/aggregate/aggregate_request.c
```diff
@@ -183,7 +182,7 @@ int parseValueFormat(uint32_t *flags, ArgsCursor *ac, QueryError *status) {
   } else if (!strcasecmp(format, "STRING")) {
     *flags &= ~QEXEC_FORMAT_EXPAND;
   } else {
-    QERR_MKBADARGS_FMT(status, "FORMAT %s is not supported", format);
+    QERR_MKBADARGS_FMT(status, "FORMAT", " %s is not supported", format);
```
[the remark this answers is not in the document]
kei-nan: All we know at this point is that format is not EXPAND nor is it STRING.
Who is to say what it contains, and if we can't say then there is the possibility of it containing user data.

src/aggregate/aggregate_request.c
```diff
@@ -522,23 +521,23 @@ static int parseQueryArgs(ArgsCursor *ac, AREQ *req, RSSearchOptions *searchOpts
     // See if this is one of our arguments which requires special handling
     if (AC_AdvanceIfMatch(ac, "SUMMARIZE")) {
       if(!ensureSimpleMode(req)) {
-        QERR_MKBADARGS_FMT(status, "SUMMARIZE is not supported on FT.AGGREGATE");
+        QERR_MKBADARGS(status, "SUMMARIZE is not supported on FT.AGGREGATE");
```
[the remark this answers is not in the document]
kei-nan: I don't understand what you are asking.
right now it will get expanded to:
`QueryError_SetError` function call

src/aggregate/aggregate_request.c
```diff
@@ -522,23 +521,23 @@ static int parseQueryArgs(ArgsCursor *ac, AREQ *req, RSSearchOptions *searchOpts
...
-        QERR_MKBADARGS_FMT(status, "Bad arguments for SUMMARIZE");
+        QERR_MKBADARGS(status, "Bad arguments for SUMMARIZE");
         return REDISMODULE_ERR;
       }
       req->reqflags |= QEXEC_F_SEND_HIGHLIGHT;
 
     } else if (AC_AdvanceIfMatch(ac, "HIGHLIGHT")) {
       if(!ensureSimpleMode(req)) {
-        QERR_MKBADARGS_FMT(status, "HIGHLIGHT is not supported on FT.AGGREGATE");
+        QERR_MKBADARGS(status, "HIGHLIGHT is not supported on FT.AGGREGATE");
```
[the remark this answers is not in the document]
kei-nan: There is no user data here, it will get translated to `QueryError_SetError`.
Not sure exactly what you are asking.

src/aggregate/aggregate_request.c
```diff
@@ -1631,10 +1630,10 @@ int AREQ_BuildPipeline(AREQ *req, QueryError *status) {
           if (AC_AdvanceIfMatch(&lstp->args, SPEC_AS_STR)) {
             int rv = AC_GetString(&lstp->args, &name, &name_len, 0);
             if (rv != AC_OK) {
-              QERR_MKBADARGS_FMT(status, "LOAD path AS name - must be accompanied with NAME");
+              QERR_MKBADARGS(status, "LOAD path AS name - must be accompanied with NAME");
```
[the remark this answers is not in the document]
kei-nan: There are no user supplied strings that get formatted in this error message.

src/aggregate/aggregate_request.c
```diff
@@ -1631,10 +1630,10 @@ int AREQ_BuildPipeline(AREQ *req, QueryError *status) {
...
               return REDISMODULE_ERR;
             } else if (!strcasecmp(name, SPEC_AS_STR)) {
-              QERR_MKBADARGS_FMT(status, "Alias for LOAD cannot be `AS`");
+              QERR_MKBADARGS(status, "Alias for LOAD cannot be `AS`");
```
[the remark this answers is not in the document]
kei-nan: There are no strings that get formatted into the error message.
So no user data can leak here.

raz-mon (approved):
🔥

src/aggregate/aggregate_request.c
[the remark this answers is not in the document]
oshadmi: Seems rather strict to me that an erroneous keyword would be considered user data, but I suppose we can go on the safe side and as long as not proven as none-user data, we will treat it as user data.

src/aggregate/aggregate_request.c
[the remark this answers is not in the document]
oshadmi: Right, I confused it with `QERR_MKBADARGS_FMT` which is calling
QueryError_Set**With**UserDataFmt

src/aggregate/aggregate_request.c
```diff
@@ -1229,7 +1230,7 @@ static ResultProcessor *buildGroupRP(PLN_GroupStep *gstp, RLookup *srclookup,
     if (!ff) {
       // No such reducer!
       Grouper_Free(grp);
-      QueryError_SetErrorFmt(err, QUERY_ENOREDUCER, "No such reducer: %s", pr->name);
+      QueryError_SetWithoutUserDataFmt(err, QUERY_ENOREDUCER, "No such reducer: %s", pr->name);
```
oshadmi, line 1233: An illegal reducer - why is it different from an illegal format, e.g., in
```QERR_MKBADARGS_FMT(status, "FORMAT", " %s is not supported", format);```
Which is calling  QueryError_Set**With**UserDataFmt

File: src/aggregate/reducer.c [hunk 14]
```diff
@@ -78,7 +78,7 @@ int ReducerOpts_GetKey(const ReducerOptions *options, const RLookupKey **out) {
     // We currently allow implicit loading only for known fields from the schema.
     // If we can't load keys, or the key we loaded is not in the schema, we fail.
     if (!options->loadKeys || !((*out)->flags & RLOOKUP_F_SCHEMASRC)) {
-      QueryError_SetErrorFmt(options->status, QUERY_ENOPROPKEY, "Property `%s` not present in document or pipeline", s);
+      QueryError_SetWithUserDataFmt(options->status, QUERY_ENOPROPKEY, "Property is not preset in document or pipeline", ": `%s`", s);
```
oshadmi, line 81:
```suggestion
      QueryError_SetWithUserDataFmt(options->status, QUERY_ENOPROPKEY, "Property is not present in document or pipeline", ": `%s`", s);
```

src/document.c
```diff
@@ -619,8 +618,8 @@ FIELD_PREPROCESSOR(vectorPreprocessor) {
   }
   if (fdata->vecLen != fs->vectorOpts.expBlobSize) {
 
-    QueryError_SetErrorFmt(status, QUERY_EBADATTR,
-                           "Could not add vector with blob size %zu (expected size %zu)", fdata->vecLen,
+    QueryError_SetWithUserDataFmt(status, QUERY_EBADATTR,
```
oshadmi, line 621: Is Blob size considered user data? Protecting user data in this context is about privacy and hiding sensitive data. Do we regard the length of some data as sensitive?

src/index.c
```diff
@@ -1092,7 +1092,7 @@ int NI_SkipTo_O(void *ctx, t_docId docId, RSIndexResult **hit) {
     // This doc-id was deleted
     return INDEXREAD_NOTFOUND;
   }
-  RS_LOG_ASSERT_FMT(nc->lastDocId == docId, "Expected docId to be %llu, got %llu", docId, nc->lastDocId);
+  RS_LOG_ASSERT_FMT(nc->lastDocId == docId, "Expected docId to be %zu, got %zu", docId, nc->lastDocId);
```
oshadmi, line 1095: Why change format from `%llu` to `%zu` - the type of `t_docId` is `uint64_t` not `size_t`

src/module.c
```diff
@@ -3082,7 +3082,7 @@ int DistAggregateCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
   IndexSpec *sp = StrongRef_Get(spec_ref);
   if (!sp) {
     // Reply with error
-    return RedisModule_ReplyWithErrorFormat(ctx, "%s: no such index", idx);
+    return RedisModule_ReplyWithErrorFormat(ctx, "No index exists with provided name %s", idx);
```
oshadmi, line 3085: Rephrasing? Why?
If rephrasing consider this
```suggestion
    return RedisModule_ReplyWithErrorFormat(ctx, "No index exists with the provided name %s", idx);
```

src/module.c
```diff
@@ -3399,7 +3399,7 @@ int DistSearchCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
```
oshadmi, line 3402: Rephrasing? Why?
If rephrasing see above suggestion.

src/query.c
```diff
@@ -1027,6 +1029,7 @@ static IndexIterator *Query_EvalUnionNode(QueryEvalCtx *q, QueryNode *qn) {
 
   // a union stage with one child is the same as the child, so we just return it
   if (QueryNode_NumChildren(qn) == 1) {
+    qn->children[0]->opts.fieldMask &= qn->opts.fieldMask;
```
oshadmi, line 1032: Should this fix be in a separate PR?
Can we test this line?

src/query_parser/v1/parser.c
```diff
@@ -1678,7 +1678,7 @@ static void yy_syntax_error(
 #define TOKEN yyminor
 /************ Begin %syntax_error code ****************************************/
 
-    QueryError_SetErrorFmt(ctx->status, QUERY_ESYNTAX,
+    QueryError_SetWithoutUserDataFmt(ctx->status, QUERY_ESYNTAX,
```
oshadmi, line 1681: Shouldn't an arbitrary user token be considered user data?

src/query_parser/v1/parser.y
```diff
@@ -32,7 +32,7 @@
 %name RSQueryParser_v1_
 
 %syntax_error {
-    QueryError_SetErrorFmt(ctx->status, QUERY_ESYNTAX,
+    QueryError_SetWithoutUserDataFmt(ctx->status, QUERY_ESYNTAX,
```
oshadmi, line 35: Shouldn't an arbitrary user token be considered user data?

src/query_parser/v2/parser.c
```diff
@@ -109,13 +109,13 @@ static void setup_trace(QueryParseCtx *ctx) {
 
 static void reportSyntaxError(QueryError *status, QueryToken* tok, const char *msg) {
   if (tok->type == QT_TERM || tok->type == QT_TERM_CASE) {
-    QueryError_SetErrorFmt(status, QUERY_ESYNTAX,
+    QueryError_SetWithoutUserDataFmt(status, QUERY_ESYNTAX,
```
oshadmi, line 112: Same question for other parsing error messages:
Shouldn't an arbitrary user token be considered user data?

src/spec.c
```diff
@@ -753,13 +753,13 @@ static int parseVectorField_flat(FieldSpec *fs, VecSimParams *params, ArgsCursor
...
     } else {
-      QERR_MKBADARGS_FMT(status, "Bad arguments for algorithm %s: %s", VECSIM_ALGORITHM_BF, AC_GetStringNC(ac, NULL));
+      QERR_MKBADARGS_FMT(status, "Bad arguments for algorithm", " %s: %s", VECSIM_ALGORITHM_BF, AC_GetStringNC(ac, NULL));
       return 0;
     }
     numParam++;
   }
   if (expNumParam > numParam) {
-    QERR_MKBADARGS_FMT(status, "Expected %d parameters but got %d", expNumParam * 2, numParam * 2);
+    QERR_MKBADARGS_FMT(status, "Expected", " %d parameters but got %d", expNumParam * 2, numParam * 2);
```
oshadmi, line 762: Why is this different than above
```
QERR_MKBADARGS_WITHOUT_USER_DATA_FMT(status, "Expected %d parameters but got %d", expNumParam * 2, numParam * 2);
```

src/spec.c
```diff
@@ -1158,14 +1157,14 @@ static int IndexSpec_AddFieldsInternal(IndexSpec *sp, StrongRef spec_ref, ArgsCu
...
-        QueryError_SetErrorFmt(status, QUERY_EBADOPTION,
+        QueryError_SetWithUserDataFmt(status, QUERY_EBADOPTION,
                                "Cannot set dynamic field to sortable - %s", fieldName);
         goto reset;
       }
 
       fs->sortIdx = sp->numSortableFields++;
       if (fs->sortIdx == -1) {
-        QueryError_SetErrorFmt(status, QUERY_ELIMIT, "Schema is limited to %d Sortable fields",
+        QueryError_SetWithoutUserDataFmt(status, QUERY_ELIMIT, "Schema is limited to %d Sortable fields",
```
oshadmi, line 1167: Here we use **Without** user data.
Why is it different than the following (**With** user data)
```
QueryError_SetWithUserDataFmt(status, QUERY_ELIMIT, "Schema is limited", " to %d TEXT fields",
```

src/vector_index.c
```diff
@@ -121,16 +121,16 @@ IndexIterator *NewVectorIterator(QueryEvalCtx *q, VectorQuery *vq, IndexIterator
     }
     case VECSIM_QT_RANGE: {
       if ((dim * VecSimType_sizeof(type)) != vq->range.vecLen) {
-        QueryError_SetErrorFmt(q->status, QUERY_EINVAL,
+        QueryError_SetWithoutUserDataFmt(q->status, QUERY_EINVAL,
```
oshadmi, line 124: Nitpick - should we be consistent with https://github.com/RediSearch/RediSearch/pull/5719#discussion_r1998550183
and consider actual and expected vector sizes as user data?

tests/pytests/test_aggregate.py
```diff
@@ -1314,7 +1314,7 @@ def test_aggregate_filter_on_missing_values():
     # Search for the documents with the indexed fields (sanity)
     # document doc1 has no value for num1, so we expect to receive the mentioned error
     (env.expect('FT.AGGREGATE', 'idx', '@tag:{val}', 'LOAD', '1', 'num1', 'FILTER', '@num1 > 2').error().
-     contains('num1: has no value, consider using EXISTS if applicable'))
+     contains('Could not lookup the value for a parameter name, consider using EXISTS if applicable for num1'))
```
oshadmi, line 1317: Nitpick:
Consider rephrasing? (fix also the code in `evalProperty` in expression.c)
Instead of
`Could not lookup `
consider
`Could not retrieve `
or
`Could not find `

tests/pytests/test_missing.py
```diff
@@ -60,7 +60,7 @@ def testMissingValidations():
     # Tests that we get an error in case of a user tries to use "ismissing(@field)"
     # when `field` is created with `NOINDEX` and `INDEXMISSING`
     env.expect('FT.CREATE', 'idx3', 'SCHEMA', 'f1', 'TAG', 'INDEXMISSING', 'NOINDEX').error().contains(
-        'Field `f1` cannot be defined with both `NOINDEX` and `INDEXMISSING`'
+        'Field cannot be defined with both `NOINDEX` and `INDEXMISSING`'
```
oshadmi, line 63: Shouldn't we get the user data as well here?

tests/pytests/test_issues.py
```diff
@@ -1307,7 +1307,7 @@ def test_mod_6783(env:Env):
 def test_mod_8589(env:Env):
   env.expect('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT', 'v', 'VECTOR', 'FLAT', '6', 'TYPE', 'FLOAT32', 'DIM', '2', 'DISTANCE_METRIC', 'L2').ok()
   env.cmd('HSET', 'doc1', 'v', '????????', 't', 'foo bar foo') # Max frequency is 2 (foo)
-  docinfo = to_dict(env.cmd(debug_cmd(), 'DOCINFO', 'idx', 'doc1'))
+  docinfo = to_dict(env.cmd(debug_cmd(), 'DOCINFO', 'idx', 'doc1', 'REVEAL'))
```
oshadmi, line 1310: Can you please explain this change?

tests/pytests/test_missing.py
kei-nan: We check the for the prefix of the error, the user data is in the end of the error message.
I'll add the suffix of the message as well, maybe even modify the message a bit.

src/module.c
kei-nan: Moving the user data to the end, I think it was a general request from Dvir.
Redis stores the first string from the error message and outputs it in the command stats.
We don't want the user data to appear in the command stats.

src/module.c
kei-nan: Redis takes the first string as the error code.
Don't want user data to be taken as an error code.
Will apply your suggestion.

src/module.c
oshadmi: I meant why rephrasing except for moving the user data to the end of the string?
could still be 
```
return RedisModule_ReplyWithErrorFormat(ctx, "No such index %s", idx);
```

src/query.c
kei-nan: Maybe that would be better.
Will open a different PR for this one.

src/query_parser/v1/parser.y
kei-nan: This is always in relation to a user query.
The problem is we don't separate from indexing errors and query errors.
Query errors go straight to the user and we don't store them.
However I think we can modify this to still indicate it holds user data.
Will make the change.

src/query_parser/v1/parser.c
kei-nan: Will change it

src/query_parser/v2/parser.c
kei-nan: The cost of making this a user data string would mean breaking the error format.
We can do it, but note that query errors currently go directly to the user.
So we would break the message format without a big gain.

Let me know if you still want me to modify the error message.
* it would mean moving the user data formatting to the end of the error message and rephrasing it.

src/spec.c
kei-nan: true, will change it.

src/spec.c
kei-nan: it shouldn't, will change it.

tests/pytests/test_issues.py
kei-nan: Shouldn't be in this PR, part of a different PR.
Will remove it.

tests/pytests/test_aggregate.py
kei-nan: went with:
Could not find

src/query.c
kei-nan: will remove from this PR

src/query_parser/v2/parser.c
oshadmi: Let's change the error message and align with messages containing user data.

src/query_parser/v2/parser.y
```diff
@@ -150,13 +150,13 @@ static void setup_trace(QueryParseCtx *ctx) {
 
 static void reportSyntaxError(QueryError *status, QueryToken* tok, const char *msg) {
   if (tok->type == QT_TERM || tok->type == QT_TERM_CASE) {
-    QueryError_SetErrorFmt(status, QUERY_ESYNTAX,
+    QueryError_SetWithoutUserDataFmt(status, QUERY_ESYNTAX,
```
oshadmi, line 153: Should also be considered user data?

oshadmi (approved):

oshadmi (approved):

PR #5722, opened 2025-03-05
MOD-8970: Define an FFI API for trie_rs corresponding to deps/triemap/triemap.h

hdoordt:

Implement an FFI module in Rust that is almost equivalent to `deps/triemap/triemap.h` insofar its functionality that is used outside the triemap dependency. The major difference between `ffi.rs` and `triemap.h` is the fact that `ffi.rs` exposes the defined structs that are used externally (`TrieMap`, `TrieMapIterator`) only behind opaque pointers, so that using code does not rely on their internals.

After analysis of the code base, it has been concluded that the only occurrences of code altering or reading the fields of these structs is upon initialization of the `TrieMapIterator`. Therefore, the `TrieMap_Iterate` function takes an additional parameter of type `tm_iter_mode`, which is meant to allow for setting the iteration mode upon initialization if `TrieMapIterator`.

A further difference between the FFI API in `triemap_rs` and `triemap.h`, is the introduction of the opaque type `TrieMapResultBuf`, accompanying functions `TrieMapResultBuf_Free` and `TrieMapResultBuf_Data`, and `TrieMap_FindPrefixes` taking an argument of type `*mut TrieMapResultBuf` to store its results. Introduction of the opaque `TrieMapResultBuf` ensures the Rust implementation is responsible for both allocating an freeing the result buffer memory, whereas `triemap.h` fully defers the responsibility of freeing the data to the caller.

The FFI module is designed to be a more-or-less drop-in replacement for `triemap.h` and wraps the trie map implementation that will follow in a later PR.

The code in ffi.rs is validated by running [cbindgen](https://github.com/mozilla/cbindgen) against it,
and validating the resulting header file is equivalent to triemap.h.
Example command:
```
cd src/redisearch_rs/trie_rs && cbindgen --output triemap_rs.h --lang C
```

To do:

- [x] Document and check invariants using debug assertions 
- [x] Add missing docs to items that weren't documented in `triemap.h`
- [x] Describe effects of includeMin and includeMax on behavior of `TrieMap_IterateRange`
- [x] Describe behaviour of `TrieMap_IterateRange` if `minlen` or `maxlen` is 0 or even negative

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/Cargo.toml
+++ b/src/redisearch_rs/Cargo.toml
@@ -29,6 +29,9 @@ version = "0.0.1"
 edition = "2024"
 license = "Redis Source Available License 2.0 (RSALv2) or the Server Side Public License v1 (SSPLv1)"
 
+[workspace.dependencies]
+libc = "0.2.170"
+
 # A profile for fast test execution that doesn't sacrifice
 # runtime checks and debuggability.
 [profile.optimised_test]
--- a/src/redisearch_rs/trie_rs/src/ffi.rs
+++ b/src/redisearch_rs/trie_rs/src/ffi.rs
@@ -18,6 +18,30 @@ pub type tm_len_t = u16;
 #[used]
 pub static mut TRIEMAP_NOTFOUND: *mut ::std::os::raw::c_void = c"NOT FOUND".as_ptr() as *mut _;
 
+/// Used by TrieMapIterator to determine type of query.
+///
+/// C equivalent:
+/// ```c
+/// typedef enum {
+///     TM_PREFIX_MODE = 0,
+///     TM_CONTAINS_MODE = 1,
+///     TM_SUFFIX_MODE = 2,
+///     TM_WILDCARD_MODE = 3,
+///     TM_WILDCARD_FIXED_LEN_MODE = 4,
+///   } tm_iter_mode;
+#[repr(C)]
+enum tm_iter_mode {
+    TM_PREFIX_MODE = 0,
+    TM_CONTAINS_MODE = 1,
+    TM_SUFFIX_MODE = 2,
+    TM_WILDCARD_MODE = 3,
+    TM_WILDCARD_FIXED_LEN_MODE = 4,
+}
+
+/// Default mode for TrieMapIterator
+#[unsafe(no_mangle)]
+static TM_ITER_MODE_DEFAULT: tm_iter_mode = tm_iter_mode::TM_PREFIX_MODE;
+
 /// Opaque type TrieMap
 pub enum TrieMap {}
 
@@ -116,14 +140,14 @@ unsafe extern "C" fn TrieMap_Find(
     str: *const c_char,
     len: tm_len_t,
 ) -> *mut c_void {
-    unsafe { TRIEMAP_NOTFOUND }
+    todo!()
 }
 
 /// Find nodes that have a given prefix. Results are placed in an array.
 ///
 /// C equivalent:
 /// ```c
-/// int TrieMap_FindPrefixes(TrieMap *t, const char *str, tm_len_t len, arrayof(void*) *results); // arrayof(void*) == void**
+/// int TrieMap_FindPrefixes(TrieMap *t, const char *str, tm_len_t len, void** *results); // arrayof(void*)
 /// ```
 #[unsafe(no_mangle)]
 unsafe extern "C" fn TrieMap_FindPrefixes(
@@ -188,6 +212,7 @@ unsafe extern "C" fn TrieMap_Iterate(
     t: *mut TrieMap,
     prefix: *const c_char,
     prefix_len: tm_len_t,
+    iter_mode: tm_iter_mode,
 ) -> *mut TrieMapIterator {
     todo!()
 }
--- a/src/redisearch_rs/trie_rs/src/ffi.rs
+++ b/src/redisearch_rs/trie_rs/src/ffi.rs
@@ -1,0 +1,302 @@
+#![allow(non_camel_case_types)]
+
+use std::ffi::{c_char, c_int, c_void};
+
+/// C equivalent:
+/// ```c
+/// typedef uint16_t tm_len_t;
+/// ```
+pub type tm_len_t = u16;
+
+/// This special pointer is returned when TrieMap_Find cannot find anything.
+///
+/// C equivalent:
+/// ```c
+/// void *TRIEMAP_NOTFOUND = "NOT FOUND";
+/// ```
+#[unsafe(no_mangle)]
+#[used]
+pub static mut TRIEMAP_NOTFOUND: *mut ::std::os::raw::c_void = c"NOT FOUND".as_ptr() as *mut _;
+
+/// Opaque type TrieMap
+pub enum TrieMap {}
+
+/// C equivalent:
+/// ```c
+/// typedef void(TrieMapRangeCallback)(const char *, size_t, void *, void *);
+/// ```
+type TrieMapRangeCallback =
+    Option<unsafe extern "C" fn(*const c_char, libc::size_t, *mut c_void, *mut c_void)>;
+
+/// Type of the functions [`TrieMapIterator_Next`], [`TrieMapIterator_NextContains`],
+/// and [`TrieMapIterator_NextWildcard`].
+///
+///  C equivalent:
+/// ```c
+/// typedef int (*TrieMapIterator_NextFunc)(TrieMapIterator *it, char **ptr, tm_len_t *len, void **value);
+/// ```
+#[expect(dead_code)]
+type TrieMapIterator_NextFunc = Option<
+    unsafe extern "C" fn(
+        it: *mut TrieMapIterator,
+        ptr: *mut *mut c_char,
+        len: *mut tm_len_t,
+        value: *mut *mut c_void,
+    ) -> c_int,
+>;
+
+/// Opaque type TrieMapIterator
+enum TrieMapIterator {}
+
+/// Create a new [`TrieMap`]. Returns an opaque pointer to the newly created trie.
+///
+/// C equivalent:
+/// ```c
+/// TrieMap *NewTrieMap();
+/// ```
+#[unsafe(no_mangle)]
+unsafe extern "C" fn NewTrieMap() -> *mut TrieMap {
+    todo!()
+}
+
+/// C equivalent:
+/// ```c
+/// typedef void (*freeCB)(void *);
+/// ```
+type freeCB = Option<unsafe extern "C" fn(*mut c_void)>;
+
+/// C equivalent:
+/// ```c
+/// typedef void *(*TrieMapReplaceFunc)(void *oldval, void *newval);
+/// ```
+type TrieMapReplaceFunc =
+    Option<unsafe extern "C" fn(oldval: *mut c_void, newval: *mut c_void) -> *mut c_void>;
+
+/// Add a new string to a trie. Returns 1 if the key is new to the trie or 0 if
+/// it already existed.
+///
+/// If value is given, it is saved as a payload inside the trie node.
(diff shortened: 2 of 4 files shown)


src/redisearch_rs/trie_rs/src/ffi.rs
```diff
@@ -0,0 +1,553 @@
...
+    TM_WILDCARD_MODE = 3,
+    TM_WILDCARD_FIXED_LEN_MODE = 4,
+}
+
+/// Default iteration mode for [`TrieMap_Iterate`].
+#[unsafe(no_mangle)]
+static TM_ITER_MODE_DEFAULT: tm_iter_mode = tm_iter_mode::TM_PREFIX_MODE;
+
+/// Opaque type TrieMap. Can be instantiated with [`NewTrieMap`].
+pub enum TrieMap {}
```
kei-nan, line 49: why is it an enum and not a struct?

hdoordt: The difference between an empty struct and an empty enum, is that the empty enum cannot be instantiated. `*mut TrieMap` represents an opaque pointer to a struct that is to be defined later (in MOD-8972), and it would make no sense to create instances of `TrieMap`, hence we disallow doing so.

src/redisearch_rs/Cargo.toml
```diff
@@ -6,3 +6,6 @@ resolver = "3"
 version = "0.0.1"
 edition = "2024"
 license = "Redis Source Available License 2.0 (RSALv2) or the Server Side Public License v1 (SSPLv1)"
+
+[workspace.dependencies]
+libc = "0.2.170"
```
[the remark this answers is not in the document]
hdoordt: The `libc` crate does not actually pull in libc, it merely provides bindings and an idiomatic API to the system's `libc`. As long as the system libc doesn't introduce any non-backwards-compatible/breaking changes, and as long the `libc` crate is tested for the [targets](https://github.com/rust-lang/libc/blob/HEAD/ci/verify-build.sh) we're building for, our crates will still build and run fine.

src/redisearch_rs/trie_rs/src/ffi.rs
```diff
@@ -0,0 +1,553 @@
...
+    Option<unsafe extern "C" fn(*const c_char, libc::size_t, *mut c_void, *mut c_void)>;
+
+/// Type of the functions [`TrieMapIterator_Next`], [`TrieMapIterator_NextContains`],
+/// and [`TrieMapIterator_NextWildcard`].
+///
+///  C equivalent:
+/// ```c
+/// typedef int (*TrieMapIterator_NextFunc)(TrieMapIterator *it, char **ptr, tm_len_t *len, void **value);
+/// ```
+#[expect(dead_code)]
```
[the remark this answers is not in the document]
hdoordt: The reason is that rust can't see that `TrieMapIterator_NextFunc` is actually just a typedef to replicate the signatures of `TrieMapIterator_Next` `TrieMapIterator_NextContains` and `TrieMapIterator_NextWildcard`, and that it's supposed to be used by foreign code. The `TrieMapIterator_NextFunc` is not being used as a parameter to any of the exposed functions either, so Rust will regard is as dead code.

`TrieMapRangeCallback` _is_ being used as a parameter for `TrieMap_IterateRange`, and is therefore not considered dead code

src/redisearch_rs/trie_rs/src/ffi.rs
LukeMathWalker: We could achieve the same result with 

```rust
#[non_exhaustive]
pub struct TrieMap {}
```

but it's just a matter of style at that point.

src/redisearch_rs/Cargo.toml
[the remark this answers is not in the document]
hdoordt: The `libc` crate does not actually pull in libc, it merely provides bindings and an idiomatic API to the system's `libc`. As long as the system libc doesn't introduce any non-backwards-compatible/breaking changes, and as long the `libc` crate is tested for the [targets](https://github.com/rust-lang/libc/blob/HEAD/ci/verify-build.sh) we're building for, our crates will still build and run fine.

Note that the actual libc version being used by Rust is determined by the version of `std`.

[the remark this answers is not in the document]
kei-nan: We have a lot of environments we need to support; I am a bit concerned about finding the right libc version that will work for all environments.

We usually support targets longer than their EOL - so concerned about libc crate stopping supporting an OS we would still need to support.

Why do we need to specify an explicit version?

src/redisearch_rs/trie_rs/src/ffi.rs
kei-nan: This feels like very implicit/hidden behaviour we are relying on.
I think using the `non_exhaustive` would be more explicit.

src/redisearch_rs/Cargo.toml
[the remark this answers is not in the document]
hdoordt: This is certainly a concern to take into account. However, if we want guarantee support for EOL targets, that may mean pinning the Rust compiler (and therefore the version of `std`) to a specific version. I propose we open a separate PR for that.

The `libc` crate may or may not actually build with the version that we pin, in which case we can downgrade the `libc` crate version. `libc` crate versions are guaranteed to build with certain versions of Rust. If the `libc` crate version builds with your Rust compiler version, then that means the version of the `libc` crate is compatible with the target's `libc`.

Again, confusingly, the version of the `libc` crate we're depending on, does not determine what version of C `libc` we support. The Rust compiler version does.

src/redisearch_rs/trie_rs/src/ffi.rs
hdoordt: This isn't the hill I'm willing to die on, but the struct Luca proposes can [actually be instantiated](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=25d16f39397d877b081a8420338702e3), be it just in the same module. So the empty `enum` is actually more clear in my view.

Furthermore, the empty `enum` is a fairly common pattern and [described well in the Nomicon](https://doc.rust-lang.org/nomicon/exotic-sizes.html#empty-types).

Again, if this isn't convincing, I'll happily swap it out for the non-exhaustive `struct`.

hdoordt: This isn't the hill I'm willing to die on, but the struct Luca proposes can [actually be instantiated](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=25d16f39397d877b081a8420338702e3), be it just in the same module. So the empty `enum` is actually more clear in my view.

Furthermore, the empty `enum` is a fairly common pattern and [described well in the Nomicon](https://doc.rust-lang.org/nomicon/exotic-sizes.html#empty-types). (Note that as the nomicon describes, it's imperative that we don't actually dereference a pointer to the empty enum without casting it to the actual struct first).

Again, if this isn't convincing, I'll happily swap it out for the non-exhaustive `struct`.

src/redisearch_rs/trie_rs/src/ffi.rs
```diff
@@ -0,0 +1,553 @@
...
+
+/// This special pointer is returned when TrieMap_Find cannot find anything.
+///
+/// C equivalent:
+/// ```c
+/// void *TRIEMAP_NOTFOUND = "NOT FOUND";
+/// ```
+#[unsafe(no_mangle)]
+#[used]
+pub static mut TRIEMAP_NOTFOUND: *mut ::std::os::raw::c_void = c"NOT FOUND".as_ptr() as *mut _;
```
raz-mon, line 21: Why do we need the mutibility here?

src/redisearch_rs/trie_rs/src/ffi.rs
```diff
@@ -0,0 +1,553 @@
...
+/// ```c
+/// void *TrieMap_Find(TrieMap *t, const char *str, tm_len_t len);
+/// ```
+#[unsafe(no_mangle)]
+unsafe extern "C" fn TrieMap_Find(
+    t: *mut TrieMap,
+    str: *const c_char,
+    len: tm_len_t,
+) -> *mut c_void {
+    debug_assert!(t.is_null(), "t cannot be NULL");
```
raz-mon, line 214:
```suggestion
    debug_assert!(!t.is_null(), "t cannot be NULL");
```

src/redisearch_rs/trie_rs/src/ffi.rs
```diff
@@ -0,0 +1,553 @@
...
+    unsafe extern "C" fn(
+        it: *mut TrieMapIterator,
+        ptr: *mut *mut c_char,
+        len: *mut tm_len_t,
+        value: *mut *mut c_void,
+    ) -> c_int,
+>;
+
+/// Opaque type TrieMapIterator. Obtained from calling [`TrieMap_Iterate`].
+enum TrieMapIterator {}
```
raz-mon, line 78: Needs `#[repr(C)]` as well?

src/redisearch_rs/trie_rs/src/ffi.rs
```diff
@@ -0,0 +1,553 @@
...
+
+/// Opaque type TrieMapResultBuf. Holds the results of [`TrieMap_FindPrefixes`].
+enum TrieMapResultBuf {}
```
raz-mon, line 81: Needs `#[repr(C)]` as well?

src/redisearch_rs/trie_rs/src/ffi.rs
hdoordt: That [wouldn't compile](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=6a717007222b03a1ac575e4039fe5353) though. Also, we just need `* mut TrieMapIterator` to be FFI-safe, not `TrieMapIterator` itself. But the layout of pointers _is_ defined. So in this case, no need.

src/redisearch_rs/trie_rs/src/ffi.rs
hdoordt: See https://github.com/RediSearch/RediSearch/pull/5722#discussion_r1991838372

raz-mon (dismissed):

src/redisearch_rs/trie_rs/src/ffi.rs
LukeMathWalker: Re-reading this conversation, there may be a misunderstanding at play.

On the Rust side, we'll have a proper `struct` that implements the required functionality. But that `struct` will have private fields since we have to uphold a variety of invariants, and we want to be able to change the internals without affecting consumers, including C consumers.

Hence, a "dummy" opaque type is needed. We'll cast to/from it whenever we cross the boundary, thus forcing all C code to use the methods exposed in this FFI file rather than peeking at fields or internal details.
This is a common (even recommended) pattern when doing C FFI in Rust—see the [relevant Rustonomicon section](https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs).

kei-nan (approved):
Approving to not block pending PRs.
Not a fan of the use of enum to illustrate a "void" type.
But don't want to block this PR too much.

src/redisearch_rs/trie_rs/src/ffi.rs
hdoordt: Yes, we intend to change the enum definition with

```rust
/// Opaque type TrieMap. Can be instantiated with [`NewTrieMap`].
#[repr(transparent)]
struct TrieMap(crate::trie::TrieMap<*mut c_void>);
```

where `crate::trie::TrieMap` is the actual Rust triemap type.

LukeMathWalker (approved):

LukeMathWalker (approved):

PR #5735, opened 2025-03-09
Support hide user data from logs config parameter

kei-nan:
We need to support the hide user data from logs redis config param
When it is true we should hide user data when writing logs
This is an initial support to integrate it with the index error changes.

1. Current: We always have user data in our log files
2. Change: Monitor the redis hide-user-data-from-logs config param and use it to determine what to output when writing to the logs
3. Outcome: We will have a reliable way to know when we should hide user data from the logs.

1. Configuration

#### Mark if applicable


tests/pytests/test_debug_commands.py
```diff
@@ -865,3 +866,17 @@ def testSearchResp2(self):
         def listResults(res):
             return [{res[i]: res[i + 1]} for i in range(1, len(res[1:]), 2)]
         self.Resp2("SEARCH", ['SORTBY', 'n'], listResults)
+
+# For now allowing access to the value through the debug command
+# Maybe in the future it should be accessible through the FT.CONFIG command and the test move to test_config.py
+# Didn't want to "break" the API by adding a new config parameter
+def test_hideUserDataFromLogs(env):
```
nafraf, line 873: Do we need to check the minimum redis version which support `hide-user-data-from-log` ?

kei-nan: Looks like it was added in 7.4:
https://github.com/redis/redis/releases/tag/7.4.0
For now, will remove the backport label for 2.6 until we get some answers from redis folks

src/notifications.c
```diff
@@ -341,20 +376,26 @@ void Initialize_KeyspaceNotifications(RedisModuleCtx *ctx) {
     REDISMODULE_NOTIFY_LOADED | REDISMODULE_NOTIFY_MODULE,
     HashNotificationCallback);
 
+  // RedisModule_SubscribeToServerEvent should exist since redis 6.0
+  // We can assume it is always present
+
```
raz-mon, line 381: I think we can remove this comment

src/debug_commands.c
```diff
@@ -1643,6 +1655,7 @@ DebugCommandType commands[] = {{"DUMP_INVIDX", DumpInvertedIndex}, // Print all
                                {"SET_MONITOR_EXPIRATION", setMonitorExpiration},
                                {"WORKERS", WorkerThreadsSwitch},
                                {"BG_SCAN_CONTROLLER", bgScanController},
+                               {"GET_HIDE_USER_DATA_FROM_LOGS", getHideUserDataFromLogs},
```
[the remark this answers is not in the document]
kei-nan: it is just for testing.
I can make the registration of the command only happen in tests if that helps.
It will complicate the debug command registration a bit but maybe long term it will be a good thing to have control over.

src/notifications.c
kei-nan:
```suggestion
```

src/debug_commands.c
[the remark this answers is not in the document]
raz-mon: My bad, thought this was a config.

raz-mon (approved):

nafraf (dismissed):

PR #5736, opened 2025-03-09
[MOD-9023] Replacing freeFlag with a reference counter

meiravgri:
This PR fixes a race condition on `freeFlag` in the coordinator mechanism. 
The issue occurred because `freeFlag` was expected to be reset by the uv thread when the coordinator reached `MRIterator_Release` but since the flag is reset from the a BG thread, this was not always the case. 
This led to an incorrect sequence of operations, ultimately causing an assertion failure.
To resolve this, freeFlag has been **replaced with a reference counter**, which:

- **Increases** before pushing a new job to the RQ.
- **Decreases** when the coordinator no longer requires more replies.
- **Decreases** again when all shards have finished pushing replies to the channel.

### Logging improvements:
* Added `RS_DEBUG_LOG_FMT` and `RS_DEBUG_LOG` macros for debug logging in `deps/rmutil/rm_assert.h`.
* when `ENABLE_ASSERT=1` it will call `RedisModule_Log` with "debug" level

### DIFF
--- a/deps/rmutil/rm_assert.h
+++ b/deps/rmutil/rm_assert.h
@@ -19,8 +19,12 @@
 
 #ifndef ENABLE_ASSERT
 #define RS_LOG_ASSERT_FMT(condition, fmt, ...) // NOP
+#define RS_DEBUG_LOG_FMT(fmt, ...) // NOP
+#define RS_DEBUG_LOG(fmt) // NOP
 #else
 #define RS_LOG_ASSERT_FMT(condition, fmt, ...) _RS_LOG_ASSERT_FMT(condition, fmt, __VA_ARGS__)
+#define RS_DEBUG_LOG_FMT(fmt, ...) RedisModule_Log(RSDummyContext, "debug", (fmt), __VA_ARGS__)
+#define RS_DEBUG_LOG(fmt) RedisModule_Log(RSDummyContext, "debug", (fmt))
 #endif
 
 #define RS_LOG_ASSERT(condition, str) RS_LOG_ASSERT_FMT(condition, str "%s", "")
@@ -36,6 +40,6 @@
 #define RS_CHECK_FUNC(funcName, ...)                                          \
     if (funcName) {                                                           \
         funcName(__VA_ARGS__);                                                \
-    } 
+    }
 
 #endif  //__REDISEARCH_ASSERT__
--- a/deps/rmutil/rm_assert.h
+++ b/deps/rmutil/rm_assert.h
@@ -20,11 +20,11 @@
 #ifndef ENABLE_ASSERT
 #define RS_LOG_ASSERT_FMT(condition, fmt, ...) // NOP
 #define RS_DEBUG_LOG_FMT(fmt, ...) // NOP
-#define RS_DEBUG_LOG(fmt) // NOP
+#define RS_DEBUG_LOG(str) // NOP
 #else
 #define RS_LOG_ASSERT_FMT(condition, fmt, ...) _RS_LOG_ASSERT_FMT(condition, fmt, __VA_ARGS__)
 #define RS_DEBUG_LOG_FMT(fmt, ...) RedisModule_Log(RSDummyContext, "debug", (fmt), __VA_ARGS__)
-#define RS_DEBUG_LOG(fmt) RedisModule_Log(RSDummyContext, "debug", (fmt))
+#define RS_DEBUG_LOG(str) RedisModule_Log(RSDummyContext, "debug", (str))
 #endif
 
 #define RS_LOG_ASSERT(condition, str) RS_LOG_ASSERT_FMT(condition, str "%s", "")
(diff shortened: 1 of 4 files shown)


src/coord/dist_aggregate.c
```diff
@@ -44,7 +44,7 @@ static bool getCursorCommand(long long cursorId, MRCommand *cmd, MRIteratorCtx *
   // command instead of a READ command (here we know it has more results)
   if (timedout && !cmd->forCursor) {
     newCmd = MR_NewCommand(4, "_FT.CURSOR", "DEL", idx, buf);
-    newCmd.depleted = true;
+    // newCmd.depleted = true;
```
alonre24, line 47: why commenting out?

src/coord/rmr/rmr.c
```diff
@@ -508,9 +508,12 @@ void MRIteratorCallback_ResetTimedOut(MRIteratorCtx *ctx) {
 
 void MRIteratorCallback_Done(MRIteratorCallbackCtx *ctx, int error) {
   // Mark the command of the context as depleted (so we won't send another command to the shard)
+  RS_ASSERT(!ctx->cmd.depleted);
   ctx->cmd.depleted = true;
   short pending = --ctx->it->ctx.pending; // Decrease `pending` before decreasing `inProcess`
-  RS_LOG_ASSERT(pending >= 0, "Pending should not reach a negative value");
+
+  RS_LOG_ASSERT_FMT(pending >= 0, "Pending (%d) should not reach a negative value. inproc: %d, freeflag: %d, channel size: %d, shard_slot: %d",
+  pending, ctx->it->ctx.inProcess, ctx->it->ctx.freeFlag, MRChannel_Size(ctx->it->ctx.chan), ctx->cmd.targetSlot);
```
alonre24, line 516: Can we always print it in debug mode rather than only when the assertion fails? perhaps it can bring more information when time comes.

src/coord/dist_aggregate.c
meiravgri: This is wrong. Here the cmd is essentially not depleted as we are going to send the shard another request...

src/coord/rmr/rmr.c
```diff
@@ -442,7 +447,9 @@ struct MRIteratorCtx {
   short pending;    // Number of shards with more results (not depleted)
   short inProcess;  // Number of currently running commands on shards
   bool timedOut;    // whether the coordinator experienced a timeout
-  bool freeFlag;    // whether the MRIterator should be freed
+  // reference counter of the iterator.
+  // When it reaches 0, both readers and the writer agree that the iterator can be released
+  int8_t itRefCount;
```
alonre24, line 452: Using int8 here implies that we assume that there will be no more than 128 shards. Not sure it is a valid assumption

meiravgri: There shouldn't be more than 3

GuyAv46 (approved):
What a catch and what a solution 🎣 🥇

raz-mon (approved):

src/coord/dist_aggregate.c
```diff
@@ -44,7 +44,6 @@ static bool getCursorCommand(long long cursorId, MRCommand *cmd, MRIteratorCtx *
   // command instead of a READ command (here we know it has more results)
   if (timedout && !cmd->forCursor) {
     newCmd = MR_NewCommand(4, "_FT.CURSOR", "DEL", idx, buf);
-    newCmd.depleted = true;
```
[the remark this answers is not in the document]
meiravgri: it surely will send a command to this shard
the command will be marked as depleted when the shard returns its reply to the DEL command

deps/rmutil/rm_assert.h
```diff
@@ -19,8 +19,12 @@
 
 #ifndef ENABLE_ASSERT
 #define RS_LOG_ASSERT_FMT(condition, fmt, ...) // NOP
+#define RS_DEBUG_LOG_FMT(fmt, ...) // NOP
+#define RS_DEBUG_LOG(str) // NOP
```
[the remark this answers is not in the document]
meiravgri: Their purpose is to be used in a testing env, not in production (even if it's only enabled with debug loglevel)

src/coord/dist_aggregate.c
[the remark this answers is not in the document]
raz-mon: But yeah should be ok

deps/rmutil/rm_assert.h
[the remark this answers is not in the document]
meiravgri: We don't know the impact on performance of calling `RedisModule_log` (even if it's not the configured log level, it still executes some code), so i prefer not to compile it at all.
We can open a ticket to measure that

meiravgri:
> Backport failed for `8.0`, because it was unable to cherry-pick the commit(s).
> 
> Please cherry-pick the changes locally and resolve any conflicts.
> 
> ```shell
> git fetch origin 8.0
> git worktree add -d .worktree/backport-5736-to-8.0 origin/8.0
> cd .worktree/backport-5736-to-8.0
> git switch --create backport-5736-to-8.0
> git cherry-pick -x 3420e93a87a6fbd4b8880075b557cf8e5d4a4ab0
> ```

Already backported

PR #5738, opened 2025-03-10
[8.0] MOD-8839: Non-production assertions (#5667)

raz-mon:
Backport #5667 to `8.0`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ include deps/readies/mk/main
 
 define HELPTEXT
 make setup         # install prerequisited (CAUTION: THIS WILL MODIFY YOUR SYSTEM)
-make fetch         # download and prepare dependant modules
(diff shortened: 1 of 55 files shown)


Makefile
```diff
@@ -67,6 +67,7 @@ make pytest        # run python tests (tests/pytests)
   ONLY_STABLE=1        # skip unstable tests
   TEST_PARALLEL=n      # test parallalization
   LOG_LEVEL=<level>    # server log level (default: debug)
+  ENABLE_ASSERT=1      # enable assertions
```
kei-nan, line 70: maybe add what the default is if left unspecified?

raz-mon: I can add that, but it will be different from master.. I'll write to myself to add it to master in another PR

kei-nan (approved):

PR #5112, opened 2024-10-13
MOD-7873, MOD-7877: Use Hidden Types In Field And Index Specs

kei-nan:
Use hidden types in the context of index name and field name and spec.

A clear and concise description of what the PR is solving, including:
1. Current: Names are accessed freely in the code and are outputted directly to the logs, exposing user data
2. Change:
     - Wrap the strings in the hidden types to try and prevent unwanted access to them.
       - The most important types are: Indexes, Fields and so on.
     - Output obfuscated strings to the logs depending on the `hide-user-data-from-log` redis config 
     - Add two sub commands to `FT.DEBUG`, `INDEXES` and `INFO` to expose index information in an obfuscated way.
3. Result: Better hide user data from being outputted.

**Which issues this PR fixes**
1. MOD-7873
2. MOD-7877

**Main objects this PR modified**
1. Index and Field Specs

**Mark if applicable**

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #5740, opened 2025-03-10
MOD-7878: Support Index Debug Commands

kei-nan:
Created PR based on changes in:
https://github.com/RediSearch/RediSearch/pull/5112/

1. Current: We don't have a command to get obfuscated information on indexes
2. Change: Add two debug commands, one to list the indexes obfuscated names, the other to get the obfuscated info for an index given its obfuscated name.
In addition, we are replacing the index unique id with a hashed name which will be more consistent across shards.
4. Outcome: Providing a way to obtain obfuscated information on indexes.

1. Debug Commands
2. IndexSpec

#### Mark if applicable


src/spec.c
```diff
@@ -1582,11 +1596,7 @@ void IndexSpec_RemoveFromGlobals(StrongRef spec_ref) {
   dictDelete(specDict_g, spec->name);
 
   // Remove spec from global aliases list
-  if (spec->uniqueId) {
-    // If uniqueid is 0, it means the index was not initialized
-    // and is being freed now during an error.
```
raz-mon, line 1587: This cannot happen?

raz-mon (commented):
Nice 👌 
See one comment

src/spec.c
kei-nan: see:
https://github.com/RediSearch/RediSearch/pull/5740/files#diff-57034127e779577bf2707c68447b4b8f44aadf163d76934b0d14799f409feec3L2645

When loading from rdb we used to set the id to 0 to indicate we had an old spec with the same name.
Since that code was modified to check at the start of the function we no longer need this code.

raz-mon (approved):

raz-mon (approved):

PR #5744, opened 2025-03-11
[MOD-8971] Introduce `LowMemoryThinVec`, a low-memory port of `thin_vec::ThinVec`

LukeMathWalker:
`LowMemoryThinVec` is pointer-sized, since it stores its length and capacity on the heap, in the same allocation used to store its elements.

The implementation is based on a fork of the `thin_vec` crate. In `low_memory_thin_vec`:
- We remove code paths that are unnecessary for our usecase (e.g. `gecko-ffi`, draining, splicing)
- We reduce the maximum capacity of the `std` variant to `u16`, matching what we expect to use in our Rust trie re-implementation.
- We leverage newer `std` APIs to simplify or remove existing unsafe blocks and move computations to compile-time via `const` annotations where possible.
- We review all existing unsafe blocks and add proper safety comments.

A fork will allow us to evolve the code in whatever direction we deem optimal as the porting work progresses.

## Reviewing this PR

Low-level data structures such as `LowMemoryThinVec` require a significant amount of `unsafe` Rust.  
To review this PR, I recommend having a look at the following resources:

- The [Rustnomicon](https://doc.rust-lang.org/nomicon/). In particular, the ["Implementing Vec"](https://doc.rust-lang.org/nomicon/vec/vec.html) section.
- [Jack Wrenn's article on safety hygiene](https://jack.wrenn.fyi/blog/safety-hygiene/).
- The documentation of all the `unsafe` functions we are calling in each `unsafe` block. 

### Coverage

We don't yet have test coverage in CI/`make` for Rust code (coming soon in a separate PR), but you can compute coverage via `cargo-llvm-cov` using this command:

```bash
# Install via `cargo install cargo-llvm-cov`
cd src/redisearch_rs/low_memory_thin_vec
# We use the `nightly` toolchain to include documentation tests,
# which are not yet supported in LLVM coverage on the stable toolchain
cargo +nightly llvm-cov --doctests --html --open
```

I've attached a precomputed coverage report for convenience: [coverage_report.zip](https://github.com/user-attachments/files/19212520/coverage_report.zip)

Summary:
image

### `miri`

Rust provides a tool called [`miri`](https://github.com/rust-lang/miri) to detect undefined behaviour. It's particularly useful when working on code with many `unsafe` blocks, such as this PR.
We plan to add `miri` to our CI pipeline in a separate PR, but you can run it locally via:

```bash
# Install via `rustup +nightly component add miri` and
# then run `cargo +nightly miri setup`. 
cargo +nightly miri test --all-targets
```

No undefined behaviour was reported.
It is not a _proof_ that there is no UB but, combined with the high test coverage, it's pretty promising.

## Remaining work

- [x] Review all `unsafe` blocks and annotate them with appropriate `// SAFETY` comments.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/low_memory_thin_vec/src/layout.rs
+++ b/src/redisearch_rs/low_memory_thin_vec/src/layout.rs
@@ -1,0 +1,66 @@
+//! Utilities for computing the layout of allocations.
+use crate::header::Header;
+use std::alloc::Layout;
+
+/// Gets the layout of the allocated memory for a `LowMemoryThinVec<T>` with the given capacity.
+pub(crate) const fn allocation_layout<T>(cap: usize) -> Layout {
+    let mut vec = Layout::new::<Header>();
+    let elements = match Layout::array::<T>(cap) {
+        Ok(elements) => elements,
+        Err(_) => {
+            // The panic message must be known at compile-time if we want `allocation_layout` to be a `const fn`.
+            // Therefore we can't capture the error (nor the faulty capacity value) in the panic message.
+            panic!(
+                "The size of the array of elements within `LowMemoryThinVec<T>` would exceed `isize::MAX`, \
+                which is the maximum size that can be allocated."
+            )
+        }
+    };
+    vec = match vec.extend(elements) {
+        Ok((layout, _)) => layout,
+        Err(_) => {
+            // The panic message must be known at compile-time if we want `allocation_layout` to be a `const fn`.
+            // Therefore we can't capture the error (nor the faulty capacity value) in the panic message.
+            panic!(
+                "The size of the allocated buffer for `LowMemoryThinVec` would exceed `isize::MAX`, \
+                which is the maximum size that can be allocated."
+            )
+        }
+    };
+    vec.pad_to_align()
+}
+
+/// Gets the alignment for the allocation owned by `LowMemoryThinVec<T>`.
+pub(crate) const fn allocation_alignment<T>() -> usize {
+    // Alignment doesn't change with capacity, so we can use an arbitrary value.
+    // Since:
+    // - the capacity value is known at compile-time
+    // - `allocation_layout` is a `const` function
+    // we can mark `alloc_align` as `const` and be sure that `alloc_align` will be
+    // computed at compile time for any `T` that may end up being used in our
+    // program as a type for the elements within `LowMemoryThinVec<T>`.
+    allocation_layout::<T>(1).align()
+}
+
+/// Gets the size for the allocation owned by `LowMemoryThinVec<T>`, given its required capacity.
+///
+/// # Panics
+///
+/// This will panic if isize::MAX is overflowed at any point.
+pub(crate) const fn allocation_size<T>(cap: usize) -> usize {
+    allocation_layout::<T>(cap).size()
+}
+
+/// Gets the padding that must be inserted between the end of the header field
+/// and the start of the elements array to ensure proper alignment.
+///
+/// # Performance
+///
+/// This will value will be computed at compile time for any `T` that may end up being used in our
+/// program as a type for the elements within `LowMemoryThinVec<T>`, since the function is `const`
+/// and takes no runtime arguments.
+pub(crate) const fn header_field_padding<T>() -> usize {
+    let alloc_align = allocation_alignment::<T>();
+    let header_size = std::mem::size_of::<Header>();
+    alloc_align.saturating_sub(header_size)
+}
--- a/src/redisearch_rs/low_memory_thin_vec/src/layout.rs
+++ b/src/redisearch_rs/low_memory_thin_vec/src/layout.rs
@@ -42,15 +42,6 @@ pub(crate) const fn allocation_alignment<T>() -> usize {
     allocation_layout::<T>(1).align()
 }
 
-/// Gets the size for the allocation owned by `LowMemoryThinVec<T>`, given its required capacity.
-///
-/// # Panics
-///
-/// This will panic if isize::MAX is overflowed at any point.
-pub(crate) const fn allocation_size<T>(cap: usize) -> usize {
-    allocation_layout::<T>(cap).size()
-}
-
 /// Gets the padding that must be inserted between the end of the header field
 /// and the start of the elements array to ensure proper alignment.
 ///
--- a/src/redisearch_rs/low_memory_thin_vec/src/layout.rs
+++ b/src/redisearch_rs/low_memory_thin_vec/src/layout.rs
@@ -64,3 +64,32 @@ pub(crate) const fn header_field_padding<T>() -> usize {
     let header_size = std::mem::size_of::<Header>();
     alloc_align.saturating_sub(header_size)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_header_field_padding() {
+        // Zero header padding needed if storing `u8`s,
+        // since the whole allocation is aligned to 2,
+        // the alignment of the header.
+        assert_eq!(header_field_padding::<u8>(), 0);
+
+        // With `u16` elements, both elements and header have the
+        // same alignment, so no padding is needed.
+        assert_eq!(header_field_padding::<u16>(), 0);
+
+        // With `u32` elements, the whole allocation is aligned to 4,
+        // but the header is 4 bytes long, so no padding is needed.
+        assert_eq!(header_field_padding::<u32>(), 0);
+
+        // With `u64` elements, the whole allocation is aligned to 8,
+        // so the header needs 4 bytes of padding to be aligned.
+        assert_eq!(header_field_padding::<u64>(), 4);
+
+        // With `u128` elements, the whole allocation is aligned to 16,
+        // so the header needs 12 bytes of padding to be aligned.
+        assert_eq!(header_field_padding::<u128>(), 12);
+    }
+}
--- a/src/redisearch_rs/low_memory_thin_vec/src/layout.rs
+++ b/src/redisearch_rs/low_memory_thin_vec/src/layout.rs
@@ -5,16 +5,13 @@ use std::alloc::Layout;
 /// Gets the layout of the allocated memory for a `LowMemoryThinVec<T>` with the given capacity.
 pub(crate) const fn allocation_layout<T>(cap: usize) -> Layout {
     let mut vec = Layout::new::<Header>();
-    let elements = match Layout::array::<T>(cap) {
-        Ok(elements) => elements,
-        Err(_) => {
-            // The panic message must be known at compile-time if we want `allocation_layout` to be a `const fn`.
-            // Therefore we can't capture the error (nor the faulty capacity value) in the panic message.
-            panic!(
-                "The size of the array of elements within `LowMemoryThinVec<T>` would exceed `isize::MAX`, \
-                which is the maximum size that can be allocated."
-            )
-        }
+    let Ok(elements) = Layout::array::<T>(cap) else {
+        // The panic message must be known at compile-time if we want `allocation_layout` to be a `const fn`.
+        // Therefore we can't capture the error (nor the faulty capacity value) in the panic message.
+        panic!(
+            "The size of the array of elements within `LowMemoryThinVec<T>` would exceed `isize::MAX`, \
+                        which is the maximum size that can be allocated."
+        )
     };
     vec = match vec.extend(elements) {
         Ok((layout, _)) => layout,
--- a/src/redisearch_rs/low_memory_thin_vec/src/layout.rs
+++ b/src/redisearch_rs/low_memory_thin_vec/src/layout.rs
@@ -47,7 +47,7 @@ pub(crate) const fn allocation_alignment<T>() -> usize {
 ///
 /// # Performance
 ///
-/// This will value will be computed at compile time for any `T` that may end up being used in our
+/// This value will be computed at compile time for any `T` that may end up being used in our
 /// program as a type for the elements within `LowMemoryThinVec<T>`, since the function is `const`
 /// and takes no runtime arguments.
 pub(crate) const fn header_field_padding<T>() -> usize {
--- a/src/redisearch_rs/low_memory_thin_vec/src/lib.rs
+++ b/src/redisearch_rs/low_memory_thin_vec/src/lib.rs
@@ -1316,9 +1316,7 @@ impl<T: Clone> LowMemoryThinVec<T> {
 impl<T> Drop for LowMemoryThinVec<T> {
     #[inline]
     fn drop(&mut self) {
-        #[cold]
-        #[inline(never)]
-        fn drop_non_singleton<T>(this: &mut LowMemoryThinVec<T>) {
+        if !self.is_singleton() {
             // First deallocate all elements, since they may
             // need own other resources that must be freed.
             // If we don't do this first, we may leak memory.
@@ -1326,7 +1324,7 @@ impl<T> Drop for LowMemoryThinVec<T> {
             // - The pointer is valid and unaliased, since it comes from a `&mut` reference.
             // - We're inside a `Drop` implementation.
             unsafe {
-                ptr::drop_in_place(&mut this[..]);
+                ptr::drop_in_place(self.as_mut_slice());
             }
 
             // SAFETY:
@@ -1335,14 +1333,10 @@ impl<T> Drop for LowMemoryThinVec<T> {
             //   we're using to deallocate it.
             unsafe {
                 dealloc(
-                    this.ptr.as_ptr() as *mut u8,
-                    allocation_layout::<T>(this.capacity()),
+                    self.ptr.as_ptr() as *mut u8,
+                    allocation_layout::<T>(self.capacity()),
                 )
             }
-        }
-
-        if !self.is_singleton() {
-            drop_non_singleton(self);
         }
     }
 }
--- a/src/redisearch_rs/low_memory_thin_vec/src/lib.rs
+++ b/src/redisearch_rs/low_memory_thin_vec/src/lib.rs
@@ -10,16 +10,16 @@
 //!   -----               |              ----
 //!                       |
 //!  +---------------+    |
-//!  | ptr (8 bytes) | ------->  Header +------------+
-//!  +---------------+    |             | 3 (len)    |
-//!                       |             | 4 (cap)    |
-//!                       |             | (padding)  |
-//!                       |             +------------+
-//!                       |      Data   | 12         |
-//!                       |             | 151        |
-//!                       |             | 2          |
-//!                       |             | (unused)   |
-//!                       |             +------------+
+//!  | ptr (8 bytes) | ------->  Header +----------------------+
+//!  +---------------+    |             | 3 (len)     (2 bytes)|
+//!                       |             | 4 (cap)     (2 bytes)|
+//!                       |             | (padding)   (4 bytes)|
+//!                       |             +----------------------+
+//!                       |      Data   | 12          (8 bytes)|
+//!                       |             | 151         (8 bytes)|
+//!                       |             | 2           (8 bytes)|
+//!                       |             | (unused)    (8 bytes)|
+//!                       |             +----------------------+
 //! ```
 //!
 //! It's pointer-sized on the stack, compared to `Vec<T>` which has 3 pointer-sized fields:
@@ -28,18 +28,18 @@
 //!    Stack              |      Heap
 //!    -----              |      ----
 //!                       |
-//!   +---------------+   |      +------------+
-//!   | ptr (8 bytes) | -------> | 12         |
-//!   | len (8 bytes) |   |      | 151        |
-//!   | cap (8 bytes) |   |      | 2          |
-//!   +---------------+   |      | (unused)   |
-//!                       |      +------------+
+//!   +---------------+   |      +--------------------+
+//!   | ptr (8 bytes) | -------> | 12        (8 bytes)|
+//!   | len (8 bytes) |   |      | 151       (8 bytes)|
+//!   | cap (8 bytes) |   |      | 2         (8 bytes)|
+//!   +---------------+   |      | (unused)  (8 bytes)|
+//!                       |      +--------------------+
 //! ```
 //!
 //! # Memory footprint
 //!
-//! The memory footprint of LMThinVecs is lower than `Vec<T>` ; notably in cases where space is reserved for
-//! a non-existence `LowMemoryThinVec<T>`. So `Vec<LowMemoryThinVec<T>>` and `Option<LowMemoryThinVec<T>>::None` will waste less
+//! The memory footprint of `LowMemoryThinVec<T>` is smaller than `Vec<T>` ; notably in cases where space is reserved for
+//! the non-existence of `LowMemoryThinVec<T>`. So `Vec<LowMemoryThinVec<T>>` and `Option<LowMemoryThinVec<T>>::None` will waste less
 //! space. Being pointer-sized also means it can be passed/stored in registers.
 //!
 //! Of course, any actually constructed `LowMemoryThinVec` will theoretically have a bigger allocation, but
@@ -977,7 +977,7 @@ impl<T> LowMemoryThinVec<T> {
         }
         // Ensure the new capacity is at least double, to guarantee exponential growth.
         let double_cap = if old_cap == 0 {
-            // skip to 4 because tiny ThinVecs are dumb; but not if that would cause overflow
+            // skip to 4 because tiny vecs are dumb; but not if that would cause overflow
             if mem::size_of::<T>() > (!0) / 8 { 1 } else { 4 }
         } else {
             old_cap.saturating_mul(2)
--- a/src/redisearch_rs/low_memory_thin_vec/src/lib.rs
+++ b/src/redisearch_rs/low_memory_thin_vec/src/lib.rs
@@ -143,7 +143,7 @@ pub struct LowMemoryThinVec<T> {
     //
     // ```rust,ignore
     // enum HeaderRef {
-    //     Singleton(&'static Header), // -> pointing at `EMPTY_HEADER`
+    //     Singleton, // -> pointing at `EMPTY_HEADER`
     //     Allocated(&Header), // -> pointing at the beginning of the allocated buffer
     // }
     // ```
@@ -1182,6 +1182,7 @@ impl<T> LowMemoryThinVec<T> {
         );
         if self.has_allocation() {
             let old_cap = self.capacity();
+            let new_layout = allocation_layout::<T>(new_cap);
             // SAFETY:
             // - `self.ptr` was allocated via the same global allocator.
             // - We're using the correct layout for the old capacity.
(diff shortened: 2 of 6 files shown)


src/redisearch_rs/low_memory_thin_vec/src/layout.rs
```diff
@@ -0,0 +1,66 @@
...
+        Ok((layout, _)) => layout,
+        Err(_) => {
+            // The panic message must be known at compile-time if we want `allocation_layout` to be a `const fn`.
+            // Therefore we can't capture the error (nor the faulty capacity value) in the panic message.
+            panic!(
+                "The size of the allocated buffer for `LowMemoryThinVec` would exceed `isize::MAX`, \
+                which is the maximum size that can be allocated."
+            )
+        }
+    };
```
kei-nan, line 29:
```suggestion
    let elements = Layout::array::<T>(cap);
    vec = match vec.extend(elements) {
        Ok((layout, _)) => layout,
        Err(_) => {
            // The panic message must be known at compile-time if we want `allocation_layout` to be a `const fn`.
            // Therefore we can't capture the error (nor the faulty capacity value) in the panic message.
            panic!(
                "The size of the allocated buffer for `LowMemoryThinVec` would exceed `isize::MAX`, \
                which is the maximum size that can be allocated."
            )
        }
    };
```

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2049 @@
...
+//!   | ptr (8 bytes) | -------> | 12         |
+//!   | len (8 bytes) |   |      | 151        |
+//!   | cap (8 bytes) |   |      | 2          |
+//!   +---------------+   |      | (unused)   |
+//!                       |      +------------+
+//! ```
+//!
+//! # Memory footprint
+//!
+//! The memory footprint of LMThinVecs is lower than `Vec<T>` ; notably in cases where space is reserved for
```
kei-nan, line 41: LMThinVecs - use updated type name or use a more generic sentence which doesn't use the struct name.

src/redisearch_rs/low_memory_thin_vec/src/layout.rs
```diff
@@ -0,0 +1,66 @@
...
+///
+/// # Performance
+///
+/// This will value will be computed at compile time for any `T` that may end up being used in our
+/// program as a type for the elements within `LowMemoryThinVec<T>`, since the function is `const`
+/// and takes no runtime arguments.
+pub(crate) const fn header_field_padding<T>() -> usize {
+    let alloc_align = allocation_alignment::<T>();
+    let header_size = std::mem::size_of::<Header>();
+    alloc_align.saturating_sub(header_size)
```
kei-nan, line 65: The calculation seems a bit off to me.
If header size is 5, padding is 3 and T size is 8 the alloc_align will be 16.
we then subtract 5 from it which is 11.
Shouldn't it have returned 3 in this case?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2049 @@
...
+    let Some(header) = NonNull::new(header) else {
+        // The allocation failed!
+        handle_alloc_error(layout)
+    };
+
+    // If `T` is a zero-sized type, we won't ever need to increase the size of
+    // the allocation. Its values take no space in memory!
+    // Therefore it's safe to set the capacity to `MAX_CAP`.
+    let capacity = if mem::size_of::<T>() == 0 {
+        MAX_CAP
```
kei-nan, line 117: Just out of curiosity, would using an option for the capacity help cleaning up some parts of the code that need to deal with the case of a T with size 0?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2049 @@
...
+    };
+
+    // Initialize the allocated buffer with a valid header value.
+    //
+    // SAFETY:
+    // - The destination and the value are properly aligned,
+    //   since the allocation was performed against a type layout
+    //   that begins with a header field.
+    unsafe {
+        header.write(Header::new(0, capacity));
```
kei-nan, line 129: Out of curiosity,
Is there a guarantee of some sort on the padding and/or elements content at this point?
e.g are they always zeroed out?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: Off the top of my head, I don't think it would make things clearer.
If you search for `size_of` invocations and/or `zero-sides`, it turns out that this is the only location we actually check for it. That's thanks to the fact that we're setting capacity to `MAX_CAP` here.

Regardless of code clarity, we can't afford an `Option` in terms of performance—we're accessing the capacity on almost all operations, so we don't want to introduce unnecessary runtime overhead on that code path.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: No, there is no guarantee.
The memory is considered to be uninitialized and must be treated as if it contains random garbage.

[`MaybeUninit`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html) is the way to go when that needs to be modelled more closely. The dedicated [Rustonomicon section](https://doc.rust-lang.org/nomicon/uninitialized.html) is also a good read.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: Fixed in fb709581d

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2049 @@
...
+//!
+//! The memory layout of a local variable of type `LowMemoryThinVec<u64>` looks as follows:
+//!
+//!```text
+//!   Stack               |              Heap
+//!   -----               |              ----
+//!                       |
+//!  +---------------+    |
+//!  | ptr (8 bytes) | ------->  Header +------------+
+//!  +---------------+    |             | 3 (len)    |
```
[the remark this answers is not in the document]
LukeMathWalker: Fixed in fb709581d

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+    ///
+    /// let vec: LowMemoryThinVec<i32> = LowMemoryThinVec::with_capacity(10);
+    /// assert_eq!(vec.capacity(), 10);
+    /// ```
+    pub fn capacity(&self) -> usize {
+        self.header_ref().capacity()
+    }
+
+    /// Returns `true` if the vector has the capacity to hold any element.
+    pub fn has_capacity(&self) -> bool {
```
kei-nan, line 387: I think this function might be a bit misleading.
We could argue that a full vector(that is not a singleton) should return false when this function is called.

When would someone use this sort of function?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+    /// While the following example is sound, there is a memory leak since
+    /// the inner vectors were not freed prior to the `set_len` call:
+    ///
+    /// ```no_run
+    /// use low_memory_thin_vec::low_memory_thin_vec;
+    ///
+    /// let mut vec = low_memory_thin_vec![low_memory_thin_vec![1, 0, 0],
+    ///                    low_memory_thin_vec![0, 1, 0],
+    ///                    low_memory_thin_vec![0, 0, 1]];
+    /// // SAFETY:
```
kei-nan, line 464: there are two safety sections.
is the second safety section referring to the set_len function or the code in the example?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+    /// let mut vec = low_memory_thin_vec![1, 2];
+    /// vec.push(3);
+    /// assert_eq!(vec, [1, 2, 3]);
+    /// ```
+    pub fn push(&mut self, val: T) {
+        let old_len = self.len();
+        if old_len == self.capacity() {
+            self.reserve(1);
+        }
+
```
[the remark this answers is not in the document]
LukeMathWalker: I'd say the latter in this case.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+        // - Pointer is within the allocation backing the vector,
+        //   since `idx` is strictly smaller than `old_len` at
+        //   this point. We've already deferred to `push` the
+        //   case where `idx == old_len`.
+        let after_idx_ptr = unsafe { data_ptr.add(idx + 1) };
+        // SAFETY:
+        // We know that we have enough capacity to shift everything
+        // to the right by 1 thanks to `reserve`.
+        unsafe {
+            ptr::copy(idx_ptr, after_idx_ptr, old_len - idx);
```
kei-nan, line 643: copying here seems very dependent on the copy direction.
does rust copy handle the copy direction correctly?
[1, 2, 3, 4] -> shift every element 1 place to the write
possible results:
[_, 1, 1, 1, 1]
[_, 1, 2 ,3, 4]

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+    {
+        let len = self.len();
+        let mut del = 0;
+        {
+            let v = &mut self[..];
+
+            for i in 0..len {
+                if !f(&mut v[i]) {
+                    del += 1;
+                } else if del > 0 {
```
kei-nan, line 1101: is it worth checking i - del != i?
can that optimize things a bit more?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+    pub fn clear(&mut self) {
+        // Drop the elements to ensure any resource they own is released.
+        // Failing to do so could lead to memory leaks or other resource leaks.
+        //
+        // SAFETY:
+        // - The pointer is valid and aligned, since it comes from a reference.
+        // - We have exclusive access to the element, the pointer
+        //   comes from `&mut self`.
+        unsafe {
+            ptr::drop_in_place(&mut self[..]);
```
kei-nan, line 906: trying to wrap my head around this code.
this function is not marked as unsafe.
What guarantees do we have we are not a singleton?
Won't we get a dangling pointer for the singelton when we try and get the data pointer to the elements?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+        self.extend(other.iter().cloned())
+    }
+}
+
+impl<T> Drop for LowMemoryThinVec<T> {
+    #[inline]
+    fn drop(&mut self) {
+        #[cold]
+        #[inline(never)]
+        fn drop_non_singleton<T>(this: &mut LowMemoryThinVec<T>) {
```
kei-nan, line 1304: why do we mark this as cold?
Doesn't it mean it is less likely to be reached?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+        }
+        for x in iter {
+            self.push(x);
+        }
+    }
+}
+
+impl<T: fmt::Debug> fmt::Debug for LowMemoryThinVec<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Debug::fmt(&**self, f)
```
kei-nan, line 1384: Can you explain what the double dereferencing does in this case?
Do we bring back one element out of the vector?
What handling is there if the vector is empty?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+    }
+}
+
+impl<T> Hash for LowMemoryThinVec<T>
```
kei-nan, line 1388: I feel like we might want to forbit using vectors as keys in a hash map.
It feels like a costly operation which we might not want to be done implicitly.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+where
+    A: PartialEq<B>,
+{
+    #[inline]
+    fn eq(&self, other: &&'a [B]) -> bool {
+        self[..] == other[..]
+    }
+}
+
+macro_rules! array_impls {
```
kei-nan, line 1460: what is this for?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: That's a good point, the name is confusing and we had another (private) function that amounted to pretty much the same goal.

Fixed in 5a785c96b.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: I'm not sure I follow.
`ptr::copy(source, destination, length)` will write the memory block `[source..source + length]` at the location specified by `destination`, thus overwriting whatever was previously stored in `[destination..destination+length]`.

You're always writing "to the right" of the destination pointer. Does that answer the question?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+
+        // We shift everything past `idx` to the left by 1, to fill the
+        // gap left by the removed element.
+        //
+        // SAFETY:
+        // - We know that the range we're copying from is initialized.
+        // - We know that the range we're copying to is within bounds.
+        // - We know that all pointers are aligned.
+        unsafe {
+            ptr::copy(after_idx_ptr, idx_ptr, old_len - idx - 1);
```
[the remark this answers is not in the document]
LukeMathWalker: Correct. The safety contract doesn't require the count to be non-zero (see [docs](https://doc.rust-lang.org/std/ptr/fn.copy.html#safety)). If it's zero, it becomes a no-op (but the pointer must be aligned nonetheless!).

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: We would have to look at the generated code (or add some micro-benchmarks) to see if the compiler is smart enough to figure it out on its own (and determine if adding an `if` clause benefits/damages performance).

I'm inclined to avoid the additional effort until this turns out to be a method we invoke in a hot-loop/performance-sensitive situation and therefore it makes sense to focus heavily on it.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: A function should be marked `unsafe` if the _caller_ of that function needs to verify some preconditions/invariants that the compiler can't check on its own.
In the case of `clear`, the only thing the caller must guarantee is the lack of aliasing, which can be modelled via the type system using `&mut self`. That's why we don't need to mark `clear` as `unsafe`—the compiler can check all our preconditions automatically.

Talking about `drop_in_place`: we're dropping a _slice_ here (notice the `self[..]` syntax; that's a slice that spans all elements currently stored in the vector). Since we're writing `&mut self[..]`, that's equivalent to `self.as_slice_mut()`.
Calling `drop_in_place` on a slice is equivalent to iterating over all the values it contains and then calling `drop` on each one, **without deallocating the memory that backs the slice.

If we're in the singleton case, the slice is empty, so that's a no-op. In the singleton case, the pointer may be dangling, but that's not an issue since the slice length is set to 0, therefore it won't ever be dereferenced.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: Yes, that's correct, since we expect a lot of vectors to be empty.
It's an optimisation we can tune based on benchmarks later on.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: It desugars to `&(*(self.deref()))` which is a very convoluted way to a `&[T]`.
It's too implicit even for my taste, so I replace all implicit conversions to slices with the corresponding method call in 4604348da

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: It's never implicit, though, is it?
You're defining a `HashMap` and you're using a vector as its key type. 
It's not something I expect to see in production code, but most of these trait implementations are very convenient in other scenarios (e.g. writing tests) so I'd recommend we keep them.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: It allows you to compare a `LowMemoryThinVec` against a fixed-length array (i.e. `[T; N]`).
Since the implementation is the same for every value of `N`, we use a macro to avoid having 32 copies of the same code block.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+{
+    fn hash<H>(&self, state: &mut H)
+    where
+        H: Hasher,
+    {
+        self[..].hash(state);
+    }
+}
+
+impl<T> PartialOrd for LowMemoryThinVec<T>
```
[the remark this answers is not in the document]
LukeMathWalker: See my reasoning in https://github.com/RediSearch/RediSearch/pull/5744#discussion_r2002937133

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+    /// assert_eq!(into_iter.as_slice(), &['b', 'c']);
+    /// ```
+    pub fn as_slice(&self) -> &[T] {
+        // SAFETY:
+        // - The data pointer is always aligned and it's safe to offset by an
+        //   index within the bounds of the vector, since it'll fall within the same
+        //   allocation.
+        //   `self.start` has been checked to be within bounds when the iterator was created/last advanced.
+        let start_ptr = unsafe { self.vec.data_raw().add(self.start) };
+        let n_remaining_elements = self.len();
```
GuyAv46, line 1775: where does `self.len()` is defined? is it `self.vec.len() - self.start`?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: Yes, it's `memmove`. You get `memcpy` with https://doc.rust-lang.org/std/intrinsics/fn.copy_nonoverlapping.html

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+    /// ```
+    ///
+    /// [`clear`]: LowMemoryThinVec::clear
+    /// [`drain`]: LowMemoryThinVec::drain
+    pub fn truncate(&mut self, len: usize) {
+        // drop any extra elements
+        while len < self.len() {
+            // Decrement the length *before* calling drop_in_place(),
+            // so that a panic on `Drop` doesn't try to re-drop the
+            // value with just-failed to drop.
```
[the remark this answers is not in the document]
LukeMathWalker: There's no way to know if `T` may or may not panic on drop, so this is unavoidable unfortunately.

src/redisearch_rs/low_memory_thin_vec/src/layout.rs
```diff
@@ -0,0 +1,86 @@
...
+        Ok(elements) => elements,
+        Err(_) => {
+            // The panic message must be known at compile-time if we want `allocation_layout` to be a `const fn`.
+            // Therefore we can't capture the error (nor the faulty capacity value) in the panic message.
+            panic!(
+                "The size of the array of elements within `LowMemoryThinVec<T>` would exceed `isize::MAX`, \
+                which is the maximum size that can be allocated."
+            )
+        }
+    };
```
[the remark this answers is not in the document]
LukeMathWalker: Yup, we can. Changed in bb6de21e3

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+        vec
+    });
+    ($($x:expr,)*) => (low_memory_thin_vec![$($x),*]);
+}
+
+impl<T> LowMemoryThinVec<T> {
+    /// Creates a new empty LowMemoryThinVec.
+    ///
+    /// This will not allocate.
+    pub fn new() -> LowMemoryThinVec<T> {
```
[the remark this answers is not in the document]
LukeMathWalker: We can, with a bit of refactoring and a dash of unsafe. Done in  63f97a5b2

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+    // For details, see:
+    // https://github.com/rust-lang/rfcs/blob/master/text/0769-sound-generic-drop.md#phantom-data
+    _phantom: PhantomData<T>,
+}
+
+// SAFETY:
+// `LowMemoryThinVec<T>` is, for all `Send` intents and purposes, equivalent to a `Vec<T>`.
+// This `Send` implementation is therefore safe for the same reasons as the one
+// provided by `Vec<T>` when `T` is `Send`.
+unsafe impl<T: Send> Send for LowMemoryThinVec<T> {}
```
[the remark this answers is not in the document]
LukeMathWalker: `Send` and `Sync` are usually automatically derived by the compiler.
Since we store a `NonNull<Header>`, the compiler can't make assumptions as to what is allowed/forbidden, so it won't derive `Send` and `Sync` for us (see [this impl](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#impl-Send-for-NonNull%3CT%3E)).
If we want to override its decision, we need to implement them manually. The traits we're implementing are marked as `unsafe` since they require the developer to uphold invariants that the compiler can't check automatically in the non-inferred case. Since the trait is `unsafe`, the `impl` block must be marked as `unsafe` too.

For a more in-depth discussion, check [the relevant nomicon section](https://doc.rust-lang.org/nomicon/send-and-sync.html).

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+    }
+
+    // Accessor conveniences
+
+    /// Return a reference to the header.
+    fn header_ref(&self) -> &Header {
+        // SAFETY:
+        // Guaranteed by the invariants on the `ptr` field.
+        // Check out [`LowMemoryThinVec::ptr`] for more details.
+        unsafe { self.ptr.as_ref() }
```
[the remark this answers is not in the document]
LukeMathWalker: There's a big semantic difference between a pointer and a reference in Rust. You can always go from a reference to a pointer, but the latter may not be possible unless a series of prerequisites are met.
The [module documentation for `ptr`](https://doc.rust-lang.org/std/ptr/index.html#pointer-to-reference-conversion) has all the details.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -0,0 +1,2050 @@
...
+        let len = self.len();
+        let mut del = 0;
+        {
+            let v = &mut self[..];
+
+            for i in 0..len {
+                if !f(&mut v[i]) {
+                    del += 1;
+                } else if del > 0 {
+                    v.swap(i - del, i);
```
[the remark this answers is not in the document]
LukeMathWalker: We can't—we would have to require `T: Clone` _and_ perform a `.clone()` invocation.
You can't do a `memcpy` unless the type implements `Copy`, since it may own resources that must not be aliased and must not be released while a live reference is still around.

Consider the case where `T` is `Vec<u64>`—a vector of vectors.
If we copy, we would end up with two pointers at its allocated buffer. The copy that stays in the truncated part will invoke the destructor, thus deallocating that buffer—we now have a dangling pointer in the "live" part of our thin vec.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: It comes from [`ExactSizeIterator`](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html), a trait that we implement just below.
And yes, it is the number of remaining elements, as defined in `size_hint`.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
kei-nan: So if I try and compare a thin vec against a fixed length array of size greater than 32 it will fail to compile?

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: Since we don't have hard benchmarking data to look at, I removed the attributes in abb1b4e77. We can revisit the decision later on.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: I've done it on more than one occasion where performance wasn't critical.
I don't think we should limit the expressiveness of the library on the off chance that it may be misused. There are many other ways to write inefficient code with the other exposed methods, and there are code reviews/microbenchmarks to catch naive implementations.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
LukeMathWalker: It turns out that const generics have improved, so we can now do a single impl that applies to arrays of arbitrary sizes: bbf7727ed

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
kei-nan: @GuyAv46 WDYT?

GuyAv46: I believe we have the same definition for hashing an `RSValue` of type vector… I agree with Luca, we won’t use vectors as a key for a hashmap unless we want to, I don’t think we need to block this option.
Let’s review PRs carefully and not let any juniors make such features by accident :)

GuyAv46: We use it in the GROUPBY step to identify unique groups

GuyAv46 (approved):
Would be nice to get 100% coverage

kei-nan (approved):

PR #5765, opened 2025-03-16
MOD-9054: Fix `total_results` field in shard RESP3 response

raz-mon:
Changes introduced in this PR:
* Removes a redundant check from the RESP2 + RESP3 response paths.
* Fixes the `total_results` field in the RESP3 response to be **CORRECT**, always.

Note: This fix should not be breaking, since it changes the response order of the MAP we use in RESP3, in which the order plays no role.
For that reason exactly (breakage), we do not fix this for RESP2.


src/aggregate/aggregate_exec.c
```diff
@@ -447,9 +447,7 @@ static void sendChunk_Resp2(AREQ *req, RedisModule_Reply *reply, size_t limit,
     }
 
     // Set `resultsLen` to be the expected number of results in the response.
-    if (ShouldReplyWithTimeoutError(rc, req)) {
-      resultsLen = 1;
-    } else if (rc == RS_RESULT_ERROR) {
+    if (rc == RS_RESULT_ERROR) {
```
meiravgri, line 450: is this correct also when no documents are returned?

meiravgri: can you add a test for it? (with FT.DEBUG FT.DEARCH TIMEOUT_AFTER_N 0) ?

src/aggregate/aggregate_exec.c
```diff
@@ -619,6 +608,9 @@ static void sendChunk_Resp3(AREQ *req, RedisModule_Reply *reply, size_t limit,
 done_3:
     RedisModule_Reply_ArrayEnd(reply); // >results
 
+    // <total_results>
+    RedisModule_ReplyKV_LongLong(reply, "total_results", req->qiter.totalResults);
+
```
meiravgri, line 613: isn't it going to change the reply fields' order?

tests/pytests/test_resp3.py
```diff
@@ -1601,3 +1601,22 @@ def test_warning_maxprefixexpansions():
     if shard['Warning']== 'Max prefix expansions limit was reached':
          n_warnings += 1
   env.assertEqual(n_warnings, 1)
+
+def test_totalResults_aggregate():
```
meiravgri, line 1605: what about resp2?

src/aggregate/aggregate_exec.c
raz-mon: Mentioned in the PR comment, only for RESP3 where the order doesn't matter AFAIK

tests/pytests/test_resp3.py
raz-mon: It didn't change there

src/aggregate/aggregate_exec.c
raz-mon: The condition is just redundant as far as I can see..
I don't want to change the behavior of possible paths, since that will be breaking

meiravgri: Yeah i see the repetition...
So in general,  do we have test with timeout error and 0 results?

src/aggregate/aggregate_exec.c
meiravgri: 🤔 i didn't know that

meiravgri: anyway, is  `qiter.totalResults`  derived from the results array length to give the most accurate value?

src/aggregate/aggregate_exec.c
raz-mon: Not sure
It's hard to say since this requires a traversal of a lot of tests. We can create a ticket for the addition of such a test

src/aggregate/aggregate_exec.c
raz-mon: Kind of.. But more no than yes, unfortunately.
Take a look in the code, it's incremented and decremented on the fly as results get propagated or discarded in the pipeline flow.
It may be more accurate on the coordinator than on the shard level, since there it is taking the length of the `results` component as the base value.
This needs more attention in order to be 100% accurate, but I didn't want to break any existing behavior since this PR is meant to go to the version branches in a patch as well.

We will need to have a continuation to regularize this for all paths and pipelines, that will probably break some things..

raz-mon: It will give a different value, that will be correct in more scenarios than the current state, relatively for free due to the map nature of the response. The "free" part is mainly since it is not breaking, and also low effort as you can see.

For instance, today using the strict timeout policy we will get a correct value in both implementations with high probability, but using the non-strict policy we almost never get a correct value in the current state. After the fix, we will get a correct value for that case instead of the non-informative `1` value that we got until now.

meiravgri (dismissed):

src/aggregate/aggregate_exec.c
oshadmi: @raz-mon Nitpick - if `total_results` is still not guaranteed to always be accurate, although being better than the current value, will users be able to rely on it (with RESP3)?

tests/pytests/test_resp3.py
```diff
@@ -1601,3 +1601,22 @@ def test_warning_maxprefixexpansions():
...
+  # Create an index
+  env.expect('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT').ok()
+
+  # Populate the index
+  n_docs = 17 * env.shardsCount
+  for i in range(n_docs):
+      conn.execute_command('HSET', f'doc{i}', 't', str(i))
+
+  # Test that the `total_results` field is correct
+  res = env.cmd('FT.AGGREGATE', 'idx', '*')
```
oshadmi, line 1621: Do we also need a test with cursor with some `COUNT` showing that also the result count of the last read is also accurate?

src/aggregate/aggregate_exec.c
raz-mon: More than they could yesterday..
As mentioned in [this comment](https://github.com/RediSearch/RediSearch/pull/5765#discussion_r1997637337), we will need to do a more complete and probably breaking work to make sure this value is not broken in some paths

tests/pytests/test_resp3.py
raz-mon: I don't think so, since we output the same variable that we did before.
But I can add it just so we are more quiet about the fix

oshadmi (approved):

PR #5766, opened 2025-03-16
MOD-9055: Make scorer name arguments case-insensitive

raz-mon:

1. Current: Only upper-case scorer name arguments are accepted.
2. Change: Make scorer name argument case insensitive.
3. Outcome: Both upper and lower case forms of the scorer name are valid.

#### Mark if applicable


src/extension.c
```diff
@@ -15,12 +15,51 @@
...
+ * @example
+ * @code
+ * char input[] = "Hello, World!";
+ * char output[50];
+ * str_toUpper(input, output, strlen(input), sizeof(output));
+ * printf("Uppercase: %s\n", output); // Output: "HELLO, WORLD!"
+ * @endcode
+ */
+static inline void str_toUpper(const char *in, char *out, size_t inLen, size_t maxLen) {
+  if (inLen < 0) {
```
oshadmi, line 53: Isn't type of `inLen` unsigned (`size_t`)? Is this reachable?

oshadmi: If `str_toUpper` is currently only used here and always gets a valid `inLen` consider dropping this if block?

src/extension.c
```diff
@@ -15,12 +15,51 @@
 #include "query.h"
 #include <err.h>
 
+#define MAX_NAME_LEN 1024
```
oshadmi, line 18: Consider making it 1025 bytes long, so it would be easier to document that a scorer max name length is 1024 ascii characters

src/extension.c
raz-mon: Yeah I wanted it for generality but there's no real need..
Will modify it

src/extension.c
```diff
@@ -15,12 +15,47 @@
...
+ *
+ * @example
+ * @code
+ * char input[] = "Hello, World!";
+ * char output[50];
+ * str_toUpper(input, output, strlen(input), sizeof(output));
+ * printf("Uppercase: %s\n", output); // Output: "HELLO, WORLD!"
+ * @endcode
+ */
+static inline void str_toUpper(const char *in, char *out, size_t inLen, size_t maxLen) {
```
kei-nan, line 51: optional:
move this function to strconv.h header.

kei-nan, line 51: optional:
move this function to strconv.h header.

kei-nan: What ensures out buffer has enough space?

raz-mon: Makes sense 👍

raz-mon:
Closing due to potential breakage.

PR #5780, opened 2025-03-17
MOD-8969: Compute test coverage for Rust code

LukeMathWalker:

We add support for computing test coverage for Rust code, both locally and in CI. This is done via a new tool, [`cargo-llvm-cov`](https://github.com/taiki-e/cargo-llvm-cov).

The coverage file is located next to the coverage report for C code and uploaded alongside it to Codecov in CI. Codecov will merge the two reports for us on the server side.

There are no user-facing changes to the pre-existing `make` commands. 
`make coverage` will now transparently compute coverage for Rust code as well as C code. If you only wish to compute Rust coverage, you can do so via `make rust-tests COV=1`.

#### Mark if applicable

### DIFF
--- a/.install/test_deps/common_installations.sh
+++ b/.install/test_deps/common_installations.sh
@@ -20,6 +20,9 @@ activate_venv() {
 	fi
 }
 
+# Tool required to compute test coverage for Rust code
+cargo install cargo-llvm-cov --locked
+
 python3 -m venv venv
 activate_venv
 source venv/bin/activate
--- a/.github/workflows/flow-coverage.yml
+++ b/.github/workflows/flow-coverage.yml
@@ -43,7 +43,7 @@ jobs:
     - name: Upload coverage
       uses: codecov/codecov-action@v5
       with:
-        files: bin/linux-x64-debug-cov/cov.info
+        files: bin/linux-x64-debug-cov/cov.info,bin/linux-x64-debug-cov/rust_cov.info
         disable_search: true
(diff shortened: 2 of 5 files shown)


LukeMathWalker:
Working as expected in CI. You can browse the coverage for the Rust files [here](https://app.codecov.io/github/RediSearch/RediSearch/commit/8881620b5fb784489f9ad3de3c1fe9d95df93ff0/tree/src/redisearch_rs?dropdown=coverage) 🎉

hdoordt (approved):

.install/test_deps/common_installations.sh
```diff
@@ -20,6 +20,9 @@ activate_venv() {
 	fi
 }
 
+# Tool required to compute test coverage for Rust code
+cargo install cargo-llvm-cov --locked
```
LukeMathWalker, line 24: Should I install `cargo-llvm-cov` here or should I do it in `install_rush.sh`? cc @kei-nan

kei-nan: My main issue is that coverage is becoming a requirement even when running without coverage.
I don't see a big difference between the placement of the bash files.
Approved the PR none the less.

LukeMathWalker: > My main issue is that coverage is becoming a requirement even when running without coverage.

In what way? Only `make coverage` (or other commands with `COV=1`) will fail if `cargo-llvm-cov` is not installed. It is optional in that sense.

PR #5781, opened 2025-03-17
MOD-8969: Run Rust tests through miri to catch undefined behaviour

LukeMathWalker:

[`miri`](https://github.com/rust-lang/miri) is an official tool to catch undefined behaviour in Rust programs.
It runs the Rust code through a (sandboxed) platform-independent interpreter, tracks which operations are being performed and compares them to Rust's operational semantics to catch UB.
This is particularly useful when working on codebases with a significant amount of `unsafe` code, as in our case.

You need to opt into running tests through `miri`, with a dedicated `RUN_MIRI` option.
In CI it's enabled exclusively for the `sanitize` check (PR event flow + merge-queue event flow).

There is no need to run `miri` on multiple platforms, since the check is platform-independent.

#### Mark if applicable

### DIFF
--- a/.install/install_rust.sh
+++ b/.install/install_rust.sh
@@ -13,18 +13,3 @@ echo "Cargo binary location: $(which cargo)"
 
 # Update to the latest stable toolchain
 rustup update
-
-# --allow-downgrade:
-#   Allow `rustup` to install an older `nightly` if the latest one
-#   is missing one of the components we need.
-# llvm-tools-preview:
-#   Required by `cargo-llvm-cov` for test coverage
-# miri:
-#   Required to run `cargo miri test` for UB detection
-# rust-src:
-#   Required to build RedisJSON with address sanitizer
-rustup toolchain install nightly \
-    --allow-downgrade \
-    --component llvm-tools-preview \
-    --component miri \
-    --component rust-src
--- a/.install/install_rust.sh
+++ b/.install/install_rust.sh
@@ -19,9 +19,12 @@ rustup update
 #   is missing one of the components we need.
 # llvm-tools-preview:
 #   Required by `cargo-llvm-cov` for test coverage
+# miri:
+#   Required to run `cargo miri test` for UB detection
 # rust-src:
 #   Required to build RedisJSON with address sanitizer
 rustup toolchain install nightly \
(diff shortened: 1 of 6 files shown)


hdoordt (dismissed):
Looks good on the Rust side of things. Can't vouch for the correctness of actual integration in CI setup.

.install/install_rust.sh
```diff
@@ -19,9 +19,12 @@ rustup update
...
 # llvm-tools-preview:
 #   Required by `cargo-llvm-cov` for test coverage
+# miri:
+#   Required to run `cargo miri test` for UB detection
 # rust-src:
 #   Required to build RedisJSON with address sanitizer
 rustup toolchain install nightly \
     --allow-downgrade \
     --component llvm-tools-preview \
+    --component miri \
```
DvirDukhan, line 29: here is a general question
since nightly is only for sanitizer, why don't we move those to test deps installation? I assume we don't use nightly features on production code

Makefile
```diff
@@ -458,6 +459,10 @@ endif
 
 rust-tests:
 	$(SHOW)cd $(REDISEARCH_RS_DIR) && $(RUST_TEST_RUNNER) test $(RUST_TEST_OPTIONS) $(TEST_NAME)
+ifeq ($(RUN_MIRI),1)
```
DvirDukhan, line 462: this is the main comment from me on this PR:
from the PR description 
```
You need to opt into running tests through miri, with a dedicated RUN_MIRI option.
In CI it's enabled exclusively for the sanitize check (PR event flow + merge-queue event flow).
```
can you explain why this is not `on` by default when SAN is used so we can avoid additional flag to the makefile?
we can avoid the additional flags in the CI as well

LukeMathWalker: I didn't reuse `SAN` because Rust supports sanitizers (see [here](https://doc.rust-lang.org/stable/unstable-book/compiler-flags/sanitizer.html) for details), and we plan to honor that flag in the future.
Running with sanitizers in Rust is a much heavier/cumbersome operation (since it requires the `build-std` unstable flag) so it's convenient to be able to run `miri` without having to enable sanitizers.

DvirDukhan (approved):

PR #5734, opened 2025-03-09
[2.10] CP TIMEOUT related bugs & Query Debug Mechanism 

meiravgri:
backport https://github.com/RediSearch/RediSearch/pull/5554 to 2.10

### Background  
In **master**, `src/module.c` combines both **coordinator** and **standalone** logic, allowing to implementation of the debug mechanism for both modes in this file. However, in **version 2.10**, we still maintain separate files for each module:  
- **Coordinator:** `coord/module.c`  
- **Standalone:** `src/module.c`  

Similarly, `debug_commands.c` was merged into a single file in master.  

Due to these structural differences, this **cherry-pick had to be done manually** to properly extract coordinator-related code and place it in the correct files.  

### Changes in Coordinator & Standalone Implementations  

| Component                      | Master  | 2.10  |
|--------------------------------|-------------------------|-----------------------------------------|
| **expose `Dist<Aggregate/Search>Command`**       | `src/module.h` | `coord/src/coord_module.h` |
| **handle _FT.DEBUG arg in cluster mode** | `src/module.c` | `coord/src/coord_module.c` |
| **`debug_commands.c`**         | registrater `FT.DEBUG FT.SEARCH/AGGREGATE` && `FT.DEBUG _FT.SEARCH/_FT.AGGREGATE`,  | Registers `FT.DEBUG FT.SEARCH/AGGREGATE` for standalone mode or `_FT.SEARCH/_FT.AGGREGATE` when built with `COORD=1` |
| **`coord/src/debug_command.c`**         | N/A | Registers `FT.DEBUG FT.SEARCH/AGGREGATE` |
| **RPTimeoutAfterCount_SimulateTimeout** | `parent->sctx->time.timeout` | `parent->sctx->timeout` |

- **`GetNumShards_UnSafe` was not cherry-picked**, as it is not in use. Cluster mode is checked testing if `RS_COORDINATOR` macro is defined.

PR #5788, opened 2025-03-18
[2.8] CP TIMEOUT related bugs & Query Debug Mechanism

meiravgri:
### Background  
This PR cherry-picks PR #5734 to 2.8, which previously cherry-picked PR #5554 to 2.10, since 2.10 is structurally closer to 2.8 than master.

However, **2.8 does not support `FT.DEBUG` in coordinator mode**, so the query debug mechanism is **only available in standalone mode**. This required **removing certain coordinator-related code** that was included in the original PR.  

---

#### **Discarded Code**  
- **Coordinator changes**: Removed all modifications to `coord/src/module.c`.  
- **Debug commands**:  
  - Support only **`FT.SEARCH` / `FT.AGGREGATE`** commands.  
  - Removed support for **`_FT.SEARCH` / `_FT.AGGREGATE`** commands (used for coordinator-shard communication).  
- **Distributed aggregation**: Removed split distributed command in `dist_aggregate.c`.  
- **Internal-only options**:  
  - Removed `COORDINATOR_FORCED_TIMEOUT`.  
  - Removed `isClusterCoord`.  

---

### **Test Adjustments**  

| Test Name  | Change |
|------------|--------|
| `testTimeoutPartialWithEmptyResults` | **Removed**: Required simulating a scenario where the **coordinator** reaches `getNextReply` before the timeout elapsed, while the **shard returns empty results**—this is **not possible to control without the debug mechanism**, which 2.8 lacks. |
| `test_resp3: test_error_with_partial_results` | In **cluster**, timeout verification only if document count < expected. In **SA**, stable debug mechanism is used. |
| `test_timeout` | **Skipped on cluster**: Only checks a bug in the sorter, fine to test in SA. |
| `test_debug_commands:TestQueryDebugCommands` | **Skipped on cluster**, added messages, **Removed `_MT` tests** as `WORKER_THREADS` can't be configured after env setup in 2.8, making them messy, and MT isn't critical for this scenario.|
| `verifyResultsResp3` | **Moved from est_debug_commands:TestQueryDebugCommands to common tests**. |

---

### DIFF
--- a/coord/src/dist_aggregate.c
+++ b/coord/src/dist_aggregate.c
@@ -235,23 +235,26 @@ typedef struct {
   int shardsProfileIdx;
 } RPNet;
 
+static void RPNet_resetCurrent(RPNet *nc) {
+    nc->current.root = NULL;
+    nc->current.rows = NULL;
+}
+
 static int getNextReply(RPNet *nc) {
   if (nc->cmd.forCursor) {
     // if there are no more than `clusterConfig.cursorReplyThreshold` replies, trigger READs at the shards.
     // TODO: could be replaced with a query specific configuration
     if (!MR_ManuallyTriggerNextIfNeeded(nc->it, clusterConfig.cursorReplyThreshold)) {
       // No more replies
-      nc->current.root = NULL;
-      nc->current.rows = NULL;
+      RPNet_resetCurrent(nc);
       return 0;
     }
   }
   MRReply *root = MRIterator_Next(nc->it);
   if (root == NULL) {
     // No more replies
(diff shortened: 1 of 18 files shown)


meiravgri:
here as a reminder to CP the follow up PR:
https://github.com/RediSearch/RediSearch/pull/5784#issuecomment-2735403139

alonre24 (approved):
Reviewd compared to the backport to 2.10 and the expected diffs described in the PR description. All looks as expected

GuyAv46 (approved):

PR #5797, opened 2025-03-19
[MOD-8972] - Implement basic TrieMap functionality in Rust

hdoordt:

**Stacked on #5744.**

Implement TrieMap type, with insertion, removal, finding, and basic iteration
Add a TrieMap implementation, that uses `LowMemoryThinVec` to store
node labels and children. This version includes implementations
for insertion, removal, finding, and ordered iteration over entries.
Additionally, `TrieMap::mem_usage` allows one to compute the current
memory usage of the trie.

Furthermore, an implementation of `std::fmt::Debug` is done,
as a means of inspecting the structure of the trie. Combined
with snapshot testing based on the `insta` crate, this allows
developers to verify the correctness of the data structure.

As a sanity check, a `proptest`-based tests is included,
which asserts the behavior of `TrieMap<T>` is identical
to that of `BTreeMap<Vec<c_char>, T>` given a randomized
input of operations.

The type of the `TrieMap` keys is `&[c_char]` to accomodate for
the fact that in C, `char` can either be unsigned or signed, depending
on the target platform, which affects ordering.

Additionally:
- Replaces the placeholder `TrieMap` in `trie_rs::ffi`
with a newtype, wrapping a `trie_rs::trie::TrieMap<*mut c_void>`.
- Forward calls to `NewTrieMap`, `TrieMap_Add`, `TrieMap_Find`,
`TrieMap_Delete`, `TrieMap_Free`, and `TrieMap_MemUsage` to their
counterpart methods of `TrieMap<*mut c_void>`.

This PR does _not_ include further methods of iterating over the trie.
Those are to be addressed in a future PR.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/low_memory_thin_vec/src/lib.rs
+++ b/src/redisearch_rs/low_memory_thin_vec/src/lib.rs
@@ -1386,8 +1386,9 @@ impl<T: Copy + std::fmt::Debug> LowMemoryThinVec<T> {
 
         // SAFETY:
         // `prefix` cannot be a reference to a slice in `self`,
-        // as that would violate borrowing rules. Therefore, we can safely
-        // memcpy the elements from `prefix` to start of the buffer.
+        // as that would violate borrowing rules. Furthermore, `T` is bound to `Copy`
+        // and therefore can simply be copied byte-for-byte and does not implement `Drop`.
+        // Therefore, we can safely memcpy the elements from `prefix` to start of the buffer.
         unsafe {
             std::ptr::copy_nonoverlapping(prefix.as_ptr(), self.data_raw(), prefix_len);
         }
--- a/src/redisearch_rs/low_memory_thin_vec/src/lib.rs
+++ b/src/redisearch_rs/low_memory_thin_vec/src/lib.rs
@@ -393,6 +393,29 @@ impl<T> LowMemoryThinVec<T> {
         self.header_ref().capacity()
     }
 
+    /// Returns the memory usage of the vector on the heap in bytes,
+    /// i.e. the size of the header and the elements, as well as any padding.
+    ///
+    /// Does not take into account any additional memory allocated by elements
+    /// of the vector. Returns 0 if the vector has no capacity.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use low_memory_thin_vec::LowMemoryThinVec;
+    ///
+    /// let vec: LowMemoryThinVec<i32> = LowMemoryThinVec::with_capacity(5);
+    /// assert_eq!(vec.mem_usage(), 24);
+    ///
+    /// assert_eq!(LowMemoryThinVec::<u64>::new().mem_usage(), 0);
+    /// ```
+    pub fn mem_usage(&self) -> usize {
+        if !self.has_allocated() {
+            return 0;
+        }
+        allocation_layout::<T>(self.capacity()).size()
+    }
+
     /// Returns `true` if the vector has allocated any memory via the global
     /// allocator.
     #[inline]
@@ -1287,6 +1310,22 @@ impl<T: Clone> LowMemoryThinVec<T> {
         }
     }
 
+    /// Creates a `LowMemoryThinVec` from a slice, cloning the elements.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use low_memory_thin_vec::LowMemoryThinVec;
+    ///
+    /// let vec = LowMemoryThinVec::from_slice(&[1, 2, 3]);
+    /// assert_eq!(vec, [1, 2, 3]);
+    /// ```
+    pub fn from_slice(slice: &[T]) -> Self {
+        let mut vec = LowMemoryThinVec::with_capacity(slice.len());
+        vec.extend_from_slice(slice);
+        vec
+    }
+
     /// Clones and appends all elements in a slice to the `LowMemoryThinVec`.
     ///
     /// Iterates over the slice `other`, clones each element, and then appends
@@ -1310,6 +1349,52 @@ impl<T: Clone> LowMemoryThinVec<T> {
     /// [`extend`]: LowMemoryThinVec::extend
     pub fn extend_from_slice(&mut self, other: &[T]) {
         self.extend(other.iter().cloned())
+    }
+}
+
+impl<T: Copy + std::fmt::Debug> LowMemoryThinVec<T> {
+    /// Reserves capacity to fit the given `prefix` slice,
+    /// moves the current elements back to make room for the prefix
+    /// and copies the prefix to the front of the vector.
+    pub fn prepend_with_slice(&mut self, prefix: &[T]) {
+        let prefix_len = prefix.len();
+        let self_len = self.len();
+        let new_len = prefix_len + self_len;
+        debug_assert!(new_len <= isize::MAX as usize);
+
+        if prefix.is_empty() {
+            return;
+        }
+
+        if self.is_empty() {
+            self.extend_from_slice(prefix);
+            return;
+        }
+
+        self.reserve(prefix_len);
+        {
+            // SAFETY:
+            // We reserved enough space for an additional `prefix_len`
+            // amount of elements.
+            let dst = unsafe { self.data_raw().add(prefix_len) };
+            // SAFETY:
+            // We have reserved enough space for `prefix_len` elements,
+            // so it is safe to copy the elements from the current vector to the
+            // newly reserved space.
+            unsafe { std::ptr::copy(self.data_raw(), dst, self_len) };
+        }
+
+        // SAFETY:
+        // `prefix` cannot be a reference to a slice in `self`,
+        // as that would violate borrowing rules. Therefore, we can safely
+        // memcpy the elements from `prefix` to start of the buffer.
+        unsafe {
+            std::ptr::copy_nonoverlapping(prefix.as_ptr(), self.data_raw(), prefix_len);
+        }
+        // SAFETY:
+        // We have reserved enough space for `prefix_len` elements,
+        // and all elements have been initialized.
+        unsafe { self.set_len(new_len) };
     }
 }
 
@@ -2038,5 +2123,15 @@ mod tests {
         v.resize(v.len(), 0);
         // No reallocation has taken place.
         assert_eq!(old_ptr, v.as_ptr());
+    }
+
+    #[test]
+    fn test_prepend_with_slice() {
+        let mut v = low_memory_thin_vec![4, 5, 6, 7];
+        v.prepend_with_slice(&[1, 2, 3]);
+        assert_eq!(v, [1, 2, 3, 4, 5, 6, 7]);
+
+        v.prepend_with_slice(&[-1, 0]);
+        assert_eq!(v, [-1, 0, 1, 2, 3, 4, 5, 6, 7]);
     }
 }
--- a/src/redisearch_rs/trie_rs/Cargo.toml
+++ b/src/redisearch_rs/trie_rs/Cargo.toml
@@ -18,7 +18,21 @@ ffi = ["redis_allocator"]
 [dependencies]
 libc.workspace = true
 low_memory_thin_vec.workspace = true
-redis-module = { workspace = true, optional = true }
+
+[target.'cfg(all(target_env="musl", target_os="linux"))'.dependencies.redis-module]
+# Statically link to the libclang on aarch64-unknown-linux-musl,
+# necessary on Alpine.
+# See https://github.com/rust-lang/rust-bindgen/issues/2360
+features = ["bindgen-static", "min-redis-compatibility-version-6-0"]
+workspace = true
+optional = true
+default-features = false
+
+
+[target.'cfg(not(all(target_env="musl", target_os="linux")))'.dependencies.redis-module]
+workspace = true
+optional = true
+default-features = true
 
 [dev-dependencies]
 insta.workspace = true
--- a/src/redisearch_rs/trie_rs/Cargo.toml
+++ b/src/redisearch_rs/trie_rs/Cargo.toml
@@ -16,3 +16,9 @@ redis_allocator = []
 
 [dependencies]
 libc.workspace = true
+low_memory_thin_vec = { path = "../low_memory_thin_vec" }
+
+[dev-dependencies]
+insta = "1.42.2"
+proptest = { version = "1.6.0", default-features = false, features = ["std"] }
+proptest-derive = { version = "0.5.1", default-features = false }
--- a/src/redisearch_rs/trie_rs/Cargo.toml
+++ b/src/redisearch_rs/trie_rs/Cargo.toml
@@ -12,11 +12,13 @@ crate-type = ["staticlib"]
 
 [features]
 default = ["redis_allocator"]
-redis_allocator = []
+redis_allocator = ["dep:redis-module"]
+ffi = ["redis_allocator"]
 
 [dependencies]
 libc.workspace = true
 low_memory_thin_vec = { path = "../low_memory_thin_vec" }
+redis-module = { workspace = true, optional = true }
 
 [dev-dependencies]
 insta = "1.42.2"
--- a/src/redisearch_rs/trie_rs/Cargo.toml
+++ b/src/redisearch_rs/trie_rs/Cargo.toml
@@ -17,7 +17,7 @@ ffi = ["redis_allocator"]
 
 [dependencies]
 libc.workspace = true
-low_memory_thin_vec = { path = "../low_memory_thin_vec" }
+low_memory_thin_vec.workspace = true
 redis-module = { workspace = true, optional = true }
 
 [dev-dependencies]
--- a/src/redisearch_rs/trie_rs/Cargo.toml
+++ b/src/redisearch_rs/trie_rs/Cargo.toml
@@ -21,6 +21,6 @@ low_memory_thin_vec = { path = "../low_memory_thin_vec" }
 redis-module = { workspace = true, optional = true }
 
 [dev-dependencies]
-insta = "1.42.2"
-proptest = { version = "1.6.0", default-features = false, features = ["std"] }
-proptest-derive = { version = "0.5.1", default-features = false }
+insta.workspace = true
+proptest = { workspace = true, features = ["std"] }
+proptest-derive.workspace = true
--- a/src/redisearch_rs/trie_rs/src/ffi.rs
+++ b/src/redisearch_rs/trie_rs/src/ffi.rs
@@ -152,12 +152,8 @@ type TrieMapReplaceFunc =
 /// Add a new string to a trie. Returns 1 if the key is new to the trie or 0 if
 /// it already existed.
 ///
-/// If value is given, it is saved as a payload inside the trie node.
-/// If the key already exists, we replace the old value with the new value, using
-/// free() to free the old value.
-///
-/// If cb is given, instead of replacing and freeing, we call the callback with
-/// the old and new value, and the function should return the value to set in the
+/// If `cb` is given, instead of replacing and freeing the value using `rm_free`,
+/// we call the callback with the old and new value, and the function should return the value to set in the
 /// node, and take care of freeing any unwanted pointers. The returned value
 /// can be NULL and doesn't have to be either the old or new value.
 ///
@@ -168,6 +164,9 @@ type TrieMapReplaceFunc =
 ///  - `str` can be NULL only if `len == 0`. It is not necessarily NULL-terminated.
 ///  - `len` can be 0. If so, `str` is regarded as an empty string.
 ///  - `value` holds a pointer to the value of the record, which can be NULL
+///  - `cb` must not free the value it returns
+///  - The Redis allocator must be initialized before calling this function,
+///    and `RedisModule_Free` must not get mutated while running this function.
 ///
 /// C equivalent:
 /// ```c
@@ -202,21 +201,35 @@ unsafe extern "C" fn TrieMap_Add(
         &[]
     };
 
-    let value = match cb {
-        Some(cb) => {
-            if let Some(old) = trie.find(key) {
-                // SAFETY: The following line is sound if the cb implementation
-                // does not introduce undefined behavior.
-                unsafe { cb(*old, value) }
+    let mut was_vacant = true;
+    trie.insert_with(key, |old| {
+        if let Some(old_value) = old {
+            was_vacant = false;
+            if let Some(cb) = cb {
+                // SAFETY: The safety requirements of this function
+                // require `cb` has the correct signature and does
+                // not free the value it returns.
+                unsafe { cb(old_value, value) }
             } else {
+                // SAFETY:
+                // The safety requirements of this function
+                // require the caller to ensure that the Redis allocator is initialized,
+                // and that `RedisModule_Free` does not get mutated while running this function.
+                let rm_free = unsafe {
+                    redis_module::raw::RedisModule_Free.expect("Redis allocator not available")
+                };
+                // SAFETY:
+                // The safety requirements of this function
+                // require the caller to ensure that the Redis allocator is properly initialized.
+                unsafe { rm_free(old_value) };
                 value
             }
+        } else {
+            value
         }
-        None => value,
-    };
+    });
 
-    let was_vacant = trie.insert(key, value).is_none();
-    was_vacant as _
+    if was_vacant { 1 } else { 0 }
 }
 
 /// Find the entry with a given string and length, and return its value, even if
--- a/src/redisearch_rs/trie_rs/src/ffi.rs
+++ b/src/redisearch_rs/trie_rs/src/ffi.rs
@@ -1,6 +1,9 @@
 #![allow(non_camel_case_types, non_snake_case)]
 
-use std::ffi::{c_char, c_int, c_void};
+use std::{
+    ffi::{c_char, c_int, c_void},
+    slice,
+};
 
 /// Holds the length of a key string in the trie.
 ///
@@ -10,7 +13,7 @@ use std::ffi::{c_char, c_int, c_void};
 /// ```
 pub type tm_len_t = u16;
 
-/// This special pointer is returned when TrieMap_Find cannot find anything.
+/// This special pointer is returned when [`TrieMap_Find`] cannot find anything.
 ///
 /// C equivalent:
 /// ```c
@@ -46,7 +49,8 @@ enum tm_iter_mode {
 static TM_ITER_MODE_DEFAULT: tm_iter_mode = tm_iter_mode::TM_PREFIX_MODE;
 
 /// Opaque type TrieMap. Can be instantiated with [`NewTrieMap`].
-pub enum TrieMap {}
+#[repr(transparent)]
+struct TrieMap(crate::trie::TrieMap<*mut c_void>);
 
 /// Callback type for passing to [`TrieMap_IterateRange`].
 ///
@@ -124,7 +128,8 @@ unsafe extern "C" fn TrieMapResultBuf_Data(buf: *mut TrieMapResultBuf) -> *mut *
 /// ```
 #[unsafe(no_mangle)]
 unsafe extern "C" fn NewTrieMap() -> *mut TrieMap {
-    todo!()
+    let map = Box::new(TrieMap(crate::trie::TrieMap::new()));
+    Box::into_raw(map)
 }
 
 /// Callback type for passing to [`TrieMap_Delete`].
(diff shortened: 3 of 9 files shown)


src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -1313,6 +1352,52 @@ impl<T: Clone> LowMemoryThinVec<T> {
...
+            // We have reserved enough space for `prefix_len` elements,
+            // so it is safe to copy the elements from the current vector to the
+            // newly reserved space.
+            unsafe { std::ptr::copy(self.data_raw(), dst, self_len) };
+        }
+
+        // SAFETY:
+        // `prefix` cannot be a reference to a slice in `self`,
+        // as that would violate borrowing rules. Therefore, we can safely
+        // memcpy the elements from `prefix` to start of the buffer.
```
LukeMathWalker, line 1390: We should note that this is only safe due to the fact that `T` is `Copy`.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -1313,6 +1352,52 @@ impl<T: Clone> LowMemoryThinVec<T> {
     }
 }
 
+impl<T: Copy + std::fmt::Debug> LowMemoryThinVec<T> {
+    /// Reserves capacity to fit the given `prefix` slice,
+    /// moves the current elements back to make room for the prefix
+    /// and copies the prefix to the front of the vector.
+    pub fn prepend_with_slice(&mut self, prefix: &[T]) {
```
LukeMathWalker, line 1359: `prepend_slice`, perhaps, rather than `prepend_with_slice`? 

Feel free to ignore the suggestion.

src/redisearch_rs/trie_rs/Cargo.toml
```diff
@@ -16,3 +16,9 @@ redis_allocator = []
 
 [dependencies]
 libc.workspace = true
+low_memory_thin_vec = { path = "../low_memory_thin_vec" }
+
+[dev-dependencies]
+insta = "1.42.2"
+proptest = { version = "1.6.0", default-features = false, features = ["std"] }
+proptest-derive = { version = "0.5.1", default-features = false }
```
LukeMathWalker, line 24: Can we centralize all version dependencies in the workspace manifest? It'll make it easier to manage in the long run.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
hdoordt: I felt `prepend_slice` is a bit ambiguous, so I'm keeping it as is.

src/redisearch_rs/trie_rs/src/ffi.rs
```diff
@@ -181,8 +186,37 @@ unsafe extern "C" fn TrieMap_Add(
...
+    } else {
+        &[]
+    };
+
+    let value = match cb {
+        Some(cb) => {
+            if let Some(old) = trie.find(key) {
+                // SAFETY: The following line is sound if the cb implementation
+                // does not introduce undefined behavior.
+                unsafe { cb(*old, value) }
```
LukeMathWalker, line 210: Can `cb` free `old`? It's not clear to me reading the function documentation.

src/redisearch_rs/trie_rs/src/ffi.rs
```diff
@@ -295,9 +385,17 @@ unsafe extern "C" fn TrieMap_Delete(
...
+    // Reconstruct the original Box<TrieMap> which will take care of freeing the memory
+    // upon dropping.
+    // SAFETY: The safety requirements of this function
+    // state the caller is to ensure that the pointer `t` is
+    // a valid TrieMap obtained from `NewTrieMap` and cannot be NULL.
+    // If that invariant is upheld, then the following line is sound.
+    let trie = unsafe { Box::from_raw(t) };
+    if let Some(func) = func {
+        let _unused = (func, trie);
+        todo!("Iterate over all nodes and free them by calling `func` given the data.");
```
LukeMathWalker, line 397: In a follow-up PR?

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,844 @@
...
+    }
+
+    impl<const N: usize> ToCCharArray<N> for [u8; N] {
+        fn c_chars(self) -> [c_char; N] {
+            self.map(|b| b as c_char)
+        }
+    }
+
+    /// Forwards to `insta::assert_debug_snapshot!`,
+    /// but is disabled in Miri.
```
LukeMathWalker, line 594: Let's capture why we need to disable it for `miri`.

src/redisearch_rs/trie_rs/src/ffi.rs
```diff
@@ -181,8 +186,37 @@ unsafe extern "C" fn TrieMap_Add(
...
+                // does not introduce undefined behavior.
+                unsafe { cb(*old, value) }
+            } else {
+                value
+            }
+        }
+        None => value,
+    };
+
+    let was_vacant = trie.insert(key, value).is_none();
```
kei-nan, line 218: This seems a bit wasteful, no?
We search and then insert if we don't find?
Can't we do a try_insert version instead and then handle the case it already existed??

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,844 @@
...
+    /// The portion of the key attached to this node.
+    ///
+    /// # Invariants
+    ///
+    /// - `label` can only be empty for the root node.
+    label: LowMemoryThinVec<c_char>,
+}
+
+impl<Data: fmt::Debug> Node<Data> {
+    pub fn insert(&mut self, key: &[c_char], data: Data) -> Option<Data> {
```
kei-nan, line 114: Does rust have tail recursion optimization?
Does every call generate a new stack frame?

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,844 @@
...
+            //
+            // Create a new root node with an empty label,
+            // insert the old root as a child of the new root,
+            // and add a new child to the empty root.
+            let new_child = Node {
+                children: ChildRefs::default(),
+                data: Some(data),
+                label: LowMemoryThinVec::from_slice(key),
+            };
+            let children = ChildRefs(low_memory_thin_vec![ChildRef {
```
kei-nan, line 229: maybe add the old root here along with the new child?

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,844 @@
...
+        let child = &mut self.children.0[child_index].node;
+
+        // If the child's label is equal to the key, we remove the child.
+        if child.label == key {
+            let data = child.data.take();
+
+            let child_is_leaf = child.children.is_empty();
+            // If there's a single grandchild,
+            // we merge the grandchild into the child.
+            child.merge_child();
```
kei-nan, line 295: you call merge_child even if the child is leaf?

src/redisearch_rs/trie_rs/src/ffi.rs
hdoordt: It is. Clarified in 30793852dbaa5ec7c6de3ddffe3eb6e30d9fa62b.

src/redisearch_rs/trie_rs/src/ffi.rs
```diff
@@ -181,8 +186,37 @@ unsafe extern "C" fn TrieMap_Add(
...
+    was_vacant as _
```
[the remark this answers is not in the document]
hdoordt: Fixed in 30793852dbaa5ec7c6de3ddffe3eb6e30d9fa62b

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,844 @@
...
+        old_parent.label = old_parent_suffix;
+
+        self.children
+            .find_or_insert(old_parent.label[0], move || old_parent);
+        None
+    }
+
+    /// Remove a child from the node.
+    /// Returns the data associated with the key if it was present.
+    fn remove_child(&mut self, key: &[c_char]) -> Option<Data> {
```
kei-nan, line 281: shouldn't you check key is not empty?

src/redisearch_rs/trie_rs/src/ffi.rs
hdoordt: Good call. Fixed in 30793852dbaa5ec7c6de3ddffe3eb6e30d9fa62b.

src/redisearch_rs/trie_rs/src/trie.rs
hdoordt: Fixed in 4ca45094e575cd88387bc59e3fa7918201237a8e.

src/redisearch_rs/trie_rs/Cargo.toml
hdoordt: Fixed in aec5bec92ed88557c63b7c2d8ecd6754748fc2fb.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
hdoordt: Fixed in https://github.com/RediSearch/RediSearch/commit/4ca45094e575cd88387bc59e3fa7918201237a8e.

src/redisearch_rs/trie_rs/src/trie.rs
hdoordt: It does, though there are no guarantees that this optimization is actually applied. I suggest we benchmark before optimizing further.

src/redisearch_rs/trie_rs/src/trie.rs
hdoordt: `merge_child` will only actually merge if the child has exactly one child. That said, d6b5c604c4dfdecd9eea8ad743b513be52b02e96 should clarify things

src/redisearch_rs/trie_rs/src/trie.rs
hdoordt: This is being done in the following line:

```rust
let child_index = self.children.index_of(*key.first()?)?;
```

In particular , `?`-operator:

```rust
key.first()?
```

performs an early return if `first()` returns `None`, i.e. if `key` is empty.

That said, better clarify that in a comment. Fixed in d6b5c604c4dfdecd9eea8ad743b513be52b02e96

src/redisearch_rs/trie_rs/src/trie.rs
hdoordt: Fixed in a5a467093

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,891 @@
...
+        &mut self,
+        first_byte: c_char,
+        create_node: CreateNode,
+    ) -> &mut Node<Data>
+    where
+        CreateNode: FnOnce() -> Node<Data>,
+    {
+        let index = match self
+            .0
+            .binary_search_by_key(&first_byte, |child| child.first_byte)
```
kei-nan, line 516: A byte can have 256 possible values.
At what point would it be better to use a table of 256 and with O(1) access the correct child?
Maybe using a bitmask to represent which children are present might allow for faster lookup.
Just a thought.
If a binary search is the approach the c code took then all is well.

hdoordt: Very interesting idea.

The [C implementation](https://github.com/RediSearch/RediSearch/blob/master/deps/triemap/triemap.c#L249-L260) seems to call `memchr` to find the child's offset.

I would expect whether it's worth the added complexity and memory use greatly depends on the data. If the node labels tend to be short, then it might not be worth it.

src/redisearch_rs/trie_rs/src/lib.rs
```diff
@@ -1,12 +1,11 @@
+#[cfg(feature = "ffi")]
 pub mod ffi;
```
hdoordt, line 2:
```suggestion
mod ffi;
```

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,891 @@
...
+        }
+
+        // In this case, the node and the key do share a common prefix.
+        //
+        // Create a new node that uses the shared prefix as its label.
+        // The prefix is then stripped from both the current label and the
+        // new key; the resulting suffixes are used as labels for the new child nodes.
+        let old_parent_suffix = LowMemoryThinVec::from_slice(&self.label[equal_up_to..]);
+        let child_suffix = LowMemoryThinVec::from_slice(&key[equal_up_to..]);
+        let shared_prefix = {
```
kei-nan, line 251: what is the difference of `shared_prefix` from `old_parent_suffix`?
Won't they contain the same value?
Why do you need both?
Won't the self.label be empty after you take from it?

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,891 @@
...
+        data
+    }
+
+    /// If `self` has exactly one child, and `self` doesn't hold
+    /// any data, merge child into `self`, by moving the child's data and
+    /// childreninto `self`. Depending on the spare capacity of `self`s label
+    /// and that of the child, we either extend `self`s label with the child's label,
+    /// or vice versa.
+    fn merge_child(&mut self) {
+        if self.data.is_some() {
```
kei-nan, line 344: why is it okay to silently fail/not merge if the node has data?

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,891 @@
...
+    /// involves file I/O, which is not supported in Miri.
+    macro_rules! assert_debug_snapshot {
+        ($($arg:tt)*) => {
+            #[cfg(not(miri))]
+            insta::assert_debug_snapshot!($($arg)*);
+        };
+    }
+
+    #[test]
+    fn test_trie() {
```
kei-nan, line 646: consider testing with unicode strings as well.

hdoordt: This is implicitly covered by the `proptest`, which uses random byte sequences as keys

src/redisearch_rs/trie_rs/src/trie.rs
hdoordt: In this case, the current node (which is the old parent, to be replaced with a new one) shares a prefix with the key. That means we need to create a new parent node that has that shared prefix as a label, and assign it the current node (i.e. the old parent) as a child. We use `std::mem::swap` to swap out the current node with the new parent, then add that old parent as a child to the new parent, along with the newly created node (`new_child`).

> what is the difference of shared_prefix from old_parent_suffix? Won't they contain the same value?
Why do you need both?
`shared_prefix` is the prefix that is being shared between the key and the current node (old parent). `old_parent_suffix` is the current node (old parent)'s label, but with that share prefix removed.


> Won't the self.label be empty after you take from it?
Yes, but it's assigned to `old_parent` after the swap, and the `shared_prefix` is assigned to `new_parent` before the swap.

src/redisearch_rs/trie_rs/src/trie.rs
kei-nan: I think it makes the code less clear.
I think the if should be removed from this function.
maybe a debug assert can be in its place.

perhaps create a `try_merge_child` function which checks the condition where the data is not empty and only then does it calls `merge_child`.

src/redisearch_rs/trie_rs/src/trie.rs
kei-nan: Re-read the code, will say the code is not trivial.
The variable names seem to imply different things.
Maybe using `old_self` and `new_self` would help clarify what is going on.
Just a suggestion.

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,891 @@
...
+
+    /// Check whether the children vector is empty.
+    fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+}
+
+/// The reference to the child node held inside its parent node.
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct ChildRef<Data> {
```
kei-nan, line 560: Do you know how this will interact with the repr(C) thin vec uses?
Will Rust be allowed to move c_char after node in the memory alignment for example?

hdoordt: The `#[repr(C)]` attribute on `LowMemoryThinVec` and its `Header` don't have any effect on the data the vector stores. Therefore, Rust will have the liberty to reorder the fields. In the end though, as the `ChildRef`s are stored in a contiguous buffer that is the `LowMemoryThinVec`, there will be padding either within the struct or between items, and that padding will be 7 bytes in any case.

src/redisearch_rs/trie_rs/src/trie.rs
hdoordt: Clarified in 9a3b257efe2c029278a3e2ed69345fbdde937a1a

src/redisearch_rs/trie_rs/src/trie.rs
hdoordt: Clarified in 9a3b257efe2c029278a3e2ed69345fbdde937a1a

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,917 @@
...
+    }
+
+    fn label_to_string_lossy(&self) -> String {
+        let slice = self.label.iter().map(|&c| c as u8).collect::<Vec<_>>();
+        String::from_utf8_lossy(&slice).into_owned()
+    }
+}
+
+impl<Data: fmt::Debug> fmt::Debug for Node<Data> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
alonre24, line 473: Cool :)
Note for the future: We should probably use this for debugability (today we have an implementation  in our debug command that is using iterator, but I believe that this "raw" scan is even better

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,917 @@
...
+            NODE_SIZE // root node
+            + thin_vec_mem_usage_for_cap::<ChildRef<i32>>(2) // 2 children
+            + 10 // Label b"hello": header (4) + padding(0) + capacity (5 * 1) + padding (1)
+            + 10, // Label b"world": header (4) + padding(0) + capacity(5 * 1) + padding(1)
+            "The size of the trie should equal the size of 1 `Node<i32>`, 1 `ChildRef<i32>` for the
+            root's child, and 1 label with a capacity of 5 elements and aligned to 2 bytes"
+        );
+    }
+
+    #[derive(proptest_derive::Arbitrary, Debug)]
```
alonre24, line 878: Where is the definition of the flow that runs here? How many operations are there, and how are they generated?

alonre24 (commented):
Very nice work!!
I mainly have questions just to make sure that I understand how things work, but overall all looks good, clean and well tested :)

src/redisearch_rs/trie_rs/src/ffi.rs
```diff
@@ -216,7 +267,36 @@ unsafe extern "C" fn TrieMap_Find(
...
+    // the intention to mimic the API of the original C implementation.
+    #[allow(static_mut_refs)]
+    // SAFETY: TRIEMAP_NOTFOUND is a pointer to a static mut `c_void`.
+    // It is only referred to by this function and is not available outside this module,
+    // except through the `extern void * TRIEMAP_NOTFOUND`.
+    // The caller is responsible for ensuring that the returned pointer is not dereferenced
+    // in case it is equal to TRIEMAP_NOTFOUND.
+    let value = *trie.find(key).unwrap_or(unsafe { &TRIEMAP_NOTFOUND });
+
+    value
```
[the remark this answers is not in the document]
kei-nan: Since we don't want to change the c code we need to provide the same behaviour as the c implementation.

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,917 @@
...
+            + self.children.0.mem_usage()
+            + self
+                .children
+                .0
+                .iter()
+                .map(|c| c.node.mem_usage())
+                .sum::<usize>()
+    }
+
+    fn label_to_string_lossy(&self) -> String {
```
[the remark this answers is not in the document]
hdoordt: This attempts to convert a slice of bytes to a UTF-8-encoded String. Not all bytes slices are valid UTF-8, and this function defers to `String::from_utf8_lossy` to do the validation and replace any [non-UTF-8 characters with the replacement character](https://doc.rust-lang.org/stable/std/char/constant.REPLACEMENT_CHARACTER.html). In that sense, it loses information.

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,917 @@
...
+    /// // Splitting at index 1
+    /// bi (A)   ->    b (-)
+    ///   \             \
+    ///    ke (B)        i (A)
+    ///                   \
+    ///                    ke (B)
+    /// ```
+    fn split(&mut self, index: usize) {
+        let suffix = LowMemoryThinVec::from_slice(&self.label.as_slice()[index..]);
+        let suffix_first_byte = suffix[0];
```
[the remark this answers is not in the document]
hdoordt: This is going to panic if the `0` is out of bounds, i.e. if the suffix is empty, even when optimizations are enabled, unless the compiler is able to prove that this function is never called with empty labels. Splitting an empty label is definitely a bug, so panicking here is good.

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,917 @@
...
+            return;
+        }
+        if self.children.is_single() {
+            let child = self.children.remove(0);
+
+            let self_label_fits_child_label =
+                self.label.capacity() - self.label.len() >= child.label.len();
+
+            let child_label_fits_self_label =
+                || child.label.capacity() - child.label.len() >= self.label.len();
```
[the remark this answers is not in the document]
hdoordt: `self_label_fits_child_label` needs to be evaluated always, but `child_label_fits_self_label` only if `self_label_fits_child_label` is `false`. This way, we're ensuring `child_label_fits_self_label` is not eagerly evaluated.

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,917 @@
...
+        data
+    }
+
+    /// If `self` has exactly one child, and `self` doesn't hold
+    /// any data, merge child into `self`, by moving the child's data and
+    /// childreninto `self`. Depending on the spare capacity of `self`s label
+    /// and that of the child, we either extend `self`s label with the child's label,
+    /// or vice versa.
+    fn merge_child_if_possible(&mut self) {
+        if self.data.is_some() {
```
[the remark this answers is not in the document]
hdoordt: If we have a trie with entries `"bike": A`, `"biker": B` and `"bikes": C`, and we remove `"bikes"`, we cannot actually merge as "bike" holds data. In case we later on remove "bike", then "bike" and "biker" _can_ be merged.

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,917 @@
...
+            //
+            // ```text
+            // bi (A)   ->    b (C)
+            //   \             \
+            //    ke (B)        i (A)
+            //                   \
+            //                    ke (B)
+            // ```
+            self.split(key.len());
+            self.data = Some(f(None));
```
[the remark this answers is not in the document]
hdoordt: No, because the signature of the callback is such that it must always return a value

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,917 @@
...
+                return;
+            };
+
+            // Suffix is not empty, therefore the insertion needs to happen
+            // in a child (or grandchild) of the current node.
+            // `find_or_insert` will determine if it becomes a new direct
+            // child or if we already have a child with the same first byte.
+            let child = self.children.find_or_insert(*first_byte, || Node {
+                children: ChildRefs::new(),
+                data: None,
```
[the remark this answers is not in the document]
hdoordt: The data will be set when calling `child.insert_or_replace_with(suffix, f)`

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,894 @@
...
+                // Replace the data attached to the current node
+                // with the new data.
+                let current = self.data.take();
+                let new = f(current);
+                self.data.replace(new);
+                return;
+            };
+
+            // Suffix is not empty, therefore the insertion needs to happen
+            // in a child (or grandchild) of the current node.
```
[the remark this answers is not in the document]
hdoordt:
```suggestion
            // in a descendant of the current node.
```

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,917 @@
...
+        // After removing the child, we attempt to merge the child into the root.
+        root.merge_child_if_possible();
+        data
+    }
+
+    /// Get a reference to the value associated with a key.
+    /// Returns `None` if the no entry for the key is present.
+    pub fn find(&self, key: &[c_char]) -> Option<&T> {
+        self.root.as_ref().and_then(|n| n.find(key))
+    }
```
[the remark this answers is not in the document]
hdoordt: > But what if the trie map has no data (serves as a trie set basically), and we want to be able to differentiate between 2.1 and 2.2? Should this use case of "trie set" be part of what we support in this crate, or will it come in next PRs?

In that case, `Data` will be `()`, i.e. `type TrieSet = TrieMap<()>`. Therefore, the type of `Node::data` will be `Option<()>`, still allowing us to differentiate.

src/redisearch_rs/trie_rs/src/ffi.rs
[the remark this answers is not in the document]
hdoordt: I agree with @kei-nan. I'm not a fan of the sentinel pointer, but we want to make the replacement as smooth as possible and should therefore stick to this API.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -2039,4 +2125,14 @@ mod tests {
...
+
+    #[test]
+    fn test_prepend_with_slice() {
+        let mut v = low_memory_thin_vec![4, 5, 6, 7];
+        v.prepend_with_slice(&[1, 2, 3]);
+        assert_eq!(v, [1, 2, 3, 4, 5, 6, 7]);
+
+        v.prepend_with_slice(&[-1, 0]);
+        assert_eq!(v, [-1, 0, 1, 2, 3, 4, 5, 6, 7]);
+    }
```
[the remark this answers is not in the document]
hdoordt: That is being done implicitly by the equality implementation.

src/redisearch_rs/low_memory_thin_vec/src/lib.rs
```diff
@@ -1313,6 +1352,53 @@ impl<T: Clone> LowMemoryThinVec<T> {
...
 
+impl<T: Copy + std::fmt::Debug> LowMemoryThinVec<T> {
+    /// Reserves capacity to fit the given `prefix` slice,
+    /// moves the current elements back to make room for the prefix
+    /// and copies the prefix to the front of the vector.
+    pub fn prepend_with_slice(&mut self, prefix: &[T]) {
+        let prefix_len = prefix.len();
+        let self_len = self.len();
+        let new_len = prefix_len + self_len;
+        debug_assert!(new_len <= isize::MAX as usize);
```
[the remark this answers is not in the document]
hdoordt: Correct, `debug_assert` are no-ops in release mode.

src/redisearch_rs/trie_rs/src/ffi.rs
```diff
@@ -147,12 +152,8 @@ type TrieMapReplaceFunc =
 /// Add a new string to a trie. Returns 1 if the key is new to the trie or 0 if
 /// it already existed.
 ///
-/// If value is given, it is saved as a payload inside the trie node.
```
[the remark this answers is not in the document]
hdoordt: The removed part of the comments was incorrect for `TrieMap_Add`. Payloads are still supported, i.e. they are the void pointers stored in the nodes (`struct TrieMap(crate::trie::TrieMap<*mut c_void>);`)

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -0,0 +1,917 @@
...
+        // ```text
+        //     bike (A)   ->      bi (-)
+        //     /  \             /       \
+        // r (B)   s (C)      ke (A)     s (D)
+        //                   /     \
+        //                 r (B)   s (C)
+        // ```
+        let Some((equal_up_to, _)) = self
+            .label
+            .iter()
```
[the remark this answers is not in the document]
alonre24: I understand now, thanks for the explanation!

src/redisearch_rs/trie_rs/src/trie.rs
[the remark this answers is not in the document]
alonre24: I understand now. I was confused to think that `self` if the node corresponding to the deleted string.

src/redisearch_rs/trie_rs/src/trie.rs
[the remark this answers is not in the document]
hdoordt: Clarified in c06d9de1f53c2808584f1116df33a26283fcfa28

alonre24 (dismissed):
:+1:

kei-nan (dismissed):

hdoordt:
@kei-nan Can you give me a pointer as to why this PR gets removed from the merge queue?

kei-nan:
> @kei-nan Can you give me a pointer as to why this PR gets removed from the merge queue?

Looks like there was a panic in one of the environments:
```thread 'main' panicked at /github/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bindgen-0.66.1/lib.rs:604:31:```
see:
https://github.com/RediSearch/RediSearch/actions/runs/14195322024/job/39769042587

LukeMathWalker (dismissed):

LukeMathWalker (approved):

PR #5804, opened 2025-03-20
[2.10] MOD-9054: Fix `total_results` field in shard RESP3 response (#5765)

raz-mon:
CP of #5765 to `2.10`

### DIFF
--- a/tests/pytests/test_resp3.py
+++ b/tests/pytests/test_resp3.py
@@ -1531,3 +1531,33 @@ def test_warning_maxprefixexpansions():
       if shard_response.get('Warning', 'None') == 'Max prefix expansions limit was reached':
           n_warnings += 1
     env.assertEqual(n_warnings, 1)
+
+# TODO: `total_results` is currently not accurate on cluster - to be fixed in MOD-9094
+@skip(cluster=True)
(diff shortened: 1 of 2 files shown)


tests/pytests/test_resp3.py
```diff
@@ -1531,3 +1531,33 @@ def test_warning_maxprefixexpansions():
...
+
+  # Test that the `total_results` field is correct
+  res = env.cmd('FT.AGGREGATE', 'idx', '*')
+  env.assertEqual(res['total_results'], n_docs)
+
+  # Test the `total_results` field for a cursor
+  res, cid = env.cmd('FT.AGGREGATE', 'idx', '*', 'WITHCURSOR', 'COUNT', '5')
+  while cid:
+    env.assertEqual(res['total_results'], 5)
+    res, cid = env.cmd('FT.CURSOR', 'READ', 'idx', cid)
```
kei-nan, line 1560: isn't it possible that the read command would return less than 5 results?

raz-mon: No, we are using the default strict timeout policy and we have a total number of results that is divided by 5 with no remainder

kei-nan (approved):

PR #5818, opened 2025-03-24
MOD-8674: Fix TLS connection to shards

raz-mon:
This PR fixes a bug in which we connected to the shards from the coordinator with TLS connections depending on whether TLS ports existed in the server configuration or not, instead of depending on the Redis `tls-cluster` config param.

Overall the fix of the condition upon which we connect to our coordinator and shards with TLS on OSS-cluster ONLY:
* Old condition: Whether there are any TLS ports configured for the cluster.
* New condition: Whether or not the `tls-cluster` config param is true or false.

From Redis Documentation:
> Cluster
When Redis Cluster is used, use tls-cluster yes in order to enable TLS for the cluster bus and cross-node connections.

From [link](https://redis.io/docs/latest/operate/oss_and_stack/management/security/encryption/#:~:text=to%20the%20master.-,Cluster,to%20enable%20TLS%20for%20the%20cluster%20bus%20and%20cross%2Dnode%20connections.,-Sentinel).

### DIFF
--- a/src/coord/rmr/conn.c
+++ b/src/coord/rmr/conn.c
@@ -525,11 +525,16 @@ static int checkTLS(char** client_key, char** client_cert, char** ca_cert, char*
   char* tlsPort = NULL;
 
   // If `tls-cluster` is not set to `yes`, we do not connect to the other nodes
-  // with TLS.
+  // with TLS on OSS-cluster. On Enterprise, we always want to connect with TLS
+  // when the tls-port is set to a non-zero value, since this is the port we
+  // get from the proxy.
   clusterTls = getRedisConfigValue(ctx, "tls-cluster");
   if (!clusterTls || strcmp(clusterTls, "yes")) {
-    ret = 0;
-    goto done;
+    tlsPort = getRedisConfigValue(ctx, "tls-port");
+    if (!IsEnterprise() || !tlsPort || !strcmp(tlsPort, "0")) {
+      ret = 0;
+      goto done;
+    }
   }
 
   *client_key = getRedisConfigValue(ctx, "tls-key-file");
--- a/src/coord/rmr/conn.c
+++ b/src/coord/rmr/conn.c
@@ -524,13 +524,12 @@ static int checkTLS(char** client_key, char** client_cert, char** ca_cert, char*
   char* clusterTls = NULL;
   char* tlsPort = NULL;
 
+  // If `tls-cluster` is not set to `yes`, we do not connect to the other nodes
+  // with TLS.
   clusterTls = getRedisConfigValue(ctx, "tls-cluster");
   if (!clusterTls || strcmp(clusterTls, "yes")) {
-    tlsPort = getRedisConfigValue(ctx, "tls-port");
-    if (!tlsPort || !strcmp(tlsPort, "0")) {
-      ret = 0;
-      goto done;
-    }
+    ret = 0;
+    goto done;
   }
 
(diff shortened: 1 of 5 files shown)


src/coord/rmr/conn.c
```diff
@@ -524,13 +524,12 @@ static int checkTLS(char** client_key, char** client_cert, char** ca_cert, char*
   char* clusterTls = NULL;
   char* tlsPort = NULL;
 
+  // If `tls-cluster` is not set to `yes`, we do not connect to the other nodes
+  // with TLS.
   clusterTls = getRedisConfigValue(ctx, "tls-cluster");
   if (!clusterTls || strcmp(clusterTls, "yes")) {
-    tlsPort = getRedisConfigValue(ctx, "tls-port");
```
GuyAv46, line 529: so previously we checked that `tls-cluster==yes` **OR** `tls-port` is defined and non-zero, and now we don't care about the port configuration and only rely on the `tls-cluster` configuration, right? Consider rephrasing the description to be a bit more clear

src/coord/rmr/conn.c
```diff
@@ -524,13 +524,12 @@ static int checkTLS(char** client_key, char** client_cert, char** ca_cert, char*
   char* clusterTls = NULL;
   char* tlsPort = NULL;
 
+  // If `tls-cluster` is not set to `yes`, we do not connect to the other nodes
+  // with TLS.
   clusterTls = getRedisConfigValue(ctx, "tls-cluster");
   if (!clusterTls || strcmp(clusterTls, "yes")) {
```
GuyAv46, line 530: Necessarily lowercase? Should we switch to `strcasecmp`?

src/coord/rmr/conn.c
```diff
@@ -524,13 +524,12 @@ static int checkTLS(char** client_key, char** client_cert, char** ca_cert, char*
   char* clusterTls = NULL;
   char* tlsPort = NULL;
```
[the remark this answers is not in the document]
raz-mon: Got back to us thanks to Enterprise..

src/coord/rmr/conn.c
raz-mon: I'm quite sure Redis works only with lowercase values

src/coord/rmr/conn.c
raz-mon: Right.
But now this is true only for OSS, on Enterprise we stay with the same logic so we won't break.
I buffed up the comments a bit, is that what you meant in "description"?

GuyAv46: The PR description :). I that we move from **only** checking `tls-port` to **only** checking `tls-cluster`

raz-mon: That's exactly what it says 🤔 
anyways, added the update that the change is for OSS only

MeirShpilraien (dismissed):

src/coord/rmr/conn.c
```diff
@@ -524,10 +524,14 @@ static int checkTLS(char** client_key, char** client_cert, char** ca_cert, char*
...
 
+  // If `tls-cluster` is not set to `yes`, we do not connect to the other nodes
+  // with TLS on OSS-cluster. On Enterprise, we always want to connect with TLS
+  // when the tls-port is set to a non-zero value, since this is the port we
+  // get from the proxy.
```
GuyAv46, line 530: and the proxy isn't setting `tls-cluster=yes`? Sound strange

GuyAv46 (approved):

PR #5837, opened 2025-03-30
MOD-9174: Fix pipeline issues

raz-mon:
# Pipeline fixes
## Pager setting time of `resultsLimit`
The pager is responsible for paging the results returned from its upstream, i.e., skip and limit them.
The setting of the `resultsLimit` field of the `queryIterator` was done in the first `Next` call, which was **too late**, since other RPs rely on this value, such as the safe-loader.
This PR fixes the pager result-processor's time of setting of the queryIterator `resultsLimit` field to construction time, rather than at the first call of the `Next` function.

## Empty result not serialized
We do not serialize empty results from the shard pipeline, as such results are not expected by the coordinator which will in turn return an error. We instead **crash in development**, or warn the user that an empty result has been received for serialization via a log, and skip the serialization of that result in **production**.

# Assertions added
This PR adds assertions to some critical places in our code:
## Healthy result id from shard
The result id exists and is not `NULL` in the coordinator response parsing logic.
## SafeLoader last result-status
The safe-loader CANNOT return `RS_RESULT_OK` upon returning the last result status, since the result it returns is not populated.

### DIFF
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -112,8 +112,15 @@ static size_t serializeResult(AREQ *req, RedisModule_Reply *reply, const SearchR
     RedisModule_Reply_Map(reply);
   }
 
-  if (dmd && (options & QEXEC_F_IS_SEARCH)) {
+  if (options & QEXEC_F_IS_SEARCH) {
     size_t n;
+    RS_LOG_ASSERT(dmd, "Document metadata NULL in result serialization.");
+    if (!dmd) {
+      // Empty results should not be serialized!
+      // We already crashed in development env. In production, log and continue
+      RedisModule_Log(req->sctx->redisCtx, "warning", "Document metadata NULL in result serialization.");
+      return 0;
+    }
     const char *s = DMD_KeyPtrLen(dmd, &n);
     if (has_map) {
       RedisModule_ReplyKV_StringBuffer(reply, "id", s, n);
--- a/src/module.c
+++ b/src/module.c
@@ -2061,6 +2061,13 @@ searchResult *newResult_resp3(searchResult *cached, MRReply *results, int j, sea
   }
 
   MRReply *result_id = MRReply_MapElement(result_j, "id");
+  if (!result_id || !MRReply_Type(result_id) == MR_REPLY_STRING) {
+    // We crash in development env, and return NULL (such that an error is raised)
+    // in production.
+    RS_ABORT("Expected id to exist, and be a string");
+    res->id = NULL;
+    return res;
+  }
   res->id = (char*)MRReply_String(result_id, &res->idLen);
   if (!res->id) {
     return res;
--- a/src/result_processor.c
+++ b/src/result_processor.c
@@ -582,10 +582,6 @@ static int rppagerNext_Limit(ResultProcessor *base, SearchResult *r) {
 static int rppagerNext_Skip(ResultProcessor *base, SearchResult *r) {
   RPPager *self = (RPPager *)base;
 
-  // Currently a pager is never called more than offset+limit times.
-  // We limit the entire pipeline to offset+limit (upstream and downstream).
-  uint32_t limit = MIN(self->remaining, base->parent->resultLimit);
-  base->parent->resultLimit = self->offset + limit;
   // If we've not reached the offset
   while (self->offset) {
     int rc = base->upstream->Next(base->upstream, r);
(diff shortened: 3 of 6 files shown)


src/aggregate/aggregate_exec.c
```diff
@@ -112,8 +112,15 @@ static size_t serializeResult(AREQ *req, RedisModule_Reply *reply, const SearchR
...
   }
 
-  if (dmd && (options & QEXEC_F_IS_SEARCH)) {
+  if (options & QEXEC_F_IS_SEARCH) {
     size_t n;
+    RS_LOG_ASSERT(dmd, "Document metadata NULL in result serialization.");
+    if (!dmd) {
+      // Empty results should not be serialized!
+      // We already crashed in development env. In production, log and continue
+      RedisModule_Log(req->sctx->redisCtx, "warning", "Document metadata NULL in result serialization.");
```
kei-nan, line 121: maybe log the document id/index name or additional details that can help debug the issue?

src/module.c
```diff
@@ -2061,6 +2061,13 @@ searchResult *newResult_resp3(searchResult *cached, MRReply *results, int j, sea
   }
 
   MRReply *result_id = MRReply_MapElement(result_j, "id");
+  if (!result_id || !MRReply_Type(result_id) == MR_REPLY_STRING) {
+    // We crash in development env, and return NULL (such that an error is raised)
+    // in production.
+    RS_ABORT("Expected id to exist, and be a string");
```
kei-nan, line 2067: maybe log the j index? and the total result count in the log message?
The index name could also help if that is available.

src/result_processor.c
```diff
@@ -835,12 +848,11 @@ static int rpSafeLoaderNext_Yield(ResultProcessor *rp, SearchResult *result_outp
   SearchResult *curr_res = GetNextResult(self);
 
   if (curr_res) {
-    SetResult(curr_res, result_output);
-  }
-  if (!curr_res || rp->parent->resultLimit <= 1) {
-    return rpSafeLoader_ResetAndReturnLastCode(self);
+    SetResult( curr_res, result_output);
```
kei-nan, line 851:
```suggestion
    SetResult(curr_res, result_output);
```

src/aggregate/aggregate_exec.c
raz-mon: There is no document id, and the index name is not so relevant here I think

src/module.c
raz-mon: We can add the `j` index, sure.
I don't think the index name is relevant here

src/module.c
```diff
@@ -2061,6 +2061,13 @@ searchResult *newResult_resp3(searchResult *cached, MRReply *results, int j, sea
   }
 
   MRReply *result_id = MRReply_MapElement(result_j, "id");
+  if (!result_id || !MRReply_Type(result_id) == MR_REPLY_STRING) {
```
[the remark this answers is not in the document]
raz-mon: No:
```
inline const char *MRReply_String(const MRReply *reply, size_t *len) {
  if (len) {
    *len = reply->len;
  }
  return reply->str;
}
```

GuyAv46 (dismissed):

src/aggregate/aggregate_exec.c
kei-nan: `r->docId`?
I think the more details the better, otherwise the log is less effective in helping us debug these issues in the future.

src/module.c
```diff
@@ -2064,7 +2064,7 @@ searchResult *newResult_resp3(searchResult *cached, MRReply *results, int j, sea
   if (!result_id || !MRReply_Type(result_id) == MR_REPLY_STRING) {
     // We crash in development env, and return NULL (such that an error is raised)
     // in production.
-    RS_ABORT("Expected id to exist, and be a string");
+    RS_LOG_ASSERT_FMT(false, "Expected id %d to exist, and be a string", j);
```
GuyAv46, line 2067: whould be nice to have `RS_ABORT_FMT`

GuyAv46 (approved):

kei-nan (approved):
In this PR a new Next call was added which creates a very unlikely call circle:
rpSafeLoader_ResetAndReturnLastCode -> rpSafeLoaderNext_Accumulate -> rpSafeLoaderNext_Yield -> rpSafeLoader_ResetAndReturnLastCode

We reviewed the flow and it seems unlikely to happen since when setting next to yield we know we have at least one result.

Approving after reviewing the flow.

PR #5838, opened 2025-03-30
MOD-9216: Fix flakiness in TLS test

raz-mon:
This PR fixes a flakiness in the TLS test due to the time it takes for the cluster topology to be refreshed and parsed.
We also lock all required packages' versions to avoid unwanted updates.

### DIFF
--- a/tests/pytests/requirements.txt
+++ b/tests/pytests/requirements.txt
@@ -1,9 +1,9 @@
-packaging >= 20.8
-gevent
-deepdiff <= 8.3.0
-redis >= 5.1.0
-RLTest >= 0.7.16
-numpy >= 1.21.6
-scipy >= 1.7.3
-faker
-distro
+packaging == 20.8
+gevent == 24.11.1
+deepdiff == 8.3.0
+redis == 5.1.0
+RLTest == 0.7.16
+numpy == 1.21.6
+scipy == 1.7.3
+faker == 37.1.0
+distro == 1.9.0
--- a/tests/pytests/test.py
+++ b/tests/pytests/test.py
@@ -4371,6 +4371,8 @@ def test_with_tls_and_non_tls_ports():
     # Upon setting `tls-cluster` to `no`, we should still be able to succeed
     # connecting the coordinator to the shards, just not in TLS mode.
     run_command_on_all_shards(env, 'CONFIG', 'SET', 'tls-cluster', 'no')
+    time.sleep(2)   # Wait a bit for the new topology to kick in (inter-node
+                    # connections to be non-tls).
 
     common_with_auth(env)
 


tests/pytests/test.py
```diff
@@ -4371,6 +4371,8 @@ def test_with_tls_and_non_tls_ports():
     # Upon setting `tls-cluster` to `no`, we should still be able to succeed
     # connecting the coordinator to the shards, just not in TLS mode.
     run_command_on_all_shards(env, 'CONFIG', 'SET', 'tls-cluster', 'no')
+    time.sleep(2)   # Wait a bit for the new topology to kick in (inter-node
```
GuyAv46, line 4374: No redis command for waiting for these changes? (Like `wait` for replicas)

tests/pytests/requirements.txt
```diff
@@ -1,9 +1,9 @@
...
-RLTest >= 0.7.16
-numpy >= 1.21.6
-scipy >= 1.7.3
-faker
-distro
+packaging == 24.2
+gevent == 24.11.1
+deepdiff == 8.3.0
+redis >= 5.2.1
+RLTest == 0.7.16
```
kei-nan, line 5: Do we really want to pin all of the python packages?
I think the problem in this specific case is RLTest.

If we pin everything now and try to upgrade in the future the amount of deltas could be big and we would be a lot of work.

pinning as we go allows us to unpin when needed and handle the specific changes the new package brings with it.

raz-mon: Not pinning leaves us exposed to flakiness due to package updates.
As it makes sense to have this automatic update when we have a little amount of packages, it is not maintainable and stable when we have a lot.
Since we don't want to start debating when are there "a lot" of packages, and we don't have any owner of this, it is most stable to pin the versions, and update them every now and then.
@DvirDukhan please give your 2 cents so we can move on with the PR, here or there.

tests/pytests/test.py
raz-mon: Maybe we can use `search.CLUSTERINFO` to see if we got the new topology.
I'll see if there's a good field for that (probably not).

raz-mon: Yup, done.

tests/pytests/test.py
```diff
@@ -4366,11 +4366,29 @@ def test_with_tls_and_non_tls_ports():
...
+    # Get the TLS ports
+    tls_ports = get_ports(env)
+    # The non-TLS ports are the TLS ports + 1500 (hard-coded in RLTest)
+    expected_ports = [port + 1500 for port in tls_ports]
 
     # Upon setting `tls-cluster` to `no`, we should still be able to succeed
     # connecting the coordinator to the shards, just not in TLS mode.
     run_command_on_all_shards(env, 'CONFIG', 'SET', 'tls-cluster', 'no')
+    while get_ports(env) != expected_ports:
+        time.sleep(0.1)
```
GuyAv46, line 4391: consider wrapping with a `TimeLimit`

GuyAv46 (dismissed):

kei-nan (dismissed):

kei-nan (approved):

DvirDukhan:
is the sanitizer flaky tests related to this PR content? https://github.com/RediSearch/RediSearch/actions/runs/14241528482/job/39912353195?pr=5838 @raz-mon

raz-mon:
@DvirDukhan This PR is supposed to fix this flakiness.
For some reason there is still some flakiness left that I need to fix

GuyAv46 (dismissed):

PR #5839, opened 2025-03-30
[MOD-9217] Fix timeout testFailOnTimeout_nonStrict ON_TIMEOUT policy

meiravgri:
set ON_TIMEOUT to RETURN so that test returns the expected profile output when the timeout is reached on non-strict timeout policy


alonre24 (approved):
Backports to 2.10 and 2.8 is not a must (good to have)

raz-mon (approved):

PR #5844, opened 2025-03-30
MOD-9222: Fix strict timeout result aggregation

raz-mon:
This PR fixes the aggregation of results when using strict timeout policy together with the safe-loader (i.e., multi-threading).
Specifically, the decrement of the `resultsLimit` was done too early, such that the safe-loader saw an already-decremented value, thus fetching one less result than needed.
The decrement was moved to the body of the loop, in case the `Next()` call succeeded, i.e., `RS_RESULT_OK` was returned from the pipeline.

A test was already encountering this bug, but was changed to use the non-strict timeout policy, thus hiding the bug. This setting was removed from the test, such that now it will use the strict timeout policy and test this behavior (Cursor + MT).


kei-nan (approved):

GuyAv46 (approved):
🔥

PR #5846, opened 2025-03-31
[2.10] MOD-8391: Report active threads-indexes upon crash

kei-nan:
Manual backport of #5403.


coord/src/dist_plan.cpp
```diff
@@ -375,6 +375,8 @@ int AGGPLN_Distribute(AGGPlan *src, QueryError *status) {
   dstp->base.getLookup = distStepGetLookup;
   BlkAlloc_Init(&dstp->alloc);
 
+  // TODO: The while condition is buggy, since it returns the `AGGPlan`, not the `PLN_BaseStep` that is actually needed
+  // Should be fixed to `DLLIST_FOREACH(it, ll) {}`.
```
raz-mon, line 379: I thought we said we'll remove this and open a ticket, no?

tests/pytests/test_crash.py
```diff
@@ -0,0 +1,37 @@
...
+                pos += 1
+            elif pos != 0 and pos < len(lines):
+                return False
+    return pos == len(lines)
+
+# we expect to see out index information about the crash in the log file
+@skip(cluster=True)
+def test_query_thread_crash():
+    env = CrashingEnv(testName="test_query_thread_crash", freshEnv=True)
+    env.assertTrue(expect_query_crash_output(env, ['search_current_thread', 'search_index:idx']))
```
raz-mon, line 37: Is the index name not supposed to be obfuscated?

kei-nan: only if hideUserDataFromLogs was set to true

coord/src/dist_plan.cpp
kei-nan: I think I removed the comment that was next to the DLL macros.
We did open a ticket in the end:
https://redislabs.atlassian.net/browse/MOD-9180?searchObjectId=553741&searchContainerId=10026&searchContentType=issue&searchSessionId=1536bd6e-d9c4-40e4-9632-18c0135f6cc8

raz-mon: I see, I thought we did this for this as well.
I think we should

tests/pytests/test_crash.py
raz-mon: Seems like something worth testing

PR #5850, opened 2025-04-01
MOD-9274: Pin CMake Version For Mac OS

kei-nan:
- pin cmake in macos to prevent errors in deps cmake files about unsupported cmake version

1. Current: brew update updates cmake to version 4 which causes deps/hiredis compilation to fail due to too low of a cmake version in `cmake_minimum_required` directive.
2. Change: Prevent the update for cmake by pinning the cmake version, we wanted a quick fix for now.
3. Outcome: CI can pass

1. macos.sh

#### Mark if applicable

### DIFF
--- a/.install/macos.sh
+++ b/.install/macos.sh
@@ -10,6 +10,9 @@ fi
 
 export HOMEBREW_NO_AUTO_UPDATE=1
 
+# Without pinning cmake, it will install the latest version(>= 4.0)
+# This leads to deps/hiredis failing to compile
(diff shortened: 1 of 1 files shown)


.install/macos.sh
```diff
@@ -10,6 +10,7 @@ fi
 
 export HOMEBREW_NO_AUTO_UPDATE=1
 
+brew pin cmake
 brew update
```
alonre24, line 14: Consider docment what the version we expect here

alonre24 (approved):

PR #5862, opened 2025-04-03
Account For Vector Set Module In Tests

kei-nan:
* expect vector sets to be loaded by default

1. Current: vector set module was implicitly added in redis and it is causing some of our tests to fail
2. Change: account for vector sets existence in our tests when using module list command
3. Outcome: CI will pass

1. tests

#### Mark if applicable


raz-mon (approved):
Do we want to condition this on something, or is the vectorSet module always going to be loaded?

PR #5869, opened 2025-04-04
[2.10] Fix RSValue String Convert Call

kei-nan:
Fix a case where null was passed into a pointer and the pointer was later dereferenced.

1. Current: We pass NULL as the length pointer, the pointer, in certain cases it will be immediately dereferenced.
2. Change: Use a local variable and pass its address
3. Outcome: No longer crash on that specific call for dereferencing null.

1. RSValue

#### Mark if applicable

### DIFF
--- a/src/aggregate/expr/expression.c
+++ b/src/aggregate/expr/expression.c
@@ -507,12 +507,13 @@ void RPEvaluator_Reply(RedisModule_Reply *reply, const char *title, const Result
   RS_LOG_ASSERT (type == RP_PROJECTOR || type == RP_FILTER, "Error");
 
   char buf[32];
+  size_t len;
   const char *literal;
   RPEvaluator *rpEval = (RPEvaluator *)rp;
   const RSExpr *expr = rpEval->eval.root;
   switch (expr->t) {
     case RSExpr_Literal:
-      literal = RSValue_ConvertStringPtrLen(&expr->literal, NULL, buf, sizeof(buf));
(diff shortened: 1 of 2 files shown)


raz-mon (commented):
Let's add a test that would have crashed before.
Profile +aggregate + apply should do

src/aggregate/expr/expression.c
```diff
@@ -507,12 +507,13 @@ void RPEvaluator_Reply(RedisModule_Reply *reply, const char *title, const Result
...
 
   char buf[32];
+  size_t len;
   const char *literal;
   RPEvaluator *rpEval = (RPEvaluator *)rp;
   const RSExpr *expr = rpEval->eval.root;
   switch (expr->t) {
     case RSExpr_Literal:
-      literal = RSValue_ConvertStringPtrLen(&expr->literal, NULL, buf, sizeof(buf));
+      literal = RSValue_ConvertStringPtrLen(&expr->literal, &len, buf, sizeof(buf));
```
DvirDukhan, line 516: we need to have input validation on `RSValue_ConvertStringPtrLen`
same as at
```c
const char *RM_StringPtrLen(const RedisModuleString *str, size_t *len) {
    if (str == NULL) {
        const char *errmsg = "(NULL string reply referenced in module)";
        if (len) *len = strlen(errmsg);
        return errmsg;
    }
    if (len) *len = sdslen(str->ptr);
```

PR #5892, opened 2025-04-08
MOD-9316: Fix Duplicate Index Loading From RDB

kei-nan:
need to consume the entirety of the rdb when loading an index, even if it is a duplicate

1. Current: Since we avoid reading from the rdb if the index is a duplicate then redis issues an error since the aux rdb part was not consumed in its entirety.
2. Change: Revert to old behaviour where we consumed the entire aux RDB part, turning on spec flags to avoid some side effects.
3. Outcome: RDB flow will behave in the expected way.

1. spec

#### Mark if applicable

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -1613,8 +1613,10 @@ void IndexSpec_RemoveFromGlobals(StrongRef spec_ref, bool removeActive) {
   // Remove spec from global index list
   dictDelete(specDict_g, (void*)spec->specName);
 
-  // Remove spec from global aliases list
-  IndexSpec_ClearAliases(spec_ref);
+  if (!spec->isDuplicate) {
+    // Remove spec from global aliases list
+    IndexSpec_ClearAliases(spec_ref);
+  }
 
   SchemaPrefixes_RemoveSpec(spec_ref);
 
@@ -2582,15 +2584,17 @@ int IndexSpec_CreateFromRdb(RedisModuleCtx *ctx, RedisModuleIO *rdb, int encver,
 
   RefManager *oldSpec = dictFetchValue(specDict_g, specName);
   if (oldSpec) {
-    // spec already exists lets just free this one
-    HiddenString_Free(specName, true);
+    // spec already exists, however we need to finish consuming the rdb so redis won't issue an error(expecting an eof but seeing remaining data)
+    // right now this can cause nasty side effects, to avoid them we will set isDuplicate to true
     RedisModule_Log(RSDummyContext, "notice", "Loading an already existing index, will just ignore.");
-    return REDISMODULE_OK;
   }
 
   IndexSpec *sp = rm_calloc(1, sizeof(IndexSpec));
   StrongRef spec_ref = StrongRef_New(sp, (RefManager_Free)IndexSpec_Free);
   sp->own_ref = spec_ref;
+  // setting isDuplicate to true will make sure index will not be removed from global
+  // cursor map and aliases.
+  sp->isDuplicate = oldSpec != NULL;
 
   // `indexError` must be initialized before attempting to free the spec
   sp->stats.indexError = IndexError_Init();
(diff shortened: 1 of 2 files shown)


nafraf:
Is it possible to add a test for this change?

src/spec.c
```diff
@@ -2582,15 +2584,17 @@ int IndexSpec_CreateFromRdb(RedisModuleCtx *ctx, RedisModuleIO *rdb, int encver,
...
     RedisModule_Log(RSDummyContext, "notice", "Loading an already existing index, will just ignore.");
-    return REDISMODULE_OK;
   }
 
   IndexSpec *sp = rm_calloc(1, sizeof(IndexSpec));
   StrongRef spec_ref = StrongRef_New(sp, (RefManager_Free)IndexSpec_Free);
   sp->own_ref = spec_ref;
+  // setting isDuplicate to true will make sure index will not be removed from global
+  // cursor map and aliases.
+  sp->isDuplicate = oldSpec != NULL;
```
oshadmi, line 2597: Consider moving this nearer to where `oldSpec` is set

src/spec.c
```diff
@@ -2680,10 +2684,20 @@ int IndexSpec_CreateFromRdb(RedisModuleCtx *ctx, RedisModuleIO *rdb, int encver,
   }
 
   sp->scan_in_progress = false;
-  dictAdd(specDict_g, (void*)sp->specName, spec_ref.rm);
+  if (sp->isDuplicate) {
```
oshadmi, line 2687: Can we add a unit test?

src/spec.c
kei-nan: We need sp to be allocated,
It is right after we have a strong ref for the index.
Do you want to move this to be before the setting of `own_ref`?

src/spec.c
```diff
@@ -2582,15 +2584,17 @@ int IndexSpec_CreateFromRdb(RedisModuleCtx *ctx, RedisModuleIO *rdb, int encver,
...
+    // right now this can cause nasty side effects, to avoid them we will set isDuplicate to true
     RedisModule_Log(RSDummyContext, "notice", "Loading an already existing index, will just ignore.");
-    return REDISMODULE_OK;
   }
 
   IndexSpec *sp = rm_calloc(1, sizeof(IndexSpec));
   StrongRef spec_ref = StrongRef_New(sp, (RefManager_Free)IndexSpec_Free);
   sp->own_ref = spec_ref;
+  // setting isDuplicate to true will make sure index will not be removed from global
+  // cursor map and aliases.
```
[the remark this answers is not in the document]
kei-nan: Yep, used the old comment
Will update it.

src/spec.h
```diff
@@ -314,6 +314,7 @@ typedef struct IndexSpec {
   bool cascadeDelete;             // (deprecated) remove keys when removing spec. used by temporary index
   bool monitorDocumentExpiration;
   bool monitorFieldExpiration;
+  bool isDuplicate;               // Marks that this index is a duplicate of an existing one
```
raz-mon, line 317: This can't really happen other than during loading from a persistence file (rdb, aof), so why have it as a field?
I think it would be better not to do have it.

raz-mon (commented):
Consider making a flow that will just consume the rdb file, and not call all these unnecessary allocations and frees, and especially not mangle with the global data-structures for no reason (e.g., prefixes)

src/spec.c
oshadmi: Not critical - Can move the `if (oldSpec)` block downward?

src/spec.c
oshadmi: Can we add a unit test that calls to the registered `aux_load` handler, like in https://github.com/redis/redis/blob/7.4/src/rdb.c#L3523,
And then to confirm the entire AUX data was consumed and the end marker was reached
as in https://github.com/redis/redis/blob/7.4/src/rdb.c#L3533 ?

src/spec.h
oshadmi: We use this field in `IndexSpec_RemoveFromGlobals`

raz-mon: It can be an argument to the function, better than an almost unused field IMO.
If we want a fast fix that does not seek to be optimal, but similar to what we had before I guess we can go forward with it, but I do think it is not ideal.

src/spec.c
```diff
@@ -2680,10 +2683,20 @@ int IndexSpec_CreateFromRdb(RedisModuleCtx *ctx, RedisModuleIO *rdb, int encver,
...
   sp->scan_in_progress = false;
-  dictAdd(specDict_g, (void*)sp->specName, spec_ref.rm);
+  if (sp->isDuplicate) {
+    // spec already exists lets just free this one
+    // Remove the new spec from the global prefixes dictionary.
+    // This is the only global structure that we added the new spec to at this point
+    SchemaPrefixes_RemoveSpec(spec_ref);
+    addPendingIndexDrop();
+    StrongRef_Release(spec_ref);
+    spec_ref = (StrongRef){oldSpec};
```
raz-mon, line 2693: Can remove

kei-nan:
```suggestion
```

src/spec.c
kei-nan: yes, but we would still need to mock all the rdb functions + rdb object itself.

kei-nan: yes, but we would still need to mock all the rdb functions + rdb object itself.

oshadmi: OK. I will create an issue for this.

src/spec.c
kei-nan: okay

kei-nan: Moving the if block downwards makes the comment not make sense for the next person that comes across this piece of code.

oshadmi (approved):

PR #5909, opened 2025-04-10
Coverage Report Without `readies` - Phase 2 - [MOD-6711]

GuyAv46:
### Key Changes in the PR
1. Delete `flow-coverage` and unify the flow into `task-test`
2. Fix event workflows to use the latest redis, instead of unstable (except on nightly)
3. Fix the redis build, add a build flag for coverage (to collect coverage info from fork child on flow tests - `REDIS_CFLAGS=-DCOVERAGE_TEST`)
4. Add a missing definition for coverage in `CMakeLists.txt`
5. Add coverage collection logic to `build.sh`
6. Fix inputs and other definitions in `build.sh`
7. Replace `BINROOT` input with `BINDIR` in `sbin/unit-tests`
8. Delete empty files
9. Fix a Rust test for `miri`

#### Mark if applicable


.github/workflows/task-test.yml
```diff
@@ -166,14 +170,19 @@ jobs:
 
       - name: Build Redis
         working-directory: redis
-        run: |
-          ${{ steps.mode.outputs.mode }} SANITIZER=${{ inputs.san }} BUILD_TLS=yes make install
+        run:  ${{ steps.mode.outputs.mode }} make install
+              BUILD_TLS=yes
+              ${{ inputs.coverage && 'REDIS_CFLAGS=-DCOVERAGE_TEST' || '' }}
+              SANITIZER=${{ inputs.san }}
```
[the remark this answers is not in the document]
GuyAv46: Incorrect. All the lines will be executed as a single command (the intended usage here)

.github/workflows/event-merge-to-queue.yml
```diff
@@ -16,40 +16,52 @@ concurrency:
...
     secrets: inherit
     uses: ./.github/workflows/task-test.yml
     with:
-      get-redis: unstable
+      get-redis: ${{ needs.get-latest-redis-tag.outputs.tag }}
       rejson-branch: master
       test-config: '' # run all tests
       san: address
       env: ubuntu-latest
-      run_miri: true
```
alonre24, line 49: Why stop running miri ?

build.sh
```diff
@@ -178,6 +184,49 @@ setup_build_environment() {
   BINDIR="${BINROOT}/${FULL_VARIANT}/${OUTDIR}"
 }
 
+#-----------------------------------------------------------------------------
+# Function: prepare_coverage_capture
+# Run lcov preparations before testing for coverage
+#-----------------------------------------------------------------------------
+prepare_coverage_capture() {
+  [[ -n $GITHUB_ACTIONS ]] && echo "::group::Code Coverage Preparation"
```
alonre24, line 192: What is the purpose of these statements? Just log messages?

build.sh
```diff
@@ -178,6 +184,49 @@ setup_build_environment() {
...
+  # Extract only the coverage of the project source files
+  lcov --output-file $BINROOT/$NAME.info --extract $BINROOT/full.info \
+    "$ROOT/src/*" \
+    "$ROOT/deps/triemap/*" \
+    "$ROOT/deps/thpool/*" \
+
+  # TODO: remove `tests`?
+  # # Remove coverage for directories we don't want
+  # lcov -o $BINROOT/tmp.info --remove $BINROOT/full.info \
+  #   "*/tests/*" \
```
alonre24, line 222: What is that? Why should we care about coverage of tests?

build.sh
```diff
@@ -178,6 +184,49 @@ setup_build_environment() {
...
+  lcov --capture --directory $BINROOT --base-directory $ROOT -o $BINROOT/test.info
+
+  # Accumulate results with the baseline captured before the test
+  lcov --add-tracefile $BINROOT/base.info --add-tracefile $BINROOT/test.info -o $BINROOT/full.info
+
+  # Extract only the coverage of the project source files
+  lcov --output-file $BINROOT/$NAME.info --extract $BINROOT/full.info \
+    "$ROOT/src/*" \
+    "$ROOT/deps/triemap/*" \
+    "$ROOT/deps/thpool/*" \
```
alonre24, line 217: I suggest covering all the deps that are not GitHub submodules
So we can see in front oof ur eyes which code we imported for no reason and clean it up later on

.github/workflows/event-merge-to-queue.yml
GuyAv46: Not stopping, only removing the explicit input.
We use `miri` if this is a sanitizer flow.

build.sh
GuyAv46: Yes, this is for a nice grouping (folding) of the log prints, to reduce noise

build.sh
GuyAv46: We don't, but we are also not collecting such info today (we don't have `tests` sub-directory under `src`, except for rust code which we collect "automatically" thanks to `cargo`, and not by this manual implementation.
On 2.10 and previous builds, we have such directories (for the coordinator for example), so we will need this step

build.sh
GuyAv46: This is the right way to work. First, describe which directories are needed for coverage, and then remove specific sub-directories that are not required.
There is no benefit for including other `deps` we don't care about. Until this PR, we didn't have any coverage report of code under `deps`, so at least we have it now for owned code

build.sh
GuyAv46: I uncommented the step to be aligned on all branches, and also ignore future `tests` subdirectories

alonre24 (approved):

PR #5913, opened 2025-04-11
get_cmake_version is not complatible with alpine, fix that

meiravgri:
In .install/verify_build_deps.sh we use the following command to get `cmake` version:
```
cmake --version | grep -oP 'cmake version \K[0-9.]+' | cut -d. -f1,2
```
However, `-P` option for `grep` is not supported on **Alpine**.
This PR replaced cmake version extraction with a compatible set of commands:
```
cmake --version | grep "cmake version" | sed -E 's/.*cmake version ([0-9]+\.[0-9]+).*/\1/'
```

## Changes
* `.install/verify_build_deps.sh:get_cmake_version` function: Replaced the `grep -oP` command with `grep` and `sed` to extract the CMake version.

### DIFF
--- a/.install/verify_build_deps.sh
+++ b/.install/verify_build_deps.sh
@@ -128,7 +128,7 @@ get_compiler_version() {
 
 # extract the version in the format X.Y
 get_cmake_version() {
-  cmake --version | grep -oP 'cmake version \K[0-9.]+' | cut -d. -f1,2
+  cmake --version | grep "cmake version" | sed -E 's/.*cmake version ([0-9]+\.[0-9]+).*/\1/'
 }
 
 # ==== Version Checkers ====


.install/verify_build_deps.sh
```diff
@@ -128,7 +128,7 @@ get_compiler_version() {
 
 # extract the version in the format X.Y
 get_cmake_version() {
-  cmake --version | grep -oP 'cmake version \K[0-9.]+' | cut -d. -f1,2
+  cmake --version | grep "cmake version" | sed -E 's/.*cmake version ([0-9]+\.[0-9]+).*/\1/'
```
nafraf, line 131: Do we need `s/.*cmake...`? Is `s/^cmake...` enough?

meiravgri: Wouldn't it be better to keep it more general?

nafraf: I don't expect changes in `cmake --version` output, but OK, It's better to keep it general. :)

PR #5918, opened 2025-04-14
MOD-9234: Pin test dependencies

raz-mon:
Pins the pytest dependencies' versions to reduce the flakiness of our CI.
Adds a pin to the `orderly-set` package, which was updated to `5.4.0`, which uses syntax supported from Python 3.9 and on. Until we remove Python 3.8 from our CI, we pin this package version to `5.3.0` such that the platforms that use Python3.8 are able to execute the test-suite.

### DIFF
--- a/tests/pytests/requirements.txt
+++ b/tests/pytests/requirements.txt
@@ -1,9 +1,9 @@
-packaging >= 20.8
-gevent
-deepdiff <= 8.3.0
-redis >= 5.1.0
-RLTest >= 0.7.16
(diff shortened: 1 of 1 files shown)


raz-mon:
CI run on all (Linux) platforms: https://github.com/RediSearch/RediSearch/actions/runs/14439600230

alonre24 (dismissed):
Thank you

alonre24 (approved):

oshadmi (approved):

PR #5926, opened 2025-04-14
MOD-9231 - Integrate Rust Benchmarks in CI

DarthB:
# MOD-9231 - Integrate Rust Benchmarks in CI

Adds a github action workflow that runs the micro benchmarks (only rust for now) and provides the output as github artifact. 

This workflow use two triggers (event for PR or push to integration):

1. If a PR is generated the benchmark is run and the benchmark results are stored as github artifact. If available, the benchmark is compared to a baseline from master.
2. For a push to master the benchmark is run and stored as baseline for comparison by upcoming pull requests.

## Notes

- The github action only runs if files of the rust project have changed.
- The github action uses reduced setup for the benchmark environment taken from [benchmark-flow.yml workflow](https://github.com/RediSearch/RediSearch/blob/master/.github/workflows/benchmark-flow.yml)

## Remarks 

- In course of the PR a leak in the C-side allocation / de-allocation shim has been fixed. see [PR 5951](https://github.com/RediSearch/RediSearch/pull/5951)
- For benchmark with bigger datasets, the Ubuntu github runner needs more overall time to be statistically reliable (a benchmark tool parameter, called `measurement_time` has been raised)
- Introduces a task to check for changes and uses this to decided if a workflow needs to run

#### Mark if applicable


.github/workflows/benchmark-rust.yml
```diff
@@ -0,0 +1,155 @@
...
+      - name: Run benchmark on master
+        if: github.ref == 'refs/heads/master' && github.event_name == 'push' && success() 
+        run: |
+          ulimit -v 12582912
+          (while true; do ps -o pid,rss,%mem,%cpu,cmd | grep -E 'cargo|bench' | grep -v grep; sleep 5; done) &
+          MONITOR_PID=$!
+          cargo bench -- --save-baseline master
+          kill $MONITOR_PID
+        working-directory: "src/redisearch_rs/trie_bencher"
+
```
LukeMathWalker, line 143: Why do we want to keep this monitor stuff?

src/redisearch_rs/trie_bencher/build.rs
```diff
@@ -6,20 +6,25 @@ fn main() {
...
         };
-        let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
+
         root.join(format!(
             "bin/{target_os}-{target_arch}-release/search-community/deps/triemap"
         ))
     };
 
     assert!(std::fs::exists(lib_dir.join("libtrie.a")).unwrap());
+    // ensure that undefined symbols are allowed - that's seems to be the case on macOS but not un Ubuntu in CI.
```
LukeMathWalker, line 24: Let's elaborate in the comment as to why we have undefined symbols and that's OK here.

src/redisearch_rs/trie_bencher/src/lib.rs
```diff
@@ -28,12 +29,37 @@ pub fn str2u8(input: &str) -> Vec<u8> {
...
+        // free the memory allocated for the C string
+        unsafe {
+            dealloc(
+                self.cstr_ptr as *mut u8,
+                Layout::from_size_align(self.c_len as usize, std::mem::align_of::<usize>())
+                    .unwrap(),
+            )
+        }
+    }
 }
```
LukeMathWalker, line 65: Wouldn't it be simpler to define an extension method on `CString`, named `as_trie_input` which returns the underlying pointer and the correct length?

The implementation would desugar to

```rust
let view = self.as_c_str();
let len = view.count_bytes();
let ptr = view.as_ptr();
(ptr, len)
```
```

File: src/redisearch_rs/trie_bencher/monitor_bench.sh [hunk 4]
```diff
@@ -0,0 +1,23 @@
...
+    echo ""
+    sleep 5
+done) &
+MONITOR_PID=$!
+
+# Run the benchmark
+cargo bench
+
+# Stop the monitoring
+kill $MONITOR_PID
```
LukeMathWalker, line 23: Do we need to keep this around?

.github/workflows/benchmark-rust.yml
```diff
@@ -0,0 +1,155 @@
+name: Rust Benchmarks
```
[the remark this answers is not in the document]
DarthB: I added a benchmark folder to run `cargo bench` in it and an `output folder` to store the specific benchmark as preparation for matrix like runs.

.github/workflows/benchmark-rust.yml
DarthB: If we don't have it and happen to implement another leak, we end up with a non-saying error message in CI. 
I extracted it in a sh script that is usable locally too.

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -0,0 +1,122 @@
...
+      - name: Setup tests dependencies
+        run: |
+          .install/test_deps/common_installations.sh ${{ steps.mode.outputs.mode }} sudo
+      - name: Install Boost
+        working-directory: .install
+        run: ./install_boost.sh ${{ env.BOOST_VERSION }} sudo
+
+      - name: Build RediSearch
+        run: make build
+            PROFILE=${{ inputs.profile_env }}
```
LukeMathWalker, line 71: I don't think this `PROFILE` argument does anything.

.github/workflows/event-push-to-integ.yml
```diff
@@ -7,40 +7,56 @@ on:
...
+      - name: Checkout
+        uses: actions/checkout@
+        with:
+          fetch-depth: 0 # required for changed-files action to work
+      - name: Check if micro-benchmarks were changed
+        id: check-benchmarks
+        uses: tj-actions/changed-files@
+        with: # Only run on changes to these paths (code, deps, and benchmarks related changes)
+          files: |
+            src/redisearch_rs/**
```
LukeMathWalker, line 47: Do we need to separate jobs? Can't we have a single job that checks out the code and then exports the required output variables in two different steps to trigger/suppress the benchmark/micro-benchmarks below?

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -0,0 +1,122 @@
...
+            SHOW=1
+      - name: Prepare automation
+        run: |
+          ./deps/readies/bin/getpy3
+          pip3 install --upgrade pip
+          python3 -m pip install -r tests/benchmarks/requirements.txt
+          VERSION=1.11.1 sudo ./deps/readies/bin/getterraform
+      - name: Prepare ReJSON Module
+        run: REJSON_BRANCH=${{ inputs.rejson_branch }} ./tests/deps/setup_rejson.sh
+
```
LukeMathWalker, line 81: This setup feels overly complicated. We should be able to get away with something as simple as what we have in `task-lint.yml`.

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -0,0 +1,122 @@
...
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Setup specific  # assuming node20 supported
+        working-directory: .install
+        run: |
+          ./install_script.sh ${{ steps.mode.outputs.mode }} sudo
+      - name: Setup tests dependencies
+        run: |
+          .install/test_deps/common_installations.sh ${{ steps.mode.outputs.mode }} sudo
```
LukeMathWalker, line 64: Do you want to hard-code the mode to `sudo` or do you want to compute it in a step upstream? We need to pick one strategy here.

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -0,0 +1,122 @@
...
+      - name: Install Boost
+        working-directory: .install
+        run: ./install_boost.sh ${{ env.BOOST_VERSION }} sudo
```
LukeMathWalker, line 67: The install script will take care of this, no need to have this step.

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -0,0 +1,122 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Setup specific  # assuming node20 supported
+        working-directory: .install
+        run: |
+          ./install_script.sh ${{ steps.mode.outputs.mode }} sudo
+      - name: Setup tests dependencies
```
LukeMathWalker, line 62: Do we actually need test dependencies here?

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -0,0 +1,122 @@
...
+            PROFILE=${{ inputs.profile_env }}
+            SHOW=1
+      - name: Prepare automation
+        run: |
+          ./deps/readies/bin/getpy3
+          pip3 install --upgrade pip
+          python3 -m pip install -r tests/benchmarks/requirements.txt
+          VERSION=1.11.1 sudo ./deps/readies/bin/getterraform
+      - name: Prepare ReJSON Module
+        run: REJSON_BRANCH=${{ inputs.rejson_branch }} ./tests/deps/setup_rejson.sh
```
LukeMathWalker, line 80: Pretty confident none of this is needed.

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -0,0 +1,122 @@
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: "rust-benchmark-results-master"
+          path: ${{ inputs.output_folder }}
+      - name: Upload benchmarks for PR comparison
+        if: github.event_name == 'pull_request'
+        uses: actions/upload-artifact@v4
+        with:
+          name: "rust-benchmark-results-pr-${{ github.event.pull_request.number }}"
+          path: ${{ inputs.output_folder }}
```
LukeMathWalker, line 122: This doesn't seem to be working, looking at https://github.com/RediSearch/RediSearch/actions/runs/14612608134/job/40993542455#step:16:12

src/redisearch_rs/trie_bencher/build.rs
```diff
@@ -5,20 +5,26 @@ fn main() {
...
-        let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
+
         root.join(format!(
             "bin/{target_os}-{target_arch}-release/search-community/deps/triemap"
         ))
     };
 
     assert!(std::fs::exists(lib_dir.join("libtrie.a")).unwrap());
+    // Allow undefined symbols in non-macOS builds to align with macOS linker behavior, which permits unresolved symbols by default.
+    // This is safe here because the undefined symbols are resolved at runtime via dynamic linking, ensuring compatibility in CI on Ubuntu.
```
LukeMathWalker, line 24:
```suggestion
    // There are several symbols exposed by `libtrie.a` that we don't
    // actually invoke (either directly or indirectly) in our benchmarks.
    // We provide a definition for the ones we need (e.g. Redis' allocation functions),
    // but we don't want to be forced to add dummy definitions for the ones we don't rely on.
    // We prefer to fail at runtime if we try to use a symbol that's undefined.
    // This is the default linker behaviour on macOS. On other platforms, the default
    // configuration is stricter: it exits with an error if any symbol is undefined. 
    // We intentionally relax it here.
```

Let's explain the _why_ in comments, not just the how.

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -0,0 +1,122 @@
...
+    
+    # For calling from other workflows
+    workflow_call:
+      inputs:
+        profile_env:
+          type: number # for default of 0
+        rejson_branch:
+          type: string
+          default: master
+          description: 'branch to use for rejson'
```
LukeMathWalker, line 34: Based on my comments below, I don't think these two are needed.

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -0,0 +1,122 @@
...
+        benchmark_folder: 
+          type: string
+          default: src/redisearch_rs/trie_bencher
+          description: folder containing the benchmark to be run with `cargo bench`
+        output_folder:
+          type: string
+          default: bin/criterion/trie
+          description: folder used to store the output of criterion
```
LukeMathWalker, line 42: I think it's premature to try to parametrise this since different setups may be required depending on the benchmark.
I would suggest to keep it simple first, and then generalize when we have multiple benchmark suites.

.github/workflows/benchmark-rust.yml
LukeMathWalker: The same reasoning applies to almost every other CI job—I don't see why benchmarks are particularly special here.
In particular, the monitor script doesn't provide us much more information than "there is a memory leak", since it wraps around the entire benchmark suite.
I would prefer to keep the CI script uncluttered, since it looks like the complexity we are bringing in outweighs the benefit to me.

DarthB:
addressed your comments and stored the done work for folder input parameters and the monitoring shell script elsewhere.

.github/workflows/event-pull_request.yml
```diff
@@ -30,6 +30,9 @@ jobs:
       rejson-branch: master
     secrets: inherit
 
+  micro-benchmarks:
+    uses: ./.github/workflows/flow-micro-benchmarks.yml
```
LukeMathWalker, line 34: Shouldn't this be filtered on path changes, too?

.github/workflows/task-check-changes.yml
```diff
@@ -0,0 +1,56 @@
...
+      RUST_CODE_CHANGED: ${{ steps.check-micro-benchmarks.outputs.any_modified }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # required for changed-files action to work
+      
+      - name: Check if code were changed
+        id: check-code
+        uses: tj-actions/changed-files@v46
```
DvirDukhan, line 33: I think this had a major CVE no? we removed this action from our repo due to that

src/redisearch_rs/trie_bencher/benches/operations.rs
```diff
@@ -64,7 +64,7 @@ fn criterion_benchmark_redis_wiki_10k(c: &mut Criterion) {
     let corpus = CorpusType::RedisBench10kNumerics;
     let terms = corpus.create_terms(true);
     let bencher =
-        OperationBencher::new("Wiki-10K".to_owned(), terms, Some(Duration::from_secs(10)));
+        OperationBencher::new("Wiki-10K".to_owned(), terms, Some(Duration::from_secs(20)));
```
DvirDukhan, line 67: why the increase?

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -0,0 +1,69 @@
...
+  workflow_call:
+    inputs:
+      env:
+        default: "${{ vars.RUNS_ON }} || ubuntu-latest"
+        type: string
+
+env:
+  BOOST_VERSION: ${{ vars.BOOST_VERSION || '1.83.0' }}
+  LLVM_VERSION: ${{ vars.LLVM_VERSION || 18 }}
+
```
DvirDukhan, line 13: I believe we removed the need for it

DvirDukhan (commented):
thanks
ny main concern is the GH action

.github/workflows/task-check-changes.yml
DarthB: [It had](https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/). Luckily v46 isn't affected anymore. It's still in the redisearch repo but [commented out](https://github.com/RediSearch/RediSearch/blob/38cb9277e341e63beeca21a90af66d6d548e158a/.github/workflows/event-push-to-integ.yml#L32). I double checked the logs and didn't find spilled secrets. Puuh.

Are you ok with pining this action to the sha1 of the commit or shall we roll our own check?

src/redisearch_rs/trie_bencher/benches/operations.rs
DarthB: In CI we got a warning from the Benchmark library [Criterion](https://github.com/bheisler/criterion.rs) the increase to 10s was enough time for a mac with M4 but not for the github runner to get enough samples from the executions.

The benchmark framework takes samples from different points of time of the measurements and if the setup is expensive that parameter needs to be raised.

.github/workflows/flow-micro-benchmarks.yml
DarthB: I saw it elsewhere, will remove it there too

DarthB:
Addressed the comments. Thanks for the feedback.

DvirDukhan (approved):

PR #5941, opened 2025-04-15
fix flaky OOM tests MOD-9406

lerman25:
Fix flakiness of OOM testing by setting tighter limits.

### DIFF
--- a/tests/pytests/common.py
+++ b/tests/pytests/common.py
@@ -23,6 +23,7 @@ from unittest.mock import ANY, _ANY
 from unittest import SkipTest
 import inspect
 import subprocess
+import math
 
 BASE_RDBS_URL = 'https://dev.cto.redis.s3.amazonaws.com/RediSearch/rdbs/'
 REDISEARCH_CACHE_DIR = '/tmp/redisearch-rdbs/'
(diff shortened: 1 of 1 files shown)


tests/pytests/common.py
```diff
@@ -857,7 +857,11 @@ def set_tight_maxmemory_for_oom(env, memory_limit_per = 0.8):
     # Get current memory consumption value
     memory_usage = env.cmd('INFO', 'MEMORY')['used_memory']
     # Set memory limit to less then memory limit
-    env.expect('config', 'set', 'maxmemory', int(memory_usage*(1/(memory_limit_per-0.01)))).ok()
+    required_memory = memory_usage * (1/memory_limit_per)
+    # Round up and add 1
+    new_memory = int(required_memory) + bool(required_memory%1) + 1
```
GuyAv46, line 862: why not use `round`?

tests/pytests/common.py
```diff
@@ -857,7 +857,11 @@ def set_tight_maxmemory_for_oom(env, memory_limit_per = 0.8):
...
+
+    env.expect('config', 'set', 'maxmemory',new_memory).ok()
```
GuyAv46, line 864: Is this the flaky part? I believe that some internal redis stuff can consume memory so the new limit may not be enough. Maybe there is a way to ensure that? or should we relax the validation if that's the case?

tests/pytests/common.py
lerman25: round rounds to nearest rather then up,
I can change to round and add 2 it's not that important

GuyAv46: [math.ceil](https://docs.python.org/3/library/math.html#math.ceil) rounds up. But I agree this is not very important

PR #5948, opened 2025-04-16
MOD-8918, MOD-8920: Score normalization result processor for BM25STD.NORM scorer

Itzikvaknin:
This PR introduces a new scorer: `BM25STD.NORM`.
It uses the same scoring formula as `BM25STD`, but applies min-max normalization to produce scores in the range `0 <= score <= 1`.

This addition does not affect the module's behavior or performance in scenarios where `BM25STD.NORM` is not used.


tests/pytests/test_scorers.py
```diff
@@ -422,7 +422,49 @@ def testNormScore(env, stretch_factor, query_param=False):
...
+    norm_scores = []
+
+    for i in range(1, len(res), 2):
+        # The same order should be kept
+        env.assertEqual(res[i], norm_res_search[i])
+
+        env.assertGreaterEqual(float(norm_res_search[i+1]), 0)
+        env.assertLessEqual(float(norm_res_search[i+1]), 1)
+
+        # The score should be normalized using the stretched tanh function
```
nafraf, line 457: Please fix the comment.

src/result_processor.c
```diff
@@ -1276,3 +1277,101 @@ void PipelineAddCrash(struct AREQ *r) {
...
+   if (self->scoreKey) {
+     RLookup_WriteOwnKey(self->scoreKey, &r->rowdata, RS_NumVal(r->score));
+   }
+  //  EXPLAIN(r->scoreExplain,
+  //        "(Score: %.2f = (Original Score: %.2f - Min Score: %.2f) / (Max Score: %.2f - Min Score: %.2f))",
+  //        r->score, oldScore, self->minValue, self->maxValue, self->minValue);
+   return RS_RESULT_OK;
+ }
+
+static int rpNormelizorNext_innerLoop(ResultProcessor *rp, SearchResult *r) {
```
nafraf, line 1326:
```suggestion
static int rpNormalizerNext_innerLoop(ResultProcessor *rp, SearchResult *r) {
```

src/result_processor.c
```diff
@@ -1276,3 +1277,101 @@ void PipelineAddCrash(struct AREQ *r) {
...
+  self->pooledResult->indexResult = NULL;
+  array_ensure_append_1(self->pool, self->pooledResult);
+
+  // we need to allocate a new result for the next iteration
+  //TODO: Consider optimizing by allocating memory once every N iterations
+  self->pooledResult = rm_calloc(1, sizeof(*self->pooledResult));
+  return RESULT_QUEUED;
+}
+
+static int rpNormelizorAccum(ResultProcessor *rp, SearchResult *r) {
```
nafraf, line 1352:
```suggestion
static int rpNormalizerAccum(ResultProcessor *rp, SearchResult *r) {
```

src/result_processor.c
```diff
@@ -1276,3 +1277,101 @@ void PipelineAddCrash(struct AREQ *r) {
...
+   RPNormelizer* self = (RPNormelizer*)rp;
+   size_t length = array_len(self->pool);
+   if (array_len(self->pool) == 0) {
+      // We've already yielded all results, return EOF
+     return RS_RESULT_EOF;
+   }
+  SearchResult *poppedResult = array_pop(self->pool);
+  *r = *poppedResult;
+  rm_free(poppedResult);
+  //  double oldScore = r->score;
```
nafraf, line 1315: leftover?

raz-mon (commented):
Nice
See some initial comments
We need many more tests. For instance, that we have the normalizer in the right place in the pipeline for different commands as expected (after the scorer), that we do not send non-zero scores to zero etc.
Also, I think this should be added under the `ENABLE_UNSTABLE_FEATURES` for 2.10 at least at this point

tests/pytests/test_profile.py
```diff
@@ -674,3 +674,18 @@ def testPofileGILTime():
...
+  #create index
+  env.cmd('ft.create', 'idx', 'SCHEMA', 't', 'TEXT')
+
+  # Populate db
+  with env.getClusterConnectionIfNeeded() as conn:
+    conn.execute_command('HSET', 'doc1', 't', 'hello world')
+    conn.execute_command('HSET', 'doc2', 't', 'hello space world')
+    conn.execute_command('HSET', 'doc3', 't', 'hello more space world')
+
+  aggregate_response = env.cmd('FT.PROFILE', 'idx', 'AGGREGATE', 'query', 'hello', 'ADDSCORES', 'SCORER', 'BM25STD.NORM')
```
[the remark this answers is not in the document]
Itzikvaknin: Caused by division by zero, one shard had only one doc in it, so the max_score is equal to min_score, resulting in denominator value of 0 (max_score-min_score), I removed the minimum from the calculation as @raz-mon suggested and this bug no longer occurs.

tests/pytests/test_scorers.py
```diff
@@ -1,8 +1,8 @@
 import math
 from time import sleep
-
 from includes import *
 from common import *
+import functools
```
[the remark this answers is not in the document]
Itzikvaknin: We don't :)

src/result_processor.c
```diff
@@ -15,7 +15,8 @@
 #include "rmutil/rm_assert.h"
 #include "util/timeout.h"
 #include "util/arr.h"
-
+#include "util/minmax.h"
+#include <float.h>
```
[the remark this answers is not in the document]
Itzikvaknin: No

src/result_processor.c
```diff
@@ -1276,3 +1277,101 @@ void PipelineAddCrash(struct AREQ *r) {
...
+  *r = *poppedResult;
+  rm_free(poppedResult);
+  //  double oldScore = r->score;
+   r->score = (r->score - self->minValue) / (self->maxValue - self->minValue);
+   if (self->scoreKey) {
+     RLookup_WriteOwnKey(self->scoreKey, &r->rowdata, RS_NumVal(r->score));
+   }
+  //  EXPLAIN(r->scoreExplain,
+  //        "(Score: %.2f = (Original Score: %.2f - Min Score: %.2f) / (Max Score: %.2f - Min Score: %.2f))",
+  //        r->score, oldScore, self->minValue, self->maxValue, self->minValue);
```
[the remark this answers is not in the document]
Itzikvaknin: Added

src/result_processor.c
```diff
@@ -1276,3 +1277,101 @@ void PipelineAddCrash(struct AREQ *r) {
...
+
+ static int rpNormelizor_Yield(ResultProcessor *rp, SearchResult *r){
+   RPNormelizer* self = (RPNormelizer*)rp;
+   size_t length = array_len(self->pool);
+   if (array_len(self->pool) == 0) {
+      // We've already yielded all results, return EOF
+     return RS_RESULT_EOF;
+   }
+  SearchResult *poppedResult = array_pop(self->pool);
+  *r = *poppedResult;
```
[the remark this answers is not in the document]
Itzikvaknin: see `SearchResult_Override`

src/result_processor.c
Itzikvaknin: Using it in next commits

src/aggregate/aggregate_request.c
```diff
@@ -1489,6 +1489,19 @@ static void buildImplicitPipeline(AREQ *req, QueryError *Status) {
        (IsOptimized(req) ? HasScorer(req) : !hasQuerySortby(&req->ap)))) {
     rp = getScorerRP(req, first);
     PUSH_RP();
+    if (req->searchopts.scorerName && !strcmp(req->searchopts.scorerName, BM25_STD_NORMALIZED_MAX_SCORER_NAME )) {
```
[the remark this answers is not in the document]
Itzikvaknin: You're right — it does make sense to return the error regardless of whether `ADDSCORES` or `WITHSCORES `is used.
That said, we apply certain optimizations that skip the scoring step when it's redundant or unnecessary, even if it's explicitly requested. This happens in the query you mentioned.

I discussed it with @raz-mon , and for now, it doesn’t seem critical.

src/result_processor.c
```diff
@@ -1317,3 +1322,96 @@ void PipelineAddCrash(struct AREQ *r) {
...
+   size_t length = array_len(self->pool);
+   if (array_len(self->pool) == 0) {
+      // We've already yielded all results, return EOF
+     return RS_RESULT_EOF;
+   }
+  SearchResult *poppedResult = array_pop(self->pool);
+  SearchResult_Override(r, poppedResult);
+  rm_free(poppedResult);
+  double oldScore = r->score;
+  r->score = r->score / self->maxValue;
```
nafraf, line 1360: This is an edge case, should we return `NaN` or `0`?
```sh
FT.CREATE idx SCHEMA t TEXT
HSET doc1 t cherry
HSET doc2 t orange

127.0.0.1:6379> FT.SEARCH idx '-@t:(cherry)' WITHSCORES SCORER BM25STD.NORM DIALECT 2
1) (integer) 1
2) "doc2"
3) "nan"
4) 1) "t"
   2) "orange"
```

Itzikvaknin: Fixed in 042c0da

raz-mon (commented):
Nice!
See some comments

src/result_processor.c
```diff
@@ -1380,6 +1377,7 @@ static int rpNormalizerNext_innerLoop(ResultProcessor *rp, SearchResult *r) {
     return rp->Next(rp, r);
   } else if (rc != RS_RESULT_OK) {
     // whoops!
+    RS_LOG_ASSERT(rc != RS_RESULT_OK, "Received an unexpected result status that is neither EOF nor OK");
```
raz-mon, line 1380: Use `RS_ABORT`, since we already know to what the condition evaluates

src/result_processor.c
```diff
@@ -1324,10 +1324,7 @@ void PipelineAddCrash(struct AREQ *r) {
 }
 
  /*******************************************************************************************************************
-  *  Max Result Processor
-  *
-  * This processor first depletes the result pipeline by iterating through all results once.
-  * It can be used to collect maximum values or any other aggregation that requires a single pass.
+  *  Max Score Normalizer Result Processor
```
raz-mon, line 1327: Let's keep the documentation, just fix it to be more accurate

src/result_processor.c
```diff
@@ -1389,18 +1387,16 @@ static int rpNormalizerNext_innerLoop(ResultProcessor *rp, SearchResult *r) {
...
 }
 
 static int rpNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
   RPNormalizer *self = (RPNormalizer *)rp;
   uint32_t chunkLimit = rp->parent->resultLimit;
-  rp->parent->resultLimit = UINT32_MAX; // we want to accumulate all results
   int rc;
   int count = 0;
-  while ((rc = rpNormalizerNext_innerLoop(rp, r)) == RESULT_QUEUED) {}
+  while (count < chunkLimit && (rc = rpNormalizerNext_innerLoop(rp, r)) == RESULT_QUEUED);
```
raz-mon, line 1399: Where is `count` updated?

tests/pytests/test_scorers.py
```diff
@@ -483,58 +483,56 @@ def testNormScore(env, stretch_factor, query_param=False):
...
-def testBM25NormMax():
-    '''
-    The normalization process should be integrated with the scoring mechanism, as we're introducing it as a scorer type: BM25STD.NORM.
-    This integration ensures that each document in the result set, derived from the index iterator, receives a consistent normalized score.
-    For instance, applying a LIMIT 0 2 clause doesn't alter the scores considered for normalization, since we exhaust the index iterator before applying the limit.
-    '''
 
-
-    def create_index(env, use_key_tags=False):
+class TestBM25NormMax:
```
raz-mon, line 487: 👍

raz-mon:
@Itzikvaknin Let's also add a test (or a few) for cursor behavior, with and without sorting.
Specifically I'm a bit worried about the behavior of the non-sorting command, e.g., on each read will we get a different normalizer?

src/result_processor.c
```diff
@@ -1317,3 +1322,98 @@ void PipelineAddCrash(struct AREQ *r) {
...
+  SearchResult_Override(r, poppedResult);
+  rm_free(poppedResult);
+  double oldScore = r->score;
+  if (r->score != 0) {
+    r->score /= self->maxValue;
+  }
+  if (self->scoreKey) {
+    RLookup_WriteOwnKey(self->scoreKey, &r->rowdata, RS_NumVal(r->score));
+  }
+  EXPLAIN(r->scoreExplain,
```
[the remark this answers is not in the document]
Itzikvaknin: `void explain(RSScoreExplain *scrExp, char *fmt, ...) {
  void *tempStr = scrExp->str;

  va_list ap;
  va_start(ap, fmt);
  rm_vasprintf((char **restrict) & scrExp->str, fmt, ap);
  va_end(ap);

  rm_free(tempStr);
}`

`rm_free(tempStr)` frees the original string previously held in `scrExp->str` after it’s replaced by a new string.

src/result_processor.c
```diff
@@ -1317,3 +1322,98 @@ void PipelineAddCrash(struct AREQ *r) {
...
+  RPNormalizer *self = (RPNormalizer *)rp;
+
+  // get the next result from upstream. `self->pooledResult` is expected to be empty and allocated.
+  int rc = rp->upstream->Next(rp->upstream, self->pooledResult);
+  // if our upstream has finished - just change the state to not accumulating, and yield
+  if (rc == RS_RESULT_EOF) {
+    rp->Next = rpNormalizer_Yield;
+    return rp->Next(rp, r);
+  } else if (rc != RS_RESULT_OK) {
+    // whoops!
```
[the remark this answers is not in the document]
raz-mon: Actually, I got confused from the "whoops!".. We should not assert here. This is fine, for instance, we may have reached the timeout time and got a timeout.
Let's remove the silly comment and the assert.

tests/pytests/test_scorers.py
```diff
@@ -624,6 +603,21 @@ def test_no_results(self):
         agg = self.env.cmd('FT.AGGREGATE', 'idx', query, 'ADDSCORES', 'SCORER', 'BM25STD.NORM')
         self.env.assertEqual(len(agg), 1)
 
+    def test_cursor(self):
+        agg_norm = self.env.cmd('FT.AGGREGATE', 'idx', 'hello world', 'ADDSCORES', 'SCORER', 'BM25STD.NORM',
+                        'LOAD', '2', '@__key', '@__score')
+
+        key_index = agg_norm[1].index('__key') + 1
+        score_index = agg_norm[1].index('__score') + 1
+        scores_std = {row[key_index]: float(row[score_index]) for row in agg_norm[1:]}
```
raz-mon, line 612:
```sug
scores_norm = {row[key_index]: float(row[score_index]) for row in agg_norm[1:]}
```

raz-mon (commented):
One bug and one small comment

raz-mon (approved):
Bravo

src/result_processor.c
```diff
@@ -1317,3 +1322,101 @@ void PipelineAddCrash(struct AREQ *r) {
...
+static int rpNormalizerNext_innerLoop(ResultProcessor *rp, SearchResult *r) {
+  RPNormalizer *self = (RPNormalizer *)rp;
+  // get the next result from upstream. `self->pooledResult` is expected to be empty and allocated.
+  int rc = rp->upstream->Next(rp->upstream, self->pooledResult);
+  // if our upstream has finished - just change the state to not accumulating, and yield
+  if (rc == RS_RESULT_EOF) {
+    rp->Next = rpNormalizer_Yield;
+    return rp->Next(rp, r);
+  } else if (rc != RS_RESULT_OK) {
+    return rc;
```
raz-mon, line 1388: We need to handle timeouts the same way the sorter does - return the partial results on non-strict timeout policy

raz-mon (dismissed):
Nicely done

src/result_processor.c
```diff
@@ -1317,3 +1322,108 @@ void PipelineAddCrash(struct AREQ *r) {
...
+  self->pooledResult = rm_calloc(1, sizeof(*self->pooledResult));
+  return RESULT_QUEUED;
+}
+
+static int rpNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
+  RPNormalizer *self = (RPNormalizer *)rp;
+  uint32_t chunkLimit = rp->parent->resultLimit;
+  rp->parent->resultLimit = UINT32_MAX; // we want to accumulate all results
+  int rc;
+  while ((rc = rpNormalizerNext_innerLoop(rp, r)) == RESULT_QUEUED);
```
oshadmi, line 1413: `while` without a body - could be misleading

src/result_processor.c
```diff
@@ -1317,3 +1322,108 @@ void PipelineAddCrash(struct AREQ *r) {
...
+ /* Create a new Max Collector processor */
+ ResultProcessor *RPNormalizer_New(const RLookupKey *rlk) {
+  RPNormalizer *ret = rm_calloc(1, sizeof(*ret));
+  ret->pooledResult = rm_calloc(1, sizeof(*ret->pooledResult));
+  ret->pool = array_new(SearchResult*, 0);
+  ret->base.Next = rpNormalizer_Accum;
+  ret->base.Free = rpNormalizer_Free;
+  ret->base.type = RP_MAX_SCORE_NORMALIZER;
+  ret->scoreKey = rlk;
+  ret->maxValue = 0;
```
oshadmi, line 1427: Redundant since allocated using `rm_calloc`?

src/aggregate/aggregate_request.c
```diff
@@ -1489,8 +1494,17 @@ static void buildImplicitPipeline(AREQ *req, QueryError *Status) {
        (IsOptimized(req) ? HasScorer(req) : !hasQuerySortby(&req->ap)))) {
     rp = getScorerRP(req, first);
     PUSH_RP();
+    const char *scorerName = req->searchopts.scorerName;
+    if (scorerName && !strcmp(scorerName, BM25_STD_NORMALIZED_MAX_SCORER_NAME )) {
```
oshadmi, line 1498: In backport - will need to check for unstable features flag, right?

oshadmi (commented):
Some comments

src/result_processor.c
```diff
@@ -1317,3 +1322,108 @@ void PipelineAddCrash(struct AREQ *r) {
...
+    int ret = self->timedOut ? RS_RESULT_TIMEDOUT : RS_RESULT_EOF;
+    self->timedOut = false;
+    return ret;
+   }
+  SearchResult *poppedResult = array_pop(self->pool);
+  SearchResult_Override(r, poppedResult);
+  rm_free(poppedResult);
+  double oldScore = r->score;
+  if (r->score != 0) {
+    r->score /= self->maxValue;
```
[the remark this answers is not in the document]
Itzikvaknin: `self->maxValue >= r->score` for every r in the result set
so its protected by `if (r->score != 0) {}`
but it does make sense switching to `if (self->maxValue != 0) {} `

src/aggregate/aggregate_request.c
Itzikvaknin: Its already checked in this PR, see `AREQ_ApplyContext`

src/aggregate/aggregate_request.c
```diff
@@ -1487,6 +1487,14 @@ static void buildImplicitPipeline(AREQ *req, QueryError *Status) {
        (IsOptimized(req) ? HasScorer(req) : !hasQuerySortby(&req->ap)))) {
     rp = getScorerRP(req, first);
     PUSH_RP();
+    if (req->searchopts.scorerName && !strcmp(req->searchopts.scorerName, BM25_STD_NORMALIZED_MIN_MAX_SCORER_NAME)) {
```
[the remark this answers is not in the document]
oshadmi: We need to enable this normalization with the unstable feature flag for backports (2.10, 8.0)

src/aggregate/aggregate_request.c
```diff
@@ -1147,9 +1147,14 @@ int AREQ_ApplyContext(AREQ *req, RedisSearchCtx *sctx, QueryError *status) {
...
 
-  if (opts->scorerName && (Extensions_GetScoringFunction(NULL, opts->scorerName) == NULL)) {
-    QueryError_SetWithoutUserDataFmt(status, QUERY_EINVAL, "No such scorer %s", opts->scorerName);
-    return REDISMODULE_ERR;
+  if (opts->scorerName) {
+    if (Extensions_GetScoringFunction(NULL, opts->scorerName) == NULL) {
+      QueryError_SetWithoutUserDataFmt(status, QUERY_EINVAL, "No such scorer %s", opts->scorerName);
+      return REDISMODULE_ERR;
+    } else if (!strcmp(opts->scorerName, BM25_STD_NORMALIZED_MAX_SCORER_NAME) && !RSGlobalConfig.enableUnstableFeatures) {
+      QueryError_SetWithoutUserDataFmt(status, QUERY_EINVAL, "Scorer %s not available when `ENABLE_UNSTABLE_FEATURES` is off", opts->scorerName);
```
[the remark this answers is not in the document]
oshadmi: On master we do not need to check for this config. When backporting to 8.0 we should use the new `search-enable-unstable-features`. When backporting to 2.10 we should use the old `ENABLE_UNSTABLE_FEATURES`

tests/pytests/test_scorers.py
```diff
@@ -435,7 +482,150 @@ def testNormScore(env, stretch_factor, query_param=False):
...
+        for row in agg[1:]:
+            self.env.assertAlmostEqual(float(row[1]), 1.0, 0.00001)
+
+    def test_no_results(self):
+        query = 'no match term'
+        res = self.env.cmd('FT.SEARCH', 'idx', query, 'WITHSCORES', 'SCORER', 'BM25STD.NORM', 'NOCONTENT')
+        self.env.assertEqual(res[0], 0)
+
+        agg = self.env.cmd('FT.AGGREGATE', 'idx', query, 'ADDSCORES', 'SCORER', 'BM25STD.NORM')
+        self.env.assertEqual(len(agg), 1)
```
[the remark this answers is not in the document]
oshadmi: Right, for zero results count should be zero also for AGGREGATE

src/aggregate/aggregate_request.c
oshadmi: On master no need to check for unstable feature flag. Only needed on backports.

nafraf (approved):

PR #5949, opened 2025-04-16
MOD-9368: Add configurable stretch factor for normalized BM25STD.TANH

raz-mon:
Adds a configurable stretch factor `search-bm25std-tanh-factor` (`BM25STD_TANH_FACTOR` for `_FT.CONFIG`) for normalized BM25STD.TANH, in the range of `[1, 10000]`, such that the user can set it according to the range relevant for his use-case.


src/config.c
```diff
@@ -443,6 +444,26 @@ CONFIG_GETTER(getIndexingMemoryLimit) {
...
+      BM25STD_TANH_STRETCH_MIN, BM25STD_TANH_STRETCH_MAX);
+    return REDISMODULE_ERR;
+  }
+  config->BM25STD_TanhStretchFactor = newStretch;
+  return REDISMODULE_OK;
+}
+
+CONFIG_GETTER(getBM25StdTanhStretch) {
+  sds ss = sdsempty();
+  return sdscatprintf(ss, "%u", config->BM25STD_TanhStretchFactor);
```
GuyAv46, line 464:
```suggestion
  return sdscatprintf(ss, "%ul", config->BM25STD_TanhStretchFactor);
```

src/ext/default.c
```diff
@@ -313,13 +313,13 @@ static double BM25StdNormScorer(const ScoringFunctionArgs *ctx, const RSIndexRes
           dmd->score);
 
   // Normalize the score
-  double normalizedScore = tanhStretched(score, NORMALIZED_BM25_STRETCH_FACTOR);
+  double normalizedScore = tanhStretched(score, RSGlobalConfig.BM25STD_TanhStretchFactor);
```
GuyAv46, line 316: query configurations that are used from the background and can be modified after module load should get copied to the query context and read from there, not from the global struct

src/ext/default.c
```diff
@@ -313,13 +313,13 @@ static double BM25StdNormScorer(const ScoringFunctionArgs *ctx, const RSIndexRes
...
 
   // Modify the explanation to include the normalization
   strExpCreateParent(ctx, &scrExp);
   EXPLAIN(scrExp,
     "Final Normalized BM25 : tanh(stretch factor 1/%d * Final BM25 %.2f)",
-    NORMALIZED_BM25_STRETCH_FACTOR, score);
+    RSGlobalConfig.BM25STD_TanhStretchFactor, score);
```
GuyAv46, line 322: query configurations that are used from the background and can be modified after module load should get copied to the query context and read from there, not from the global struct

deps/rmutil/args.c
```diff
@@ -136,10 +136,10 @@ int AC_GetLongLong(ArgsCursor *ac, long long *ll, int flags) {
 GEN_AC_FUNC(AC_GetUnsignedLongLong, unsigned long long, 0, LLONG_MAX, 1)
 GEN_AC_FUNC(AC_GetUnsigned, unsigned, 0, UINT_MAX, 1)
 GEN_AC_FUNC(AC_GetInt, int, INT_MIN, INT_MAX, 0)
-GEN_AC_FUNC(AC_GetU32, uint32_t, 0, UINT32_MAX, 1)
+GEN_AC_FUNC(AC_GetU8, uint8_t, 0, UINT8_MAX, 1)
+GEN_AC_FUNC(AC_GetU16, uint16_t, 0, UINT16_MAX, 1)
```
Itzikvaknin, line 140: Looks like this isn't used in this PR — is it a generalization of the current changes? Also, what's the reason for removing u32?

raz-mon: Meant to remove the 16 one that I added.. Fixing thanks

src/ext/default.c
raz-mon: Right thanks

src/config.c
```diff
@@ -1231,6 +1252,12 @@ RSConfigOptions RSGlobalConfigOptions = {
                       " any queries on the affected index will result in an error. The default is 80 percent.",
          .setValue = setIndexingMemoryLimit,
          .getValue = getIndexingMemoryLimit},
+        {.name = "BM25STD_TANH_STRETCH",
+          .helpText = "Set the BM25STD.TANH stretch factor. This is the an integer value that divides the argument"
```
nafraf, line 1256:
```suggestion
          .helpText = "Set the BM25STD.TANH stretch factor. This is an integer value that divides the argument"
```

src/config.c
raz-mon: Seems that `sds` does not support that option

GuyAv46: 😮

src/config.h
```diff
@@ -158,7 +160,7 @@ typedef struct {
   // Control user data obfuscation in logs
   bool hideUserDataFromLog;
   // Stretch factor for BM25STD.TANH scoring function
-  uint64_t BM25STD_TanhStretchFactor;
+  uint64_t BM25STD_TanhFactor;
```
GuyAv46, line 163: redundant and duplicating here (we have a `requestConfigParams` section)

raz-mon: Right forgot to clean that

GuyAv46 (approved):

nafraf (approved):

PR #5955, opened 2025-04-17
[2.10] MOD-9368: Add configurable stretch factor for normalized BM25STD.TANH…

raz-mon:
CP of #5949 to `2.10`.

### DIFF
--- a/tests/pytests/test_config.py
+++ b/tests/pytests/test_config.py
@@ -1,6 +1,7 @@
 from RLTest import Env
 from includes import *
 from common import skip, config_cmd
+import math
 
 not_modifiable = 'Not modifiable at runtime'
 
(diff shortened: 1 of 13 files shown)


tests/pytests/test_config.py
```diff
@@ -188,6 +192,7 @@ def test_arg_num(arg_name, arg_value):
     test_arg_num('BG_INDEX_SLEEP_GAP', 15)
     test_arg_num('MINSTEMLEN', 3)
     test_arg_num('INDEX_CURSOR_LIMIT', 128)
+    test_arg_num('BM25STD_TANH_FACTOR', 8)
```
oshadmi, line 195: Ideally would be as in master (4 instead of 8 - although using `noDefaultModuleArgs` and testing the same behavior)

PR #5958, opened 2025-04-19
Fix scores when document has no text - [MOD-9423]

GuyAv46:

Fix an edge case where the document has no text at all, causing its `maxFreq` and `len` to be zero, and attempting to use this values to normalize the scores yields `inf` for non-zero scores, and `nan` for zero scores (`x/0` and `0/0` respectively). This causes the obviously not matching document to score very high, yielding meaningless results that are preferred over better matches, or just messing with the sorting when comparing scores to `nan`

#### Mark if applicable

### DIFF
--- a/src/ext/default.c
+++ b/src/ext/default.c
@@ -103,6 +103,10 @@ static inline double tfIdfInternal(const ScoringFunctionArgs *ctx, const RSIndex
     return 0;
   }
   uint32_t norm = normMode == NORM_MAXFREQ ? dmd->maxFreq : dmd->len;
+  if (norm == 0) {
+    EXPLAIN(scrExp, "Document %s is 0", normMode == NORM_MAXFREQ ? "max frequency" : "length");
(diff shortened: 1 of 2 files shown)


kei-nan (approved):

raz-mon (approved):

src/ext/default.c
```diff
@@ -103,6 +103,10 @@ static inline double tfIdfInternal(const ScoringFunctionArgs *ctx, const RSIndex
     return 0;
   }
   uint32_t norm = normMode == NORM_MAXFREQ ? dmd->maxFreq : dmd->len;
+  if (norm == 0) {
```
oshadmi, line 106: What about BM25? Can some fields also be zero there?

oshadmi: Seems OK
https://github.com/RediSearch/RediSearch/blob/2.10/src/ext/default.c#L157

PR #5986, opened 2025-04-22
Avoid lazy expiration in background indexing for CRDT [MOD-9486]

lerman25:
Currently, during the background scan, using RM_KeyType to check the key type causes the key to expire in CRDT, due to CRDT's RM_KeyType function.
This continues the work from #5274.


src/spec.c
DvirDukhan, line 1: I think you need to verify that the scanner didn't open the key, since it might be left open.

PR #5996, opened 2025-04-23
Varint oxidization

zeenix:

1. Current: The `varint` module is currently in C.
2. Change: This change ports all the implementation of `varint` module to Rust, along with adding a more idiomatic Rust API.
3. Outcome: This PR is the first step towards the complete port of the Inverted Index to Rust.


src/redisearch_rs/encode_decode/src/varint/vector_writer.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /// Write an integer to the vector.
+    ///
+    /// # Return Value
+    ///
+    /// The number of bytes written to the vector.
+    pub fn write(&mut self, value: u32) -> std::io::Result<usize> {
+        // Assumes `value` is greater than `last_value` (copied over from the C code).
+        let bytes_written = write(value - self.last_value, &mut self.buffer)?;
```
LukeMathWalker, line 27: Worth calling this out in the function documentation.
Should we include a `debug_assert` here to ensure the invariant isn't violated in tests? Unless we have tests exercising this specific scenario already.

src/redisearch_rs/encode_decode/src/buffer.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// Redefines the `BufferWriter` struct from `buffer.h`
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct BufferWriter {
+    pub buf: *mut Buffer,
+    pub pos: *mut u8,
+}
+
+impl From<BufferWriter> for Cursor<Vec<u8>> {
```
DarthB, line 39:
```suggestion
impl From<BufferWriter> for Cursor<&mut [u8]> {
```

That would fit the Reader and is a bit more generic

zeenix: but what's the lifetime of the reference then?

zeenix: Also to keep in mind these are supposed to be transient until the whole of II is ported.

DarthB: no action required

src/redisearch_rs/encode_decode/src/varint/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl DerefMut for VarintBuf {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
```
DarthB, line 123: some unit tests as sanity checks:
a == decode(encode(a))
use a with numbers that need different amount of bytes

src/redisearch_rs/encode_decode/src/varint/c_api.rs
```diff
@@ -0,0 +1,144 @@
...
+    FieldMask,
+    buffer::{BufferReader, BufferWriter},
+};
+
+#[unsafe(no_mangle)]
+extern "C" fn ReadVarint(mut b: NonNull<BufferReader>) -> u32 {
+    // Safety: The caller is responsible for ensuring that the pointer is valid.
+    let buffer_reader = unsafe { b.as_mut() };
+    let mut cursor = buffer_reader.as_cursor();
+    let val = read(&mut cursor).unwrap();
```
DarthB, line 17:
```suggestion
    let val = read(&mut cursor).map_err(|e| panic!("failed decode varint: {}", e).unwrap();
```

src/redisearch_rs/encode_decode/src/varint/c_api.rs
```diff
@@ -0,0 +1,144 @@
...
+#[unsafe(no_mangle)]
+// `improper_ctypes_definitions` would be triggered because of u128 crossing the FFI boundary but
+// that's no longer an issue:
+// https://blog.rust-lang.org/2024/03/30/i128-layout-update/#compatibility
+#[allow(improper_ctypes_definitions)]
+extern "C" fn ReadVarintFieldMask(mut b: NonNull<BufferReader>) -> FieldMask {
+    // Safety: The caller is responsible for ensuring that the pointer is valid.
+    let buffer_reader = unsafe { b.as_mut() };
+    let mut cursor = buffer_reader.as_cursor();
+    let val = read_field_mask(&mut cursor).unwrap();
```
DarthB, line 32: as above... and so on

src/redisearch_rs/encode_decode/src/varint/c_api.rs
```diff
@@ -0,0 +1,144 @@
...
+extern "C" fn VVW_Write(mut writer: NonNull<VectorWriter>, value: u32) -> usize {
+    // Safety: The caller is responsible for ensuring that the pointer is valid.
+    unsafe { writer.as_mut() }.write(value).unwrap()
+}
+
+#[unsafe(no_mangle)]
+extern "C" fn VVW_GetByteData(writer: *const VectorWriter) -> *const u8 {
+    if writer.is_null() {
+        return std::ptr::null();
+    }
```
DarthB, line 83:
```suggestion
   #[cfg(debug_assertions)]
    if writer.is_null() {
        return std::ptr::null();
    }
```

DarthB: Or were this checks on the c-side too?

src/redisearch_rs/encode_decode/src/varint/c_api.rs
zeenix: panic from the unwrap() contains the error and location (one can backtrace to find more).

src/redisearch_rs/encode_decode/src/varint/c_api.rs
zeenix: Actually I think at least in this case null is acceptable. Otherwise I would have only taken `NonNull` instead of a raw pointer. I'll revert this, ok?

src/redisearch_rs/encode_decode/src/varint/c_api.rs
DarthB: I had some outdated information here. Nowadays FFI panic output to stderr is much better than what I had in mind.

DarthB:
We should rebase on master again - the changes in https://github.com/RediSearch/RediSearch/commit/a0fb945982559db5d6841f8172259f30bd30692e change the structure for our ffi modules (all are moved to the redissearch_rs crate)

src/redisearch_rs/encode_decode/src/buffer.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+impl From<Cursor<Vec<u8>>> for BufferWriter {
+    fn from(cursor: Cursor<Vec<u8>>) -> Self {
+        let pos = cursor.position() as usize;
+        let mut buffer_vec = ManuallyDrop::new(cursor.into_inner());
+        let buffer_ptr = buffer_vec.as_mut_ptr();
+
+        BufferWriter {
```
DarthB, line 61: Check for regressions because of `Box::new`

zeenix: Oh, that's right. We loose the original buffer pointer when going back to C. I think we need a new type here that provides `Write` impl.

src/redisearch_rs/encode_decode/src/varint/c_api.rs
```diff
@@ -0,0 +1,134 @@
...
+#[unsafe(no_mangle)]
+#[allow(improper_ctypes_definitions)]
+extern "C" fn WriteVarintFieldMask(value: FieldMask, b: NonNull<BufferWriter>) -> usize {
+    let mut wrapper = BufferWriterWrapper::from(b);
+    write_field_mask(value, &mut wrapper).unwrap()
+}
+
+#[unsafe(no_mangle)]
+extern "C" fn NewVarintVectorWriter(cap: usize) -> NonNull<VectorWriter> {
+    let vector_writer = Box::leak(Box::new(VectorWriter::new(cap)));
```
LukeMathWalker, line 57: I assume `Box::into_raw` would be more aligned with the intent here?
`Box::into_non_null` would be even better, but it's still on `nightly` unfortunately.

src/redisearch_rs/encode_decode/src/varint/c_api.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    // Safety: The pointer is valid because we just created it.
+    unsafe { NonNull::new_unchecked(vector_writer) }
+}
+
+#[unsafe(no_mangle)]
+extern "C" fn VVW_Write(mut writer: NonNull<VectorWriter>, value: u32) -> usize {
+    // Safety: The caller is responsible for ensuring that the pointer is valid.
+    unsafe { writer.as_mut() }.write(value).unwrap()
+}
```
LukeMathWalker, line 67: Let's be defensive wherever the pointer may end up being null.
Let's get a raw pointer, check (if debug assertions are enabled) if it's not null, and the convert it into a `NonNull` instance (using `new_unchecked` if debug assertions are disabled).

The more breadcrumbs we leave for ourselves when debugging, the better.

src/redisearch_rs/encode_decode/src/varint/vector_writer.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /// The number of members written to the vector.
+    #[inline(always)]
+    pub fn count(&self) -> usize {
+        self.n_members
+    }
+
+    /// Truncate the vector.
+    pub fn truncate(&mut self) -> usize {
```
LukeMathWalker, line 68: Let's call this `shrink_to_fit`, rather than `truncate`, since the latter has a different meaning in the Rust universe.
We can just map one name to the other in the C API layer, for now.

src/redisearch_rs/encode_decode/src/varint/c_api.rs
zeenix: > Let's be defensive wherever the pointer may end up being null.

But `NonNull` will already panic here if called with a null pointer from the C side. If we're inside the function, we know that the pointer is at least non-null.

LukeMathWalker:
Let's split this PR in 3 smaller ones:
1. `varint`and `VectorWriter` pure Rust implementations, with docs, tests and microbenchmarks
2. `Buffer`. Let's add tests and document our safety invariants, following the same style we use in `trie_rs`/`low_memory_thin_vec` (basically [this style](https://jack.wrenn.fyi/blog/safety-hygiene/)).
3. The C integration

PR #6001, opened 2025-04-23
MOD-9506: Lazily subscribe to key-space notifications

raz-mon:
This PR changes the time of the subscription of the RediSearch module to the key-space notifications to index creation time, i.e., a "lazy" subscription, in order to avoid unnecessary performance penalties in cases where the module is not used.


raz-mon:
Checked manually, and will be verified in further performance testing.
We can possibly add a test using a new notification that will turn on some flag, for testing purposes only.

GuyAv46 (dismissed):

oshadmi (approved):

src/spec.c
```diff
@@ -487,6 +488,13 @@ IndexSpec *IndexSpec_CreateNew(RedisModuleCtx *ctx, RedisModuleString **argv, in
     IndexSpec_SetTimeoutTimer(sp, StrongRef_Demote(spec_ref));
   }
 
+  // (Lazily) Subscribe to keyspace notifications, now that we have at least one
+  // spec
+  if (!RS_KeyspaceEvents_Initialized) {
+    Initialize_KeyspaceNotifications(ctx);
+    RS_KeyspaceEvents_Initialized = true;
+  }
```
MeirShpilraien, line 496: Why not moving this logic into `Initialize_ServerEventNotifications` instead of writing it 3 times?

src/notifications.c
```diff
@@ -376,14 +376,16 @@ void Initialize_KeyspaceNotifications(RedisModuleCtx *ctx) {
     REDISMODULE_NOTIFY_EXPIRED | REDISMODULE_NOTIFY_EVICTED |
     REDISMODULE_NOTIFY_LOADED | REDISMODULE_NOTIFY_MODULE,
     HashNotificationCallback);
+}
```
MeirShpilraien, line 379: I wonder if we should switch to use the `staticCtx` on this function, if we accidentally call it with a ctx that do not have the module pointer (ctx->module), it will cause Redis to crash when fire the key space notification. `staticCtx` has the module ctx initialized for sure.

MeirShpilraien (commented):
:+1: 
2 small comments.

src/spec.c
raz-mon: `Initialize_ServerEventNotifications` is called on server load where we don't want to subscribe to these notifications.
I can gather these lines to a function if that's what you mean

src/notifications.c
raz-mon: Using `RSDummyContext` 👍

GuyAv46 (dismissed):

src/notifications.c
```diff
@@ -369,21 +369,26 @@ void ConfigChangedCallback(RedisModuleCtx *ctx, RedisModuleEvent eid, uint64_t e
...
 
-void Initialize_KeyspaceNotifications(RedisModuleCtx *ctx) {
-  RedisModule_SubscribeToKeyspaceEvents(ctx,
-    REDISMODULE_NOTIFY_GENERIC | REDISMODULE_NOTIFY_HASH |
-    REDISMODULE_NOTIFY_TRIMMED | REDISMODULE_NOTIFY_STRING |
-    REDISMODULE_NOTIFY_EXPIRED | REDISMODULE_NOTIFY_EVICTED |
-    REDISMODULE_NOTIFY_LOADED | REDISMODULE_NOTIFY_MODULE,
-    HashNotificationCallback);
+void Initialize_KeyspaceNotifications() {
+  if (!RS_KeyspaceEvents_Initialized) {
```
MeirShpilraien, line 373:
```suggestion
  static bool RS_KeyspaceEvents_Initialized = false;
  if (!RS_KeyspaceEvents_Initialized) {

```
You can put `RS_KeyspaceEvents_Initialized` here as a static variable and then you do not need to define it as global.

src/module.h
```diff
@@ -48,10 +48,12 @@ void RediSearch_CleanupModule(void);
...
 int SpellCheckCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc);
 
-/** Module-level dummy context for certain dummy RM_XXX operations */
+// Module-level dummy context for certain dummy RM_XXX operations
 extern RedisModuleCtx *RSDummyContext;
-/** Indicates that RediSearch_Init was called */
+// Indicates that RediSearch_Init was called
 extern int RS_Initialized;
+// Indicates that RediSearch has subscribed to keyspace events
+extern bool RS_KeyspaceEvents_Initialized;
```
MeirShpilraien, line 56: Also you do not need this.

MeirShpilraien (dismissed):
Small suggestion but not mandatory.

src/notifications.c
raz-mon: Right 👍

MeirShpilraien (approved):

PR #6002, opened 2025-04-23
Avoid lazy expiration in background indexing for CRDT [MOD-9486]

lerman25:
Currently, during the background scan, using RM_KeyType to check the key type causes the key to expire in CRDT, due to CRDT's RM_KeyType function.
To solve, if we are on crdt, we open the key in the scan with the noeffect flag. 
This continues the work from https://github.com/RediSearch/RediSearch/pull/5274.


DvirDukhan:
Did we validate it with the CRDT team?

alonre24 (approved):

PR #6005, opened 2025-04-23
Fix Cursor Purge - [MOD-9408]

GuyAv46:

Fixing the cursors API so one can safely call `FT.CURSOR DEL` on some cursor currently being read by another connection on another thread.

Instead of attempting to delete the requested cursor, if we see that it is currently being used, we mark it for deletion. When the using thread finishes with the cursor, it will delete the cursor even if it was not depleted.

The issue can only happen using `WORKERS` or cluster mode (Coordinator).

1. MOD-9432
2. MOD-9433
3. MOD-9434
4. MOD-9435

1. Cursor API internals

#### Mark if applicable

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -59,12 +59,12 @@ static void Cursor_RemoveFromIdle(Cursor *cur) {
 }
 
 /* Assumed to be called under the cursors global lock or upon server shut down. */
-static void Cursor_FreeInternal(Cursor *cur, khiter_t khi) {
+static void Cursor_FreeInternal(Cursor *cur) {
   CursorList *cl = getCursorList(cur->is_coord);
+  khiter_t khi = kh_get(cursors, cl->lookup, cur->id);
+
   /* Decrement the used count */
   RS_LOG_ASSERT(khi != kh_end(cl->lookup), "Iterator shouldn't be at end of cursor list");
-  RS_LOG_ASSERT(kh_get(cursors, cl->lookup, cur->id) != kh_end(cl->lookup),
-                                                    "Cursor was not found");
   kh_del(cursors, cl->lookup, khi);
   RS_LOG_ASSERT(kh_get(cursors, cl->lookup, cur->id) == kh_end(cl->lookup),
                                                     "Failed to delete cursor");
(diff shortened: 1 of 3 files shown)


src/cursor.c
```diff
@@ -281,26 +315,19 @@ Cursor *Cursors_TakeForExecution(CursorList *cl, uint64_t cid) {
...
-  } else {
-    rc = REDISMODULE_ERR;
-  }
+  int rc = CursorList_Delete_Internal(cl, cid, false);
   CursorList_Unlock(cl);
   return rc;
 }
 
 int Cursor_Free(Cursor *cur) {
-  return Cursors_Purge(getCursorList(cur->is_coord), cur->id);
```
alonre24, line 303: Can we unify `Cursors_Purge` and `Cursor_Free`? Actually, neither of them is really freeing the memory, right? Better name would be `Cursor_Delete` IMO

src/cursor.c
```diff
@@ -87,6 +87,32 @@ static void Cursor_FreeInternal(Cursor *cur, khiter_t khi) {
   rm_free(cur);
 }
 
+static int CursorList_Delete_Internal(CursorList *cl, uint64_t cid, bool owned) {
```
MeirShpilraien, line 90: Why not split this function to 3 logical units:
1. Mark cursor for deletion
2. Unlink cursor (remove from idle list)
3. Free cursor structure

I think that it is more elegant instead of sending a boolean value indicating where we were called from. WDYT?

tests/cpptests/test_cpp_cursors.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+  ASSERT_EQ(Cursors_Purge(&g_CursorsList, id), REDISMODULE_OK) << "Should be able to mark for deletion";
+  ASSERT_EQ(Cursors_TakeForExecution(&g_CursorsList, id), nullptr) << "Cursor already deleted";
+  ASSERT_TRUE(cur->delete_mark);
+
+  ASSERT_EQ(Cursors_GetInfoStats().total_user, 1) << "Cursor should be alive";
+  ASSERT_EQ(Cursor_Free(cur), REDISMODULE_OK) << "Cursor should be deleted";
+  ASSERT_EQ(Cursors_GetInfoStats().total_user, 0) << "Cursor should be deleted";
+
+}
```
alonre24, line 66: Let's also add a short stress test with MT that simulates the faulty scenario - both here in unit test and as a flow test 
(Something like @MeirShpilraien's reproduction):

```
from multiprocessing import Process
import redis
r = redis.Redis()



def del_cursor(cursor):
    r = redis.Redis()
    r.execute_command('FT.CURSOR', 'DEL', 'idx', cursor)
    print('hello', name)

while True:
    cursor = r.execute_command('FT.AGGREGATE', 'idx', 'foo', 'WITHCURSOR', 'COUNT', '500000')[1]

    p = Process(target=del_cursor, args=(str(cursor),))
    p.start()

    while cursor != 0:
        cursor = r.execute_command('FT.CURSOR', 'READ', 'idx', cursor)[1]
```

alonre24 (commented):
Good catch and a clean fix
Let's add a short stress test in this opportunity

src/cursor.c
GuyAv46: I didn't want to rename the function, so I don't have to change it in all the files that use it. It's not _that_ bad in my opinion.

`Cursors_Purge` and `Cursor_Free` are **NOT** the same thing.
- `Cursors_Purge` is a **request** to delete a cursor, freeing it if it's idle.
- `Cursors_Free` is for releasing an owned cursor. It **always** freeing memory.

src/cursor.c
GuyAv46: Since in the two places we own the cursor, it is also not idle, most of this function is redundant, and we only need to call `Cursor_FreeInternal`.
I'll move the rest of the logic into purge

tests/cpptests/test_cpp_cursors.cpp
GuyAv46: I really don't think this is the right time or place

MeirShpilraien (approved):

PR #6019, opened 2025-04-26
run against vecsim main

meiravgri:
this branch purpose is to run CI checks against vecsim main banch to early identify any buils breaks


.gitmodules
```diff
@@ -14,6 +14,7 @@
 [submodule "deps/VectorSimilarity"]
 	path = deps/VectorSimilarity
 	url = https://github.com/RedisAI/VectorSimilarity.git
+	branch = main
```
JoanFM: should we perhaps have some marker of the VecSim related tests and have a separate set of tests dedicated to this?

Isn't there a big risk of blocking too many PR merges if there is a bug in main that one cannot fix here?

meiravgri: redisearch is not going to point to vecsim main
This is a temporary change to allow me to utilize the CI to run against different platforms

alonre24 (dismissed):

alonre24 (approved):

PR #6027, opened 2025-04-27
rust code reviewer bot

DvirDukhan:
# Rust Code Review Slash Command

## Overview
This PR adds a slash command `/rust_review` that triggers a Rust code review for the PR. This makes the Rust code review process more intentional and on-demand, rather than running automatically on every PR.

## Changes
- Created a new workflow `event-comment-rust-review.yml` that triggers when someone comments `/rust_review` on a PR
- Removed the automatic Rust code review from the pull request event handler
- Leverages the existing `task-rust-code-review.yml` workflow for the actual code review process
- Supports both `/rust_review` and `rust_review` formats for backward compatibility

## How to Use
1. Open a PR that contains Rust code changes
2. Comment `/rust_review` on the PR
3. The workflow will automatically run and post a detailed code review as a comment

## Benefits
- Makes the code review process more intentional
- Reduces CI load by only running when requested
- Provides flexibility to request a review at any time during the PR lifecycle
- Maintains the same high-quality code review using OpenAI's GPT-4

## Testing
The workflow has been tested and confirmed to work correctly when a comment containing `/rust_review` is added to a PR.


.github/workflows/event-pull_request.yml
```diff
@@ -46,12 +46,14 @@ jobs:
       env: ${{ vars.RUNS_ON || 'ubuntu-latest' }}
     secrets: inherit
 
+# Rust code review removed - now triggered by slash command /rust_review
```
benoitdion, line 49: nothing removed?

.github/workflows/task-rust-code-review.yml
```diff
@@ -0,0 +1,184 @@
...
+        with:
+          python-version: '3.10'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install requests openai
+
+      - name: Get changed Rust files
+        id: changed-files
```
benoitdion, line 35: there's https://github.com/marketplace/actions/changed-files too though this doesn't look too complicated

.github/workflows/task-rust-code-review.yml
```diff
@@ -0,0 +1,184 @@
...
+          echo "rust_files_changed=true" >> $GITHUB_OUTPUT
+          echo "files=$FILES" >> $GITHUB_OUTPUT
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+
+      - name: Generate code review
+        if: steps.changed-files.outputs.rust_files_changed == 'true'
+        id: generate-review
+        run: |
+          cat > review_script.py << 'EOF'
```
benoitdion, line 58: consider moving to a file instead of inline within the workflow

.github/workflows/task-rust-code-review.yml
```diff
@@ -0,0 +1,184 @@
...
+              print("No Rust files to review")
+              exit(0)
+
+          files = files_str.split()
+
+          # Function to get file content and diff
+          def get_file_diff(file_path):
+              try:
+                  # Get the diff for this file
+                  result = subprocess.run(
```
benoitdion, line 86: consider increasing the lines of context (`-U`)

DvirDukhan (commented):
/rust_review

.github/workflows/task-rust-code-review.yml
```diff
@@ -0,0 +1,73 @@
...
+          pip install requests openai
+
+      - name: Get changed Rust files
+        id: changed-files
+        run: |
+          PR_NUMBER=${{ inputs.pr_number }}
+          REPO=${{ inputs.repository }}
+
+          # Get the list of changed Rust files in the PR
+          FILES=$(gh pr view $PR_NUMBER --repo $REPO --json files --jq '.files[].path' | grep '\.rs$' || echo "")
```
LukeMathWalker: I don't think we should limit the context to Rust files—if the PR touches both Rust and C files, it's probably important to have both in the context.

LukeMathWalker: As in, trigger the job only if Rust files are changed, but then include C files too in the context.

.github/scripts/rust_code_review.py
```diff
@@ -0,0 +1,144 @@
...
+                capture_output=True,
+                text=True,
+                check=True
+            )
+        return result.stdout
+    except subprocess.CalledProcessError:
+        return f"Error getting diff for {file_path}"
+
+
+def generate_review_prompt(all_diffs):
```
LukeMathWalker: It'd be interesting (and valuable) to experiment with the prompt.
This is fairly generic, but we can probably refine it further over time. Would it be possible to allow the issue comment to specify a custom review prompt?

.github/workflows/event-comment-rust-review.yml
```diff
@@ -0,0 +1,31 @@
+name: Rust Code Review on Comment
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  check-comment:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.issue.pull_request && (contains(github.event.comment.body, '/rust_review') || contains(github.event.comment.body, 'rust_review')) }}
```
LukeMathWalker, line 15: Should we check who the comment author is?
E.g. ensure it has write access to the repository.

.github/scripts/rust_code_review.py
DvirDukhan: can you give an example?

LukeMathWalker: Let's consider the task of writing an FFI module. There are specific pitfalls to guard against—e.g. never assume that a pointer you've been handed over from C is non-null. We always want to check it at runtime or check it as a debug assertion.
This kind of "style" constraints aren't necessarily caught by our existing linters, so AI could be useful.

PR #6039, opened 2025-04-28
[8.0] Fix OOM behavior  

lerman25:
Change OOM behavior such that:
1. Won't fail queries after background scan fail
2.  Raise warning with RESP3

### DIFF
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -630,6 +630,8 @@ done_3:
     } else if (rc == RS_RESULT_ERROR) {
       // Non-fatal error
       RedisModule_Reply_SimpleString(reply, QueryError_GetUserError(req->qiter.err));
+    } else if (req->sctx->spec && req->sctx->spec->scan_failed_OOM) {
+    RedisModule_Reply_SimpleString(reply, QUERY_WINDEXING_FAILURE);
     } else if (req->qiter.err->reachedMaxPrefixExpansions) {
       RedisModule_Reply_SimpleString(reply, QUERY_WMAXPREFIXEXPANSIONS);
     }
@@ -872,8 +874,6 @@ static int buildRequest(RedisModuleCtx *ctx, RedisModuleString **argv, int argc,
     QueryError_SetWithUserDataFmt(status, QUERY_ENOINDEX, "No such index", " %s", indexname);
     goto done;
   }
-  // OOM should be checked before
-  RS_ASSERT(!(sctx->spec->scan_failed_OOM));
 
   CurrentThread_SetIndexSpec(sctx->spec->own_ref);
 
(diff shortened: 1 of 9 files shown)


src/module.c
```diff
@@ -3221,7 +3206,7 @@ int TagValsCommandHandler(RedisModuleCtx *ctx, RedisModuleString **argv, int arg
   }
   RS_AutoMemory(ctx);
 
-  VERIFY_ACL(ctx, argv[1], CHECK_OOM)
+  VERIFY_ACL(ctx, argv[1], IGNORE_OOM)
```
DvirDukhan, line 3209: as a general comment, since it is out of scope for the PR, I would have query validation function and call it to `VERIFY_ACL`, `handle_OOM` etc.. different PR
piggy back is abuse :)

src/query_error.h
```diff
@@ -71,6 +71,7 @@ extern "C" {
      an incomplete index.")                                                                      \
 
 #define QUERY_WMAXPREFIXEXPANSIONS "Max prefix expansions limit was reached"
+#define QUERY_WINDEXING_FAILURE "Index contains partial data due to OOM indexing failure"
```
DvirDukhan, line 74:
```suggestion
#define QUERY_WINDEXING_FAILURE "Index contains partial data due to indexing failure because not enough memory"
```

tests/pytests/test_index_oom.py
```diff
@@ -237,56 +252,11 @@ def test_change_config_during_bg_indexing(env):
   memory_ratio = used_memory / max_memory
   env.assertAlmostEqual(memory_ratio, 0.8, delta=0.1)
 
-@skip(cluster=True)
```
DvirDukhan, line 240: I would leave this test if we can make it run. if not comment it out and make it run in the future. I think it is important not to delete it

src/aggregate/aggregate_exec.c
```diff
@@ -630,6 +630,8 @@ static void sendChunk_Resp3(AREQ *req, RedisModule_Reply *reply, size_t limit,
     } else if (rc == RS_RESULT_ERROR) {
       // Non-fatal error
       RedisModule_Reply_SimpleString(reply, QueryError_GetUserError(req->qiter.err));
+    } else if (req->sctx->spec && req->sctx->spec->scan_failed_OOM) {
+    RedisModule_Reply_SimpleString(reply, QUERY_WINDEXING_FAILURE);
```
alonre24, line 634: Are these options mutually exclusive? I believe there can be a query in which we will have both `QUERY_WMAXPREFIXEXPANSIONS` and `QUERY_WMAXPREFIXEXPANSIONS`

tests/pytests/test_index_oom.py
```diff
@@ -11,6 +11,9 @@
 
 @skip(cluster=True)
 def test_stop_background_indexing_on_low_mem(env):
+    # Change the memory limit to 80% so it can be tested without colliding with redis memory limit
```
alonre24, line 14: What do you mean by "not colliding with redis memory limit"?
Also, do we have any tests with the new default (100%)? If not, I think we should have a basic one

tests/pytests/test_index_oom.py
alonre24: Unless we plan to change the behavior in the near future, I don't see a reason to leave this test commented...

DvirDukhan (approved):

PR #6102, opened 2025-05-07
Fix build for C23

zeenix:
C23 is now the default in gcc (version 15) and is being shipped by default in some Linux distributions (e.g Fedora 42).

#### Mark if applicable

### DIFF
--- a/src/util/circular_buffer.c
+++ b/src/util/circular_buffer.c
@@ -33,10 +33,10 @@ CircularBuffer CircularBuffer_New(size_t item_size, uint cap) {
   CircularBuffer cb = rm_calloc(1, sizeof(_CircularBuffer) + item_size * cap);
 
   cb->read       = cb->data;                      // initial read position
-  cb->write      = ATOMIC_VAR_INIT(0);            // write offset into data
+  cb->write      = 0;                             // write offset into data
   cb->item_cap   = cap;                           // buffer capacity
   cb->item_size  = item_size;                     // item size
-  cb->item_count = ATOMIC_VAR_INIT(0);            // no items in buffer
+  cb->item_count = 0;                             // no items in buffer
   cb->end_marker = cb->data + (item_size * cap);  // end of data marker
 
   return cb;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -282,7 +282,11 @@ if(COORD_TYPE STREQUAL "rlec")
     set_target_properties(redisearch PROPERTIES OUTPUT_NAME "module-enterprise")
 endif()
 
-set_target_properties(redisearch PROPERTIES LINKER_LANGUAGE CXX)
+set_target_properties(redisearch
+    PROPERTIES
+    LINKER_LANGUAGE CXX
+    C_STANDARD 17
+    C_STANDARD_REQUIRED ON)
 setup_shared_object_target(redisearch "")
 target_link_libraries(redisearch
         redisearch-geometry
--- a/tests/ctests/test_khtable.c
+++ b/tests/ctests/test_khtable.c
@@ -31,7 +31,7 @@ static uint32_t myHash(const KHTableEntry *e) {
   return ((const MyEntry *)e)->hash;
 }
 
-static KHTableEntry *myAlloc() {
+static KHTableEntry *myAlloc(void *ctx) {
   return calloc(1, sizeof(MyEntry));
 }
 


zeenix:
Just for the record, I ran [this workflow](https://github.com/RediSearch/RediSearch/actions/runs/14884141590/workflow) to further test this doesn't break things for older compilers and [it succeeds](https://github.com/RediSearch/RediSearch/actions/runs/14884141590).

src/util/circular_buffer.c
```diff
@@ -33,10 +33,10 @@ CircularBuffer CircularBuffer_New(size_t item_size, uint cap) {
   CircularBuffer cb = rm_calloc(1, sizeof(_CircularBuffer) + item_size * cap);
 
   cb->read       = cb->data;                      // initial read position
-  cb->write      = ATOMIC_VAR_INIT(0);            // write offset into data
+  cb->write      = 0;                             // write offset into data
```
kei-nan, line 36: I think the cmake should dictate we compile with the right c version if we want to remove this usage.

kei-nan:
The title is a bit misleading - we are talking about c standard and not the c++ standard.

src/util/circular_buffer.c
zeenix: I'm not sure I follow, sorry. By default it doesn't seem to choose anything and leave it to the system compilers.

zeenix:
> The title is a bit misleading - we are talking about c standard and not the c++ standard.

I thought that `ATOMIC_VAR_INIT` is purely a [C++ API](https://en.cppreference.com/w/cpp/atomic/ATOMIC_VAR_INIT)  even though we're using it from C but seems [you're right and this is because of C standard change](https://en.cppreference.com/w/c/atomic/ATOMIC_VAR_INIT). I'll update the title and commit log..

src/util/circular_buffer.c
zeenix: I think the change is harmless anyway:

> This macro was a part of early draft design for C11 atomic types. It is not needed in C11, and is deprecated in C17 and removed in C23. 

So I'd suggest we just merge it, especially given it passed the workflow using older Ubuntu and compilers. It makes the build work on my Fedora 42 so unblocks me. I think you'll agree that my times are better spent figuring out the bigger issue with memory in #5996. :)

raz-mon (approved):
@zeenix Let's run the full CI on the PR before merging just to see that all compilers of our supported versions "eat" this change fine

zeenix:
> @zeenix Let's run the full CI on the PR (before merging or in the merge queue) just to see that all compilers of our supported versions "eat" this change fine

For sure! :+1: How do you do that? :)

zeenix:
@raz-mon [all good](https://github.com/RediSearch/RediSearch/actions/runs/14903970812). :partying_face:

src/util/circular_buffer.c
kei-nan: Setting something to 0 is very implicit behaviour.
The fact it compiled doesn't necessarily means all is well.
If we are saying it is obsolete starting from a specific c standard version then we need to ensure we will compile only with that version or higher.

This probably requires a change in the CMakeLists.txt
See:
https://cmake.org/cmake/help/latest/prop_tgt/C_STANDARD.html

zeenix: > Setting something to 0 is very implicit behaviour.

From what I understand, the explicit API was unnecessary in C11 already, as the spec says:

> It is not needed in C11

Looking around, I saw [other projects also just remove the explicit calls](https://code.videolan.org/videolan/vlc/-/merge_requests/1658/diffs) and it seems to be [the general wisdom on the internets](https://stackoverflow.com/questions/48841767/how-does-c17-want-me-to-initialize-my-atomics). :)

> The fact it compiled doesn't necessarily means all is well.

For sure. I was more relying on the tests to catch issue. Do those workflows not run any tests?

> If we are saying it is obsolete starting from a specific c standard version then we need to ensure we will compile only with that version or higher.

We could totally do that, although given what I shared above, I think it's not really needed. Just let me know if you still think I should add the requirement bump to C17 and I'll do that.

zeenix: @kei-nan 

```C
#define ATOMIC_VAR_INIT(_VI) { _VI }
```

I rest my case. :wink:

kei-nan: According to chatgpt it was officially deprecated in c17:
image
I prefer to avoid as many risks as possible here.

If specifying the c standard(and specifying it is required) will formalize the project requirement from future environments then I think we should specify it.

zeenix: just to be clear, that's the actual definition in the header file. So my change is only removing an indirection.

zeenix: > Let us be explicit in our requirements from the compiler.

Done.

kei-nan (approved):

PR #6103, opened 2025-05-07
Yield to redis while indexing - [MOD-9220]

BenGoldberger:
This PR should fix the issue that a shard may not respond to "PING" while indexing **on load.**
this is done by yielding to redis after certain amount of indexing operations. the amount of operations between "yields" is configurable with `FT.CONFIG SET INDEXER_YIELD_EVERY_OPS <num operations>`.
It is possible to check the number of "yields" was called with `FT.DEBUG YIELDS_ON_LOAD_COUNTER`


src/debug_commands.c
```diff
@@ -1715,6 +1715,47 @@ DEBUG_COMMAND(getHideUserDataFromLogs) {
...
+    return RedisModule_ReplyWithError(ctx, NODEBUG_ERR);
+  }
+  
+  if (argc > 3) {
+    return RedisModule_WrongArity(ctx);
+  }
+  
+  // Check if we need to reset the counter
+  if (argc == 3) {
+    const char *subCmd = RedisModule_StringPtrLen(argv[2], NULL);
```
alonre24, line 1746: Is there a validation that `argc >= 2` ?

tests/pytests/common.py
```diff
@@ -84,7 +84,12 @@ def getConnectionByEnv(env):
 def waitForIndex(env, idx = 'idx'):
     waitForRdbSaveToFinish(env)
     while True:
-        res = env.cmd('ft.info', idx)
+        try:
+            res = env.cmd('ft.info', idx)
+        except redis_exceptions.ResponseError as e:
+            if 'redis is busy' in str(e).lower():
+                continue
+            raise e
```
GuyAv46, line 92: Consider handling the exception this way, so we don't skip the `sleep` call at the end of the loop
```suggestion
        except redis_exceptions.ResponseError as e:
            if 'redis is busy' not in str(e).lower():
                raise e
```

src/document.c
```diff
@@ -757,10 +761,21 @@ static PreprocessorFunc preprocessorMap[] = {
     [IXFLDPOS_GEOMETRY] = geometryPreprocessor,
     };
 
+
 int IndexerBulkAdd(RSAddDocumentCtx *cur, RedisSearchCtx *sctx,
                    const DocumentField *field, const FieldSpec *fs, FieldIndexerData *fdata,
-                   QueryError *status) {
+                   QueryError *status) {  
+  
+  static size_t opCounter = 0;
```
alonre24, line 769: Can't we expose `opCounter` for the debug command instead of having both `opCounter` and `g_yieldCallCounter`?

src/document.c
```diff
@@ -757,10 +761,21 @@ static PreprocessorFunc preprocessorMap[] = {
...
+  // Yield to Redis every RSGlobalConfig.indexerYieldEveryOps operations
+  if (RedisModule_Yield && ++opCounter >= RSGlobalConfig.indexerYieldEveryOps) {
+    opCounter = 0;
+    IncrementYieldCounter(); // Track that we called yield
+    RedisModule_Yield(sctx->redisCtx, REDISMODULE_YIELD_FLAG_CLIENTS, NULL);
```
GuyAv46, line 774: Wouldn't it be better if we wrap the call to `RedisModule_Yield` with the function that also tracks the number of calls? This is not the only place we call yield, and it also makes it easy to forget using it in other places.

tests/pytests/test_debug_commands.py
```diff
@@ -68,7 +69,7 @@ def testDebugHelp(self):
         self.env.expect(debug_cmd(), 'help').equal(help_list)
 
         arity_2_cmds = ['GIT_SHA', 'DUMP_PREFIX_TRIE', 'GC_WAIT_FOR_JOBS', 'DELETE_LOCAL_CURSORS', 'SHARD_CONNECTION_STATES',
-                        'PAUSE_TOPOLOGY_UPDATER', 'RESUME_TOPOLOGY_UPDATER', 'CLEAR_PENDING_TOPOLOGY', 'INFO', 'INDEXES', 'GET_HIDE_USER_DATA_FROM_LOGS']
+                        'PAUSE_TOPOLOGY_UPDATER', 'RESUME_TOPOLOGY_UPDATER', 'CLEAR_PENDING_TOPOLOGY', 'INFO', 'INDEXES', 'GET_HIDE_USER_DATA_FROM_LOGS', 'INDEXING_YIELD_COUNTER']
```
alonre24, line 72: Please add a sanity test for this new debug subcommand as we have for other debug commands

src/document.c
```diff
@@ -757,10 +761,21 @@ static PreprocessorFunc preprocessorMap[] = {
...
+  }
+
```
alonre24, line 776: We want to have the opCounter before every indexing operation **per field** (not per document)

tests/pytests/test_issues.py
```diff
@@ -1437,3 +1437,55 @@ def test_mod_9423(env:Env):
...
+    env.broadcast('SAVE')
+    env.broadcast('DEBUG RELOAD NOSAVE')
+    waitForIndex(env, 'idx')
+    env.expect(config_cmd(), 'GET', 'INDEXER_YIELD_EVERY_OPS').equal([['INDEXER_YIELD_EVERY_OPS', f'{yield_every_n_ops}']])
+    
+    final_count = env.cmd(debug_cmd(), 'INDEXING_YIELD_COUNTER')
+    expected_min_yields = num_docs // yield_every_n_ops
+    env.assertGreaterEqual(final_count, expected_min_yields, 
+                          message=f"Expected at least {expected_min_yields} yields, got {final_count}")
+
```
alonre24, line 1491: Nice test!
Let's please add another test where we have multiple fields (one of each type), and test with multiple indexes

alonre24 (commented):
Nice work!
Added some comments and suggestions.
The most important issue is that we want to count for every field indexing operation, and not just for every document.
Also, let's run the micro-benchmarks action to validate we have no regression in heavy indexing loads

src/debug_commands.c
BenGoldberger: I think it implied. you cant reach this function here without argc>=2

src/document.c
BenGoldberger: I agree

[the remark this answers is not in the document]
BenGoldberger: yes, from what I understand thats the wanted behavior, there is a possibility for a lot of indexing from notification.

alonre24 (commented):
Last comments and let's go :)

GuyAv46 (dismissed):

fcostaoliveira:
### Automated performance analysis summary

This comment was automatically generated given there is performance data available.

In summary:
- Detected a total of 1 stable tests between versions.
- Detected a total of 2 regressions bellow the regression water line 5.0.

You can check a comparison in detail via the [grafana link](https://benchmarksrediscom.grafana.net/d//3Ejv2wZnk/?var-branch=master&var-branch=yield-to-redis-while-indexing)

### Comparison between master and yield-to-redis-while-indexing.

Time Period from 30 days ago. (environment used: oss-standalone)

|                            Test Case                             |Baseline master (median obs. +- std.dev)|Comparison yield-to-redis-while-indexing (median obs. +- std.dev)|% change (higher-better)|           Note           |
|------------------------------------------------------------------|----------------------------------------|----------------------------------------------------------------:|------------------------|--------------------------|
|search-high-cardinality-negation-term-baseline                    | 29 +- 6.5% (7 datapoints)              |                                                               27|-6.8%                   |waterline=6.5%. REGRESSION|
|vecsim-arxiv-titles-384-angular-filters-m16-ef-128-fulltext-filter| 623 +- 3.3% (7 datapoints)             |                                                              640|2.8%                    |No Change                 |
|vecsim-arxiv-titles-384-angular-filters-m16-ef-128-tag-filter     | 77385 +- 4.8% (7 datapoints)           |                                                            71531|-7.6%                   |REGRESSION                |

GuyAv46 (dismissed):
Very nice!
Let’s improve the description of the PR - How often do we yield, what’s the granularity (yield after every field of every document indexed to every single index), when do we yield (both on load/server events, BG indexing, regular notification indexing), what new configurations were introduced and how to use them and so on

alonre24 (dismissed):

BenGoldberger:
Notice that I only count the yield while loading and not every yield.
I renamed the debug command.

tests/pytests/test_issues.py
```diff
@@ -1437,3 +1437,123 @@ def test_mod_9423(env:Env):
...
+    waitForIndex(env, 'idx')
+    waitForIndex(env, 'idx2')
+    waitForIndex(env, 'idx3')
+
+    # Check that yield was called
+    yields_count = env.cmd(debug_cmd(), 'YIELDS_ON_LOAD_COUNTER')
+    env.assertGreater(yields_count, 0, message="Yield should have been called at least once")
+    
+    # Verify the number of yields 
+    expected_min_yields = 7 * num_docs // yield_every_n_ops
```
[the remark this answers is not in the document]
BenGoldberger: there are 2 text fileds

alonre24 (approved):

src/indexer.c
```diff
@@ -106,6 +109,7 @@ static void writeCurEntries(RSAddDocumentCtx *aCtx, RedisSearchCtx *ctx) {
     if (invidx) {
       entry->docId = aCtx->doc->docId;
       RS_LOG_ASSERT(entry->docId, "docId should not be 0");
+      IndexerYieldWhileLoading(ctx->redisCtx);
```
GuyAv46, line 112: Do we really want to potentially yield on **every term in the document**? Even if we don't yield, we call `RedisModule_GetContextFlags(ctx)` every time. This can significantly affect the indexing performance. CC @alonre24

src/indexer.c
```diff
@@ -395,12 +398,16 @@ int IndexDocument(RSAddDocumentCtx *aCtx) {
  * This helps keep Redis responsive during long indexing operations.
  * @param ctx The Redis context
  */
-void IndexerYield(RedisModuleCtx *ctx) {
+void IndexerYieldWhileLoading(RedisModuleCtx *ctx) {
```
GuyAv46, line 401: Are we using it anywhere else? Let's make it `static` if not

src/indexer.c
```diff
@@ -395,12 +398,16 @@ int IndexDocument(RSAddDocumentCtx *aCtx) {
...
   static size_t opCounter = 0;
-  
-  // Yield to Redis every RSGlobalConfig.indexerYieldEveryOps operations
-  if (++opCounter >= RSGlobalConfig.indexerYieldEveryOps) {
+
+  // If server is loading, Yield to Redis every RSGlobalConfig.indexerYieldEveryOps operations
+  if (RedisModule_Yield
```
GuyAv46, line 405: Why do we check this? When is it not defined?

src/indexer.c
```diff
@@ -395,12 +398,16 @@ int IndexDocument(RSAddDocumentCtx *aCtx) {
...
+   && RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_LOADING
```
GuyAv46, line 406: Can we avoid testing this here? This function cannot be inlined, so it will significantly affect performance. We should test it once higher up and propagate the decision here

src/indexer.c
BenGoldberger: with redis version < 7

src/indexer.c
alonre24: I agree that it may add some overhead. Looking at the benchmark report, I don't see anything unusual on load flows. But on the other hand, it is not that reliable either... So I guess it does make sense to test it once and avoid jumping to redis code frame each time

src/indexer.c
GuyAv46: So this is only relevant to 2.6. Let's not clutter the code with unnecessary checks

src/indexer.c
GuyAv46: I don't think we have any benchmark scenario for indexing real-world text dataset, keep me honest

src/indexer.c
```diff
@@ -398,13 +400,11 @@ int IndexDocument(RSAddDocumentCtx *aCtx) {
...
-void IndexerYieldWhileLoading(RedisModuleCtx *ctx) {
+static void IndexerYieldWhileLoading(RedisModuleCtx *ctx, bool isLoading) {
   static size_t opCounter = 0;
 
   // If server is loading, Yield to Redis every RSGlobalConfig.indexerYieldEveryOps operations
-  if (RedisModule_Yield
-   && RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_LOADING
-   && ++opCounter >= RSGlobalConfig.indexerYieldEveryOpsWhileLoading) 
+  if (isLoading && ++opCounter >= RSGlobalConfig.indexerYieldEveryOpsWhileLoading) 
   {
```
GuyAv46, line 408:
```suggestion
  if (isLoading && ++opCounter >= RSGlobalConfig.indexerYieldEveryOpsWhileLoading) {
```

src/spec.c
```diff
@@ -3001,6 +3003,7 @@ static void Indexes_LoadingEvent(RedisModuleCtx *ctx, RedisModuleEvent eid, uint
       Indexes_ScanAndReindex();
     }
     workersThreadPool_OnEventEnd(true);
+    g_isLoading = false;
```
GuyAv46, line 3006: set to `false` on `REDISMODULE_SUBEVENT_LOADING_FAILED` as well

GuyAv46 (changes requested):

GuyAv46 (dismissed):
Great job!

alonre24 (dismissed):

alonre24 (approved):

PR #6108, opened 2025-05-08
MOD-9612: Do not check for early timeouts in non-strict timeout policy

raz-mon:
This PR skips the early timeout check we perform between building the iterator tree and building the pipeline.
In this way - we will still fail early upon using the strict timeout policy, but will continue to fail (stop collecting results, and return the collected ones) a bit later in case of non-strict timeout policy.

Related #6100


alonre24 (approved):

oshadmi (dismissed):
Great!

oshadmi (approved):

PR #6110, opened 2025-05-08
[2.8] Get redis 7.4 for packaging

nafraf:

1. Current: Redis last tag (8.0.0) is used for packaging, but it fails
2. Change: Use Redis 7.4.x is used for packaging
3. Outcome: Packaging works

1. MOD-...
2. #...

1. ...

#### Mark if applicable

### DIFF
--- a/.github/workflows/flow-build-artifacts.yml
+++ b/.github/workflows/flow-build-artifacts.yml
@@ -38,23 +38,22 @@ on:
         default: all
 
 jobs:
-  get-latest-redis-tag:
-    uses: ./.github/workflows/task-get-latest-tag.yml
-    with:
-      repo: redis/redis
-      prefix: '7.4'
-
+  
   setup:
     # Sets SHA and Validates the reference Input (branch or tag)
     runs-on: ubuntu-latest
-    needs: [get-latest-redis-tag]
     outputs:
       sha: ${{ steps.set-sha.outputs.sha }}
-      redis-ref: ${{ needs.get-latest-redis-tag.outputs.tag }}
+      redis-ref: ${{ steps.get-latest-redis-tag.outputs.tag }}
     steps:
       - uses: actions/checkout@v4
       - id: set-sha
         run: echo "sha=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
+      - id: get-latest-redis-tag
+        uses: ./.github/workflows/task-get-latest-tag.yml
+        with:
+          repo: redis/redis
+          prefix: '7.4'
       - name: Validate Reference
         shell: python
         run: |
--- a/.github/workflows/flow-build-artifacts.yml
+++ b/.github/workflows/flow-build-artifacts.yml
@@ -38,22 +38,23 @@ on:
         default: all
 
 jobs:
-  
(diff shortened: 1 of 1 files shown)


oshadmi (dismissed):

.github/workflows/flow-build-artifacts.yml
```diff
@@ -38,21 +38,23 @@ on:
         default: all
 
 jobs:
+  get-latest-redis-tag:
```
raz-mon, line 41: Can we put this inside the `steps` of the `setup` step?

oshadmi: `setup` has it in `needs` - so it is clear it needs to do `get-latest-redis-tag`
And it can can be reused.
If we "inline" it as a step - will not have a dedicated name

raz-mon: Yeah I saw the `needs`.
Since the `get-redis` variable is exposed from the `setup` step it seems cleaner to me.

GuyAv46: IMO, the current implementation is nice, but we can also fix the redis version to a specific patch version and remove the need to check for the latest release, as we don't really care about the redis version as long as it's compatible with the module version.
We don't use it for tests, only for loading the module and getting the `COMMANDS` info

oshadmi: Isn't it what we currently have? A specific job that will get the redis version compatible with the module version? The ability to also take the latest patch from the compatible version is not a problem, is it?

GuyAv46: I'm saying we can pin the redis version to 7.4.3 hard-coded, without any job

GuyAv46: It's not a problem, but it also not necessary

oshadmi: I am also fine with latest 7.4.x. Eventually, the packed module would also be loaded with and run against newer versions than 7.4.3.

GuyAv46: True, but in this workflow, it doesn't matter. We should **test** the module with the latest compatible version, but in the pack workflow we shouldn't care

raz-mon: Yeah we can do that as well I see no harm
Let's just keep the ref-setting logic in the `setup` step

GuyAv46: It shouldn't be possible to call another workflow from within a step, and we also shouldn't duplicate the logic of `get-latest-tag`.
We should do one of the two:
1. Use `get-latest-tag`, external to `setup`
2. Use some specific and acceptable redis tag

GuyAv46: Current state fails https://github.com/RediSearch/RediSearch/actions/runs/14957133919

raz-mon: That's what I was worried about..
Both options sound good to me 👍

raz-mon (approved):

oshadmi (approved):

PR #6117, opened 2025-05-12
[MOD-9688] Use `&[u8]` as key type rather than `&[c_char]` to get the expected ordering behaviour in iterators.

LukeMathWalker:

The label type in `TrieMap` has been changed from `&[c_char]` to `&[u8]` to match the ordering behaviour in the C version:

```C
static int __cmp_chars(const void *p1, const void *p2) {
  return *(char *)p1 - *(char *)p2;
} 
```

Using `&[u8]` allows us to simplify the code a bit.

#### Mark if applicable


src/redisearch_rs/trie_rs/tests/integration/iter.rs
```diff
@@ -40,67 +38,75 @@ macro_rules! assert_prefixed_iterators {
...
+        trie,
+        b"",
+        vec![
+            ("abcabc".as_bytes(), 1),
+            ("fußball straße".as_bytes(), 2),
+            ("grüßen".as_bytes(), 3),
+            ("бълга123".as_bytes(), 0),
+        ]
+    );
+}
```
LukeMathWalker, line 59: This test fails on `master`, but it's fixed by this PR.

src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -79,19 +79,20 @@ impl<Data> TrieMap<Data> {
             // After removing the child, we attempt to merge the child into the root.
             root.merge_child_if_possible();
             data
-        }
+        };
+        data
```
alonre24, line 83: Is change related to the replacement of c_char with u8?

src/redisearch_rs/trie_bencher/build.rs
```diff
@@ -18,7 +18,6 @@ fn main() {
     let lib_dir = {
         let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
         let target_arch = match target_arch.as_str() {
-            "aarch64" => "arm64v8",
```
alonre24, line 21: Relates to the build changes, right? not to the u8 change

src/redisearch_rs/trie_bencher/src/bencher.rs
```diff
@@ -266,11 +263,11 @@ fn load_c_benchmark<M: Measurement>(group: &mut BenchmarkGroup<'_, M>, contents:
 }
 
 pub fn rust_load_from_terms(keys: &[String]) -> RustTrieMap {
-    let words = keys.iter().map(|s| str2boxed_c_char(s)).collect::<Vec<_>>();
+    let words = keys.iter().map(|s| s.as_bytes()).collect::<Vec<_>>();
     rust_load(&words)
 }
 
-fn rust_load(words: &[Box<[c_char]>]) -> RustTrieMap {
+fn rust_load(words: &[&[u8]]) -> RustTrieMap {
```
alonre24, line 270: Why not `Box`ed anymore?

alonre24 (commented):
Nice cleanup!
Looks straight forward as far as I understand, added a few questions

src/redisearch_rs/trie_rs/tests/integration/iter.rs
alonre24: Why is that? How come the change in this PR enabled this test to pass?

src/redisearch_rs/trie_bencher/src/bencher.rs
LukeMathWalker: We no longer have to, we can work directly with the slice.

src/redisearch_rs/trie_bencher/build.rs
LukeMathWalker: Indeed. Without this, the microbenchmarks won't work in CI (nor locally, if using `build.sh` to do the build). Same for the change in the workflow.

src/redisearch_rs/trie_rs/src/trie.rs
LukeMathWalker: This one slipped through the cracks, removed.

src/redisearch_rs/trie_rs/tests/integration/iter.rs
LukeMathWalker: If we're using signed characters:
```
бълга123  => [-48, -79, -47, -118, -48, -69, -48, -77, -48, -80, 49, 50, 51]
abcabc    => [ 97,  98,  99,   97,  98,  99]
```
thus `бълга123` is smaller than `abcabc`.
If we encode them as unsigned characters, instead, `бълга123` becomes `[208, 177, 209, 138, 208, 187, 208, 179, 208, 176, 49, 50, 51]`, thus flipping the order and matching our expectations.

alonre24 (approved):

PR #6118, opened 2025-05-12
MOD-9671: Use wheel for gevent

Itzikvaknin:
We had `gevent<=24.11.1` in requirements.txt, but this version is not available as a prebuilt wheel on some platforms.

Since pip prioritizes the latest version satisfying the requirement, it attempts to install 24.11.1 even if it means building from source. For example, on Ubuntu Focal (aarch64), the latest wheel available is 22.10.2, but pip chooses to download and compile 24.11.1 from source. This leads to a failure due to a syntax error in gevent's .pyx source file — an issue outside our control.

Using `--only-binary=gevent` forces pip to install the latest available wheel, avoiding the build and resolving the problem.

Recommendation: In such cases, we should bound the version from below as well as from above, to ensure pip picks a compatible version with an available binary.

CI completions: https://github.com/RediSearch/RediSearch/actions/runs/14969913856/job/42048324188

### DIFF
--- a/.install/test_deps/common_installations.sh
+++ b/.install/test_deps/common_installations.sh
@@ -51,7 +51,7 @@ pip install -q --upgrade setuptools
 echo "pip version: $(pip --version)"
 echo "pip path: $(which pip)"
 
-pip install -q --only-binary=gevent -rtests/pytests/requirements.txt
+pip install -q -r tests/pytests/requirements.txt
 
 # List installed packages
 pip list
--- a/.install/test_deps/common_installations.sh
+++ b/.install/test_deps/common_installations.sh
@@ -51,7 +51,7 @@ pip install -q --upgrade setuptools
 echo "pip version: $(pip --version)"
 echo "pip path: $(which pip)"
 
-pip install -q -r tests/pytests/requirements.txt
+pip install -q --only-binary=gevent -rtests/pytests/requirements.txt
 
 # List installed packages
 pip list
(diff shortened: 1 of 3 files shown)


alonre24 (dismissed):

.install/test_deps/common_installations.sh
```diff
@@ -51,7 +51,7 @@ pip install -q --upgrade setuptools
 echo "pip version: $(pip --version)"
 echo "pip path: $(which pip)"
 
-pip install -q -r tests/pytests/requirements.txt
+pip install -q --only-binary=gevent -rtests/pytests/requirements.txt
```
GuyAv46, line 54: Can we add this requirement to the `requirements.txt` file itself, instead of when we call `pip`?

Itzikvaknin: Done 👍

GuyAv46 (dismissed):

tests/pytests/requirements.txt
```diff
@@ -1,3 +1,4 @@
+--only-binary=gevent
 packaging <= 24.2
 gevent <= 24.11.1
```
GuyAv46, line 3:
```suggestion
packaging <= 24.2
--only-binary=gevent
gevent <= 24.11.1
```
```suggestion
packaging <= 24.2
gevent <= 24.11.1; --only-binary=gevent
```

GuyAv46 (approved):

GuyAv46 (approved):

PR #6119, opened 2025-05-12
[MOD-9694] Trie prefixes iterator

LukeMathWalker:

Introduce a new iterator that returns all of the prefixes of a given term, `PrefixesIter`.
It's equivalent to `FindPrefixes` in `deps/triemap.h`.

The test module has been broken down into iterator-specific modules. New tests are under `tests/trie/iter/prefixes.rs`. All the other test files are just chunks of the old `tests/trie/iter.rs` file.

#### Mark if applicable


src/redisearch_rs/trie_bencher/src/bencher.rs
```diff
@@ -162,6 +162,42 @@ impl OperationBencher {
...
+    c: &mut BenchmarkGroup<'_, M>,
+    terms: &[String],
+    target: &str,
+) {
+    let target = target.into_cstring();
+    let view = target.as_view();
+    let map = c_load_from_terms(terms);
+    c.bench_function("C", |b| {
+        // We are leaking memory here, since we don't free the pointer to the result array.
+        b.iter_with_large_drop(|| map.find_prefixes(view).unwrap())
```
LukeMathWalker, line 199: This can be fixed by moving `util/arr.*` to a subfolder, adding a `CMakeLists.txt` file, build them as a standalone static library, binding them to Rust and then calling `array_free` here.

It feels like a lot of changes for benching code that will be deleted once we swap the impl, but open to making them if you think it's worthwhile. cc @raz-mon @joanFM

LukeMathWalker:
The C version was only returning values, ignoring the keys.
I've modified the Rust version to match its behaviour and now benches are where we like them to be:

```text
Wiki-1K|Find prefixes/Rust: [21.289 ns 21.407 ns 21.526 ns]
Wiki-1K|Find prefixes/C:    [32.131 ns 34.159 ns 37.129 ns]
```

src/redisearch_rs/trie_rs/tests/integration/iter/prefixes.rs
```diff
@@ -0,0 +1,49 @@
...
+    trie.insert(b"", b"".to_vec());
+    trie.insert(b"apple", b"apple".into());
+
+    assert_eq!(prefixes(&trie, b""), vec![b""]);
+}
+
+#[test]
+fn non_empty_prefixes() {
+    let mut trie = TrieMap::new();
+    trie.insert(b"apple", b"apple".to_vec());
```
JoanFM, line 29: for readability of the goal, I would maybe have a different value than the key? Something like "apple-value" or sthg like this?

JoanFM: It would make it more clear to see if `key`or `value` is extracted?

src/redisearch_rs/trie_bencher/src/bencher.rs
LukeMathWalker: Done in 9517ea36f

LukeMathWalker: Found a simpler way in e5cfc7a58 to avoid touching as many C files.

src/redisearch_rs/trie_rs/tests/integration/iter/prefixes.rs
JoanFM: Yes yes I know is clear, but IMO is more straightforward to clarify with the value. But I guess there is already a pattern on other tests so it's okey

src/redisearch_rs/trie_bencher/src/bencher.rs
```diff
@@ -162,6 +162,38 @@ impl OperationBencher {
         load_c_benchmark(&mut group, &self.keys);
         group.finish();
     }
+
+    /// Benchmark the find prefixes iterator.
+    ///
+    /// The benchmark group will be marked with the given label.
+    pub fn find_prefixes_group(&self, c: &mut Criterion, target: &str, label: &str) {
```
JoanFM, line 169: Just a comment: Not sure what is the strategy to do the benchmarking. I wonder if we should be using more than one key as a query to the `operations` to benchmark. 

Ignore my comment if this has been the case for other operations.

LukeMathWalker: We generally try to exercise all the different "modes" of the operation.
If you look at the `operations.rs` file, you can see how we classify inputs based on what part of the algorithm they "hit".

For prefixes, we could add inputs that exit earlier as an additional data point.
Alternatively, we can add a dedicated corpus with a higher opportunity for matches, but it may be overkill.

LukeMathWalker: I've added another bench in 58a4161d6 using the full Gutenberg dataset, which is ~10x the size.
We can see that the timing scales (roughly) with the depth of the traversal, which is what we expect.

src/util/arr/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+# Build the `arr` module as a standalone static library
+# This is a temporary requirement to allow us to benchmark the
+# Rust implementation of the triemap against the original C implementation.
+file(GLOB ARR_SOURCES "arr.c")
+add_library(arr STATIC ${ARR_SOURCES})
+target_include_directories(arr PRIVATE . ../..)
```
raz-mon, line 6: Why is this needed ?

src/redisearch_rs/trie_rs/tests/integration/iter/prefixes.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    assert_eq!(prefixes(&trie, b"apples"), vec![b"apple"]);
+
+    // If the target is stored as a term in the trie,
+    // it is returned as a valid prefix of itself.
+    assert_eq!(prefixes(&trie, b"ban"), vec![b"ban"]);
+
+    assert_eq!(
+        prefixes(&trie, b"bananas"),
+        vec!["ban".as_bytes(), b"banana"]
```
raz-mon, line 45: Any reason for using both `as_bytes` and `b""`?

raz-mon (commented):
LGTM 🔥
2 small questions

src/util/arr/CMakeLists.txt
LukeMathWalker: We need to invoke `array_new_sz` [here](https://github.com/RediSearch/RediSearch/pull/6119/files#diff-a01d2d8e17d9a560a800037f2d3d372a69cf4e11d6ec3292e0f4be69950b44beR53) and `array_free` [down here](https://github.com/RediSearch/RediSearch/pull/6119/files#diff-a01d2d8e17d9a560a800037f2d3d372a69cf4e11d6ec3292e0f4be69950b44beR86).
Since we're building `deps/triemap.c` as a static lib, it will refer to the symbols in `arr`, but it won't include their definition. We need to build the `arr` module as its own static lib to get those definitions in.

JoanFM (approved):

raz-mon (approved):

PR #6156, opened 2025-05-19
[MOD-9735] Track the number of unique keys in the trie.

LukeMathWalker:

Track (eagerly) the number of unique keys in the trie without requiring a full traversal.
It boils down to incrementing/decrementing a field whenever we perform a removal or an insertion.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/trie_rs/src/trie.rs
+++ b/src/redisearch_rs/trie_rs/src/trie.rs
@@ -19,11 +19,16 @@ use std::fmt;
 pub struct TrieMap<Data> {
     /// The root node of the trie.
     root: Option<Node<Data>>,
+    /// The number of unique keys stored in this map.
+    n_unique_keys: usize,
 }
 
 impl<Data> Default for TrieMap<Data> {
     fn default() -> Self {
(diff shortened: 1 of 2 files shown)


src/redisearch_rs/trie_rs/src/trie.rs
```diff
@@ -104,12 +113,23 @@ impl<Data> TrieMap<Data> {
     where
         F: FnOnce(Option<Data>) -> Data,
     {
+        let mut has_cardinality_increased = false;
+        let wrapped_f = |old_data: Option<Data>| {
+            if old_data.is_none() {
+                has_cardinality_increased = true;
```
kei-nan, line 119: what happens if no user data is stored in the container and only strings are stored?

LukeMathWalker: It would still work, because the `Data` generic parameter would be set to `()` and stored nodes would have value set to `Some(())` to distinguish them from prefix nodes with no payload.

kei-nan (approved):

PR #6159, opened 2025-05-19
[MOD-9734] Trie contains iterator

LukeMathWalker:
## Describe the changes in the pull request

Stacked on top of #6155.

An iterator that returns all entries that contain a given fragment. The traversal is optimized to skip the check whenever we know that ancestors verified the predicate.
The performance is on par with the C version.

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #6160, opened 2025-05-19
[MOD-9762] FFI bindings for Rust trie iterators

LukeMathWalker:

Stacked on top of #6159 .

The existing FFI bindings were limited to trie operations—create, insert, delete, find and free.
We now add bindings for our iterator implementations—plain, prefix, suffix, contains, wildcard and range.

The FFI code is used to automatically generate a C header file, which will replace `deps/triemap.h`.
The iterator interface has been simplified down a single next method, using enum dispatch internally.

We have some smoke tests against the FFI code, but it'll be primarily exercised by C/C++/Python tests that rely on the triemap for their functionality.

#### Mark if applicable


src/redisearch_rs/redis_mock/src/allocator.rs
```diff
@@ -34,7 +34,7 @@ const ALIGNMENT: usize = std::mem::align_of::<usize>();
 ///
 /// If size is zero, the behavior is implementation defined (null pointer may be returned,
 /// or some non-null pointer may be returned that shall not be dereferenced).
-pub(crate) extern "C" fn alloc_shim(size: usize) -> *mut c_void {
+pub extern "C" fn alloc_shim(size: usize) -> *mut c_void {
```
kei-nan, line 37: which crates need access to these functions?

LukeMathWalker: The test binary in `redisearch_rs` needs to bind the Redis symbols to these functions via the macro (https://github.com/RediSearch/RediSearch/pull/6160/files) which in turn requires these symbols to be reachable from outside the crate they are defined into.

kei-nan (dismissed):

kei-nan (approved):

PR #6163, opened 2025-05-19
[MOD-9396] Fix indexError leak

meiravgri:
This PR fixes a memory leak in the IndexError object related to the handling of the global `NA_rstr` (previously of type 'RedisModuleString*') that was used for "N/A" responses in IndexError objects.

The issue occurred because each new `IndexError` object incremented the reference count of the NA_rstr via `RedisModule_HoldString()` but never decreased it. While not causing an immediate memory leak, this created a significant problem when the reference count reached `MAX_INT - 1.` At this threshold, Redis would begin treating the string as stack-allocated and start duplicating it rather than incrementing its reference count. These duplicated strings were never freed, leading to a memory leak that grew over time.

## Fix
The fix replaces the use of a global `NA_rstr` for uninitialized keys with simply using `NULL` instead. This avoids the reference count issue entirely while maintaining the same functionality.

### DIFF
--- a/src/info/index_error.c
+++ b/src/info/index_error.c
@@ -77,11 +77,8 @@ void IndexError_RaiseBackgroundIndexFailureFlag(IndexError *error) {
     error->background_indexing_OOM_failure = true;
 }
 
-// TODO2: consider setting error->key to something that is not null to distinguish between
-// a destroyed object and a valid one. then use RS_ASSERT at the beginning of each function
-// of this API to check that error->key is NULL or a valid key.
 
-// Destroy the index error, the error.key cannot be used after this function is called.
+// Destroy the index error, the error object cannot be used after this function is called.
 void IndexError_Destroy(IndexError* error) {
     if (!error->key) return;
 
@@ -189,12 +186,11 @@ void IndexError_Combine(IndexError *error, const IndexError *other) {
         // Prefer the other error.
         // copy/add error count later.
         IndexError_ClearLastError(error);
-        if (error->key) {
-            RedisModule_FreeString(RSDummyContext, error->key);
-            // if we got here, it means that the other->key is not N/A
-            RS_ASSERT(other->key != NULL);
-            error->key = RedisModule_HoldString(RSDummyContext, other->key);
-        }
+        if (error->key) RedisModule_FreeString(RSDummyContext, error->key);
+        
+        // if we got here, it means that the other->key is not N/A
+        RS_ASSERT(other->key != NULL);
+        error->key = RedisModule_HoldString(RSDummyContext, other->key);
         error->last_error_without_user_data = rm_strdup(other->last_error_without_user_data);
         error->last_error_with_user_data = rm_strdup(other->last_error_with_user_data);
         error->last_error_time = other->last_error_time;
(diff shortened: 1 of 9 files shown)


src/info/index_error.c
```diff
@@ -17,6 +17,7 @@
 
 extern RedisModuleCtx *RSDummyContext;
 
+char* const DEADBEEF = (char*)0xDEADBEEF;
 char* const NA = "N/A";
 char* const OK = "OK";
```
meiravgri, line 22: is this in use?

src/info/index_error.c
```diff
@@ -17,6 +17,7 @@
 
 extern RedisModuleCtx *RSDummyContext;
 
+char* const DEADBEEF = (char*)0xDEADBEEF;
 char* const NA = "N/A";
```
meiravgri, line 21: can they be const char * const?

src/info/index_error.c
```diff
@@ -26,40 +27,41 @@ char* const IndexingErrorKey_String = "last indexing error key";
...
-    if (error->last_error_without_user_data != NA) {
+    RS_ASSERT(error->last_error_without_user_data != DEADBEEF && error->last_error_with_user_data != DEADBEEF);
+    if (error->last_error_without_user_data && error->last_error_without_user_data != NA) {
         rm_free(error->last_error_without_user_data);
     }
-    if (error->last_error_with_user_data != NA) {
+    if (error->last_error_with_user_data && error->last_error_with_user_data != NA) {
         rm_free(error->last_error_with_user_data);
     }
 }
```
meiravgri, line 54: can we check that error->key != NULL in the begining of the function and skip the NA check ?

src/info/index_error.c
```diff
@@ -71,38 +73,48 @@ void IndexError_AddError(IndexError *error, ConstErrorMessage withoutUserData, C
...
-    if (error.last_error_with_user_data != NA && error.last_error_with_user_data != NULL) {
-      rm_free(error.last_error_with_user_data);
-      error.last_error_with_user_data = NA;
-    }
-    if (error.key != NA_rstr) {
-        RedisModule_FreeString(RSDummyContext, error.key);
-        error.key = RedisModule_HoldString(RSDummyContext, NA_rstr);
-    }
+
+// Destroy the index error, the error object cannot be used after this function is called.
```
meiravgri, line 81: move to the header

src/info/index_error.c
```diff
@@ -71,38 +73,48 @@ void IndexError_AddError(IndexError *error, ConstErrorMessage withoutUserData, C
...
+void IndexError_Destroy(IndexError* error) {
+    if (!error->key) return;
```
meiravgri, line 83:
```suggestion
    if (!error->key) {
     RS_ASSERT(error->last_error_without_user_data == NULL) 
     RS_ASSERT(error->last_error_with_user_data == NULL)
     return
    }
```

src/info/index_error.c
```diff
@@ -71,38 +73,48 @@ void IndexError_AddError(IndexError *error, ConstErrorMessage withoutUserData, C
...
+
+    RedisModule_FreeString(RSDummyContext, error->key);
+    error->key = NULL;
+    RS_ASSERT(error->last_error_without_user_data != NULL && error->last_error_without_user_data != DEADBEEF);
```
meiravgri, line 87: DEADBEEF assert can be moved before `if (!error->key)`, no?

meiravgri (commented):
awesomeness

meiravgri:
We decided on a more conservative solution to avoid introducing big changes in patch versions:
https://github.com/RediSearch/RediSearch/pull/6193

PR #6167, opened 2025-05-20
MOD-9787: Add More Diagnostics When Active Queries Are Not Empty

kei-nan:
- Try to print active cursors and queries information on exit
- On startup, print the base address of the library
- Fix reference count bug for cursors

1. Active Queries Container

#### Related to
1. MOD-9774
2. MOD-9775 (?)

#### Mark if applicable

### DIFF
--- a/src/info/info_redis/types/blocked_queries.c
+++ b/src/info/info_redis/types/blocked_queries.c
@@ -9,6 +9,7 @@
 #include "info/info_redis/types/blocked_queries.h"
 #include "rmutil/rm_assert.h"
 #include "redismodule.h"
+#include "rmutil/rm_assert.h"
 
 BlockedQueries *BlockedQueries_Init() {
   BlockedQueries* blockedQueries = rm_calloc(1, sizeof(BlockedQueries));
@@ -17,14 +18,37 @@ BlockedQueries *BlockedQueries_Init() {
   return blockedQueries;
 }
 
+static size_t PrintActiveQueries(BlockedQueries *blockedQueries) {
+  size_t count = 0;
+  DLLIST_FOREACH(node, &blockedQueries->queries) {
+    BlockedQueryNode *at = DLLIST_ITEM(node, BlockedQueryNode, llnode);
+    IndexSpec *sp = StrongRef_Get(at->spec);
+    ++count; // increment regardless if sp is valid, the fact we have a valid node is problematic
+    if (!sp) {
+      continue;
+    }
+    RedisModule_Log(NULL, "warning", "Active query on index %s, started at %ld", IndexSpec_FormatName(sp, RSGlobalConfig.hideUserDataFromLog), at->start);
+  }
+  return count;
+}
+
+static size_t PrintActiveCursors(BlockedQueries *blockedQueries) {
+  size_t count = 0;
+  DLLIST_FOREACH(node, &blockedQueries->cursors) {
+    BlockedCursorNode *at = DLLIST_ITEM(node, BlockedCursorNode, llnode);
+    IndexSpec *sp = StrongRef_Get(at->spec);
+    ++count; // increment regardless if sp is valid, the fact we have a valid node is problematic
+    RedisModule_Log(NULL, "warning", "Active cursor %zu, on index %s, started at %ld", at->cursorId, sp ? IndexSpec_FormatName(sp, RSGlobalConfig.hideUserDataFromLog) : "n/a", at->start);
+  }
+  return count;
+}
+
 void BlockedQueries_Free(BlockedQueries *blockedQueries) {
-  // Assert that the lists are empty
-  RS_LOG_ASSERT((blockedQueries->queries.prev == blockedQueries->queries.next) &&
-                (blockedQueries->queries.next == &blockedQueries->queries),
-                "Active queries list is not empty");
-  RS_LOG_ASSERT((blockedQueries->cursors.prev == blockedQueries->cursors.next) &&
-              (blockedQueries->cursors.next == &blockedQueries->cursors),
-              "Active cursor list is not empty");
+  const size_t numQueries = PrintActiveQueries(blockedQueries);
+  const size_t numCursors = PrintActiveCursors(blockedQueries);
+  RS_LOG_ASSERT_FMT(numQueries == 0 && numCursors == 0, 
+    "There are %zu active queries and %zu active cursors. This is a bug. Please report it to https://github.com/redisearch/redisearch//issues", 
+    numQueries, numCursors);
   rm_free(blockedQueries);
 }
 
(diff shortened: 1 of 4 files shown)


src/module-init/module-init.c
```diff
@@ -119,6 +121,12 @@ int RediSearch_Init(RedisModuleCtx *ctx, int mode) {
   // Print version string!
   DO_LOG("notice", "RediSearch version %d.%d.%d (Git=%s)", REDISEARCH_VERSION_MAJOR,
          REDISEARCH_VERSION_MINOR, REDISEARCH_VERSION_PATCH, RS_GetExtraVersion());
+#ifdef __USE_GNU
+  // we print the base addess to allow easier translation of backtrace symbols
+  Dl_info info;
+  dladdr((void *)RediSearch_Init, &info);
+  DO_LOG("notice", "RediSearch base address: %p", &info.dli_fbase);
```
GuyAv46, line 128: Should be on debug level

src/info/info_redis/types/blocked_queries.c
```diff
@@ -49,10 +73,15 @@ BlockedCursorNode* BlockedQueries_AddCursor(BlockedQueries* blockedQueries, Weak
...
-  StrongRef_Release(blockedQueryNode->spec);
+  if (blockedQueryNode->spec.rm) {
+    StrongRef_Release(blockedQueryNode->spec);
+  }
   dllist_delete(&blockedQueryNode->llnode);
 }
 
 void BlockedQueries_RemoveCursor(BlockedCursorNode* blockedCursorNode) {
+  if (blockedCursorNode->spec.rm) {
+    StrongRef_Release(blockedCursorNode->spec);
```
GuyAv46, line 84: I think it’s better to have the bug fix on its own, and CP it to where we need it.
Also, we should add a mechanism on sanitizer shutdown to make sure we don’t leak indexes

src/info/info_redis/types/blocked_queries.c
```diff
@@ -49,10 +73,15 @@ BlockedCursorNode* BlockedQueries_AddCursor(BlockedQueries* blockedQueries, Weak
 }
 
 void BlockedQueries_RemoveQuery(BlockedQueryNode* blockedQueryNode) {
-  StrongRef_Release(blockedQueryNode->spec);
+  if (blockedQueryNode->spec.rm) {
+    StrongRef_Release(blockedQueryNode->spec);
+  }
```
GuyAv46, line 78: Can it ever be false?

src/info/info_redis/types/blocked_queries.c
```diff
@@ -17,14 +18,37 @@ BlockedQueries *BlockedQueries_Init() {
...
+  return count;
+}
+
+static size_t PrintActiveCursors(BlockedQueries *blockedQueries) {
+  size_t count = 0;
+  DLLIST_FOREACH(node, &blockedQueries->cursors) {
+    BlockedCursorNode *at = DLLIST_ITEM(node, BlockedCursorNode, llnode);
+    IndexSpec *sp = StrongRef_Get(at->spec);
+    ++count; // increment regardless if sp is valid, the fact we have a valid node is problematic
+    RedisModule_Log(NULL, "warning", "Active cursor %zu, on index %s, started at %ld", at->cursorId, sp ? IndexSpec_FormatName(sp, RSGlobalConfig.hideUserDataFromLog) : "n/a", at->start);
```
GuyAv46, line 41: Can we print the full command? And the query itself?

src/info/info_redis/types/blocked_queries.c
```diff
@@ -17,14 +18,37 @@ BlockedQueries *BlockedQueries_Init() {
...
+static size_t PrintActiveQueries(BlockedQueries *blockedQueries) {
+  size_t count = 0;
+  DLLIST_FOREACH(node, &blockedQueries->queries) {
+    BlockedQueryNode *at = DLLIST_ITEM(node, BlockedQueryNode, llnode);
+    IndexSpec *sp = StrongRef_Get(at->spec);
+    ++count; // increment regardless if sp is valid, the fact we have a valid node is problematic
+    if (!sp) {
+      continue;
+    }
+    RedisModule_Log(NULL, "warning", "Active query on index %s, started at %ld", IndexSpec_FormatName(sp, RSGlobalConfig.hideUserDataFromLog), at->start);
```
GuyAv46, line 30: Can we print the full command?

src/info/info_redis/types/blocked_queries.c
```diff
@@ -17,14 +18,37 @@ BlockedQueries *BlockedQueries_Init() {
...
-  RS_LOG_ASSERT((blockedQueries->queries.prev == blockedQueries->queries.next) &&
-                (blockedQueries->queries.next == &blockedQueries->queries),
-                "Active queries list is not empty");
-  RS_LOG_ASSERT((blockedQueries->cursors.prev == blockedQueries->cursors.next) &&
-              (blockedQueries->cursors.next == &blockedQueries->cursors),
-              "Active cursor list is not empty");
+  const size_t numQueries = PrintActiveQueries(blockedQueries);
+  const size_t numCursors = PrintActiveCursors(blockedQueries);
+  RS_LOG_ASSERT_FMT(numQueries == 0 && numCursors == 0, 
+    "There are %zu active queries and %zu active cursors. This is a bug. Please report it to https://github.com/redisearch/redisearch//issues",
```
GuyAv46, line 50:
```suggestion
    "There are %zu active queries and %zu active cursors. This is a bug. Please report it to https://github.com/RediSearch/RediSearch/issues", 
```

src/info/info_redis/types/blocked_queries.c
kei-nan: right now no, related to the fact the ownership of the query object is passed to the query thread.
We can clone the query string but we can't obfuscate it so it isn't something we can do in production at this time.

src/info/info_redis/types/blocked_queries.c
kei-nan: same problem as I mentioned before.
The query ast is passed to the query thread, it isn't ref counted.
We don't want to print a non obfuscated string.

src/info/info_redis/types/blocked_queries.c
kei-nan: I guess not, since it is being called from the main thread which managed the spec lifetime.

src/module-init/module-init.c
kei-nan: do we set debug logging level on our CI?

src/info/info_redis/types/blocked_queries.c
GuyAv46: It's more of a question about whether we expect to create a node with no reference

src/module-init/module-init.c
GuyAv46: yes

src/info/info_redis/types/blocked_queries.c
GuyAv46: Even not temporarily for debugging purposes on master only? Sounds a bit extreme

src/info/info_redis/types/blocked_queries.c
GuyAv46: This code is not for production. Mayebe add an `#ifdef` to include it on CI and development builds

src/info/info_redis/types/blocked_queries.c
kei-nan: okay, will add the explain string for the query ast.

src/info/info_redis/types/blocked_queries.c
kei-nan: okay, will add the explain string for the query ast.
will output it only if hide user data from logs is off.

PR #6175, opened 2025-05-20
MOD-9768: Add Rust abstractions for Buffer*

zeenix:

This PR provides the Rust bindings to the `Buffer` API that is currently implemented in C. This will be useful for oxidization of the modules that rely on this API (e.g Varint). Except for the memory growth in `BufferWriter`, all other functionality is implemented in Rust.

This is the first part of the split of #5996.

#### Mark if applicable

### DIFF
--- a/.github/workflows/task-test.yml
+++ b/.github/workflows/task-test.yml
@@ -50,7 +50,6 @@ env:
   # flags, and therefore reuse the same intermediate build artifacts.
   RUSTFLAGS: "-D warnings"
   RUST_BACKTRACE: "full"
-  CARGO_PROFILE_OPTIMISED_TEST_BUILD_OVERRIDE_DEBUG: "true"
 
 jobs:
   common-flow:
--- a/.github/workflows/task-test.yml
+++ b/.github/workflows/task-test.yml
@@ -50,6 +50,7 @@ env:
   # flags, and therefore reuse the same intermediate build artifacts.
   RUSTFLAGS: "-D warnings"
   RUST_BACKTRACE: "full"
+  CARGO_PROFILE_OPTIMISED_TEST_BUILD_OVERRIDE_DEBUG: "true"
 
 jobs:
   common-flow:
--- a/src/redisearch_rs/buffer/Cargo.toml
+++ b/src/redisearch_rs/buffer/Cargo.toml
@@ -11,15 +11,3 @@ workspace = true
 crate-type = ["staticlib", "rlib"]
 
 [dependencies]
-
-[target.'cfg(all(target_env="musl", target_os="linux"))'.dependencies.redis-module]
-# Statically link to the libclang on aarch64-unknown-linux-musl,
-# necessary on Alpine.
-# See https://github.com/rust-lang/rust-bindgen/issues/2360
-features = ["bindgen-static", "min-redis-compatibility-version-6-0"]
-workspace = true
-default-features = false
-
-[target.'cfg(not(all(target_env="musl", target_os="linux")))'.dependencies.redis-module]
-workspace = true
-default-features = true
--- a/src/redisearch_rs/buffer/Cargo.toml
+++ b/src/redisearch_rs/buffer/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "buffer"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[lints]
+workspace = true
+
+[lib]
+crate-type = ["staticlib", "rlib"]
+
+[dependencies]
+
+[target.'cfg(all(target_env="musl", target_os="linux"))'.dependencies.redis-module]
+# Statically link to the libclang on aarch64-unknown-linux-musl,
+# necessary on Alpine.
+# See https://github.com/rust-lang/rust-bindgen/issues/2360
+features = ["bindgen-static", "min-redis-compatibility-version-6-0"]
+workspace = true
+default-features = false
+
+[target.'cfg(not(all(target_env="musl", target_os="linux")))'.dependencies.redis-module]
+workspace = true
+default-features = true
--- a/src/redisearch_rs/buffer/Cargo.toml
+++ b/src/redisearch_rs/buffer/Cargo.toml
@@ -8,7 +8,7 @@ license.workspace = true
 workspace = true
 
 [lib]
-crate-type = ["staticlib", "rlib"]
+crate-type = ["rlib"]
 
 [dependencies]
 ffi.workspace = true
--- a/src/redisearch_rs/buffer/Cargo.toml
+++ b/src/redisearch_rs/buffer/Cargo.toml
@@ -11,3 +11,4 @@ workspace = true
 crate-type = ["staticlib", "rlib"]
 
 [dependencies]
+ffi.workspace = true
--- a/src/redisearch_rs/buffer/src/lib.rs
+++ b/src/redisearch_rs/buffer/src/lib.rs
@@ -7,25 +7,45 @@
  * GNU Affero General Public License v3 (AGPLv3).
 */
 
+//! The `Buffer` API Rust wrappers.
+//!
+//! This crate provides Rust wrappers for the `Buffer`, `BufferReader` and `BufferWriter` structs
+//! from `buffer.h`.
+
 use std::{
     ptr::{NonNull, copy_nonoverlapping},
     slice,
 };
 
 /// Redefines the `Buffer` struct from `buffer.h`
-///
-/// Allocated by C, we never want to free it.
 #[repr(C)]
 pub struct Buffer {
+    /// A pointer to the underlying data buffer. This is typically allocated by C and hence should
+    /// not be freed by Rust code.
     pub data: NonNull<u8>,
+    /// The capacity of the buffer (i-e allocated size)
     pub capacity: usize,
+    /// The length of the buffer (i-e the total used memory from allocated memory)
     pub len: usize,
 }
 
 /// Redefines the `BufferReader` struct from `buffer.h`
+///
+/// Provides read functionality over a Buffer with position tracking.
+///
+/// # Safety Invariants
+///
+/// * `buf` must point to a valid, properly initialized `Buffer`.
+/// * `pos` must not exceed the length of the Buffer.
+/// * The `BufferReader` does not own the `Buffer`, only borrows it.
+///
+/// It's best to use this through the `std::io::Read` implementation.
 #[repr(C)]
 pub struct BufferReader {
+    /// A pointer to a [`Buffer`]. If the `BufferReader` is coming from the C side, we must not
+    /// free or move it.
     pub buf: NonNull<Buffer>,
+    /// The read position in the `buf` pointer.
     pub pos: usize,
 }
 
@@ -58,6 +78,13 @@ impl std::io::Read for BufferReader {
 impl Buffer {
     /// Creates a new `Buffer` with the given pointer, length, and capacity.
     ///
+    /// # Safety
+    ///
+    /// This function is safe to call, but creates a Buffer that maintains unsafe invariants:
+    /// * `data` must be a valid pointer to a memory region of at least `capacity` bytes.
+    /// * The memory region must remain valid for the lifetime of the Buffer.
+    /// * The first `len` bytes of the memory region must be initialized.
+    ///
     /// # Panics
     ///
     /// Panics if `len` is greater than `capacity`.
@@ -70,7 +97,16 @@ impl Buffer {
         }
     }
 
-    /// The internal buffer as a slice.
+    /// Returns the initialized portion of the buffer as a slice.
+    ///
+    /// # Safety
+    ///
+    /// This method assumes the safety invariants of `Buffer` are upheld:
+    /// * `data` points to a valid memory region of at least `capacity` bytes.
+    /// * The first `len` bytes of that memory are initialized.
+    ///
+    /// If these invariants are violated, this method may return an invalid slice,
+    /// leading to undefined behavior.
     pub fn as_slice(&self) -> &[u8] {
         // Safety: `self.ptr` is a valid pointer, if C side gave us one.
         let data = unsafe { self.data.as_ref() };
@@ -78,7 +114,17 @@ impl Buffer {
         unsafe { slice::from_raw_parts(data, self.len) }
     }
 
-    /// The internal buffer as a mutable slice.
+    /// Returns the initialized portion of the buffer as a mutable slice.
+    ///
+    /// # Safety
+    ///
+    /// This method assumes the safety invariants of Buffer are upheld:
+    /// * `data` points to a valid memory region of at least `capacity` bytes.
+    /// * The first `len` bytes of that memory are initialized.
+    /// * The caller has exclusive access to the buffer (no aliasing).
+    ///
+    /// If these invariants are violated, this method may return an invalid mutable slice,
+    /// leading to undefined behavior.
     pub fn as_mut_slice(&mut self) -> &mut [u8] {
         // Safety: `self.ptr` is a valid pointer, if C side gave us one.
         let data = unsafe { self.data.as_mut() };
@@ -86,27 +132,31 @@ impl Buffer {
         unsafe { slice::from_raw_parts_mut(data, self.len) }
     }
 
-    /// The length of the buffer.
+    /// Returns the length of the buffer (number of initialized bytes).
     pub fn len(&self) -> usize {
         self.len
     }
 
-    /// If the buffer is empty.
+    /// Returns true if the buffer is empty (length is zero).
     pub fn is_empty(&self) -> bool {
         self.len == 0
     }
 
-    /// The capacity of the buffer.
+    /// Returns the total capacity of the buffer (maximum number of bytes it can hold).
     pub fn capacity(&self) -> usize {
         self.capacity
     }
 
-    /// The remaining capacity of the buffer.
+    /// Returns the remaining capacity of the buffer (capacity - length).
     pub fn remaining_capacity(&self) -> usize {
         self.capacity - self.len
     }
 
     /// Advance the buffer by `n` bytes.
+    ///
+    /// This increases the buffer's length by `n` bytes, effectively marking
+    /// more of the buffer as "initialized" without actually writing any data.
+    /// Typically used after directly writing to the buffer's memory.
     ///
     /// # Panics
     ///
@@ -118,6 +168,37 @@ impl Buffer {
 }
 
 /// Redefines the `BufferWriter` struct from `buffer.h`
+///
+/// A writer for appending data to a `Buffer`. Provides a mechanism to write bytes to a `Buffer`
+/// while tracking the current write position with a cursor.
+///
+/// # Safety Invariants
+///
+/// * `buf` must point to a valid, properly initialized `Buffer`.
+/// * `cursor` must point to a valid position within the Buffer's allocated memory.
+/// * The `BufferWriter` does not own the `Buffer`, only borrows it.
+/// * If the buffer is grown (implicitly) as part of a write operation, all previously obtained
+///   pointers into the buffer are invalidated.
+///
+/// # Usage
+///
+/// It's best to use this through the `std::io::Write` implementation, which handles
+/// buffer growth and cursor position management automatically.
+///
+/// ```no_run
+/// # use std::io::Write;
+/// # use buffer::{Buffer, BufferWriter};
+/// # use std::ptr::NonNull;
+/// # let buffer_ptr: NonNull<Buffer> = todo!();
+/// # let buffer = unsafe { buffer_ptr.as_mut() };
+/// let mut writer = BufferWriter {
+///     buf: buffer_ptr,
+///     cursor: buffer.data,  // Start writing at the beginning
+/// };
+///
+/// // Write data to the buffer
+/// writer.write(b"Hello, world!").unwrap();
+/// ```
 #[repr(C)]
 #[derive(Debug, Clone, Copy)]
 pub struct BufferWriter {
@@ -162,7 +243,27 @@ impl std::io::Write for BufferWriter {
 #[cfg(not(test))]
 unsafe extern "C" {
     /// Ensure that at least extraLen new bytes can be added to the buffer.
-    /// Returns the number of bytes added, or 0 if the buffer is already large enough.
+    ///
+    /// This function grows the buffer's capacity if needed to accommodate at least `extraLen`
+    /// additional bytes beyond the current length. If the buffer already has enough capacity, no
+    /// action is taken.
+    ///
+    /// # Safety
+    ///
+    /// This function must be called with a valid buffer pointer. It may reallocate the buffer's
+    /// memory, which invalidates any existing pointers into the buffer. After calling this
+    /// function, any previously obtained pointers or references to the buffer's data may be
+    /// invalid.
+    ///
+    /// The caller must ensure:
+    /// * `b` points to a valid `Buffer` struct.
+    /// * The `Buffer`'s data pointer is valid and was allocated by the C side.
+    /// * After calling this function, update any pointers that may have been invalidated.
+    ///
+    /// # Returns
+    ///
+    /// Returns the number of bytes added to the capacity, or 0 if the buffer
+    /// already had enough capacity.
     fn Buffer_Grow(b: NonNull<Buffer>, extraLen: usize) -> usize;
 }
 
--- a/src/redisearch_rs/buffer/src/lib.rs
+++ b/src/redisearch_rs/buffer/src/lib.rs
@@ -209,7 +209,7 @@ impl BufferWriter {
 impl std::io::Write for BufferWriter {
     fn write(&mut self, bytes: &[u8]) -> std::io::Result<usize> {
         // Safety: We assume `buf` is a valid pointer to a properly initialized Buffer.
-        let buffer = unsafe { &mut *(self.0.buf as *mut Buffer) };
+        let buffer = unsafe { self.buffer_mut() };
 
         // Check if we need to grow the buffer to accommodate the new data
         debug_assert!(
@@ -226,7 +226,7 @@ impl std::io::Write for BufferWriter {
             unsafe { Buffer_Grow(self.0.buf, bytes.len()) };
 
             // Safety: We assume `buf` is a valid pointer to a properly initialized Buffer.
-            let buffer = unsafe { &mut *(self.0.buf as *mut Buffer) };
+            let buffer = unsafe { self.buffer_mut() };
 
             // The buffer has relocated, so we need to update our cursor.
             // Safety: After growth, `buffer.len()` points just past the end of the valid data,
@@ -246,7 +246,7 @@ impl std::io::Write for BufferWriter {
         unsafe { copy_nonoverlapping(src, dest, bytes.len()) };
 
         // Safety: We assume `buf` is a valid pointer to a properly initialized Buffer.
-        let buffer = unsafe { &mut *(self.0.buf as *mut Buffer) };
+        let buffer = unsafe { self.buffer_mut() };
         // Update the buffer length.
         // Safety: We've ensured that the buffer has enough capacity.
         unsafe { buffer.advance(bytes.len()) };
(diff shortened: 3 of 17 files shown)


src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+        // Safety: We assume `buf` is a valid pointer to a properly initialized Buffer.
+        let buffer = unsafe { self.buf.as_mut() };
+
+        // Check if there's enough data to read
+        if self.pos + dest_buf.len() > buffer.len {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::UnexpectedEof,
+                "BufferReader: not enough data",
+            ));
+        }
```
LukeMathWalker, line 63: I think this is overly strict—`read`'s contract states that it can read _up to_ the capacity of the destination buffer, but there is no expectation that the buffer will be filled, hence why we return the number of bytes that have actually been read.

LukeMathWalker: We should compute the minimum between `buffer.len - self.pos` and `dest_buf.len` and then read that many bytes into the destination buffer. No error scenario.

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+    /// Advance the buffer by `n` bytes.
+    ///
+    /// This increases the buffer's length by `n` bytes, effectively marking
+    /// more of the buffer as "initialized" without actually writing any data.
+    /// Typically used after directly writing to the buffer's memory.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `n` exceeds the remaining capacity of the buffer.
+    pub fn advance(&mut self, n: usize) {
```
LukeMathWalker, line 177: This should be unsafe, in the same way `Vec::set_len` is unsafe (https://doc.rust-lang.org/std/vec/struct.Vec.html#method.set_len). The caller must guarantee additional invariants that are not tracked by the type system—i.e. that the next `n` elements are initialized.

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+    ///
+    /// This function is safe to call, but creates a Buffer that maintains unsafe invariants:
+    /// * `data` must be a valid pointer to a memory region of at least `capacity` bytes.
+    /// * The memory region must remain valid for the lifetime of the Buffer.
+    /// * The first `len` bytes of the memory region must be initialized.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `len` is greater than `capacity`.
+    pub fn new(data: NonNull<u8>, len: usize, capacity: usize) -> Self {
```
LukeMathWalker, line 100: This should be unsafe, for the same reasons `Vec::from_raw_parts` is unsafe.

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+        let buffer = unsafe { self.buf.as_mut() };
+
+        // Check if we need to grow the buffer to accommodate the new data
+        if buffer.len + bytes.len() > buffer.capacity {
+            // Safety: `Buffer_Grow` is a C function that increases the buffer's capacity. It
+            // expects a valid buffer pointer and returns the number of bytes added to capacity. If
+            // it returns non-zero, the buffer was successfully grown.
+            //
+            // Note that this may invalidate previously held pointers into the buffer.
+            if unsafe { Buffer_Grow(self.buf, bytes.len()) } != 0 {
```
LukeMathWalker, line 234: Shouldn't we handle the error case here?

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+/// Redefines the `Buffer` struct from `buffer.h`
+#[repr(C)]
+pub struct Buffer {
+    /// A pointer to the underlying data buffer. This is typically allocated by C and hence should
+    /// not be freed by Rust code.
+    pub data: NonNull<u8>,
+    /// The capacity of the buffer (i-e allocated size)
+    pub capacity: usize,
+    /// The length of the buffer (i-e the total used memory from allocated memory)
+    pub len: usize,
```
LukeMathWalker, line 29: Should these fields be public? 
We have read-only accessors for capacity and length.
We also have `BufferWriter` to write into a buffer. 
There should be no need for public access outside this crate.

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+//! from `buffer.h`.
+
+use std::{
+    ptr::{NonNull, copy_nonoverlapping},
+    slice,
+};
+
+/// Redefines the `Buffer` struct from `buffer.h`
+#[repr(C)]
+pub struct Buffer {
```
LukeMathWalker, line 22: I'm assuming that we want the user to guarantee that all elements up to `len` are initialized whenever we get a `Buffer` on our hands.

src/redisearch_rs/buffer/src/tests.rs
```diff
@@ -0,0 +1,281 @@
...
+        let test_data = b"MoreData";
+        assert_eq!(writer.write(test_data).unwrap(), 8);
+
+        // Buffer should have grown
+        assert!(buffer.capacity > initial_capacity);
+        assert_eq!(buffer.len, 18);
+
+        // Check the written data
+        let expected = b"HelloWorldMoreData";
+        assert_eq!(&buffer.as_slice()[..18], expected);
```
LukeMathWalker, line 204: We should verify that the cursor writer has been updated to be a pointer inside the new allocation.

src/redisearch_rs/buffer/src/tests.rs
```diff
@@ -0,0 +1,281 @@
...
+            cursor: NonNull::new(buffer.data.as_ptr().add(initial_capacity)).unwrap(),
+        };
+
+        // Write 1 more byte - should trigger grow
+        let test_data = b"!";
+        assert_eq!(writer.write(test_data).unwrap(), 1);
+
+        // Buffer should have grown
+        assert!(buffer.capacity > initial_capacity);
+        assert_eq!(buffer.len, initial_capacity + 1);
```
LukeMathWalker, line 236: We should verify that the cursor writer has been updated to be a pointer inside the new allocation.

src/redisearch_rs/buffer/src/lib.rs
zeenix: Oh, actually `0` doesn't mean an error. The only reason `Buffer_Grow` can fail is OOM I think and I don't think we handle that as an error anywhere (at least not in C). The docs are misleading. I'll fix them..

zeenix: Looked again at the varint C code and remembered that it's not using `Grow` directly but through `Buffer_Reserve` (an inline function so we can't use directly in Rust) that indeed does return `0` as an error if `buffer.len + bytes.len() > buffer.capacity`. However, `Grow` function doesn't return 0. I'll turn this into a debug assertion.

src/redisearch_rs/buffer/src/lib.rs
zeenix: sure. I assume you mean that we we should add a `Safety` section in the doc comment with this and other invariants listed?

LukeMathWalker: Yup!

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+    slice,
+};
+
+/// Redefines the `Buffer` struct from `buffer.h`
+///
+/// # Safety
+///
+/// * `data` must point to a valid, properly initialized `Buffer`.
+/// * `capacity` must not exceed the length of the Buffer.
+/// * `len` must not exceed the length of the Buffer.
```
chesedo, line 27:
```suggestion
/// * `data` must point to a valid, properly initialized `Buffer`.
/// * `capacity` must not exceed the capacity of the Buffer.
/// * `len` must not exceed the length or capacity of the Buffer.
```

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+
+        // Check if we need to grow the buffer to accommodate the new data
+        if buffer.len + bytes.len() > buffer.capacity {
+            // Safety: `Buffer_Grow` is a C function that increases the buffer's capacity. It
+            // expects a valid buffer pointer and returns the number of bytes added to capacity.
+            // This number can be 0 if the buffer is already large enough but since it reallocates
+            // unconditionally, we need to update our cursor in either case.
+            //
+            // Note that this may invalidate previously held pointers into the buffer.
+            unsafe { Buffer_Grow(self.buf, bytes.len()) };
```
chesedo, line 247: Is it possible that this `Buffer_Grow` might move the `self.buf` to a new location in memory (when there are not enough `bytes.len()` capacity after the buffer for it to grow into)?

And if it does move, then where does the `buffer` variable from line 237 point to?

My reason for asking is because I think your next use of `buffer` on line 251 could point to an old memory location, so the `cursor` could point to invalid memory.

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+/// let mut writer = unsafe { BufferWriter::for_buffer(buffer_ptr) };
+///
+/// // Write data to the buffer
+/// writer.write(b"Hello, world!").unwrap();
+/// ```
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
+pub struct BufferWriter {
+    pub buf: NonNull<Buffer>,
+    pub cursor: NonNull<u8>,
```
chesedo, line 216: Question for educational purposes :smile: : why does the writer need a cursor while the reader only needed to keep track of the pos?

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+        Self {
+            data,
+            len,
+            capacity,
+        }
+    }
+
+    /// Returns the initialized portion of the buffer as a slice.
+    ///
+    /// # Safety
```
JonasKruckenberg, line 111: dunno if restating the safety requirements here is necessary as its not required of callers of `as_slice` to uphold these but of callers of `new` (and this method is marked safe too)

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+    ///
+    /// This method assumes the safety invariants of Buffer are upheld:
+    /// * `data` points to a valid memory region of at least `capacity` bytes.
+    /// * The first `len` bytes of that memory are initialized.
+    /// * The caller has exclusive access to the buffer (no aliasing).
+    ///
+    /// If these invariants are violated, this method may return an invalid mutable slice,
+    /// leading to undefined behavior.
```
JonasKruckenberg, line 138: same as above

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+    }
+}
+
+impl std::io::Write for BufferWriter {
+    fn write(&mut self, bytes: &[u8]) -> std::io::Result<usize> {
+        // Safety: We assume `buf` is a valid pointer to a properly initialized Buffer.
+        let buffer = unsafe { self.buf.as_mut() };
+
+        // Check if we need to grow the buffer to accommodate the new data
+        if buffer.len + bytes.len() > buffer.capacity {
```
JonasKruckenberg, line 240: maybe use checked addition here too, to avoid nasty wrapping bugs if (for whatever reason) people call this will really large slices?

JonasKruckenberg (approved):
Looks good overall, just a few minor (optional) comment

src/redisearch_rs/buffer/src/lib.rs
zeenix: The "length" here is the actual length/size of the buffer, so I think my docs are correct. :)

src/redisearch_rs/buffer/src/lib.rs
zeenix: True. :+1:

src/redisearch_rs/buffer/src/lib.rs
zeenix: That's a great question. I'd also like to know but this coming from the existing C impl. It's inconsistent for sure. We need to keep in mind that this API will eventually be removed once all the C code that uses Buffers has been oxidized.

src/redisearch_rs/buffer/src/lib.rs
zeenix: I'm not sure if 32-bit systems are targetted by Redis and for 64-bit machines, this will be one hell of a huge buffer that is extremely unlikely to happen IMO. Perhaps one of Redis folks can chime in here? @kei-nan ?

zeenix: I can certainly add a `debug_assert` here though.

src/redisearch_rs/buffer/src/lib.rs
zeenix: So the `buffer` here is the wrapper type (`Buffer`), which is **not** moved by `Buffer_Grow` call. Only the actual underlying buffer is and `Buffer_Grow` updates the contents of the `Buffer` struct accordingly.

chesedo: Ah, I see. Thanks

src/redisearch_rs/buffer/src/lib.rs
chesedo: Isn't length the size used of the available capacity? Hence the `debug_assert!` later to make sure `len <= cap`. This means that capacity should always equal or exceed the length of the buffer.

src/redisearch_rs/buffer/src/lib.rs
chesedo: Since this is only a wrapper (and not a C code port), why does it then need to follow the C code?

zeenix: It's not only a wrapper. The pointer comes from the C side. See #5996 for the use of it.

src/redisearch_rs/buffer/src/lib.rs
zeenix: > /// * `len` must not exceed the length or capacity of the Buffer

Hmm.. I think you've a point about this part though. I think this should be "`len` must not exceed the capacity"

zeenix: I update the docs here and also moved the safety comment for each field to their own docs, since the fields are no longer public.

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+    /// # Safety
+    ///
+    /// It must point to a valid, properly initialized `Buffer`.
+    data: NonNull<u8>,
+    /// The capacity of the buffer (i-e allocated size)
+    ///
+    /// # Safety
+    ///
+    /// This should ideally be actual allocated length of the underlying byte buffer
+    /// but if not, it must not exceed it.
```
LukeMathWalker, line 36:
```suggestion
    /// The capacity must not exceed the size of the allocation behind [`Self::data`].
```

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+    capacity: usize,
+    /// The length of the buffer (i-e the total used memory from allocated memory)
+    ///
+    /// # Safety
+    ///
+    /// Must not exceed the `capacity`.
```
LukeMathWalker, line 42:
```suggestion
    /// - Must not exceed the `capacity`.
    /// - All elements up to [`Self::len`] must be correctly initialized.
```

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+
+use std::{
+    cmp,
+    ptr::{NonNull, copy_nonoverlapping},
+    slice,
+};
+
+/// Redefines the `Buffer` struct from `buffer.h`
+#[repr(C)]
+pub struct Buffer {
```
LukeMathWalker, line 23: As we discussed last week, let's make this a:

```rust
#[repr(transparent)]
pub struct Buffer(ffi::Buffer);
```

using bindgen.

LukeMathWalker: Same strategy for the other two structs.

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+/// Redefines the `BufferWriter` struct from `buffer.h`
+///
+/// A writer for appending data to a `Buffer`. Provides a mechanism to write bytes to a `Buffer`
+/// while tracking the current write position with a cursor.
+///
+/// # Safety Invariants
+///
+/// * `buf` must point to a valid, properly initialized `Buffer`.
+/// * `cursor` must point to a valid position within the Buffer's allocated memory.
+/// * The `BufferWriter` does not own the `Buffer`, only borrows it.
```
LukeMathWalker, line 181:
```suggestion
/// * The `BufferWriter` does not own the `Buffer`, only borrows it mutably.
```

src/redisearch_rs/buffer/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+# Statically link to the libclang on aarch64-unknown-linux-musl,
+# necessary on Alpine.
+# See https://github.com/rust-lang/rust-bindgen/issues/2360
+features = ["bindgen-static", "min-redis-compatibility-version-6-0"]
+workspace = true
+default-features = false
+
+[target.'cfg(not(all(target_env="musl", target_os="linux")))'.dependencies.redis-module]
+workspace = true
+default-features = true
```
LukeMathWalker, line 25: Do we actually need this dep?

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    /// Returns the initialized portion of the buffer as a slice.
+    pub fn as_slice(&self) -> &[u8] {
+        // Safety: We assume `self.data` is a valid pointer as per `Buffer`'s invariants.
+        let data = unsafe { self.data.as_ref() };
+
+        // Safety: We assume `self.len` is a valid length within the allocated memory.
+        // Creates a slice referencing the initialized portion of the buffer.
+        unsafe { slice::from_raw_parts(data, self.len) }
```
LukeMathWalker, line 123:
```suggestion
        unsafe { slice::from_raw_parts(self.data.as_ptr(), self.len) }
```
Same as below, you don't want to create an intermediate reference.

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+            // Safety: We assume `buf` is a valid pointer to a properly initialized `Buffer`.
+            cursor: unsafe { buffer.as_ref() }.data,
+        }
+    }
+}
+
+impl std::io::Write for BufferWriter {
+    fn write(&mut self, bytes: &[u8]) -> std::io::Result<usize> {
+        // Safety: We assume `buf` is a valid pointer to a properly initialized Buffer.
+        let buffer = unsafe { self.buf.as_mut() };
```
LukeMathWalker, line 225: We can't hold a mut reference to a buffer that spans across the call to `Buffer_Grow`, since it will be invalidated if we actually try to grow and need to reallocate.
We should encapsulate its usage:

```rust
        let must_grow = {
            // Safety: We assume `buf` is a valid pointer to a properly initialized Buffer.
            let buffer = unsafe { self.buf.as_mut() };

            // Check if we need to grow the buffer to accommodate the new data
            debug_assert!(
                buffer.len.checked_add(bytes.len()).is_some(),
                "Buffer overflow"
            );
            buffer.len + bytes.len() > buffer.capacity
        };
        if must_grow { /* */}
```

and then recreate it just in time when we need it to update the length:

```rust
        // Safety: We assume `buf` is a valid pointer to a properly initialized Buffer.
        let buffer = unsafe { self.buf.as_mut() };
        // Update the buffer length.
        buffer.len += bytes.len();
```

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+    fn read(&mut self, dest_buf: &mut [u8]) -> std::io::Result<usize> {
+        // Safety: We assume `buf` is a valid pointer to a properly initialized Buffer.
+        let buffer = unsafe { self.buf.as_mut() };
+
+        // Safety: `self.pos` was just checked to be within the limits of the buffer's
+        // initialized range, so this pointer arithmetic is safe.
+        let src = unsafe { buffer.data.add(self.pos) };
+
+        // Safety: We just verified that `src` points to valid memory within the buffer.
+        let src = unsafe { src.as_ref() };
```
LukeMathWalker, line 76:
```suggestion
        let src = src.as_ptr();
```
is enough.

src/redisearch_rs/buffer/src/lib.rs
zeenix: > We can't hold a mut reference to a buffer that spans across the call to `Buffer_Grow`, since it will be invalidated if we actually try to grow and need to reallocate.

The thing is that the `buffer` (which is a pointer to `Buffer`) itself won't be invalidated and that's why I wasn't sure what MIRI was talking about, but I guess since it is unable to reason about what will happen inside `Buffer_Grow`, it assumes that it will be invalidated.

src/redisearch_rs/buffer/src/lib.rs
zeenix: yeah, was the plan but then MIRI came in the way. :cry:

src/redisearch_rs/buffer/src/lib.rs
LukeMathWalker: Semantically equivalent, so that's fine for me.

.github/workflows/task-test.yml
```diff
@@ -50,6 +50,7 @@ env:
   # flags, and therefore reuse the same intermediate build artifacts.
   RUSTFLAGS: "-D warnings"
   RUST_BACKTRACE: "full"
+  CARGO_PROFILE_OPTIMISED_TEST_BUILD_OVERRIDE_DEBUG: "true"
```
kei-nan, line 53: Do we really need this?
It isn't clear what this does or why we need this in the test flow.
At the very least please add a comment explaining what this does.

src/redisearch_rs/buffer/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+name = "buffer"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[lints]
+workspace = true
+
+[lib]
+crate-type = ["staticlib", "rlib"]
```
LukeMathWalker, line 11: I don't think we need to compile this as a static lib.

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+    ffi::c_char,
+    ptr::{NonNull, copy_nonoverlapping},
+    slice,
+};
+
+/// Thin wrapper around the `Buffer` struct from `buffer.h`.
+#[repr(transparent)]
+pub struct Buffer(ffi::Buffer);
+
+/// Redefines the `BufferReader` struct from `buffer.h`
```
LukeMathWalker, line 26:
```suggestion
/// Thin wrapper around the `BufferReader` struct from `buffer.h`
```

src/redisearch_rs/ffi/.gitignore
```diff
@@ -0,0 +1,3 @@
+data/*
+target/
+flamegraph.svg
```
LukeMathWalker, line 3: ?

src/redisearch_rs/ffi/build.rs
```diff
@@ -28,22 +28,6 @@ fn main() {
         println!("cargo:rustc-link-arg=-Wl,--unresolved-symbols=ignore-in-object-files");
     }
 
-    let bin_root = {
```
kei-nan, line 31: why did we remove this?

tests/cpptests/test_cpp_index.cpp
```diff
@@ -1549,13 +1545,12 @@ TEST_F(IndexTest, testVarintFieldMask) {
   BufferWriter bw = NewBufferWriter(&b);
   for (int i = 0; i < sizeof(t_fieldMask); i++, x |= x << 8) {
     size_t sz = WriteVarintFieldMask(x, &bw);
-    ASSERT_EQ(expected[i], sz);
```
kei-nan, line 1552: Can we still have an assert here?

.github/workflows/task-test.yml
zeenix: I added the reasoning in the commit log and forgot that nobody looks at those and are squashed in the end anyway. :facepalm: I'll copy it over as a comment.

src/redisearch_rs/ffi/.gitignore
zeenix: c&p mistake. I created the `ffi` crate by copying an existing one.

src/redisearch_rs/ffi/build.rs
zeenix: because build fails otherwise. The libraries don't seem to be there before this crate is built (or more specifically this build script is run). Without it, the build seems to work fine. @LukeMathWalker I was actually meaning to ask you about this since this was your code. How did you ensure that the libs are always built/made available before the build script runs?

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+        // Safety: We assume `buf` is a valid pointer to a properly initialized Buffer.
+        let buffer = unsafe { &*self.0.buf };
+
+        // Safety: `self.pos` was just checked to be within the limits of the buffer's
+        // initialized range, so this pointer arithmetic is safe.
+        let src = unsafe { buffer.data.add(self.0.pos) } as *const u8;
+
+        // Safety: We just verified that `src` points to valid memory within the buffer.
+        let dest = dest_buf.as_mut_ptr();
+        let len = cmp::min(dest_buf.len(), buffer.offset - self.0.pos);
```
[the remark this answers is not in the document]
zeenix: > can we get a negative number here?

No. Everything here is a `usize`.

> Doesn't look like we avoid reading data if buffer.offset == self.0.pos

I don't think it matters. copying would just be a noop.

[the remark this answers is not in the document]
zeenix: yeah but by `len` so if it's `0`, there is no incrementing happening.

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+    pub unsafe fn buffer_mut(&mut self) -> &mut Buffer {
+        // Safety: We assume `buf` is a valid pointer to a properly initialized `Buffer`.
+        unsafe { &mut *(self.0.buf as *mut Buffer) }
+    }
+}
+
+impl std::io::Write for BufferWriter {
+    fn write(&mut self, bytes: &[u8]) -> std::io::Result<usize> {
+        // Safety: We assume `buf` is a valid pointer to a properly initialized Buffer.
+        let buffer = unsafe { &mut *(self.0.buf as *mut Buffer) };
```
LukeMathWalker, line 212: Use `.buffer_mut()`

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+
+        // Safety:
+        // * `src` is a valid pointer to the input slice.
+        // * `dest` is a valid pointer within the buffer (upheld by our invariants).
+        // * We've either verified or ensured that the buffer has enough capacity.
+        // * The regions don't overlap (copy_nonoverlapping requirement).
+        unsafe { copy_nonoverlapping(src, dest, bytes.len()) };
+
+        // Safety: We assume `buf` is a valid pointer to a properly initialized Buffer.
+        let buffer = unsafe { &mut *(self.0.buf as *mut Buffer) };
```
LukeMathWalker, line 249: Use `buffer_mut()`.

LukeMathWalker (dismissed):

kei-nan (dismissed):

LukeMathWalker (approved):

PR #6177, opened 2025-05-21
Optimize chunk comparisons in wildcard matching

LukeMathWalker:

Use the optimized `is_prefix` routine from `memchr` to compare literal chunks.

Bench results:

```text
foobar vs foobar        time:   [2.6312 ns 2.6336 ns 2.6355 ns]
                        change: [-40.342% -39.767% -39.379%] (p = 0.00 < 0.05)
                        Performance has improved.

fo?bar vs foobar        time:   [6.2195 ns 6.2738 ns 6.3359 ns]
                        change: [-11.746% -10.794% -9.7779%] (p = 0.00 < 0.05)
                        Performance has improved.

fo?bar vs foobarz       time:   [582.61 ps 589.51 ps 601.84 ps]
                        change: [+2.6597% +6.9849% +12.974%] (p = 0.00 < 0.05)
                        Performance has regressed.

foo*baz vs foobarbaz    time:   [8.5217 ns 8.6821 ns 8.9057 ns]
                        change: [-20.867% -19.625% -17.899%] (p = 0.00 < 0.05)
                        Performance has improved.

foo*bar*red??*l*bs? vs foobarbazredxxlxxbsx
                        time:   [25.760 ns 25.843 ns 25.939 ns]
                        change: [-16.295% -15.976% -15.638%] (p = 0.00 < 0.05)
                        Performance has improved.

* vs randomkey          time:   [1.4787 ns 1.4890 ns 1.5035 ns]
                        change: [+1.5281% +2.3489% +3.0888%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/wildcard/src/lib.rs
+++ b/src/redisearch_rs/wildcard/src/lib.rs
@@ -14,6 +14,8 @@
 //! You can create a [`WildcardPattern`] from a pattern using [`WildcardPattern::parse`] and
 //! then rely on [`WildcardPattern::matches`] to determine if a string matches the pattern.
 
+use memchr::arch::all::is_prefix;
+
 mod fmt;
 
 #[derive(Copy, Clone, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


JonasKruckenberg (commented):
Looks good to me, could replicate the same relative benchmark gains locally too. Only question is the "* vs randomkey" regression 🤔 how common is that use case? I guess given the other gains this would be a worthwhile tradeoff but 11% is a relevant regression

src/redisearch_rs/wildcard/src/lib.rs
```diff
@@ -245,18 +247,21 @@ impl<'pattern> WildcardPattern<'pattern> {
...
+                    };
+                    if !is_prefix(&key[i_k..], &chunk[..min_len]) {
+                        try_backtrack!(bt_state, i_t, i_k, 'outer);
+                    }
+                    if is_partial_match {
+                        return MatchOutcome::PartialMatch;
+                    } else {
+                        i_t += 1;
+                        i_k += chunk.len();
                     }
```
zeenix, line 264: nitpick: with early exit in the if block, `else` block becomes redundant and best not to nest when not needed for readability:

```suggestion
                    if is_partial_match {
                        return MatchOutcome::PartialMatch;
                    }

                    i_t += 1;
                    i_k += chunk.len();
```

zeenix (approved):
lgtm otherwise

LukeMathWalker:
> Looks good to me, could replicate the same relative benchmark gains locally too. Only question is the "* vs randomkey" regression 🤔 how common is that use case? I guess given the other gains this would be a worthwhile tradeoff but 11% is a relevant regression

That's a degenerate case—if you want all entries, you should do a plain search rather than a wildcard search. So I'd say it's an acceptable trade-off.

PR #6179, opened 2025-05-21
Shadow tester

chesedo:

This provides a skeleton for a new shadow tester. The idea is to have a tester which can run a bunch of queries against a master version of the Redisearch `.so` file and the `.so` file of changes being worked on. This tester then makes sure that two redis servers (with each `.so` file as a module) return the exact same responses for a set of queries.

In the future this can be expanded to run more queries, etc.

#### Mark if applicable


src/redisearch_rs/tools/shadow_tester/src/args.rs
```diff
@@ -0,0 +1,20 @@
...
+    #[arg(short, long)]
+    pub baseline_so: PathBuf,
+
+    /// Path to the changeset .so file. The responses from this module should match the baseline's.
+    #[arg(short, long)]
+    pub changeset_so: PathBuf,
+
+    /// Command to start up RLTest which will start a Redis server for each .so file.
+    #[arg(short, long)]
+    pub rltest_command: String,
```
LukeMathWalker, line 19: Isn't this a path too?

chesedo: Yes, in our usage but technically it can be a command too. Eg I think it will work when given `python3 -m RLTest ...` too. So I'm torn between making it a path or a string.

src/redisearch_rs/tools/shadow_tester/src/redis_client.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn drop(&mut self) {
+        println!("killing RLTest server");
+
+        kill_process(&mut self.server_process).expect("to be able to kill the RLTest server");
+    }
+}
+
+/// Kill the `runtests.sh` process correctly.
+/// Technically this has an issue since the `runtests.sh` process does not correctly stop the
+/// RLTest process it spawns when it is killed. To kill the RLTest process, one has to manually
```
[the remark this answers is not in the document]
chesedo: Oh, this is actually correct. The issue is quite complicated so I should properly expand the comment more. Basically this is what happens:

- `runtests.sh` spawns
- `RLTest` which internally spawns
- `redis-server`

Now, when `runtests.sh` receives a kill signal (from this function) it does not stop the `RLTest` process it spawned. However, manually killing the `RLTest` process still causes the `redis-server` process to hang around. This means manually killing the `redis-server` process too.

However, if one kills the `redis-server` process without killing the `RLTest` process first, then that causes the `RLTest` to reach an end naturally. Hence saving on manual process kill.

In the bigger context that means the comment is correct: `runtests.sh` does not the the `RLTest` it spawned. And the very next line then stating `redis-server` has to be killed is correct since that automatically stops the `RLTest` process too.

I'll figure out how I can reword the comment.

src/redisearch_rs/tools/shadow_tester/src/args.rs
```diff
@@ -0,0 +1,20 @@
+use std::path::PathBuf;
+
+use clap::Parser;
+
+#[derive(Parser)]
+#[command(version, about)]
+pub struct Options {
+    /// Path to the baseline .so file. The responses from this module will be considered the source
+    /// of truth.
+    #[arg(short, long)]
```
JonasKruckenberg, line 10: I think it doesn't add much value to have short (-b,-c) versions of these flags. They will only cause confusion

chesedo: Fair point. WIll change

src/redisearch_rs/tools/shadow_tester/src/args.rs
chesedo: Giving it a command won't work. I'm changing this to a path too.

src/redisearch_rs/tools/shadow_tester/src/redis_client.rs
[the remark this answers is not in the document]
LukeMathWalker: Oh, I had missed the intermediate process—thanks for the explanation!

PR #6183, opened 2025-05-21
Shadow tests CI

JonasKruckenberg:

builds off #6179 

This PR implements the necessary GitHub Actions workflow to run the shadow tests in CI when the `enforce:refactor` label is present on a pull request.

## TODOS

- [ ] should the label be `enforce:refactor` or something like `enforce:shadow-tests` or `enforce:query-diff`
- [ ] Should the new tests be called "Query Diff Tests" or "Shadow Tests"? We should update the naming across the tool & CI accordingly.

#### Mark if applicable

### DIFF
--- a/.github/workflows/event-pull_request.yml
+++ b/.github/workflows/event-pull_request.yml
@@ -84,6 +84,17 @@ jobs:
       rejson-branch: master
       test-config: QUICK=1
 
+  query-diff:
+    # needs: [check-what-changed, get-latest-redis-tag]
+    if: contains(github.event.pull_request.labels.*.name, 'enforce:refactor')
+    secrets: inherit
+    uses: ./.github/workflows/task-query-diff.yml
+    with:
+      san: address
+      get-redis: ${{ needs.get-latest-redis-tag.outputs.tag }}
+      rejson-branch: master
+      test-config: QUICK=1
+
   pr-validation:
     needs:
       - get-latest-redis-tag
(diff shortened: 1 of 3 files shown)


.github/workflows/task-query-diff.yml
```diff
@@ -0,0 +1,61 @@
...
+          ls ${{ env.baseline_so }}
+          ls ${{ env.changeset_so }}
+          cargo run --manifest-path src/redisearch_rs/Cargo.toml \
+            -p shadow_tester \
+            --baseline_so ${{ env.baseline_so }} \
+            --changeset_so ${{ env.changeset_so }} \
+            --rltest-path tests/pytests/runtests.sh
+        env:
+          baseline_so: bin/query_diff/module-enterprise.so
+          changeset_so: bin/linux-x64-release/search-community/redisearch.so
```
LukeMathWalker, line 61: Enterprise and search-community are not 1:1 since they use a different coordinator.
Don't we have snapshots the search community build?

JonasKruckenberg: Yeah I figured as much, but I don't think there are snapshots for the non-enterprise version. Or at least non I could find with my nonexistent access to the S3 bucket 🤷🏻‍♂️ 
The only file I could reverse engineer is `s3://redismodules/redisearch/snapshots/redisearch.Linux-ubuntu24.04-x86_64.master.zip` (plus ofc for other platforms) and unzipping that only give you a .json manifest and the enterprise module

.github/workflows/event-pull_request.yml
```diff
@@ -84,6 +84,21 @@ jobs:
       rejson-branch: master
       test-config: QUICK=1
 
+  query-diff:
+    needs: [check-what-changed, get-latest-redis-tag]
+    if: >
+      (
+        (!github.event.pull_request.draft && needs.check-what-changed.outputs.CODE_CHANGED == 'true') ||
+        contains(github.event.pull_request.labels.*.name, 'enforce:refactor')
+      )
```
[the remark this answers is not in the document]
JonasKruckenberg: Kept it as is bc I didn't know how to simulate labels in act, but found a way now so all good

PR #6194, opened 2025-05-22
Allow warning to compile boost qvm with clang17

DarthB:
Small improvements to build script for clang17

Update build.sh:
allow missing-template-arg-list-template-kw

Update src/redisearch_rs/build.rs:
enhance assert message for missing lib


raz-mon (approved):

DarthB:
The fix breaks later clang versions in CI, so minimum clang is 18 now.

PR #6201, opened 2025-05-23
use github api to get artifact from correct workflow run

DarthB:

Fixes the retrieval of the artifact zip from master by using the github API, such that the PRs actually download something and compare their new benchmarks.

### DIFF
--- a/.github/workflows/flow-micro-benchmarks.yml
+++ b/.github/workflows/flow-micro-benchmarks.yml
@@ -18,12 +18,36 @@ jobs:
         run: ./install_script.sh sudo
       - name: Build RediSearch
         run: ./build.sh
-      - name: Get Latest Baseline
-        uses: actions/download-artifact@v4
-        with:
-          name: rust-benchmark-results-master
-          path: bin/redisearch_rs/criterion
-        continue-on-error: true # Prevent failure if no baseline exists yet
+        
+       # Step to fetch the latest artifact from master
+      - name: Get Latest Baseline Artifact URL
+        id: get-artifact-url
+        if: github.event_name == 'pull_request'
+        run: |
+          RUN_ID=$(curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            -H "Accept: application/vnd.github+json" \
+            "https://api.github.com/repos/${{ github.repository }}/actions/runs?per_page=1&branch=master&status=success" \
+            | jq -r '.workflow_runs[0].id')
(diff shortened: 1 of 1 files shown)


sbin/fetch_micro_bench_artifact.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
LukeMathWalker, line 1: This can probably be significantly simplified by using GitHub's CLI (`gh`) rather than going through the raw API.

LukeMathWalker (approved):

JoanFM:
I do not think that adding more of these scripts is better than running the micro benchmarks on every merge to master. I would vote for having this second option

LukeMathWalker:
> I do not think that adding more of these scripts is better than running the micro benchmarks on every merge to master. I would vote for having this second option

I'd be happy to go down that path—it simplifies everything. It's mostly a matter of CI cost/usage. If that's not an issue, let's do it.

JoanFM:
> > I do not think that adding more of these scripts is better than running the micro benchmarks on every merge to master. I would vote for having this second option
> 
> I'd be happy to go down that path—it simplifies everything. It's mostly a matter of CI cost/usage. If that's not an issue, let's do it.

I think if that becomes an issue we should fix then, but do not early optimize this

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -17,13 +17,18 @@ jobs:
...
         run: ./install_script.sh sudo
       - name: Build RediSearch
-        run: ./build.sh
-      - name: Get Latest Baseline
-        uses: actions/download-artifact@v4
+        run: ./build.sh  
+      - name: Download Latest Baseline Artifact from master
+        id: get-artifact
+        if: github.event_name == 'pull_request'
+        uses: dawidd6/action-download-artifact@4c1e823582f43b179e2cbb49c3eade4e41f992e2 # refs @v10
```
JoanFM, line 24: why do we need to change the action ?

DarthB: We use the external action to access the master branch from the PR. 
The action supports this and allows us to avoid using an external script file.
Internally, it first retrieves the workflow runs from master and then searches the artifact.

JoanFM: but the old action did not do the same?

DarthB: If you refer to [actions/download-artifact@v4](https://github.com/actions/download-artifact): 
That action only works on the current workflow. You can specify a run-id, but then we need script calls against the GitHub API to figure that id out.

JoanFM (approved):

PR #6202, opened 2025-05-25
Fix Coverage Deadlock In GC During Tests

kei-nan:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -189,10 +189,14 @@ setup_build_environment() {
 # Run lcov preparations before testing for coverage
 #-----------------------------------------------------------------------------
 prepare_coverage_capture() {
-  [[ -n $GITHUB_ACTIONS ]] && echo "::group::Code Coverage Preparation"
+  if [[ -n $GITHUB_ACTIONS ]]; then
+    echo "::group::Code Coverage Preparation"
+  fi
   lcov --zerocounters      --directory $BINROOT --base-directory $ROOT
   lcov --capture --initial --directory $BINROOT --base-directory $ROOT -o $BINROOT/base.info
-  [[ -n $GITHUB_ACTIONS ]] && echo "::endgroup::"
+  if [[ -n $GITHUB_ACTIONS ]]; then
+    echo "::endgroup::"
+  fi
 }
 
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -189,10 +189,14 @@ setup_build_environment() {
 # Run lcov preparations before testing for coverage
 #-----------------------------------------------------------------------------
 prepare_coverage_capture() {
-  [[ -n $GITHUB_ACTIONS ]] && echo "::group::Code Coverage Preparation"
+  if [[ -n $GITHUB_ACTIONS ]]; then
+    echo "::group::Code Coverage Preparation"
+  fi
```
GuyAv46, line 194: Lets fix it on the `capture_coverage` as well

src/fork_gc.c
```diff
@@ -66,7 +66,7 @@ static void FGC_sendFixed(ForkGC *fgc, const void *buff, size_t len) {
     perror("broken pipe, exiting GC fork: write() failed");
     // just exit, do not abort(), which will trigger a watchdog on RLEC, causing adverse effects
     RedisModule_Log(fgc->ctx, "warning", "GC fork: broken pipe, exiting");
-    exit(1);
+    RedisModule_ExitFromChild(1);
```
GuyAv46, line 69: good catch

src/fork_gc.c
```diff
@@ -290,7 +290,12 @@ static void sendHeaderString(ForkGC *gc, void *arg) {
 }
 
 static void FGC_reportProgress(ForkGC *gc) {
+#if defined(COV) || defined(COVERAGE)
+  // Don't send heartbeats when running coverage tests, as it can mess with the coverage report.
+  return;
+#else
   RedisModule_SendChildHeartbeat(gc->progress);
+#endif
```
GuyAv46, line 298: Consider simplifying the logic (no `else` case)
```suggestion
#if !defined(COV) && !defined(COVERAGE)
  // Don't send heartbeats when running coverage tests, as it can mess with the coverage report.
  RedisModule_SendChildHeartbeat(gc->progress);
#endif
```

PR #6209, opened 2025-05-26
Shadow tests - `FT.SEARCH` query

chesedo:

This provides a skeleton for a new shadow tester. The idea is to have a tester which can run a bunch of queries against a master version of the Redisearch `.so` file and the `.so` file of changes being worked on. This tester then makes sure that two redis servers (with each `.so` file as a module) return the exact same responses for a set of queries.

In the future this can be expanded to run more queries, etc.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/tools/shadow_tester/src/test_runner.rs
+++ b/src/redisearch_rs/tools/shadow_tester/src/test_runner.rs
@@ -84,7 +84,7 @@ impl TestRunner {
     }
 }
 
-/// Helper to turn a [redis::Arg] into a [String]
+/// Helper to turn a [`redis::Arg`] into a [`String`]
 fn arg_to_str(arg: Arg<&[u8]>) -> String {
     let Arg::Simple(s) = arg else {
         panic!("Expected a simple argument");


src/redisearch_rs/tools/shadow_tester/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "shadow_tester"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+clap = { version = "4.5.38", features = ["derive"] }
+pretty_assertions = "1.4.1"
+redis = "0.31.0"
```
LukeMathWalker, line 9: All dependencies should be declared at the workspace level and then inherited.

src/redisearch_rs/tools/shadow_tester/src/args.rs
```diff
@@ -0,0 +1,21 @@
...
+    pub baseline_so: PathBuf,
+
+    /// Path to the changeset .so file. The responses from this module should match the baseline's.
+    #[arg(long)]
+    pub changeset_so: PathBuf,
+
+    /// Path to a script which start up a RLTest server which will start a Redis server for each
+    /// .so file.
+    #[arg(long)]
+    pub rltest_path: PathBuf,
```
LukeMathWalker, line 20:
```suggestion
    pub rltest: PathBuf,
```
For naming consistency with the other options.

src/redisearch_rs/tools/shadow_tester/src/datasets.rs
```diff
@@ -0,0 +1,127 @@
...
+        ],
+    ));
+
+    Ok(setup_queries)
+}
+
+/// Gets the 1984 book text from the cache if it exists, otherwise fetches it from the Gutenberg
+/// Project and stores it in the cache.
+fn get_1984_text() -> Result<String, ureq::Error> {
+    let cache_path = PathBuf::from("cache/1984.txt");
```
LukeMathWalker, line 106: This way, the cache path depends on the current directory when the tool is invoked. Is that the intention?

src/redisearch_rs/tools/shadow_tester/src/datasets.rs
```diff
@@ -0,0 +1,127 @@
...
+pub fn get_1984_index() -> Result<Vec<(String, Vec<String>)>, ureq::Error> {
+    let text = get_1984_text()?;
+
+    let mut part_num = 0;
+    let mut part_len = 0;
+    let mut index_pos = 0;
+    let mut part_text = vec![];
+
+    let mut setup_queries = vec![
+        ("flushall".to_string(), vec![]),
```
LukeMathWalker, line 25: What does this do/why do we need it?

src/redisearch_rs/tools/shadow_tester/src/datasets.rs
```diff
@@ -0,0 +1,127 @@
...
+                "part_num".to_string(),
+                "NUMERIC".to_string(),
+                "SORTABLE".to_string(),
+                "index_pos".to_string(),
+                "NUMERIC".to_string(),
+                "SORTABLE".to_string(),
+                "length".to_string(),
+                "NUMERIC".to_string(),
+                "SORTABLE".to_string(),
+            ],
```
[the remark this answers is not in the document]
LukeMathWalker: Or, at the very least, avoid the `to_string` everywhere.

src/redisearch_rs/tools/shadow_tester/src/datasets.rs
chesedo: No, not really intended. Just used a simpler implementation from the `trie_bencher` to get the file. Do you have a specific static location for the cache folder?

src/redisearch_rs/tools/shadow_tester/src/datasets.rs
chesedo: Since the `redis-server` does not shutdown correctly, this is to make sure the DB is clean on consecutive local runs. However, `flushall` actually does way to much and will interfere once we have other datasets too. I'll find queries to just clean up this dataset.

src/redisearch_rs/tools/shadow_tester/src/datasets.rs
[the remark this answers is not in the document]
chesedo: So having it all as one string gives this error

```
An error was signalled by the server - ResponseError: wrong number of arguments for '_FT.CREATE' command
```

The `redis` crate (correctly) binds the entire string as a single argument :sweat_smile: 

My alternative is using this

```
        (
            "FT.CREATE".to_string(),
            "idx:1984 ON HASH PREFIX 1 1984:part: SCHEMA text TEXT WEIGHT 1.0 part_num NUMERIC SORTABLE index_pos NUMERIC SORTABLE length NUMERIC SORTABLE".split(" ").map(|s| s.to_string()).collect(),
        ),
```

src/redisearch_rs/tools/shadow_tester/src/test_runner.rs
```diff
@@ -0,0 +1,119 @@
...
+            println!("{diff}");
+
+            false
+        } else {
+            true
+        }
+    }
+}
+
+/// Helper to turn a [redis::Arg] into a [String]
```
LukeMathWalker, line 87:
```suggestion
/// Helper to turn a [`redis::Arg`] into a [`String`]
```

LukeMathWalker (dismissed):

chesedo:
We can look at this again when we have more time

PR #6211, opened 2025-05-26
MOD-9835: Fix unicode_tolower()

dor-forer:

1. Current: 
In `unicode_tolower()`, for characters where the lowercase version occupies more bytes than its uppercase version, the `for` loop which traverse the utf-8 string is exceeding the limits of the allocated memory when the length of the lowercase version of the term is longer than its original form.
In `tag_strtolower()` and `rm_normalize()`, the length passed to `unicode_tolower()` is the length of the original term, but it should be the length of the term after unescaping it.

2. Change: 
- In `unicode_tolower()`: 
    - Replace the `for` loop for a `while` loop which check that the current utf-8 symbols is inside the valid limits.
    - Break the loop if a NULL character is found
- In `tag_strtolower()` and `rm_normalize()`: 
    - Pass to `unicode_tolower()` the updated length after the term is unescaped.
     
3. Outcome: 

1. MOD-9835

1. ...

#### Mark if applicable

### DIFF
--- a/tests/pytests/test_multibyte_char_terms.py
+++ b/tests/pytests/test_multibyte_char_terms.py
@@ -1506,7 +1506,7 @@ def test_utf8_lowercase_longer_than_uppercase_tags(env):
             'FT.SEARCH', 'idx', f'@t:{{{t1_lower}}}', 'NOCONTENT', 'DIALECT', dialect)
         env.assertEqual(res, [1, '{doc}:lower:1'])
 
-        # 63 characteres, occupying 126 bytes in UTF-8 + 1 byte for the
+        # 63 characters, occupying 126 bytes in UTF-8 + 1 byte for the
         # null terminator, so the total length is 127 bytes
         t2 = 'İ' * 63
         # 63 characters, occupying 189 bytes in UTF-8 + 1 byte for the
@@ -1586,7 +1586,7 @@ def test_utf8_lowercase_longer_than_uppercase_texts(env):
             'FT.SEARCH', 'idx', f'@t:({t1_lower})', 'NOCONTENT', 'DIALECT', dialect)
         env.assertEqual(res, [1, '{doc}:lower:1'])
 
-        # 63 characteres, occupying 126 bytes in UTF-8 + 1 byte for the
+        # 63 characters, occupying 126 bytes in UTF-8 + 1 byte for the
         # null terminator, so the total length is 127 bytes
         t2 = 'İ' * 63
         # 63 characters, occupying 189 bytes in UTF-8 + 1 byte for the


src/util/strconv.h
```diff
@@ -115,57 +116,64 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
...
-    else {
+    } else {
       // If no transformation is needed, just copy the unicode codepoint
-      u_buffer[i] = codepoint;
-      ++i;
+      u_buffer[i++] = codepoint;
     }
   }
-
+  RS_LOG_ASSERT_FMT(i <= u_len, "i (%u) should be less equal to u_len (%zd)", i, u_len);
```
meiravgri, line 163:
```suggestion
  RS_LOG_ASSERT_FMT(i <= u_len, "i (%u) should be equal to u_len (%zd)", i, u_len);
```

src/util/strconv.h
```diff
@@ -115,57 +116,64 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
...
     encoded_char = nu_utf8_read(encoded_char, &codepoint);
+    if (codepoint == 0) {
+      // If we reach the end of the string, break
+      break;
+    }
     // Transform unicode codepoint to lower case
     const char *map = nu_tolower(codepoint);
 
     // Read the transformed codepoint and store it in the unicode buffer
-    // map would be NULL if no transformation is needed,
```
meiravgri, line 143: revert?

src/util/strconv.h
```diff
@@ -115,57 +116,64 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
...
     nu_writenstr(u_buffer, u_len, encoded, nu_utf8_write);
   } else {
     reencoded_len = 0;
+    // reallocate the string to fit the transformed string
+    // rm_free(encoded);
+    // encoded = (char *)rm_malloc(reencoded_len + 1);
+    // nu_writenstr(u_buffer, u_len, encoded, nu_utf8_write);
+
+    // encoded[reencoded_len] = '\0'; // Null-terminate the string
+    // // Update the length to the new length
```
meiravgri, line 176: Remove and consider adding a comment to explain why we would enter the else (or elaborate the documentation at the header regarding return value = 0).

src/util/strconv.h
nafraf: Reverted.

src/query.c
```diff
@@ -1077,7 +1077,6 @@ typedef IndexIterator **IndexIteratorArray;
  * should be performed. If true, the string remains case-sensitive.
  */
 static void tag_strtolower(char *str, size_t *len, int caseSensitive) {
-  size_t origLen = *len;
```
meiravgri, line 1080: To avoid redundant memory accesses, consider storing the updated length in a local variable and updating `*len` only once at the end.

src/util/strconv.h
```diff
@@ -149,23 +155,30 @@ static size_t unicode_tolower(char *encoded, size_t in_len) {
...
   }
-
+  RS_LOG_ASSERT_FMT(i == u_len, "i (%u) should be equal to u_len (%zd)", i, u_len);
   // Encode Unicode codepoints back to utf8 string
   ssize_t reencoded_len = nu_bytenlen(u_buffer, u_len, nu_utf8_write);
   if (reencoded_len > 0 && reencoded_len <= in_len) {
     nu_writenstr(u_buffer, u_len, encoded, nu_utf8_write);
   } else {
     reencoded_len = 0;
+    // If the reencoded length is greater than in_len, we cannot write it back
```
meiravgri, line 172: 🚀

src/util/strconv.h
```diff
@@ -177,7 +190,7 @@ static size_t unicode_tolower(char *encoded, size_t in_len) {
 }
 
 
-// strndup + unescape + fold
+// strndup + unescape + tolower
 static char *rm_normalize(const char *s, size_t len) {
```
meiravgri, line 194: no need to update len here as in `tag_strtolower?`

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -1434,3 +1441,92 @@ def testTagSearch(env):
...
+    env.expect('HSET', '{doc}:1', 't', t).equal(1)
+    if not env.isCluster():
+        res = env.cmd(debug_cmd(), 'DUMP_TAGIDX', 'idx', 't')
+        # The conversion to lowercase occupies more space than the original
+        # term, so the term is stored in its original form
+        env.assertEqual(res, [[t, [1]]])
+
+    env.expect('HSET', '{doc}:2', 't', t_lower).equal(1)
+    if not env.isCluster():
+        res = env.cmd(debug_cmd(), 'DUMP_TAGIDX', 'idx', 't')
```
meiravgri, line 1463: Once reallocation is enabled, `DUMP_TAGIDX` will return only one term and the assert is expected to fail, right?

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -1434,3 +1441,92 @@ def testTagSearch(env):
...
+        env.assertEqual(res, [[t, [1]], [t_lower, [2]]])
+
+    for dialect in range(1, 5):
+        res = conn.execute_command(
+            'FT.SEARCH', 'idx', f'@t:{{{t_escaped}}}', 'NOCONTENT', 'DIALECT', dialect)
+        env.assertEqual(res, [1, '{doc}:1'])
+
+        res = conn.execute_command(
+            'FT.SEARCH', 'idx', f'@t:{{{t_escaped_lower}}}', 'NOCONTENT', 'DIALECT', dialect)
+        env.assertEqual(res, [1, '{doc}:2'])
```
meiravgri, line 1475: Once reallocation is enabled, should each query return both docs?

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -1434,3 +1441,92 @@ def testTagSearch(env):
 
     res = conn.execute_command('FT.SEARCH', 'idx', '@t:{fussball}', 'NOCONTENT')
     env.assertEqual(res, [1, 'doc:2s'])
+
+def testLongTags(env):
```
meiravgri, line 1445:
```suggestion
def test_utf8_lowercase_longer_than_uppercase_tags(env):
```

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -1434,3 +1441,92 @@ def testTagSearch(env):
...
+        env.assertEqual(res, [1, '{doc}:1'])
+
+        # Search lowercase escaped term
+        res = conn.execute_command(
+            'FT.SEARCH', 'idx', f'@t:({t_escaped_lower})', 'NOCONTENT', 'DIALECT', dialect)
+        env.assertEqual(res, [1, '{doc}:2'])
+
+        # If we don't escape the term, each word is treated as a separate term,
+        # so the search will return no results
+        res = conn.execute_command(
```
meiravgri, line 1527: With tag, we got a syntax error when we tried to execute the query without escaping 🤔 
1. Was it because we used dialect 1?
2. This problem doesn't exist with text?

meiravgri (commented):
great!! some small comemnts and don't listen to copilot

tests/pytests/test_multibyte_char_terms.py
nafraf: Yes.

src/query.c
nafraf: OK. Done.

src/util/strconv.h
nafraf: I think we don't need it. Since in `tag_strtolower` we update the contents of the input string, we also update its length, but for `rm_normalize` we duplicate the input string and return a new pointer.

tests/pytests/test_multibyte_char_terms.py
meiravgri: cool
maybe add a small comment as a reminder

src/query.c
```diff
@@ -1077,25 +1077,27 @@ typedef IndexIterator **IndexIteratorArray;
...
     }
     *str++ = *p++;
   }
   *str = '\0';
 
   if (!caseSensitive) {
-    size_t newLen = unicode_tolower(origStr, *len);
+    size_t newLen = unicode_tolower(origStr, length);
     if (newLen) {
       origStr[newLen] = '\0';
```
meiravgri, line 1096: Just a small thought — I think this part might be a bit sensitive down the line when reallocation is added. If the buffer is resized to exactly `newLen`, then writing `origStr[newLen] = '\0';` would write one byte past the end. Would be easy to miss, and could cause a subtle overflow 😅.

Maybe worth adding a test to catch this kind of edge case early? The Turkish 'İ' (U+0130) is a good one — it’s 2 bytes uppercase and becomes 3 bytes lowercase. 😊

tests/pytests/test_multibyte_char_terms.py
nafraf: 1. No, it also occurs with other dialects.
2. It doesn't happens with text, because for TEXT the minus `-` is considered as NOT operator, but for TAG it is not allowed to have the minus inside the '{}'

<details>

  <summary>Test with redisearch 2.10.17</summary>

```sh
127.0.0.1:6379> FT.CREATE idx schema t tag x text NOSTEM
OK
127.0.0.1:6379> FT.explaincli idx '@x:(hello - world)'  dialect 1
1) @x:INTERSECT {
2)   @x:hello
3)   @x:NOT{
4)     @x:world
5)   }
6) }
7) 
127.0.0.1:6379> FT.explaincli idx '@x:(hello - world)'  dialect 2
1) @x:INTERSECT {
2)   @x:hello
3)   @x:NOT{
4)     @x:world
5)   }
6) }
7) 
127.0.0.1:6379> FT.explaincli idx '@t:{hello - world}'  dialect 1
(error) Syntax error at offset 17 near world
127.0.0.1:6379> FT.explaincli idx '@t:{hello - world}'  dialect 2
(error) Syntax error at offset 10 near hello
```

</details>

meiravgri: So how come this one doesn't throw a syntax error in the tag test? what am i missing?
```
        if dialect > 1:
            res = conn.execute_command(
                'FT.SEARCH', 'idx', f'@t:{{"{t}"}}', 'NOCONTENT', 'DIALECT', '2')
            # The term is stored in its original form, so the search will return the
            # document with the original term
            env.assertEqual(res, [1, '{doc}:1'])
```

nafraf: In that case, we are using TAG auto escaping, the user don't need to escape the special character if the text is between double quotes.
The description is in the last example of [Tag field](https://redis.io/docs/latest/develop/interact/search-and-query/query/exact-match/) section.

tests/pytests/test_multibyte_char_terms.py
nafraf: OK. Done.

tests/pytests/test_multibyte_char_terms.py
```diff
@@ -1434,3 +1441,176 @@ def testTagSearch(env):
...
+            # The term is stored in its original form, so the search will return the
+            # document with the original term
+            env.assertEqual(res, [1, '{doc}:1'])
+
+            # Search lowercase term
+            res = conn.execute_command(
+                'FT.SEARCH', 'idx', f'@t:{{"{t_lower}"}}', 'NOCONTENT', 'DIALECT', dialect)
+            env.assertEqual(res, [1, '{doc}:2'])
+
+        # Test the edge cases where the length is around the limit to require
```
meiravgri, line 1492: By “reallocation” I meant cases where the lowercase output is longer than the input — a single-character string like İ would cover that.
Is there a reason to use the more complex loop if a simpler test would make the case clearer? (Especially since the previous test already covers sequences that exceed the stack allocation size.)

meiravgri (approved):
well done!

src/util/strconv.h
```diff
@@ -115,27 +116,32 @@ static char *rm_strndup_unescape(const char *s, size_t len) {
...
+  uint32_t u_stack_buffer[SSO_MAX_LENGTH];
+  uint32_t *u_buffer = u_stack_buffer;
+
   ssize_t u_len = nu_strtransformnlen(encoded, in_len, nu_utf8_read,
                                               nu_tolower, nu_casemap_read);
 
-  if (u_len >= (SSO_MAX_LENGTH - 1)) {
-    u_buffer = (uint32_t *)rm_malloc(sizeof(*u_buffer) * (u_len + 1));
+  if (u_len > (SSO_MAX_LENGTH - 1)) {
+    u_buffer = (uint32_t *)rm_malloc(sizeof(uint32_t) * (u_len + 1));
```
oshadmi: Why change `sizeof(*u_buffer)` ?

nafraf: Change reverted.

PR #6220, opened 2025-05-27
[MOD-9798] Refactor `inverted_index`

chesedo:

This PR refactors `inverted_index` to prepare it for porting to Rust. The refactor does not change functionality in any way. But makes the following changes:
1. Removes the dependency on `forward_index.h` by moving `InvertedIndex_WriteForwardIndexEntry` to `forward_index`.
1. Removes definitions which are not implemented from `inverted_index.h`
1. Makes public functions that are only used by `inverted_index` private.

#### Mark if applicable

### DIFF
--- a/src/inverted_index.h
+++ b/src/inverted_index.h
@@ -12,7 +12,6 @@
 #include "redisearch.h"
 #include "buffer.h"
 #include "doc_table.h"
-#include "forward_index.h"
 #include "index_iterator.h"
 #include "index_result.h"
 #include "spec.h"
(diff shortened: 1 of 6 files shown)


src/inverted_index.h
```diff
@@ -63,11 +63,6 @@ typedef union {
   const NumericFilter *filter;
 } IndexDecoderCtx;
 
-/**
- * Called when an entry is removed
- */
-typedef void (*RepairCallback)(const RSIndexResult *res, void *arg);
```
zeenix, line 70: :+1: I wonder if there would be a way to automatically figure out all the used C API so we can remove all of it at once and then not have to worry about this in the future. :thinking:

zeenix (approved):
LGTM.

GuyAv46 (approved):

PR #5619, opened 2025-02-08
Refactor Iterators

GuyAv46:
## Describe the changes in the pull request

A clear and concise description of what the PR is solving, including:
1. Current: The current state briefly
2. Change: What is the change
3. Outcome: Adding the outcome

#### Which additional issues this PR fixes
1. MOD-...
2. #...

#### Main objects this PR modified
1. ...

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #6221, opened 2025-05-27
add IDList iterator refactor - [MOD-8846]

JoanFM:

It contains the code for the IDListIterator refactor.

Implementation has been taken from https://github.com/RediSearch/RediSearch/pull/5619/files

#### Mark if applicable

### DIFF
--- a/src/iterators/idlist_iterator.c
+++ b/src/iterators/idlist_iterator.c
@@ -38,6 +38,7 @@ static IteratorStatus IL_Read(QueryIterator *base) {
 static IteratorStatus IL_SkipTo(QueryIterator *base, t_docId docId) {
   IdListIterator *it = (IdListIterator *)base;
   if (isEof(it) || it->offset >= it->size) {
+    setEof(it, 1);
     return ITERATOR_EOF;
   }
 
--- a/src/iterators/idlist_iterator.c
+++ b/src/iterators/idlist_iterator.c
@@ -1,0 +1,125 @@
+/*
+ * Copyright Redis Ltd. 2016 - present
+ * Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
+ * the Server Side Public License v1 (SSPLv1).
+ */
+
+#include "idlist_iterator.h"
+
+static inline void setEof(IdListIterator *it, int value) {
+  it->base.atEOF = value;
+}
+
+static inline int isEof(const IdListIterator *it) {
+  return it->base.atEOF;
+}
+
+static size_t IL_NumEstimated(QueryIterator *base) {
+  IdListIterator *it = (IdListIterator *)base;
+  return (size_t)it->size;
+}
+ 
+/* Read the next entry from the iterator, into hit *e.
+*  Returns ITERATOR_EOF if at the end */
+static IteratorStatus IL_Read(QueryIterator *base) {
+  IdListIterator *it = (IdListIterator *)base;
+  if (isEof(it) || it->offset >= it->size) {
+    setEof(it, 1);
+    return ITERATOR_EOF;
+  }
+
+  base->lastDocId = it->docIds[it->offset++];
+  it->base.current->docId = base->lastDocId;
+  return ITERATOR_OK;
+}
+ 
+/* Skip to a docid, potentially reading the entry into hit, if the docId
+* matches */
+static IteratorStatus IL_SkipTo(QueryIterator *base, t_docId docId) {
+  IdListIterator *it = (IdListIterator *)base;
+  if (isEof(it) || it->offset >= it->size) {
+    return ITERATOR_EOF;
+  }
+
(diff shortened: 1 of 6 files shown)


src/iterators/idlist_iterator.c
```diff
@@ -0,0 +1,126 @@
...
+ * the Server Side Public License v1 (SSPLv1).
+ */
+
+#include "idlist_iterator.h"
+
+static inline void setEof(IdListIterator *it, int value) {
+  it->base.atEOF = value;
+}
+
+static inline int isEof(const IdListIterator *it) {
```
BenGoldberger, line 13:
```suggestion
static inline bool isEof(const IdListIterator *it) {
```

src/iterators/idlist_iterator.c
```diff
@@ -0,0 +1,126 @@
...
+  IdListIterator *it = (IdListIterator *)base;
+  return (size_t)it->size;
+}
+ 
+/* Read the next entry from the iterator, into hit *e.
+*  Returns ITERATOR_EOF if at the end */
+static IteratorStatus IL_Read(QueryIterator *base) {
+  IdListIterator *it = (IdListIterator *)base;
+  if (isEof(it) || it->offset >= it->size) {
+    setEof(it, 1);
```
BenGoldberger, line 27:
```suggestion
    setEof(it, true);
```

src/iterators/idlist_iterator.c
```diff
@@ -0,0 +1,126 @@
...
+  it->base.current->docId = base->lastDocId;
+  return ITERATOR_OK;
+}
+ 
+/* Skip to a docid, potentially reading the entry into hit, if the docId
+* matches */
+static IteratorStatus IL_SkipTo(QueryIterator *base, t_docId docId) {
+  IdListIterator *it = (IdListIterator *)base;
+  if (isEof(it) || it->offset >= it->size) {
+    setEof(it, 1);
```
BenGoldberger, line 41:
```suggestion
    setEof(it, true);
```

src/iterators/idlist_iterator.c
```diff
@@ -0,0 +1,126 @@
...
+    return ITERATOR_EOF;
+  }
+
+  if (docId > it->docIds[it->size - 1]) {
+    setEof(it, 1);
```
BenGoldberger, line 46:
```suggestion
    setEof(it, true);
```

src/iterators/idlist_iterator.c
```diff
@@ -0,0 +1,126 @@
...
+    if (docId < did) {
+      top = i - 1;
+    } else {
+      bottom = i + 1;
+    }
+  }
+  if (did < docId) did = it->docIds[++i];
+  it->offset = i + 1;
+  if (it->offset >= it->size) {
+    setEof(it, 1);
```
BenGoldberger, line 69:
```suggestion
    setEof(it, true);
```

tests/cpptests/micro-benchmarks/benchmark_idlist_iterator.cpp
```diff
@@ -0,0 +1,112 @@
+/*
+ * Copyright Redis Ltd. 2016 - present
+ * Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
+ * the Server Side Public License v1 (SSPLv1).
+ */
+
+#define MICRO_BENCHMARKS
```
BenGoldberger, line 7: why do we need to define this?

JoanFM: because we have the 2 definitions for both old and new versions, when we run benchmarks we compile and run both versions at the same time that contain duplicated structs

tests/cpptests/micro-benchmarks/benchmark_idlist_iterator.cpp
```diff
@@ -0,0 +1,112 @@
...
+      }
+    }
+    std::sort(docIds.begin(), docIds.end());
+  }
+};
+bool BM_IdListIterator::initialized = false;
+
+// Translation - exponential range from 2 to 20 (double each time), then 25, 50, 75, and 100.
+// This is the number of docIds to iterate on in each scenario
+#define DOCIDS_SCENARIOS() RangeMultiplier(2)->Range(2, 20)->DenseRange(25, 100, 25)
```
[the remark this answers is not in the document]
GuyAv46: For the `Read` benchmark, it shouldn't have any effect, so I suggest a high number so we don't perform the rewind too often (it will hurt the performance)
For skipping, I think we can have a parameter for how much to jump, not sure if it's interesting though.
Remember that the benchmark gives us the average time it takes to perform a single read/skip, so there is no need to make the number of results a parameter if we don't expect it to affect the performance of a single op

PR #6224, opened 2025-05-27
ResultProcessor Rust Scaffolding

JonasKruckenberg:

This is a first draft of `ResultProcessor`s in Rust. This PR includes the scaffolding and infrastructure as well as the `RPCounter` rewrite as a demonstration.

It still has a few rough edges, but should illustrate the approach that we settled on: We ideally want consumers to just implement a `ResultProcessor` trait on their types, however the current C implementation models the chain/stack of RPs as an intrusive singly-linked list. This means in order to remain compatible with existing C RPs Rust RPs will also need to opt into intrusive linking. This necessitates including the `result_processor::Header` type as the first struct member.

When all RPs are rewritten in Rust at some point, the fields of `result_processor::Header` and the `upstream` method impl can just be switched to not require the `unsafe extern C` pointers anymore

We believe this is the best approach out of all that we tried.

THIS PR IS NOT DONE! DO NOT MERGE!

#### Mark if applicable


src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,26 @@
...
+    // THIS MUST BE THE FIRST FIELD
+    header: Header,
+    count: usize,
+}
+
+impl ResultProcessor for Counter {
+    fn next(&mut self) -> crate::Result<Option<SearchResult>> {
+        let mut upstream = self.header.upstream().unwrap();
+
+        while let Some(_) = upstream.next()? {
```
DarthB, line 14:
```suggestion
        while (upstream.next()?).is_some()
```

src/redisearch_rs/redisearch_rs/src/result_processor/counter.rs
```diff
@@ -0,0 +1,17 @@
+use super::{result_processor_alloc, result_processor_free, result_processor_next};
+use result_processor::{ResultProcessorType, counter::Counter};
+use std::ptr::NonNull;
+
+#[unsafe(no_mangle)]
+extern "C" fn RPCounter_New() -> NonNull<result_processor::Header> {
```
chesedo, line 6: This is more of a team wide question: where do we want the `extern "C"` code to be located? Personally I would like them all to be in a single `ffi.rs` file or under an `ffi` module. This just makes them easier to find if one thinks of finding something by the function it does.

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    //     ERROR,
+    //     // Not a return code per se, but a marker signifying the end of the 'public'
+    //     // return codes. Implementations can use this for extensions.
+    //     MAX,
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+pub trait ResultProcessor {
+    fn next(&mut self) -> Result<Option<SearchResult>>;
```
DarthB, line 30:
```suggestion
    fn next(&mut self, sr: &mut SearchResult) -> Result<Option<SearchResult>>;
```

To my best knowledge the SearchResult is owned by the caller of the pipeline.

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,26 @@
...
+            self.count += 1;
+        }
```
chesedo, line 16: Since `upstream` is the previous result processor, this means `Counter` counts the number of result processors right?

src/redisearch_rs/redisearch_rs/src/result_processor/counter.rs
DarthB: This crate `redisearch_rs` is the place to put all the extern "C" code - we link it statically and only want one lib to link -> no changes to build scripts.

src/redisearch_rs/redisearch_rs/src/result_processor/mod.rs
```diff
@@ -0,0 +1,45 @@
+mod counter;
+
+use result_processor::{ResultProcessor, SearchResult};
+use std::{mem::MaybeUninit, ptr::NonNull};
+
+/// Utility function that implement the `ResultProcessorNext` FFI function
+/// for any type that implements `ResultProcessor`
+unsafe extern "C" fn result_processor_next<T: ResultProcessor>(
+    ptr: NonNull<result_processor::Header>,
+    mut out: NonNull<MaybeUninit<SearchResult>>,
```
DarthB, line 10: Probably will become `NonNull` as we cannot port everything to Rust yet. `SearchResult` has some dependencies and I will bind more than port for now.
Anyway - it's fine to get a vertical slice as soon as possible

src/redisearch_rs/redisearch_rs/src/result_processor/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+        Err(_) => {
+            // TODO convert rust error into ret code
+            return 1;
+        }
+    }
+}
+
+/// Utility function that implement the `ResultProcessorFree` FFI function
+/// MUST BE KEPT IN SYNC WITH `result_processor_alloc` below
+unsafe extern "C" fn result_processor_free<T>(ptr: NonNull<result_processor::Header>) {
```
DarthB, line 34: trait bounds

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,26 @@
+use crate::{Header, ResultProcessor, SearchResult};
+
+#[repr(C)]
+pub struct Counter {
+    // THIS MUST BE THE FIRST FIELD
+    header: Header,
```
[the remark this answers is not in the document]
DarthB: We need to have the same layout as the `ResultProcessor` struct in C:
```
typedef struct {
  ResultProcessor base;
  size_t count;
} RPCounter;
```

They use this pattern to enable interface/trait like functionality in C.

src/redisearch_rs/result_processor/src/counter.rs
DarthB: next() returns a `SearchResult` so it counts the numbers of results in the previous RP, IIRC.

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    Max,
+}
+
+pub type FFIResultProcessorNext =
+    unsafe extern "C" fn(NonNull<Header>, res: NonNull<MaybeUninit<SearchResult>>) -> i32;
+
+pub type FFIResultProcessorFree = unsafe extern "C" fn(NonNull<Header>);
+
+#[repr(C)]
+pub struct Header {
```
DarthB, line 65: That is fine for now.
However, as reminder: We'll use C as source of truth and `Bindgen` to generate the ffi code from a C header afterwards we can wrap like this:
```
#[repr(transparent)]
pub struct Header(ffi::ResultProcessor)

impl Header {
    // ...
}
```

src/redisearch_rs/redisearch_rs/Cargo.toml
```diff
@@ -20,8 +20,9 @@ test = false
 bench = false
 
 [features]
-default = []
+default = ["trie", "result_processor"]
```
DarthB, line 23: Ensure this is temporary

src/redisearch_rs/redisearch_rs/Cargo.toml
```diff
@@ -20,8 +20,9 @@ test = false
...
 trie = []
```
DarthB, line 24: Out of scope question: Why is the dependency not optional?

DarthB (changes requested):
Nice work! Added some comments mainly for upcoming work and explanations. 

Let's push this to a vertical slice to test the C-integration as PoC.

We will then create separate PRs for merging.

LukeMathWalker:
You can break the cycle by making `next` a concrete method for `Upstream` rather than implementing the trait for it. I've given it a go in https://github.com/RediSearch/RediSearch/commit/41755ed79b7e77bff83ab76c3010849ecbbb93e9

PR #6227, opened 2025-05-28
[MOD-9806] Index encoder & decoder interfaces

chesedo:

This makes Rust interfaces for the inverted index encoder and decoder. The idea is to lay the ground work to be able to port all the index encoders and decoders inside the inverted index to Rust.

#### Mark if applicable


src/inverted_index.c
```diff
@@ -433,6 +435,57 @@ IndexEncoder InvertedIndex_GetEncoder(IndexFlags flags) {
   }
 }
 
+/* Write a forward-index entry to an index writer */
+size_t InvertedIndex_WriteEntryGeneric_RS(InvertedIndex *idx, IndexEncoderRS *encoder,
```
chesedo, line 439: This is basically a copy of the function below but calls Rust binding instead

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -0,0 +1,50 @@
+use std::io::{Read, Seek, Write};
```
LukeMathWalker, line 10: Let's avoid using `impl` trait notation in argument position—it doesn't allow the caller to use the turbofish syntax to specify the (implicit) generic parameters.
Let's use explicit generics.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+    Record(RSIndexResult),
+    /// The record was filtered out and should not be returned.
+    FilteredOut,
+    /// There is nothing more left on the reader to decode.
+    EndOfStream,
+}
+
+/// Decoder to read records from an index
+pub trait Decoder {
+    /// Decode the next record from the reader. The offset is the base value for any delta being
```
LukeMathWalker: It's not clear to me, reading the docs, what offset actually means here.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+
+/// Decoder to read records from an index
+pub trait Decoder {
+    /// Decode the next record from the reader. If any delta values are decoded, then they should
+    /// add to the `base` document ID to get the actual document ID.
+    fn decode<R: Read>(&self, reader: R, base: t_docId) -> std::io::Result<DecoderResult>;
+
+    /// Like `[Decoder::decode]`, but seeks to a specific document ID and returns it.
+    ///
+    /// Returns `None` if there is no record greater than or equal to the target document ID.
```
```suggestion
    /// Like `[Decoder::decode]`, but it skips all entries whose document ID is lower than `target`.
    ///
    /// Returns `None` if no record has a document ID greater than or equal to `target`.
```

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn seek<R: Read + Seek + Copy>(
```
LukeMathWalker: `Copy` is a strange bound for a reader—what's the rationale?

LukeMathWalker: If it's due to the need to invoke `decode` multiple times, we should change `decode`'s signature to take `&mut R`, rather than `R`. Probably the right call anyway.

chesedo: Yeah, it's for the loop... Let me change it to `&mut R`

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn encode<W: Write + Seek>(
+        writer: W,
+        delta: t_docId,
+        record: &RSIndexResult,
+    ) -> std::io::Result<usize>;
+}
+
+pub enum DecoderResult {
+    /// The record was successfully decoded.
+    Record(RSIndexResult),
```
LukeMathWalker, line 46: Are we 100% positive that no data is borrowed from the index? Will it all be owned by the caller?

chesedo: This is the only current decoder call where data can potentially be owned by the index

https://github.com/RediSearch/RediSearch/blob/e09f5c1d7871367ac16bf460e10301987aafc172/src/inverted_index.c#L930-L933

However, I'm thinking we should be able to change this to rather just assign the returned value to the index (if needed even). This also assumes that the decoder calls are not used to merge multiple decoded values into the same record (ie different decoder calls decode different parts of the same record). From my understanding this assumption is same since only a single decoder exists for an index so it will always decode exactly the same parts of a record.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+            match self.decode(reader, base)? {
+                DecoderResult::Record(record) if record.docId >= target => {
+                    return Ok(Some(record));
+                }
+                DecoderResult::Record(_) | DecoderResult::FilteredOut => continue,
+                DecoderResult::EndOfStream => return Ok(None),
+            }
+        }
+    }
+}
```
LukeMathWalker: Two questions:
- Do we expect (some) decoders to provide a specialized implementation of this method?
- Assuming the answer to the previous question is yes, is that why we require `Seek`?

chesedo: Yes, two decoders use this.

# 1. The frequency offset decoder

https://github.com/RediSearch/RediSearch/blob/e09f5c1d7871367ac16bf460e10301987aafc172/src/inverted_index.c#L568-L594

This might need a small update from the default impl

# 2. The raw ids decoder

https://github.com/RediSearch/RediSearch/blob/e09f5c1d7871367ac16bf460e10301987aafc172/src/inverted_index.c#L724-L770

It will need `Seek` for the binary search

LukeMathWalker (approved):

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: This is the only place it is called.

https://github.com/RediSearch/RediSearch/blob/e09f5c1d7871367ac16bf460e10301987aafc172/src/inverted_index.c#L1100-L1119

The default impl should be the same as this fallback (`else` branch) when the seeker is not set.

src/redisearch_rs/ffi/build.rs
```diff
@@ -33,6 +33,7 @@ fn main() {
     let deps = root.join("deps");
     let bindings = bindgen::Builder::default()
```
JoanFM: maybe we could create a list of binding interfaces outside for better readability?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+
+pub use ffi::{RSIndexResult, t_docId};
+
+/// Encoder to write a record into an index
+pub trait Encoder {
+    /// Write the record to the writer and return the number of bytes written. The delta is the
+    /// pre-computed difference between the current document ID and the last document ID written.
+    fn encode<W: Write + Seek>(
+        writer: W,
+        delta: t_docId,
```
JoanFM: Isn't this interface a little opinionated? Shouldn't we call `docId` or `doc`? And then let the implementation decide wether it is the delta or not that is actually used?

JoanFM: I think it would be nice to perhaps decouple the Encoding part with the Delta Computation or call it DeltaEncoder if it is an overkill

src/redisearch_rs/ffi/build.rs
chesedo: Hmm, what do you mean by "outside"?

JoanFM: I mean have a:

`let include_api_headers = ['....']`

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: From what I can tell, this is always called with a `docId` (via `InvertedIndex_WriteEntryGeneric`). And since this is taking the delta, that means it is already decoupled from the delta computation right?

JoanFM: I am not so familiar with the current code, but it is forced here that u need to provide a delta. Looking at the code is the way is currently handled, so it may be a good option, and any other may be an overkill

chesedo: No worries. So this is where all the current encoders are defined

https://github.com/RediSearch/RediSearch/blob/74451de0aac2c97bc6b684ae9b6b49c2fe00fb96/src/inverted_index.c#L149-L229

They all use delta. And it also makes sense since the inverted index will need to map some term to its original document ID (and using deltas causes it to store less data in memory).

JoanFM: maybe having a `DocIdDelta`struct that t_docID can deref to and from automatically? It would clarify a little more the signature IMO. Not sure if the extra complexity could be worth though

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: To answer the question as to whether `RSIndexResult` will need lifetimes when written in Rust, the answer is no.

This is the C definition of `RSIndexResult`

https://github.com/RediSearch/RediSearch/blob/74451de0aac2c97bc6b684ae9b6b49c2fe00fb96/src/redisearch.h#L336-L374

The only pointers are `dmd` and `metrics`. However none of the encoders will these fields. So when we get to converting this to Rust we can make two new types:
1. Which contains the actual encoding / decoding data where everything is owned
2. Another type stored by the indexer. It will contain the base type in 1 with other metadata fields like metrics that might have lifetimes.

But this means whatever the encoders / decoders works with won't need lifetimes.

src/redisearch_rs/ffi/build.rs
LukeMathWalker: https://github.com/RediSearch/RediSearch/pull/6244

src/redisearch_rs/inverted_index/src/lib.rs
LukeMathWalker: ^ I think that'd be a nice improvement, to clarify what's going on.

chesedo: Oh, you mean because `delta` is technically not a doc ID :thinking: . That touches on a redisearch team convention that we won't want to address with the Rust port.

chesedo: Wait @LukeMathWalker , you would want a delta struct? I agree it will help, but not sure that this is the right time to do it?

LukeMathWalker: Using strongly-typed wrappers is definitely in scope IMO, if it makes the code more misuse resistant.
As long as we do the right conversions at the FFI boundary, I don't see an issue. Especially considering that the delta computation is internal to the II and that'll be entirely in Rust soon-ish.

chesedo: Okay, but then I won't wrap `t_docID`. I'll wrap `usize`. Then the FFI boundary can take care of the conversion. Any issues with that?

chesedo: But that also brings up another question: where do I put the wrapping struct?

LukeMathWalker: > Okay, but then I won't wrap `t_docID`. I'll wrap `usize`. Then the FFI boundary can take care of the conversion. Any issues with that?

No issues on my side, and I think that's the correct way to go about it.

> But that also brings up another question: where do I put the wrapping struct?

In the Rust module for now.

chesedo: > In the Rust module for now.

Do you mean `inverted_index`? Then we can just move it to some common crate later.

JoanFM: I would also go as far as naming `DeltaEncoder`

LukeMathWalker: > Do you mean inverted_index? Then we can just move it to some common crate later.

Yup.

> I would also go as far as naming DeltaEncoder

I think that'd be misleading, because we are encoding multiple data types (not just doc ids) and those are not delta-encoded.

chesedo: Done

LukeMathWalker: We want a newtype, not a type alias—we don't gain any safety from an alias imo.

chesedo: Fair, updated

LukeMathWalker (approved):

JoanFM (approved):
Actually I think it does not affect, but could you double check @chesedo ?

chesedo:
> Actually I think it does not affect, but could you double check @chesedo ?

No, it won't affect. So it will be safe to merge this one

LukeMathWalker (dismissed):

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use std::io::{Read, Seek, Write};
+
+pub use ffi::{RSIndexResult, t_docId};
+
+/// A delta is the difference between document IDs. It is mostly used to save space in the index
+/// because document IDs are usually sequential and the difference between them is small. With the
+/// help of encoding, we can store the difference (delta) efficiently instead of the full document
```
zeenix: nit: which encoding? Is there a missing `this` here? :thinking:

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// Write the record to the writer and return the number of bytes written. The delta is the
+    /// pre-computed difference between the current document ID and the last document ID written.
+    fn encode<W: Write + Seek>(
+        writer: W,
+        delta: Delta,
+        record: &RSIndexResult,
+    ) -> std::io::Result<usize>;
+}
+
+pub enum DecoderResult {
```
zeenix, line 44: why a `Result` enum and not using a new `Error` enum? For nesting, you could either have an `io::Error` variant or have nested `io::Result<Result<RSIndex, DecoderError>>`.

zeenix (changes requested):

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Only some of the encoders use the delta. Some like the `RawDocIDOnly` just don't. Hence why the next line has "can" and implies "can optionally". I'll update the comment.

zeenix: Thanks for explaining. Perhaps it should say "help of encoders"?

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: I don't see the filtered cases and end-of-stream as error cases. Decoding happened successfully with these and so we are returning that success. Granted end-of-stream can technically be replaced with an `Option` like an iter's `.next()`. But then the return type becomes `io::Result<Option<DecoderResult>>`. I prefer to avoid this nesting, but it can work.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+*/
+
+use std::io::{Read, Seek, Write};
+
+pub use ffi::{RSIndexResult, t_docId};
+
+/// A delta is the difference between document IDs. It is mostly used to save space in the index
+/// because document IDs are usually sequential and the difference between them is small. With the
+/// help of encoding, we can store the difference (delta) efficiently instead of the full document
+/// ID.
```
chesedo:
```suggestion
/// A delta is the difference between document IDs. It is mostly used to save space in the index
/// because document IDs are usually sequential and the difference between them are small. With the
/// help of encoding, we can optionally store the difference (delta) efficiently instead of the full document
/// ID.
```

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: I updated the comment. In terms of English grammar I would not update it to "encoders"

src/redisearch_rs/inverted_index/src/lib.rs
zeenix: > I don't see the filtered cases and end-of-stream as error cases.

Right, the name confused me before my second cup of :coffee: :laughing: 

> Granted end-of-stream can technically be replaced with an `Option` like an iter's `.next()`. But then the return type becomes `io::Result<Option<DecoderResult>>`. I prefer to avoid this nesting, but it can work.

Does the callers really need to differentiate between filtered-out and EOS? If not, this could just be an `io::Result<Option<RSIndexResult>>`.

zeenix: > Does the callers really need to differentiate between filtered-out and EOS? 

As in, if they need to take different paths based on which of these 2 variants was returned?

chesedo: Yes, for filtered out the callers might call it again to try and get the next one. For EOS the callers should know that calling it again won't yield any different results.

zeenix: > For EOS the callers should know that calling it again won't yield any different results.

Right.

zeenix (approved):

PR #6230, opened 2025-05-28
Add micro-benchmarks to CI - [MOD - 8849]

JoanFM:

Adds "micro-benchmark" runs to CI:

- Create one benchmark binary for each of the micro benchmark test.
- Add option to run the micro benchmarks in `Makefile` and `build.sh`
- Add `flow-micro-benchmark.yml` and  `flow-micro-benchmark-runner.yml` to run the actions in CI and collect its results.
- Separate the logic between "Rust Micro Benchmarks" and "Micro Benchmarks"

The "diff" is weird because I renamed "flow-microbenchmark" to "flow-rust-micro-benchmark" but then I created it again to add the regular microbenchmarks.

#### Mark if applicable


JoanFM:
@dor-forer here you may have some hints for me

.github/workflows/flow-micro-benchmarks-runner.yml
```diff
@@ -0,0 +1,175 @@
...
+      - name: Print runner info
+        run: |
+          printf "Runner lscpu:\n$(lscpu)\n"
+          printf "Runner lsmem:\n$(lsmem)\n"
+          printf "Runner nproc:\n$(nproc)\n"
+          printf "Runner uname:\n$(uname -a)\n"
+          printf "Runner arch:\n$(arch)\n"
+      - name: Install benchmark dependencies
+        run: |
+          cd .install
```
JoanFM, line 73:
```suggestion
          export HOME=/home/ubuntu
          cd .install
```

.github/workflows/flow-micro-benchmarks-runner.yml
```diff
@@ -0,0 +1,176 @@
...
+          sudo apt install python3-pip -y
+          pip3 install --upgrade pip PyYAML setuptools redisbench-admin
+        env:
+          DEBIAN_FRONTEND: noninteractive
+      - name: Run Micro Benchmark
+        env:
+          ARCH: ${{ inputs.architecture }}
+        timeout-minutes: 300
+        run: |
+          ./build.sh run_microbenchmarks
```
JoanFM, line 88:
```suggestion
          export HOME=/home/ubuntu
          ./build.sh run_microbenchmarks
```

.github/workflows/flow-micro-benchmarks-runner.yml
```diff
@@ -0,0 +1,177 @@
...
+          sudo apt install python3-pip -y
+          pip3 install --upgrade pip PyYAML setuptools redisbench-admin
+        env:
+          DEBIAN_FRONTEND: noninteractive
+      - name: Run Micro Benchmark
+        env:
+          ARCH: ${{ inputs.architecture }}
+        timeout-minutes: 300
+        run: |
+          export HOME=/home/ubuntu
```
JoanFM, line 86:
```suggestion
          export HOME=/home/ubuntu
          export PATH="$HOME/.cargo/bin:$PATH"
```

build.sh
```diff
@@ -640,6 +644,49 @@ run_tests() {
...
+#-----------------------------------------------------------------------------
+run_micro_benchmarks() {
+  if [[ "$RUN_MICRO_BENCHMARKS" != "1" ]]; then
+    return 0
+  fi
+
+  echo "Running micro-benchmarks..."
+
+  # Update the binary directory path for ARM architectures
+  UPDATED_BINDIR=$(echo "$BINDIR" | sed 's/arm64v8/aarch64/')
```
GuyAv46: Why do we need to update the path? Can't we use the one we generate for the binaries?

build.sh
```diff
@@ -640,6 +644,49 @@ run_tests() {
...
+
+  # Check if micro-benchmarks directory exists
+  MICRO_BENCH_DIR="$UPDATED_BINDIR/micro-benchmarks"
+  if [[ ! -d "$MICRO_BENCH_DIR" ]]; then
+    echo "Error: Micro-benchmarks directory not found at $MICRO_BENCH_DIR"
+    echo "Make sure to build with BUILD_TESTS=1"
```
GuyAv46: Since `RUN_MICRO_BENCHMARKS=1` sets `BUILD_TESTS=1`, do we need this check?

build.sh
```diff
@@ -640,6 +644,49 @@ run_tests() {
...
+    HAS_FAILURES=1
+    return 1
+  fi
+
+  # Run each benchmark executable
+  echo "Running benchmarks from $MICRO_BENCH_DIR"
+  cd "$MICRO_BENCH_DIR" || return 1
```
GuyAv46: Can the `cd` ever fail, if we already validated `MICRO_BENCH_DIR` is a valid directory?

build.sh
```diff
@@ -640,6 +644,49 @@ run_tests() {
...
+
+  for benchmark in benchmark_*; do
+    if [[ -x "$benchmark" ]]; then
+      benchmark_name=$(basename "$benchmark" | sed 's/^benchmark_//')
```
GuyAv46: Kind of an overkill. `benchmark` should be a base name only, and there is a simpler way to remove a prefix without `sed`
```suggestion
      benchmark_name=${benchmark#benchmark_}
```

build.sh
```diff
@@ -640,6 +644,49 @@ run_tests() {
...
+      echo "Running benchmark_$benchmark_name..."
```
GuyAv46:
```suggestion
      echo "Running $benchmark..."
```

build.sh
```diff
@@ -60,6 +61,9 @@ parse_arguments() {
       RUN_RUST_TESTS|run_rust_tests)
         RUN_RUST_TESTS=1
         ;;
+      RUN_MICRO_BENCHMARKS|run_micro_benchmarks|RUN_MICROBENCHMARKS|run_microbenchmarks)
```
GuyAv46, line 64: nice

Makefile
```diff
@@ -698,3 +699,23 @@ sanbox:
 	@docker run -it -v $(PWD):/search -w /search --cap-add=SYS_PTRACE --security-opt seccomp=unconfined $(SANBOX_ARGS) redisfab/clang:16-$(ARCH)-bullseye bash
 
 .PHONY: box sanbox
+
+UPDATED_BINDIR:=$(subst arm64v8,aarch64,$(BINDIR))
+
+micro-benchmarks:
```
GuyAv46: Do we need this definition? We are going to replace this `Makefile` completely soon and make it call the `./build.sh` script. I'd avoid changing this file and the logic duplication

.github/workflows/benchmark-trigger.yml
```diff
@@ -23,6 +23,22 @@ jobs:
...
 
+  rust-micro-benchmarks:
+    name: Trigger Rust Micro Benchmarks
+    if: >
+      (
+        github.event.action == 'labeled' &&
+        github.event.label.name == 'action:run-micro-benchmark'
+      ) || (
+        contains(fromJson('["opened", "reopened", "synchronize"]'), github.event.action) &&
+        contains(github.event.pull_request.labels.*.name, 'action:run-micro-benchmark')
```
GuyAv46, line 34: Should we keep the same label for both micro benchmarks? Or should we make this job be triggered by `action:run-rust-micro-benchmark`?

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -1,64 +1,100 @@
-name: Run a Micro Benchmark Flow
```
GuyAv46: We'd better give it a nice name. It is displayed in the GitHub UI

.github/workflows/flow-rust-micro-benchmarks.yml
GuyAv46, line 1: This file was just renamed, right?

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -1,64 +1,100 @@
...
+        required: false
+        default: 'all'
+        description: 'Run only on specific architecture'
+  workflow_dispatch:
+    inputs:
+      architecture:
+          type: choice
+          options:
+            - all
+            - arm64
```
GuyAv46: We should align the variant with what we have in other workflows
```suggestion
            - aarch64
```

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -1,64 +1,100 @@
...
+        id: set-matrix
         run: |
-          if [ -d bin/redisearch_rs/criterion ]; then
-            echo "baseline_exists=true" >> $GITHUB_OUTPUT
+          # Define the full matrix as a JSON string
+          FULL_MATRIX='
+          {
+            "include": [
+              {
+                "architecture": "arm64",
```
GuyAv46: We should align the variant with what we have in other workflows
```suggestion
                "architecture": "aarch64",
```

.github/workflows/flow-micro-benchmarks.yml
```diff
@@ -1,64 +1,100 @@
...
+    uses: ./.github/workflows/flow-micro-benchmarks-runner.yml
+    secrets: inherit
+    strategy:
+      matrix: ${{ fromJson(needs.prepare_runner_configurations.outputs.matrix) }}
+    with:
+      architecture: ${{ matrix.architecture }}
+      instance-type: ${{ matrix.instance-type }}
+      ami-id: ${{ matrix.ami-id }}
+
+  compare_micro_benchmarks:
```
GuyAv46, line 75: Will this job comment on the PR? Where do we see the comparison? Does it also upload the results somewhere, like we have in VecSim?

.github/workflows/flow-micro-benchmarks-runner.yml
```diff
@@ -0,0 +1,180 @@
...
+env:
+  TAGS: | # Make sure there is no trailing comma!
+    [
+      {"Key": "Name",         "Value": "redisearch-ci-runner"},
+      {"Key": "Environment",  "Value": "CI"},
+      {"Key": "Run ID",       "Value": "${{ github.run_id }}"},
+      {"Key": "PR",           "Value": "${{ github.event.number }}"},
+      {"Key": "Owner",        "Value": "${{ github.actor }}"},
+      {"Key": "Project",      "Value": "${{ github.repository }}"}
+    ]
```
GuyAv46: Just so it's easy to distinguish between machines we launch on different occasions, even though I think this is the only place we have it in search today, let's add an indication its a micro benchmark instance

.github/workflows/benchmark-trigger.yml
JoanFM: I am open to any suggestion, we can split now that they have so much different purpose

GuyAv46:
We should also set the Grafana panels, right? The link in the comparison report links to the "end-to-end" benchmarks

.github/workflows/flow-micro-benchmarks.yml
JoanFM: The upload happens inside the ``runner``.

This job will only do the PR comment

.github/workflows/flow-rust-micro-benchmarks.yml
JoanFM: yes

build.sh
JoanFM: Not sure why but it works like this in my mac

build.sh
JoanFM: fair enough

Makefile
JoanFM: ah I thought it was the other way around. I can remove yes

JoanFM:
> We should also set the Grafana panels, right? The link in the comparison report links to the "end-to-end" benchmarks

Makes sense, not sure if it is part of the redis-compare call. @fcostaoliveira any idea?

build.sh
GuyAv46: We should generate `BINDIR` once and in a single place. Changing it in different locations/scripts makes it hard to track and maintain.
If there is an issue in the way we first calculate it, we should fix it there instead

.github/workflows/benchmark-trigger.yml
GuyAv46: I think let's keep it this way. We can add later more fine-grained labels for only specific jobs, like we have on VecSim (e.g., Rust only, iterators only...)

.github/workflows/flow-micro-benchmarks-runner.yml
```diff
@@ -0,0 +1,180 @@
...
+          # Create full variant string for the build directory
+          FULL_VARIANT="${OS_NAME}-${ARCH}-${FLAVOR}"
+
+          # Set coordinator type (assuming OSS)
+          COORD="oss"
+          OUTDIR="search-community"
+
+          # Set the final BINDIR
+          BINDIR="bin/${FULL_VARIANT}/${OUTDIR}"
+
```
dor-forer, line 117: We have the get-platform script under sbin folder if you want to use it

.github/workflows/flow-micro-benchmarks-runner.yml
```diff
@@ -0,0 +1,180 @@
...
+          JSON_FILES=$(find $UPDATED_BINDIR/micro-benchmarks -name "*_results.json")
+
+          if [ -z "$JSON_FILES" ]; then
+            echo "No benchmark result files found in $UPDATED_BINDIR/micro-benchmarks"
+            exit 1
+          fi
+
+          # Find all JSON result files
+          JSON_FILES=$(find $BINDIR -name "*_results.json")
+
```
dor-forer: Maybe give the second JSON_FILES a different name to avoid confusion

JoanFM: I am not sure why there are 2 JSON_FILES, removing these second ones

.github/workflows/flow-micro-benchmarks-runner.yml
JoanFM: I'd rather keep as is now

dor-forer (approved):

JoanFM:
> > We should also set the Grafana panels, right? The link in the comparison report links to the "end-to-end" benchmarks
> 
> Makes sense, not sure if it is part of the redis-compare call. @fcostaoliveira any idea?

@GuyAv46 parameter has been added to the compare

PR #6236, opened 2025-05-29
IDList Iterator follow-up comments

JoanFM:

Continuation of #6221 to handle new comments

1. MOD-...
2. #...

1. ...

#### Mark if applicable


src/iterators/idlist_iterator.c
```diff
@@ -23,6 +23,7 @@ static size_t IL_NumEstimated(QueryIterator *base) {
 *  Returns ITERATOR_EOF if at the end */
 static IteratorStatus IL_Read(QueryIterator *base) {
   IdListIterator *it = (IdListIterator *)base;
+  RS_ASSERT(it->offset < it->size);
```
GuyAv46, line 26: This assertion is invalid; With the current implementation, we may have `it->offset >= it->size` and the following check will catch it

tests/cpptests/micro-benchmarks/benchmark_idlist_iterator.cpp
```diff
@@ -39,64 +41,65 @@ class BM_IdListIterator : public benchmark::Fixture {
...
 };
-bool BM_IdListIterator::initialized = false;
 
-// Translation - exponential range from 2 to 20 (double each time), then 25, 50, 75, and 100.
-// This is the number of docIds to iterate on in each scenario
-#define DOCIDS_SCENARIOS() RangeMultiplier(2)->Range(2, 20)->DenseRange(25, 100, 25)
+template <typename IteratorType>
+bool BM_IdListIterator<IteratorType>::initialized = false;
+
+#define DOCIDS_SCENARIOS() Arg(10)->Arg(100)->Arg(1000)->Arg(100000)
```
GuyAv46, line 61: For the Read benchmark, it shouldn't have any effect, so I suggest a high number so we don't perform the rewind too often (it will hurt the performance)
For skipping, I think we can have a parameter for how much to jump, not sure if it's interesting though.
Remember that the benchmark gives us the average time it takes to perform a single read/skip, so there is no need to make the number of results a parameter if we don't expect it to affect the performance of a single op

tests/cpptests/micro-benchmarks/benchmark_idlist_iterator.cpp
```diff
@@ -58,7 +58,9 @@ class BM_IdListIterator : public benchmark::Fixture {
 template <typename IteratorType>
 bool BM_IdListIterator<IteratorType>::initialized = false;
 
-#define DOCIDS_SCENARIOS() Arg(10)->Arg(100)->Arg(1000)->Arg(100000)
+#define READ_DOCIDS_SCENARIOS() Arg(100)
+#define SKIPTO_DOCIDS_SCENARIOS() Arg(100)->Arg(1000)
+
```
GuyAv46, line 63: I think there was a misunderstanding
1. Let's remove these definitions completely
2. Set `numDocuments` in `Setup` to a high number so the benchmark has enough data to test

src/iterators/idlist_iterator.c
```diff
@@ -57,6 +56,9 @@ static IteratorStatus IL_SkipTo(QueryIterator *base, t_docId docId) {
       break;
     }
     if (docId < did) {
+      if (i == 0) {
+         break;
+      }
```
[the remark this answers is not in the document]
JoanFM: Adding it, the benchmark looked better

src/iterators/idlist_iterator.c
```diff
@@ -125,4 +119,81 @@ QueryIterator *IT_V2(NewIdListIterator) (t_docId *ids, t_offset num, double weig
...
+  it->docIds = docIds;
+  it->size = num_results;
+  it->offset = 0;
+
+  ret->lastDocId = 0;
+  setEof(ret, false);
+  ret->type = METRIC_ITERATOR;
+  ret->current = NewMetricResult();
+  // If we interested in yielding score
+  if (yields_metric) {
```
GuyAv46: This option is redundant.
In my opinion, we should have only one of the following options:
1. Have a single iterator type that gets a metric array on construction, and then choose between `MR_Read` and `IL_Read` (and so on) based on whether a valid array was passed or `NULL`
2. Have two explicit types (as you implemented), but the metrics iterator does not support a missing metrics array (the caller should use the id-list iterator instead)

src/iterators/idlist_iterator.c
```diff
@@ -125,4 +119,81 @@ QueryIterator *IT_V2(NewIdListIterator) (t_docId *ids, t_offset num, double weig
...
+
+  IndexResult_Free(base->current);
+
+  if (it->docIds) {
+    rm_free(it->docIds);
+  }
+
+  if (mi->metricList) {
+    rm_free(mi->metricList);
+  }
```
GuyAv46: `free` (and `rm_free`) can handle `NULL` input. Consider removing the check (we also don't expect `NULL` values here)

src/iterators/idlist_iterator.c
```diff
@@ -125,4 +119,81 @@ QueryIterator *IT_V2(NewIdListIterator) (t_docId *ids, t_offset num, double weig
...
+    SetYield(base);
+  }
+  return rc;
+}
+
+static void MR_Free(QueryIterator *self) {
+  MetricIterator *mi = (MetricIterator *)self;
+  if (mi == NULL) {
+    return;
+  }
```
GuyAv46: I think we can skip this check (we can only reach this static function from the API, which means we have something to free)

src/iterators/idlist_iterator.c
```diff
@@ -125,4 +119,81 @@ QueryIterator *IT_V2(NewIdListIterator) (t_docId *ids, t_offset num, double weig
...
+  return rc;
+}
+
+static IteratorStatus MR_SkipTo(QueryIterator *base, t_docId docId) {
+  MetricIterator *mr = (MetricIterator *)base;
+  IdListIterator *it = &mr->base;
+  int rc = IL_SkipTo(base, docId);
+  if (ITERATOR_OK == rc || ITERATOR_NOTFOUND == rc) {
+    base->current->num.value = mr->metricList[it->offset - 1];
```
GuyAv46: This line can be part of `SetYield`, no?

src/iterators/idlist_iterator.c
```diff
@@ -125,4 +119,81 @@ QueryIterator *IT_V2(NewIdListIterator) (t_docId *ids, t_offset num, double weig
...
+  ResultMetrics_Reset(mr->current);
+  ResultMetrics_Add(mr->current, NULL, RS_NumVal(mr->current->num.value));
+}
+
+static IteratorStatus MR_Read(QueryIterator *base) {
+  MetricIterator *mr = (MetricIterator *)base;
+  IdListIterator *it = &mr->base;
+  IteratorStatus rc = IL_Read(base);
+  if (ITERATOR_OK == rc) {
+    base->current->num.value = mr->metricList[it->offset - 1];
```
GuyAv46: This line can be part of `SetYield`, no?

src/iterators/idlist_iterator.c
JoanFM: I changed the code such that if yields metric is not returned, then IDListIterator is returned

src/iterators/idlist_iterator.c
```diff
@@ -73,12 +73,10 @@ static IteratorStatus IL_SkipTo(QueryIterator *base, t_docId docId) {
...
   IndexResult_Free(self->current);
-  if (it->docIds) {
-      rm_free(it->docIds);
-  }
+  rm_free(it->docIds);
   rm_free(self);
 }
- 
+
 static int cmp_docids(const void *p1, const void *p2) {
```
GuyAv46: Fix the indentation in this function

src/iterators/idlist_iterator.c
```diff
@@ -142,58 +145,46 @@ static IteratorStatus MR_SkipTo(QueryIterator *base, t_docId docId) {
...
-  if (mi->metricList) {
-    rm_free(mi->metricList);
-  }
-
+  rm_free(it->docIds);
+  rm_free(mi->metricList);
   rm_free(mi);
 }
 
 QueryIterator *IT_V2(NewMetricIterator)(t_docId *docIds, double *metric_list, size_t num_results, Metric metric_type, bool yields_metric) {
```
GuyAv46: We said we should align the API here regarding array types (should be a simple allocation and not `array_t`, and also align on the ownership of the array (do we need to duplicate it or can we take ownership over it?)

src/iterators/idlist_iterator.c
```diff
@@ -142,58 +145,46 @@ static IteratorStatus MR_SkipTo(QueryIterator *base, t_docId docId) {
...
     ret->SkipTo = MR_SkipTo;
+    ret->Rewind = IL_Rewind;
+    ret->Free = MR_Free;
+    ret->NumEstimated = IL_NumEstimated;
   } else {
-    ret->Read = IL_Read;
-    ret->SkipTo = IL_SkipTo;
+    IdListIterator *it = rm_new(IdListIterator);
+    ret = &it->base;
+    IL_Init(it, docIds, num_results, 1.0);
```
GuyAv46: In this case (`yields_metric == false`), we currently leak the metrics array

src/iterators/idlist_iterator.c
GuyAv46: Do we really need the `yields_metric` option? Can we just use the `NewIdListIterator` if we don't need the metrics?
I see that previously, when `yields_metric == false`, we still set the result metric value. I think it wasn't ever used, so it makes sense to stop doing it. Unless something stops working

JoanFM: I am respecting the interface to ease the transition

JoanFM: I thought it be easier not to change the API so that the transition from old iterators to new iterators is smoother.

src/iterators/idlist_iterator.c
```diff
@@ -70,62 +69,112 @@ static IteratorStatus IL_SkipTo(QueryIterator *base, t_docId docId) {
...
   ret->Rewind = IL_Rewind;
+  return ret;
+}
 
+static void SetYield(QueryIterator *base, double value) {
+  base->current->num.value = value;
+  if (!base->current->metrics) {
+    ResultMetrics_Add(base->current, NULL, RS_NumVal(value));
+  } else {
+    ResultMetrics_UpdateDouble(base->current, NULL, value);
```
GuyAv46: Sadly, we can't do this optimization this way.
The metric and `RSValue` may be shared at this point (after we previously returned them) is the caller took a reference to the result. This is why we must allocate a new RSValue every time.
We can have an optimization that checks if we have the only reference to this value, and then reuse it

src/iterators/idlist_iterator.c
```diff
@@ -70,62 +69,109 @@ static IteratorStatus IL_SkipTo(QueryIterator *base, t_docId docId) {
...
 }
- 
+
 static int cmp_docids(const void *p1, const void *p2) {
   const t_docId *d1 = p1, *d2 = p2;
-
-  return (int)(*d1 - *d2);
+  if (*d1 < *d2) return -1;
+  if (*d1 > *d2) return 1;
+  return 0;
```
GuyAv46: Dead code now. Can be removed

src/iterators/idlist_iterator.c
```diff
@@ -70,62 +69,109 @@ static IteratorStatus IL_SkipTo(QueryIterator *base, t_docId docId) {
...
   ret->Read = IL_Read;
   ret->SkipTo = IL_SkipTo;
   ret->Rewind = IL_Rewind;
+  return ret;
+}
 
+static void SetYield(QueryIterator *base, double value) {
+  base->current->num.value = value;
+  ResultMetrics_Reset(base->current);
+  ResultMetrics_Add(base->current, NULL, RS_NumVal(value));
```
GuyAv46: The `RLookupKey*` is mandatory here. We removed it from the base `QueryIterator`, so we should add it to the metric iterator. It is set after the construction, so in this PR, we should just set it to `NULL` in the constructor and use it here.

tests/cpptests/micro-benchmarks/benchmark_idlist_iterator.cpp
```diff
@@ -11,103 +11,105 @@
 
 #include <random>
 #include <vector>
+#include <iostream>
```
GuyAv46: leftovers?

GuyAv46 (dismissed):

src/iterators/idlist_iterator.c
```diff
@@ -9,59 +9,58 @@
...
-  if (docId > it->docIds[it->size - 1]) {
-    setEof(it, true);
+  if (it->offset >= it->size || docId > it->docIds[it->size - 1]) {
+    setEof(base, true);
     return ITERATOR_EOF;
   }
 
-  int64_t top = it->size - 1, bottom = it->offset;
-  int64_t i;
+  t_offset top = MIN(it->size, it->offset + (docId - base->lastDocId) + 1);
```
GuyAv46: I thought about it a bit more, and we don't need the +1 here. We can add a test with a fully dense list and show that we can skip between any pair of IDs.
We should also comment about this non-trivial limit
```suggestion
  t_offset top = MIN(it->size, it->offset + (docId - base->lastDocId));
```

src/iterators/idlist_iterator.c
```diff
@@ -9,59 +9,58 @@
...
-  it->base.current->docId = base->lastDocId;
+  base->current->docId = base->lastDocId;
   return ITERATOR_OK;
 }
- 
+
 /* Skip to a docid, potentially reading the entry into hit, if the docId
 * matches */
 static IteratorStatus IL_SkipTo(QueryIterator *base, t_docId docId) {
   IdListIterator *it = (IdListIterator *)base;
```
GuyAv46, line 42: Let's add an assertion `RS_ASSERT(base->lastDocId < docId)`

tests/cpptests/test_cpp_iterator_idlist.cpp
```diff
@@ -20,7 +20,14 @@ class IDListIteratorCommonTest : public ::testing::TestWithParam<std::vector<t_d
 
   void SetUp() override {
     docIds = GetParam();
-    iterator_base = IT_V2(NewIdListIterator)(docIds.data(), docIds.size(), 1.0);
+    // do a copy of the docIds vector and sort them before passing them to the iterator
+    std::sort(docIds.begin(), docIds.end(), cmp_docids);
```
GuyAv46, line 28: Each test sorts `docIds` again at the moment. Let's remove the sorting from the tests.

tests/cpptests/micro-benchmarks/benchmark_metric_iterator.cpp
```diff
@@ -0,0 +1,191 @@
+#define MICRO_BENCHMARKS
```
GuyAv46:
```suggestion
/*
 * Copyright (c) 2006-Present, Redis Ltd.
 * All rights reserved.
 *
 * Licensed under your choice of the Redis Source Available License 2.0
 * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
 * GNU Affero General Public License v3 (AGPLv3).
*/
```

tests/cpptests/test_cpp_iterator_idlist.cpp
GuyAv46, line 1: Fix license header

tests/cpptests/test_cpp_iterator_metric.cpp
```diff
@@ -0,0 +1,255 @@
+/*
+* Copyright Redis Ltd. 2016 - present
+* Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
+* the Server Side Public License v1 (SSPLv1).
+*/
```
GuyAv46, line 5:
```suggestion
/*
 * Copyright (c) 2006-Present, Redis Ltd.
 * All rights reserved.
 *
 * Licensed under your choice of the Redis Source Available License 2.0
 * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
 * GNU Affero General Public License v3 (AGPLv3).
*/
```

src/iterators/idlist_iterator.c
```diff
@@ -9,59 +9,58 @@
...
   }
 
-  int64_t top = it->size - 1, bottom = it->offset;
-  int64_t i;
+  t_offset top = MIN(it->size, it->offset + (docId - base->lastDocId) + 1);
+  t_offset bottom = it->offset;
+  t_offset i = 0;
   t_docId did;
-  while (bottom <= top) {
+  while (bottom < top) {
```
GuyAv46: We can have a `do..while` here because this condition is always true at first

tests/cpptests/test_cpp_iterator_idlist.cpp
JoanFM: why would we? if it is a prerequisite of the Iterator? Wdym?

GuyAv46: I meant keep the sorting in the `Setup` only, we don't need to sort the sorted list again in every test

src/iterators/idlist_iterator.c
```diff
@@ -9,123 +9,167 @@
...
 
+  ret->lastDocId = 0;
+  setEof(ret, false);
+  ret->type = METRIC_ITERATOR;
+  ret->current = NewMetricResult();
+  ret->Read = MR_Read;
+  ret->SkipTo = MR_SkipTo;
+  ret->Rewind = IL_Rewind;
+  ret->Free = MR_Free;
+  ret->NumEstimated = IL_NumEstimated;
```
BenGoldberger, line 173: can the use of idList function be a potential problem? (those function do cast the pointer to idlistIterator)

JoanFM: The base of the Metric is an IDListIterator. And the base of IDList is a QueryIterator, so all pointer casts are valid

BenGoldberger (approved):

PR #6237, opened 2025-05-29
MOD-9609: Pass Private Data Context Back To UnblockClient API

kei-nan:
Weird not so well documented behaviour.
Redis has `RedisModule_BlockClientSetPrivateData` which we used after blocking the client.
We assumed based on the documentation of the `BlockClient` API that the free callback will be called with the context we set.
Apparently it uses the context that gets passed to `UnblockClient`, since we passed NULL it didn't call the free function for the query nodes.
This caused the server to crash during the exit flow, we think RLTest sometimes missed this and sometimes didn't
Now getting the context and passing it back to unblock client.
Saw that the free function now gets called.

1. Current: Low coverage due to server crashing on exit.
2. Change: Passing the right context to the unblock api call.
3. Outcome: Server stops crashing, better coverage.

1. Blocking Client Flow

#### Mark if applicable

### DIFF
--- a/src/info/info_redis/types/blocked_queries.c
+++ b/src/info/info_redis/types/blocked_queries.c
@@ -90,6 +90,7 @@ void BlockedQueries_RemoveQuery(BlockedQueryNode* blockedQueryNode) {
 }
 
 void BlockedQueries_RemoveCursor(BlockedCursorNode* blockedCursorNode) {
+  StrongRef_Release(blockedCursorNode->spec);
   if (blockedCursorNode->query) {
     rm_free(blockedCursorNode->query);
   }


src/aggregate/aggregate_exec.c
```diff
@@ -974,7 +975,8 @@ static int buildPipelineAndExecute(AREQ *r, RedisModuleCtx *ctx, QueryError *sta
       rs_timersub(&time, &r->qiter.initTime, &time);
       rs_timeradd(&time, &r->qiter.GILTime, &r->qiter.GILTime);
     }
-    workersThreadPool_AddWork((redisearch_thpool_proc)AREQ_Execute_Callback, BCRctx);
+    const int rc = workersThreadPool_AddWork((redisearch_thpool_proc)AREQ_Execute_Callback, BCRctx);
+    RS_ASSERT_ALWAYS(rc == 0);
```
GuyAv46, line 979: Why do we want to assert that (and always)?

GuyAv46 (commented):
Awesome catch!

src/aggregate/aggregate_exec.c
GuyAv46: `workersThreadPool_AddWork` only returns non-zero on memory allocation failure

GuyAv46: Since it cannot happen today, I think it is better to have `RS_ASSERT` only, until we have a reason to use it always

Itzikvaknin (dismissed):
Very nice!

GuyAv46 (approved):

PR #6238, opened 2025-06-02
refactor NotIterator - [MOD-9677]

JoanFM:

This will be the continuation of #6307 

The PR does:

- [x] Provide the functionality of the **NotIterator**
- [x] Adds testing
- [x] Adds micro-benchmarks

#### Mark if applicable

## Note 

I strongly believe that the OldIterator has a bug. 

The Old implementation assumes that there are no DocID in the `childIterator` that does not exist in the `wildcard iterator`(the child iterator (the negation functions) seems to be assumed to be a subgroup of the WildCard iterator). 

This test fails for OldIterator:

```cpp
#include "gtest/gtest.h"
#include "redisearch.h"
#include "index.h"
#include "query.h"
#include "micro-benchmarks/deprecated_iterator_util.h"
#include <vector>
#include <algorithm>
#include <iostream>

class NotIteratorOldTest : public ::testing::Test {
protected:
  IndexIterator *iterator_base;
  std::vector<t_docId> childDocIds = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
  std::vector<t_docId> wcDocIds = {3, 4, 9};
  std::vector<t_docId> resultSet;
  t_docId maxDocId = 10;

  void SetUp() override {
    // Create the child iterator with the specified document IDs
    IndexIterator *child = (IndexIterator *)new MockOldIterator(childDocIds);

    // Create the wildcard iterator with the specified document IDs
    IndexIterator *wcii = (IndexIterator *)new MockOldIterator(wcDocIds);

    // Set up the timeout
    struct timespec timeout = {LONG_MAX, 999999999}; // "infinite" timeout

    // Create the NOT iterator with the child and wildcard iterators
    iterator_base = _New_NotIterator_With_WildCardIterator(child, wcii, maxDocId, 1.0, timeout);

    // Compute the expected result set
    // For the optimized NOT iterator, the result set is all documents in the wildcard iterator
    // that are not in the child iterator
    resultSet.clear();
    for (auto wcId : wcDocIds) {
      if (std::find(childDocIds.begin(), childDocIds.end(), wcId) == childDocIds.end()) {
        resultSet.push_back(wcId);
      }
    }

    // Print test setup information
    std::cout << "Child Doc IDs: ";
    for (auto id : childDocIds) {
      std::cout << id << " ";
    }
    std::cout << std::endl;

    std::cout << "Wildcard Doc IDs: ";
    for (auto id : wcDocIds) {
      std::cout << id << " ";
    }
    std::cout << std::endl;

    std::cout << "Expected Result Set: ";
    if (resultSet.empty()) {
      std::cout << "(empty)";
    } else {
      for (auto id : resultSet) {
        std::cout << id << " ";
      }
    }
    std::cout << std::endl;
  }

  void TearDown() override {
    iterator_base->Free(iterator_base);
  }
};

TEST_F(NotIteratorOldTest, ReadOptimized) {
  RSIndexResult *hit = iterator_base->current;
  int rc;

  std::cout << "=== READING RESULTS ===\n";
  size_t count = 0;

  // Read all results from the iterator
  while ((rc = iterator_base->Read(iterator_base->ctx, &hit)) == INDEXREAD_OK) {
    std::cout << "Read result: docId=" << hit->docId << std::endl;
    // Verify the result is in our expected result set
    //ASSERT_EQ(hit->docId, resultSet[count]);
    count++;
  }
  ASSERT_EQ(count, 0);

  // Verify we reached EOF
  ASSERT_EQ(rc, INDEXREAD_EOF);

  // Verify we read the expected number of results
  ASSERT_EQ(count, resultSet.size()) << "Expected to read " << resultSet.size() << " documents";
  std::cout << "Read " << count << " results (expected " << resultSet.size() << ")" << std::endl;

  // Try reading again after EOF
  rc = iterator_base->Read(iterator_base->ctx, &hit);
  ASSERT_EQ(rc, INDEXREAD_EOF);
  std::cout << "=== FINISHED READING ===\n";
}
```

This seems to be OK for most of the times because the QueryEvalNode filters the Not Iterator conditions before creating the ChildIterator, but there may be some edge cases where this is not true.

This makes the Benchmarks look bad for this NEW implementation.

However, I added a set of tests where the conditions are met (_Subset) in the micro benchmarks, that show that the performance is comparable or slightly better for the new implementation

### DIFF
--- a/src/iterators/not_iterator.c
+++ b/src/iterators/not_iterator.c
@@ -86,43 +86,30 @@ static IteratorStatus NI_Read_Optimized(QueryIterator *base) {
     return ITERATOR_EOF;
   }
 
-  IteratorStatus rc;
-
-  // If child has not read any element, read one
-  if (ni->child->lastDocId == 0) {
-    // Read at least the first entry of the child
-    rc = ni->child->Read(ni->child);
-    if (rc == ITERATOR_TIMEOUT) return ITERATOR_TIMEOUT;
-  }
-
   // Advance to the next potential docId
-  rc = ni->wcii->Read(ni->wcii);
+  IteratorStatus rc = ni->wcii->Read(ni->wcii);
   if (rc == ITERATOR_TIMEOUT) {
     return ITERATOR_TIMEOUT;
   }
   // Iterate through all the documents present in the wcii until we find one that is not in child
   while (!ni->wcii->atEOF) {
-    // Case 1: Current docID is less than child's docID or child is exhauster.
-    // This means we found a document that is not in the child iterator
     if (ni->child->atEOF || ni->wcii->lastDocId < ni->child->lastDocId) {
+      // Case 1: Current docID is less than child's docID or child is exhauster.
+      // This means we found a document that is not in the child iterator
       base->lastDocId = base->current->docId = ni->wcii->lastDocId;
       return ITERATOR_OK; // Found a valid difference element
-    }
-
-    // Case 2: Current docID is equal to child's docID.
-    // If we're at the same docId as the child, we need to advance both and in next loop iteration we will see how they compare
-    // TODO: Joan here we are not doing this
-    else if (ni->wcii->lastDocId == ni->child->lastDocId) {
+    } else if (ni->wcii->lastDocId == ni->child->lastDocId) {
+      // Case 2: Current docID is equal to child's docID.
+      // If we're at the same docId as the child, we need to advance both and in next loop iteration we will see how they compare
       rc = ni->child->Read(ni->child);
       if (rc == ITERATOR_TIMEOUT) return rc;
       rc = ni->wcii->Read(ni->wcii);
       if (rc == ITERATOR_TIMEOUT) {
         return ITERATOR_TIMEOUT;
       }
-    }
-    // Case 3: Current docID is ahead of child's docID
-    // This means we need to advance the child until it catches up, and in next loop we will see how they compare
-    else { //(ni->child->lastDocId < ni->wcii->lastDocId)
+    } else { //(ni->child->lastDocId < ni->wcii->lastDocId)
+      // Case 3: Current docID is ahead of child's docID
+      // This means we need to advance the child until it catches up, and in next loop we will see how they compare
       while (!ni->child->atEOF && ni->child->lastDocId < ni->wcii->lastDocId) {
         rc = ni->child->Read(ni->child);
         if (rc == ITERATOR_TIMEOUT) return rc;
@@ -140,6 +127,7 @@ static IteratorStatus NI_Read_Optimized(QueryIterator *base) {
  * NOTFOUND. If we don't or we're at the end - return OK */
 static IteratorStatus NI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId) {
   NotIterator *ni = (NotIterator *)base;
+  RS_ASSERT(base->lastDocId < docId);
   // do not skip beyond max doc id
   if (base->atEOF) {
     return ITERATOR_EOF;
@@ -148,7 +136,6 @@ static IteratorStatus NI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId)
     base->atEOF = true;
     return ITERATOR_EOF;
   }
-  RS_ASSERT(base->lastDocId < docId);
 
   // Case 1: Child is ahead or at EOF - docId is not in child
   if (ni->child->lastDocId > docId || ni->child->atEOF) {
@@ -185,6 +172,7 @@ static IteratorStatus NI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId)
  */
 static IteratorStatus NI_SkipTo_Optimized(QueryIterator *base, t_docId docId) {
   NotIterator *ni = (NotIterator *)base;
+  RS_ASSERT(base->lastDocId < docId);
 
   // Check if we've reached the end or if docId exceeds maximum
   if (base->atEOF) {
@@ -194,8 +182,6 @@ static IteratorStatus NI_SkipTo_Optimized(QueryIterator *base, t_docId docId) {
     base->atEOF = true;
     return ITERATOR_EOF;
   }
-  RS_ASSERT(base->lastDocId < docId);
-
   // Handle relative positions of wildcard and child iterators
   IteratorStatus rc = ni->wcii->SkipTo(ni->wcii, docId);
   if (rc == ITERATOR_TIMEOUT) return ITERATOR_TIMEOUT;
@@ -204,13 +190,12 @@ static IteratorStatus NI_SkipTo_Optimized(QueryIterator *base, t_docId docId) {
     return rc;
   }
 
-  // Case 1: Wildcard is behind child
   if (ni->wcii->lastDocId < ni->child->lastDocId || ni->child->atEOF) {
+    // Case 1: Wildcard is behind child
     // Wildcard found a document before child's position - valid result
     base->lastDocId = base->current->docId = ni->wcii->lastDocId;
-  }
-  // Case 2: Both iterators at same position
-  else if (ni->wcii->lastDocId == ni->child->lastDocId) {
+  } else if (ni->wcii->lastDocId == ni->child->lastDocId) {
+    // Case 2: Both iterators at same position
     // Both at same position - find next valid result
     rc = NI_Read_Optimized(base);
     if (rc == ITERATOR_OK) {
@@ -218,9 +203,8 @@ static IteratorStatus NI_SkipTo_Optimized(QueryIterator *base, t_docId docId) {
     } else if (rc == ITERATOR_EOF) {
       RS_ASSERT(base->atEOF);
     }
-  }
-  // Case 3: Wildcard is ahead of child
-  else { // ni->wcii->lastDocId > ni->child->lastDocId
+  } else { // ni->wcii->lastDocId > ni->child->lastDocId
+    // Case 3: Wildcard is ahead of child
     // Wildcard advanced past child - check if child has this new docID
     IteratorStatus child_rc = ni->child->SkipTo(ni->child, ni->wcii->lastDocId);
     if (child_rc == ITERATOR_TIMEOUT) return ITERATOR_TIMEOUT;


JoanFM:
Putting it as draft until #6240 and #6241 are merged so that this PR is smaller

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,237 @@
...
+  if (ni->wcii) {
+    ni->wcii->Free(ni->wcii);
+  }
+  IndexResult_Free(base->current);
+  rm_free(base);
+}
+
+static size_t NI_NumEstimated(QueryIterator *base) {
+  NotIterator *ni = (NotIterator *)base;
+  return ni->maxDocId - ni->child->NumEstimated(ni->child);
```
GuyAv46: Thinking about it, this may lead to underestimating, while we usually expect the estimation to overestimate. Maybe we'd better keep the estimation as it is (don't consult with the child)

GuyAv46: also lets test this funcion

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,237 @@
...
+}
+
+static IteratorStatus NI_ReadSorted_O(QueryIterator *base); // forward decl
+static IteratorStatus NI_ReadSorted_NO(QueryIterator *base); // forward decl
```
GuyAv46: Let's rename all these functions.
- No need to specify "sorted"
- Better names for optimized non-optimized implementations, instead of `_NO` sufixes

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,237 @@
...
+
+/* SkipTo for NOT iterator - Non-optimized version. If we have a match - return
+ * NOTFOUND. If we don't or we're at the end - return OK */
+static IteratorStatus NI_SkipTo_NO(QueryIterator *base, t_docId docId) {
```
GuyAv46: Let's add an assertion `RS_ASSERT(base->lastDocId < docId)`

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,237 @@
...
+  NotIterator *ni = (NotIterator *)base;
+  // do not skip beyond max doc id
+  if (docId > ni->maxDocId) {
```
GuyAv46: No early return on EOF?

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,237 @@
...
+    // read the next entry from the child
+    IteratorStatus rc = ni->child->SkipTo(ni->child, docId);
+    if (rc == ITERATOR_TIMEOUT) return ITERATOR_TIMEOUT;
+    if (rc != ITERATOR_OK) {
+      base->lastDocId = base->current->docId = docId;
+      return ITERATOR_OK;
+    }
+  }
+  // If the child docId is the one we are looking for, it's an anti match!
+  // We need to return NOTFOUND and set hit to the next valid docId
```
GuyAv46:
```suggestion
  // We need to return NOTFOUND and set the current result to the next valid docId
```

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,237 @@
...
+  IteratorStatus rc = NI_ReadSorted_NO(base);
+  if (rc == ITERATOR_OK) {
+    return ITERATOR_NOTFOUND;
+  }
+  return rc;
+}
+
+/* SkipTo for NOT iterator - Optimized version. If we have a match - return
+ * NOTFOUND. If we don't or we're at the end - return OK */
+static IteratorStatus NI_SkipTo_O(QueryIterator *base, t_docId docId) {
```
GuyAv46: Let's add an assertion `RS_ASSERT(base->lastDocId < docId)`

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,237 @@
...
+  rc = NI_ReadSorted_O(base);
+  if (rc == ITERATOR_OK) {
+    return ITERATOR_NOTFOUND;
+  }
+  return rc;
+}
+
+/* Read from a NOT iterator - Non-Optimized version. This is applicable only if
+ * the only or leftmost node of a query is a NOT node. We simply read until max
+ * docId, skipping docIds that exist in the child */
```
GuyAv46: Incorrect, let's fix this outdated comment
```suggestion
/* Read from a NOT iterator - Non-Optimized version. We simply read until max
 * docId, skipping docIds that exist in the child */
```

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  IteratorStatus rc;
+
+  if (ni->child->lastDocId == 0) {
+    // Read at least the first entry of the child
+    rc = ni->child->Read(ni->child);
+    if (rc == ITERATOR_TIMEOUT) return ITERATOR_TIMEOUT;
+  }
+  //Advance and see if it si valid
```
GuyAv46:
```suggestion
  // Advance and see if it is valid
```

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,237 @@
...
+    rc = ni->child->Read(ni->child);
+    if (rc == ITERATOR_TIMEOUT) return rc;
+    // Check for timeout with low granularity (MOD-5512)
+    //TODO(Joan): Ticket about magic number? Should it be a config?
+    if (TimedOut_WithCtx_Gran(&ni->timeoutCtx, 5000)) {
+      base->atEOF = true;
+      return ITERATOR_TIMEOUT;
+    }
+    // The lastDocID that we proposed is not valid, so we need to try another one
+    rc = ni->wcii->Read(ni->wcii);
```
GuyAv46: Unlike the non-optimized implementation, the wildcard iterator can surpass the child iterator here. We have to address that, otherwise we may miss relevant results

src/iterators/not_iterator.h
```diff
@@ -0,0 +1,40 @@
...
+extern "C" {
+#endif
+
+typedef struct {
+  QueryIterator base;         // base index iterator
+  QueryIterator *wcii;        // wildcard index iterator
+  QueryIterator *child;       // child index iterator
+  t_docId maxDocId;
+  TimeoutCtx timeoutCtx;
+} NotIterator, NotContext;
```
GuyAv46:
```suggestion
} NotIterator;
```

src/iterators/not_iterator.h
```diff
@@ -0,0 +1,40 @@
...
+* @param weight - the weight of the node (assigned to the returned result)
+* @param timeout - the timeout for the iterator
+* @param q - the query context
+*/
+QueryIterator *IT_V2(NewNotIterator)(QueryIterator *it, t_docId maxDocId, double weight, struct timespec timeout, QueryEvalCtx *q);
+
+#ifdef __cplusplus
+}
+#endif
+
```
GuyAv46: Clean whitespace

src/iterators/not_iterator.c
JoanFM: Yes, it needs better logic and better testing. Will put as draft for now

tests/cpptests/micro-benchmarks/benchmark_not_iterator.cpp
```diff
@@ -0,0 +1,365 @@
...
+      this->SetName(#BaseClass "<" #a "," #b ", " #c ">/" #Method);       \
+    }                                                                     \
+                                                                          \
+   protected:                                                             \
+    void BenchmarkCase(::benchmark::State&) BENCHMARK_OVERRIDE;           \
+  };
+
+  #define BENCHMARK_TEMPLATE3_DEFINE_F(BaseClass, Method, a, b, c)    \
+  BENCHMARK_TEMPLATE3_PRIVATE_DECLARE_F(BaseClass, Method, a, b, c)   \
+  void BENCHMARK_PRIVATE_CONCAT_NAME(BaseClass, Method)::BenchmarkCase
```
GuyAv46: CAn't we use the generic `BENCHMARK_TEMPLATE_DEFINE_F` macro? We already require C++17

tests/cpptests/iterator_util.h
```diff
@@ -31,74 +33,94 @@ class MockIterator {
...
-            base.atEOF = true;
-            return whenDone;
-        }
-        base.lastDocId = base.current->docId = docIds[nextIndex++];
-        return ITERATOR_OK;
+      if (sleepTime.has_value()) {
+        std::this_thread::sleep_for(sleepTime.value());
+      }
+      readCount++;
+      if (nextIndex >= docIds.size() || base.atEOF || docIds.size() == 0) {
```
GuyAv46: If `docIds.size() == 0` then `nextIndex >= docIds.size()`, no? Do we need this validation?

tests/cpptests/iterator_util.h
```diff
@@ -31,74 +33,94 @@ class MockIterator {
...
+      docIds.erase(new_end, docIds.end());
+    }
+
+    template<typename... Args>
+    MockIterator(std::chrono::milliseconds sleep, Args&&... args)
+      : docIds({std::forward<Args>(args)...}), whenDone(ITERATOR_EOF), nextIndex(0), readCount(0), sleepTime(sleep) {
+      setBase(&base);
+      auto new_end = std::unique(docIds.begin(), docIds.end());
+      docIds.erase(new_end, docIds.end());
+    }
```
GuyAv46: We should stick to a single constructor and reuse it (like we had before), avoiding logic duplication

tests/cpptests/iterator_util.h
```diff
@@ -31,74 +33,94 @@ class MockIterator {
...
 
     template<typename... Args>
-    MockIterator(IteratorStatus st, Args&&... ids_args)
-        : docIds({ids_args...}), whenDone(st), nextIndex(0), readCount(0) {
-        setBase(&base);
-        std::sort(docIds.begin(), docIds.end());
```
GuyAv46: We should keep the sorting here. Some places already rely on this

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,355 @@
...
+  return ITERATOR_EOF;
+}
+
+QueryIterator *IT_V2(NewNotIterator)(QueryIterator *it, t_docId maxDocId, double weight, struct timespec timeout, QueryEvalCtx *q) {
+  NotIterator *ni = rm_calloc(1, sizeof(*ni));
+  QueryIterator *ret = &ni->base;
+  bool optimized = q && q->sctx->spec->rule && q->sctx->spec->rule->index_all;
+  if (optimized) {
+    //TODO(Joan): Once all the WildCards are migrated, call the generic one
+    ni->wcii = IT_V2(NewWildcardIterator_NonOptimized)(q->docTable->maxDocId, q->docTable->size);
```
GuyAv46: This API is ready

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,355 @@
...
+  if (ni->wcii) {
+    ni->wcii->Free(ni->wcii);
+  }
+  IndexResult_Free(base->current);
+  rm_free(base);
+}
+
+static size_t NI_NumEstimated(QueryIterator *base) {
+  NotIterator *ni = (NotIterator *)base;
+  return ni->maxDocId;
```
GuyAv46: If we have a wildcard iterator, we better use its estimation. `maxDocId` can grow much higher than the actual number of documents. We also have the doc table size in the construction, we can hold it as well as a better estimation

tests/cpptests/micro-benchmarks/benchmark_not_iterator.cpp
JoanFM: It does not have option to have 3 template parameters, they have for up to 2

tests/cpptests/iterator_util.h
JoanFM: I think I did not realize I removed that

tests/cpptests/test_cpp_iterator_not.cpp
```diff
@@ -547,7 +567,7 @@ INSTANTIATE_TEST_SUITE_P(
       }()
     ),
     ::testing::Values(std::nullopt),
-    ::testing::Values(false)
+    ::testing::Values(false, true)
```
GuyAv46:
```suggestion
    ::testing::Bool()
```

tests/cpptests/test_cpp_iterator_not.cpp
```diff
@@ -449,7 +469,7 @@ INSTANTIATE_TEST_SUITE_P(
       std::vector<t_docId>{50, 60, 70, 80, 90, 100, 600, 750, 950, 1200}
     ),
     ::testing::Values(std::nullopt),
-    ::testing::Values(false)
+    ::testing::Values(false, true)
```
GuyAv46:
```suggestion
    ::testing::Bool()
```

tests/cpptests/test_cpp_iterator_not.cpp
```diff
@@ -376,7 +394,7 @@ INSTANTIATE_TEST_SUITE_P(
       }()
     ),
     ::testing::Values(std::nullopt, 100),
-    ::testing::Values(false)
+    ::testing::Values(false, true)
```
GuyAv46:
```suggestion
    ::testing::Bool()
```

tests/cpptests/test_cpp_iterator_not.cpp
```diff
@@ -109,6 +111,9 @@ class NotIteratorCommonTest : public ::testing::TestWithParam<std::tuple<std::ve
 };
 
 TEST_P(NotIteratorCommonTest, Read) {
+  if (optimized && opt_max_doc_id.has_value()) {
+    GTEST_SKIP() << "For optimized version the maxDocId is not necessarily respected (smaller than the max of Child or WCDocId): Is it worth adding this check? return base->lastDocId < ni->maxDocId ? ITERATOR_EOF : ITERATOR_OK; ";
+  }
```
GuyAv46, line 116: I agree that this scenario is not relevant.
Can we define the test combinations in a way that avoids redundant combinations? (Maybe taking the maximal id from the wildcard IDs vector if `optimized == false`?)

src/iterators/not_iterator.c
```diff
@@ -74,6 +74,71 @@ static IteratorStatus NI_Read_NotOptimized(QueryIterator *base) {
...
+  if (ni->child->lastDocId == 0) {
+    // Read at least the first entry of the child
+    rc = ni->child->Read(ni->child);
+    if (rc == ITERATOR_TIMEOUT) return ITERATOR_TIMEOUT;
+  }
+
+  // Advance to the next potential docId
+  rc = ni->wcii->Read(ni->wcii);
+  if (rc == ITERATOR_TIMEOUT) {
+    base->atEOF = true;
```
GuyAv46: we shouldn't mark the iterator as EOF on timeout

src/iterators/not_iterator.c
```diff
@@ -74,6 +74,71 @@ static IteratorStatus NI_Read_NotOptimized(QueryIterator *base) {
...
+    // Case 3: Current docID is ahead of child's docID
+    // This means we need to advance the child until it catches up, and in next loop we will see how they compare
+    else { //(ni->child->lastDocId < ni->wcii->lastDocId)
+      while (!ni->child->atEOF && ni->child->lastDocId < ni->wcii->lastDocId) {
+        rc = ni->child->Read(ni->child);
+        if (rc == ITERATOR_TIMEOUT) return rc;
+      }
+    }
+    if (TimedOut_WithCtx_Gran(&ni->timeoutCtx, 5000)) {
+      base->atEOF = true;
```
GuyAv46: we shouldn't mark the iterator as EOF on timeout

src/iterators/not_iterator.c
```diff
@@ -74,6 +74,71 @@ static IteratorStatus NI_Read_NotOptimized(QueryIterator *base) {
...
+
+    // Case 2: Current docID is equal to child's docID.
+    // If we're at the same docId as the child, we need to advance both and in next loop iteration we will see how they compare
+    // TODO: Joan here we are not doing this
+    else if (ni->wcii->lastDocId == ni->child->lastDocId) {
+      rc = ni->child->Read(ni->child);
+      if (rc == ITERATOR_TIMEOUT) return rc;
+      rc = ni->wcii->Read(ni->wcii);
+      if (rc == ITERATOR_TIMEOUT) {
+        base->atEOF = true;
```
GuyAv46: we shouldn't mark the iterator as EOF on timeout

src/iterators/not_iterator.c
```diff
@@ -112,14 +177,113 @@ static IteratorStatus NI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId)
...
+      // Child doesn't have this document - valid result
+      base->lastDocId = base->current->docId = ni->wcii->lastDocId;
+      return base->lastDocId == docId ? ITERATOR_OK : ITERATOR_NOTFOUND;
+    }
+  }
+
+  // Default fallback (should not reach here in normal execution)
+  // LCOV_EXCL_START
+  return ITERATOR_EOF;
+  // LCOV_EXCL_STOP
```
GuyAv46: Let's avoid using these magic comments in production code. If we shouldn't reach here, can we rewrite the logic so we don't have redundant fallbacks?

src/iterators/not_iterator.c
```diff
@@ -112,14 +177,113 @@ static IteratorStatus NI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId)
...
+ * - ni: The NotIterator
+ * - docId: The original requested document ID
+ *
+ * Returns:
+ * - ITERATOR_OK if a valid result was found and it matches the original docId
+ * - ITERATOR_NOTFOUND if a valid result was found but it's not the original docId
+ * - ITERATOR_EOF if we've reached the end
+ * - ITERATOR_TIMEOUT if the operation timed out
+ */
+static IteratorStatus NI_HandleRelativePositions(QueryIterator *base, NotIterator *ni, t_docId docId) {
```
GuyAv46: This function is very complex and is only used once, so I suspect it has redundancies.
The implementation in #5619 is much shorter. Is there a problem with it?

src/iterators/not_iterator.c
JoanFM: I agree, but in this case the old implementation did. Just FYI

tests/cpptests/test_cpp_iterator_not.cpp
JoanFM: I can do, but to me is easier to understand a clear skip message than a hard-to-follow logic in the SetUp. Your call

src/iterators/not_iterator.c
JoanFM: It seems to not account for the situation where the subset of ChildIDs is not in the WCIds

JoanFM: Tests fail if changing the implementation for this:

```cpp
IteratorStatus NI_SkipTo_Optimized(QueryIterator *base, t_docId docId) {
  // do not skip beyond max doc id
  NotIterator *ni = (NotIterator *)base;
  if (docId > ni->maxDocId) {
    base->atEOF = true;
    return ITERATOR_EOF;
  }

  RSIndexResult *wr;
  IteratorStatus rc = ni->wcii->SkipTo(ni->wcii, docId);
  if (rc != ITERATOR_OK && rc != ITERATOR_OK) {
    return rc;
  } else if (rc == ITERATOR_OK) {
    // A valid wildcard result was found. Let's check if the child has it
    if (ni->child->lastDocId > docId || ni->child->atEOF) {
      // If the child is ahead of the skipto id, it means the child doesn't have this id.
      // So we are okay!
      base->lastDocId = base->current->docId = docId;
      return ITERATOR_OK;
    } else if (ni->child->lastDocId < docId) {
      // read the next entry from the child
      rc = ni->child->SkipTo(ni->child, docId);
      if (rc == ITERATOR_TIMEOUT) return ITERATOR_TIMEOUT;
      if (rc != ITERATOR_OK) {
        base->lastDocId = base->current->docId = docId;
      }
    }
  }

  // If the wildcard iterator is missing the docId, or the child iterator has it,
  // We need to return NOTFOUND and set hit to the next valid docId
  rc = NI_Read_Optimized(base);
  if (rc == ITERATOR_OK) {
    return ITERATOR_NOTFOUND;
  }
  return rc;
}
```

GuyAv46: I see...

src/iterators/not_iterator.c
```diff
@@ -112,14 +174,108 @@ static IteratorStatus NI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId)
...
+    return base->lastDocId == docId ? ITERATOR_OK : ITERATOR_NOTFOUND;
+  }
+  // Case 2: Both iterators at same position
+  else if (ni->wcii->lastDocId == ni->child->lastDocId) {
+    // Both at same position - find next valid result
+    rc = NI_Read_Optimized(base);
+    if (rc == ITERATOR_OK) {
+      return ITERATOR_NOTFOUND;
+    } else if (rc == ITERATOR_EOF) {
+      base->atEOF = true;
```
GuyAv46: The call to `NI_Read_Optimized` already set it up
```suggestion
```
(We can assert that instead)
```suggestion
    } else if (rc == ITERATOR_EOF) {
      RS_ASSERT(base->atEOF);
```

src/iterators/not_iterator.c
```diff
@@ -112,14 +174,108 @@ static IteratorStatus NI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId)
...
+        return ITERATOR_NOTFOUND;
+      } else if (rc == ITERATOR_EOF) {
+        base->atEOF = true;
+      }
+    } else if (rc == ITERATOR_NOTFOUND || rc == ITERATOR_EOF) {
+      // Child doesn't have this document - valid result
+      base->lastDocId = base->current->docId = ni->wcii->lastDocId;
+      return base->lastDocId == docId ? ITERATOR_OK : ITERATOR_NOTFOUND;
+    }
+  }
```
GuyAv46: At this point, `child->lastDocId >= docId` (if it was initially <, we skipped it to `docId` and now it's `child->lastDocId >= docId`)
This means that this path necessarily handles `ITERATOR_NOTFOUND` case

src/iterators/not_iterator.c
```diff
@@ -112,14 +174,108 @@ static IteratorStatus NI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId)
...
+    rc = ni->child->SkipTo(ni->child, ni->wcii->lastDocId);
+    if (rc == ITERATOR_TIMEOUT) return ITERATOR_TIMEOUT;
+
+    if (rc == ITERATOR_OK) {
+      // Child has this document - find next valid result
+      rc = NI_Read_Optimized(base);
+      if (rc == ITERATOR_OK) {
+        return ITERATOR_NOTFOUND;
+      } else if (rc == ITERATOR_EOF) {
+        base->atEOF = true;
```
GuyAv46: Should've been handled by `NI_Read_Optimized`

src/iterators/not_iterator.c
```diff
@@ -112,14 +174,108 @@ static IteratorStatus NI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId)
...
+      }
+    } else if (rc == ITERATOR_NOTFOUND || rc == ITERATOR_EOF) {
+      // Child doesn't have this document - valid result
+      base->lastDocId = base->current->docId = ni->wcii->lastDocId;
+      return base->lastDocId == docId ? ITERATOR_OK : ITERATOR_NOTFOUND;
```
GuyAv46: Cannot be OK
```suggestion
      return ITERATOR_NOTFOUND;
```

src/iterators/not_iterator.c
```diff
@@ -112,14 +174,108 @@ static IteratorStatus NI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId)
...
+    if (rc == ITERATOR_OK) {
+      return ITERATOR_NOTFOUND;
+    } else if (rc == ITERATOR_EOF) {
+      base->atEOF = true;
+    }
+  }
+  // Case 3: Wildcard is ahead of child
+  else { // ni->wcii->lastDocId > ni->child->lastDocId
+    // Wildcard advanced past child - check if child has this new docID
+    rc = ni->child->SkipTo(ni->child, ni->wcii->lastDocId);
```
GuyAv46: This flow may require calling `child->SkipTo` twice:
1. `docId > child->lastDocId` so we skip to it (from `NI_SkipTo_Optimized`)
2. We call `wcii->SkipTo` (always, at the start of the function)
3. And now here

I believe removing the first call will still work and improve the performance in this case.
I also suggest inlining this function so these observations are easier to see

src/iterators/not_iterator.c
JoanFM: not if the ChildID is atEOF right? If I do not do this, some tests fail

JoanFM: What I can do is if Child is atEOF, directly call wc_>SkipTo

src/iterators/not_iterator.c
JoanFM: Removing the first call makes SkipTo fail for "Child not Subset of WildCard case".

JoanFM: I will do some refactor to simplify, it can be handled in a signle fucntion

src/iterators/not_iterator.c
```diff
@@ -74,6 +74,68 @@ static IteratorStatus NI_Read_NotOptimized(QueryIterator *base) {
...
+  }
+
+  IteratorStatus rc;
+
+  // If child has not read any element, read one
+  if (ni->child->lastDocId == 0) {
+    // Read at least the first entry of the child
+    rc = ni->child->Read(ni->child);
+    if (rc == ITERATOR_TIMEOUT) return ITERATOR_TIMEOUT;
+  }
```
GuyAv46: This step seems redundant with your implementation. Seems like we will necessarily reach case 3 and will exit the while loop by the same condition

src/iterators/not_iterator.c
```diff
@@ -74,6 +74,68 @@ static IteratorStatus NI_Read_NotOptimized(QueryIterator *base) {
...
+      }
+    }
+    // Case 3: Current docID is ahead of child's docID
+    // This means we need to advance the child until it catches up, and in next loop we will see how they compare
+    else { //(ni->child->lastDocId < ni->wcii->lastDocId)
+      while (!ni->child->atEOF && ni->child->lastDocId < ni->wcii->lastDocId) {
+        rc = ni->child->Read(ni->child);
+        if (rc == ITERATOR_TIMEOUT) return rc;
+      }
+    }
```
GuyAv46: Let's finalize the implementation here

- [ ] remove the `TODO`s left (address them if not done already)
- [ ] Move each case comment into its case block. At the moment, it's hard to see that the conditions are connected to each other

src/iterators/not_iterator.c
```diff
@@ -112,14 +174,94 @@ static IteratorStatus NI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId)
...
+    // Both at same position - find next valid result
+    rc = NI_Read_Optimized(base);
+    if (rc == ITERATOR_OK) {
+      return ITERATOR_NOTFOUND;
+    } else if (rc == ITERATOR_EOF) {
+      RS_ASSERT(base->atEOF);
+    }
+  }
+  // Case 3: Wildcard is ahead of child
+  else { // ni->wcii->lastDocId > ni->child->lastDocId
```
GuyAv46: Same comments about the case comments and `else..if` layout

tests/cpptests/test_cpp_iterator_not.cpp
```diff
@@ -134,6 +139,9 @@ TEST_P(NotIteratorCommonTest, SkipToEOF) {
 }
```
GuyAv46, line 139: Fix skip-to id (skipping backwards is undefined). This means we should also push the assertion about it to the beginning of both SkipTo implementations

GuyAv46 (approved):

PR #6245, opened 2025-06-03
[MOD-9978] Use offset rather than a cursor in BufferWriter

LukeMathWalker:

Align `BufferWriter` with `BufferReader`: track the current position using an offset rather than a cursor pointer.
This reduces the amount of bookkeeping we need to—when the buffer grows, we no longer need to update the cursor to point at the new allocation. The offset remains valid.

#### Mark if applicable

### DIFF
--- a/src/varint.c
+++ b/src/varint.c
@@ -63,9 +63,8 @@ size_t WriteVarint(uint32_t value, BufferWriter *w) {
   size_t pos = varintEncode(value, varint);
   size_t nw = VARINT_LEN(pos);
 
-  size_t mem_growth = 0;
   // we assume buffer reserve will not fail
-  mem_growth = Buffer_Reserve(w->buf, nw);
+  size_t mem_growth = Buffer_Reserve(w->buf, nw);
 
   memcpy(BufferWriter_Current(w), VARINT_BUF(varint, pos), nw);
 


src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -138,31 +138,20 @@ impl Buffer {
...
-/// ```no_run
-/// # use std::io::Write;
-/// # use buffer::{Buffer, BufferWriter};
-/// # use std::ptr::NonNull;
-/// let buffer_ptr: NonNull<Buffer> = todo!();
-/// let mut writer = unsafe { BufferWriter::for_buffer(buffer_ptr) };
-///
-/// // Write data to the buffer
-/// writer.write(b"Hello, world!").unwrap();
-/// ```
```
chesedo, line 165: Is this removed because of the `no_run`?

LukeMathWalker: It's removed because the signature of `for_buffer` has been changed to take a proper `&mut Buffer` rather than a `NonNull<Buffer>`, and that scenario is already tested in `tests.rs`.

tests/cpptests/test_cpp_index.cpp
```diff
@@ -1173,7 +1173,7 @@ TEST_F(IndexTest, testBuffer) {
   ASSERT_TRUE(w.buf->cap == 2) << "Wrong capacity";
   ASSERT_TRUE(w.buf->data != NULL);
   ASSERT_TRUE(Buffer_Offset(w.buf) == 0);
-  ASSERT_TRUE(w.buf->data == w.pos);
```
chesedo, line 1176: Does this mean this PR also updated all direct field accesses?

src/redisearch_rs/buffer/src/lib.rs
chesedo: I see. Does that mean we prefer proper tests over doc tests?

LukeMathWalker: Not as a blanket policy, but in this case I don't think the doc test was adding much.

tests/cpptests/test_cpp_index.cpp
LukeMathWalker: Correct!

src/buffer.c
```diff
@@ -94,7 +94,7 @@ size_t BufferWriter_Seek(BufferWriter *b, size_t offset) {
   if (offset > b->buf->cap) {
     return b->buf->offset;
   }
-  b->pos = b->buf->data + offset;
+  b->pos = offset;
```
JoanFM, line 97: It seems then that we have duplicated information?

src/buffer.h
```diff
@@ -105,7 +105,7 @@ position if where is outside bounds
 
 typedef struct {
   Buffer *buf;
-  char *pos;
+  size_t pos;
```
JoanFM, line 108: how does it differ from the `Buffer` internal `offset`?

chesedo (dismissed):

src/buffer.h
LukeMathWalker: `Buffer`'s `offset` tracks the number of elements stored in the buffer.
`BufferWriter`'s `pos` tracks the position where the next element will be written. 

The two don't have to match—you may choose to overwrite existing elements by setting `pos` to a smaller number than `Buffer`'s `offset`.

src/buffer.c
LukeMathWalker: Setting `b->pos` to `offset` is what one would reasonably expect from a seek operation.
The weirdness, here, comes from the line below:

```c
b->buf->offset = offset;
```

where we basically "forget" that we (potentially) had initialized elements after the offset we are seeking to.
Changing this has implications that echo throughout the codebase, so I've avoided bundling it in this PR.

src/varint.c
```diff
@@ -64,11 +65,9 @@ size_t WriteVarint(uint32_t value, BufferWriter *w) {
 
   size_t mem_growth = 0;
   // we assume buffer reserve will not fail
-  if (!!(mem_growth = Buffer_Reserve(w->buf, nw))) {
-    w->pos = w->buf->data + w->buf->offset;
-  }
+  mem_growth = Buffer_Reserve(w->buf, nw);
```
GuyAv46: Nitpick: remove the declaration of `mem_growth` and declare it in this line

LukeMathWalker: Done in d7496765b

PR #6248, opened 2025-06-03
Refactor Decoders API

GuyAv46:
## Split from #6219

Cherry picked from commit 621675ba742278fe00833fbfaeac0317c56adf0d
Related to MOD-8848

1. decoders API
2. Introducing `IndexBlockReader`

#### Mark if applicable

### DIFF
--- a/src/inverted_index.c
+++ b/src/inverted_index.c
@@ -525,7 +525,7 @@ static void IndexReader_AdvanceBlock(IndexReader *ir) {
  ******************************************************************************/
 
 #define DECODER(name) \
-  static bool name(BufferReader *br, const IndexDecoderCtx *ctx, RSIndexResult *res, t_docId offset)
+  static bool name(IndexBlockReader *br, const IndexDecoderCtx *ctx, RSIndexResult *res)
 
 /**
  * Skipper implements SkipTo. It is an optimized version of DECODER which reads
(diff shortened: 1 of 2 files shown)


src/inverted_index.h
```diff
@@ -51,6 +51,11 @@ typedef struct InvertedIndex {
   };
 } InvertedIndex;
 
+typedef struct IndexBlockReader {
+  BufferReader br;
+  t_docId curOffset;
```
JoanFM: Could u add a comment on what this `curOffset`represents?

I understand is the translation from the inner `br->pos` into the offset in the Buffer as "DocIDs" and not as "bytes"?

src/inverted_index.c
```diff
@@ -525,52 +525,51 @@ static void IndexReader_AdvanceBlock(IndexReader *ir) {
  ******************************************************************************/
 
 #define DECODER(name) \
-  static bool name(BufferReader *br, const IndexDecoderCtx *ctx, RSIndexResult *res, t_docId offset)
```
JoanFM, line 528: can we name it block reader? all these `br.br`seems confusing

src/inverted_index.h
JoanFM: Looking at the code I would call it `curDocId`

GuyAv46: Our inverted indexes don't store the docIDs in their 64-bit representation, but the deltas from some known value. Usually, it's the delta from the previous ID, and most of the decoders will set it to the currently extracted ID as a preparation for the next read. But this is not always the case; The "RawDocIds" stores the offsets from the first ID in the block, and it doesn't update the offset on every read.
It will be incorrect to call it `curDocId` in this case. In general, this value is the current offset (or delta) to extract the current id from the encoded entry

JoanFM: would Delta be a valid one? I see is not the DocID but is the info needed foe the DocId. but the fact that even when the Reader is instantiated it takes the lastid from a previous block, indicates to me that Offset seems a little missleading?

GuyAv46: I think offset is a better name because the delta is the encoded data that we need to extract and add to the offset to get the ID. Offset is also a term that is used throughout the code base

JoanFM (approved):

src/inverted_index.h
GuyAv46: On second thought, "offset" is not the right term. Renamed to `curBaseID`

src/inverted_index.c
```diff
@@ -971,7 +974,13 @@ static bool IndexReader_ReadWithSeeker(IndexReader *ir, t_docId docId) {
   bool found = false;
   while (!found) {
     // try and find docId using seeker
-    found = ir->decoders.seeker(&ir->br, &ir->decoderCtx, ir, docId, ir->record);
+    IndexBlockReader reader = (IndexBlockReader){
+      .buffReader = ir->br,
+      .curBaseId = (ir->decoders.decoder != readRawDocIdsOnly) ? ir->lastId : IR_CURRENT_BLOCK(ir).firstId,
+    };
+    found = ir->decoders.seeker(&reader, &ir->decoderCtx, docId, ir->record);
+    ir->br = reader.buffReader;
```
JoanFM, line 982: just for my education, this is because the reader may have "moved" to another block right?

JoanFM: Wouldn't perhaps make more sense for the `IndexReader` to actually have an IndexBlockReader instead of  a BufferReader as member?

GuyAv46: No, each block has a different buffer, so a `BuffReader` will point to the same block in this case.
We copy the buffer state as it was changed (advanced) in the seeker callback, and we want `ir->br` to be in the same state.
The better thing to do is to have a `BlockReader` as part of the IndexReader struct, but instead of doing that, we do it in the refactor #6219

JoanFM: I do not see the IndexReader contain a BlockReader in the PR now?

PR #6250, opened 2025-06-03
MOD-9917 MOD-9918 initial rust result processor bindings and infrastructure

JonasKruckenberg:

This PR defines the `ResultProcessor` trait and implements basic infrastructure required for writing result processors in Rust. To make reviewing easier, this PR *does not* include any result processor implementations and *does not* impact the redisearch binary at all.

#### Mark if applicable


src/redisearch_rs/result_processor/src/ffi.rs
```diff
@@ -0,0 +1,236 @@
...
+
+        match me.result_processor.next(cx, res) {
+            Ok(Some(())) => RPStatus::RS_RESULT_OK,
+            Ok(None) => RPStatus::RS_RESULT_EOF,
+            Err(Error::Paused) => RPStatus::RS_RESULT_PAUSED,
+            Err(Error::TimedOut) => RPStatus::RS_RESULT_TIMEDOUT,
+            Err(Error::Error) => RPStatus::RS_RESULT_ERROR,
+        }
+    }
+
```
DarthB:
```
/// Frees a result processor that Rust has allocated
/// 
/// Safety-Invariant:
/// 1. Caller ensures that the `ptr` has been allocated by Rust, e.g., via [RPCounter_New] or a similar method.
```

src/redisearch_rs/result_processor/src/ffi.rs
```diff
@@ -0,0 +1,236 @@
...
+    unsafe extern "C" fn result_processor_free(ptr: *mut Header) {
+        debug_assert!(ptr.is_aligned());
+
+        // Safety: This function is called through the ResultProcessor "VTable" which - through the generics
+        //  and constructor of this type - ensures that we can safely cast to `Self` here.
+        //  For all other safety guarantees we have to trust the QueryIterator implementation to be correct.
```
DarthB:
```suggestion
        // Safety: 
        // 1. Guarantees that we can cast to Self here, because we ensure with [ResultProcessor<P>]
        // to support the "VTable"-like Layout that prepends the header to the concrete fields of a ResultProcessor
        // implementation, i.e. we trust C to call this on the correct `ptr`.
```

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        // Safety: Refer to the pinning comment of ffi::ResultProcessor for why
+        // this necessary and reasonably safe.
+        let result_processor = unsafe { Pin::new_unchecked(result_processor) };
+
+        Some(Upstream { result_processor })
+    }
+}
+
+/// The previous result processor in the pipeline
```
DarthB: add why we use the new-type in the comments

src/redisearch_rs/result_processor/src/ffi.rs
```diff
@@ -0,0 +1,236 @@
...
+    ///
+    /// The returned `Box` will own the raw pointer, in particular dropping the `Box`
+    /// will deallocate the `ResultProcessor`. This function should only be used by the [`Self::result_processor_free`].
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure the pointer was previously created through [`Self::into_ptr`]. Furthermore,
+    /// callers have to be careful to never call this method twice for the same pointer, otherwise a double-free
+    /// or other memory corruptions will occur.
+    /// The caller *must* also ensure that `ptr` continues to be treated as pinned.
```
DarthB:
```suggestion
    /// 1. The caller must ensure the pointer was previously created through [`Self::into_ptr`]. 
    /// 2. The caller must never call this method twice for the same pointer or free the pointer.
    /// 3. The caller must also ensure that `ptr` continues to be treated as pinned.
```

src/redisearch_rs/result_processor/src/ffi.rs
```diff
@@ -0,0 +1,236 @@
...
+/// Properties of Result Processors that are accessed by C code through FFI. This type is named `Header` because it
+/// must be the first member of the [`ResultProcessor`] struct in order to guarantee that we can cast a `*mut ResultProcessor<P>`
+/// to a `*mut Header` (which is what the C code expects to receive).
+///
+/// # Pinning Safety Contract
+///
+/// Result processors intrusively linked, where one result processor has the pointer to the
+/// previous (forming an intrusively singly-linked list). This places a big safety invariant on all code
+/// that touches this data structure: Result processors must *never* be moved while part of a QueryIterator
+/// chain. Accidentally moving a processor will remove in a broken link and undefined behaviour.
```
DarthB:
```suggestion
/// chain. Accidentally moving a processor will create a broken link and undefined behaviour.
```

src/redisearch_rs/result_processor/src/ffi.rs
```diff
@@ -0,0 +1,236 @@
...
+type ResultProcessorNext =
+    unsafe extern "C" fn(*mut Header, res: *mut MaybeUninit<SearchResult>) -> RPStatus;
+type ResultProcessorFree = unsafe extern "C" fn(*mut Header);
+
+/// Properties of Result Processors that are accessed by C code through FFI. This type is named `Header` because it
+/// must be the first member of the [`ResultProcessor`] struct in order to guarantee that we can cast a `*mut ResultProcessor<P>`
+/// to a `*mut Header` (which is what the C code expects to receive).
+///
+/// # Pinning Safety Contract
+///
```
DarthB: I would start with a short sentence explaining pinning as that may not be known by everyone, e.g.
```
/// In Rust, "pinning" a value prevents it from being moved in memory by the compiler. 
/// This is crucial for scenarios like LinkedList. From C perspective it prohibits `memcpy` operations.
```

src/redisearch_rs/result_processor/src/ffi.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// The caller must ensure the pointer was previously created through [`Self::into_ptr`]. Furthermore,
+    /// callers have to be careful to never call this method twice for the same pointer, otherwise a double-free
+    /// or other memory corruptions will occur.
+    /// The caller *must* also ensure that `ptr` continues to be treated as pinned.
+    #[inline]
+    unsafe fn from_ptr(ptr: *mut Self) -> Pin<Box<Self>> {
+        // This function must be kept in sync with `Self::into_ptr` above.
+
+        // Safety: The caller has to ensure they never call this function twice in a memory-unsafe way. But since this
+        // is not public anyway, it should be easier to verify.
```
DarthB:
```suggestion
        // Safety: 
        // 1 -> We know Rust is the owner and therefore a Drop via `Box` is good
        // 2 -> We never create two Boxes and thus a double free in Rust
```

src/redisearch_rs/result_processor/src/ffi.rs
```diff
@@ -0,0 +1,236 @@
...
+        let b = unsafe { Box::from_raw(ptr.cast()) };
+        // Safety: Also responsibility of the caller.
```
DarthB:
```suggestion
        // Safety: 3-> Caller has to uphold the pin contract
```

src/redisearch_rs/result_processor/src/ffi.rs
```diff
@@ -0,0 +1,236 @@
...
+        unsafe { Pin::new_unchecked(b) }
+    }
+
+    unsafe extern "C" fn result_processor_next(
```
DarthB: Use numbered invariants as method documentation and adapt the Safety comments internally like in other review comments.

DarthB (dismissed):
Please re-add the `RPCounter` Rust implementation as an example.

src/redisearch_rs/result_processor/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "result_processor"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[dependencies]
+pin-project-lite = "0.2"
```
LukeMathWalker: Let's use workspace dependencies everywhere.

src/redisearch_rs/result_processor/src/ffi.rs
```diff
@@ -0,0 +1,236 @@
+/*
```
LukeMathWalker: For types/signatures whose source of truth is defined in `C` (e.g. ResultProcessorType) we should prefer using the definition generated by `bindgen` rather than hand-writing it.
It should be enough to add the specific header you need to the build script for the `ffi` crate. See https://github.com/RediSearch/RediSearch/blob/d4ad39fb210e6695979fe912039fa22efdc48d43/src/redisearch_rs/ffi/build.rs#L40

src/redisearch_rs/result_processor/src/ffi.rs
```diff
@@ -0,0 +1,236 @@
...
+    free: Option<ResultProcessorFree>,
+
+    // ResultProcessors *must* be !Unpin to ensure they are not moved about breaking the linked list.
+    _unpin: PhantomPinned,
+}
+
+pin_project! {
+    #[derive(Debug)]
+    #[repr(C)]
+    pub struct ResultProcessor<P> {
```
LukeMathWalker: Style comment/question: can we avoid using the same name for both the trait and the struct? 🤔

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Pull the next [`SearchResult`] from this result processor into the provided `res` location.
+    ///
+    /// Returns `Ok(Some(()))` if a search result was successfully pulled from the processor
+    /// and `Ok(None)` to indicate the end of search results has been reached.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Err(_)` for expectional error cases.
+    pub fn next(&mut self, res: &mut MaybeUninit<ffi::SearchResult>) -> Result<Option<()>, Error> {
+        let next = self.result_processor.next.unwrap();
```
LukeMathWalker: Let's add an `expect` here.
It shouldn't happen, but if it happens...

src/redisearch_rs/result_processor/src/ffi.rs
```diff
@@ -0,0 +1,236 @@
...
+                ty: P::TYPE,
+                giltime: timespec {
+                    tv_sec: 0,
+                    tv_nsec: 0,
+                },
+                next: Some(Self::result_processor_next),
+                free: Some(Self::result_processor_free),
+                _unpin: PhantomPinned,
+            },
+            result_processor,
```
LukeMathWalker: Let's explain why it's OK for `parent` and `upstream` to be set to a null pointer.

src/redisearch_rs/result_processor/src/ffi.rs
JonasKruckenberg: Yeah I totally agree, I just dislike the code it generates :/ In particular (though I'm really not sure its 100% necessary for our case) I would like to have `PhantomPinned` on the `ResultProcessor` struct to prevent as much funny aliasing problems as possible (see https://github.com/rust-lang/rust/issues/63818)

LukeMathWalker: I think it's fine to make an exception where manual modelling can give us a benefit (e.g. `ResultProcessor`).
But other than that, I'd stick to that policy.

src/redisearch_rs/result_processor/src/ffi.rs
JonasKruckenberg: I agree it's unfortunate, do you have a better name idea? `FFIResultProcessorWrapper`?

src/redisearch_rs/result_processor/Cargo.toml
JonasKruckenberg: changed 👍🏻

src/redisearch_rs/result_processor/src/ffi.rs
JonasKruckenberg: Changed the crate to use generated bindings for all types except the `Header`

src/redisearch_rs/result_processor/src/ffi.rs
LukeMathWalker: `ResultProcessorEnvelope`? I don't know 🤷🏻

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+    /// "VTable" function. Frees the processor and any internal data related to it.
+    free: Option<unsafe extern "C" fn(self_: *mut Header)>,
+    /// ResultProcessor *must* be !Unpin to ensure they can never be moved, and they never receive
+    /// LLVM `noalias` annotations; See <https://github.com/rust-lang/rust/issues/63818>.
+    /// FIXME: Remove once <https://github.com/rust-lang/rust/issues/63818> is closed and replace with the recommended fix.
+    _unpin: PhantomPinned,
+}
+
+// Header duplicates ffi::ResultProcessor, so we need to make sure it has the exact same size, alignment and field layout.
+const _: () = {
```
LukeMathWalker: Let's move this to a test module.

JonasKruckenberg: sure thing

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use crate::ResultProcessor;
+
+#[derive(Debug)]
+pub struct Counter {
```
LukeMathWalker:
```suggestion
/// A processor to track the number of entries yielded by the previous processor in the chain.
pub struct Counter {
```

src/redisearch_rs/result_processor/src/ffi.rs
JonasKruckenberg: Chose `ResultProcessorWrapper` should be alright

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,50 @@
...
+
+impl ResultProcessor for Counter {
+    const TYPE: ffi::ResultProcessorType = ffi::ResultProcessorType_RP_COUNTER;
+
+    fn next(
+        &mut self,
+        mut cx: crate::Context,
+        res: &mut std::mem::MaybeUninit<ffi::SearchResult>,
+    ) -> Result<Option<()>, crate::Error> {
+        let mut upstream = cx.upstream().unwrap();
```
LukeMathWalker:
```suggestion
        let mut upstream = cx.upstream().expect("There is no processor upstream of this counter.");
```

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,50 @@
...
+
+        while upstream.next(res).is_ok() {
+            self.count += 1;
+
+            // Safety: C function call
```
LukeMathWalker: We'll have to do better than this 😁

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// Execution paused due to rate limiting (or manual pause from ext. thread??)
+    Paused,
+    /// Execution halted because of timeout
+    TimedOut,
+    /// Aborted because of error. The QueryState (parent->status) should have
+    /// more information.
+    Error,
+}
+
+/// This is the main trait that Rust result processors need to implement
```
LukeMathWalker:
```suggestion
///  Result processors are in charge of transforming the entries retrieved from the inverted index when
/// trying to fulfil a user-provided query.
///
/// The processing kind is varied—e.g. it may entail scoring, filtering, sorting, paginating, etc.
///
/// # Structure
/// 
/// Result processors are structured as a chain, assembled by the query planner.
/// Processing is lazy: when the last processor in the chain is asked to yield a result, it
/// will in turn ask for entries from the previous processor, recursively until it reached
/// the beginning of the chain.
/// At the head of the chain, you will always find an index iterator, yielding entries from
/// the database indexes.
///
/// # Result storage
///
/// The processing output is accumulated in a [`SearchResult`] object.
```

As a general rule of thumb, assume that the C code will be gone once we are done and write docs as if the Rust code is the only documentation source.

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,357 @@
...
+        // to document data. Callers *MUST* ensure they are eventually freed.
+        next: Option<unsafe extern "C" fn(self_: *mut Header, res: *mut ffi::SearchResult) -> c_int>,
+        // "VTable" function. Frees the processor and any internal data related to it.
+        free: Option<unsafe extern "C" fn(self_: *mut Header)>,
+        // ResultProcessor *must* be !Unpin to ensure they can never be moved, and they never receive
+        // LLVM `noalias` annotations; See <https://github.com/rust-lang/rust/issues/63818>.
+        // FIXME: Remove once <https://github.com/rust-lang/rust/issues/63818> is closed and replace with the recommended fix.
+        _unpin: PhantomPinned,
+    }
+}
```
LukeMathWalker: What's stopping us from moving the source of truth for this type from C to Rust? 
As long as we don't change its definition, we should be in a good place.

The `_unpin` field will just disappear when `cbindgen` does its work.

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,94 @@
...
+            .upstream()
+            .expect("There is no processor upstream of this counter.");
+
+        while upstream.next(res)?.is_some() {
+            self.count += 1;
+
+            // Safety: This function should only be called on initialized SearchResults. Luckily,
+            // a ResultProcessor returning `RPStatus_RS_RESULT_OK` means "Result is filled with valid data"
+            // so it is safe to call this function inside this loop.
+            #[cfg(not(test))] // can we link Rust unit tests against the redis code?
```
LukeMathWalker: We can define the symbol in our test binary, usually to a mocking implementation. 
Check out https://github.com/RediSearch/RediSearch/pull/6257/files#diff-7b00489cf4d31bb840014ef87237cfe2f519b817ebd2207285fb5d8b61b47d5cR365-R383 as a reference example.

That's why, generally speaking, we prefer to have tests under the `tests` folder rather than colocated via a `mod tests`.

JonasKruckenberg:
@Itzikvaknin without looking at the actual sorter impl too much, here is a rough sketch for how you'd implement a "two-phase" RP (or any state machine really) in Rust. The trick is to basically define all possible stages of the state machine as an `enum` and then in the `next` method you match on that enum to determine what logic to run.

If you want to switch the state of the state machine (the phase of the RP) then you simply set the value of the `state` field. Also note that you can ofc (and likely would for real RPs) factor out the actual implementation of each phase into its own method.


```rust
struct NotSorter {
    shared_data: u32,
    state: NotSorterState,
}

enum NotSorterState {
    // We can have variant, specific state too
    Accum { state: u64 },
    Yield { other_state: u32 },
}

impl NotSorter {
    pub fn new() -> Self {
        Self {
            shared_data: 0,
            // The sorter starts out accumulating search results
            state: NotSorterState::Accum { state: 0 },
        }
    }
}

impl ResultProcessor for NotSorter {
    const TYPE: ffi::ResultProcessorType = ffi::ResultProcessorType_RP_COUNTER;

    fn next(
        &mut self,
        mut cx: crate::Context,
        res: &mut ffi::SearchResult,
    ) -> Result<Option<()>, crate::Error> {
        let mut upstream = cx
            .upstream()
            .expect("There is no processor upstream of this counter.");

        // if the impls for Accum and Yield grow too complex, they can be factorered out into methods too...
        match self.state {
            NotSorterState::Accum { state } => {
                // TODO accumulation & sorting logic here:
                while upstream.next(res)?.is_some() {
                    // TODO insert the search result into the minmax heap
                }

                // after we accumulated all values & inserted them into the minmax heap
                // we want to switch to the second "yield" phase.
                // We can do so by simply changing the value of the `state` field since
                // that is what we matched on above
                // Note that we can pass variant specific state here if we wanted to (rn we just pass 46)
                self.state = NotSorterState::Yield { other_state: 46 };
            }
            NotSorterState::Yield { other_state } => {
                // TODO do the yielding here
            }
        }

        todo!()
    }
}
```

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: I don't think this is necessary or good to include since it would be just talking about internal implementation details.

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: Nothing, but IMO we should do this in a follow up PR once the Counter RP and associated FFI machinery is merged too.

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: addressed

src/redisearch_rs/result_processor/src/ffi.rs
JonasKruckenberg: I think the explanation is fine as-is now.

src/redisearch_rs/result_processor/src/counter.rs
JonasKruckenberg: The pin projection type sadly cannot be made `pub` so the test doesn't work outside its defining crate :/.

src/redisearch_rs/Cargo.lock
```diff
@@ -916,6 +916,26 @@ version = "2.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e3148f5046208a5d56bcfc03053e3ca6334e51da8dfb19b6cdc8b306fae3283e"
 
+[[package]]
```
kei-nan, line 941: what does the pin-project package do?

src/redisearch_rs/ffi/Cargo.toml
```diff
@@ -5,6 +5,9 @@ edition.workspace = true
 license-file.workspace = true
 publish.workspace = true
 
+[lib]
+doctest = false
```
kei-nan, line 9: what is this config?

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+use std::{
+    marker::PhantomPinned,
+    pin::Pin,
+    ptr::{self, NonNull},
+};
+
+/// Errors that can be returned by [`ResultProcessor`]
+#[derive(Debug)]
+pub enum Error {
+    /// Execution paused due to rate limiting (or manual pause from ext. thread??)
```
kei-nan: The question marks at the end don't make me feel confident about this comment :smile: 
Can we verify why this value is needed and update the comment?

JonasKruckenberg: This is literally a copy-paste from the C header file ;)
https://github.com/RediSearch/RediSearch/blob/c7847be2b7f89769b2713d1ec1a8571f89d21a5b/src/result_processor.h#L149
I'm fine with removing it from this change though and adding it back later if necessary.

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+        I: Iterator<Item = SearchResult>,
+    {
+        const TYPE: ffi::ResultProcessorType = ffi::ResultProcessorType_RP_MAX + 1;
+
+        fn next(&mut self, _cx: Context, out: &mut ffi::SearchResult) -> Result<Option<()>, Error> {
+            if let Some(res) = self.iter.next() {
+                *out = res;
+                Ok(Some(()))
+            } else {
+                Ok(None)
```
kei-nan, line 409: This will cause the iteration to stop?
Should this be documented?
Should the user always stop calling next if it got None?
I guess I am a bit worried we lose some information about why a result was not obtained.

src/redisearch_rs/ffi/Cargo.toml
JonasKruckenberg: The `bindgen` tool that is used to generate Rust code from C header files will attempt to keep doc comments present in C. Since Rust will attempt to run all code blocks in doc comments (stuff like
```
/*
* this is a comment with a code snippet
* ```
* ``` 
*/
```
will actually be run as a test case when running `cargo test`)
but the C header file comments weren't written with that in mind ofc, so tests fail and break. Setting `doctest = false` will just instruct Rust to skip running these tests for the `ffi` module.

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+    pub fn upstream(&mut self) -> Option<Upstream<'_>> {
+        // Safety: We have to trust that the upstream pointer set by our QueryIterator parent
+        // is correct.
+        let result_processor = unsafe { self.result_processor.upstream.as_mut()? };
+
+        // Safety: Refer to the pinning comment of ffi::ResultProcessor for why
+        // this necessary and reasonably safe.
+        let result_processor = unsafe { Pin::new_unchecked(result_processor) };
+
+        Some(Upstream { result_processor })
```
kei-nan: When can none be returned?

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: Good point! This is actually a pretty established pattern in Rust ([`Iterator`](https://doc.rust-lang.org/std/iter/index.html)) where `Some(val)` means "here is your value" and `None` means "I have no more values to give you, please don't call me again".

But this should be documented on the `ResultProcessor` trait, assuming everybody knows this isn't great!

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: edit: it appears that `RS_RESULT_PAUSED` is actually not used anywhere in the codebase, so removing should be fine. But I'm happy to defer this decision to you.

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: updated the method description to better explain this 👍🏻

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: updated the method description to better explain this 👍🏻

> I guess I am a bit worried we lose some information about why a result was not obtained.

The mapping should be straightforward without loosing information:
```
  // Result is filled with valid data
Ok(Some()) =>  RS_RESULT_OK

  // Result is empty, and the last result has already been returned.
Ok(None) =>  RS_RESULT_EOF

  // Execution paused due to rate limiting (or manual pause from ext. thread??)
Err(Error::Paused) => RS_RESULT_PAUSED

  // Execution halted because of timeout
Err(Error::Timedout) =>  RS_RESULT_TIMEDOUT

  // Aborted because of error. The QueryState (parent->status) should have
  // more information.
Err(Error::Error) =>  RS_RESULT_ERROR
```

src/redisearch_rs/Cargo.lock
JonasKruckenberg: This is probably more complex than can be explained in a GH comment, but here we go (not gonna assume any knowledge): The chain of result processors forms an intrusive singly linked list where each node in the list stores the pointer to the previous node. The safety of this hinges on one very important invariant though: 
Each node *must* remain in the same memory location so we don't create dangling pointers. 

This easy enough to uphold in C (simply never pass by value) but more difficult in Rust since the compiler is free to move values as it sees fit (Rust calls this [*trivially moveable*](https://doc.rust-lang.org/nightly/std/pin/index.html#what-is-moving)). As a result a Rust *reference* (`&T` or `&mut T`) is a stable "handle" to a value but a pointer (`*const T` or `*mut T`) is not.

For intrusive data types like this we need to tell the compiler "don't move this please I have pointers to it" which is called [*pinning*](https://doc.rust-lang.org/nightly/std/pin/index.html) in Rust. We basically wrap a reference in the `Pin<T>` type (`Pin<&mut T>`) which disallows moving the *pointee* from its location in memory. 

Crucially though, the way `Pin` disallows is *not magic*, it simply doesn't implement any methods and traits that would allow a caller to move the value. Unfortunately this  means banning all mutable access to the value `T` (you cannot get a `&mut T` from a `Pin<&mut T>` for example) since with a `&mut T` you can always move the value very easily (via [`mem::replace`](https://doc.rust-lang.org/nightly/std/mem/fn.replace.html) for example). 

So Rust has another technique that lets you treat memory locations a "pinned" while still being able to mutate them through ["pin projections"](https://blog.yoshuawuyts.com/safe-pin-projections-through-view-types/#what-are-pin-projections) essentially a "proxy type" that behaves like a regular Rust type (including allowing moving & mutation) but will forward all accesses and mutations to the backing, actually pinned type.

The `pin-project` crate is a utility crate that allows us to automate the creation of this "proxy type", but simply writing
```rust
#[pin_project]
struct Pinned {
   a: u64
}
```

src/redisearch_rs/result_processor/src/ffi.rs
JonasKruckenberg: I don't understand this comment

src/redisearch_rs/result_processor/src/ffi.rs
JonasKruckenberg: changed 👍🏻

src/redisearch_rs/result_processor/src/counter.rs
JonasKruckenberg: This comment also doesn't make sense for this PR anymore, but I have fixed it in #6261

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,577 @@
...
+    /// Converts a heap-allocated `ResultProcessor` into a raw pointer.
+    ///
+    /// The caller is responsible for the memory previously managed by the `Box`, in particular
+    /// the caller should properly destroy the `ResultProcessor` and deallocate the memory by calling
+    /// `Self::from_ptr`.
+    ///
+    /// # Safety
+    ///
+    /// The caller *must* continue to treat the pointer as pinned.
+    pub unsafe fn into_ptr(me: Pin<Box<Self>>) -> *mut Self {
```
DarthB, line 271:
```suggestion
    [inline]
    pub unsafe fn into_ptr(me: Pin<Box<Self>>) -> *mut Self {
```

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,577 @@
...
+            Err(Error::Error) => ffi::RPStatus_RS_RESULT_ERROR as c_int,
+        }
+    }
+
+    unsafe extern "C" fn result_processor_free(me: *mut Header) {
+        debug_assert!(me.is_aligned());
+
+        // Safety: This function is called through the ResultProcessor "VTable" which - through the generics
+        //  and constructor of this type - ensures that we can safely cast to `Self` here.
+        //  For all other safety guarantees we have to trust the QueryIterator implementation to be correct.
```
DarthB:
```suggestion
        //  For safety guarantees, like that the pointer points to a valid Result Processor, we have to trust the QueryIterator implementation to be correct.
```

You mean it must use the proper construction methods or something else too?

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,577 @@
...
+}
+
+#[cfg(test)]
+pub(crate) mod test {
+    #![expect(unused)]
+
+    use super::*;
+    use ffi::{RPStatus_RS_RESULT_OK, SearchResult};
+
+    // Header duplicates ffi::ResultProcessor, so we need to make sure it has the exact same size, alignment and field layout.
```
DarthB:
```suggestion
    // Compile time check for Header that duplicates ffi::ResultProcessor: We need to make sure it has the same size, alignment, and field layout.
```

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,577 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
```
DarthB, line 9: We should add a crate-level docs. For now, without the rp_counter example, it could boil down to:
1. Main Usage ResultProcessor Trait
2. Shortly explain the interplay of ResultProcessorWrapper, Header, close with importance of PIN
3. Link Pin and explain what pin_project does
4. Mention compile time check for layout and unit tests

JonasKruckenberg: I don't think the pinning is important here, as it is a transient requirement (we can make it go away when everything is rust) plus one that safe consumer code does not need to worry about whatsoever as it is handled by the library itself. 

But the other comments I think I have addressed?

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: changed 👍🏻

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: I updated the comment

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: updated the comment

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: I removed the error variant for now, we can add it back later when this functionality is actually used

src/redisearch_rs/Cargo.lock
JonasKruckenberg: edit: note that this complexity is only transient (while we have to interop between C and Rust) and completely hidden from RP implementors and consumers (they don't need to worry about pinning at all)

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: I updated the method description too

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,606 @@
...
+    /// will deallocate the `ResultProcessor`. This function should only be used by the [`Self::result_processor_free`].
+    ///
+    /// # Safety
+    ///
+    /// 1. The caller must ensure the pointer was previously created through [`Self::into_ptr`].
+    /// 2. The caller has to be careful to never call this method twice for the same pointer, otherwise a
+    ///    double-free or other memory corruptions will occur.
+    /// 3. The caller *must* also ensure that `ptr` continues to be treated as pinned.
+    #[inline]
+    unsafe fn from_ptr(ptr: *mut Self) -> Pin<Box<Self>> {
```
DarthB, line 291: Inline that method in `result_processor_free` and re-use the documentation

JonasKruckenberg: no, its intentionally kept separate to keep the `from_ptr/into_pr` pair since those methods must be in sync with the types they cast to and from (also see the first comment in each of those methods)

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,606 @@
...
+        // This function must be kept in sync with `Self::into_ptr` above.
+
+        // Safety: The caller has to ensure they never call this function twice in a memory-unsafe way. But since this
+        // is not public anyway, it should be easier to verify.
+        let b = unsafe { Box::from_raw(ptr.cast()) };
+        // Safety: Also responsibility of the caller.
```
DarthB:
```suggestion
        // Safety: 3-> Caller has to uphold the pin contract
```

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,606 @@
...
+
+        match me.result_processor.next(cx, res) {
+            Ok(Some(())) => ffi::RPStatus_RS_RESULT_OK as c_int,
+            Ok(None) => ffi::RPStatus_RS_RESULT_EOF as c_int,
+            Err(Error::TimedOut) => ffi::RPStatus_RS_RESULT_TIMEDOUT as c_int,
+            Err(Error::Error) => ffi::RPStatus_RS_RESULT_ERROR as c_int,
+        }
+    }
+
+    unsafe extern "C" fn result_processor_free(me: *mut Header) {
```
DarthB, line 334:
```suggestion
    /// The entry point for C to free result processors generated by Rust.
    ///
    /// [Header] is equivalent with the C-side `ResultProcessor` struct. For more details see [Self::from_ptr].
    unsafe extern "C" fn result_processor_free(me: *mut Header) {
```

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,606 @@
...
+        // This function must be kept in sync with `Self::into_ptr` above.
+
+        // Safety: The caller has to ensure they never call this function twice in a memory-unsafe way. But since this
+        // is not public anyway, it should be easier to verify.
+        let b = unsafe { Box::from_raw(ptr.cast()) };
+        // Safety: Also responsibility of the caller.
+        unsafe { Pin::new_unchecked(b) }
+    }
+
+    unsafe extern "C" fn result_processor_next(
```
DarthB, line 304:
```suggestion
    /// The entry point for C to trigger the `next` call on result processors implemented in Rust.
    ///
    /// See [Self::from_ptr] for Safety information
    unsafe extern "C" fn result_processor_next(
```

src/redisearch_rs/result_processor/src/ffi.rs
DarthB: I think it's because of the outdated referenced code that it has moved - I don't understand it either anymore.

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,606 @@
...
+    /// 1. The caller must ensure the pointer was previously created through [`Self::into_ptr`].
+    /// 2. The caller has to be careful to never call this method twice for the same pointer, otherwise a
+    ///    double-free or other memory corruptions will occur.
+    /// 3. The caller *must* also ensure that `ptr` continues to be treated as pinned.
+    #[inline]
+    unsafe fn from_ptr(ptr: *mut Self) -> Pin<Box<Self>> {
+        // This function must be kept in sync with `Self::into_ptr` above.
+
+        // Safety: The caller has to ensure they never call this function twice in a memory-unsafe way. But since this
+        // is not public anyway, it should be easier to verify.
```
DarthB:
```suggestion
         // Safety: 
        // 1 -> We know Rust is the owner and therefore a deallocation on the Rust side is sound
        // 2 -> If the caller fulfills this we won't double free here.
```

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -0,0 +1,724 @@
...
+
+        fn next(&mut self, _cx: Context, out: &mut ffi::SearchResult) -> Result<Option<()>, Error> {
+            if let Some(res) = self.iter.next() {
+                *out = res;
+                Ok(Some(()))
+            } else {
+                Ok(None)
+            }
+        }
+    }
```
DarthB, line 412: Remove not used code, store somewhere and readd if needed, e.g. for the Test Entry Mock ticket

PR #6252, opened 2025-06-04
[MOD-9991] Check security advisories and license compatibility for Rust 

LukeMathWalker:

Introduce an additional step to our linting workflow to be notified when:
- There is code with known security vulnerabilities in our dependency tree
- Some dependencies use a license that hasn't been explicitly allowed

To get the scanning tool to work correctly, I've had to:

- Switch from `license` to `license-file` in our manifests, since Redis' license is not in https://spdx.org/licenses/
- Commit the lock file
- Mark all workspace package as `publish = false`

#### Mark if applicable


.github/workflows/task-lint.yml
```diff
@@ -33,11 +33,31 @@ jobs:
         id: fmt
         continue-on-error: true
         run: make fmt CHECK=1
-
+      - uses: EmbarkStudios/cargo-deny-action@v2
```
DvirDukhan, line 36: I would argue this needs a dedicated place, but I don't see any better option

DvirDukhan (commented):
minor comment

.github/workflows/task-lint.yml
LukeMathWalker: We could create a separate workflow, but I don't think it makes a big difference—lint is still one of the fastest to complete and all the checks in here broadly fit into the same category.

DvirDukhan (approved):

DvirDukhan:
@LukeMathWalker what is the risk to backport this one to 8.2?

LukeMathWalker:
> @LukeMathWalker what is the risk to backport this one to 8.2?

Minimal I'd say. We might have to regenerate the lock file, but that's easily handled.

DvirDukhan (approved):

PR #6254, opened 2025-06-04
[MOD-10002] Migrate `RSNumericRecord` to Rust

chesedo:
This moves the `RSNumericRecord` definition from C to Rust. This will make it easier to define `RSIndexResult` in Rust and then eventually to port the inverted index to Rust.

The Rust definition has exactly the same structure as the existing C definition. There no C code needs to be updated.

#### Mark if applicable


LukeMathWalker (approved):

src/redisearch_rs/redisearch_rs/src/types/mod.rs
```diff
@@ -0,0 +1,3 @@
+//! This module contains pure Rust types that we want to expose to C code.
+
+pub use inverted_index::types::RSNumericRecord;
```
zeenix: I'm not sure we want this. Rust users of this API should IMO use it throught the `inverted_index mode` mod. I'd also skip the `types` mod.

src/redisearch_rs/headers/triemap.h
zeenix, line 1: Might be a good idea to already take part of the changes from [my next PR branch](https://github.com/RediSearch/RediSearch/commit/7b41c9d56e14722a4a5c45faf674be61f5660023) (e.g renaming of this header to `redisearch_rs.h`).

zeenix (changes requested):

src/redisearch_rs/redisearch_rs/src/types/mod.rs
chesedo: Wait, what is "this"? I need the re-export in `redisearch_rs` to generate the C version of this. All the Rust crates will use it via `inverted_index` instead (until it is moved to a better place).

chesedo: I like mods to isolate common pieces of code. Is there a good reason to skip it?

src/redisearch_rs/headers/triemap.h
chesedo: Imo we should make another small PR for that if we want it in.

src/redisearch_rs/headers/triemap.h
```diff
@@ -74,6 +74,13 @@ typedef struct LowMemoryThinVecCVoid TrieMapResultBuf;
  */
 typedef void (*TrieMapRangeCallback)(const char*, size_t, void*, void*);
 
+/**
+ * Represents a numeric value in an index record.
+ */
+typedef struct RSNumericRecord {
```
alonre24: Why is that part of triemap.h?

src/redisearch.h
```diff
@@ -17,6 +17,8 @@
 #include "util/dllist.h"
 #include "stemmer.h"
 
+#include "triemap.h" // Import all the types defined in Rust
```
alonre24: Why do we need all the triemap types here rather than just `RSNumericRecord`?

src/redisearch_rs/ffi/src/lib.rs
```diff
@@ -21,7 +21,8 @@
     clippy::multiple_unsafe_ops_per_block,
     clippy::undocumented_unsafe_blocks,
     clippy::missing_safety_doc,
-    clippy::len_without_is_empty
+    clippy::len_without_is_empty,
+    rustdoc::broken_intra_doc_links
```
alonre24, line 25: Why is that needed?

alonre24 (commented):
Few questions

src/redisearch_rs/headers/triemap.h
LukeMathWalker: Renaming is necessary, since exposing `RSNumericRecord` via `triemap.h` is... weird 😁

chesedo: Fair, I'll just add it to this PR. Ideally we should get small stuff like this in as soon as we know about them.

src/redisearch_rs/headers/triemap.h
chesedo: This is the same a Zeeshan's comment. Basically this file has the wrong name. I'll give it a better name in this PR

src/redisearch_rs/ffi/src/lib.rs
chesedo: The changes in #6251 are also trying to check the doc links in FFI types. And these can be invalid for the FFI types since the `ffi` crate does not dependent on the `redisearch_rs` where these links might link to. So this opts out of the check only for this crate.

src/redisearch_rs/Cargo.toml
```diff
@@ -1,11 +1,11 @@
 [workspace]
 members = [
+    "entrypoint/*",
```
LukeMathWalker: Perhaps `c_entrypoint` is a better name, thinking about it now.

src/redisearch.h
```diff
@@ -16,6 +16,7 @@
 #include <time.h>
 #include "util/dllist.h"
 #include "stemmer.h"
+#include "inverted_index_rs.h"
```
LukeMathWalker: Looking at where the type was defined in C, we probably want a `type_rs.h` rather than `inverted_index_rs.h` for these specific definitions that are not used exclusively by the inverted index.

src/redisearch_rs/headers/triemap.h
chesedo: Done, it is now under `types_rs.h` (and the old `triemap.h` still exists too)

src/redisearch_rs/headers/triemap.h
chesedo: I've moved this to `types_rs.h`

src/redisearch_rs/Cargo.toml
chesedo: Done https://github.com/RediSearch/RediSearch/pull/6254/commits/1091bc3a89200a0890ce40b7172c9bf69cc5bb3c

src/redisearch.h
chesedo: Done https://github.com/RediSearch/RediSearch/pull/6254/commits/7a1eca49f26fd6db9f86b555c67ed85bf7b5e86f

src/redisearch.h
chesedo: This is now `types_rs.h` which will only contain types exposed by Rust

alonre24 (approved):
Not sure I'm following all the changes in the project structure and their purpose, but overall looks OK to me. We can discuss it later on.

src/redisearch_rs/redisearch_rs/src/types/mod.rs
zeenix: > I like mods to isolate common pieces of code. Is there a good reason to skip it?

I don't see the need to isolate types from the rest. In Rust everything is a type.

zeenix (changes requested):
I disagree with the `types` module still.

src/redisearch_rs/redisearch_rs/src/types/mod.rs
chesedo: I removed / flattened it

zeenix (approved):

src/redisearch_rs/ffi/build.rs
```diff
@@ -33,7 +33,8 @@ fn main() {
         let redis_modules = root.join("deps").join("RedisModulesSDK");
         let src = root.join("src");
         let deps = root.join("deps");
-        [redis_modules, src, deps]
+        let redisearch_rs = root.join("src").join("redisearch_rs").join("headers");
```
chesedo, line 36: Can I remove this?

chesedo: No, this is still needed. In this file we are making Rust "bindings" for the `redisearch.h` C code. But `redisearch.h` has an include for `types_rs.h` with this PR. This line adds the `src/redisearch_rs/headers` to the C include path to correctly resolve `types_rs.h`

chesedo: image

PR #6255, opened 2025-06-04
[MOD-9999] Migrate `RSAggregateResult` to Rust

chesedo:
This moves the `RSAggregateResult` definition to Rust in preparation of moving the `RSIndexResult` to Rust too. This will make it easier to port the inverted index to Rust.

The Rust definition is exactly the same as the old C definition, so nothing in the C code needs to be updated.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/headers/types_rs.h
+++ b/src/redisearch_rs/headers/types_rs.h
@@ -13,6 +13,28 @@ typedef struct RSQueryTerm RSQueryTerm;
 
 
 /**
+ * Represents an aggregate array of values in an index record.
+ */
+typedef struct RSAggregateResult {
+  /**
+   * The number of child records
+   */
+  int numChildren;
+  /**
+   * The capacity of the records array. Has no use for extensions
+   */
+  int childrenCap;
+  /**
+   * An array of records
+   */
+  void **children;
+  /**
+   * A map of the aggregate type of the underlying records
+   */
+  uint32_t typeMask;
+} RSAggregateResult;
+
+/**
  * Represents a numeric value in an index record.
  */
 typedef struct RSNumericRecord {
--- a/src/redisearch_rs/headers/types_rs.h
+++ b/src/redisearch_rs/headers/types_rs.h
@@ -17,15 +17,139 @@ typedef struct RSQueryTerm RSQueryTerm;
 typedef struct RSIndexResult RSIndexResult;
 
 
-typedef enum RSResultType {
-  Union = 1,
-  Intersection = 2,
-  Term = 4,
-  Virtual = 8,
-  Numeric = 16,
-  Metric = 32,
-  HybridMetric = 64,
-} RSResultType;
+enum RSResultType
+#ifdef __cplusplus
+  : uint32_t
+#endif // __cplusplus
+ {
+  RSResultType_Union = 1,
+  RSResultType_Intersection = 2,
+  RSResultType_Term = 4,
+  RSResultType_Virtual = 8,
+  RSResultType_Numeric = 16,
+  RSResultType_Metric = 32,
+  RSResultType_HybridMetric = 64,
+};
+#ifndef __cplusplus
+typedef uint32_t RSResultType;
+#endif // __cplusplus
+
+/**
+ * Represents a set of flags of some type `T`.
+ * `T` must have the `#[bitflags]` attribute applied.
+ *
+ * A `BitFlags<T>` is as large as the `T` itself,
+ * and stores one flag per bit.
+ *
+ * ## Comparison operators, [`PartialOrd`] and [`Ord`]
+ *
+ * To make it possible to use `BitFlags` as the key of a
+ * [`BTreeMap`][std::collections::BTreeMap], `BitFlags` implements
+ * [`Ord`]. There is no meaningful total order for bitflags,
+ * so the implementation simply compares the integer values of the bits.
+ *
+ * Unfortunately, this means that comparing `BitFlags` with an operator
+ * like `<=` will compile, and return values that are probably useless
+ * and not what you expect. In particular, `<=` does *not* check whether
+ * one value is a subset of the other. Use [`BitFlags::contains`] for that.
+ *
+ * ## Customizing `Default`
+ *
+ * By default, creating an instance of `BitFlags<T>` with `Default` will result
+ * in an empty set. If that's undesirable, you may customize this:
+ *
+ * ```
+ * # use enumflags2::{BitFlags, bitflags};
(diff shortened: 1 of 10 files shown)


src/redisearch_rs/redisearch_rs/cbindgen.toml
```diff
@@ -4,6 +4,13 @@ autogen_warning = "/* Warning, this file is autogenerated by cbindgen from `src/
 cpp_compat = true
 pragma_once = true
 
+after_includes = """
+/**
+ * Forward declaration of RSIndexResult. It will be correctly defined in `redisearch.h`
```
LukeMathWalker:
```suggestion
 * Forward declaration of RSIndexResult. It will be defined in `redisearch.h`
```

src/redisearch_rs/inverted_index/src/types/mod.rs
```diff
@@ -7,7 +7,27 @@
...
+
+/// Represents an aggregate array of values in an index record.
+/// cbindgen:rename-all=CamelCase
+#[repr(C)]
+pub struct RSAggregateResult {
+    /// The number of child records
+    pub num_children: c_int,
+
+    /// The capacity of the records array. Has no use for extensions
+    pub children_cap: c_int,
```
LukeMathWalker: Do we need to keep it as a `c_int`? Considering what we are tracking, I would assume that a `u16` would do just fine?

src/redisearch_rs/inverted_index/src/types/mod.rs
```diff
@@ -7,7 +7,27 @@
...
+
+    /// An array of records
+    pub children: *mut *mut RSIndexResult,
+
+    /// A map of the aggregate type of the underlying records
+    pub type_mask: u32,
```
LukeMathWalker: This is not "just" a `u32`.
It's a bitflags, and we should model it as such. `bindgen` should then be able to translate it to a "plain" u32 over the FFI boundary.

src/redisearch_rs/inverted_index/src/types/mod.rs
chesedo: I would say yes for the time being. We need the C output to be `int` (to match the old declaration) since it can be 16 or 32 bits depending on the compiling machine. Using `i16` / `u16` could therefore change the alignment

LukeMathWalker: I know that it would be a change—my argument here is that it should not be an `int` in C either. I'm also OK with doing the change in another PR if we want to keep this a pure refactoring.

chesedo: I see. Let me make another PR to change the C to always be a `u16`. Would like to keep this one focused on the refactor

src/redisearch_rs/inverted_index/src/types/mod.rs
chesedo: I could not get the `bitflags` type to be exported by `cbindgen` for this. So just used an enum instead

src/redisearch_rs/inverted_index/src/types/mod.rs
```diff
@@ -7,8 +7,45 @@
...
+use std::ffi::c_int;
+
 /// Represents a numeric value in an index record.
 /// cbindgen:field-names=[value]
 #[allow(rustdoc::broken_intra_doc_links)] // The field rename above breaks the intra-doc link
 #[repr(C)]
 pub struct RSNumericRecord(pub f64);
+
+#[bitflags]
+#[repr(u8)]
```
LukeMathWalker: This changes the representation from `u32` to `u8`. Is that something we're happy with?
I'd say _yes_, since `u8` is enough, but double-checking.

chesedo: This should be `u32`. I'll update it. I'll make all the field size updated in another PR...

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -57,6 +58,39 @@ pub struct RSTermRecord {
...
+/// cbindgen:rename-all=CamelCase
+#[repr(C)]
+pub struct RSAggregateResult {
+    /// The number of child records
+    pub num_children: c_int,
+
+    /// The capacity of the records array. Has no use for extensions
+    pub children_cap: c_int,
+
+    /// An array of records
```
zeenix, line 87: An array of **pointers to** records?

zeenix (dismissed):
LGTM otherwise.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo:
```suggestion
    /// An array of pointers to records
```

chesedo: I updated the comment https://github.com/RediSearch/RediSearch/pull/6255/commits/e04cf6b12c9698037e3010cafbfff48cfdad7dec

src/redisearch_rs/headers/types_rs.h
```diff
@@ -11,6 +11,167 @@
...
+ * Ideally, that would be the case. However, because `const fn`s cannot
+ * have trait bounds in current Rust, this would prevent us from providing
+ * most `const fn` APIs. As a workaround, we define `BitFlags` with two
+ * type parameters, with a default for the second one:
+ *
+ * ```ignore
+ * struct BitFlags<T, N = <T as BitFlag>::Numeric> {
+ *     value: N,
+ *     marker: PhantomData<T>,
+ * }
```
alonre24, line 140: This practice is not needed for us in `RSResultTypeMask`?

zeenix: @alonre24 this is generated code from Rust and these docs are verbatim copied by the cbindgen tool. @chesedo and I agreed last evening that it'd be great to disable this copying and he'll look into that today.

chesedo: Yeah, we probably won't use any `const` functions. Note that this documentation comes from the `enumflags2` crate via its `BitFlags` type.

It's coming from this line

https://github.com/RediSearch/RediSearch/blob/e04cf6b12c9698037e3010cafbfff48cfdad7dec/src/redisearch_rs/inverted_index/src/lib.rs#L74-L76

chesedo: Turning it off is also a no go. `cbindgen` only allows turning all documentation off and not the documentation of a single type.

alonre24 (dismissed):

zeenix (approved):

PR #6256, opened 2025-06-04
Migrate `RSTermRecord` to Rust

chesedo:
Migrates the `RSTermRecord` to Rust. This will allow `RSIndexResult` to be migrated to Rust which will make it easier to implement the inverse index in Rust.

The structure of the Rust `RSTermRecord` is exactly the same as the old C definition. So nothing needs to be updated in the C code.

#### Mark if applicable


JonasKruckenberg (approved):

src/redisearch_rs/headers/types_rs.h
zeenix, line 1: I had assumed that we didn't go for this whole "putting types" in a module. I really think, we need to split wrt modules/API. So everything related to Inverted Index should be in a crate/header named `encode_decode` or `inverted_index`.

zeenix (commented):
LGTM otherwise.

src/redisearch.h
```diff
@@ -234,7 +234,7 @@ typedef int (*RSQueryTokenExpander)(RSQueryExpanderCtx *ctx, RSToken *token);
 typedef void (*RSFreeFunction)(void *);
 
 /* A single term being evaluated in query time */
-typedef struct {
+typedef struct RSQueryTerm {
```
alonre24, line 237: Not migrating `RSQueryTerm` as well? why?

src/redisearch_rs/headers/types_rs.h
chesedo: Yeah, that might be the final result. In the intermediate we can't have two `inverted_index.h` files (one of the old C and another for the bits that have migrated to Rust)

zeenix: Sure. I'll resolve this. It's not something to address in this PR anyway.

src/redisearch.h
chesedo: It's used behind a pointer and not needed for the inverted index work. So it's is left out to keep the change simple

alonre24 (approved):

PR #6257, opened 2025-06-04
[MOD-10008] Reduce the amount of unsafe code by using more idiomatic Rust definitions

LukeMathWalker:

Correctness is our paramount concern when manipulating C types in Rust.
When defining idiomatic Rust wrappers, we are caught between a rock and a hard place: 
- If we create thin new types around the generated bindings (in the `ffi` crate) we get layout correctness for free. If the C code changes, the bindings will change and any discrepancy will cause a compilation error in Rust. At the same time, we're working with raw pointers everywhere, thus forcing us to use a lot of `unsafe` code where mistakes can be made more easily.
- If we use a "Rust-native" definition, we reduce the amount of `unsafe` code but we risk layout drift.

This PR tries to strike a balance for the `buffer.*` module.
We stick to wrapping the `ffi` type for `Buffer`, but we use more idiomatic Rust types for reader and writer. We add compile-time assertions to guard against layout drift.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/buffer/src/reader.rs
+++ b/src/redisearch_rs/buffer/src/reader.rs
@@ -73,6 +73,9 @@ impl<'a> std::io::Read for BufferReader<'a> {
 }
 
 // Check, at compile-time, that `BufferReader` and `ffi::BufferReader` have the same representation.
+// This check will alert us if the C or the Rust definition changed without a corresponding patch
+// to the representation in the other language.
+// In that scenario, you should make sure to align both on the same memory layout.
 const _: () = {
     use std::mem::offset_of;
 
--- a/src/redisearch_rs/buffer/src/writer.rs
+++ b/src/redisearch_rs/buffer/src/writer.rs
@@ -106,6 +106,9 @@ impl<'a> std::io::Write for BufferWriter<'a> {
 }
 
 // Check, at compile-time, that `BufferWriter` and `ffi::BufferWriter` have the same representation.
+// This check will alert us if the C or the Rust definition changed without a corresponding patch
+// to the representation in the other language.
+// In that scenario, you should make sure to align both on the same memory layout.
 const _: () = {
     use std::mem::offset_of;
 


src/redisearch_rs/buffer/src/reader.rs
```diff
@@ -0,0 +1,93 @@
...
+        // No out-of-bounds risk, thanks to `BufferReader`'s invariant 1.
+        let n_bytes = dest_buf.write(&self.buffer.as_slice()[self.position..])?;
+        self.position += n_bytes;
+
+        Ok(n_bytes)
+    }
+}
+
+// Check, at compile-time, that `BufferReader` and `ffi::BufferReader` have the same representation.
+const _: () = {
```
zeenix, line 79: why not an ignored test instead?

zeenix (approved):
it would have been better to do the splitting & merging of files in separate commit(s) so it's easier to see the main change for reviewers.

LGTM otherwise.

src/redisearch_rs/buffer/src/writer.rs
```diff
@@ -0,0 +1,135 @@
...
+///
+/// # Usage
+///
+/// It's best to use this through the `std::io::Write` implementation, which handles
+/// buffer growth and cursor position management automatically.
+///
+/// # Invariants
+///
+/// 1. Position is smaller than or equal to the length of the buffer we're writing into.
+/// 2. `BufferWriter` has the same memory layout of [`ffi::BufferWriter`].
```
chesedo:
```suggestion
/// 2. `BufferWriter` has the same memory layout as [`ffi::BufferWriter`].
```

src/redisearch_rs/buffer/src/reader.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::Buffer;
+
+/// A cursor to read from a [`Buffer`].
+///
+/// It's best to use this through the `std::io::Read` implementation.
+///
+/// # Invariants
+///
+/// 1. Position is smaller than or equal to the length of the buffer we're reading from.
+/// 2. `BufferReader` has the same memory layout of [`ffi::BufferReader`].
```
chesedo:
```suggestion
/// 2. `BufferReader` has the same memory layout as [`ffi::BufferReader`].
```

chesedo (commented):
nit: small comment improvement

src/redisearch_rs/buffer/src/reader.rs
LukeMathWalker: I didn't think I could have a const fn as a test, but today I learned it's possible. 
Done in afe41ac09

zeenix: That's not what I meant. :) I didn't know that either.

I was talking about:

```rust
#[test]
#[ignore]
```

which tests the code to compile but not actually run.

chesedo (approved):

src/redisearch_rs/buffer/src/reader.rs
LukeMathWalker: While that would be technically correct (i.e. the current test desugars to a no-op at runtime), I think it'd be confusing to see that `#[ignore]` for folks who are not too well-versed into the semantics of const evaluation. 
Since the penalty is negligible, I prefer to leave it as is.

zeenix: I'm just wondering if declaring the test fn as const is doing what you are thinking. Did you find any docs on this. I couldn't. I'd be fine with just turning it back to `const _: () = ` that you originally did.

LukeMathWalker: It seems to, but then it only checks it if tests are built. So yeah, probably better to revert back. Done in 5e61a40d6

src/redisearch_rs/buffer/src/reader.rs
```diff
@@ -0,0 +1,102 @@
...
+    const SIZE_MATCHES: bool = size_of::<BufferReader>() == size_of::<ffi::BufferReader>();
+    const ALIGN_MATCHES: bool = align_of::<BufferReader>() == align_of::<ffi::BufferReader>();
+
+    // Field offset checks
+    const BUFFER_OFFSET_MATCHES: bool =
+        offset_of!(BufferReader<'static>, buffer) == offset_of!(ffi::BufferReader, buf);
+    const POSITION_OFFSET_MATCHES: bool =
+        offset_of!(BufferReader<'static>, position) == offset_of!(ffi::BufferReader, pos);
+
+    // Conditional compilation failure on mismatch
```
kei-nan, line 92: What should be done if the panic occurs?
What steps should we do?
Revert this PR?

Maybe add a comment with some directions.

LukeMathWalker: Done in 1eab83f8d

LukeMathWalker:
I've rebased on the latest `master`. Can you give it a final look @kei-nan?

kei-nan (approved):

PR #6260, opened 2025-06-05
Optional Iterator refactor - [MOD-8844]

BenGoldberger:
It contains the code for the Optional Iterator refactor.

Implementation has been taken from https://github.com/RediSearch/RediSearch/pull/5619/files

### DIFF
--- a/src/iterators/optional_iterator.c
+++ b/src/iterators/optional_iterator.c
@@ -1,0 +1,109 @@
+#include "optional_iterator.h"
+#include "empty_iterator.h"
+
+static void OI_Free(QueryIterator *base) {
+  OptionalIterator *nc = (OptionalIterator *)base;
+  if (nc->child) {
+    nc->child->Free(nc->child);
+  }
+  // Only free our virtual result, not the child's result
+  if (nc->virt) {
+    IndexResult_Free(nc->virt);
+  }
+  rm_free(base);
+}
+
+static size_t OI_NumEstimated(QueryIterator *base) {
+  OptionalIterator *nc = (OptionalIterator *)base;
+  return nc->maxDocId;
+}
+
+static void OI_Rewind(QueryIterator *base) {
+  OptionalIterator *nc = (OptionalIterator *)base;
+  base->atEOF = false;
+  base->lastDocId = 0;
+  nc->virt->docId = 0;
+  if (nc->child) {
+    nc->child->Rewind(nc->child);
+  }
+}
+
+
+
+// SkipTo for OPTIONAL iterator - Non-optimized version.
+static IteratorStatus OI_SkipTo_NO(QueryIterator *base, t_docId docId) {
+  OptionalIterator *nc = (OptionalIterator *)base;
+
+  if (docId > nc->maxDocId || base->atEOF) {
+    base->atEOF = true;
+    return ITERATOR_EOF;
+  }
+
+  if (docId > nc->child->lastDocId) {
+    IteratorStatus rc = nc->child->SkipTo(nc->child, docId);
+    if (rc == ITERATOR_TIMEOUT) return rc;
+  }
+
+  if (docId > 0 && docId == nc->child->lastDocId) {
+    // Has a real hit on the child iterator
+    nc->base.current = nc->child->current;
+    nc->base.current->weight = nc->weight;
+  } else {
+    // Virtual hit
+    nc->virt->docId = docId;
+    nc->base.current = nc->virt;
+  }
+  // Set the current ID
+  base->lastDocId = docId;
+  return ITERATOR_OK;
+}
+
+// Read from an OPTIONAL iterator - Non-Optimized version.
+static IteratorStatus OI_ReadSorted_NO(QueryIterator *base) {
+  OptionalIterator *nc = (OptionalIterator *)base;
+  if (base->atEOF || base->lastDocId >= nc->maxDocId) {
+    base->atEOF = true;
+    return ITERATOR_EOF;
+  }
+
+  // Increase the size by one
+  base->lastDocId++;
+
+  if (base->lastDocId > nc->child->lastDocId && !base->atEOF) {
+    IteratorStatus rc = nc->child->Read(nc->child);
+    if (rc == ITERATOR_TIMEOUT) return rc;
+  }
+
+  if (base->lastDocId != nc->child->lastDocId) {
+    nc->base.current = nc->virt;
+  } else {
+    nc->base.current = nc->child->current;
+    nc->base.current->weight = nc->weight;
+  }
+
+  nc->base.current->docId = base->lastDocId;
+  return ITERATOR_OK;
+}
+
+// Create a new OPTIONAL iterator - Non-Optimized version.
+QueryIterator *IT_V2(NewOptionalIterator_NonOptimized)(QueryIterator *it, t_docId maxDocId, size_t numDocs, double weight) {
+  OptionalIterator *nc = rm_calloc(1, sizeof(*nc));
+  nc->child = it ? it : IT_V2(NewEmptyIterator)();
+  nc->maxDocId = maxDocId;
+  nc->virt = NewVirtualResult(weight, RS_FIELDMASK_ALL);
+  nc->virt->freq = 1;
+  nc->weight = weight;
+
+  QueryIterator *ret = &nc->base;
+  ret->type = OPTIONAL_ITERATOR;
+  ret->atEOF = false;
+  ret->lastDocId = 0;
+  ret->current = nc->virt;
+  ret->NumEstimated = OI_NumEstimated;
+  ret->Free = OI_Free;
+  ret->Read = OI_ReadSorted_NO;
+  ret->SkipTo = OI_SkipTo_NO;
+  ret->Rewind = OI_Rewind;
+
+  return ret;
+}
--- a/tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
+++ b/tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
@@ -1,0 +1,243 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+#include "benchmark/benchmark.h"
+#include "iterator_util.h"
+#include "redismock/util.h"
+
+#include <random>
+#include <vector>
+
+#include "src/iterators/iterator_api.h"
+#include "src/iterators/optional_iterator.h"
+#include "src/iterators/idlist_iterator.h"
+#include "src/iterators/empty_iterator.h"
+
+class BM_OptionalIterator : public benchmark::Fixture {
+public:
+  std::vector<t_docId> childDocIds;
+  static bool initialized;
+  const t_docId maxDocId = 100000;
+  const size_t numDocs = 50000;
+  const double weight = 1.0;
+
+  void SetUp(::benchmark::State &state) {
+    if (!initialized) {
+      RMCK::init();
+      initialized = true;
+    }
+
+    // Generate child docIds based on benchmark parameters
+    auto numChildDocs = state.range(0);
+    std::mt19937 rng(42);
+    std::uniform_int_distribution<t_docId> dist(1, maxDocId);
+
+    childDocIds.resize(numChildDocs);
+    for (int i = 0; i < numChildDocs; ++i) {
+      childDocIds[i] = dist(rng);
+    }
+    std::sort(childDocIds.begin(), childDocIds.end());
+    childDocIds.erase(std::unique(childDocIds.begin(), childDocIds.end()), childDocIds.end());
+  }
+
+  void TearDown(::benchmark::State &state) {
+    // Nothing to clean up
+  }
+
+  QueryIterator* createOptionalWithChild() {
+    QueryIterator *child = IT_V2(NewIdListIterator)(childDocIds.data(), childDocIds.size(), 1.0);
+    return IT_V2(NewOptionalIterator_NonOptimized)(child, maxDocId, numDocs, weight);
+  }
+
+  QueryIterator* createOptionalEmpty() {
+    return IT_V2(NewOptionalIterator_NonOptimized)(NULL, maxDocId, numDocs, weight);
+  }
+};
+bool BM_OptionalIterator::initialized = false;
+
+// Benchmark scenarios: different numbers of child documents
+#define CHILD_DOCS_SCENARIOS() RangeMultiplier(2)->Range(2, 1024)->DenseRange(1500, 5000, 500)
+
+BENCHMARK_DEFINE_F(BM_OptionalIterator, ReadEmpty)(benchmark::State &state) {
+  QueryIterator *iterator_base = createOptionalEmpty();
+
+  for (auto _ : state) {
+    int rc = iterator_base->Read(iterator_base);
+    if (rc == ITERATOR_EOF) {
+      iterator_base->Rewind(iterator_base);
+    }
+    benchmark::DoNotOptimize(iterator_base->current);
+  }
+  iterator_base->Free(iterator_base);
+}
+
+BENCHMARK_DEFINE_F(BM_OptionalIterator, ReadWithChild)(benchmark::State &state) {
+  QueryIterator *iterator_base = createOptionalWithChild();
+
+  for (auto _ : state) {
+    int rc = iterator_base->Read(iterator_base);
+    if (rc == ITERATOR_EOF) {
+      iterator_base->Rewind(iterator_base);
+    }
+    benchmark::DoNotOptimize(iterator_base->current);
+  }
+  iterator_base->Free(iterator_base);
+}
+
+BENCHMARK_DEFINE_F(BM_OptionalIterator, SkipToEmpty)(benchmark::State &state) {
+  QueryIterator *iterator_base = createOptionalEmpty();
+
+  t_docId docId = 1;
+  for (auto _ : state) {
+    int rc = iterator_base->SkipTo(iterator_base, docId);
+    docId += 100;
+    if (rc == ITERATOR_EOF || docId > maxDocId) {
+      iterator_base->Rewind(iterator_base);
+      docId = 1;
+    }
+    benchmark::DoNotOptimize(rc);
+  }
+  iterator_base->Free(iterator_base);
+}
+
+BENCHMARK_DEFINE_F(BM_OptionalIterator, SkipToWithChild)(benchmark::State &state) {
+  QueryIterator *iterator_base = createOptionalWithChild();
+
+  t_docId docId = 1;
+  for (auto _ : state) {
+    int rc = iterator_base->SkipTo(iterator_base, docId);
+    docId += 100;
+    if (rc == ITERATOR_EOF || docId > maxDocId) {
+      iterator_base->Rewind(iterator_base);
+      docId = 1;
+    }
+    benchmark::DoNotOptimize(rc);
+  }
+  iterator_base->Free(iterator_base);
+}
+
+BENCHMARK_DEFINE_F(BM_OptionalIterator, SkipToRealHits)(benchmark::State &state) {
+  QueryIterator *iterator_base = createOptionalWithChild();
+
+  size_t childIndex = 0;
+  for (auto _ : state) {
+    if (childIndex >= childDocIds.size()) {
+      iterator_base->Rewind(iterator_base);
+      childIndex = 0;
+    }
+
+    t_docId target = childDocIds[childIndex++];
+    int rc = iterator_base->SkipTo(iterator_base, target);
+    benchmark::DoNotOptimize(rc);
+  }
+  iterator_base->Free(iterator_base);
+}
+
+BENCHMARK_DEFINE_F(BM_OptionalIterator, SkipToVirtualHits)(benchmark::State &state) {
+  QueryIterator *iterator_base = createOptionalWithChild();
+
+  // Generate docIds that are NOT in childDocIds (virtual hits)
+  std::vector<t_docId> virtualDocIds;
+  for (t_docId i = 1; i <= maxDocId; i += 137) { // Use prime step to avoid patterns
+    if (std::find(childDocIds.begin(), childDocIds.end(), i) == childDocIds.end()) {
+      virtualDocIds.push_back(i);
+    }
+  }
+
+  size_t virtualIndex = 0;
+  for (auto _ : state) {
+    if (virtualIndex >= virtualDocIds.size()) {
+      iterator_base->Rewind(iterator_base);
+      virtualIndex = 0;
+    }
+
+    t_docId target = virtualDocIds[virtualIndex++];
+    int rc = iterator_base->SkipTo(iterator_base, target);
+    benchmark::DoNotOptimize(rc);
+  }
+  iterator_base->Free(iterator_base);
+}
+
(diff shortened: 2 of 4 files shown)


tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+BENCHMARK_DEFINE_F(BM_OptionalIterator, ReadWithChild)(benchmark::State &state) {
+  QueryIterator *iterator_base = createOptionalWithChild();
+
+  for (auto _ : state) {
+    IteratorStatus rc = iterator_base->Read(iterator_base);
+    if (rc == ITERATOR_EOF) {
+      iterator_base->Rewind(iterator_base);
+    }
+    benchmark::DoNotOptimize(iterator_base->current);
```
JoanFM: what does this do?

tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+class BM_OptionalIterator : public benchmark::Fixture {
+public:
+  std::vector<t_docId> childDocIds;
+  static bool initialized;
+  const t_docId maxDocId = 100000;
+  const size_t numDocs = 50000;
+  const double weight = 1.0;
+
+  void SetUp(::benchmark::State &state) {
```
JoanFM, line 34: I suggest do the creation in SetUP and Release in TearDown, and to support 2 versions, you could use template parameter of the IteratorType.

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,109 @@
+#include "optional_iterator.h"
+#include "empty_iterator.h"
+
+static void OI_Free(QueryIterator *base) {
+  OptionalIterator *nc = (OptionalIterator *)base;
```
JoanFM: can we use another name? `nc`stands for ? maybe oi?

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  if (base->lastDocId != nc->child->lastDocId) {
+    nc->base.current = nc->virt;
+  } else {
+    nc->base.current = nc->child->current;
+    nc->base.current->weight = nc->weight;
+  }
+
+  nc->base.current->docId = base->lastDocId;
```
JoanFM: u can use `base`directly also here

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,109 @@
...
+  // Increase the size by one
+  base->lastDocId++;
+
+  if (base->lastDocId > nc->child->lastDocId && !base->atEOF) {
+    IteratorStatus rc = nc->child->Read(nc->child);
+    if (rc == ITERATOR_TIMEOUT) return rc;
+  }
+
+  if (base->lastDocId != nc->child->lastDocId) {
+    nc->base.current = nc->virt;
```
JoanFM: I would use `base`directly

tests/cpptests/test_cpp_iterator_optional.cpp
```diff
@@ -0,0 +1,375 @@
...
+  OptionalIterator *oi = (OptionalIterator *)iterator_base;
+
+  // Read first document (should be virtual)
+  ASSERT_EQ(iterator_base->Read(iterator_base), ITERATOR_OK);
+  ASSERT_EQ(iterator_base->current, oi->virt);
+  ASSERT_EQ(iterator_base->current->weight, weight);
+}
+
+// Test edge cases
+class OptionalIteratorEdgeCasesTest : public ::testing::Test {
```
JoanFM: I think I am missing a Test showing how it behaves when Child Times Out

tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
BenGoldberger: in theory it tells the compiler to not optimize the variable, it something that the AI tool suggested.

JoanFM: I am not sure I would use it, or we should use it everywhere. Wdyt @GuyAv46

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,109 @@
...
+  // Only free our virtual result, not the child's result
+  if (oi->virt) {
+    IndexResult_Free(oi->virt);
+  }
+  rm_free(base);
+}
+
+static size_t OI_NumEstimated(QueryIterator *base) {
+  OptionalIterator *oi = (OptionalIterator *)base;
+  return oi->maxDocId;
```
GuyAv46, line 17: OK for now, but we should also consult with the wildcard child if it is set

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,109 @@
...
+  oi->virt->docId = 0;
+  if (oi->child) {
+    oi->child->Rewind(oi->child);
+  }
+}
+
+
+
+// SkipTo for OPTIONAL iterator - Non-optimized version.
+static IteratorStatus OI_SkipTo_NO(QueryIterator *base, t_docId docId) {
```
GuyAv46: Let's improve the name (`*_NO` is hard to understand, we can afford to give it the full "notOptimized" suffix

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,109 @@
...
+    base->atEOF = true;
+    return ITERATOR_EOF;
+  }
+
+  if (docId > oi->child->lastDocId) {
+    IteratorStatus rc = oi->child->SkipTo(oi->child, docId);
+    if (rc == ITERATOR_TIMEOUT) return rc;
+  }
+
+  if (docId > 0 && docId == oi->child->lastDocId) {
```
GuyAv46: `docId > 0` is always true
```suggestion
  if (docId == oi->child->lastDocId) {
```

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,109 @@
...
+  OptionalIterator *oi = (OptionalIterator *)base;
+  if (base->atEOF || base->lastDocId >= oi->maxDocId) {
+    base->atEOF = true;
+    return ITERATOR_EOF;
+  }
+
+  // Increase the size by one
+  base->lastDocId++;
+
+  if (base->lastDocId > oi->child->lastDocId && !base->atEOF) {
```
GuyAv46: `!base->atEOF` is always true here
```suggestion
  if (base->lastDocId > oi->child->lastDocId) {
```

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  base->current->docId = base->lastDocId;
+  return ITERATOR_OK;
+}
+
+// Create a new OPTIONAL iterator - Non-Optimized version.
+QueryIterator *IT_V2(NewOptionalIterator_NonOptimized)(QueryIterator *it, t_docId maxDocId, size_t numDocs, double weight) {
+  OptionalIterator *oi = rm_calloc(1, sizeof(*oi));
+  oi->child = it ? it : IT_V2(NewEmptyIterator)();
```
GuyAv46:
```suggestion
  oi->child = it ?: IT_V2(NewEmptyIterator)();
```
I wonder if this case is even possible, and whether we should have an empty iterator as a child or simply return a wildcard iterator instead of the optional

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,109 @@
...
+    base->current = oi->child->current;
+    base->current->weight = oi->weight;
+  }
+
+  base->current->docId = base->lastDocId;
+  return ITERATOR_OK;
+}
+
+// Create a new OPTIONAL iterator - Non-Optimized version.
+QueryIterator *IT_V2(NewOptionalIterator_NonOptimized)(QueryIterator *it, t_docId maxDocId, size_t numDocs, double weight) {
```
GuyAv46: Let's have a "generic" constructor, the optimization should be chosen automatically when possible, and not be part of the API

tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include "redismock/util.h"
+#include "iterator_util.h"
+
+#include <random>
+#include <vector>
+
+#include "src/iterators/iterator_api.h"
+#include "src/iterators/optional_iterator.h"
+#include "src/iterators/idlist_iterator.h"
+#include "src/iterators/empty_iterator.h"
```
GuyAv46: Why do we need them here?

tests/cpptests/test_cpp_iterator_optional.cpp
```diff
@@ -0,0 +1,481 @@
...
+  ASSERT_EQ(iterator_base->SkipTo(iterator_base, maxDocId + 1), ITERATOR_EOF);
+  ASSERT_TRUE(iterator_base->atEOF);
+}
+
+TEST_F(OptionalIteratorEmptyTest, SkipToZero) {
+  // Test skipping to docId 0 (should set to 0)
+  ASSERT_EQ(iterator_base->SkipTo(iterator_base, 0), ITERATOR_OK);
+  ASSERT_EQ(iterator_base->current->docId, 0);
+  ASSERT_EQ(iterator_base->lastDocId, 0);
+}
```
GuyAv46: Not required. Lets drop this "feature"

tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
```diff
@@ -0,0 +1,236 @@
...
+BENCHMARK_REGISTER_F(BM_OptionalIteratorEmpty, ReadEmpty)->CHILD_DOCS_SCENARIOS();
+BENCHMARK_REGISTER_F(BM_OptionalIteratorWithChild, ReadWithChild)->CHILD_DOCS_SCENARIOS();
+BENCHMARK_REGISTER_F(BM_OptionalIteratorEmpty, SkipToEmpty)->CHILD_DOCS_SCENARIOS();
+BENCHMARK_REGISTER_F(BM_OptionalIteratorWithChild, SkipToWithChild)->CHILD_DOCS_SCENARIOS();
+BENCHMARK_REGISTER_F(BM_OptionalIteratorWithChild, SkipToRealHits)->CHILD_DOCS_SCENARIOS();
+BENCHMARK_REGISTER_F(BM_OptionalIteratorWithChild, SkipToVirtualHits)->CHILD_DOCS_SCENARIOS();
+BENCHMARK_REGISTER_F(BM_OptionalIteratorWithChild, MixedOperations)->CHILD_DOCS_SCENARIOS();
+BENCHMARK_REGISTER_F(BM_OptionalIteratorWithChild, FullIteration)->CHILD_DOCS_SCENARIOS();
+BENCHMARK_REGISTER_F(BM_OptionalIteratorWithChild, RewindPerformance)->CHILD_DOCS_SCENARIOS();
+BENCHMARK_REGISTER_F(BM_OptionalIteratorEmpty, CreateAndDestroy)->CHILD_DOCS_SCENARIOS();
```
GuyAv46: Let's remove the `CHILD_DOCS_SCENARIOS` in the cases of `BM_OptionalIteratorEmpty`

tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
```diff
@@ -0,0 +1,236 @@
...
+    // Just advance the iterator
+  }
+
+  for (auto _ : state) {
+    iterator_base->Rewind(iterator_base);
+    benchmark::DoNotOptimize(iterator_base->lastDocId);
+  }
+}
+
+BENCHMARK_DEFINE_F(BM_OptionalIteratorEmpty, CreateAndDestroy)(benchmark::State &state) {
```
GuyAv46: We shouldn't care for this scenario

tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
```diff
@@ -0,0 +1,236 @@
...
+    // Iterate through all documents
+    while (temp_iterator->Read(temp_iterator) == ITERATOR_OK) {
+      benchmark::DoNotOptimize(temp_iterator->current->docId);
+    }
+
+    temp_iterator->Free(temp_iterator);
+  }
+}
+
+BENCHMARK_DEFINE_F(BM_OptionalIteratorWithChild, RewindPerformance)(benchmark::State &state) {
```
GuyAv46: We shouldn't care for this scenario

tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+    if (rc == ITERATOR_EOF) {
+      iterator_base->Rewind(iterator_base);
+    }
+
+    benchmark::DoNotOptimize(rc);
+  }
+}
+
+BENCHMARK_DEFINE_F(BM_OptionalIteratorWithChild, FullIteration)(benchmark::State &state) {
```
GuyAv46: Does this scenario give interesting results? I don't think we should care about it

tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
```diff
@@ -0,0 +1,236 @@
...
+      virtualIndex = 0;
+    }
+
+    t_docId target = virtualDocIds[virtualIndex++];
+    IteratorStatus rc = iterator_base->SkipTo(iterator_base, target);
+    benchmark::DoNotOptimize(rc);
+  }
+}
+
+BENCHMARK_DEFINE_F(BM_OptionalIteratorWithChild, MixedOperations)(benchmark::State &state) {
```
GuyAv46: Do we need this scenario? Isn't it just the average of `Read` and `SkipTo`?

tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
```diff
@@ -0,0 +1,236 @@
...
+  for (auto _ : state) {
+    IteratorStatus rc = iterator_base->SkipTo(iterator_base, docId);
+    docId += 100;
+    if (rc == ITERATOR_EOF || docId > maxDocId) {
+      iterator_base->Rewind(iterator_base);
+      docId = 1;
+    }
+    benchmark::DoNotOptimize(rc);
+  }
+}
```
GuyAv46:
```suggestion
BENCHMARK_DEFINE_F(BM_OptionalIteratorWithChild, SkipToWithChild)(benchmark::State &state) {
  t_offset step = 10;
  for (auto _ : state) {
    IteratorStatus rc = iterator_base->SkipTo(iterator_base, iterator_base->lastDocId + step);
    if (rc == ITERATOR_EOF) {
      iterator_base->Rewind(iterator_base);
    }
    benchmark::DoNotOptimize(rc);
  }
}
```

tests/cpptests/micro-benchmarks/benchmark_optional_iterator.cpp
GuyAv46: Either with templates or set both new and old in `SetUp`. Eventually we will delete the old one

src/iterators/optional_iterator.c
BenGoldberger: in theory yes, but in we dont enforce it in the test, so it leads to segfault if not checked

src/iterators/optional_iterator.c
BenGoldberger: since the non optimize version doesnt use the wildcard iterator I left it for now

src/iterators/optional_iterator.c
GuyAv46: So let's add an assertion (that `docId` > `base->lastDocId`) and drop this validation

JoanFM: agree here with Guy

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,109 @@
...
+  oi->virt->docId = 0;
+  if (oi->child) {
+    oi->child->Rewind(oi->child);
+  }
+}
+
+
+
+// SkipTo for OPTIONAL iterator - Non-optimized version.
+static IteratorStatus OI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId) {
```
JoanFM, line 35: can you add a comment up here, explaining the logic? In this SkipTo, there is no ITERATOR_NOTFOUND? It always matches?

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,109 @@
...
+
+// Read from an OPTIONAL iterator - Non-Optimized version.
+static IteratorStatus OI_ReadSorted_NotOptimized(QueryIterator *base) {
+  OptionalIterator *oi = (OptionalIterator *)base;
+  if (base->atEOF || base->lastDocId >= oi->maxDocId) {
+    base->atEOF = true;
+    return ITERATOR_EOF;
+  }
+
+  // Increase the size by one
```
JoanFM: it is not the size

JoanFM: maybe, point to the next one?

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,109 @@
...
+  base->lastDocId++;
+
+  if (base->lastDocId > oi->child->lastDocId) {
+    IteratorStatus rc = oi->child->Read(oi->child);
```
JoanFM, line 76: Is reading one time enough? Or should we read until lastDocID is surpassed?

JoanFM: If we Read until lastDocId is surpassed I am not sure how we would behave in next iteration

tests/cpptests/index_utils.h
```diff
@@ -47,3 +48,32 @@ size_t CalculateNumericInvertedIndexMemory(NumericRangeTree *rt, NumericRangeNod
 size_t NumericRangeGetMemory(const NumericRangeNode *Node);
 
 NumericRangeTree *getNumericTree(IndexSpec *spec, const char *field);
+
+class MockQueryEvalCtx {
```
JoanFM, line 52: I think for the tests is better and more explanatory to have a constructor where you can inject the Iterator u want than having all these mocks that are quite hard to follow. Then you can do some testing on the constructor if needed.

src/iterators/optional_iterator.c
BenGoldberger: we would return `oi->virt` until child->lastDocId

tests/cpptests/index_utils.h
BenGoldberger: I'm not sure I follow. do you want me to add a constructor for the mock?

JoanFM: check here https://github.com/RediSearch/RediSearch/pull/6307

I wonder if it is easier to follow to have just some constructor that u can use to inject your iterator, instead of trying to mock all these structures.

Then you can consider what u may see best.

src/iterators/optional_iterator.c
BenGoldberger: Yes similar to the wildcard iterator

tests/cpptests/index_utils.h
BenGoldberger: I think this implementation  is quite simple for this benchmark

tests/cpptests/test_cpp_iterator_optional.cpp
```diff
@@ -0,0 +1,404 @@
...
+  // Should be virtual hit
+  OptionalIterator *oi = (OptionalIterator *)iterator_base;
+  ASSERT_EQ(iterator_base->current, oi->virt);
+}
+
+TEST_F(OptionalIteratorTest, SkipToSequence) {
+  // Test skipping to various docIds in sequence
+  t_docId targets[] = {5, 15, 25, 35, 45, 55, 65, 75, 85, 95};
+
+  for (size_t i = 0; i < sizeof(targets) / sizeof(targets[0]); i++) {
```
GuyAv46:
```suggestion
  for (auto &target : targets) {
```

GuyAv46 (dismissed):

src/iterators/optional_iterator.c
GuyAv46: `nc` stands for "not context". Sowing the history of where it was copied from :)

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,112 @@
...
+  }
+}
+
+
+
+// SkipTo for OPTIONAL iterator - Non-optimized version.
+// Skip to a specific docId. If the child has a hit on this docId, return it.
+// Otherwise, return a virtual hit.
+static IteratorStatus OI_SkipTo_NotOptimized(QueryIterator *base, t_docId docId) {
+  assert(docId > 0);
```
JoanFM: RS_ASSERT I believe is best. I guess somehow we skip the assertion in prod

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,112 @@
...
+
+  // Point to next doc
+  base->lastDocId++;
+
+  if (base->lastDocId > oi->child->lastDocId) {
+    IteratorStatus rc = oi->child->Read(oi->child);
+    if (rc == ITERATOR_TIMEOUT) return rc;
+  }
+
+  if (base->lastDocId != oi->child->lastDocId) {
```
JoanFM: very nitpicky, ignore if u want, but what about having positive logical here as in the SkipTo?

Also, why is the oi-virt-docId not updated?

tests/cpptests/test_cpp_iterator_optional.cpp
```diff
@@ -0,0 +1,404 @@
...
+  // After reading all docs, should return EOF
+  ASSERT_EQ(iterator_base->Read(iterator_base), ITERATOR_EOF);
+  ASSERT_TRUE(iterator_base->atEOF);
+}
+
+TEST_F(OptionalIteratorWithEmptyChildTest, SkipToVirtualHits) {
+  // Skip to various docIds - all should be virtual hits
+  t_docId targets[] = {5, 15, 25, 35, 45};
+
+  for (size_t i = 0; i < sizeof(targets) / sizeof(targets[0]); i++) {
```
JoanFM: same as Guy's comment on target loops.

JoanFM (changes requested):

src/iterators/optional_iterator.c
GuyAv46: Good catch! I'd also change it to validate that the requested ID is ahead of the last ID
```suggestion
  RS_ASSERT(docId > base->lastDocId);
```

src/iterators/optional_iterator.c
BenGoldberger: it probably should be updated

JoanFM (approved):

GuyAv46 (approved):

src/iterators/optional_iterator.c
```diff
@@ -0,0 +1,112 @@
...
+  } else {
+    oi->virt->docId = base->lastDocId;
+    base->current = oi->virt;
+  }
+  return ITERATOR_OK;
+}
+
+// Create a new OPTIONAL iterator - Non-Optimized version.
+QueryIterator *IT_V2(NewOptionalIterator)(QueryIterator *it, t_docId maxDocId, size_t numDocs, double weight) {
+  assert(it != NULL);
```
GuyAv46, line 92: On a second look, this may happen (for example, when searching for `~foo` but no document has `foo` so there is no inverted index for it)
I suggest we return a wildcard iterator in this case.
We have to check the implications on `FT.PROFILE` and `FT.EXPLAIN` when we start using the new iterators

GuyAv46: alternativly, use `RS_ASSERT` and we will have to handle this case when using the API

PR #6261, opened 2025-06-05
MOD-10017 Rust Counter result processor

JonasKruckenberg:

This PR ports the RPCounter to Rust *without* exposing it to C yet. This implements all the machinery and setup, as well as tests required to make sure the new RP behaves exactly as the old RP. Enabling the new RP and switching the codebase over is left for a followup PR.

stacked on top of #6250 

Note: 
A lot of the design is transient, once more code has been ported to Rust we will revisit certain APIs and update them accordingly. (See MOD-9920, MOD-10230, and MOD-10229)


src/redisearch_rs/c_entrypoint/result_processor_ffi/build.rs
```diff
@@ -0,0 +1,22 @@
...
+
+fn main() {
+    let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+
+    cbindgen::Builder::new()
+        .with_crate(crate_dir)
+        .with_config(Config::from_file("cbindgen.toml").expect("Failed to find cbindgen config"))
+        .generate()
+        .expect("Unable to generate bindings")
+        .write_to_file("../../headers/result_processor.h");
```
chesedo: If this is going to be called from the C code while the `result_processor.h` still exists in C then you will want to rename this to `result_processor_rs.h` for the time being.

src/redisearch_rs/c_entrypoint/result_processor_ffi/cbindgen.toml
```diff
@@ -0,0 +1,13 @@
+sys_includes = ["time.h"]
```
chesedo: You might be able to drop this

JonasKruckenberg: The result processor definition has a `giltime: timespec` field where `timespec` comes from `libc` 🤔 I assumed I needed it because of that

src/redisearch_rs/c_entrypoint/result_processor_ffi/build.rs
JonasKruckenberg: good point 👍🏻

src/redisearch_rs/c_entrypoint/result_processor_ffi/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+#![allow(non_camel_case_types, non_snake_case)]
```
chesedo: These are not used in this module right?

JonasKruckenberg: yup, removed!

src/redisearch_rs/c_entrypoint/result_processor_ffi/cbindgen.toml
JonasKruckenberg: was able to drop it, the type isn't defined by Rust yet anyways

src/redisearch_rs/c_entrypoint/result_processor_ffi/build.rs
JonasKruckenberg: changed

src/redisearch_rs/c_entrypoint/result_processor_ffi/cbindgen.toml
chesedo: Well, it does not seem that any thing in the generated header uses `time` directly. The dependency should pull in `time` where it is used

chesedo (approved):

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,119 @@
...
+            unsafe { NonNull::new(Box::into_raw(Pin::into_inner_unchecked(upstream))).unwrap() };
+
+        let mut rp = Box::pin(ResultProcessorWrapper::new(Counter::new()));
+
+        // Emulate what `QITR_PushRP` would be doing
+        rp.header.upstream = upstream.as_ptr().cast();
+
+        let mut rp = NonNull::new(unsafe { ResultProcessorWrapper::into_ptr(rp) }).unwrap();
+
+        let cx = unsafe { Context::from_raw(rp.cast()) };
```
DarthB: add unsafe wrapper `get_context`

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,119 @@
...
+            },
+            flags: 0,
+        };
+
+        // Set up the result processors on the heap
+        let upstream = Box::pin(ResultProcessorWrapper::new(from_iter([INIT, INIT, INIT])));
+        let upstream =
+            unsafe { NonNull::new(Box::into_raw(Pin::into_inner_unchecked(upstream))).unwrap() };
+
+        let mut rp = Box::pin(ResultProcessorWrapper::new(Counter::new()));
```
DarthB:
```suggestion
        let mut rp = Counter::new_c_compatabile();
```
Let's move this unsafe stuff in an impl block and use safe methods there. This way `RPCounter_New` becomes even shorter.

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        // Emulate what `QITR_PushRP` would be doing
+        rp.header.upstream = upstream.as_ptr().cast();
+
+        let mut rp = NonNull::new(unsafe { ResultProcessorWrapper::into_ptr(rp) }).unwrap();
```
DarthB: wrap in a safe method too

DarthB (dismissed):
Great work. 

We should also hide as much unsafe in safe methods as possible.

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,119 @@
...
+        assert_eq!(unsafe { rp.as_mut() }.result_processor.count, 3);
+
+        // we need to correctly free both result processors at the end
+        unsafe {
+            let mut rp = rp.cast::<crate::Header>();
+            (rp.as_mut().free.unwrap())(rp.as_ptr());
+
+            let mut upstream = upstream.cast::<crate::Header>();
+            (upstream.as_mut().free.unwrap())(upstream.as_ptr());
+        }
```
DarthB: Can you explain that in more detail? Isn't the C code calling free after the result processor pipeline is run? Why do we need this in pure Rust Code?

JonasKruckenberg: This is a test case, there is no C code. We leak the allocation at the start of the test case (`ResultProcessorWrapper::into_ptr`) and so at the end we need to correctly free it to avoid leaking memory (so tests run fine AND to make Miri happy)

src/redisearch_rs/result_processor/src/counter.rs
JonasKruckenberg: This is purposefully not safe, because as the safety comment mentions the caller has to correctly treat the pointer as pinned and free it after it is done using the RP.

src/redisearch_rs/result_processor/src/counter.rs
JonasKruckenberg: I prefer to keep allocations explicit and not hide them behind methods, allocations can be feel like hidden "magic" otherwise. I also think this will make developers coming from C more comfortable with the function

src/redisearch_rs/result_processor/src/counter.rs
DarthB: So, a free call is required for Miri?! These free calls where what puzzled me in the first place.

src/redisearch_rs/result_processor/src/counter.rs
DarthB: I see your point, but as test code is "user-code," I think: In idiomatic Rust, we strive to avoid showing the internals of allocation whenever possible. Can't we force the creation of a pinned Box on the Rust side, e.g., by using `ResultProcessWrapper` as a builder?

src/redisearch_rs/result_processor/src/counter.rs
JonasKruckenberg: yup

src/redisearch_rs/result_processor/src/counter.rs
JonasKruckenberg: we have a safe constructor for this now

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -252,6 +266,8 @@ where
                 },
                 next: Some(Self::result_processor_next),
                 free: Some(Self::result_processor_free),
+                inner_ty_id: TypeId::of::<P>(),
+                inner_ty_name: type_name::<P>(),
```
LukeMathWalker: Missing cfg here?

src/redisearch_rs/result_processor/src/test_utils.rs
```diff
@@ -0,0 +1,210 @@
...
+impl Chain {
+    pub const fn new() -> Self {
+        Self {
+            result_processors: Vec::new(),
+        }
+    }
+
+    /// Append a new result processor at the end of the chain. It will have its `upstream`
+    /// field set to the previous last result processor.
+    pub fn append<P>(&mut self, mut result_processor: Pin<Box<ResultProcessorWrapper<P>>>)
```
LukeMathWalker: I think it'd nicer to take `ResultProcessorWrapper<P>` as input here, moving the box-pinning inside `append`.

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: good catch 👍🏻

src/redisearch_rs/result_processor/src/test_utils.rs
JonasKruckenberg: My thinking and personal style preference is that I want allocations to be explicit when dealing with unsafe code, that's why I have been particular about always allocating the Box at the "root" of the call tree. IMO this makes it easier for C people to grok too.

But in the end this doesn't really matter to me, if this is important to you I'm fine changing it

LukeMathWalker: For production code, I agree.
But for testing code, I think we should strive to minimise noise.

JonasKruckenberg: changed 👍🏻

src/redisearch_rs/c_entrypoint/result_processor_ffi/tests/counter.rs
```diff
@@ -0,0 +1,89 @@
...
+//! This file contains tests to ensure the FFI functions behave as expected.
+
+use result_processor_ffi::counter::*;
+use std::ptr;
+
+/// Mock implementation of `SearchResult_Clear` for tests
+///
+/// this doesn't actually free anything, so will leak resources but hopefully this is fine for the few Rust
+/// tests for now
+// FIXME replace with SearchResult::clear once `ffi::SearchResult` is ported to Rust
```
Itzikvaknin, line 19: Can you please add a ticket or reference to show how you are tracking it?

src/redisearch_rs/c_entrypoint/result_processor_ffi/src/counter.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use result_processor::{ResultProcessorWrapper, counter::Counter};
+
+/// Crate a new heap-allocated `Counter` result processor
+///
+/// # Safety
+///
+/// - The caller must never move the allocated result processor from its original allocation.
+/// - The caller must ensure to call the `Free` VTable function to properly destroy the type.
+#[unsafe(no_mangle)]
```
Itzikvaknin, line 18: isnt it always unsafe?

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        Ok(None)
+    }
+}
+
+impl Default for Counter {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
Itzikvaknin, line 49: Is it necessary?

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -0,0 +1,75 @@
...
+    ) -> Result<Option<()>, crate::Error> {
+        let mut upstream = cx
+            .upstream()
+            .expect("There is no processor upstream of this counter.");
+
+        while upstream.next(res)?.is_some() {
+            self.count += 1;
+
+            // Safety: This function should only be called on initialized SearchResults. Luckily,
+            // a ResultProcessor returning `RPStatus_RS_RESULT_OK` means "Result is filled with valid data"
```
Itzikvaknin, line 34: How is it inferred? (`is_some()` -> `RPStatus_RS_RESULT_OK`)

src/redisearch_rs/result_processor/src/test_utils.rs
```diff
@@ -0,0 +1,222 @@
...
+impl Drop for Chain {
+    fn drop(&mut self) {
+        for mut ptr in self.result_processors.drain(..) {
+            unsafe { (ptr.as_mut().free.unwrap())(ptr.as_ptr()) }
+        }
+    }
+}
+
+/// Return the default value for [`ffi::SearchResult`]
+// FIXME replace with Default::default once `ffi::SearchResult` is ported to Rust
```
Itzikvaknin: Same monitoring comment

src/redisearch_rs/result_processor/src/test_utils.rs
```diff
@@ -0,0 +1,222 @@
...
+        flags: 0,
+    };
+    SEARCH_RESULT_INIT
+}
+
+/// Mock implementation of `SearchResult_Clear` for tests
+///
+/// this doesn't actually free anything, so will leak resources but hopefully this is fine for the few Rust
+/// tests for now
+// FIXME replace with SearchResult::clear once `ffi::SearchResult` is ported to Rust
```
Itzikvaknin: See above

Itzikvaknin (approved):

PR #6271, opened 2025-06-06
ci: run unit, flow standalone and flow coordinator tests in parallel

JoanFM:

Tries to parallelize Unit, StandAlone and Coordinator jobs so that PR and merges are faster


.github/workflows/task-test.yml
```diff
@@ -125,32 +108,8 @@ jobs:
...
-          echo "CXX=$(dirname $CLANG_BIN)/clang++-$CLANG_VERSION" >> $GITHUB_ENV
-          echo "LD=$CLANG_BIN" >> $GITHUB_ENV
-      - name: Get Redis (node20)
-        if: ${{ steps.node20.outputs.supported == 'true' }}
-        uses: actions/checkout@v4
-        with:
-          repository: redis/redis
-          ref: ${{ inputs.get-redis }}
-          path: redis
       - name: Get Redis (node20 not supported)
```
kei-nan, line 111: is it possible that git will only be installed by the install scripts?
Or are we sure it will be around before anything gets installed in the docker container?
Maybe the installation and compiling be separated into two yml files?

JoanFM: the problem of having the "Checkout phase" in a separate file (like an Inner GH action) is that those cannot use other "uses" external actions (not sure why I just hit that limit while working on this).

As per the presence of `git`. In the current GH flow, how does it differ?

JoanFM: For the separation in different actions, it seems that these composite actions do not allow to use external "uses" (such as checkout and so on).

As per the git issue, I see it is installed in the installation scripts, but these steps are only set for cloning the repo (in presence or not of a specific NodeJS). The only difference I see is when some `pre-steps`scripts are passed (but I do not see any automatic way of doing so).

kei-nan: Why does it need to be a composite action file?
Can it be a regular yml file we can then reuse?
Right now it looks like only this flow is using this composite action?

Can we maybe try and structure/group the actions more logically?
- installation/setup
- building
- running tests with the built binary
@GuyAv46 WDYT? You are the most familiar with this flow.

JoanFM: If you do not have a composable action, then your other YAML runs as a separate job (which means you lose all the installs and so on). I already tried and it was my first intention but could not find the way

GuyAv46: I tried to use composite action before, I think it would be nice to have a setup action with all the different cases.

The issue is that it can be used only after checking out the repo, and for us it means we must install the pre-script dependencies first, and that takes most of the "meat" out of the composite action. We also have slightly different installations for testing and artifact building, so I'm not sure it will be simple to use a setup action that fits all purposes.

- An action is run on a given runner as part of the job steps, whereas a workflow runs different jobs on different runners; therefore, we can't call a workflow from a step.
- Therefore, no
- We can try giving the action more inputs to fit other setups, but I'm not sure if it will make it any clearer or simpler to maintain

JoanFM: I know this is not perfect and has a lot of repetition (specially for this node20 check) but I think is not a bad balance of parallelization and logic isolation.

GuyAv46: To be honest, I'm not sure if splitting the standalone and cluster testing is the right thing to do in order to speed up the CI. The CI's weak point at the moment is flaky setups and tests, and waiting for ARM runners that we don't have enough of. Splitting the testing job into 2 will require double the instances and double the setup we have to run on each instance.

From what I see, we'd better 
1. Ask @adamiBs for more ARM instances. At the moment, we may wait for instances for almost an hour, while all the other jobs are already done.
2. Remove the standalone test configuration multiplier - each configuration takes ~6 minutes, and there is very little benefit for running all the tests again with a different configuration each time. We need to write a few dedicated tests for these configurations.
3. Make the setup step more robust, so we have fewer failures on these steps
4. Improve the coordinator's environment setup. IIRC @MeirShpilraien had an idea for how to speed up this process. This is the main reason the coordinator tests are running significantly slower than the standalone ones. We may be able to boost the coordinator tests' step run time up to the standalone time

JoanFM: Okey, then I am going to put it as draft or close this

PR #6275, opened 2025-06-07
Fix skipTo target id

GuyAv46:

Fix `SkipTo` target ID in the new iterator miro-benchmarks.

#### Main Fix:
Use the iterator's `lastDocId` value to ensure we request a skip forward, regardless of the previous `SkipTo` call status

#### Mark if applicable

### DIFF
--- a/tests/cpptests/micro-benchmarks/benchmark_idlist_iterator.cpp
+++ b/tests/cpptests/micro-benchmarks/benchmark_idlist_iterator.cpp
@@ -62,13 +62,11 @@ BENCHMARK_DEFINE_F(BM_IdListIterator, Read)(benchmark::State &state) {
 BENCHMARK_DEFINE_F(BM_IdListIterator, SkipTo)(benchmark::State &state) {
   QueryIterator *iterator_base = IT_V2(NewIdListIterator)(docIds.data(), docIds.size(), 1.0);
 
-  t_docId docId = 10;
+  t_offset step = 10;
   for (auto _ : state) {
-    IteratorStatus rc = iterator_base->SkipTo(iterator_base, docId);
-    docId += 10;
+    IteratorStatus rc = iterator_base->SkipTo(iterator_base, iterator_base->lastDocId + step);
     if (rc == ITERATOR_EOF) {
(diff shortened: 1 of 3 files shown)


tests/cpptests/micro-benchmarks/benchmark_idlist_iterator.cpp
```diff
@@ -92,15 +90,13 @@ BENCHMARK_DEFINE_F(BM_IdListIterator, Read_Old)(benchmark::State &state) {
 
 BENCHMARK_DEFINE_F(BM_IdListIterator, SkipTo_Old)(benchmark::State &state) {
   IndexIterator *iterator_base = NewIdListIterator(docIds.data(), docIds.size(), 1.0);
-  RSIndexResult *hit;
+  RSIndexResult *hit = iterator_base->current;
```
JoanFM, line 93: is the current properly set to docID 0? when instantiated? should we add this to the unittest?

JoanFM (changes requested):

tests/cpptests/micro-benchmarks/benchmark_idlist_iterator.cpp
JoanFM: well not in unit tests, but are we sure?

JoanFM: I would do `lastDocID = 0`at the beginning, and then update with `hit`after every call to SkipTo (which I believe that according to interface could be more ensured)?

GuyAv46: When initializing and rewinding. In the new implementation, I rely on the `lastDocId`, which should be zeroed more strictly.
I can replace the old implementation benchmarks with a more "manual" tracking so we don't have to worry about current implementation behavior

JoanFM: Yes, this is my concern, the old one

JoanFM (approved):

PR #6279, opened 2025-06-08
lock before opening indexes if info - [MOD-10007]

BenGoldberger:
Move the locking of the spec in IndexInfoCommand->fillReplyWithIndexInfo
Hopefully this will fix the bug and some flaky tests

### DIFF
--- a/src/info/info_command.c
+++ b/src/info/info_command.c
@@ -116,7 +116,10 @@ void fillReplyWithIndexInfo(RedisSearchCtx* sctx, RedisModule_Reply *reply, bool
 
   RedisModule_ReplyKV_Array(reply, "attributes"); // >attributes
   size_t geom_idx_sz = 0;
-
+  
+  // Lock the spec
+  RedisSearchCtx_LockSpecRead(sctx);
+  
   for (int i = 0; i < sp->numFields; i++) {
     RedisModule_Reply_Map(reply); // >>field
 
@@ -228,9 +231,6 @@ void fillReplyWithIndexInfo(RedisSearchCtx* sctx, RedisModule_Reply *reply, bool
   }
 
   RedisModule_Reply_ArrayEnd(reply); // >attributes
-
-  // Lock the spec
-  RedisSearchCtx_LockSpecRead(sctx);
 
   REPLY_KVINT("num_docs", sp->stats.numDocuments);
   REPLY_KVINT("max_doc_id", sp->docs.maxDocId);


src/info/info_command.c
```diff
@@ -116,7 +116,10 @@ void fillReplyWithIndexInfo(RedisSearchCtx* sctx, RedisModule_Reply *reply, bool
 
   RedisModule_ReplyKV_Array(reply, "attributes"); // >attributes
   size_t geom_idx_sz = 0;
-
+  
+  // Lock the spec
```
kei-nan: Should this be locked even earlier?
Are we sure `renderIndexOptions` or `renderIndexDefinitions` don't need the lock as well?

kei-nan (dismissed):

tests/pytests/test_gc.py
```diff
@@ -278,6 +280,136 @@ def testAutoMemory_MOD_3951():
 
     # This test should catch some leaks on the sanitizer
 
+def testConcurrentFTInfoDuringIndexDeletion(env):
```
alonre24, line 283: Nice test! Though I'm not sure that this is the right use case to reproduce the corruption that we saw, when the index is deleted, the GC will immediately exit and deallocate itself, without accessing the spec data structures. 
The right use case to test IMO is deleting documents and having the GC running while we call ft.info

tests/pytests/test_gc.py
```diff
@@ -278,6 +280,136 @@ def testAutoMemory_MOD_3951():
...
+                    results['info_calls'] += 1
+                    results['errors'] += 1
+                    # Expected errors when index is being deleted:
+                    # - "Unknown index name" or "no such index"
+                    error_msg = str(e).lower()
+                    if 'unknown index' in error_msg or 'no such index' in error_msg:
+                        # These are expected errors during index deletion
+                        pass
+                    else:
+                        # Unexpected error - log it but don't fail the test
```
alonre24: why without failing the test if the error is unexpected?

tests/pytests/test_gc.py
```diff
@@ -278,6 +280,136 @@ def testAutoMemory_MOD_3951():
...
+        threads.append(thread)
+        thread.start()
+
+    # Let the threads run for a short time to establish baseline
+    time.sleep(0.5)
+
+    # Now delete the indexes while FT.INFO calls are running
+    for idx_name in index_names:
+        try:
+            env.expect('FT.DROP', idx_name).equal('OK')
```
alonre24: Always use `DROPINDEX`, `DROP` command is deprecated

tests/pytests/test_gc.py
```diff
@@ -278,6 +280,136 @@ def testAutoMemory_MOD_3951():
...
+            env.debugPrint(f"Dropped index {idx_name}", force=True)
+        except Exception as e:
+            env.debugPrint(f"Error dropping index {idx_name}: {e}", force=True)
```
alonre24: why do we need these debug prints?

src/fork_gc.c
```diff
@@ -93,17 +94,21 @@ static void FGC_sendTerminator(ForkGC *fgc) {
 }
 
 static int __attribute__((warn_unused_result)) FGC_recvFixed(ForkGC *fgc, void *buf, size_t len) {
-  while (len) {
-    ssize_t nrecvd = read(fgc->pipefd[GC_READERFD], buf, len);
+  while (poll(fgc->pollfd_read, 1, 180000) == 1) {
```
alonre24, line 98: Please document what this call to `poll` does here

src/fork_gc.c
```diff
@@ -93,17 +94,21 @@ static void FGC_sendTerminator(ForkGC *fgc) {
...
+    ssize_t nrecvd = read(fgc->pipe_read_fd, buf, len);
     if (nrecvd > 0) {
       buf += nrecvd;
       len -= nrecvd;
-    } else if (nrecvd < 0 && errno != EINTR) {
+    } else if (nrecvd <= 0 && errno != EINTR) {
```
alonre24, line 103: What should we do if `nrecvd <= 0 && errno == EINTR` ? 
Also, should we increase the log level here? In which scenarios are we expected to reach here?

src/fork_gc.c
```diff
@@ -93,17 +94,21 @@ static void FGC_sendTerminator(ForkGC *fgc) {
...
       RedisModule_Log(fgc->ctx, "verbose", "ForkGC - got error while reading from pipe (%s)", strerror(errno));
       return REDISMODULE_ERR;
     }
+    if (len == 0) {
+      return REDISMODULE_OK;
+    }
   }
-  return REDISMODULE_OK;
+  RedisModule_Log(fgc->ctx, "verbose", "ForkGC - got error while reading from pipe (%s)", strerror(errno));
```
alonre24: Consider increasing the log level here and elaborating the log message with more relevant info for troubleshooting if possible

src/fork_gc.c
```diff
@@ -1300,12 +1305,17 @@ static int periodicCb(void *privdata) {
...
+  int rc = pipe(pipefd);  // create the pipe
   if (rc == -1) {
     RedisModule_Log(ctx, "warning", "Couldn't create pipe - got errno %d, aborting fork GC", errno);
     IndexSpecRef_Release(early_check);
     return 1;
   }
+  gc->pipe_read_fd = pipefd[GC_READERFD];
+  gc->pipe_write_fd = pipefd[GC_WRITERFD];
+  gc->pollfd_read[0].fd = gc->pipe_read_fd;
+  gc->pollfd_read[0].events = POLLIN;
```
alonre24, line 1320: Can you document the purpose of `gc->pollfd_read`?

tests/pytests/test_gc.py
```diff
@@ -278,6 +280,145 @@ def testAutoMemory_MOD_3951():
...
+            env.expect('FT.DROPINDEX', idx_name).equal('OK')
+        except Exception as e:
+            env.assertTrue(False, message=f"Unexpected error dropping index {idx_name}: {e}")
+
+        # Force GC to clean up the deleted index
+        try:
+            forceBGInvokeGC(env, idx_name)
+        except Exception as e:
+            # Expected - index might already be cleaned up
+            pass
```
alonre24: Why calling `forceBGInvokeGC` and not `forceInvokeGC`? Also, don't we expect that the call will fail in all cases since the index is supposed to be deleted? Let's try to keep this test deterministic and predictable as much as possible.

tests/pytests/test_gc.py
```diff
@@ -278,6 +280,145 @@ def testAutoMemory_MOD_3951():
...
+    # Verify that we made a reasonable number of FT.INFO calls
+    env.assertGreater(results['info_calls'], 50,
+                     message=f"Expected at least 50 FT.INFO calls, got {results['info_calls']}")
+
+    # Verify that we had both successful calls and expected errors
+    env.assertGreater(results['successful_calls'], 0,
+                     message=f"Expected some successful FT.INFO calls, got {results['successful_calls']}")
+
+    # Log the results for debugging
+    # env.debugPrint(f"FT.INFO test results: {results['info_calls']} total calls, {results['successful_calls']} successful, {results['errors']} errors", force=True)
```
alonre24: remove

tests/pytests/test_gc.py
```diff
@@ -278,6 +280,145 @@ def testAutoMemory_MOD_3951():
...
+    for thread in threads:
+        thread.join(timeout=5.0)  # 5 second timeout for thread cleanup
+
+    # Verify that we made a reasonable number of FT.INFO calls
+    env.assertGreater(results['info_calls'], 50,
+                     message=f"Expected at least 50 FT.INFO calls, got {results['info_calls']}")
+
+    # Verify that we had both successful calls and expected errors
+    env.assertGreater(results['successful_calls'], 0,
+                     message=f"Expected some successful FT.INFO calls, got {results['successful_calls']}")
```
alonre24: These checks are naturally flaky, and we should avoid them. Try to come up with a different validation that is valid in all cases and does not rely on timing.

src/fork_gc.c
BenGoldberger: if nrecvd < 0 we got read error
if nrecvd==0 read returned EOF. but we haven't reach the expected len param

src/fork_gc.c
BenGoldberger: it just a struct for the poll function, I initialized it here instead of doing it with every call to  `FGC_recvFixed`

tests/pytests/test_gc.py
BenGoldberger: I the gc will no run in the background can the module still answer `info` calls? (is it still a scenario that both processes accesses the same dict?)

tests/pytests/test_gc.py
BenGoldberger: I can assert greater then 0 for both cases? 
that would be the less flaky

BenGoldberger: we can remove the checks. but I think we want to know that the info calls were called

src/fork_gc.c
alonre24: I understand. But you haven't answered my questions :)

tests/pytests/test_gc.py
alonre24: Right, got it

tests/pytests/test_gc.py
alonre24: If you can come up with a testing mechanism that will ensure no flakiness and still verify that info calls were done, that would be optimal. But if not, we should remove these checks

src/fork_gc.c
BenGoldberger: It the same behavior as before, except we now handle the EOF scenario, that was a potential endless loop before.
We can increase the log level.

tests/pytests/test_gc.py
BenGoldberger: We do sleep for a second to make sure some info calls gets through...

src/fork_gc.c
alonre24: I'm wondering if we should seize the opportunity to make sure there are no other options to get an infinite loop.

tests/pytests/test_gc.py
alonre24: That helps, but still leaves room for flakiness in noisy environments. Can we use some kind of a shared variable to check that at least one info call was made instead of relying on the sleep?

src/fork_gc.c
BenGoldberger: I dont see how. len is decreased, and if not we return, I'll change the len condition to return if len <=0 to make sure we exit the loop.

And poll will timeout after 3 minutes

tests/pytests/test_gc.py
BenGoldberger: I changed the sleep to a while loop, and deleted one of the asserts

src/gc.c
```diff
@@ -170,7 +170,7 @@ static void GCContext_UnblockClient(void* data) {
 }
 
 void GCContext_WaitForAllOperations(RedisModuleBlockedClient* bc) {
-  redisearch_thpool_add_work(gcThreadpool_g, GCContext_UnblockClient, bc, THPOOL_PRIORITY_HIGH);
+  redisearch_thpool_add_work(gcThreadpool_g, GCContext_UnblockClient, bc, THPOOL_PRIORITY_LOW);
```
alonre24: What is the reason for this change?

tests/pytests/test_gc.py
```diff
@@ -278,6 +280,140 @@ def testAutoMemory_MOD_3951():
...
+    threads = []
+    for idx_name in index_names:
+        thread = threading.Thread(target=ft_info_worker, args=(idx_name,))
+        thread.daemon = True
+        threads.append(thread)
+        thread.start()
+
+    # Let the threads run for a short time to establish baseline
+    while results['info_calls'] < 10:
+        time.sleep(0.1)
```
alonre24, line 364: Would it be better to move it to after we delete the docs so we verify info calls were executed in parallel to GC?

tests/pytests/test_gc.py
```diff
@@ -278,6 +280,140 @@ def testAutoMemory_MOD_3951():
...
+            env.expect('FT.DROPINDEX', idx_name).equal('OK')
+        except Exception as e:
+            env.assertTrue(False, message=f"Unexpected error dropping index {idx_name}: {e}")
+
+        # Force GC to clean up the deleted index
+        try:
+            forceBGInvokeGC(env, idx_name)
+        except Exception as e:
+            # Expected - index might already be cleaned up
+            pass
```
alonre24: Can we validate the expected error, and fail if an unexpected error is caught?

tests/pytests/test_gc.py
BenGoldberger: I think here it grantee to run in parallel, if it was after the deletion in theory the deletion could have finished before the info workers started

src/fork_gc.c
```diff
@@ -64,7 +65,7 @@ static void FGC_updateStats(ForkGC *gc, RedisSearchCtx *sctx,
 // Buff shouldn't be NULL.
 static void FGC_sendFixed(ForkGC *fgc, const void *buff, size_t len) {
   RS_LOG_ASSERT(len > 0, "buffer length cannot be 0");
-  ssize_t size = write(fgc->pipefd[GC_WRITERFD], buff, len);
```
kei-nan, line 67: @BenGoldberger how are these GC changes related to the info changes and the locking placement?

kei-nan: If they are not, then maybe move them to a different PR?

BenGoldberger: It a small fix to an issue that came up while testing this, so I think it fine to include it here.

alonre24: The GC changes are not related directly, but they are needed to deflake tests. We could move it to a different PR, but since it is a small fix and PR is ready, I suggest we leave it as is. @BenGoldberger, please update the PR description and the MOD tickets that are related

PR #6280, opened 2025-06-08
Stop GC if OOM [MOD-10069]

lerman25:
This PR introduces a stop mechanism for the GC if OOM.
Before the GC fork, we check if `used_memory`  is lower then memory limit (either `maxmemory`, `max_process_mem` or `total_system_memory`), if not we set a gc retry and stop the callback

### DIFF
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -1314,6 +1314,7 @@ static int periodicCb(void *privdata) {
   if (isOutOfMemory(ctx)) {
     RedisModule_Log(ctx, "warning", "Not enough memory for GC fork, skipping GC job");
     gc->retryInterval.tv_sec = RSGlobalConfig.gcConfigParams.forkGc.forkGcRetryInterval;
+    IndexSpecRef_Release(early_check);
     RedisModule_ThreadSafeContextUnlock(ctx);
     return 1;
   }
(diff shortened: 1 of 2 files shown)


src/fork_gc.c
```diff
@@ -1243,10 +1243,35 @@ FGCError FGC_parentHandleFromChild(ForkGC *gc) {
   return status;
 }
 
+static inline bool isOutOfMemory(RedisModuleCtx *ctx) {
+  #define MIN_NOT_0(a,b) (((a)&&(b))?MIN((a),(b)):MAX((a),(b)))
+  RedisModuleServerInfoData *info = RedisModule_GetServerInfo(ctx, "memory");
```
GuyAv46, line 1249: Did we verify that all of these APIs are safe to call without the GIL?

src/fork_gc.c
```diff
@@ -1243,10 +1243,35 @@ FGCError FGC_parentHandleFromChild(ForkGC *gc) {
...
+  return used_memory > maxmemory;
+}
+
 static int periodicCb(void *privdata) {
   ForkGC *gc = privdata;
   RedisModuleCtx *ctx = gc->ctx;
 
+  // Check if we are out of memory before even trying to fork
+  if (isOutOfMemory(ctx)) {
+    RedisModule_Log(ctx, "warning", "fork failed - not enough memory, aborting fork GC");
```
GuyAv46: IMO, "fork failed" is misleading. Maybe "Skipping GC job" and maybe additional info we can share (e.g, index name)

GuyAv46 (commented):
Nice and simple

Do we plan to check the memory limits (including COW) while the fork is already running, and stop the GC after it has begun?

src/fork_gc.c
lerman25: Agree about the "fork failed". 
Adding index name is not aligned with the other warnings in this function, should I still add this information?

src/fork_gc.c
lerman25: Not safe, moving call to after taking the GIL and adding comment to function.

GuyAv46 (approved):

PR #6281, opened 2025-06-09
Use the environment to locate bash

chesedo:

This makes the scripts more portable. On some systems `bash` is not located at `/bin/bash` so this uses the `env` command instead to correctly locate the location of `bash` on all systems.

#### Mark if applicable


alonre24 (approved):
SHEBANG

PR #6287, opened 2025-06-09
MOD-9769: Implement varint encoding/decoding in Rust

zeenix:

This PR re-implements Varint encoding & decoding in Rust. It also adds benchmarks that compare the
performance and memory usage of C vs Rust implementations.

A recent sample run of the benchmarks on my laptop:

```
❯ cargo bench

Varint|Encode/Rust      time:   [660.06 ns 662.24 ns 665.10 ns]
                        change: [+0.9330% +2.4229% +3.9848%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 10 outliers among 100 measurements (10.00%)
  7 (7.00%) high mild
  3 (3.00%) high severe
Varint|Encode/C         time:   [665.43 ns 666.27 ns 667.36 ns]
                        change: [+2.6780% +4.0123% +5.3601%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 17 outliers among 100 measurements (17.00%)
  1 (1.00%) high mild
  16 (16.00%) high severe

Varint|Decode/Rust      time:   [135.24 ns 135.37 ns 135.50 ns]
Found 12 outliers among 100 measurements (12.00%)
  4 (4.00%) low severe
  3 (3.00%) low mild
  2 (2.00%) high mild
  3 (3.00%) high severe
Varint|Decode/C         time:   [319.02 ns 319.54 ns 320.19 ns]
Found 11 outliers among 100 measurements (11.00%)
  2 (2.00%) low severe
  1 (1.00%) low mild
  1 (1.00%) high mild
  7 (7.00%) high severe

Varint|Encode FieldMask/Rust
                        time:   [672.07 ns 673.37 ns 675.06 ns]
Found 12 outliers among 100 measurements (12.00%)
  1 (1.00%) high mild
  11 (11.00%) high severe
Varint|Encode FieldMask/C
                        time:   [668.89 ns 670.55 ns 672.26 ns]
Found 19 outliers among 100 measurements (19.00%)
  7 (7.00%) high mild
  12 (12.00%) high severe

Varint|Decode FieldMask/Rust
                        time:   [239.62 ns 239.89 ns 240.21 ns]
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe
Varint|Decode FieldMask/C
                        time:   [351.79 ns 352.80 ns 353.91 ns]
Found 19 outliers among 100 measurements (19.00%)
  8 (8.00%) low severe
  2 (2.00%) high mild
  9 (9.00%) high severe

Varint Vector Writer|Vector Writer/Rust
                        time:   [582.04 ns 585.19 ns 588.42 ns]
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe
Varint Vector Writer|Vector Writer/C
                        time:   [658.85 ns 660.69 ns 662.67 ns]
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe
```

As you can see the performance is similar, except in case of:

* Vector writer, which is about 100 ns faster than the C implementation. I suspect this is due to
  Rust side using a `Vec` and C side using the internal Buffer API and the former being more
  efficient than latter at allocations/growth.

* Decoder, which is unbelievably faster than the C implementation. The reason is that to be able to
  call the C decoder functions from Rust, I had to create non-inlined wrappers around the
  `static inline` ones and hence that removes the advantage of inlining from the C side. If you add
  `#[inline(never)]` to the Rust functions, the benchmark results become quite comparable for Varint
  and Rust's FieldMask decoding becomes 2x slower than its C counterpart:

```
Varint|Decode/Rust      time:   [309.14 ns 310.67 ns 312.31 ns]
                        change: [+126.58% +129.49% +132.24%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
Varint|Decode/C         time:   [310.92 ns 311.30 ns 311.77 ns]
                        change: [-3.3777% -1.5450% +1.1390%] (p = 0.23 > 0.05)
                        No change in performance detected.
Found 20 outliers among 100 measurements (20.00%)
  1 (1.00%) high mild
  19 (19.00%) high severe

Varint|Decode FieldMask/Rust
                        time:   [695.86 ns 696.72 ns 697.73 ns]
                        change: [+186.12% +190.35% +193.93%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 18 outliers among 100 measurements (18.00%)
  2 (2.00%) low mild
  2 (2.00%) high mild
  14 (14.00%) high severe
Varint|Decode FieldMask/C
                        time:   [352.47 ns 353.46 ns 354.49 ns]
                        change: [-1.0641% +0.2702% +1.6579%] (p = 0.75 > 0.05)
                        No change in performance detected.
Found 16 outliers among 100 measurements (16.00%)
  5 (5.00%) low severe
  2 (2.00%) low mild
  5 (5.00%) high mild
  4 (4.00%) high severe
```

### Side notes:

* This PR was split from #5996, rebased and then benchmarks were added on top.

* Once this PR is merged, the next step would be to drop the use of the C implementation in favor of this.

#### Mark if applicable


src/redisearch_rs/encode_decode/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+license-file.workspace = true
+
+[lints]
+workspace = true
+
+[lib]
+crate-type = ["staticlib", "rlib"]
+
+[dependencies]
+ffi.workspace = true
```
chesedo: I don't see this being used in the crate. Should be removable.

src/redisearch_rs/encode_decode/src/varint/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+/// Read an encoded integer from the given reader.
+pub fn read<R>(mut read: R) -> io::Result<u32>
+where
+    R: Read,
+{
+    let mut buffer = [0; 1];
+    read.read_exact(&mut buffer)?;
+    let mut c = buffer[0];
+
+    let mut val = (c & 127) as u32;
```
chesedo, line 40: Can we make the `127` a byte constant so that it is easier to see this is the value `0111_1111` and therefore easier to see the effect of this AND operation

src/redisearch_rs/encode_decode/src/varint/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+        val = (val << 7) | ((c & 127) as FieldMask);
+    }
+
+    Ok(val)
+}
+
+/// Encode an integer into a varint format and write it to the given writer.
+///
+/// The number of bytes written.
+pub fn write<W>(value: u32, mut write: W) -> io::Result<usize>
```
chesedo:
```suggestion
/// Encode an integer into a varint format and write it to the given writer.
///
/// # Return Value
///
/// The number of bytes written.
pub fn write<W>(value: u32, mut write: W) -> io::Result<usize>
```

src/redisearch_rs/encode_decode/src/varint/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+impl DerefMut for VarintBuf {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
chesedo, line 162: Shouldn't we have one or two tests just to make sure the write output is correct?

src/redisearch_rs/varint_bencher/src/c_varint.rs
```diff
@@ -0,0 +1,323 @@
...
+    #[inline(always)]
+    pub fn bytes_data(&self) -> *const u8 {
+        // SAFETY: self.0 is a valid VarintVectorWriter pointer, accessing buf.data field
+        unsafe { (*self.0).buf.data as *const u8 }
+    }
+
+    /// Truncate the vector writer.
+    /// Returns the new capacity.
+    #[inline(always)]
+    pub fn shrink_to_fit(&mut self) -> usize {
```
chesedo: Are all these methods used / needed?

src/varint.c
```diff
@@ -120,3 +120,76 @@ size_t VVW_Write(VarintVectorWriter *w, uint32_t i) {
...
+// Non-inline wrapper functions for FFI to ensure these are available as exported symbols.
+uint32_t ReadVarintNonInline(BufferReader *b) {
+  return ReadVarint(b);
+}
+
+t_fieldMask ReadVarintFieldMaskNonInline(BufferReader *b) {
+  return ReadVarintFieldMask(b);
+}
+
+t_fieldMask ReadVarintFieldMask(BufferReader *b) {
```
chesedo: I understand the need for the inline functions (to have a comparable benchmark). But why are these needed? Did the C code not have readers for the field masks?

chesedo: Nvm, I found them in the headers

src/redisearch_rs/varint_bencher/README.md
```diff
@@ -0,0 +1,117 @@
...
+between Rust and C, focusing on memory efficiency and space usage.
+
+## Overview
+
+This crate provides:
+- **Memory usage analysis**: Compare space efficiency between Rust and C varint
+  implementations
+- **Compression ratios**: Analyze how well varint encoding compresses different
+  value ranges
+- **Implementation correctness**: Verify that Rust and C implementations produce
```
chesedo, line 13: Where are the Rust and C results compared?

chesedo (commented):
Are we okay with the field mask decoding being slower in Rust (when inlining is disabled)? Do we know why it is slower?

Also did the memory usage show they have the same compression?

zeenix:
> Are we okay with the field mask decoding being slower in Rust (when inlining is disabled)?

We don't disable inlining so I don't think we need to be concerned.

> Also did the memory usage show they have the same compression?

Yes. I should have included the output of that here. I'll add that soon.

src/redisearch_rs/varint_bencher/README.md
zeenix: To ensure they're doing the exact same thing?

src/redisearch_rs/varint_bencher/src/c_varint.rs
zeenix: Good question. I'll check and remove ones that aren't used.

src/redisearch_rs/encode_decode/src/varint/mod.rs
zeenix: I thought I had one. I'll add it.

zeenix: wait. there is one: `test_variant` function below. Maybe I misunderstood your question/comment? :thinking:

src/redisearch_rs/encode_decode/src/varint/mod.rs
zeenix: I'd rather keep it as if for now, so it's easy to see that the Rust impl is identical to the C one.

chesedo: Hmm, but we don't really care about the implementation being the same. We only care about the results / effects being the same. And it will make the code easier to read / reason which means it's an improvement.

src/redisearch_rs/encode_decode/src/varint/mod.rs
chesedo: `test_variant` only checks the length of the buffer and not the actual values written to the buffer

src/redisearch_rs/varint_bencher/README.md
chesedo: Oh, sorry I meant where in the code? I didn't see anything that actually does the comparison

src/redisearch_rs/encode_decode/src/varint/mod.rs
zeenix: I thinks it makes it easier for the reviewer (especially folks not too familiar with Rust) to easily compare. I'm not objecting to your suggestion but I think it'd be best to do after this PR.

src/redisearch_rs/encode_decode/src/varint/mod.rs
zeenix: right, it doesn't test the encoded value, only checks the decoded value after reading it back.

chesedo:
> > Are we okay with the field mask decoding being slower in Rust (when inlining is disabled)?
> 
> We don't disable inlining so I don't think we need to be concerned.

My concern is more of, "what if that means the Rust inlined version is slower than the C version too?". We currently only know the inlined Rust version is faster than the non-inlined C version. And since the non-inlined Rust version is slower than the non-inlined C version, then maybe than means Rust is slower when inlined too.

src/redisearch_rs/encode_decode/src/varint/mod.rs
chesedo: I really don't see how anyone (a Rust or C person) would find `127` easier than a `0111_1111` for a binary operation

src/redisearch_rs/encode_decode/src/varint/mod.rs
chesedo: Yes, okay, this is the point I'm trying to make: we only know encoding + decoding gives the original value. That just means decoding does a 100% correct inverse of encoding. We don't actually know that the value encoding puts on the buffer is correct (except for the length). Encoding might have done a little endian while it was expected to do have a "big endian" instead (or whatever).

src/redisearch_rs/encode_decode/src/varint/mod.rs
zeenix: > I really don't see how anyone (a Rust or C person) would find `127` easier than a `0111_1111` for a binary operation

I didn't say they would. :)

src/redisearch_rs/encode_decode/src/varint/mod.rs
zeenix: > Yes, okay, this is the point I'm trying to make:

Yeah, I got it. :)

src/redisearch_rs/varint_bencher/README.md
zeenix: oops, sorry my bad. I read "Where" as "Why". :facepalm: You're right. This is from a previous iteration that was never pushed. I'll update.

zeenix:
> My concern is more of, "what if that means the Rust inlined version is slower than the C version too?". We currently only know the inlined Rust version is faster than the non-inlined C version. And since the non-inlined Rust version is slower than the non-inlined C version, then maybe than means Rust is slower when inlined too.

I understand but given that:

* We would never put `#[inline(never)]` on the Rust functions and hence always let the compiler+linker inline where it makes sense.
* We're talking a few hundred nanoseconds for millions of iterations here in the worst case scenario.

I'm not sure there is anything to worry about here. Besides, the code is identical so I'm not sure what can be done here.

src/redisearch_rs/encode_decode/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+version.workspace = true
+edition.workspace = true
+license-file.workspace = true
+publish.workspace = true
+
+[lints]
+workspace = true
+
+[lib]
+crate-type = ["staticlib", "rlib"]
```
chesedo: Just confirming: why is this a static library on its own?

src/redisearch_rs/encode_decode/src/varint/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+        vbuf[pos] = 128 | ((value & 127) as u8);
+        value >>= 7;
+    }
+
+    pos
+}
+
+/// A buffer used for encoding varints.
+#[repr(transparent)]
+struct VarintBuf([u8; 24]);
```
chesedo, line 139: Note for a future optimization: even a 128-bit int will only need 19 bytes to be encoded in this way

src/redisearch_rs/varint_bencher/src/main.rs
```diff
@@ -0,0 +1,275 @@
...
+                    -diff_pct
+                );
+            } else {
+                println!(
+                    "  C varint encoding is {:.1}% more space efficient",
+                    diff_pct
+                );
+            }
+        }
+        show_byte_differences(rust_encoded_bytes, c_encoded_bytes, "Varint");
```
chesedo: Do we need the extra messages? Why not just call `show_byte_differences`?

src/redisearch_rs/encode_decode/src/varint/mod.rs
chesedo: Let's leave this for a future improvement. I already made another note for a potential improvement too

src/redisearch_rs/encode_decode/Cargo.toml
zeenix: How would dynamic linking work? The library will have to be installed in a canonical location for the dynamic linker to load it at runtime then. As a static library, it just becomes part of the Redisearch binary.

src/redisearch_rs/encode_decode/src/varint/mod.rs
zeenix: good point but:

* we should keep in mind that this is only the stack of the using functions, not allocated.
* AFAIK multiple of 8 is preferred for sizes, for alignment reasons.

chesedo: Ahh, good point

src/redisearch_rs/encode_decode/Cargo.toml
chesedo: I mean it does not go through the `redisearch_rs` like all the other exposed functions. I might be missing something here, but don't remember this being needed on the trie crate :thinking:

zeenix: Ah you're right. The thing is that this work started before we had a `redisearch_rs` and `trie` did indeed this this (and where I copied this from) before that (see commit a0fb945982559db5d6841f8172259f30bd30692e). I'll update..

chesedo (approved):

src/redisearch_rs/varint_bencher/src/main.rs
```diff
@@ -0,0 +1,238 @@
...
+    if differences > 5 {
+        println!("    ... and {} more differences", differences - 5);
+    } else if differences == 0 && rust_bytes.len() != c_bytes.len() {
+        println!("    All common bytes identical, but different lengths");
+    } else if differences == 0 {
+        println!("    No differences found (this shouldn't happen)");
+    } else {
+        println!("    Total differences: {}", differences);
+    }
+
```
dor-forer, line 55: What is the 5? If it is just a random number, maybe it can be configurable or fixed?

src/redisearch_rs/varint_bencher/src/c_varint.rs
```diff
@@ -0,0 +1,311 @@
...
+        };
+        read(&mut temp_buffer)
+    }
+
+    /// Convenience wrapper for single field mask decoding that returns FieldMask.
+    /// Used by tests and non-performance-critical code.
+    pub fn read_to_field_mask(data: &[u8]) -> FieldMask {
+        let mut temp_buffer = ffi::Buffer {
+            data: data.as_ptr() as *mut i8,
+            offset: data.len(),
```
dor-forer, line 244: Isn't the offset supposed to be set to 0?

src/redisearch_rs/varint_bencher/src/c_varint.rs
```diff
@@ -0,0 +1,311 @@
...
+    /// Used by main.rs for memory analysis.
+    #[inline(always)]
+    pub fn write_to_vec(value: u32) -> Vec<u8> {
+        // SAFETY: RedisModule_Alloc is a valid function pointer provided by Redis module system
+        let alloc_fn = unsafe { crate::RedisModule_Alloc.unwrap() };
+
+        // SAFETY: Calling Redis allocator with valid size parameter
+        let data_ptr = unsafe { alloc_fn(16) };
+        if data_ptr.is_null() {
+            return Vec::new();
```
dor-forer, line 138: Don't we want to panic when we can't allocate? Otherwise, it will be hard to understand why it failed.

zeenix:
Just (clean) rebased on current master.

src/redisearch_rs/varint_bencher/src/bencher.rs
```diff
@@ -0,0 +1,353 @@
...
+    values.extend((0..100).map(|i| i * 1000));
+
+    values
+}
+
+fn encode_rust_benchmark<M: Measurement>(group: &mut BenchmarkGroup<'_, M>, values: &[u32]) {
+    let mut buf = Vec::with_capacity(1024);
+    group.bench_function("Rust", |b| {
+        b.iter(|| {
+            for &value in values {
```
dor-forer: Should buf be cleared or reset before each value, like in c benchmark?

src/varint.h
```diff
@@ -19,38 +19,23 @@
 extern "C" {
 #endif
 /* Read an encoded integer from the buffer. It is assumed that the buffer will not overflow */
-static inline uint32_t ReadVarint(BufferReader *b) {
+uint32_t ReadVarint(BufferReader *b);
```
GuyAv46: Shouldn't we keep an inlined variant until the migration is complete?
Consider not removing these implementations, and use them in `src/varint.c` in the `*NonInline` variants

src/redisearch_rs/encode_decode/src/varint/vector_writer.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// The number of members written to the vector.
+    #[inline(always)]
+    pub fn count(&self) -> usize {
+        self.n_members
+    }
+
+    /// Truncate the vector.
+    pub fn shrink_to_fit(&mut self) -> usize {
```
dor-forer, line 79: Maybe document what is the returned value

src/varint.h
```diff
@@ -19,38 +19,23 @@
...
-}
+t_fieldMask ReadVarintFieldMaskNonInline(BufferReader *b);
 
 size_t WriteVarint(uint32_t value, BufferWriter *w);
 
 size_t WriteVarintFieldMask(t_fieldMask value, BufferWriter *w);
 
+uint32_t ReadVarintRaw(const char **pos, const char *end);
+
+t_fieldMask ReadVarintFieldMaskRaw(const char **pos, const char *end);
```
GuyAv46: Where do we use the "raw" implementations?

src/redisearch_rs/varint_bencher/.gitignore
```diff
@@ -0,0 +1,5 @@
+target/
+flamegraph.svg
+*.profdata
+benchmark_data/
+criterion_reports/
```
dor-forer: empty line

src/varint.c
```diff
@@ -120,3 +120,76 @@ size_t VVW_Write(VarintVectorWriter *w, uint32_t i) {
...
+
+  uint32_t val = c & 127;
+  while (c >> 7) {
+    ++val;
+    c = BUFFER_READ_BYTE(b);
+    val = (val << 7) | (c & 127);
+  }
+
+  return val;
+}
```
GuyAv46: should be kept in the h file

src/varint.c
```diff
@@ -120,3 +120,76 @@ size_t VVW_Write(VarintVectorWriter *w, uint32_t i) {
...
+
+  t_fieldMask val = c & 127;
+  while (c >> 7) {
+    ++val;
+    c = BUFFER_READ_BYTE(b);
+    val = (val << 7) | (c & 127);
+  }
+
+  return val;
+}
```
GuyAv46: should be kept in the h file

src/redisearch_rs/varint_bencher/src/c_varint.rs
zeenix: AFAIK, `offset` is the length (the amount of bytes used) in the allocated buffer.

src/redisearch_rs/varint_bencher/src/c_varint.rs
zeenix: Yeah, you're right. Although keep in mind that all this will be removed as part of the follow-up PR that removes the C impl.

src/redisearch_rs/varint_bencher/src/main.rs
zeenix: Sure but I'm not sure it's worth it. We'll be dropping the C impl soon after and this code will also be deleted then.

src/redisearch_rs/varint_bencher/src/bencher.rs
zeenix: The `Buffer` wrapper is a C-specific thing and not needed for/by Rust so there is nothing to reset.

zeenix: Although now that you mention it, perhaps the C benchmark doesn't need to reset the buffer. I'll check..

src/redisearch_rs/varint_bencher/src/c_varint.rs
dor-forer: Ok,  I misunderstood, I thought it was the beginning of the buffer. Thanks.

src/redisearch_rs/encode_decode/src/varint/mod.rs
GuyAv46: I tend to agree with @chesedo on this one. We are all used to bits :)
But keeping it as 127 is also fine

src/redisearch_rs/encode_decode/src/varint/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+    io::{self, Read, Write},
+    ops::{Deref, DerefMut},
+};
+
+/// Read an encoded integer from the given reader.
+pub fn read<R>(mut read: R) -> io::Result<u32>
+where
+    R: Read,
+{
+    let mut buffer = [0; 1];
```
GuyAv46, line 36: What's the type here? Or what enforces the type of the buffer?
(I found [fn read_exact(&mut self, buf: &mut [u8]) -> Result<()>](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact) - so I guess it's a `u8`?)

src/redisearch_rs/encode_decode/src/varint/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+    read.read_exact(&mut buffer)?;
+    let mut c = buffer[0];
+
+    let mut val = (c & 127) as u32;
+    while c >> 7 != 0 {
```
GuyAv46, line 41: this can be `while c & 128 != 0`. I wonder if it affects the performance, or if the compiler is smart enough to take the most performing operation

src/redisearch_rs/encode_decode/src/varint/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+        read.read_exact(&mut buffer)?;
+        c = buffer[0];
+        val = (val << 7) | ((c & 127) as u32);
+    }
+
+    Ok(val)
+}
+
+/// Same as `read` but reads the value into a `FieldMask`.
+// FIXME: The logic is identical to `read` so we can use a macro here to avoid code duplication.
```
GuyAv46, line 52: Can't it be a generic implementation by the return type?

src/redisearch_rs/encode_decode/src/varint/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+        pos -= 1;
+        value -= 1;
+        vbuf[pos] = 128 | ((value & 127) as u8);
+        value >>= 7;
+    }
+
+    pos
+}
+
+// FIXME: The logic is identical to `encode` so we can use a macro here to avoid code duplication.
```
GuyAv46, line 122: Can't we use a generic implementation?
Shouldn't we make this one `#[inline(always)]` as well (I see you follow the existing implementation, but I think it's a mistake in the current implementation. Not sure if we'll see a difference though)

src/redisearch_rs/encode_decode/src/varint/vector_writer.rs
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    /// Write an integer to the vector.
+    ///
+    /// # Return Value
+    ///
+    /// The number of bytes written to the vector.
+    pub fn write(&mut self, value: u32) -> std::io::Result<usize> {
+        self.buffer.reserve_exact(16);
```
GuyAv46: Using `reserve_exact` because we want to be memory efficient?
I know this is the most similar implementation to what we have now, but personally, I don't trust that a lot of thinking or benchmarking went into the C implementation.

src/varint.h
zeenix: sure.

src/varint.h
zeenix: Sorry, that was an intermediate change that I somehow forgot to remove. Will remove it. :)

src/redisearch_rs/encode_decode/src/varint/mod.rs
zeenix: The usage tells the compiler the type so it can infer it. This is a `u8`, yes.

src/redisearch_rs/encode_decode/src/varint/mod.rs
zeenix: I can try but as I wrote to @chesedo here before, I want to keep the code exactly the same in this PR so its easier to compare and ensure its the same code, just in Rust.

src/redisearch_rs/encode_decode/src/varint/mod.rs
zeenix: I did try a generic impl first but you'd be surprised by how many generic constraints were needed. :) It ended up more LOCs than how it is now. :) That's why I think macro would be better approach here.

src/redisearch_rs/encode_decode/src/varint/mod.rs
zeenix: > Can't we use a generic implementation?

Same here.

> Shouldn't we make this one `#[inline(always)]` as well (I see you follow the existing implementation, but I think it's a mistake in the current implementation. Not sure if we'll see a difference though)

Oh good point. I think these functions should be inlined but I don't have a strong feeling either way. It does make a difference in performance though. As you can see in the results I shared, `encode` performs better than `encode_field_mask`.

src/redisearch_rs/encode_decode/src/varint/vector_writer.rs
zeenix: I added this to be close to the existing impl, yes. I also doubt it's needed. I'll see if if can be safely removed.

src/redisearch_rs/varint_bencher/src/bencher.rs
zeenix: Oh, I didn't realize that this is the writer part. It indeed does need to reset at least the `offset` but only resetting the offset bit does seem to make a small difference (C encoding becomes 10 nanoseconds faster on my machine).

zeenix:
All review comments addressed and rebased.

dor-forer (dismissed):

src/redisearch_rs/encode_decode/src/varint/vector_writer.rs
```diff
@@ -32,7 +32,6 @@ impl VectorWriter {
     ///
     /// The number of bytes written to the vector.
     pub fn write(&mut self, value: u32) -> std::io::Result<usize> {
-        self.buffer.reserve_exact(16);
```
GuyAv46: So now the call to `write` will reallocate the buffer if needed, right?

GuyAv46 (dismissed):

src/redisearch_rs/encode_decode/src/varint/vector_writer.rs
zeenix: yes, since it's being passed as the `&mut Vec<u8>` instead of as a mutable slice (`&mut [u8]`).

zeenix: `write` is taking a `Write` impl and the `Write` impl of `&mut Vec<T>` extends the buffer. This can be throughly test in my next PR where the C code and tests start using the Rust impl as well. I'll rebase that branch and ensure this before marking this for merging..

zeenix: OK. All tests seem to pass when C impl is removed so I think we're safe. :)

chesedo (approved):

src/redisearch_rs/encode_decode/Cargo.toml
zeenix: So turns out that the [CI breakage here](https://github.com/RediSearch/RediSearch/actions/runs/15608693488/job/43964191671?pr=6287) is caused by removing this. I tried to link against the `libredisearch_rs.a` but encountered scarry looking errors. :ghost: So for now I've reverted the change. So I'll have to leave this part to you. :)

zeenix: Nope, [it still fails](https://github.com/RediSearch/RediSearch/actions/runs/15614525613/job/43983775113?pr=6287). :( I'll revert the revert. :)

zeenix:
There is a strange linker error with the benchmarking code when building for coverage. I ran out of time before I could fix it (and I was under the impression that coverage job is still broken and hence optional). 

However, given that https://github.com/RediSearch/RediSearch/pull/6308 passes all checks and is based on this PR, I'd just close this and get that one merged. Up to you folks of course.

src/redisearch_rs/varint_bencher/src/bencher.rs
LukeMathWalker: We should create a fresh buffer for every iteration here (or resetting the writing cursor); otherwise, we end up measuring the growth algorithm of the buffer.

PR #6296, opened 2025-06-10
Implement API for optimized wildcard iterator - [MOD-9567]

GuyAv46:

Implement an API for an optimized wildcard iterator, using the new QueryIterator API.

The implementation itself isn't changing - we use an inverted index iterator or an empty iterator if the optimization is enabled, and the non-optimized implementation if not.

This PR doesn't include any new tests or benchmarks, as no new component is implemented, and all the internal iterators (non-optimized wildcard, inverted index, empty) are already tested

#### Mark if applicable

### DIFF
--- a/src/iterators/wildcard_iterator.c
+++ b/src/iterators/wildcard_iterator.c
@@ -75,19 +75,24 @@ QueryIterator *IT_V2(NewWildcardIterator_NonOptimized)(t_docId maxId, size_t num
   return ret;
 }
 
+QueryIterator *IT_V2(NewWildcardIterator_Optimized)(const RedisSearchCtx *sctx) {
+  RS_ASSERT(sctx->spec->rule->index_all);
+  if (sctx->spec->existingDocs) {
+    return NewInvIndIterator_GenericQuery(sctx->spec->existingDocs, sctx,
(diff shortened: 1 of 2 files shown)


src/iterators/wildcard_iterator.c
```diff
@@ -72,3 +74,25 @@ QueryIterator *IT_V2(NewWildcardIterator_NonOptimized)(t_docId maxId, size_t num
   ret->NumEstimated = WI_NumEstimated;
   return ret;
 }
+
+QueryIterator *IT_V2(NewWildcardIterator_Optimized)(const RedisSearchCtx *sctx) {
+  RS_ASSERT(sctx->spec->rule->index_all);
+  if (sctx->spec->existingDocs) {
+    return NewInvIndIterator_GenericQuery(sctx->spec->existingDocs, sctx,
```
JoanFM, line 81: I am just wondering, what about the other type of constructors defined in `inverted_index_iterator.c`? They are not used?

GuyAv46: When we maintain the “existing ids” inverted index, we only store the docIds, no additional text/numeric data. So for this iterator we only need the “generic” API.
The other inverted index APIs and logic are already being tested, as part of #6219

JoanFM (approved):

PR #6300, opened 2025-06-11
fix `ffi` CI build issues

JonasKruckenberg:
This PR is a simple one-line fix for the Rust `ffi` crate build issues. We don't need the `FP_` constants and since they cause trouble, lets just disable generation of them.

### DIFF
--- a/src/redisearch_rs/ffi/build.rs
+++ b/src/redisearch_rs/ffi/build.rs
@@ -46,7 +46,10 @@ fn main() {
     let mut bindings = bindgen::Builder::default();
 
     for header in headers {
-        bindings = bindings.header(header.display().to_string());
+        bindings = bindings
+            .header(header.display().to_string())
+            .allowlist_file(header.display().to_string());
+
         println!("cargo:rerun-if-changed={}", header.display());
     }
     for include in includes {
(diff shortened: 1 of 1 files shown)


src/redisearch_rs/ffi/build.rs
```diff
@@ -58,6 +58,9 @@ fn main() {
 
     let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
     bindings
+        // FP_NAN, FP_ZERO etc. are causing duplicate definition errors on linux, we don't need them anyways
+        // so just blocklist them all
+        .blocklist_item("FP_.*")
```
chesedo: For my current benchmarking code I'm finding that the [allowlist](https://rust-lang.github.io/rust-bindgen/allowlisting.html) works a bit better to not pull in the whole world.
image

chesedo: So basically only allow the files in the `headers` array. The `types_rs.h` blocklist is not needed but we will use it soon to prevent duplicate Rust structs in different crates

JonasKruckenberg: sure yeah, I like that more. Its more work when porting but cleaner overall
Edit: we can just use the array outright for most of the heavy lifting

src/redisearch_rs/ffi/build.rs
```diff
@@ -58,6 +61,7 @@ fn main() {
 
     let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
     bindings
+        .allowlist_file(".*/types_rs.h")
```
chesedo, line 64: Do you need this one? Would like to rather remove it

JonasKruckenberg: yup, doesn't build without as we moved types to Rust now

PR #6301, opened 2025-06-11
Migrate `RSIndexResult` to Rust

chesedo:
This PR moves the `RSIndexResult` struct to be defined in Rust. This is the last step to be able to more easily migrate the inverted index to Rust too.

This only differs to the C type in that the anonymous "data" union field now has a name. This required updating all the C files.

#### Mark if applicable


src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -91,6 +95,50 @@ pub struct RSAggregateResult {
     pub type_mask: RSResultTypeMask,
 }
 
+/// Holds the actual data of an ['IndexResult']
+#[repr(C)]
+pub union RSIndexResultData {
```
zeenix, line 97: Can't we switch to enums?

chesedo: Yeah, I forgot. Let me check...

chesedo: Okay, so switching to a Rust `enum` is going to cause `cbindgen` to add a whole new tag field to determine the enum variant. But there is already a `result_type` field for that. I propose we rather switch to enums once all this code is in Rust.

chesedo: image

zeenix: That's unfortunate but is it that bad for the C side? :thinking: 

> propose we rather switch to enums once all this code is in Rust.

If we go that route, we should add a FIXME/TODO comment and add a Jira ticket for the conversion.

However, I wonder if it'd be better to make this change and live with a slightly worse (than it already is) C API, which is nicer/safe on the Rust side already?

src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
```diff
@@ -9,4 +9,6 @@
 
 //! This module contains pure Rust types that we want to expose to C code.
 
-pub use inverted_index::{RSAggregateResult, RSNumericRecord, RSResultType, RSTermRecord};
+pub use inverted_index::{
+    RSAggregateResult, RSIndexResult, RSNumericRecord, RSResultType, RSTermRecord,
```
alonre24: Do we still need to expose all of these, or just `RSIndexResult`?

src/redisearch_rs/headers/types_rs.h
```diff
@@ -202,3 +194,64 @@ typedef struct RSTermRecord {
...
+  double value;
+} RSNumericRecord;
+
+/**
+ * Holds the actual data of an ['IndexResult']
+ */
+typedef union RSIndexResultData {
+  struct RSAggregateResult agg;
+  struct RSTermRecord term;
+  struct RSNumericRecord num;
```
alonre24, line 211: `RSVirtualRecord` not needed?

src/redisearch.h
```diff
@@ -290,48 +290,6 @@ typedef struct RSYieldableMetric{
   struct RSValue *value;
 } RSYieldableMetric;
 
-#pragma pack(16)
```
alonre24, line 293: Do we get this pack in Rust as well?

src/redisearch_rs/headers/types_rs.h
chesedo: No, nothing used it because the result type would indicate it is virtual.

src/redisearch.h
chesedo: Yes, Rust has `packed(N)`, but the [`cbindgen` does not support it](https://github.com/mozilla/cbindgen/issues/433) when creating the C files. I did a test to confirm the memory sizes and field offsets of the Rust version without packing is exactly the same as the C one with packing.

src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
chesedo: Great catch, updated in https://github.com/RediSearch/RediSearch/pull/6301/commits/a767d0956ec35eddd1902683a75c4d460292c47e

src/redisearch.h
alonre24: I see. Please add it to the PR description in case someone wonders in the future :)

alonre24 (approved):

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -37,18 +37,21 @@ impl From<Delta> for usize {
...
 pub struct RSNumericRecord(pub f64);
 
 /// Represents the encoded offsets of a term in a document. You can read the offsets by iterating
 /// over it with RSOffsetVector_Iterator
+#[derive(Clone, Copy)]
 #[repr(C)]
 pub struct RSOffsetVector {
     pub data: *mut c_char,
     pub len: u32,
 }
```
LukeMathWalker: Let's not derive `Copy` for types that contain raw pointers, it's a recipe for bugs down the line.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -37,18 +37,21 @@ impl From<Delta> for usize {
...
 
 /// Represents a single record of a document inside a term in the inverted index
+#[derive(Clone, Copy)]
```
LukeMathWalker: Same as above, re:Copy.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -77,6 +80,7 @@ pub type RSResultTypeMask = BitFlags<RSResultType, u32>;
 /// Represents an aggregate array of values in an index record.
 /// cbindgen:rename-all=CamelCase
 #[repr(C)]
+#[derive(Clone, Copy)]
```
LukeMathWalker: Same as above, re:Copy.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Okay, can do. This would force the `data` field union to be an enum instead (which Zeeshan also suggested).

chesedo: We are doing `ManuallyDrop` instead on the `union` types.

src/redisearch.h
chesedo: Description updated

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Created ticket MOD-10158

LukeMathWalker (approved):

PR #6307, opened 2025-06-12
[MOD-8843] - refactor NotIterator - NonOptimized 

JoanFM:

Split from #6238 

1. MOD-...
2. #...

1. ...

#### Mark if applicable


tests/cpptests/iterator_util.h
```diff
@@ -31,74 +33,90 @@ class MockIterator {
...
 
     template<typename... Args>
-    MockIterator(IteratorStatus st, Args&&... ids_args)
-        : docIds({ids_args...}), whenDone(st), nextIndex(0), readCount(0) {
-        setBase(&base);
-        std::sort(docIds.begin(), docIds.end());
-        auto new_end = std::unique(docIds.begin(), docIds.end());
-        docIds.erase(new_end, docIds.end());
+    MockIterator(IteratorStatus st, std::optional<std::chrono::milliseconds> sleep, Args&&... ids_args)
+      : docIds({ids_args...}), whenDone(st), nextIndex(0), readCount(0), sleepTime(sleep) {
```
GuyAv46: Should it also be forwarded?
```suggestion
      : docIds({std::forward<Args>(ids_args)...}), whenDone(st), nextIndex(0), readCount(0), sleepTime(sleep) {
```

tests/cpptests/micro-benchmarks/benchmark_not_iterator.cpp
```diff
@@ -0,0 +1,365 @@
...
+   public:                                                                \
+    BaseClass##_##Method##_Benchmark() {                                  \
+      this->SetName(#BaseClass "<" #a "," #b ", " #c ">/" #Method);       \
+    }                                                                     \
+                                                                          \
+   protected:                                                             \
+    void BenchmarkCase(::benchmark::State&) BENCHMARK_OVERRIDE;           \
+  };
+
+  #define BENCHMARK_TEMPLATE3_DEFINE_F(BaseClass, Method, a, b, c)    \
```
GuyAv46:
```suggestion
#define BENCHMARK_TEMPLATE3_DEFINE_F(BaseClass, Method, a, b, c)      \
```

tests/cpptests/micro-benchmarks/benchmark_not_iterator.cpp
```diff
@@ -0,0 +1,365 @@
...
+#include <random>
+#include <vector>
+
+#include "src/iterators/iterator_api.h"
+#include "src/iterators/not_iterator.h"
+
+#include "deprecated_iterator_util.h"
+#include "src/index.h"
+
+#define BENCHMARK_TEMPLATE3_PRIVATE_DECLARE_F(BaseClass, Method, a, b, c) \
```
GuyAv46: Consider moving these macros to `iterator_util.h`

tests/cpptests/micro-benchmarks/benchmark_not_iterator.cpp
```diff
@@ -0,0 +1,365 @@
...
+  Args({2'000'000, 1'000'000}) -> \
+  Args({10'000'000, 50'000'000}) -> \
+  Args({50'000'000, 10'000'000})
+
+#define SUBSET_SCENARIOS() \
+  Args({1'000, 100'000}) -> \
+  Args({10'000, 500'000}) -> \
+  Args({100'000, 1'000'000}) -> \
+  Args({1'000'000, 2'000'000}) -> \
+  Args({10'000'000, 50'000'000})
```
GuyAv46: I think we can have better scenarios
1. The case where the child has more results than the wildcard is possible (with optimized wildcard only), but it is very extreme. We can also expect at most the size of a block, because only the last block may contain irrelevant IDs that were not cleaned. Bottom line - this is an important edge case to test, but not so much to benchmark
2. The number of docs the iterator returns should be high, so the benchmark doesn't have to rewind too often and skew the benchmark results. We should expect the same result from `{100'000, 1'000'000}` and `{10'000, 100'000}`. A better parameter would be the ratio between the wildcard result set size and the child result set size.
I suggest having a single value representing this ratio, ranging from 1 (1:1) to 100,000 (1:100,000). You can then choose the right number of docs for the child to get this ratio, or for every ID in the range of the wildcard, ["flip a coin"](https://github.com/RediSearch/RediSearch/pull/6322/files#diff-7c9e405ba2c510317a116ce21303d5a006be9a76ff21ea7d4fff61ac4e32560aR49) to decide whether to take it for the child or not. Use `RangeMultiplier(x)->Range(1, y)` for a scenario generation

tests/cpptests/test_cpp_iterator_not.cpp
```diff
@@ -0,0 +1,649 @@
...
+  void SetUp() override {
+    // Get child document IDs from parameter
+    std::tie(childDocIds, wcDocIds, optimized) = GetParam();
+    // Consider wcDocIds only if optimized is true
+
+    // Find the maximum document ID
+    maxDocId = 0;
+    for (auto id : childDocIds) {
+      maxDocId = std::max(maxDocId, id);
+    }
```
GuyAv46:
```suggestion
    maxDocId = *std::max_element(childDocIds.begin(), childDocIds.end());
```
```suggestion
    maxDocId = *std::ranges::max_element(childDocIds);
```

tests/cpptests/test_cpp_iterator_not.cpp
```diff
@@ -0,0 +1,649 @@
...
+  ::testing::Combine(
+    ::testing::Values(
+      std::vector<t_docId>{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
+      std::vector<t_docId>{2, 4, 6, 8, 10},
+      std::vector<t_docId>{5, 10, 15, 20, 25, 30},
+      std::vector<t_docId>{1, 3, 5, 7, 9, 11, 13, 15},
+      std::vector<t_docId>{1, 2, 3, 4, 5, 6, 100, 150},
+      std::vector<t_docId>{1, 2, 3, 6, 10, 15},
+      std::vector<t_docId>{500, 600, 700, 800, 900, 1000},
+      // Add long random list
```
GuyAv46: Why? What are we testing here?

tests/cpptests/test_cpp_iterator_not.cpp
```diff
@@ -0,0 +1,649 @@
...
+
+#include "gtest/gtest.h"
+#include "iterator_util.h"
+
+#include <random>
+#include <vector>
+
+#include "src/iterators/not_iterator.h"
+
+class NotIteratorCommonTest : public ::testing::TestWithParam<std::tuple<std::vector<t_docId>, std::vector<t_docId>, bool>> {
```
GuyAv46: This test suite is very thorough, but I believe it has a lot of redundancy.
I suggest having a common test with a single fixed (can be random) setup, no need for a large combination matrix for testing basic behavior.
For specific scenarios worth checking, use short and specific child iterators to cause the desired scenario.
As it is now, it's hard to follow what are we trying to test

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,165 @@
...
+  if (ni->wcii) {
+    ni->wcii->Free(ni->wcii);
+  }
+  IndexResult_Free(base->current);
+  rm_free(base);
+}
+
+static size_t NI_NumEstimated(QueryIterator *base) {
+  NotIterator *ni = (NotIterator *)base;
+  return ni->wcii ? ni->wcii->NumEstimated(ni->wcii) : ni->maxDocId;
```
GuyAv46, line 37: We better store the number of docs in the index and use it here instead of `maxDocId`

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,165 @@
...
+    rc = ni->child->Read(ni->child);
+    if (rc == ITERATOR_TIMEOUT) return rc;
+    // Check for timeout with low granularity (MOD-5512)
+    if (TimedOut_WithCtx_Gran(&ni->timeoutCtx, 5000)) {
+      base->atEOF = true;
+      return ITERATOR_TIMEOUT;
+    }
+  }
+  base->atEOF = true;
+  return ITERATOR_EOF;
```
GuyAv46, line 74: We can add a test scenario that leads us here

tests/cpptests/iterator_util.h
```diff
@@ -31,74 +33,90 @@ class MockIterator {
     size_t nextIndex;
     IteratorStatus whenDone;
     size_t readCount;
+    std::optional<std::chrono::milliseconds> sleepTime; // Sleep for this duration before returning from Read/SkipTo
```
GuyAv46: Why do we want the mock to sleep for? And where do we use it? I see the timeout-related code is still uncovered

src/iterators/not_iterator.c
```diff
@@ -0,0 +1,165 @@
...
+      return ITERATOR_OK;
+    }
+  }
+
+  // If we are here, Child has DocID (either already lastDocID == docId or the SkipTo returned OK)
+  // We need to return NOTFOUND and set the current result to the next valid docId
+  base->current->docId = base->lastDocId = docId;
+  IteratorStatus rc = NI_Read_NotOptimized(base);
+
+  return rc == ITERATOR_OK ? ITERATOR_NOTFOUND : rc;
```
GuyAv46, line 112: In case of timeout/eof (`rc != ITERATOR_OK`), we should reset `base->lastDocId`

tests/cpptests/test_cpp_iterator_not.cpp
JoanFM: Just gaining certainity about potential random scenarios.

tests/cpptests/iterator_util.h
JoanFM: I will review the uncovered things, but the idea is to test the ITERATOR_TIMEOUT scenarios

src/iterators/not_iterator.c
JoanFM: in case of EOF also?

src/iterators/not_iterator.c
JoanFM: ok, this can be an scenario where maxDocID is smaller than the last child Read. Will do

src/iterators/not_iterator.c
JoanFM: why should be done here? and not in the Read?

src/iterators/not_iterator.c
JoanFM: Would it make sense then u mean to change maxDocId in the constructor and return it here? or we may want to have another wcii-estimated and still loop until maxDocID?. First option seems best perhaps?

src/iterators/not_iterator.c
JoanFM: In `IDList`iterator for instance, this is not reset in EOF case

JoanFM:
Since @BenGoldberger had approved previously the Optimized+NonOptimized implementation, I proceed with merging when ready

PR #6311, opened 2025-06-13
Rust Implementation of RSSortingVector - MOD-9985

DarthB:

The PR introduces two changes:
Adds `RSSortingVector` Rust implementation + Unit Tests

## More detailed changes

- Implements the `RSSortingVector` as a Boxed Slice in Rust
- Extracts C-Code `normalizeStr` which implements utf case folding based on nulibI
- Add an `RSValueTrait` that is implemented by mocks and the final FFI implementation 
- Adds unit tests with mock type of `RSValue`

## Referenced PRs

Previous PRs:
- [x] Relies on Refactorings in: https://github.com/RediSearch/RediSearch/pull/6312

- [x] Relies on API changes in: https://github.com/RediSearch/RediSearch/pull/6426
  String normalization becomes caller's responsibility in the C API for `RSSortingVector`

Upcoming PRs:
- FFI Layer: https://github.com/RediSearch/RediSearch/pull/6459
- Full swap of C with Rust Implementation: https://github.com/RediSearch/RediSearch/pull/6397

Archived PRs:
- Draft for a new-type usable to ensure sortable strings are normalized in: https://github.com/RediSearch/RediSearch/pull/6427
  Implement the new type `NormalizedString` in Rust to ensure normalized Strings in `RSSortingVector`
- Old Linking approach: https://github.com/RediSearch/RediSearch/pull/6342

### DIFF
--- a/src/redisearch_rs/sorting_vector/src/lib.rs
+++ b/src/redisearch_rs/sorting_vector/src/lib.rs
@@ -87,7 +87,7 @@ impl<T: RSValueTrait> RSSortingVector<T> {
                 len: self.values.len(),
             });
             #[cfg(not(debug_assertions))]
-            return Err(IndexOutOfBounds);
+            return Err(IndexOutOfBounds {});
         }
     }
 


src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    /// returns the approximate memory size of the RSValue instance.
+    fn memsize() -> usize {
+        std::mem::size_of::<Self>()
+    }
+}
+
+/// A [`RSSortingVector`] is a vector of a type T implementing [`RSValueTrait`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct RSSortingVector<T: RSValueTrait + Clone> {
```
JonasKruckenberg: `RSValueTrait` appears to be used exclusively with `+ Clone`. Just make `Clone` a super trait.

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+
+impl<T: RSValueTrait + Clone> RSSortingVector<T> {
+    /// Creates a new [`RSSortingVector`] with the given length.
+    pub fn new(len: usize) -> Self {
+        Self {
+            values: vec![T::create_null(); len].into_boxed_slice(),
+        }
+    }
+
+    /// Checks if the index is valid and decrements the reference count of previous RSValue instances.
```
JonasKruckenberg: This comment doesn't seem applicable anymore?

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    pub fn iter(&self) -> Iter<'_, T> {
+        self.values.iter()
+    }
+
+    pub fn iter_mut(&mut self) -> IterMut<'_, T> {
+        self.values.iter_mut()
+    }
+
+    /// Set a number (double) at the given index
+    pub fn put_num(&mut self, idx: usize, num: f64) {
```
JonasKruckenberg: Should we rename `put` to `insert` to mimic the Rust stdlib [`Vec::insert`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.insert)?

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// Creates a new [`RSSortingVector`] with the given length.
+    pub fn new(len: usize) -> Self {
+        Self {
+            values: vec![T::create_null(); len].into_boxed_slice(),
+        }
+    }
+
+    /// Checks if the index is valid and decrements the reference count of previous RSValue instances.
+    /// Returns `true` if the index is valid , `false` otherwise.
+    fn sanity_check(&self, idx: usize) -> bool {
```
LukeMathWalker:
```suggestion
    fn in_bounds(&self, idx: usize) -> bool {
```

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    fn as_num(&self) -> Option<f64>;
+
+    /// gets the type of the RSValue instance, it either null, string, number, or reference.
+    fn get_type(&self) -> ffi::RSValueType;
+
+    /// returns true if the RSValue is stored as a pointer on the heap (the C implementation)
+    fn is_ptr_type() -> bool;
+
+    /// returns the approximate memory size of the RSValue instance.
+    fn memsize() -> usize {
```
LukeMathWalker:
```suggestion
    fn mem_size() -> usize {
```

src/redisearch_rs/sorting_vector/src/tests.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    fn is_ptr_type() -> bool {
+        // Mock implementation, return false
+        false
+    }
+}
+
+#[test]
+fn test_rssortingvector_creation() {
```
LukeMathWalker:
```suggestion
fn test_creation() {
```

We're in the `sorting_vector` crate, no need to prefix every test name with it.

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    pub fn iter_mut(&mut self) -> IterMut<'_, T> {
+        self.values.iter_mut()
+    }
+
+    /// Set a number (double) at the given index
+    pub fn put_num(&mut self, idx: usize, num: f64) {
+        if !self.sanity_check(idx) {
+            return;
+        }
```
JonasKruckenberg: I don't think we should be keeping the current "silently ignore out of bounds accesses" behavior. At the very least we should an error log to these sanity checks, but ideally just omit them and let the slice indexing below panic.

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+fn walk_down_rsvalue_ref_chain<T: RSValueTrait>(origin: &T) -> &T {
+    let mut loop_var = origin;
+    while loop_var.get_ref().is_some() {
+        loop_var = loop_var.get_ref().unwrap();
+    }
+    loop_var
+}
+
+#[cfg(test)]
+mod tests;
```
LukeMathWalker: Tests don't use private APIs, so let's move them to the `tests` folder.

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+        }
+    }
+
+    /// Checks if the index is valid and decrements the reference count of previous RSValue instances.
+    /// Returns `true` if the index is valid , `false` otherwise.
+    fn sanity_check(&self, idx: usize) -> bool {
+        idx < self.values.len()
+    }
+
+    pub fn iter(&self) -> Iter<'_, T> {
```
JonasKruckenberg, line 95: doc comment needed

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+        self.values.iter()
+    }
+
+    pub fn iter_mut(&mut self) -> IterMut<'_, T> {
```
JonasKruckenberg, line 100: doc comment needed

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /// Set a value at the given index
+    pub fn put_val(&mut self, idx: usize, value: T) {
+        if !self.sanity_check(idx) {
+            return;
+        }
+
+        self.values[idx] = value;
+    }
```
LukeMathWalker: The strategy we're using to handle out-of-bound indexes is not ideal: there is no way for the caller to determine if something has actually been inserted or not.

We should either:
- Use the idiomatic Rust approach, i.e. panic on index out of bounds. We would move the check for being in bounds to the calling code/FFI layer.
- Return a boolean set to `true` if something was inserted, `false` otherwise. Minimal disruption to calling code.

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    fn memsize() -> usize {
+        std::mem::size_of::<Self>()
+    }
+}
+
+/// A [`RSSortingVector`] is a vector of a type T implementing [`RSValueTrait`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct RSSortingVector<T: RSValueTrait + Clone> {
+    values: Box<[T]>,
+}
```
LukeMathWalker: What's the purpose of this type? It's not clear from the docs here.

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    pub fn put_val(&mut self, idx: usize, value: T) {
+        if !self.sanity_check(idx) {
+            return;
+        }
+
+        self.values[idx] = value;
+    }
+
+    /// Set a reference to the value at the given index
+    pub fn put_val_as_ref(&mut self, idx: usize, value: T) {
```
JonasKruckenberg: I don't think the Rust signature makes sense here. This method could never be called with Rust types correctly (you would always need a `.clone()`) when calling. Is there an example of C code using this?

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+        if !self.sanity_check(idx) {
+            return;
+        }
+
+        self.values[idx] = T::create_ref(value);
+    }
+
+    /// Set a null value at the given index
+    pub fn put_null(&mut self, idx: usize) {
```
JonasKruckenberg: This doesn't appear in the C code, is this necessary?

src/redisearch_rs/sorting_vector/src/lib.rs
JonasKruckenberg: I'd put module declarations (even tests) at the top

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: It is used in `SortingVector_RdbLoad` therefore added it to Rust, too.

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: Good idea

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: Two examples, first from documents.c - I think the idea is to generate a `RSValue` and move it directly, the SortingVector takes ownership. Should work, or do I miss something?

```c
if (FieldSpec_IsSortable(fs)) {
    if (field->unionType != FLD_VAR_T_ARRAY) {
      bool unf = (fs->options & FieldSpec_UNF) != 0;
      RSSortingVector_PutStr(aCtx->sv, fs->sortIdx, c, unf);
    } else if (field->multisv) {
      RSSortingVector_PutRSVal(aCtx->sv, fs->sortIdx, field->multisv);
      field->multisv = NULL;
    }
  }
```

and I need it in FFI to add the intelligence they have for normalization there.

```rust
if is_normalized {
        // Safety: Caller must ensure 2 -> str is a valid C string
        let len = unsafe { libc::strlen(str) };

        // Safety: Caller must ensure 2 -> str is a valid C string
        let dupl = unsafe { ffi::rust_strdup(str) };

        // Safety: RS_StringVal is filled with a duplicated string, which is a valid C string.
        let val = unsafe { ffi::RS_StringVal(dupl, len as u32) };

        vec.put_val(idx, RSValueFFI(val));
    } else {
        // Safety: Caller must ensure 2 -> str is a valid C string
        let normalized_str = unsafe { ffi::normalizeStr(str) };

        // Safety: normalized_str is a valid C string, and we can safely get its length.
        let len = unsafe { libc::strlen(normalized_str) };

        // Safety: normalized_str is a valid C string, and we can safely use it to create an RSValue.
        let val = unsafe { ffi::RS_StringVal(normalized_str, len as u32) };
        vec.put_val(idx, RSValueFFI(val));
    }
```

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/728c1c992a0e1b3731a3639fd76b6f7142f9e352

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/a38142a0e844484d1e719920e5e7434842174385

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/a38142a0e844484d1e719920e5e7434842174385

src/redisearch_rs/sorting_vector/src/tests.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/ff6ffe92abf6b743b628817614a426f5ff6f2ba9

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/ff6ffe92abf6b743b628817614a426f5ff6f2ba9

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/a38142a0e844484d1e719920e5e7434842174385

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/728c1c992a0e1b3731a3639fd76b6f7142f9e352

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/a38142a0e844484d1e719920e5e7434842174385

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/a38142a0e844484d1e719920e5e7434842174385

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/0a48072569417fdcb87a40469a0202df66353376

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/6575df3f10ec9038c693dda1307e5dc6f1b21eec

src/redisearch_rs/sorting_vector/src/lib.rs
LukeMathWalker: Looks good, just a naming remark: let's go for `try_insert_*` rather than `insert_*`, since we return an error for out-of-bounds accesses.

src/redisearch_rs/sorting_vector/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "sorting_vector"
+version.workspace = true
+edition.workspace = true
+license-file.workspace = true
+publish.workspace = true
+
+[dependencies]
+pin-project.workspace = true
+libc = { workspace = true, features = ["extra_traits"] }
```
LukeMathWalker: Is this used anywhere?

src/redisearch_rs/sorting_vector/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "sorting_vector"
+version.workspace = true
+edition.workspace = true
+license-file.workspace = true
+publish.workspace = true
+
+[dependencies]
+pin-project.workspace = true
```
LukeMathWalker: Is this used anywhere?

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+pub enum Error {
+    /// The index is out of bounds for the sorting vector.
+    OutOfBounds,
+
+    /// A byte string was given as an argument that is not valid UTF-8.
+    StringUtfError,
+
+    /// A string was given as an argument that contains a null byte, which is not allowed.
+    StringNulError,
+}
```
LukeMathWalker: Let's avoid the "big enum black hole" error pattern.
Let's use specific error types for each operation kind:
- `IndexOutOfBounds` as the error type for insert operations. It'd be ideal to capture the index we tried to insert and the length of the collection, although we probably want to do that just when `debug_assertions` are enabled to avoid bloating the `Result` size.
   ```rust
   pub struct IndexOutOfBounds {
       #[cfg(debug_assertions)]
       index: usize,
       #[cfg(debug_assertions)]
       length: usize
   }
   ```
- `InsertNormalizedStringError` with the three cases:
   ```rust
   pub enum InsertNormalizedStringError {
        IndexOutOfBounds(IndexOutOfBounds),  
        Utf8Error(Utf8Error),
        NulError(NulError),
    }
    ```

All error types must implement the `Error` trait, forwarding the appropriate sources.

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+
+/// This is the Rust implementation of [`RSSortingVector`] which acts as a cache for sortable fields in a document.
+///
+/// A [`RSSortingVector`] is a boxed slice of a type T implementing [`RSValueTrait`]. That means it has a constant length.
```
LukeMathWalker:
```suggestion
/// [`RSSortingVector`] acts as a cache for sortable fields in a document.
///
/// It has a constant length, determined upfront on creation. It can't be resized.
```

Document as if the C code doesn't exist.

LukeMathWalker: Open questions that come to mind reading the docs and the comments:
- Do we expect a single vector instance to contain values of different kinds?
- Is the vector itself sorted in some way?
We should clarify them.

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    /// Set a number (double) at the given index
+    pub fn insert_num(&mut self, idx: usize, num: f64) -> Result<(), Error> {
+        self.in_bounds(idx)?;
+        self.values[idx] = T::create_num(num);
+        Ok(())
+    }
+
+    /// Set a string at the given index
+    pub fn insert_string(&mut self, idx: usize, str: &str) -> Result<(), Error> {
```
LukeMathWalker: Is this used when the caller guarantees that the string has already been normalized? Or do we have a legit usecase for non-normalized strings?

DarthB: Yes. There is an option structure that indicates if the string is already normalized. That's how that has been handled in C, they used this `unf` flag. I should prove the assumption that sometimes Redis queries return normalized strings and sometimes not, maybe we can rid of this.

```rust
if (str && FieldSpec_IsSortable(fs)) {
    if (field->unionType != FLD_VAR_T_ARRAY) {
      bool unf = (fs->options & FieldSpec_UNF) != 0;
      //...
```

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: using `thiserror` is fine or do you prefer another approach?

LukeMathWalker: It's fine 👍🏻

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/a0ec3028e6748e4864e7bb9719ed680ee924e932

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+        self.values.iter()
+    }
+
+    /// Returns a mutable iterator over the values in the sorting vector.
+    pub fn iter_mut(&mut self) -> IterMut<'_, T> {
+        self.values.iter_mut()
+    }
+
+    /// Set a number (double) at the given index
+    pub fn try_insert_num(&mut self, idx: usize, num: f64) -> Result<(), InsertSortingVectorError> {
```
LukeMathWalker: All operations apart from `try_insert_string_and_normalize` should use `IndexOutOfBounds` as their error type, since that's the only failure mode.

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+pub struct IndexOutOfBounds {
+    #[cfg(debug_assertions)]
+    pub index: usize,
+    #[cfg(debug_assertions)]
+    pub len: usize,
+}
+
+/// Errors that can be returned by [`RSSortingVector`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum InsertSortingVectorError {
```
LukeMathWalker:
```suggestion
pub enum InsertAndNormalizeError {
```

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/757cf1f22e134bba1036781259a2fe983a2f8d5a

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/757cf1f22e134bba1036781259a2fe983a2f8d5a

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+pub enum InsertAndNormalizeError {
+    #[error("{0}.")]
+    OutOfBounds(#[from] IndexOutOfBounds),
+
+    #[error("String argument is not valid UTF-8: {0}")]
+    Utf8Error(#[from] Utf8Error),
+
+    #[error("String argument contains a null byte: {0}")]
+    NulError(#[from] NulError),
+}
```
LukeMathWalker:
```suggestion
/// Error that can be returned by [`RSSortingVector::try_insert_string_and_normalize`].
#[derive(Error, Debug, Clone, PartialEq, Eq)]
pub enum InsertAndNormalizeError {
    #[error(transparent)]
    OutOfBounds(#[from] IndexOutOfBounds),

    #[error(transparent)]
    Utf8Error(#[from] Utf8Error),

    #[error(transparent)]
    NulError(#[from] NulError),
}
```

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/fc13588a0dc9d32aafbd477e21742ee76b54e2f8

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/2ff10f1b6dc0b40adc18b27702ec84cd55c9cc69

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    /// Set a string at the given index, the string is normalized before being set.
+    pub fn try_insert_string<S: AsRef<str>>(
+        &mut self,
+        _idx: usize,
+        _str: S,
+    ) -> Result<(), IndexOutOfBounds> {
+        unimplemented!(
+            "We cannot yet use `NormalizedString` here, as the RSValue string allocations still happen\
+                in the C layer and the workaround for linking C code for allocations is not ready."
```
JonasKruckenberg:
```suggestion
        unimplemented!(
            "We cannot yet allocate RSValues in Rust. See MOD-10347 for details."
```

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+        &mut self,
+        _idx: usize,
+        _str: S,
+    ) -> Result<(), IndexOutOfBounds> {
+        self.in_bounds(_idx)?;
+        let casemapper = CaseMapper::new();
+        let normalized = casemapper.fold_string(_str.as_ref()).into_owned();
+        self.values[_idx] = T::create_string(normalized);
+        Ok(())
+    }
```
LukeMathWalker: Can we avoid mixing test helpers with production code?

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/386aec0ba4024161a89e770f8b67835b8aa63657

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: Due to `RSValue` and its C-side string allocations, this is a temporary fix. Nonetheless, I eliminated this second method.

https://github.com/RediSearch/RediSearch/pull/6311/commits/386aec0ba4024161a89e770f8b67835b8aa63657

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+        for idx in 0..self.values.len() {
+            if self.values[idx].is_null() {
+                continue;
+            }
+            if T::is_ptr_type() {
+                // count the size of the struct if not null, like in C
+                sz += T::mem_size();
+            }
+
+            // the original behavior would by-pass references in the middle of the chain
```
raz-mon, line 189: When will the comments about the previous implementations be removed?

src/redisearch_rs/sorting_vector/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+            self.values.len() * T::mem_size()
+        };
+
+        for idx in 0..self.values.len() {
+            if self.values[idx].is_null() {
+                continue;
+            }
+            if T::is_ptr_type() {
+                // count the size of the struct if not null, like in C
+                sz += T::mem_size();
```
raz-mon, line 186: Isn't this double-counting for the second case above?

raz-mon (commented):
LGTM basically
A couple of questions:
* I guess it would be good to have as full an API as possible at this point, such that we can test the most and add less additions later on when the RSValue is supported. That would basically require an extended mock for the RSValue - do you think that's worthwhile, or prefer to have bigger additions to the API and tests later on?
* Some comments relate to the previous code. While that is comfortable for the review and the transition, we should at some point remove them. Probably the best time would be prior to merging, so that we don't forget it like that.

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/9555ae6b956311c45ad3a40e83a84765fbb78f6e

src/redisearch_rs/sorting_vector/src/lib.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6311/commits/9555ae6b956311c45ad3a40e83a84765fbb78f6e

DarthB:
> * I guess it would be good to have as full an API as possible at this point, such that we can test the most and add less additions later on when the RSValue is supported. That would basically require an extended mock for the RSValue - do you think that's worthwhile, or prefer to have bigger additions to the API and tests later on?

I checked the `RSValue` API and propose the following:
1. add `increment` and `decrement` to the trait in this PR (we will need it soon for `RLookupRow`)
2. Have a follow-up PR for the methods that are needed in `RLookup`

We plan to port the `RSValue` type in MOD-10347 and attempt to minimize the workload associated with that trait workaround.

> * Some comments relate to the previous code. While that is comfortable for the review and the transition, we should at some point remove them. Probably the best time would be prior to merging, so we don't forget it.

I would like to address this size calculation when we port `RSValue` - MOD-10347.

raz-mon (approved):
LGTM
Note the rdbLoad function as well, that is not implemented here

PR #6317, opened 2025-06-15
MOD-10031: RPHybridMerger

Itzikvaknin:


src/result_processor.c
```diff
@@ -1426,3 +1427,216 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+// Structure to hold both SearchResult and scores from both upstreams
+typedef struct {
+  SearchResult *searchResult; // The actual search result
+  double scores[2];           // scores[0] = upstream1, scores[1] = upstream2
+  bool hasScores[2];          // hasScores[0] = has upstream1, hasScores[1] = has upstream2
+} HybridSearchResult;
+
+// Constructor for HybridSearchResult
+static HybridSearchResult* HybridSearchResult_New(SearchResult *searchResult) {
+  HybridSearchResult* hybridResult = rm_calloc(1, sizeof(HybridSearchResult));
```
kei-nan, line 1606: rm_calloc memsets the allocated memory with 0, not sure you need the extra initialization below.

src/result_processor.c
```diff
@@ -1426,3 +1427,216 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+
+ /*******************************************************************************************************************
+  *  Hybrid Merger Result Processor
+  *
+  * This result processor merges results from two upstream processors using a hybrid scoring function.
+  * It takes results from both upstreams and applies the provided function to combine their scores.
+  *******************************************************************************************************************/
+ typedef struct {
+   ResultProcessor base;
+   double (*hybridScoring_function)(double, double, bool, bool);
```
kei-nan: Consider using a typedef for the callback type.
also some documentation on the callback parameters would be nide.

src/result_processor.c
```diff
@@ -1426,3 +1427,216 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+  *  Hybrid Merger Dictionary Types
+  *
+  * Single dictionary for storing both search results and scores:
+  * keyPtr -> HybridSearchResult mapping (contains both SearchResult and scores)
+  *******************************************************************************************************************/
+
+// Structure to hold both SearchResult and scores from both upstreams
+typedef struct {
+  SearchResult *searchResult; // The actual search result
+  double scores[2];           // scores[0] = upstream1, scores[1] = upstream2
```
kei-nan: maybe the magic number 2 should be a macro so it would be easier to find where we rely on the number?

src/result_processor.c
```diff
@@ -1426,3 +1427,216 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+   hybridResult->scores[upstreamIndex] = r->score;
+   hybridResult->hasScores[upstreamIndex] = true;
+ }
+
+ /* Helper function to consume results from a single upstream */
+ static int ConsumeFromUpstream(SearchResult *tempResult, dict *hybridResults, size_t window, ResultProcessor *upstream, int upstreamIndex) {
+   size_t consumed = 0;
+   int rc = RS_RESULT_OK;
+
+   while (consumed < window && rc == RS_RESULT_OK) {
```
kei-nan: I think this loop structure can be simplified a bit.
move function call to loop condition

src/result_processor.c
```diff
@@ -1426,3 +1427,216 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+        // Handle timeout case: switch to yield phase and return.
+        self->timedOut = true;
+        self->iterator = dictGetIterator(self->hybridResults);
+        rp->Next = RPHybridMerger_Yield;
+        return rp->Next(rp, r);
+      }
+    }
+  }
+   // Initialize iterator for yield phase
+   self->iterator = dictGetIterator(self->hybridResults);
```
kei-nan: what should happen if rc == RS_RESULT_TIMEDOUT but policy is strict?
right now you are still returning results and swallowing the timeout rc code.

src/result_processor.c
```diff
@@ -1426,3 +1427,216 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+     // No more results to yield - return appropriate status
+     int ret = self->timedOut ? RS_RESULT_TIMEDOUT : RS_RESULT_EOF;
+     return ret;
+   }
+
+   HybridSearchResult *hybridResult = dictGetVal(entry);
+   RS_ASSERT(hybridResult && hybridResult->searchResult);
+   RS_ASSERT(hybridResult->hasScores[0] || hybridResult->hasScores[1]);
+
+   double hybridScore = self->hybridScoring_function(hybridResult->scores[0], hybridResult->scores[1],
```
kei-nan: how is this supposed to work for RRF?
It needs access to all of the results in order to provide a score, no?

Itzikvaknin: They can be ranked either based on the sorter's output or when consumed from the upstream sorter.
I'll leave that decision to the pipeline builder and the integration logic.

kei-nan: You are scrambling the results ordering by using a hash table.
We would need to sort them possibly a second time.
Or will the ranker will give them a score which is their rank?
Please clarify.

Itzikvaknin: The order in which `HybridMerger` receives results from its sorter upstream determines the rank — the first `Next()` call gives rank 1, the second gives rank 2, and so on.

We can assign the rank either inside the scorer’s yield function or right after calling `upstream->Next()` in `HybridMerger`.

Both of these steps happen before inserting into the hashtable, so it's not a problem. It's just a matter of where we choose to do it.

kei-nan: So this is something that is still missing from the PR?
So the scores that get passed to the callback can just be integers, no?
If they just mean the rank...

src/result_processor.c
Itzikvaknin: As done in `RPSort` and `RPMaxScoreNormalizer` - Stop Accumulating results and return `RS_RESULT_TIMEDOUT` in the first `Next()` call  
Changed it to behave that way

src/result_processor.c
Itzikvaknin: Integer specifically for RRF, for hybrid it can be double
And yes, handling ranks is not in this PR

tests/cpptests/test_cpp_resultprocessor.cpp
```diff
@@ -11,6 +11,8 @@
 #include "result_processor.h"
 #include "query.h"
 #include "gtest/gtest.h"
+#include "config.h"
+#include "module.h"
```
kei-nan: do we really need to include these files?
seems like quite a bit of an overhead.
not critical but maybe consider minimizing what we include.

tests/cpptests/test_cpp_resultprocessor.cpp
```diff
@@ -88,3 +104,981 @@ TEST_F(ResultProcessorTest, testProcessorChain) {
...
+    int counter = 0;
+    static int NextFn(ResultProcessor *rp, SearchResult *res) {
+      MockUpstream1 *p = static_cast<MockUpstream1 *>(rp);
+      if (p->counter >= 3) return RS_RESULT_EOF;
+
+      res->docId = ++p->counter;
+      res->score = 2.0;
+
+      // Set up document metadata with keys
+      static RSDocumentMetadata dmd[3] = {0};
```
kei-nan: why does it need to be static variable.

src/result_processor.c
```diff
@@ -1426,3 +1430,217 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+   ResultProcessor *upstreams[HYBRID_MERGER_UPSTREAMS];
+   size_t window;
+   dict *hybridResults;  // keyPtr -> HybridSearchResult mapping
+   dictIterator *iterator; // Iterator for yielding results
+   SearchResult *pooledResult; // Reusable result for accumulation
+   bool timedOut; // Flag to track if we timed out during accumulation
+ } RPHybridMerger;
+
+ /* Helper function to store a result from an upstream into the hybrid merger's dictionary */
+ static void StoreUpstreamResult(SearchResult *r, dict *hybridResults, int upstreamIndex) {
```
kei-nan: Can we get coverage for this function?

src/result_processor.c
```diff
@@ -1426,3 +1430,217 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+   hybridResult->hasScores[upstreamIndex] = true;
+ }
+
+ /* Helper function to consume results from a single upstream */
+ static int ConsumeFromUpstream(SearchResult *tempResult, dict *hybridResults, size_t maxResults, ResultProcessor *upstream, int upstreamIndex) {
+   size_t consumed = 0;
+   int rc = RS_RESULT_OK;
+   SearchResult_Clear(tempResult);
+   while (consumed < maxResults && (rc = upstream->Next(upstream, tempResult)) == RS_RESULT_OK) {
+       StoreUpstreamResult(tempResult, hybridResults, upstreamIndex);
```
kei-nan: coverage seems to be missing for this function.

kei-nan: Also, shouldn't we give the ranking here?

src/result_processor.c
```diff
@@ -15,6 +15,9 @@
 #include "rmutil/rm_assert.h"
 #include "util/timeout.h"
 #include "util/arr.h"
+
+#define HYBRID_MERGER_UPSTREAMS 2
```
raz-mon: Why have this as a hardcoded macro and not just have an array of upstreams for the merger to handle?

src/result_processor.c
```diff
@@ -1426,3 +1430,217 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
   ret->scoreKey = rlk;
   return &ret->base;
 }
+
+ /*******************************************************************************************************************
+  *  Hybrid Merger Dictionary Types
+  *
+  * Single dictionary for storing both search results and scores:
+  * keyPtr -> HybridSearchResult mapping (contains both SearchResult and scores)
```
raz-mon: Consider having the mapping from `docId --> Result`
Can optimize memory of the dictionary

src/result_processor.c
```diff
@@ -1426,3 +1430,217 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+  SearchResult *searchResult; // The actual search result
+  double scores[HYBRID_MERGER_UPSTREAMS];           // scores[0] = upstream1, scores[1] = upstream2
+  bool hasScores[HYBRID_MERGER_UPSTREAMS];          // hasScores[0] = has upstream1, hasScores[1] = has upstream2
+} HybridSearchResult;
+
+// Constructor for HybridSearchResult
+static HybridSearchResult* HybridSearchResult_New(SearchResult *searchResult) {
+  HybridSearchResult* hybridResult = rm_calloc(1, sizeof(HybridSearchResult));
+  hybridResult->searchResult = searchResult;
+  return hybridResult;
```
raz-mon, line 1611: Is coverage off or is this really not covered?
If the latter, add coverage or remove if not needed

src/result_processor.c
```diff
@@ -1426,3 +1430,217 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+}
+
+// Destructor for HybridSearchResult
+static void HybridSearchResult_Free(HybridSearchResult* hybridResult) {
+  if (hybridResult) {
+    srDtor(hybridResult->searchResult);
+    rm_free(hybridResult);
```
raz-mon, line 1620: Coverage?

src/result_processor.c
```diff
@@ -1426,3 +1430,217 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+ /* Helper function to store a result from an upstream into the hybrid merger's dictionary */
+ static void StoreUpstreamResult(SearchResult *r, dict *hybridResults, int upstreamIndex) {
+   const char *keyPtr = r->dmd->keyPtr;
+
+   // Check if we've seen this document before
+   HybridSearchResult *hybridResult = dictFetchValue(hybridResults, keyPtr);
+
+   if (!hybridResult) {
+     // First time seeing this document - create new hybrid result with copied SearchResult
+     SearchResult *resultCopy = SearchResult_Copy(r);
```
raz-mon: Why copy instead of using this result?

src/result_processor.c
```diff
@@ -1426,3 +1430,217 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+     hybridResult = HybridSearchResult_New(resultCopy);
+     dictAdd(hybridResults, (void*)keyPtr, hybridResult);
```
raz-mon, line 1666: Didn't get to the RRF scoring part yet, but basically the mapping value should contain the rank for the scoring.
We know the rank since we get the results sorted - this is an assumption of this result-processor

src/result_processor.c
```diff
@@ -1426,3 +1430,217 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+     return ret;
+   }
+
+   HybridSearchResult *hybridResult = dictGetVal(entry);
+   RS_ASSERT(hybridResult && hybridResult->searchResult);
+   RS_ASSERT(hybridResult->hasScores[0] || hybridResult->hasScores[1]);
+
+   double hybridScore = self->hybridScoring_function(hybridResult->scores, hybridResult->hasScores, HYBRID_MERGER_UPSTREAMS);
+
+   SearchResult_Override(r, hybridResult->searchResult);
```
raz-mon, line 1714: We should have the results sorted as well
We can add a `sorter` RP as the downstream of this result processor so we won't overload this RP..

src/result_processor.c
```diff
@@ -1426,3 +1430,217 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+
+ /* Accumulation phase - consume window results from both upstreams */
+ static int RPHybridMerger_Accum(ResultProcessor *rp, SearchResult *r) {
+   RPHybridMerger *self = (RPHybridMerger *)rp;
+
+  // The array keeps unconsumed indices at the beginning, and consumed indices are moved to the end.
+  int upstreamIndices[HYBRID_MERGER_UPSTREAMS];
+  for (int i = 0; i < HYBRID_MERGER_UPSTREAMS; i++) {
+    upstreamIndices[i] = i;
+  }
```
raz-mon: This seems unnecessary, the mapping is trivial and equivalent to regular enumeration

src/result_processor.h
```diff
@@ -311,6 +314,26 @@ void PipelineAddCrash(struct AREQ *r);
...
+
+ // Parameters:
+ //   scores: Array of score values from each upstream
+ //   hasScores: Array of boolean flags indicating if each score is available
+ //   numUpstreams: Number of upstream processors
+ // Returns: Combined score calculated from the array of optional input scores
+ typedef double (*HybridScoringFunction)(double *scores, bool *hasScores, size_t numUpstreams);
+
+
+ ResultProcessor *RPHybridMerger_New(HybridScoringFunction hybridScoring_function,
```
raz-mon: Intuitively I would implement the RRF-hybrid-scoring-function in this PR as well so this RP will be usable after this PR.
I assume you didn't do that in order not to blow up the PR.. Legit consideration but it will yield some tail for the PR

tests/cpptests/test_cpp_resultprocessor.cpp
```diff
@@ -88,3 +104,981 @@ TEST_F(ResultProcessorTest, testProcessorChain) {
...
+  } upstream2;
+
+
+
+  // Create hybrid merger with window size 4
+  ResultProcessor *upstreams[] = {&upstream1, &upstream2};
+  ResultProcessor *hybridMerger = RPHybridMerger_New(
+    hybridScoringFunction,
+    upstreams,
+    2,  // numUpstreams
```
raz-mon: The number of upstreams is hardcoded in a macro, and also an argument to the merger constructor - let's have it only an argument

tests/cpptests/test_cpp_resultprocessor.cpp
```diff
@@ -88,3 +104,981 @@ TEST_F(ResultProcessorTest, testProcessorChain) {
...
+  }
+
+  // Should have processed 3 unique documents
+  ASSERT_EQ(3, count);
+
+  SearchResult_Destroy(&r);
+  QITR_FreeChain(&qitr);
+}
+
+TEST_F(ResultProcessorTest, testHybridMergerDifferentDocuments) {
```
raz-mon: Please add documentation as to what each test tests

raz-mon (commented):
Nice work!
See some comments.
Mainly:
- Coverage.
- Sorting of the results. Perhaps not here but we should address it in a separate RP if so, and that needs to be reflected in the breakdown.
- Implementation of the RRF scorer - same as the above.

Regarding the PR comment, let's refine it to be a bit more accurate. e.g., the hybrid-merger does not do anything simultaneously. It would be best to have a very detailed PR comment depicting the new RP's behavior to detail, and even some of the implementation details.

src/result_processor.c
Itzikvaknin: `HybridMerger` can be in the coordinator, can we rely on `docId` uniqueness ?

raz-mon:
Another thing that I think the merger needs to do, is to merge the upstream's pipelines in a way.
This means RLookUps, QueryIterators, and maybe more - some research is needed to understand exactly how this merging should be done, but for sure we need something of that nature here. I won't be trivial, let's discuss it

src/result_processor.c
Itzikvaknin: The main reason was the difference between dynamic and static upstream array sizes (and a few other minor cases).
I also wanted to avoid over-generalizing the implementation, which could make the purpose of this PR less clear.

src/result_processor.c
Itzikvaknin: It changes during the loop to avoid unnecessary work

src/result_processor.c
Itzikvaknin: Copy from tempResult to hybridResults dict

`
   SearchResult_Clear(tempResult);
   while (consumed < maxResults && (rc = upstream->Next(upstream, tempResult)) == RS_RESULT_OK) {
       StoreUpstreamResult(tempResult, hybridResults, upstreamIndex);
       ...
   }
`

src/result_processor.c
Itzikvaknin: I agree

tests/cpptests/test_cpp_resultprocessor.cpp
Itzikvaknin: Removed `module.h` since `RSDummyContext` can be replaced by `NULL`
Config is required for `RSTimeoutPolicy`

tests/cpptests/test_cpp_resultprocessor.cpp
Itzikvaknin: So it keeps its value between different calls to NextFn

src/result_processor.c
raz-mon: 100% correct

src/result_processor.c
Itzikvaknin: Added the rank to the `HybridSearchResult`

src/hybrid_scoring.h
```diff
@@ -0,0 +1,15 @@
...
+ } HybridScoringType;
+
+ typedef double (*HybridScoringFunction)(ScoringFunctionArgs *ctx, const double *scores, const bool *hasScores, const size_t numUpstreams);
+
+ /* Get scoring function based on scoring type */
+HybridScoringFunction GetScoringFunction(HybridScoringType scoringType);
+
+double RRFScore(ScoringFunctionArgs *ctx,const double *ranks, const bool *has_rank, const size_t num_sources);
+
+double HybridLinearScore(ScoringFunctionArgs *ctx, const double *scores, const bool *has_score, const size_t num_sources);
```
raz-mon: Let's have the same name convention, e.g., `HybridRRFScore`

src/hybrid_scoring.c
```diff
@@ -0,0 +1,77 @@
...
+ * Compute Reciprocal Rank Fusion (RRF) score for a document.
+ *
+ * RRF is used to combine multiple ranked lists into a single score.
+ * Each system contributes to the score as 1 / (k + rank), where lower
+ * ranks (i.e., higher relevance) contribute more.
+ *
+ * Formula:
+ *     RRF_score = Σ (1 / (k + rank_i)) for all i where has_rank[i] is true
+ *
+ * - ranks[i] is assumed to be 1-based (i.e., 1 is the best rank).
```
raz-mon, line 26: Why not 0-based?

src/hybrid_scoring.c
```diff
@@ -0,0 +1,77 @@
...
+ * - A typical value for k is 60, which dampens the effect of lower rankings.
+ *
+ * @param ranks       Array of rank values (e.g., ranks[i] is the rank assigned by system i).
+ * @param has_rank    Array of booleans indicating whether a rank was assigned by system i.
+ * @param num_sources Number of rank sources (i.e., size of ranks[] and has_rank[]).
+ * @param k           Damping constant, Higher values reduce the influence of low-ranked documents.
+ *
+ * @return RRF score as a double. Higher scores indicate higher relevance.
+ */
+double RRFScore(ScoringFunctionArgs *ctx,const double *ranks, const bool *has_rank, const size_t num_sources) {
```
raz-mon:
```suggestion
double RRFScore(ScoringFunctionArgs *ctx, const double *ranks, const bool *has_rank, const size_t num_sources) {
```

src/hybrid_scoring.c
```diff
@@ -0,0 +1,77 @@
...
+ /* Get scoring function based on scoring type */
+ HybridScoringFunction GetScoringFunction(HybridScoringType scoringType) {
+   switch (scoringType) {
+     case HYBRID_SCORING_LINEAR:
+       // Linear scoring not implemented yet
+       return HybridLinearScore;
+     case HYBRID_SCORING_RRF:
+       return RRFScore;
+     default:
+       return HybridLinearScore; // Default fallback to hybrid linear scoring
```
raz-mon: IIRC, the default should be RRF
But anyways shouldn't there be an assertion here, since this shouldn't happen?..
I would expect that and an error setting rather than this fallback

src/redisearch.h
```diff
@@ -397,6 +397,14 @@ typedef struct {
...
   uint64_t tanhFactor;
+
+  /* Hybrid scoring function parameters
+   * For RRF scoring: */
+  size_t RRF_k;
+  /* For linear hybrid scoring: */
+  double *linearWeights;
+  size_t numScores;
+
 } ScoringFunctionArgs;
```
raz-mon:
```suggestion
  /* Hybrid scoring function parameters For RRF scoring: */
  size_t RRF_k;

  /* For linear hybrid scoring: */
  double *linearWeights;
  size_t numScores;
} ScoringFunctionArgs;
```

src/hybrid_scoring.c
raz-mon: Also, if you have `ranks` to be an array using the `arr.h` API, you won't need the `num_sources` argument

src/result_processor.c
```diff
@@ -1426,3 +1428,245 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
   ret->scoreKey = rlk;
   return &ret->base;
 }
+
+ /*******************************************************************************************************************
+  *  Hybrid Merger Dictionary Types
```
raz-mon:
```sugestion
  *  Hybrid Merger Dictionary Type
```

Also, move to above the new dictionary type

src/result_processor.c
```diff
@@ -1426,3 +1428,245 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+  * keyPtr -> HybridSearchResult mapping (contains both SearchResult and scores)
+  *******************************************************************************************************************/
+
+// Structure to hold both SearchResult and scores from multiple upstreams
+typedef struct {
+  SearchResult *searchResult; // The actual search result
+  double *scores;             // Dynamic array of scores from upstreams
+  bool *hasScores;           // Dynamic array of flags indicating if each score is available
+  size_t numSources;       // Number of upstreams
+} HybridSearchResult;
```
raz-mon: Also this I think can be moved to the HybridMerger section

raz-mon: Also, indentation of comments

src/result_processor.c
```diff
@@ -1426,3 +1428,245 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+ } RPHybridMerger;
+
+ /* Helper function to store a result from an upstream into the hybrid merger's dictionary */
+ static void StoreUpstreamResult(SearchResult *r, dict *hybridResults, int upstreamIndex, size_t numUpstreams, double score) {
+   const char *keyPtr = r->dmd->keyPtr;
+
+   // Check if we've seen this document before
+   HybridSearchResult *hybridResult = dictFetchValue(hybridResults, keyPtr);
+
+   if (!hybridResult) {
```
raz-mon, line 1663: We will (later) need to merge between the results, not just discard the second

src/result_processor.c
```diff
@@ -1426,3 +1428,245 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+  // Busy wait loop - continuously try to consume from upstreams until all are consumed
+  while (numUnconsumed > 0) {
+    for (int i = 0; i < numUnconsumed; i++) {
+      int upstreamIndex = upstreamIndices[i];
+      int rc = ConsumeFromUpstream(self, self->window, self->upstreams[upstreamIndex], upstreamIndex);
+
+      if (rc == RS_RESULT_DEPLETING) {
+        // Upstream is still active but not ready to provide results. Skip to the next.
+        continue;
+      } else if (rc == RS_RESULT_OK || rc == RS_RESULT_EOF) {
```
raz-mon: Let's add a comment why ok also means done (`window` results reached - can assert this btw)

src/result_processor.c
```diff
@@ -1426,3 +1428,245 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+
+ /* Accumulation phase - consume window results from all upstreams */
+ static int RPHybridMerger_Accum(ResultProcessor *rp, SearchResult *r) {
+   RPHybridMerger *self = (RPHybridMerger *)rp;
+
+  // The array keeps unconsumed indices at the beginning, and consumed indices are moved to the end.
+  int *upstreamIndices = rm_malloc(self->numUpstreams * sizeof(int));
+  for (size_t i = 0; i < self->numUpstreams; i++) {
+    upstreamIndices[i] = i;
+  }
```
raz-mon: I think let's get rid of this, and from all the index-swapping logic, and just remember we're done with upstream `i`.
WDYT?

src/result_processor.c
```diff
@@ -1426,3 +1428,245 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+        }
+        continue;
+      } else if (rc == RS_RESULT_TIMEDOUT){
+          if (rp->parent->timeoutPolicy == TimeoutPolicy_Return) {
+            // Handle timeout case: switch to yield phase and return.
+            self->timedOut = true;
+            self->iterator = dictGetIterator(self->hybridResults);
+            rp->Next = RPHybridMerger_Yield;
+            rm_free(upstreamIndices);
+            return rp->Next(rp, r);
```
raz-mon: Timeout means consumed as well
Also, when returning here, we leave "dangliing" results in the other upstream (potentially) - should be addressed as well.
If we want to discard them, let's add a comment for that here. I think it can be weird since RRF will have no meaning with one upstream..
Give it a think..

src/result_processor.c
```diff
@@ -1426,3 +1428,245 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+          } else { //strict
```
raz-mon:
```suggestion
          } else { // Strict
```

src/result_processor.c
```diff
@@ -1426,3 +1428,245 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+   RPHybridMerger *ret = rm_calloc(1, sizeof(*ret));
+
+   RS_ASSERT(numUpstreams > 0);
+   ret->numUpstreams = numUpstreams;
+
+   ret->scoringType = scoringType;
+   ret->scoringCtx = scoringCtx;
+
+   // Allocate and copy the array for upstreams
+   ret->upstreams = rm_malloc(numUpstreams * sizeof(ResultProcessor*));
```
raz-mon: If it already gets `upstreams`, let's use it

raz-mon (commented):
Looking good, see a few more comments

src/result_processor.c
Itzikvaknin: Removed it

src/result_processor.c
Itzikvaknin: I think its a good place for it, all the things between this and `dictType dictTypeHybridSearchResult` are type descriptors for the dict

raz-mon: Things like `HybridSearchResult` are more general and related to the RP more than the dict
Your call

src/hybrid_scoring.c
Itzikvaknin: Once there is more than one array and all are expected to have the same length, I prefer keeping it explicit

src/result_processor.c
Itzikvaknin: Reorganised

src/hybrid_scoring.c
```diff
@@ -0,0 +1,78 @@
...
+ * @param k           Damping constant, Higher values reduce the influence of low-ranked documents.
+ *
+ * @return RRF score as a double. Higher scores indicate higher relevance.
+ */
+double HybridRRFScore(ScoringFunctionArgs *ctx, const double *ranks, const bool *has_rank, const size_t num_sources) {
+
+    double score = 0.0;
+    for (size_t i = 0; i < num_sources; ++i) {
+        if (has_rank[i]) {
+            score += 1.0 / (ctx->RRF_k + ranks[i]);
```
kei-nan: maybe add an assertion that RRF_k is not zero at the start of the function.

src/result_processor.c
```diff
@@ -1426,3 +1428,230 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+            return rc;
+          }
+      }
+    }
+  }
+
+  // Free the consumed tracking array
+  rm_free(consumed);
+
+   // Initialize iterator for yield phase
```
kei-nan: indentation here seems off, either here of in rm_free above this line.

tests/cpptests/test_cpp_hybridmerger.cpp
```diff
@@ -0,0 +1,1748 @@
...
+ * Intersection: No intersection (different documents from each upstream)
+ * Emptiness: Both upstreams have documents
+ * Timeout: No timeout
+ * Expected behavior: Each document gets weighted score from only its contributing upstream (0.4*1.0=0.4 or 0.6*3.0=1.8)
+ */
+TEST_F(HybridMergerTest, testHybridMergerDifferentDocuments) {
+  QueryIterator qitr = {0};
+
+  // Mock upstream1: generates 3 docs with score 1.0
+  struct MockUpstream1 : public ResultProcessor {
```
kei-nan: why all the code duplication?
Can't you declare the mock stream type once with arguments in its ctor and reuse the class?

tests/cpptests/test_cpp_hybridmerger.cpp
```diff
@@ -0,0 +1,1748 @@
...
+      // Set up document metadata with keys
+      static RSDocumentMetadata dmd[3] = {0};
+      static const char* keys[] = {"doc1", "doc2", "doc3"};
+      dmd[p->counter - 1].keyPtr = (char*)keys[p->counter - 1];
+      res->dmd = &dmd[p->counter - 1];
+
+      return RS_RESULT_OK;
+    }
+    MockUpstream1() {
+      memset(this, 0, sizeof(*this));
```
kei-nan: in C++ you shouldn't memset the struct

src/hybrid_scoring.c
Itzikvaknin: Ranks start at 1, so the denominator is guaranteed to be non-zero

src/redisearch.h
```diff
@@ -397,6 +397,7 @@ typedef struct {
 
   /* Tanh factor (used only in the `BM25STD.TANH` scorer)*/
   uint64_t tanhFactor;
+
```
raz-mon: Remove

src/result_processor.c
```diff
@@ -1590,3 +1592,230 @@ ResultProcessor *RPDepleter_New(StrongRef sync_ref) {
...
+  size_t numSources;          // Number of upstream processors (array size)
+} HybridSearchResult;
+
+// Constructor for HybridSearchResult
+static HybridSearchResult* HybridSearchResult_New(SearchResult *searchResult, size_t numUpstreams) {
+  HybridSearchResult* hybridResult = rm_calloc(1, sizeof(HybridSearchResult));
+  hybridResult->searchResult = searchResult;
+  hybridResult->numSources = numUpstreams;
+  hybridResult->scores = rm_calloc(numUpstreams, sizeof(double));
+  hybridResult->hasScores = rm_calloc(numUpstreams, sizeof(bool));
```
raz-mon, line 1610: Since we know the length of this array in advance and each upstream already has its index, you can use remove `hasScores` and just use `scores` with some initial value that cannot be a valid score, e.g., `-1`.
`hasScores` is fine as well just not necessary, FYI

src/result_processor.c
```diff
@@ -1590,3 +1592,230 @@ ResultProcessor *RPDepleter_New(StrongRef sync_ref) {
...
+   size_t consumed = 0;
+   int rc = RS_RESULT_OK;
+   SearchResult *r = rm_calloc(1, sizeof(*r));
+   while (consumed < maxResults && (rc = upstream->Next(upstream, r)) == RS_RESULT_OK) {
+       double score = r->score;
+       consumed++;
+       if (self->hybridScoringCtx->scoringType == HYBRID_SCORING_RRF) {
+         score = consumed;
+       }
+       StoreUpstreamResult(r, self->hybridResults, upstreamIndex, self->numUpstreams, score);
```
raz-mon, line 1689: Not sure how much I love this, mainly since I believe we will soon want to have an option to return the score as well e.g., for profile or `WITHSCORES`, so overriding it may not be future-proof.
FYI at this point, but it's worth thinking if this will need to be changed once the above is needed and considering having the explicit rank.

src/result_processor.c
raz-mon: I think we should wait for the other upstreams at this point so that we don't fold the pipeline when a depleter is working in the background.
I.e., wait for all to be consumed before we return results.
If we see bad timeout behavior we can fix for more complex mechanisms like ref-counting the results of the depleter

tests/cpptests/test_cpp_hybridmerger.cpp
```diff
@@ -0,0 +1,996 @@
...
+ * Intersection: N/A (one upstream empty)
+ * Emptiness: First upstream empty, second upstream has documents
+ * Timeout: No timeout
+ * Expected behavior: Only documents from second upstream with weighted score (0.5*5.0=2.5)
+ */
+TEST_F(HybridMergerTest, testHybridMergerEmptyUpstream1) {
+  QueryIterator qitr = {0};
+
+  // Create upstreams: first empty, second with documents
+  MockUpstream upstream1(0, {}, {}); // Empty scores and docIds
```
raz-mon, line 281: Is EOF returned on the second `Next` call to `upstream1`?

raz-mon: What about half-empty, i.e., with half the results of `upstream2`?

tests/cpptests/test_cpp_hybridmerger.cpp
```diff
@@ -0,0 +1,996 @@
...
+ * Test that hybrid merger with 4 upstreams using linear scoring (full intersection)
+ *
+ * Scoring function: Hybrid linear
+ * Number of upstreams: 4
+ * Intersection: Full intersection (same documents from all upstreams)
+ * Emptiness: All upstreams have documents
+ * Timeout: No timeout
+ * Expected behavior: Each document gets combined score from all 4 upstreams (0.1*1.0 + 0.2*2.0 + 0.3*3.0 + 0.4*4.0 = 3.0)
+ */
+TEST_F(HybridMergerTest, testHybridMergerLinear4Upstreams) {
```
raz-mon: What value does this test have, given that we tested 3 upstreams above?

raz-mon (commented):
Test comments
Nice work @Itzikvaknin

src/result_processor.c
kei-nan: I think the bigger problem if we don't wait s that the thread will continue running while the query is being folded.
So not sure how ref-counting will help.

tests/cpptests/test_cpp_hybridmerger.cpp
Itzikvaknin: Is EOF returned on the second Next call to upstream1? returns EOF on the first `Next()` call
What about half-empty, i.e., with half the results of upstream2? Will add this test case

tests/cpptests/test_cpp_hybridmerger.cpp
Itzikvaknin: leftover.. removed

src/result_processor.c
Itzikvaknin: Lets leave it that way

src/result_processor.c
Itzikvaknin: For now, it fits the requirements. Let's keep an eye on it in case they change

src/result_processor.c
Itzikvaknin: I agree. Maybe we can check how does it behave currently if the client closes while the query is still running in the background

src/result_processor.c
```diff
@@ -1740,15 +1745,14 @@ dictType dictTypeHybridSearchResult = {
...
             self->timedOut = true;
-            self->iterator = dictGetIterator(self->hybridResults);
-            rp->Next = RPHybridMerger_Yield;
-            return rp->Next(rp, r);
+            consumed[i] = true;
+            numConsumed++;
+            continue;
           } else { // Strict
+            rm_free(consumed);
             return rc;
```
raz-mon: This early return is still problematic, from the same reason the other path was fixed

tests/cpptests/test_cpp_hybridmerger.cpp
```diff
@@ -879,6 +879,49 @@ TEST_F(HybridMergerTest, testHybridMergerLinear3Upstreams) {
...
+ * Test that hybrid merger correctly handles partial intersection with linear scoring
+ *
+ * Scoring function: Hybrid linear
+ * Number of upstreams: 2
+ * Intersection: Partial intersection (documents 2,3 appear in both upstreams)
+ * Emptiness: Both upstreams have documents
+ * Timeout: No timeout
+ * Expected behavior: Documents 2,3 get combined scores from both upstreams
+ */
+TEST_F(HybridMergerTest, testHybridMergerPartialIntersection) {
```
raz-mon, line 902: No such test for RRF?

Itzikvaknin: added

tests/cpptests/test_cpp_hybridmerger.cpp
```diff
@@ -1041,3 +1048,57 @@ TEST_F(HybridMergerTest, testRRFScoring3Upstreams) {
...
+  ResultProcessor *rpTail = qitr.endProc;
+  int result;
+  size_t count = 0;
+
+  // Try to get results - should return error immediately due to query iterator error
+  while ((result = rpTail->Next(rpTail, &r)) == RS_RESULT_OK) {
+    count++;
+    SearchResult_Clear(&r);
+    // Safety check to avoid infinite loop in case of unexpected behavior
+    if (count > 10) break;
```
raz-mon: ??

tests/cpptests/test_cpp_hybridmerger.cpp
```diff
@@ -1041,3 +1048,57 @@ TEST_F(HybridMergerTest, testRRFScoring3Upstreams) {
...
+  QITR_PushRP(&qitr, hybridMerger);
+
+  // Process and verify that the most severe error (RS_RESULT_ERROR) is returned
+  SearchResult r = {0};
+  ResultProcessor *rpTail = qitr.endProc;
+  int result;
+  size_t count = 0;
+
+  // Try to get results - should return error immediately due to query iterator error
+  while ((result = rpTail->Next(rpTail, &r)) == RS_RESULT_OK) {
```
raz-mon: We know the first call will be an error no?
Why the loop?

tests/cpptests/test_cpp_hybridmerger.cpp
```diff
@@ -1041,3 +1048,57 @@ TEST_F(HybridMergerTest, testRRFScoring3Upstreams) {
...
+  qitr.err = &qerr;
+
+  // Set error in query iterator at the beginning
+  QueryError_SetError(&qerr, QUERY_EGENERIC, "Test error in query iterator");
+
+  // Create upstreams with different behaviors:
+  // upstream1: returns RS_RESULT_TIMEDOUT after 1 call
+  // upstream2: returns RS_RESULT_EOF (empty upstream)
+  MockUpstream upstream1(1, {1.0, 2.0}, {1, 2}); // timeoutAfterCount=1 (timeout after 1 call)
+  MockUpstream upstream2(0, {}, {}); // empty upstream (returns EOF)
```
raz-mon, line 1132: Missing the argument for the `errorAfterCount=0, so this will not return an error

src/result_processor.c
```diff
@@ -1696,15 +1696,24 @@ dictType dictTypeHybridSearchResult = {
...
    RPHybridMerger *self = (RPHybridMerger *)rp;
-   if (unlikely(!self->iterator)) {
+
+   if (QueryError_HasError(rp->parent->err)) {
+     return RS_RESULT_ERROR;
+   }
+
+   if (self->timedOut && rp->parent->timeoutPolicy == TimeoutPolicy_Fail){
+     return RS_RESULT_TIMEDOUT;
+   }
```
raz-mon: This should happen once after the accumulation phase, as it won't change throughout the yielding process

src/result_processor.c
```diff
@@ -1696,15 +1696,24 @@ dictType dictTypeHybridSearchResult = {
...
+
+   if (!self->iterator) {
     // Initialize iterator for yield phase
     self->iterator = dictGetIterator(self->hybridResults);
    }
```
raz-mon: This as well

src/result_processor.c
Itzikvaknin: Moved

src/result_processor.c
Itzikvaknin: Moved

tests/cpptests/test_cpp_hybridmerger.cpp
Itzikvaknin: Removed

tests/cpptests/test_cpp_hybridmerger.cpp
Itzikvaknin: Just reusing the same pattern across all the tests

raz-mon (approved):
Nice work 💪

src/hybrid_scoring.h
```diff
@@ -0,0 +1,51 @@
...
+#include "redisearch.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+ typedef enum {
+   HYBRID_SCORING_LINEAR,
+   HYBRID_SCORING_RRF
+ } HybridScoringType;
```
kei-nan:
```suggestion
typedef enum {
  HYBRID_SCORING_LINEAR,
  HYBRID_SCORING_RRF
} HybridScoringType;
```

src/result_processor.c
```diff
@@ -1589,3 +1592,229 @@ ResultProcessor *RPDepleter_New(StrongRef sync_ref) {
...
+  hybridResult->searchResult = searchResult;
+  hybridResult->numSources = numUpstreams;
+  hybridResult->scores = rm_calloc(numUpstreams, sizeof(double));
+  hybridResult->hasScores = rm_calloc(numUpstreams, sizeof(bool));
+  return hybridResult;
+}
+
+// Destructor for HybridSearchResult
+static void HybridSearchResult_Free(HybridSearchResult* hybridResult) {
+  if (hybridResult) {
```
kei-nan, line 1616: checking pointer is valid seems a bit weird.
maybe an assert would be better?
any case where the pointer would be null seems like a logical flow issue.

tests/cpptests/test_cpp_hybridmerger.cpp
```diff
@@ -0,0 +1,1103 @@
...
+ * Test that hybrid merger correctly handles error precedence with three upstreams returning different states
+ *
+ * Scoring function: Hybrid linear
+ * Number of upstreams: 3
+ * Intersection: N/A (error handling test)
+ * Emptiness: Mixed (one upstream EOF, one timeout, one error)
+ * Timeout: Mixed (one upstream EOF, one timeout, one error)
+ * Expected behavior: Return RS_RESULT_ERROR (most severe error) when one upstream returns error, regardless of other upstream states
+ */
+TEST_F(HybridMergerTest, testHybridMergerErrorPrecedence) {
```
kei-nan, line 1073: 👍

src/result_processor.c
Itzikvaknin: We also validate the pointer in `srDtor`, so i followed the same logic here since its just a wrapper of `SearchResult`

kei-nan (approved):

raz-mon (approved):

PR #6322, opened 2025-06-15
Refactor Intersection Iterator - [MOD-8842]

GuyAv46:

Implementation is based on #5619 

Micro-benchmarks show an improvement of up to 37%

#### Mark if applicable

### DIFF
--- a/src/iterators/intersection_iterator.c
+++ b/src/iterators/intersection_iterator.c
@@ -22,10 +22,13 @@ static IteratorStatus II_AgreeOnDocId(IntersectionIterator *it) {
   const t_docId docId = it->base.lastDocId;
   AggregateResult_Reset(it->base.current);
   for (int i = 0; i < it->num_its; i++) {
+    RS_ASSERT(it->its[i]->lastDocId <= docId);
     if (it->its[i]->lastDocId < docId) {
+      // Advance the iterator to the requested docId
       IteratorStatus rc = it->its[i]->SkipTo(it->its[i], docId);
       if (rc != ITERATOR_OK) {
         if (rc == ITERATOR_EOF) {
+          // Some child iterator reached EOF, so the intersection is also at EOF
           it->base.atEOF = true;
         } else if (rc == ITERATOR_NOTFOUND) {
           it->base.lastDocId = it->its[i]->lastDocId;
@@ -71,8 +74,16 @@ static IteratorStatus II_Read_CheckRelevancy(QueryIterator *base) {
   if (base->atEOF) {
     return ITERATOR_EOF;
   }
+  const t_docId prevLastDocId = base->lastDocId;
+
   base->lastDocId++; // advance the last docId. Current docId is at least this
-  return II_Read_Internal_CheckRelevancy(it);
+  IteratorStatus rc = II_Read_Internal_CheckRelevancy(it);
+  if (rc != ITERATOR_OK) {
+    // If we didn't find a relevant result, we need to reset the lastDocId to the previous value
+    base->lastDocId = prevLastDocId;
+    AggregateResult_Reset(base->current);
+  }
+  return rc;
 }
 
 static IteratorStatus II_Read(QueryIterator *base) {
@@ -80,15 +91,25 @@ static IteratorStatus II_Read(QueryIterator *base) {
   if (base->atEOF) {
     return ITERATOR_EOF;
   }
+  const t_docId prevLastDocId = base->lastDocId;
+
   base->lastDocId++; // advance the last docId. Current docId is at least this
-  return II_Read_Internal(it);
+  IteratorStatus rc = II_Read_Internal(it);
+  if (rc != ITERATOR_OK) {
+    // If we didn't find a relevant result, we need to reset the lastDocId to the previous value
+    base->lastDocId = prevLastDocId;
+    AggregateResult_Reset(base->current);
+  }
+  return rc;
 }
 
 static IteratorStatus II_SkipTo_CheckRelevancy(QueryIterator *base, t_docId docId) {
+  RS_ASSERT(base->lastDocId < docId);
   IntersectionIterator *it = (IntersectionIterator *)base;
   if (base->atEOF) {
     return ITERATOR_EOF;
   }
+  const t_docId prevLastDocId = base->lastDocId;
 
   base->lastDocId = docId;
 
@@ -98,22 +119,31 @@ static IteratorStatus II_SkipTo_CheckRelevancy(QueryIterator *base, t_docId docI
       return ITERATOR_OK;
     }
     // Agreed on docId, but not relevant - need to read the next valid result.
-    it->base.lastDocId++; // advance the last docId to the next possible value
+    base->lastDocId++; // advance the last docId to the next possible value
   } else if (ITERATOR_NOTFOUND != rc) {
+    base->lastDocId = prevLastDocId; // reset the lastDocId to the previous value
+    AggregateResult_Reset(base->current);
     return rc; // Unexpected - bubble up
   }
 
   // Not found - but we need to read the next valid result.
   rc = II_Read_Internal_CheckRelevancy(it);
   // Return rc, switching OK to NOTFOUND
-  return (rc == ITERATOR_OK) ? ITERATOR_NOTFOUND : rc; // Unexpected - bubble up
+  if (rc == ITERATOR_OK) {
+    return ITERATOR_NOTFOUND;
+  }
+  base->lastDocId = prevLastDocId; // reset the lastDocId to the previous value
+  AggregateResult_Reset(base->current);
+  return rc; // Unexpected - bubble up
 }
 
 static IteratorStatus II_SkipTo(QueryIterator *base, t_docId docId) {
+  RS_ASSERT(base->lastDocId < docId);
   IntersectionIterator *it = (IntersectionIterator *)base;
   if (base->atEOF) {
     return ITERATOR_EOF;
   }
+  const t_docId prevLastDocId = base->lastDocId;
 
   base->lastDocId = docId;
 
@@ -122,6 +152,10 @@ static IteratorStatus II_SkipTo(QueryIterator *base, t_docId docId) {
     // Not found - but we need to read the next valid result
     rc = II_Read_Internal(it);
     if (rc == ITERATOR_OK) rc = ITERATOR_NOTFOUND;
+  }
+  if (!(ITERATOR_OK == rc || ITERATOR_NOTFOUND == rc)) {
+    base->lastDocId = prevLastDocId; // reset the lastDocId to the previous value
+    AggregateResult_Reset(base->current);
   }
   return rc;
 }
@@ -136,7 +170,7 @@ static void II_Rewind(QueryIterator *base) {
 
   base->atEOF = false;
   base->lastDocId = 0;
-  base->current->docId = 0;
+  AggregateResult_Reset(base->current);
 
   // rewind all child iterators
   for (int i = 0; i < ii->num_its; i++) {
--- a/src/iterators/intersection_iterator.c
+++ b/src/iterators/intersection_iterator.c
@@ -21,7 +21,7 @@ static inline bool II_currentIsRelevant(IntersectionIterator *it) {
 static IteratorStatus II_AgreeOnDocId(IntersectionIterator *it) {
   const t_docId docId = it->base.lastDocId;
   AggregateResult_Reset(it->base.current);
-  for (int i = 0; i < it->num_its; i++) {
+  for (uint32_t i = 0; i < it->num_its; i++) {
     RS_ASSERT(it->its[i]->lastDocId <= docId);
     if (it->its[i]->lastDocId < docId) {
       // Advance the iterator to the requested docId
--- a/src/iterators/intersection_iterator.c
+++ b/src/iterators/intersection_iterator.c
@@ -69,7 +69,7 @@ static IteratorStatus II_AgreeOnDocId(IntersectionIterator *it, t_docId *curTarg
   return ITERATOR_OK;
 }
 
-static inline IteratorStatus II_Read_Internal_CheckRelevancy(IntersectionIterator *it, t_docId docId) {
+static inline IteratorStatus II_Find_Consensus_WithRelevancyCheck(IntersectionIterator *it, t_docId docId) {
   IteratorStatus rc;
   do { // retry until we agree on the docId
     rc = II_AgreeOnDocId(it, &docId);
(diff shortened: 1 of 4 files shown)


JoanFM (commented):
There is no need for IT_V2 ? I think

src/iterators/intersection_iterator.c
```diff
@@ -0,0 +1,271 @@
...
+  for (size_t i = 0; i < it->num_its; ++i) {
+    QueryIterator *cur = it->its[i];
+    if (!cur) {
+      // If the current iterator is empty, then the entire query will fail
+      it->num_expected = 0;
+      return false;
+    }
+    size_t amount = cur->NumEstimated(cur);
+    if (amount < it->num_expected) {
+      it->num_expected = amount;
```
JoanFM, line 248: If it is 0, we should also return false, we also know it would fail?

src/iterators/intersection_iterator.c
```diff
@@ -0,0 +1,271 @@
...
+    RS_ASSERT(it->its[i]->lastDocId <= docId);
+    if (it->its[i]->lastDocId < docId) {
+      // Advance the iterator to the requested docId
+      IteratorStatus rc = it->its[i]->SkipTo(it->its[i], docId);
+      if (rc != ITERATOR_OK) {
+        if (rc == ITERATOR_EOF) {
+          // Some child iterator reached EOF, so the intersection is also at EOF
+          it->base.atEOF = true;
+        } else if (rc == ITERATOR_NOTFOUND) {
+          it->base.lastDocId = it->its[i]->lastDocId;
```
JoanFM: I miss this logic, how are you sure you need to move here? I may miss some logic surrounding this? Maybe a better function naming can provide more context

src/iterators/intersection_iterator.c
```diff
@@ -0,0 +1,271 @@
...
+        }
+        return rc;
+      }
+    }
+    AggregateResult_AddChild(it->base.current, it->its[i]->current);
```
JoanFM, line 67: Also why is this?

src/iterators/intersection_iterator.c
```diff
@@ -0,0 +1,271 @@
...
+static inline bool II_currentIsRelevant(IntersectionIterator *it) {
+  return IndexResult_IsWithinRange(it->base.current, it->max_slop, it->in_order);
+}
+
+static IteratorStatus II_AgreeOnDocId(IntersectionIterator *it) {
+  const t_docId docId = it->base.lastDocId;
+  AggregateResult_Reset(it->base.current);
+  for (uint32_t i = 0; i < it->num_its; i++) {
+    RS_ASSERT(it->its[i]->lastDocId <= docId);
+    if (it->its[i]->lastDocId < docId) {
```
JoanFM, line 47: doesn't it play a role if the child is atEOF already?

JoanFM: I guess this is checked elsewhere in the non static code

JoanFM:
I cannot think now of a better way, but these multiple level loop seems hard to follow. Isn't there a way to make the loop over `AgreeOnId` more explicit?

src/iterators/intersection_iterator.c
GuyAv46: I don't feel comfortable with freeing all the iterators without attempting to query them based on an estimation. I also don't think it's a case worth optimizing

src/iterators/intersection_iterator.c
GuyAv46: I'll add more comments

src/iterators/intersection_iterator.c
GuyAv46: If a child returns EOF, we also return EOF, so it is not expected before we read from it

GuyAv46:
> There is no need for IT_V2? I think

I renamed the constructor `NewIntersectionIterator` instead of `NewIntersectIterator` to be more aligned with `NewUnionIterator` and more grammatically correct, so the `IT_V2` macro is not required

> I cannot think now of a better way, but these multiple level loop seems hard to follow. Isn't there a way to make the loop over AgreeOnId more explicit?

I don't think so, and it's better to maintain the current separation due to the differences in the outer loop when validation is required.
We must have 2 loops for the core logic 
```
while (no agreement) {
    for (every child) {
        check if the current child has the current target ID
    }
}
```

src/iterators/intersection_iterator.c
```diff
@@ -0,0 +1,294 @@
...
+
+static inline IteratorStatus II_Read_Internal_CheckRelevancy(IntersectionIterator *it, t_docId docId) {
+  IteratorStatus rc;
+  do { // retry until we agree on the docId
+    rc = II_AgreeOnDocId(it, &docId);
+    if (rc != ITERATOR_OK) {
+      continue;
+    }
+    if (!II_currentIsRelevant(it)) {
+      rc = II_ReadFromFirstChild(it, &docId);
```
JoanFM, line 80: This read is about "unblocking" the logic for the next AgreeOnDocId to start from fresh start?

src/iterators/intersection_iterator.c
```diff
@@ -0,0 +1,294 @@
...
+    }
+    // Hit!
+    it->base.lastDocId = it->base.current->docId;
+    return ITERATOR_OK;
+
+  } while (rc == ITERATOR_OK || rc == ITERATOR_NOTFOUND);
+  return rc;
+}
+
+static inline IteratorStatus II_Read_Internal(IntersectionIterator *it, t_docId docId) {
```
JoanFM: For better understanding of the code, would `II_Find_Consensus` be ab better name? (in counter to the ReadFirstChild)

src/iterators/intersection_iterator.c
```diff
@@ -0,0 +1,294 @@
...
+  // rewind all child iterators
+  for (int i = 0; i < ii->num_its; i++) {
+    ii->its[i]->Rewind(ii->its[i]);
+  }
+}
+
+static void II_Free(QueryIterator *base) {
+  IntersectionIterator *ii = (IntersectionIterator *)base;
+
+  for (int i = 0; i < ii->num_its; i++) {
```
JoanFM: As per copilot, here also `size_t`?

src/iterators/intersection_iterator.c
```diff
@@ -0,0 +1,294 @@
...
+ * If they do, aggregate their results into `current` and return ITERATOR_OK.
+ * If any of the iterators is at EOF, set `atEOF` to true and return ITERATOR_EOF.
+ * If any of the iterators is not at the requested docId, advance it to the requested `docId` and
+ * return ITERATOR_NOTFOUND. The caller may retry calling this function in that case.
+ * In case of an error, return the error code.
+ */
+static IteratorStatus II_AgreeOnDocId(IntersectionIterator *it, t_docId *curTarget) {
+  const t_docId docId = *curTarget;
+
+  for (uint32_t i = 0; i < it->num_its; i++) {
```
JoanFM, line 45: or size_t here or uint32_t in the other parts?

tests/cpptests/test_cpp_iterator_intersection.cpp
```diff
@@ -0,0 +1,353 @@
...
+        unsigned numChildren;
+        std::tie(numChildren, resultSet) = GetParam();
+        // Set docIds so the intersection of all children is resultSet.
+        // Make sure that some ids are unique to each child
+        docIds.resize(numChildren);
+        t_docId id = 1;
+        for (auto &childIds : docIds) {
+            // Copy the resultSet to each child as a base
+            childIds = resultSet;
+            // Add some unique ids to each child. Mock constructors will ensure that the ids are unique and sorted.
```
JoanFM: I think there is no damage to do the uniqueness and sorting here even if it is done twice. For tests clarity should come before optimization.

tests/cpptests/test_cpp_iterator_intersection.cpp
```diff
@@ -0,0 +1,353 @@
...
+            for (size_t i = 0; i < 100; i++) {
+                childIds.push_back(id++);
+            }
+        }
+        // Create children iterators
+        auto children = (QueryIterator **)rm_malloc(sizeof(QueryIterator *) * numChildren);
+        for (unsigned i = 0; i < numChildren; i++) {
```
JoanFM, line 44: Somehow it would be nice to "randomize" the original order of the Children iterator to validate the sorting in constructor

tests/cpptests/test_cpp_iterator_intersection.cpp
```diff
@@ -0,0 +1,353 @@
...
+INSTANTIATE_TEST_SUITE_P(IntersectionIterator, IntersectionIteratorCommonTest, ::testing::Combine(
+    ::testing::Values(2, 5, 25),
+    ::testing::Values(
+        std::vector<t_docId>{1, 2, 3, 40, 50},
+        std::vector<t_docId>{5, 6, 7, 24, 25, 46, 47, 48, 49, 50, 51, 234, 2345, 3456, 4567, 5678, 6789, 7890, 8901, 9012, 12345, 23456, 34567, 45678, 56789},
+        std::vector<t_docId>{9, 25, 30, 40, 50, 60, 70, 80, 90, 100, 110, 120, 130, 140, 150, 160, 170, 180, 190, 200, 210, 220, 230, 240, 250}
+    )
+));
+
+class IntersectionIteratorTest : public ::testing::Test {
```
JoanFM, line 162: Isn't there a way to use mocks so that we do not need to create an InvertedIndex.

JoanFM: Some sort of basic CheckRelevancy? this is the only logic we should test in such a unit test

src/iterators/intersection_iterator.c
GuyAv46: At this point, all the iterators agree on the ID, but it's irrelevant, so we read from the first child to get the next target ID. This is also the best child to read from when `in_order == false`, as the first child is expected to have the least amount of results

tests/cpptests/test_cpp_iterator_intersection.cpp
GuyAv46: For these tests, I need the proper encoding of the entries to use the relevancy test. It's simpler to use actual indexes than to implement a mock that does the same thing.
The mock is currently more designed to be light and fast for the benchmarks, and also to get a set of IDs with no additional metadata

tests/cpptests/test_cpp_iterator_intersection.cpp
GuyAv46: With the current setup, they won't be sorted initially (depends on the setup). I added a check that they are sorted by the constructor

tests/cpptests/test_cpp_iterator_intersection.cpp
GuyAv46: I find it less readable to add unnecessary steps. It makes it look like we expect the results to be unsorted twice for some reason.
The mock constructor handles any set of results as input. Why not rely on it?

src/iterators/intersection_iterator.c
```diff
@@ -0,0 +1,294 @@
...
+    base->SkipTo = II_SkipTo_CheckRelevancy;
+  }
+  base->Free = II_Free;
+  base->Rewind = II_Rewind;
+
+  if (!allValid) {
+    // Some of the iterators are NULL, so the intersection will always be empty.
+    base->Free(base);
+    base = IT_V2(NewEmptyIterator)();
+  }
```
BenGoldberger, line 293: why not check for if before and return?

GuyAv46: So I can call `base->Free(base);` without implementing another cleanup mechanism (we need to free all the children).
This is not a common case (if it can even happen), so I don't see a reason to optimize it

src/iterators/intersection_iterator.c
```diff
@@ -0,0 +1,295 @@
...
+  return rc;
+}
+
+static IteratorStatus II_Read(QueryIterator *base) {
+  IntersectionIterator *it = (IntersectionIterator *)base;
+  if (base->atEOF) {
+    return ITERATOR_EOF;
+  }
+  t_docId target;
+  IteratorStatus rc = II_ReadFromFirstChild(it, &target);
```
BenGoldberger, line 123: is this optimization needed?
won't `II_AgreeOnDocId` do the same thing on the first iteration?

src/iterators/intersection_iterator.c
```diff
@@ -0,0 +1,295 @@
...
+  RS_ASSERT(base->lastDocId < docId);
+  IntersectionIterator *it = (IntersectionIterator *)base;
+  if (base->atEOF) {
+    return ITERATOR_EOF;
+  }
+
+  IteratorStatus rc = II_AgreeOnDocId(it, &docId);
+  if (ITERATOR_OK == rc) {
+    it->base.lastDocId = it->base.current->docId;
+  } else if (ITERATOR_NOTFOUND == rc) {
```
BenGoldberger, line 170: isn't a single iteration of `II_Find_Consensus`? why cant we just call `II_Find_Consensus`

src/iterators/intersection_iterator.c
```diff
@@ -0,0 +1,295 @@
...
+    }
+    // Agreed on docId, but not relevant - need to read the next valid result.
+    rc = II_ReadFromFirstChild(it, &docId);
+    if (rc != ITERATOR_OK) return rc; // If we failed to read, bubble up the error
+  } else if (ITERATOR_NOTFOUND != rc) {
+    return rc; // Unexpected - bubble up
+  }
+
+  // Not found - but we need to read the next valid result.
+  // `docId` is set to the next valid docId (by `II_AgreeOnDocId` or `II_ReadFromFirstChild`).
```
BenGoldberger, line 151: again it looks like a single iteration of `II_Find_Consensus_WithRelevancyCheck`
I wonder if it needed?

JoanFM (approved):

src/iterators/intersection_iterator.c
GuyAv46: Almost the same, but if we hit the requested ID, we want to return `ITERATOR_OK`; otherwise, the next one we have consensus on, but return `ITERATOR_NOTFOUND`.

src/iterators/intersection_iterator.c
GuyAv46: This is kind of manually doing the first iteration of `Find_Consensus`, but we must do it to differentiate between a hit on the requested ID and other results

src/iterators/intersection_iterator.c
GuyAv46: `II_AgreeOnDocId` needs a target ID.
The alternative (which was my original implementation) is to start with `it->lastDocId + 1`, but this is a bit less efficient;
1. It adds another iteration of `II_AgreeOnDocId` if the first child doesn't have this ID
2. It requires a `child->SkipTo` call, which is usually less performant than a `Read` (especially with `it->SkipTo(it->lastDocId + 1)`, which will return the same result as `it->Read()`)

src/iterators/intersection_iterator.c
BenGoldberger: then maybe we can add a boolean for the first iteration?
what do you think? it may have a cleaner result

BenGoldberger (approved):

PR #6140, opened 2025-05-15
[MOD-9560]  Revert marking index as faulty after reaching OOM

lerman25:
This PR changes the following: 
- When a background scan for Index fails due to OOM, queries on that index will not longer return an error but a partial result, based on what was indexed during the scan. 
- Resp3 warning for each query.

Continuing the work of #6114 , #6053 , #5778 .

PR #6114, opened 2025-05-11
[MOD-9372 , MOD-9733] Stop indexing OOM - Add wait before OOM

lerman25:
This PR adds a wait-before-OOM option, extending the work in #5778.
In Enterprise we want to give the Resource Manager time to raise memory limits when background indexing hits OOM.
If 2 consecutive OOM were reached - the scan stops. 

We therefore introduce the BG_INDEX_OOM_PAUSE_TIME config, which sets how long the background-indexing thread sleeps before it restarts the scan.

- Enterprise default: 5 s
- OSS default: 0 s (no pause, no restart)

While the scan is paused, a Redis log entry is written for visibility.

For testing, 2 new command are introduced for the `BG_SCAN_CONTROLLER` (https://github.com/RediSearch/RediSearch/pull/5672 ) - `SET_PAUSE_BEFORE_OOM_RESET` and `DEBUG_SCANNER_UPDATE_CONFIG`.
The first - pauses before we wait and restart the scan
The latter - updates the debug indexes scanner if a config was changed after it's initialization. 
Added scanner status PAUSED_BEFORE_OOM_RESET to support the above.

PR #6323, opened 2025-06-15
[2.10] Backport OOM indexing to all branches [MOD-9715]

lerman25:
This PR backports all OOM indexing PRs to 2.8.
https://github.com/RediSearch/RediSearch/pull/5672, https://github.com/RediSearch/RediSearch/pull/5778, https://github.com/RediSearch/RediSearch/pull/5941, https://github.com/RediSearch/RediSearch/pull/6053, https://github.com/RediSearch/RediSearch/pull/6114, https://github.com/RediSearch/RediSearch/pull/6140, https://github.com/RediSearch/RediSearch/pull/6203, https://github.com/RediSearch/RediSearch/pull/6276,#6297

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -650,6 +650,37 @@ CONFIG_GETTER(getIndexerYieldEveryOps) {
   return sdscatprintf(ss, "%u", config->indexerYieldEveryOpsWhileLoading);
 }
 
+// SET MEMORY LIMIT PERCENTAGE
+CONFIG_SETTER(setIndexingMemoryLimit) {
+  uint8_t newLimit;
+  int acrc = AC_GetU8(ac, &newLimit, AC_F_GE0);
+  CHECK_RETURN_PARSE_ERROR(acrc);
+  if (newLimit > 100) {
+    QueryError_SetWithoutUserDataFmt(status, QUERY_ELIMIT, "Memory limit for indexing cannot be greater then 100%%");
+    return REDISMODULE_ERR;
+  }
+  config->indexingMemoryLimit = newLimit;
+  return REDISMODULE_OK;
+}
+
+CONFIG_GETTER(getIndexingMemoryLimit) {
+  sds ss = sdsempty();
+  return sdscatprintf(ss, "%u", config->indexingMemoryLimit);
+}
+
+CONFIG_SETTER(setBgOOMpauseTimeForRetry) {
+  uint32_t newPauseTime;
+  int acrc = AC_GetU32(ac, &newPauseTime, AC_F_GE0);
+  CHECK_RETURN_PARSE_ERROR(acrc);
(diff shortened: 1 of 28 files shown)


src/config.c
```diff
@@ -650,6 +650,37 @@ CONFIG_GETTER(getIndexerYieldEveryOps) {
...
 }
 
+// SET MEMORY LIMIT PERCENTAGE
+CONFIG_SETTER(setIndexingMemoryLimit) {
+  uint8_t newLimit;
+  int acrc = AC_GetU8(ac, &newLimit, AC_F_GE0);
+  CHECK_RETURN_PARSE_ERROR(acrc);
+  if (newLimit > 100) {
+    QueryError_SetError(status, QUERY_ELIMIT, "Memory limit for indexing cannot be greater then 100%%");
+    return REDISMODULE_ERR;
```
alonre24, line 660: typo in `100%%`? 
This isn't covered in other branches as well?

src/spec.c
```diff
@@ -2802,9 +2964,17 @@ int IndexSpec_UpdateDoc(IndexSpec *spec, RedisModuleCtx *ctx, RedisModuleString
     return REDISMODULE_ERR;
   }
 
+  QueryError status = {0};
+
+  if(spec->scan_failed_OOM) {
+    QueryError_SetError(&status, QUERY_INDEXBGOOMFAIL, "Index background scan did not complete due to OOM. New documents will not be indexed.");
```
alonre24, line 2970:
```suggestion
    QueryError_SetError(&status, QUERY_INDEXBGOOMFAIL, "Index background scan did not complete due to OOM. Index may contain only partial data.");
```
Nitpick: This error message sounds more appropriate as we are not setting the index as faulty anymore, WDYT?

src/config.c
lerman25: Without the extra % it doesn't print the % to the cli
Should be covered automatically by the test_config no?

src/spec.c
lerman25: Will change in all version

lerman25: It's to notify that the index will not receive new docs (I understood that's what should happen).
If that's not the case I'll fix that

src/config.c
alonre24: Oh I see. Seems like, but it was marked by codecov as if it was not covered for some reason

src/spec.c
alonre24: What if the memory limit increased after a while, and some will try to insert a document synchronously via HSET?

lerman25: It will not insert the doc to the index

src/config.c
lerman25: It is missing, I will fix it

alonre24 (approved):

PR #6325, opened 2025-06-16
MOD-10030: Add background depleter result processor

raz-mon:
Adds a background depleter result processor.

The new result processor is responsible for depleting its upstream in the background, such that the downstream can do other tasks in the meanwhile.

Main characteristics:
* Upon the first call to the depleter, the depleting thread is spawned, and `RS_RESULT_DEPLETING` is ALWAYS returned.
* While the BG thread is still depleting the pipeline (via its upstream), `RS_RESULT_DEPLETING` is returned to the downstream caller.
* When the BG thread is finished, results are returned one by one until the collected results are depleted.

Some points for optimization which we may wish to exploit at some point:
* Returning results that are ready prior to the BG thread being finished with its task.
* Returning the full list of results all at once.
* There are more opportunities here, that require some tailor made behavior.

### DIFF
--- a/src/module-init/module-init.c
+++ b/src/module-init/module-init.c
@@ -33,6 +33,7 @@
 #include "profile.h"
 #include "info/info_redis/info_redis.h"
 
+#define DEPLETER_POOL_SIZE 4
 
 /**
  * Check if we can run under the current AOF configuration. Returns true
@@ -166,6 +167,8 @@ int RediSearch_Init(RedisModuleCtx *ctx, int mode) {
   }
   DO_LOG("verbose", "threadpool has %lu high-priority bias that always prefer running queries "
                     "when possible", RSGlobalConfig.highPriorityBiasNum);
+
+  depleterPool = redisearch_thpool_create(DEPLETER_POOL_SIZE, DEFAULT_HIGH_PRIORITY_BIAS_THRESHOLD, LogCallback, "depleter");
 
   IndexAlias_InitGlobal();
 
--- a/src/result_processor.c
+++ b/src/result_processor.c
@@ -15,6 +15,7 @@
 #include "rmutil/rm_assert.h"
 #include "util/timeout.h"
 #include "util/arr.h"
+#include <stdatomic.h>
 /*******************************************************************************************************************
  *  General Result Processor Helper functions
  *******************************************************************************************************************/
@@ -1333,7 +1334,7 @@ void PipelineAddCrash(struct AREQ *r) {
    *
    * The processor works in two phases:
    * 1. Accumulation: Gather all results from upstream and find the max score.
-   * 2. Yield: Normalize each result’s score by division with the max score, then pass
+   * 2. Yield: Normalize each result's score by division with the max score, then pass
    *    it downstream.
   *******************************************************************************************************************/
  typedef struct {
@@ -1424,5 +1425,89 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
   ret->base.Free = RPMaxScoreNormalizer_Free;
   ret->base.type = RP_MAX_SCORE_NORMALIZER;
   ret->scoreKey = rlk;
+  return &ret->base;
+}
+
+/*******************************************************************************************************************
+ *  Depleter Result Processor
+ *
+ *  The RPDepleter result processor offloads the task of consuming all results from its upstream processor
+ *  into a background thread, storing them in an internal array. While the background thread is running,
+ *  calls to Next() return RS_RESULT_DEPLETING. Once depleting is complete, Next() yields results one by one
+ *  from the internal array, and finally returns RS_RESULT_EOF.
+ *
+ *  This processor is useful for scenarios where the entire result set must be materialized before further
+ *  processing, and where background depleting can improve responsiveness or parallelism.
+ *
+ *  Threading: The depleting thread is dispatched via the global depleterPool thread pool. The processor
+ *  ensures thread-safe signaling using atomic variables.
+ *******************************************************************************************************************/
+typedef struct {
+  ResultProcessor base;                /* Base result processor struct */
+  arrayof(SearchResult *) results;     /* Array of pointers to SearchResult, filled by the depleting thread */
+  atomic_int done_depleting;           /* Set to 1 when depleting is finished */
+  uint cur_idx;                        /* Current index for yielding results */
+} RPDepleter;
+
+/**
+ * Frees all resources associated with the RPDepleter, including the results array.
+ */
+static void RPDepleter_Free(ResultProcessor *base) {
+  RPDepleter *self = (RPDepleter *)base;
+  array_free_ex(self->results, srDtor(*(char **)ptr));
+  rm_free(self);
+}
+
+/**
+ * Background thread function: consumes all results from upstream and stores them in the results array.
+ * Signals completion by setting done_depleting to 1.
+ */
+static void RPDepleter_Deplete(void *arg) {
+  RPDepleter *self = (RPDepleter *)arg;
+
+  // Deplete the pipeline into the `self->results` array.
+  SearchResult *r = rm_calloc(1, sizeof(*r));
+  while (self->base.upstream->Next(self->base.upstream, r) == RS_RESULT_OK) {
+    array_append(self->results, r);
+    r = rm_calloc(1, sizeof(*r));
+  }
+  rm_free(r);
+  atomic_store(&self->done_depleting, 1);
+}
+
+/**
+ * Next function for RPDepleter. On the first call, dispatches the depleting thread and returns
+ * RS_RESULT_DEPLETING until the background thread is done. Once depleting is complete, yields results
+ * from the internal array, and finally returns RS_RESULT_EOF.
+ */
+static int RPDepleter_Next(ResultProcessor *base, SearchResult *r) {
+  RPDepleter *self = (RPDepleter *)base;
+  // First call to next will start the depleting thread, and return `RS_RESULT_DEPLETING`.
+  redisearch_thpool_add_work(depleterPool, RPDepleter_Deplete, self, THPOOL_PRIORITY_HIGH);
+
+  // If the depleting thread is done, return the results.
+  if (atomic_load(&self->done_depleting) == 1) {
+    if (self->cur_idx == array_len(self->results)) {
+      // We've reached the end of the array, return EOF.
+      return RS_RESULT_EOF;
+    }
+    // Return the next result in the array.
+    SearchResult *current = self->results[self->cur_idx++];
+    SearchResult_Override(r, current);    // Copy result data to output
+    return RS_RESULT_OK;
+  }
+  return RS_RESULT_DEPLETING;
+}
+
+/**
+ * Constructs a new RPDepleter processor, wrapping the given upstream processor.
+ * The returned processor takes ownership of result depleting and yielding.
+ */
+ResultProcessor *RPDepleter_New() {
+  RPDepleter *ret = rm_calloc(1, sizeof(*ret));
+  ret->results = array_new(SearchResult*, 0);
+  ret->base.Next = RPDepleter_Next;
+  ret->base.Free = RPDepleter_Free;
+  ret->base.type = RP_DEPLETER;
   return &ret->base;
 }
--- a/src/result_processor.c
+++ b/src/result_processor.c
@@ -1434,13 +1434,7 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
  *  The RPDepleter result processor offloads the task of consuming all results from its upstream processor
  *  into a background thread, storing them in an internal array. While the background thread is running,
  *  calls to Next() return RS_RESULT_DEPLETING. Once depleting is complete, Next() yields results one by one
- *  from the internal array, and finally returns RS_RESULT_EOF.
- *
- *  This processor is useful for scenarios where the entire result set must be materialized before further
- *  processing, and where background depleting can improve responsiveness or parallelism.
- *
- *  Threading: The depleting thread is dispatched via the global depleterPool thread pool. The processor
- *  ensures thread-safe signaling using atomic variables.
+ *  from the internal array, and finally returns the last return code from the upstream.
  *******************************************************************************************************************/
 typedef struct {
   ResultProcessor base;                // Base result processor struct
@@ -1452,7 +1446,7 @@ typedef struct {
 } RPDepleter;
 
 /**
- * Frees all resources associated with the RPDepleter, including the results array.
+ * Desctructor
  */
 static void RPDepleter_Free(ResultProcessor *base) {
   RPDepleter *self = (RPDepleter *)base;
@@ -1498,9 +1492,8 @@ static int RPDepleter_Next_Yield(ResultProcessor *base, SearchResult *r) {
 }
 
 /**
- * Next function for RPDepleter. Called on the first Next call.
- * Dispatches the depleting thread, sets the yielding Next function, and returns
- * `RS_RESULT_DEPLETING`.
+ * Next function for RPDepleter.
+ * Returns `RS_RESULT_DEPLETING` until the BG thread is done depleting.
  */
 static int RPDepleter_Next_Dispatch(ResultProcessor *base, SearchResult *r) {
   RPDepleter *self = (RPDepleter *)base;
--- a/src/result_processor.c
+++ b/src/result_processor.c
@@ -1443,10 +1443,12 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
  *  ensures thread-safe signaling using atomic variables.
  *******************************************************************************************************************/
 typedef struct {
-  ResultProcessor base;                /* Base result processor struct */
-  arrayof(SearchResult *) results;     /* Array of pointers to SearchResult, filled by the depleting thread */
-  atomic_int done_depleting;           /* Set to 1 when depleting is finished */
-  uint cur_idx;                        /* Current index for yielding results */
+  ResultProcessor base;                // Base result processor struct
+  arrayof(SearchResult *) results;     // Array of pointers to SearchResult, filled by the depleting thread
+  atomic_int done_depleting;           // Set to 1 when depleting is finished
+  uint cur_idx;                        // Current index for yielding results
+  RPStatus last_rc;                    // Last return code from upstream
+  bool first_call;                     // Whether the first call to Next has been made
 } RPDepleter;
 
 /**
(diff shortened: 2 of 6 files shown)


src/module-init/module-init.c
```diff
@@ -167,6 +168,8 @@ int RediSearch_Init(RedisModuleCtx *ctx, int mode) {
   DO_LOG("verbose", "threadpool has %lu high-priority bias that always prefer running queries "
                     "when possible", RSGlobalConfig.highPriorityBiasNum);
 
+  depleterPool = redisearch_thpool_create(DEPLETER_POOL_SIZE, DEFAULT_HIGH_PRIORITY_BIAS_THRESHOLD, LogCallback, "depleter");
```
kei-nan, line 171: Why use another thread pool?
Can't we use an existing thread pool?

src/result_processor.c
```diff
@@ -1426,3 +1427,107 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+  RPStatus last_rc;                    // Last return code from upstream
+  bool first_call;                     // Whether the first call to Next has been made
+} RPDepleter;
+
+/**
+ * Destructor
+ */
+static void RPDepleter_Free(ResultProcessor *base) {
+  RPDepleter *self = (RPDepleter *)base;
+  array_free_ex(self->results, srDtor(*(char **)ptr));
```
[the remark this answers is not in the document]
raz-mon: Not correct.

src/result_processor.c
```diff
@@ -1426,3 +1427,107 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+/**
+ * Background thread function: consumes all results from upstream and stores them in the results array.
+ * Signals completion by setting done_depleting to 1.
+ */
+static void RPDepleter_Deplete(void *arg) {
+  RPDepleter *self = (RPDepleter *)arg;
+  RPStatus rc;
+
+  // Deplete the pipeline into the `self->results` array.
+  SearchResult *r = rm_calloc(1, sizeof(*r));
```
kei-nan, line 1493: maybe keeping an array of search results will be better?
Instead of keeping pointers keep actual objects, wdyt?
Will reduce memory fragmentation.

src/result_processor.c
```diff
@@ -1426,3 +1427,107 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+    self->first_call = false;
+    // TODO: We may want this to be added to the `workers` thread pool instead
+    // of a new one. In case there are no workers, we do the depleting in the
+    // current thread.
+    redisearch_thpool_add_work(depleterPool, RPDepleter_Deplete, self, THPOOL_PRIORITY_HIGH);
+    return RS_RESULT_DEPLETING;
+  }
+
+  // If the depleting thread is done, return the results.
+  if (atomic_load(&self->done_depleting) == 1) {
```
kei-nan: There is no wait here, the calling thread is going to continually call Next?
Seems like wasting the cpu time.
Try and use a conditional variable and sleep on the variable for a bit here to avoid consuming the cpu too much.

kei-nan (changes requested):

src/module-init/module-init.c
raz-mon: Yes, we can basically use the `workers` thread-pool.
I didn't do that for two reasons:
1. Didn't want to mix up the different job types.
2. A user that does not set the workers config will get very poor performance from the `FT.HYBRID` command, since it will be the main-thread that will deplete both pipelines. We can see this as an integral part of the command, thus have a new and dedicated thread-pool for it.

I'm quite on the fence about what we should use among them, due to the above.

src/result_processor.c
raz-mon: Will consider

src/result_processor.c
raz-mon: I don't want it to sleep.
I want it to check which of its branches are done first and handle it while the other is working.
Otherwise I may sleep on one branch, while the other is done and ready for processing.

src/module-init/module-init.c
kei-nan: I know in vector sets there were talks about in some setup we don't allow multiple threads(I think it was tenant related)
So I don't think we should break this behaviour here.

1. But we are creating threads that might never be used if the user never uses the new command.
2. That is the same for a regular search.

src/result_processor.c
kei-nan: But the result processor which will use this result processor has no idea it needs to sleep due to the abstract nature of the result processor type.

So right now no one will sleep, which is also bad.

src/module-init/module-init.c
raz-mon: Regarding that setup - that doesn't make sense. Today we use multi-threading for several operations regardless of any configuration - e.g., indexing. So that has no affect on us, or else we need to change more existing code as well.

1. I'm quite sure we don't create them until they are actually used. See `thpool.c`.
2. Correct but this is a different use-case, that is running two queries, which we don't currently have, in which the single-command latency is highly affected by it. I understand this argument and it's the reason I'm on the fence with this (actually tending towards the `workers` thpool TBH). It doesn't have to be final now, since we're only merging to the feature-branch at this point.

src/result_processor.c
raz-mon: I don't agree.
The result processor that is the downstream of this result processor HAS to be aware of this upstream type to some extent since it needs to perform the logic that checks which upstream is ready and start working on it.
They will have to be somewhat coupled. We are trying to have them as least coupled as possible but this is an important part I think we need to have this way.

src/result_processor.c
```diff
@@ -1426,3 +1427,107 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+
+  // Depleting thread is done, it's safe to return the results.
+  if (self->cur_idx == array_len(self->results)) {
+    // We've reached the end of the array, return the last code from the upstream.
+    return self->last_rc;
+  }
+  // Return the next result in the array.
+  SearchResult *current = self->results[self->cur_idx++];
+  SearchResult_Override(r, current);    // Copy result data to output
+  return RS_RESULT_OK;
```
Itzikvaknin, line 1525: you can free `current` here, consider using `array_pop`

src/result_processor.c
```diff
@@ -1426,3 +1427,107 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+ *
+ *  The RPDepleter result processor offloads the task of consuming all results from its upstream processor
+ *  into a background thread, storing them in an internal array. While the background thread is running,
+ *  calls to Next() return RS_RESULT_DEPLETING. Once depleting is complete, Next() yields results one by one
+ *  from the internal array, and finally returns the last return code from the upstream.
+ *******************************************************************************************************************/
+typedef struct {
+  ResultProcessor base;                // Base result processor struct
+  arrayof(SearchResult *) results;     // Array of pointers to SearchResult, filled by the depleting thread
+  atomic_int done_depleting;           // Set to 1 when depleting is finished
```
Itzikvaknin: `atomic_bool` ?

src/result_processor.c
```diff
@@ -1426,3 +1427,107 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+ * Background thread function: consumes all results from upstream and stores them in the results array.
+ * Signals completion by setting done_depleting to 1.
+ */
+static void RPDepleter_Deplete(void *arg) {
+  RPDepleter *self = (RPDepleter *)arg;
+  RPStatus rc;
+
+  // Deplete the pipeline into the `self->results` array.
+  SearchResult *r = rm_calloc(1, sizeof(*r));
+  while ((rc = self->base.upstream->Next(self->base.upstream, r)) == RS_RESULT_OK) {
```
Itzikvaknin, line 1494: Do we have any guidance on which result processors are supposed to handle timeouts?

src/result_processor.c
```diff
@@ -1426,3 +1427,107 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+  atomic_int done_depleting;           // Set to 1 when depleting is finished
+  uint cur_idx;                        // Current index for yielding results
+  RPStatus last_rc;                    // Last return code from upstream
+  bool first_call;                     // Whether the first call to Next has been made
+} RPDepleter;
+
+/**
+ * Destructor
+ */
+static void RPDepleter_Free(ResultProcessor *base) {
```
Itzikvaknin, line 1477: Can this function be called while `RPDepleter_Deplete` is running in a different thread?

src/result_processor.c
```diff
@@ -1426,3 +1427,107 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+  SearchResult *current = self->results[self->cur_idx++];
+  SearchResult_Override(r, current);    // Copy result data to output
+  return RS_RESULT_OK;
+}
+
+/**
+ * Next function for RPDepleter.
+ * Returns `RS_RESULT_DEPLETING` until the BG thread is done depleting.
+ */
+static int RPDepleter_Next_Dispatch(ResultProcessor *base, SearchResult *r) {
```
Itzikvaknin, line 1539:
```suggestion
static int RPDepleter_Next_Dispatch(ResultProcessor *rp, SearchResult *r) {
```

src/result_processor.c
```diff
@@ -1426,3 +1427,107 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+  RPDepleter *self = (RPDepleter *)base;
+  array_free_ex(self->results, srDtor(*(char **)ptr));
+  rm_free(self);
+}
+
+/**
+ * Background thread function: consumes all results from upstream and stores them in the results array.
+ * Signals completion by setting done_depleting to 1.
+ */
+static void RPDepleter_Deplete(void *arg) {
```
Itzikvaknin, line 1488: Do we need to mask `rp->parent->resultLimit = UINT32_MAX;`?
Otherwise the name may be misleading

tests/cpptests/test_cpp_agg.cpp
```diff
@@ -324,3 +324,222 @@ TEST_F(AggTest, AvoidingCompleteResultStructOpt) {
 
   RedisModule_FreeThreadSafeContext(ctx);
 }
+
+extern "C" {
```
Itzikvaknin: Why here and not in`tests/cpptests/test_cpp_resultprocessor.cpp`?

src/result_processor.c
raz-mon: Can't use `array_pop` since I'm taking the next result from the beginning, not the end.
Regarding freeing, right we can, but it will be freed soon so not sure how much of a difference it'll make.
Will consider

src/result_processor.c
raz-mon: Same same.. Some platforms don't like the bool I think so I tend to use int sometimes.
Doesn't really matter for this case

src/result_processor.c
raz-mon: We propagate this decision higher, by returning the last return code after the last result

src/result_processor.c
raz-mon: I don't think so.
This is called upon pipeline folding which is called upon command folding, which is called after the pipeline is done working so that the BG thread is done working

src/module-init/module-init.c
kei-nan: The additional downside is if the user increases the WORKERS then this pool doesn't benefit from that at all.
I think we shouldn't make things more complicated for our users at this point.
That is at least my take.

src/result_processor.c
raz-mon: Why?
This follows the current API and other next calls

src/result_processor.c
kei-nan: But the same one(Ranker/Scorer/Merger) will be used in the coordinator, where it won't be coupled to this result processor.
And there we won't want it to wait.

I recommend you look at how future promise work.
I think we can solve this better.

src/result_processor.c
raz-mon: Well, I think we want to deplete the pipeline as it is, not actually force the depletion of the index - that depends on the pipeline.
For our use-case, we'll have a sorter that will actually deplete the index so we're good on that front.
For instance let's say the pipeline needs `100` results, why should I return more than that from the depleter? It's goal is to deplete a **pipeline**, not an **index**.

I think the name is ok, it's just a different context 😉

tests/cpptests/test_cpp_agg.cpp
raz-mon: Wasn't aware of it, you're right it's a better match, moving

src/result_processor.c
kei-nan: Nice catch, how does the deletion flow works if it is running in another process.
I think there should be a shared context between the two threads which is ref counted.

src/result_processor.c
raz-mon: Why won't we want it there?
It will be there as well

src/result_processor.c
kei-nan: how will the context and timeout be incremented across threads?
It doesn't sound thread safe.

src/result_processor.c
kei-nan: maybe we can store the results in the reverse order.
e,g push_back to add and pop to take it out?

src/module-init/module-init.c
raz-mon: I also like the fact that there will be one config param that controls the query BG work.
Will contemplate a bit more and take a decision. Anyways we can always change this until we release the feature.

src/result_processor.c
raz-mon: See where `AREQ_Free` is called, I don't see a risk with it

src/result_processor.c
kei-nan: there won't be an additional thread to wait for.
The next call will actually bring you back results from the cursor.

src/result_processor.c
kei-nan: what happens if it gets deleted while the thread is still running and getting results from the pipeline?

src/result_processor.c
raz-mon: We currently plan for there to be.
Let's align on that and continue from there...

src/result_processor.c
raz-mon: @kei-nan, see my two comments, I don't see a scenario where this can happen currently. Do you?

src/result_processor.c
raz-mon: @kei-nan This is not a concern of this PR, but of the pipeline building PR.

src/result_processor.c
raz-mon: If we do not append results to the end, but push them in the front of the array, that will imply **many** unnecessary copies.
I don't see a good reason to do that

raz-mon: We can change the sorting order but that's too patchy for this IMO

src/result_processor.c
kei-nan: Yes,
If cursor is used then the cursor can get deleted while this thread is still running.
We also have client disconnects, index gets deleted and a bunch of other data races we need to address here.
You are moving read locks across threads, meaning two threads rely on a single read lock.

In general I think we need this code to be airtight threading wise.
This is the most delicate parts of the code which will be extremely hard to debug afterward.
Much more than logic issues.

src/result_processor.c
kei-nan: I disagree, maybe we should schedule a meeting and discuss this in a wider forum.

src/result_processor.c
kei-nan: I don't think we do.
There is no point for a thread in the coordinator side past the rpnet processing.
Which is before this result processor.
Maybe should talk about this in a wider forum.

raz-mon: This is the current design as depicted in the [design-doc](https://redislabs.atlassian.net/wiki/spaces/DX/pages/5164236959/Hybrid+Search+RRF+and+Linear+Combination+-+Design+Doc).
The idea is to have the sorting of the results of the separate queries from the different shards done concurrently (may be optimized to a special sorter that limits the results returned from a cursor).
Seems you want to redesign

src/result_processor.c
raz-mon: How do the scenarios you mentioned cause issues with the release of the result-processor?
As long as the downstream (new HybridMerger) will handle the downstream correctly - wait for the second thread to finish or kill it - I still don't see an issue here.

kei-nan: I think we discussed this virtually, for now resolving this.

src/result_processor.c
kei-nan: We said the pipeline builder will be responsible for the context the result processors get.
Please ensure a task exists for this part so it won't get lost in all the future tasks we need to do.

src/result_processor.c
kei-nan: okay

src/result_processor.c
kei-nan: Went with condition variable, resolving.

src/result_processor.c
```diff
@@ -1426,3 +1430,159 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+ * Constructs a new RPDepleter processor
+ */
+ResultProcessor *RPDepleter_New(StrongRef sync_ref) {
+  RPDepleter *ret = rm_calloc(1, sizeof(*ret));
+  ret->results = array_new(SearchResult*, 0);
+  ret->base.Next = RPDepleter_Next_Dispatch;
+  ret->base.Free = RPDepleter_Free;
+  ret->base.type = RP_DEPLETER;
+  ret->first_call = true;
+  ret->sync_ref = sync_ref;
```
kei-nan, line 1590: should it increment the ref count here?

src/result_processor.c
```diff
@@ -1426,3 +1430,159 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+ *
+ *  The RPDepleter result processor offloads the task of consuming all results from its upstream processor
+ *  into a background thread, storing them in an internal array. While the background thread is running,
+ *  calls to Next() return RS_RESULT_DEPLETING. Once depleting is complete, Next() yields results one by one
+ *  from the internal array, and finally returns the last return code from the upstream.
+ *******************************************************************************************************************/
+typedef struct {
+  ResultProcessor base;                // Base result processor struct
+  arrayof(SearchResult *) results;     // Array of pointers to SearchResult, filled by the depleting thread
+  atomic_int done_depleting;           // Set to 1 when depleting is finished
```
kei-nan: done_depleting is changed and checked under the lock, don't think it needs to be atomic anymore.

tests/cpptests/test_cpp_resultprocessor.cpp
```diff
@@ -88,3 +91,266 @@ TEST_F(ResultProcessorTest, testProcessorChain) {
   ASSERT_EQ(2, numFreed);
   RLookup_Cleanup(&lk);
 }
+
+TEST_F(ResultProcessorTest, RPDepleter_Basic) {
```
kei-nan: Maybe move the new result processor tests to a different file?
I think it will make things more organized.

src/result_processor.c
raz-mon: No, the ref-count should be incremented by the caller.
Can emphasize that with a comment

src/result_processor.c
raz-mon: For sure, forgot to update that tks

tests/cpptests/test_cpp_resultprocessor.cpp
raz-mon: Separated to a new test file

src/result_processor.c
raz-mon: There indeed exists such a task

src/result_processor.c
```diff
@@ -1426,3 +1430,159 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+  // Check if depleting is already done.
+  // We do this while holding the mutex so that we don't miss a signal.
+  if (self->done_depleting == true) {
+    pthread_mutex_unlock(&sync->mutex);
+    base->Next = RPDepleter_Next_Yield;
+    return base->Next(base, r);
+  }
+
+  // Wait on condition variable for any depleter to signal completion
+  pthread_cond_wait(&sync->cond, &sync->mutex);
```
Itzikvaknin, line 1563: Why should the depleter signal completion? Am I missing something?
The goal is to be able to query whether it's ready — I don't see a need for blocking calls in this.

Itzikvaknin (dismissed):

src/result_processor.c
raz-mon: We want the depleter of one pipeline to wake up when the other pipeline has finished.
Both will block on the same condition variable and wake up on it

src/module-init/module-init.c
raz-mon: Decided not to go with the `workers` thread-pool at this point due to possible problems with adding more jobs to the queue when `WORKERS` was changed to `0` by the main-thread, and since problems can arise when we have only one worker (deadlock). More such problems may arise from this, so we continue with a new thread-pool. If we want we may change this later on prior to releasing.

kei-nan (approved):

PR #6332, opened 2025-06-17
MOD-10153: New APPLY function case()

nafraf:
## Description:
Add a conditional case() function to aggregation pipeline APPLY stage for ternary conditional logic.

1. Current: 
   Query error when trying to APPLY expression on results of a UNION QUERY where results could be present in the TEXT query but not in the VECTOR query.
2. Change: 
  Implement new function `case(condition, if_true, if_false)`
3. Outcome: 

1. MOD-10153

1. ...

#### Mark if applicable

### DIFF
--- a/tests/pytests/test_case.py
+++ b/tests/pytests/test_case.py
@@ -96,6 +96,7 @@ def testInvalidApplyFunction(env):
         'case(invalid_function(@vector_distance), 1, 0)',
         'case(!invalid_function(@vector_distance), 1, 0)',
         'case(!!invalid_function(@vector_distance), 1, 0)',
+        'case(!!!invalid_function(@vector_distance), 1, 0)',
     ]
     for apply_expr in invalid_apply_exprs:
         env.expect(
@@ -307,9 +308,9 @@ def testNestedCaseDepth(env):
         """ Generate a recursive CASE structure with the specified depth in the form: CASE(0, "a", "b") """
         if depth <= 1:
             return f"case(0, 'a{depth}', 'b')"
-        
+
         return f"case(0, 'a{depth}', {generate_case_expr(depth-1)})"
-    
+
     conn = getConnectionByEnv(env)
 
     # Create an index and add a doc
@@ -319,7 +320,7 @@ def testNestedCaseDepth(env):
     # Test deep nesting of case
     depth = 20
     case_expr = generate_case_expr(depth)
-    
+
     res = conn.execute_command(
         'FT.AGGREGATE', 'idx_depth', '*',
         'APPLY', case_expr, 'AS', 'nested_value',


src/aggregate/functions/string.c
```diff
@@ -276,6 +276,14 @@ int func_exists(ExprEval *ctx, RSValue *argv, size_t argc, RSValue *result) {
   return EXPR_EVAL_OK;
 }
 
+int func_case(ExprEval *ctx, RSValue *argv, size_t argc, RSValue *result) {
+  // This function is never directly called for CASE expressions
+  // The actual implementation is in evalFuncCase in expression.c
+  // This is just a placeholder for function registration
```
oshadmi, line 282: If signature would have `f` we could have moved the body of `evalFuncCase` to here and call it from `evalFunc`?, but due to the special nature of `case` that must not evaluate all args, we must have this special handling that requires `f` to properly process its args?

nafraf: Yes.
We need first to evaluate the condition to avoid to evaluate the unneeded, NULLABLE, expression.

Another option could be to evaluate only the needed expressions in `evalFuncCase` (condition, selected branch) and pass them as arguments to func_case. But the code of func_case would be only to return the evaluated branch based on `condition`

tests/pytests/test_case.py
```diff
@@ -0,0 +1,322 @@
...
+def _prepare_index(env, idx, dim=4):
+    conn = getConnectionByEnv(env)
+    env.expect('FT.CREATE', idx, 'SCHEMA',
+               'v', 'VECTOR', 'FLAT', '6', 'DIM', dim, 'DISTANCE_METRIC', 'L2', 'TYPE', 'FLOAT32',
+               't', 'TEXT').ok()
+
+    conn.execute_command('HSET', 'doc1', 'v', create_np_array_typed([1.1] * dim).tobytes(), 't', 'pizza')
+    conn.execute_command('HSET', 'doc2', 'v', create_np_array_typed([1.4] * dim).tobytes())
+    conn.execute_command('HSET', 'doc3', 't', 'pizzas')
+    conn.execute_command('HSET', 'doc4', 't', 'beer')
```
oshadmi: What about a doc which doesn't have neither one of the fields?
(will require a nested case?)

tests/pytests/test_case.py
oshadmi, line 1: Can add unit tests for `case`?

src/aggregate/expr/expression.c
```diff
@@ -20,18 +20,49 @@ static void setReferenceValue(RSValue *dst, RSValue *src) {
 }
 
 extern int func_exists(ExprEval *ctx, RSValue *argv, size_t argc, RSValue *result);
+extern int func_case(ExprEval *ctx, RSValue *argv, size_t argc, RSValue *result);
+
+static int evalFuncCase(ExprEval *eval, const RSFunctionExpr *f, RSValue *result) {
+  // Evaluate the condition
+  RSValue condVal = RSVALUE_STATIC;
+  int rc = evalInternal(eval, f->args->args[0], &condVal);
+  if (rc != EXPR_EVAL_OK) {
```
oshadmi, line 29: Couldn't the condition be evaluated also to NULL, which should then return the `else` value? 
(`evalInternal` of a missing property could return `EXPR_EVAL_NULL`)

oshadmi: Is there a test with null/missing property in the condition?

tests/pytests/test_case.py
nafraf: Tests added, nested case is not required because the docs always contains `@__score`

tests/pytests/test_case.py
nafraf: Unit tests added.

src/aggregate/expr/expression.c
nafraf: Tests for NULL and missing properties were added.

There is a weird behavior for missing fields but it is not related to the `case`, it also occurs on `master`. If I run FT.AGGREGATE like the following I get error message saying that `@vector_distance` was not found, which is expected.

```sh
'FT.AGGREGATE', 'idx', '@v:[VECTOR_RANGE .7 $vector]=>{$YIELD_DISTANCE_AS: vector_distance} | @t:(pizza)',
        'ADDSCORES', 'SCORER', 'BM25STD',
        'PARAMS', '2', 'vector', create_np_array_typed([1.2] * dim).tobytes(),
        'LOAD', '3', '@vector_distance', '@__key', 't',
        'APPLY', '@vector_distance', 'AS', 'has_vector_distance',
        'SORTBY', '2', '@__key', 'ASC',
        'DIALECT', '2')
```
`Could not find the value for a parameter name, consider using EXISTS if applicable for vector_distance`

But if I remove the `SORTBY`, I get only the 2 results which contains the `@vector_distance`, but the count of results is wrong:
```sh
[1, 
['vector_distance', '0.0400000177324', '__score', '1.05949604453', '__key', 'doc1', 't', 'pizza', 'has_vector_distance', '0.0400000177324'], 
['vector_distance', '0.159999892116', '__score', '0', '__key', 'doc2', 'has_vector_distance', '0.159999892116']
]
```

nafraf: We should report the same error in both cases, right?

oshadmi: This is unrelated to this PR supporting `APPLY CASE`, right?
So let's create a separate issue (Query without SORTBY returns a "Could not find the value" error  and does not return an error with SORTBY?)

Regarding the AGGREGATE count, it might be inaccurate (try `WITHCOUNT`)

tests/cpptests/test_cpp_expr.cpp
```diff
@@ -401,3 +401,241 @@ TEST_F(ExprTest, testPropertyFetch) {
...
+  // Test with complex expressions returning different types
+  ctx.assign("case(1, 3.14 * 2, 'pi_doubled')");
+  ASSERT_TRUE(ctx) << ctx.error();
+  ASSERT_EQ(EXPR_EVAL_OK, ctx.eval());
+  res = RSValue_Dereference(&ctx.result());
+  ASSERT_EQ(RSValue_Number, res->t);
+  ASSERT_FLOAT_EQ(6.28, res->numval);
+}
+
+TEST_F(ExprTest, testEvalFuncCaseConditionEvaluationError) {
```
oshadmi: Why is the name of the test has Error? Isn't it successful?

What about testing a condition which is a `@missing_field`?
How is it being evaluated - since it is NULL it should return the `else` value, right?

tests/cpptests/test_cpp_expr.cpp
```diff
@@ -401,3 +401,241 @@ TEST_F(ExprTest, testPropertyFetch) {
...
+  ctx.lookup = &lk;
+  ctx.srcrow = &rr;
+
+  // Test that only the selected branch is evaluated
+  // When condition is true, only the true branch should be evaluated
+  ASSERT_EQ(EXPR_EVAL_OK, ctx.bindLookupKeys());
+  ASSERT_EQ(EXPR_EVAL_OK, ctx.eval());
+  auto res = RSValue_Dereference(&ctx.result());
+  ASSERT_EQ(RSValue_Number, res->t);
+  ASSERT_EQ(15, res->numval);  // @foo + 10 = 5 + 10 = 15
```
oshadmi: What about returning a `@missing_field` in one of the returned values?
Shouldn't it error with "Could not find the value"

tests/pytests/test_case.py
```diff
@@ -0,0 +1,375 @@
...
+        'PARAMS', '2', 'vector', create_np_array_typed([1.2] * dim).tobytes(),
+        'LOAD', '3', '@vector_distance', '@__key', 't',
+        'SORTBY', '2', '@__key', 'ASC',
+        'DIALECT', '2')
+    nresults = 5
+    env.assertEqual(res[0], nresults)
+    v_dist = [float(row[row.index('vector_distance') + 1] if 'vector_distance' in row else '0') for row in res[1:]]
+    t_score = [float(row[row.index('__score') + 1] if '__score' in row else '0') for row in res[1:]]
+
+    # Test APPLY with case function
```
oshadmi, line 34: This is just prepare for test?
Isn't the test below after
```# Test APPLY with case function and linear combination of vector distance and text score```

File: tests/cpptests/test_cpp_expr.cpp [hunk 7]
```diff
@@ -401,3 +401,241 @@ TEST_F(ExprTest, testPropertyFetch) {
...
+
+  ctx.assign("case(0, 42, 'string_result')");
+  ASSERT_TRUE(ctx) << ctx.error();
+  ASSERT_EQ(EXPR_EVAL_OK, ctx.eval());
+  res = RSValue_Dereference(&ctx.result());
+  ASSERT_EQ(RSValue_String, res->t);
+  ASSERT_STREQ("string_result", res->strval.str);
+
+  // Test with complex expressions returning different types
+  ctx.assign("case(1, 3.14 * 2, 'pi_doubled')");
```
oshadmi, line 608: What about different types in cond, e.g., ```<string> != <numeric>```

What about returning a bool, e.g., 
```
case(0, 1==1, 2!=2)
case(0, 1!=1, exists(@mising))
```

tests/pytests/test_case.py
```diff
@@ -0,0 +1,375 @@
...
+        env.expect(
+            'FT.AGGREGATE', 'idx', '@t:(pizza|beer)',
+            'LOAD', '1', '@t',
+            'APPLY', apply_expr, 'AS', 'final_score',
+            'DIALECT', '2').error().contains("Error converting string")
+
+    # Invalid function name in APPLY clause
+    invalid_apply_exprs = [
+        'case(invalid_function(@vector_distance), 1, 0)',
+        # 'case(!invalid_function(@vector_distance), 1, 0)',
```
oshadmi: Can uncomment this now since it is fixed

src/aggregate/expr/expression.c
```diff
@@ -20,18 +20,49 @@ static void setReferenceValue(RSValue *dst, RSValue *src) {
 }
 
 extern int func_exists(ExprEval *ctx, RSValue *argv, size_t argc, RSValue *result);
+extern int func_case(ExprEval *ctx, RSValue *argv, size_t argc, RSValue *result);
+
+static int evalFuncCase(ExprEval *eval, const RSFunctionExpr *f, RSValue *result) {
+  // Evaluate the condition
+  RSValue condVal = RSVALUE_STATIC;
+  int rc = evalInternal(eval, f->args->args[0], &condVal);
```
raz-mon, line 28: Is there some validation for the existence of this argument?
If not let's add some assertion/validation, since we will later crash

src/aggregate/expr/expression.c
```diff
@@ -20,18 +20,49 @@ static void setReferenceValue(RSValue *dst, RSValue *src) {
...
+  if (rc != EXPR_EVAL_OK) {
+    RSValue_Clear(&condVal);
+    return rc;
+  }
+
+  // Determine which branch to evaluate based on the condition
+  RSValue *expr = RSValue_Dereference(&condVal);
```
raz-mon: Not needed, since it occurs inside `BoolTest` as well

src/aggregate/expr/expression.c
```diff
@@ -20,18 +20,49 @@ static void setReferenceValue(RSValue *dst, RSValue *src) {
...
+  int condition = RSValue_BoolTest(expr);
+  RSValue_Clear(&condVal);
+
+  // Evaluate only the branch we need
+  int branchIndex = condition ? 1 : 2;
+  rc = evalInternal(eval, f->args->args[branchIndex], result);
```
raz-mon, line 40: Also, worth asserting that these args exist?

raz-mon: Or just add an assertion in `evalInternal`.. Probably that's best.

src/aggregate/expr/expression.c
```diff
@@ -20,18 +20,49 @@ static void setReferenceValue(RSValue *dst, RSValue *src) {
...
+}
 
 static int evalFunc(ExprEval *eval, const RSFunctionExpr *f, RSValue *result) {
   int rc = EXPR_EVAL_ERR;
 
+  // Special handling for func_case
+  if (f->Call == func_case) {
+    return evalFuncCase(eval, f, result);
+  }
+
```
raz-mon: Why do we need special handling?
Please add to the comment as well

src/aggregate/functions/string.c
raz-mon, line 282: Why is this needed?

tests/pytests/test_case.py
```diff
@@ -0,0 +1,375 @@
...
+
+    # Test basic case function with exists condition
+    res = conn.execute_command(
+        'FT.AGGREGATE', 'idx', '*',
+        'LOAD', '2', '@t', '@__key',
+        'APPLY', 'case(exists(@t), 1, 0)', 'AS', 'has_text',
+        'SORTBY', '2', '@__key', 'ASC',
+        'DIALECT', '2')
+
+    env.assertEqual(res[0], 5)  # 4 documents total
```
oshadmi:
```suggestion
    env.assertEqual(res[0], 5)  # 5 documents total
```

tests/cpptests/test_cpp_expr.cpp
```diff
@@ -401,3 +401,241 @@ TEST_F(ExprTest, testPropertyFetch) {
...
+
+TEST_F(ExprTest, testEvalFuncCase) {
+  TEvalCtx ctx;
+
+  // Basic case function tests - condition evaluates to true
+  ctx.assign("case(1, 42, 99)");
+  ASSERT_TRUE(ctx) << ctx.error();
+  ASSERT_EQ(EXPR_EVAL_OK, ctx.eval());
+  auto res = RSValue_Dereference(&ctx.result());
+  ASSERT_EQ(RSValue_Number, res->t);
```
raz-mon: Consider creating a macro/function for this - repeats heavily

raz-mon (commented):
VERY NICE 🔥 
See a few comments

tests/pytests/test_case.py
```diff
@@ -0,0 +1,375 @@
...
+        'GROUPBY', '1', '@product_class',
+        'REDUCE', 'COUNT', '0', 'AS', 'count',
+        'SORTBY', '2', '@product_class', 'ASC',
+        'DIALECT', '2')
+
+    # Expected results:
+    # - budget_electronics: prod1
+    # - available_items: prod2, prod3
+    # - other: prod4
+    # Total groups should be 3
```
oshadmi: For complex conditions consider not using groupby
Expected seems to be 
prod1 first case true ==> budget_electronics
prod2 first case false, second case false ==> other
prod3 first case false, second case true ==> available_items
prod4 first case false, second case true ==> available_items

src/aggregate/expr/expression.c
oshadmi: `func_case` is registered with exactly 3 args and it is being checked, so no need to validate.
Maybe we can debug assert that `f->Call == func_case`

tests/cpptests/test_cpp_expr.cpp
oshadmi: ➕ 1️⃣

tests/cpptests/test_cpp_expr.cpp
nafraf: Unit tests added: 
- `testEvalFuncCaseWithDifferentTypeComparison`
- `ctx.assign("case(1, exists(@mising), 0)");`
- `ctx.assign("case(0, 1==1, 2!=2)");`

src/aggregate/functions/string.c
nafraf: To register `case()` function, and it helps to validate the number of arguments of the function.

tests/pytests/test_case.py
nafraf: The section above obtains the vector distances and scores without `APPLY`.
The list `apply_expr_and_expected_results[]` contains the different expressions and expected results to test.
The `for`, execute the tests.

tests/pytests/test_case.py
nafraf: Done.

src/aggregate/expr/expression.c
nafraf: We don't need to validate the args, the register of `case()` function give us that "for free".
Unit tests were added to confirm the argument validation.

tests/cpptests/test_cpp_expr.cpp
nafraf: Test renamed.
More tests were added to check NULL behavior.

tests/cpptests/test_cpp_expr.cpp
nafraf: Yes. It should return an error.
Unit tests now check that it returns the error, flow tests were also added to confirm the error message.

tests/pytests/test_case.py
nafraf: Done.

src/aggregate/expr/expression.c
nafraf: Thanks!. Removed.

src/aggregate/expr/expression.c
nafraf: Comments improved.

tests/cpptests/test_cpp_expr.cpp
nafraf: OK. Done. Thanks.

src/aggregate/expr/expression.c
nafraf: Yes. It's unrelated to this PR.

tests/pytests/test_case.py
nafraf: OK. Tests modified.

src/aggregate/expr/expression.c
oshadmi: Wanted to create an issue but it behaves the same?
Using 2.10.15
Using these vector bytes
```
import numpy
>>> numpy.array(1.2).tobytes()
b'333333\xf3?'
>>> numpy.array(1.1).tobytes()
b'\x9a\x99\x99\x99\x99\x99\xf1?'
>>> numpy.array(1.4).tobytes()
b'ffffff\xf6?'
```

And this index

```
127.0.0.1:6379> FT.CREATE idx SCHEMA v VECTOR FLAT 6 DIM 4 DISTANCE_METRIC L2 TYPE FLOAT32 t TEXT x NUMERIC
OK
127.0.0.1:6379> HSET doc1 x 1 v b'\x9a\x99\x99\x99\x99\x99\xf1?' t pizza
(integer) 3
127.0.0.1:6379> HSET doc2 x 1 v b'ffffff\xf6?'
(integer) 2
127.0.0.1:6379> HSET doc3 x 1 t pizzas
(integer) 2
127.0.0.1:6379> HSET doc4 x 1 t beer
(integer) 2
127.0.0.1:6379> HSET doc5 x 1 
(integer) 1
```
With and without SORTBY behaves the same
```
127.0.0.1:6379> FT.AGGREGATE idx '@v:[VECTOR_RANGE .7 $vector]=>{$YIELD_DISTANCE_AS: vector_distance} | @t:(pizza)' ADDSCORES SCORER BM25STD PARAMS 2 vector b'333333\xf3?' LOAD 3 @vector_distance @__key t APPLY @vector_distance AS has_vector_distance SORTBY 2 @__key ASC DIALECT 2
(error) vector_distance: has no value, consider using EXISTS if applicable

127.0.0.1:6379> FT.AGGREGATE idx '@v:[VECTOR_RANGE .7 $vector]=>{$YIELD_DISTANCE_AS: vector_distance} | @t:(pizza)' ADDSCORES SCORER BM25STD PARAMS 2 vector b'333333\xf3?' LOAD 3 @vector_distance @__key t APPLY @vector_distance AS has_vector_distance DIALECT 2
(error) vector_distance: has no value, consider using EXISTS if applicable
127.0.0.1:6379> 



```

tests/pytests/test_case.py
```diff
@@ -0,0 +1,395 @@
...
+        'DIALECT', '2')
+
+    # Expected: 3 groups - low (1 doc), medium (2 docs), high (1 doc)
+    env.assertEqual(res[0], 3)
+    categories = {row[1]: int(row[3]) for row in res[1:]}
+    env.assertEqual(categories['low'], 1)
+    env.assertEqual(categories['medium'], 2)
+    env.assertEqual(categories['high'], 1)
+
+def testNestedCaseAndGroup(env):
```
oshadmi, line 192: What about a test with much deeper nestedness?
```
def generate_case(depth):
    """
    Generate a recursive CASE structure with the specified depth
    in the form: CASE(0, "a", "b")
    """
    if depth <= 0:
        return '"b"'
    
    if depth == 1:
        return 'CASE(0, "a", "b")'
    
    return f'CASE(0, "a", {generate_case(depth-1)})'
```

And then test with `generate_case(50)` which should evaluate to `"b"`

oshadmi: Added test `testNestedCaseDepth`

oshadmi (approved):

PR #6333, opened 2025-06-17
[2.10] Coverage Report Without `readies` - Phase 2 - [MOD-6711]

GuyAv46:
# Description
Backport of #5909 to `2.10`.


.github/workflows/event-nightly.yml
```diff
@@ -25,16 +25,20 @@ jobs:
...
+      test-config: QUICK=1
+      get-redis: unstable
 
   sanitize:
     secrets: inherit
     uses: ./.github/workflows/task-test.yml
     with:
       # Consider running against unstable Redis (will require dynamic linking to libstdc++
-      get-redis: ${{ needs.get-latest-redis-tag.outputs.tag }}
+      get-redis: unstable
```
alonre24, line 41: The changes in nightly don't really matter since it will not run on non master branches, right?

GuyAv46: right. My GHA yaml analyzer was screaming that this input is not defined, so I removed it. But it has no effect

PR #6336, opened 2025-06-17
[MOD-10177] Rust benchmark for `numeric` encoder and decoder

chesedo:

This creates a Rust benchmark runner for the C numeric encoder and decoder only. This will serve as a base for more encoder / decoder benchmarks.

#### Mark if applicable


src/redisearch_rs/inverted_index_bencher/src/main.rs
```diff
@@ -0,0 +1,18 @@
...
+*/
+
+fn main() {
+    compute_and_report_memory_usage();
+}
+
+fn compute_and_report_memory_usage() {
+    // This function is a placeholder for the original logic. The actual implementation would go
+    // here once the inverted index is done, but for now, we will just print a message.
+    println!("This function would compute and report memory usage.");
```
chesedo: What do we want to compute here if we even want to compute the memory usage?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -136,6 +138,82 @@ pub struct RSIndexResult {
...
+impl RSIndexResult {
+    pub fn numeric(num: f64) -> Self {
+        Self {
+            doc_id: 0,
+            dmd: std::ptr::null(),
+            field_mask: 0,
+            freq: 0,
+            offsets_sz: 0,
+            data: RSIndexResultData {
+                num: ManuallyDrop::new(RSNumericRecord(num)),
```
gdesmott, line 154: What's the rationale for using `ManuallyDrop`  here?

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -47,6 +47,19 @@ impl Buffer {
...
+    pub fn from_array<const N: usize>(a: [u8; N]) -> Buffer {
+        let ptr = Box::into_raw(Box::new(a)).cast::<u8>();
+        let ptr = NonNull::new(ptr).unwrap();
+
+        // SAFETY: We are creating a new Buffer with a pointer to an array of N bytes. Since this is
+        // a static array, we know that the memory is valid, initialized, and won't be dropped. We
+        // also know the length and capacity are both N and that this stays within the memory region
+        // of the pointer. All this this together ensures the safety of `Buffer::new` is satisfied.
+        unsafe { Buffer::new(ptr, N, N) }
+    }
```
LukeMathWalker: This is problematic: when the buffer grows, it'll invoke the `Buffer_Grow` method and we'll end up reallocating via C something that was allocated from Rust.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: It's for the `union RSIndexResultData`.  It's basically to prevent a Rust `Drop` call on any union variant when the type goes out of scope.

https://github.com/RediSearch/RediSearch/blob/341f4d4e532782d6559ef5f8256dfb59bf6d0d17/src/redisearch_rs/inverted_index/src/lib.rs#L97-L101

image

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -70,6 +83,11 @@ impl Buffer {
         self.0.offset == 0
     }
 
+    /// Resets the buffer's length to zero, effectively clearing it.
+    pub fn reset(&mut self) {
```
LukeMathWalker: Let's call it clear, given https://doc.rust-lang.org/std/vec/struct.Vec.html#method.clear

src/redisearch_rs/buffer/src/lib.rs
chesedo: The grow actually fails. I was thinking of adding a comment about that or this method needs to be in some smarter place (it's going to be used a lot even outside of tests)

src/redisearch_rs/inverted_index_bencher/build.rs
```diff
@@ -0,0 +1,146 @@
...
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use std::env;
+use std::fs::read_dir;
+use std::path::{Path, PathBuf};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Always attempt to link static libraries if they exist, independent of bindgen
+    try_link_static_libraries();
```
LukeMathWalker: We should fail if we can't link.

src/redisearch_rs/inverted_index_bencher/build.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    // Generate C bindings - fail build if this doesn't work
+    generate_c_bindings()?;
```
LukeMathWalker, line 19: Why do we want to generate C bindings here? We could add them to the existing FFI crate, since it's going to be a dependency in any case.

src/redisearch_rs/inverted_index_bencher/src/main.rs
LukeMathWalker: In this case, I don't think we care, since we are porting the encoding scheme 1:1.
This was interesting for `TrieMap` since it wasn't a 1:1 port and it wasn't easy to reason, in the abstract, about the size changes.

src/redisearch_rs/inverted_index_bencher/src/ffi.rs
```diff
@@ -0,0 +1,131 @@
...
+    use super::*;
+
+    #[test]
+    fn test_encode_numeric() {
+        let mut buffer = Buffer::from_array([0; 66]);
+        // Test cases for all the different numeric encodings. These cases can be moved to the Rust
+        // implementation tests verbatim.
+        let tests = [
+            // Tiny int tests
+            (2.0, 0, vec![0b010_00_000]),
```
LukeMathWalker, line 87: Any reason for the peculiar grouping strategy used here for bits? 👀

src/redisearch_rs/inverted_index_bencher/src/bencher.rs
```diff
@@ -0,0 +1,140 @@
...
+            group.finish();
+        }
+    }
+}
+
+fn numeric_c_encode<M: Measurement>(group: &mut BenchmarkGroup<'_, M>, values: &[f64]) {
+    let mut buffer = Buffer::from_array([0; 16]);
+
+    group.bench_function("C", |b| {
+        b.iter(|| {
```
LukeMathWalker: Use `iter_batched_ref` with a fresh buffer for every iteration. It should be big enough to avoid memory growth.

src/redisearch_rs/inverted_index_bencher/src/ffi.rs
chesedo: The first byte is the header byte. So they all have the same grouping as their respective header layout

https://github.com/RediSearch/RediSearch/blob/341f4d4e532782d6559ef5f8256dfb59bf6d0d17/src/inverted_index.c#L262-L267

src/redisearch_rs/buffer/src/lib.rs
LukeMathWalker: Can we have a constructor that uses the C allocator? And then we simply write values into it and the machinery takes care of using the right alloc/realloc.

chesedo: That should be possible. I'm just not that familiar with that path. So will need to look into it first.

src/redisearch_rs/inverted_index_bencher/build.rs
chesedo: Technically no, we don't need to generate them here. But we only need the C bindings for `inverted_index.h` and only for a temporary duration until the C `inverted_index.h` is no more. So it feels like it would only pollute the FFI crate if we add them there instead.

src/redisearch_rs/inverted_index_bencher/src/bencher.rs
chesedo: I've tried giving the buffer 16kB of memory. That is not enough for the float benches (each float takes 11 bytes) and already takes 24Gb of memory on my system

src/redisearch_rs/inverted_index_bencher/src/main.rs
chesedo: Great, that means we'll only use `main.rs` to compare the C and Rust outputs right?

src/redisearch_rs/inverted_index_bencher/src/bencher.rs
LukeMathWalker: That sounds like a memory leak 👀

src/redisearch_rs/inverted_index_bencher/src/main.rs
LukeMathWalker: I don't think we need a main at all, we have tests for correctness.

src/redisearch_rs/inverted_index_bencher/src/bencher.rs
chesedo: Ofcourse, I never reclaim the memory from this pointer

https://github.com/RediSearch/RediSearch/blob/afe6347f46f7f42f1f59dda8bb4dc659a419d1cc/src/redisearch_rs/buffer/src/lib.rs#L52-L61

src/redisearch_rs/buffer/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6336/commits/99c95f4d699791e12b412193918ecd450d23e2e9

src/redisearch_rs/inverted_index_bencher/src/bencher.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6336/commits/2ea2e13fb39988415e915961aa1d48f4116aa2a7

src/redisearch_rs/inverted_index_bencher/src/main.rs
chesedo: Yeah, removed in https://github.com/RediSearch/RediSearch/pull/6336/commits/c38afd94bd05234019ae9722e8635fe26f99cf6e

src/redisearch_rs/inverted_index_bencher/src/ffi.rs
```diff
@@ -22,12 +28,52 @@ mod bindings {
     include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 }
 
+/// An extra wrapper around the ['Buffer`] which allows the C code to grow it correctly and for its
+/// memory to be cleaned up when it goes out of scope.
+pub struct BufferWrapper(Buffer);
```
LukeMathWalker: I would suggest calling it `TestBuffer` to clarify its purpose.

src/redisearch_rs/inverted_index_bencher/src/bencher.rs
```diff
@@ -108,17 +108,19 @@ impl NumericBencher {
...
+            || BufferWrapper::new(),
+            |mut buffer| {
+                for &value in values {
+                    let mut record = inverted_index::RSIndexResult::numeric(value);
+                    let grew_size = encode_numeric(&mut buffer, &mut record, 684);
+
+                    black_box(grew_size);
+                }
+            },
+            BatchSize::LargeInput,
```
LukeMathWalker: Your input is small, I'd say.

src/redisearch_rs/inverted_index_bencher/src/ffi.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+/// An extra wrapper around the ['Buffer`] which allows the C code to grow it correctly and for its
+/// memory to be cleaned up when it goes out of scope.
+pub struct TestBuffer(Buffer);
+
+impl TestBuffer {
+    pub fn new() -> Self {
+        let layout = Layout::array::<u8>(0).unwrap();
+        let data = unsafe { alloc(layout) };
```
gdesmott, line 41: This  will break when we use `buffer.c` functions, like `Buffer_Grow`.
I think we need to use `allocator.rs` here instead, or `Buffer_Init` ?

src/redisearch_rs/inverted_index_bencher/build.rs
```diff
@@ -0,0 +1,137 @@
...
+            return rerun_if_c_changes(&path);
+        } else if let Some(extension) = path.extension() {
+            if extension == "c" || extension == "h" {
+                println!("cargo:rerun-if-changed={}", path.display());
+            }
+        }
+    }
+
+    Ok(())
+}
```
BenGoldberger, line 137: maybe we want those function in some util?
I see they are implemented in different places.

src/redisearch_rs/inverted_index_bencher/src/ffi.rs
chesedo: I've replaced this with a `with_capacity` function in https://github.com/RediSearch/RediSearch/pull/6336/commits/f8d0076c4f2c7ac3b0ed905333d540d2634a28c8. We should rather use it with a capacity big enough so that `Buffer_Grow` never needs to be called by the benchmarks.

src/redisearch_rs/inverted_index_bencher/build.rs
chesedo: I've been thinking the same while working on this. However I see we are still finding ways to improve these with each copy. So I would rather want us to improve it a few more times before moving it to a util location where no one is going to improve it.

Once it is in the util location we can come and update all these to use the util instead.

BenGoldberger (approved):

PR #6341, opened 2025-06-18
chore: make preflight

JonasKruckenberg:

This PR adds a simple `preflight` rule to the makefile to allow contributors to run as many
of the CI checks locally as possible.

#### Mark if applicable


build.sh
```diff
@@ -88,6 +89,17 @@ parse_arguments() {
       QUICK=*)
         QUICK="${arg#*=}"
         ;;
+      LINT|lint)
+        LINT=1
+        ;;
+      PREFLIGHT|preflight)
+        LINT=1
+        COV=1
+        SAN=1
```
GuyAv46, line 98: These parameters indicate how to run the tests and how to build.
1. They are not compatible with each other
2. Setting any of them means you don't run tests in "normal" mode

JonasKruckenberg: ah yeah of course, I think leaving the `COV` param off is fine, it doesn't add much value anyways. But running with sanitizers would be nice I think. Do the C/C++ tests execute differently based on whether sanitizers are enabled or not? (except for being slower)

GuyAv46: Currently, the `build.sh` script is limited in the test types it can run in a single invocation, and there is no way to bundle all the jobs we run on the CI in a single execution.
I think we can invoke it in a loop when we refactor the Makefile, then we can run each case sequentially (CC @BenGoldberger)

GuyAv46: And yes, we build the C++ unit tests differently on sanitizer

JonasKruckenberg: hmm okay, this was a nice idea but maybe not the right approach. thanks all for checking!

PR #6346, opened 2025-06-18
[MOD-10235] Polish varint implementation

LukeMathWalker:

This PR refactors the Rust varint implementation. There are no functional changes.
In particular:
- We use a trait-based design, with macros to avoid code duplication.
- We remove the direct dependency on `FieldMask`, and therefore on the `ffi` crate, by providing an encoding/decoding implementation for both `u64` and `u128`. We can thus stop mirroring in the varint test suite the platform-specific knobs that are used, on the C side, to pick the correct size of `FieldMask`. 
- More extensive documentation all around.
- More test cases, including errors and property-based tests. We now have 100% test coverage for varint encoding/decoding.
- Rename `encode_decode` to `varint`, since we want to have small self-contained crates to minimize compile times. The existing `varint` crate is renamed to `varint_ffi`.

### Benchmarks

Benchmarks have been refactored to measure performance based on the expected encoded size, rather than grouping everything together.
You can compare performance before/after the refactoring by executing benches for the first commit on this branch and then re-running them for the latest commit on this branch. No significant changes were measured on my local machine.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/varint/tests/varint.rs
+++ b/src/redisearch_rs/varint/tests/varint.rs
@@ -7,7 +7,7 @@
  * GNU Affero General Public License v3 (AGPLv3).
 */
 
-use std::{io::Cursor, u128};
+use std::io::Cursor;
 
 use varint::*;
 


src/redisearch_rs/encode_decode/src/varint/mod.rs
```diff
@@ -7,343 +7,197 @@
...
+//! 00 00 00 00
+//! ```
+//!
+//! When varint-encoded, it requires two bytes:
+//!
+//! ```text
+//! the first continuation bit
+//! ∨
+//! 10 00 00 00
+//! 00 00 00 01
```
chesedo: It might be worth mentioning that this uses little-endian encoding. Or to be explicit that the byte blocks are backwards.

src/redisearch_rs/encode_decode/src/varint/vector_writer.rs
```diff
@@ -48,7 +48,7 @@ impl VectorWriter {
         // If the value we're trying to encode is smaller than the last value,
         // we wrap around rather than underflowing.
         let diff = value.wrapping_sub(self.last_value);
-        let size = write(diff, &mut self.buffer)?;
+        let size = u32::write_as_varint(diff, &mut self.buffer)?;
```
chesedo:
```suggestion
        let size = diff.write_as_varint(&mut self.buffer)?;
```
Why aren't we doing the normal method thing for all of these?

src/redisearch_rs/encode_decode/tests/varint.rs
```diff
@@ -0,0 +1,262 @@
+use std::u128;
+
+use encode_decode::varint::*;
+
+#[test]
+fn test_u32() {
+    let values: [u32; 12] = [123456789, 987654321, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+    let expected_lens = [4, 5, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1];
+
+    for (i, value) in values.iter().enumerate() {
```
chesedo:
```suggestion
    for (i, value) in values.into_iter().enumerate() {
```
You don't need to access `values` again so can just take ownership of it. It will also making accessing the `value`s cleaner

src/redisearch_rs/encode_decode/tests/varint.rs
```diff
@@ -0,0 +1,262 @@
...
+///
+/// The decoding process won't panic, but it'll output a non-sensical number.
+fn test_size_confusion() {
+    let mut buf = [0u8; 24];
+    let input = u128::MAX;
+    let n_written_bytes = u128::write_as_varint(input, buf.as_mut_slice()).unwrap();
+    assert_eq!(n_written_bytes, 19);
+
+    let output = u32::read_as_varint(buf.as_slice()).unwrap();
+    assert_eq!(output, 4294967295);
```
chesedo:
```suggestion
    assert_eq!(output, u32::MAX);
```

src/redisearch_rs/encode_decode/tests/varint.rs
```diff
@@ -0,0 +1,262 @@
...
+    let test_cases: &[(u32, Vec<u8>)] = &[
+        (0, vec![0x00]),
+        (1, vec![0x01]),
+        // (127, vec![0x7F]),
+        // (128, vec![0x80, 0x00]),
+        // (129, vec![0x80, 0x01]),
+        // (255, vec![0x80, 0x7F]),
+        // (256, vec![0x81, 0x00]),
+        // (16383, vec![0xFE, 0x7F]),
+        // (16384, vec![0xFF, 0x00]),
```
chesedo: Not worth testing?

chesedo (commented):
Just small nit picky stuff

src/redisearch_rs/encode_decode/src/varint/mod.rs
LukeMathWalker: It is big endian encoding, but I messed up the representation 😛 Fixed in [41270e9](https://github.com/RediSearch/RediSearch/pull/6346/commits/41270e9f87fa29063c51847c97c2eea8b42dc1b2)

src/redisearch_rs/encode_decode/src/varint/vector_writer.rs
LukeMathWalker: Switched in [41270e9](https://github.com/RediSearch/RediSearch/pull/6346/commits/41270e9f87fa29063c51847c97c2eea8b42dc1b2)

src/redisearch_rs/encode_decode/tests/varint.rs
LukeMathWalker: Switched over in [41270e9](https://github.com/RediSearch/RediSearch/pull/6346/commits/41270e9f87fa29063c51847c97c2eea8b42dc1b2)

src/redisearch_rs/encode_decode/tests/varint.rs
LukeMathWalker: Fixed in [41270e9](https://github.com/RediSearch/RediSearch/pull/6346/commits/41270e9f87fa29063c51847c97c2eea8b42dc1b2)

src/redisearch_rs/encode_decode/tests/varint.rs
LukeMathWalker: Leftover while doing some experiments 😅 Fixed in [41270e9](https://github.com/RediSearch/RediSearch/pull/6346/commits/41270e9f87fa29063c51847c97c2eea8b42dc1b2)

src/redisearch_rs/encode_decode/src/varint/mod.rs
chesedo: Fair, I meant that after the encoding it looks more little endian like with the least significant bits appearing first

src/redisearch_rs/varint/tests/varint.rs
```diff
@@ -0,0 +1,278 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use std::{io::Cursor, u128};
```
lerman25: Is the use for u128 necessary?

src/redisearch_rs/varint_bencher/benches/core_operations.rs
```diff
@@ -14,12 +14,12 @@ use std::time::Duration;
 use varint_bencher::VarintBencher;
 
 fn benchmark_core_operations(c: &mut Criterion) {
-    let bencher = VarintBencher::new("Varint".to_owned(), Duration::from_secs(10));
+    let bencher = VarintBencher::new(Duration::from_secs(5));
```
lerman25, line 17: Is the change from 10 to 5 intentional?

lerman25 (commented):
Great, few questions

src/redisearch_rs/varint_bencher/benches/core_operations.rs
LukeMathWalker: Yes, since the bench cases are now smaller we shouldn't need 10 seconds to achieve statistical significance.
I've just triggered a micro-benchmark run in CI to double-check that's the case in the CI environment too.

src/redisearch_rs/varint/tests/varint.rs
LukeMathWalker: It isn't—I removed it b8ed78e20.

src/redisearch_rs/varint_bencher/benches/core_operations.rs
LukeMathWalker: I can confirm that the change doesn't cause issues, given the logs in https://github.com/RediSearch/RediSearch/actions/runs/15872068421/job/44750799229?pr=6346

lerman25 (approved):

PR #6352, opened 2025-06-19
Refactor Optional Iterator Optimized mode - [MOD-9678]

GuyAv46:

Implement the optimized variant of the optional iterator

1. New implementation
2. A few small fixes
3. Implement new tests and benchmark cases
4. Fix SkipToOld benchmark

#### Mark if applicable

### DIFF
--- a/src/iterators/optional_iterator.c
+++ b/src/iterators/optional_iterator.c
@@ -24,7 +24,7 @@ static void OI_Free(QueryIterator *base) {
 
 static size_t OI_NumEstimated(QueryIterator *base) {
   OptionalIterator *oi = (OptionalIterator *)base;
-  return oi->wcii ? oi->wcii->NumEstimated(oi->wcii) : oi->numDocs;
+  return oi->wcii ? oi->wcii->NumEstimated(oi->wcii) : oi->maxDocId;
 }
 
 static void OI_Rewind(QueryIterator *base) {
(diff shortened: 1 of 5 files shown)


src/iterators/optional_iterator.c
```diff
@@ -88,12 +174,16 @@ static IteratorStatus OI_ReadSorted_NotOptimized(QueryIterator *base) {
...
+  RS_ASSERT(it != NULL);
   OptionalIterator *oi = rm_calloc(1, sizeof(*oi));
+  bool optimized = q && q->sctx->spec->rule && q->sctx->spec->rule->index_all;
+  if (optimized) {
+    oi->wcii = IT_V2(NewWildcardIterator_Optimized)(q->sctx);
+  }
   oi->child = it;
-  oi->maxDocId = maxDocId;
   oi->virt = NewVirtualResult(weight, RS_FIELDMASK_ALL);
+  oi->maxDocId = q->docTable->maxDocId;
```
JoanFM, line 187: Here you are assuming q is not NULL and in line 180 (bool optimized = ) you are checking it for non-NULL, what guarantees do there exist here?

GuyAv46: You're right. I'll add an assertion for that.
In practice, we always pass a valid query context, and the support for NULL was added for convenience in unit tests. We always pass a valid (mock) context now for the optional iterator tests and benchmarks so we don't need to support NULL input anymore

JoanFM (approved):

PR #6358, opened 2025-06-22
Performance Improvments to inverted index iterator SkipTo and Read - [MOD-10070]

GuyAv46:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

### DIFF
--- a/src/doc_table.h
+++ b/src/doc_table.h
@@ -147,6 +147,20 @@ bool DocTable_IsDocExpired(DocTable* t, const RSDocumentMetadata* dmd, struct ti
 // missing predicate - one of the fields did expire -> entry is valid in the context of missing
 bool DocTable_VerifyFieldExpirationPredicate(const DocTable *t, t_docId docId, const t_fieldIndex* fieldIndices, size_t fieldCount, enum FieldExpirationPredicate predicate, const struct timespec* expirationPoint);
 
+static inline bool DocTable_CheckFieldExpirationPredicate(const DocTable *t, t_docId docId, t_fieldIndex field, enum FieldExpirationPredicate predicate, const struct timespec* expirationPoint) {
+  if (!t->ttl) return true;
+  return TimeToLiveTable_VerifyDocAndField(t->ttl, docId, field, predicate, expirationPoint);
+}
+static inline bool DocTable_CheckFieldMaskExpirationPredicate(const DocTable *t, t_docId docId, uint32_t fieldMask, enum FieldExpirationPredicate predicate, const struct timespec* expirationPoint, const t_fieldIndex* ftIdToFieldIndex) {
+  if (!t->ttl) return true;
+  return TimeToLiveTable_VerifyDocAndFieldMask(t->ttl, docId, fieldMask, predicate, expirationPoint, ftIdToFieldIndex);
+}
+static inline bool DocTable_CheckWideFieldMaskExpirationPredicate(const DocTable *t, t_docId docId, t_fieldMask fieldMask, enum FieldExpirationPredicate predicate, const struct timespec* expirationPoint, const t_fieldIndex* ftIdToFieldIndex) {
+  if (!t->ttl) return true;
+  return TimeToLiveTable_VerifyDocAndWideFieldMask(t->ttl, docId, fieldMask, predicate, expirationPoint, ftIdToFieldIndex);
+}
+
+
 /** Get the docId of a key if it exists in the table, or 0 if it doesn't */
 t_docId DocTable_GetId(const DocTable *dt, const char *s, size_t n);
 
--- a/src/iterators/inverted_index_iterator.c
+++ b/src/iterators/inverted_index_iterator.c
@@ -46,82 +46,41 @@ size_t InvIndIterator_NumEstimated(QueryIterator *base) {
   return it->idx->numDocs;
 }
 
-#define FIELD_MASK_BIT_COUNT (sizeof(t_fieldMask) * 8)
-
 // Used to determine if the field mask for the given doc id are valid based on their ttl:
 // it->filterCtx.predicate
 // returns true if the we don't have expiration information for the document
 // otherwise will return the same as DocTable_VerifyFieldExpirationPredicate
 // if predicate is default then it means at least one of the fields need to not be expired for us to return true
 // if predicate is missing then it means at least one of the fields needs to be expired for us to return true
-static inline bool VerifyFieldMaskExpirationForDocId(InvIndIterator *it, t_docId docId, t_fieldMask docFieldMask) {
-  // If there isn't a ttl information then the doc fields are valid
-  if (!it->sctx || !it->sctx->spec || !DocTable_HasExpiration(&it->sctx->spec->docs, docId)) {
-    return true;
+static inline bool VerifyFieldMaskExpirationForCurrent(InvIndIterator *it) {
+  // If the field is not a mask, then we can just check the single field
+  if (!it->filterCtx.field.isFieldMask) {
+    return DocTable_CheckFieldExpirationPredicate(
+      &it->sctx->spec->docs,
+      it->base.current->docId,
+      it->filterCtx.field.value.index,
+      it->filterCtx.predicate,
+      &it->sctx->time.current
+    );
+  } else if (it->idx->flags & Index_WideSchema) {
+    return DocTable_CheckWideFieldMaskExpirationPredicate(
+      &it->sctx->spec->docs,
+      it->base.current->docId,
+      it->base.current->fieldMask & it->filterCtx.field.value.mask,
+      it->filterCtx.predicate,
+      &it->sctx->time.current,
+      it->sctx->spec->fieldIdToIndex
+    );
+  } else {
+    return DocTable_CheckFieldMaskExpirationPredicate(
+      &it->sctx->spec->docs,
+      it->base.current->docId,
+      it->base.current->fieldMask & it->filterCtx.field.value.mask,
+      it->filterCtx.predicate,
+      &it->sctx->time.current,
+      it->sctx->spec->fieldIdToIndex
+    );
   }
-
-
-  // TODO: move to constructor
-
-  // doc has expiration information, create a field id array to check for expiration predicate
-  size_t numFieldIndices = 0;
-  // Use a stack allocated array for the field indices, if the field mask is not a single field
-  t_fieldIndex fieldIndicesArray[FIELD_MASK_BIT_COUNT];
-  t_fieldIndex* sortedFieldIndices = fieldIndicesArray;
-  if (it->filterCtx.field.isFieldMask) {
-    const t_fieldMask relevantMask = docFieldMask & it->filterCtx.field.value.mask;
-    numFieldIndices = IndexSpec_TranslateMaskToFieldIndices(it->sctx->spec,
-                                                            relevantMask,
-                                                            fieldIndicesArray);
-  } else if (it->filterCtx.field.value.index != RS_INVALID_FIELD_INDEX) {
-    sortedFieldIndices = &it->filterCtx.field.value.index;
-    numFieldIndices = 1;
-  }
-  return DocTable_VerifyFieldExpirationPredicate(&it->sctx->spec->docs, docId,
-                                                 sortedFieldIndices, numFieldIndices,
-                                                 it->filterCtx.predicate, &it->sctx->time.current);
-}
-
-IteratorStatus InvIndIterator_Read(QueryIterator *base) {
-  InvIndIterator *it = (InvIndIterator *)base;
-  if (base->atEOF) {
-    return ITERATOR_EOF;
-  }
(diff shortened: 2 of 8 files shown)


src/doc_table.h
```diff
@@ -147,6 +147,20 @@ bool DocTable_IsDocExpired(DocTable* t, const RSDocumentMetadata* dmd, struct ti
 // missing predicate - one of the fields did expire -> entry is valid in the context of missing
 bool DocTable_VerifyFieldExpirationPredicate(const DocTable *t, t_docId docId, const t_fieldIndex* fieldIndices, size_t fieldCount, enum FieldExpirationPredicate predicate, const struct timespec* expirationPoint);
 
+static inline bool DocTable_CheckFieldExpirationPredicate(const DocTable *t, t_docId docId, t_fieldIndex field, enum FieldExpirationPredicate predicate, const struct timespec* expirationPoint) {
```
kei-nan, line 151: please add some documentation to these functions, when and for what they are used.
Also what does it mean for them to return true or false.

src/inverted_index/inverted_index.c
```diff
@@ -462,15 +462,18 @@ IndexEncoder InvertedIndex_GetEncoder(IndexFlags flags) {
 size_t InvertedIndex_WriteEntryGeneric(InvertedIndex *idx, IndexEncoder encoder, t_docId docId,
                                        RSIndexResult *entry) {
   size_t sz = 0;
-  bool same_doc = 0;
-  if (idx->lastId && idx->lastId == docId) {
+  bool same_doc = false;
```
kei-nan:
```suggestion
  const bool same_doc = idx->lastId == docId;
```

src/inverted_index/inverted_index.c
```diff
@@ -462,15 +462,18 @@ IndexEncoder InvertedIndex_GetEncoder(IndexFlags flags) {
...
+  RS_ASSERT(docId > 0);
+  if (idx->lastId == docId) {
```
kei-nan:
```suggestion
  if (same_doc) {
```

src/inverted_index/inverted_index.c
```diff
@@ -462,15 +462,18 @@ IndexEncoder InvertedIndex_GetEncoder(IndexFlags flags) {
...
     if (encoder != encodeNumeric) {
       // do not allow the same document to be written to the same index twice.
       // this can happen with duplicate tags for example
       return 0;
     } else {
       // for numeric it is allowed (to support multi values)
-      same_doc = 1;
+      same_doc = true;
```
kei-nan:
```suggestion
```

src/iterators/inverted_index_iterator.c
```diff
@@ -275,22 +402,62 @@ static QueryIterator *NewInvIndIterator(InvertedIndex *idx, RSIndexResult *res,
...
+  base->NumEstimated = InvIndIterator_NumEstimated;
+  base->Free = InvIndIterator_Free;
+  base->Rewind = InvIndIterator_Rewind;
+
+  // Choose the Read and SkipTo methods for best performance
+  skipMulti = skipMulti && (idx->flags & Index_HasMultiValue);
+  bool hasSeeker = it->decoders.seeker != NULL;
+  // TODO: better estimation
+  // check if the specific field/fieldMask has expiration, according to the `filterCtx`
+  bool hasExpiration = sctx && sctx->spec->docs.ttl && sctx->spec->monitorFieldExpiration &&
```
kei-nan: maybe move the logic to a function with some documentation?

src/iterators/inverted_index_iterator.c
```diff
@@ -275,22 +402,62 @@ static QueryIterator *NewInvIndIterator(InvertedIndex *idx, RSIndexResult *res,
...
+                       (!filterCtx->field.isFieldMask && filterCtx->field.value.index != RS_INVALID_FIELD_INDEX);
+
+  // Read function choice:
+  // skip multi     |  no                   |  yes
+  // ------------------------------------------------------------------------
+  // no expiration  |  Read_Default         |  Read_SkipMulti
```
kei-nan, line 442: I am familiar with expiration but less so with multi.
Maybe give some indication in the comment when should we expect the best performance and when we don't.

src/ttl_table.c
```diff
@@ -132,3 +132,170 @@ bool TimeToLiveTable_VerifyDocAndFields(TimeToLiveTable *table, t_docId docId, c
   }
   return false;
 }
+
+bool TimeToLiveTable_VerifyDocAndField(TimeToLiveTable *table, t_docId docId, t_fieldIndex field, enum FieldExpirationPredicate predicate, const struct timespec* expirationPoint) {
```
kei-nan, line 136: why do we need these new functions?
Aren't the existing functions enough?

src/iterators/inverted_index_iterator.c
```diff
@@ -275,22 +402,62 @@ static QueryIterator *NewInvIndIterator(InvertedIndex *idx, RSIndexResult *res,
   it->gcMarker = idx->gcMarker;
   it->decoders = InvertedIndex_GetDecoder(idx->flags);
   it->decoderCtx = *decoderCtx;
-  it->skipMulti = skipMulti;
+  it->skipMulti = skipMulti; // Original request, regardless of what implementation is chosen
```
JoanFM, line 419: add the comment in the .h struct description.

src/iterators/inverted_index_iterator.c
JoanFM: agree, let's have a `pickStrategy`method

src/ttl_table.h
```diff
@@ -31,4 +35,12 @@ bool TimeToLiveTable_HasExpiration(TimeToLiveTable *table, t_docId docId);
 bool TimeToLiveTable_HasDocExpired(TimeToLiveTable *table, t_docId docId, const struct timespec* expirationPoint);
 bool TimeToLiveTable_VerifyDocAndFields(TimeToLiveTable *table, t_docId docId, const t_fieldIndex* sortedFieldIndices, size_t fieldCount, enum FieldExpirationPredicate predicate, const struct timespec* expirationPoint);
 
+bool TimeToLiveTable_VerifyDocAndField(TimeToLiveTable *table, t_docId docId, t_fieldIndex field, enum FieldExpirationPredicate predicate, const struct timespec* expirationPoint);
```
JoanFM: what does it differ from the function just above?

JoanFM: Personal style, I prefer to call `fieldIdx`otherwise it seems to me I am working with some field object

src/ttl_table.c
```diff
@@ -132,3 +132,170 @@ bool TimeToLiveTable_VerifyDocAndFields(TimeToLiveTable *table, t_docId docId, c
...
+  if (fieldWithExpirationCount == 0) {
+    // the document has no fields with expiration times, there exists at least one valid field
+    return true;
+  }
+
+  for (size_t currentFieldIndex = 0; currentFieldIndex < fieldWithExpirationCount; currentFieldIndex++) {
+    FieldExpiration* fieldExpiration = &ttlEntry->fieldExpirations[currentFieldIndex];
+    if (field == fieldExpiration->index) {
+      // the field has an expiration time
+      const bool expired = DidExpire(&fieldExpiration->point, expirationPoint);
```
JoanFM, line 156: I guess you could skip the DidExpire check in some predicate cases?

src/ttl_table.c
```diff
@@ -132,3 +132,170 @@ bool TimeToLiveTable_VerifyDocAndFields(TimeToLiveTable *table, t_docId docId, c
...
+    return true;
+  }
+
+  size_t predicateMisses = 0;
+  size_t currentFieldIndex = 0;
+  int bitIndex;
+  while ((bitIndex = ffs(fieldMask))) {
+    bitIndex--;  // ffs returns 1-based index, we need 0-based
+    fieldMask &= ~(1U << bitIndex);  // Clear the bit we just processed
+    t_fieldIndex fieldIndexToCheck = ftIdToFieldIndex[bitIndex];
```
JoanFM, line 198: can we abstract away part of this code below with the Wide version? Seems like some duplication occurs. And would help have smaller methods and more clarity?

src/inverted_index/inverted_index.c
JoanFM: I like @kei-nan suggestion

src/ttl_table.c
```diff
@@ -132,3 +132,170 @@ bool TimeToLiveTable_VerifyDocAndFields(TimeToLiveTable *table, t_docId docId, c
...
+    return true;
+  }
+
+  for (size_t currentFieldIndex = 0; currentFieldIndex < fieldWithExpirationCount; currentFieldIndex++) {
+    FieldExpiration* fieldExpiration = &ttlEntry->fieldExpirations[currentFieldIndex];
+    if (field == fieldExpiration->index) {
+      // the field has an expiration time
+      const bool expired = DidExpire(&fieldExpiration->point, expirationPoint);
+      if (expired) {
+        return (predicate == FIELD_EXPIRATION_MISSING);
```
JoanFM, line 159: I find this logic with the predicator hard to understand. What do these options imply?

src/ttl_table.c
```diff
@@ -132,3 +132,170 @@ bool TimeToLiveTable_VerifyDocAndFields(TimeToLiveTable *table, t_docId docId, c
...
+      } else {
+        return (predicate == FIELD_EXPIRATION_DEFAULT);
```
JoanFM: I do not understand, if it did not expire, shouldn't it just return TRUE? Like it is valid, why is there any other consideration here?

src/ttl_table.c
```diff
@@ -132,3 +132,170 @@ bool TimeToLiveTable_VerifyDocAndFields(TimeToLiveTable *table, t_docId docId, c
   }
   return false;
 }
+
+bool TimeToLiveTable_VerifyDocAndField(TimeToLiveTable *table, t_docId docId, t_fieldIndex field, enum FieldExpirationPredicate predicate, const struct timespec* expirationPoint) {
+  dictEntry *entry = dictFind(table, (void*)docId);
+  if (!entry) {
```
JoanFM, line 138: Just to have common logic, couldn't we just create a FieldMask with the field bit to 1 and call the generic version?

tests/cpptests/test_cpp_index.cpp
```diff
@@ -224,13 +224,13 @@ TEST_P(IndexFlagsTest, testRWFlags) {
   } else {
     ASSERT_EQ(1, idx->size);
   }
-  ASSERT_EQ(199, idx->lastId);
+  ASSERT_EQ(200, idx->lastId);
```
JoanFM, line 227: why this?

src/iterators/inverted_index_iterator.c
GuyAv46: Now part of `ShouldSkipMulti` documentation

src/ttl_table.c
GuyAv46: Not very simple, as half of the lines are loop operations (`break`, `continue`, `return`) and there are API differences between 32-bit and 128-bit that are hard to generalize in C. I'll add a comment about unifying the two field-mask implementations when refactoring to Rust

src/ttl_table.c
GuyAv46: because the setup is a bit complex (`t_fieldIndex` is not necessarily the bit offset in `t_fieldMask`) and I prefer to have a dedicated logic for this case, as for the iterator over this inverted index will always has the same field input and may call this function for each `Read` and `SkipTo`, potentially more than once per API call

src/ttl_table.c
GuyAv46: For performance reasons, I prefer to have a dedicated API per input type instead of pre-processing each case to have a single API, which also require translating the filed-mask into an array although it may not be used (due to early returns, or expiration status that was found for some bit)

src/ttl_table.h
GuyAv46: The above function will eventually be removed. It gets an array of `t_fieldIndex` while this one gets a single index

tests/cpptests/test_cpp_index.cpp
GuyAv46: This test was setting the first ID to be 0, which is undefined behavior and shouldn't happen in the code in practice.
Changing the first ID to be 1 also shifts the `lastId` by one

src/ttl_table.c
kei-nan: This is a lot of code duplication.
I understand not wanting to translate fields but Isn't there a way to have some common helper functions between all these functions to reduce the code duplication?

JoanFM (approved):
PR #6519 should allow to run microbenchmarks

src/ttl_table.c
GuyAv46: Not really, not in a meaningful way.
I can write a macro but I don't think it will be more readable

PR #6371, opened 2025-06-24
[MOD-10184] add some optimizations to simplify some patterns in iterators

JoanFM:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

### DIFF
--- a/src/iterators/intersection_iterator.c
+++ b/src/iterators/intersection_iterator.c
@@ -259,7 +259,6 @@ static QueryIterator *IntersectionIteratorReducer(QueryIterator **its, size_t *n
   // Remove all wildcard iterators from the array
   size_t current_size = *num;
   size_t write_idx = 0;
-  size_t empty_iterator_idx = 0;
   bool all_wildcards = true;
   for (size_t read_idx = 0; read_idx < current_size; read_idx++) {
     if (IsWildcardIterator(its[read_idx])) {
@@ -275,17 +274,19 @@ static QueryIterator *IntersectionIteratorReducer(QueryIterator **its, size_t *n
   }
   *num = write_idx;
 
+  // Check for empty iterators
   for (size_t ii = 0; ii < write_idx; ++ii) {
     if (!its[ii] || its[ii]->type == EMPTY_ITERATOR) {
-      empty_iterator_idx = ii;
       ret = its[ii] ? its[ii] : IT_V2(NewEmptyIterator)();
+      its[ii] = NULL; // Mark as taken
       break;
     }
   }
 
   if (ret) {
+    // Free all non-NULL iterators
     for (size_t ii = 0; ii < write_idx; ++ii) {
-      if (its[ii] && ii != empty_iterator_idx) {
+      if (its[ii]) {
         its[ii]->Free(its[ii]);
       }
     }
--- a/src/query.c
+++ b/src/query.c
@@ -1480,7 +1480,9 @@ static IndexIterator *Query_EvalTagNode(QueryEvalCtx *q, QueryNode *qn) {
       array_free(total_its);
     }
   }
-  int quickExit = q->notSubtree ? 1 : 0;
+  // We want to get results with all the matching children (`quickExit == false`), unless:
+  // 1. We are a `Not` sub-tree, so we only care about the set of IDs
+  bool quickExit = q->notSubtree;
   ret = NewUnionIterator(iters, n, quickExit, qn->opts.weight, QN_TAG, NULL, q->config);
 
 done:


src/hybrid_reader.c
```diff
@@ -443,11 +443,27 @@ void HybridIterator_Free(struct indexIterator *self) {
...
 
+static IndexIterator* HybridIteratorReducer(HybridIteratorParams *hParams) {
+  IndexIterator* ret = NULL;
+  if (hParams->childIt && hParams->childIt->type == EMPTY_ITERATOR) {
+    ret = hParams->childIt;
+  } else if (hParams->childIt && hParams->childIt->type == WILDCARD_ITERATOR) {
+    hParams->childIt->Free(hParams->childIt);
+    hParams->qParams.searchMode = VECSIM_STANDARD_KNN;
+    hParams->childIt = NULL;
+  }
```
GuyAv46: Can we cover these cases?
We should also handle the case where `hParams->childIt == NULL`

src/iterators/intersection_iterator.c
```diff
@@ -231,65 +232,119 @@ static int cmpIter(QueryIterator **it1, QueryIterator **it2) {
 }
 
 // Set estimation for number of results. Returns false if the query is empty (some of the iterators are NULL)
```
GuyAv46: fix comment as well

src/iterators/intersection_iterator.c
```diff
@@ -231,65 +232,119 @@ static int cmpIter(QueryIterator **it1, QueryIterator **it2) {
...
+}
+
+/**
+ * Reduce the intersection iterator by applying these rules:
+ * 1. If any of the iterators is an empty iterator, return the empty iterator and update the number of children
+ * 2. Remove all wildcard iterators since they would not contribute to the intersection
+ * 3. If there is only one left child iterator, return it
+ * 4. Otherwise, return NULL and let the caller create the intersection iterator
+*/
+static QueryIterator *IntersectionIteratorReducer(QueryIterator **its, size_t *num, int max_slop, bool in_order, double weight) {
```
GuyAv46:
```suggestion
static QueryIterator *IntersectionIteratorReducer(QueryIterator **its, size_t *num) {
```

src/iterators/intersection_iterator.c
```diff
@@ -231,65 +232,119 @@ static int cmpIter(QueryIterator **it1, QueryIterator **it2) {
...
+    }
+  }
+  *num = write_idx;
+
+  for (size_t ii = 0; ii < write_idx; ++ii) {
+    if (!its[ii] || its[ii]->type == EMPTY_ITERATOR) {
+      has_empty_iterator = true;
+      empty_iterator_idx = ii;
+      if (!its[empty_iterator_idx]) {
+        its[empty_iterator_idx] = IT_V2(NewEmptyIterator)();
```
GuyAv46:
```suggestion
        ret = IT_V2(NewEmptyIterator)();
```

src/iterators/intersection_iterator.c
```diff
@@ -231,65 +232,119 @@ static int cmpIter(QueryIterator **it1, QueryIterator **it2) {
...
+ * 3. If there is only one left child iterator, return it
+ * 4. Otherwise, return NULL and let the caller create the intersection iterator
+*/
+static QueryIterator *IntersectionIteratorReducer(QueryIterator **its, size_t *num, int max_slop, bool in_order, double weight) {
+  QueryIterator *ret = NULL;
+
+  // Remove all wildcard iterators from the array
+  size_t current_size = *num;
+  size_t write_idx = 0;
+  size_t empty_iterator_idx = 0;
```
GuyAv46: redundant

src/iterators/intersection_iterator.c
```diff
@@ -231,65 +232,119 @@ static int cmpIter(QueryIterator **it1, QueryIterator **it2) {
...
+      if (its[ii] && ii != empty_iterator_idx) {
+        its[ii]->Free(its[ii]);
+      }
+    }
+    ret = its[empty_iterator_idx];
+  } else {
+    // Handle edge cases after wildcard removal
+    if (write_idx == 0) {
+      // All iterators were wildcards, return NULL
+      ret = IT_V2(NewEmptyIterator)();
```
GuyAv46:
```suggestion
      // All iterators were wildcards, return Wildcard
      ret = <wildcard constructor, or reuse one of them>;
```

src/iterators/intersection_iterator.c
```diff
@@ -231,65 +232,119 @@ static int cmpIter(QueryIterator **it1, QueryIterator **it2) {
...
+    } else if (write_idx == 1) {
+      // Only one iterator left, return it directly
+      ret = its[0];
+    }
+  }
+
+  if (ret != NULL) {
+    rm_free(its);
+  }
```
GuyAv46, line 306: I think it would be simpler to handle the reduction after we construct the intersection, so we can call its free function and avoid manually handling cleanup. This could become complicated if we ever add more allocations to the iterator

src/iterators/not_iterator.c
```diff
@@ -226,12 +226,44 @@ static IteratorStatus NI_SkipTo_Optimized(QueryIterator *base, t_docId docId) {
...
+static QueryIterator* NotIteratorReducer(QueryIterator *it, t_docId maxDocId, double weight, struct timespec timeout, QueryEvalCtx *q) {
+  QueryIterator *ret = NULL;
+  if (!it || it->type == EMPTY_ITERATOR) {
+    if (q) {
+      bool optimized = q->sctx && q->sctx->spec && q->sctx->spec->rule && q->sctx->spec->rule->index_all;
+      if (optimized) {
+        ret = IT_V2(NewWildcardIterator_Optimized)(q->sctx, 1);
+      } else {
+        ret = IT_V2(NewWildcardIterator)(q, 1);
+      }
```
GuyAv46: No need to check this. Use only the generic API. Also pass correct weight
```suggestion
      ret = IT_V2(NewWildcardIterator)(q, weight);
```

src/iterators/not_iterator.c
```diff
@@ -226,12 +226,44 @@ static IteratorStatus NI_SkipTo_Optimized(QueryIterator *base, t_docId docId) {
...
+/*
+ * Reduce the not iterator by applying these rules:
+ * 1. If the child is an empty iterator or NULL, return a wildcard iterator
+ * 2. If the child is a wildcard iterator, return an empty iterator
+ * 3. Otherwise, return NULL and let the caller create the not iterator
+ */
+static QueryIterator* NotIteratorReducer(QueryIterator *it, t_docId maxDocId, double weight, struct timespec timeout, QueryEvalCtx *q) {
+  QueryIterator *ret = NULL;
+  if (!it || it->type == EMPTY_ITERATOR) {
+    if (q) {
```
GuyAv46: Can we assert that instead? We should return a wildcard in this case, and we must have it for that. Otherwise the logic is wrong

src/iterators/not_iterator.c
```diff
@@ -226,12 +226,44 @@ static IteratorStatus NI_SkipTo_Optimized(QueryIterator *base, t_docId docId) {
...
+    return ret;
+  }
   NotIterator *ni = rm_calloc(1, sizeof(*ni));
-  QueryIterator *ret = &ni->base;
-  bool optimized = q && q->sctx->spec->rule && q->sctx->spec->rule->index_all;
+  ret = &ni->base;
+  bool optimized = q && q->sctx && q->sctx->spec && q->sctx->spec->rule && q->sctx->spec->rule->index_all;
   if (optimized) {
-    ni->wcii = IT_V2(NewWildcardIterator_Optimized)(q->sctx);
+    ni->wcii = IT_V2(NewWildcardIterator_Optimized)(q->sctx, 1);
```
GuyAv46:
```suggestion
    ni->wcii = IT_V2(NewWildcardIterator_Optimized)(q->sctx, weight);
```

src/iterators/not_iterator.c
```diff
@@ -226,12 +226,44 @@ static IteratorStatus NI_SkipTo_Optimized(QueryIterator *base, t_docId docId) {
...
   }
   ni->child = it ? it : IT_V2(NewEmptyIterator)();
```
GuyAv46: Now we can just set it
```suggestion
  ni->child = it;
```

src/iterators/optional_iterator.c
```diff
@@ -173,22 +174,51 @@ static IteratorStatus OI_Read_NotOptimized(QueryIterator *base) {
...
+static QueryIterator* OptionalIteratorReducer(QueryIterator *it, QueryEvalCtx *q, double weight) {
+  QueryIterator *ret = NULL;
+  if (!it || it->type == EMPTY_ITERATOR) {
+    // If the child is NULL, we return a wildcard iterator. All will be virtual hits
+    bool optimized = q->sctx->spec->rule && q->sctx->spec->rule->index_all;
+    if (optimized) {
+      ret = IT_V2(NewWildcardIterator_Optimized)(q->sctx, weight);
+    } else {
+      ret = IT_V2(NewWildcardIterator)(q, weight);
+    }
```
GuyAv46: Use generic API
```suggestion
    ret = IT_V2(NewWildcardIterator)(q, weight);
```

src/iterators/union_iterator.c
```diff
@@ -372,8 +373,57 @@ static void UI_Free(QueryIterator *base) {
...
+          if (i != j && its[j]) {
+            its[j]->Free(its[j]);
+          }
+        }
+        break;
+      }
+    }
+  }
+
+  if (write_idx == 1) {
```
GuyAv46, line 412: Handle + test `write_idx == 0`

src/iterators/wildcard_iterator.c
```diff
@@ -75,24 +86,29 @@ QueryIterator *IT_V2(NewWildcardIterator_NonOptimized)(t_docId maxId, size_t num
...
 
 // Returns a new wildcard iterator.
 // If the spec tracks all existing documents, it will return an iterator over those documents.
 // Otherwise, it will return a non-optimized wildcard iterator
-QueryIterator *IT_V2(NewWildcardIterator)(const QueryEvalCtx *q) {
-  if (q->sctx->spec->rule->index_all == true) {
-    return IT_V2(NewWildcardIterator_Optimized)(q->sctx);
+QueryIterator *IT_V2(NewWildcardIterator)(const QueryEvalCtx *q, double weight) {
+  QueryIterator *ret = NULL;
+  if (q && q->sctx && q->sctx->spec && q->sctx->spec->rule && q->sctx->spec->rule->index_all == true) {
```
GuyAv46: Most of these can never happen. Consider avoiding the excessive validation

src/query.c
```diff
@@ -915,10 +915,14 @@ static IndexIterator *Query_EvalWildcardNode(QueryEvalCtx *q, QueryNode *qn) {
 
 static IndexIterator *Query_EvalNotNode(QueryEvalCtx *q, QueryNode *qn) {
   RS_LOG_ASSERT(qn->type == QN_NOT, "query node type should be not")
+  IndexIterator *child = NULL;
+  if (qn) {
```
GuyAv46: Redundant. We already assert that

src/query.c
```diff
@@ -915,10 +915,14 @@ static IndexIterator *Query_EvalWildcardNode(QueryEvalCtx *q, QueryNode *qn) {
...
+    q->notSubtree = true;
+    child = Query_EvalNode(q, qn->children[0]);
+    q->notSubtree = false;
```
GuyAv46: Consider adding a test for this case (`Not(expr(Not(...)))`)
```suggestion
    bool currently_notSubtree = q->notSubtree;
    q->notSubtree = true;
    child = Query_EvalNode(q, qn->children[0]);
    q->notSubtree = currently_notSubtree;
```

src/query.c
```diff
@@ -1092,8 +1096,8 @@ static IndexIterator *Query_EvalUnionNode(QueryEvalCtx *q, QueryNode *qn) {
     rm_free(iters);
     return ret;
   }
-
-  IndexIterator *ret = NewUnionIterator(iters, n, 0, qn->opts.weight, QN_UNION, NULL, q->config);
+  int quickExit = q->notSubtree ? 1 : 0;
+  IndexIterator *ret = NewUnionIterator(iters, n, quickExit, qn->opts.weight, QN_UNION, NULL, q->config);
```
GuyAv46: Consider adding a comment on why we want a quick exit if and only if we are a not-subtree
```suggestion
  bool quickExit = q->notSubtree;
  IndexIterator *ret = NewUnionIterator(iters, n, quickExit, qn->opts.weight, QN_UNION, NULL, q->config);
```

src/query.c
```diff
@@ -1480,8 +1484,8 @@ static IndexIterator *Query_EvalTagNode(QueryEvalCtx *q, QueryNode *qn) {
       array_free(total_its);
     }
   }
-
-  ret = NewUnionIterator(iters, n, 0, qn->opts.weight, QN_TAG, NULL, q->config);
+  int quickExit = q->notSubtree ? 1 : 0;
```
GuyAv46: Consider adding a comment on why we want a quick exit if and only if we are a not-subtree
```suggestion
  bool quickExit = q->notSubtree;
```

src/query_parser/v1/parser.y
```diff
@@ -91,6 +91,27 @@ static int one_not_null(void *a, void *b, void *out) {
...
 
+// optimize NOT nodes: NOT(NOT(A)) = A
+// if the child is a NOT node, return its child instead of creating a double negation
+static inline struct RSQueryNode* not_step(struct RSQueryNode* child) {
+    if (!child) {
+        return NULL;
+    }
+
+    // If the child is a NOT node, return its child (double negation elimination)
+    if (child->type == QN_NOT && QueryNode_NumChildren(child) == 1) {
```
GuyAv46: This is implied if it's a Not node. We can assert that if you want
```suggestion
    if (child->type == QN_NOT) {
```

src/query_parser/v2/parser.y
```diff
@@ -147,6 +147,27 @@ static inline struct RSQueryNode* union_step(struct RSQueryNode* B, struct RSQue
...
 
+// optimize NOT nodes: NOT(NOT(A)) = A
+// if the child is a NOT node, return its child instead of creating a double negation
+static inline struct RSQueryNode* not_step(struct RSQueryNode* child) {
+    if (!child) {
+        return NULL;
+    }
+
+    // If the child is a NOT node, return its child (double negation elimination)
+    if (child->type == QN_NOT && QueryNode_NumChildren(child) == 1) {
```
GuyAv46: This is implied if it's a Not node. We can assert that if you want
```suggestion
    if (child->type == QN_NOT) {
```

tests/cpptests/iterator_util.h
```diff
@@ -36,7 +36,7 @@ class MockIterator {
     std::optional<std::chrono::nanoseconds> sleepTime; // Sleep for this duration before returning from Read/SkipTo
 private:
     void Init() {
-      base.type = READ_ITERATOR;
+      base.type = ID_LIST_ITERATOR;
```
GuyAv46: Nitpick - I guess this is more correct than `READ_ITERATOR`, but consider using `MAX_ITERATOR`

tests/cpptests/test_cpp_iterator_not.cpp
```diff
@@ -635,3 +639,173 @@ TEST_F(NotIteratorNoChildTest, Rewind) {
...
+  RedisSearchCtx* sctx = (RedisSearchCtx*)rm_calloc(1, sizeof(RedisSearchCtx));
+  sctx->spec = spec;
+
+  DocTable* docTable = (DocTable*)rm_calloc(1, sizeof(DocTable));
+  docTable->maxDocId = maxDocId;
+  docTable->size = maxDocId;
+
+  QueryEvalCtx* qctx = (QueryEvalCtx*)rm_calloc(1, sizeof(QueryEvalCtx));
+  qctx->sctx = sctx;
+  qctx->docTable = docTable;
```
GuyAv46: Use the existing `MockQueryEvalCtx` instead
```suggestion
  MockQueryEvalCtx mockQctx;
```

tests/cpptests/test_cpp_iterator_not.cpp
```diff
@@ -635,3 +639,173 @@ TEST_F(NotIteratorNoChildTest, Rewind) {
...
```
GuyAv46: Use the existing `MockQueryEvalCtx` instead
```suggestion
  MockQueryEvalCtx mockQctx;
```

tests/cpptests/test_cpp_iterator_not.cpp
```diff
@@ -635,3 +639,173 @@ TEST_F(NotIteratorNoChildTest, Rewind) {
...
```
GuyAv46: Use the existing `MockQueryEvalCtx` instead
```suggestion
  std::vector<t_docId> wildcard = {1, 2, 3};
  MockQueryEvalCtx mockQctx(wildcard);
```

tests/cpptests/test_cpp_iterator_not.cpp
```diff
@@ -635,3 +639,173 @@ TEST_F(NotIteratorNoChildTest, Rewind) {
...
```
GuyAv46: Use the existing `MockQueryEvalCtx` instead
```suggestion
  std::vector<t_docId> wildcard = {1, 2, 3};
  MockQueryEvalCtx mockQctx(wildcard);
```

src/hybrid_reader.c
JoanFM: are there specific unit tests for this iterator?

src/iterators/intersection_iterator.c
JoanFM: but this is independent from the iterator constructor. If the optim comes after construction, you may need to Free the iterator with its children except for one or semantics like these that may actually complicate things?

src/iterators/not_iterator.c
JoanFM: what is that?

src/hybrid_reader.c
GuyAv46: Yes, in `test_cpp_index.cpp`. You can also write a flow (Python) test, as this code is reachable

src/iterators/intersection_iterator.c
GuyAv46: You can take it and set its position to NULL or an empty iterator and then free the intersection

JoanFM: but if I keep the iterator as valid and I just remove the WildCard iterator, then we are in the same position, I have to handle a new `its`. I think it is better to keep as here

src/iterators/intersection_iterator.c
```diff
@@ -249,51 +249,51 @@ static void II_SetEstimation(IntersectionIterator *it) {
...
+      all_wildcards = false;
       its[write_idx++] = its[read_idx];
     }
   }
   *num = write_idx;
 
   for (size_t ii = 0; ii < write_idx; ++ii) {
     if (!its[ii] || its[ii]->type == EMPTY_ITERATOR) {
-      has_empty_iterator = true;
       empty_iterator_idx = ii;
```
GuyAv46: If we "take" the iterator here (set `its[ii] = NULL`), we don't need to track this index, and we can simplify the rest of the logic

src/iterators/intersection_iterator.c
```diff
@@ -309,7 +309,7 @@ static QueryIterator *IntersectionIteratorReducer(QueryIterator **its, size_t *n
 
 QueryIterator *NewIntersectionIterator(QueryIterator **its, size_t num, int max_slop, bool in_order, double weight) {
   RS_ASSERT(its && num > 0);
-  QueryIterator *ret = IntersectionIteratorReducer(its, &num, max_slop, in_order, weight);
+  QueryIterator *ret = IntersectionIteratorReducer(its, &num);
   if (ret != NULL) {
     return ret;
```
GuyAv46, line 315:
```suggestion
    ret->current->weight = weight;
    return ret;
```

src/query.c
GuyAv46: missed this one

src/query.c
```diff
@@ -1092,8 +1095,8 @@ static IndexIterator *Query_EvalUnionNode(QueryEvalCtx *q, QueryNode *qn) {
     rm_free(iters);
     return ret;
   }
-
-  IndexIterator *ret = NewUnionIterator(iters, n, 0, qn->opts.weight, QN_UNION, NULL, q->config);
+  bool quickExit = q->notSubtree;
```
GuyAv46, line 1095:
```suggestion
  // We want to get results with all the matching children (`quickExit == false`), unless:
  // 1. We are a `Not` sub-tree, so we only care about the set of IDs
  bool quickExit = q->notSubtree;
```

src/query.c
GuyAv46:
```suggestion
  // We want to get results with all the matching children (`quickExit == false`), unless:
  // 1. We are a `Not` sub-tree, so we only care about the set of IDs
  bool quickExit = q->notSubtree;
```

GuyAv46 (approved):

PR #6374, opened 2025-06-25
Fix lints that will be introduced in Rust 1.88.0

LukeMathWalker:

Every new Rust release introduces new lints in `clippy` and the compiler itself.
This PR fixes the newly reported violations from 1.88.0.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/trie_rs/src/iter/range.rs
+++ b/src/redisearch_rs/trie_rs/src/iter/range.rs
@@ -293,42 +293,42 @@ impl<'tm, Data> RangeIter<'tm, Data> {
                 self.stack.reserve(node.children().len());
 
                 let mut max_index = node.children().len();
-                if let Some(max) = child_max {
-                    if let Some(first) = max.first() {
-                        max_index = match node.children_first_bytes().binary_search(first) {
-                            Ok(i) => {
-                                self.stack.push(StackEntry {
-                                    node: &node.children()[i],
-                                    was_visited: false,
-                                    min: child_min,
-                                    max: child_max,
-                                });
-                                i
-                            }
-                            Err(i) => i,
-                        };
-                    }
+                if let Some(max) = child_max
+                    && let Some(first) = max.first()
+                {
+                    max_index = match node.children_first_bytes().binary_search(first) {
+                        Ok(i) => {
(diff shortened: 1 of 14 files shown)


LukeMathWalker:
Keeping it as a draft, since we can't merge until 1.88.0 is released, given that let-chains aren't stable on 1.87.x.

src/redisearch_rs/trie_rs/src/iter/range.rs
```diff
@@ -293,42 +293,42 @@ impl<'tm, Data> RangeIter<'tm, Data> {
...
-                                    min: child_min,
-                                    max: child_max,
-                                });
-                                i
-                            }
-                            Err(i) => i,
-                        };
-                    }
+                if let Some(max) = child_max
+                    && let Some(first) = max.first()
```
chesedo, line 297: Maybe this could be a `match` instead?

src/redisearch_rs/trie_rs/src/iter/range.rs
```diff
@@ -293,42 +293,42 @@ impl<'tm, Data> RangeIter<'tm, Data> {
...
-                                        min: child_min,
-                                        max: child_max,
-                                    });
-                                    i + 1
-                                }
-                                Err(i) => i,
-                            };
-                    }
+                if let Some(min) = child_min
+                    && let Some(first) = min.first()
```
chesedo, line 317: Maybe a `match` will be simpler?

chesedo (commented):
Just a small nit that migth not be beneficial

src/redisearch_rs/trie_rs/src/iter/range.rs
LukeMathWalker: Since we only care about one branch, I'd say no.

src/redisearch_rs/trie_rs/src/iter/range.rs
LukeMathWalker: Since we only care about one branch, I'd say no.

src/redisearch_rs/trie_rs/src/iter/range.rs
chesedo: Makes sense

PR #6376, opened 2025-06-25
MOD-10224: Add vector filter validation

nafraf:
**Description:** 
Add vector filter validation and related 

1. Current: 
    - There is not function to validate if the vector (VSIM) filter is correct.
2. Change:
    - Introduced explicitWeight option in QueryNodeOptions to track user-defined weights.
    - Implemented QAST_CheckIsValidAsVectorFilter to validate vector filter queries.
    - Created query_test_utils.h for C++ wrappers and utility functions.
    - Added test cases for vector filter validation in test_cpp_vector_filter.cpp.
    - Refactored existing tests to utilize new utility functions.
3. Outcome: 
    - `QAST_CheckIsValidAsVectorFilter()` can be used to validate if VSIM filter clause is valid

1. MOD-10224

1. ...

#### Mark if applicable

### DIFF
--- a/src/query.c
+++ b/src/query.c
@@ -154,6 +154,7 @@ QueryNode *NewQueryNode(QueryNodeType type) {
     .maxSlop = -1,
     .inOrder = 0,
     .weight = 1,
+    .explicitWeight = false,
     .distField = NULL,
   };
   return s;
@@ -1828,6 +1829,54 @@ int QAST_CheckIsValid(QueryAST *q, IndexSpec *spec, RSSearchOptions *opts, Query
   return QueryNode_CheckIsValid(q->root, spec, opts, status);
 }
 
+static int QueryNode_CheckIsValidAsVectorFilter(QueryNode *n, QueryError *status) {
+  int withChildren = 1;
+  int res = REDISMODULE_OK;
+  switch(n->type) {
+    case QN_VECTOR:
+      if (n->vn.vq->type == VECSIM_QT_KNN) {
+        QueryError_SetError(status, QUERY_ESYNTAX, "VECTOR KNN queries are not allowed in a vector filter");
+        res = REDISMODULE_ERR;
+      } else if (n->vn.vq->type == VECSIM_QT_RANGE) {
+        QueryError_SetError(status, QUERY_ESYNTAX, "VECTOR RANGE queries are not allowed in a vector filter");
+        res = REDISMODULE_ERR;
+      }
+      break;
+    case QN_NULL:
+    case QN_MISSING:
+    case QN_MAX:
+      withChildren = 0;
+      break;
+    default:
+      if (n->opts.explicitWeight) {
+        QueryError_SetError(status, QUERY_ESYNTAX, "Weight attribute is not allowed in a vector filter");
+        res = REDISMODULE_ERR;
+      }
+      break;
+  }
+  // Handle children
+  if (withChildren && res == REDISMODULE_OK) {
+    for (size_t ii = 0; ii < QueryNode_NumChildren(n); ++ii) {
+      res = QueryNode_CheckIsValidAsVectorFilter(n->children[ii], status);
+      if (res == REDISMODULE_ERR)
+        break;
+    }
+  }
+  return res;
+}
+
+// Checks whether QAST is valid as vector filter.
+// It can't contain:
+// Vector KNN search =>[KNN num_neighbours @field $vector]
+// Vector RANGE search @<vector_field>:[VECTOR_RANGE (<radius> | $<radius_param>) $<vector_blob_param> $<vector_query_params>]
+// Weight attribute  =>{weight: X}
+int QAST_CheckIsValidAsVectorFilter(QueryAST *q, QueryError *status) {
+  if (!q || !q->root) {
+    return REDISMODULE_OK;
+  }
+  return QueryNode_CheckIsValidAsVectorFilter(q->root, status);
+}
+
 /* Set the field mask recursively on a query node. This is called by the parser to handle
  * situations like @foo:(bar baz|gaz), where a complex tree is being applied a field mask */
 void QueryNode_SetFieldMask(QueryNode *n, t_fieldMask mask) {
@@ -2207,6 +2256,7 @@ static int QueryNode_ApplyAttribute(QueryNode *qn, QueryAttribute *attr, QueryEr
       return res;
     }
     qn->opts.weight = d;
+    qn->opts.explicitWeight = true;
     res = 1;
 
   } else if (STR_EQCASE(attr->name, attr->namelen, PHONETIC_ATTR)) {
--- a/src/query.h
+++ b/src/query.h
@@ -130,6 +130,7 @@ int QAST_EvalParams(QueryAST *q, RSSearchOptions *opts, unsigned int dialectVers
 int QueryNode_EvalParams(dict *params, QueryNode *node, unsigned int dialectVersion, QueryError *status);
 
 int QAST_CheckIsValid(QueryAST *q, IndexSpec *spec, RSSearchOptions *opts, QueryError *status);
+int QAST_CheckIsValidAsVectorFilter(QueryAST *q, QueryError *status);
 
 /* Return a string representation of the QueryParseCtx parse tree. The string should be freed by the
  * caller */
--- a/tests/cpptests/test_cpp_vector_filter.cpp
+++ b/tests/cpptests/test_cpp_vector_filter.cpp
@@ -1,0 +1,83 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+#include "src/redisearch.h"
+#include "src/spec.h"
+#include "query_test_utils.h"
+
+#include "gtest/gtest.h"
+
+#include <algorithm>
+
+class VectorFilterTest : public ::testing::Test {
+protected:
+  void SetUp() override {
+  }
+};
+
+bool isValidAsVectorFilter(const char *qt, RedisSearchCtx &ctx) {
+  QASTCXX ast;
+  ast.setContext(&ctx);
+  return ast.isValidAsVectorFilter(qt);
+}
+
+#define assertValidVectorFilter(qt, ctx) ASSERT_TRUE(isValidAsVectorFilter(qt, ctx))
+#define assertInvalidVectorFilter(qt, ctx) ASSERT_FALSE(isValidAsVectorFilter(qt, ctx))
+
+TEST_F(VectorFilterTest, testVectorFilter) {
+  // Create an index spec with a title field and a vector field
+  static const char *args[] = {
+    "SCHEMA",
+    "title", "text",
+    "body", "text",
+    "v", "vector", "HNSW", "6", "TYPE", "FLOAT32", "DIM", "4", "DISTANCE_METRIC", "L2",
+    "v2", "vector", "HNSW", "6", "TYPE", "FLOAT32", "DIM", "4", "DISTANCE_METRIC", "L2"};
+
+  QueryError err = {QUERY_OK};
+  StrongRef ref = IndexSpec_ParseC("idx", args, sizeof(args) / sizeof(const char *), &err);
+  ASSERT_EQ(err.code, QUERY_OK) << QueryError_GetUserError(&err);
+
+  RedisSearchCtx ctx = SEARCH_CTX_STATIC(NULL, (IndexSpec *)StrongRef_Get(ref));
+
+  // Invalid queries with KNN
+  assertInvalidVectorFilter("*=>[KNN 10 @vec_field $BLOB]", ctx);
+  assertInvalidVectorFilter("@title:hello =>[KNN 10 @vec_field $BLOB]", ctx);
+
+  // Invalid queries with range
+  assertInvalidVectorFilter("@v:[VECTOR_RANGE 0.01 $BLOB]", ctx);
+  assertInvalidVectorFilter("hello | @v:[VECTOR_RANGE 0.01 $BLOB]", ctx);
+
+  // Invalid queries with weight
+  assertInvalidVectorFilter("@title:hello => {$weight: 2.0}", ctx);
+  assertInvalidVectorFilter("hello | @title:hello => {$weight: 2.0}", ctx);
+  assertInvalidVectorFilter("@title:'' => {$weight: 2.0}", ctx);
+  assertInvalidVectorFilter("( @title:(foo bar) @body:lol => {$weight: 2.0;} )=> {$slop:2; $inorder:true}", ctx);
+  assertInvalidVectorFilter("( @title:(foo bar) @body:lol )=> {$weight:2.0; $inorder:true}", ctx);
+
+  // Complex queries with range
+  assertInvalidVectorFilter("@v:[VECTOR_RANGE 0.01 $BLOB] @title:foo OR bar", ctx);
+  assertInvalidVectorFilter("(@v:[VECTOR_RANGE 0.01 $BLOB] @title:foo) => { $weight: 2.0 }", ctx);
+  assertInvalidVectorFilter("@v:[VECTOR_RANGE 0.01 $BLOB] @title:foo OR bar @v:[VECTOR_RANGE 0.04 $BLOB2]", ctx);
+  assertInvalidVectorFilter("(@v:[VECTOR_RANGE 0.01 $BLOB] @title:foo) => [KNN 5 @v $BLOB2]", ctx);
+  assertInvalidVectorFilter("@v:[VECTOR_RANGE 0.01 $BLOB] => [KNN 5 @v2 $BLOB2 AS second_score]", ctx);
+  assertInvalidVectorFilter("@v:[VECTOR_RANGE 0.01 $BLOB]=>{$yield_distance_as: score1;} => [KNN 5 @v2 $BLOB2 AS second_score]", ctx);
+  assertInvalidVectorFilter("@v:[VECTOR_RANGE 0.01 $BLOB]=>{$yield_distance_as: score1;} => [KNN 5 @v2 $BLOB2] => {$yield_distance_as:second_score;}", ctx);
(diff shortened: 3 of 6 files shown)


src/query.c
```diff
@@ -1828,6 +1829,53 @@ int QAST_CheckIsValid(QueryAST *q, IndexSpec *spec, RSSearchOptions *opts, Query
   return QueryNode_CheckIsValid(q->root, spec, opts, status);
 }
 
+static int QueryNode_CheckIsValidAsVectorFilter(QueryNode *n, QueryError *status) {
+  int withChildren = 1;
```
JoanFM: curiousity, why not a boolean?

tests/cpptests/test_cpp_vector_filter.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include "src/spec.h"
+#include "query_test_utils.h"
+
+#include "gtest/gtest.h"
+
+#include <algorithm>
+
+class VectorFilterTest : public ::testing::Test {
+protected:
+  void SetUp() override {
```
JoanFM: you do not need this empty SetUp right?

tests/cpptests/test_cpp_vector_filter.cpp
```diff
@@ -0,0 +1,86 @@
...
+bool isValidAsVectorFilter(const char *qt, RedisSearchCtx &ctx) {
+  QASTCXX ast;
+  ast.setContext(&ctx);
+  return ast.isValidAsVectorFilter(qt);
+}
+
+#define assertValidVectorFilter(qt, ctx) ASSERT_TRUE(isValidAsVectorFilter(qt, ctx))
+#define assertInvalidVectorFilter(qt, ctx) ASSERT_FALSE(isValidAsVectorFilter(qt, ctx))
+
+TEST_F(VectorFilterTest, testVectorFilter) {
```
JoanFM: small suggestion, why don't we put the ctx construction in the setup, and then you can separate in different tests where you test the different type of validations?

tests/cpptests/test_cpp_vector_filter.cpp
nafraf: Right. Removed.

src/query.c
nafraf: Right. Modified to be boolean.

tests/cpptests/test_cpp_vector_filter.cpp
nafraf: I prefer to continue using the same pattern used for all the other tests, a single schema for each test. But now I created two tests: one for valid tests and another for the invalid cases.
Thanks.

src/query.c
```diff
@@ -1828,6 +1829,53 @@ int QAST_CheckIsValid(QueryAST *q, IndexSpec *spec, RSSearchOptions *opts, Query
...
+  return res;
+}
+
+// Checks whether QAST is valid as vector filter.
+// It can't contain:
+// Vector KNN search =>[KNN num_neighbours @field $vector]
+// Vector RANGE search @<vector_field>:[VECTOR_RANGE (<radius> | $<radius_param>) $<vector_blob_param> $<vector_query_params>]
+// Weight attribute  =>{weight: X}
+int QAST_CheckIsValidAsVectorFilter(QueryAST *q, QueryError *status) {
+  if (!q || !q->root) {
```
raz-mon: Is the coverage report faulty or do we actually not enter here at all?
Consider asserting instead

raz-mon: I see it is covered in `QAST_CheckIsValid`, so it's probably a faulty report - just validate please

raz-mon (commented):
Nicely and cleanly done! 💪 

My main comment: With this implementation, we would need to traverse the AST again after the first traversal, done in `QAST_CheckIsValid`, which is probably ok for most queries, but not optimal. Consider implementing it such that we will perform all validations in one traversal, e.g., by invoking a check on a node according to some bit-field of checks relevant for this traversal, or something of that sort.

src/query.c
nafraf: `QAST_CheckIsValidAsVectorFilter()` was removed, now all the validations are done in `QAST_CheckIsValid()`

nafraf:
> My main comment: With this implementation, we would need to traverse the AST again after the first traversal, done in `QAST_CheckIsValid`, which is probably ok for most queries, but not optimal. Consider implementing it such that we will perform all validations in one traversal, e.g., by invoking a check on a node according to some bit-field of checks relevant for this traversal, or something of that sort.

@raz-mon Thanks.  `QAST_CheckIsValid` was modified to support the vector filter validation, the validation is done if QAST `isVectorFilter` is true.

src/query.c
```diff
@@ -1792,6 +1794,17 @@ static int QueryNode_CheckIsValid(QueryNode *n, IndexSpec *spec, RSSearchOptions
...
       break;
+    case QN_VECTOR:
+      if (validationFlags & QAST_DISABLE_VECTOR_QUERIES) {
+        if (n->vn.vq->type == VECSIM_QT_KNN) {
+          QueryError_SetError(status, QUERY_ESYNTAX, "VECTOR KNN queries are not allowed in a vector filter");
+          res = REDISMODULE_ERR;
+        } else if (n->vn.vq->type == VECSIM_QT_RANGE) {
+          QueryError_SetError(status, QUERY_ESYNTAX, "VECTOR RANGE queries are not allowed in a vector filter");
+          res = REDISMODULE_ERR;
+        }
```
raz-mon: I wonder if this is necessary.. Wouldn't it be enough to just return a single error for any vector node?
E.g., `Vector queries are not allowed in a vector filter`.

src/query.c
```diff
@@ -1801,14 +1814,19 @@ static int QueryNode_CheckIsValid(QueryNode *n, IndexSpec *spec, RSSearchOptions
...
     case QN_FUZZY:
     case QN_LEXRANGE:
-    case QN_VECTOR:
     case QN_GEOMETRY:
       break;
   }
+
   // Handle children
   if (withChildren && res == REDISMODULE_OK) {
+    if ((validationFlags & QAST_NO_WEIGHT_ATTRIBUTE) && n->opts.explicitWeight) {
```
raz-mon: Why is this validation done only for the `withChildren && res == REDISMODULE_OK` cases?

src/query.h
```diff
@@ -130,6 +139,7 @@ int QAST_EvalParams(QueryAST *q, RSSearchOptions *opts, unsigned int dialectVers
 int QueryNode_EvalParams(dict *params, QueryNode *node, unsigned int dialectVersion, QueryError *status);
 
 int QAST_CheckIsValid(QueryAST *q, IndexSpec *spec, RSSearchOptions *opts, QueryError *status);
+int QAST_CheckIsValidAsVectorFilter(QueryAST *q, QueryError *status);
```
raz-mon: Leftover

src/query.c
oshadmi: We assume `QAST_DISABLE_VECTOR_QUERIES` is only passed from a vector filter validation.
I suppose later it would be replaced by some hierarchy of the query nodes (e.g., HYBRID node would have a VECTOR child node, etc.)

src/query.c
oshadmi: We do not support query attributes for `ismissing`, right?

src/query.h
nafraf: Thanks. Removed.

src/query.c
nafraf: It does not make sense, but the parser accept it:
```sh
127.0.0.1:6379> FT.explaincli myidx '(ismissing(@t))=>{$slop:2; $inorder:true; $weight:0.4}' withscores dialect 2
1) ISMISSING{t} => { $weight: 0.4; $slop: 2; $inorder: true; }
2) 
127.0.0.1:6379> FT.explaincli myidx '(ismissing(@t))=>{$weight:0.4}' withscores dialect 2
1) ISMISSING{t} => { $weight: 0.4; }
``

File: src/query.c (hunk 5 above)
Comment (nafraf) [reply src/query.c]: Error message updated.

File: src/query.c [hunk 8]
```diff
@@ -1796,11 +1796,8 @@ static int QueryNode_CheckIsValid(QueryNode *n, IndexSpec *spec, RSSearchOptions
       break;
     case QN_VECTOR:
       if (validationFlags & QAST_DISABLE_VECTOR_QUERIES) {
-        if (n->vn.vq->type == VECSIM_QT_KNN) {
-          QueryError_SetError(status, QUERY_ESYNTAX, "VECTOR KNN queries are not allowed in a vector filter");
-          res = REDISMODULE_ERR;
-        } else if (n->vn.vq->type == VECSIM_QT_RANGE) {
-          QueryError_SetError(status, QUERY_ESYNTAX, "VECTOR RANGE queries are not allowed in a vector filter");
+        if ((n->vn.vq->type == VECSIM_QT_KNN) || (n->vn.vq->type == VECSIM_QT_RANGE)) {
```
raz-mon: Since these are the two only cases, this condition will always be evaluated to `true` no?
I think you can throw the error once we find a `QN_VECTOR` node

src/query.c
```diff
@@ -1818,13 +1815,14 @@ static int QueryNode_CheckIsValid(QueryNode *n, IndexSpec *spec, RSSearchOptions
       break;
   }
 
+  // Check for weight attribute
+  if (res == REDISMODULE_OK && (validationFlags & QAST_NO_WEIGHT_ATTRIBUTE) && n->opts.explicitWeight) {
```
raz-mon: Nitpick, consider putting this before the switch-case

src/query.c
```diff
@@ -1818,13 +1815,14 @@ static int QueryNode_CheckIsValid(QueryNode *n, IndexSpec *spec, RSSearchOptions
...
+    QueryError_SetError(status, QUERY_ESYNTAX, "Weight attribute is not allowed");
```
raz-mon: I think the previous error message is better.. Consider having the flag be for vector filter validation, such that we can check for the same flag and raise the same error here (as done above for the Vector part).
Let's add the error to `QUERY_XERRS` in `query_error.h`.

raz-mon (commented):
LGTM 💪 
Some small comments

src/query.c
nafraf: You are right. Done.

src/query.c
nafraf: Done.

src/query.c
nafraf: OK. Now the error messages are in `QUERY_XERRS`, we have 3 different error messages.
The tests were modified to check the validation and the error code returned by the validation.

tests/cpptests/test_cpp_query_validation.cpp
```diff
@@ -17,30 +17,58 @@
...
+    ADD_FAILURE() << "Query should be invalid but was valid: " << qt;
+    return false;
+  }
+  if (actual_err_code != error) {
+    ADD_FAILURE() << "Error code mismatch for query '" << qt
+                  << "': expected " << error
+                  << " but got " << actual_err_code
+                  << " Error message: " << ast.getError();
+    return false;
+  }
```
raz-mon: Redundant?

tests/cpptests/test_cpp_query_validation.cpp
```diff
@@ -51,34 +79,31 @@ TEST_F(QueryValidationTest, testInvalidVectorFilter) {
   RedisSearchCtx ctx = SEARCH_CTX_STATIC(NULL, (IndexSpec *)StrongRef_Get(ref));
 
   // Invalid queries with KNN
-  assertInvalidVectorFilter("*=>[KNN 10 @vec_field $BLOB]", ctx);
-  assertInvalidVectorFilter("@title:hello =>[KNN 10 @vec_field $BLOB]", ctx);
+  assertInvalidHybridVectorFilterQuery("*=>[KNN 10 @v $BLOB]", ctx);
+  assertInvalidHybridVectorFilterQuery("@title:hello =>[KNN 10 @v2 $BLOB]", ctx);
```
raz-mon: I see `v` and `v2` are the same - any reason for testing/creating both?

raz-mon (approved):
Nicely done

tests/cpptests/test_cpp_query_validation.cpp
nafraf: OK. I removed the `if(valid)` condition, for valid queries, the comparison of the error codes is enough to return `false`.

tests/cpptests/test_cpp_query_validation.cpp
nafraf: No. Test refactored to use only `v`. Thanks.

src/query_error.h
```diff
@@ -69,7 +69,10 @@ extern "C" {
   X(QUERY_EUNKNOWNINDEX, "Unknown index name")                                                   \
   X(QUERY_EDROPPEDBACKGROUND, "The index was dropped before the query could be executed")        \
   X(QUERY_EALIASCONFLICT, "Alias conflicts with an existing index name")                         \
-  X(QUERY_INDEXBGOOMFAIL, "Index background scan did not complete due to OOM")                             \
+  X(QUERY_INDEXBGOOMFAIL, "Index background scan did not complete due to OOM")                   \
+  X(QUERY_EHYBRID_VSIM_FILTER_INVALID_QUERY, "Vector queries are not allowed in VSIM filter")    \
```
oshadmi:
```suggestion
  X(QUERY_EHYBRID_VSIM_FILTER_INVALID_QUERY, "Vector queries are not allowed in FT.HYBRID VSIM subquery FILTER")    \
```

src/query_error.h
```diff
@@ -69,7 +69,10 @@ extern "C" {
...
+  X(QUERY_EHYBRID_VSIM_FILTER_INVALID_WEIGHT, "Weight attribute is not allowed in VSIM filter")  \
+  X(QUERY_EHYBRID_SEARCH_INVALID_QUERY, "Vector queries are not allowed in HYBRID SEARCH")       \
```
oshadmi:
```suggestion
  X(QUERY_EHYBRID_SEARCH_INVALID_QUERY, "Vector queries are not allowed in FT.HYBRID SEARCH subquery")       \
```

tests/cpptests/test_cpp_query_validation.cpp
```diff
@@ -0,0 +1,206 @@
...
+  assertInvalidHybridVectorFilterWeight("hello | @title:hello => {$weight: 2.0}", ctx);
+  assertInvalidHybridVectorFilterWeight("@title:'hello' => {$weight: 2.0}", ctx);
+  assertInvalidHybridVectorFilterWeight("( @title:(foo bar) @body:lol => {$weight: 2.0;} )=> {$slop:2; $inorder:true}", ctx);
+  assertInvalidHybridVectorFilterWeight("( @title:(foo bar) @body:lol )=> {$weight:2.0; $inorder:true}", ctx);
+  assertInvalidHybridVectorFilterWeight("(ismissing(@body))=> {$weight: 2.0}", ctx);
+  assertInvalidHybridVectorFilterWeight("(@body:'')=> {$weight: 2.0}", ctx);
+  assertInvalidHybridVectorFilterWeight("(@v:[VECTOR_RANGE 0.01 $BLOB] @title:foo) => { $weight: 2.0 }", ctx);
+
+  // // Complex queries with range
+  assertInvalidHybridVectorFilterQuery("@v:[VECTOR_RANGE 0.01 $BLOB] @title:foo OR bar", ctx);
```
oshadmi, line 94: Add test with VECTOR_RANGE on the right hand side of an OR?

tests/cpptests/test_cpp_query_validation.cpp
```diff
@@ -0,0 +1,206 @@
...
+  // Invalid queries with KNN
+  assertInvalidHybridSearchQuery("*=>[KNN 10 @v $BLOB]", ctx);
+  assertInvalidHybridSearchQuery("(@title:hello)=>[KNN 10 @v $BLOB]", ctx);
+
+  // Invalid queries with range
+  assertInvalidHybridSearchQuery("@v:[VECTOR_RANGE 0.01 $BLOB]", ctx);
+  assertInvalidHybridSearchQuery("hello | @v:[VECTOR_RANGE 0.01 $BLOB]", ctx);
+
+  // Complex queries with range
+  assertInvalidHybridSearchQuery("@v:[VECTOR_RANGE 0.01 $BLOB] @title:foo OR bar", ctx);
```
oshadmi, line 161: Add test with VECTOR_RANGE on the right hand side of the OR

src/query_error.h
nafraf: OK. Error messages fixed.

src/query_error.h
nafraf: OK. Error messages fixed.

tests/cpptests/test_cpp_query_validation.cpp
nafraf: Done.

tests/cpptests/test_cpp_query_validation.cpp
nafraf: Test added.

src/query_error.h
```diff
@@ -70,9 +70,9 @@ extern "C" {
   X(QUERY_EDROPPEDBACKGROUND, "The index was dropped before the query could be executed")        \
   X(QUERY_EALIASCONFLICT, "Alias conflicts with an existing index name")                         \
   X(QUERY_INDEXBGOOMFAIL, "Index background scan did not complete due to OOM")                   \
-  X(QUERY_EHYBRID_VSIM_FILTER_INVALID_QUERY, "Vector queries are not allowed in VSIM filter")    \
-  X(QUERY_EHYBRID_VSIM_FILTER_INVALID_WEIGHT, "Weight attribute is not allowed in VSIM filter")  \
-  X(QUERY_EHYBRID_SEARCH_INVALID_QUERY, "Vector queries are not allowed in HYBRID SEARCH")       \
+  X(QUERY_EHYBRID_VSIM_FILTER_INVALID_QUERY, "Vector queries are not allowed in FT.HYBRID VSIM subquery FILTER")  \
+  X(QUERY_EHYBRID_VSIM_FILTER_INVALID_WEIGHT, "Weight attribute is not allowed in FT.HYBRID VSIM subquery FILTER") \
```
oshadmi: Nitpick: Make it also plural for weight similar to vector queries error?

nafraf: OK. Done.

oshadmi (dismissed):

raz-mon (approved):

nafraf:
This PR is closed because its changes were applied to `feature-RRF` branch.
PR #6487

PR #6351, opened 2025-06-19
[MOD-9801] Migrate numeric encoding/decoding to Rust

chesedo:
## Describe the changes in the pull request
This makes a Rust implementation of the C numeric encoder and decoder used by the inverted index. None of the other encoders/decoders are migrated with this PR

## Benchmark results
Note: inlining was turned off in Rust to be comparable with the C calls

TLDR: The Rust encoding is only slower for small deltas. Since the difference is very small (2 nanoseconds) and the encoding is not as important as the decoding, this is acceptable. The Rust decoder is only slower for the biggest deltas (taking 7 bytes) for only some types like the positive integers with value sizes of 4. This is an edge case as most deltas will be smaller. Therefore this is also acceptable.

### Encoding
The Rust encoder is slower for all types when the delta is zero. But is faster in all other cases.

<details><summary>TinyInt</summary>
<pre><code>
Encode Numeric | Tiny/C/Value size: 0/Delta size: 0
                        time:   [14.969 ns 15.071 ns 15.231 ns]
Found 16 outliers among 100 measurements (16.00%)
  5 (5.00%) low mild
  2 (2.00%) high mild
  9 (9.00%) high severe
Encode Numeric | Tiny/Rust/Value size: 0/Delta size: 0
                        time:   [16.891 ns 16.920 ns 16.963 ns]
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe
</code></pre>

<pre><code>
Encode Numeric | Tiny/C/Value size: 0/Delta size: 4
                        time:   [23.357 ns 23.558 ns 23.930 ns]
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe
Encode Numeric | Tiny/Rust/Value size: 0/Delta size: 4
                        time:   [16.401 ns 16.430 ns 16.465 ns]
Found 20 outliers among 100 measurements (20.00%)
  5 (5.00%) high mild
  15 (15.00%) high severe
</code></pre>

<pre><code>
Encode Numeric | Tiny/C/Value size: 0/Delta size: 7
                        time:   [31.260 ns 31.406 ns 31.582 ns]
Found 13 outliers among 100 measurements (13.00%)
  1 (1.00%) low severe
  3 (3.00%) low mild
  6 (6.00%) high mild
  3 (3.00%) high severe
Encode Numeric | Tiny/Rust/Value size: 0/Delta size: 7
[...]

PR #6378, opened 2025-06-26
[MOD-10286] Implement frequency only encoder/decoder in Rust

gdesmott:

Follow up on https://github.com/RediSearch/RediSearch/pull/6351 by implementing the frequency only encoder and decoder in Rust.

## Benchmarks results

Both Rust encoder and decoder are faster than the C version.

```
Encode - FreqsOnly/C    time:   [299.58 ns 300.40 ns 301.49 ns]
Encode - FreqsOnly/Rust time:   [215.73 ns 216.14 ns 216.70 ns]

Decode - FreqsOnly/C    time:   [10.752 ns 10.787 ns 10.824 ns]
Decode - FreqsOnly/Rust time:   [8.1762 ns 8.1857 ns 8.1950 ns]
```

#### Mark if applicable


src/cndict_loader.c
```diff
@@ -6,7 +6,7 @@
  * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
  * GNU Affero General Public License v3 (AGPLv3).
 */
-#include "buffer.h"
+#include "buffer/buffer.h"
```
chesedo: These should not be needed because you have `${root}/src/buffer` in the include directories in the CMakeLIsts file.

https://github.com/RediSearch/RediSearch/pull/6378/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR211

I know the Rust ffi crate will fail to build. Just add the new header location to its includes section as well

https://github.com/RediSearch/RediSearch/blob/b2b8e5c2c734b94ab87ec661f614a9c1467c902a/src/redisearch_rs/ffi/build.rs#L32-L49

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -121,3 +121,13 @@ impl Buffer {
         self.0.offset += n;
     }
 }
+
+impl std::fmt::Debug for Buffer {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Buffer")
+            .field("len", &self.len())
+            .field("capacity", &self.capacity())
+            .field("data", &self.as_slice())
```
chesedo: Do we want to print all the data in a debug?

gdesmott: In general I only use `Debug` when hacking/debugging so tend to go with "display as much info as available" as it may be useful at some point.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -309,3 +328,90 @@ pub trait Decoder {
...
+        let (decoded_values, _bytes_consumed) = qint_decode::<2, _>(&mut reader).unwrap();
+        let [delta, freq] = decoded_values;
+
+        let record = RSIndexResult::freqs_only(base + delta as u64, freq);
+        Ok(Some(DecoderResult::Record(record)))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
chesedo: We just put tests in a separate file to ensure we only test the public interface of a module. When we want to test a private function / method then we'll use inline tests.

src/redisearch_rs/inverted_index_bencher/benches/numeric.rs
```diff
@@ -19,5 +19,14 @@ fn benchmark_numeric(c: &mut Criterion) {
...
+fn benchmark_freqs_only(c: &mut Criterion) {
+    let bencher = FreqsOnlyBencher::new();
+
+    bencher.encoding(c);
+    bencher.decoding(c);
+}
+
+// FIXME: should we move this to lib.rs? We can only have one criterion_main
+// and I'm not sure it's worth having a sub module for each type of benches.
+criterion_group!(benches, benchmark_numeric, benchmark_freqs_only);
```
chesedo: I was thinking we make a whole new file for each encoder / decoder under `benches` and then we make a section like this for each

https://github.com/RediSearch/RediSearch/blob/b5b49a96835391d68f1dac3d2c8f4939b118d0b3/src/redisearch_rs/inverted_index_bencher/Cargo.toml#L15-L17

src/redisearch_rs/inverted_index_bencher/src/bencher.rs
```diff
@@ -355,3 +355,146 @@ fn numeric_rust_decode<M: Measurement>(group: &mut BenchmarkGroup<'_, M>, input:
...
+        self.rust_decode(&mut group);
+
+        group.finish();
+    }
+
+    fn c_encode<M: Measurement>(&self, group: &mut BenchmarkGroup<'_, M>) {
+        // Use a single buffer big enough to hold all encoded values
+        let buffer_size = self.test_values.iter().map(|test| test.encoded.len()).sum();
+
+        group.bench_function(BenchmarkId::new("C", ""), |b| {
```
chesedo:
```suggestion
        group.bench_function("C", |b| {
```
These are the same if you don't use parameter groups (the second argument)

src/redisearch_rs/buffer/src/lib.rs
chesedo: I'm worried about a 1Kb slice being printed to the terminal if this is accidentally called in prod code. Even worse if it contains sensitive data.

src/redisearch_rs/inverted_index_bencher/benches/numeric.rs
gdesmott: Aren't criterion groups designed for exactly this use case?

I don't really mind but that's gonna add extra boilerplate.

src/redisearch_rs/buffer/src/lib.rs
gdesmott: That's a fair concern. On the other hand I found those quite useful while working on tests.

What about making those `#[cfg(test)]` then?

src/redisearch_rs/inverted_index/src/lib.rs
gdesmott: Right, I meant to move it but forgot.

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -121,3 +121,16 @@ impl Buffer {
...
 }
+
+impl std::fmt::Debug for Buffer {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut debug = f.debug_struct("Buffer");
+        let debug = debug
+            .field("len", &self.len())
+            .field("capacity", &self.capacity());
+        // We don't want to accidentally output huge or sensitive data in production code.
+        #[cfg(test)]
```
LukeMathWalker: Would `debug_assertions` be a better gate?

src/redisearch_rs/inverted_index/src/freqs_only/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+pub struct FreqsOnly;
+
+impl Encoder for FreqsOnly {
+    fn encode<W: Write + Seek>(
+        mut writer: W,
+        delta: Delta,
+        record: &RSIndexResult,
+    ) -> std::io::Result<usize> {
+        // Truncating delta to u32 as qint_encode expects u32 values.
+        // The inverted index will create new block if the delta exceeds u32::MAX.
```
LukeMathWalker: Is this true for all deltas? Or is it true only for some specific encoders?
Can we somehow lift this guarantee up into the type system?

src/redisearch_rs/inverted_index/src/freqs_only/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+        let (decoded_values, _bytes_consumed) = qint_decode::<2, _>(reader).unwrap();
+        let [delta, freq] = decoded_values;
+
+        let record = RSIndexResult::freqs_only(base + delta as u64, freq);
+        Ok(Some(DecoderResult::Record(record)))
+    }
+}
+
+#[cfg(test)]
+mod tests;
```
LukeMathWalker: Since tests only use the public API, move them to a top-level `tests` folder.

src/redisearch_rs/inverted_index/src/freqs_only/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+}
+
+impl Decoder for FreqsOnly {
+    fn decode<R: Read>(
+        &self,
+        reader: &mut R,
+        base: t_docId,
+    ) -> std::io::Result<Option<DecoderResult>> {
+        let (decoded_values, _bytes_consumed) = qint_decode::<2, _>(reader).unwrap();
```
LukeMathWalker:
```suggestion
        let (decoded_values, _bytes_consumed) = qint_decode::<2, _>(reader)?;
```

src/redisearch_rs/inverted_index/src/freqs_only/tests.rs
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use std::io::Cursor;
+
+use crate::{Decoder, DecoderResult, Delta, Encoder, RSIndexResult, freqs_only::FreqsOnly};
+
```
LukeMathWalker: We should test the unhappy case: reader/writer failures, delta bigger than u32::MAX.

src/redisearch_rs/buffer/src/lib.rs
gdesmott: Probably as this could be helpful when debugging the full module directly.

src/redisearch_rs/inverted_index/src/freqs_only/mod.rs
gdesmott: > Is this true for all deltas? Or is it true only for some specific encoders?

It's true for all non-numeric encoders, see https://github.com/RediSearch/RediSearch/blob/master/src/inverted_index/inverted_index.c#L494

```c
 // For non-numeric encoders the maximal delta is UINT32_MAX (since it is encoded with 4 bytes)
 // For numeric encoder the maximal delta has to fit in 7 bytes (since it is encoded with 0-7 bytes)
``` 

I'll make this clearer in the comment.

>  Can we somehow lift this guarantee up into the type system?

We could have `pub struct Delta32(u32);`  and use it in all those encoders to enforce this.

gdesmott: Humm but then we'd need another function `encode32()` in the `Encoder` trait which does not seem great.

Do you have any other idea? If not I think the best option is to document this and enforce it in each of those encoders.

LukeMathWalker: One strategy would be to make `Delta` an associated type of the `Encoder` trait, thus allowing you to specify different delta types for each implementer. But we would have to see how that fits into the broader scheme of things.

gdesmott: I think that may not be very convenient to use but let's see when we start actually using all this code in II.

I changed the code to explicit check for delta overflows and panic accordingly. Should we go with that for now?

LukeMathWalker: Yup, that works for now.

src/inverted_index/inverted_index.h
```diff
@@ -10,7 +10,7 @@
 #define __INVERTED_INDEX_H__
 
 #include "redisearch.h"
-#include "buffer.h"
+#include "buffer/buffer.h"
```
chesedo: Remember to remove this change from all the C files (except for the test files where it is not possible)

gdesmott: I tried actually but iirc that was breaking CI.

But let me double check.

chesedo (dismissed):

src/inverted_index/inverted_index.h
chesedo: If it still breaks then it might be some `build.rs` that is missing an update

gdesmott: Indeed, I had to edit `build.rs` as well: https://github.com/RediSearch/RediSearch/pull/6378/commits/6deeb1341853d6976e0457c1f14946cc257107f0

gdesmott:
Rebased on top of  `master` now that #6351 is merged and fixed to fit the latest API changes.

src/redisearch_rs/buffer/src/lib.rs
```diff
@@ -121,3 +121,16 @@ impl Buffer {
...
+impl std::fmt::Debug for Buffer {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut debug = f.debug_struct("Buffer");
+        let debug = debug
+            .field("len", &self.len())
+            .field("capacity", &self.capacity());
+        // We don't want to accidentally output huge or sensitive data in production code.
+        #[cfg(debug_assertions)]
+        let debug = debug.field("data", &self.as_slice());
+        debug.finish()
```
GuyAv46, line 134: This syntax means we assert if we reach here from a release build?

src/redisearch_rs/inverted_index_bencher/src/ffi.rs
```diff
@@ -154,4 +182,45 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_encode_freqs_only() {
+        // Test cases for the frequencies only encoder and decoder. These cases can be moved to the Rust
+        // implementation tests verbatim.
+        let tests = [
+            // (frequency, delta, expected encoding)
+            (0, 0, vec![0, 0, 0]),
+            (0, 1, vec![0, 1, 0]),
```
GuyAv46, line 193: Not sure if it matters, but I don't think `frequency 0` is possible (why do we add this document to this inverted index if it doesn't have this term?)

src/redisearch_rs/inverted_index/tests/freqs_only.rs
```diff
@@ -0,0 +1,113 @@
...
+    Decoder, DecoderResult, Delta, Encoder, RSIndexResult, freqs_only::FreqsOnly,
+};
+
+#[test]
+fn test_encode_freqs_only() {
+    // Test cases for the frequencies only encoder and decoder.
+    let tests = [
+        // (frequency, delta, expected encoding)
+        (0, 0, vec![0, 0, 0]),
+        (0, 1, vec![0, 1, 0]),
```
GuyAv46, line 22: I don't think we should expect frequency 0, but I don't see any harm in testing it

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -38,6 +39,15 @@ impl From<Delta> for usize {
...
 }
 
+// Non-numeric encoders only accept deltas that fit in a `u32`.
+impl TryFrom<Delta> for u32 {
+    type Error = std::num::TryFromIntError;
+
+    fn try_from(delta: Delta) -> Result<Self, Self::Error> {
+        u32::try_from(delta.0)
+    }
+}
```
GuyAv46: Shouldn't this be covered by a test?

src/inverted_index/inverted_index.c
```diff
@@ -373,6 +373,11 @@ ENCODER(encodeNumeric) {
   return sz;
 }
 
+// Wrapper around the private static `encodeFreqsOnly` function to expose it to benchmarking.
+size_t encode_freqs_only(BufferWriter *bw, t_docId delta, RSIndexResult *res) {
+  encodeFreqsOnly(bw, delta, res);
```
chesedo:
```suggestion
  return encodeFreqsOnly(bw, delta, res);
```

src/redisearch_rs/inverted_index/src/freqs_only/mod.rs
chesedo, line 1: I ended up having the numeric one as `numeric.rs`. It just did not seem like anything else will be under its module

https://github.com/RediSearch/RediSearch/blob/9948b8da5a927fa05ee9948e6c68843d92026814/src/redisearch_rs/inverted_index/src/numeric.rs

src/redisearch_rs/inverted_index/src/freqs_only/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+*/
+
+use std::io::{Read, Seek, Write};
+
+use ffi::t_docId;
+use qint::{qint_decode, qint_encode};
+
+use crate::{Decoder, DecoderResult, Delta, Encoder, RSIndexResult};
+
+/// Encode and decode only the delta ID and frequencies of a record, without any other data.
```
chesedo:
```suggestion
/// Encode and decode only the delta and frequencies of a record, without any other data.
```

src/redisearch_rs/buffer/src/lib.rs
gdesmott: No, see it like a `#ifdef` on this single line.

That means the data field will only be present in debug builds but not in release ones.

src/inverted_index/inverted_index.c
gdesmott: My fix got lost during the rebase, thanks!

src/redisearch_rs/buffer/src/lib.rs
gdesmott: The rationale for this is to prevent accidentally printing huge and/or sensitive data in release builds, see https://github.com/RediSearch/RediSearch/pull/6378#discussion_r2169128764

src/redisearch_rs/inverted_index_bencher/src/ffi.rs
gdesmott: Indeed, that makes sense. :)

So we can either keep it as it and add a comment explaining that actual records should have a frequency higher than 1.
Or we can enforce it, either by panicing if the freq is 0 or even enforce it  through the API (`NonZeroU32`). What do you think?

src/redisearch_rs/inverted_index/src/lib.rs
gdesmott: It is here: https://github.com/RediSearch/RediSearch/blob/rs_ii_enc_dec/src/redisearch_rs/inverted_index/tests/freqs_only.rs#L107

src/redisearch_rs/inverted_index/src/freqs_only/mod.rs
gdesmott: Makes sense.

gdesmott: Actually I'm wondering if it would not make more sense to have all those encoders and decoders in specific modules for proper namespacing.

`encoder::FreqsOnly` seems a clearer API than just `FreqsOnly`.

chesedo: We might. I know we at least don't want them all in the same file since the numeric one is huge. I know the encoders & decoders are currently implemented for the same struct which will make an `encoder::*` & `decoder::*` namespace hard (where will the base struct go). But I suspect we'll have to base the decoders of a new struct soonish (the II will need each decoder to have a different context for that field masks)

gdesmott: Let's keep it as it for now then and reconsider once your progress with your II work.

src/redisearch_rs/inverted_index/src/lib.rs
GuyAv46: Codecov reports it as uncovered, so we should look into it (out of scope for this PR)

chesedo (approved):

src/redisearch_rs/inverted_index/src/lib.rs
gdesmott: Actually the coverage is right. This test checks [this code path](https://github.com/RediSearch/RediSearch/blob/rs_ii_enc_dec/src/redisearch_rs/inverted_index/src/freqs_only.rs#L32) not this code.

I double checked and we don't need this `TryFrom`. We may need something like this if we enforce the two Delta sizes in the API but we agreed to defer that for now (see https://github.com/RediSearch/RediSearch/pull/6378#discussion_r2174848606 ).

So I just removed this code for now.

gdesmott:
> We need to understand why the coverage report misses some of the functions

Actually the coverage report is correct: https://github.com/RediSearch/RediSearch/pull/6378#discussion_r2189158791

chesedo (approved):

PR #6383, opened 2025-06-27
MOD-10359  Implement shard window ratio optimization

benoitdion:
This PR introduces a new **shard window ratio optimization** feature for RediSearch vector search operations, designed to reduce coordinator workload in cluster deployments by allowing shards to return fewer results than the requested K value.

---

### **🎯 What This PR Does**

**Core Feature**: Adds support for the `$SHARD_K_RATIO` parameter in KNN vector queries, which optimizes cluster performance by reducing the number of results each shard needs to process and return to the coordinator.

#### **Query Syntax**
```
*=>[KNN 10 @vector $query_vec]=>{$SHARD_K_RATIO: 0.5}
```

**Formula**: `effective_k_per_shard = max(top_k / num_shards, ceil(top_k × ratio))`

This ensures:
- **Minimum guarantee**: Each shard returns at least `top_k / num_shards` results
- **Optimization**: When `ceil(top_k × ratio) > top_k / num_shards`, use the larger value
- **Correctness**: The coordinator can still return exactly K results to the user

**Example**: With K=100, ratio=0.3, and 3 shards:
- **Before**: Each shard returns 100 results (300 total to coordinator)
- **After**: Each shard returns 34 results (102 total to coordinator) 
- **User gets**: Exactly 100 results as requested
---

#### **New Files Added**
- `src/shard_window_ratio.h` - Core optimization logic and command modification functions
- `src/shard_window_ratio.c` - Implementation of KNN command modification for shards
- `tests/ctests/coord_tests/test_shard_window_ratio.c` - C unit tests for parameter validation
- `tests/pytests/test_shard_window_ratio.py` - Python integration tests

#### **Core Implementation**
- **Query Parser Enhancement**: Extended both v1 and v2 parsers to recognize `$SHARD_K_RATIO` parameter
- **Coordinator Logic**: Added effective K calculation and command modification for shard distribution
- **Parameter Validation**: Validates ratio must be > 0 and ≤ 1.0
- **Command Support**: Works with both `FT.SEARCH` and `FT.AGGREGATE` operations

---

#### **Deployment Behavior**
- **Cluster Mode**: Optimization active - shards process `effective_k`, coordinator returns exactly K results
- **Standalone Mode**: Parameter ignored - always returns K results (no optimization needed)

#### **Parameter Handling**
- **Literal K**: `KNN 50` - modifies the literal value in query string for shards
  Example: 
```
"KNN 50 @vec $blob" → "KNN 25 @vec $blob"
```
- **Parameter K**: `KNN $k` - replaces parameter reference with literal value in query string
  Example: 
```
"KNN $k @vec $blob" → "KNN 25 @vec $blob"
```

#### **Validation**
- Ratio must be > 0.0 and ≤ 1.0
- Invalid ratios return descriptive error messages
- Backward compatibility maintained - existing queries work unchanged

---

#### **Unit Tests (C)**
- Parameter validation for valid/invalid ratio values
- Attribute name case sensitivity testing
- Error message validation
- Backward compatibility verification
- Multiple attribute handling

#### **Integration Tests (Python)**
- End-to-end functionality with `FT.SEARCH` and `FT.AGGREGATE`
- Profile validation to ensure correct shard behavior
- Parameter vs literal K value handling
- Hybrid query support
- Edge cases (insufficient documents per shard)
- Both cluster and standalone mode testing

### **🚀 Usage Examples**

```bash
# Basic usage with 50% ratio
FT.SEARCH idx "*=>[KNN 10 @vec $query]=>{$SHARD_K_RATIO: 0.5}" PARAMS 2 query <vector_blob>

# With additional vector parameters
FT.SEARCH idx "*=>[KNN 20 @vec $query]=>{$SHARD_K_RATIO: 0.3; $YIELD_DISTANCE_AS: dist}" PARAMS 2 query <vector_blob>

# Using parameter-based K value
FT.SEARCH idx "*=>[KNN $k @vec $query]=>{$SHARD_K_RATIO: 0.4}" PARAMS 4 k 15 query <vector_blob>

# Works with FT.AGGREGATE too
FT.AGGREGATE idx "*=>[KNN 10 @vec $query]=>{$SHARD_K_RATIO: 0.6}" PARAMS 2 query <vector_blob>
```


src/module.h
```diff
@@ -136,6 +138,18 @@ bool debugCommandsEnabled(RedisModuleCtx *ctx);
 specialCaseCtx *prepareOptionalTopKCase(const char *query_string, RedisModuleString **argv, int argc, uint dialectVersion,
                              QueryError *status);
 
+/**
+ * Get the effective K value for a KNN context based on whether this is a shard command.
+ * For shard commands, applies shard window ratio optimization.
+ * For coordinator commands, returns original K.
+ */
+static inline size_t getEffectiveK(const knnContext *ctx, bool isShardCommand) {
+  if (!isShardCommand || ctx->shardWindowRatio >= 1.0) {
```
benoitdion: `ctx->shardWindowRatio >= 1.0` should never happen if we validate earlier in the pipeline?

src/shard_window_ratio.h
```diff
@@ -0,0 +1,78 @@
...
+ * @return true if this is a shard-level command, false otherwise
+ */
+static inline bool isShardLevelCommand(RedisModuleString **argv, int argc) {
+  if (argc == 0 || argv == NULL) {
+    return false;
+  }
+
+  size_t cmdLen;
+  const char *cmdName = RedisModule_StringPtrLen(argv[0], &cmdLen);
+  return (cmdLen > 3 && cmdName[0] == '_' && strncasecmp(cmdName + 1, "FT.", 3) == 0);
```
benoitdion: is there a better way?

src/module.c
```diff
@@ -1693,6 +1695,9 @@ void SpecialCaseCtx_Free(specialCaseCtx* ctx) {
   if (!ctx) return;
   if(ctx->specialCaseType == SPECIAL_CASE_KNN) {
     QueryNode_Free(ctx->knn.queryNode);
+    if (ctx->knn.pq) {
+      heap_destroy(ctx->knn.pq);
```
benoitdion: unrelated to the core of this PR, potential memory leak fix

src/shard_window_ratio.h
oshadmi: @raz-mon Is there some API to detect whether a command is an internal command (starting from redis>= 8.0)?

raz-mon: It's currently checked similarly [here](https://github.com/RediSearch/RediSearch/blob/b1332489e7dde8535c3774201e4ad1652b4f08d8/src/aggregate/aggregate_exec.c#L937-L938), and a flag is turned on.
But the term "internal" is abused a bit, since we have underscored commands that are not internal (e.g., `_FT.CONFIG`) in the sense that they are not disabled for non-internal connections.
Anyways, it seems that the check here is for shard commands rather than internal commands, for which we have the `QEXEC_F_INTERNAL` flag but no dedicated API (would be nice to have).

tests/pytests/test_shard_window_ratio.py
```diff
@@ -0,0 +1,439 @@
...
+                       message=f"With K={k}, ratio={ratio}: expected {expected_total} results ({mode_desc}), got {actual_result_count}")
+
+        # Validate shard result processing for this scenario
+        profile_dict = to_dict(profile_info)
+        _validate_individual_shard_results(env, profile_dict, scenario)
+
+
+
+def test_ft_aggregate_shard_window_ratio_profile():
+    """Test FT.AGGREGATE with shard window ratio and profile metrics"""
```
meiravgri: what about ft.search?

tests/pytests/test_shard_window_ratio.py
```diff
@@ -0,0 +1,439 @@
...
+    query_vec = create_np_array_typed([50] * dim, 'FLOAT32')
+
+    # Test scenarios with different characteristics
+    # Expected results depend on cluster vs standalone mode
+    scenarios = [
+        {
+            'k': 3,
+            'ratio': 0.3,
+            'expected_effective_k': 1,
+            'expected_total_results_cluster': 3,  # User always gets K results in cluster
```
meiravgri: probably will have to set it later as it depends on a previous entry of this dict
```suggestion
            'expected_total_results_cluster': expected_effective_k * num_shards,  # User always gets K results in cluster
```

src/shard_window_ratio.h
```diff
@@ -0,0 +1,78 @@
...
+ * - If ratio is valid (0.0 < ratio < 1.0), applies optimization
+ * - Ensures result is at least 1
+ *
+ * @param originalK The original K value requested
+ * @param ratio The shard window ratio (any value, function handles validation)
+ * @return Effective K value (originalK if no optimization, calculated value otherwise)
+ */
+static inline size_t calculateEffectiveK(size_t originalK, double ratio) {
+  // No optimization if ratio is invalid or >= 1.0
+  if (ratio <= 0.0 || ratio >= 1.0) {
```
meiravgri: why would numShards be 0

src/module.c
```diff
@@ -1771,6 +1778,8 @@ void setKNNSpecialCase(searchRequestCtx *req, specialCaseCtx *knn_ctx) {
 // was found).
 specialCaseCtx *prepareOptionalTopKCase(const char *query_string, RedisModuleString **argv, int argc, uint dialectVersion,
                                         QueryError *status) {
+  // Automatically detect if this is a shard-level command (_FT.*) vs coordinator command (FT.*)
+  bool isShardCommand = isShardLevelCommand(argv, argc);
```
meiravgri: it is not used here

src/module.c
```diff
@@ -1825,7 +1834,9 @@ specialCaseCtx *prepareOptionalTopKCase(const char *query_string, RedisModuleStr
       goto cleanup;
     }
     specialCaseCtx *ctx = SpecialCaseCtx_New();
-    ctx->knn.k = k;
+    ctx->knn.originalK = k;  // Always store original K
+    ctx->knn.shardWindowRatio = queryVectorNode.vq->shardWindowRatio;  // Store ratio for later calculation
```
meiravgri:
```suggestion
    ctx->knn.shardKRatio = queryVectorNode.vq->shardKRatio;  // Store ratio for later calculation
```

src/module.c
meiravgri: Although searchRequestCtx_Free->SpecialCaseCtx_Free is called in rscParseRequest, the heap is only created in `searchResultReducer` (later in the query flow) and also released there:

```
  if (rCtx.pq) {
    heap_destroy(rCtx.pq);
  }
  if (rCtx.reduceSpecialCaseCtxKnn &&
      rCtx.reduceSpecialCaseCtxKnn->knn.pq) {
    heap_destroy(rCtx.reduceSpecialCaseCtxKnn->knn.pq);
    rCtx.reduceSpecialCaseCtxKnn->knn.pq = NULL;
  }
```
I am removing this to prevent coverage hits.

src/module.c
```diff
@@ -120,7 +122,7 @@ bool ACLUserMayAccessIndex(RedisModuleCtx *ctx, IndexSpec *sp) {
   bool ret = true;
   HiddenUnicodeString **prefixes = sp->rule->prefixes;
   RedisModuleString *prefix;
-  for (uint i = 0; i < array_len(prefixes); i++) {
+  for (unsigned int i = 0; i < array_len(prefixes); i++) {
```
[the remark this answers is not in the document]
meiravgri: thought i reverted it

src/module.c
```diff
@@ -2102,7 +2107,7 @@ searchResult *newResult_resp3(searchResult *cached, MRReply *results, int j, sea
   }
 
   MRReply *result_id = MRReply_MapElement(result_j, "id");
-  if (!result_id || !MRReply_Type(result_id) == MR_REPLY_STRING) {
+  if (!result_id || MRReply_Type(result_id) != MR_REPLY_STRING) {
```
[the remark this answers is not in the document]
meiravgri: i think it's clearer

src/coord/dist_aggregate.c
```diff
@@ -17,7 +17,9 @@
 #include "util/timeout.h"
 #include "resp3.h"
 #include "coord/config.h"
+#include "config.h"  // For MIN_SHARD_WINDOW_RATIO and MAX_SHARD_WINDOW_RATIO constants
```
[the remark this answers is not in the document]
meiravgri: it does compile without this include but it i snot included directly from shard_window_ratio.h
it is required for MAX_SHARD_WINDOW_RATIO
IMO healtier to include all required headers rather than relying on existing headers to indirectly include it, that might be removed in the future

src/query_node.h
```diff
@@ -126,7 +126,7 @@ typedef struct {
   struct GeometryQuery *geomq;
 } QueryGeometryNode;
 
-typedef struct {
+typedef struct {// TODO: consider remove this struct, not in use
   struct VectorQuery *vq;
 } QueryVectorNode;
```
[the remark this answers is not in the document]
meiravgri: i see all other query node types do that... :(

src/coord/rmr/command.c
```diff
@@ -175,6 +175,54 @@ void MRCommand_ReplaceArg(MRCommand *cmd, int index, const char *newArg, size_t
...
+    return;
+  }
+
+  char *oldArg = cmd->strs[index];
+  size_t oldArgLen = cmd->lens[index];
+
+  // Validate position and length
+  if (pos + oldLen > oldArgLen) {
+    return;
+  }
```
GuyAv46: Consider asserting this instead of checking

src/coord/rmr/command.c
```diff
@@ -175,6 +175,54 @@ void MRCommand_ReplaceArg(MRCommand *cmd, int index, const char *newArg, size_t
   MRCommand_ReplaceArgNoDup(cmd, index, news, len);
 }
 
+void MRCommand_ReplaceArgSubstring(MRCommand *cmd, int index, size_t pos, size_t oldLen, const char *newStr, size_t newLen) {
+  if (index < 0 || index >= cmd->num) {
+    return;
+  }
```
GuyAv46: Consider asserting this instead of checking

src/coord/rmr/command.h
```diff
@@ -78,6 +78,10 @@ void MRCommand_SetPrefix(MRCommand *cmd, const char *newPrefix);
 void MRCommand_ReplaceArg(MRCommand *cmd, int index, const char *newArg, size_t len);
 void MRCommand_ReplaceArgNoDup(MRCommand *cmd, int index, const char *newArg, size_t len);
 
+/** Replace a substring within an argument at a specific position
+ * OPTIMIZATION: Avoids reallocation when new string is same/shorter length */
+void MRCommand_ReplaceArgSubstring(MRCommand *cmd, int index, size_t pos, size_t oldLen, const char *newStr, size_t newLen);
```
GuyAv46: Let's improve the comment here. It's not clear what each parameter is supposed to be

src/coord/rmr/command.c
```diff
@@ -175,6 +175,54 @@ void MRCommand_ReplaceArg(MRCommand *cmd, int index, const char *newArg, size_t
...
+  memcpy(newArg, oldArg, pos);                           // Copy before
+  memcpy(newArg + pos, newStr, newLen);                  // Copy new substring
+  memcpy(newArg + pos + newLen, oldArg + pos + oldLen,   // Copy after
+         oldArgLen - pos - oldLen);
+
+  newArg[newArgLen] = '\0';
+
+  // Replace the argument
+  MRCommand_ReplaceArgNoDup(cmd, index, newArg, newArgLen);
+}
```
GuyAv46: We should cover this fallback with a test

src/module.c
```diff
@@ -3290,6 +3293,30 @@ static int prepareCommand(MRCommand *cmd, searchRequestCtx *req, RedisModuleBloc
...
+  // Handle KNN with shard ratio optimization for both multi-shard and standalone
+  if (req->specialCases) {
+    for (size_t i = 0; i < array_len(req->specialCases); ++i) {
+      if (req->specialCases[i]->specialCaseType == SPECIAL_CASE_KNN) {
+        specialCaseCtx* knnCtx = req->specialCases[i];
+        KNNVectorQuery *knn_query = &knnCtx->knn.queryNode->vn.vq->knn;
+        double ratio = knn_query->shardWindowRatio;
+
+        // Apply optimization only if ratio is valid and < 1.0 (ratio = 1.0 means no optimization)
+        if (ratio < 1.0) {
```
GuyAv46:
```suggestion
        if (ratio < MAX_SHARD_WINDOW_RATIO) {
```

src/module.c
```diff
@@ -3290,6 +3293,30 @@ static int prepareCommand(MRCommand *cmd, searchRequestCtx *req, RedisModuleBloc
...
+          // Calculate effective K based on deployment mode
+          size_t effectiveK = calculateEffectiveK(knn_query->k, ratio, NumShards);
+
+          // Modify the command to replace KNN k (shards will ignore $SHARD_K_RATIO)
+          if (modifyKNNCommand(cmd, knn_query->k, effectiveK, &knnCtx->knn) != 0) {
+            QueryError_SetError(status, QUERY_EGENERIC, "Failed to modify KNN command for shard distribution");
+            return REDISMODULE_ERR;
+          }
+        }
+        break; // Only handle first KNN context
```
GuyAv46: A bit misleading - what other KNN context can be found? We only support a single KNN per query

src/coord/rmr/command.c
```diff
@@ -175,6 +175,54 @@ void MRCommand_ReplaceArg(MRCommand *cmd, int index, const char *newArg, size_t
...
+  }
+
+  // Calculate new total length
+  size_t newArgLen = oldArgLen - oldLen + newLen;
+
+  // OPTIMIZATION: If new string is same or shorter, modify in-place to avoid reallocation
+  if (newArgLen <= oldArgLen) {
+    // Move the "after" part to make room for new substring
+    if (newLen != oldLen) {
+      memmove(oldArg + pos + newLen, oldArg + pos + oldLen, oldArgLen - pos - oldLen);
```
GuyAv46: Since the use case is for literals in the query string, a better performing option could be padding the remaining characters with spaces instead of moving the entire query string

src/shard_window_ratio.h
```diff
@@ -0,0 +1,73 @@
...
+
+  return effectiveK > 0 ? effectiveK : 1;  // Ensure at least 1 result
+}
+
+/**
+ * Modify KNN command for shard distribution by replacing K value.
+ *
+ * This function handles two cases:
+ * 1. Literal K (e.g., "KNN 50") - uses saved position for exact replacement
+ * 2. Parameter K (e.g., "KNN $k") - modifies parameter value in PARAMS section
```
GuyAv46: Not sure if this is safe - what if this value is being used for another purpose in the query?
Consider creating a new parameter instead (and modify the query to use it). Alternatively, we can replace the parameter with the new literal value

src/shard_window_ratio.h
```diff
@@ -0,0 +1,73 @@
...
+ * - Optimization: If ceil(top_k × ratio) > top_k/#shards, use the larger value
+ *
+ * @param originalK The original K value requested
+ * @param ratio The shard window ratio (any value, function handles validation)
+ * @param numShards The number of shards in the cluster
+ * @return Effective K value per shard
+ */
+static inline size_t calculateEffectiveK(size_t originalK, double ratio, size_t numShards) {
+  // No optimization if ratio is invalid or > 1.0, or if numShards is 0
+  if (ratio <= 0.0 || ratio > 1.0 || numShards == 0) {
```
GuyAv46: Consider asserting.
For the ratio validations, we'd better use the macros and not `0.0` and `1.0`.
For `numShards`, I don't think it can be 0 if we got here, so an assertion is better

src/shard_window_ratio.h
```diff
@@ -0,0 +1,73 @@
...
+    return originalK;
+  }
+
+  // Calculate minimum K per shard to ensure we can return full originalK results
+  // Use ceiling division: (originalK + numShards - 1) / numShards
+  size_t minKPerShard = (originalK + numShards - 1) / numShards;
```
GuyAv46, line 51: We recently added a macro `CEIL_DIV` for that
```suggestion
  size_t minKPerShard = CEIL_DIV(originalK, numShards);
```

src/shard_window_ratio.h
```diff
@@ -0,0 +1,73 @@
...
+
+  // Calculate ratio-based K per shard
+  double exactRatioK = (double)originalK * ratio;
+  size_t ratioKPerShard = (size_t)ceil(exactRatioK);
+
+  // Apply PRD formula: max(top_k/#shards, ceil(top_k × ratio))
+  size_t effectiveK = (ratioKPerShard > minKPerShard) ? ratioKPerShard : minKPerShard;
+
+  return effectiveK > 0 ? effectiveK : 1;  // Ensure at least 1 result
```
GuyAv46: `minKPerShard` is already greater than 0, so the last validation is redundant

tests/ctests/coord_tests/test_shard_window_ratio.c
```diff
@@ -0,0 +1,351 @@
...
+#include "shard_window_ratio.h"
+#include <string.h>
+#include <math.h>
+
+// Test helper functions
+static QueryNode* createTestVectorNode() {
+    QueryNode* node = rm_calloc(1, sizeof(QueryNode));
+    node->type = QN_VECTOR;
+    node->vn.vq = rm_calloc(1, sizeof(VectorQuery));
+    node->vn.vq->knn.shardWindowRatio = 1.0; // Initialize with DEFAULT_SHARD_WINDOW_RATIO
```
GuyAv46: So let's do exactly that, or the right macro
```suggestion
    node->vn.vq->knn.shardWindowRatio = DEFAULT_SHARD_WINDOW_RATIO;
```

tests/pytests/test_shard_window_ratio.py
```diff
@@ -0,0 +1,268 @@
...
+    conn = getConnectionByEnv(env)
+
+    # Add test documents - ensure we have enough for all test scenarios
+    p = conn.pipeline(transaction=False)
+    for i in range(1, num_docs + 1):
+        vector = get_unique_vector(dim, datatype)
+        p.execute_command('HSET', f'doc{i}', 'v', vector.tobytes())
+    p.execute()
+
+    waitForIndex(env, 'idx')
```
GuyAv46: redundant when we first set the schema and then the documents

tests/pytests/test_shard_window_ratio.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+        for malformed_query in malformed_queries:
+            res = env.expect(cmd, 'idx', malformed_query['query'],
+                    'PARAMS', 2, 'query_vec', query_vec.tobytes(), 'nocontent')
+            ValidateError(env, res,
+                        message=f"{cmd} expected error for {malformed_query['error']} in: {malformed_query['query']}",
+                        expected_error_message=f"Syntax error")
+
+def test_ft_profile_shard_result_validation_scenarios():
```
GuyAv46, line 106: Relevant for standalone?

src/shard_window_ratio.c
```diff
@@ -0,0 +1,110 @@
...
+ * @return 0 on success, -1 if parameter not found
+ */
+static int modifyParameterKInParams(MRCommand *cmd, const char *paramName, size_t effectiveK) {
+    // Scan command arguments to find parameter name
+    for (size_t i = 0; i < cmd->num - 1; i++) {  // -1 because we check i+1
+        if (strcmp(cmd->strs[i], paramName) == 0) {
+            // Found parameter name, value is at next index
+            char effectiveK_str[32];
+            snprintf(effectiveK_str, sizeof(effectiveK_str), "%zu", effectiveK);
+            MRCommand_ReplaceArg(cmd, i + 1, effectiveK_str, strlen(effectiveK_str));
```
GuyAv46: `snprintf` returns the string length
```suggestion
            size_t len = snprintf(effectiveK_str, sizeof(effectiveK_str), "%zu", effectiveK);
            MRCommand_ReplaceArg(cmd, i + 1, effectiveK_str, len);
```

src/shard_window_ratio.c
```diff
@@ -0,0 +1,110 @@
...
+        if (queryStr && strstr(queryStr, "KNN")) {
+
+            // Get saved position information
+            size_t k_pos = knnNode->vn.vq->knn.k_literal_pos;
+            size_t k_len = knnNode->vn.vq->knn.k_literal_len;
+
+            // Verify position is valid and K value matches
+            size_t queryLen;
+            MRCommand_ArgStringPtrLen(cmd, i, &queryLen);
+            if (k_pos > 0 && k_pos + k_len <= queryLen) {
```
GuyAv46: Can it be false? Consider asserting

src/shard_window_ratio.c
```diff
@@ -0,0 +1,110 @@
...
+    // Find the query string in the command arguments
+    for (size_t i = 0; i < cmd->num; i++) {
+        const char* queryStr = MRCommand_ArgStringPtrLen(cmd, i, NULL);
+        if (queryStr && strstr(queryStr, "KNN")) {
+
+            // Get saved position information
+            size_t k_pos = knnNode->vn.vq->knn.k_literal_pos;
+            size_t k_len = knnNode->vn.vq->knn.k_literal_len;
+
+            // Verify position is valid and K value matches
```
GuyAv46: why?

src/shard_window_ratio.c
```diff
@@ -0,0 +1,110 @@
...
+            size_t queryLen;
+            MRCommand_ArgStringPtrLen(cmd, i, &queryLen);
+            if (k_pos > 0 && k_pos + k_len <= queryLen) {
+                char originalK_str[32];
+                snprintf(originalK_str, sizeof(originalK_str), "%zu", originalK);
+
+                if (strncmp(queryStr + k_pos, originalK_str, k_len) == 0) {
```
GuyAv46: Why do we need to validate this? Also, using `strncmp` is a bad practice if we are not sure the lengths are equal (unless we intentionally want to check if one is a prefix of the other)

src/shard_window_ratio.c
```diff
@@ -0,0 +1,110 @@
...
+                    // Use efficient substring replacement - no need to copy entire query!
+                    char effectiveK_str[32];
+                    snprintf(effectiveK_str, sizeof(effectiveK_str), "%zu", effectiveK);
+                    size_t newK_len = strlen(effectiveK_str);
```
GuyAv46:
```suggestion
                    size_t newK_len = snprintf(effectiveK_str, sizeof(effectiveK_str), "%zu", effectiveK);
```

src/shard_window_ratio.c
```diff
@@ -0,0 +1,110 @@
...
+
+// Note: We don't remove $SHARD_K_RATIO from query strings
+// as shards can safely ignore this attribute as a no-op
+
+int modifyKNNCommand(MRCommand *cmd, size_t originalK, size_t effectiveK, knnContext *knnCtx) {
+    // Fast path: No modification needed if K values are the same
+    QueryNode *knnNode = knnCtx->queryNode;
+    if (!cmd || !knnNode || originalK == effectiveK) {
+        return 0;
+    }
```
GuyAv46: Relevant? Consider asserting

GuyAv46 (changes requested):

src/coord/rmr/command.c
meiravgri: probably not relevant for the actual (and cuurently - only) usage, as the new k is equal or smaller, but you are right that it should be covered

tests/pytests/test_shard_window_ratio.py
meiravgri: Yes 
It's a valid input (and also used by the coordinator to distribute the command to the shards), so these tests validate it doesn't break the current behaviour.

tests/pytests/test_shard_window_ratio.py
meiravgri: IBA (i blame augment)

tests/ctests/coord_tests/test_shard_window_ratio.c
meiravgri: lets

src/shard_window_ratio.h
meiravgri: Nice catch!

src/shard_window_ratio.h
meiravgri: is k = 0 a valid input?

src/shard_window_ratio.h
meiravgri: only defined in  .c files

src/coord/rmr/command.c
meiravgri: done

src/module.c
meiravgri: you are reffering to the comment, am i right?
we indeed should stop the loop after we found the knn context

GuyAv46: Yes

src/shard_window_ratio.h
GuyAv46: Hmm, yes, actually. How are we handling this case?

meiravgri: no special handling in this pr

GuyAv46: Consider adding a test. Seems like this function will correctly return 0 in this case, if we reach here

meiravgri: already did :) see test_shrad_window_ratio.c and also new test in test_shrad_window_ratio.py
also note i changed this line to simply return effectiveK

GuyAv46: Noted :)

src/module.c
```diff
@@ -2892,6 +2894,7 @@ static int searchResultReducer(struct MRCtx *mc, int count, MRReply **replies) {
   if (rCtx.reduceSpecialCaseCtxKnn &&
       rCtx.reduceSpecialCaseCtxKnn->knn.pq) {
     heap_destroy(rCtx.reduceSpecialCaseCtxKnn->knn.pq);
+    rCtx.reduceSpecialCaseCtxKnn->knn.pq = NULL;
```
alonre24, line 2897: Is that needed due to the changes, or better practice regardless?

src/shard_window_ratio.c
```diff
@@ -0,0 +1,35 @@
...
+    // Fast path: No modification needed if K values are the same
+    if (originalK == effectiveK) {
+        return;
+    }
+    // Get saved position information
+    size_t k_pos = vq->knn.k_token_pos;
+    size_t k_len = vq->knn.k_token_len;
+
+    char effectiveK_str[32];
+    size_t newK_len = snprintf(effectiveK_str, sizeof(effectiveK_str), "%zu", effectiveK);
```
alonre24, line 31: We know for sure that there will be no overflow when writing to `effectiveK_str`?

src/special_case_ctx.h
```diff
@@ -0,0 +1,41 @@
+/*
```
alonre24: This file should go under coord dir

src/vector_index.h
```diff
@@ -93,6 +94,13 @@ typedef struct {
   size_t vecLen;                 // vector length
   size_t k;                      // number of vectors to return
   VecSimQueryReply_Order order;  // specify the result order.
+  double shardWindowRatio;       // shard window ratio for distributed queries
+
+  // Position tracking for K value modification (shard ratio optimization)
+  // For literal K (e.g., "KNN 10"): stores position and length of numeric value
+  // For parameter K (e.g., "KNN $k"): stores position and length INCLUDING the '$' prefix
+  size_t k_token_pos;            // Byte offset where K token starts in original query
+  size_t k_token_len;            // Length of K token
```
alonre24, line 103: Can you explain why that is needed?

alonre24 (commented):
reviewed code. todo: tests

src/shard_window_ratio.c
meiravgri: as long as size_t = 2^64, yes

src/module.c
meiravgri: better practice to avoid potential double free

src/vector_index.h
meiravgri: The 3 lines above aren't sufficient?

src/module.c
meiravgri: not related to this pr

PR #6388, opened 2025-06-29
MOD-10250: RRF RESP3 format response

raz-mon:
Two main contributions:
1. `serializeResult_hybrid`: A `SearchResult` serializer matching the needs of the new `FT.HYBRID` command response. This serializer maintains a consistent format, regardless of the RESP protocol used by the client, i.e., uses the RESP3 API only, such that the fallbacks defined in Redis are used in case RESP2 is used by the current client.
2. `sendChunk_hybrid`: A new pipeline activation function, which creates the client response and dispatches it, also in a RESP-agnostic format. It uses `serializeResult_hybrid` to serialize the results received from the pipeline.

Note 1: This PR is a base PR to be used later on when the `FT.HYBRID` command handler is implemented, after the pipeline building logic is added. When it will be used it will be properly tested, and probably tweaked here and there. The counterpart in the coordinator will be added once the cluster solution is implemented.

Note 2: Many options that are available in `FT.SEARCH`, `FT.AGGREGATE` are not supported for the `FT.HYBRID` command (at this point), e.g., profiling, cursors, and many arguments like `WITHPAYLOAD` etc.
In order not to maintain unused code, we start without support for these options at all, and add support for whatever is needed/wanted instead.

Other changes:
1. Some touchups to the existing reply-related code, mainly changing to use the RESP3->2 fallbacks defined in Redis, and removing unnecessary labels.

### DIFF
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -144,7 +144,7 @@ static size_t serializeResult(AREQ *req, RedisModule_Reply *reply, const SearchR
       RedisModule_Reply_Array(reply);
       RedisModule_Reply_Double(reply, r->score);
       SEReply(reply, r->scoreExplain);
-  	  RedisModule_Reply_ArrayEnd(reply);
+      RedisModule_Reply_ArrayEnd(reply);
     }
   }
 


src/aggregate/aggregate_exec.c
```diff
@@ -281,6 +281,118 @@ static size_t serializeResult(AREQ *req, RedisModule_Reply *reply, const SearchR
...
+  if (options & QEXEC_F_SEND_SCORES) {
+    RedisModule_Reply_SimpleString(reply, "score");
+    if (!(options & QEXEC_F_SEND_SCOREEXPLAIN)) {
+      // This will become a string in RESP2
+      RedisModule_Reply_Double(reply, r->score);
+    } else {
+      RedisModule_Reply_Array(reply);
+        RedisModule_Reply_Double(reply, r->score);
+        SEReply(reply, r->scoreExplain);
+	  RedisModule_Reply_ArrayEnd(reply);
```
kei-nan: indentation

src/aggregate/aggregate_exec.c
```diff
@@ -281,6 +281,118 @@ static size_t serializeResult(AREQ *req, RedisModule_Reply *reply, const SearchR
...
+    }
+    RedisModule_Reply_MapEnd(reply); // >required_fields
+  }
+
+  if (!(options & QEXEC_F_SEND_NOFIELDS)) {
+    const RLookup *lk = cv->lastLk;
+
+    RedisModule_ReplyKV_Map(reply, "attributes"); // >attributes
+
+    if (r->flags & Result_ExpiredDoc) {
```
kei-nan, line 314: I don't think this is possible anymore, for this flag to be on.
Maybe assert instead?

src/aggregate/aggregate_exec.c
raz-mon: Fixed for original as well

src/aggregate/aggregate_exec.c
```diff
@@ -281,6 +281,118 @@ static size_t serializeResult(AREQ *req, RedisModule_Reply *reply, const SearchR
...
+      if (rlk && (rlk->flags & RLOOKUP_T_NUMERIC) && v && v->t != RSVALTYPE_DOUBLE && !RSValue_IsNull(v)) {
+        double d;
+        RSValue_ToNumber(v, &d);
+        RSValue_SetNumber(&rsv, d);
+        v = &rsv;
+      }
+      if (need_map) {
+        RedisModule_Reply_CString(reply, req->requiredFields[currentField]); // key name
+      }
+      reeval_key(reply, v);
```
kei-nan: any chance for a comment explaining what is going on here?
why does this line depend on need_map variable?
what is reeval_key?

src/aggregate/aggregate_exec.c
raz-mon: Why not? E.g., for a deleted document

src/aggregate/aggregate_exec.c
raz-mon: This whole section is fully internal.
I thought to keep it since we will probably need it soon, but on second thought, I will remove it and we will add whatever support we will need later on when we support the cluster solution.

src/aggregate/aggregate_exec.c
```diff
@@ -545,7 +752,7 @@ static void sendChunk_Resp2(AREQ *req, RedisModule_Reply *reply, size_t limit,
     }
 
 done_2_err:
-    finishSendChunk(req, results, &r, cursor_done);
+    finishSendChunk(req, results, &r, cursor_done, clock() - req->initClock);
```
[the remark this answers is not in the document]
raz-mon: It's used only in the function in this case

src/aggregate/aggregate_exec.c
```diff
@@ -141,8 +142,8 @@ static size_t serializeResult(AREQ *req, RedisModule_Reply *reply, const SearchR
       RedisModule_Reply_Double(reply, r->score);
     } else {
       RedisModule_Reply_Array(reply);
-        RedisModule_Reply_Double(reply, r->score);
-        SEReply(reply, r->scoreExplain);
+      RedisModule_Reply_Double(reply, r->score);
+      SEReply(reply, r->scoreExplain);
 	  RedisModule_Reply_ArrayEnd(reply);
```
kei-nan: RedisModule_Reply_ArrayEnd indentation seems off

src/reply.c
```diff
@@ -320,25 +321,24 @@ int RedisModule_Reply_EmptyArray(RedisModule_Reply *reply) {
...
     json_add(reply, false, "[]");
-    RedisModule_ReplyWithArray(reply->ctx, 0);
   }
   _RedisModule_Reply_Next(reply);
   return REDISMODULE_OK;
 }
 
 int RedisModule_Reply_Set(RedisModule_Reply *reply) {
   int type;
+  // Reply with a set in RESP3, and an array in RESP2 (default fallback)
```
Itzikvaknin, line 336: How does `RedisModule_ReplyWithSet` work with `_RedisModule_Reply_Push(reply, REDISMODULE_REPLY_ARRAY)`? in the previous code the was a match between the types

src/reply.c
```diff
@@ -354,11 +354,8 @@ int RedisModule_Reply_SetEnd(RedisModule_Reply *reply) {
     json_add_close(reply, " ]");
   }
   int count = _RedisModule_Reply_Pop(reply);
-  if (reply->resp3) {
-    RedisModule_ReplySetSetLength(reply->ctx, count);
-  } else {
-    RedisModule_ReplySetArrayLength(reply->ctx, count);
-  }
+  // Set the length of the set/array
+  RedisModule_ReplySetSetLength(reply->ctx, count);
```
Itzikvaknin, line 358: Same as above, how do we know resp2 is handled implicitly?

src/aggregate/aggregate_exec.c
Itzikvaknin:
```suggestion
	    RedisModule_Reply_ArrayEnd(reply);
```

src/aggregate/aggregate_exec.c
```diff
@@ -281,6 +281,86 @@ static size_t serializeResult(AREQ *req, RedisModule_Reply *reply, const SearchR
...
+      // Excludes hidden fields, fields not included in RETURN and, score and language fields.
+      SchemaRule *rule = (req->sctx && req->sctx->spec) ? req->sctx->spec->rule : NULL;
+      int excludeFlags = RLOOKUP_F_HIDDEN;
+      int requiredFlags = (req->outFields.explicitReturn ? RLOOKUP_F_EXPLICITRETURN : 0);
+      int skipFieldIndex[lk->rowlen]; // Array has `0` for fields which will be skipped
+      memset(skipFieldIndex, 0, lk->rowlen * sizeof(*skipFieldIndex));
+      size_t nfields = RLookup_GetLength(lk, &r->rowdata, skipFieldIndex, requiredFlags, excludeFlags, rule);
+
+      RedisModule_Reply_Map(reply);
+        int i = 0;
```
Itzikvaknin: Indentation?

src/reply.c
raz-mon: Not sure I understand the question.
`RedisModule_ReplyWithSet` opens a set in RESP3, and an array in RESP2, the same as `RedisModule_ReplyWithArray`.

src/reply.c
raz-mon: From the API implementation, see https://github.com/redis/redis/blob/03816c15f75962b8e4662cee22be6f7ba6d5d9dd/src/module.c#L3253-L3254

src/reply.c
Itzikvaknin: its a bit odd that the opening `RedisModule_ReplyWithSet` handles the protocol inference but we should do it for the closing `_RedisModule_Reply_Push(reply, type);`

src/aggregate/aggregate_exec.c
```diff
@@ -435,6 +515,101 @@ void finishSendChunk(AREQ *req, SearchResult **results, SearchResult *r, bool cu
...
 
+/**
+ * Activates the pipeline embedded in `req`, and serializes the appropriate
+ * response to the client, according to the RESP protocol used (2/3).
+ *
+ * Note: Currently this is used only by the `FT.HYBRID` command, that does
+ * not support cursors and profiling, thus this function does not handle
+ * those cases. Support should be added as these features are added.
+ */
+static void sendChunk_hybrid(AREQ *req, RedisModule_Reply *reply, size_t limit,
```
Itzikvaknin, line 526: Are you sure we should separate at the `sendChunk` level? It’s very high-level, and it might be too high if the only difference is in serialization

src/reply.c
raz-mon: For the set I see no issues
It is mainly for the map vs. array, where the closing changes according to the protocol used
That is something the implementors probably overlooked

src/aggregate/aggregate_exec.c
raz-mon: In theory we could have used `sendChunk_Resp3`, and that would have been 100% consistent with the `FT.SEARCH/AGGREGATE` responses, but we want a whole different result structure according to the PRD so we need a new pipeline activation function or many conditions. I prefer the former at this point.
If we see they get closer and closer as the feature proceeds (which very much may happen..), we can move to the latter

src/aggregate/aggregate_exec.c
```diff
@@ -414,7 +494,7 @@ static bool hasTimeoutError(QueryError *err) {
   return QueryError_GetCode(err) == QUERY_ETIMEDOUT;
 }
 
-void finishSendChunk(AREQ *req, SearchResult **results, SearchResult *r, bool cursor_done) {
+static void finishSendChunk(AREQ *req, SearchResult **results, SearchResult *r, bool cursor_done, clock_t duration) {
   if (results) {
     destroyResults(results);
```
Itzikvaknin, line 499: How long can `destroyResults(results);` take? do you want to exclude it from time measuring?

Itzikvaknin (approved):

src/aggregate/aggregate_exec.c
raz-mon: In most cases (other than errors, timeouts which are not the common case) `results = NULL` here, so this takes no-time thus fine not to count

src/aggregate/aggregate_exec.c
```diff
@@ -281,6 +281,86 @@ static size_t serializeResult(AREQ *req, RedisModule_Reply *reply, const SearchR
...
+      int excludeFlags = RLOOKUP_F_HIDDEN;
+      int requiredFlags = (req->outFields.explicitReturn ? RLOOKUP_F_EXPLICITRETURN : 0);
+      int skipFieldIndex[lk->rowlen]; // Array has `0` for fields which will be skipped
+      memset(skipFieldIndex, 0, lk->rowlen * sizeof(*skipFieldIndex));
+      size_t nfields = RLookup_GetLength(lk, &r->rowdata, skipFieldIndex, requiredFlags, excludeFlags, rule);
+
+      RedisModule_Reply_Map(reply);
+      int i = 0;
+      for (const RLookupKey *kk = lk->head; kk; kk = kk->next) {
+        if (!kk->name || !skipFieldIndex[i++]) {
```
kei-nan, line 329: the fact that a lookup key doesn't have a name is okay?

raz-mon: I didn't write this - existing logic.

kei-nan (approved):

PR #6390, opened 2025-06-29
MOD-10299: Wrap Common AREQ Members Usage With Functions

kei-nan:
Would like to separate the pipeline structure out of the AREQ struct.
This is needed to allow the hybrid its own pipeline while not giving it other stuff it does not need like the index iterators.
This PR wraps common pipeline related members in an AREQ function so it will be easier to replace in a future PR.

1. Current: Direct usage of AREQ pipeline members
2. Change: Wrap the usage with a function when possible
3. Outcome: Better encapsulation of some AREQ members.

1. AREQ

#### Mark if applicable

### DIFF
--- a/src/aggregate/aggregate.h
+++ b/src/aggregate/aggregate.h
@@ -279,7 +279,7 @@ static inline RedisSearchCtx *AREQ_SearchCtx(AREQ *req) {
   return req->sctx;
 }
 
-static inline AGGPlan *AREQ_Plan(AREQ *req) {
+static inline AGGPlan *AREQ_AGGPlan(AREQ *req) {
   return &req->ap;
 }
 
--- a/src/aggregate/aggregate.h
+++ b/src/aggregate/aggregate.h
@@ -95,18 +95,18 @@ typedef enum {
 
 } QEFlags;
 
-#define IsCount(r) ((r)->reqflags & QEXEC_F_NOROWS)
-#define IsSearch(r) ((r)->reqflags & QEXEC_F_IS_SEARCH)
-#define IsProfile(r) ((r)->reqflags & QEXEC_F_PROFILE)
-#define IsOptimized(r) ((r)->reqflags & QEXEC_OPTIMIZE)
-#define IsFormatExpand(r) ((r)->reqflags & QEXEC_FORMAT_EXPAND)
+#define IsCount(r) (AREQ_RequestFlags(r) & QEXEC_F_NOROWS)
+#define IsSearch(r) (AREQ_RequestFlags(r) & QEXEC_F_IS_SEARCH)
+#define IsProfile(r) (AREQ_RequestFlags(r) & QEXEC_F_PROFILE)
+#define IsOptimized(r) (AREQ_RequestFlags(r) & QEXEC_OPTIMIZE)
+#define IsFormatExpand(r) (AREQ_RequestFlags(r) & QEXEC_FORMAT_EXPAND)
 #define IsWildcard(r) ((r)->ast.root->type == QN_WILDCARD)
 #define HasScorer(r) ((r)->optimizer->scorerType != SCORER_TYPE_NONE)
 #define HasLoader(r) ((r)->stateflags & QEXEC_S_HAS_LOAD)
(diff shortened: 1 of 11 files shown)


src/aggregate/aggregate.h
```diff
@@ -259,6 +259,30 @@ int AREQ_ApplyContext(AREQ *req, RedisSearchCtx *sctx, QueryError *status);
...
+
+static inline QueryProcessingCtx *AREQ_QueryProcessingCtx(AREQ *req) {
+  return &req->qiter;
+}
+
+static inline RedisSearchCtx *AREQ_SearchCtx(AREQ *req) {
+  return req->sctx;
+}
+
+static inline AGGPlan *AREQ_Plan(AREQ *req) {
```
raz-mon:
```suggestion
static inline AGGPlan *AREQ_AGGPlan(AREQ *req) {
```

src/profile.c
```diff
@@ -137,13 +137,14 @@ void Profile_Print(RedisModule_Reply *reply, void *ctx) {
 
       //Print total GIL time
         if (profile_verbose){
+          QueryProcessingCtx *qctx = AREQ_QueryProcessingCtx(req);
```
raz-mon, line 140: Your call, I would remove the other name if we're moving to use this one (the other is used mainly today.

src/coord/dist_aggregate.c
```diff
@@ -443,7 +443,8 @@ static int rpnetNext(ResultProcessor *self, SearchResult *r) {
     MRReply *fields = MRReply_MapElement(result, "extra_attributes");
     RS_LOG_ASSERT(fields && MRReply_Type(fields) == MR_REPLY_MAP, "invalid fields record");
 
-    processResultFormat(&nc->areq->reqflags, rows);
+    uint32_t reqflags = (uint32_t)AREQ_RequestFlags(nc->areq);
+    processResultFormat(&reqflags, rows);
```
raz-mon: Use the correct address

kei-nan:
```suggestion
    processResultFormat(&nc->areq->reqflags, rows);
```

src/profile.c
kei-nan: I think it would cause the size of this PR to be massive and would make it very hard to review.
I think it would be better to do that in chunks.

kei-nan: I think it would cause the size of this PR to be massive and would make it very hard to review.
I think it would be better to do that in chunks.

src/aggregate/aggregate.h
kei-nan: okay

src/aggregate/aggregate.h
```diff
@@ -259,6 +259,30 @@ int AREQ_ApplyContext(AREQ *req, RedisSearchCtx *sctx, QueryError *status);
...
+}
+
+static inline void AREQ_AddRequestFlags(AREQ *req, QEFlags flags) {
+  req->reqflags |= flags;
+}
+
+static inline void AREQ_RemoveRequestFlags(AREQ *req, QEFlags flags) {
+  req->reqflags &= ~flags;
+}
+
```
Itzikvaknin, line 273:
```suggestion

static inline bool AREQ_HasRequestFlags(const AREQ *req, QEFlags flags) {
  return (req->reqflags & flags) != 0;
}
```

src/aggregate/aggregate_request.c
```diff
@@ -48,13 +48,13 @@ static bool ensureSimpleMode(AREQ *areq) {
  * name argument must not contain any user data, as it is used for error formatting
  */
 static int ensureExtendedMode(AREQ *areq, const char *name, QueryError *status) {
-  if (areq->reqflags & QEXEC_F_IS_SEARCH) {
+  if (AREQ_RequestFlags(areq) & QEXEC_F_IS_SEARCH) {
```
Itzikvaknin, line 51: Maybe use `AREQ_HasRequestFlags` as suggested in `aggregate.h`

Itzikvaknin (approved):

PR #4370, opened 2024-01-17
Add verbosity to sanitizer [MOD-6314]

alonre24:
**Describe the changes in the pull request**

To have the ability to investigate why we are getting signal 11 in sanitizer logs occasionally, we added the `verbosity` flag to the sanitizer options. This is causing the sanitizer to generate logs on each run, not just upon failures as it was. Hence, we fail CI in the sanitizer job only if we get a failure (not relying on whether a log file is generated).
Moreover, we removed the `WARN` flag that caused the sanitizer job not to fail if the log indicated that signal 11 was caught so that the job would fail in such case.
In addition:
* remove the redundant `NOFAIL` and `PLATFORM_MODE` flags from test scripts that are no longer in use.
* never fail fast in test jobs in CI, not just in sanitizer
* collect logs for both shards and sanitizer upon failure has united into a single step. 

**Mark if applicable**

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #6395, opened 2025-06-29
Codebase cleanup

GuyAv46:
1. Remove any `printf` left in the code
2. Relax sanitizer options

#### Mark if applicable


sbin/unit-tests
```diff
@@ -51,8 +51,8 @@ END
 setup_sanitizer() {
     if [[ -n $SAN ]]; then
         ASAN_LOG=${LOGS_DIR}/${TEST_NAME}.asan.log
-        export ASAN_OPTIONS="detect_odr_violation=0:alloc_dealloc_mismatch=0:halt_on_error=0:detect_leaks=1:log_path=${ASAN_LOG}:verbosity=1:log_thread=1"
```
JoanFM, line 54: why don't we want these options?

JoanFM (dismissed):
LGTM

sbin/unit-tests
GuyAv46: We added them 17 months ago here #4370 to catch some flaky behavior he had back then, but this never yielded any results, so I think it's time to revert them.
Apparently, there is no `log_thread` option (or not anymore). I couldn't find any documentation for it, and we have the sanitizer complains about them ([here for example](https://github.com/RediSearch/RediSearch/actions/runs/15954699591/job/44999423062#step:18:437))

>   WARNING: found 2 unrecognized flag(s):
      log_thread
      log_thread

PR #6399, opened 2025-06-30
MOD-10254: Separate Aggregation Pipeline From AREQ

kei-nan:
This PR separates the aggregation pipeline from the AREQ struct.
It moves the pipeline construction functions to their own file.
This is needed in order to support the hybrid request struct.

1. Current: Aggregation pipeline exists implicitly inside the AREQ struct
2. Change: Move the pipeline to its own struct so it could be used inside the hybrid request.
3. Outcome: Better separation and code reusability

1. AREQ

#### Mark if applicable


src/aggregate/aggregate.h
```diff
@@ -123,20 +123,35 @@ typedef enum {
   QEXEC_S_ITERDONE = 0x02,
 } QEStateFlags;
 
-typedef struct AREQ {
+typedef struct AggregationPipeline {
```
Itzikvaknin: When freeing this struct it should free these QITRs (pipelines):
tail (starts with RPHybridMerger), 
each subquery pipeline

src/aggregate/aggregate_request.c
```diff
@@ -1206,14 +1206,20 @@ int AREQ_ApplyContext(AREQ *req, RedisSearchCtx *sctx, QueryError *status) {
   return REDISMODULE_OK;
 }
 
-void AREQ_Free(AREQ *req) {
-  // First, free the result processors
-  ResultProcessor *rp = AREQ_QueryProcessingCtx(req)->endProc;
+void AggregationPipeline_Free(AggregationPipeline *pipeline) {
+  ResultProcessor *rp = pipeline->qctx.endProc;
```
Itzikvaknin: Consider using `QITR_FreeChain`

src/aggregate/aggregate_pipeline.c
```diff
@@ -0,0 +1,550 @@
...
+  // Copy timeout policy to the parent struct of the result processors
+  pipeline->qctx.timeoutPolicy = timeoutPolicy;
+
+  pipeline->stateflags |= outStateflags;
+  return REDISMODULE_OK;
+error:
+  return REDISMODULE_ERR;
+}
+
+int BuildPipeline(AggregationPipeline *pipeline, QOptimizer* optimizer, RSSearchOptions* searchOpts, QueryError *status, RSTimeoutPolicy timeoutPolicy) {
```
ofiryanai: there's `buildPipeline` in [line 369](https://github.com/RediSearch/RediSearch/pull/6399/files#diff-25a88c1e1d4bb23a187edb3492ebd0ec3146c36e790ac92b01758c0522fdcac4R369)
I think its too confusing to have 2 methods defer only on capital letter. maybe add `_buildPipeline` to the private one

src/aggregate/aggregate.h
kei-nan: used QITR_FreeChain

src/aggregate/aggregate_pipeline.c
kei-nan: changed the public function name so it will be clearer.

Itzikvaknin (dismissed):

src/aggregate/aggregate.h
raz-mon: Let's add a comment that this is all the fields necessary for building a pipeline

src/aggregate/aggregate.h
```diff
@@ -123,20 +123,35 @@ typedef enum {
...
   /* plan containing the logical sequence of steps */
   AGGPlan ap;
 
+  /** Context for iterating over the queries themselves */
+  QueryProcessingCtx qctx;
+
+  /** Context, owned by request */
+  RedisSearchCtx *sctx;
```
raz-mon: Let's take a look if this is used outside of the pipeline a lot. If so, it's probably best to put it in the `AREQ`

src/aggregate/aggregate.h
```diff
@@ -123,20 +123,35 @@ typedef enum {
   QEXEC_S_ITERDONE = 0x02,
 } QEStateFlags;
 
-typedef struct AREQ {
+typedef struct AggregationPipeline {
   /* plan containing the logical sequence of steps */
   AGGPlan ap;
 
+  /** Context for iterating over the queries themselves */
+  QueryProcessingCtx qctx;
```
raz-mon: I think it makes most sense to have only the `QueryProcessingCtx` and the `AGGPlan` in this new struct, since the separation is clearer than the current one and more natural for the other commands

src/pipeline/pipeline.h
```diff
@@ -0,0 +1,64 @@
...
+
+  /** Maximum number of results to return */
+  size_t maxResultsLimit;
+
+  /** Language for highlighting **/
+  RSLanguage language;
+} AggregationPipelineParams;
+
+
+typedef struct IndexingPipelineParams {
```
kei-nan: add a bit of comments

src/pipeline/pipeline_construction.h
```diff
@@ -0,0 +1,22 @@
...
+
+/** Build the document search and scoring part of the pipeline.
+ *  This creates the initial pipeline components that execute the query against
+ *  the index to find matching documents and calculate their relevance scores. */
+void QueryPipeline_BuildIndexingPart(QueryPipeline *pipeline, const IndexingPipelineParams *params);
+
+/** Build the result processing and output formatting part of the pipeline.
+ *  This creates pipeline components that process search results through operations
+ *  like filtering, sorting, grouping, field loading, and output formatting. */
+int QueryPipeline_BuildAggregationPart(QueryPipeline *pipeline, const AggregationPipelineParams *params, uint32_t *outStateFlags);
```
kei-nan: Add a comment that this function requires at least one existing upstream in the pipeline.

src/aggregate/aggregate_request.c
```diff
@@ -1802,3 +1267,38 @@ void AREQ_Free(AREQ *req) {
   rm_free(req->args);
   rm_free(req);
 }
+
+int AREQ_BuildPipeline(AREQ *req, QueryError *status) {
+  QueryPipeline_Initialize(&req->pipeline, req-> reqConfig.timeoutPolicy, status);
```
raz-mon:
```suggestion
  QueryPipeline_Initialize(&req->pipeline, req->reqConfig.timeoutPolicy, status); 
```

src/aggregate/aggregate.h
```diff
@@ -101,14 +103,12 @@ typedef enum {
...
 #define IsWildcard(r) ((r)->ast.root->type == QN_WILDCARD)
-#define HasScorer(r) ((r)->optimizer->scorerType != SCORER_TYPE_NONE)
+#define HasScorer(opt) ((opt)->scorerType != SCORER_TYPE_NONE)
 #define HasLoader(r) ((r)->stateflags & QEXEC_S_HAS_LOAD)
 #define IsScorerNeeded(r) ((r)->reqflags & (QEXEC_F_SEND_SCORES | QEXEC_F_SEND_SCORES_AS_FIELD))
 #define HasScoreInPipeline(r) ((r)->reqflags & QEXEC_F_SEND_SCORES_AS_FIELD)
 #define IsInternal(r) ((r)->reqflags & QEXEC_F_INTERNAL)
 #define IsDebug(r) ((r)->reqflags & QEXEC_F_DEBUG)
-// Get the index search context from the result processor
-#define RP_SCTX(rpctx) ((rpctx)->parent->sctx)
```
raz-mon, line 111: Not sure removing the `RedisSearchCtx` from the shared context to the result processors is a good move, since it is clearer that it is a shared resource of the pipeline using it, and clearly sets the bounds of its usage (to the result processors that are attached to it)

src/aggregate/aggregate_request.c
```diff
@@ -33,11 +33,11 @@ extern RSConfig RSGlobalConfig;
  *   formatting
  * @param status the error object
  */
-static bool ensureSimpleMode(AREQ *areq) {
-  if(AREQ_RequestFlags(areq) & QEXEC_F_IS_AGGREGATE) {
+static bool ensureSimpleMode(AREQ *req) {
+  if (req->reqflags & QEXEC_F_IS_AGGREGATE) {
```
raz-mon: Why not use the API?

src/aggregate/aggregate_request.c
```diff
@@ -33,11 +33,11 @@ extern RSConfig RSGlobalConfig;
...
     return false;
   }
-  AREQ_AddRequestFlags(areq, QEXEC_F_IS_SEARCH);
+  req->reqflags |= QEXEC_F_IS_SEARCH;
```
raz-mon: Why not use the API?

src/pipeline/pipeline_construction.c
```diff
@@ -0,0 +1,525 @@
...
+#define PUSH_RP()                                      \
+  rpUpstream = pushRP(&pipeline->qctx, rp, rpUpstream); \
+  rp = NULL;
+
+/**
+ * Builds the document search and scoring pipeline that executes queries against the index.
+ * This creates the initial pipeline components that find matching documents and calculate
+ * their relevance scores, providing the foundation for subsequent aggregation and filtering stages.
+ */
+void QueryPipeline_BuildIndexingPart(QueryPipeline *pipeline, const IndexingPipelineParams *params) {
```
raz-mon: `Indexing` --> `Index`/`Implicit`
`Index` is not so good since we have scoring as well. I would go with `Implicit` or something of that sort.
Goes for the other places `Indexing` is used as well.

src/pipeline/pipeline_construction.c
```diff
@@ -0,0 +1,525 @@
...
+    rp = RPHighlighter_New(params->language, params->outFields, lookup);
+    PUSH_RP();
+  }
+
+  return REDISMODULE_OK;
+error:
+  return REDISMODULE_ERR;
+}
+
+int QueryPipeline_BuildAggregationPart(QueryPipeline *pipeline, const AggregationPipelineParams *params, uint32_t *outStateFlags) {
```
raz-mon: Documentation.
Specifically, what part of the command does this function handle, so it's difference from the above is clear.

raz-mon (commented):
👍 
See a few comments

src/aggregate/aggregate.h
kei-nan: We did have talks where different result processors across multiple threads will need their own context
And that means that the pipeline can't have a single context anymore.

raz-mon: OK I don't see a blocker for it so let's proceed with it.

raz-mon (dismissed):

Itzikvaknin (approved):
Looks good!
Consider adding a test that demonstrates the boundaries of each new component and shows what this decoupling enables — for example, executing the pipeline without an `AREQ`

PR #6405, opened 2025-06-30
[MOD-10298] - add check version to json open key

BenGoldberger:
add check that json api version is >=5 before using openKeyWithFlags

### DIFF
--- a/src/document_basic.c
+++ b/src/document_basic.c
@@ -220,14 +220,19 @@ int Document_LoadSchemaFieldJson(Document *doc, RedisSearchCtx *sctx, QueryError
   }
 
   RedisModule_CloseKey(k);
+  const char *keyName = RedisModule_StringPtrLen(doc->docKey, NULL);
 
-  RedisJSON jsonRoot = japi->openKeyWithFlags(ctx, doc->docKey, DOCUMENT_OPEN_KEY_INDEXING_FLAGS);
+  RedisJSON jsonRoot = NULL;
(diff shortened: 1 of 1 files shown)


src/document_basic.c
```diff
@@ -220,14 +220,19 @@ int Document_LoadSchemaFieldJson(Document *doc, RedisSearchCtx *sctx, QueryError
...
 
   RedisModule_CloseKey(k);
+  const char *keyName = RedisModule_StringPtrLen(doc->docKey, NULL);
 
-  RedisJSON jsonRoot = japi->openKeyWithFlags(ctx, doc->docKey, DOCUMENT_OPEN_KEY_INDEXING_FLAGS);
+  RedisJSON jsonRoot = NULL;
+  if (japi_ver >= 5) {
+    jsonRoot = japi->openKeyWithFlags(ctx, doc->docKey, DOCUMENT_OPEN_KEY_QUERY_FLAGS);
+  } else {
+    jsonRoot = japi->openKeyFromStr(ctx, keyName);
```
alonre24: Why is that the fallback? should we test that `japi_ver` supports `openKeyFromStr` ?

BenGoldberger: openKeyFromStr is in every version
this is the fallback I see in other cases https://github.com/RediSearch/RediSearch/blob/115d8eeac5a587120695ee3acf2c210a805d531f/src/rlookup.c#L696-L705

alonre24 (dismissed):

PR #6408, opened 2025-07-01
MOD-9985 - Extract the UTF case folding code

DarthB:

We extract the normalizeStr function that does utf case folding outside of sortable.h/c. 

## Why do we need this

This part won't be ported to Rust yet.

Linking a crate like ICU adds more complexity and also a Miri error. We plan to rely on the libnu which is used by C.

## Referenced PRs:

- https://github.com/RediSearch/RediSearch/pull/6311

### DIFF
--- a/src/normalize.c
+++ b/src/normalize.c
@@ -1,0 +1,50 @@
+#include <string.h>
+#include <stdint.h>
+
+#include "libnu/libnu.h"
+
+#include "normalize.h"
+
+#ifndef rm_calloc 
+#ifndef REDIS_MODULE_TARGET
+#include <stdlib.h>
+#define rm_calloc calloc
+#else
+#include "redismodule.h"
+static inline void *rm_calloc(size_t nelem, size_t elemsz) {
+    return RedisModule_Calloc(nelem, elemsz);
+}
+#endif
+#endif
+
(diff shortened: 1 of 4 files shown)


src/normalize.c
```diff
@@ -0,0 +1,50 @@
...
+#ifndef REDIS_MODULE_TARGET
+#include <stdlib.h>
+#define rm_calloc calloc
+#else
+#include "redismodule.h"
+static inline void *rm_calloc(size_t nelem, size_t elemsz) {
+    return RedisModule_Calloc(nelem, elemsz);
+}
+#endif
+#endif
```
LukeMathWalker, line 18: Why do we need to inline this definition? It should be enough to import it, no?

DarthB: Old PR.

To answer: We needed that for linking C code into Rust, but we excluded all the dependencies from the Redis Module.
I don't come up with a nicer solution for linking this part when it's invoked by linked C code. Here, the `normalizeStr` method is calling `rm_calloc`. So we also needed the symbols for the `rm_calloc` variant in case it is used.

We won't use it in https://github.com/RediSearch/RediSearch/pull/6311 anymore.

PR #6410, opened 2025-07-01
Fix FieldSpec Cleanup on failed RDB load - [MOD-10307]

GuyAv46:

Fix an issue when loading a schema from RDB, and we decide to abort for some reason.

We need to make sure the number of `IndexError_Init` calls matches the number of `IndexError_Clear` calls we later do, otherwise we may free the global "N/A" string.

#### Mark if applicable

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -3142,6 +3142,7 @@ int Indexes_RdbLoad(RedisModuleIO *rdb, int encver, int when) {
   for (size_t i = 0; i < nIndexes; ++i) {
     if (IndexSpec_CreateFromRdb(ctx, rdb, encver, &status) != REDISMODULE_OK) {
       RedisModule_LogIOError(rdb, "warning", "RDB Load: %s", QueryError_GetDisplayableError(&status, RSGlobalConfig.hideUserDataFromLog));
+      QueryError_ClearError(&status);
       return REDISMODULE_ERR;
     }
   }


src/spec.c
```diff
@@ -2917,7 +2912,11 @@ int IndexSpec_CreateFromRdb(RedisModuleCtx *ctx, RedisModuleIO *rdb, int encver,
 
   sp->numFields = LoadUnsigned_IOError(rdb, goto cleanup);
   sp->fields = rm_calloc(sp->numFields, sizeof(FieldSpec));
-  int maxSortIdx = -1;
+  // First, initialise fields IndexError before loading them.
+  // If some fields are not loaded correctly, we will attempt to free the spec and cleanup the fields.
+  for (int i = 0; i < sp->numFields; i++) {
+    sp->fields[i].indexError = IndexError_Init();
+  }
```
alonre24: Will that fix the leak in `test_rdb_memory_limit`? Can we remove the skip for sanitizer there to validate (or test it in some other way)?

alonre24 (commented):
nice catch

src/spec.c
GuyAv46: Will re-enable the test.
I'm not sure why this is new or flaky. I'd think any short-read test with multiple fields should cause this issue to happen

alonre24: Its not flaky, it's new from my PR from yesterday, and this is probably the only test that checks rdb failure with sanitizer...

GuyAv46: Oh we don't run the short read test on sanitizer... well that's why we didn't catch it before

PR #6411, opened 2025-07-01
Fix missing profile info - [MOD-6327]

GuyAv46:

1. Current: Profiling info is added to the profiling info array when we are done consuming the reply (and only if we do)
2. Change: Take the profiling info only, and right when we take a reply from the channel
3. Outcome: Profiling info is available even if we don't consume the entire reply. Profile info left in the channel will still be missing

We now split the profiling info from the full reply and move the ownership over it to the profiling info array. The rest of the reply is still owned by the RPNet result processor.
This relies on `hiredis` being able to handle the `NULL` we set instead of the sub-section of the profiling data, which now has to be released separately.

1. MOD-8157

1. shards reply ownership on `FT.PROFILE AGGREGATE`

#### Mark if applicable

### DIFF
--- a/src/coord/dist_aggregate.c
+++ b/src/coord/dist_aggregate.c
@@ -266,30 +266,6 @@ static int getNextReply(RPNet *nc) {
     return 1;
   }
 
-  MRReply *rows = MRReply_ArrayElement(root, 0);
-  if (nc->cmd.forProfiling && nc->cmd.protocol == 3) {
-    /* On RESP3, FT.PROFILE AGGREGATE returns:
-      [
-        {
-          "Results": { <FT.AGGREGATE reply> },
-          "Profile": { <profile data> }
-        },
-        cursor_id
-      ]
-     * So we need to extract the "Results" map from the first element of the array
-     */
-
-    rows = MRReply_MapElement(rows, "results");
-  }
-  if (   rows == NULL
-      || (MRReply_Type(rows) != MR_REPLY_ARRAY && MRReply_Type(rows) != MR_REPLY_MAP)
-      || MRReply_Length(rows) == 0) {
-    MRReply_Free(root);
-    root = NULL;
-    rows = NULL;
-    RedisModule_Log(RSDummyContext, "warning", "An empty reply was received from a shard");
-  }
-
   // For profile command, extract the profile data from the reply
   if (nc->cmd.forProfiling) {
     // if the cursor id is 0, this is the last reply from this shard, and it has the profile data
@@ -314,6 +290,35 @@ static int getNextReply(RPNet *nc) {
       }
       array_append(nc->shardsProfile, profile_data);
     }
+  }
+
+  MRReply *rows = MRReply_ArrayElement(root, 0);
+
+  if (nc->cmd.forProfiling && nc->cmd.protocol == 3) {
+    /* On RESP3, FT.PROFILE AGGREGATE returns:
+      [
+        {
+          "Results": { <FT.AGGREGATE reply> },
+          "Profile": { <profile data> }
+        },
+        cursor_id
+      ]
+     * So we need to extract the "Results" map from the first element of the array
+     */
+
+    rows = MRReply_MapElement(rows, "results");
+  }
+  if (   rows == NULL
+      || (MRReply_Type(rows) != MR_REPLY_ARRAY && MRReply_Type(rows) != MR_REPLY_MAP)
+      || MRReply_Length(rows) == 0) {
+    if (rows == NULL || MRReply_Length(rows) == 0) {
+      RedisModule_Log(RSDummyContext, "verbose", "An empty reply was received from a shard");
+    } else {
+      RedisModule_Log(RSDummyContext, "warning", "An unexpected reply was received from a shard");
+    }
+    MRReply_Free(root);
+    root = NULL;
+    rows = NULL;
   }
 
   // invariant: either rows == NULL or at least one row exists
--- a/src/coord/dist_aggregate.c
+++ b/src/coord/dist_aggregate.c
@@ -22,12 +22,12 @@
 #include "aggregate/aggregate_debug.h"
 #include "info/info_redis/threads/current_thread.h"
 
-#include <err.h>
+#define CURSOR_EOF 0
 
 // Get cursor command using a cursor id and an existing aggregate command
 // Returns true if the cursor is not done (i.e., not depleted)
 static bool getCursorCommand(long long cursorId, MRCommand *cmd, MRIteratorCtx *ctx) {
-  if (cursorId == 0) {
+  if (cursorId == CURSOR_EOF) {
     // Cursor was set to 0, end of reply chain. cmd->depleted will be set in `MRIteratorCallback_Done`.
     return false;
   }
@@ -100,9 +100,12 @@ static void netCursorCallback(MRIteratorCallbackCtx *ctx, MRReply *rep) {
     size_t len = MRReply_Length(rep);
     if (isResp3) {
       bail_out = len != 2; // (map, cursor)
-      if (!bail_out) bail_out = MRReply_Type(MRReply_ArrayElement(rep, 0)) != MR_REPLY_MAP;
       if (bail_out) {
         RedisModule_Log(RSDummyContext, "warning", "Expected reply of length 2, got %ld", len);
+      }
+      if (!bail_out) bail_out = MRReply_Type(MRReply_ArrayElement(rep, 0)) != MR_REPLY_MAP;
+      if (bail_out) {
+        RedisModule_Log(RSDummyContext, "warning", "Expected reply of type map, got %d", MRReply_Type(MRReply_ArrayElement(rep, 0)));
       }
     } else {
       bail_out = len != 2 && len != 3; // (results, cursor) or (results, cursor, profile)
@@ -122,7 +125,7 @@ static void netCursorCallback(MRIteratorCallbackCtx *ctx, MRReply *rep) {
   long long cursorId;
   MRReply* cursor = MRReply_ArrayElement(rep, 1);
   if (!MRReply_ToInteger(cursor, &cursorId)) {
-    cursorId = 0;
+    cursorId = CURSOR_EOF;
   }
 
   // Push the reply down the chain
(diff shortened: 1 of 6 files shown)


src/coord/rmr/reply.c
```diff
@@ -222,15 +222,34 @@ inline MRReply *MRReply_ArrayElement(const MRReply *reply, size_t idx) {
   return reply->element[idx];
 }
 
-inline MRReply *MRReply_MapElement(const MRReply *reply, const char *key) {
-  if (reply->type != MR_REPLY_MAP) return NULL;
-  for (int i = 0; i < reply->elements; i += 2) {
+inline MRReply *MRReply_TakeArrayElement(const MRReply *reply, size_t idx) {
+  // TODO: check out of bounds
+  MRReply *ret = reply->element[idx];
```
Itzikvaknin, line 227: DO?

src/coord/dist_aggregate.c
```diff
@@ -290,7 +290,33 @@ static int getNextReply(RPNet *nc) {
     RedisModule_Log(RSDummyContext, "warning", "An empty reply was received from a shard");
   }
 
-  // invariant: either rows == NULL or least one row exists
+  // For profile command, extract the profile data from the reply
```
Itzikvaknin, line 248: Nice

src/coord/dist_aggregate.c
```diff
@@ -316,6 +292,35 @@ static int getNextReply(RPNet *nc) {
...
+    rows = MRReply_MapElement(rows, "results");
+  }
+  if (   rows == NULL
+      || (MRReply_Type(rows) != MR_REPLY_ARRAY && MRReply_Type(rows) != MR_REPLY_MAP)
+      || MRReply_Length(rows) == 0) {
+    if (rows == NULL || MRReply_Length(rows) == 0) {
+      RedisModule_Log(RSDummyContext, "verbose", "An empty reply was received from a shard");
+    } else {
+      RedisModule_Log(RSDummyContext, "warning", "An unexpected reply was received from a shard");
+    }
```
Itzikvaknin: Whats the coverage status here?
It shows both branches are not covered, but how can 319-321 be covered in that case?

Itzikvaknin (commented):
Nice
Only coverage issue :)

src/coord/rmr/reply.c
GuyAv46: Did :)
Added an assertion

src/coord/dist_aggregate.c
```diff
@@ -125,29 +126,8 @@ static void netCursorCallback(MRIteratorCallbackCtx *ctx, MRReply *rep) {
...
-  {
-    MRReply *results = MRReply_ArrayElement(rep, 0);
-    if (results && MRReply_Type(results) == MR_REPLY_ARRAY && MRReply_Length(results) > 1) {
-      MRIteratorCallback_AddReply(ctx, rep); // to be picked up by getNextReply
-      // User code now owns the reply, so we can't free it here ourselves!
-      rep = NULL;
-    }
-  }
+  MRIteratorCallback_AddReply(ctx, rep); // to be picked up by getNextReply
+  rep = NULL; // User code now owns the reply, so we can't free it here ourselves!
```
GuyAv46: Essential validation is done previously. We need to pass the reply to the thread so we don't lose the profile info, even if there are no results.
This logic was partially done by the reader already, and now only by the reader and after extracting the profile data

raz-mon: Is this needed? I see you removed the freeing code

src/coord/dist_aggregate.c
```diff
@@ -266,38 +243,57 @@ static int getNextReply(RPNet *nc) {
...
-
-    rows = MRReply_MapElement(rows, "results");
-  }
-  if (   rows == NULL
-      || (MRReply_Type(rows) != MR_REPLY_ARRAY && MRReply_Type(rows) != MR_REPLY_MAP)
-      || MRReply_Length(rows) == 0) {
+  // For profile command, extract the profile data from the reply
+  if (nc->cmd.forProfiling) {
+    // if the cursor id is 0, this is the last reply from this shard, and it has the profile data
+    if (0 == MRReply_Integer(MRReply_ArrayElement(root, 1))) {
```
raz-mon: Since we use this id in several parts of the code consider adding a macro for it e.g.,:
```suggestion
    if (CURSOR_DONE_ID == MRReply_Integer(MRReply_ArrayElement(root, 1))) {
```

raz-mon: Also, an assertion that we have this element may be handy

src/coord/dist_aggregate.c
```diff
@@ -266,38 +243,57 @@ static int getNextReply(RPNet *nc) {
...
+        //   {
+        //     "Results": { <FT.AGGREGATE reply> },
+        //     "Profile": { <profile data> }
+        //   },
+        //   cursor_id
+        // ]
+        MRReply *data = MRReply_ArrayElement(root, 0);
+        profile_data = MRReply_TakeMapElement(data, "profile");
+      } else {
+        // RESP2
```
raz-mon, line 264: A short description of the RESP2 response format would be good here as well (like for RESP3 above)

src/coord/dist_aggregate.c
```diff
@@ -266,38 +243,57 @@ static int getNextReply(RPNet *nc) {
...
+      meta = MRReply_MapElement(meta, "results"); // profile has an extra level
+    }
+    rows = MRReply_MapElement(meta, "results");
+  } else { // RESP2
+    rows = MRReply_ArrayElement(root, 0);
+  }
+
+  // invariant: either rows == NULL or at least one row exists
+  const size_t empty_rows_len = nc->cmd.protocol == 3 ? 0 : 1; // RESP2 has the first element as the number of results.
+  RS_ASSERT(!rows || MRReply_Type(rows) == MR_REPLY_ARRAY);
```
raz-mon: According to the comment above, should we also assert that is has length >= 1 in the second condition?

src/coord/dist_aggregate.c
```diff
@@ -419,14 +403,10 @@ static int rpnetNext(ResultProcessor *self, SearchResult *r) {
   }
 
   // invariant: at least one row exists
-
-  bool resp3 = MRReply_Type(rows) == MR_REPLY_MAP;
   if (new_reply) {
     if (resp3) { // RESP3
       nc->curIdx = 0;
-      MRReply *results = MRReply_MapElement(rows, "results");
-      RS_LOG_ASSERT(results, "invalid results record: missing 'results' key");
```
raz-mon, line 428: Always good to make sure (again) 😅

tests/pytests/test_issues.py
```diff
@@ -1533,27 +1533,57 @@ def test_mod_8809_multi_index_multi_fields(env:Env):
...
+
     yields_count = env.cmd(debug_cmd(), 'YIELDS_ON_LOAD_COUNTER')
     expected_min_yields = 7 * num_docs // yield_every_n_ops
-    env.assertGreaterEqual(yields_count, expected_min_yields, 
+    env.assertGreaterEqual(yields_count, expected_min_yields,
                           message=f"Expected at least {expected_min_yields} yields, got {yields_count}")
 
+def mod_8157(env:Env):
+    """
+    Test missing profile info on aggregate query
```
raz-mon, line 1562: 😍

tests/pytests/test_profile.py
```diff
@@ -667,25 +667,22 @@ def testPofileGILTime():
...
-    total_GIL_index[-1] += 1
-    total_GIL_time = access_nested_list(res, total_GIL_index)
-
-    env.assertGreaterEqual(float(total_GIL_time), 0)
-    env.assertGreaterEqual(float(rp_GIL_time), 0)
-    env.assertGreaterEqual(float(total_GIL_time), float(rp_GIL_time))
-  except Exception:
-    print(f"::error title=GIL report test failure:: res: {res}")
+  env.assertTrue(recursive_contains(res, 'Threadsafe-Loader'), message=f"res: {res}")
+  env.assertTrue(recursive_contains(res, 'Total GIL time'), message=f"res: {res}")
```
raz-mon, line 671: 😍

src/coord/dist_aggregate.c
```diff
@@ -125,29 +126,8 @@ static void netCursorCallback(MRIteratorCallbackCtx *ctx, MRReply *rep) {
   }
 
   // Push the reply down the chain
-  if (isResp3) // RESP3
-  {
-    MRReply *map = MRReply_ArrayElement(rep, 0);
-    MRReply *results = NULL;
-    if (map && MRReply_Type(map) == MR_REPLY_MAP) {
-      results = MRReply_MapElement(map, "results");
-      if (cmd->forProfiling) results = MRReply_MapElement(results, "results"); // profile has an extra level
```
raz-mon, line 134: Since you're already changing the validation logic here, it's a good chance to add a clear comment on what validations are actually done on the reply here.

raz-mon (commented):
Very nice! 💪 
So there are a couple of issues here:
* Didn’t send last response to pipeline thread if the results was empty (e.g., when exactly 1000 results are sent from a shard.
* We saved the profile data only when finishing to read the reply in rpnetNext --> If we stop earlier (timeout, limit), we don’t get the profile data.
* Anything else?

tests/pytests/test_issues.py
```diff
@@ -1533,27 +1533,57 @@ def test_mod_8809_multi_index_multi_fields(env:Env):
...
+    reply = env.cmd('FT.PROFILE', 'idx', 'AGGREGATE', 'QUERY', '*')
+    if env.protocol == 2:
+        # RESP2 returns a list of lists
+        profile = reply[1]
+    else:
+        # RESP3 returns a dictionary
+        profile = reply['Profile']
+    profile = to_dict(profile)
+    env.assertContains('Shards', profile)
+    env.assertEqual(len(profile['Shards']), env.shardsCount)
```
alonre24: Can we check with a timeout as well?

alonre24 (commented):
LGTM, let's consider if it safe to backport it

raz-mon (approved):
💪

alonre24 (approved):

GuyAv46:
We had some reply structure changes since `2.X` versions, so this PR won't be backported to these versions for now

PR #6413, opened 2025-07-01
MOD-10301: Add index locking/freeing mechanism for index-space consistency

raz-mon:
Adds an index-spec locking & freeing mechanism that will allow index-space consistency among the two subqueries of the `FT.HYBRID` command.
The locking flow should be as follows:
1. Pipeline thread locks the index for read.
2. Both depleter threads lock the index for read.
3. Pipeline thread releases the index lock, such that the depleter threads can release the index-lock and catch the GIL without deadlocks.
4. Depleter threads release the index lock before loading the documents.

New behavior of `RPDepleter`:
* In case the depleter also performs the index locking logic, it will now return `RS_RESULT_DEPLETING` until the lock is taken by both background threads (depleters), such that the pipeline thread will not sleep until it releases the lock, such that there will be no deadlock when the depleters perform the Loader logic.

Important notes:
1. Assumes the index lock has already been taken by the pipeline thread (main thread or BG).
2. Assumes that each subquery pipeline has its own `RediSearchCtx`, with `sctx->flags & RS_CTX_UNSET = 1`, such that the locking and unlocking mechanism works properly.
3. An option to add the depleters without the locking logic is enabled, by creating the depleters with `RPDepleter_New(ref, false)`, such that we can later exploit this in the cluster solution (the coordinator should not lock the indexes`.
4. In case there is not loader in the depleter's pipeline, the index lock is released either in `RPIdx` (can be unlocked there even with a loader in case of depletion of the index) or when the depleter thread is done its work. Either way, when the depleter is done with its execution, we know the index is unlocked (from the thread's POV), and index-space consistency among the two subquery pipelines is guaranteed.

### DIFF
--- a/src/result_processor.c
+++ b/src/result_processor.c
@@ -1498,7 +1498,7 @@ static void RPDepleter_Deplete(void *arg) {
     // Lock the index for read
     RedisSearchCtx_LockSpecRead(RP_SCTX(&self->base));
     // Increment the counter
-    __atomic_fetch_add(&((DepleterSync *)StrongRef_Get(self->sync_ref))->num_locked, 1, __ATOMIC_SEQ_CST);
+    atomic_fetch_add(&((DepleterSync *)StrongRef_Get(self->sync_ref))->num_locked, 1);
   }
 
   // Deplete the pipeline into the `self->results` array.
@@ -1566,7 +1566,9 @@ static int RPDepleter_Next_Dispatch(ResultProcessor *base, SearchResult *r) {
 
   if (self->take_index_lock && !self->index_released) {
     // Load the atomic counter
-    int num_locked = __atomic_load_n(&((DepleterSync *)StrongRef_Get(self->sync_ref))->num_locked, __ATOMIC_SEQ_CST);
+    int num_locked = atomic_load(&((DepleterSync *)StrongRef_Get(self->sync_ref))->num_locked);
+    // We currently support only 2 upstream depleters. When we move on to support
+    // more subqueries - this will be generalized.
     if (num_locked == 2) {
       // Release the index
       RedisSearchCtx_UnlockSpec(RP_SCTX(base));
--- a/src/result_processor.c
+++ b/src/result_processor.c
@@ -1436,10 +1436,14 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
 /*******************************************************************************************************************
  *  Depleter Result Processor
  *
- *  The RPDepleter result processor offloads the task of consuming all results from its upstream processor
- *  into a background thread, storing them in an internal array. While the background thread is running,
- *  calls to Next() return RS_RESULT_DEPLETING. Once depleting is complete, Next() yields results one by one
- *  from the internal array, and finally returns the last return code from the upstream.
+ *  The RPDepleter result processor offloads the task of consuming all results from
+ *  its upstream processor into a background thread, storing them in an internal
+ *  array. While the background thread is running, calls to Next() wait on a shared
+ *  condition variable and return RS_RESULT_DEPLETING. The thread can be awakened
+ *  either by its own depleting thread completing or by another RPDepleter's thread
+ *  signaling completion. Once depleting is complete for this processor, Next()
+ *  yields results one by one from the internal array, and finally returns the last
+ *  return code from the upstream.
  *******************************************************************************************************************/
 typedef struct {
   ResultProcessor base;             // Base result processor struct
@@ -1448,7 +1452,7 @@ typedef struct {
   uint cur_idx;                     // Current index for yielding results
   RPStatus last_rc;                 // Last return code from upstream
   bool first_call;                  // Whether the first call to Next has been made
-  StrongRef sync_ref;               // Reference to shared synchronization object
+  StrongRef sync_ref;               // Reference to shared synchronization object (DepleterSync)
   bool take_index_lock;             // Whether or not the depleter should take the index lock
   bool index_released;              // Whether or not the index-spec has been released by the pipeline thread yet
 } RPDepleter;
@@ -1457,6 +1461,7 @@ typedef struct {
 typedef struct {
   pthread_cond_t cond;
   pthread_mutex_t mutex;
+  uint n_depleters;       // Number of depleters to sync
   atomic_int num_locked;  // Number of depleters that have locked the index
 } DepleterSync;
 
@@ -1469,10 +1474,11 @@ static void DepleterSync_Free(void *obj) {
 }
 
 // Create a new shared sync object for a pipeline
-StrongRef DepleterSync_New() {
-  DepleterSync *sync = rm_malloc(sizeof(DepleterSync));
+StrongRef DepleterSync_New(uint n_depleters) {
+  DepleterSync *sync = rm_calloc(1, sizeof(DepleterSync));
   pthread_cond_init(&sync->cond, NULL);
   pthread_mutex_init(&sync->mutex, NULL);
+  sync->n_depleters = n_depleters;
   return StrongRef_New(sync, DepleterSync_Free);
 }
 
@@ -1519,7 +1525,6 @@ static void RPDepleter_Deplete(void *arg) {
 
   // Signal completion under mutex protection
   DepleterSync *sync = (DepleterSync *)StrongRef_Get(self->sync_ref);
-  RS_LOG_ASSERT(sync, "Invalid sync reference");
   pthread_mutex_lock(&sync->mutex);
   self->done_depleting = true;
   pthread_cond_broadcast(&sync->cond);  // Wake up all waiting depleters
@@ -1566,10 +1571,9 @@ static int RPDepleter_Next_Dispatch(ResultProcessor *base, SearchResult *r) {
 
   if (self->take_index_lock && !self->index_released) {
     // Load the atomic counter
-    int num_locked = atomic_load(&((DepleterSync *)StrongRef_Get(self->sync_ref))->num_locked);
-    // We currently support only 2 upstream depleters. When we move on to support
-    // more subqueries - this will be generalized.
-    if (num_locked == 2) {
+    DepleterSync *sync = (DepleterSync *)StrongRef_Get(self->sync_ref);
+    int num_locked = atomic_load(&sync->num_locked);
+    if (num_locked == sync->n_depleters) {
       // Release the index
       RedisSearchCtx_UnlockSpec(RP_SCTX(base));
       // Mark the index as released
@@ -1582,7 +1586,6 @@ static int RPDepleter_Next_Dispatch(ResultProcessor *base, SearchResult *r) {
 
   // On subsequent calls, wait on condition variable for any depleter to complete
   DepleterSync *sync = (DepleterSync *)StrongRef_Get(self->sync_ref);
-  RS_LOG_ASSERT(sync, "Invalid sync reference");
   pthread_mutex_lock(&sync->mutex);
 
   // Check if depleting is already done.
@@ -1622,6 +1625,8 @@ ResultProcessor *RPDepleter_New(StrongRef sync_ref, bool take_index_lock) {
   ret->base.type = RP_DEPLETER;
   ret->first_call = true;
   ret->sync_ref = sync_ref;
+  // Make sure the sync reference is valid
+  RS_LOG_ASSERT(StrongRef_Get(sync_ref), "Invalid sync reference");
   ret->take_index_lock = take_index_lock;
   return &ret->base;
 }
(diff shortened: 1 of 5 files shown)


src/result_processor.c
```diff
@@ -1442,19 +1442,22 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
-  uint cur_idx;                        // Current index for yielding results
-  RPStatus last_rc;                    // Last return code from upstream
-  bool first_call;                     // Whether the first call to Next has been made
-  StrongRef sync_ref;                  // Reference to shared synchronization object
+  ResultProcessor base;             // Base result processor struct
+  arrayof(SearchResult *) results;  // Array of pointers to SearchResult, filled by the depleting thread
+  bool done_depleting;              // Set to `true` when depleting is finished (under lock)
+  uint cur_idx;                     // Current index for yielding results
+  RPStatus last_rc;                 // Last return code from upstream
+  bool first_call;                  // Whether the first call to Next has been made
```
kei-nan, line 1457: all these booleans feel like good candidates for a bitmask replacement.
Also I think some should move to the `DepleterSync` struct.

src/result_processor.c
```diff
@@ -1491,6 +1494,13 @@ static void RPDepleter_Deplete(void *arg) {
   RPDepleter *self = (RPDepleter *)arg;
   RPStatus rc;
 
+  if (self->take_index_lock) {
+    // Lock the index for read
+    RedisSearchCtx_LockSpecRead(RP_SCTX(&self->base));
+    // Increment the counter
+    atomic_fetch_add(&((DepleterSync *)StrongRef_Get(self->sync_ref))->num_locked, 1);
```
kei-nan: I wonder if using a semaphore would be cleaner or even using the existing/different conditional variable.

src/result_processor.c
```diff
@@ -1548,6 +1564,22 @@ static int RPDepleter_Next_Dispatch(ResultProcessor *base, SearchResult *r) {
...
+  if (self->take_index_lock && !self->index_released) {
+    // Load the atomic counter
+    int num_locked = atomic_load(&((DepleterSync *)StrongRef_Get(self->sync_ref))->num_locked);
+    // We currently support only 2 upstream depleters. When we move on to support
+    // more subqueries - this will be generalized.
+    if (num_locked == 2) {
+      // Release the index
+      RedisSearchCtx_UnlockSpec(RP_SCTX(base));
+      // Mark the index as released
+      self->index_released = true;
```
kei-nan: shouldn't this boolean exist in the sync context instead in the depleter itself?
Right now it seems like each depleter will release the lock when he sees the count reached 2.
Also please use the number of upstreams/childrens/instances instead of a magic number "2".

src/result_processor.c
raz-mon: I don't see how a semaphore is good here, nor how we can use the existing cond-var.
What use of these do you see?

src/result_processor.c
```diff
@@ -1442,19 +1442,22 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
-  bool first_call;                     // Whether the first call to Next has been made
-  StrongRef sync_ref;                  // Reference to shared synchronization object
+  ResultProcessor base;             // Base result processor struct
+  arrayof(SearchResult *) results;  // Array of pointers to SearchResult, filled by the depleting thread
+  bool done_depleting;              // Set to `true` when depleting is finished (under lock)
+  uint cur_idx;                     // Current index for yielding results
+  RPStatus last_rc;                 // Last return code from upstream
+  bool first_call;                  // Whether the first call to Next has been made
+  StrongRef sync_ref;               // Reference to shared synchronization object
+  bool take_index_lock;             // Whether or not the depleter should take the index lock
```
Itzikvaknin:
```suggestion
  StrongRef sync_ref;               // Reference to shared synchronization object of type DepleterSync
```

src/result_processor.c
```diff
@@ -1548,6 +1564,22 @@ static int RPDepleter_Next_Dispatch(ResultProcessor *base, SearchResult *r) {
     return RS_RESULT_DEPLETING;
   }
 
+  if (self->take_index_lock && !self->index_released) {
+    // Load the atomic counter
+    int num_locked = atomic_load(&((DepleterSync *)StrongRef_Get(self->sync_ref))->num_locked);
+    // We currently support only 2 upstream depleters. When we move on to support
+    // more subqueries - this will be generalized.
+    if (num_locked == 2) {
```
Itzikvaknin: Consider adding the number of depleters sharing the `DepleterSync` as context/argument

src/result_processor.c
```diff
@@ -1442,19 +1442,22 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
  *  from the internal array, and finally returns the last return code from the upstream.
```
Itzikvaknin: Maybe worth mentioning how it works with multiple `RPDepleter` sharing the `DepleterSync`

src/result_processor.c
Itzikvaknin: Can `StrongRef_Get(self->sync_ref)` fail?
Implicit "error" handling here may be a risk

src/result_processor.c
```diff
@@ -1548,6 +1564,22 @@ static int RPDepleter_Next_Dispatch(ResultProcessor *base, SearchResult *r) {
...
   }
 
+  if (self->take_index_lock && !self->index_released) {
+    // Load the atomic counter
+    int num_locked = atomic_load(&((DepleterSync *)StrongRef_Get(self->sync_ref))->num_locked);
+    // We currently support only 2 upstream depleters. When we move on to support
+    // more subqueries - this will be generalized.
+    if (num_locked == 2) {
+      // Release the index
+      RedisSearchCtx_UnlockSpec(RP_SCTX(base));
```
Itzikvaknin, line 1596: If polling is used, this part will be executed as many times as there are depleters (?)
Can it lead to multiple unlocks?

src/result_processor.c
raz-mon: It exists in the header actually.
Will put it here as well

src/result_processor.c
raz-mon: Good question
Not really, since this would mean that we received a bad ref in the constructor.
I have some assertions in different places, I'll just put one in the constructor and remove the rest.

src/result_processor.c
raz-mon: I actually wanted to explicit about our current limitation for 2 upstreams, but since it is used in more places I'll generalize it.
We will have to remember that we cannot just generalize to more upstreams with the current logic.
I'll add a comment on that.

src/result_processor.c
raz-mon: Yes, it should.
Behaves the same but makes more sense.
2 is not a magic number! it's the current hard limit of upstreams we can have!
I'm changing it only because I want to use it in many places, WITH AN IMPORTANT comment that we cannot just generalize for many upstreams without updating the logic of the RP.

src/result_processor.c
raz-mon: I don't think it's needed here, only 3 booleans

src/result_processor.c
raz-mon: Updated such that we do this once (wouldn't matter to do more but obviously better not to)

src/result_processor.c
```diff
@@ -1569,23 +1569,23 @@ static int RPDepleter_Next_Dispatch(ResultProcessor *base, SearchResult *r) {
...
+  if (self->take_index_lock && !sync->index_released) {
     // Load the atomic counter
-    DepleterSync *sync = (DepleterSync *)StrongRef_Get(self->sync_ref);
     int num_locked = atomic_load(&sync->num_locked);
     if (num_locked == sync->n_depleters) {
       // Release the index
       RedisSearchCtx_UnlockSpec(RP_SCTX(base));
       // Mark the index as released
-      self->index_released = true;
+      sync->index_released = true;
```
Itzikvaknin, line 1598: Should we use it under locking/ atomic var?
`atomic_fetch_add`

src/spec.c
```diff
@@ -2336,6 +2336,7 @@ if (scanner->isDebug) { \
 #define IF_DEBUG_PAUSE_CHECK_ON_OOM(scanner, ctx) IF_DEBUG_PAUSE_CHECK(scanner, ctx, pauseOnOOM, DEBUG_INDEX_SCANNER_CODE_PAUSED_ON_OOM)
 
 static void Indexes_ScanAndReindexTask(IndexesScanner *scanner) {
+  printf("Got here (task), scanner: %p\n", scanner);
```
Itzikvaknin: Leftovers?

src/result_processor.c
raz-mon: No, only used from the pipeline thread

src/spec.c
raz-mon: Yes!

src/result_processor.c
```diff
@@ -1436,25 +1436,35 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+  bool first_call;                  // Whether the first call to Next has been made
+  StrongRef sync_ref;               // Reference to shared synchronization object (DepleterSync)
+  bool take_index_lock;             // Whether or not the depleter should take the index lock
 } RPDepleter;
 
 // Shared synchronization object for all RPDepleter instances of a pipeline
 typedef struct {
   pthread_cond_t cond;
   pthread_mutex_t mutex;
+  uint n_depleters;       // Number of depleters to sync
```
kei-nan:
```suggestion
  uint32_t num_depleters;       // Number of depleters to sync
```

src/result_processor.c
```diff
@@ -1466,10 +1476,11 @@ static void DepleterSync_Free(void *obj) {
 }
 
 // Create a new shared sync object for a pipeline
-StrongRef DepleterSync_New() {
-  DepleterSync *sync = rm_malloc(sizeof(DepleterSync));
+StrongRef DepleterSync_New(uint n_depleters) {
```
kei-nan:
```suggestion
StrongRef DepleterSync_New(uint32_t num_depleters) {
```

src/result_processor.c
```diff
@@ -1548,9 +1571,23 @@ static int RPDepleter_Next_Dispatch(ResultProcessor *base, SearchResult *r) {
     return RS_RESULT_DEPLETING;
   }
 
-  // On subsequent calls, wait on condition variable for any depleter to complete
   DepleterSync *sync = (DepleterSync *)StrongRef_Get(self->sync_ref);
-  RS_LOG_ASSERT(sync, "Invalid sync reference");
+
+  if (self->take_index_lock && !sync->index_released) {
+    // Load the atomic counter
+    int num_locked = atomic_load(&sync->num_locked);
```
kei-nan, line 1592: maybe add assertion it isn't above the depleter count?

src/result_processor.c
```diff
@@ -1436,25 +1436,35 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
-  bool first_call;                     // Whether the first call to Next has been made
-  StrongRef sync_ref;                  // Reference to shared synchronization object
+  ResultProcessor base;             // Base result processor struct
+  arrayof(SearchResult *) results;  // Array of pointers to SearchResult, filled by the depleting thread
+  bool done_depleting;              // Set to `true` when depleting is finished (under lock)
+  uint cur_idx;                     // Current index for yielding results
+  RPStatus last_rc;                 // Last return code from upstream
+  bool first_call;                  // Whether the first call to Next has been made
+  StrongRef sync_ref;               // Reference to shared synchronization object (DepleterSync)
+  bool take_index_lock;             // Whether or not the depleter should take the index lock
```
kei-nan: consider moving this to the sync struct as well.

src/result_processor.c
```diff
@@ -1436,25 +1436,35 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
 } RPDepleter;
 
 // Shared synchronization object for all RPDepleter instances of a pipeline
 typedef struct {
   pthread_cond_t cond;
   pthread_mutex_t mutex;
+  uint n_depleters;       // Number of depleters to sync
+  atomic_int num_locked;  // Number of depleters that have locked the index
+  bool index_released;    // Whether or not the index-spec has been released by the pipeline thread yet
```
kei-nan, line 1478: maybe add a bit more comments here, explaining why we need this and what it means to solve?

src/result_processor.c
raz-mon: ok

kei-nan (approved):

Itzikvaknin (approved):

PR #6428, opened 2025-07-03
MOD-10041 Parse FT.HYBRID command

ofiryanai:
## Parse the new FT.HYBRID command

See the new synax and discussion in [the PRD](https://redislabs.atlassian.net/wiki/spaces/DX/pages/3981803720/PRD+Hybrid+Search+RRF+and+Linear+Combination#:~:text=The%20current%20syntax%20used%20by%20the%20FT.SEARCH%20and%20FT.AGGREGATE%20provides%20limitations%20for%20what%20can%20be%20expressed%20in%20the%20query%20string%20and%20complex%20combinations%20and%20fusion%20logic.).

1. MOD-10106

#### Mark if applicable

- [V] This PR introduces API changes


src/hybrid/parse_hybrid.c
```diff
@@ -0,0 +1,489 @@
...
+  QueryError_Init(&req->tailError);
+  Pipeline_Initialize(req->tailPipeline, requests[0]->pipeline->qctx.timeoutPolicy, &req->tailError);
+
+  // Initialize pipelines for each individual request
+  for (size_t i = 0; i < nrequests; i++) {
+    QueryError_Init(&req->errors[i]);
+    Pipeline_Initialize(requests[i]->pipeline, requests[i]->reqConfig.timeoutPolicy, &req->errors[i]);
+  }
+  return req;
+}
```
Itzikvaknin: duplicate of `src/hybrid/hybrid_request.c`?

src/aggregate/aggregate.h
```diff
@@ -150,7 +150,7 @@ typedef struct AREQ {
   ConcurrentSearchCtx conc;
 
   /** The pipeline for this request */
-  Pipeline pipeline;
+  Pipeline *pipeline;
```
kei-nan: Why does it need to be a pointer?

src/hybrid/hybrid_request.c
```diff
@@ -101,39 +101,82 @@ int HybridRequest_BuildPipeline(HybridRequest *req, const HybridPipelineParams *
...
+        // Check if we need to manually free the thread-safe context
+        if (hybridReq->requests[i]->sctx && hybridReq->requests[i]->sctx->redisCtx) {
+
+            // Free the search context
+            RedisModuleCtx *thctx = hybridReq->requests[i]->sctx->redisCtx;
+            RedisSearchCtx *sctx = hybridReq->requests[i]->sctx;
+            SearchCtx_Free(sctx);
+            // Free the thread-safe context
+            if (thctx) {
+                RedisModule_FreeThreadSafeContext(thctx);
```
kei-nan: Are we sure free will only get called in the main thread context?
Since we are calling a redis api function here.

src/hybrid/parse_hybrid.c
```diff
@@ -0,0 +1,418 @@
...
+
+    // Unknown argument that's not VSIM - this is an error
+    QueryError_SetWithUserDataFmt(status, QUERY_EPARSEARGS, "Unknown parameter", " `%s` in SEARCH", cur);
+    return REDISMODULE_ERR;
+  }
+
+  return REDISMODULE_OK;
+}
+
+static int parseVectorSubquery(ArgsCursor *ac, AREQ *vectorRequest, QueryError *status) {
```
kei-nan, line 73: Doesn't look like the vectorRequest is used in the function, consider removing it.

src/hybrid/parse_hybrid.c
```diff
@@ -0,0 +1,418 @@
...
+    combineCtx->scoringType = HYBRID_SCORING_RRF;
+  } else {
+    combineCtx->scoringType = HYBRID_SCORING_RRF;
+  }
+
+  // Parse parameters based on scoring type
+  if (combineCtx->scoringType == HYBRID_SCORING_LINEAR) {
+    // For LINEAR, we expect exactly 2 weight values
+    combineCtx->linearCtx.linearWeights = rm_calloc(2, sizeof(double));
+    combineCtx->linearCtx.numWeights = 2;
```
kei-nan: try and avoid the magic numbers, I think we defined a value for the number of possible requests

src/hybrid/parse_hybrid.c
```diff
@@ -0,0 +1,418 @@
...
+
+  // If there are remaining arguments, parse them into the aggregate plan
+  bool hasMerge = false;
+  if (remainingArgs > 0) {
+    hasMerge = true;
+    mergeAreq = AREQ_New();
+
+    AGPLN_Init(AREQ_AGGPlan(mergeAreq));
+    RSSearchOptions_Init(&mergeAreq->searchopts);
+    if (parseAggPlan(mergeAreq, &ac, status) != REDISMODULE_OK) {
```
kei-nan: not a fan of the merge being an AREQ, I think it can lead to issues if not now then in the future.
IMHO we need a clear separation.
Maybe the function should accept all the arguments it requires from AREQ, if there are too many we can have a struct to hold all the arguments.

src/hybrid/parse_hybrid.c
```diff
@@ -0,0 +1,418 @@
...
+
+  if (hybridParams) {
+    HybridScoringContext_Free(hybridParams->scoringCtx);
+    rm_free(hybridParams);
+  }
+
+  return NULL;
+}
+
+int execHybrid(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
```
kei-nan: This function needs a better name and or documentation

src/hybrid/parse_hybrid.c
```diff
@@ -0,0 +1,418 @@
...
+// Copy the request configuration from the source to the destination
+static void copyReqConfig(AREQ *dest, const AREQ *src) {
+  dest->reqConfig.queryTimeoutMS = src->reqConfig.queryTimeoutMS;
+  dest->reqConfig.dialectVersion = src->reqConfig.dialectVersion;
+  dest->reqConfig.timeoutPolicy = src->reqConfig.timeoutPolicy;
+  dest->reqConfig.printProfileClock = src->reqConfig.printProfileClock;
+  dest->reqConfig.BM25STD_TanhFactor = src->reqConfig.BM25STD_TanhFactor;
+}
+
+HybridRequest* parseHybridRequest(RedisModuleCtx *ctx, RedisModuleString **argv, int argc,
```
kei-nan: some documentation would help

src/hybrid/parse_hybrid.c
```diff
@@ -0,0 +1,418 @@
...
+
+    // Not a stopping keyword, advance to next argument
+    AC_Advance(ac);
+  }
+
+  // TODO: Parse additional vector parameters (method, FILTER, etc.)
+  return REDISMODULE_OK;
+}
+
+static int parseCombine(ArgsCursor *ac, HybridScoringContext *combineCtx, QueryError *status) {
```
kei-nan, line 121: some documentation would be nice

tests/cpptests/redismock/util.h
```diff
@@ -85,6 +85,9 @@ class ArgvList {
   ArgvList(RedisModuleCtx *ctx, const char **s, size_t n) : m_ctx(ctx) {
     m_list = CreateArgv(ctx, s, n);
   }
+
+  explicit ArgvList(RedisModuleCtx *ctx, const char* command);
```
kei-nan: I think tsplitting a command should be a static function of this class which will return ArgvList, I think that will clean things and clarify stuff.

src/hybrid/parse_hybrid.c
ofiryanai: this is used in @Itzikvaknin 's [PR](https://github.com/RediSearch/RediSearch/pull/6482). on this PR this is just a stub

src/hybrid/parse_hybrid.c
ofiryanai: done

src/hybrid/parse_hybrid.c
ofiryanai: done

src/hybrid/parse_hybrid.c
ofiryanai: done

tests/cpptests/redismock/util.h
ofiryanai: This adds support for parsing full command strings like `ArgvList(ctx, "FT.HYBRID SEARCH hello")` vs individual args `ArgvList(ctx, "FT.HYBRID", "SEARCH", "hello")`. Not currently used but keeping for future reference. Can remove if preferred to avoid scope creep

src/hybrid/parse_hybrid.c
ofiryanai: extracted to `HYBRID_REQUEST_NUM_SUBQUERIES` constant
done

src/aggregate/aggregate.h
nafraf: To transfer the ownership to the hybridRequest.
Since we are using `mergeAreq` to create `tailPipeline`, when we free `mergeAreq` we need to free the AREQ without freeing its pipeline.
If we solve https://github.com/RediSearch/RediSearch/pull/6428#discussion_r2224529842 we can revert this change

src/hybrid/parse_hybrid.c
nafraf: But we are using AREQ also as argument of the functions invoked by `parseAggPlan()`, we'll need to refactor a lot of functions, it will require a big effort and could be risky.

src/aggregate/aggregate.h
nafraf: Change reverted.

src/hybrid/parse_hybrid.c
nafraf: Now we have a new structure `ParseAggPlanContext`, and the functions used to parse the hybrid were modified to avoid the use of `AREQ`

src/aggregate/aggregate.h
```diff
@@ -260,12 +284,12 @@ static inline QEFlags AREQ_RequestFlags(const AREQ *req) {
...
 
-static inline void AREQ_AddRequestFlags(AREQ *req, QEFlags flags) {
-  req->reqflags |= flags;
+static inline void AREQ_AddRequestFlags(uint32_t *reqflags, QEFlags flags) {
+  *reqflags |= flags;
 }
 
-static inline void AREQ_RemoveRequestFlags(AREQ *req, QEFlags flags) {
-  req->reqflags &= ~flags;
+static inline void AREQ_RemoveRequestFlags(uint32_t *reqflags, QEFlags flags) {
```
kei-nan: not sure about this specific change, we lose the context of the AREQ but function name still has AREQ...
We lose the abstraction layer for the AREQ since we will pass reqflags directly and not hide where the member sits.

I suggest to revert to previous function signatures.
In the parse functions you could set the flags directly without this function.

src/aggregate/aggregate_request.c
```diff
@@ -1291,8 +1319,8 @@ int AREQ_BuildPipeline(AREQ *req, QueryError *status) {
       .optimizer = req->optimizer,
     },
     .outFields = &req->outFields,
-    .maxResultsLimit = IsSearch(req) ? req->maxSearchResults : req->maxAggregateResults,
+    .maxResultsLimit = IsSearch(req) ? RSGlobalConfig.maxSearchResults : RSGlobalConfig.maxAggregateResults,
```
kei-nan: Isn't this something the user can specify per request?
Are we sure we aren't changing existing behaviour?
Are we still running in the main thread here?

src/hybrid/hybrid_request.c
```diff
@@ -129,10 +130,49 @@ HybridRequest *HybridRequest_New(AREQ **requests, size_t nrequests) {
...
+      AREQ_Free(req->requests[i]);
+    }
+    array_free(req->requests);
+
     array_free(req->errors);
-    Pipeline_Clean(&req->pipeline);  // Cleans up the merger and aggregation pipeline
+
+    // Free the scoring context resources
+    if (req->hybridParams) {
+      HybridScoringContext_Free(req->hybridParams->scoringCtx);
```
kei-nan: shouldn't this be released by the merger?

src/param.c
```diff
@@ -52,3 +52,35 @@ void Param_DictFree(dict *d) {
   dictReleaseIterator(iter);
   dictRelease(d);
 }
+
+dict *Param_DictClone(dict *source) {
```
kei-nan, line 56: is this function still needed?

tests/cpptests/redismock/util.h
kei-nan: I just wanted a it as a static function.
e.g static ArgvList Create(const char *cmd)

But if you are saying it is unused the consider removing it.

kei-nan (changes requested):
Hi,
Left some additional comments

src/aggregate/aggregate.h
nafraf: OK. Change reverted.
Now we are using a macro when we don't have access to AREQ.
`#define REQFLAGS_AddFlags(reqflags, flags) (*(reqflags) |= (flags))`

src/aggregate/aggregate_request.c
nafraf: It was an unintentionally change. Reverted.

src/hybrid/hybrid_request.c
nafraf: Yes. Fixed.

src/param.c
nafraf: Yes. We use it to clone `mergeSearchopts.params` to vector and search request.

kei-nan (approved):

PR #6429, opened 2025-07-03
[MOD-10326] Rust InvertedIndex - Writing only

chesedo:
Implements an `InvertedIndex` in Rust which only allows writing new records. Reading records will happen in another PR.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/inverted_index/src/freqs_only.rs
+++ b/src/redisearch_rs/inverted_index/src/freqs_only.rs
@@ -12,7 +12,7 @@ use std::io::{Read, Seek, Write};
 use ffi::t_docId;
 use qint::{qint_decode, qint_encode};
 
-use crate::{Decoder, DecoderResult, Delta, Encoder, RSIndexResult};
+use crate::{Decoder, DecoderResult, Encoder, RSIndexResult};
 
 /// Encode and decode only the delta and frequencies of a record, without any other data.
 /// The delta and frequency are encoded using [qint encoding](qint).
@@ -20,16 +20,14 @@ use crate::{Decoder, DecoderResult, Delta, Encoder, RSIndexResult};
 pub struct FreqsOnly;
 
 impl Encoder for FreqsOnly {
+    type DeltaType = u32;
+
     fn encode<W: Write + Seek>(
         &self,
         mut writer: W,
-        delta: Delta,
+        delta: Self::DeltaType,
         record: &RSIndexResult,
     ) -> std::io::Result<usize> {
-        let delta = delta
-            .0
-            .try_into()
-            .expect("FreqsOnly encoder only supports deltas that fit in u32");
         let bytes_written = qint_encode(&mut writer, [delta, record.freq])?;
         Ok(bytes_written)
     }
--- a/src/redisearch_rs/inverted_index/src/freqs_only.rs
+++ b/src/redisearch_rs/inverted_index/src/freqs_only.rs
@@ -23,7 +23,7 @@ impl Encoder for FreqsOnly {
     type DeltaType = u32;
 
     fn encode<W: Write + Seek>(
-        &self,
+        &mut self,
         mut writer: W,
         delta: Self::DeltaType,
         record: &RSIndexResult,
--- a/src/redisearch_rs/inverted_index/src/freqs_only.rs
+++ b/src/redisearch_rs/inverted_index/src/freqs_only.rs
@@ -20,12 +20,12 @@ use crate::{Decoder, DecoderResult, Encoder, RSIndexResult};
 pub struct FreqsOnly;
 
 impl Encoder for FreqsOnly {
-    type DeltaType = u32;
+    type Delta = u32;
 
     fn encode<W: Write + Seek>(
         &mut self,
         mut writer: W,
-        delta: Self::DeltaType,
+        delta: Self::Delta,
         record: &RSIndexResult,
     ) -> std::io::Result<usize> {
         let bytes_written = qint_encode(&mut writer, [delta, record.freq])?;
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -331,8 +331,10 @@ pub trait Encoder {
     /// allowing the inverted index to work around their limitations accordingly (i.e. by creating new blocks).
     type Delta: IdDelta;
 
-    /// Does this encoder allow the same document to appear in the index multiple times. Defaults
-    /// to `false`.
+    /// Does this encoder allow the same document to appear in the index multiple times. We need to
+    /// allow duplicates to support multi-value JSON fields.
+    ///
+    /// Defaults to `false`.
     const ALLOW_DUPLICATES: bool = false;
 
     /// The suggested number of entries that can be written in a single block. Defaults to 100.
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -382,7 +382,7 @@ pub struct InvertedIndex<E> {
 
     /// Number of unique documents in the index. This is not the total number of entries, but rather the
     /// number of unique documents that have been indexed.
-    num_docs: usize,
+    n_unique_docs: usize,
 
     /// The encoder to use when adding new entries to the index
     encoder: E,
@@ -442,7 +442,7 @@ impl<E: Encoder> InvertedIndex<E> {
     pub fn new(encoder: E) -> Self {
         Self {
             blocks: Vec::new(),
-            num_docs: 0,
+            n_unique_docs: 0,
             encoder,
         }
     }
@@ -502,7 +502,7 @@ impl<E: Encoder> InvertedIndex<E> {
         self.blocks.push(block);
 
         if !same_doc {
-            self.num_docs += 1;
+            self.n_unique_docs += 1;
         }
 
         Ok(buf_growth + mem_growth)
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -431,6 +431,7 @@ impl<E: Encoder> InvertedIndex<E> {
         let bytes_written = E::encode(buffer, Delta::new(delta as _), record)?;
 
         block.num_entries += 1;
+        block.last_doc_id = doc_id;
         self.last_doc_id = Some(doc_id);
 
         if !same_doc {
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -489,6 +489,10 @@ impl<E: Encoder> InvertedIndex<E> {
         let writer = block.writer();
         let _bytes_written = self.encoder.encode(writer, delta, record)?;
 
+        // We don't use `_bytes_written` returned by the encoder to determine by how much memory
+        // grew because the buffer might have had enough capacity for the bytes in the encoding.
+        // Instead we took the capacity of the buffer before the write and now check by how much it
+        // has increased (if any).
         let buf_growth = block.buffer.capacity() - buf_cap;
 
         block.num_entries += 1;
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -319,7 +319,11 @@ pub trait Encoder {
     /// Calculate the delta using the block it will be added to. Returns `None` if the delta is too
     /// big for this encoder and a new block should be created.
     fn calculate_delta(block: &IndexBlock, doc_id: t_docId) -> Option<Self::DeltaType> {
-        let delta = doc_id - block.last_doc_id;
+        debug_assert!(
+            doc_id >= block.last_doc_id,
+            "documents should be encoded in the order of their IDs"
+        );
+        let delta = doc_id.wrapping_sub(block.last_doc_id);
 
         if delta > u32::MAX as _ {
             // If the delta is larger than `u32::MAX`, we cannot encode it with this encoder.
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -318,6 +318,10 @@ impl PartialEq for RSIndexResult {
 
 /// Encoder to write a record into an index
 pub trait Encoder {
+    /// Does this encoder allow the same document to appear in the index multiple times. Defaults
+    /// to `false`.
+    const ALLOW_DUPLICATES: bool = false;
+
     /// Write the record to the writer and return the number of bytes written. The delta is the
     /// pre-computed difference between the current document ID and the last document ID written.
     ///
@@ -371,6 +375,7 @@ pub trait Decoder {
 
 pub struct InvertedIndex<E> {
     buffer: Vec<u8>,
+    last_doc_id: Option<t_docId>,
     _encoder: PhantomData<E>,
 }
 
@@ -378,16 +383,29 @@ impl<E: Encoder> InvertedIndex<E> {
     pub fn new() -> Self {
         Self {
             buffer: Vec::new(),
+            last_doc_id: None,
             _encoder: Default::default(),
         }
     }
 
     pub fn add_record(&mut self, record: &RSIndexResult) -> std::io::Result<usize> {
-        let buffer = Cursor::new(&mut self.buffer);
+        let doc_id = record.doc_id;
 
-        E::encode(buffer, Delta::new(record.doc_id as _), record)?;
+        if !E::ALLOW_DUPLICATES && self.last_doc_id.map(|d| d == doc_id).unwrap_or_default() {
+            // The encoder does not allow writting the same document to the same index twice. This
+            // can happen when the index is created with duplicate tags for example.
+            return Ok(0);
+        }
 
-        Ok(0)
+        let buffer_pos = self.buffer.len();
+        let mut buffer = Cursor::new(&mut self.buffer);
+        buffer.set_position(buffer_pos as _);
+
+        let bytes_written = E::encode(buffer, Delta::new(doc_id as _), record)?;
+
+        self.last_doc_id = Some(doc_id);
+
+        Ok(bytes_written)
     }
 }
 
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -390,6 +390,15 @@ struct IndexBlock {
 }
 
 impl IndexBlock {
+    fn new(doc_id: t_docId) -> Self {
+        Self {
+            first_doc_id: doc_id,
+            last_doc_id: doc_id,
+            num_entries: 0,
+            buffer: Vec::new(),
+        }
+    }
+
     fn writer(&mut self) -> Cursor<&mut Vec<u8>> {
         let pos = self.buffer.len();
         let mut buffer = Cursor::new(&mut self.buffer);
@@ -448,21 +457,17 @@ impl<E: Encoder> InvertedIndex<E> {
 
     fn get_last_block(&mut self, doc_id: t_docId) -> &mut IndexBlock {
         if self.blocks.is_empty() {
-            let block = IndexBlock {
-                first_doc_id: doc_id,
-                last_doc_id: doc_id,
-                num_entries: 0,
-                buffer: Vec::new(),
-            };
+            let block = IndexBlock::new(doc_id);
             self.blocks.push(block);
-        } else if self.blocks.last().unwrap().num_entries >= E::BLOCK_ENTRIES {
+        } else if self
+            .blocks
+            .last()
+            .expect("we just confirmed there are blocks")
+            .num_entries
+            >= E::BLOCK_ENTRIES
+        {
             // We need to create a new block since the last one is full
-            let block = IndexBlock {
-                first_doc_id: doc_id,
-                last_doc_id: doc_id,
-                num_entries: 0,
-                buffer: Vec::new(),
-            };
+            let block = IndexBlock::new(doc_id);
             self.blocks.push(block);
         }
 
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -10,7 +10,7 @@
 use std::{
     ffi::{c_char, c_int},
     fmt::Debug,
-    io::{Read, Seek, Write},
+    io::{Cursor, Read, Seek, Write},
     mem::ManuallyDrop,
 };
 
@@ -367,3 +367,28 @@ pub trait Decoder {
         }
     }
 }
+
+pub struct InvertedIndex {
+    buffer: Vec<u8>,
+}
+
+impl InvertedIndex {
+    pub fn new() -> Self {
+        InvertedIndex { buffer: Vec::new() }
+    }
+
+    pub fn add_record<E: Encoder>(
+        &mut self,
+        encoder: E,
+        record: &RSIndexResult,
+    ) -> std::io::Result<usize> {
+        let buffer = Cursor::new(&mut self.buffer);
+
+        E::encode(buffer, Delta::new(record.doc_id as _), record)?;
+
+        Ok(0)
+    }
+}
+
+#[cfg(test)]
+mod tests;
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -318,6 +318,8 @@ impl PartialEq for RSIndexResult {
 
 /// Encoder to write a record into an index
 pub trait Encoder {
+    type DeltaType: TryFrom<u64> + Default;
+
     /// Does this encoder allow the same document to appear in the index multiple times. Defaults
     /// to `false`.
     const ALLOW_DUPLICATES: bool = false;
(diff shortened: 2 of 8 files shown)


gdesmott:
@chesedo you can rebase on top of  `master` now that my branch has been merged. That should reduce the diff.

src/redisearch_rs/inverted_index/src/freqs_only.rs
```diff
@@ -12,24 +12,22 @@ use std::io::{Read, Seek, Write};
...
 
 impl Encoder for FreqsOnly {
+    type DeltaType = u32;
+
     fn encode<W: Write + Seek>(
-        &self,
+        &mut self,
         mut writer: W,
-        delta: Delta,
+        delta: Self::DeltaType,
```
gdesmott: Can we remove `Delta` then?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -317,19 +317,37 @@ impl PartialEq for RSIndexResult {
...
         writer: W,
-        delta: Delta,
+        delta: Self::DeltaType,
         record: &RSIndexResult,
     ) -> std::io::Result<usize>;
+
+    /// Calculate the delta using the block it will be added to. Returns `None` if the delta is too
+    /// big for this encoder and a new block should be created.
+    fn calculate_delta(block: &IndexBlock, doc_id: t_docId) -> Option<Self::DeltaType> {
+        let delta = doc_id - block.last_doc_id;
```
gdesmott: Should we enforce `doc_id > last_doc_id`?

src/redisearch_rs/inverted_index/src/numeric.rs
```diff
@@ -186,6 +189,11 @@ impl Numeric {
 
         self
     }
+
+    /// Returns the number of entries encoded by this encoder.
+    pub fn num_entries(&self) -> usize {
```
gdesmott, line 188: This seems to be used only by tests.

Is that really worth making the encoder stateful for this?
I'm not fan of `encode()` now taking `&mut self`.

gdesmott (commented):
Still need to look at the actual II code.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -367,3 +385,167 @@ pub trait Decoder {
...
+
+                Default::default()
+            }
+        };
+
+        let buf_cap = block.buffer.capacity();
+        let writer = block.writer();
+        let _bytes_written = self.encoder.encode(writer, delta, record)?;
+
+        let buf_growth = block.buffer.capacity() - buf_cap;
```
gdesmott, line 527: Probably worth a comment explaining the logic here.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -317,19 +317,37 @@ impl PartialEq for RSIndexResult {
 
 /// Encoder to write a record into an index
 pub trait Encoder {
+    type DeltaType: TryFrom<u64> + Default;
+
+    /// Does this encoder allow the same document to appear in the index multiple times. Defaults
+    /// to `false`.
+    const ALLOW_DUPLICATES: bool = false;
```
gdesmott, line 338: I'm wondering it it would not make more sense to have those in another trait II specific trait, as the actual encoders won't use this.

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -0,0 +1,222 @@
+use crate::{Encoder, InvertedIndex, RSIndexResult};
+
+/// Dummy encoder which allows defaults for testing
```
gdesmott:
```suggestion
/// Dummy encoder which allows defaults for testing, encoding only the delta
```

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -367,3 +385,167 @@ pub trait Decoder {
...
+
+        let same_doc = match (
+            E::ALLOW_DUPLICATES,
+            self.last_doc_id().map(|d| d == doc_id).unwrap_or_default(),
+        ) {
+            (true, true) => true,
+            (false, true) => {
+                // The encoder does not allow writting the same document to the same index twice. This
+                // can happen when the index is created with duplicate tags for example.
+                return Ok(0);
```
gdesmott, line 487: Would it be worth having a more explicit API and return an enum? Something like:

```rust
enum AddRecordResult {
  Added(usize),
  AlreadyThere,
}
```

src/redisearch_rs/inverted_index/src/freqs_only.rs
chesedo: Yes, it seems so. I can't find anything in the inverted index that will need it again. Good catch :partying_face:

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Hmm, the current C code does not and just underflows instead. I think a debug assert will make sense to make sure the code is structured correctly though.

src/redisearch_rs/inverted_index/src/numeric.rs
chesedo: Yes, numeric index will eventually use it to find some median here

https://github.com/RediSearch/RediSearch/blob/b77f80f4cb7c8b8ef5081b67c96bda17d7f183d3/src/numeric_index.c#L88-L90

chesedo: The `fieldMask` is also used for only some of the encoders / indices

https://github.com/RediSearch/RediSearch/blob/b77f80f4cb7c8b8ef5081b67c96bda17d7f183d3/src/inverted_index/inverted_index.h#L48-L51

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Only the numeric encoder will set this as it is the only encoder which allows the same doc to be written multiple times

https://github.com/RediSearch/RediSearch/blob/b77f80f4cb7c8b8ef5081b67c96bda17d7f183d3/src/inverted_index/inverted_index.c#L466-L475

src/redisearch_rs/inverted_index/src/numeric.rs
gdesmott: Would it not make more sense to have the object encapsulating the actual encoder (`InvertedIndex` atm) handle this?

Just a suggestion, as it seems to be a better design to me. Feel free to ignore it.

src/redisearch_rs/inverted_index/src/lib.rs
gdesmott: I think what bother me here is merging the low level encoders (numeric, freq only, etc) and the II wrapping those encoders.

The later seems different enough to have its own trait I think. It could depend of the `Encoder` trait so we can keep using its API.

Again, just a design suggestion.

src/redisearch_rs/inverted_index/src/numeric.rs
chesedo: I wanted to initially, But that would require `InvertedIndex` to identify each encoder. Ie the `InvertedIndex` would have to go "oh, my encoder is the numeric encoder so I should increase the `numEntries`". I don't like this because:
1. `InvertedIndex` knowing the encoder breaks the abstraction of the encoders
2. The `InvertedIndex` will have a `numEntries` field (or an enum) but won't always need to use it (since only some encoders need it)

gdesmott: Agreed, that makes sense.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Hmmm, I don't see the current usages of the code benefiting from such an enum. Ie. I think they will just convert the `AlreadyThere` variant to `0` anyway

src/redisearch_rs/inverted_index/src/freqs_only.rs
```diff
@@ -12,24 +12,22 @@ use std::io::{Read, Seek, Write};
...
-use crate::{Decoder, DecoderResult, Delta, Encoder, RSIndexResult};
+use crate::{Decoder, DecoderResult, Encoder, RSIndexResult};
 
 /// Encode and decode only the delta and frequencies of a record, without any other data.
 /// The delta and frequency are encoded using [qint encoding](qint).
 #[derive(Default)]
 pub struct FreqsOnly;
 
 impl Encoder for FreqsOnly {
+    type DeltaType = u32;
```
LukeMathWalker:
```suggestion
    type Delta = u32;
```
No need to use Polish naming conventions.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -367,3 +370,171 @@ pub trait Decoder {
...
+pub struct InvertedIndex<E> {
+    /// The blocks of the index. Each block contains a set of entries for a specific range of
+    /// document IDs. The entries and blocks themselves are ordered by document ID, so the first
+    /// block contains entries for the lowest document IDs, and the last block contains entries for
+    /// the highest document IDs.
+    blocks: Vec<IndexBlock>,
+
+    /// Number of unique documents in the index. This is not the total number of entries, but rather the
+    /// number of unique documents that have been indexed.
+    num_docs: usize,
```
```suggestion
    n_unique_docs: usize,
```

LukeMathWalker: It makes code using it a lot more obvious.

src/redisearch_rs/inverted_index/src/lib.rs
gdesmott: Fair enough.

src/redisearch_rs/inverted_index/src/freqs_only.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6429/commits/2eba35de4463f5b89a6196405d4c28d1c9fb8bc3

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -20,6 +20,30 @@ pub use ffi::{RSDocumentMetadata, RSQueryTerm, RSYieldableMetric, t_docId, t_fie
 pub mod freqs_only;
 pub mod numeric;
 
+/// Trait used to correctly derive the delta needed for different encoders
+pub trait Delta
```
```suggestion
pub trait DocIdDelta
```

Perhaps?

chesedo: I saw the "Perhaps?" too late. So rename done :sweat_smile:

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -20,6 +20,30 @@ pub use ffi::{RSDocumentMetadata, RSQueryTerm, RSYieldableMetric, t_docId, t_fie
...
+where
+    Self: Sized,
+{
+    /// Try to convert a `u64` into this delta type. If the `u64` will be too big for this delta
+    /// type, then `None` should be returned. Returning `None` will cause the [`InvertedIndex`]
+    /// to make a new block to reset the delta.
+    fn from_u64(delta: u64) -> Option<Self>;
+
+    /// The delta value when the [`InvertedIndex`] has to reset it for a new block.
+    fn reset() -> Self;
```
LukeMathWalker: Is this ever different from 0 for that specific delta type?

chesedo: No, it's always the type's zero

chesedo: I did want to name it `zero()` initially but the comments flowed better with the `reset()`

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -367,3 +381,179 @@ pub trait Decoder {
         }
     }
 }
+
+/// An inverted index is a data structure that maps terms to their occurrences in documents. It is
+/// used to efficiently search for documents that contain specific terms.
+pub struct InvertedIndex<E> {
```
gdesmott, line 396: Any specific reason you did not use `pub struct InvertedIndex<E: Encoder>` here?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -367,3 +381,179 @@ pub trait Decoder {
...
+    pub fn new(encoder: E) -> Self {
+        Self {
+            blocks: Vec::new(),
+            n_unique_docs: 0,
+            encoder,
+        }
+    }
+
+    /// Add a new record to the index and return by how much memory grew. It is expected that
+    /// the document ID of the record is greater than or equal the last document ID in the index.
```
gdesmott, line 474: This is true only for encoders returning the last id as base.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Yip, this is the idiomatic Rust way with the constraint being on the `impl`... if I remember correctly :sweat_smile:

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Correct, but I don't see why callers would want to exploit that. The way I understand it is that II will write records in increasing DocId order.

src/redisearch_rs/inverted_index/src/lib.rs
gdesmott: @LukeMathWalker out of curiosity, do you know the rational for this?

LukeMathWalker: Two main reasons:
- The type may offer some basic functionality even if the bound is not met, thus allowing each method to have more granular constraints
- You don't have to repeat the bound every single time you're passing that type around

gdesmott: That makes sense, thanks!

src/redisearch_rs/inverted_index/src/lib.rs
LukeMathWalker: I would still go for `zero()`, since it's much clearer what the expectations are on the implementer. But up to you really.

src/redisearch_rs/inverted_index/src/numeric.rs
```diff
@@ -202,13 +196,39 @@ impl Default for Numeric {
     }
 }
 
+/// The [`Numeric`] encoder only supports encoding deltas that fit within 7 bytes
+#[derive(Debug, PartialEq)]
+pub struct NumericDelta(pub u64);
```
LukeMathWalker: The wrapped value shouldn't be pub.

chesedo: Yeah, that's lazy old me trying to find a way to make it easy to construct in tests :sweat_smile: . I'll add a `new()` or something

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Eg of what I mean with the comment. I realize "reset" equates to "zero" in all cases. But in terms of code flow "reset" just seems to take less cognitive reasoning

https://github.com/RediSearch/RediSearch/blob/ccedfbfaa6fedac85bf0139662c26a70b2ab60cc/src/redisearch_rs/inverted_index/src/lib.rs#L494-L503

The comment has "reset" and it is clear that we are resetting at the end of the scope. Now I can make the comment say something about "create a new zero delta" but that feels harder to follow

src/redisearch_rs/inverted_index/src/numeric.rs
```diff
@@ -222,6 +222,13 @@ impl IdDelta for NumericDelta {
     }
 }
 
+impl NumericDelta {
+    /// Creates a new [`NumericDelta`] from a given delta value
+    pub fn new(delta: u64) -> Self {
+        Self(delta)
+    }
+}
```
LukeMathWalker: Not all u64s are valid for this type, no? We should reject the ones larger than 7 bytes here too.

chesedo: Correct, but if I also do the 7 bytes check here, then it won't be possible to have this test

https://github.com/RediSearch/RediSearch/blob/ccedfbfaa6fedac85bf0139662c26a70b2ab60cc/src/redisearch_rs/inverted_index/tests/numeric.rs#L576-L600

Only the tests and benchmarks make use of this `new()` method -> https://github.com/RediSearch/RediSearch/pull/6429/commits/ccedfbfaa6fedac85bf0139662c26a70b2ab60cc

LukeMathWalker: What we're missing is an `inner()` method that returns the underlying `u64`, which is what you'd need in those tests to write assertions.

It doesn't matter where it's invoked _now_, it's `pub` and it becomes a way to bypass the invariants that the type is supposed to enforce.

src/redisearch_rs/inverted_index/src/lib.rs
LukeMathWalker: I would phrase it as:

```rust
// The delta is too large for this encoder. We need to create a new block.
// Since the new block is empty, we'll start with `delta` equal to 0.

File: src/redisearch_rs/inverted_index/src/numeric.rs (hunk 15 above)
Comment (chesedo) [reply src/redisearch_rs/inverted_index/src/numeric.rs]: Ahhh, yes that will work :+1:

Comment (chesedo) [reply src/redisearch_rs/inverted_index/src/numeric.rs]: Done in https://github.com/RediSearch/RediSearch/pull/6429/commits/db12f0acb3c51493316d7ecc2765dec31bf75dd6

File: src/redisearch_rs/inverted_index/src/lib.rs [hunk 16]
```diff
@@ -317,19 +322,28 @@ impl PartialEq for RSIndexResult {
 
 /// Encoder to write a record into an index
 pub trait Encoder {
+    type Delta: IdDelta;
```
LukeMathWalker, line 332:
```suggestion
    /// Document ids are represented as `u64`s and stored using delta-encoding.
    /// 
    /// Some encoders can't encode arbitrarily large id deltas—e.g. they might be limited to `u32::MAX` or 
    /// another subset of the `u64` value range.
    ///
    /// This associated type can be used by each encoder to specify which range they support, thus
    /// allowing the inverted index to work around their limitations accordingly (i.e. by creating new blocks).
    type Delta: IdDelta;
```

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -317,19 +322,28 @@ impl PartialEq for RSIndexResult {
...
+
+    /// Does this encoder allow the same document to appear in the index multiple times. Defaults
+    /// to `false`.
+    const ALLOW_DUPLICATES: bool = false;
```
LukeMathWalker: I don't think the way this is worded is correct—isn't the duplication check restricted to the current block? Could the same document appear multiple times in _different_ blocks? Or are we guaranteed that's not the case?

chesedo: So the same doc can't appear in different blocks because of two things:
1. Docs are inserted in incremental DocId order
2. When two consecutive docs have the same ID then a new block will never be created. Even if it reached the recommended capacity

https://github.com/RediSearch/RediSearch/blob/db12f0acb3c51493316d7ecc2765dec31bf75dd6/src/redisearch_rs/inverted_index/src/lib.rs#L536-L547

This means in all cases that a doc appears multiple times in the index, the doc is appearing multiple times in one singular block

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: That's perfect. Done in https://github.com/RediSearch/RediSearch/pull/6429/commits/722dc1f863342e8da43532a7ac133f2a8835ceb6

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -317,19 +322,35 @@ impl PartialEq for RSIndexResult {
...
+    ///
+    /// Some encoders can't encode arbitrarily large id deltas—e.g. they might be limited to `u32::MAX` or
+    /// another subset of the `u64` value range.
+    ///
+    /// This associated type can be used by each encoder to specify which range they support, thus
+    /// allowing the inverted index to work around their limitations accordingly (i.e. by creating new blocks).
+    type Delta: IdDelta;
+
+    /// Does this encoder allow the same document to appear in the index multiple times. Defaults
+    /// to `false`.
```
alonre24: The purpose of allowing duplicates is to support multi-value in JSON fields - consider adding this to the comments

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -367,3 +388,179 @@ pub trait Decoder {
...
+
+impl IndexBlock {
+    const SIZE: usize = std::mem::size_of::<Self>();
+
+    /// Make a new index block with the given initial doc ID. This returns the block and how much
+    /// memory grew by.
+    pub fn new(doc_id: t_docId) -> (Self, usize) {
+        let this = Self {
+            first_doc_id: doc_id,
+            last_doc_id: doc_id,
```
alonre24, line 443: Is this expected to be a dummy `doc_id`? Since `num_entries` is zero and the buffer is empty at this point

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -367,3 +388,179 @@ pub trait Decoder {
...
+
+        let same_doc = match (
+            E::ALLOW_DUPLICATES,
+            self.last_doc_id().map(|d| d == doc_id).unwrap_or_default(),
+        ) {
+            (true, true) => true,
+            (false, true) => {
+                // The encoder does not allow writing the same document to the same index twice. This
+                // can happen when the index is created with duplicate tags for example.
+                return Ok(0);
```
alonre24: I don't understand that part. Is it expected to have duplicate tags for an index that does not allow duplicates?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -367,3 +388,179 @@ pub trait Decoder {
...
+            "documents should be encoded in the order of their IDs"
+        );
+        let delta = doc_id.wrapping_sub(delta_base);
+
+        let delta = match E::Delta::from_u64(delta) {
+            Some(delta) => delta,
+            None => {
+                // The delta is too large for this encoder. We need to create a new block.
+                // Since the new block is empty, we'll start with `delta` equal to 0.
+                let (new_block, block_size) = IndexBlock::new(doc_id);
```
alonre24, line 508: I'm not sure that I got the logic here - when will `wrapping_sub(delta_base)` is expected to return None?

src/redisearch_rs/inverted_index/tests/freqs_only.rs
```diff
@@ -101,13 +99,3 @@ fn test_decode_freqs_only_empty_input() {
     let kind = res.unwrap_err().kind();
     assert_eq!(kind, std::io::ErrorKind::UnexpectedEof);
 }
-
-#[test]
-#[should_panic]
-fn test_encode_freqs_only_delta_overflow() {
```
alonre24, line 107: What has changed? Why is this test no longer valid?

alonre24 (commented):
Looks good, few questions

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: No, this is priming a new block. It is only called here

https://github.com/RediSearch/RediSearch/blob/722dc1f863342e8da43532a7ac133f2a8835ceb6/src/redisearch_rs/inverted_index/src/lib.rs#L542-L553

Which in turn is called here, and then the block is written to

https://github.com/RediSearch/RediSearch/blob/722dc1f863342e8da43532a7ac133f2a8835ceb6/src/redisearch_rs/inverted_index/src/lib.rs#L489-L516

alonre24: I see. It's a bit awkward IMO, but clear. If you can find a way in which we slightly change the logic so there will not be this "in-between" state, I believe it would be more readable. If not, let's just clearly document it

src/redisearch_rs/inverted_index/src/lib.rs
alonre24: Let's just sharpen the comment - we allow duplicate doc IDs even if `ALLOW_DUPLICATES` is false, but we ignore them intentionally, since the input is redundant

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6429/commits/83696ea16e14c9a088156557d105240a441386d7

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6429/commits/5156434af09a7851af7f478250e30dd1c02a7d7f

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6429/commits/69acbfa282cbaf74bb8780ed5c058f5bee14d7b9

src/redisearch_rs/inverted_index/tests/freqs_only.rs
chesedo: The code has been updated to no longer panic when the delta overflows. Instead it gracefully makes a new block. The logic that governs this is tested here

https://github.com/RediSearch/RediSearch/blob/722dc1f863342e8da43532a7ac133f2a8835ceb6/src/redisearch_rs/inverted_index/src/tests.rs#L233-L241

alonre24 (approved):

PR #6431, opened 2025-07-03
`RLookupKey` transliteration

JonasKruckenberg:
First "ugly" porting of `RLookupKey` to Rust. This PR just adds the struct and most basic methods. Leaving future work to followup PRs.

Note that there is one diversion from the C code: The `RLookupKey` flag `NameAlloc` is now transient too (i.e. only an argument to the methods and not persisted into the key itself) since the allocation state is already tracked via the `_name` and `_path` `Cow` fields.

### DIFF
--- a/src/redisearch_rs/rlookup/src/lookup.rs
+++ b/src/redisearch_rs/rlookup/src/lookup.rs
@@ -73,7 +73,8 @@ const GET_KEY_FLAGS: RLookupKeyFlags =
     make_bitflags!(RLookupKeyFlag::{Override | Hidden | ExplicitReturn | ForceLoad});
 
 /// Flags do not persist to the key, they are just options to [`RLookup::get_key_read`], [`RLookup::get_key_write`], or [`RLookup::get_key_load`].
-const TRANSIENT_FLAGS: RLookupKeyFlags = make_bitflags!(RLookupKeyFlag::{Override | ForceLoad});
+const TRANSIENT_FLAGS: RLookupKeyFlags =
+    make_bitflags!(RLookupKeyFlag::{Override | ForceLoad | NameAlloc});
 
 /// RLookup key
 ///
--- a/src/redisearch_rs/rlookup/src/lookup.rs
+++ b/src/redisearch_rs/rlookup/src/lookup.rs
@@ -48,7 +48,7 @@ pub enum RLookupKeyFlag {
     /// field for another consumer. Don't output this field.
     Hidden = 0x100,
 
-    /// The opposite of F_HIDDEN. This field is specified as an explicit return in
+    /// The opposite of [`RLookupKeyFlag::Hidden`]. This field is specified as an explicit return in
     /// the RETURN list, so ensure that this gets emitted. Only set if
     /// explicitReturn is true in the aggregation request.
     ExplicitReturn = 0x200,
@@ -139,7 +139,7 @@ pub struct RLookupKey<'a> {
 
     /// The name of this key.
     pub name: *const c_char,
-    /// The length of this key, without the null-terminator.
+    /// The length of this key in bytes, without the null-terminator.
     /// Should be used to avoid repeated `strlen` computations.
     pub name_len: usize,
 
@@ -274,6 +274,28 @@ mod tests {
 
         let key = RLookupKey::new(name, RLookupKeyFlags::empty());
         assert_eq!(key.name, name.as_ptr());
+        assert!(matches!(key._name, Cow::Borrowed(_)));
+    }
+
+    // Assert that creating a RLookupKey with the NameAlloc flag indeed allocates a new string
+    #[test]
+    fn rlookupkey_new_utf8_with_namealloc() {
+        let name = c"🔍🔥🎶";
+
+        let key = RLookupKey::new(name, make_bitflags!(RLookupKeyFlag::NameAlloc));
+        assert_ne!(key.name, name.as_ptr());
+        assert_eq!(key.name_len, 12); // 3 characters, 4 bytes each
+        assert!(matches!(key._name, Cow::Owned(_)));
+    }
+
+    // Assert that creating a RLookupKey *without* the NameAlloc flag keeps the provided string
+    #[test]
+    fn rlookupkey_new_utf8_without_namealloc() {
+        let name = c"🔍🔥🎶";
+
+        let key = RLookupKey::new(name, RLookupKeyFlags::empty());
+        assert_eq!(key.name, name.as_ptr());
+        assert_eq!(key.name_len, 12); // 3 characters, 4 bytes each
         assert!(matches!(key._name, Cow::Borrowed(_)));
     }
 }
--- a/src/redisearch_rs/rlookup/src/lookup.rs
+++ b/src/redisearch_rs/rlookup/src/lookup.rs
@@ -120,38 +120,38 @@ const TRANSIENT_FLAGS: RLookupKeyFlags =
 #[repr(C)]
 pub struct RLookupKey<'a> {
     /// Index into the dynamic values array within the associated `RLookupRow`.
-    dstidx: u16,
+    pub dstidx: u16,
 
     /// If the source for this key is a sorting vector, this is the index
     /// into the `RSSortingVector` within the associated `RLookupRow`.
-    svidx: u16,
+    pub svidx: u16,
 
     /// Various flags dictating the behavior of looking up the value of this key.
     /// Most notably, `Flags::SVSRC` means the source is an `RSSortingVector` and
     /// `Self::svidx` should be used to look up the value.
-    flags: RLookupKeyFlags,
+    pub flags: RLookupKeyFlags,
 
     /// The path of this key.
     ///
     /// For fields *not* loaded from a [`FieldSpec`][ffi::FieldSpec], this points to the *same* string
     /// as `Self::path`.
-    path: *const c_char,
+    pub path: *const c_char,
 
     /// The name of this key.
-    name: *const c_char,
+    pub name: *const c_char,
     /// The length of this key, without the null-terminator.
     /// Should be used to avoid repeated `strlen` computations.
-    name_len: usize,
+    pub name_len: usize,
 
     /// Pointer to next field in the list
     #[pin]
-    next: Option<NonNull<RLookupKey<'a>>>,
+    pub next: Option<NonNull<RLookupKey<'a>>>,
 
     // Private Rust fields
     /// The actual "owning" strings, we need to hold onto these
     /// so the pointers above stay valid. Note that you
-    /// MUST NEVER MOVE THESE BEFORE THE name AND path FIELDS LESS
-    /// YOU WANT POTENTIALLY RISK UB
+    /// MUST NEVER MOVE THESE BEFORE THE name AND path FIELDS UNLESS
+    /// YOU WANT TO POTENTIALLY RISK UB
     #[pin]
     _name: Cow<'a, CStr>,
     #[pin]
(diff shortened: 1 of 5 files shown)


src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    /// Pointer to next field in the list
+    #[pin]
+    next: Option<NonNull<RLookupKey<'a>>>,
+
+    // Private Rust fields
+    /// The actual "owning" strings, we need to hold onto these
+    /// so the pointers above stay valid.
+    _name: Cow<'a, CStr>,
+    _path: Option<Cow<'a, CStr>>,
```
LukeMathWalker: Are we saying that the type is self-referential?

JonasKruckenberg: yup! thats what all the pinning stuff is for again. But this is temporary only, the "pointer only" fields will go away after we encapsulate Rlookup

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,248 @@
...
+            svidx: 0,
+            flags: flags & !TRANSIENT_FLAGS,
+            name: name.as_ptr().cast(),
+            path: name.as_ptr().cast(),
+            name_len: name.count_bytes(),
+            _name: name,
+            _path: None,
+            next: None,
+        }
+    }
```
LukeMathWalker: Don't we need `name` to be pinned _before_ we call `as_ptr`?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,248 @@
...
+        let ptr = Box::into_raw(unsafe { Pin::into_inner_unchecked(me) });
+
+        // Safety: we know the ptr we get from Box::into_raw is never null
+        unsafe { NonNull::new_unchecked(ptr) }
+    }
+
+    /// Constructs a `Pin<Box<RLookupKey>>` from a raw pointer.
+    ///
+    /// The returned `Box` will own the raw pointer, in particular dropping the `Box`
+    /// will deallocate the `ResultProcessor`. This function should only be used by [`RLookup::drop`].
```
DarthB:
```suggestion
    /// will deallocate the `RLookupKey`. This function should only be used by [`RLookup::drop`].
```

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: I mean like teeeechnically, maybe? I don't think in this situation it really doesn't matter. The pointer is being treated *as pinned* throughout (never moved out of).

LukeMathWalker: Let's err on the side of correctness—it won't make a difference to C callers (we'll use unsafe either way) but it clarifies the requirements.

JonasKruckenberg: Sure, what's your suggestion?

LukeMathWalker: Thinking about it

```rust
enum PinnedCStr<'a> {
    Borrowed(&'a Pin<Box<CStr>>),
    Owned(Pin<Box<CStr>>),
}
```

seems like the kind of semantic we're looking for here, with `&'a Pin<Box<CStr>>` as the input parameter type.

LukeMathWalker: But my pin-fu is quite rusty, so it might or might not fit.

JonasKruckenberg: I don't think this pulls its weight honestly, the `Cow<CStr>` is already pinned anyways through `pin_project`. We must be careful about only ever returning `&CStr` OR `Pin<&mut CStr>` (so consumers cannot move out of the memory location) but internally this doesn't pose much of a hazard I don't think.

JonasKruckenberg: so the concern we have is that someone could accidentally move out of `_name` in the process invalidating the `name` ptr field right? The way I see it is that, since `name` is already pinned *after* construction through `pin_project`, and we never hand out `&mut Cow<CStr>`, `*mut Cow<CStr>`, or `NonNull<Cow<CStr>>` there is not way for *safe* Rust to move out of `_name`. 

What we're concerned about the are 2 scenarios:
1. `Cow::Borrowed` - The CStr is borrowed immutably. When the string came from C FFI code, it has to promise to never memcpy it. If the string came from Rust, the livetime enforces that no mutable borrow can exist for the same CStr while we hold it. No risk of moving out of the string therefore. 
2. `Cow::Owned` - The CStr is owned by the RLookupKey. Here we just need to make sure to never hand out a mutable reference to the `Cow` so callers cannot use that too move the string (this applies to 1. too ofc).

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,278 @@
...
+
+// ===== impl RLookupKey =====
+
+// SAFETY NOTICE
+//
+// This type contains self-referential fields (e.g. `name` points to memory owned by `_name`) and therefore
+// must be pinned at all times. This means in practice, to only ever hand out one of two types of references to the
+// string: either an immutable `&CStr` - safe Rust cannot move out of an immutable reference - or a pinned mutable
+// reference `Pin<&mut CStr>` which safe Rust also cannot move out of.
+// This means you may NEVER EVER hand out a `&mut CStr` EVER.
```
JonasKruckenberg, line 169: @LukeMathWalker

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,278 @@
...
+    borrow::Cow,
+    ffi::{CStr, c_char},
+    pin::Pin,
+    ptr::NonNull,
+};
+
+#[bitflags]
+#[repr(u32)]
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum RlookupKeyFlag {
```
LukeMathWalker:
```suggestion
pub enum RLookupKeyFlag {
```

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,278 @@
...
+#[derive(Debug, PartialEq, Eq)]
+#[pin_project]
+#[repr(C)]
+pub struct RLookupKey<'a> {
+    /// Index into the dynamic values array within the associated `RLookupRow`.
+    dstidx: u16,
+
+    /// If the source for this key is a sorting vector, this is the index
+    /// index into the `RSSortingVector` within the associated `RLookupRow`.
+    svidx: u16,
```
LukeMathWalker:
```suggestion
    /// If the source for this key is a sorting vector, this is the index
    /// into the `RSSortingVector` within the associated `RLookupRow`.
    svidx: u16,
```

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,279 @@
...
+    name_len: usize,
+
+    /// Pointer to next field in the list
+    #[pin]
+    next: Option<NonNull<RLookupKey<'a>>>,
+
+    // Private Rust fields
+    /// The actual "owning" strings, we need to hold onto these
+    /// so the pointers above stay valid. Note that you
+    /// MUST NEVER MOVE THESE BEFORE THE name AND path FIELDS LESS
```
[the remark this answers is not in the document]

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,279 @@
...
+
+// SAFETY NOTICE
+//
+// This type contains self-referential fields (e.g. `name` points to memory owned by `_name`) and therefore
+// must be pinned at all times. This means in practice, to only ever hand out one of two types of references to the
+// string: either an immutable `&CStr` - safe Rust cannot move out of an immutable reference - or a pinned mutable
+// reference `Pin<&mut CStr>` which safe Rust also cannot move out of.
+// This means you may NEVER EVER hand out a `&mut CStr` EVER.
+impl<'a> RLookupKey<'a> {
+    /// Constructs a new `RLookupKey` using te provided `CStr` and flags.
```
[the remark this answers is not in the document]

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    /// The path of this key.
+    ///
+    /// For fields *not* loaded from a [`FieldSpec`][ffi::FieldSpec], this points to the *same* string
+    /// as `Self::path`.
+    pub path: *const c_char,
+
+    /// The name of this key.
+    pub name: *const c_char,
+    /// The length of this key, without the null-terminator.
```
oshadmi: The bytes length? (using `count_bytes()` when setting it)

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,279 @@
...
+        let ptr = unsafe { RLookupKey::into_ptr(key) };
+        let key = unsafe { RLookupKey::from_ptr(ptr) };
+
+        assert_eq!(*key, RLookupKey::new(c"test", RLookupKeyFlags::empty()));
+    }
+
+    // Assert that creating a RLookupKey with the NameAlloc flag indeed allocates a new string
+    #[test]
+    fn rlookupkey_new_with_namealloc() {
+        let name = c"test";
```
oshadmi, line 263: Do we need a test with UTF-8 CStr? 😺 to make sure length is byte-length?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,279 @@
...
+    ForceLoad = 0x40,
+
+    /// This key is unresolved. Its source needs to be derived from elsewhere
+    Unresolved = 0x80,
+
+    /// This field is hidden within the document and is only used as a transient
+    /// field for another consumer. Don't output this field.
+    Hidden = 0x100,
+
+    /// The opposite of F_HIDDEN. This field is specified as an explicit return in
```
oshadmi:
```suggestion
    /// The opposite of HIDDEN. This field is specified as an explicit return in
```

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    // Private Rust fields
+    /// The actual "owning" strings, we need to hold onto these
+    /// so the pointers above stay valid. Note that you
+    /// MUST NEVER MOVE THESE BEFORE THE name AND path FIELDS UNLESS
+    /// YOU WANT TO POTENTIALLY RISK UB
+    #[pin]
+    _name: Cow<'a, CStr>,
+    #[pin]
+    _path: Option<Cow<'a, CStr>>,
```
oshadmi, line 158: Is this the "ugly" part as mentioned in the PR description?
Where we kind of mimic the C behavior of reusing the path pointer to point to the name when they are identical, etc.? And later on, this could be improved to more idiomatic way such as using an enum with single Cow for single value or with double Cow, one for name and one for path, etc., ?

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: bytes yeah, updated the comment to reflect this 👍🏻

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: Good call, added two new test cases for this

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: changed to a Rust doc comment link

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: Yes, part of it is mimicking the C behavior with the `name` and `path` fields, so existing C code remains happy while at the same time having these two field so that Rust code is happy too. After cleanup we'll remove the `name`/`path` raw pointer fields in favor of these two `Cow` fields.

PR #6432, opened 2025-07-03
add `RlookupKey::update_from_field_spec`

JonasKruckenberg:
This PR adds the `RlookupKey::update_from_field_spec` translating the C `setKeyByFieldSpec` function to Rust. This also adds temporary definitions for `FieldSpecOption` and `FieldSpecType` until MOD-10333 is addressed to keep the Rust code clean.

Stacks on top of #6431


src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -0,0 +1,498 @@
...
+
+        if fs_options.contains(FieldSpecOption::Sortable) {
+            self.flags |= RlookupKeyFlag::SvSrc;
+            self.svidx = u16::try_from(fs.sortIdx).unwrap();
+
+            if fs_options.contains(FieldSpecOption::Unf) {
+                // If the field is sortable and not normalized (UNF), the available data in the
+                // sorting vector is the same as the data in the document.
+                self.flags |= RlookupKeyFlag::ValAvailable;
+            }
```
DarthB: The data is the same if it is already normalized, or do I misunderstand?
I think, this is the other way round, see [unf query in c](https://github.com/RediSearch/RediSearch/commit/fa57fd2de377424327afe43f989562210bb226f7#diff-90e39f5f0dfc1344b1b22db34414d3623c2d53cfedcb89e870e24fbb55c93d36L392-R393) and [original string put code](https://github.com/RediSearch/RediSearch/commit/fa57fd2de377424327afe43f989562210bb226f7#diff-2619b7927348152fd98b110f12b1f3683f0a2e13f0f1a793b5b179cd9c704fb7L78-L79)

```suggestion
            if !fs_options.contains(FieldSpecOption::Unf) {
                // If the field is sortable and normalized (not UNF), the available data in the
                // sorting vector is the same as the data in the document.
                self.flags |= RlookupKeyFlag::ValAvailable;
            }
```

JonasKruckenberg: well, idk 🤷‍♂️ its just what the C code does https://github.com/RediSearch/RediSearch/blob/master/src/rlookup.c#L105

see the IsUNF function here too https://github.com/RediSearch/RediSearch/blob/master/src/field_spec.h#L148

(i don't actually know what UNF is supposed to mean)

DarthB: The places in C we referenced are contradicting 🤷‍♂️

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -239,10 +242,62 @@ impl<'a> RLookupKey<'a> {
...
+            // Safety: we received the pointer from the field spec and have to assume it is valid
+            let path_ptr =
+                unsafe { ffi::HiddenString_GetUnsafe(fs.fieldPath, ptr::from_mut(&mut path_len)) };
+
+            debug_assert!(!path_ptr.is_null());
+            // Safety: We assume the `path_ptr` and `length` information returned by the field spec
+            // point to a valid null-terminated C string. Importantly `length` here is value as returned by
+            // `strlen` so **does not** include the null terminator (that is why we do `path_len + 1` below)
+            let bytes = unsafe { slice::from_raw_parts(path_ptr.cast::<u8>(), path_len + 1) };
+            let path = CStr::from_bytes_with_nul(bytes)
```
oshadmi: We can consider using `from_bytes_with_nul_unchecked` if we know the pointer points to a valid null-terminated char string with no embedded nulls, but I suppose this is safer and the performance gain is relatively small (or consider using checked version only when cfg has debug_assertions?)

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -239,10 +242,62 @@ impl<'a> RLookupKey<'a> {
...
+            }
+        };
+        self._path = Some(path);
+        self.path = self._path.as_ref().unwrap().as_ptr();
+
+        let fs_options = FieldSpecOptions::from_bits(fs.options()).unwrap();
+
+        if fs_options.contains(FieldSpecOption::Sortable) {
+            self.flags |= RLookupKeyFlag::SvSrc;
+            self.svidx = u16::try_from(fs.sortIdx).unwrap();
```
oshadmi: Similar question here - can we rely on the values coming from C and be more relaxed when reading them, e.g., use `from_unchecked`? If not, can add a clear error message using `expect`?

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: While I am a bit hesitant about "proliferating" the use of unchecked methods like this. Ideally we would go for correctness and then tune and optimize for performance later, but I defer to your experience with this code. If you say this is likely important I am fine with using the unchecked version and adding debug assertions to make this safe for `debug` builds.

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: yes we can, see above.

src/redisearch_rs/rlookup/src/lookup.rs
oshadmi: Actually this is probably not a critical path so we can go for correctness and see later if we need to optimize.

src/redisearch_rs/rlookup/src/lookup.rs
oshadmi: UNF is Unformatted 😊

oshadmi: > if (FieldSpec_IsUnf(fs)) {
>       // If the field is sortable and not normalized (UNF), the available data in the
>       // sorting vector is the same as the data in the document.
>       key->flags |= RLOOKUP_F_VAL_AVAILABLE;
>     }

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: okay 👍🏻 lets go with that for now

oshadmi (approved):

PR #6433, opened 2025-07-03
RLookup: KeyList linked-list implementation

JonasKruckenberg:
This implements the "linked-list of keys" component and associated methods for `RLookup`. This type abstracts the actual storage of keys from the Rest of the RLookup implementation.

Stacks on top of #6431

### DIFF
--- a/src/redisearch_rs/rlookup/src/lookup.rs
+++ b/src/redisearch_rs/rlookup/src/lookup.rs
@@ -166,6 +166,8 @@ pub struct RLookupKey<'a> {
 #[derive(Debug)]
 #[repr(C)]
 struct KeyList<'a> {
+    // The head and tail nodes of this linked-list.
+    // FIXME [MOD-10314] make this more type-safe when we no longer have direct field access from C
     head: Option<NonNull<RLookupKey<'a>>>,
     tail: Option<NonNull<RLookupKey<'a>>>,
     // Length of the data row. This is not necessarily the number
@@ -312,7 +314,7 @@ impl<'a> RLookupKey<'a> {
         assert_eq!(
             self.name_len,
             self._name.count_bytes(),
-            "{ctx}`key.name_len` did not match `key._name` lentgh"
+            "{ctx}`key.name_len` did not match `key._name` length"
         );
 
         if ptr::eq(self, tail) {
--- a/src/redisearch_rs/rlookup/src/lookup.rs
+++ b/src/redisearch_rs/rlookup/src/lookup.rs
@@ -191,18 +191,6 @@ struct LinkInner<'a> {
     _unpin: PhantomPinned,
 }
 
-/// Iterates over the keys in a [`KeyList`] by reference.
-pub struct Iter<'list, 'a> {
-    _rlookup: &'list KeyList<'a>,
-    curr: Option<NonNull<RLookupKey<'a>>>,
-}
-
-/// Iterates over the keys in a [`KeyList`] by mutable reference.
-pub struct IterMut<'list, 'a> {
-    _rlookup: &'list mut KeyList<'a>,
-    curr: Option<NonNull<RLookupKey<'a>>>,
-}
-
 // ===== impl RLookupKey =====
 
 // SAFETY NOTICE
@@ -339,52 +327,6 @@ impl<'a> KeyList<'a> {
         // Safety: We treat the pointer as pinned internally and never hand out references that could be moved out of (in safe Rust)
         // publicly.
         unsafe { Pin::new_unchecked(key) }
-    }
-
-    /// Returns an Iterator over the `RLookupKey`s in this `KeyList`.
-    ///
-    /// The iteration order is **not specified** and must not be relied on for correctness.
-    pub fn iter(&self) -> Iter<'_, 'a> {
-        #[cfg(debug_assertions)]
-        self.assert_valid("KeyList::iter");
-
-        Iter {
-            _rlookup: self,
-            curr: self.head,
-        }
-    }
-
-    /// Returns an Iterator over the `RLookupKey`s in this `KeyList`.
-    ///
-    /// The iteration order is **not specified** and must not be relied on for correctness.
-    pub fn iter_mut(&mut self) -> IterMut<'_, 'a> {
-        #[cfg(debug_assertions)]
-        self.assert_valid("KeyList::iter_mut");
-
-        IterMut {
-            curr: self.head,
-            _rlookup: self,
-        }
-    }
-
-    /// Find a [`RLookupKey`] in this `KeyList` by its [`name`][RLookupKey::name]
-    /// and return an immutable reference to it if found.
-    fn find_by_name(&self, name: &'a CStr) -> Option<&RLookupKey<'a>> {
-        #[cfg(debug_assertions)]
-        self.assert_valid("KeyList::find");
-
-        // FIXME [MOD-10315] replace with more efficient search
-        self.iter().find(|key| key._name.as_ref() == name)
-    }
-
-    /// Find a [`RLookupKey`] in this `KeyList` by its [`name`][RLookupKey::name]
-    /// and return a mutable reference to it if found.
-    fn find_by_name_mut(&mut self, name: &'a CStr) -> Option<Pin<&mut RLookupKey<'a>>> {
-        #[cfg(debug_assertions)]
-        self.assert_valid("KeyList::find_mut");
-
-        // FIXME [MOD-10315] replace with more efficient search
-        self.iter_mut().find(|key| key._name.as_ref() == name)
     }
 
     /// Asserts as many of the linked list's invariants as possible.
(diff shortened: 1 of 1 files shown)


src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -240,6 +277,280 @@ impl<'a> RLookupKey<'a> {
...
+    pub const fn new() -> Self {
+        Self {
+            head: None,
+            tail: None,
+            rowlen: 0,
+        }
+    }
+
+    fn push(&mut self, mut key: RLookupKey<'a>) -> Pin<&mut RLookupKey<'a>> {
+        key.dstidx = u16::try_from(self.rowlen).unwrap();
```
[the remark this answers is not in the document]
JonasKruckenberg: not a concern right now

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -157,6 +157,26 @@ pub struct RLookupKey<'a> {
     _path: Option<Cow<'a, CStr>>,
 }
 
+#[derive(Debug)]
+#[repr(C)]
+struct KeyList<'a> {
+    head: Option<NonNull<RLookupKey<'a>>>,
+    tail: Option<NonNull<RLookupKey<'a>>>,
+    // Length of the data row. This is not necessarily the number
+    // of lookup keys
```
DarthB: As the only mutator is `push` for now the rowlen is the number of `RLookupKey` will that change in follow up PRs?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -275,4 +615,104 @@ mod tests {
...
+        unsafe {
+            assert!(foo.as_ref().next.has_next());
+        }
+        unsafe {
+            assert!(!bar.as_ref().next.has_next());
+        }
+    }
+
+    #[test]
+    fn keylist_find() {
```
DarthB, line 917: Test the not there case too

JonasKruckenberg: good point, addressed it now 👍🏻

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: Yes, the `overrideKey` method (because the list is singly linked) can only add keys, but those keys are "hidden" and act more as "tombstone values"

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -157,6 +160,48 @@ pub struct RLookupKey<'a> {
...
 }
 
+/// An append-only list of [`RlookupKey`]s.
+///
+/// This type allows the creation and retrieval of [`RlookupKey`]s.
+#[derive(Debug)]
+#[repr(C)]
+struct KeyList<'a> {
+    head: Option<NonNull<RLookupKey<'a>>>,
+    tail: Option<NonNull<RLookupKey<'a>>>,
```
LukeMathWalker, line 174: Since both are either `None` or `Some` at the same time, wouldn't it be better to model them as an `Option<ListPointers>` where 

```rust
struct KeyList<'a> {
    pointers: Option<ListPointers<'a>>,
    row_length: u32
}

struct ListPointers<'a> {
    head: NonNull<RLookupKey<'a>>,
    tail: NonNull<RLookupKey<'a>>
}
```

Name open to bikeshedding, but you get the idea.

LukeMathWalker: Alternatively:

```rust
struct KeyList<'a>(Option<KeyListInner<'a>>);

struct KeyListInner<'a> {
    head: NonNull<RLookupKey<'a>>,
    tail: NonNull<RLookupKey<'a>>,
    row_length: u32
}
```

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -157,6 +160,48 @@ pub struct RLookupKey<'a> {
...
+pub struct Cursor<'list, 'a> {
+    _rlookup: &'list KeyList<'a>,
+    curr: Option<NonNull<RLookupKey<'a>>>,
+}
+
+/// A cursor over a [`KeyList`] with editing operations.
+pub struct CursorMut<'list, 'a> {
+    _rlookup: &'list mut KeyList<'a>,
+    curr: Option<NonNull<RLookupKey<'a>>>,
+}
```
LukeMathWalker:
```suggestion
/// A cursor over a [`KeyList`].
pub struct Cursor<'list, 'a> {
    _rlookup: &'list KeyList<'a>,
    current: Option<NonNull<RLookupKey<'a>>>,
}

/// A cursor over a [`KeyList`] with editing operations.
pub struct CursorMut<'list, 'a> {
    _rlookup: &'list mut KeyList<'a>,
    current: Option<NonNull<RLookupKey<'a>>>,
}
```

Let's avoid abbreviations where they aren't strictly necessary.

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -240,6 +283,378 @@ impl<'a> RLookupKey<'a> {
...
+    }
+
+    /// Insert a `RLookupKey` into this `KeyList` and return a mutable reference to it.
+    ///
+    /// The key will be owned by the list and freed when dropping the list.
+    fn push(&mut self, mut key: RLookupKey<'a>) -> Pin<&mut RLookupKey<'a>> {
+        #[cfg(debug_assertions)]
+        self.assert_valid("KeyList::push before");
+
+        key.dstidx = u16::try_from(self.rowlen).unwrap();
```
LukeMathWalker: Why is it OK to unwrap here?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -240,6 +283,378 @@ impl<'a> RLookupKey<'a> {
...
+
+        let mut c = self.cursor_front_mut();
+        while let Some(key) = c.current() {
+            if key._name.as_ref() == name {
+                return Some(c);
+            }
+            c.move_next();
+        }
+        None
+    }
```
LukeMathWalker: In both cases we're not returning a reference to the key, but rather a cursor with its current position set to that key.

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -240,6 +283,378 @@ impl<'a> RLookupKey<'a> {
...
+        }
+    }
+
+    /// Returns `true` if this node is currently linked to a [`List`].
+    #[cfg(test)]
+    pub fn has_next(&self) -> bool {
+        self.next().is_some()
+    }
+
+    fn unlink(&mut self) {
```
LukeMathWalker: Docs—and I assume we need to detail how this ties in with `Drop` above?

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: Yes, I would want this, but unfortunately C code currently directly accesses these fields so we're tied to this layout. (See MOD-10314 for the work item to clean this up)

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: oh no 😂 I mean fair enough, but do you know how many merge conflicts this is gonna cause 😢

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: It probably isn't but the C code does a silent truncating cast from `u32` to `u16` so having an explicit cast seems better. I guess I can improve the error message though.

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: I just moved it into `Drop` proper, and added a comment

src/redisearch_rs/rlookup/src/lookup.rs

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: wasn't as bad as I thought!

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: I added a more helpful panic message.

src/redisearch_rs/rlookup/src/lookup.rs
LukeMathWalker: Gotcha—let's add a comment with a link to the ticket then.

JonasKruckenberg: added comment 👍🏻

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -239,6 +266,392 @@ impl<'a> RLookupKey<'a> {
...
+    /// Update the pointer to the next node
+    #[inline]
+    fn set_next(
+        &mut self,
+        next: Option<NonNull<RLookupKey<'a>>>,
+    ) -> Option<NonNull<RLookupKey<'a>>> {
+        mem::replace(self.next.get_mut(), next)
+    }
+
+    fn assert_valid(&self, tail: &Self, ctx: &str) {
```
oshadmi, line 296: Can be conditioned on `#[cfg(debug_assertions)]` ?

oshadmi: or `#[cfg(any(debug_assertions, test))]`

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -239,6 +266,392 @@ impl<'a> RLookupKey<'a> {
...
+
+        // Safety: RLookup never hands out mutable references to the key (except `Pin<&mut T>` which is fine)
+        // and never copies, or memmoves the memory internally.
+        let mut ptr = unsafe { RLookupKey::into_ptr(Box::pin(key)) };
+
+        if let Some(mut tail) = self.tail.take() {
+            // if we have a tail we also must have a head
+            debug_assert!(self.head.is_some());
+
+            // Safety: We know we can borrow tail here, since we mutably borrow the RLookup
```
oshadmi: We mutably borrow KeyList or RLookup?

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: I made the function itself explicitly not `#[cfg(debug_assertions)]` because I figured we might want the option to assert consistency at very specific points even in release builds. 
But at the moment at least all *callsites* of this function are guarded behind `#[cfg(debug_assertions)]` so this will only ever be called in `debug` mode.

Edit: I added the `cfg` to this one too, its easy to remove later if we end up needing it.

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: Good catch, leftover from a refactor. It's borrowing the `KeyList` of course. Changed 👍🏻

oshadmi (approved):
LGTM 👍🏼

oshadmi (approved):

PR #6434, opened 2025-07-03
Initial `RLookup` transliteration - MOD-10393

JonasKruckenberg:
First "ugly" porting of `RLookup` to Rust. This diff contains nothing but the crate setup, type definitions and `RLookup` constructors/destructor.

Stacks on top of #6431, #6432, #6433

PR #6435, opened 2025-07-03
add `RLookup::get_key_read` method.

JonasKruckenberg:
This diff adds the `RLookup::get_key_read`, and `RLookup::gen_key_from_spec` methods translating the C `RLookup_GetKey_Read`, `RLookup_GetKey_ReadEx`, and `genKeyFromSpec` respectively.

Stacks on top of #6431, #6432, #6433, #6434


src/redisearch_rs/rlookup/src/bindings.rs
```diff
@@ -78,11 +97,9 @@ impl IndexSpecCache {
 
     pub fn find_field(&self, name: &CStr) -> Option<&ffi::FieldSpec> {
         self.fields().iter().find(|field| {
-            debug_assert!(!field.fieldName.is_null());
```
kei-nan, line 81: is this no longer true?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -760,14 +767,83 @@ impl<'a> RLookup<'a> {
...
+        // Safety: The non-lexical lifetime analysis of current Rust, incorrectly handles borrows in early
+        // return statements, expanding the borrow of `self` to the scope of the entire method. This is
+        // obviously not correct as `key` is either found and borrowed in which case we never reach the code below
+        // this early return OR it is *not* found and *not* borrowed which means the code below is fine to
+        // borrow self again. The current compiler is not smart enough to get this though, so we create a disjoint
+        // borrow below.
+        // TODO remove once <https://github.com/rust-lang/rust/issues/54663> is fixed.
+        let me = unsafe { &*(self as *const Self) };
+        if let Some(c) = me.keys.find_by_name(name) {
+            return Some(c.into_current().unwrap());
```
kei-nan: into_current().unwrap() seems odd.
looks like something is created and then something is extracted from it.
maybe the creation is a redundant step?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -1309,6 +1546,40 @@ mod tests {
         drop(unsafe { Box::from_raw(value.cast_mut().cast::<UserString>()) });
     }
 
+    /// Returns the following:
+    /// - 0, if the strings are the same
+    /// - a negative value if left is less than right
+    /// - a positive value if left is greater than right
+    #[unsafe(no_mangle)]
+    extern "C" fn HiddenString_CompareC(
```
kei-nan: should this function exist in this file or should it maybe go to a separate file?

src/redisearch_rs/rlookup/src/lookup.rs
DarthB: A small wrapper type `Cursor` is generated; in release, those things should be optimized out. We did this as the first step toward an idiomatic Rust implementation. 

We introduced the `Cursor` and `CursorMut` types, which provide idiomatic Rust iterator functionality on top of the `RLookupKey` structure. `RLookupKey` still adheres to the C-linked list memory schema, allowing legacy for-loops in the remaining C code. 

Functions like `find_by_name` return a `Cursor`; therefore, we need these two steps for now: Transform the `Cursor` into an `Option<&RLookupKey>` and `unwrap` the latter (the cursor always points to a valid reference when returned as Some from `find_by_name`.

src/redisearch_rs/rlookup/src/bindings.rs
DarthB: Not sure why that went away, readded: https://github.com/RediSearch/RediSearch/pull/6435/commits/f8424d7d484451b363c6403037543f347c36422f

src/redisearch_rs/rlookup/src/lookup.rs
DarthB: I extracted it in another file: https://github.com/RediSearch/RediSearch/pull/6435/commits/f8424d7d484451b363c6403037543f347c36422f

nafraf (approved):

PR #6437, opened 2025-07-03
add `RLookup::get_key_load` - MOD-10397

JonasKruckenberg:
This diff adds the `RLookup::get_key_load` method translating the C `RLookup_GetKey_Load` and `RLookup_GetKey_LoadEx` functions.

Stacks on top of #6431, #6432, #6433, #6434

### DIFF
--- a/src/redisearch_rs/rlookup/src/lookup.rs
+++ b/src/redisearch_rs/rlookup/src/lookup.rs
@@ -17,7 +17,7 @@ use std::{
     ffi::{CStr, c_char},
     mem,
     ops::DerefMut,
-    pin::Pin,
+    pin::{Pin, pin},
     ptr::{self, NonNull},
     slice,
 };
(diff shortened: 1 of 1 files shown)


src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -997,6 +994,107 @@ impl<'a> RLookup<'a> {
...
+        //    (no need to load it from the document).
+
+        // Safety: The non-lexical lifetime analysis of current Rust, incorrectly handles borrows in early
+        // return statements, expanding the borrow of `self` to the scope of the entire method. This is
+        // obviously not correct as `key` is either found and borrowed in which case we never reach the code below
+        // this early return OR it is *not* found and *not* borrowed which means the code below is fine to
+        // borrow self again. The current compiler is not smart enough to get this though, so we create a disjoint
+        // borrow below.
+        // TODO remove once <https://github.com/rust-lang/rust/issues/54663> is fixed.
+        let me = unsafe { &mut *(self as *mut Self) };
```
LukeMathWalker: This is dynamite waiting to explode. Can we restructure the rest of the method to work around the issue and avoid this unsafe block?

JonasKruckenberg: unsure, fundamentally these methods are all `get_or_insert` shaped. To my knowledge no easy solution exists for this that's not polonius. That only leaves large restructuring as an option, which idk... also doesn't feel great.

JonasKruckenberg: or `polonius-the-crab`

LukeMathWalker: A third workaround, at least in this phase, would be to go for "successive non-idiomatic lookups, as `polonius-the-crab` calls them.
At this stage, I'd be more focused on obvious correctness than perf, especially given how convoluted this "get or insert" is in practice.

DarthB: I invested some time in a previous PR, `get_key_write`, which has the same false positive pattern. 
And came to that solution too: `To my knowledge no easy solution exists for this that's not polonius.`

However, I think `polonius-the-crab` is a good fit here. With anonymous lifetimes in macros, we get some security against mistakes during refactoring.

nafraf (approved):

PR #6438, opened 2025-07-03
Simplify legacy RDB support - [MOD-10352]

GuyAv46:

1. Current: We create modern types from legacy RDBs, just to be ignored and released when the loading event ends
2. Change: Remove legacy code, keep only RDB loading logic to consume the RDB without attempting to create the actual data structures
3. Outcome: Simpler support, without needing to implement dedicated constructors (especially as we migrate to Rust)

1. Legacy types

#### Mark if applicable

### DIFF
--- a/src/legacy_types.c
+++ b/src/legacy_types.c
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+#include "legacy_types.h"
+#include "util/misc.h"
+#include "rmutil/rm_assert.h"
+#include <stdbool.h>
+
+#define LEGACY_ENC_VER 1
+#define LEGACY_LEGACY_ENC_VER 0
+
+// RDB load callback cannot return NULL, as it indicates an error
+void *dummyNoNull = (void*)0xDEADBEEF;
+
+// Dummy no-op functions for type methods
+void GenericType_DummyRdbSave(RedisModuleIO *rdb, void *value) {
+  RS_ABORT("Attempted to save a legacy type to RDB");
+}
+
+void GenericType_DummyFree(void *value) {
+  RS_ASSERT(value == dummyNoNull);
+}
+
+// Consume an inverted index type from RDB
+void *InvertedIndex_RdbLoad_Consume(RedisModuleIO *rdb, int encver) {
+  if (encver > LEGACY_ENC_VER) {
+    return NULL;
+  }
+
+  RedisModule_LoadUnsigned(rdb); // Consume the flags of the index
+  RedisModule_LoadUnsigned(rdb); // Consume the lastId of the index
+  RedisModule_LoadUnsigned(rdb); // Consume the number of documents in the index
+  size_t n_blocks = RedisModule_LoadUnsigned(rdb); // Load the number of blocks in the index
+
+  for (size_t i = 0; i < n_blocks; i++) {
+    RedisModule_LoadUnsigned(rdb); // Consume the firstId of the block
+    RedisModule_LoadUnsigned(rdb); // Consume the lastId of the block
+    RedisModule_LoadUnsigned(rdb); // Consume the number of entries in the block
+    RedisModule_Free(RedisModule_LoadStringBuffer(rdb, NULL)); // Consume the buffer of the block
+  }
+  return dummyNoNull;
+}
+
+// Consume a numeric index type from RDB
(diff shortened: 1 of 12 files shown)


src/module.c
```diff
@@ -1234,19 +1235,15 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx) {
     return REDISMODULE_ERR;
   }
 
-  // register trie type
+  // register trie-map type
```
raz-mon: Isn't this a dictionary type? With `Trie`s as its values?
If so, why `trie-map`?

src/trie/trie_type.c
```diff
@@ -374,10 +374,6 @@ void TrieType_GenericSave(RedisModuleIO *rdb, Trie *tree, int savePayloads) {
   }
 }
 
-void TrieType_Digest(RedisModuleDigest *digest, void *value) {
```
raz-mon, line 377: 🥇

src/legacy_types.c
```diff
@@ -0,0 +1,112 @@
...
+  for (size_t i = 0; i < n_tags; i++) {
+    RedisModule_Free(RedisModule_LoadStringBuffer(rdb, NULL)); // Consume the tag value
+    InvertedIndex_RdbLoad_Consume(rdb, encver); // Consume the inverted index for the tag
+  }
+  return dummyNoNull;
+}
+
+int RegisterLegacyTypes(RedisModuleCtx *ctx) {
+
+  RedisModuleTypeMethods tm = {
```
raz-mon, line 86: The `mem_usage` method is not set now, can this be a problem in any way?

src/numeric_index.h
```diff
@@ -123,16 +115,9 @@ void NumericRangeTree_Free(NumericRangeTree *t);
...
-
 NumericRangeTree *openNumericKeysDict(IndexSpec* spec, RedisModuleString *keyName, bool create_if_missing);
 
-int NumericIndexType_Register(RedisModuleCtx *ctx);
-void *NumericIndexType_RdbLoad(RedisModuleIO *rdb, int encver);
-void NumericIndexType_RdbSave(RedisModuleIO *rdb, void *value);
-void NumericIndexType_Digest(RedisModuleDigest *digest, void *value);
-void NumericIndexType_Free(void *value);
-unsigned long NumericIndexType_MemUsage(const void *value);
+unsigned long NumericIndexType_MemUsage(const NumericRangeTree *tree);
```
raz-mon, line 120: Used/Needed?

src/legacy_types.c
GuyAv46: According to the [documentation](https://redis.io/docs/latest/develop/reference/modules/modules-api-ref/#section-modules-data-types) (and I also verified in the `redis` code), it is an optional callback
```
  RedisModuleTypeMethods tm = {
      .version = REDISMODULE_TYPE_METHOD_VERSION,
      .rdb_load = myType_RDBLoadCallBack,
      .rdb_save = myType_RDBSaveCallBack,
      .aof_rewrite = myType_AOFRewriteCallBack,
      .free = myType_FreeCallBack,

      // Optional fields
      .digest = myType_DigestCallBack,
      .mem_usage = myType_MemUsageCallBack,
      .aux_load = myType_AuxRDBLoadCallBack,
      .aux_save = myType_AuxRDBSaveCallBack,
      .free_effort = myType_FreeEffortCallBack,
      .unlink = myType_UnlinkCallBack,
      .copy = myType_CopyCallback,
      .defrag = myType_DefragCallback

      // Enhanced optional fields
      .mem_usage2 = myType_MemUsageCallBack2,
      .free_effort2 = myType_FreeEffortCallBack2,
      .unlink2 = myType_UnlinkCallBack2,
      .copy2 = myType_CopyCallback2,
  }
```

src/numeric_index.h
GuyAv46: This method, specifically, is in use today, so I only made the type explicit

src/module.c
GuyAv46: Good catch. I saw that we register the trie later, and this is a dictionary+trie, so I assumed it's the trie-map

src/numeric_index.h
raz-mon: Where is it in use? I see only a debug command calling it..

GuyAv46: Isn’t that a good enough reason 😅? It is returning the correct memory consumption of the tree. We can also start using it in info commands

raz-mon: Wouldn't it always be `0`, since we don't use this type in any of the versions this is going to be backported to?

GuyAv46: This function returns the memory consumption of the numeric tree, regardless of whether we register it as a type known to redis or not

raz-mon: Oh it measures the memory usage of a tree it takes as an argument. I thought it measured the key-space usage of the index that was saved there.
OK

PR #6460, opened 2025-07-07
add assert when adding to dict on ft.create

BenGoldberger:
Add assert to create function that the dictAdd retuned OK, and log if not, so that we'll be alerted if there is an issue in the future

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -537,6 +537,7 @@ IndexSpec *IndexSpec_CreateNew(RedisModuleCtx *ctx, RedisModuleString **argv, in
     RedisModule_Log(ctx, "warning", "Failed adding index to global dictionary");
     StrongRef_Release(spec_ref);
     RS_ABORT("dictAdd shouldn't fail here - index shouldn't exists in the dictionary");
+    return NULL;
   }
   // Start the garbage collector
   IndexSpec_StartGC(ctx, spec_ref, sp);
--- a/src/spec.c
+++ b/src/spec.c
@@ -533,11 +533,10 @@ IndexSpec *IndexSpec_CreateNew(RedisModuleCtx *ctx, RedisModuleString **argv, in
   }
 
   // Add the spec to the global spec dictionary
-  int res = dictAdd(specDict_g, name, spec_ref.rm);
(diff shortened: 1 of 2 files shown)


tests/pytests/test_multithread.py
```diff
@@ -470,8 +470,8 @@ def expected(conns):
 def test_change_workers_number():
 
     def check_threads(expected_num_threads_alive, expected_n_threads):
-        env.assertEqual(getWorkersThpoolStats(env)['numThreadsAlive'], expected_num_threads_alive)
-        env.assertEqual(getWorkersThpoolNumThreads(env), expected_n_threads)
+        env.assertEqual(getWorkersThpoolStats(env)['numThreadsAlive'], expected_num_threads_alive, depth=1)
+        env.assertEqual(getWorkersThpoolNumThreads(env), expected_n_threads, depth=1)
```
GuyAv46: Let's also add a `message` with the variable name so we can see which of the assertions had failed.

src/spec.c
```diff
@@ -533,7 +533,12 @@ IndexSpec *IndexSpec_CreateNew(RedisModuleCtx *ctx, RedisModuleString **argv, in
   }
 
   // Add the spec to the global spec dictionary
-  dictAdd(specDict_g, name, spec_ref.rm);
+  int res = dictAdd(specDict_g, name, spec_ref.rm);
+  RS_ASSERT(res == DICT_OK);
+  if (res != DICT_OK) {
+    RedisModule_Log(ctx, "warning", "Failed adding index to global dictionary");
+    StrongRef_Release(spec_ref);
+  }
```
GuyAv46:
```suggestion
  if (dictAdd(specDict_g, name, spec_ref.rm) != DICT_OK) {
    RedisModule_Log(ctx, "warning", "Failed adding index to global dictionary");
    StrongRef_Release(spec_ref);
    RS_ABORT("dictAdd shouldn't fail here - index shouldn't exists in the dictionary");
  }
```

GuyAv46: Also, shouldn't we return (NULL) here?

PR #6470, opened 2025-07-08
add `CursorMut::override_current` - MOD-10391

JonasKruckenberg:
This PR adds the `CursorMut::override_current` method translating the C `overrideKey` function.

stacks on top of #6433

### DIFF
--- a/src/redisearch_rs/rlookup/src/lookup.rs
+++ b/src/redisearch_rs/rlookup/src/lookup.rs
@@ -276,7 +276,8 @@ impl<'a> RLookupKey<'a> {
         }
     }
 
-    pub fn from_parts(
+    /// Construct an `RLookupKey` from its main parts. Prefer Self::new if you are unsure which to use.
+    fn from_parts(
         name: Cow<'a, CStr>,
         path: Option<Cow<'a, CStr>>,
         dstidx: u16,
@@ -285,13 +286,13 @@ impl<'a> RLookupKey<'a> {
         debug_assert_eq!(
             matches!(name, Cow::Owned(_)),
             flags.contains(RLookupKeyFlag::NameAlloc),
-            "RLookupKeyFlag::NameAlloc but name is Cow::Borrowed"
+            "`RLookupKeyFlag::NameAlloc` was provided, but `name` was not `Cow::Owned`"
         );
         if let Some(path) = &path {
             debug_assert_eq!(
                 matches!(path, Cow::Owned(_)),
                 flags.contains(RLookupKeyFlag::NameAlloc),
-                "RLookupKeyFlag::NameAlloc but path is Cow::Borrowed"
+                "`RLookupKeyFlag::NameAlloc` was provided, but `path` was not `Cow::Owned`"
             );
         }
 
@@ -300,7 +301,9 @@ impl<'a> RLookupKey<'a> {
             svidx: 0,
             flags: flags & !TRANSIENT_FLAGS,
             name: name.as_ptr(),
-            path: name.as_ptr(),
+            // if a separate path was provided we should set the pointer accordingly
+            // if not, we fall back to the name as usual
+            path: path.as_ref().map_or(name.as_ptr(), |path| path.as_ptr()),
             name_len: name.count_bytes(),
             _name: name,
             _path: path,
@@ -727,21 +730,22 @@ impl<'list, 'a> CursorMut<'list, 'a> {
         Some(unsafe { Pin::new_unchecked(curr) })
     }
 
-    /// Override the [`RLookupKey`] currently being pointed to by this cursor (if any) with the given flags.
+    /// Override the [`RLookupKey`] at this cursor position and extend it with the given flags.
     ///
-    /// The new key will inherit the `name`, `path`, and `dstidx` of the current key but receive a
-    /// **new pointer identity**. The new key is returned.
+    /// The new key will inherit the `name`, `path`, and `dstidx`, and the `flags` of the key at the current position, but
+    /// receive a **new pointer identity**. The *new key* is returned.
+    ///
+    /// The old key remains as a hidden tombstone in the linked list.
     #[cfg_attr(not(test), expect(unused, reason = "used by later stacked PRs"))]
     pub fn override_current(
         mut self,
         flags: RLookupKeyFlags,
     ) -> Option<Pin<&'list mut RLookupKey<'a>>> {
         let mut old = self.current()?;
+        println!("overriding key {old:?}");
 
         let new = {
             let mut old = old.as_mut().project();
-
-            *old.flags |= RLookupKeyFlag::Hidden;
 
             *old.name = ptr::null();
             *old.name_len = usize::MAX;
@@ -750,7 +754,12 @@ impl<'list, 'a> CursorMut<'list, 'a> {
             *old.path = ptr::null();
             let path = mem::take(old._path.deref_mut());
 
-            RLookupKey::from_parts(name, path, *old.dstidx, flags)
+            let new = RLookupKey::from_parts(name, path, *old.dstidx, *old.flags | flags);
+
+            // Mark the old key as hidden, so it won't show up in iteration.
+            *old.flags |= RLookupKeyFlag::Hidden;
+
+            new
         };
 
         // Safety: we treat the pointer as pinned below and only hand out a pinned mutable reference.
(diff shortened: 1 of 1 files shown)


src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -222,6 +223,38 @@ impl<'a> RLookupKey<'a> {
...
     }
 
+    pub fn from_parts(
+        name: Cow<'a, CStr>,
+        path: Option<Cow<'a, CStr>>,
+        dstidx: u16,
+        flags: RLookupKeyFlags,
+    ) -> Self {
+        debug_assert_eq!(
+            matches!(name, Cow::Owned(_)),
```
DarthB, line 287:
```suggestion
            matches!(name, Cow::Borrowed(_)),
```

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -222,6 +223,38 @@ impl<'a> RLookupKey<'a> {
...
+            flags.contains(RLookupKeyFlag::NameAlloc),
+            "RLookupKeyFlag::NameAlloc but name is Cow::Borrowed"
+        );
+        if let Some(path) = &path {
+            debug_assert_eq!(
+                matches!(path, Cow::Owned(_)),
```
DarthB, line 293:
```suggestion
                matches!(path, Cow::Borrowed(_)),
```

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -934,4 +1037,44 @@ mod tests {
...
+        // new key should have provided keys
+        assert!(found.flags.contains(RLookupKeyFlag::Numeric));
+        // new key should not inherit any old flags
+        assert!(!found.flags.contains(RLookupKeyFlag::Unresolved));
+
+        let mut c = keylist.cursor_front();
+
+        // we expect the first item to be the tombstone of the old key
+        assert!(c.current().unwrap()._name.is_empty());
+        assert!(c.current().unwrap().flags.contains(RLookupKeyFlag::Hidden));
```
DarthB:
```suggestion
        // we expect the first item to be the tombstone of the old key
        assert!(c.current().unwrap().is_tombstone());
```

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -934,4 +1037,44 @@ mod tests {
 
         assert!(keylist.find_by_name(c"baz").is_none());
     }
+
+    #[test]
+    fn keylist_override_key() {
```
DarthB: Seems like a test for find and after push too. Proposal: Just use the Cursor interface here and add another test for the part with `found`.

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -624,6 +672,61 @@ impl<'a> CursorMut<'_, 'a> {
         Some(unsafe { Pin::new_unchecked(curr) })
     }
 
+    /// Override the [`RLookupKey`] currently being pointed to by this cursor (if any) with the given flags.
+    ///
+    /// The new key will inherit the `name`, `path`, and `dstidx` of the current key but receive a
+    /// **new pointer identity**. The new key is returned.
```
DarthB:
```suggestion
    /// Override the [`RLookupKey`] at this cursor position and extend it with the flags given as an argument.
    ///
    /// The new key will inherit the `name`, `path`, and `dstidx`, and the flags of the key at the current position, but 
    /// receive a **new pointer identity**. The new key is returned.
    ///
    The old key remains as a tombstone in the linked list and is now only used for traversal.
```

JonasKruckenberg: changed the comment somewhat, because your suggestion was actually not quite accurate (the flags are not extended, they are overridden). But otherwise addressed.

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: addressed 👍🏻

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: The assertion is correct as-is, `from_parts` expects the caller to do the "maybe allocation", but as a sanity check we make sure that the given flags and the given `Cow` variant match.

I improved the panic message though, hopefully its clearer now.

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: see above

JonasKruckenberg:
> Nice work. I have to main points:
> 
> 1. I think you missed this line in C:
> 
> ```
> new->flags |= old->flags & RLOOKUP_F_NAMEALLOC;
> ```


The handling of the NAMEALLOC flag is actually the one thing that #6431 changes. The NAMEALLOC flag is now treated as *transient* and never stored in the key (one exists as an argument to the methods). So we don't actually need the masking.
As to the flag inheritance: Yup realized that was missing too now

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: Update: nvm, I just forgot you are right

src/redisearch_rs/rlookup/src/lookup.rs

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -624,6 +673,68 @@ impl<'a> CursorMut<'_, 'a> {
...
+    pub fn override_current(
+        mut self,
+        flags: RLookupKeyFlags,
+    ) -> Option<Pin<&'list mut RLookupKey<'a>>> {
+        let mut old = self.current()?;
+
+        let new = {
+            let mut old = old.as_mut().project();
+
+            *old.flags |= RLookupKeyFlag::Hidden;
```
DarthB: remove that line it comes later

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -628,6 +677,66 @@ impl<'a> CursorMut<'_, 'a> {
...
 
+    /// Override the [`RLookupKey`] at this cursor position and extend it with the given flags.
+    ///
+    /// The new key will inherit the `name`, `path`, and `dstidx`, and the `flags` of the key at the current position, but
+    /// receive a **new pointer identity**. The new key is returned.
+    ///
+    /// The old key remains as a hidden tombstone in the linked list.
+    #[cfg_attr(not(test), expect(unused, reason = "used by later stacked PRs"))]
+    pub fn override_current(
+        mut self,
```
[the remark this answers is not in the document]
JonasKruckenberg: `override_current` returns a `&'list mut RLookupKey` where `'list` is *not* the lifetime of the cursor, but the lifetime of the upstream `KeyList`. This is required for later methods to use the return value more easily.

This means though, that `override_current` does *not* borrow the cursor, which would allow consumers to obtain multiple mutable references to the same key. For this reason `override_current` must be consumed to prevent this (alternatively `override_current` could return `'_` borrows instead of `'list` which would tie up the cursor for as long as the borrow lives, but that doesn't work because of the above mentioned requirements)

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -224,6 +225,39 @@ impl<'a> RLookupKey<'a> {
...
+                "`RLookupKeyFlag::NameAlloc` was provided, but `path` was not `Cow::Owned`"
+            );
+        }
+
+        Self {
+            dstidx,
+            svidx: 0,
+            flags: flags & !TRANSIENT_FLAGS,
+            name: name.as_ptr(),
+            path: name.as_ptr(),
```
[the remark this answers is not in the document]

DarthB (commented):
LGTM

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -679,6 +730,67 @@ impl<'a> CursorMut<'_, 'a> {
...
+    /// receive a **new pointer identity**. The *new key* is returned.
+    ///
+    /// The old key remains as a hidden tombstone in the linked list.
+    #[cfg_attr(not(test), expect(unused, reason = "used by later stacked PRs"))]
+    pub fn override_current(
+        mut self,
+        flags: RLookupKeyFlags,
+    ) -> Option<Pin<&'list mut RLookupKey<'a>>> {
+        let mut old = self.current()?;
+        println!("overriding key {old:?}");
```
Itzikvaknin: Leftover?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -1158,6 +1356,58 @@ mod tests {
...
+            make_bitflags!(RLookupKeyFlag::Unresolved),
+        ));
+        keylist
+            .cursor_front_mut()
+            .override_current(make_bitflags!(RLookupKeyFlag::Numeric));
+
+        let mut c = keylist.cursor_front();
+
+        // we expect the first item to be the tombstone of the old key
+        assert!(c.current().unwrap().is_tombstone());
```
Itzikvaknin, line 1403: Nice :thumbsup:

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -1158,6 +1356,58 @@ mod tests {
...
+
+        // and the next item to be the new key
+        c.move_next();
+        assert_eq!(c.current().unwrap()._name.as_ref(), c"foo");
+    }
```
Itzikvaknin, line 1408: Please add a test for the tail logic.
It has specific handling, so better to keep it separate from the general case

Itzikvaknin (dismissed):
Look good :muscle:

src/redisearch_rs/rlookup/src/lookup.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6470/commits/e086dfa0ea9f689d72f803f75cf05d0d196963c3

src/redisearch_rs/rlookup/src/lookup.rs
DarthB: Great idea! https://github.com/RediSearch/RediSearch/pull/6470/commits/e086dfa0ea9f689d72f803f75cf05d0d196963c3

DarthB (approved):

Itzikvaknin (approved):

PR #6481, opened 2025-07-09
[MOD-9680] - Adapt OnReopen mechanism to new LEAF iterators

JoanFM:
Want to start adapting the OnReopen mechanism to the new iterators.

This is the first step of porting the OnReopen mechanism of the leaf Iterators (Checking Tag, Term and Numeric index) adding as follows:

- `Revalidate` and `CheckAbort` API to iterator: This is another API similar to `Read`, `SkipTo`and so on. This will be a dummy OK call for most of leaf iterators except for InvIndexIterators.

- InvIndexIterators expansion to have specialized types depending on the index structure underlyining, to have custom `CheckAbort` specialization.

`Revalidate`can return:
- OK (Nothing happened everything good)
- ABORTED (Abort the complete chain, the Index was deleted)
- MOVED (The lastDocId of this specific iterator is invalid and the iterator has advanced to another one).

## Left out

It is left out of the PR the work of the Non-leaf iterators: NOTITERATOR, OPTIONALITERATOR, UNIONITERATOR and INTERSECTIONITERATOR. 

These iterators would need to act differently depending on the ouptut of their children `Revalidate`output


src/result_processor.c
```diff
@@ -104,6 +104,7 @@ static int rpidxNext(ResultProcessor *base, SearchResult *res) {
     // If we need to read the iterators and we didn't lock the spec yet, lock it now
     // and reopen the keys in the concurrent search context (iterators' validation)
     RedisSearchCtx_LockSpecRead(RP_SCTX(base));
+    //it->ReOpen(it->ctx);
```
JoanFM, line 107: or it->Revalidate when changed

src/iterators/inverted_index_iterator.c
```diff
@@ -60,6 +61,118 @@ size_t InvIndIterator_NumEstimated(QueryIterator *base) {
...
+    return VALIDATE_OK;
+  }
+
+  // For numeric fields, we need to get the field name from the filter context
+  // We need to use the field spec directly from the numeric filter
+  const NumericFilter *filter = it->decoderCtx.filter;
+  if (!filter) {
+    // No filter means we can't check for revision changes
+    return VALIDATE_OK;
+  }
```
GuyAv46: If that's the case, we'll need to take it from another place (maybe take a pointer to the field spec). Only the iterators over the edges of the numeric range have a filter, but we intentionally set it to NULL if we know all the results in the range are going to match

src/iterators/inverted_index_iterator.c
```diff
@@ -60,6 +61,118 @@ size_t InvIndIterator_NumEstimated(QueryIterator *base) {
...
+}
+
+static ValidateStatus InvIndIterator_Revalidate(QueryIterator *base) {
+  // Here we should apply the specifics of Term, Tag and Numeric
+
+  InvIndIterator *it = (InvIndIterator *)base;
+  if (base->atEOF) {
+    // Save time and state if we are already at the end
+    return VALIDATE_OK;
+  }
```
GuyAv46: Probably unsafe (we may rewind, and if we didn't validate, we may access invalid pointers. Let's avoid this pretty insignificant optimization

src/iterators/inverted_index_iterator.c
```diff
@@ -60,6 +61,118 @@ size_t InvIndIterator_NumEstimated(QueryIterator *base) {
...
+    base->Rewind(base);
+    IteratorStatus rc = base->SkipTo(base, lastDocId);
+    if (rc == ITERATOR_NOTFOUND) {
+      ret = VALIDATE_MOVED;
+    } else if (rc == ITERATOR_EOF) {
+      ret = VALIDATE_OK;
+    }
+  }
+
+  return ret;
```
GuyAv46: A bit hard to follow. Consider either assuming `ret == VALIDATE_OK` and only change it if we need to (the `rc == ITERATOR_NOTFOUND` case), or have a clearer default
```suggestion
  if (ret != VALIDATE_OK) {
    return ret;
  }

  // the gc marker tells us if there is a chance the keys have undergone GC while we were asleep
  if (it->gcMarker == it->idx->gcMarker) {
    // no GC - we just go to the same offset we were at
    size_t offset = it->blockReader.buffReader.pos;
    SetCurrentBlockReader(it);
    it->blockReader.buffReader.pos = offset;
  } else {
    // if there has been a GC cycle on this key while we were asleep, the offset might not be valid
    // anymore. This means that we need to seek the last docId we were at

    // keep the last docId we were at
    t_docId lastDocId = base->lastDocId;
    // reset the state of the reader
    base->Rewind(base);
    IteratorStatus rc = base->SkipTo(base, lastDocId);
    if (rc == ITERATOR_NOTFOUND) {
      ret = VALIDATE_MOVED;
    }
  }

  return ret;
```

src/iterators/inverted_index_iterator.c
```diff
@@ -60,6 +61,118 @@ size_t InvIndIterator_NumEstimated(QueryIterator *base) {
...
+}
+
+static ValidateStatus TermCheckAbort(QueryIterator *base) {
+  InvIndIterator *it = (InvIndIterator *)base;
+  if (!it->sctx) {
+    return VALIDATE_OK;
+  }
+  InvertedIndex *idx = Redis_OpenInvertedIndex(it->sctx, base->current->data.term.term->str,
+      base->current->data.term.term->len, false, NULL);
+  if (!idx || idx == TRIEMAP_NOTFOUND || it->idx != idx) {
```
GuyAv46: CAnnot return `TRIEMAP_NOTFOUND`
```suggestion
  if (!idx || it->idx != idx) {
```

src/iterators/inverted_index_iterator.c
```diff
@@ -60,6 +61,118 @@ size_t InvIndIterator_NumEstimated(QueryIterator *base) {
...
+
+static ValidateStatus TagCheckAbort(QueryIterator *base) {
+  TagInvIndIterator *it = (TagInvIndIterator *)base;
+  if (!it->base.sctx) {
+    return VALIDATE_OK;
+  }
+  size_t sz;
+  InvertedIndex *idx = TagIndex_OpenIndex(it->tagIdx, base->current->data.term.term->str,
+      base->current->data.term.term->len, false, &sz);
+  if (!idx || idx == TRIEMAP_NOTFOUND || it->base.idx != idx) {
```
GuyAv46: Cannot return `NULL`
```suggestion
  if (idx == TRIEMAP_NOTFOUND || it->base.idx != idx) {
```

tests/cpptests/test_cpp_iterator_index.cpp
```diff
@@ -6,65 +6,64 @@
...
-    TYPE_NUMERIC,
-    TYPE_GENERIC,
-} IndexIteratorType;
+typedef enum IndexType {
+    INDEX_TYPE_TERM_FULL,
+    INDEX_TYPE_NUMERIC_FULL,
+    INDEX_TYPE_TERM,
+    INDEX_TYPE_NUMERIC,
+    INDEX_TYPE_GENERIC,
+} IndexType;
```
GuyAv46: why?

tests/cpptests/test_cpp_iterator_index.cpp
```diff
@@ -6,65 +6,64 @@
...
 
-class IndexIteratorTest : public ::testing::TestWithParam<std::tuple<IndexIteratorType, bool>> {
+class IndexIteratorTest : public ::testing::TestWithParam<IndexType> {
```
GuyAv46: Seems like a rebase was deleting all the TTL variant tests. Need to revert

tests/cpptests/test_cpp_iterator_index.cpp
JoanFM: merge issue it seems

tests/cpptests/test_cpp_iterator_index.cpp
JoanFM: added back

src/iterators/inverted_index_iterator.c
JoanFM: what do I need from another place? the FieldSpec?

GuyAv46 (approved):

PR #6469, opened 2025-07-08
RLookupRow - Refactor C-Code - MOD-10394

DarthB:
## Describe the changes in the pull request

Removed not used C Code and replaced direct field accesses with inline functions for `RLookupRow`.

First of all, stacked PRs for `RLookupRow`

Follow Up PRs: #6472 #6476 #6484

PR #6484, opened 2025-07-10
RLookupRow - implementation of `write_(own_)key` functions - MOD-10390

DarthB:

Implementation of the `write_own_key` and `write_key` methods for the RLookupRow.

Stacks on top of #6469 #6472 #6476


DarthB:
A noop merge probably caused by the same base

PR #6490, opened 2025-07-14
[MOD-10444] Rust code for `RSAggregateResult` management

chesedo:
This adds the Rust code to move the `RSAggregateResult` internal management to Rust. The switch to Rust does not happen in this PR and will happen in a separate PR instead.

#### Mark if applicable


src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -100,6 +114,109 @@ pub struct RSAggregateResult {
     pub type_mask: RSResultTypeMask,
 }
 
+impl RSAggregateResult {
+    /// Create a new aggregate result with th given capacity
```
```suggestion
    /// Create a new aggregate result with the given capacity
```

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -100,6 +114,109 @@ pub struct RSAggregateResult {
...
+            }
+
+            ptr as *mut *mut RSIndexResult
+        } else {
+            ptr::null_mut()
+        }
+    }
+
+    fn free_children(&mut self) {
+        if !self.children.is_null() && self.children_cap > 0 {
```
gdesmott: `children != NULL <-> children_cap > 0` right? Should we enforce that with a (debug) assert?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -367,3 +511,6 @@ pub trait Decoder {
         }
     }
 }
+
+#[cfg(test)]
+mod tests;
```
gdesmott: Any reason to not have those in the existing `/tests/` dir instead?

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -0,0 +1,140 @@
...
+#[unsafe(no_mangle)]
+pub extern "C" fn IndexResult_ConcatMetrics(
+    _parent: *mut RSIndexResult,
+    _child: *mut RSIndexResult,
+) {
+    // Do nothing since the code will call this
+}
+
+impl RSAggregateResult {
+    fn get(&self, index: usize) -> Option<&RSIndexResult> {
```
gdesmott: Why not have this as proper public API?

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -0,0 +1,140 @@
...
+        if index > self.num_children as _ {
+            return None;
+        }
+
+        let child_ptr = unsafe { *self.children.add(index) };
+
+        if child_ptr.is_null() {
```
gdesmott: `ptr::add()` returns `NonNull`

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -0,0 +1,140 @@
...
+
+            agg.get(index)
+        } else {
+            None
+        }
+    }
+}
+
+#[test]
+fn pushing_to_aggregate_result() {
```
gdesmott: Did you check with `valgrind` if we don't leak memory in those tests?
I find it useful when writing code manually allocating memory.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -100,6 +114,109 @@ pub struct RSAggregateResult {
...
+    pub fn new(cap: usize) -> Self {
+        Self {
+            num_children: 0,
+            children_cap: cap as c_int,
+            children: Self::new_children(cap),
+            type_mask: RSResultTypeMask::empty(),
+        }
+    }
+
+    fn push(&mut self, result: &mut RSIndexResult) {
```
gdesmott: Do you actually need  `&mut RsIndexResult` here?

You can cast `&RsIndexResult` to a `*mut` using `result as *const _ as *mut _;`. 
I know it's kind of hacky but afaik it's a common pattern which will save us from relying on `mut` in the Rust API.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -100,6 +114,109 @@ pub struct RSAggregateResult {
...
+        if self.num_children >= self.children_cap {
+            self.grow()
+        }
+
+        // SAFETY: `num_children` is within the valid range of the `children` pointer, else we would
+        // have grown above to make it valid.
+        let slot = unsafe { self.children.add(self.num_children as _) };
+
+        // SAFETY: we just used `add` to ensure `slot` is a valid pointer and is "aligned correctly"
```
gdesmott, line 142: This actually assumes that `result` stays allocated during the whole `RSAggregateResult` lifetime, right?

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -0,0 +1,140 @@
...
+            None
+        } else {
+            let child = unsafe { &*child_ptr };
+            Some(child)
+        }
+    }
+}
+
+impl RSIndexResult {
+    fn union(doc_id: t_docId, cap: usize, weight: f64) -> Self {
```
gdesmott: Why have this in the tests? We already have such similar constructors.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Yes, it is always an IFF with the current code. Do you mean adding a debug assertion just above this line and then just checking one of them?

gdesmott: yeah

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: I initially thought there was something private the tests needed. But see that is not the case.

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: I might be missing something. How do you see `ptr::add()` returns `NonNull`?

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Yes, the C code has the same assumption so I kept it

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: I was going to move it in one of the later PRs to keep the public interface as small as possible

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6490/commits/e9640e7f0493fe063f61d386859f249d7d177902

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: I just wanted to keep the public API as small as possible. But moved it here to have the tests in `/tests/` -> https://github.com/RediSearch/RediSearch/pull/6490/commits/e9640e7f0493fe063f61d386859f249d7d177902

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Nope, `mut` is not needed. Nice, trick!! Done in -> https://github.com/RediSearch/RediSearch/pull/6490/commits/788d99821c04dc989e1c346305f6687b0fb4bf7e

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: It got moved in https://github.com/RediSearch/RediSearch/pull/6490/commits/e9640e7f0493fe063f61d386859f249d7d177902

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6490/commits/c6140c6fff4a742e0cd759f9a1405c18af8c71ed

src/redisearch_rs/inverted_index/src/tests.rs
gdesmott: oh, I was looking at the `NonNull<T>` doc :sweat_smile:

src/redisearch_rs/inverted_index/src/lib.rs
gdesmott: Not great for safety but that will go away as we keep porting code to Rust.

Maybe make it clearer in the function doc as I'm sure we'll miss that at some point when using this. :)

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: Ahhh, I see. I'm purposefully not cleaning up the children pointers because that only has to happen if the type was copied (deep cloned). However, I should be able to have the tests clean up the memory correctly.

I'll add an update for this

chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6490/commits/7c5c48d81ca0acdb075d93bcdeaa59ef7bbc7721

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6490/commits/bbd4f46eb042920cd12e62e608e78ad5fbfa9458

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -100,6 +114,144 @@ pub struct RSAggregateResult {
...
+        };
+
+        self.resize(new_cap)
+    }
+
+    fn resize(&mut self, new_cap: c_int) {
+        let new_children = Self::new_children(new_cap as _);
+
+        // Copy existing items to new array
+        if !self.children.is_null() && self.num_children > 0 {
```
gdesmott: As in `get` you can check only `num_children` et assert the ptr is not NULL.

chesedo: Perfect, done in https://github.com/RediSearch/RediSearch/pull/6490/commits/ec4c5df0ecae3be1f276f6bda71b434abc20af9f

gdesmott (dismissed):

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -8,18 +8,32 @@
...
 pub mod numeric;
 
+// Manually define some C functions, because we'll create a circular dependency if we use the FFI
+// crate to make them automatically.
+unsafe extern "C" {
+    /// Adds the metrics of a child [`RSIndexResult`] to the parent [`RSIndexResult`].
+    ///
+    /// # Safety
+    /// Both should be valid `RSIndexResult` instances.
+    #[allow(improper_ctypes)]
```
LukeMathWalker, line 33: What improper types are we passing around?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -100,6 +114,149 @@ pub struct RSAggregateResult {
     pub type_mask: RSResultTypeMask,
 }
 
+impl RSAggregateResult {
```
LukeMathWalker, line 117: We are casting `cap` to `usize` left and right. I think it's about time we changed it to be a `usize` in the struct definition.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -100,6 +114,149 @@ pub struct RSAggregateResult {
...
+            unsafe {
+                ptr::copy_nonoverlapping(self.children, new_children, self.num_children as _);
+            }
+        }
+
+        // Deallocate the old array
+        self.free_children();
+
+        self.children = new_children;
+        self.children_cap = new_cap;
```
LukeMathWalker, line 211: Shouldn't we use realloc here?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -219,6 +407,37 @@ impl RSIndexResult {
...
+    fn is_aggregate(&self) -> bool {
+        matches!(
+            self.result_type,
+            RSResultType::Intersection | RSResultType::Union | RSResultType::HybridMetric
+        )
+    }
+
+    /// Adds a result if this is an aggregate type. Else nothing happens to the added result.
+    ///
+    /// # Undefined behaviour
```
LukeMathWalker, line 421:
```suggestion
    /// # Safety
```

chesedo:
This is replaced with #6492 and #6494

PR #6492, opened 2025-07-15
[MOD-10485] Guard access to `RSAggregateResult` internals

chesedo:
We need to change the internal layout of this struct to be able to have it allocated on the Rust memory side. This change causes all the internals to be accessed via some method which will hide this change once it happens.

There are only a few direct accesses left inside `index_result.*` which will be moved with the change that will move the allocation to the Rust side.

#### Mark if applicable

### DIFF
--- a/src/hybrid_reader.c
+++ b/src/hybrid_reader.c
@@ -94,7 +94,7 @@ static void insertResultToHeap_Aggregate(HybridIterator *hr, RSIndexResult *res,
   AggregateResult_AddChild(res, vec_res);
   AggregateResult_AddChild(res, child_res);
   RSIndexResult *hit = IndexResult_DeepCopy(res);
-  AggregateResult_Reset(res); // Reset the current result.
+  IndexResult_ResetAggregate(res); // Reset the current result.
   ResultMetrics_Add(hit, hr->base.ownKey, RS_NumVal(vec_res->data.num.value));
 
   if (hr->topResults->count < hr->query.k) {
--- a/src/hybrid_reader.c
+++ b/src/hybrid_reader.c
@@ -11,7 +11,7 @@
 #include "VecSim/vec_sim.h"
 #include "VecSim/query_results.h"
 
-#define VECTOR_RESULT(p) (p->type == RSResultType_Metric ? p : p->data.agg.children[0])
+#define VECTOR_RESULT(p) (p->type == RSResultType_Metric ? p : AggregateResult_Get(&p->data.agg, 0))
 
 static VecSimQueryReply_Code prepareResults(HybridIterator *hr); // forward declaration
 
@@ -104,7 +104,8 @@ static void insertResultToHeap_Aggregate(HybridIterator *hr, RSIndexResult *res,
   }
   // Set new upper bound.
   RSIndexResult *worst = mmh_peek_max(hr->topResults);
-  *upper_bound = worst->data.agg.children[0]->data.num.value;
+  const RSIndexResult *first = AggregateResult_Get(&worst->data.agg, 0);
+  *upper_bound = first->data.num.value;
 }
 
 static void insertResultToHeap(HybridIterator *hr, RSIndexResult *res, RSIndexResult *child_res,
--- a/src/index_result.c
+++ b/src/index_result.c
@@ -134,10 +134,19 @@ RSIndexResult *IndexResult_DeepCopy(const RSIndexResult *src) {
 
       ret->data.agg.children = rm_malloc(numChildren * sizeof(RSIndexResult *));
       ret->data.agg.childrenCap = numChildren;
+
       // deep copy recursively all children
-      for (int i = 0; i < numChildren; i++) {
-        ret->data.agg.children[i] = IndexResult_DeepCopy(src->data.agg.children[i]);
+      int i = 0;
+      RSAggregateResultIter *iter = AggregateResult_Iter(&src->data.agg);
+      RSIndexResult *child = NULL;
+
+      while (AggregateResultIter_Next(iter, &child)) {
+        ret->data.agg.children[i] = IndexResult_DeepCopy(child);
+        i++;
       }
+
+      AggregateResultIter_Free(iter);
+
       break;
 
     // copy term results
@@ -209,12 +218,15 @@ void IndexResult_Free(RSIndexResult *r) {
   ResultMetrics_Free(r);
   if (r->type == RSResultType_Intersection || r->type == RSResultType_Union || r->type == RSResultType_HybridMetric) {
     // for deep-copy results we also free the children
-    if (r->isCopy && r->data.agg.children) {
-      int numChildren = AggregateResult_NumChildren(&r->data.agg);
+    if (r->isCopy) {
+      RSAggregateResultIter *iter = AggregateResult_Iter(&r->data.agg);
+      RSIndexResult *child = NULL;
 
-      for (int i = 0; i < numChildren; i++) {
-        IndexResult_Free(r->data.agg.children[i]);
+      while (AggregateResultIter_Next(iter, &child)) {
+        IndexResult_Free(child);
       }
+
+      AggregateResultIter_Free(iter);
     }
     rm_free(r->data.agg.children);
     r->data.agg.children = NULL;
@@ -303,12 +315,15 @@ void result_GetMatchedTerms(RSIndexResult *r, RSQueryTerm *arr[], size_t cap, si
   switch (r->type) {
     case RSResultType_Intersection:
     case RSResultType_Union:
+      RSAggregateResultIter *iter = AggregateResult_Iter(&r->data.agg);
+      RSIndexResult *child = NULL;
 
-      int numChildren = AggregateResult_NumChildren(&r->data.agg);
-
-      for (int i = 0; i < numChildren; i++) {
-        result_GetMatchedTerms(r->data.agg.children[i], arr, cap, len);
+      while (AggregateResultIter_Next(iter, &child)) {
+        result_GetMatchedTerms(child, arr, cap, len);
       }
+
+      AggregateResultIter_Free(iter);
+
       break;
     case RSResultType_Term:
       if (r->data.term.term) {
@@ -454,14 +469,20 @@ int IndexResult_IsWithinRange(RSIndexResult *ir, int maxSlop, int inOrder) {
   RSOffsetIterator iters[num];
   uint32_t positions[num];
   int n = 0;
-  for (int i = 0; i < num; i++) {
+
+  RSAggregateResultIter *iter = AggregateResult_Iter(r);
+  RSIndexResult *child = NULL;
+
+  while (AggregateResultIter_Next(iter, &child)) {
     // collect only iterators for nodes that can have offsets
-    if (RSIndexResult_HasOffsets(r->children[i])) {
-      iters[n] = RSIndexResult_IterateOffsets(r->children[i]);
+    if (RSIndexResult_HasOffsets(child)) {
+      iters[n] = RSIndexResult_IterateOffsets(child);
       positions[n] = 0;
       n++;
     }
   }
+
+  AggregateResultIter_Free(iter);
 
   // No applicable offset children - just return 1
   if (n == 0) {
--- a/src/index_result.c
+++ b/src/index_result.c
@@ -182,18 +182,6 @@ void Term_Free(RSQueryTerm *t) {
   }
 }
 
-void IndexResult_Init(RSIndexResult *h) {
-
-  h->docId = 0;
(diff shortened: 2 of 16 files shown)


src/redisearch_rs/c_entrypoint/types_ffi/cbindgen.toml
```diff
@@ -17,6 +17,3 @@ typedef __uint128_t t_fieldMask;
 [parse]
 parse_deps = true
 include = ["enumflags2", "inverted_index"]
-
-[export]
-include = ["RSIndexResult"]
```
gdesmott, line 22: What's the rational for this change? I did not look closely at cbindgen yet.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -100,6 +100,84 @@ pub struct RSAggregateResult {
...
+    }
+
+    /// Reset the aggregate result, clearing all children and resetting the type mask
+    pub fn reset(&mut self) {
+        self.num_children = 0;
+        self.type_mask = RSResultTypeMask::empty();
+    }
+}
+
+pub struct RSAggregateResultIter<'a> {
```
gdesmott, line 163: I'd document this if only to have full doc coverage.

src/redisearch_rs/c_entrypoint/types_ffi/cbindgen.toml
chesedo: This was a temp solution to get cbindgen to export the `RSIndexResult` since there were no functions in the `lib.rs` that used it.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Thin but simple https://github.com/RediSearch/RediSearch/pull/6492/commits/985fbb4f5f28f2fd1e4a0f4566fd119d80e44add

gdesmott (dismissed):

src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
```diff
@@ -9,4 +9,168 @@
...
+/// # Safety
+///
+/// The following invariants must be upheld when calling this function:
+/// - `iter` must point to a valid `RSAggregateResultIter` and cannot be NULL.
+/// - The iterator must have been created using [`AggregateResult_Iter`].
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn AggregateResultIter_Free(iter: *mut RSAggregateResultIter<'static>) {
+    debug_assert!(!iter.is_null(), "iter must not be null");
+
+    // SAFETY: Caller is to ensure `iter` is non-null and was allocated using `AggregateRusult_Iter`
```
gdesmott:
```suggestion
    // SAFETY: Caller is to ensure `iter` is non-null and was allocated using `AggregateResult_Iter`
```

src/index_result.c
```diff
@@ -129,12 +130,23 @@ RSIndexResult *IndexResult_DeepCopy(const RSIndexResult *src) {
...
+
       // deep copy recursively all children
-      for (int i = 0; i < src->data.agg.numChildren; i++) {
-        ret->data.agg.children[i] = IndexResult_DeepCopy(src->data.agg.children[i]);
+      int i = 0;
+      RSAggregateResultIter *iter = AggregateResult_Iter(&src->data.agg);
+      RSIndexResult *child = NULL;
+
+      while (AggregateResultIter_Next(iter, &child)) {
+        ret->data.agg.children[i] = IndexResult_DeepCopy(child);
```
meiravgri, line 144: Should we implement setter methods as well?

for example:
AggregateResult_SetChildren(agg, children_array)
AggregateResult_SetCapacity(agg, capacity)
AggregateResult_SetChild(agg, index, child)

src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
```diff
@@ -9,4 +9,168 @@
 
 //! This module contains pure Rust types that we want to expose to C code.
 
-pub use inverted_index::RSIndexResult;
+use inverted_index::{RSAggregateResult, RSAggregateResultIter, RSIndexResult};
+
+/// Get the result at the specified index in the aggregate result. This will return a `NULL` pointer
+/// if the index is out of bounds.
```
meiravgri, line 15: Are all calling functions checking for null before dereferencing?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -88,16 +88,95 @@ pub type RSResultTypeMask = BitFlags<RSResultType, u32>;
...
+        } else {
+            None
+        }
+    }
+
+    /// Reset the aggregate result, clearing all children and resetting the type mask
+    pub fn reset(&mut self) {
+        self.num_children = 0;
+        self.type_mask = RSResultTypeMask::empty();
+    }
```
meiravgri: Does `reset()` need to free the individual child results pointed to by the `children` array, or are they managed elsewhere?

src/hybrid_reader.c
```diff
@@ -104,7 +104,8 @@ static void insertResultToHeap_Aggregate(HybridIterator *hr, RSIndexResult *res,
   }
   // Set new upper bound.
   RSIndexResult *worst = mmh_peek_max(hr->topResults);
-  *upper_bound = worst->data.agg.children[0]->data.num.value;
+  const RSIndexResult *first = AggregateResult_Get(&worst->data.agg, 0);
+  *upper_bound = first->data.num.value;
```
meiravgri, line 108: dont we have getters for RSIndexResult ?

src/offset_vector.c
```diff
@@ -159,8 +169,10 @@ RSOffsetIterator RSIndexResult_IterateOffsets(const RSIndexResult *res) {
     case RSResultType_Union:
     default:
       // if we only have one sub result, just iterate that...
-      if (res->data.agg.numChildren == 1) {
-        return RSIndexResult_IterateOffsets(res->data.agg.children[0]);
+      int numChildren = AggregateResult_NumChildren(&res->data.agg);
```
meiravgri: why int?

src/offset_vector.c
```diff
@@ -112,21 +112,31 @@ static RSOffsetIterator _aggregateResult_iterate(const RSAggregateResult *agg) {
   _RSAggregateOffsetIterator *it = mempool_get(pool);
   it->res = agg;
 
-  if (agg->numChildren > it->size) {
-    it->size = agg->numChildren;
+  int numChildren = AggregateResult_NumChildren(agg);
```
meiravgri: why int

src/offset_vector.c
```diff
@@ -197,7 +209,7 @@ uint32_t _aoi_Next(void *ctx, RSQueryTerm **t) {
   int minIdx = -1;
   uint32_t minVal = RS_OFFSETVECTOR_EOF;
   uint32_t *offsets = it->offsets;
-  register int num = it->res->numChildren;
+  register int num = AggregateResult_NumChildren(it->res);
```
meiravgri: int?

src/optimizer_reader.c
```diff
@@ -167,7 +167,7 @@ int OPT_Read(void *ctx, RSIndexResult **e) {
         if (numericRes->type == RSResultType_Numeric) {
           *it->pooledResult = *numericRes;
         } else {
-          RSIndexResult *child = numericRes->data.agg.children[0];
+          const RSIndexResult *child = AggregateResult_Get(&numericRes->data.agg, 0);
           RS_LOG_ASSERT(child->type == RSResultType_Numeric, "???");
           *it->pooledResult = *(child);
```
meiravgri, line 172: is it safe to copy RSIndexResult content? doesnt it contain any pointers to allocated memory?

src/index_result.c
chesedo: We can, they will just be temporary as all the setters will be removed for Rust ones in the next few PRs. Here is an example of the PR coming directly after this one which will remove it. So I opted to rather keep this PR simpler.

https://github.com/RediSearch/RediSearch/pull/6494/files#diff-e1e6ef6053333d33179fc1c24305e5dac4b8619a16c6109b8c64828aa2ddd73bR140

src/hybrid_reader.c
chesedo: No, not yet. I'm already working on that in another PR [MOD-10328]

src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
chesedo: Lol, no none are checking for `null` first :sweat_smile: 

Here is an example of some of the calling functions

https://github.com/RediSearch/RediSearch/pull/6492/files#diff-e1e6ef6053333d33179fc1c24305e5dac4b8619a16c6109b8c64828aa2ddd73bR260-R276

There are nine of these in total. Some do have a bound check on the index value though. So they won't return null. I can add a guard to the others if they are not too complicated (they would have had out of bounds access before this PR anyway)

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: This is only called from here

https://github.com/RediSearch/RediSearch/pull/6492/files#diff-9244c87ebf28dd7d48554f498aea31ac56ba0117f26ceb666eed1fc155eaf83cR59

So, based on the old code, it should not free the children. I'm going to update the comment to make it clearer that this does not free the underlying children.

chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6492/commits/62f88f3475f4eaf4de82b638be899cd624264fbf

src/offset_vector.c
chesedo: Nice catch! I updated them all in https://github.com/RediSearch/RediSearch/pull/6492/commits/2ddc25ccdd5e2e95fa4e7131ac9c1a9e972788a3

src/optimizer_reader.c
chesedo: Hmm, I think the copy in the old code is wrong. @GuyAv46 line 172 should do a deep copy right?

GuyAv46: A shallow copy should be fine, but we lose the deep content that may be needed for the scorer.
But in this case, we assert the child type is `RSResultType_Numeric`, so there is no deep content, so it is safe

chesedo: Okay, thanks :pray:

chesedo: @GuyAv46 another question: why does the code just assume there will be a child at index `0`?

GuyAv46: It relies on the iterator's tree structure. With the Optimus iterator, it should be either only numeric or an aggregation of numeric and something else

chesedo: I see, thanks again!

src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
chesedo: Okay, these are the `NULL` I would have to check:

https://github.com/RediSearch/RediSearch/pull/6492/files#diff-d18e15d2cd21d003258f732c45fb62d4203794990d6b29902b44bb44d3e0695aR452
This one seems safe because (from my understanding) a hybrid metric will always have at least two items - @GuyAv46 please confirm I did not mess this up from a previous conversation :sweat_smile: 

https://github.com/RediSearch/RediSearch/pull/6492/files#diff-8e484f7004c4ee0dfd1910d4558da288b76efa154c74e251470ae1b73ca8c32fR14
and
https://github.com/RediSearch/RediSearch/pull/6492/files#diff-8e484f7004c4ee0dfd1910d4558da288b76efa154c74e251470ae1b73ca8c32fR107
These are again a hybrid so at least two items will exist.

https://github.com/RediSearch/RediSearch/pull/6492/files#diff-44d0ecccd515b7d79844f7a36a43ed8ef6089d03e24c84e9b51909ef1f18c873R170
Was already answered here https://github.com/RediSearch/RediSearch/pull/6492#discussion_r2225786737

If these are correct then no `NULL` checks are needed :crossed_fingers:

GuyAv46: Yes, this is another example where we know (assume) the result layout based on the iterator that owns and creates it. Should be safe

chesedo: Awesome, thank you!

meiravgri (approved):

PR #6493, opened 2025-07-16
[MOD-10379] Implement the encodeFreqsFields and  encodeFreqsFieldsWide encoder/decoder in Rust

gdesmott:

Implement the freqs fields encoder and decoder in Rust.

## Benchmarks results

Both Rust encoder and decoder are faster than the C version, in normald and wide mode.

```
Encode - FreqsFields/C    time:   [2.6159 µs 2.6176 µs 2.6194 µs]
Encode - FreqsFields/Rust time:   [1.9075 µs 1.9083 µs 1.9092 µs]

Decode - FreqsFields/C    time:   [10.721 ns 10.746 ns 10.774 ns]
Decode - FreqsFields/Rust time:   [8.9110 ns 8.9224 ns 8.9346 ns]

Encode - FreqsFields Wide/C    time:   [4.4013 µs 4.4071 µs 4.4137 µs]
Encode - FreqsFields Wide/Rust time:   [3.1365 µs 3.1388 µs 3.1412 µs]

Decode - FreqsFields Wide/C    time:   [46.058 ns 46.124 ns 46.195 ns]
Decode - FreqsFields Wide/Rust time:   [22.759 ns 22.810 ns 22.892 ns]
```

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/inverted_index/src/freqs_fields.rs
+++ b/src/redisearch_rs/inverted_index/src/freqs_fields.rs
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use std::io::{Read, Seek, Write};
+
+use ffi::{t_docId, t_fieldMask};
+use qint::{qint_decode, qint_encode};
+use varint::VarintEncode;
+
+use crate::{Decoder, DecoderResult, Encoder, RSIndexResult};
+
+/// Encode and decode the delta, frequency and field mask of a record.
+///
+/// This encoder supports field masks fitting in a `u32`.
+/// Use [`FreqsFieldsWide`] for `u128` field masks.
+///
+/// The delta, frequency, field mask are encoded using [qint encoding](qint).
+///
+/// This encoder only supports delta values that fit in a `u32`.
+#[derive(Default)]
+pub struct FreqsFields;
+
+impl Encoder for FreqsFields {
+    type Delta = u32;
+
+    fn encode<W: Write + Seek>(
+        &mut self,
+        mut writer: W,
+        delta: Self::Delta,
+        record: &RSIndexResult,
+    ) -> std::io::Result<usize> {
+        let field_mask = record
+                .field_mask
+                .try_into()
+                .expect("Need to use the wide variant of the FreqsFields encoder to support field masks bigger than u32");
+        let bytes_written = qint_encode(&mut writer, [delta, record.freq, field_mask])?;
+
+        Ok(bytes_written)
+    }
+}
+
+impl Decoder for FreqsFields {
+    fn decode<R: Read>(&self, reader: &mut R, base: t_docId) -> std::io::Result<DecoderResult> {
+        let (decoded_values, _bytes_consumed) = qint_decode::<3, _>(reader)?;
+        let [delta, freq, field_mask] = decoded_values;
+
+        let record = RSIndexResult::term()
+            .doc_id(base + delta as t_docId)
+            .field_mask(field_mask as t_fieldMask)
+            .frequency(freq);
+        Ok(DecoderResult::Record(record))
+    }
+}
+
+/// Encode and decode the delta, frequency and field mask of a record.
+///
+/// This encoder supports larger field masks fitting in a `u128`.
+/// Use [`FreqsFields`] for `u32` field masks.
+///
+/// The delta and frequency are encoded using [qint encoding](qint).
+/// The field mask is then encoded using [varint encoding](varint).
+///
+/// This encoder only supports delta values that fit in a `u32`.
+#[derive(Default)]
+pub struct FreqsFieldsWide;
+
+impl Encoder for FreqsFieldsWide {
+    type Delta = u32;
+
+    fn encode<W: Write + Seek>(
+        &mut self,
+        mut writer: W,
+        delta: Self::Delta,
+        record: &RSIndexResult,
+    ) -> std::io::Result<usize> {
+        let mut bytes_written = qint_encode(&mut writer, [delta, record.freq])?;
+        bytes_written += record.field_mask.write_as_varint(&mut writer)?;
+        Ok(bytes_written)
+    }
+}
+
+impl Decoder for FreqsFieldsWide {
+    fn decode<R: Read>(&self, reader: &mut R, base: t_docId) -> std::io::Result<DecoderResult> {
+        let (decoded_values, _bytes_consumed) = qint_decode::<2, _>(reader)?;
+        let [delta, freq] = decoded_values;
+        let field_mask = t_fieldMask::read_as_varint(reader)?;
+
+        let record = RSIndexResult::term()
+            .doc_id(base + delta as t_docId)
+            .field_mask(field_mask)
+            .frequency(freq);
+        Ok(DecoderResult::Record(record))
+    }
+}
--- a/src/redisearch_rs/inverted_index/tests/freqs_fields.rs
+++ b/src/redisearch_rs/inverted_index/tests/freqs_fields.rs
@@ -1,0 +1,204 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use std::io::Cursor;
+
+use ffi::t_fieldMask;
+use inverted_index::{
+    Decoder, DecoderResult, Encoder,
+    freqs_fields::{FreqsFields, FreqsFieldsWide},
+};
+
+#[test]
+fn test_encode_freqs_fields() {
+    // Test cases for the freqs fields encoder and decoder.
+    let tests = [
+        // (delta, frequency, field mask, expected encoding)
+        (0, 1, 1, vec![0, 0, 1, 1]),
+        (
+            10,
+            5,
+            u32::MAX as t_fieldMask,
+            vec![48, 10, 5, 255, 255, 255, 255],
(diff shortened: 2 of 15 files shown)


src/redisearch_rs/inverted_index/src/freqs_fields.rs
```diff
@@ -0,0 +1,87 @@
...
+impl Encoder for FreqsFields {
+    type Delta = u32;
+
+    fn encode<W: Write + Seek>(
+        &mut self,
+        mut writer: W,
+        delta: Self::Delta,
+        record: &RSIndexResult,
+    ) -> std::io::Result<usize> {
+        let bytes_written = if self.wide {
```
chesedo: I'm not a fan of doing a logic check like this at runtime. We lose a few CPU cycles each time we do the check but already knew which version we wanted when we constructed this type.

Maybe you can move the `bool` to a template parameter if you don't want to make a completely new struct for the wide variant?

src/redisearch_rs/inverted_index/tests/freqs_fields.rs
```diff
@@ -0,0 +1,181 @@
...
+
+use std::io::Cursor;
+
+use inverted_index::{Decoder, DecoderResult, Encoder, freqs_fields::FreqsFields};
+
+#[test]
+fn test_encode_freqs_fields() {
+    // Test cases for the freqs fields encoder and decoder.
+    let tests = [
+        // (delta, frequency, field mask, term offsets vector, expected encoding)
```
chesedo:
```suggestion
        // (delta, frequency, field mask, expected encoding)
```

src/redisearch_rs/inverted_index/tests/freqs_fields.rs
```diff
@@ -0,0 +1,181 @@
...
+#[should_panic]
+fn test_encode_freqs_fields_field_mask_overflow() {
+    // Encoder only supports 32 bits field mask and will panic if larger
+    let mut buf = [0u8; 3];
+    let buf = &mut buf[0..1];
+    let mut cursor = Cursor::new(buf);
+
+    let record = inverted_index::RSIndexResult::term(0, u32::MAX as u128 + 1, 1);
+    let _res = FreqsFields::default().encode(&mut cursor, 0, &record);
+}
```
chesedo: We should increase buffer size here because if this does not panic on the field mask being to big (in the future) then it will still panic because the buffer is too small (based on the test above these inputs need a buffer size of 4)

src/redisearch_rs/inverted_index_bencher/Cargo.toml
```diff
@@ -26,6 +26,10 @@ ffi.workspace = true
 inverted_index.workspace = true
 itertools.workspace = true
 redis_mock.workspace = true
+varint_ffi = { path = "../c_entrypoint/varint_ffi" }
+
+[dev-dependencies]
+varint_ffi = { path = "../c_entrypoint/varint_ffi" }
```
chesedo: Why is the extra dev-dependency one needed?

src/redisearch_rs/inverted_index/src/freqs_fields.rs
gdesmott: Good point.

I think having two different structs makes the more sense then. What do you think?

src/redisearch_rs/inverted_index_bencher/Cargo.toml
gdesmott: To ensure, this symbol used by the C code is there, see

```rust
    // The encode C implementation relies on this symbol. Re-export it to ensure it is not discarded by the linker.
    #[allow(unused_imports)]
    pub use varint_ffi::WriteVarintFieldMask;
```

src/redisearch_rs/inverted_index_bencher/benches/encoding_decoding.rs
```diff
@@ -26,5 +26,20 @@ fn benchmark_freqs_only(c: &mut Criterion) {
...
-criterion_group!(benches, benchmark_numeric, benchmark_freqs_only);
+fn benchmark_freqs_fields(c: &mut Criterion) {
+    let bencher = benchers::freqs_fields::Bencher::new(false);
+    bencher.encoding(c);
+    bencher.decoding(c);
+
+    let bencher = benchers::freqs_fields::Bencher::new(true);
+    bencher.encoding(c);
+    bencher.decoding(c);
+}
```
chesedo: The `new(false|true)` is not intuitive at first glance. Maybe make a `default()` and `wide()` versions like earlier?

src/redisearch_rs/inverted_index_bencher/src/ffi.rs
```diff
@@ -102,10 +102,69 @@ pub fn read_freqs(buffer: &mut Buffer, base_id: u64) -> (bool, inverted_index::R
...
+    buffer: &mut TestBuffer,
+    record: &mut inverted_index::RSIndexResult,
+    delta: u64,
+) -> usize {
+    let mut buffer_writer = BufferWriter::new(&mut buffer.0);
+
+    unsafe {
+        bindings::encode_freqs_fields_wide(&mut buffer_writer as *const _ as *mut _, delta, record)
+    }
+}
```
chesedo: What's this one for? I think we only use the above one right?

src/redisearch_rs/inverted_index/src/freqs_fields.rs
chesedo: Yes, two structs does make more sense then. I think making two benchers will also make more sense then.

src/redisearch_rs/inverted_index_bencher/Cargo.toml
chesedo: Is the symbol defined in a test or something? I'm confused why it won't be there with just the `[dependencies]` definition

src/redisearch_rs/inverted_index/tests/freqs_fields.rs
gdesmott: Good catch.

src/redisearch_rs/inverted_index_bencher/src/ffi.rs
chesedo: This will be become invalid if we make two benchers

gdesmott: Leftover from one of my many refactoring. :)

src/redisearch_rs/inverted_index_bencher/Cargo.toml
gdesmott: Oh I see, indeed we can remove the `dev-depencies` when making it a hard dep.
I think things got mixed up when rebasing and re-ordering patches.

Fixed.

src/redisearch_rs/inverted_index/src/freqs_fields.rs
gdesmott: I split the two structs.

I'm not convinced it's worth splitting benchers as well though. We don't care about the runtime cost there and they are very similar so sharing the same object greatly reduces code duplication.

src/redisearch_rs/inverted_index/tests/freqs_fields.rs
```diff
@@ -0,0 +1,192 @@
...
+#[should_panic]
+fn test_encode_freqs_fields_field_mask_overflow() {
+    // Encoder only supports 32 bits field mask and will panic if larger
+    let mut buf = [0u8; 100];
+    let buf = &mut buf[0..1];
+    let mut cursor = Cursor::new(buf);
+
+    let record = inverted_index::RSIndexResult::term(0, u32::MAX as u128 + 1, 1);
+    let _res = FreqsFields::default().encode(&mut cursor, 0, &record);
+}
```
chesedo: Won't the slice of the buf we are taking still be another potential panic point because the slice length is only 1?

It should be possible to remove the slice entirely since `Write` is implemented for a `Cursor` over an array

https://doc.rust-lang.org/std/io/struct.Cursor.html#impl-Write-for-Cursor%3C%5Bu8;+N%5D%3E

gdesmott: We can remove it indeed. Good catch.

gdesmott:
@chesedo : I did not realize the field mask size was arch dependent so I added a commit to handle this.

chesedo:
> @chesedo : I did not realize the field mask size was arch dependent so I added a commit to handle this.

:sweat_smile: I missed this too. Thanks!

chesedo (approved):

src/redisearch_rs/inverted_index_bencher/src/ffi.rs
```diff
@@ -102,10 +102,58 @@ pub fn read_freqs(buffer: &mut Buffer, base_id: u64) -> (bool, inverted_index::R
...
+    record: &mut inverted_index::RSIndexResult,
+    delta: u64,
+    wide: bool,
+) -> usize {
+    let mut buffer_writer = BufferWriter::new(&mut buffer.0);
+
+    if wide {
+        unsafe {
+            bindings::encode_freqs_fields_wide(
+                &mut buffer_writer as *const _ as *mut _,
```
GuyAv46: From the `encode_freqs_fields_wide` signature, I get that we need to pass a pointer to `BufferWriter`. But can you explain this syntax?

GuyAv46 (approved):

src/redisearch_rs/inverted_index_bencher/src/ffi.rs
gdesmott: Indeed, it's a bit tricky. :)

We need to pass a pointer to `bindings::BufferWriter` while writer here is a `buffer::BufferWriter`. Those can be safely casted into each other as [they have the same memory layout](https://github.com/RediSearch/RediSearch/blob/2bd48e6aaa43d5b5eaa9db76ef0945e6bf750c71/src/redisearch_rs/buffer/src/writer.rs#L22).

 `as *const _ as *mut _`  is a classic trick to by pass the type checker in such case as pointers can be safely casted into each other (deferencing those is actually unsafe).
 
 I made this clearer in this new commit: https://github.com/RediSearch/RediSearch/pull/6493/commits/3586faf25b44015a677e03740e5b139a96b6ae9b

chesedo (dismissed):

GuyAv46 (dismissed):

gdesmott:
Merge blocked on #6495

chesedo (approved):

PR #6495, opened 2025-07-16
[MOD-10446] Have Rust constructors for all the index types

chesedo:
This makes Rust constructors for all the `RSIndexResult` types. Nothiing uses these yet as the `Clone` and `Drop` implementation for `RSIndexResult` is still missing (will be in a next PR).

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/ffi/src/lib.rs
+++ b/src/redisearch_rs/ffi/src/lib.rs
@@ -31,3 +31,9 @@ include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 
 /// Rust implementation of `t_fieldMask` from `redisearch.h`
 pub type FieldMask = t_fieldMask;
+
+#[cfg(target_arch = "x86_64")]
+pub const RS_FIELDMASK_ALL: FieldMask = u128::MAX;
+
+#[cfg(not(target_arch = "x86_64"))]
+pub const RS_FIELDMASK_ALL: FieldMask = u64::MAX;
--- a/src/redisearch_rs/ffi/src/lib.rs
+++ b/src/redisearch_rs/ffi/src/lib.rs
@@ -32,8 +32,8 @@ include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 /// Rust implementation of `t_fieldMask` from `redisearch.h`
 pub type FieldMask = t_fieldMask;
 
-#[cfg(target_arch = "x86_64")]
+#[cfg(target_pointer_width = "64")]
 pub const RS_FIELDMASK_ALL: FieldMask = u128::MAX;
 
-#[cfg(not(target_arch = "x86_64"))]
+#[cfg(target_pointer_width = "32")]
 pub const RS_FIELDMASK_ALL: FieldMask = u64::MAX;
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -191,17 +191,17 @@ pub struct RSIndexResult {
 
 impl Default for RSIndexResult {
     fn default() -> Self {
-        Self::virt(0, 0, 0.0)
+        Self::virt()
     }
 }
 
 impl RSIndexResult {
     /// Create a new virtual index result
-    pub fn virt(doc_id: t_docId, field_mask: FieldMask, weight: f64) -> Self {
+    pub fn virt() -> Self {
         Self {
-            doc_id,
+            doc_id: 0,
             dmd: ptr::null(),
-            field_mask,
+            field_mask: 0,
             freq: 0,
             offsets_sz: 0,
             data: RSIndexResultData {
@@ -210,14 +210,13 @@ impl RSIndexResult {
             result_type: RSResultType::Virtual,
             is_copy: false,
             metrics: ptr::null_mut(),
-            weight,
+            weight: 0.0,
         }
     }
 
-    /// Create a new numeric index result with the given numeric value
-    pub fn numeric(doc_id: t_docId, num: f64) -> Self {
+    /// Create a new numeric index result with the given number
+    pub fn numeric(num: f64) -> Self {
         Self {
-            doc_id,
             field_mask: RS_FIELDMASK_ALL,
             freq: 1,
             data: RSIndexResultData {
@@ -229,10 +228,9 @@ impl RSIndexResult {
         }
     }
 
-    /// Create a new metric index result with the given document ID
-    pub fn metric(doc_id: t_docId) -> Self {
+    /// Create a new metric index result
+    pub fn metric() -> Self {
         Self {
-            doc_id,
             field_mask: RS_FIELDMASK_ALL,
             data: RSIndexResultData {
                 num: ManuallyDrop::new(RSNumericRecord(0.0)),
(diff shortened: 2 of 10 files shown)


src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -253,58 +274,109 @@ pub struct RSIndexResult {
...
+    fn default() -> Self {
+        Self::virt(0, 0, 0.0)
+    }
+}
+
 impl RSIndexResult {
-    /// Create a new numeric index result with the given numeric value
-    pub fn numeric(doc_id: t_docId, num: f64) -> Self {
+    /// Create a new virtual index result
+    pub fn virt(doc_id: t_docId, field_mask: FieldMask, weight: f64) -> Self {
```
gdesmott: Shouldn't we remove `pub fn freqs_only()` then? It's just another virt constructor.

src/redisearch_rs/ffi/src/lib.rs
```diff
@@ -31,3 +31,9 @@ include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 
 /// Rust implementation of `t_fieldMask` from `redisearch.h`
 pub type FieldMask = t_fieldMask;
+
+#[cfg(target_arch = "x86_64")]
+pub const RS_FIELDMASK_ALL: FieldMask = u128::MAX;
+
+#[cfg(not(target_arch = "x86_64"))]
+pub const RS_FIELDMASK_ALL: FieldMask = u64::MAX;
```
gdesmott, line 39: Isn't `t_fieldMask` always a `u128`?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -253,58 +274,109 @@ pub struct RSIndexResult {
...
+            ..Default::default()
+        }
+    }
+
+    /// Create a new hybrid metric index result with the given document ID
+    pub fn hybrid_metric(doc_id: t_docId) -> Self {
+        Self {
+            doc_id,
+            data: RSIndexResultData {
+                agg: ManuallyDrop::new(RSAggregateResult::with_capacity(2)),
```
gdesmott, line 276: why 2?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -253,58 +274,109 @@ pub struct RSIndexResult {
...
+            },
+            result_type: RSResultType::HybridMetric,
+            weight: 1.0,
+            ..Default::default()
+        }
+    }
+
+    /// Create a new term index result with the given document ID, term pointer, and weight
+    pub fn term(doc_id: t_docId, term: *mut RSQueryTerm, weight: f64) -> Self {
```
gdesmott: My version is https://github.com/RediSearch/RediSearch/pull/6493/commits/c14d8b73f9e0113424c480d527446e4a8c5bb95c

I guess we should agree on what should be passed to those constructors and what's should be default.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Oh, I did not even realize. Maybe we should have a builder pattern on these to be able to set specific values after construction? (that would mean all of them take in no args or maybe just doc id)

gdesmott: That's a very good idea, a builder pattern sounds like the right solution here. (not a blocker for this MR).

src/redisearch_rs/ffi/src/lib.rs
chesedo: Nope, it based off the arch

https://github.com/RediSearch/RediSearch/blob/0c8a28b0ad094119ce32872542b49da9e293a057/src/redisearch.h#L40-L48

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: It's just based on the C version, but I'm not sure why it is 2 :sweat_smile: This reminds me of something I saw earlier in the week which seemed to indicate that a hybrid metric only ever has 2 entries. I'll reach out to Guy to confirm

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Yes, let's put our heads together on it. We should be able to construct all the default ones the same as this C file, but still have an easy way to make ones for tests.

https://github.com/RediSearch/RediSearch/blob/0c8a28b0ad094119ce32872542b49da9e293a057/src/index_result.c

Small details that are important are things like setting the fieldmask for numeric records correctly (to ALL). I only found the Rust constructor (and therefore decoder) was wrong when I compared the inverted index output to the C in a benchmark.

src/redisearch_rs/ffi/src/lib.rs
gdesmott: This is wrong then?
https://github.com/RediSearch/RediSearch/blob/0c8a28b0ad094119ce32872542b49da9e293a057/src/redisearch_rs/headers/types_rs.h#L16

src/redisearch_rs/ffi/src/lib.rs
```diff
@@ -31,3 +31,9 @@ include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 
 /// Rust implementation of `t_fieldMask` from `redisearch.h`
 pub type FieldMask = t_fieldMask;
+
+#[cfg(target_arch = "x86_64")]
```
gdesmott: Would it be best to use [target_pointer_width](https://doc.rust-lang.org/reference/conditional-compilation.html#target_pointer_width) instead?

src/redisearch_rs/inverted_index/src/lib.rs
gdesmott: > We should be able to construct all the default ones the same as this C file, but still have an easy way to make ones for tests.

The builder pattern is the right solution here then.

chesedo: Yip, agree. Let me make that change

src/redisearch_rs/ffi/src/lib.rs
chesedo: Nice, yes, I didn't know about this one. Thanks!

src/redisearch_rs/ffi/src/lib.rs
chesedo: Not, sure. `bindgen` would have generated the Rust type from the C code. And then `cbindgen` generated that from the Rust type again. I would say it appearing in the `types_rs.h` again is an error as it should only appear in the `redisearch.h` file.

src/redisearch_rs/ffi/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6495/commits/cdd4419844ba048eb1b868589c3e84c577315fe9

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Builder done in https://github.com/RediSearch/RediSearch/pull/6495/commits/b75aa0be5393e20237c57b706bbd022e4fe1189b

I'm not sure if we want to make a proper builder struct yet though. It is possible, but I worry it might make the code too complex.

gdesmott: Yeah I think that's good for now.

src/redisearch_rs/ffi/src/lib.rs
chesedo: Fixed in #6503

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -155,43 +189,131 @@ pub struct RSIndexResult {
...
+                agg: ManuallyDrop::new(RSAggregateResult::with_capacity(2)),
+            },
+            result_type: RSResultType::HybridMetric,
+            weight: 1.0,
+            ..Default::default()
+        }
+    }
+
+    /// Create a new term index result with the given term pointer
+    pub fn term(term: *mut RSQueryTerm) -> Self {
```
```suggestion
    pub fn term(term: *const RSQueryTerm) -> Self {
```

gdesmott: Also I'm wondering if it would not be best to default to  `null` and have a explicit `term()`.

chesedo: The aim was to have a seamless replacement for this

https://github.com/RediSearch/RediSearch/blob/f232aba9844504d1433716217cf77b841945a216/src/index_result.c#L52-L68

But I see only two of the callers pass in a `term` so defaulting to `null` makes more sense.

chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6495/commits/b1ec5754e1b0aca773660a9d2911912bb44af613

gdesmott (approved):

chesedo:
> General question: Could we benefit from adding explicit unit tests that verify each constructor sets the correct field values and that builder pattern chaining works as expected?

We can but there are so many pieces of code that will call this that it is not worth adding more tests in my opinion

PR #6502, opened 2025-07-17
Wraps RedisModule Resources and Functions to Rust

DarthB:
1. MOD-...
2. #...

1. ...

#### Mark if applicable


src/redisearch_rs/ffi/src/redis_module/resources.rs
JonasKruckenberg, line 1: <del>

Okay, couple things:
1. Resources should be `NonNull` instead of `*mut`
2. remove the trait, I don't think it adds any value (there is no valid use case for being generic over resources since they are so different)
3. you simplify the macro. below is my sketch of how I would do it
```rust
macro_rules! define_redis_resource_ {
    (
        $name:ident(NonNull<$ty:ty>),
        constructor( $($param:ident: $param_ty:ty),+ ) -> Self $constructor:expr,
        destructor($dparam:ident:_) $destructor:expr
    ) => {
        // Define the struct to wrap the raw pointer
        #[derive(Debug)]
        pub struct $name(std::ptr::NonNull<$ty>);

        impl $name {
            pub fn new($($param: $param_ty),*) -> Self {
                $constructor
            }

            pub fn with<F, R>(&self, f: F) -> R
            where
                F: FnOnce(&$ty) -> R,
            {
                f(unsafe { self.0.as_ref() })
            }

            pub fn with_mut<F, R>(&mut self, f: F) -> R
            where
                F: FnOnce(&mut $ty) -> R,
            {
                f(unsafe { self.0.as_mut() })
            }
        }

        impl Drop for $name {
            fn drop(&mut self) {
                let $dparam = self.0;
                $destructor
            }
        }
    };
}
```

which can be used like this:

```rust
// this makes the syntax more familiar to readers (IMO that's a good thing even though its technically fake imposter syntax), 
// and clearly labels the ctor and dtor 
define_redis_resource_! {
    RedisKey(NonNull<crate::RedisModuleKey>),
    constructor(ctx: RedisModuleCtxPtr, name: RedisModuleStringPtr,  mode: i32) -> Self {
        let ptr = unsafe { crate::RedisModule_OpenKey.unwrap()(ctx, name, mode).cast::<crate::RedisModuleKey>() };
        Self(NonNull::new(ptr).unwrap())
    },
    destructor(ptr:_) {
        unsafe { crate::RedisModule_CloseKey.unwrap()(ptr.as_ptr()) }
    }
}
```

</del>


Edit: NVM. I played with the code some more and I don't think a macro is necessary or good to have in this case. Couple reasons:
1. Most resources have pretty different usage patterns, and ctor requirements (naming, behavior, etc.)
2. Hand-written bindings allow us more idiomatic usage of the types (see the Cursor that implements Iterator) 
3. Just simple, plain handwritten bindings are actually smaller (256LoC vs. 462LoC while doing a bit more too) and (IMO) more understandable


see https://github.com/RediSearch/RediSearch/compare/master...jonas/simplify-redis-module-wrapper for my take on how to simplify this (not complete, maybe not correct, but hopefully you get the gist)

src/redisearch_rs/rlookup/src/bindings.rs
```diff
@@ -43,3 +57,427 @@ pub enum FieldSpecType {
...
+pub struct RedisString {
+    // the context in which the string was created, used for freeing the string
+    ctx: *mut RedisModuleCtx,
+    // the actual string, wrapped in a NonNull pointer for safety
+    str: NonNull<ffi::RedisModuleString>,
+}
+
+impl Drop for RedisString {
+    fn drop(&mut self) {
+        // Safety: Assumption: c-side initialized the ftor and ftor is never changed.
```
gdesmott: what's `ftor`?

src/redisearch_rs/rlookup/src/bindings.rs
```diff
@@ -43,3 +57,427 @@ pub enum FieldSpecType {
...
+        let gs = unsafe { RedisModule_FreeString }.unwrap();
+
+        // Safety: RAII pattern ensures that free is called only once
+        unsafe { gs(self.ctx, self.str.as_ptr()) }
+    }
+}
+
+impl RedisString {
+    #[expect(unused, reason = "Used by follow-up PRs")]
+    pub fn from_raw_parts(ctx: *mut RedisModuleCtx, ptr: *const c_char, len: libc::size_t) -> Self {
```
gdesmott, line 269: Should we add  `debug_assert!(!ptr.is_null())` ?

src/redisearch_rs/rlookup/src/bindings.rs
```diff
@@ -43,3 +57,427 @@ pub enum FieldSpecType {
...
+
+/// A new-type wrapper around `RedisModuleString` that ensures proper memory management.
+/// It is designed to be used with the RAII pattern, where the string is automatically freed when it goes out of scope.
+///
+/// Use [`RedisString::from_raw_parts`] to create an instance from raw parts. The c string is accessible via [`RedisString::as_c_str`].
+/// The methods [`RedisString::to_i64`] and [`RedisString::to_f64`] can be used to convert the string to integers or floating-point numbers, respectively.
+#[derive(Debug)]
+pub struct RedisString {
+    // the context in which the string was created, used for freeing the string
+    ctx: *mut RedisModuleCtx,
```
gdesmott, line 249: What guarantees this pointer stays valid during the object lifetime?
Should this be documented in the safety pre-conditions?

src/redisearch_rs/rlookup/src/bindings.rs
```diff
@@ -43,3 +57,427 @@ pub enum FieldSpecType {
...
+            ctx,
+            str: NonNull::new(ptr).unwrap(),
+        }
+    }
+
+    /// Returns a pointer to the underlying `RedisModuleString` object.
+    /// This is useful for passing the string to C functions that expect a `RedisModuleString
+    /// Safety: The returned pointer is still managed by this object and should only be used with RedisModule functions.
+    #[expect(unused, reason = "Used by follow-up PRs")]
+    pub unsafe fn get_ptr(&mut self) -> *mut ffi::RedisModuleString {
```
gdesmott: Why not `as_ptr()`? That's the idiomatic name.

src/redisearch_rs/rlookup/src/bindings.rs
```diff
@@ -43,3 +57,427 @@ pub enum FieldSpecType {
...
+
+        // Safety: By RAII we assure to provide a valid pointer to the string and we assume then the c-side will be safe to call.
+        let ptr = unsafe { gs(raw_str, ptr::from_mut(&mut len)) };
+
+        if ptr.is_null() {
+            None
+        } else {
+            // Safety: We assume the ptr and len given by the c-side are valid.
+            let bytes = unsafe { slice::from_raw_parts(ptr.cast(), len) };
+            Some(CStr::from_bytes_with_nul(bytes).unwrap())
```
gdesmott: `RedisModuleString` is guaranteed to be NULL terminated?

gdesmott (commented):
The general design makes sense to me.

Just wondering if it would make sense to have a wrapper around `RedisModuleCtx` so we can have proper methods on it.

src/redisearch_rs/rlookup/src/bindings.rs
DarthB: Use function pointer - Somehow for me they are ftors :-D

src/redisearch_rs/rlookup/src/bindings.rs
DarthB: Yeah, added it at two places

src/redisearch_rs/rlookup/src/bindings.rs
DarthB: Added comments

src/redisearch_rs/rlookup/src/bindings.rs
DarthB: No, but some follow-up code expects this. I just used, `ok()` instead of `unwrap` here

src/redisearch_rs/rlookup/src/bindings.rs
```diff
@@ -50,6 +66,75 @@ pub enum FieldSpecType {
...
+#[bitflags]
+#[repr(u32)] // should be c_unit
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum KeyMode {
+    /// Open the key for reading.
+    Read = 1 << 0,
+
+    /// Open the key for writing.
+    Write = 1 << 1,
+
```
Itzikvaknin, line 81: Whats the reason for that gap?

src/redisearch_rs/rlookup/src/bindings.rs
```diff
@@ -140,6 +225,468 @@ impl AsRef<ffi::IndexSpecCache> for IndexSpecCache {
...
+    // the actual string, wrapped in a NonNull pointer for safety
+    str: NonNull<ffi::RedisModuleString>,
+}
+
+impl Drop for RedisString {
+    fn drop(&mut self) {
+        // Safety: Assumption: c-side initialized the function ptr and it is is never changed.
+        let gs = unsafe { RedisModule_FreeString }.unwrap();
+
+        // Safety: RAII pattern ensures that free is called only once
```
Itzikvaknin, line 260: from the Rust side, the assumption "is never changed" includes free?

src/redisearch_rs/rlookup/src/bindings.rs
```diff
@@ -140,6 +225,468 @@ impl AsRef<ffi::IndexSpecCache> for IndexSpecCache {
...
+        // is returned when the reply is not a string or an error.
+        let ptr = unsafe { gs(self.0.as_ptr(), ptr::from_mut(&mut len)) };
+
+        if ptr.is_null() {
+            None
+        } else {
+            // Safety: The c-side, RedisModule_CallReplyStringPtr, must returns a valid pointer to a c string.
+            let bytes = unsafe { slice::from_raw_parts(ptr.cast(), len) };
+
+            // We ass
```
Itzikvaknin: ?

src/redisearch_rs/rlookup/src/bindings.rs
DarthB: For `RedisKey` the value of the 2nd byte is reserved or for the keys unused.

src/redisearch_rs/rlookup/src/bindings.rs
DarthB: Comment addressed in https://github.com/RediSearch/RediSearch/pull/6502/commits/40ae8cdd31dc82de2844135457fb17367f5348a4

src/redisearch_rs/rlookup/src/bindings.rs
DarthB: https://github.com/RediSearch/RediSearch/pull/6502/commits/40ae8cdd31dc82de2844135457fb17367f5348a4

Itzikvaknin (approved):
Consider using the `redis-module` crate when working with Redis types

PR #6504, opened 2025-07-17
[MOD-10380] Implement the encodeFieldsOnly and encodeFieldsOnlyWide encoder/decoder in Rust

gdesmott:

Implement the fields only encoder and decoder in Rust.

#### Mark if applicable


src/redisearch_rs/inverted_index/src/fields_only.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+impl Decoder for FieldsOnly {
+    fn decode<R: Read>(&self, reader: &mut R, base: t_docId) -> std::io::Result<DecoderResult> {
+        let (decoded_values, _bytes_consumed) = qint_decode::<2, _>(reader)?;
+        let [delta, field_mask] = decoded_values;
+
+        let record = RSIndexResult::term()
+            .doc_id(base + delta as u64)
```
GuyAv46: Why not as `t_docId` (if it's not implicit anyway)?
```suggestion
            .doc_id(base + delta as t_docId)
```
In this case, the `as` is on the delta only, right? (not the sum result).

gdesmott: > In this case, the as is on the delta only, right? (not the sum result).

Yes, we need to cast it as `delta` is a `u32`.

> Why not as t_docId (if it's not implicit anyway)?

Good point.
`t_docId` is actually `u64` but you're right it's clearer that way.

I changed it and in https://github.com/RediSearch/RediSearch/pull/6493 as well.

gdesmott:
Merged blocked on https://github.com/RediSearch/RediSearch/pull/6493

chesedo (approved):

PR #6512, opened 2025-07-21
pin redisbench-admin version

JoanFM:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

### DIFF
--- a/.github/workflows/flow-micro-benchmarks-runner.yml
+++ b/.github/workflows/flow-micro-benchmarks-runner.yml
@@ -75,7 +75,7 @@ jobs:
           cd ..
           # Then install Python dependencies
           sudo apt install python3-pip -y
-          pip3 install --upgrade pip PyYAML setuptools redisbench-admin>=0.11.37
+          pip3 install --upgrade pip PyYAML setuptools redisbench-admin==0.11.40
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/flow-micro-benchmarks-runner.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           cd ..
           # Then install Python dependencies
           sudo apt install python3-pip -y
-          pip3 install --upgrade pip PyYAML setuptools redisbench-admin>=0.11.37
+          pip3 install --upgrade pip PyYAML setuptools redisbench-admin==0.11.40
```
alonre24: can we refactor the flows so that [tests/benchmarks/requirements.txt](https://github.com/RediSearch/RediSearch/pull/6512/files#diff-bd5cd5e9e92cbeddea2eea6d970654ab35d9856526bdcf4cc99dad6f4dbdda1b) will apply for all the flows that are using redisbench-admin? (and then we can pin the version only in one place)

JoanFM: I think we can do. Maybe we will be installing more than needed. But I believe it could still be worth it

alonre24 (approved):

PR #6513, opened 2025-07-21
ci: pin a Rust nightly version

gdesmott:
Should prevent accidental breaking of the CI.

#### Mark if applicable

### DIFF
--- a/.install/test_deps/common_installations.sh
+++ b/.install/test_deps/common_installations.sh
@@ -20,6 +20,8 @@ activate_venv() {
 	fi
 }
 
+# retrieve nightly version from build.sh
+NIGHTLY_VERSION=$(grep "NIGHTLY_VERSION=" build.sh | cut -d'=' -f2 | tr -d '"')
 # --allow-downgrade:
 #   Allow `rustup` to install an older `nightly` if the latest one
 #   is missing one of the components we need.
@@ -29,7 +31,7 @@ activate_venv() {
 #   Required to run `cargo miri test` for UB detection
 # rust-src:
 #   Required to build RedisJSON with address sanitizer
-rustup toolchain install nightly \
+rustup toolchain install $NIGHTLY_VERSION \
     --allow-downgrade \
     --component llvm-tools-preview \
     --component miri \
@@ -40,7 +42,7 @@ cargo install cargo-llvm-cov --locked
 # Make sure `miri` is fully operational before running tests with it.
 # See https://github.com/rust-lang/miri/blob/master/README.md#running-miri-on-ci
 # for more details.
-cargo +nightly miri setup
+cargo +$NIGHTLY_VERSION miri setup
 
 python3 -m venv venv
 activate_venv
--- a/build.sh
+++ b/build.sh
@@ -552,11 +552,16 @@ run_rust_tests() {
     export RUSTFLAGS="${RUSTFLAGS:+${RUSTFLAGS} }-D warnings"
   fi
 
+  # Pin a specific working version of nightly to prevent breaking the CI because
+  # regressions in a nightly build.
+  # Make sure to synchronize updates across all modules: Redis and RedisJSON.
+  NIGHTLY_VERSION="nightly-2025-07-30"
+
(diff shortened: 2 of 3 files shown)


raz-mon (approved):

.install/test_deps/common_installations.sh
```diff
@@ -40,7 +44,7 @@ cargo install cargo-llvm-cov --locked
 # Make sure `miri` is fully operational before running tests with it.
 # See https://github.com/rust-lang/miri/blob/master/README.md#running-miri-on-ci
 # for more details.
-cargo +nightly miri setup
+cargo +$NIGHTLY_VERSION miri setup
```
chesedo, line 45: Won't this mean the `build.sh` also needs updating?

gdesmott: Indeed, `+nightly` does not pick our version unfortunately. 

Should be ok now.

chesedo (approved):

gdesmott:
@raz-mon : I'm afraid this is not going to work.

`RedisJSON` relies on nightly as well and no longer find it as it looks for the latest one:
 https://github.com/RediSearch/RediSearch/actions/runs/16418681005/job/46391047723?pr=6513#step:18:231

So we need a way to pin the same version across repos. I'm not sure what would be a convenient way to do that?

Or we could install the latest nightly as part of the `RedisJSON`  setup. If we just need to build it and use the module directly then it's fine having two different nightly version.

gdesmott:
@chesedo @DarthB any suggestion?

chesedo:
> @chesedo @DarthB any suggestion?

No, sorry I don't see an easy way around this. The proper solution is having a unified system for all the CIs, but that is a big undertaking for another time.

Maybe using a pre-build (Rust) image for all the CI is a step in that direction and would be quicker to make. But that still requires some co-ordination.

gdesmott:
@alonre24 I opened https://github.com/RedisJSON/RedisJSON/issues/1362 on RedisJSON.

Another (ugly) option would be to install both nightly versions as a workaround; see my second commit in this PR.

gdesmott:
@alonre24 : I  tested locally with version `nightly-2025-07-30` which works fine.

So once RedisJSON has pinned this version as well we can merge this.

build.sh
```diff
@@ -539,11 +539,15 @@ run_rust_tests() {
     export RUSTFLAGS="${RUSTFLAGS:+${RUSTFLAGS} }-D warnings"
   fi
 
+  # Pin a specific working version of nightly to prevent breaking the CI because
+  # regressions in a nightly build.
+  NIGHTLY_VERSION="nightly-2025-07-30"
```
alonre24, line 558: Please add that changing this nightly version requires alignment with other modules. 
Also, will this `NIGHTLY_VERSION` be used when building not for testing (that is, when calling simply `.build.sh`)?

gdesmott: > Please add that changing this nightly version requires alignment with other modules.

Updated comment.

>  Also, will this NIGHTLY_VERSION be used when building not for testing (that is, when calling simply .build.sh)?

No, nightly is only used when explicitly enabling code coverage or sanitizer. See https://github.com/RediSearch/RediSearch/blob/master/build.sh#L543

alonre24: In the Redis open source CI, both search and JSON are built as submodules. In particular, they call `./build.sh` for building search. So as far as I understand, we will not be aligned with JSON build unless we use nightly as the default build as well, no?

gdesmott: It's fine having RediSearch built using Rust stable and JSON with nightly. Both will be used as modules by redis core so the Rust version used to build this module does not really matter.

We just need to make sure the CI runners have the  proper versions of the compiler installed, stable and nightly, so it can build everything.

gdesmott:
For the record here are the PR pinning the same version in RediJSON and redismodule-rs:
- https://github.com/RedisJSON/RedisJSON/pull/1375
- https://github.com/RedisLabsModules/redismodule-rs/pull/418

chesedo (approved):

PR #6313, opened 2025-06-13
[MOD-10562] - Multiple I/O uvloop threads for Coordinator

JoanFM:
### Current
The MRCluster implementation is single-threaded, with all I/O operations running on a single thread.

### Change
This PR refactors the MRCluster architecture to support multiple I/O threads:

1. **MRCluster Structure**: Remains the main entrypoint of the "coordinator". The coordinator still has a `cluster_g` global object. However, this global object right now can be seen as a Pool of `IORuntimeCtx` that will make sure that all operations balanced between different IORuntimes in a round robin fashion, thus working in different IO threads.

2. **IORuntimeCtx**: Introduced new structure encapsulating I/O runtime context with:
   - Connection manager
   - Cluster topology
   - Work queue
   - UV loop for async I/O (plus different UV primitives)

The main interfaces for IORuntimeCtx are:
 - **IORuntimeCtx_Schedule**: Schedule a given callback with private data to be queued to run in the IORuntime UV thread
 - **IORuntimeCtx_Schedule_Topology**: Schedule operations to update the topology (there is a mechanism for Topology Updates to go before other updates).

The **IORuntimeCtx** still starts the thread in a lazy manner.

3. **MRCtx** and **MRIterator** get assigned a IORuntimeIdx which make sure all the operations are handled in a single thread.
    
4. **Thread Management**: Improve thread creation, synchronization, and proper shutdown
5. **Topology Management**: Modified to propagate updates across all I/O contexts

### Outcome
Aim to improve scalability by distributing work across multiple I/O threads while maintaining backward compatibility.

#### Main objects this PR modified
1. MRCluster
2. IORuntimeCtx (new)
3. MR_* API functions

So:

1 MRCluster -> (NUMBER_IO_THREADS) IORuntimeCtx -> 1 MRConnManager -> (CONN_PER_SHARD/NUMBER_IO_THREADS) Connections (ConnPool)

## Notes

The Number of IO Threads will remain constant, no configuration setting will be added. When trying to add it a potential deadlock was seen between the main thread holding the GIL while joining a thread and connection callbacks also trying to acquire the GIL, so if this is added again this will need to be considered.

PR #6514, opened 2025-07-21
refactor pending queues

JoanFM:
Some additional work on top of #6313

### DIFF
--- a/src/coord/rmr/io_runtime_ctx.h
+++ b/src/coord/rmr/io_runtime_ctx.h
@@ -52,7 +52,7 @@ typedef struct {
   // Request queue and topology requests
   MRWorkQueue *queue;
   struct queueItem *pendingTopo; // The pending topology to be applied
-  arrayof(uv_async_t *) pendingQueues;
+  bool pendingItems; // Are there any pending items waiting for Topology to be applied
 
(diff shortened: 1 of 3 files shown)


src/coord/rmr/io_runtime_ctx.h
```diff
@@ -52,7 +52,7 @@ typedef struct {
   // Request queue and topology requests
   MRWorkQueue *queue;
   struct queueItem *pendingTopo; // The pending topology to be applied
-  arrayof(uv_async_t *) pendingQueues;
+  bool pendingItems; // Are there any pending items waiting for Topology to be applied
```
GuyAv46, line 55: We can check the size of the queue instead. WDYT?

JoanFM: it seems to me that a boolean reflects better how the underlying logic works. An Async call triggers everything in the Queue. But I can change it

GuyAv46: OK, I also think this is an acceptable redundancy.
Alternatively, we can always attempt to trigger the async handle; worst case, it has nothing to do.

GuyAv46 (approved):

PR #6516, opened 2025-07-21
[MOD-10431] accumulate conn state from different IO Runtimes

JoanFM:
Follow up from #6313 

This PR makes sure that the SHARD_CONNECTION_STATES command gives the real status of all the connections from all the IO Runtimes

### DIFF
--- a/src/coord/rmr/conn.c
+++ b/src/coord/rmr/conn.c
@@ -213,11 +213,11 @@ void MRConnManager_FillStateDict(MRConnManager *mgr, dict *stateDict) {
     if (existingEntry) {
       // Replace existing entry - the old value will be freed by the destructor
       dictReplace(stateDict, key, newList);
-      rm_free(key); // dictReplace doesn't take ownership of key if it already exists
     } else {
       // Add new entry
       dictAdd(stateDict, key, newList);
     }
+    rm_free(key);
   }
 
   dictReleaseIterator(it);
--- a/src/coord/rmr/conn.c
+++ b/src/coord/rmr/conn.c
@@ -139,24 +139,90 @@ void MRConnManager_Free(MRConnManager *mgr) {
   dictRelease(mgr->map);
 }
 
-void MRConnManager_ReplyState(MRConnManager *mgr, RedisModuleCtx *ctx) {
-  RedisModule_ReplyWithMap(ctx, dictSize(mgr->map));
-  dictIterator *it = dictGetIterator(mgr->map);
+void MRConnManager_ReplyState(dict *stateDict, RedisModuleCtx *ctx) {
+  RedisModule_ReplyWithMap(ctx, dictSize(stateDict));
+  dictIterator *it = dictGetIterator(stateDict);
   dictEntry *entry;
-  while ((entry = dictNext(it))) {
-    MRConnPool *pool = dictGetVal(entry);
-    RedisModuleString *key = RedisModule_CreateStringPrintf(ctx, "%s:%d", pool->conns[0]->ep.host,
-                                                                          pool->conns[0]->ep.port);
-    RedisModule_ReplyWithString(ctx, key);
-    RedisModule_FreeString(ctx, key);
-    RedisModule_ReplyWithArray(ctx, pool->num);
 
-    for (size_t i = 0; i < pool->num; i++) {
-      RedisModule_ReplyWithCString(ctx, MRConnState_Str(pool->conns[i]->state));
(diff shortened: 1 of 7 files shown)


src/coord/rmr/conn.h
```diff
@@ -64,7 +64,9 @@ typedef struct {
 
 void MRConnManager_Init(MRConnManager *mgr, int nodeConns);
 
-void MRConnManager_ReplyState(MRConnManager *mgr, RedisModuleCtx *ctx);
+void MRConnManager_ReplyState(dict *stateDict, RedisModuleCtx *ctx);
+
+void MRConnManager_FillStateDict(MRConnManager *mgr, dict *stateDict);
```
DvirDukhan: add comments about those functions

src/coord/rmr/conn.c
```diff
@@ -139,25 +139,91 @@ void MRConnManager_Free(MRConnManager *mgr) {
   dictRelease(mgr->map);
 }
 
-void MRConnManager_ReplyState(MRConnManager *mgr, RedisModuleCtx *ctx) {
-  RedisModule_ReplyWithMap(ctx, dictSize(mgr->map));
+void MRConnManager_ReplyState(dict *stateDict, RedisModuleCtx *ctx) {
+  RedisModule_ReplyWithMap(ctx, dictSize(stateDict));
```
DvirDukhan, line 144: do we need null validation here for `stateDict`?

src/coord/rmr/rmr.c
```diff
@@ -323,27 +322,63 @@ struct ReplyClusterInfoCtx {
...
+  RedisModuleBlockedClient *bc = mt_bc->bc;
   RedisModuleCtx *ctx = RedisModule_GetThreadSafeContext(bc);
-  MRConnManager_ReplyState(&ioRuntime->conn_mgr, ctx);
-  IORuntimeCtx_RequestCompleted(ioRuntime);
-  RedisModule_FreeThreadSafeContext(ctx);
-  RedisModule_BlockedClientMeasureTimeEnd(bc);
-  RedisModule_UnblockClient(bc, NULL);
-  rm_free(replyClusterInfoCtx);
+  pthread_mutex_lock(&mt_bc->lock);
+  MRConnManager_FillStateDict(&ioRuntime->conn_mgr, mt_bc->replyDict);
```
DvirDukhan, line 346: why do we need this section?

src/coord/rmr/rmr.c
```diff
@@ -323,27 +322,63 @@ struct ReplyClusterInfoCtx {
...
 void MR_GetConnectionPoolState(RedisModuleCtx *ctx) {
   RedisModuleBlockedClient *bc = RedisModule_BlockClient(ctx, NULL, NULL, NULL, 0);
   RedisModule_BlockedClientMeasureTimeStart(bc);
-  struct ReplyClusterInfoCtx *replyClusterInfoCtx = rm_new(struct ReplyClusterInfoCtx);
-  size_t idx = MRCluster_AssignRoundRobinIORuntimeIdx(cluster_g);
-  replyClusterInfoCtx->bc = bc;
-  replyClusterInfoCtx->ioRuntime = cluster_g->io_runtimes_pool[idx];
-  IORuntimeCtx_Schedule(replyClusterInfoCtx->ioRuntime, uvGetConnectionPoolState, replyClusterInfoCtx);
+  struct MultiThreadedRedisBlockedCtx *mt_bc = rm_new(struct MultiThreadedRedisBlockedCtx);
+  mt_bc->pending_replies = cluster_g->num_io_threads;
```
DvirDukhan: is this correct?

src/coord/rmr/rmr.c
JoanFM: We need to fill the Dict dynamically. (So we have either an empty dict or a partially filleed one, where each connection manager "appends" its info).

src/coord/rmr/rmr.c
JoanFM: Maybe I can rename, pending_replies may not the best name. But yes, first every IO Runtime needs to fill its information. Once this goes down to 0, you consider the accumulated response is ready to reply and unblock client.

src/coord/rmr/conn.c
JoanFM: I will add an RS_ASSERT. This is only called once and it is clear it does not happen.

DvirDukhan (approved):

PR #6523, opened 2025-07-22
[MOD-10543] Make a reader which can skip duplicates

chesedo:
Just a wrapping inverted index reader which is able to skip over documents with duplicate IDs when reading

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -832,6 +832,47 @@ impl<'a, D: Decoder> IndexReader<'a, D> {
     }
 }
 
+/// A reader that skips duplicate records in the index. It is used to ensure that the same document
+/// ID is not returned multiple times in the results. This is useful when the index contains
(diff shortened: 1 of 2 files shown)


gdesmott (approved):

alonre24 (commented):
Can you refer me to the equivalent logic in the C code? Do we need micro benchmarks here?

chesedo:
> Can you refer me to the equivalent logic in the C code? Do we need micro benchmarks here?

It's for this code in the index reader. During the design we decided to put this logic in a wrapper rather than the reader itself. This way things which won't need it won't need to spend extra cycles on the boolean flag.

https://github.com/RediSearch/RediSearch/blob/cf1b1e29035cb954ec8fdf7217c7ca76fcdf7b92/src/inverted_index/inverted_index.c#L1026-L1035

We don't need a benchmark here as there are a limited number of ways to write this. If this is slow then it will mean using a different design all together which will make the benchmark obsolete.

alonre24 (dismissed):

alonre24 (approved):

gdesmott (approved):

PR #6510, opened 2025-07-21
[MOD-10327] Rust inverted index reader

chesedo:
## Describe the changes in the pull request
Implements an inverted index reader in Rust. This only has the reader and is missing the code to:
- skip to
- verify field mask expiration for doc ID
- skip duplicates (#6523)
- filter out records (#6524)

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #6524, opened 2025-07-23
[MOD-10544] Make a reader which filters based on the field mask

chesedo:
A wrapper around an inverted index reader which can filter based on the field mask

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -832,5 +832,39 @@ impl<'a, D: Decoder> IndexReader<'a, D> {
     }
 }
 
+/// A reader that filters out records that do not match a given field mask. It is used to
+/// filter records in an index based on their field mask, allowing only those that match the
+/// specified mask to be returned.
+pub struct FilterMaskReader<I> {
+    /// Mask which a record needs to match to be valid
+    mask: t_fieldMask,
+
+    /// The inner reader that will be used to read the records from the index.
+    inner: I,
+}
+
+impl<I: Iterator<Item = RSIndexResult>> FilterMaskReader<I> {
+    /// Create a new filter mask reader with the given mask and inner iterator
+    pub fn new(mask: t_fieldMask, inner: I) -> Self {
+        Self { mask, inner }
+    }
+}
+
+impl<I: Iterator<Item = RSIndexResult>> Iterator for FilterMaskReader<I> {
+    type Item = RSIndexResult;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            let next = self.inner.next()?;
+
+            if next.field_mask & self.mask == 0 {
+                continue;
+            }
+
+            return Some(next);
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests;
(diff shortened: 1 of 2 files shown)


src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -677,5 +677,39 @@ impl<E: Encoder> InvertedIndex<E> {
...
+/// specified mask to be returned.
+pub struct FilterMaskReader<I> {
+    /// Mask which a record needs to match to be valid
+    mask: t_fieldMask,
+
+    /// The inner reader that will be used to read the records from the index.
+    inner: I,
+}
+
+impl<I: Iterator<Item = RSIndexResult>> FilterMaskReader<I> {
```
gdesmott, line 846: At some point, maybe not now, maybe it could make sense to have a result reader trait bound to `Iterator<Item = RSIndexResult>`?

gdesmott (approved):

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Do you mean a new trait that has `Iterator<Item = RSIndexResult>` as the super trait? Then yes, I think it will match 1-to-1 with the new iterator redis made.

gdesmott: yes exactly.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -677,5 +677,39 @@ impl<E: Encoder> InvertedIndex<E> {
     }
 }
 
+/// A reader that filters out records that do not match a given field mask. It is used to
+/// filter records in an index based on their field mask, allowing only those that match the
+/// specified mask to be returned.
+pub struct FilterMaskReader<I> {
```
gdesmott, line 838: Should we get rid of `DecoderResult` and have `Decoder::decoder()` returns `RSIndexResult` directly then?

chesedo: Yes, that change is happening in #6510 over here -> https://github.com/RediSearch/RediSearch/pull/6510/files#diff-2ee9bad447a86c8ba40c7a36e41c24038e3d75a710404be5e297d5bc34313fa4R556

gdesmott: I thought I saw it somewhere but did not find it. Getting lost in all those PRs. :sweat_smile:

chesedo:
> Can you refer me to the equivalent logic in the C code? Do we need micro benchmarks here?

This is for all the decoders which do a filter on their very last step like this

https://github.com/RediSearch/RediSearch/blob/cf1b1e29035cb954ec8fdf7217c7ca76fcdf7b92/src/inverted_index/inverted_index.c#L592-L598

And this

https://github.com/RediSearch/RediSearch/blob/cf1b1e29035cb954ec8fdf7217c7ca76fcdf7b92/src/inverted_index/inverted_index.c#L600-L606

We decided to do the split in the design so that the decoders do just decoding and filtering is decoupled into another struct. After this a similar wrapper will be needed for the numeric filtering.

Oh, and we don't need to micro-benchmark this since there is little to optimize in it. We are more likely to switch to a whole different design if this turns out to be slow which will make the micro-benchmark obsolete.

PR #6529, opened 2025-07-23
Hybrid implicit load [MOD-10249]

ofiryanai:
Add Implicit load to key + score according to PRD

HybridRequest tail

### DIFF
--- a/src/hybrid/hybrid_request.c
+++ b/src/hybrid/hybrid_request.c
@@ -8,9 +8,13 @@
 extern "C" {
 #endif
 
+// Field names for implicit LOAD step
+#define HYBRID_IMPLICIT_DOC_ID "doc_id"
+#define HYBRID_IMPLICIT_COMBINED_SCORE "combined_score"
+
 /**
  * Create an implicit LOAD step for hybrid search when no explicit LOAD is provided.
- * The implicit LOAD includes @__key (document ID) and @__score (combined score).
+ * The implicit LOAD includes doc_id (document ID) and combined_score (combined score).
  *
  * @param plan The aggregation plan to get RLookup from
  * @return Newly allocated PLN_LoadStep with implicit fields, or NULL on failure
@@ -23,7 +27,7 @@ static PLN_LoadStep* CreateImplicitLoadStep(AGGPlan *plan) {
 
     lstp->base.type = PLN_T_LOAD;
     lstp->base.dtor = loadDtor;
-    lstp->nkeys = 2;  // @__key and @__score
+    lstp->nkeys = 2;  // doc_id and combined_score
     lstp->keys = rm_calloc(2, sizeof(*lstp->keys));
 
     if (!lstp->keys) {
@@ -37,9 +41,9 @@ static PLN_LoadStep* CreateImplicitLoadStep(AGGPlan *plan) {
         lookup = AGPLN_GetLookup(plan, NULL, AGPLN_GETLOOKUP_FIRST);
     }
 
-    // Create RLookupKeys for @__key and @__score
-    lstp->keys[0] = RLookup_GetKey_Load(lookup, UNDERSCORE_KEY, UNDERSCORE_KEY, RLOOKUP_F_NOFLAGS);
-    lstp->keys[1] = RLookup_GetKey_Load(lookup, UNDERSCORE_SCORE, UNDERSCORE_SCORE, RLOOKUP_F_NOFLAGS);
+    // Create RLookupKeys for doc_id and combined_score
+    lstp->keys[0] = RLookup_GetKey_Load(lookup, HYBRID_IMPLICIT_DOC_ID, UNDERSCORE_KEY, RLOOKUP_F_NOFLAGS);
+    lstp->keys[1] = RLookup_GetKey_Load(lookup, HYBRID_IMPLICIT_COMBINED_SCORE, UNDERSCORE_SCORE, RLOOKUP_F_NOFLAGS);
 
     // Initialize args as empty since this is implicit (no user-provided arguments)
     lstp->args = (ArgsCursor){0};  // Zero-initialize for empty cursor
--- a/src/hybrid/hybrid_request.c
+++ b/src/hybrid/hybrid_request.c
@@ -1,10 +1,51 @@
 #include "hybrid/hybrid_request.h"
 #include "pipeline/pipeline.h"
 #include "hybrid/hybrid_scoring.h"
+#include "document.h"
+#include "aggregate/aggregate_plan.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+/**
+ * Create an implicit LOAD step for hybrid search when no explicit LOAD is provided.
+ * The implicit LOAD includes @__key (document ID) and @__score (combined score).
+ *
+ * @param plan The aggregation plan to get RLookup from
+ * @return Newly allocated PLN_LoadStep with implicit fields, or NULL on failure
+ */
+static PLN_LoadStep* CreateImplicitLoadStep(AGGPlan *plan) {
+    PLN_LoadStep *lstp = rm_calloc(1, sizeof(*lstp));
+    if (!lstp) {
+        return NULL;
+    }
+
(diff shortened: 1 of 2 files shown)


src/hybrid/hybrid_request.c
```diff
@@ -73,6 +118,16 @@ int HybridRequest_BuildPipeline(HybridRequest *req, const HybridPipelineParams *
     ResultProcessor *merger = RPHybridMerger_New(params->scoringCtx, depleters, req->nrequests);
     QITR_PushRP(&req->pipeline.qctx, merger);
 
+    // Create implicit LOAD step if none exists (MOD-10249: automatic @__key and @__score return)
+    // This should happen after merge operation, for the final tail pipeline
+    if (!loadStep) {
```
kei-nan, line 89: consider moving this to the else clause of the if below this one.

ofiryanai: see the comment in the if below this one. I considered it, and thought that adding it will make the comment and the flow less understandable. btw, this needs to move soon anyway, we figured out yesterday that this implicit load should be in each subquery rather than on the tail.

ofiryanai: I think this is outdated, resolving for now

src/hybrid/hybrid_request.c
```diff
@@ -49,12 +56,20 @@ int HybridRequest_BuildPipeline(HybridRequest *req, const HybridPipelineParams *
...
         if (loadStep) {
-            // Add a loader to load the fields specified in the LOAD step
-            ResultProcessor *loader = RPLoader_New(AREQ_SearchCtx(areq), AREQ_RequestFlags(areq), lastLookup, loadStep->keys, loadStep->nkeys, false, &areq->stateflags);
-            QITR_PushRP(qctx, loader);
+          keys = loadStep->keys;
+          nkeys = loadStep->nkeys;
+        } else {
+          // If load was not specified, implicitly load doc key
+          keys = array_new(RLookupKey *, 1);
+          array_append(keys, RLookup_GetKey_Load(lookup, HYBRID_IMPLICIT_DOC_ID, UNDERSCORE_KEY, RLOOKUP_F_NOFLAGS));
```
kei-nan: I think the loader creates its own copy of the array.
So I think there is a memory leak here.
Consider creating the array outside of the loop and freeing it afterwards if it was allocated.

ofiryanai: Good catch! I've fixed the issue by using stack allocation instead of heap allocation for the keys array

src/hybrid/hybrid_request.c
```diff
@@ -70,7 +88,13 @@ int HybridRequest_BuildPipeline(HybridRequest *req, const HybridPipelineParams *
 
     // Create the hybrid merger that combines results from all depleter processors
     // This is where the magic happens - results from different search modalities are merged
-    ResultProcessor *merger = RPHybridMerger_New(params->scoringCtx, depleters, req->nrequests);
+    const RLookupKey *scoreKey = NULL;
+    if (!loadStep) {
+        // implicit load score as well as key
+        RLookup *lookup = AGPLN_GetLookup(&req->pipeline.ap, NULL, AGPLN_GETLOOKUP_FIRST);
```
kei-nan: since the pipeline plan could be empty not sure you will get a lookup here.
maybe pass the score key string to the hybrid merger ctor which will use its lookup to load the score key?

src/hybrid/hybrid_request.c
```diff
@@ -1,11 +1,18 @@
...
+#include "aggregate/aggregate_plan.h"
+#include "rmutil/args.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+// Field names for implicit LOAD step
+#define HYBRID_IMPLICIT_DOC_ID "doc_id"
+#define HYBRID_IMPLICIT_COMBINED_SCORE "combined_score"
```
raz-mon: Didn't we say we'll use the regular `"score"` field name?

raz-mon: Is this used?

src/hybrid/hybrid_request.c
```diff
@@ -1,11 +1,18 @@
...
+#include "document.h"
+#include "aggregate/aggregate_plan.h"
+#include "rmutil/args.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+// Field names for implicit LOAD step
+#define HYBRID_IMPLICIT_DOC_ID "doc_id"
```
raz-mon: We need the key, not the doc-id

src/hybrid/hybrid_request.c
```diff
@@ -70,7 +88,13 @@ int HybridRequest_BuildPipeline(HybridRequest *req, const HybridPipelineParams *
 
     // Create the hybrid merger that combines results from all depleter processors
     // This is where the magic happens - results from different search modalities are merged
```
raz-mon: Remove this comment please

tests/cpptests/test_cpp_hybrid.cpp
```diff
@@ -468,3 +468,138 @@ TEST_F(HybridRequestTest, testHybridRequestBuildPipelineTail) {
...
+    PLN_LoadStep *requestLoadStep = (PLN_LoadStep *)AGPLN_FindStep(&areq->pipeline.ap, NULL, NULL, PLN_T_LOAD);
+    EXPECT_EQ(nullptr, requestLoadStep) << "Request " << i << " should not have PLN_LoadStep (uses RPLoader instead)";
+  }
+
+  // Verify individual request pipeline structures include loaders for document key loading
+  std::vector<ResultProcessorType> expectedIndividualPipeline = {RP_DEPLETER, RP_LOADER, RP_INDEX};
+  for (size_t i = 0; i < hybridReq->nrequests; i++) {
+    AREQ *areq = hybridReq->requests[i];
+    std::string pipelineName = "Request " + std::to_string(i) + " pipeline with implicit LOAD";
+    VerifyPipelineChain(areq->pipeline.qctx.endProc, expectedIndividualPipeline, pipelineName);
```
raz-mon, line 549: This can be added to the loop above.
A validation of the keys loaded would be good as well

tests/cpptests/test_cpp_hybrid.cpp
```diff
@@ -468,3 +468,138 @@ TEST_F(HybridRequestTest, testHybridRequestBuildPipelineTail) {
...
+  }
+
+  // Clean up
```
raz-mon, line 571: Let's validate we have a loader that loads the key in this case.
We should also make sure we get the scores as well

src/hybrid/hybrid_request.c
ofiryanai: done

src/hybrid/hybrid_request.c
ofiryanai: removed

src/hybrid/hybrid_request.c
ofiryanai: done

src/hybrid/hybrid_request.c
ofiryanai: Thanks for the suggestion! However, I'll always get a lookup here because `AGPLN_Init` is called in `HybridRequest_New`, which initializes the aggregation plan with a lookup even when it starts empty. I think its sensible to assume at this stage that HybridRequest was initialized.
Also, the HybridMerger RP doesn't have its own lookup to get keys from. As far as I understands RPs don't own lookups, AggPlans do

ofiryanai (commented):
responded to CR and made some changes. @raz-mon @kei-nan please re-review

tests/cpptests/test_cpp_hybrid.cpp
ofiryanai: done

tests/cpptests/test_cpp_hybrid.cpp
ofiryanai: done

raz-mon (dismissed):

PR #6532, opened 2025-07-24
Disable SHARD_K_RATIO when unstable features are not enabled

meiravgri:
Ensure that the SHARD_K_RATIO feature is disabled if the ENABLE_UNSTABLE_FEATURES flag is not set. Add a test to verify this behavior.

### DIFF
--- a/src/shard_window_ratio.h
+++ b/src/shard_window_ratio.h
@@ -42,7 +42,7 @@ static inline size_t calculateEffectiveK(size_t originalK, double ratio, size_t 
   // We should not get here if numShards == 1
   RS_LOG_ASSERT(numShards > 1, "Should not calculate effective K for single shard");
 
-  if (ratio == MAX_SHARD_WINDOW_RATIO) {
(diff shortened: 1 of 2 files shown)


src/shard_window_ratio.h
```diff
@@ -42,7 +42,7 @@ static inline size_t calculateEffectiveK(size_t originalK, double ratio, size_t
   // We should not get here if numShards == 1
   RS_LOG_ASSERT(numShards > 1, "Should not calculate effective K for single shard");
 
-  if (ratio == MAX_SHARD_WINDOW_RATIO) {
+  if (ratio == MAX_SHARD_WINDOW_RATIO || !(RSGlobalConfig.enableUnstableFeatures)) {
```
alonre24, line 45: We should test the flag also sooner, upon parsing the query. `...=>[KNN ... ] => {$SHARD_K_RATIO: X}` should return syntax error if `enableUnstableFeatures` is false

PR #6537, opened 2025-07-24
Refactor "on reopen" logic - implement Revalidate for the internal iterators - [MOD-10534]

GuyAv46:

Implement new `Revalidate` API for the union, intersection, optional, and not iterators.

This PR also aligns the behavior of the `Revalidate` API regarding some edge cases with the return code (see the enum in `iterator_api.h` for the full details)

#### Mark if applicable

### DIFF
--- a/src/iterators/intersection_iterator.c
+++ b/src/iterators/intersection_iterator.c
@@ -378,7 +378,6 @@ QueryIterator *NewIntersectionIterator(QueryIterator **its, size_t num, int max_
   ret->atEOF = false;
   ret->lastDocId = 0;
   ret->current = NewIntersectResult(num, weight);
-  ret->isAborted = false;
   ret->NumEstimated = II_NumEstimated;
   if (max_slop < 0 && !in_order) {
     // No slop and no order means every result is relevant, so we can use the fast path
--- a/src/iterators/intersection_iterator.c
+++ b/src/iterators/intersection_iterator.c
@@ -308,6 +308,48 @@ static QueryIterator *IntersectionIteratorReducer(QueryIterator **its, size_t *n
   return ret;
 }
 
+static ValidateStatus II_Revalidate(QueryIterator *base) {
+  IntersectionIterator *ii = (IntersectionIterator *)base;
+  bool any_child_moved = false;
+  t_docId max_child_docId = 0;
+
+  // Step 1: Revalidate all children and track status
+  for (uint32_t i = 0; i < ii->num_its; i++) {
+    QueryIterator *child = ii->its[i];
+    ValidateStatus child_status = child->Revalidate(child);
+
+    if (child_status == VALIDATE_ABORTED) {
+      return VALIDATE_ABORTED; // Intersection fails if any child fails
+    }
+
+    if (child_status == VALIDATE_MOVED) {
+      any_child_moved = true;
+      // Track the maximum docId among moved children
+      if (child->lastDocId > max_child_docId) {
+        max_child_docId = child->lastDocId;
+      }
+    }
+  }
+
+  // Step 2: Handle the result based on child status
+  if (!any_child_moved) {
+    // All children returned OK - simply return OK
+    return VALIDATE_OK;
+  }
+
+  // Step 3: At least one child moved - need to find new intersection position
+  // Skip to the maximal docId among moved children
+  IteratorStatus skip_rc = base->SkipTo(base, max_child_docId);
+
+  // Step 4: Return based on skip result
+  if (skip_rc == ITERATOR_OK || skip_rc == ITERATOR_NOTFOUND) {
+    return VALIDATE_MOVED;
+  } else {
+    // Skip failed (EOF, TIMEOUT, etc.) - return OK (iterator still valid at original position)
+    return VALIDATE_OK;
+  }
+}
+
 QueryIterator *NewIntersectionIterator(QueryIterator **its, size_t num, int max_slop, bool in_order, double weight) {
   RS_ASSERT(its && num > 0);
   QueryIterator *ret = IntersectionIteratorReducer(its, &num);
@@ -347,6 +389,7 @@ QueryIterator *NewIntersectionIterator(QueryIterator **its, size_t num, int max_
   }
   ret->Free = II_Free;
   ret->Rewind = II_Rewind;
+  ret->Revalidate = II_Revalidate;
 
   return ret;
 }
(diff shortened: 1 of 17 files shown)


src/iterators/not_iterator.c
```diff
@@ -226,6 +226,74 @@ static IteratorStatus NI_SkipTo_Optimized(QueryIterator *base, t_docId docId) {
   return rc;
 }
 
+// Revalidate for NOT iterator - Non-optimized version.
+static ValidateStatus NI_Revalidate_NotOptimized(QueryIterator *base) {
+  NotIterator *ni = (NotIterator *)base;
+
+  // 1. Revalidate the child iterator
+  ValidateStatus child_status = ni->child->Revalidate(ni->child);
```
JoanFM, line 234: so if VALIDATE_MOVED from child we do not care?

src/iterators/not_iterator.c
```diff
@@ -226,6 +226,74 @@ static IteratorStatus NI_SkipTo_Optimized(QueryIterator *base, t_docId docId) {
...
+    // When child is aborted, NOT iterator becomes "NOT nothing" = everything
+    ni->child->Free(ni->child);
+    ni->child = IT_V2(NewEmptyIterator)();
+  }
+
+  // 3. Branch on wildcard status
+  if (wcii_status == VALIDATE_OK) {
+    // Same logic as non-optimized case
+    RS_LOG_ASSERT(child_status != VALIDATE_MOVED || ni->child->lastDocId > base->lastDocId, "Moved but still not beyond lastDocId");
+    base->atEOF = ni->wcii->atEOF; // sync EOF state with wildcard iterator
```
JoanFM: is it needed? if it was not on EOF before, can it change with a VALIDATE_OK result?

src/iterators/not_iterator.c
GuyAv46: With the Not iterator, the child should always be ahead of the internal counter/wildcard, as it has to mark the next invalid result. The only exceptions are initially when they are both on id 0, and when the child is at EOF (and then the counter/wildcard will get ahead of the child).
If we got `VALIDATE_MOVED`, it means the child is **at a new valid result**, so it's not any of the exceptions. If it was ahead/at the same ID as the interlan wildcard, it's even "more ahead" now. This should not affect the current result

JoanFM:
Is the `isAborted`flag actually used? Seems is only written and never read. Maybe we can remove it?

src/iterators/not_iterator.c
GuyAv46: `VALIDATE_OK` may indicate EOF, so yeah.
I may need to add `VALIDATE_EOF` if we cannot handle EOF under `VALIDATE_OK`. But `VALIDATE_MOVED` has to mean `current` is a **new** and **valid** result.

src/iterators/union_iterator.c
```diff
@@ -420,6 +420,63 @@ static QueryIterator *UnionIteratorReducer(QueryIterator **its, int *num, bool q
...
+  uint32_t write_idx = 0;
+  for (uint32_t read_idx = 0; read_idx < ui->num_orig; read_idx++) {
+    QueryIterator *child = ui->its_orig[read_idx];
+    RS_ASSERT(child != NULL);
+
+    ValidateStatus child_status = child->Revalidate(child);
+
+    if (child_status == VALIDATE_ABORTED) {
+      // Free the aborted child
+      child->Free(child);
```
JoanFM, line 451: set here all_child_ok = false;

src/iterators/union_iterator.c
```diff
@@ -420,6 +420,63 @@ static QueryIterator *UnionIteratorReducer(QueryIterator **its, int *num, bool q
...
+    } else {
+      // Keep this child - copy it to write position if needed
+      if (write_idx != read_idx) {
+        ui->its_orig[write_idx] = child;
+      }
+      write_idx++;
+      all_child_ok = all_child_ok && (child_status == VALIDATE_OK);
+    }
+  }
+  if (write_idx == ui->num_orig && all_child_ok) {
```
JoanFM: and here u can simply check all_child_ok

src/iterators/union_iterator.c
```diff
@@ -420,6 +431,60 @@ static QueryIterator *UnionIteratorReducer(QueryIterator **its, int *num, bool q
...
+    return VALIDATE_ABORTED;
+  }
+
+  // Use UI_SyncIterList to update `its` array and heap - remove exhausted iterators
+  UI_SyncIterList(ui);
+
+  // Update current result - reset and rebuild if we have active children
+  AggregateResult_Reset(ui->base.current);
+  // Find the minimum docId among active children to update current result
+  t_docId minId = UINT64_MAX;
```
JoanFM, line 474: just a note in general. Would be nice to have a MAX_DOC_ID defined

JoanFM (approved):

JoanFM:
LGTM as per logic and tests.

Only detail is I would remove isAborted flag as not used

JoanFM (dismissed):

src/iterators/not_iterator.c
GuyAv46: Correction after PR was changed:
`VALIDATE_OK` means the iterator was not changed, and if it wasn't EOF before the revalidation, `current` is still valid and has the same docID. Specifically, the EOF status is not expected to change if `VALIDATE_OK` is returned

src/iterators/intersection_iterator.c
```diff
@@ -308,6 +308,50 @@ static QueryIterator *IntersectionIteratorReducer(QueryIterator **its, size_t *n
...
+        max_child_docId = child->lastDocId;
+      }
+    }
+  }
+
+  // Step 2: Handle the result based on child status
+  if (!any_child_moved) {
+    // All children returned OK - simply return OK
+    return VALIDATE_OK;
+  } else if (base->atEOF) {
```
JoanFM, line 340: If it is already at EOF, is it needed to do all the previous revalidation on all the children?

JoanFM (approved):

src/iterators/intersection_iterator.c
GuyAv46: We should revalidate in case we get rewound. I think we should take the safe approach for now, but if we ever see it will be beneficial to revalidate only if needed, we can add a ”needs revalidation” flag and revalidate on rewind

JoanFM (dismissed):

JoanFM (dismissed):

PR #6561, opened 2025-07-29
ci: lint: check git diff

gdesmott:
Should prevent commits where the main code has been updated but the generated code derivated from it was not committed.

This happened multiple times while changing the Rust code while the generated C bindings were not updated.

### DIFF
--- a/.github/workflows/task-lint.yml
+++ b/.github/workflows/task-lint.yml
@@ -25,6 +25,13 @@ jobs:
         id: lint
         continue-on-error: true
         run: make lint
+
+      - name: Diff
+        id: git-diff
+        continue-on-error: true
+        # error if there is uncommitted changes as it's likely generated code which has not been committed
+        run: git diff --exit-code
+
       - name: License header
         id: license-header
         continue-on-error: true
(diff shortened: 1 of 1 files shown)


gdesmott:
Here is how it fails: https://github.com/RediSearch/RediSearch/actions/runs/16593337232/job/46934123255?pr=6561

DarthB:
> Here is how it fails: https://github.com/RediSearch/RediSearch/actions/runs/16593337232/job/46934123255?pr=6561

What do you think about adding an additional step and output in the "merge" step? Or are there situations when the solution differs from rerunning the build locally and committing?

gdesmott:
> What do you think about adding an additional step and output in the "merge" step? Or are there situations when the solution differs from rerunning the build locally and committing?

I'm not sure where is the best place to put this. 

I set it after the clippy run as we need the Rust code to be fully built before checking and figured out it was not worth re-doing another build just to check this.
But I'm open to better suggestions.

.github/workflows/task-lint.yml
```diff
@@ -24,7 +24,8 @@ jobs:
       - name: Lint
         id: lint
         continue-on-error: true
-        run: make lint
+        # error if there is uncommitted changes as it's likely generated code which has not been committed
+        run: make lint && git diff --exit-code
```
chesedo: I would just suggest making this two steps

gdesmott: I was assuming that each step was independent but they're not. So indeed it's better this way.

chesedo (approved):

PR #6571, opened 2025-07-30
[MOD-10530] Redesign Decoder API to allow borrowing from the II

gdesmott:
We need to change the Decoder trait so records can borrow from the inverted index.

This is required to prevent extra copies when decoding the offsets vector.

#### Mark if applicable


chesedo (dismissed):
Nice and simple! :partying_face:

GuyAv46 (dismissed):

chesedo (approved):

PR #6575, opened 2025-07-31
rust iterator trait

DvirDukhan:

This PR introduces the rqe iterator trait


src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,87 @@
...
+
+use inverted_index::RSIndexResult;
+
+/**
+ * An iterator has two successful states: `OK` and `NotFound`.
+ * `OK` indicates that the iterator has found a valid result, in either "read" or "skip_to",
+ * `NotFound` indicates that a "skip_to(doc_id)" operation has found a result with a doc_id greater than the requested `doc_id`.
+ */
+pub enum RQEIteratorStatus {
+    OK,
```
DvirDukhan: should it be something like 
```suggestion
    OK(&RSIndexResult)
```
so we can get the result directly from the call, without the need to call `current`?

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,87 @@
...
+pub trait RQEIterator {
+    /**
+     * Read the next entry from the iterator.
+     *  On a successful read, the iterator must:
+     *  1. Set its `lastDocId` member to the new current result id
+     *  2. Set its `current` index result property to its current result, for the caller to access if desired
+     *  @returns Ok(RQEIteratorStatus::OK)
+     *  @returns Err(RQEIteratorError) for any error.
+     */
+    fn read(&mut self) -> Result<RQEIteratorStatus, RQEIteratorError>;
```
DvirDukhan: do we want to use `Result` or keep C style?

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Returns the last doc id that was read or skipped to.
+     */
+    fn last_doc_id(&self) -> u64;
+
+    /**
+     * Returns true if the iterator has more results to read, meaning, not at EOF.
+     */
+    fn has_next(&self) -> bool;
```
DvirDukhan, line 73: changed `atEOF` to `has_next`

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,87 @@
...
+    NotFound,
+}
+
+pub enum RQEIteratorError {
+    EOF,
+    TimedOut,
+}
+
+pub enum RQEValidateStatus {
+    Moved, // The iterator is still valid but lastDocID changed, and `current` is a new valid result or at EOF. If not at EOF, the `current` result should be used before the next read, or it will be overwritten.
```
JoanFM: we would also need the OK result

DvirDukhan: why? if we use `Result` you get `OK` natively

JoanFM: ah Ok makes sense

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
chesedo: Yes, that will work better. Then you can remove the `current()` method too.

I will want to remove the borrow here though

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
chesedo: Probably because the iterators can error internally. Eg. EOF error

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
DvirDukhan: why to move instead of borrow?

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
chesedo: Nice!!

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
chesedo: Borrowing makes code complicated because it introduces lifetimes which will end up going everywhere. So it is easier to start without it and then add it when you find you will have to use it.

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,76 @@
...
+use ffi::t_docId;
+use inverted_index::RSIndexResult;
+
+/// An iterator has two successful states: `OK` and `NotFound`.
+/// `OK` indicates that the iterator has found a valid result, in either "read" or "skip_to",
+/// `NotFound` indicates that a "skip_to(doc_id)" operation has found a result with a doc_id greater than the requested `doc_id`.
+pub enum RQEIteratorStatus {
+    OK(RSIndexResult),
+    NotFound(RSIndexResult),
+}
```
chesedo:
```suggestion
/// The result of an iterator read or skipping operation
pub enum RQEIteratorStatus {
    /// The iterator has found a valid result, in either `read()` or `skip_to()`
    OK(RSIndexResult),
    
    /// A `skip_to()` has found a result with a `doc_id` greater than the requested `doc_id`
    NotFound(RSIndexResult),
}
```
Each variant can have its documentation associated with it in Rust. This means you don't have to look at the enum in the IDE to know what each variant does / is for

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,76 @@
...
+    Aborted,
+}
+
+pub trait RQEIterator {
+    /// Read the next entry from the iterator.
+    /// On a successful read, the iterator must:
+    /// 1. Set its `lastDocId` member to the new current result id
+    /// 2. Set its `current` index result property to its current result, for the caller to access if desired
+    /// returns Ok(RQEIteratorStatus::OK) with a reference to the current result or Err(RQEIteratorError) for any error.
+    fn read(&mut self) -> Result<RQEIteratorStatus, RQEIteratorError>;
```
LukeMathWalker: If my understanding of the docs is correct, this will never return `RQEIteratorStatus::NotFound`, won't it?

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    /// Skip to the next ID of the iterator, which is greater or equal to `docId`.
+    /// It is assumed that when `skip_to` is called, `self.lastDocId() < docId`.
+    /// On a successful read, the iterator must:
+    /// 1. Set its `lastDocId` member to the new current result id
+    /// 2. Set its `current` index result property to its current result, for the caller to access if desired
+    /// returns Ok(RQEIteratorStatus::OK) if the iterator has found `docId`.
+    /// returns Ok(RQEIteratorStatus::NotFound) if the iterator has only found a result greater than `docId`.
+    /// returns Err(RQEIteratorError) for any error.
+    fn skip_to(&mut self, doc_id: t_docId) -> Result<RQEIteratorStatus, RQEIteratorError>;
```
chesedo:
```suggestion
    /// Skip to the next record in the iterator with an ID greater or equal to the given `docId`.
    ///
    /// It is assumed that when `skip_to` is called, `self.lastDocId() < docId`.
    ///
    /// On a successful read, the iterator must:
    /// 1. Set its `lastDocId` member to the new current result id
    /// 2. Set its `current` index result property to its current result, for the caller to access if desired
    ///
    /// Return `Ok(RQEIteratorStatus::OK)` if the iterator has found a record with the `docId` and `Ok(RQEIteratorStatus::NotFound)`
    /// if the iterator found a result greater than `docId`.
    fn skip_to(&mut self, doc_id: t_docId) -> Result<RQEIteratorStatus, RQEIteratorError>;
```
This is an example of how documentation is generally structured. There are three main changes in it
1. The first sentence is just a short description
2. The paragraphs after it contain more details with possible examples, edge cases, etc
3. The return reads as a sentence. Generally, you won't have a new sentence / point for each return type. The error is also implied and mostly redundant. The return comment is also mostly before any code examples that might be in the comment.

Here are some more [examples with the different elements in a comment](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components)

src/redisearch_rs/rqe_iterators/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "rqe_iterators"
+version.workspace = true
+edition.workspace = true
+license-file.workspace = true
+publish.workspace = true
+
+[dependencies]
+enumflags2.workspace = true
```
chesedo: This can be removed too for the time being

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
DvirDukhan: indeed

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,81 @@
...
+pub enum RQEIteratorStatus {
+    /// The iterator has found a valid result, in either `read()` or `skip_to()`
+    OK(RSIndexResult),
+    
+    /// A `skip_to()` has found a result with a `doc_id` greater than the requested `doc_id`
+    NotFound(RSIndexResult),
+}
+
+/// An iterator failure indications
+pub enum RQEIteratorError {
```
JoanFM, line 23: does it make sense to have EOF as an error? Isn't it a pretty regular return statement? Isn't it going to make the coding of inner iterators (Union, Intersection, etc ...) a little more complex?

DvirDukhan: so do you suggest moving it to the status? we remain with timeout => should we consider this as an error or not?

JoanFM: I think TimeOut an error makes sense IMO

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
LukeMathWalker: Then I would recommend a different return type: `Result<Option<RSIndexResult>, TimeoutError>` (or `Option<Result<RSIndexResult, TimeoutError>>`, slightly different semantics for time outs).
The `None` case encodes the EoF scenario, as it's idiomatic for Rust iterators.
You don't have a redundant variant in the success case, so the caller code doesn't have to "discard" it.

DvirDukhan: removed the EOF completely, `read` and `skip` now return an `Option`

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,79 @@
...
+use inverted_index::RSIndexResult;
+
+/// The result of an iterator read or skipping operation
+pub enum RQEIteratorStatus {
+    /// The iterator has found a valid result, in either `read()` or `skip_to()`
+    OK(RSIndexResult),
+
+    /// A `skip_to()` has found a result with a `doc_id` greater than the requested `doc_id`
+    NotFound(RSIndexResult),
+}
```
LukeMathWalker:
```suggestion
/// The outcome of [`RQEIterator::skip_to`].
pub enum SkipToOutcome {
    /// The iterator has a valid entry for the requested `doc_id`.
    Found(RSIndexResult),

    /// The iterator doesn't have an entry for the requested `doc_id`, but there are entries with an id greater than the requested one.
    NotFound(RSIndexResult),
}
```

Since we now use this enum exclusively for `skip_to`, we can make the docs more precise.

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,79 @@
...
+
+/// An iterator failure indications
+pub enum RQEIteratorError {
```
LukeMathWalker, line 23: Do you want to keep it as an enum for future extensibility? Since you're left with one variant, I'd convert it to a struct:

```rust
#[derive(Debug)]
struct TimedOut;
```

We should implement `Display` and the `Error` trait for it.

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// The iterator has reached the time limit for execution.
+    TimedOut,
+}
+
+/// The status of the iterator after a call to `revalidate`
+pub enum RQEValidateStatus {
+    /// The iterator is still valid and at the same position.
+    OK,
+    /// The iterator is still valid but lastDocID changed, and `current` is a new valid result or at EOF. If not at EOF, the `current` result should be used before the next read, or it will be overwritten.
```
LukeMathWalker: You no longer have `current` in this design, so this should be updated I believe.

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,79 @@
...
+/// An iterator failure indications
+pub enum RQEIteratorError {
+    /// The iterator has reached the time limit for execution.
+    TimedOut,
+}
+
+/// The status of the iterator after a call to `revalidate`
+pub enum RQEValidateStatus {
+    /// The iterator is still valid and at the same position.
+    OK,
```
LukeMathWalker:
```suggestion
    Ok,
```

Enum variant names should be camelcased.

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// This function returns Ok with the current result or Err(RQEIteratorError) for any error.
+    fn read(&mut self) -> Result<Option<RSIndexResult>, RQEIteratorError>;
+
+    /// Skip to the next record in the iterator with an ID greater or equal to the given `docId`.
+    ///
+    /// It is assumed that when `skip_to` is called, `self.lastDocId() < docId`.
+    ///
+    /// On a successful read, the iterator must:
+    /// 1. Set its `lastDocId` member to the new current result id
+    /// 2. Set its `current` index result property to its current result, for the caller to access if desired
```
LukeMathWalker: 2. is no longer needed, right?

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
JoanFM: I think not. We expect the next Read call to handle it. We do not return a value in this status

LukeMathWalker: That's what I mean—`current` doesn't exist anymore, so the documentation must be updated to reflect the way this is supposed to work in this design (i.e. call `read`).

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,79 @@
...
+    TimedOut,
+}
+
+/// The status of the iterator after a call to `revalidate`
+pub enum RQEValidateStatus {
+    /// The iterator is still valid and at the same position.
+    Ok,
+    /// The iterator is still valid but lastDocID changed, and `current` is a new valid result or at EOF. If not at EOF, the `current` result should be used before the next read, or it will be overwritten.
+    Moved,
+    /// The iterator is no longer valid, and should not be used or rewound. Should be freed.
```
chesedo: What does "Should be freed" refer to here?

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// Return `Ok(RQEIteratorStatus::OK)` if the iterator has found a record with the `docId` and `Ok(RQEIteratorStatus::NotFound)`
+    /// if the iterator found a result greater than `docId`. 'None" will be returned if the iterator has reached the end of the index.
+    fn skip_to(&mut self, doc_id: t_docId) -> Result<Option<RQEIteratorStatus>, RQEIteratorError>;
+
+    /// Called when the iterator is being revalidated after a concurrent index change.
+    ///
+    /// The iterator should check if it is still valid.
+    /// Return Ok if the iterator is still valid, Moved if the iterator is still valid, but the lastDocId has changed (moved forward) or Aborted if the iterator is no longer valid
+
+    fn revalidate(&mut self) -> RQEValidateStatus;
```
chesedo: All comments in general can benefit with an update

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
DvirDukhan: I want to keep it as enum for future extensibility. I think that in the future we should have a unified enum for the error, globally.

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
DvirDukhan: let's start by removing. If from some reason down the road we will need to return it, we will

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,75 @@
...
+    TimedOut,
+}
+
+/// The status of the iterator after a call to `revalidate`
+pub enum RQEValidateStatus {
+    /// The iterator is still valid and at the same position.
+    Ok,
+    /// The iterator is still valid but its internal state has changed.
+    Moved,
+    /// The iterator is no longer valid, and should not be used or rewound.
```
chesedo:
```suggestion
    /// The iterator is no longer valid, and should not be used or should be rewound.
```

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,75 @@
...
+    Aborted,
+}
+
+pub trait RQEIterator {
+    /// Read the next entry from the iterator.
+    ///
+    /// On a successful read, the iterator must set its `last_doc_id` property to the new current result id
+    /// This function returns Ok with the current result or Err(RQEIteratorError) for any error.
+    fn read(&mut self) -> Result<Option<RSIndexResult>, RQEIteratorError>;
```
chesedo: A return of `None` should also be documented

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///
+    /// Return `Ok(SkipToOutcome::Found)` if the iterator has found a record with the `docId` and `Ok(SkipToOutcome::NotFound)`
+    /// if the iterator found a result greater than `docId`. 'None" will be returned if the iterator has reached the end of the index.
+    fn skip_to(&mut self, doc_id: t_docId) -> Result<Option<SkipToOutcome>, RQEIteratorError>;
+
+    /// Called when the iterator is being revalidated after a concurrent index change.
+    ///
+    /// The iterator should check if it is still valid.
+    /// Return Ok if the iterator is still valid, Moved if the iterator is still valid, its internal state has changed or Aborted if the iterator is no longer valid
+
```
chesedo:
```suggestion
    /// Return Ok if the iterator is still valid; Moved if the iterator is still valid, but its internal state has changed; or Aborted if the iterator is no longer valid
```
Technically you can also remove this line since the same info is on the enum variants

PR #6576, opened 2025-07-31
[MOD-10377] Implement the encodeFull and encodeFullWide encoder/decoder in Rust

gdesmott:
Implement the full encoder and decoder in Rust.

## Benchmarks results

Performances of the Rust code is either better or about the same as the C version, except for non wide decoding which is slightly slower but not by much.

```
Encode - Full/C         time:   [31.873 µs 38.856 µs 44.635 µs]
Encode - Full/Rust      time:   [10.524 µs 14.999 µs 19.694 µs]

Decode - Full/C         time:   [21.062 ns 21.107 ns 21.152 ns]
Decode - Full/Rust      time:   [25.832 ns 25.875 ns 25.921 ns]

Encode - Full Wide/C    time:   [13.949 µs 14.042 µs 14.144 µs]
Encode - Full Wide/Rust time:   [14.237 µs 14.319 µs 14.412 µs]

Decode - Full Wide/C    time:   [58.386 ns 58.560 ns 58.781 ns]
Decode - Full Wide/Rust time:   [40.989 ns 41.049 ns 41.118 ns]
``` 

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/inverted_index_bencher/Cargo.toml
+++ b/src/redisearch_rs/inverted_index_bencher/Cargo.toml
@@ -28,8 +28,6 @@ inverted_index.workspace = true
 itertools.workspace = true
 redis_mock.workspace = true
 varint_ffi = { path = "../c_entrypoint/varint_ffi" }
-
-[dev-dependencies]
 types_ffi = { path = "../c_entrypoint/types_ffi" }
 
 [target.'cfg(all(target_env="musl", target_os="linux"))'.dependencies.redis-module]
--- a/src/redisearch_rs/inverted_index_bencher/benches/encoding_decoding.rs
+++ b/src/redisearch_rs/inverted_index_bencher/benches/encoding_decoding.rs
@@ -52,6 +52,16 @@ fn benchmark_doc_ids_only(c: &mut Criterion) {
     bencher.decoding(c);
 }
 
+fn benchmark_full(c: &mut Criterion) {
+    let bencher = benchers::full::Bencher::default();
+    bencher.encoding(c);
+    bencher.decoding(c);
+
+    let bencher = benchers::full::Bencher::wide();
+    bencher.encoding(c);
+    bencher.decoding(c);
+}
+
 criterion_group!(
     benches,
     benchmark_numeric,
@@ -59,5 +69,7 @@ criterion_group!(
     benchmark_freqs_fields,
     benchmark_fields_only,
     benchmark_doc_ids_only,
+    benchmark_full,
 );
+
 criterion_main!(benches);
--- a/src/redisearch_rs/inverted_index_bencher/src/benchers/full.rs
+++ b/src/redisearch_rs/inverted_index_bencher/src/benchers/full.rs
@@ -1,0 +1,244 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use std::{io::Cursor, ptr::NonNull, time::Duration, vec};
+
+use buffer::Buffer;
+use criterion::{
+    BatchSize, BenchmarkGroup, Criterion, black_box,
+    measurement::{Measurement, WallTime},
+};
+use ffi::t_fieldMask;
+use inverted_index::{
+    Decoder, Encoder,
+    full::{Full, FullWide},
+    test_utils::TestTermRecord,
+};
+use itertools::Itertools;
+
+use crate::ffi::{TestBuffer, encode_full, read_freq_offsets_flags};
+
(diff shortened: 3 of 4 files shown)


src/redisearch_rs/inverted_index/src/full.rs
```diff
@@ -0,0 +1,194 @@
...
+///
+/// record must have `result_type` set to `RSResultType::Term`.
+#[inline(always)]
+fn offsets<'a>(record: &'a RSIndexResult<'_>) -> &'a [u8] {
+    // SAFETY: caller ensured the proper result_type.
+    let term = record.as_term().unwrap();
+    if term.offsets.data.is_null() {
+        &[]
+    } else {
+        assert_eq!(term.offsets.len, record.offsets_sz);
```
chesedo, line 46: Does this mean `offsets_sz` is just a redundant value of the offset length?

src/redisearch_rs/inverted_index/src/full.rs
```diff
@@ -0,0 +1,194 @@
...
+/// Create a [`RSIndexResult`] from the given parameters and read its offsets from the reader.
+#[inline(always)]
+fn decode_term_record_offsets<'a>(
+    cursor: &mut Cursor<&'a [u8]>,
+    base: t_docId,
+    delta: u32,
+    field_mask: t_fieldMask,
+    freq: u32,
+    offsets_sz: u32,
+) -> std::io::Result<RSIndexResult<'a>> {
```
chesedo, line 92: Is there a performance hit when the `inline` is missing?

src/redisearch_rs/inverted_index_bencher/Cargo.toml
```diff
@@ -28,6 +28,7 @@ inverted_index.workspace = true
 itertools.workspace = true
 redis_mock.workspace = true
 varint_ffi = { path = "../c_entrypoint/varint_ffi" }
+types_ffi = { path = "../c_entrypoint/types_ffi" }
```
chesedo, line 31: Why is this needed?

src/redisearch_rs/inverted_index/src/full.rs
gdesmott: Yes, see https://github.com/RediSearch/RediSearch/blob/master/src/redisearch_rs/inverted_index/src/lib.rs#L392

src/redisearch_rs/inverted_index_bencher/Cargo.toml
gdesmott: Because of https://github.com/RediSearch/RediSearch/blob/10377-full/src/redisearch_rs/inverted_index_bencher/src/benchers/full.rs#L27

We need to ensure this symbol is present.

src/redisearch_rs/inverted_index/src/full.rs
gdesmott: The Rust decode bench is about 18% slower, but still faster than the C one, so maybe that's ok?

src/redisearch_rs/inverted_index/src/full.rs
chesedo: Okay, we should remember this when switching to Rust enums

src/redisearch_rs/inverted_index/src/full.rs
chesedo: Yes, I was more interested in whether the `inline` might be redundant. The code base will look weird if we end up having it on each function

src/redisearch_rs/inverted_index_bencher/Cargo.toml
chesedo: It's a bit of an unfortunate dependency but okay

chesedo (approved):

src/redisearch_rs/inverted_index/src/full.rs
gdesmott: I removed it.

gdesmott: The performance hit seems to be bigger with the other offsets decoder using this function, so I'd rather keep it if that's ok with you?

src/redisearch_rs/inverted_index/src/test_utils.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        // SAFETY: we asserted the type above
+        let a_term_record = self.0.as_term().unwrap();
+        // SAFETY: we checked that other has the same type as self
+        let b_term_record = other.0.as_term().unwrap();
+
+        // SAFETY: `len` is guaranteed to be a valid length for the data pointer.
+        let a_offsets = unsafe {
+            std::slice::from_raw_parts(
+                a_term_record.offsets.data as *const i8,
```
GuyAv46, line 88: Why `i8` and not `u8`?

GuyAv46 (approved):

src/redisearch_rs/inverted_index/src/test_utils.rs
gdesmott: Because [`c_char`](https://doc.rust-lang.org/std/ffi/type.c_char.html) is actually  a `i8`. 
It does not really matter as we just handle it as a opaque 8 bits type.

PR #6577, opened 2025-07-31
remove default-members from workspace

gdesmott:
Instead explicitly exclude requiring C code.
    
This should prevent CI not running on new crates as we no longer have to
maintain the default members list.

Also fix clippy warnings in crates that were not checked by the CI.

#### Mark if applicable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -279,8 +279,8 @@ run:
 
 lint:
 	@echo "Running linters..."
-	@cd $(ROOT)/src/redisearch_rs && cargo clippy -- -D warnings
-	@cd $(ROOT)/src/redisearch_rs && RUSTDOCFLAGS="-Dwarnings" cargo doc
+	@cd $(ROOT)/src/redisearch_rs && cargo clippy --workspace --exclude ffi --exclude inverted_index_bencher --exclude trie_bencher --exclude varint_bencher -- -D warnings
+	@cd $(ROOT)/src/redisearch_rs && RUSTDOCFLAGS="-Dwarnings" cargo doc --workspace --exclude ffi --exclude inverted_index_bencher --exclude trie_bencher --exclude varint_bencher
 
 fmt:
 ifeq ($(CHECK),1)
--- a/Makefile
+++ b/Makefile
@@ -277,10 +277,15 @@ run:
 		fi; \
 	fi
 
+# Function to extract RUST_EXCLUDE_CRATES from build.sh
+define get_rust_exclude_crates
+$(shell grep "RUST_EXCLUDE_CRATES=" build.sh | cut -d'=' -f2 | tr -d '"')
+endef
+
 lint:
 	@echo "Running linters..."
-	@cd $(ROOT)/src/redisearch_rs && cargo clippy --workspace --exclude ffi --exclude inverted_index_bencher --exclude trie_bencher --exclude varint_bencher -- -D warnings
-	@cd $(ROOT)/src/redisearch_rs && RUSTDOCFLAGS="-Dwarnings" cargo doc --workspace --exclude ffi --exclude inverted_index_bencher --exclude trie_bencher --exclude varint_bencher
+	@cd $(ROOT)/src/redisearch_rs && cargo clippy --workspace $(call get_rust_exclude_crates) -- -D warnings
+	@cd $(ROOT)/src/redisearch_rs && RUSTDOCFLAGS="-Dwarnings" cargo doc --workspace $(call get_rust_exclude_crates)
 
 fmt:
 ifeq ($(CHECK),1)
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -279,8 +279,8 @@ run:
 
 lint:
 	@echo "Running linters..."
-	@cd $(ROOT)/src/redisearch_rs && cargo clippy -- -D warnings
-	@cd $(ROOT)/src/redisearch_rs && RUSTDOCFLAGS="-Dwarnings" cargo doc
+	@cd $(ROOT)/src/redisearch_rs && cargo clippy --workspace --exclude ffi --exclude inverted_index_bencher --exclude trie_bencher --exclude varint_bencher -- -D warnings
+	@cd $(ROOT)/src/redisearch_rs && RUSTDOCFLAGS="-Dwarnings" cargo doc --workspace --exclude ffi --exclude inverted_index_bencher --exclude trie_bencher --exclude varint_bencher
```
LukeMathWalker: Can we centralise the list in one place?

build.sh
```diff
@@ -38,6 +38,10 @@ RUST_PROFILE=""  # Which profile should be used to build/test Rust code
 RUN_MIRI=0       # Run Rust tests through miri to catch undefined behavior
 RUST_DENY_WARNS=0 # Deny all Rust compiler warnings
 
+# Rust code is built first, so exclude crates linking on C code as the internal lib is not built yet.
```
LukeMathWalker:
```suggestion
# Rust code is built first, so exclude crates that link C code, since the static libraries they depend on haven't been built yet.
```

build.sh
```diff
@@ -38,6 +38,10 @@ RUST_PROFILE=""  # Which profile should be used to build/test Rust code
...
+# Keep the exclude list synced with the clippy and rustdoc exclude lists in Makefile.
+RUST_EXCLUDE_CRATES="--exclude ffi --exclude inverted_index_bencher --exclude trie_bencher --exclude varint_bencher"
```
LukeMathWalker: Do we need to exclude `ffi`?

build.sh
```diff
@@ -560,7 +564,7 @@ run_rust_tests() {
 
   # Run cargo test with the appropriate filter
   cd "$RUST_DIR"
-  RUSTFLAGS="${RUSTFLAGS:--D warnings}" cargo $RUST_EXTENSIONS test --profile=$RUST_PROFILE $RUST_TEST_OPTIONS $TEST_FILTER -- --nocapture
+  RUSTFLAGS="${RUSTFLAGS:--D warnings}" cargo $RUST_EXTENSIONS test $RUST_EXCLUDE_CRATES --profile=$RUST_PROFILE $RUST_TEST_OPTIONS $TEST_FILTER -- --nocapture
```
LukeMathWalker: We should test everything. By the time we are testing, everything has been built.

build.sh
gdesmott: It was not in the default members but indeed we don't have to exclude it.

build.sh
```diff
@@ -40,7 +40,7 @@ RUST_DENY_WARNS=0 # Deny all Rust compiler warnings
 
 # Rust code is built first, so exclude crates that link C code, since the static libraries they depend on haven't been built yet.
 # Keep the exclude list synced with the clippy and rustdoc exclude lists in Makefile.
-RUST_EXCLUDE_CRATES="--exclude ffi --exclude inverted_index_bencher --exclude trie_bencher --exclude varint_bencher"
+RUST_EXCLUDE_CRATES="--exclude inverted_index_bencher --exclude trie_bencher --exclude varint_bencher"
```
LukeMathWalker: Sorry for doing a double-pass 🙈 But I don't think `trie_bencher` depends on C code anymore. Neither does `varint_bencher`.

build.sh
LukeMathWalker: Although it doesn't really matter since they are all benchmarking crates in the end.

Makefile
gdesmott: A bit ugly because of Makefile but I found a way to parse it from `build.sh`.

build.sh
gdesmott: CI was failing if those were not excluded actually: https://github.com/RediSearch/RediSearch/actions/runs/16648289545/job/47114160791

build.sh
gdesmott: Indeed that seems to work. Let's see if CI passes.

build.sh
LukeMathWalker: Ah, I see, we don't replicate the target selection logic to pick up the static library from the appropriate folder. Fine to exclude them, there are no tests there anyway.

gdesmott:
@LukeMathWalker I fixed the CI, should we merge this then?

PR #6578, opened 2025-07-31
[MOD-10677] Add lifetime to RSOffsetVector

gdesmott:
Splitting out some commits from https://github.com/RediSearch/RediSearch/pull/6576 to reduce PR sizes.

### DIFF
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -147,13 +147,18 @@ pub struct RSTermRecord<'a> {
     pub offsets: RSOffsetVector<'a>,
 }
 
-impl RSTermRecord<'_> {
+impl<'a> RSTermRecord<'a> {
     /// Create a new term record without term pointer and offsets.
     pub fn new() -> Self {
         Self {
             term: ptr::null_mut(),
             offsets: RSOffsetVector::empty(),
         }
+    }
+
+    /// Create a new term with the given term pointer and offsets.
+    pub fn with_term(term: *mut RSQueryTerm, offsets: RSOffsetVector<'a>) -> RSTermRecord<'a> {
+        Self { term, offsets }
     }
 }
 
@@ -435,7 +440,7 @@ impl Default for RSIndexResult<'_> {
     }
 }
 
-impl RSIndexResult<'_> {
+impl<'a> RSIndexResult<'a> {
     /// Create a new virtual index result
     pub fn virt() -> Self {
         Self {
@@ -522,6 +527,21 @@ impl RSIndexResult<'_> {
                 term: ManuallyDrop::new(RSTermRecord::new()),
             },
             result_type: RSResultType::Term,
+            ..Default::default()
+        }
+    }
+
+    pub fn term_with_term_ptr(
+        term: *mut RSQueryTerm,
+        offsets: RSOffsetVector<'a>,
+    ) -> RSIndexResult<'a> {
+        let offsets_sz = offsets.len;
+        Self {
+            data: RSIndexResultData {
+                term: ManuallyDrop::new(RSTermRecord::with_term(term, offsets)),
+            },
+            result_type: RSResultType::Term,
+            offsets_sz,
             ..Default::default()
         }
     }
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -148,7 +148,7 @@ pub struct RSTermRecord<'a> {
 }
 
 impl RSTermRecord<'_> {
-    /// Create a new term record with the given term pointer
+    /// Create a new term record without term pointer and offsets.
     pub fn new() -> Self {
         Self {
             term: ptr::null_mut(),
@@ -515,7 +515,7 @@ impl RSIndexResult<'_> {
         }
     }
 
-    /// Create a new term index result with the given term pointer
+    /// Create a new term index result.
     pub fn term() -> Self {
         Self {
             data: RSIndexResultData {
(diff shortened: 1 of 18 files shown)


src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -213,24 +234,28 @@ pub type RSResultTypeMask = BitFlags<RSResultType, u32>;
...
 #[derive(Debug, PartialEq)]
-pub struct RSAggregateResult {
+pub struct RSAggregateResult<'a> {
     /// The records making up this aggregate result
     ///
     /// The `RSAggregateResult` is part of a union in [`RSIndexResultData`], so it needs to have a
     /// known size. The std `Vec` won't have this since it is not `#[repr(C)]`, so we use our
     /// own `LowMemoryThinVec` type which is `#[repr(C)]` and has a known size instead.
-    records: LowMemoryThinVec<*const RSIndexResult>,
+    records: LowMemoryThinVec<*const RSIndexResult<'a>>,
```
chesedo, line 243: This is cool!! I would not have expected Rust to keep track of the lifetime of a raw pointer

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -506,6 +531,21 @@ impl RSIndexResult {
         }
     }
 
+    pub fn term_with_term_ptr(
+        term: *mut RSQueryTerm,
+        offsets: RSOffsetVector<'a>,
+    ) -> RSIndexResult<'a> {
```
chesedo, line 538: Missing public docs :smile:

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -801,7 +841,11 @@ pub trait Encoder {
 pub trait Decoder {
     /// Decode the next record from the reader. If any delta values are decoded, then they should
     /// add to the `base` document ID to get the actual document ID.
-    fn decode(&self, cursor: &mut Cursor<&[u8]>, base: t_docId) -> std::io::Result<RSIndexResult>;
+    fn decode(
+        &self,
+        cursor: &mut Cursor<&[u8]>,
+        base: t_docId,
+    ) -> std::io::Result<RSIndexResult<'_>>;
```
chesedo: Hmm, does this ellicit lifetime correctly bind to the slice pointer? :thinking: Won't it be on `&self`?

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -313,18 +313,21 @@ fn reading_records() {
         .expect("to be able to read from the buffer")
         .expect("to get a record");
     assert_eq!(record, RSIndexResult::virt().doc_id(10));
+    drop(record);
```
chesedo, line 316: Why are these needed?

src/redisearch_rs/inverted_index/src/lib.rs
gdesmott: Actually it does not, pointers don't have lifetimes, that's why I had to set this `PhantomData`  but the end result is the same.

src/redisearch_rs/inverted_index/src/lib.rs
gdesmott: Very good point. I think it's worth making it explicit, if only for clarity.

src/redisearch_rs/inverted_index/src/tests.rs
gdesmott: We need to drop to release the borrow.

```
error[E0499]: cannot borrow `ir` as mutable more than once at a time
   --> inverted_index/src/tests.rs:317:18
    |
311 |     let record = ir
    |                  -- first mutable borrow occurs here
...
317 |     let record = ir
    |                  ^^ second mutable borrow occurs here
...
335 | }
    | - first borrow might be used here, when `record` is dropped and runs the `Drop` code for type `RSIndexResult`
    ```

Comment (chesedo) [reply src/redisearch_rs/inverted_index/src/tests.rs:316]: Hmm, that feels wrong. Are you still getting this error after adding the explicit lifetime?

Comment (gdesmott) [reply src/redisearch_rs/inverted_index/src/tests.rs:316]: Yes, that does not help.

I think that's a limitation of the borrow checker not considering that the previous binding is drop before re-calling the method.

Comment (GuyAv46) [reply src/redisearch_rs/inverted_index/src/tests.rs:316]: Very strange. Feels like a pretty basic use case

File: src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs [hunk 5]
```diff
@@ -41,7 +41,7 @@ pub trait RQEIterator {
     /// On a successful read, the iterator must set its `last_doc_id` property to the new current result id
     /// This function returns Ok with the current result for valid results, or None if the iterator is depleted.
     /// The function will return Err(RQEIteratorError) for any error.
-    fn read(&mut self) -> Result<Option<RSIndexResult>, RQEIteratorError>;
+    fn read(&mut self) -> Result<Option<RSIndexResult<'_>>, RQEIteratorError>;
```
GuyAv46, line 44: Is this a placeholder? The lifetime should match `self`, right?

GuyAv46 (dismissed):

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
gdesmott: Yes, this is equivalent to: `fn read<'a>(&'a mut self) -> Result<Option<RSIndexResult<'a>>, RQEIteratorError>;` 

This kind of lifetime can be elided, see https://doc.rust-lang.org/nomicon/lifetime-elision.html

chesedo (approved):

PR #6579, opened 2025-07-31
MOD-10533: Merge SearchResults

Itzikvaknin:
Introduces `MergeSearchResults()` that combines multiple SearchResult objects from different sources into a single result with hybrid score

### DIFF
--- a/src/hybrid/hybrid_search_result.h
+++ b/src/hybrid/hybrid_search_result.h
@@ -67,8 +67,6 @@ double applyHybridScoring(HybridSearchResult *hybridResult, int8_t targetIndex, 
  */
 SearchResult* mergeSearchResults(HybridSearchResult *hybridResult, HybridScoringContext *scoringCtx);
 
-
-
 #ifdef __cplusplus
 }
 #endif


src/hybrid/hybrid_search_result.c
```diff
@@ -0,0 +1,164 @@
...
+#include <assert.h>
+
+/**
+ * Merge flags from source into target (in-place).
+ * Modifies target by ORing it with source flags.
+ */
+void MergeFlags(uint8_t *target_flags, const uint8_t *source_flags) {
+  if (!target_flags || !source_flags) {
+    return;
+  }
```
raz-mon: Assert

src/hybrid/hybrid_search_result.c
```diff
@@ -0,0 +1,164 @@
...
+
+  array_free(values);
+  return result;
+}
+
+/**
+ * Main function to merge SearchResults from multiple upstreams into a single comprehensive result.
+ * Finds the primary result, computes hybrid score, merges flags, and returns the merged result.
+ * This function transfers ownership of the primary result from the HybridSearchResult to the caller.
+ */
```
raz-mon: What's the definition of `primary result`?
Let's be clear on this, and why it matches our wanted behavior

src/hybrid/hybrid_search_result.c
```diff
@@ -0,0 +1,164 @@
...
+  int8_t targetIndex = -1;
+  for (size_t i = 0; i < hybridResult->numSources; i++) {
+    if (hybridResult->hasResults[i] && hybridResult->searchResults[i]) {
+      primary = hybridResult->searchResults[i];
+      targetIndex = (int8_t)i;
+      break;
+    }
+  }
+
+  if (!primary || targetIndex == -1) {
```
raz-mon: One condition is enough here..

src/hybrid/hybrid_search_result.c
```diff
@@ -0,0 +1,164 @@
...
+ */
+SearchResult* MergeSearchResults(HybridSearchResult *hybridResult, HybridScoringContext *scoringCtx) {
+  RS_ASSERT(hybridResult && scoringCtx);
+
+  // Find the primary result (first non-null result)
+  SearchResult *primary = NULL;
+  int8_t targetIndex = -1;
+  for (size_t i = 0; i < hybridResult->numSources; i++) {
+    if (hybridResult->hasResults[i] && hybridResult->searchResults[i]) {
+      primary = hybridResult->searchResults[i];
```
raz-mon, line 129: Let add an assertion for `primary`

src/hybrid/hybrid_search_result.c
```diff
@@ -0,0 +1,164 @@
...
+
+  if (!primary || targetIndex == -1) {
+    return NULL;
+  }
+
+  // Apply hybrid scoring to compute hybrid score and merge explanations
+  double hybridScore = ApplyHybridScoring(hybridResult, targetIndex, scoringCtx);
+
+  // Update primary result's score
+  primary->score = hybridScore;
```
raz-mon: This overriding matches our current wanted behavior that does not expose the scores of the upstream branches, but will soon be extended such that we will be able to save and further use the upstream branches' scores in the tail pipeline.
Consider adding a comment on that, for clarity.

src/hybrid/hybrid_search_result.c
```diff
@@ -0,0 +1,164 @@
...
+
+  // Merge flags from all upstreams
+  for (size_t i = 0; i < hybridResult->numSources; i++) {
+    if (hybridResult->hasResults[i] && hybridResult->searchResults[i] && i != targetIndex) {
```
raz-mon: Having both `hybridResult->hasResults[i] && hybridResult->searchResults[i]` seems redundant.
Consider adding assertions for the constraints you wish always to be true, and removing redundant conditions.

src/hybrid/hybrid_search_result.c
```diff
@@ -0,0 +1,164 @@
...
+double ApplyHybridScoring(HybridSearchResult *hybridResult, int8_t targetIndex, HybridScoringContext *scoringCtx) {
+  if (!hybridResult || !scoringCtx) {
+    return 0.0;
+  }
+  RS_ASSERT(hybridResult->hasResults[targetIndex]);
+
+  // Extract values from SearchResults
+  arrayof(double) values = array_newlen(double, hybridResult->numSources);
+  for (size_t i = 0; i < hybridResult->numSources; i++) {
+    if (hybridResult->hasResults[i] && hybridResult->searchResults[i]) {
```
raz-mon: Same comment about redundant conditions.
```suggestion
    if (hybridResult->hasResults[i]) {
      RS_ASSERT(hybridResult->searchResults[i]);
      ...
```

src/score_explain.c
```diff
@@ -57,4 +57,6 @@ void explain(RSScoreExplain *scrExp, char *fmt, ...) {
   va_end(ap);
 
   rm_free(tempStr);
-}
\ No newline at end of file
+}
+
```
raz-mon: One line

tests/ctests/test_hybrid_merge.c
```diff
@@ -0,0 +1,71 @@
...
+#include "rmutil/alloc.h"
+#include "test_util.h"
+#include "redisearch.h"
+#include "rlookup.h"
+#include "value.h"
+#include <math.h>
+#include <string.h>
+#include <stdio.h>
+#include <assert.h>
+
```
raz-mon: This file name is a bit weird w.r.t to the tests that are actually here.
Consider embedding these tests in [tests/cpptests/test_cpp_hybridmerger.cpp](https://github.com/RediSearch/RediSearch/pull/6579/files?show-viewed-files=true&file-filters%5B%5D=#diff-170964ee10f784ac04ffc49d257166ad091dbd320465b052d839223931f0bc4e)

raz-mon (commented):
Nice 👍 
See some comments

src/result_processor.c
```diff
@@ -1721,28 +1694,31 @@ dictType dictTypeHybridSearchResult = {
...
 
-   // Update the score for this upstream
-   hybridResult->scores[upstreamIndex] = score;
-   hybridResult->hasScores[upstreamIndex] = true;
+   r->score = score;
+   HybridSearchResult_StoreResult(hybridResult, r, upstreamIndex);
  }
 
+
+
```
ofiryanai, line 1722: remove new lines

src/hybrid/hybrid_search_result.c
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/**
+ * Store a SearchResult from a source into the HybridSearchResult.
+ * Updates the score of the SearchResult and marks the source as having results.
+ */
+void HybridSearchResult_StoreResult(HybridSearchResult* hybridResult, SearchResult* searchResult, int sourceIndex) {
+  if (!hybridResult || !searchResult ) {
+    return;
+  }
```
ofiryanai: assert instead, it's reasonable to assume they're not null in Release

src/hybrid/hybrid_search_result.c
```diff
@@ -0,0 +1,164 @@
...
+
+
+/**
+ * Apply hybrid scoring to compute combined score from multiple sources.
+ * Supports both RRF (with ranks) and Linear (with scores) hybrid scoring.
+ */
+double ApplyHybridScoring(HybridSearchResult *hybridResult, int8_t targetIndex, HybridScoringContext *scoringCtx) {
+  if (!hybridResult || !scoringCtx) {
+    return 0.0;
+  }
```
ofiryanai: RS_ASSERT so it will be trimmed in Release

src/hybrid/hybrid_search_result.c
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/**
+ * Constructor for HybridSearchResult.
+ * Allocates memory for storing SearchResults from numSources sources.
+ */
+HybridSearchResult* HybridSearchResult_New(size_t numSources) {
+  if (numSources == 0) {
+    return NULL;
+  }
```
ofiryanai: RS_ASSERT

ofiryanai (dismissed):
A bit too defensive guards in some of the function IMO. But bottom line looks good.

src/hybrid/hybrid_search_result.c
Itzikvaknin: Im not sure I understand what do you mean,
Added  `RS_ASSERT(primary && targetIndex != -1);`  right after the this loop

src/hybrid/hybrid_search_result.c
Itzikvaknin: This is the searchResult score, I believe you’re referring to the score field in `RLookupRow`
It is indeed written to that field, and this happens in `RPHybridMerger_Yield`:

```
// Add score as field if scoreKey is provided
if (self->scoreKey) {
    RLookup_WriteOwnKey(self->scoreKey, &r->rowdata, RS_NumVal(r->score));
}
```
I don’t think it should be discussed in this function

tests/cpptests/test_cpp_hybridmerger.cpp
```diff
@@ -1283,3 +1284,45 @@ TEST_F(HybridMergerTest, testHybridMergerRRFFlagMerging) {
...
+
+/*
+ * Test MergeFlags function with Result_ExpiredDoc flag
+ */
+TEST_F(HybridMergerTest, testMergeFlags_ExpiredDoc) {
+  uint8_t target_flags = 0;  // No flags initially
+  uint8_t source_flags = Result_ExpiredDoc;  // Source has expired flag
+
+  // Test merging expired flag
+  MergeFlags(&target_flags, &source_flags);
```
ofiryanai: my eyes are little hurting from the capital letters in utility function, please consider renaming to mergeFlags.. same for other methods in the file it's defined

Itzikvaknin: changed

ofiryanai (approved):

src/hybrid/hybrid_search_result.c
raz-mon: I mean an assertion for `primary` - `RS_ASSERT(primary)`
And removing the second condition in the `if` - that part was less clear

tests/ctests/test_hybrid_merge.c
Itzikvaknin: Moved to `test_cpp_hybridmerger.cpp`

src/hybrid/hybrid_search_result.c
Itzikvaknin: Changed to
```
/**
 * Main function to merge SearchResults from multiple upstreams into a single comprehensive result.
 *
 * PRIMARY RESULT SELECTION:
 * The "primary result" is the first non-null SearchResult found in index order (0, 1, 2...).
 * This prefers search results (index 0) over vector results (index 1) when both exist for RSIndexResult.
 *
 * The primary result is the SearchResult we merge into and return to the downstream processor.
 * This function transfers ownership of the primary result from the HybridSearchResult to the caller.
 */
```

src/hybrid/hybrid_search_result.h
```diff
@@ -0,0 +1,76 @@
...
+ * PRIMARY RESULT SELECTION:
+ * The "primary result" is the first non-null SearchResult found in index order (0, 1, 2...).
+ * This prefers search results (index 0) over vector results (index 1) when both exist for RSIndexResult.
+ *
+ * The primary result is the SearchResult we merge into and return to the downstream processor.
+ * This function transfers ownership of the primary result from the HybridSearchResult to the caller.
+ */
+SearchResult* mergeSearchResults(HybridSearchResult *hybridResult, HybridScoringContext *scoringCtx);
+
+
```
raz-mon: Remove newlines

src/hybrid/hybrid_search_result.c
raz-mon: I'm aware this is the score in the searchResult.
We may want to save the score not only via the RLookUp from which it will also be replied to the client.
Since we decided to expose it to the pipeline only when explicitly requested by the user we may not need that.

raz-mon (approved):

PR #6580, opened 2025-07-31
FFI adaptions for RLookupKey to work with CBindgen

DarthB:

1. Current: This PR resolves an issue where cbindgen could not generate the `RLookupKey` type for the C side due to its use of `Cow` and `Option` in fields.
2. Change:
  - Added Types for `Cow` and `Option` that are representable in C, i.e. `[repr(u8)]`
  - Added some renaming to `cbindgen.toml` to use `CStr` as `const char*`.
4. Outcome: cbindgen exports now work as expected.

#### Mark if applicable


LukeMathWalker:
Can't we simplify things by defining our own `Cow` with a well-known representation?

E.g.
```rust
#[repr(u8)]
pub enum Cow<'a, T> {
    Borrowed(&'a T),
    Owned(T),
}
```

DarthB:
But we would add another type as a workaround for `CBindgen`. Isn't the more minor change separating the structs?

I want to get rid of the `Cow` fields for two reasons:
1. `CBindgen` code generation does not work with generics.
2. The `Cow` fields are implementation details we don't want to communicate back to C.

LukeMathWalker:
> But we would add another type as a workaround for `CBindgen`. Isn't the more minor change separating the structs?
> 
> I want to get rid of the `Cow` fields for two reasons:
> 
> 1. `CBindgen` code generation does not work with generics.

This isn't an issue, your field has fully specified parameters (i.e. `cbindgen` knows how to represent `Cow<'a, CStr>`). I've tested it before making that comment.

> 2. The `Cow` fields are implementation details we don't want to communicate back to C.

True, but they are part of the type and I don't see why we should refactor the Rust type along somewhat artificial lines to have a "header" type in this case.

DarthB:
Another advantage of a type representable in C is that we don't have to fear `sizeof` and `memcpy`.
The tradeoff is that we need to reimplement types from the standard library partially. 

I didn't want to overwrite the history here yet:
https://github.com/RediSearch/RediSearch/compare/master...tj/alt2/ffi_rlookup

LukeMathWalker:
> Another advantage of a type representable in C is that we don't have to fear `sizeof` and `memcpy`. The tradeoff is that we need to reimplement types from the standard library partially.
> 
> I didn't want to overwrite the history here yet: [master...tj/alt2/ffi_rlookup](https://github.com/RediSearch/RediSearch/compare/master...tj/alt2/ffi_rlookup)

That looks pretty good to me—the moment we no longer have to do FFI, we delete the auxiliary types and replace everything with "stock" `Option`/`Cow`.

LukeMathWalker (approved):

src/redisearch_rs/headers/rlookup_rs.h
```diff
@@ -94,3 +97,144 @@ enum RLookupOption
...
+  Owned_CStr,
+};
+#ifndef __cplusplus
+typedef uint8_t CBCow_CStr_Tag;
+#endif // __cplusplus
+
+typedef union CBCow_CStr {
+  CBCow_CStr_Tag tag;
+  struct {
+    CBCow_CStr_Tag borrowed_tag;
```
raz-mon, line 131: Perhaps worth understanding why we have extra tags

raz-mon (approved):
Nice 👌 
Just one comment to check up

src/redisearch_rs/headers/rlookup_rs.h
DarthB: Every tag starts at offset 0.
Worth noting, `CStr` is actually a fat pointer. Addressed here: https://github.com/RediSearch/RediSearch/pull/6580/commits/1982ff59da9f35901e935414a3e536323e0c4525

raz-mon (approved):

PR #6583, opened 2025-08-03
MOD-10633 Add scorer and sorter by setting search flag for first subquery

ofiryanai:
Fix Hybrid flow that was missing scorer and sorter for the first subquery.

Now it is added by setting the flag of the search AREQ as if it's SEARCH rather than AGGREGATE.

The default Search command adds Scorer Sortby and Loader.
The first flag is to trigger this behavior. The second is to not create Loader, which is already added later "manually" in `HybridRequest_BuildPipeline`

### DIFF
--- a/src/aggregate/aggregate.h
+++ b/src/aggregate/aggregate.h
@@ -119,7 +119,7 @@ typedef struct {
 // Context structure for parseAggPlan to reduce parameter count
 typedef struct {
   AGGPlan *plan;                    // Aggregation plan
-  uint32_t reqflags;               // Request flags
+  uint32_t *reqflags;               // Request flags
   RequestConfig *reqConfig;         // Request configuration
   RSSearchOptions *searchopts;      // Search options
   size_t *prefixesOffset;           // Prefixes offset
--- a/src/aggregate/aggregate_request.c
+++ b/src/aggregate/aggregate_request.c
@@ -269,15 +269,15 @@ static int handleCommonArgs(ParseAggPlanContext *papCtx, ArgsCursor *ac, QueryEr
 
     if (arng->isLimited && arng->limit == 0) {
       // LIMIT 0 0 - only count
-      REQFLAGS_AddFlags(&papCtx->reqflags, QEXEC_F_NOROWS);
-      REQFLAGS_AddFlags(&papCtx->reqflags, QEXEC_F_SEND_NOFIELDS);
+      REQFLAGS_AddFlags(papCtx->reqflags, QEXEC_F_NOROWS);
+      REQFLAGS_AddFlags(papCtx->reqflags, QEXEC_F_SEND_NOFIELDS);
       // TODO: unify if when req holds only maxResults according to the query type.
       //(SEARCH / AGGREGATE)
-    } else if ((arng->limit > *papCtx->maxSearchResults) && (IsSearch(papCtx) || IsHybridTail(papCtx))) {
+    } else if ((arng->limit > *papCtx->maxSearchResults) && (*papCtx->reqflags & (QEXEC_F_IS_SEARCH | QEXEC_F_IS_HYBRID_SEARCH_SUBQUERY))) {
       QueryError_SetWithoutUserDataFmt(status, QUERY_ELIMIT, "LIMIT exceeds maximum of %llu",
                              *papCtx->maxSearchResults);
       return ARG_ERROR;
-    } else if ((arng->limit > *papCtx->maxAggregateResults) && !(IsSearch(papCtx) || IsHybridTail(p˚apCtx))) {
+    } else if ((arng->limit > *papCtx->maxAggregateResults) && !(*papCtx->reqflags & (QEXEC_F_IS_SEARCH | QEXEC_F_IS_HYBRID_SEARCH_SUBQUERY))) {
       QueryError_SetWithoutUserDataFmt(status, QUERY_ELIMIT, "LIMIT exceeds maximum of %llu",
                              *papCtx->maxAggregateResults);
       return ARG_ERROR;
@@ -301,13 +301,13 @@ static int handleCommonArgs(ParseAggPlanContext *papCtx, ArgsCursor *ac, QueryEr
       return ARG_ERROR;
     }
   } else if (AC_AdvanceIfMatch(ac, "WITHCURSOR")) {
-    if (parseCursorSettings(&papCtx->reqflags, papCtx->cursorConfig, ac, status) != REDISMODULE_OK) {
+    if (parseCursorSettings(papCtx->reqflags, papCtx->cursorConfig, ac, status) != REDISMODULE_OK) {
       return ARG_ERROR;
     }
   } else if (AC_AdvanceIfMatch(ac, "_NUM_SSTRING")) {
-    REQFLAGS_AddFlags(&papCtx->reqflags, QEXEC_F_TYPED);
+    REQFLAGS_AddFlags(papCtx->reqflags, QEXEC_F_TYPED);
   } else if (AC_AdvanceIfMatch(ac, "WITHRAWIDS")) {
-    REQFLAGS_AddFlags(&papCtx->reqflags, QEXEC_F_SENDRAWIDS);
+    REQFLAGS_AddFlags(papCtx->reqflags, QEXEC_F_SENDRAWIDS);
   } else if (AC_AdvanceIfMatch(ac, "PARAMS")) {
     if (parseParams(&(papCtx->searchopts->params), ac, status) != REDISMODULE_OK) {
       return ARG_ERROR;
@@ -316,14 +316,14 @@ static int handleCommonArgs(ParseAggPlanContext *papCtx, ArgsCursor *ac, QueryEr
     if (parseRequiredFields(papCtx->requiredFields, ac, status) != REDISMODULE_OK) {
       return ARG_ERROR;
     }
-    REQFLAGS_AddFlags(&papCtx->reqflags, QEXEC_F_REQUIRED_FIELDS);
+    REQFLAGS_AddFlags(papCtx->reqflags, QEXEC_F_REQUIRED_FIELDS);
   } else if(AC_AdvanceIfMatch(ac, "DIALECT")) {
     dialect_specified = true;
     if (parseDialect(&papCtx->reqConfig->dialectVersion, ac, status) != REDISMODULE_OK) {
       return ARG_ERROR;
     }
   } else if(AC_AdvanceIfMatch(ac, "FORMAT")) {
-    if (parseValueFormat(&papCtx->reqflags, ac, status) != REDISMODULE_OK) {
+    if (parseValueFormat(papCtx->reqflags, ac, status) != REDISMODULE_OK) {
       return ARG_ERROR;
     }
   } else if (AC_AdvanceIfMatch(ac, "_INDEX_PREFIXES") && papCtx->prefixesOffset) {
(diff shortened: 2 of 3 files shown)


tests/cpptests/test_cpp_hybrid.cpp
```diff
@@ -58,8 +58,12 @@ IndexSpec* CreateTestIndexSpec(RedisModuleCtx *ctx, const char* indexName, Query
 }
 
 // Helper function to create a basic AREQ for testing
-AREQ* CreateTestAREQ(RedisModuleCtx *ctx, const char* query, IndexSpec *spec, QueryError *status) {
+AREQ* CreateTestAREQ(RedisModuleCtx *ctx, const char* query, IndexSpec *spec, QueryError *status, bool isSearchSubquery = false) {
   AREQ *req = AREQ_New();
+  if (isSearchSubquery) {
+    AREQ_AddRequestFlags(req, QEXEC_F_IS_SEARCH);
+    AREQ_AddRequestFlags(req, QEXEC_F_SEND_NOFIELDS);
```
raz-mon: Won't this block the serialization of the rlookup later on in `serializeResult`?

ofiryanai: I think it shouldn't. I see now that in `serializeResult_hybrid` there's a check of `QEXEC_F_SEND_NOFIELDS` but this should be read from another options in HybridRequest:

```
typedef struct {
    arrayof(AREQ*) requests;
    size_t nrequests;
    QueryError tailPipelineError;
    QueryError *errors;
    Pipeline *tailPipeline;
    RequestConfig reqConfig;
    HybridPipelineParams *hybridParams;
} HybridRequest;
```

The option that's modified in the line you commented is the options of the first AREQ in `requests`, and is meant for creation of the search subquery. The options that should be used in `serializeResult` is in `hybridParams->aggregationParams->common->reqflags`

ofiryanai: Fixed

src/aggregate/aggregate_request.c
```diff
@@ -274,12 +274,12 @@ static int handleCommonArgs(ParseAggPlanContext *papCtx, ArgsCursor *ac, QueryEr
       // TODO: unify if when req holds only maxResults according to the query type.
       //(SEARCH / AGGREGATE)
     } else if ((arng->limit > *papCtx->maxSearchResults) &&
-               (*papCtx->reqflags & QEXEC_F_IS_SEARCH)) {
+               (*papCtx->reqflags & QEXEC_F_IS_SEARCH | QEXEC_F_IS_HYBRID)) {
```
raz-mon:
```suggestion
               (*papCtx->reqflags & (QEXEC_F_IS_SEARCH | QEXEC_F_IS_HYBRID))) {
```
For the one below as well

src/hybrid/parse_hybrid.c
```diff
@@ -513,7 +513,7 @@ HybridRequest* parseHybridCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
   size_t mergeMaxAggregateResults = RSGlobalConfig.maxAggregateResults;
 
   AREQ **requests = NULL;
-  searchRequest->reqflags |= QEXEC_F_IS_AGGREGATE;
+  searchRequest->reqflags |= QEXEC_F_IS_HYBRID;
```
Itzikvaknin: Why not 
```
searchRequest->reqflags |= (QEXEC_F_IS_SEARCH | QEXEC_F_IS_HYBRID)
```
?

src/pipeline/pipeline_construction.c
```diff
@@ -283,7 +283,7 @@ void Pipeline_BuildQueryPart(Pipeline *pipeline, const QueryPipelineParams *para
    *  * there is no subsequent sorter within this grouping */
   const int reqflags = params->common.reqflags;
   if ((reqflags & (QEXEC_F_SEND_SCORES | QEXEC_F_SEND_SCORES_AS_FIELD)) ||
-      ((reqflags & QEXEC_F_IS_SEARCH) && !(reqflags & QEXEC_F_NOROWS) &&
+      ((reqflags & (QEXEC_F_IS_SEARCH | QEXEC_F_IS_HYBRID)) && !(reqflags & QEXEC_F_NOROWS) &&
```
Itzikvaknin: It seems we are adding `| QEXEC_F_IS_HYBRID` so that it works like a search.
Maybe we should think again whether this means a new type of query, or just marks that it runs in a hybrid context.

ofiryanai: That's what Raz asked, beforehand the search subquery of HybridRequest was of type QEXEC_F_IS_SEARCH and he asked to mark it as QEXEC_F_IS_HYBRID and add this flag along the code path of pipeline construction.

src/hybrid/parse_hybrid.c
ofiryanai: The search subquery has one type, QEXEC_F_IS_HYBRID. Throughout the pipeline construction HYBRID is "by accident" behaves exactly like SEARCH so on every check I added both option in `|`

raz-mon (approved):

src/pipeline/pipeline_construction.c
Itzikvaknin: We’re introducing QEXEC_F_IS_HYBRID, so vector subqueries will have this flag as well.
They would be marked as QEXEC_F_IS_HYBRID | QEXEC_F_IS_AGGREGATE,
allowing them to reach code paths that wouldn’t run with QEXEC_F_IS_AGGREGATE alone.

ofiryanai: Added [commit](https://github.com/RediSearch/RediSearch/pull/6583/commits/a891e5484b6cd59c898236e3f6931092f5c4222b) that helps convey that a query is either Search or Hybrid, not both at the same time.
However I do agree that it's weird that the Vector Subquery is currently marked as `QEXEC_F_IS_AGGREGATE`, and if I change it to `QEXEC_F_IS_HYBRID` the pipeline construction will add unneeded Scorer an Sortby to its pipeline.
Maybe we need to introduce 2 types, `QEXEC_F_IS_HYBRID_SEARCH_SUBQUERY` and `QEXEC_F_IS_HYBRID_VECTOR_SUBQUERY`.
@raz-mon What do you think?

raz-mon: Yeah sounds good, let's be explicit

src/aggregate/aggregate_request.c
```diff
@@ -47,14 +47,14 @@ static bool ensureSimpleMode(AREQ *req) {
  * found in the document - was not requested.
  * name argument must not contain any user data, as it is used for error formatting
  */
-static int ensureExtendedMode(uint32_t *reqflags, const char *name, QueryError *status) {
-  if (*reqflags & QEXEC_F_IS_SEARCH) {
+static int ensureExtendedMode(uint32_t reqflags, const char *name, QueryError *status) {
+  if (reqflags & QEXEC_F_IS_SEARCH) {
```
raz-mon: Why was this done?
Bug?

ofiryanai: Because of the change of `ParseAggPlanContext.reqflag` in src/aggregate/aggregate.h from uint32_t* to uint32_t. Needed to fix some compilation issues in order to use `IsSearch` and similar macros (That does not expect uint32*, which is reasonable. Not sure why it was a pointer to begin with)

Itzikvaknin (approved):

src/aggregate/aggregate_request.c
raz-mon: But now the change below does nothing to the flags

ofiryanai: I see it now 😱 fixed. good catch

src/aggregate/aggregate.h
```diff
@@ -33,8 +33,10 @@ extern "C" {
 typedef struct Grouper Grouper;
 struct QOptimizer;
 
-// A query can be either a search, an aggregate or a hybrid. So QEXEC_F_IS_AGGREGATE,
-// QEXEC_F_IS_SEARCH and QEXEC_F_IS_HYBRID are mutually exclusive (Only one can be set).
+/*
+ * A query can be of one type. So QEXEC_F_IS_AGGREGATE, QEXEC_F_IS_SEARCH and QEXEC_F_IS_HYBRID_TAIL
```
nafraf:
```suggestion
 * A query can be of one type. So QEXEC_F_IS_AGGREGATE, QEXEC_F_IS_SEARCH, QEXEC_F_IS_HYBRID_TAIL
```

src/aggregate/aggregate_request.c
```diff
@@ -269,17 +269,15 @@ static int handleCommonArgs(ParseAggPlanContext *papCtx, ArgsCursor *ac, QueryEr
...
       // LIMIT 0 0 - only count
-      REQFLAGS_AddFlags(papCtx->reqflags, QEXEC_F_NOROWS);
-      REQFLAGS_AddFlags(papCtx->reqflags, QEXEC_F_SEND_NOFIELDS);
+      REQFLAGS_AddFlags(&papCtx->reqflags, QEXEC_F_NOROWS);
+      REQFLAGS_AddFlags(&papCtx->reqflags, QEXEC_F_SEND_NOFIELDS);
       // TODO: unify if when req holds only maxResults according to the query type.
       //(SEARCH / AGGREGATE)
-    } else if ((arng->limit > *papCtx->maxSearchResults) &&
-               (*papCtx->reqflags & QEXEC_F_IS_SEARCH)) {
+    } else if ((arng->limit > *papCtx->maxSearchResults) && (IsSearch(papCtx) || IsHybridSearchSubquery(papCtx))) {
```
nafraf: Why do we need to consider `IsHybridSearchSubquery(papCtx)`?
Are we going to support LIMIT in the search subquery?

ofiryanai: you're right, this more belongs to the tail parsing. changed it to `IsHybridTail`

src/aggregate/aggregate_request.c
```diff
@@ -273,13 +273,11 @@ static int handleCommonArgs(ParseAggPlanContext *papCtx, ArgsCursor *ac, QueryEr
       REQFLAGS_AddFlags(papCtx->reqflags, QEXEC_F_SEND_NOFIELDS);
       // TODO: unify if when req holds only maxResults according to the query type.
       //(SEARCH / AGGREGATE)
-    } else if ((arng->limit > *papCtx->maxSearchResults) &&
-               (*papCtx->reqflags & QEXEC_F_IS_SEARCH)) {
+    } else if ((arng->limit > *papCtx->maxSearchResults) && (*papCtx->reqflags & (QEXEC_F_IS_SEARCH | QEXEC_F_IS_HYBRID_SEARCH_SUBQUERY))) {
```
nafraf, line 276: Can't we use `IsSearch()` and `sHybridSearchSubquery()`?

```suggestion
    } else if ((arng->limit > *papCtx->maxSearchResults) && (IsSearch(papCtx) || IsHybridSearchSubquery(papCtx))) {
```

Itzikvaknin (approved):

src/aggregate/aggregate_request.c
ofiryanai: there was some problem with the macros, they're expecting reqflag of type `uint32_t` but in `ParseAggPlanContext` it's of type `uint32_t*`

ofiryanai: I tried to change `ParseAggPlanContext` to hold it as direct uint32_t but this refactor broke the flow tests and this ticket is on the side for too long. since its not functionally critical I'll leave this refactor to another opportunity.

PR #6589, opened 2025-08-04
[MOD-10381] Implement the encodeFieldsOffsets and encodeFieldsOffsetsWide encoder/decoder in Rust

gdesmott:
Implement the fields offsets encoder and decoder in Rust.

## Benchmarks results

Rust encoders are as fast the C ones.
Decoders are faster than the C versions.

``` 
Encode - FieldsOffsets/C    time:   [4.4944 µs 4.5374 µs 4.5909 µs]
Encode - FieldsOffsets/Rust time:   [4.6087 µs 4.6325 µs 4.6577 µs]

Decode - FieldsOffsets/C    time:   [17.598 ns 17.620 ns 17.643 ns]
Decode - FieldsOffsets/Rust time:   [16.283 ns 16.291 ns 16.299 ns]

Encode - FieldsOffsets Wide/C    time:   [6.4141 µs 6.4460 µs 6.4820 µs]
Encode - FieldsOffsets Wide/Rust time:   [6.3252 µs 6.3704 µs 6.4202 µs]

Decode - FieldsOffsets Wide/C    time:   [56.175 ns 56.336 ns 56.533 ns]
Decode - FieldsOffsets Wide/Rust time:   [25.958 ns 25.979 ns 26.000 ns]
``` 

#### Mark if applicable

### DIFF
--- a/src/inverted_index/inverted_index.c
+++ b/src/inverted_index/inverted_index.c
@@ -749,7 +749,7 @@ DECODER(readFlagsWide) {
   return res->fieldMask & ctx->wideMask;
 }
 
-DECODER(readFlagsOffsets) {
+DECODER(readFieldsOffsets) {
   uint32_t delta, mask;
   qint_decode3(&blockReader->buffReader, &delta, &mask, &res->offsetsSz);
   res->fieldMask = mask;
@@ -759,7 +759,7 @@ DECODER(readFlagsOffsets) {
   return mask & ctx->mask;
 }
 
-DECODER(readFlagsOffsetsWide) {
+DECODER(readFieldsOffsetsWide) {
   uint32_t delta;
   qint_decode2(&blockReader->buffReader, &delta, &res->offsetsSz);
   res->fieldMask = ReadVarintFieldMask(&blockReader->buffReader);
@@ -875,13 +875,13 @@ bool read_flags_wide(IndexBlockReader *blockReader, const IndexDecoderCtx *ctx, 
 }
 
 // Wrapper around the private static `readFlagsOffsets` function to expose it to benchmarking.
-bool read_flags_offsets(IndexBlockReader *blockReader, const IndexDecoderCtx *ctx, RSIndexResult *res) {
-  return readFlagsOffsets(blockReader, ctx, res);
+bool read_fields_offsets(IndexBlockReader *blockReader, const IndexDecoderCtx *ctx, RSIndexResult *res) {
+  return readFieldsOffsets(blockReader, ctx, res);
 }
 
 // Wrapper around the private static `readFlagsOffsetsWide` function to expose it to benchmarking.
-bool read_flags_offsets_wide(IndexBlockReader *blockReader, const IndexDecoderCtx *ctx, RSIndexResult *res) {
-  return readFlagsOffsetsWide(blockReader, ctx, res);
+bool read_fields_offsets_wide(IndexBlockReader *blockReader, const IndexDecoderCtx *ctx, RSIndexResult *res) {
+  return readFieldsOffsetsWide(blockReader, ctx, res);
 }
 
 // Wrapper around the private static `readNumeric` function to expose it to benchmarking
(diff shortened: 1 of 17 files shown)


src/inverted_index/inverted_index.c
```diff
@@ -932,6 +942,16 @@ bool read_flags_wide(IndexBlockReader *blockReader, const IndexDecoderCtx *ctx,
   return readFlagsWide(blockReader, ctx, res);
 }
 
+// Wrapper around the private static `readFlagsOffsets` function to expose it to benchmarking.
+bool read_flags_offsets(IndexBlockReader *blockReader, const IndexDecoderCtx *ctx, RSIndexResult *res) {
```
GuyAv46: Consider fixing the name, at least for the wrapper

src/inverted_index/inverted_index.c
```diff
@@ -932,6 +942,16 @@ bool read_flags_wide(IndexBlockReader *blockReader, const IndexDecoderCtx *ctx,
...
+  return readFlagsOffsets(blockReader, ctx, res);
+}
+
+// Wrapper around the private static `readFlagsOffsetsWide` function to expose it to benchmarking.
+bool read_flags_offsets_wide(IndexBlockReader *blockReader, const IndexDecoderCtx *ctx, RSIndexResult *res) {
```
GuyAv46: same here

src/inverted_index/inverted_index.c
gdesmott: What do you mean? I named it from the existing `readFlagsOffsets()`.

GuyAv46: Exactly
The encoder is called `encodeFieldsOffsets` but the decoder `readFlagsOffsets`. Seems like a typo and that they both should say “fields”

gdesmott: Oh right. That confused me as well at first and then I got used to it. :)

Renamed.

GuyAv46 (approved):

gdesmott:
@chesedo : changed a few things here:
- the bencher was using the wrong decoder (always the wide one) so most benches were actually failing.
- we were not catching those errors as the benches were not asserting decoding to succeed.
- I inlined the result creation as an optimization.

With those changes the Rust decoders are now faster than the C ones! :fireworks:

src/redisearch_rs/inverted_index/src/full.rs
```diff
@@ -114,10 +110,25 @@ fn decode_term_record_offsets<'a>(
...
+        result_type: RSResultType::Term,
+        doc_id: base + delta as t_docId,
+        field_mask,
+        freq,
+        offsets_sz,
+        is_copy: false,
+        dmd: std::ptr::null(),
+        metrics: std::ptr::null_mut(),
+        weight: 0.0,
+    };
```
chesedo: Isn't there a `RSIndexResult` constructor we can use instead? Maybe worth adding one?

chesedo (dismissed):

src/redisearch_rs/inverted_index/src/full.rs
gdesmott: I extended `term_with_term_ptr()` for this.

I also added a small optimization suggested by Luca. See the two last commits.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -534,19 +535,28 @@ impl<'a> RSIndexResult<'a> {
...
     }
 
-    /// Create a new `RSIndexResult` with a given `term` and `offsets`.
+    /// Create a new `RSIndexResult` with a given `term`, `offsets`, `doc_id`, `field_mask`, and `freq`.
     pub fn term_with_term_ptr(
         term: *mut RSQueryTerm,
         offsets: RSOffsetVector<'a>,
+        doc_id: t_docId,
+        field_mask: t_fieldMask,
+        freq: u32,
```
chesedo, line 544: We can use the "builder" pattern setters for these right?

src/redisearch_rs/inverted_index/src/full.rs
chesedo: Did you forget to updatet this call? :see_no_evil:

src/redisearch_rs/inverted_index/src/lib.rs
gdesmott: Yes, but the whole point of those changes is to create the result in a single call as it greatly improves the bench performance.

chesedo: Interesting, I would have assumed the compiler just inlined them all since they are simple.

gdesmott: yeah me too :\

src/redisearch_rs/inverted_index/src/full.rs
gdesmott: oups...

PR #6592, opened 2025-08-04
Add `Revalidate` support for optimized wildcard and missing values indexes - [MOD-10732]

GuyAv46:

1. Add two new dedicated APIs for wildcard and missing, removing the generic API, which is no longer needed
2. Each API sets its `Revalidate` mechanism - wildcard is compared with the current spec's wildcard index, and missing performs a lookup in the missing value indexes dictionary

#### Mark if applicable

### DIFF
--- a/src/iterators/inverted_index_iterator.c
+++ b/src/iterators/inverted_index_iterator.c
@@ -118,6 +118,39 @@ static ValidateStatus TagCheckAbort(QueryIterator *base) {
   return VALIDATE_OK;
 }
 
+static ValidateStatus WildcardCheckAbort(QueryIterator *base) {
+  // Check if the wildcard iterator is still valid
+  InvIndIterator *wi = (InvIndIterator *)base;
+  RS_ASSERT(wi->idx);
+  RS_ASSERT(wi->sctx && wi->sctx->spec);
+
+  if (wi->sctx->spec->existingDocs != wi->idx) {
+    return VALIDATE_ABORTED;
+  }
+  return VALIDATE_OK;
+}
+
+static ValidateStatus MissingCheckAbort(QueryIterator *base) {
+  // Check if the missing iterator is still valid
+  InvIndIterator *mi = (InvIndIterator *)base;
+  RS_ASSERT(mi->idx);
+  RS_ASSERT(mi->sctx && mi->sctx->spec);
+  RS_ASSERT(mi->sctx->spec->missingFieldDict);
+  RS_ASSERT(mi->sctx->spec->numFields > mi->filterCtx.field.value.index);
+
+  const HiddenString *fieldName = mi->sctx->spec->fields[mi->filterCtx.field.value.index]->fieldName;
+  const InvertedIndex *missingII = dictFetchValue(mi->sctx->spec->missingFieldDict, fieldName);
+
+  if (mi->idx != missingII) {
+    // The inverted index was collected entirely by GC.
+    // All the documents that were inside were deleted and new ones were added.
+    // We will not continue reading those new results and instead abort reading
+    // for this specific inverted index.
+    return VALIDATE_ABORTED;
+  }
+  return VALIDATE_OK;
+}
+
 static ValidateStatus InvIndIterator_Revalidate(QueryIterator *base) {
   // Here we should apply the specifics of Term, Tag and Numeric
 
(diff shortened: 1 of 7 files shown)


tests/cpptests/test_cpp_iterator_index.cpp
```diff
@@ -360,7 +357,7 @@ class IndexIteratorTestExpiration : public ::testing::TestWithParam<IndexFlags>
 
           // Make every even document ID field expired
           for (size_t i = 2; i <= n_docs; i += 2) {
-              if (flags & Index_StoreNumeric || flags == Index_DocIdsOnly) {
+              if (flags & Index_StoreNumeric) {
```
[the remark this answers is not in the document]
GuyAv46: It should not be handled specially, just fall to the `else` block

src/iterators/inverted_index_iterator.h
```diff
@@ -86,14 +86,12 @@ QueryIterator *NewInvIndIterator_NumericQuery(const InvertedIndex *idx, const Re
...
 
-// Returns an iterator for a generic index, suitable for queries
-// The returned iterator will yield "virtual" records. For term/numeric indexes, it is best to use
-// the specific functions NewInvIndIterator_TermQuery/NewInvIndIterator_NumericQuery
-// It is assumed that the index will not be dropped while the iterator is in use
-// (for general inverted indexes like terms, tags, this is NOT guaranteed,
-// if all the documents are deleted, the index will be dropped by the garbage collector).
-QueryIterator *NewInvIndIterator_GenericQuery(const InvertedIndex *idx, const RedisSearchCtx *sctx, t_fieldIndex fieldIndex,
-                                              enum FieldExpirationPredicate predicate, double weight);
+// Returns an iterator for a wildcard index (optimized for wildcard queries) - mainly to revalidate the index
```
JoanFM, line 89: what does this comment mean? mainly to revalidate the index?

src/iterators/inverted_index_iterator.h
```diff
@@ -86,14 +86,12 @@ QueryIterator *NewInvIndIterator_NumericQuery(const InvertedIndex *idx, const Re
...
+QueryIterator *NewInvIndIterator_WildcardQuery(const InvertedIndex *idx, const RedisSearchCtx *sctx, double weight);
+
+// Returns an iterator for a missing index - revalidate the missing index was not deleted
+// Result is a virtual result with a weight of 0.0, and a field mask of RS_FIELDMASK_ALL
```
JoanFM, line 93: What is a missing index exactly? what is this mention about Revalidate? U mean these 2 constructors are meant to handle 2 different ways of revalidation?

src/iterators/inverted_index_iterator.c
```diff
@@ -118,6 +114,39 @@ static ValidateStatus TagCheckAbort(QueryIterator *base) {
   return VALIDATE_OK;
 }
 
+static ValidateStatus WildcardCheckAbort(QueryIterator *base) {
+  // Check if the wildcard iterator is still valid
+  InvIndIterator *wi = (InvIndIterator *)base;
+  RS_ASSERT(wi->idx);
+  RS_ASSERT(wi->sctx && wi->sctx->spec);
+
+  if (wi->sctx->spec->existingDocs != wi->idx) {
```
JoanFM, line 123: this would mean the index has been invalidated I guess?

tests/cpptests/micro-benchmarks/benchmark_index_iterator.cpp
```diff
@@ -132,8 +132,6 @@ class BM_IndexIterator : public benchmark::Fixture {
         if (flags == Index_StoreNumeric) {
             FieldFilterContext fieldCtx = {.field = {false, 0}, .predicate = FIELD_EXPIRATION_DEFAULT};
             iterator = NewInvIndIterator_NumericQuery(index, &q_mock->sctx, &fieldCtx, nullptr, nullptr, -INFINITY, INFINITY);
```
JoanFM, line 134: these Missing and Generic Queries are not relevant anymore for benchmarking?

JoanFM: I guess since they only differ in the Revalidate logic, we assume they are not worth benchmarking? However I think this is assuming this will always remind like this. I think we should add them to keep it future proof?

src/iterators/inverted_index_iterator.h
GuyAv46: `Missing` is a feature that maintains an inverted index per field in the schema, containing all the IDs that have no value for this field.
Simple example - `n NUMERIC` and `doc:1 = {"foo":"bar"}`
We have a separate dictionary for these indexes (not sure why, but that's the case today), so we need a custom helper to check if the index we iterate on was not replaced

src/iterators/inverted_index_iterator.c
GuyAv46: Similar to any inverted index, it means the `existingDocs` was either dropped (and is now `NULL`) or dropped and then recreated. We abort in these cases.
We can also decide to just take the new index, but I don't want to change the behavior in this PR

tests/cpptests/micro-benchmarks/benchmark_index_iterator.cpp
GuyAv46: In practice, `Index_DocIdsOnly` cases use either the TagQuery or TermQuery APIs, so this is what we do in the benchmark as well now.
Note that I didn't drop the `Index_DocIdsOnly` scenario, just the special treatment it got in the setup

src/iterators/inverted_index_iterator.h
GuyAv46: An attempt to explain why we need a special API for the wildcard inverted index

JoanFM (approved):

PR #6594, opened 2025-08-04
[MOD-10681] Fix cursor list empty

lerman25:
Currently, `CursorList_Empty` does not acquire the `CursorList_Lock`, despite calling `Cursor_FreeInternal`, which requires it. This can lead to a use-after-free error, as observed in [MOD-10681](https://redislabs.atlassian.net/browse/MOD-10681).

This PR addresses the issue by making `CursorList_Empty` thread-safe. It adopts the same logic used in `Cursors_Purge`:
- If a cursor is idle, it is freed immediately.
- If it is active, it is marked as deleted and will be safely freed on its next use.

This ensures correctness and safety in multi-threaded environments where cursor usage and cleanup may occur concurrently.

[MOD-10681]: https://redislabs.atlassian.net/browse/MOD-10681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/cpptests/test_cpp_cursors.cpp
+++ b/tests/cpptests/test_cpp_cursors.cpp
@@ -10,16 +10,15 @@
 
 #include "gtest/gtest.h"
 #include "cursor.h"
+#include <vector>
+#include <algorithm>
 
 #define is_Idle(cur) ((cur)->pos != -1)
 
 class CursorsTest : public ::testing::Test {};
 
 bool IdInArray(uint64_t id, const uint64_t *arr, int size) {
-  for (int i = 0; i < size; ++i) {
-    if (arr[i] == id) return true;
-  }
-  return false;
+  return std::find(arr, arr + size, id) != arr + size;
 }
 
 
@@ -125,7 +124,7 @@ TEST_F(CursorsTest, OwnershipAPI) {
   // Verify that the idle cursors are freed immediately, and the active ones are marked for deletion
   constexpr int numCursors = 5;
   constexpr int numIdle = numCursors / 2 + numCursors % 2;
-  uint64_t ids[numIdle] = {0};
+  std::vector<uint64_t> ids;
 
   for (int i = 0; i < numCursors; ++i) {
     cur = Cursors_Reserve(&g_CursorsList, dummy, 1000, NULL);
@@ -134,9 +133,10 @@ TEST_F(CursorsTest, OwnershipAPI) {
     ASSERT_FALSE(is_Idle(cur));
     if (i % 2 == 0) {
       ASSERT_EQ(Cursor_Pause(cur), REDISMODULE_OK) << "Cursor should be paused";
-      ids[i % 2] = cur->id;
+      ids.push_back(cur->id);
     }
   }
+
 
 
   ASSERT_EQ(Cursors_GetInfoStats().total_user, numCursors) << "All cursors should be alive";
@@ -156,7 +156,7 @@ TEST_F(CursorsTest, OwnershipAPI) {
     // Assert mark delete
 
     ASSERT_TRUE(cur->delete_mark) << "Cursor should be marked for deletion";
-    ASSERT_FALSE(IdInArray(cur->id, ids, numIdle)) << "Cursor should not be in the deleted array";
+    ASSERT_FALSE(IdInArray(cur->id, ids.data(), ids.size())) << "Cursor should not be in the deleted array";
     // Pause the cursor
     ASSERT_EQ(Cursor_Pause(cur), REDISMODULE_OK) << "Cursor should be paused";
   }


src/cursor.c
```diff
@@ -357,23 +357,22 @@ void Cursors_RenderStatsForInfo(CursorList *cl, CursorList *cl_coord, const Inde
...
-    Cursor_FreeInternal(c);
+    Cursor *cur = kh_val(cl->lookup, ii);
+    if (Cursor_IsIdle(cur)) {
+      // Since the cursor is idle, we can free it.
+      Cursor_RemoveFromIdle(cur);
+      Cursor_FreeInternal(cur);
+    } else {
+      // Since the cursor is not idle, we mark it for deletion.
+      // The next time the cursor is accessed, it will be freed.
+      cur->delete_mark = true;
```
GuyAv46, line 374: Let's extend `CursorsTest` (a new test, or extend `OwnershipAPI`) to cover this case, simulating the scenario

src/cursor.c
```diff
@@ -357,23 +357,22 @@ void Cursors_RenderStatsForInfo(CursorList *cl, CursorList *cl_coord, const Inde
 }
 #endif // FTINFO_FOR_INFO_MODULES
 
-void CursorList_Destroy(CursorList *cl) {
```
GuyAv46, line 360: If we remove `CursorList_Destroy`, let's remove it from the H file as well

GuyAv46 (commented):
Greate implementation

src/cursor.c
```diff
@@ -357,23 +357,22 @@ void Cursors_RenderStatsForInfo(CursorList *cl, CursorList *cl_coord, const Inde
...
+    if (Cursor_IsIdle(cur)) {
+      // Since the cursor is idle, we can free it.
+      Cursor_RemoveFromIdle(cur);
+      Cursor_FreeInternal(cur);
+    } else {
+      // Since the cursor is not idle, we mark it for deletion.
+      // The next time the cursor is accessed, it will be freed.
+      cur->delete_mark = true;
+    }
   }
```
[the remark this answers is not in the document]
GuyAv46: It was already the case, `khash` can handle that.
We should replace the `khash` and array implementation with `dict.h` and `arr.h` in the future

src/cursor.h
```diff
@@ -130,11 +130,6 @@ static inline CursorList *GetGlobalCursor(uint64_t cid) {
...
 void CursorList_Init(CursorList *cl, bool is_coord);
 
-/**
- * Clear the cursor list
- */
-void CursorList_Destroy(CursorList *cl);
-
 /**
  * Empty the cursor list.
  * It is assumed that this function is called from the main thread, and that
```
GuyAv46: Let's update this comment (the last sentence is not true anymore)

tests/cpptests/test_cpp_cursors.cpp
```diff
@@ -72,4 +74,41 @@ TEST_F(CursorsTest, OwnershipAPI) {
...
+  ASSERT_EQ(Cursors_GetInfoStats().total_user, 1) << "Cursor should still be alive";
+  ASSERT_EQ(Cursors_TakeForExecution(&g_CursorsList, id), nullptr) << "Cursor already deleted";
+  ASSERT_TRUE(cur->delete_mark) << "Cursor should be marked for deletion";
+
+  // When cursor is paused, it should actually be freed due to delete_mark
+  ASSERT_EQ(Cursors_GetInfoStats().total_user, 1) << "Cursor should be alive";
+  ASSERT_EQ(Cursor_Pause(cur), REDISMODULE_OK) << "Pausing the cursor should actually free it";
+  ASSERT_EQ(Cursors_GetInfoStats().total_user, 0) << "Cursor should be deleted";
+
+  // Case 4: CursorList_Empty with explicit cursor free
```
GuyAv46, line 105: Consider adding a scenario with multiple cursors, some idle and some not, to test the `khash` iterator deleting while iterating, + keeping some of the cursors for concurrent release

tests/cpptests/test_cpp_cursors.cpp
```diff
@@ -111,4 +120,48 @@ TEST_F(CursorsTest, OwnershipAPI) {
...
+  uint64_t ids[numIdle] = {0};
+
+  for (int i = 0; i < numCursors; ++i) {
+    cur = Cursors_Reserve(&g_CursorsList, dummy, 1000, NULL);
+    ASSERT_TRUE(cur != NULL);
+    ASSERT_FALSE(cur->delete_mark);
+    ASSERT_FALSE(is_Idle(cur));
+    if (i % 2 == 0) {
+      ASSERT_EQ(Cursor_Pause(cur), REDISMODULE_OK) << "Cursor should be paused";
+      ids[i % 2] = cur->id;
```
GuyAv46: currently only setting index `0`
```suggestion
      ids[i / 2] = cur->id;
```
Also, consider using a `std::vector` or a `std::set`, then just push back

tests/cpptests/test_cpp_cursors.cpp
```diff
@@ -15,6 +15,15 @@
 
 class CursorsTest : public ::testing::Test {};
 
+bool IdInArray(uint64_t id, const uint64_t *arr, int size) {
+  for (int i = 0; i < size; ++i) {
+    if (arr[i] == id) return true;
+  }
+  return false;
```
GuyAv46:
```suggestion
  return std::find(arr, arr + size, id) != arr + size;
```

PR #6572, opened 2025-07-30
Iterators refactor - Replace Old Iterators implementation with the new ones - [MOD-8850]

GuyAv46:
## Describe the changes in the pull request

Replacing old implementation usage with the new one, and deleting the old implementation + any now-dead code.

#### Additional changes
1. Move Optimus and Hybrid iterators to the `iterators` directory, and align their flow to the new API (mainly changing the struct and adding the new `Revalidate` API)
2. Remove all the old micro benchmarks and their mock helpers
3. Removing the old "OnReopen" concurrent mechanism
4. Delete cursed `index.c` file

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #6597, opened 2025-08-05
[CI] Fix `test_not:test_not_optimized` flakiness

GuyAv46:
### The Issue
There were a few flaws in the test; the first was the randomness of choosing names, which could yield a faulty name set occasionally, which is now removed

The second is the data set:
1. It has only 473 last names, so with 1005 docs, all appear at least twice, and many 3 times
2. Some names can be both first and last (like "Adam"), so when searching for them, multiple docs may return
3. Some names have the same stem (like "Adam" and "Adams"), so searching for any of them may also return docs with the other
4. Some names contain punctuation, and some of them even have the pair variant - "O'Reilly" and "Reilly". The first will be indexed as two terms - "O" and "Reilly", so searching for "Reilly" will also find "O'Reilly"

All of these cases may yield more or fewer results than expected, and this is why searching for
```python
    res = env.cmd('FT.SEARCH', 'idx', f'-(@t:{name})', 'LIMIT', '0', '0')
```
May return an unexpected number of results.

### The Goal
All of this isn't even worth digging in, because the test goal is to test the negation logic, not the probability for name collisions.

There are three searches in the test:
1. Search for literally `"-t | t"` - `t` is the field name, but here we search for the letter, so `"t"` actually returns 0 results, and `"-t"` is a wildcard
2. Search for `-@t:123` - "123" is not a valid name 
3. Search for `-@t:{name}`

The cases are (or meant to be):
1. Search for a query or its negation - should yield all docs
2. Search for negation of a non-existing term - should yield all docs* 
3. Searching for the negation of a specific term combination - should return all but one document.

*(After #6572 it's also optimized, but will keep for the backports)

### The Fix
Instead of indexing `TEXT`, we'll index `NUMERIC` values so the values are not fluid like text, and also not random

#### Mark if applicable

### DIFF
--- a/tests/pytests/test_not.py
+++ b/tests/pytests/test_not.py
@@ -1,7 +1,4 @@
 from common import *
-import faker
-from random import shuffle
-from faker.providers.person.en import Provider
 
 def test_not_optimized():
     """Tests the optimized version of the NOT iterator, which holds an optimized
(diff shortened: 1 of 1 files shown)


JoanFM (commented):
Can we check if we can remove dependencies from tests? or these faker is used elsewhere?

GuyAv46:
We are using it in `test_missing.py` and `test_existing.py`

PR #6598, opened 2025-08-06
rust flat union iterator

DvirDukhan:

1. The current state briefly
2. What is the change

1. #...
2. MOD...

1. ...
2. ...

**Mark if applicable**

### DIFF
--- a/src/redisearch_rs/rqe_iterators/src/union_iterator.rs
+++ b/src/redisearch_rs/rqe_iterators/src/union_iterator.rs
@@ -23,7 +23,7 @@ enum UnionIteratorMode {
 }
 
 /// Represents a union iterator that aggregates results from multiple child iterators
-struct UnionIterator {
+pub struct UnionIterator {
     /// The active (not depleted) children iterators that are currently being processed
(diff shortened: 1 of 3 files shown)


src/redisearch_rs/rqe_iterators/src/union_iterator.rs
```diff
@@ -0,0 +1,350 @@
...
+
+        let mut results = LowMemoryThinVec::with_capacity(self.active_children.len());
+        // Iterate over all active children iterators
+        for child in self.active_children.iter_mut() {
+            assert!(
+                child.last_doc_id() >= last_id,
+                "Child iterator's last doc id is less than the union iterator's last doc id"
+            );
+            // Read the next result from the child iterator
+            match child.read() {
```
GuyAv46: Unconditional read?
Let's say we have 2 children, and last read returned 1 from the first and 2 from the second, so the union returned 1.
Let's also say the next result from the first child is 3
If we now read from both children, they'll both advance, and we will miss 2 as a valid result of the union

PR #6605, opened 2025-08-06
[MOD-10382] Implement the  encodeOffsetsOnly encoder/decoder in Rust

gdesmott:
Implement the offsets only encoder and decoder in Rust.

## Benchmarks results

Rust encoder is almost as fast as the C one.
Rust decoder is faster than the C version.

```
Encode - OffsetsOnly/C    time:   [729.34 ns 730.81 ns 732.39 ns]
Encode - OffsetsOnly/Rust time:   [719.61 ns 721.94 ns 724.72 ns]

Decode - OffsetsOnly/C    time:   [18.146 ns 18.172 ns 18.202 ns]
Decode - OffsetsOnly/Rust time:   [10.978 ns 10.988 ns 11.004 ns]
```

#### Mark if applicable

### DIFF
--- a/src/inverted_index/inverted_index.c
+++ b/src/inverted_index/inverted_index.c
@@ -870,7 +870,7 @@ DECODER(readFieldsOffsetsWide) {
   return res->fieldMask & ctx->wideMask;
 }
 
-DECODER(readOffsets) {
+DECODER(readOffsetsOnly) {
   uint32_t delta;
   qint_decode2(&blockReader->buffReader, &delta, &res->offsetsSz);
   blockReader->curBaseId = res->docId = delta + blockReader->curBaseId;
@@ -984,9 +984,9 @@ bool read_fields_offsets_wide(IndexBlockReader *blockReader, const IndexDecoderC
   return readFieldsOffsetsWide(blockReader, ctx, res);
 }
 
-// Wrapper around the private static `readOffsets` function to expose it to benchmarking.
-bool read_offsets(IndexBlockReader *blockReader, const IndexDecoderCtx *ctx, RSIndexResult *res) {
-  return readOffsets(blockReader, ctx, res);
+// Wrapper around the private static `readOffsetsOnly` function to expose it to benchmarking.
+bool read_offsets_only(IndexBlockReader *blockReader, const IndexDecoderCtx *ctx, RSIndexResult *res) {
+  return readOffsetsOnly(blockReader, ctx, res);
 }
 
 // Wrapper around the private static `readNumeric` function to expose it to benchmarking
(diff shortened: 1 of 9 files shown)


chesedo (dismissed):

src/inverted_index/inverted_index.c
```diff
@@ -979,6 +984,11 @@ bool read_fields_offsets_wide(IndexBlockReader *blockReader, const IndexDecoderC
   return readFieldsOffsetsWide(blockReader, ctx, res);
 }
 
+// Wrapper around the private static `readOffsets` function to expose it to benchmarking.
+bool read_offsets(IndexBlockReader *blockReader, const IndexDecoderCtx *ctx, RSIndexResult *res) {
+  return readOffsets(blockReader, ctx, res);
```
GuyAv46: Can we align the names again? (both encoder and decoder "offsets" or "offsets only")

src/redisearch_rs/inverted_index/src/offsets_only.rs
```diff
@@ -0,0 +1,61 @@
...
+impl Decoder for OffsetsOnly {
+    fn decode<'a>(
+        &self,
+        cursor: &mut Cursor<&'a [u8]>,
+        base: t_docId,
+    ) -> std::io::Result<RSIndexResult<'a>> {
+        let (decoded_values, _bytes_consumed) = qint_decode::<2, _>(cursor)?;
+        let [delta, offsets_sz] = decoded_values;
+
+        let record = decode_term_record_offsets(cursor, base, delta, 0, 1, offsets_sz)?;
```
GuyAv46, line 58: Shouldn't the field mask be "all"?

gdesmott: Should we? The C implementation does not set it, see https://github.com/RediSearch/RediSearch/blob/master/src/inverted_index/inverted_index.c#L868

src/inverted_index/inverted_index.c
gdesmott: renamed to `readOffsetsOnly` as we already have `readDocIdsOnly`.

src/redisearch_rs/inverted_index/src/offsets_only.rs
GuyAv46: It does not set it, but it also is not setting it to 0, and we set it to `RS_FIELDMASK_ALL` on the reused result. See https://github.com/RediSearch/RediSearch/blob/24f63e9ccb58495c82de39981b8b5a5dcc454765/src/iterators/inverted_index_iterator.c#L642 and any other `NewTokenRecord` call in the file.
On the other hand, I'm not sure it matters; we only use the field mask for filtering in the decoder.
I think we can keep it this way

GuyAv46 (dismissed):

chesedo (dismissed):

chesedo (approved):

PR #6607, opened 2025-08-07
[MOD-10158] Guard access to `RSIndexResult`'s data field

chesedo:

We want to remove the `union` on this field to replace it with a proper Rust `enum`. Doing so will change the layout of the field a bit. This PR makes it easier to update that layout by putting all the accessors of the data field behind function getters.

#### Mark if applicable

### DIFF
--- a/src/inverted_index/inverted_index.c
+++ b/src/inverted_index/inverted_index.c
@@ -828,7 +828,12 @@ DECODER(readNumeric) {
     if (NumericFilter_IsNumeric(f)) {
       return NumericFilter_Match(f, value);
     } else {
-      return isWithinRadius(f->geoFilter, value, &value);
+      int filtered = isWithinRadius(f->geoFilter, value, &value);
+
+      // Update the value with the new calculated distance
+      IndexResult_SetNumValue(res, value);
+
+      return filtered;
     }
   }
 
--- a/src/inverted_index/inverted_index.c
+++ b/src/inverted_index/inverted_index.c
@@ -232,7 +232,8 @@ void InvertedIndex_Free(void *ctx) {
 // 1. Encode the full data of the record, delta, frequency, field mask and offset vector
 ENCODER(encodeFull) {
   uint32_t offsets_len;
-  const char *offsets_data = RSOffsetVector_GetData(&res->data.term.offsets, &offsets_len);
+  const RSTermRecord *term = IndexResult_TermRef(res);
+  const char *offsets_data = RSOffsetVector_GetData(&term->offsets, &offsets_len);
   size_t sz = qint_encode4(bw, delta, res->freq, (uint32_t)res->fieldMask, res->offsetsSz);
   sz += Buffer_Write(bw, offsets_data, offsets_len);
   return sz;
@@ -240,7 +241,8 @@ ENCODER(encodeFull) {
 
 ENCODER(encodeFullWide) {
   uint32_t offsets_len;
-  const char *offsets_data = RSOffsetVector_GetData(&res->data.term.offsets, &offsets_len);
+  const RSTermRecord *term = IndexResult_TermRef(res);
+  const char *offsets_data = RSOffsetVector_GetData(&term->offsets, &offsets_len);
   size_t sz = qint_encode3(bw, delta, res->freq, res->offsetsSz);
   sz += WriteVarintFieldMask(res->fieldMask, bw);
   sz += Buffer_Write(bw, offsets_data, offsets_len);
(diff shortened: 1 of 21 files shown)


gdesmott (dismissed):

src/inverted_index/inverted_index.c
```diff
@@ -775,25 +784,27 @@ DECODER(readNumeric) {
   Buffer_Read(&blockReader->buffReader, &delta, header.encCommon.deltaEncoding);
   blockReader->curBaseId = res->docId = blockReader->curBaseId + delta;
 
+  double value = 0;
```
meiravgri, line 787: Since the switch case doesn't have a default case i suggest to set value to the original stored value
```suggestion
  double value = IndexResult_NumValue(res);
```

src/inverted_index/inverted_index.c
```diff
@@ -802,20 +813,22 @@ DECODER(readNumeric) {
...
+  IndexResult_SetNumValue(res, value);
+
   const NumericFilter *f = ctx->filter;
   if (f) {
     if (NumericFilter_IsNumeric(f)) {
-      return NumericFilter_Match(f, res->data.num.value);
+      return NumericFilter_Match(f, value);
     } else {
-      return isWithinRadius(f->geoFilter, res->data.num.value, &res->data.num.value);
+      return isWithinRadius(f->geoFilter, value, &value);
```
meiravgri: Can this call changes the underlying value of `&value` ? if so we should call `IndexResult_SetNumValue` again after it returns

src/iterators/hybrid_reader.c
```diff
@@ -12,11 +12,11 @@
 #include "VecSim/query_results.h"
 #include "wildcard_iterator.h"
 
-#define VECTOR_RESULT(p) (p->type == RSResultType_Metric ? p : AggregateResult_Get(&p->data.agg, 0))
+#define VECTOR_VALUE(p) (p->type == RSResultType_Metric ? IndexResult_NumValue(p) : IndexResult_NumValue(AggregateResult_Get(IndexResult_AggregateRef(p), 0)))
 
 static int cmpVecSimResByScore(const void *p1, const void *p2, const void *udata) {
   const RSIndexResult *e1 = p1, *e2 = p2;
-  double score1 = VECTOR_RESULT(e1)->data.num.value, score2 = VECTOR_RESULT(e2)->data.num.value;
+  double score1 = VECTOR_VALUE(e1), score2 = VECTOR_VALUE(e2);
```
meiravgri: why the renaming?

src/inverted_index/inverted_index.c
chesedo: This will preserve the old behaviour. But I think it is technically incorrect to just use the last decoded value hence why I'm using `0` instead

src/inverted_index/inverted_index.c
chesedo: Yes, it seems the internal code does update `value` although I can't figure out why it does though. @GuyAv46 why does `isWithinRadiusLonLat` update the `distance` and is it correct for the numeric decoder to have its decoded value updated?

src/iterators/hybrid_reader.c
chesedo: The new name just seemed better since the macro no longer returns the RS result, but the actual numeric value

src/inverted_index/inverted_index.c
meiravgri: 1. What is the value before we get to the switch case?
2. How can we or the user distinguish between 0.0 as a valid value and a "bug" 0.0?

src/inverted_index/inverted_index.c
GuyAv46: `isWithinRadius` gets the filter as the first parameter, the **encoded geo location** stored as a double, and an optional output parameter for the calculated distance.
So basically, we used to both check the encoded location and replace the value with the distance from the query.
Although it is not actually used today, we probably want to keep this behavior and set the result value to the distance, and not keep the encoded location.
We also should improve the documentation and variable names in these functions so it's clearer... :(

src/iterators/hybrid_reader.c
meiravgri: vectors don’t really have a value so "vector value" feels a bit off... 
Consider something like VECTOR_SCORE

src/inverted_index/inverted_index.c
chesedo: 1. In the original code the value before the switch would either have been the value of the last decoded record (which I think is wrong) or 0 if this is the first record being decoded because of this function.

https://github.com/RediSearch/RediSearch/blob/dfc639627aa2ac2ce2f1d864b583b9273d383aa8/src/index_result.c#L87-L99

2. I don't see the current code making a distinction between a valid and "bug" value of 0.0. And it already uses 0.0 for the first decoded record

src/iterators/hybrid_reader.c
chesedo: Ahh, I see. Done in https://github.com/RediSearch/RediSearch/pull/6607/commits/5470bc04f7fcdfab099bc7b140242b219256e512

src/inverted_index/inverted_index.c
chesedo: Thanks @GuyAv46! I made it set the calculated distance in https://github.com/RediSearch/RediSearch/pull/6607/commits/1fc02d429c4698fe2379913e7ebe6342c94caee2

PR #6610, opened 2025-08-07
MOD-10550 MOD-10645 Window / Limit defaults

ofiryanai:
## Set defaults logic of Window, KNN K and Limit

Some of the changes:
- **LIMIT → KNN K fallback**: When LIMIT is explicitly provided but K is not, K uses LIMIT value
- **LIMIT → WINDOW fallback**: When LIMIT is explicitly provided but WINDOW is not, WINDOW uses LIMIT value
- **KNN K <= WINDOW constraint**: Prevent larger K that will query extra results from Vector Index that the Hybrid Merger will never yield 
- **Explicit parameter tracking**: `hasExplicitK` and `hasExplicitWindow` flags track user intent
- **`HybridScoringContext` constructors**: Easier, single point of initialization that sets the defaults
- **RRF K parsing fix**: Changed to double to support decimals like in PRD example (`COMBINE RRF 4 WINDOW 40 K 1.5`)


src/hybrid/hybrid_scoring.h
```diff
@@ -7,6 +7,11 @@
 extern "C" {
 #endif
 
+
+// Default constants for hybrid search parameters
+#define HYBRID_DEFAULT_WINDOW 20
+#define HYBRID_DEFAULT_RRF_K 1
```
raz-mon: Not 60?

src/hybrid/hybrid_scoring.h
```diff
@@ -18,18 +23,24 @@ typedef struct {
...
 
 typedef struct {
-  size_t k;
-  size_t window;
+  double k;
 } HybridRRFContext;
 
 typedef struct {
   HybridScoringType scoringType;
+  size_t window;
```
raz-mon: Why do we want a `window` for Linear? I would expect an `FT.AGGREGATE` behavior for this case, i.e., all results.
I think it makes more sense to have the `window` as part of the `HybridRRFContext` such that it matches the command

src/hybrid/parse_hybrid.c
```diff
@@ -391,6 +398,11 @@ static int parseCombine(ArgsCursor *ac, HybridScoringContext *combineCtx, QueryE
 
   // Parse parameters based on scoring type
   if (combineCtx->scoringType == HYBRID_SCORING_LINEAR) {
+    // Store original weights if they exist
```
raz-mon:
```suggestion
    // Store original weights if they exist (take ownership)
```

src/hybrid/parse_hybrid.c
```diff
@@ -452,7 +471,13 @@ static int parseCombine(ArgsCursor *ac, HybridScoringContext *combineCtx, QueryE
   return REDISMODULE_OK;
 
 error:
-  HybridScoringContext_Free(combineCtx);
+  // Restore original state and clean up any allocated resources
+  if (combineCtx->scoringType == HYBRID_SCORING_LINEAR && combineCtx->linearCtx.linearWeights) {
+    rm_free(combineCtx->linearCtx.linearWeights);
+    combineCtx->linearCtx.linearWeights = originalWeights;
+  }
+  combineCtx->scoringType = originalType;
```
raz-mon: Any reason to do this?

src/hybrid/parse_hybrid.c
```diff
@@ -466,6 +491,59 @@ static void copyRequestConfig(RequestConfig *dest, const RequestConfig *src) {
   dest->BM25STD_TanhFactor = src->BM25STD_TanhFactor;
 }
 
+// Helper function to get LIMIT value from parsed aggregation pipeline
+static size_t getLimitFromPipeline(Pipeline *pipeline) {
+  if (!pipeline) return 0;
```
raz-mon:
```suggestion
  RS_ASSERT(pipeline);
```

src/hybrid/parse_hybrid.c
```diff
@@ -466,6 +491,59 @@ static void copyRequestConfig(RequestConfig *dest, const RequestConfig *src) {
...
+ * @param vectorRequest The vector request containing KNN parameters
+ * @param hybridParams The hybrid parameters containing WINDOW settings
+ */
+static void applyLimitParameterFallbacks(Pipeline *mergePipeline,
+                                       AREQ *vectorRequest,
+                                       HybridPipelineParams *hybridParams) {
+  size_t limitValue = getLimitFromPipeline(mergePipeline);
+  bool hasExplicitLimit = hasExplicitLimitInPipeline(mergePipeline);
+
+  // Apply LIMIT → KNN K fallback ONLY if K was not explicitly set AND LIMIT was explicitly provided
```
raz-mon, line 515: The limit in case of explicit `KNN K` or `window` is `UINT32_MAX`?
Also, I wonder what happens in the mixed cases, e.g., explicit LIMIT and WINDOW with no KNN K.
Do they all compile?

tests/cpptests/test_cpp_hybrid_defaults.cpp
```diff
@@ -0,0 +1,352 @@
...
+  }
+
+  void TearDown() override {
+    if (sctx) {
+      SearchCtx_Free(sctx);
+      sctx = nullptr;
+    }
+    if (ctx) {
+      // Drop the index to clean up
+      RedisModuleCallReply *reply = RedisModule_Call(ctx, "FT.DROPINDEX", "c", index_name.c_str());
```
raz-mon: Don't think this is needed, the env will get destroyed with the index

tests/cpptests/test_cpp_hybrid_defaults.cpp
```diff
@@ -0,0 +1,352 @@
...
+      if (reply) {
+        RedisModule_FreeCallReply(reply);
+      }
+      RedisModule_FreeThreadSafeContext(ctx);
+      ctx = nullptr;
+    }
+  }
+
+  void verifyDefaults(HybridRequest* result, size_t expectedWindow,
```
raz-mon:
```suggestion
  void verifyParams(HybridRequest* result, size_t expectedWindow,
```

src/hybrid/hybrid_scoring.h
ofiryanai: yes it should be 60, thanks for validating not sure when it changed

src/hybrid/hybrid_scoring.h
ofiryanai: I think window for linear is to make sure we don't get infinite results. basically it's the limit of the subqueries. why would it be different in linear and RRF?
anyways, sounds like a product question

src/hybrid/parse_hybrid.c
ofiryanai: what should happen in that case is that KNN K takes the explicit LIMIT value, but WINDOW follows the other explicit value. let me add test case for it if there's none yet

tests/cpptests/test_cpp_hybrid_defaults.cpp
raz-mon: Can we also verify the `KNN K` value?

tests/cpptests/test_cpp_hybrid_defaults.cpp
```diff
@@ -0,0 +1,352 @@
...
+
+  ASSERT_EQ(status.code, QUERY_OK) << "Parse failed: " << (status.detail ? status.detail : "NULL");
+
+  // Verify: window=15, hasExplicitWindow=true
+  verifyDefaults(result, 15, true);
+
+  HybridRequest_Free(result);
+}
+
+// LIMIT=0 doesn't affect defaults
```
raz-mon: Why not?
That doesn't seem right

tests/cpptests/test_cpp_hybrid_defaults.cpp
```diff
@@ -0,0 +1,352 @@
...
+  
+  ASSERT_EQ(status.code, QUERY_OK) << "Parse failed: " << (status.detail ? status.detail : "NULL");
+  
+  // Verify: window=10000, hasExplicitWindow=false
+  verifyDefaults(result, 10000, false);
+  
+  HybridRequest_Free(result);
+}
+
+// Large LIMIT doesn't override explicit
```
raz-mon: What's the threshold for overriding?
I don't remember there was one..

tests/cpptests/test_cpp_parse_hybrid.cpp
```diff
@@ -286,7 +280,114 @@ TEST_F(ParseHybridTest, testWithCombineRRF) {
   HybridRequest_Free(result);
 }
 
+TEST_F(ParseHybridTest, testWithCombineRRFWithK) {
```
raz-mon, line 217: Isn't this tested in `test_hybrid_defaults`?
Same for the test below

src/hybrid/parse_hybrid.c
ofiryanai: dont think so, removed

src/hybrid/parse_hybrid.c
ofiryanai: added the test.
re UINT32_MAX, I think it might. don't see a problem with that, to what value would you limit that otherwise?

tests/cpptests/test_cpp_hybrid_defaults.cpp
ofiryanai: there's no threshold just a test that augment considered an edge case

src/hybrid/parse_hybrid.c
```diff
@@ -90,7 +89,7 @@ static int parseSearchSubquery(ArgsCursor *ac, AREQ *sreq, QueryError *status) {
   return REDISMODULE_OK;
 }
 
-static int parseKNNClause(ArgsCursor *ac, VectorQuery *vq, QueryAttribute **attributes, QueryError *status) {
+static int parseKNNClause(ArgsCursor *ac, VectorQuery *vq, QueryAttribute **attributes, ParsedVectorData *pvd, QueryError *status) {
```
nafraf: We are invoking: `parseKNNClause(ac, vq, &pvd->attributes, pvd, status)`
Do we still need `attributes` as a separate argument?

src/hybrid/hybrid_scoring.h
raz-mon: The window is a part of the RRF scoring specifically.
I don't see how it's related to the linear case, where the `LIMIT` should be the one to limit the number of results

tests/cpptests/test_cpp_parse_hybrid.cpp
ofiryanai: they test the defaults logic, this tests the basic parsing. because a ticket to validate the parsing was added

tests/cpptests/test_cpp_hybrid_defaults.cpp
ofiryanai: Actually, I'm not sure whats the expected behavior on LIMIT 0 0 on Hybrid. counting results seems non obvious (should we count results that appear in both subqueries once or twice?) and missing the purpose of this command.
I'm removing this test and if this requires further handling lets make it on separate ticket to avoid scope creep

src/hybrid/hybrid_request.c
```diff
@@ -189,8 +189,10 @@ void HybridRequest_Free(HybridRequest *req) {
 
     // Free the scoring context resources
     if (req->hybridParams) {
-      // The scoring context is freed by the hybrid merger
-      // HybridScoringContext_Free(req->hybridParams->scoringCtx);
+      // Free the scoring context - it's owned by the HybridRequest
+      if (req->hybridParams->scoringCtx) {
+        HybridScoringContext_Free(req->hybridParams->scoringCtx);
+      }
```
Itzikvaknin:
```suggestion
      HybridScoringContext_Free(req->hybridParams->scoringCtx);
```

We perform a null check here:
```void HybridRequest_Free(HybridRequest *req) {
    if (!req) return;
```

src/hybrid/hybrid_scoring.c
```diff
@@ -79,6 +79,68 @@ double HybridLinearScore(HybridScoringContext *scoringCtx, const double *scores,
...
+ * Creates a new HybridScoringContext configured for RRF scoring.
+ *
+ * @param k RRF k parameter (supports floating-point values)
+ * @param window Window size for result processing
+ * @param hasExplicitWindow Whether window was explicitly set by user
+ * @return Allocated HybridScoringContext or NULL on failure
+ */
+HybridScoringContext* HybridScoringContext_NewRRF(double k, size_t window, bool hasExplicitWindow) {
+    HybridScoringContext *ctx = rm_calloc(1, sizeof(HybridScoringContext));
+    if (!ctx) return NULL;
```
Itzikvaknin: The check for successful allocation can be removed

src/hybrid/hybrid_scoring.c
```diff
@@ -79,6 +79,68 @@ double HybridLinearScore(HybridScoringContext *scoringCtx, const double *scores,
...
+    return ctx;
+}
+
+/**
+ * Constructor with default RRF values.
+ * Creates a new HybridScoringContext with default RRF configuration.
+ *
+ * @return Allocated HybridScoringContext or NULL on failure
+ */
+HybridScoringContext* HybridScoringContext_NewDefault(void) {
```
Itzikvaknin, line 134: Nice ⚡

src/hybrid/parse_hybrid.c
```diff
@@ -125,6 +124,7 @@ static int parseKNNClause(ArgsCursor *ac, VectorQuery *vq, QueryAttribute **attr
       }
       vq->knn.k = (size_t)kValue;
       hasK = true;  // codespell:ignore
+      pvd->hasExplicitK = true;
```
Itzikvaknin, line 127: If there is a KNN clause, it must include an explicit K.
This check should be moved to `parseVectorSubquery`, where `parseKNNClause` is called, to keep the syntax validation in the right place

src/hybrid/parse_hybrid.c
```diff
@@ -310,14 +310,15 @@ static int parseVectorSubquery(ArgsCursor *ac, AREQ *vreq, QueryError *status) {
...
+  vq->knn.k = HYBRID_DEFAULT_KNN_K;
   vq->knn.order = BY_SCORE;
+  pvd->hasExplicitK = false;
 
   if (AC_IsAtEnd(ac)) goto final;
 
   // Parse optional KNN or RANGE clause
   if (AC_AdvanceIfMatch(ac, "KNN")) {
-    if (parseKNNClause(ac, vq, &pvd->attributes, status) != REDISMODULE_OK) {
+    if (parseKNNClause(ac, vq, &pvd->attributes, pvd, status) != REDISMODULE_OK) {
```
Itzikvaknin: If we move the `pvd->hasExplicitK = true;` here, I think we can remove the `pvd` argument

src/hybrid/parse_hybrid.c
```diff
@@ -391,6 +398,11 @@ static int parseCombine(ArgsCursor *ac, HybridScoringContext *combineCtx, QueryE
 
   // Parse parameters based on scoring type
   if (combineCtx->scoringType == HYBRID_SCORING_LINEAR) {
+    // Store original weights if they exist (take ownership)
```
Itzikvaknin: Should nullify the source?

src/hybrid/parse_hybrid.c
```diff
@@ -429,7 +447,7 @@ static int parseCombine(ArgsCursor *ac, HybridScoringContext *combineCtx, QueryE
 
         if (strcasecmp(paramName, "K") == 0) {
           double k;
-          if (AC_GetDouble(&params, &k, 0) != AC_OK) {
+          if (AC_GetDouble(&params, &k, 0) != AC_OK || k <= 0) {
```
Itzikvaknin, line 437: Ranks start at 1, so maybe `k<=0` is too strict

src/hybrid/parse_hybrid.c
```diff
@@ -466,6 +487,59 @@ static void copyRequestConfig(RequestConfig *dest, const RequestConfig *src) {
...
+ *
+ * @param mergePipeline The pipeline to extract LIMIT from
+ * @param vectorRequest The vector request containing KNN parameters
+ * @param hybridParams The hybrid parameters containing WINDOW settings
+ */
+static void applyLimitParameterFallbacks(Pipeline *mergePipeline,
+                                       AREQ *vectorRequest,
+                                       HybridPipelineParams *hybridParams) {
+  size_t limitValue = getLimitFromPipeline(mergePipeline);
+  bool hasExplicitLimit = hasExplicitLimitInPipeline(mergePipeline);
```
Itzikvaknin:
```suggestion
static void applyLimitParameterFallbacks(Pipeline *mergePipeline,
                                       ParsedVectorData *pvd,
                                       HybridPipelineParams *hybridParams) {
  size_t limitValue = getLimitFromPipeline(mergePipeline);
  bool hasExplicitLimit = hasExplicitLimitInPipeline(mergePipeline);
```

tests/cpptests/test_cpp_parse_hybrid.cpp
```diff
@@ -278,15 +265,115 @@ TEST_F(ParseHybridTest, testWithCombineRRF) {
...
+
+  // Verify the request was parsed successfully
+  ASSERT_TRUE(result != NULL);
+  ASSERT_EQ(status.code, QUERY_OK);
+
+  // Verify RRF scoring type was set with custom WINDOW value
+  assertRRFScoringCtx(HYBRID_DEFAULT_RRF_K, 25);
+
+  // Verify hasExplicitWindow flag is true (WINDOW was specified)
+  ASSERT_TRUE(result->hybridParams->scoringCtx->hasExplicitWindow);
```
Itzikvaknin: Can we use something like: 
```
testLimitWindowKLogic(RMCK::ArgvList, expected_limit, expected_window, expected_knn_k(optional))
```
?

Itzikvaknin (commented):
Nice ⚡

src/hybrid/parse_hybrid.c
raz-mon: I think in general `KNN K <= WINDOW` and that `KNN K <= LIMIT` should be maintained, even when they are explicitly passed since more results won't be looked at anyway. WDYT?

src/hybrid/hybrid_scoring.c
```diff
@@ -79,6 +79,64 @@ double HybridLinearScore(HybridScoringContext *scoringCtx, const double *scores,
...
+ *
+ * @param weights Array of weight values to copy
+ * @param numWeights Number of weights in the array
+ * @return Allocated HybridScoringContext or NULL on failure
+ */
+HybridScoringContext* HybridScoringContext_NewLinear(const double *weights, size_t numWeights) {
+    if (!weights || numWeights == 0) return NULL;
+
+    HybridScoringContext *ctx = rm_calloc(1, sizeof(HybridScoringContext));
+    if (!ctx) return NULL;
```
raz-mon: Also here, let's remove this we don't check the success of the every allocation

src/hybrid/parse_hybrid.c
```diff
@@ -466,6 +474,60 @@ static void copyRequestConfig(RequestConfig *dest, const RequestConfig *src) {
...
+  }
+  return 0;
+}
+
+// Helper function to check if LIMIT was explicitly provided
+static bool hasExplicitLimitInPipeline(Pipeline *pipeline) {
+  if (!pipeline) return false;
+
+  PLN_ArrangeStep *arrangeStep = AGPLN_GetArrangeStep(&pipeline->ap);
+  return (arrangeStep && arrangeStep->isLimited);
```
raz-mon: Notice that a `KNN` query sets this as well. If that's still fine for the check you want to check here, let's add that to the comment

src/hybrid/parse_hybrid.c
```diff
@@ -466,6 +474,60 @@ static void copyRequestConfig(RequestConfig *dest, const RequestConfig *src) {
...
+ *
+ * @param mergePipeline The pipeline to extract LIMIT from
+ * @param vectorRequest The vector request containing KNN parameters
+ * @param hybridParams The hybrid parameters containing WINDOW settings
+ */
+static void applyLimitParameterFallbacks(Pipeline *mergePipeline,
+                                       AREQ *vectorRequest,
+                                       HybridPipelineParams *hybridParams) {
+  size_t limitValue = getLimitFromPipeline(mergePipeline);
+  bool hasExplicitLimit = hasExplicitLimitInPipeline(mergePipeline);
```
raz-mon: Will this not get the default value of `10`, since the KNN step sets the `isLimited` field as well?

src/result_processor.c
```diff
@@ -1802,7 +1802,14 @@ dictType dictTypeHybridSearchResult = {
       if (consumed[i]) {
         continue;
       }
-      size_t window = self->hybridScoringCtx->scoringType == HYBRID_SCORING_RRF ? self->hybridScoringCtx->rrfCtx.window : SIZE_MAX;
+      size_t window;
+      if (self->hybridScoringCtx->scoringType == HYBRID_SCORING_RRF) {
+        window = self->hybridScoringCtx->rrfCtx.window;
+      } else {
+        // For LINEAR scoring, use the aggregation pipeline limit or default max aggregate results
+        window = (self->base.parent && self->base.parent->resultLimit > 0) ?
```
raz-mon: Notice we initialize `resultLimit` with either the limit or `req->maxAggregateResults`, so always get the first case here. Also there is no way we'd get here when `resultLimit <= 0`.

src/hybrid/hybrid_scoring.h
ofiryanai: I thought it might be better to set a limit by default and that window is that mechanism as it has sane limit. Anyway its fixed now I reverted the change (manually) and now window is back to RRF only

tests/cpptests/test_cpp_parse_hybrid.cpp
ofiryanai: EDIT: extracted some setup and teardown repetitive code in defaults tests

src/hybrid/parse_hybrid.c
ofiryanai: fixed

src/hybrid/parse_hybrid.c
ofiryanai: 0<k<1 is aggressive, but a valid input according to chatgpt

src/hybrid/parse_hybrid.c
ofiryanai: got rid of `originalWeights`, it really was redundant

src/hybrid/parse_hybrid.c
ofiryanai: I disagree. in the future we might decide that we can write KNN to set EF_RUNTIME for example, but not K as it has a default. that's not the case now but this is where K is set so this is where it belong to mark it as set IMO

src/hybrid/hybrid_scoring.c
ofiryanai: removed

src/hybrid/hybrid_request.c
ofiryanai: Removed the surrounding if

src/hybrid/parse_hybrid.c
ofiryanai: you're right, refactored the params of the function and used pvd inside it to get its attributes. fixed

tests/cpptests/test_cpp_hybrid_defaults.cpp
ofiryanai: removed the test case

tests/cpptests/test_cpp_hybrid_defaults.cpp
ofiryanai: done

tests/cpptests/test_cpp_parse_hybrid.cpp
```diff
@@ -316,8 +403,47 @@ TEST_F(ParseHybridTest, testComplexSingleLineCommand) {
...
+
+  // Verify the request was parsed successfully
+  ASSERT_TRUE(result != NULL);
+  ASSERT_EQ(status.code, QUERY_OK);
+
+  // Verify the structure contains expected number of requests
+  ASSERT_EQ(result->nrequests, 2);
+
+  // Verify RRF scoring type was set with explicit WINDOW value (30), not LIMIT fallback
+  assertRRFScoringCtx(HYBRID_DEFAULT_RRF_K, 30);
```
raz-mon, line 293: Why not have the window = 15 here?

raz-mon: Basically meaning to enforce `WINDOW <= LIMIT`
Also the window will practically be 15 since the knn is 15

src/hybrid/parse_hybrid.c
ofiryanai: I don't think it should be enforced. if a user asks it explicitly he might have some reason, I don't want to be too defensive

tests/cpptests/test_cpp_parse_hybrid.cpp
ofiryanai: This is what I understood from the PRD, please raise it to Adriano.

src/hybrid/hybrid_scoring.c
ofiryanai: removed

src/hybrid/parse_hybrid.c
raz-mon: It's a basic optimization that is expected IMO
What reason do we have to request for more results than what we'll actually take?

src/result_processor.c
ofiryanai: removed the whole if

src/hybrid/parse_hybrid.c
ofiryanai: renamed for clarity

src/hybrid/parse_hybrid.c
ofiryanai: After discussing F2F aggreed that WINDOW might be larger than LIMIT but KNN K should not exceed WINDOW (it can but it's a waste). changing the behavior

tests/cpptests/test_cpp_parse_hybrid.cpp
ofiryanai: as discussed, WINDOW <= LIMIT constraint is not always correct

src/hybrid/parse_hybrid.c
```diff
@@ -466,6 +474,82 @@ static void copyRequestConfig(RequestConfig *dest, const RequestConfig *src) {
...
+
+  PLN_ArrangeStep *arrangeStep = AGPLN_GetArrangeStep(&pipeline->ap);
+  if (arrangeStep && arrangeStep->isLimited && arrangeStep->limit > 0) {
+    return (size_t)arrangeStep->limit;
+  }
+  return 0;
+}
+
+// Helper function to check if LIMIT was explicitly provided
+static bool tailHasExplicitLimitInPipeline(Pipeline *pipeline) {
```
Itzikvaknin, line 489: Why  specifically `tail`?

tests/cpptests/test_cpp_hybrid_defaults.cpp
Itzikvaknin: ⬆️

src/result_processor.c
```diff
@@ -1802,7 +1802,18 @@ dictType dictTypeHybridSearchResult = {
...
+        window = self->hybridScoringCtx->rrfCtx.window;
+      } else {
+        // For LINEAR scoring, use the tail pipeline limit
+        // If parent is NULL or resultLimit is 0 (e.g., in tests), use a reasonable default
+        if (self->base.parent && self->base.parent->resultLimit > 0) {
+          window = self->base.parent->resultLimit;
+        } else {
+          window = 1000; // Default window size for LINEAR scoring
+        }
+      }
```
raz-mon: Where did the default `window=1000` for LINEAR come from?

Also, is there a situation in which there is no parent, also in the tests?
Same for the `resultLimit > 0` condition.. We shouldn't get here in either case.

src/hybrid/parse_hybrid.c
ofiryanai: Raz commented that in vector pipeline there's a limit in the pipeline although it wasn't explicitly set by the user there. I changed the function title to try to be more specific

src/result_processor.c
ofiryanai: yes, on HybridMergerTest we create a merger and run only that. 1000 was there in previous commits, when merger was first written and wasn't connected to the rest of the code yet.

ofiryanai: Some Merger tests also fail if I remove the `self->base.parent->resultLimit > 0` condition

raz-mon: It doesn't make sense to have this in our production code. This doesn't go hand by hand with our design and requirement at all.
This code assumes that these exist, it should be asserted much earlier

ofiryanai: ok, removed dependency on parent RP, this really introduced extra dependency for no reason. Set it to infinity instead and `ConsumeFromUpstream` will stop when upstream will drain all results (they have default/explicit Limit configured)

Itzikvaknin (approved):
Nice ⚡

Itzikvaknin (approved):

PR #6611, opened 2025-08-07
[MOD-10385] Implement the encodeRawDocIdsOnly encoder/decoder in Rust

gdesmott:
Implement the raw doc ids only encoder and decoder in Rust.

## Benchmarks results

The Rust encoder is almost as fast as the C one.
The Rust decoder is faster.

```
Encode - RawDocIdsOnly/C    time:   [46.431 ns 46.543 ns 46.668 ns]
Encode - RawDocIdsOnly/Rust time:   [49.432 ns 49.652 ns 49.936 ns]

Decode - RawDocIdsOnly/C    time:   [16.655 ns 16.677 ns 16.705 ns]
Decode - RawDocIdsOnly/Rust time:   [13.684 ns 13.725 ns 13.780 ns]
```

#### Mark if applicable


src/redisearch_rs/inverted_index/tests/raw_doc_ids_only.rs
```diff
@@ -0,0 +1,116 @@
...
+        (256, vec![0, 1, 0, 0]),
+        (65536, vec![0, 0, 1, 0]),
+        (u16::MAX as u32, vec![255, 255, 0, 0]),
+        (u32::MAX, vec![255, 255, 255, 255]),
+    ];
+    let doc_id = 4294967296;
+
+    for (delta, expected_encoding) in tests {
+        let mut buf = Cursor::new(Vec::new());
+        let record = RSIndexResult::term().doc_id(doc_id).frequency(1);
```
chesedo: We can use the default / virt type for these tests right? And then just use the C mocks above.

chesedo: Nvm, I see why we need `term`...

src/redisearch_rs/inverted_index/src/raw_doc_ids_only.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+impl Decoder for RawDocIdsOnly {
+    fn decode<'a>(
+        &self,
+        cursor: &mut Cursor<&'a [u8]>,
+        base: t_docId,
+    ) -> std::io::Result<RSIndexResult<'a>> {
+        let mut delta_bytes = [0u8; 4];
+        std::io::Read::read_exact(cursor, &mut delta_bytes)?;
```
GuyAv46, line 45: Is there no way to initially have `delta_bytes` as an undefined `[u8; 4]` (not initialized to 0)? Probably doesn't matter much, but since on a successful read we overwrite the entry array, will Rust optimize this?

src/redisearch_rs/inverted_index/src/raw_doc_ids_only.rs
```diff
@@ -0,0 +1,53 @@
...
+        let delta = u32::from_ne_bytes(delta_bytes);
+
+        let record = RSIndexResult::term()
+            .doc_id(base + delta as t_docId)
+            .frequency(1);
```
GuyAv46: I think this should be the default of `RSIndexResult::term()`, maybe in a future PR?

src/redisearch_rs/inverted_index/tests/raw_doc_ids_only.rs
```diff
@@ -0,0 +1,116 @@
...
+fn test_encode_raw_doc_ids_only() {
+    // Test cases for the raw doc ids encoder and decoder.
+    let tests = [
+        // (delta, expected encoding - raw 4-byte little-endian)
+        (0, vec![0, 0, 0, 0]),
+        (10, vec![10, 0, 0, 0]),
+        (256, vec![0, 1, 0, 0]),
+        (65536, vec![0, 0, 1, 0]),
+        (u16::MAX as u32, vec![255, 255, 0, 0]),
+        (u32::MAX, vec![255, 255, 255, 255]),
```
GuyAv46, line 52: Can we use `to_ne_bytes` for a more portable validation? Not that we really support big endian, so it's OK if not. Just wondering

src/redisearch_rs/inverted_index/src/raw_doc_ids_only.rs
gdesmott: Not really, Rust requires all values to be initialized before use for memory safety.

One option would be to use [MaybeUninit](https://doc.rust-lang.org/beta/std/mem/union.MaybeUninit.html) but that would make the code less safe for very few gain, if any as the compiler may optimize the init anyway.

src/redisearch_rs/inverted_index/src/raw_doc_ids_only.rs
gdesmott: That's a good idea. I added a commit for this.

src/redisearch_rs/inverted_index/tests/raw_doc_ids_only.rs
gdesmott: We could but I think that's just going to make tests more complicated for no real gain. 
At lot more work will likely be required if we ever want to support big endian, so I'd rather keep things simpler for now.

I had to use `to/from_ne_bytes`  as the code required an explicit conversion. But maybe we should use `le_bytes` instead so it's clearer that we only support LE?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -488,7 +488,6 @@ impl<'a> RSIndexResult<'a> {
                 num: ManuallyDrop::new(RSNumericRecord(0.0)),
             },
             result_type: RSResultType::Metric,
-            weight: 1.0,
```
GuyAv46: Why is this change needed?

gdesmott: It's not, I accidentally removed it. Good catch.

src/redisearch_rs/inverted_index/tests/raw_doc_ids_only.rs
GuyAv46: `le_bytes` will work on big-endian machines as well, right? I don't remember where exactly we broke the endianess generic. Maybe after we finish the migration, it will be solved, and we'll support big-endian again.
`ne_bytes` seems enough as long as we don't serialize the inverted index and expect to port the serialization between machines

GuyAv46: Anyway, let's not complicate the tests for no good reason

gdesmott: Sure it will. 

It's really just a matter if the binary format should be endian dependent or not. But as you said, as look as this is only used an internal local representation it does not really matter.

GuyAv46 (approved):

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -453,7 +454,7 @@ impl<'a> RSIndexResult<'a> {
             doc_id: 0,
             dmd: ptr::null(),
             field_mask: 0,
-            freq: 0,
+            freq: 1,
```
chesedo, line 457: This affects the frequency of everything since `default()` uses `virt()`

gdesmott: @GuyAv46 : is this ok for all result types to have `freq` defaulting to 1? 
Or should this be only for term results?

GuyAv46: Should be fine.
Where are we concerned about this value? For aggregation `RSAggregateResult`, the frequency is not used (and I'm not convinced that the summing we do even makes sense. Until recently, it wasn't even reset after each read).

gdesmott: I think it's fine as well.

Maybe we should consider moving those type specific files to the enum once https://github.com/RediSearch/RediSearch/pull/6615 is done.

PR #6612, opened 2025-08-07
MOD-10533: Sync RLookups - tail and upstreams 

Itzikvaknin:
### RLookup Synchronization:
- Use new `RLookup_CloneInto()` to copy upstream lookup into tail pipeline
- Aligns field indices (`dstidx`, `svidx`) between search subquery and tail pipeline

### Pipeline Refactoring:
- Moved `Pipeline_BuildAggregationPart` `case PLN_T_LOAD` logic into a separate function
- Reused this new function in both `Pipeline_BuildAggregationPart` and `HybridRequest_BuildPipeline`
- In `HybridRequest_BuildPipeline`, used a cloned load step for this logic

### Test Changes:
- Refactor `AddLoadStepToPlan` to create unprocessed LOAD steps (with `ArgsCursor` instead of pre-populated keys)
- Tests now verify that `BuildPipeline` correctly processes LOAD steps from unprocessed to processed state

### DIFF
--- a/src/aggregate/aggregate_plan.c
+++ b/src/aggregate/aggregate_plan.c
@@ -192,9 +192,8 @@ PLN_LoadStep *PLNLoadStep_Clone(const PLN_LoadStep *original) {
   // Pre-allocate keys array based on the number of arguments
   if (original->args.argc > 0) {
     cloned->keys = rm_calloc(original->args.argc, sizeof(RLookupKey*));
-  } else {
-    cloned->keys = NULL;
   }
+  // else - cloned->keys is already NULL
 
 
   return cloned;
--- a/src/hybrid/hybrid_request.c
+++ b/src/hybrid/hybrid_request.c
@@ -93,11 +93,13 @@ int HybridRequest_BuildPipeline(HybridRequest *req, const HybridPipelineParams *
         // Add the load step to the aggplan for proper cleanup
         AGPLN_AddStep(&areq->pipeline.ap, &subqueryLoadStep->base);
 
-        QueryError status = {0};
         // Process the LOAD step (explicit or implicit) using the unified function
         ResultProcessor *loader = processLoadStep(subqueryLoadStep, lookup, AREQ_SearchCtx(areq), AREQ_RequestFlags(areq),
-                                                 RLOOKUP_F_NOFLAGS, false, &areq->stateflags, &status);
-        if (status.code != QUERY_OK) {
+                                                 RLOOKUP_F_NOFLAGS, false, &areq->stateflags, &req->errors[i]);
+        if (req->errors[i].code != QUERY_OK) {
+            StrongRef_Release(sync_ref);
+            array_free(depleters);
+            // Note: HybridRequest_Free is called by the caller on failure
             return REDISMODULE_ERR;
         }
         if (loader) {
@@ -120,7 +122,7 @@ int HybridRequest_BuildPipeline(HybridRequest *req, const HybridPipelineParams *
     // and contain only keys from the loading step
     //Init lookup since we dont call buildQueryPart
     RLookup *lookup = AGPLN_GetLookup(&req->tailPipeline->ap, NULL, AGPLN_GETLOOKUP_FIRST);
-    RLookup_Init(lookup, NULL);
+    RLookup_Init(lookup, IndexSpec_GetSpecCache(params->aggregationParams.common.sctx->spec));
     RLookup_CloneInto(lookup, AGPLN_GetLookup(&req->requests[SEARCH_INDEX]->pipeline.ap, NULL, AGPLN_GETLOOKUP_FIRST));
 
 
--- a/src/rlookup.c
+++ b/src/rlookup.c
@@ -1005,17 +1005,17 @@ RLookupKey *RLookupKey_Clone(const RLookupKey *src) {
     dst->name = rm_strndup(src->name, src->name_len);
     dst->flags = src->flags | RLOOKUP_F_NAMEALLOC;  // Always set allocation flag
 
-    if (src->path && src->path != src->name) {
-      dst->path = rm_strdup(src->path);
-    } else if (src->path == src->name) {
-      dst->path = dst->name;  // Point to our allocated name
-    } else {
-      dst->path = NULL;
-    }
+    if (src->path) {
+      if (src->path == src->name) {
+        dst->path = dst->name;  // Point to our allocated name
+      } else {
+        dst->path = rm_strdup(src->path);
+      }
+    } //else - dst->path is already NULL
   } else {
     //source has no name
-    dst->name = NULL;
-    dst->path = NULL;
+    //dst->name = NULL;
+    //dst->path = NULL;
     dst->flags = src->flags;  // Don't set NAMEALLOC if no strings
   }
 


src/hybrid/hybrid_request.c
```diff
@@ -12,6 +12,33 @@ extern "C" {
...
+    implicitLoadStep->base.type = PLN_T_LOAD;
+    implicitLoadStep->base.alias = NULL;
+    implicitLoadStep->base.flags = 0;
+    implicitLoadStep->base.dtor = loadDtor; // Use standard destructor
+
+    // Create ArgsCursor with static array - no memory management needed
+    ArgsCursor_InitCString(&implicitLoadStep->args, implicitArgv, 1);
+
+    // Pre-allocate keys array for 1 key (same as handleLoad)
+    implicitLoadStep->nkeys = 0;
```
ofiryanai, line 38: shouldn't `nkeys` be 1? if it's 0 on purpose please explain it also in the comment on the line, currently I understand it should be 1

tests/cpptests/test_cpp_hybrid.cpp
```diff
@@ -283,7 +285,7 @@ TEST_F(HybridRequestTest, testHybridRequestPipelineBuildingBasic) {
   // Verify individual request pipeline structures
   // First request should have implicit scorer and sorter added
   // Actual order: DEPLETER -> LOADER -> SORTER -> SCORER -> INDEX
-  std::vector<ResultProcessorType> expectedFirstRequestPipeline = {RP_DEPLETER, RP_LOADER, RP_SORTER, RP_SCORER, RP_INDEX};
+  std::vector<ResultProcessorType> expectedFirstRequestPipeline = {RP_DEPLETER,RP_SORTER, RP_LOADER, RP_SCORER, RP_INDEX};
```
ofiryanai: Why did the order of the RPs change? I’m not sure how the Sorter operates, but since the loader batches all results before passing them downstream, it could hurt performance.
Also I think the comments are unnecessary, the RP list in the declaration of the variable is clear enough IMO. Please update the comment or delete it

ofiryanai (commented):
Looking good

src/hybrid/hybrid_request.c
Itzikvaknin: `nkeys` is the number of `RLookupKey`s that the step holds. 
Since this function creates an unprocessed loadstep, no `RLookupKey`s have been created yet, so it is zero.

The allocation of 1 (was changed to `implicitLoadStep->args.argc`) follows the load step handling already present in the codebase.

tests/cpptests/test_cpp_hybrid.cpp
Itzikvaknin: Reverted to `{RP_DEPLETER, RP_LOADER, RP_SORTER, RP_SCORER, RP_INDEX};`

src/hybrid/hybrid_request.c
```diff
@@ -84,11 +111,16 @@ int HybridRequest_BuildPipeline(HybridRequest *req, const HybridPipelineParams *
 
     // Release the sync reference as depleters now hold their own references
     StrongRef_Release(sync_ref);
+    //Init lookup since we dont call buildQueryPart
+    RLookup *lookup = AGPLN_GetLookup(&req->tailPipeline->ap, NULL, AGPLN_GETLOOKUP_FIRST);
+    RLookup_Init(lookup, NULL);
```
Itzikvaknin: Add `spcache`

src/rlookup.c
```diff
@@ -987,3 +987,81 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
...
+  dst->dstidx = src->dstidx;
+  dst->svidx = src->svidx;
+  dst->name_len = src->name_len;
+
+  // Cloned keys ALWAYS allocate their own strings for complete independence
+  if (src->name) {
+    dst->name = rm_strndup(src->name, src->name_len);
+    dst->flags = src->flags | RLOOKUP_F_NAMEALLOC;  // Always set allocation flag
+
+    if (src->path && src->path != src->name) {
```
Itzikvaknin: is `src->path != src->name` needed/wanted?
Follow `RLookupKey` creation, maybe share some logic with `RLookupKey` override flow

src/rlookup.c
```diff
@@ -987,3 +987,81 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
...
+    }
+  } else {
+    //source has no name
+    dst->name = NULL;
+    dst->path = NULL;
+    dst->flags = src->flags;  // Don't set NAMEALLOC if no strings
+  }
+
+  // Don't copy the next pointer - caller will handle linking
+  dst->next = NULL;
```
Itzikvaknin: `rm_calloc` sets it to `null`

src/rlookup.c
```diff
@@ -987,3 +987,81 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
...
+
+  return dst;
+}
+
+void RLookup_CloneInto(RLookup *dst, const RLookup *src) {
+  if (!dst || !src) {
```
Itzikvaknin: Assert

src/rlookup.c
```diff
@@ -987,3 +987,81 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
...
+    return;
+  }
+
+  // dst should be clean after RLookup_Init - only spcache should be set
+  // Assert that dst is in the expected clean state from initialization
+  RS_ASSERT(dst->head == NULL && dst->tail == NULL && dst->rowlen == 0 && dst->options == 0);
+
+
+  // Preserve dst's initialization state (spcache from init)
+  dst->spcache = dst->spcache;
```
Itzikvaknin: Thread-safe?
Increase refcount

src/rlookup.c
```diff
@@ -987,3 +987,81 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
...
+
+  // Reset dst structure but preserve initialization
+  dst->head = NULL;
+  dst->tail = NULL;
```
Itzikvaknin: Redundent

tests/cpptests/test_cpp_rlookup.cpp
```diff
@@ -71,3 +71,304 @@ TEST_F(RLookupTest, testRow) {
   RLookupRow_Cleanup(&rr);
   RLookup_Cleanup(&lk);
 }
+
```
Itzikvaknin, line 74: Add descriptive comment

src/aggregate/aggregate_plan.c
```diff
@@ -174,6 +174,46 @@ PLN_ArrangeStep *AGPLN_GetOrCreateArrangeStep(AGGPlan *pln) {
   return ret;
 }
 
+PLN_LoadStep *PLNLoadStep_Clone(const PLN_LoadStep *original) {
+  if (!original) {
+    return NULL;
+  }
```
Itzikvaknin: assert

src/aggregate/aggregate_plan.c
```diff
@@ -174,6 +174,46 @@ PLN_ArrangeStep *AGPLN_GetOrCreateArrangeStep(AGGPlan *pln) {
...
+    RS_LOG_ASSERT(original->keys, "original->keys is NULL");
+    // Processed: Clone the keys array
+    cloned->nkeys = original->nkeys;
+    cloned->keys = rm_calloc(original->nkeys, sizeof(RLookupKey*));
+    for (size_t i = 0; i < original->nkeys; i++) {
+      cloned->keys[i] = RLookupKey_Clone(original->keys[i]);
+    }
+    // Args might be invalid at this point, so don't copy them
+    memset(&cloned->args, 0, sizeof(ArgsCursor));
+  } else {
```
Itzikvaknin: Can remove it, we call clone only on unprocessed loadstep

src/aggregate/aggregate_plan.c
```diff
@@ -174,6 +174,46 @@ PLN_ArrangeStep *AGPLN_GetOrCreateArrangeStep(AGGPlan *pln) {
...
+    // Unprocessed: Clone the ArgsCursor and pre-allocate keys array
+    cloned->args = original->args; // Shallow copy of ArgsCursor
```
raz-mon: Can this shallow cause double-free?

raz-mon: Also, what about `argc`? The `build_pipeline` logic depends on it

src/pipeline/pipeline_construction.c
```diff
@@ -440,41 +519,15 @@ int Pipeline_BuildAggregationPart(Pipeline *pipeline, const AggregationPipelineP
...
-          RLookupKey *kk = RLookup_GetKey_LoadEx(curLookup, name, name_len, path, loadFlags);
-          // We only get a NULL return if the key already exists, which means
-          // that we don't need to retrieve it again.
-          if (kk) {
-            lstp->keys[lstp->nkeys++] = kk;
-          }
+        // Process the complete LOAD step
+        rp = processLoadStep(lstp, curLookup, params->common.sctx, params->common.reqflags,
+                            loadFlags, forceLoad, outStateFlags, status);
+        if (!rp && (lstp->nkeys || lstp->base.flags & PLN_F_LOAD_ALL)) {
```
raz-mon: Why is the second part of the condition needed?

src/aggregate/aggregate_plan.c
Itzikvaknin: `argc` is part of `args`.
It’s just an offset and a list of string pointers
From what I’ve seen in the codebase, `ArgsCursor ` users don’t take ownership of the values and don’t free them

src/rlookup.c
Itzikvaknin: I think its I think it clones the `RLookupKey` regardless of how it was created
If `src->path != src->name`, they are different, so both should be copied
If `src->path == src->name`, then path points to the same location as name (pointer equality), so we replicate that
Otherwise, in the empty case, it remains empty

src/pipeline/pipeline_construction.c
Itzikvaknin: It is not :)

src/hybrid/hybrid_request.c
```diff
@@ -93,9 +93,13 @@ int HybridRequest_BuildPipeline(HybridRequest *req, const HybridPipelineParams *
         // Add the load step to the aggplan for proper cleanup
         AGPLN_AddStep(&areq->pipeline.ap, &subqueryLoadStep->base);
 
+        QueryError status = {0};
```
raz-mon: Do we not already have a `QueryError` to populate?
I think this error will not be propagated properly to the error returning code.

src/rlookup.c
raz-mon: Count incremented?

src/rlookup.c
raz-mon: OK
Nitpick but would probably write it as follows:
```
if (src->path) {
  if (src->path != src->name) {
    ...
  } else {
  ...
  }
} else {
  ..
}
```

src/hybrid/hybrid_request.c
raz-mon: Up

src/aggregate/aggregate_plan.c
```diff
@@ -174,6 +174,32 @@ PLN_ArrangeStep *AGPLN_GetOrCreateArrangeStep(AGGPlan *pln) {
...
+  cloned->base.getLookup = original->base.getLookup;
+
+
+  cloned->args = original->args; // Shallow copy of ArgsCursor
+  cloned->nkeys = 0;
+  // Pre-allocate keys array based on the number of arguments
+  if (original->args.argc > 0) {
+    cloned->keys = rm_calloc(original->args.argc, sizeof(RLookupKey*));
+  } else {
+    cloned->keys = NULL;
```
raz-mon: Redundant (`rm_calloc`)

src/hybrid/hybrid_request.c
Itzikvaknin: Switched to `&req->errors[i]`

src/rlookup.c
Itzikvaknin: It’s a redundant line, removed
`RLookup_CloneInto` does not affect the initialization state, which is set by `RLookup_Init`

src/hybrid/hybrid_request.c
Itzikvaknin: Done

src/rlookup.c
```diff
@@ -987,3 +987,70 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
...
+      if (src->path == src->name) {
+        dst->path = dst->name;  // Point to our allocated name
+      } else {
+        dst->path = rm_strdup(src->path);
+      }
+    } //else - dst->path is already NULL
+  } else {
+    //source has no name
+    //dst->name = NULL;
+    //dst->path = NULL;
```
ofiryanai, line 1018: please remove commented out code

src/aggregate/aggregate_plan.c
```diff
@@ -174,6 +174,31 @@ PLN_ArrangeStep *AGPLN_GetOrCreateArrangeStep(AGGPlan *pln) {
...
+
+  cloned->args = original->args; // Shallow copy of ArgsCursor
+  cloned->nkeys = 0;
+  // Pre-allocate keys array based on the number of arguments
+  if (original->args.argc > 0) {
+    cloned->keys = rm_calloc(original->args.argc, sizeof(RLookupKey*));
+  }
+  // else - cloned->keys is already NULL
+
+
```
ofiryanai, line 198: not sure I understand the comment. please remove some extra newlines

Itzikvaknin: I thought `NULL`  by `rm_calloc` is too implicit, WDYT? same for the above ⬆️

raz-mon (approved):

PR #6613, opened 2025-08-07
Update vectors memory in total index info memory

alonre24:

As this field is a "best-effort" and non-accurate, we keep this track and add the vector index info to it.

#### Mark if applicable

### DIFF
--- a/src/info/indexes_info.c
+++ b/src/info/indexes_info.c
@@ -31,16 +31,21 @@ TotalIndexesInfo IndexesInfo_TotalInfo() {
     }
     // Lock for read
     pthread_rwlock_rdlock(&sp->rwlock);
-    size_t cur_mem = IndexSpec_TotalMemUsage(sp, 0, 0, 0);
-    info.total_mem += cur_mem;
-    if (info.min_mem > cur_mem) info.min_mem = cur_mem;
-    if (info.max_mem < cur_mem) info.max_mem = cur_mem;
-    info.indexing_time += sp->stats.totalIndexTime;
 
     // Vector indexes stats
     VectorIndexStats vec_info = IndexSpec_GetVectorIndexesStats(sp);
     info.fields_stats.total_vector_idx_mem += vec_info.memory;
     info.fields_stats.total_mark_deleted_vectors += vec_info.marked_deleted;
+
+    // Todo: today, this is a best effort does not account for the entire index memory accurately.
+    // In particular, it does not account for the vector index memory.
+    size_t cur_mem = IndexSpec_TotalMemUsage(sp, 0, 0, 0);
+    info.total_mem += cur_mem;
+    info.total_mem += vec_info.memory;
+
+    if (info.min_mem > cur_mem) info.min_mem = cur_mem;
+    if (info.max_mem < cur_mem) info.max_mem = cur_mem;
+    info.indexing_time += sp->stats.totalIndexTime;
 
     if (sp->gc) {
       ForkGCStats gcStats = ((ForkGC *)sp->gc->gcCtx)->stats;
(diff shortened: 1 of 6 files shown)


tests/pytests/test_info_modules.py
```diff
@@ -505,6 +505,8 @@ def test_redis_info_modules_vecsim():
   info = env.cmd('INFO', 'MODULES')
   field_infos = [to_dict(env.cmd(debug_cmd(), 'VECSIM_INFO', f'idx{i}', 'vec')) for i in range(1, 4)]
   env.assertEqual(info['search_used_memory_vector_index'], sum(field_info['MEMORY'] for field_info in field_infos))
+  # Validate that vector indexes are accounted in the total index memory
+  env.assertGreater(info['search_used_memory_indexes'], info['search_used_memory_vector_index'])
```
meiravgri, line 509: Consider elaborating the cpp tests to achieve a more accurate comparison
Refer to `test_cpp_llapi.cpp:testInfoSize`

src/info/indexes_info.c
```diff
@@ -31,17 +31,22 @@ TotalIndexesInfo IndexesInfo_TotalInfo() {
...
-    if (info.max_mem < cur_mem) info.max_mem = cur_mem;
-    info.indexing_time += sp->stats.totalIndexTime;
 
     // Vector indexes stats
     VectorIndexStats vec_info = IndexSpec_GetVectorIndexesStats(sp);
     info.fields_stats.total_vector_idx_mem += vec_info.memory;
     info.fields_stats.total_mark_deleted_vectors += vec_info.marked_deleted;
 
+    // Todo: today, this is a best effort does not account for the entire index memory accurately.
+    // In particular, it does not account for the vector index memory.
```
meiravgri: If you prefer not to include the vector index memory directly in `IndexSpec_TotalMemUsage`, I'd suggest refining the comment:
```suggestion
    // Currently, we don't track memory usage at allocation time. 
    // Instead, `IndexSpec_TotalMemUsage` provides an estimate by aggregating the memory usage of specific index components. 
// It does not include the memory used by the vector index, so we need to account for it manually here.
```

src/spec.h
```diff
@@ -678,6 +678,8 @@ size_t IndexSpec_collect_numeric_overhead(IndexSpec *sp);
 /**
  * @return all memory used by the index `sp`.
  * Uses the sizes of the doc-table, tag and text overhead if they are not `0`.
+ * TODO: fIX so this will account for the entire index memory, currently it is a best effort that
+ * account only for part of the actual memory.
```
meiravgri: i would say that the future task it get rid of this and have an allocator utility :)

tests/pytests/test_info_modules.py
alonre24: I think we can settle with that. Those LLAPI unit tests are a hassle, and we have good coverage of the vector allocator in vecsim...

meiravgri: So how can we cover IndexSpec_TotalMemUsage with vector_overhead=0?

meiravgri: also do we have a test that compares the `IndexSpec_TotalMemUsage` increase with vecsim allocator debug info ?

alonre24: > So how can we cover IndexSpec_TotalMemUsage with vector_overhead=0?

From LLAPI it is always passed with `vector_overhead=0`, but in the actual code paths, it is not expected to be zero. I could set it to be `ret+=vector_overhead` instead of using the ternary operator for that matter.


> also do we have a test that compares the `IndexSpec_TotalMemUsage` increase with vecsim allocator debug info ?

In this test, we compare the `search_used_memory_vector_index` to the vecsim alloactor info (line 507), and then we validate that the total index memory accounts for the vector fields memory, so it is covered in that sense.

meiravgri: IMO, Tests should cover all utilities (such as `IndexSpec_TotalMemUsage`) capabilities, not only those that are currently used, for future-proof purposes

meiravgri (approved):

PR #6614, opened 2025-08-07
Add SVS-VAMANA index to info fields telemetry [MOD-10104]

alonre24:
#### Mark if applicable

### DIFF
--- a/tests/pytests/test_info_modules.py
+++ b/tests/pytests/test_info_modules.py
@@ -501,12 +501,13 @@ def test_redis_info_modules_vecsim():
   env.expect('FT.CREATE', 'idx1', 'SCHEMA', 'vec', 'VECTOR', 'HNSW', '6', 'TYPE', 'FLOAT16', 'DIM', '2', 'DISTANCE_METRIC', 'L2').ok()
   env.expect('FT.CREATE', 'idx2', 'SCHEMA', 'vec', 'VECTOR', 'HNSW', '6', 'TYPE', 'FLOAT16', 'DIM', '2', 'DISTANCE_METRIC', 'L2').ok()
   env.expect('FT.CREATE', 'idx3', 'SCHEMA', 'vec', 'VECTOR', 'FLAT', '6', 'TYPE', 'FLOAT16', 'DIM', '2', 'DISTANCE_METRIC', 'L2').ok()
+  env.expect('FT.CREATE', 'idx4', 'SCHEMA', 'vec', 'VECTOR', 'SVS-VAMANA', '6', 'TYPE', 'FLOAT16', 'DIM', '2', 'DISTANCE_METRIC', 'L2').ok()
 
   set_doc().equal(1) # Add a document for the first time
   env.expect(debug_cmd(), 'WORKERS', 'DRAIN').ok()
 
   info = env.cmd('INFO', 'MODULES')
-  field_infos = [to_dict(env.cmd(debug_cmd(), 'VECSIM_INFO', f'idx{i}', 'vec')) for i in range(1, 4)]
+  field_infos = [to_dict(env.cmd(debug_cmd(), 'VECSIM_INFO', f'idx{i}', 'vec')) for i in range(1, 5)]
   env.assertEqual(info['search_used_memory_vector_index'], sum(field_info['MEMORY'] for field_info in field_infos))
   env.assertEqual(info['search_gc_marked_deleted_vectors'], 0)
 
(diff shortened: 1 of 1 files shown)


src/info/info_redis/info_redis.c
```diff
@@ -160,6 +160,13 @@ void AddToInfo_Fields(RedisModuleInfoCtx *ctx, TotalIndexesFieldsInfo *aggregate
       RedisModule_InfoAddFieldLongLong(ctx, "Flat", RSGlobalStats.fieldsStats.numVectorFieldsFlat);
     if (RSGlobalStats.fieldsStats.numVectorFieldsHNSW > 0)
       RedisModule_InfoAddFieldLongLong(ctx, "HNSW", RSGlobalStats.fieldsStats.numVectorFieldsHNSW);
+    if (RSGlobalStats.fieldsStats.numVectorFieldsSvsVamana > 0) {
+      RedisModule_InfoAddFieldLongLong(ctx, "SVS_VAMANA",
+                                       RSGlobalStats.fieldsStats.numVectorFieldsSvsVamana);
+      if (RSGlobalStats.fieldsStats.numVectorFieldsSvsVamanaCompressed > 0)
+        RedisModule_InfoAddFieldLongLong(ctx, "SVS_VAMANA_Compressed",
+                                         RSGlobalStats.fieldsStats.numVectorFieldsSvsVamanaCompressed);
+    }
```
meiravgri, line 169: I find it kinda confusing that SVS_VAMANA includes SVS_VAMANA_Compressed
I suggest to have SVS_VAMANA_NoCompression and SVS_VAMANA_Compressed
wdyt?

tests/pytests/test_info_modules.py
```diff
@@ -64,7 +67,7 @@ def testInfoModulesBasic(env):
   env.assertEqual(field_info_to_dict(fieldsInfo['search_fields_tag']), get_search_field_info('Tag', 2, Sortable=1, CaseSensitive=1))
   env.assertEqual(field_info_to_dict(fieldsInfo['search_fields_numeric']), get_search_field_info('Numeric', 2, NoIndex=1))
   env.assertEqual(field_info_to_dict(fieldsInfo['search_fields_geo']), get_search_field_info('Geo', 1))
-  env.assertEqual(field_info_to_dict(fieldsInfo['search_fields_vector']), get_search_field_info('Vector', 2, Flat=1, HNSW=1))
+  env.assertEqual(field_info_to_dict(fieldsInfo['search_fields_vector']), get_search_field_info('Vector', 4, Flat=1, HNSW=1, SVS_VAMANA=2, SVS_VAMANA_Compressed=1))
```
meiravgri, line 70: Do we have a test that also checks we count gc_marked_deleted_vectors and used_memory_vector_index correctly?
I suggest to incrementally add svs changes and making sure they take effect (for example, create an svs  index and ensure the `used_memory_vector_index` increased)

src/info/info_redis/info_redis.c
alonre24: I think that this aligned with the other fields stats that we have (for instance, we have fields_tag, and then field_tag_sortable, which is the subset of tags fields that are also sortable). This logic applies to SVS vector field and SVS vector which is also compressed.

tests/pytests/test_info_modules.py
alonre24: Actually, I don't think that we count the marked deleted vector of svs as part of the vector stats today. We should address that and confirm it here after we implement it in vecsim. Regarding the used memory of vector indexes - it should be accounted - I will add a test

src/info/info_redis/info_redis.c
meiravgri: ok

tests/pytests/test_info_modules.py
meiravgri: Yep, svs should implement Tombstone
```
template <typename DataType, typename DistType>
VecSimIndexStatsInfo VecSimTieredIndex<DataType, DistType>::statisticInfo() const {
    auto stats = VecSimIndexStatsInfo{
        .memory = this->getAllocationSize(),
        .numberOfMarkedDeleted = 0, // Default value if cast fails
    };

    // If backend implements VecSimIndexTombstone, get number of marked deleted
    if (auto tombstone = dynamic_cast<VecSimIndexTombstone *>(this->backendIndex)) {
        stats.numberOfMarkedDeleted = tombstone->getNumMarkedDeleted();
    }

    return stats;
}
```

meiravgri: or to implement `statisticInfo`. The number of marked deleted vectors is available via `changes_num` field of the svs index.

alonre24: Right
I opened a ticket for that

PR #6616, opened 2025-08-10
Avoid using SVS shared lib as default [MOD-10830] 

alonre24:

Avoid building against Intel's pre-compiled library for SVS by default; use the open source version instead. The pre-compiled shred lib that contains LVQ and LeanVec implementations is going to be used in tests and for building the module for RE only.

#### Mark if applicable

### DIFF
--- a/.github/workflows/task-test.yml
+++ b/.github/workflows/task-test.yml
@@ -182,6 +182,7 @@ jobs:
           SAN: ${{ inputs.san }}
           REDIS_VER: ${{ inputs.get-redis }}
           ENABLE_ASSERT: 1
+          SVS_SHARED_LIB: 1
         run: make build TESTS=1
       - name: Unit tests
         timeout-minutes: ${{ fromJSON(inputs.test-timeout) }}
--- a/.github/workflows/task-test.yml
+++ b/.github/workflows/task-test.yml
@@ -182,7 +182,7 @@ jobs:
           SAN: ${{ inputs.san }}
           REDIS_VER: ${{ inputs.get-redis }}
           ENABLE_ASSERT: 1
-          SVS_SHARED_LIB: 1
+          SVS_PRE_COMPILED_LIB: 1
         run: make build TESTS=1
       - name: Unit tests
         timeout-minutes: ${{ fromJSON(inputs.test-timeout) }}
--- a/build.sh
+++ b/build.sh
@@ -22,6 +22,7 @@ FORCE=0          # Force clean build flag
 VERBOSE=0        # Verbose output flag
 QUICK=${QUICK:-0} # Quick test mode (subset of tests)
 COV=${COV:-0}    # Coverage mode (for building and testing)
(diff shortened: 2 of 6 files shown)


alonre24:
Build artifacts - https://github.com/RediSearch/RediSearch/actions/runs/16874189116
In Linux x86 platforms we now see "SVS LVQ implementation found" for enterprise build and in OSS we see "SVS LVQ implementation not found" (meaning not using the shared pre-compiled lib)

build.sh
```diff
@@ -311,6 +315,10 @@ prepare_cmake_arguments() {
   if [[ "$OS_NAME" == "macos" ]]; then
     CMAKE_BASIC_ARGS="$CMAKE_BASIC_ARGS -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++"
   fi
+
+  if [[ "$SVS_SHARED_LIB" == "0" ]]; then
+    CMAKE_BASIC_ARGS="$CMAKE_BASIC_ARGS -DSVS_SHARED_LIB=OFF"
```
GuyAv46, line 320: Consider turning off by default in `CMakeLists.txt`. It makes no sense to have it off by default here but on by default in the CMake files. Let's not start the configuration spaghetti we just got rid of

alonre24: It is not ON by default in VecSim cmake, it is conditionally ON based on the platform (linux+x86+glibc>=2.26) and unless it was set externally.

build.sh
```diff
@@ -22,6 +22,7 @@ FORCE=0          # Force clean build flag
 VERBOSE=0        # Verbose output flag
 QUICK=${QUICK:-0} # Quick test mode (subset of tests)
 COV=${COV:-0}    # Coverage mode (for building and testing)
+SVS_SHARED_LIB=${SVS_SHARED_LIB:-0} # Build SVS as a shared library
```
meiravgri: we need to rename this flag in VecSim 
```suggestion
SVS_SHARED_LIB=${SVS_SHARED_LIB:-0} # Use pre-compiled SVS binary
```

.github/workflows/task-test.yml
```diff
@@ -182,6 +182,7 @@ jobs:
           SAN: ${{ inputs.san }}
           REDIS_VER: ${{ inputs.get-redis }}
           ENABLE_ASSERT: 1
+          SVS_SHARED_LIB: 1
```
meiravgri: I'm not comfortable with the fact that the CI never runs with the default value (SVS_SHARED_LIB=0)

build.sh
GuyAv46: Is it defined in the VecSim CMake?

GuyAv46 (dismissed):

.github/workflows/task-test.yml
alonre24: I am not either, but believe it is the best way to go as long as we have to choose one, and given that we have that CI does run with SVS_SHARED_LIB=0 for non x86+Linux platforms

meiravgri (approved):

PR #6620, opened 2025-08-11
[MOD-5935] Log DocTable capacity growth for memory diagnostics

lerman25:
Added a notice log in DocTable to record capacity increases:
This is meant to help diagnose a specific customer case where a small number of documents combined with a high volume of updates causes the DocTable size to grow disproportionately to the actual data stored (like in [MOD-5935](https://redislabs.atlassian.net/browse/MOD-5935)). The log will help confirm whether excessive capacity growth is contributing to the observed memory usage.

[MOD-5935]: https://redislabs.atlassian.net/browse/MOD-5935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/doc_table.c
```diff
@@ -132,6 +132,12 @@ static inline void DocTable_Set(DocTable *t, t_docId docId, RSDocumentMetadata *
 
     // We clear new extra allocation to Null all list pointers
     size_t memsetSize = (t->cap - oldcap) * sizeof(DMDChain);
+
+    // Log DocTable capacity growth to help diagnose cases where a small number of documents
+    // combined with frequent updates cause disproportionate memory usage.
+    // This allows us to confirm if unexpected memory spikes are due to capacity increases.
+    RedisModule_Log(RSDummyContext, "notice", "DocTable capacity increase from %zu to %zu", oldcap, t->cap);
```
alonre24: Nice
I would just add a comment about the reason WHY we are not shrinking (avoiding rehashing..), and refer to the relevant config if we do want to shrink

PR #6635, opened 2025-08-12
MOD-10835 Fix SORTBY 0 by not creating Arrange AggPlanStep

ofiryanai:
Previously Sortby 0 was implemented by adding Arrange AggStepPlan with `noSort` flag. This was buggy, instead we avoid creating this step so build pipeline don't try to create a RP.

Removed 2 unit tests that checked the flag set on parsing and behavior when its set.


src/aggregate/aggregate_request.c
```diff
@@ -287,9 +287,16 @@ static int handleCommonArgs(ParseAggPlanContext *papCtx, ArgsCursor *ac, QueryEr
...
     }
   } else if (AC_AdvanceIfMatch(ac, "SORTBY")) {
-    PLN_ArrangeStep *arng = AGPLN_GetOrCreateArrangeStep(papCtx->plan);
-    if ((parseSortby(arng, ac, status, papCtx)) != REDISMODULE_OK) {
-      return ARG_ERROR;
+    const char *firstArg;
+    bool isSortby0 = AC_GetString(ac, &firstArg, NULL, AC_F_NOADVANCE) == AC_OK
+                        && !strcmp(firstArg, "0");
+    if (isSortby0 && *papCtx->reqflags & QEXEC_F_IS_HYBRID_TAIL) {
+      AC_Advance(ac);  // Advance without adding SortBy step to the plan
```
raz-mon, line 294: I don't think we want this for the existing commands, nor for the branches of the hybrid command.
Let's enable this for the tail of `FT.HYBRID` only

oshadmi:
> I'm wondering if we want to have this in phase 1. It introduces a relatively weird behavior in which we sort in the branches but not in the tail, which doesn't seem useful to me at this point. It's better to add stuff when we're sure they're useful and needed, since afterwards it's breaking to remove.

@raz-mon At least for RRF it makes sense to sort branches (sub-queries) since it will affect the ranking
And not necessarily sort the tail and return up to WINDOW results in some arbitrary order.

src/aggregate/aggregate_request.c
ofiryanai: That's exactly the second part of the condition in the `if`

src/aggregate/aggregate_request.c
```diff
@@ -287,9 +287,16 @@ static int handleCommonArgs(ParseAggPlanContext *papCtx, ArgsCursor *ac, QueryEr
       return ARG_ERROR;
     }
   } else if (AC_AdvanceIfMatch(ac, "SORTBY")) {
-    PLN_ArrangeStep *arng = AGPLN_GetOrCreateArrangeStep(papCtx->plan);
-    if ((parseSortby(arng, ac, status, papCtx)) != REDISMODULE_OK) {
-      return ARG_ERROR;
+    const char *firstArg;
+    bool isSortby0 = AC_GetString(ac, &firstArg, NULL, AC_F_NOADVANCE) == AC_OK
+                        && !strcmp(firstArg, "0");
+    if (isSortby0 && *papCtx->reqflags & QEXEC_F_IS_HYBRID_TAIL) {
```
Itzikvaknin, line 293:
```suggestion
    if (isSortby0 && *papCtx->reqflags & IsHybridTail(papCtx)) {
```

ofiryanai: not working, because reqflags is `uint32_t*` in`papCtx` instead of direct `uint32_t`. tried to change it in previous PR and reverted from all the bugs

PR #6636, opened 2025-08-13
Fix "has map" performance - [MOD-9687]

GuyAv46:
## Describe the changes in the pull request

Fix "has map" performance by using the cached boolean value `resp3` of the reply, instead of checking with the redis API for every serialized result.

The solution assumes redis version >= 7 (which is true for `2.8` and later)
The macros replaced are:
```
#define _ReplyMap(ctx) (RedisModule_ReplyWithMap != NULL && is_resp3(ctx))
#define _ReplySet(ctx) (RedisModule_ReplyWithSet != NULL && is_resp3(ctx))
```
But since `2.8` and later versions reject redis < 7.1, the API is available, so they can be reduced to the `is_resp3(ctx)` check.

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #6637, opened 2025-08-13
More resp3 cleanup

GuyAv46:

Inhancement of #6636, but more aggressive, so it won't be backported.

#### Mark if applicable


GuyAv46:
Some unused variables may be in use in other branches (didn't check).
The crucial performance issue was solved in the minimal PR, so this one is more so we don't use this API in future features

PR #6640, opened 2025-08-13
[MOD-10845] Improve the lifetimes of `RSIndexResult`

chesedo:
This gives the lifetimes better names and makes a separate lifetime for the aggregate children. This work makes the `Ref` and `Owned` structs needed for MOD-10746 easier

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
+++ b/src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
@@ -78,9 +78,12 @@ pub unsafe extern "C" fn IndexResult_SetNumValue(result: *mut RSIndexResult, val
 /// The following invariant must be upheld when calling this function:
 /// - `result` must point to a valid `RSIndexResult` and cannot be NULL.
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn IndexResult_TermRef(
-    result: *const RSIndexResult<'_>,
-) -> Option<&RSTermRecord<'_>> {
+pub unsafe extern "C" fn IndexResult_TermRef<'index, 'aggregate_children>(
+    result: *const RSIndexResult<'index, 'aggregate_children>,
+) -> Option<&'index RSTermRecord<'index>>
+where
+    'aggregate_children: 'index,
+{
     debug_assert!(!result.is_null(), "result must not be null");
 
     // SAFETY: Caller is to ensure that the pointer `result` is a valid, non-null pointer to
@@ -98,9 +101,12 @@ pub unsafe extern "C" fn IndexResult_TermRef(
 /// The following invariant must be upheld when calling this function:
 /// - `result` must point to a valid `RSIndexResult` and cannot be NULL.
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn IndexResult_TermRefMut(
-    result: *mut RSIndexResult<'_>,
-) -> Option<&mut RSTermRecord<'_>> {
+pub unsafe extern "C" fn IndexResult_TermRefMut<'index, 'aggregate_children>(
+    result: *mut RSIndexResult<'index, 'aggregate_children>,
+) -> Option<&'index mut RSTermRecord<'index>>
+where
+    'aggregate_children: 'index,
+{
     debug_assert!(!result.is_null(), "result must not be null");
 
     // SAFETY: Caller is to ensure that the pointer `result` is a valid, non-null pointer to
@@ -118,9 +124,9 @@ pub unsafe extern "C" fn IndexResult_TermRefMut(
 /// The following invariant must be upheld when calling this function:
 /// - `result` must point to a valid `RSIndexResult` and cannot be NULL.
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn IndexResult_AggregateRef(
-    result: *const RSIndexResult<'_>,
-) -> Option<&RSAggregateResult<'_>> {
+pub unsafe extern "C" fn IndexResult_AggregateRef<'index, 'aggregate_children>(
+    result: *const RSIndexResult<'index, 'aggregate_children>,
+) -> Option<&'index RSAggregateResult<'index, 'aggregate_children>> {
     debug_assert!(!result.is_null(), "result must not be null");
 
     // SAFETY: Caller is to ensure that the pointer `result` is a valid, non-null pointer to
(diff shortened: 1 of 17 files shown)


src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -398,17 +403,17 @@ pub struct RSVirtualResult;
...
-    pub agg: ManuallyDrop<RSAggregateResult<'a>>,
-    pub term: ManuallyDrop<RSTermRecord<'a>>,
+pub union RSIndexResultData<'index, 'aggregate_children> {
+    pub agg: ManuallyDrop<RSAggregateResult<'index, 'aggregate_children>>,
+    pub term: ManuallyDrop<RSTermRecord<'index>>,
     pub num: ManuallyDrop<RSNumericRecord>,
     pub virt: ManuallyDrop<RSVirtualResult>,
 }
 
 /// The result of an inverted index
```
gdesmott: Worth documenting the exact semantic of both lifetimes here.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -241,22 +244,22 @@ pub type RSResultTypeMask = BitFlags<RSResultType, u32>;
...
     /// own `LowMemoryThinVec` type which is `#[repr(C)]` and has a known size instead.
-    records: LowMemoryThinVec<*const RSIndexResult<'a>>,
+    records: LowMemoryThinVec<*const RSIndexResult<'index, 'children>>,
 
     /// A map of the aggregate type of the underlying records
     type_mask: RSResultTypeMask,
     /// The lifetime is actually on `RsIndexResult` but it is stored as a pointer which does not
     /// support lifetimes. So use a PhantomData to carry the lifetime for it instead.
-    _phantom: PhantomData<&'a ()>,
+    _phantom: PhantomData<&'children ()>,
```
gdesmott, line 245: It's unclear to me why we need this for `'children` but not `'index`?

gdesmott: I asked Claude Code who provided a helpful explanation! FYI

```
 pub fn iter(&'index self) -> RSAggregateResultIter<'index, 'children>

  The 'index lifetime is actually used directly in the struct's methods! When you call iter() on an RSAggregateResult, it borrows self for the 'index lifetime
  and returns an iterator that lives for 'index.

  This means:
  - 'index is genuinely used by the struct through its iter() method signature
  - 'children is only used in the type parameters of the pointer content, but since pointers don't carry lifetime info, it needs PhantomData

  So the 'index lifetime is "consumed" by the method signature &'index self, while 'children has no direct usage in any field or method signature of the struct
  itself - it only appears in the generic parameters of the pointed-to type, which doesn't count for lifetime parameter usage checking.

  This is a subtle but important distinction in Rust's lifetime system!
  ```

File: src/redisearch_rs/inverted_index/src/lib.rs (hunk 1 above)
Comment (chesedo) [reply src/redisearch_rs/inverted_index/src/lib.rs]: Done in https://github.com/RediSearch/RediSearch/pull/6640/commits/865e5cfa3b70b45764788e55cb60f84aa662db0a

File: src/redisearch_rs/inverted_index/src/lib.rs (hunk 2 above)
Comment (chesedo) [reply src/redisearch_rs/inverted_index/src/lib.rs:245]: Done in https://github.com/RediSearch/RediSearch/pull/6640/commits/865e5cfa3b70b45764788e55cb60f84aa662db0a

File: src/redisearch_rs/inverted_index/src/lib.rs [hunk 3]
```diff
@@ -397,18 +404,22 @@ impl<'a> IntoIterator for RSAggregateResult<'a> {
 pub struct RSVirtualResult;
 
 /// Holds the actual data of an ['IndexResult']
+///
+/// The `'index` lifetime is linked to the [`IndexBlock`] when decoding borrows from the block.
+/// While the `'aggregate_children` lifetime is linked to [`RSAggregateResult`]s that is holding
```
```suggestion
/// While the `'aggregate_children` lifetime is linked to [`RSAggregateResult`] that is holding
```

gdesmott (approved):

gdesmott (dismissed):

src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
```diff
@@ -78,9 +78,12 @@ pub unsafe extern "C" fn IndexResult_SetNumValue(result: *mut RSIndexResult, val
 /// The following invariant must be upheld when calling this function:
 /// - `result` must point to a valid `RSIndexResult` and cannot be NULL.
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn IndexResult_TermRef(
-    result: *const RSIndexResult<'_>,
-) -> Option<&RSTermRecord<'_>> {
+pub unsafe extern "C" fn IndexResult_TermRef<'index, 'aggregate_children>(
+    result: *const RSIndexResult<'index, 'aggregate_children>,
+) -> Option<&'index RSTermRecord<'index>>
```
LukeMathWalker: This isn't quite correct. The lifetime of the returned reference is tied to the lifetime of `result`.
Morally, this takes as input a `&'a RSIndexResult<'index, 'aggregate_children>` and returns a `&'a RSTermRecord<'index>`, where `'index` must outlive `'a`.

I think it'd be easier, at this stage, to return a `*const RSTermRecord<'index>` or `Option<NonNull<RSTermRecord<'index>>>`.

LukeMathWalker: Same logic applies to a few other places in this FFI layer.

LukeMathWalker: The correct version of this, with lifetimes, would be:

```rust
#[unsafe(no_mangle)]
pub unsafe extern "C" fn IndexResult_TermRef<'result, 'index, 'aggregate_children>(
    result: *const RSIndexResult<'index, 'aggregate_children>,
) -> Option<&'result RSTermRecord<'index>>
where
    'aggregate_children: 'result,
    'index: 'result
{
    debug_assert!(!result.is_null(), "result must not be null");

    // SAFETY: Caller is to ensure that the pointer `result` is a valid, non-null pointer to
    // an `RSIndexResult`.
    let result: &'result RSIndexResult<'index, 'aggregate_children> = unsafe { &*result };

    result.as_term()
}
```

LukeMathWalker: If we choose to return a raw pointer,

```rust
#[unsafe(no_mangle)]
pub unsafe extern "C" fn IndexResult_TermRef<'result, 'index, 'aggregate_children>(
    result: *const RSIndexResult<'index, 'aggregate_children>,
) -> *const RSTermRecord<'index>
where
    'aggregate_children: 'result,
    'index: 'result,
{
    debug_assert!(!result.is_null(), "result must not be null");

    // SAFETY: Caller is to ensure that the pointer `result` is a valid, non-null pointer to
    // an `RSIndexResult`.
    let result: &'result RSIndexResult<'index, 'aggregate_children> = unsafe { &*result };

    result.as_term().map_or(std::ptr::null(), |r| r as *const _)
}
```

chesedo: Ahhh nice, I realized this when I made the change, but was not aware how I would link a lifetime to the `*const` result. Changes in https://github.com/RediSearch/RediSearch/pull/6640/commits/5be05454b4d4301bb6a6dbaca29e8ef75cee7951

meiravgri (approved):

PR #6651, opened 2025-08-13
MOD-10748: Fix ACLUserMayAccessIndex

nafraf:

1. Current: 
   In Redis, the "master" client (such as replication) may not have an associated user, and `RedisModule_GetCurrentUserName()` will return NULL in such cases.
   Redisearch was not validating the previous case.
2. Change: 
   Allow access in ACLUserMayAccessIndex when no user is associated
   Similar approach was used by RedisTimeSeries in [CheckKeyIsAllowedByAcls](https://github.com/RedisTimeSeries/RedisTimeSeries/blob/e3acedeac7a202aa769f1b95bb69b21ded32194a/src/module.h#L37
)
3. Outcome:

#### Mark if applicable

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -113,6 +113,15 @@ bool ACLUserMayAccessIndex(RedisModuleCtx *ctx, IndexSpec *sp) {
     return true;
   }
   RedisModuleString *user_name = RedisModule_GetCurrentUserName(ctx);
+
+  if (!user_name) {
+    // In Redis, the "master" client (such as replication or internal server
+    // operations) may not have an associated user, and
+    // RedisModule_GetCurrentUserName will return NULL in such cases.
+    // If we are in a context without a user_name, allow access.
+    return true;
+  }
+
   RedisModuleUser *user = RedisModule_GetModuleUserFromUserName(user_name);
 
   if (!user) {
--- a/src/module.c
+++ b/src/module.c
@@ -118,14 +118,14 @@ bool ACLUserMayAccessIndex(RedisModuleCtx *ctx, IndexSpec *sp) {
     // In Redis, the "master" client (such as replication or internal server
     // operations) may not have an associated user, and
     // RedisModule_GetCurrentUserName will return NULL in such cases.
-    // If we are in a context without a user_name, allow access.
+    // We thus allow full access to the super-user.
     return true;
   }
 
   RedisModuleUser *user = RedisModule_GetModuleUserFromUserName(user_name);
 
   if (!user) {
-    RedisModule_Log(ctx, "verbose", "No user found");
+    RedisModule_Log(ctx, "warning", "No user found for current client");
     RedisModule_FreeString(ctx, user_name);
     return false;
   }


src/module.c
```diff
@@ -113,6 +113,15 @@ bool ACLUserMayAccessIndex(RedisModuleCtx *ctx, IndexSpec *sp) {
     return true;
   }
   RedisModuleString *user_name = RedisModule_GetCurrentUserName(ctx);
+
+  if (!user_name) {
+    // In Redis, the "master" client (such as replication or internal server
+    // operations) may not have an associated user, and
+    // RedisModule_GetCurrentUserName will return NULL in such cases.
+    // If we are in a context without a user_name, allow access.
+    return true;
```
GuyAv46, line 122: Is this the only reason to not have a user? Feels like a backdoor.
We later return false if there is no user by this user name, how is this different?
Can we check in another way that we have a “master client”? Maybe redis can simply add an allowed user to this client?

raz-mon: From some checks we did we found that when there is no user, that means the "super-user".
For instance - https://github.com/redis/redis/blob/eac48279ad21b8612038953fefa0dcf926773efc/src/acl.c#L2452.
This is also the logic [here](https://github.com/redis/redis/blob/eac48279ad21b8612038953fefa0dcf926773efc/src/acl.c#L3226-L3227).
@oranagra WDYT?

raz-mon (approved):

src/module.c
oranagra: i agree, here's another evidence:
https://github.com/redis/redis/blob/eac48279ad21b8612038953fefa0dcf926773efc/src/acl.c#L1859
```c
    /* If there is no associated user, the connection can run anything. */
    if (u == NULL) return ACL_OK;
```
used by `RM_ACLCheckCommandPermissions`

it's a shame that `RM_GetModuleUserFromUserName` returns NULL when it doesn't find the requested user, or does't have a case that an empty string maps to a valid `RedisModuleUser` instance with a null `user` member. then everything would have worked?

in any case when `RM_GetCurrentUserName` returns NULL
```
    if (ctx->client == NULL || ctx->client->user == NULL || ctx->client->user->name == NULL) {
        return NULL;
    }
```
it means that either the context is detached from a client, and thus should not be limited by any user of the current client, or **the current client is without a user, which means it's the super user**.
i think it's ok to assume that when it returns NULL, we're unrestricted.

oshadmi, line 122: How can we test this? Can we use `DEBUG` command with `MARK-INTERNAL-CLIENT` that uses super user

raz-mon:
Reproduced by sending a write command on Enterprise on an existing index:
```
FT.CREATE idx SCHEMA t TEXT
FT.ALTER idx SCHEMA ADD t2 TEXT
```
@oshadmi This will be tested once we run our flow-tests on RE, something that is long due and planned for the short term AFAIK. For now tested manually.

src/module.c
raz-mon: See below comment.

PR #6678, opened 2025-08-20
[MOD-10922] Port Empty iterator to Rust

gdesmott:
First iterator ported to Rust with all the boilerplate for tests and benchmarking.

No point to actually compare benchmarks it as this iterator is basically a no-op but here it is anyway:

```
Iterator - Empty - Read/C    time:   [2.5167 ns 2.5183 ns 2.5201 ns]
Iterator - Empty - Read/Rust time:   [22.430 ps 22.491 ps 22.561 ps]

Iterator - Empty - SkipTo/c    time:   [2.3584 ns 2.3602 ns 2.3626 ns]
Iterator - Empty - SkipTo/Rust time:   [22.525 ps 22.568 ps 22.615 ps]
```

#### Mark if applicable


src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
chesedo, line 1: I'm not quite convinced with this file name. Won't it make more sense to have the interface in `lib.rs`?

src/redisearch_rs/rqe_iterators_bencher/build.rs
```diff
@@ -0,0 +1,28 @@
...
+    // Generate C bindings - fail build if this doesn't work
+    let root = git_root().expect("Could not find git root");
+    let headers = ["iterator_api.h", "empty_iterator.h"]
+        .iter()
+        .map(|h| root.join("src").join("iterators").join(h))
+        .collect::<Vec<_>>();
+    generate_c_bindings(headers, ".*/iterators/.*.h")?;
+
+    Ok(())
+}
```
chesedo, line 28: I love how simple this is now :star_struck:

src/redisearch_rs/rqe_iterators/src/rqe_iterator.rs
gdesmott: Agreed, I moved it.

chesedo (dismissed):

src/redisearch_rs/rqe_iterators_bencher/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+# Statically link to the libclang on aarch64-unknown-linux-musl,
+# necessary on Alpine.
+# See https://github.com/rust-lang/rust-bindgen/issues/2360
+features = ["bindgen-static", "min-redis-compatibility-version-6-0"]
+workspace = true
+default-features = false
+
+[target.'cfg(not(all(target_env="musl", target_os="linux")))'.dependencies.redis-module]
+workspace = true
+default-features = true
```
LukeMathWalker: Can this blurb be moved into the workspace Cargo.toml, so we can just `.workspace = true` in each crate rather than having to copy-paste it?

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -7,4 +7,77 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+/// The status of the iterator after a call to `revalidate`
+pub enum RQEValidateStatus {
+    /// The iterator is still valid and at the same position.
+    Ok,
+    /// The iterator is still valid but its internal state has changed.
+    Moved,
+    /// The iterator is no longer valid, and should not be used or should be rewound.
```
GuyAv46: Incorrect translation of the original comment from `iterator_api.h`
```suggestion
    /// The iterator is no longer valid, and should not be used or rewound. Should be freed.
```

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -7,4 +7,77 @@
...
+    Aborted,
+}
+
+pub trait RQEIterator {
+    /// Read the next entry from the iterator.
+    ///
+    /// On a successful read, the iterator must set its `last_doc_id` property to the new current result id
+    /// This function returns Ok with the current result for valid results, or None if the iterator is depleted.
+    /// The function will return Err(RQEIteratorError) for any error.
+    fn read(&mut self) -> Result<Option<RSIndexResult<'_, '_>>, RQEIteratorError>;
```
GuyAv46, line 47: We should consider handling EOF as an error, so the happy path always has a result. WDYT?

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -7,4 +7,77 @@
...
+    /// Returns an upper-bound estimation for the number of results the iterator is going to yield.
+    fn num_estimated(&self) -> usize;
+
+    /**************** properties ****************/
+
+    /// Returns the last doc id that was read or skipped to.
+    fn last_doc_id(&self) -> t_docId;
+
+    /// Returns true if the iterator has more results to read, meaning, not at EOF.
+    fn has_next(&self) -> bool;
```
GuyAv46: Rant:
I don't like that we changed `atEOF` to `has_next` after we just made this change in the C refactor. `has_next` is ambiguous, as we don't know why we don't have next if we return false (and in the past it was used in a few unexpected ways after timeout, or if the iterator has aborted)

src/redisearch_rs/rqe_iterators_bencher/benches/iterators.rs
```diff
@@ -0,0 +1,22 @@
...
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! Benchmark iterators
+
+use criterion::{Criterion, criterion_group, criterion_main};
+use rqe_iterators_bencher::benchers;
+
+fn benchmark_empty(c: &mut Criterion) {
```
GuyAv46, line 15: Is this file the long-term benchmark suite, after we remove the C implementation?

src/redisearch_rs/rqe_iterators/src/lib.rs
gdesmott: Ack. I phrased `Should be dropped.` as we no longer have an explicit free.

src/redisearch_rs/rqe_iterators/src/lib.rs
gdesmott: I don't know why it's been changed, I've not been involved in the design of this first API.

Do you want to change it yourself? That will give you a chance to do an easy first PR on the Rust code before starting to work on the IDs iterator. :)

src/redisearch_rs/rqe_iterators_bencher/benches/iterators.rs
gdesmott: Not this file as each iterator will  have its own in the `benches/` directory but yeah, all benchmarks will live in this crate.

src/redisearch_rs/rqe_iterators/src/lib.rs
GuyAv46: Since this is the first actual PR, let's rename it now, actually

src/redisearch_rs/rqe_iterators/src/lib.rs
gdesmott: I guess this API has been designed to mimic [FallibleIterator](https://docs.rs/fallible-iterator/latest/fallible_iterator/trait.FallibleIterator.html). 

@LukeMathWalker : was it?

LukeMathWalker: Yup. And, more generally, the way the iterator trait works in Rust—https://doc.rust-lang.org/std/iter/trait.Iterator.html#tymethod.next. It's idiomatic to use `None` to signal the end of the iteration.

GuyAv46: But in the simple iterator, it returns an option and not a result wrapping an option. In our case, where we have a few different errors (EOF and timeout at the moment), it feels more intuitive to handle the happy path with `OK(RSIndexResult)` and `Err(why no result)` in any other case.
On the other hand, unwrapping EOF with `?` as an error doesn't make sense (at least in some cases like union iterator), so let's keep it this way

LukeMathWalker: Yeah, the `Result` is due to our iterator being fallible, which goes back to the signature of `FallibleIterator::next` that @gdesmott was linking above: https://docs.rs/fallible-iterator/latest/fallible_iterator/trait.FallibleIterator.html#tymethod.next

src/redisearch_rs/rqe_iterators_bencher/Cargo.toml
gdesmott: Actually we do not need this here. 
Neither in the II bencher: https://github.com/RediSearch/RediSearch/pull/6686

src/redisearch_rs/rqe_iterators/src/empty.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    fn rewind(&mut self) {}
+
+    fn num_estimated(&self) -> usize {
+        0
+    }
+
+    fn last_doc_id(&self) -> t_docId {
+        0
+    }
```
LukeMathWalker, line 42: I do wonder if modelling the return type as `Option<t_docId>` would be more correct?

LukeMathWalker: But it's likely to worsen performance, so better to hold off for now.

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -7,4 +7,77 @@
...
+    /// Returns an upper-bound estimation for the number of results the iterator is going to yield.
+    fn num_estimated(&self) -> usize;
+
+    /**************** properties ****************/
+
+    /// Returns the last doc id that was read or skipped to.
+    fn last_doc_id(&self) -> t_docId;
+
+    /// Returns true if the iterator has more results to read, meaning, not at EOF.
+    fn at_eof(&self) -> bool;
```
LukeMathWalker: This feels backwards to me. If you return `true` to `at_eof` it reads to me as if you're _at_ the end of the file, hence no more entries to be returned.

gdesmott: Good point.
I renamed it from `has_next()` but did not flip the doc and implementation. Fixed.

GuyAv46 (approved):

LukeMathWalker (dismissed):

chesedo (approved):

PR #6680, opened 2025-08-20
ii: add reader constructor from an InvertedIndex

gdesmott:
InvertedIndex.blocks is private so there is no way for user code to create the reader itself.

#### Mark if applicable


src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -1046,6 +1046,11 @@ impl<E: Encoder> InvertedIndex<E> {
             )
         }
     }
+
+    /// Create a new [`IndexReader`] for this inverted index.
+    pub fn reader<D: Decoder>(&self, decoder: D) -> IndexReader<'_, D> {
```
gdesmott: the `decoder` is a bit redudant here as we already have the encoder.

@chesedo : IIRC we discussed at some point to have API mapping encoder <-> decoder. Maybe we should add that?

chesedo: Yeah, this should use the `encoder` to figure out what the `decoder` should be

gdesmott: Something like this then?

chesedo (approved):

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Yes, this will work. We can figure out how to give it filter arguments later

PR #6682, opened 2025-08-21
[MOD-10746] Make `Borrowed` and `Owned` variants for the term and aggregate types

chesedo:
This is an idiomatic Rust change to rather use a boolean field to know if a variable is copy or not. Doing this makes the drop (memory free) code easier and will also make the deep copy code easier.

#### Mark if applicable


src/redisearch_rs/inverted_index/src/index_result.rs
```diff
@@ -98,28 +92,66 @@ impl RSOffsetVector<'_> {
...
+}
+
+impl Drop for RSTermRecord<'_> {
+    fn drop(&mut self) {
+        match self {
+            RSTermRecord::Borrowed { term, .. } => {
+                if !term.is_null() {
+                    // SAFETY: we know the C type is the same because the Rust type is autogenerated from it
+                    unsafe {
+                        Term_Free(*term);
```
gdesmott, line 152: I keep being confused why the term is free in the borrowed case. May be worth properly documenting ownership in the enum definition.

src/redisearch_rs/c_entrypoint/types_ffi/cbindgen.toml
```diff
@@ -33,3 +33,4 @@ exclude = ["Header"]
 
 [export.rename]
 "LowMemoryThinVec______RSIndexResult" = "LowMemoryThinVecRSIndexResult"
+"LowMemoryThinVec_____RSIndexResult" = "LowMemoryThinVecRSIndexResultOwned"
```
gdesmott, line 39: It's unclear to me what is the difference between `LowMemoryThinVec______RSIndexResult` and `LowMemoryThinVec_____RSIndexResult`?  (one less `_`).

src/redisearch_rs/inverted_index/src/index_result.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6682/commits/a6bd6f8ed25c165d8bdc242181b9cb23b739a6f0

src/redisearch_rs/c_entrypoint/types_ffi/cbindgen.toml
chesedo: cbindgen made a new version of this because an extra `LowMemoryThinVec` was introduced. I added a comment in https://github.com/RediSearch/RediSearch/pull/6682/commits/a6bd6f8ed25c165d8bdc242181b9cb23b739a6f0

src/redisearch_rs/inverted_index/src/index_result.rs
```diff
@@ -193,62 +263,77 @@ pub type RSResultKindMask = BitFlags<RSResultKind, u8>;
...
+#[repr(u8)]
+#[derive(Debug, PartialEq)]
+pub enum RSAggregateResult<'index> {
+    Borrowed {
+        /// The records making up this aggregate result
+        ///
+        /// The `RSAggregateResult` is part of a union in [`RSResultData`], so it needs to have a
+        /// known size. The std `Vec` won't have this since it is not `#[repr(C)]`, so we use our
+        /// own `LowMemoryThinVec` type which is `#[repr(C)]` and has a known size instead.
+        records: LowMemoryThinVec<&'index RSIndexResult<'index>>,
```
gdesmott, line 282: Shouldn't this be records: LowMemoryThinVec<&'children RSIndexResult<'index>>, ?

We no longer need 2 different lifetimes?

src/redisearch_rs/inverted_index/src/index_result.rs
gdesmott: What's confusing me is why are we calling `Term_Free()` in the borrowed case? We should document exactly what is being owned there.

src/redisearch_rs/inverted_index/src/index_result.rs
chesedo: No, this is what caused that `'static` lifetime issue that other day. So `&'children RSIndexResult<'index, 'children>` will implicitly require `'index` to life longer than `'children` (ie `where 'index: 'children`) else you would have a dangling pointer access if `'index` was cleaned up but the child still tried to access it.

So in that `'static` lifetime issue `next_record()` would return `RSIndexResult<'index, 'static>` because the decoders will never have any children. Because of the implicit requirement that meant `'index` had to be `'static` hence why we had that compile error.

gdesmott: Right, that makes sense. Very tricky, nice catch!

Probably worth adding a comment there explaining why we use `'index` for the result reference then as it's really not that obvious. :)

gdesmott (commented):
Is valgrind happy? :)

src/redisearch_rs/inverted_index/src/index_result.rs
chesedo: Yeah, Luca actually found this. I'll see how to comment this

src/redisearch_rs/inverted_index/src/index_result.rs
gdesmott: Thanks, it's much clearer now :+1:

chesedo: I expanded the comments more in https://github.com/RediSearch/RediSearch/pull/6682/commits/73aa705d12d1ab9f5dfe1a4c10972de7d832576f

This is starting to become clear that `Borrowed` actually only means borrowing the offsets (even for the aggregate) and that everything else is owned. I'm thinking we should be able to find a better enum variant name here. Maybe `BorrowedOffsets`?

chesedo:
> Is valgrind happy? :)

Yes :partying_face:

src/redisearch_rs/inverted_index/src/index_result.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6682/commits/4fda2a2781eebcdd36e251b6424aa9a033dce811

gdesmott (dismissed):

src/iterators/hybrid_reader.c
```diff
@@ -92,7 +92,7 @@ static void insertResultToHeap_Aggregate(HybridIterator *hr, RSIndexResult *chil
   RSIndexResult *res = NewHybridResult();
   AggregateResult_AddChild(res, IndexResult_DeepCopy(vec_res));
   AggregateResult_AddChild(res, IndexResult_DeepCopy(child_res));
-  res->data.hybrid_metric.isCopy = true; // Mark as copy, so when we free it, it will also free its children.
+  res->data.hybrid_metric.tag = RSAggregateResult_Owned; // Mark as copy, so when we free it, it will also free its children.
```
GlenDC, line 95: why this comment is needed, is that not what the tag implies or is this implicit behavior specific to this usage?

GlenDC:
Btw might be nice to add a link directly to the ticket (https://redislabs.atlassian.net/browse/MOD-10746) as ticket id in title is not clickable and neither there is a bot that seems to be doing it. Small QoL thing, but in worst case copy pasting the ID works as well of course.

src/iterators/hybrid_reader.c
chesedo: It links directly to this cleanup code

https://github.com/RediSearch/RediSearch/blob/c365c578d5068bbfbca8c60ba2cb7b1f41bb1219/src/index_result.c#L228-L238

The comment makes it clearer to those unfamiliar with the code that we are setting this to ensure the two deep copies created on the two lines before this are freed correctly. This is explicit since these two logic pieces are so far apart

chesedo:
> Btw might be nice to add a link directly to the ticket (https://redislabs.atlassian.net/browse/MOD-10746) as ticket id in title is not clickable and neither there is a bot that seems to be doing it. Small QoL thing, but in worst case copy pasting the ID works as well of course.

I wonder if there is a way to automate this... :thinking:

lerman25 (approved):

PR #6606, opened 2025-08-07
[MOD-10766] Fix build with clang 20

gdesmott:
WIP, making sure CI fully pass when using clang 20.

Will be needed for #6478

PR #6689, opened 2025-08-22
Private link libstdc++ to address focal issue

ethanglaser:

This privately links redisearch.so to libstdc++. This is necessary because when installing gcc-11 on focal, a newer libstdc++ is installed with this, introducing new symbols. When using this build in the default user env with gcc-9 and lacking the newer libstdc++, missing symbol error occurs. This change adds the symbols privately and there is no runtime dependency on this library.

#### Mark if applicable

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -322,7 +322,7 @@ extract_debug_symbols(redisearch)
 add_dependencies(redisearch VectorSimilarity)
 add_dependencies(redisearch uv_a)
 
-target_link_options(redisearch PRIVATE -static-libstdc++ -static-libgcc)
+target_link_options(redisearch PRIVATE -static-libstdc++)
 
 
 #----------------------------------------------------------------------------------------------


gdesmott:
We also experienced linking issues when building with clang 20, see https://github.com/RediSearch/RediSearch/pull/6606#issuecomment-3163596160

Could this change help for this problem as well?

ethanglaser:
> We also experienced linking issues when building with clang 20, see [#6606 (comment)](https://github.com/RediSearch/RediSearch/pull/6606#issuecomment-3163596160)
> 
> Could this change help for this problem as well?

It's possible that something like this could help. However, it sounds like we'll not be proceeding with this option is it leads to complications with dynamic linking elsewhere in redis. Closing this PR for now. CC @alonre24

gdesmott:
Can you please comment here or CC me when you'll find another solution?

PR #6369, opened 2025-06-24
Small touchups relating to the inverted index work

chesedo:
## Describe the changes in the pull request
This add a virtual data type to go with virtual `RSIndexResult`s. This will make some of the following Rust PRs easier.

It also fixes some cargo definition to correctly get the `publish` property from the workspace.

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #6691, opened 2025-08-22
apply QOL to RSResultData: Numeric+Metric

GlenDC:
Quality of Life (QoL) patch for redissearch_rs's RSResultData:

Variants of RSResultData updated:

- [X] Numeric: `f64`
- [X] Metric: `f64`
- [ ] Virtual: TODO


GlenDC:
@chesedo can you explain what RSVirtualResult is about? Why was that an empty struct to begin with?

chesedo (dismissed):

chesedo:
> @chesedo can you explain what RSVirtualResult is about? Why was that an empty struct to begin with?

We initially removed it in #6301, but then discovered some decoders need the virtual variant so added it back in #6369. I just used an empty struct to make the C union type easier but am no longer sure how this would have been easier. A non-data variant should have worked then too.

GlenDC:
Done @chesedo . this PR is ready. Remove the Result type as well.

GlenDC:
> LGTM Why codecov complaining about uncovered code?

It's a small code cleanup and the test code is a bit smaller because of that. So that makes it look like we have less coverage, despite nothing having changed logic wise. It's a statistical truth but not something with meaning we have to worry about here. The code is smaller and the same, and what was tested before is still tested.

PR #6672, opened 2025-08-18
MOD-10975: Fix fd leak when OOM

nafraf:
## Describe the changes in the pull request

A clear and concise description of what the PR is solving, including:
1. Current: 
    `periodicCb()` function returns early without closing the pipe file descriptors that were created.
2. Change: 
    Missing close() calls in the out-of-memory error path were added.
3. Outcome:
    Fix fd leak

#### Main objects this PR modified
1. ...

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #6702, opened 2025-08-24
[2.10] MOD-10975: Fix fd leak when OOM (#6672)

nafraf:
# Description
Manual backport of #6672 to `2.10`.
(cherry picked from commit 56f6f0e)

Changes:
- Fix conflict in `test_gc_oom()` to create index before adding the docs. 
- Fix conflict in `test_gc_oom()` to call set_tight_maxmemory_for_oom(env) instead of `env.expect('config', 'set', 'maxmemory', 1).ok()`

### DIFF
--- a/tests/pytests/test_gc.py
+++ b/tests/pytests/test_gc.py
@@ -4,6 +4,16 @@ import platform
 from time import sleep
 import threading
 import time
+import psutil
+
+
+def get_open_file_count(pid):
+    """Get open file count using psutil library"""
+    try:
+        p = psutil.Process(pid)
+        return p.num_fds()
+    except psutil.NoSuchProcess:
+        return 0
 
 @skip(cluster=True)
 def testBasicGC(env):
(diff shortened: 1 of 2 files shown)


tests/pytests/test_gc.py
```diff
@@ -420,15 +430,26 @@ def test_gc_oom(env):
...
+    open_files_before_gc = get_open_file_count(redis_server_pid)
+    env.assertGreater(open_files_before_gc, 0)
+
+    set_tight_maxmemory_for_oom(env)
+
+    # Delete them all
     for i in range(num_docs):
         env.expect('DEL', f'doc{i}').equal(1)
-
-    env.expect('config', 'set', 'maxmemory', 1).ok()
```
raz-mon, line 430: Is there a reason for this discrepancy? CC @lerman25

raz-mon (approved):

tests/pytests/test_gc.py
lerman25: I'm not sure, my bad that I didn't document it.
My guess is that set_tight/unlimited functions in common.py weren't backported to 2.X at the time,
I don't see any other reason.

PR #6715, opened 2025-08-26
Use uv to manage the Python environment for e2e tests.

LukeMathWalker:

Improve the reproducibility of our Python setup.

#### Mark if applicable


LukeMathWalker:
Current blocker: `psutil` doesn't provide prebuilt wheels for aarch64 for 5.x.y. We need to upgrade RLTest to `psutil` 6.x.y to avoid having to build from source on aarch64 systems.

.install/amazon_linux_2.sh
```diff
@@ -5,11 +5,7 @@ set -e
 export DEBIAN_FRONTEND=noninteractive
 
 $MODE yum update -y
-$MODE amazon-linux-extras enable python3.8
-$MODE yum install -y python3.8 python38-devel which
-
-# Prefer python3.8 via PATH (do NOT fight the /usr/bin/python3 owned by RPMs)
-$MODE ln -sf /usr/bin/python3.8 /usr/local/bin/python3
+$MODE yum install -y which
```
oshadmi, line 8: `amazon-linux-extras` is not required since python installation is not done here?
Will it not be needed elsewhere (where python is being installed) since `uv` will take care of that?

.install/install_python.sh
```diff
@@ -0,0 +1,14 @@
...
+processor=$(uname -m)
+OS_TYPE=$(uname -s)
+
+curl --proto '=https' --tlsv1.2 -LsSf https://astral.sh/uv/install.sh | sh
+# Add the newly installed `uv` to the PATH
+export PATH="$HOME/.local/bin:$PATH"
+
+# Verify uv is in path
+uv -vV
+# Print where `cargo` is located for debugging purposes
```
oshadmi:
```suggestion
# Print where `uv` is located for debugging purposes
```

.install/test_deps/common_installations.sh
```diff
@@ -6,17 +6,21 @@ MODE=$1 # whether to install using sudo or not
...
 		# causes $PATH issues on platforms like Debian and Alpine,
-		# shadowing the pre-existing source command to make `cargo` available.
+		# shadowing the pre-existing source command to make some of our tools available.
 		# We work around it by appending the required lines to the shell profile
 		# _after_ the venv activation script
+
+		# cargo
 		echo '. "$HOME/.cargo/env"' >> ~/.bash_profile
+		# uv
+		echo 'export PATH="$HOME/.local/bin:$PATH"' >> ~/.bash_profile
```
oshadmi, line 23: Could this add more unintended stuff except for `uv` to the head of the PATH?

.install/ubuntu_24.04.sh
```diff
@@ -5,4 +5,11 @@ MODE=$1 # whether to install using sudo or not
...
 $MODE apt install -yqq git wget build-essential lcov openssl libssl-dev \
-    python3 python3-venv python3-dev unzip rsync clang curl libclang-dev
+    unzip rsync clang curl libclang-dev
+
+# We need Python headers to build psutil@5.x.y from
+# source, since it only started providing wheels for aarch64
+# in version 6.w.z.
+if [ "$(uname -m)" = "aarch64" ]; then
+    $MODE apt install -y python3-dev
+fi
```
oshadmi, line 15: Isn't it needed in other installation scripts, e.g., in .install/ubuntu_22.04.sh ?

LukeMathWalker: It doesn't seem to be needed, most likely because Python headers are included by default in the distro.

.install/amazon_linux_2.sh
LukeMathWalker: Correct—the Python we use will come from `uv`, so there's no need to bring in `amazon-linux-extras` and we don't seem to need anything else from that bundle.

.install/test_deps/common_installations.sh
LukeMathWalker: In theory, it could. 
The alternative would be to specify every binary by hand (`uv`, `uvx`, their `python3.x` shim), making the setup less robust to potential changes in future `uv` versions.
I'm happy to go either way, as you prefer.

oshadmi: Let's keep it this way and go with robustness since I suppose all local bin are installed by us in CI and should be fine.

.install/ubuntu_24.04.sh
oshadmi: I suggest to run https://github.com/RediSearch/RediSearch/actions/workflows/flow-build-artifacts.yml on all platforms and architectures using this branch to validate this.

oshadmi: I see the workflow passed
https://github.com/RediSearch/RediSearch/actions/runs/17292692727

LukeMathWalker: Yes, that's the one I kicked off last week and it came out ✅

PR #6728, opened 2025-08-27
document Rust dependencies policy

gdesmott:
#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/CONTRIBUTING.md
+++ b/src/redisearch_rs/CONTRIBUTING.md
@@ -11,4 +11,4 @@ thiserror.workspace = true
 Dependency versions should be updated:
 - ASAP in case of security advisories.
 - Whenever we need newer features or bug fixes released in a newer version.
-- Once in a while, via `cargo update`, if neither of the two things above have happened.
+- Once in a while, via [`cargo upgrade`](https://crates.io/crates/cargo-edit), if neither of the two things above have happened.


src/redisearch_rs/CONTRIBUTING.md
```diff
@@ -0,0 +1,14 @@
+## Dependencies
+
+Dependencies should be added to the `Cargo.toml` file in the root of the workspace.
+They can then be used in the crates `Cargo.toml` files using:
```
LukeMathWalker:
```suggestion
They can then be used by workspace members via:
```

src/redisearch_rs/CONTRIBUTING.md
```diff
@@ -0,0 +1,14 @@
...
+
+```toml
+[dependencies]
+thiserror.workspace = true
+```
+
+They should be updated:
+- ASAP in case of advisories security vulnerabilities.
+- whenever we need newer features or bug fixes from the new version.
+- once in a while if neither of the two things above have happened.
```
LukeMathWalker:
```suggestion
Dependency versions should be updated:
- ASAP in case of security advisories.
- Whenever we need newer features or bug fixes released in a newer version.
- Once in a while, via `cargo update`, if neither of the two things above have happened.
```

gdesmott: That implies we should not explicitly upgrade the version in `Cargo.toml` when not required? `cargo update` won't bump those.

LukeMathWalker: Fair point—we should `cargo upgrade` here.

LukeMathWalker (approved):

PR #6729, opened 2025-08-27
[MOD-11101] Remove encoder arg from `InvertedIndex_WriteForwardIndexEntry`

chesedo:
This removes the `encoder` argument from `InvertedIndex_WriteForwardIndexEntry` (and all the functions calling it) to align it with the incoming Rust code.

This argument is redundant since the inverted index (which is already an argument) already knows which encoder it should use.

#### Mark if applicable

### DIFF
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -97,11 +97,11 @@ static void writeCurEntries(RSAddDocumentCtx *aCtx, RedisSearchCtx *ctx) {
   IndexSpec *spec = ctx->spec;
   ForwardIndexIterator it = ForwardIndex_Iterate(aCtx->fwIdx);
   ForwardIndexEntry *entry = ForwardIndexIterator_Next(&it);
-  IndexEncoder encoder = InvertedIndex_GetEncoder(aCtx->specFlags);
 
   while (entry != NULL) {
     bool isNew;
     InvertedIndex *invidx = Redis_OpenInvertedIndex(ctx, entry->term, entry->len, 1, &isNew);
+    IndexEncoder encoder = InvertedIndex_GetEncoder(InvertedIndex_Flags(invidx));
     if (isNew && strlen(entry->term) != 0) {
       IndexSpec_AddTerm(spec, entry->term, entry->len);
     }
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -300,9 +300,8 @@ static void writeMissingFieldDocs(RSAddDocumentCtx *aCtx, RedisSearchCtx *sctx, 
     }
     // Add docId to inverted index
     t_docId docId = aCtx->doc->docId;
-    IndexEncoder enc = InvertedIndex_GetEncoder(Index_DocIdsOnly);
     RSIndexResult rec = {.data.tag = RSResultData_Virtual, .docId = docId, .offsetsSz = 0, .freq = 0};
-    aCtx->spec->stats.invertedSize +=InvertedIndex_WriteEntryGeneric(iiMissingDocs, enc, &rec);
+    aCtx->spec->stats.invertedSize +=InvertedIndex_WriteEntryGeneric(iiMissingDocs, &rec);
   }
   dictReleaseIterator(iter);
   dictRelease(df_fields_dict);
@@ -321,9 +320,8 @@ static void writeExistingDocs(RSAddDocumentCtx *aCtx, RedisSearchCtx *sctx) {
   }
 
   t_docId docId = aCtx->doc->docId;
-  IndexEncoder enc = InvertedIndex_GetEncoder(Index_DocIdsOnly);
   RSIndexResult rec = {.data.tag = RSResultData_Virtual, .docId = docId, .offsetsSz = 0, .freq = 0};
-  aCtx->spec->stats.invertedSize += InvertedIndex_WriteEntryGeneric(sctx->spec->existingDocs, enc, &rec);
+  aCtx->spec->stats.invertedSize += InvertedIndex_WriteEntryGeneric(sctx->spec->existingDocs, &rec);
 }
 
 /**
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -27,9 +27,8 @@ extern void IncrementYieldCounter(void);
 
 #include <unistd.h>
 
-static void writeIndexEntry(IndexSpec *spec, InvertedIndex *idx, IndexEncoder encoder,
-                            ForwardIndexEntry *entry) {
-  size_t sz = InvertedIndex_WriteForwardIndexEntry(idx, encoder, entry);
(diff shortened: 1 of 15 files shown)


tests/cpptests/micro-benchmarks/benchmark_index_iterator.cpp
```diff
@@ -84,10 +84,14 @@ class BM_IndexIterator : public benchmark::Fixture {
     }
 
     void createIndex(IndexFlags flags) {
+        if (flags == (Index_DocIdsOnly | Index_Temporary)) {
+            // Special case reserved for `Index_DocIdsOnly` with raw doc IDs
+            RSGlobalConfig.invertedIndexRawDocidEncoding = true; // Enable raw doc ID encoding, until the benchmark's tearDown
```
GlenDC, line 90: What happened with

```
RS_ASSERT_ALWAYS(encoder != InvertedIndex_GetEncoder(Index_DocIdsOnly)); // Ensure we are using the raw doc ID encoder
            encoder = InvertedIndex_GetEncoder(Index_DocIdsOnly);
            for (size_t i = 0; i < ids.size(); ++i) {
                RSIndexResult rec = {.docId = ids[i], .data = {.tag = RSResultData_Virtual}};
                InvertedIndex_WriteEntryGeneric(index, encoder, &rec);
            }
```

no longer required?

GlenDC (approved):
LGTM, one question, probably fine but I miss some context there

tests/cpptests/micro-benchmarks/benchmark_index_iterator.cpp
chesedo: That piece of code will be excercised by the `} else if (flags == Index_DocIdsOnly) {` case. Ie both the `else if` I removed and `else if` above it are now combined. This is possible because here we are setting the `RSGlobalConfig.invertedIndexRawDocidEncoding` early. This will cause the new combined case to correctly use the raw doc id encoding because of this

https://github.com/RediSearch/RediSearch/blob/07f7ed625f839328e68438aa75755a925053974c/src/inverted_index/inverted_index.c#L609-L614

gdesmott (dismissed):

tests/cpptests/micro-benchmarks/benchmark_index_iterator.cpp
GuyAv46: We can still keep the assertion here just to make sure this config is what is changing the behavior of `InvertedIndex_GetEncoder`

src/indexer.c
```diff
@@ -109,7 +107,7 @@ static void writeCurEntries(RSAddDocumentCtx *aCtx, RedisSearchCtx *ctx) {
       entry->docId = aCtx->doc->docId;
       RS_LOG_ASSERT(entry->docId, "docId should not be 0");
       IndexerYieldWhileLoading(ctx->redisCtx);
-      writeIndexEntry(spec, invidx, encoder, entry);
+      writeIndexEntry(spec, invidx, entry);
```
GuyAv46, line 110: This is the most critical change in this PR. We now call `InvertedIndex_GetEncoder` for every term in the document instead of once.
Should we worry about the performance? Or will it become a no-op once we move this code to Rust?

tests/cpptests/micro-benchmarks/benchmark_index_iterator.cpp
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6729/commits/af1e1f54feb893da8c63ac8d5104db446d4e42ee

src/indexer.c
chesedo: Correct, this will add overhead with each call. And it is going away in Rust because the encoder will be stored on the index

https://github.com/RediSearch/RediSearch/blob/fa08b0bb63d04b74284b0014f6f16e8e91d93036/src/redisearch_rs/inverted_index/src/lib.rs#L134-L147

GuyAv46: So when we start using the Rust inverted indexes, we should compare the indexing performance to the one before this PR

tests/cpptests/micro-benchmarks/benchmark_index_iterator.cpp
```diff
@@ -84,30 +84,30 @@ class BM_IndexIterator : public benchmark::Fixture {
...
                 InvertedIndex_WriteNumericEntry(index, ids[i], static_cast<double>(i));
             }
         } else if (flags == Index_DocIdsOnly) {
-            // Populate the index with document IDs only
-            for (size_t i = 0; i < ids.size(); ++i) {
-                RSIndexResult rec = {.docId = ids[i], .data = {.tag = RSResultData_Virtual}};
-                InvertedIndex_WriteEntryGeneric(index, encoder, &rec);
+            if (flags == (Index_DocIdsOnly | Index_Temporary)) {
+                // Ensure we are using the raw doc ID encoder
+                RS_ASSERT_ALWAYS(InvertedIndex_GetEncoder(InvertedIndex_Flags(index)) != InvertedIndex_GetEncoder(Index_DocIdsOnly));
```
GuyAv46: This assertion is unreachable (if we are here, `flags` is never `(Index_DocIdsOnly | Index_Temporary)`).
Let's move it to the beginning of `createIndex`

tests/cpptests/micro-benchmarks/benchmark_index_iterator.cpp
```diff
@@ -84,30 +84,30 @@ class BM_IndexIterator : public benchmark::Fixture {
     }
 
     void createIndex(IndexFlags flags) {
+        if (flags == (Index_DocIdsOnly | Index_Temporary)) {
+            // Special case reserved for `Index_DocIdsOnly` with raw doc IDs
+            RSGlobalConfig.invertedIndexRawDocidEncoding = true; // Enable raw doc ID encoding, until the benchmark's tearDown
+        }
```
GuyAv46:
```suggestion
        if (flags == (Index_DocIdsOnly | Index_Temporary)) {
            IndexEncoder default_id_only_encoder = InvertedIndex_GetEncoder(flags);
            // Special case reserved for `Index_DocIdsOnly` with raw doc IDs
            RSGlobalConfig.invertedIndexRawDocidEncoding = true; // Enable raw doc ID encoding, until the benchmark's tearDown
            RS_ASSERT_ALWAYS(default_id_only_encoder != InvertedIndex_GetEncoder(flags));
        }
```

src/indexer.c
chesedo: Yes, good point

tests/cpptests/micro-benchmarks/benchmark_index_iterator.cpp
chesedo: Lol, obviously :man_facepalming: . I accepted your suggestion

GuyAv46 (approved):

PR #6730, opened 2025-08-27
[MOD-11154]  Create unified memory checker

lerman25:
Create memory checker in RediSearch for OOM checks.
Should be later optimized for minimal overhead

### DIFF
--- a/src/util/redis_mem_info.h
+++ b/src/util/redis_mem_info.h
@@ -45,4 +45,4 @@ static inline float RedisMemory_GetUsedMemoryRatio(void) {
 // GIL must be held before calling this function
 // Returns 0 if maxmemory is 0
 // TODO: remove this function and use RedisMemory_GetUsedMemoryRatio instead after benchmarking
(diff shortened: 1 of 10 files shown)


src/util/redis_mem_info.h
```diff
@@ -0,0 +1,48 @@
...
+static inline float RedisMemory_GetUsedMemoryRatio(void) {
+  return RedisModule_GetUsedMemoryRatio();
+}
+
+// Get the used memory ratio from Redis server info.
+// Same function as before
+// GIL must be held before calling this function
+// Returns 0 if maxmemory is 0
+// TODO: remove this function and use RedisMemory_GetUsedMemoryRatio instead after benchmarking
+float Unified_GetUsedMemoryRatio(RedisModuleCtx *ctx);
```
alonre24:
```suggestion
float RedisMemory_GetUsedMemoryRatioUnified(RedisModuleCtx *ctx);
```
Consider keeping the convention that functions in this unit are prefixed with `RedisMemory`

PR #6735, opened 2025-08-27
MOD-10702: rough draft of WIP of step1 for MOD-10702

GlenDC:
Very early WIP draft, only here
to double check if this is the direction we want;
Does not work yet (segfaults)

### DIFF
--- a/src/inverted_index/inverted_index.c
+++ b/src/inverted_index/inverted_index.c
@@ -1287,3 +1287,104 @@ unsigned long InvertedIndex_MemUsage(const InvertedIndex *idx) {
   }
   return ret;
 }
+
+void InvertedIndex_RepairBlocks(
+    const InvertedIndex *idx, DocTable *dt,
+    IndexRepairParams *index_params,
+    IndexRepairBlocksParams *block_params
+) {
+    IndexRepairBlocksParams block_params_s = {0};
+    if (!block_params) {
+      block_params = &block_params_s;
+    }
+
+    IndexRepairParams index_param_s = {0};
+    if (!index_params) {
+      index_params = &index_param_s;
+    }
+
+    block_params->numBlocks = InvertedIndex_NumBlocks(idx);
+
+    if (block_params->fixedBlocks) {
+        array_clear(block_params->fixedBlocks);
+    } else {
+        block_params->fixedBlocks = array_new(FixedIndexBlock, 10);
+    }
+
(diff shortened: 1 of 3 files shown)


src/inverted_index/inverted_index.c
```diff
@@ -1216,3 +1216,104 @@ size_t IndexBlock_Repair(IndexBlock *blk, DocTable *dt, IndexFlags flags, IndexR
   IndexResult_Free(res);
   return frags;
 }
+
+void InvertedIndex_RepairBlocks(
+    const InvertedIndex *idx, DocTable *dt,
+    IndexRepairParams *index_params,
+    IndexRepairBlocksParams *block_params
+) {
```
chesedo, line 1295: Will we have access to `DocTable`, `IndexRepairParams`, and `IndexRepairBlocksParams` in Rust? How much of them are possible to remove?

src/inverted_index/inverted_index.c
```diff
@@ -1216,3 +1216,104 @@ size_t IndexBlock_Repair(IndexBlock *blk, DocTable *dt, IndexFlags flags, IndexR
...
+        block_params->lastblkNumEntries = IndexBlock_NumEntries(blk) + index_params->entriesCollected;
+      }
+    }
+}
+
+void InvertedIndex_RemRepairBlocksParams(IndexRepairBlocksParams block_params) {
+    array_free(block_params.fixedBlocks);
+    array_free(block_params.blockList);
+    array_free(block_params.deletedBlocks);
+}
```
chesedo, line 1390: Will we want this in Rust?

src/inverted_index/inverted_index.c
chesedo: Okay, I found the `IndexRepairBlocksParams` in the header file. Does it make sense to remove the other two?

src/inverted_index/inverted_index.c
chesedo: Nvm, I see we are managing this type. So this seems good

src/inverted_index/inverted_index.c
GlenDC: @chesedo over time more I guess, but in a first step this would be I think the easiest? How do you see this?

GlenDC:
Replaced by https://github.com/RediSearch/RediSearch/pull/6758

PR #6737, opened 2025-08-27
MOD-10407: Handle warnings and errors of the subquery pipelines

Itzikvaknin:

* **Implement error & warning handling for hybridRequest:**
   - **Error handling**: Only one error is returned, tail has priority over subqueries, logic is in `HREQ_GetError`
   - **Warning handling**: We return the warning and its source in the payload, concatenating `(SOURCE)` to the message, for example: `'Max prefix expansions limit was reached (VSIM)'`

* **Add FT.HYBRID to _FT.DEBUG**: Allowing adding timeout processor in each subquery and in the tail
   - `TIMEOUT_AFTER_N_SEARCH <N>`: Timeout after N results from search component
   - `TIMEOUT_AFTER_N_VSIM <N>`: Timeout after N results from vector component
   - `TIMEOUT_AFTER_N_TAIL <N>`: Timeout after N results from tail pipeline (merger)

* **Tests in test_hybrid_errors.py:**
   - Debug timeout tests using `TIMEOUT_AFTER_N_*` parameters
   - Real timeout tests with large datasets
   - Warning and error handling tests
   - **Note**: Tests are skipped on cluster until FT.HYBRID is implemented for cluster mode
   - **Note**: Timeout in the tail should be tested once MOD-11004 is merged

* **Minor memory fixes**: Fixed memory leaks in error handling paths


src/hybrid/hybrid_exec.c
```diff
@@ -265,14 +298,18 @@ void sendChunk_hybrid(HybridRequest *hreq, RedisModule_Reply *reply, size_t limi
...
     }
-    if (rc == RS_RESULT_TIMEDOUT) {
-      RedisModule_Reply_SimpleString(reply, QueryError_Strerror(QUERY_ETIMEDOUT));
-    } else if (rc == RS_RESULT_ERROR) {
-      // Non-fatal error
-      RedisModule_Reply_SimpleString(reply, QueryError_GetUserError(qctx->err));
-    } else if (qctx->err->reachedMaxPrefixExpansions) {
-      RedisModule_Reply_SimpleString(reply, QUERY_WMAXPREFIXEXPANSIONS);
+
+    bool TimeoutInSubquery = false;
```
kei-nan:
```suggestion
    bool timeoutInSubquery = false;
```

src/hybrid/hybrid_exec.c
```diff
@@ -265,14 +298,18 @@ void sendChunk_hybrid(HybridRequest *hreq, RedisModule_Reply *reply, size_t limi
...
+    for(size_t i = 0; i < hreq->nrequests; i++) {
```
kei-nan:
```suggestion
    for (size_t i = 0; i < hreq->nrequests; ++i) {
```

src/hybrid/hybrid_exec.c
```diff
@@ -265,14 +298,18 @@ void sendChunk_hybrid(HybridRequest *hreq, RedisModule_Reply *reply, size_t limi
...
+      QueryError* err = &hreq->errors[i];
+      char* suffix = i == 0 ? SEARCH_SUFFIX : VSIM_SUFFIX;
```
kei-nan:
```suggestion
      const char* suffix = i == 0 ? SEARCH_SUFFIX : VSIM_SUFFIX;
```

src/hybrid/hybrid_exec.c
```diff
@@ -265,14 +298,18 @@ void sendChunk_hybrid(HybridRequest *hreq, RedisModule_Reply *reply, size_t limi
...
+      int subQueryReturnCode = hreq->subqueriesReturnCodes[i];
```
kei-nan:
```suggestion
      const int subQueryReturnCode = hreq->subqueriesReturnCodes[i];
```

src/hybrid/hybrid_exec.c
```diff
@@ -265,14 +298,18 @@ void sendChunk_hybrid(HybridRequest *hreq, RedisModule_Reply *reply, size_t limi
...
+
+      TimeoutInSubquery = handleQueryError(reply, err, subQueryReturnCode, suffix, false);
```
kei-nan, line 308: what is TimeoutInSubquery used for?
do we need it?

kei-nan: TimeoutInSubquery gets overwritten in second iteration.
So at first you could get a true but on the second iteration it would be a false.
maybe just count the number of timeouts or use a logical/binary or.

src/hybrid/hybrid_request.c
```diff
@@ -164,14 +164,14 @@ int HybridRequest_BuildPipeline(HybridRequest *req, const HybridPipelineParams *
...
-
     // Restore the LOAD step to the tail pipeline for proper cleanup
     if (loadStep) {
         AGPLN_AddStep(&req->tailPipeline->ap, &loadStep->base);
     }
+    if (rc != REDISMODULE_OK) {
+        return rc;
+    };
+
     return REDISMODULE_OK;
```
kei-nan:
```suggestion
return rc;
```

src/hybrid/hybrid_request.h
```diff
@@ -23,6 +23,7 @@ typedef struct HybridRequest {
     RequestConfig reqConfig;
     HybridPipelineParams *hybridParams;
     clock_t initClock;  // For timing execution
+    int *subqueriesReturnCodes;  // Array to store return codes from each subquery
```
kei-nan:
```suggestion
    arrayof(int) subqueriesReturnCodes;  // Array to store return codes from each subquery
```

src/debug_commands.c
```diff
@@ -1875,6 +1880,7 @@ DebugCommandType commands[] = {{"DUMP_INVIDX", DumpInvertedIndex}, // Print all
                                {"_FT.AGGREGATE", RSAggregateCommandShard}, // internal use only, in SA use FT.AGGREGATE
                                {"FT.SEARCH", DistSearchCommand_DebugWrapper},
                                {"_FT.SEARCH", RSSearchCommandShard}, // internal use only, in SA use FT.SEARCH
+                               {"FT.HYBRID", HybridCommand_DebugWrapper},
```
kei-nan, line 1883: do we also want to support _FT.HYBRID?

src/hybrid/hybrid_request.c
```diff
@@ -192,6 +192,9 @@ HybridRequest *HybridRequest_New(AREQ **requests, size_t nrequests) {
     // Initialize error tracking for each individual request
     hybridReq->errors = array_new(QueryError, nrequests);
 
+    // Initialize return codes array for tracking subqueries final states
+    hybridReq->subqueriesReturnCodes = rm_calloc(nrequests, sizeof(int));
```
kei-nan: maybe use array macro to handle the array?

src/debug_commands.c
Itzikvaknin: Once we have it :)

src/hybrid/hybrid_request.c
Itzikvaknin: Avoided using the array macro since it uses on malloc, and I need initialization

src/hybrid/hybrid_request.h
Itzikvaknin: Avoided using the array macro since this array is not managed by the macro (header etc.)

src/hybrid/hybrid_exec.c
```diff
@@ -31,6 +31,38 @@
...
+    rm_free(expanded_warning);
+  } else {
+    RedisModule_Reply_SimpleString(reply, message);
+  }
+}
+
+static inline bool handleQueryError(RedisModule_Reply *reply, QueryError *err, int returnCode, const char *suffix, bool ignoreTimeout) {
+  bool timeoutOccurred = false;
+
+  if (returnCode == RS_RESULT_TIMEDOUT && !ignoreTimeout) {
```
kei-nan, line 58: should it be possible to get timeout and prefix max expansion warnings?
Do we want to hide other warnings we might have encountered?

src/hybrid/hybrid_exec.c
```diff
@@ -212,9 +244,10 @@ void sendChunk_hybrid(HybridRequest *hreq, RedisModule_Reply *reply, size_t limi
     startPipelineHybrid(hreq, rp, &results, &r, &rc);
 
     // If an error occurred, or a timeout in strict mode - return a simple error
-    if (ShouldReplyWithError(rp->parent->err, hreq->reqConfig.timeoutPolicy, false)) {  // hybrid doesn't support profiling yet
-      // TODO: take to error using HREQ_GetError
-      RedisModule_Reply_Error(reply, QueryError_GetUserError(qctx->err));
+    QueryError err = {0};
+    HREQ_GetError(hreq, &err);
+    if (ShouldReplyWithError(&err, hreq->reqConfig.timeoutPolicy, false)) {  // hybrid doesn't support profiling yet
```
kei-nan: consider removing the profiling comment, not sure what it has to do with timeout.

src/hybrid/hybrid_exec.c
```diff
@@ -265,14 +298,18 @@ void sendChunk_hybrid(HybridRequest *hreq, RedisModule_Reply *reply, size_t limi
...
-    } else if (qctx->err->reachedMaxPrefixExpansions) {
-      RedisModule_Reply_SimpleString(reply, QUERY_WMAXPREFIXEXPANSIONS);
+
+    bool timeoutInSubquery = false;
+    for (size_t i = 0; i < hreq->nrequests; ++i) {
+      QueryError* err = &hreq->errors[i];
+      const char* suffix = i == 0 ? SEARCH_SUFFIX : VSIM_SUFFIX;
+      const int subQueryReturnCode = hreq->subqueriesReturnCodes[i];
+
+      timeoutInSubquery = timeoutInSubquery || handleQueryError(reply, err, subQueryReturnCode, suffix, false);
```
kei-nan: I think the order need to be reveresed.
If timeoutInSubquery is true you still want the handleQueryError to run on the current error.

src/debug_commands.c
kei-nan: I think there is a handler for it.
it currently goes to the same command handler as FT.HYBRID.

src/result_processor.c
```diff
@@ -1801,10 +1801,14 @@ dictType dictTypeHybridSearchResult = {
         // Upstream is still active but not ready to provide results. Skip to the next.
         continue;
       }
+
+      // Store the final return code for this upstream
+      self->upstreamReturnCodes[i] = rc;
```
kei-nan, line 1824: is there any point to the timeout and error variables remaining if you are storing the return codes for all the upstreams?

src/hybrid/hybrid_exec.c
Itzikvaknin: Nice catch :)

src/result_processor.c
Itzikvaknin: Removed it and error as well

src/hybrid/hybrid_exec.c
Itzikvaknin: No, only one warning can be returned in `FT.AGGREGATE`, so I assumed we should behave the same

tests/pytests/test_rrf.py
```diff
@@ -32,26 +31,10 @@ def run_test_scenario(env, no_tag_search_query, with_tag_search_query):
...
-            key = access_nested_list(result, key_index)
-            score = float(access_nested_list(result, score_index))
-
-            results[key] = score
-        return results
-
-    results_no_tag = get_results(hybrid_res_no_tag)
-    results_with_tag = get_results(hybrid_res_with_tag) #type: ignore
+    results_no_tag = get_results_from_hybrid_response(hybrid_res_no_tag)
+    results_with_tag = get_results_from_hybrid_response(hybrid_res_with_tag) #type: ignore
```
raz-mon: What does this comment mean?

tests/pytests/test_hybrid_errors.py
```diff
@@ -0,0 +1,295 @@
...
+    dim = 2
+    conn = env.getClusterConnectionIfNeeded()
+    env.expect('FT.CREATE idx SCHEMA description TEXT embedding VECTOR FLAT 6 TYPE FLOAT32 DIM 2 DISTANCE_METRIC L2').ok
+
+    # Load test data
+    for doc_id, doc_data in test_data.items():
+        conn.execute_command('HSET', doc_id, 'description', doc_data['description'], 'embedding', doc_data['embedding'])
+
+# Debug timeout tests using TIMEOUT_AFTER_N_* parameters
+#TODO: remove once FT.HYBRID for cluster is implemented
```
raz-mon:
```suggestion
#TODO: remove skip once FT.HYBRID for cluster is implemented
```

tests/pytests/test_hybrid_errors.py
```diff
@@ -0,0 +1,295 @@
...
+    'doc:4': {
+        'description': "blue shoes",
+        'embedding': np.array([1.0, 1.0]).astype(np.float32).tobytes()
+    }
+}
+
+# Query vector for testing
+query_vector = np.array([1.2, 0.2]).astype(np.float32).tobytes()
+
+def get_warnings(response):
```
raz-mon, line 28:
```suggestion
def get_warning(response):
```

tests/pytests/test_hybrid_errors.py
```diff
@@ -0,0 +1,295 @@
...
+    env.expect('_FT.DEBUG', 'FT.HYBRID', 'idx', 'SEARCH', 'running', 'VSIM', '@embedding', '$BLOB', 'PARAMS', '2', 'BLOB', query_vector, 'TIMEOUT_AFTER_N_TAIL', '1', 'DEBUG_PARAMS_COUNT', '2').error().contains('Timeout limit was reached')
+
+#TODO: remove once FT.HYBRID for cluster is implemented
+@skip(cluster=True)
+def test_debug_timeout_return_search():
+    """Test RETURN policy with search timeout using debug parameters"""
+    env = Env(enableDebugCommand=True, moduleArgs='ON_TIMEOUT RETURN')
+    setup_basic_index(env)
+    response = env.cmd('_FT.DEBUG', 'FT.HYBRID', 'idx', 'SEARCH', 'running', 'VSIM', '@embedding', '$BLOB', 'PARAMS', '2', 'BLOB', query_vector, 'TIMEOUT_AFTER_N_SEARCH', '1', 'DEBUG_PARAMS_COUNT', '2')
+    env.assertTrue('Timeout limit was reached (SEARCH)' in get_warnings(response))
```
raz-mon: It's equality that is actually tested here right? I.e., this will be the whole `warning` section.
Consider asserting equality if so

raz-mon: Also, this is a great place to assert the behavior of non-strict timeouts with some results.
I.e., set the timeout to be after 2 results, and make sure we get them in the command output

tests/pytests/test_hybrid_errors.py
```diff
@@ -0,0 +1,295 @@
...
+
+#TODO: remove once FT.HYBRID for cluster is implemented
+@skip(cluster=True)
+def test_debug_timeout_return_vsim():
+    """Test RETURN policy with vector similarity timeout using debug parameters"""
+    env = Env(enableDebugCommand=True, moduleArgs='ON_TIMEOUT RETURN')
+    setup_basic_index(env)
+    response = env.cmd('_FT.DEBUG', 'FT.HYBRID', 'idx', 'SEARCH', 'running', 'VSIM', '@embedding', '$BLOB', 'PARAMS', '2', 'BLOB', query_vector, 'TIMEOUT_AFTER_N_VSIM', '1', 'DEBUG_PARAMS_COUNT', '2')
```
raz-mon, line 92: Same as above

tests/pytests/test_hybrid_errors.py
```diff
@@ -0,0 +1,295 @@
...
+
+#TODO: remove once FT.HYBRID for cluster is implemented
+@skip(cluster=True)
+def test_debug_timeout_return_both():
+    """Test RETURN policy with both components timeout using debug parameters"""
+    env = Env(enableDebugCommand=True, moduleArgs='ON_TIMEOUT RETURN')
+    setup_basic_index(env)
+    response = env.cmd('_FT.DEBUG', 'FT.HYBRID', 'idx', 'SEARCH', 'running', 'VSIM', '@embedding', '$BLOB', 'PARAMS', '2', 'BLOB', query_vector, 'TIMEOUT_AFTER_N_SEARCH', '1','TIMEOUT_AFTER_N_VSIM', '2', 'DEBUG_PARAMS_COUNT', '4')
+    env.assertTrue('Timeout limit was reached (SEARCH)' in get_warnings(response))
+    env.assertTrue('Timeout limit was reached (VSIM)' in get_warnings(response))
```
raz-mon, line 103: How will the warning section look like now?
Can you show a raw command and reply?

tests/pytests/test_hybrid_errors.py
```diff
@@ -0,0 +1,295 @@
...
+    """Test RETURN policy returns partial results when components timeout"""
+    env = Env(enableDebugCommand=True, moduleArgs='ON_TIMEOUT RETURN')
+    setup_basic_index(env)
+    # VSIM returns doc:2 and doc:4 (without timeout), SEARCH returns doc:3 (without timeout)
+    response = env.cmd('_FT.DEBUG', 'FT.HYBRID', 'idx', 'SEARCH', 'gear', 'VSIM', \
+                       '@embedding', '$BLOB', 'PARAMS', '2', 'BLOB', query_vector, \
+                       'TIMEOUT_AFTER_N_SEARCH', '1', 'TIMEOUT_AFTER_N_VSIM', '1', 'DEBUG_PARAMS_COUNT', '4')
+    results = get_results_from_hybrid_response(response)
+    env.assertTrue('doc:3' in results.keys())
+    env.assertTrue(('doc:2' in results.keys()) ^ ('doc:4' in results.keys()))
```
raz-mon: Let's have a tight assertion, i.e., make sure we get exactly what we expected, and not more

raz-mon: Also a comment why these only are expected would be nice

tests/pytests/test_hybrid_errors.py
```diff
@@ -0,0 +1,295 @@
...
+    def test_vector_only_fail(self):
+        """Test real timeout - vector only with FAIL policy"""
+        env = Env(moduleArgs='ON_TIMEOUT FAIL')
+        self._create_index(env)
+
+        # Populate only vectors
+        query_vector = load_vectors_to_redis(env, self.num_docs, query_vec_index=0, vec_size=self.dim, seed=10)
+
+        # Test vector timeout with FAIL policy using FT.HYBRID
+        env.expect('FT.HYBRID', 'idx', 'SEARCH', '*', 'VSIM', '@vector', '$BLOB',
```
raz-mon: Note that this is prone to be flaky. I would probably make one very big test that will test this "real" scenario, such that it will not be flaky. Anyways it's a bonus over the above

tests/pytests/test_hybrid_errors.py
raz-mon: Did that below - nice 👌 
But I would add more clarity to it, by setting the timeout here to be after `n > 1` results, such that you can check the behavior in a finer granularity

src/result_processor.c
```diff
@@ -1687,18 +1686,37 @@ dictType dictTypeHybridSearchResult = {
...
-   const RLookupKey *scoreKey;  // Key for writing score as field when QEXEC_F_SEND_SCORES_AS_FIELD is set
- } RPHybridMerger;
+ ResultProcessor base;
+ HybridScoringContext *hybridScoringCtx;  // Store by pointer - RPHybridMerger is responsible for freeing it
+ ResultProcessor **upstreams;  // Dynamic array of upstream processors
+ size_t numUpstreams;         // Number of upstream processors
+ dict *hybridResults;  // keyPtr -> HybridSearchResult mapping
+ dictIterator *iterator; // Iterator for yielding results
+ const RLookupKey *scoreKey;  // Key for writing score as field when QEXEC_F_SEND_SCORES_AS_FIELD is set
+ int *upstreamReturnCodes;  // Final return codes from each upstream
```
raz-mon: Why `int` and not the type of the return codes' enum, `RPStatus *`?
Same for the `.h` file

src/result_processor.c
```diff
@@ -1801,14 +1819,13 @@ dictType dictTypeHybridSearchResult = {
         // Upstream is still active but not ready to provide results. Skip to the next.
         continue;
       }
+
+      // Store the final return code for this upstream
+      self->upstreamReturnCodes[i] = rc;
+
       // Currently continues processing other upstreams.
       // TODO: Update logic to stop processing further results — we want to return immediately on timeout or error.
```
raz-mon: Handle?

src/result_processor.h
```diff
@@ -357,7 +357,8 @@ StrongRef DepleterSync_New(unsigned int num_depleters, bool take_index_lock);
  ResultProcessor *RPHybridMerger_New(HybridScoringContext *hybridScoringCtx,
                                      ResultProcessor **upstreams,
                                      size_t numUpstreams,
-                                     const RLookupKey *scoreKey);
+                                     const RLookupKey *scoreKey,
+                                     int *subqueriesReturnCodes);
```
raz-mon: Why not keep this as a field of the result-processor?

tests/pytests/test_hybrid_errors.py
```diff
@@ -0,0 +1,295 @@
...
+    env = Env(enableDebugCommand=True)
+    setup_basic_index(env)
+    conn = env.getClusterConnectionIfNeeded()
+    conn.execute_command('HSET', 'doc:5', 'description', 'runo')
+    conn.execute_command(config_cmd(), 'SET', 'MAXPREFIXEXPANSIONS', '1')
+
+    # Only SEARCH returns results, VSIM returns empty
+    response = env.cmd('FT.HYBRID', 'idx', 'SEARCH', 'run*', 'VSIM', \
+                       '@embedding', '$BLOB', 'RANGE', '2', 'RADIUS', '0.01', 'PARAMS', '2', 'BLOB', query_vector)
+    env.assertTrue('Max prefix expansions limit was reached (SEARCH)' in get_warnings(response))
```
raz-mon, line 135: We return the warnings only in RESP3 or profile, not in RESP2 as far as I remember, so how do we get them here?

src/hybrid/hybrid_exec.c
```diff
@@ -31,6 +31,38 @@
...
+static inline bool handleQueryError(RedisModule_Reply *reply, QueryError *err, int returnCode, const char *suffix, bool ignoreTimeout) {
+  bool timeoutOccurred = false;
+
+  if (returnCode == RS_RESULT_TIMEDOUT && !ignoreTimeout) {
+    ReplyWarning(reply, QueryError_Strerror(QUERY_ETIMEDOUT), suffix);
+    timeoutOccurred = true;
+  } else if (returnCode == RS_RESULT_ERROR) {
+    // Non-fatal error
+    ReplyWarning(reply, QueryError_GetUserError(err), suffix);
+  } else if (err->reachedMaxPrefixExpansions) {
```
raz-mon, line 64: This isn't reflected by the `rc`?

src/hybrid/hybrid_exec.c
```diff
@@ -31,6 +31,38 @@
...
+    char *expanded_warning = NULL;
+    rm_asprintf(&expanded_warning, "%s %s", message, suffix);
+    RedisModule_Reply_SimpleString(reply, expanded_warning);
+    rm_free(expanded_warning);
+  } else {
+    RedisModule_Reply_SimpleString(reply, message);
+  }
+}
+
+static inline bool handleQueryError(RedisModule_Reply *reply, QueryError *err, int returnCode, const char *suffix, bool ignoreTimeout) {
```
raz-mon: Add some documentation pls

src/hybrid/hybrid_exec.c
```diff
@@ -31,6 +31,38 @@
 
 #include <time.h>
 
+#define SEARCH_SUFFIX "(SEARCH)"
+#define VSIM_SUFFIX "(VSIM)"
+#define POST_PROCESSING_SUFFIX "(POST PROCESSING)"
+
+static inline void ReplyWarning(RedisModule_Reply *reply, const char *message, const char *suffix) {
```
raz-mon, line 39: Docs pls

src/hybrid/hybrid_exec.c
```diff
@@ -265,14 +298,17 @@ void sendChunk_hybrid(HybridRequest *hreq, RedisModule_Reply *reply, size_t limi
...
-      RedisModule_Reply_SimpleString(reply, QueryError_GetUserError(qctx->err));
-    } else if (qctx->err->reachedMaxPrefixExpansions) {
-      RedisModule_Reply_SimpleString(reply, QUERY_WMAXPREFIXEXPANSIONS);
+
+    bool timeoutInSubquery = false;
+    for (size_t i = 0; i < hreq->nrequests; ++i) {
+      QueryError* err = &hreq->errors[i];
+      const char* suffix = i == 0 ? SEARCH_SUFFIX : VSIM_SUFFIX;
+      const int subQueryReturnCode = hreq->subqueriesReturnCodes[i];
+      timeoutInSubquery = handleQueryError(reply, err, subQueryReturnCode, suffix, false) || timeoutInSubquery;
```
raz-mon:
```suggestion
    for (size_t i = 0; i < hreq->nrequests && !timeoutInSubquery; ++i) {
      QueryError* err = &hreq->errors[i];
      const char* suffix = i == 0 ? SEARCH_SUFFIX : VSIM_SUFFIX;
      const int subQueryReturnCode = hreq->subqueriesReturnCodes[i];
      timeoutInSubquery = handleQueryError(reply, err, subQueryReturnCode, suffix, false);
```

src/hybrid/hybrid_exec.c
```diff
@@ -265,14 +298,17 @@ void sendChunk_hybrid(HybridRequest *hreq, RedisModule_Reply *reply, size_t limi
     if (sctx->spec && sctx->spec->scan_failed_OOM) {
       RedisModule_Reply_SimpleString(reply, QUERY_WINDEXING_FAILURE);
     }
```
raz-mon, line 305: Should be part of the below logic?

src/hybrid/hybrid_exec.c
```diff
@@ -265,14 +298,17 @@ void sendChunk_hybrid(HybridRequest *hreq, RedisModule_Reply *reply, size_t limi
...
+
+    bool timeoutInSubquery = false;
+    for (size_t i = 0; i < hreq->nrequests; ++i) {
+      QueryError* err = &hreq->errors[i];
+      const char* suffix = i == 0 ? SEARCH_SUFFIX : VSIM_SUFFIX;
+      const int subQueryReturnCode = hreq->subqueriesReturnCodes[i];
+      timeoutInSubquery = handleQueryError(reply, err, subQueryReturnCode, suffix, false) || timeoutInSubquery;
     }
+    // Handle main query errors (POST PROCESSING)
+    handleQueryError(reply, qctx->err, rc, POST_PROCESSING_SUFFIX, timeoutInSubquery);
```
raz-mon: Can you show an example of how we want the warning to show in case of multiple errors (subqueries, tail)? This is weird to me that the subqueries errors will appear and then the tail right after it. No separators?

src/hybrid/hybrid_request.c
```diff
@@ -192,6 +188,9 @@ HybridRequest *HybridRequest_New(AREQ **requests, size_t nrequests) {
...
```
raz-mon:
```suggestion
    hybridReq->subqueriesReturnCodes = rm_calloc(nrequests, sizeof(RPStatus));
```

src/hybrid/hybrid_request.c
```diff
@@ -244,7 +243,13 @@ void HybridRequest_Free(HybridRequest *req) {
     }
     array_free(req->requests);
 
-    array_free(req->errors);
+    array_free_ex(req->errors, QueryError_ClearError((QueryError*)ptr));
+
+    // Free the return codes array
+    if (req->subqueriesReturnCodes) {
```
raz-mon: `if` needed?

src/hybrid/hybrid_exec.c
```diff
@@ -265,14 +298,17 @@ void sendChunk_hybrid(HybridRequest *hreq, RedisModule_Reply *reply, size_t limi
     if (sctx->spec && sctx->spec->scan_failed_OOM) {
```
raz-mon, line 303: `warning` --> `warnings`?
For line 296, just couldn't comment there (thanks GitHub)

src/hybrid/hybrid_debug.c
```diff
@@ -0,0 +1,276 @@
...
+  RedisSearchCtx *sctx = NewSearchCtxC(ctx, indexname, true);
+  if (!sctx) {
+    QueryError_SetWithUserDataFmt(&status, QUERY_ENOINDEX, "No such index", " %s", indexname);
+    return QueryError_ReplyAndClear(ctx, &status);
+  }
+
+  // Create debug hybrid request using the same sctx
+  HybridRequest_Debug *debug_req = HybridRequest_Debug_New(ctx, argv, argc, sctx, indexname, &status);
+  if (!debug_req) {
+    // parseHybridCommand takes ownership of sctx but doesn't free it on error - we need to clean it up
```
Itzikvaknin, line 243: parseHybridCommand should free it also on error

tests/pytests/test_hybrid_errors.py
Itzikvaknin: new command new response structure 💪

src/result_processor.h
Itzikvaknin: It is read by `sendChunk_hybrid`, so I prefer to pass it to the merger to fill, instead of reading from the pipeline internals.

src/result_processor.c
Itzikvaknin: Updated MOD-11004, Will handle it there

tests/pytests/test_rrf.py
Itzikvaknin: Removed

tests/pytests/test_hybrid_errors.py
Itzikvaknin: Changed all occurences

tests/pytests/test_hybrid_errors.py
Itzikvaknin: Its warnings, 
Will change the response keyword

tests/pytests/test_hybrid_errors.py
Itzikvaknin:
```
127.0.0.1:6379> FT.HYBRID idx SEARCH run* VSIM @embedding $BLOB FILTER "run*" PARAMS 2 BLOB "\x00\x00\x80\x3f\x00\x00\x80\x3f" 
 1) format
 2) STRING
 3) results
 4) 1) 1) attributes
       2) 1) 1) "__key"
             2) "doc:2"
             3) "__score"
             4) "0.032522474881"
    2) 1) attributes
       2) 1) 1) "__key"
             2) "doc:3"
             3) "__score"
             4) "0.032522474881"
 5) total_results
 6) (integer) 2
 7) warning
 8) 1) Max prefix expansions limit was reached (SEARCH)
    2) Max prefix expansions limit was reached (VSIM)
 9) execution_time
10) "0.161"
```

tests/pytests/test_hybrid_errors.py
Itzikvaknin: Left the XOR, so that it remains agnostic to the phase where the timeout occurred

src/hybrid/hybrid_request.c
Itzikvaknin: Done

src/hybrid/hybrid_request.c
Itzikvaknin: No, removed

src/hybrid/hybrid_exec.c
Itzikvaknin: I dont think so 

```
typedef enum {
  // Result is filled with valid data
  RS_RESULT_OK = 0,
  // Result is empty, and the last result has already been returned.
  RS_RESULT_EOF,
  // Execution paused due to rate limiting (or manual pause from ext. thread??)
  RS_RESULT_PAUSED,
  // Execution halted because of timeout
  RS_RESULT_TIMEDOUT,
  // Aborted because of error. The QueryState (parent->status) should have
  // more information.
  RS_RESULT_ERROR,
  // Depleting process has begun.
  RS_RESULT_DEPLETING,
  // Not a return code per se, but a marker signifying the end of the 'public'
  // return codes. Implementations can use this for extensions.
  RS_RESULT_MAX
} RPStatus;
```

src/result_processor.c
Itzikvaknin: Changed

src/hybrid/hybrid_exec.c
Itzikvaknin: We call `handleQueryError` on each upstream error, since they’re independent (e.g. timeout in one/both, timeout + MaxPrefixExpansions).

src/hybrid/hybrid_exec.c
Itzikvaknin: The logic below handles the pipeline level, this part is index level, so we should keep them separate IMO

tests/pytests/test_hybrid_errors.py
Itzikvaknin: Changed to single setup step 
```
  self.dim = 128
  self.num_docs = 100000
  self.doc_len = 20
```

src/debug_commands.c
Itzikvaknin: Added

src/hybrid/hybrid_exec.c
raz-mon: I now notice that this was added to RESP3 only in `FT.SEARCH/AGGREGATE`, and as an additional component of the array as well. I think that's breaking and shouldn't have been done there as well, but once it's there I guess we can have it the same here as well.

src/hybrid/hybrid_exec.c
raz-mon: I don't see how that relates to the comment

src/result_processor.h
raz-mon: I think both make sense in a way, but I would have the merger hold this data, as it is the one that manages the depletion of its upstreams, and thus it's the one that actually affects them so it should be the one to say their value.
It can expose an API for them, I think that makes more sense.
For instance, what happens if someone tries to use these values before they were updated? Only the merger knows he has yet to update the value

raz-mon (approved):

PR #6743, opened 2025-08-28
[MOD-11103] Inverted Index Reader and API

chesedo:
Creates a reader around the inverted index to prevent the iterator from accessing the index directly. This also establishes the API which the Rust implementation needs to expose for the reader.

#### Mark if applicable

### DIFF
--- a/src/inverted_index/inverted_index.c
+++ b/src/inverted_index/inverted_index.c
@@ -1287,3 +1287,186 @@ unsigned long InvertedIndex_MemUsage(const InvertedIndex *idx) {
   }
   return ret;
 }
+
+// pointer to the current block while reading the index
+#define CURRENT_BLOCK(ir) (InvertedIndex_BlockRef((ir)->idx, (ir)->currentBlock))
+#define CURRENT_BLOCK_READER_AT_END(ir) BufferReader_AtEnd(&(ir)->blockReader.buffReader)
+
+static inline void SetCurrentBlockReader(IndexReader *ir) {
+  ir->blockReader = (IndexBlockReader) {
+    NewBufferReader(IndexBlock_Buffer(CURRENT_BLOCK(ir))),
+    IndexBlock_FirstId(CURRENT_BLOCK(ir)),
+  };
+}
+
+static inline void AdvanceBlock(IndexReader *ir) {
+  ir->currentBlock++;
+  SetCurrentBlockReader(ir);
+}
+
+// A while-loop helper to advance the iterator to the next block or break if we are at the end.
+static __attribute__((always_inline)) inline bool NotAtEnd(IndexReader *ir) {
+  if (!CURRENT_BLOCK_READER_AT_END(ir)) {
+    return true; // still have entries in the current block
+  }
+  if (ir->currentBlock + 1 < InvertedIndex_NumBlocks(ir->idx)) {
+    // we have more blocks to read, so we can advance to the next block
+    AdvanceBlock(ir);
+    return true;
+  }
+  // no more blocks to read, so we are at the end
+  return false;
+}
+
+IndexReader *NewIndexReader(const InvertedIndex *idx, IndexDecoderCtx *ctx) {
+  IndexReader *ir = rm_calloc(1, sizeof(IndexReader));
+
+  ir->idx = idx;
+  ir->currentBlock = 0;
+  ir->gcMarker = InvertedIndex_GcMarker(idx);
+  ir->decoders = InvertedIndex_GetDecoder(InvertedIndex_Flags(idx));
+  ir->decoderCtx = *ctx;
+
+  SetCurrentBlockReader(ir);
+
+  return ir;
+}
+
+void IndexReader_Free(IndexReader *ir) {
+  rm_free(ir);
+}
+
+void IndexReader_Reset(IndexReader *ir) {
+  ir->currentBlock = 0;
+  ir->gcMarker = InvertedIndex_GcMarker(ir->idx);
+  SetCurrentBlockReader(ir);
+}
+
+size_t IndexReader_NumEstimated(const IndexReader *ir) {
+  return InvertedIndex_NumDocs(ir->idx);
+}
+
+
+bool IndexReader_IsIndex(const IndexReader *ir, const InvertedIndex *idx) {
+  RS_ASSERT(ir->idx);
+
+  return ir->idx == idx;
+}
+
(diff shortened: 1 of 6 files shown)


src/iterators/inverted_index_iterator.c
```diff
@@ -10,55 +10,25 @@
 #include "inverted_index_iterator.h"
 #include "redis_index.h"
 
-// pointer to the current block while reading the index
```
gdesmott, line 13: You can try running the `./tests/cpptests/test_cpp_iterator_index.cpp` to check if/how this affects performance.

src/inverted_index/inverted_index.c
```diff
@@ -1287,3 +1287,186 @@ unsigned long InvertedIndex_MemUsage(const InvertedIndex *idx) {
...
+  if (ir->currentBlock + 1 < InvertedIndex_NumBlocks(ir->idx)) {
+    // we have more blocks to read, so we can advance to the next block
+    AdvanceBlock(ir);
+    return true;
+  }
+  // no more blocks to read, so we are at the end
+  return false;
+}
+
+IndexReader *NewIndexReader(const InvertedIndex *idx, IndexDecoderCtx *ctx) {
```
gdesmott, line 1321: Could we stack allocate it instead? That would be faster.

chesedo: Hmm, that will mean the Rust version will have to expose its internals (to know the size). I will want to avoid that if possible.

gdesmott: right, let's check performance first then.

src/iterators/inverted_index_iterator.c
chesedo: I get a permission denied with that command. But was able to run this `./bin/linux-x64-release/search-community/micro-benchmarks/benchmark_index_iterator`.

The results are about the same as master. Some are faster and others slower. And these change between different runs.

src/inverted_index/inverted_index.c
```diff
@@ -1287,3 +1287,186 @@ unsigned long InvertedIndex_MemUsage(const InvertedIndex *idx) {
...
+  };
+}
+
+static inline void AdvanceBlock(IndexReader *ir) {
+  ir->currentBlock++;
+  SetCurrentBlockReader(ir);
+}
+
+// A while-loop helper to advance the iterator to the next block or break if we are at the end.
+static __attribute__((always_inline)) inline bool NotAtEnd(IndexReader *ir) {
```
GlenDC, line 1308: why always inline? Is it proven to be actually positive? Not better to let the compiler figure this one out?

GlenDC (approved):

src/inverted_index/inverted_index.c
chesedo: All the logic in this file was just moved as is from the iterator file. I'm not sure why it was always inlined there.

@GuyAv46 I see you made this change in #6358

tests/cpptests/test_cpp_iterator_index.cpp
```diff
@@ -992,25 +992,24 @@ TEST_P(InvIndIteratorRevalidateTest, RevalidateAfterIndexDisappears) {
...
-            *((InvertedIndex **)&invIt->idx) = dummyIdx;
+            // Temporarily replace the iterator's index pointer
+            IndexReader_SwapIndex(invIt->reader, dummyIdx);
 
             // Now Revalidate should return VALIDATE_ABORTED because the stored index
             // doesn't match what the lookup returns
             ASSERT_EQ(iterator->Revalidate(iterator), VALIDATE_ABORTED);
 
+            // Restore the original index pointer for proper cleanup
+            IndexReader_SwapIndex(invIt->reader, dummyIdx);
```
alonre24, line 1009: Shouldn't it be `IndexReader_SwapIndex(invIt->reader, originalIdx)`? If so, how come the test did not fail on that?

tests/cpptests/test_cpp_iterator_index.cpp
```diff
@@ -1159,8 +1157,4 @@ TEST_P(InvIndIteratorRevalidateTest, RevalidateAfterDocumentDeleted) {
     }
 
     DocTable_Free(&tempDocTable);
-
-    // Restore the original index state
-    InvertedIndex_SetNumDocs(idx, InvertedIndex_NumDocs(idx) + repairParams.entriesCollected); // Restore original numDocs
-    InvertedIndex_SetGcMarker(idx, originalIndexGcMarker);
```
alonre24, line 1165: Not needed?

tests/cpptests/test_cpp_iterator_index.cpp
```diff
@@ -1048,9 +1047,8 @@ TEST_P(InvIndIteratorRevalidateTest, RevalidateAfterDocumentDeleted) {
     // 4. This should cause SkipTo(thirdDocId) to return ITERATOR_NOTFOUND
 
     InvIndIterator *invIt = (InvIndIterator *)iterator;
-    InvertedIndex *idx = const_cast<InvertedIndex *>(invIt->idx);
-    uint32_t originalGcMarker = invIt->gcMarker;
-    uint32_t originalIndexGcMarker = InvertedIndex_GcMarker(idx);
```
alonre24, line 1053: Was `originalIndexGcMarker` redundant here?

alonre24 (commented):
LGTM
Can we confirm that the inv index iterator's cpp unit tests passing ?

tests/cpptests/test_cpp_iterator_index.cpp
chesedo: No, line 1002 swapped the original index with the dummy. So by this line the `dummyIdx` pointer points to the original which this now swaps back.

tests/cpptests/test_cpp_iterator_index.cpp
chesedo: Yes, I could not figure out why this is needed. Nothing is tested after this code.

tests/cpptests/test_cpp_iterator_index.cpp
chesedo: Because I removed setting the original GC marker back (other review comment about the code not being needed), it is no longer needed to get the original marker here. Hence why it is removed.

chesedo:
> LGTM Can we confirm that the inv index iterator's cpp unit tests passing ?

Yip, all the tests passed locally and the CI is also green

PR #6744, opened 2025-08-28
MOD-11132: Fix direct blob issue in FT.HYBRID

Itzikvaknin:
Before: 
```
FT.HYBRID idx SEARCH "run*" VSIM @embedding "\x00\x00\x80\x3f\x00\x00\x80\x3f"
(error) Error parsing vector similarity query: query vector blob size (0) does not match index's expected size (8)
```

After:
```
FT.HYBRID idx SEARCH green VSIM @embedding "\x9a\x99\x99\x3f\xcd\xcc\x4c\x3e" KNN 2 K 1
 1) format
 2) STRING
 3) results
 4) 1) 1) attributes
       2) 1) 1) "__key"
             2) "doc:2"
             3) "__score"
             4) "0.016393442623"
 5) total_results
 6) (integer) 1
 7) warning
 8) (empty array)
 9) execution_time
10) "1.135"
```

Syntactical and logical correctness are verified in `tests/pytests/test_hybrid_vector.py`


tests/pytests/common.py
```diff
@@ -942,3 +943,58 @@ def assertEqual_dicts_on_intersection(env, d1, d2, message=None, depth=0):
...
+        score_index = recursive_index(result, '__score')
+        score_index[-1] += 1
+
+        key = access_nested_list(result, key_index)
+        score = float(access_nested_list(result, score_index))
+
+        results[key] = score
+    return results
+
+def populate_db_with_faker_text(env, num_docs, doc_len=5, seed=12345, offset=0):
```
nafraf, line 973: Are we using this function in this PR?

tests/pytests/test_hybrid_vector.py
```diff
@@ -0,0 +1,71 @@
...
+from includes import *
+from common import *
+
+"""
+VECTOR SPACE LAYOUT:
+====================
+
+The test data creates a 2D vector space with 4 documents positioned as follows:
+
+    doc:3 ●────────────● doc:4
```
nafraf, line 11: Nice!! 🥇

tests/pytests/common.py
Itzikvaknin: No, i took the `common.py` changes as is from  https://github.com/RediSearch/RediSearch/pull/6737 to avoid merges, we use this function in https://github.com/RediSearch/RediSearch/pull/6737 to populate the db with alot of documents and then get a real timeout

kei-nan (approved):

nafraf (approved):

PR #6745, opened 2025-08-28
MOD-11047: Rename RRF K parameter to CONSTANT

nafraf:
## Description
Rename the RRF parameter `K` to `CONSTANT` to avoid confusion (with KNN parameter `K`)

Before:
```shell
SEARCH '@text:(both)' 
   VSIM @vector $BLOB
   KNN 2 K 10
   COMBINE RRF 2 K 50
```

After:
```shell
SEARCH '@text:(both)' 
   VSIM @vector $BLOB
   KNN 2 K 10
   COMBINE RRF 2 CONSTANT 50
```


Itzikvaknin (approved):
LGTM

PR #6746, opened 2025-08-28
[MOD-11082] remove offsets_tz var in RSIndexResult

GlenDC:
this var is duplicate info as it is the same as the offset length, so no need to store it separately

Closes [MOD-11082](https://redislabs.atlassian.net/browse/MOD-11082)

[MOD-11082]: https://redislabs.atlassian.net/browse/MOD-11082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/forward_index.c
+++ b/src/forward_index.c
@@ -272,7 +272,6 @@ int forwardIndexTokenFunc(void *ctx, const Token *tokInfo) {
 size_t InvertedIndex_WriteForwardIndexEntry(InvertedIndex *idx, ForwardIndexEntry *ent) {
   RSIndexResult rec = {.data.term_tag = RSResultData_Term,
                        .docId = ent->docId,
-                       .offsetsSz = VVW_GetByteLength(ent->vw),
                        .freq = ent->freq,
                        .fieldMask = ent->fieldMask};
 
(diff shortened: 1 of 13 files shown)


src/forward_index.c
```diff
@@ -272,7 +272,6 @@ int forwardIndexTokenFunc(void *ctx, const Token *tokInfo) {
 size_t InvertedIndex_WriteForwardIndexEntry(InvertedIndex *idx, ForwardIndexEntry *ent) {
   RSIndexResult rec = {.data.term_tag = RSResultData_Term,
                        .docId = ent->docId,
-                       .offsetsSz = VVW_GetByteLength(ent->vw),
                        .freq = ent->freq,
                        .fieldMask = ent->fieldMask};
```
chesedo, line 276: Shouldn't this set something on `.data.term`?

GlenDC: This happens below @chesedo , the data will be set and the data is the truth of source for the offset as well. Am I misunderstand something here?

chesedo: Ahh, I missed that. Thanks!

GlenDC:
@alonre24 this PR requires a reviewer from you or someone of your team.

GlenDC:
Rebased it @chesedo

gdesmott:
> LGTM Haven't we migrated all the encoders / decoders already? @gdesmott

We have a Rust verison of all encoders/decoders but are still using the C ones atm. The switch will be part of the swap to the Rust II.

PR #6747, opened 2025-08-29
[MOD-11098] Have a wrapper to track the number of entries in an inverted index

chesedo:
Rather than tracking this in the numeric encoder, this creates a new wrapper to track the entries instead.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -333,6 +333,55 @@ impl<E: Encoder + DecodedBy> InvertedIndex<E> {
     }
 }
 
+pub struct EntriesTrackingIndex<E> {
+    /// The underlying inverted index that stores the entries.
+    index: InvertedIndex<E>,
+
+    /// The total number of entries in the index. This is not the number of unique documents, but
+    /// rather the total number of entries added to the index.
+    number_of_entries: usize,
+}
+
+impl<E: Encoder> EntriesTrackingIndex<E> {
+    /// Create a new entries tracking index with the given encoder.
+    pub fn new(encoder: E) -> Self {
+        Self {
+            index: InvertedIndex::new(encoder),
+            number_of_entries: 0,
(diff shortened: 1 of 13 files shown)


gdesmott (commented):
What are the use cases for this?
is it worth having an extra trait rather than implementing it directly on the II?

gdesmott (approved):

gdesmott (approved):

LukeMathWalker:
How does this fit into the bigger picture?
Will we be tracking the number of entries for _all_ index kinds? Or will it be limited to numeric? If the latter, how does the rest of the C code interact with this wrapper where they now expect a pointer to an inverted index (i.e. a specific concrete type)?

GlenDC (approved):

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -333,6 +333,57 @@ impl<E: Encoder + DecodedBy> InvertedIndex<E> {
...
+            index: InvertedIndex::new(encoder),
+            number_of_entries: 0,
+        }
+    }
+
+    /// Add a new record to the index and return by how much memory grew. It is expected that
+    /// the document ID of the record is greater than or equal the last document ID in the index.
+    ///
+    /// The total number of entries in the index is incremented by one.
+    pub fn add_record(&mut self, record: &RSIndexResult) -> std::io::Result<usize> {
```
alonre24, line 360: Who is expected to call that in prod code? It should be bindded to `encode()` somehow?

chesedo: This will be called when writing entries to an inverted index. More specifically, when the `StoreNumeric` flag is set on an inverted index then it will be routed through this wrapper.

We don't really want to bind this to the encoder since that start giving the encoders a responsibility of more than just decoding. Hence why this PR also removes this logic from the encoder.

alonre24 (approved):

PR #6750, opened 2025-08-31
[MOD-11155] Create basic query pause debug command

lerman25:
This PR adds a basic query pause mechanism to debug commands, using existing mock timeout mechanism. 
This will help create stable tests for Out Of Memory query handling.

This PR also restructure the result processors files, such that the debug related code is clearly separated. 

This PR introduces `RPPauseAfterCount` result processor, which is create when a `_FT.DEBUG <QUERY>` is given `PAUSE_AFTER_N` parameter. 
This RP will pause the execution of the query, until a resume command is sent. 
The resume command is sent by a new debug command - `QUERY_CONTROLLER`.
The `QUERY_CONTROLLER` can be given (currently) either  `SET_PAUSE_RP_RESUME` or  `GET_IS_RP_PAUSED`.
`SET_PAUSE_RP_RESUME` - will resume the execution of the query .
`GET_IS_RP_PAUSED` - will indicate if a query is paused. 

This mechanism only supports one query at a time, and must be used with **WORKERS** > 0.

### DIFF
--- a/src/aggregate/aggregate_debug.c
+++ b/src/aggregate/aggregate_debug.c
@@ -82,10 +82,19 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
     return REDISMODULE_ERR;
   }
 
+  // Handle crash
   if (crash) {
     PipelineAddCrash(&debug_req->r);
   }
 
+  // Error handling: Verify internal_only is used with timeout
+  if (internal_only && !AC_IsInitialized(&timeoutArgs)) {
+    QueryError_SetError(status, QUERY_EPARSEARGS,
+                        "INTERNAL_ONLY must be used with TIMEOUT_AFTER_N");
+    return REDISMODULE_ERR;
+  }
+
+  // Handle timeout
   if (AC_IsInitialized(&timeoutArgs)) {
     unsigned long long results_count = -1;
     if (AC_GetUnsignedLongLong(&timeoutArgs, &results_count, AC_F_GE0) != AC_OK) {
@@ -115,13 +124,10 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
       // Take this into account when adding more debug types that are modifying the rp pipeline.
       PipelineAddTimeoutAfterCount(&debug_req->r, results_count);
     }
-  } else {
-    if (internal_only) {
-      QueryError_SetError(status, QUERY_EPARSEARGS,
-                          "INTERNAL_ONLY must be used with TIMEOUT_AFTER_N");
-      return REDISMODULE_ERR;
-    }
   }
+
+  // Handle out of memory
+
 
   return REDISMODULE_OK;
 }
--- a/src/aggregate/aggregate_debug.c
+++ b/src/aggregate/aggregate_debug.c
@@ -54,6 +54,7 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
   ArgsCursor timeoutArgs = {0};
   int crash = 0;
   int internal_only = 0;
+  ArgsCursor pauseArgs = {0};
   ACArgSpec debugArgsSpec[] = {
       // Getting TIMEOUT_AFTER_N as an array to use AC_IsInitialized API.
       {.name = "TIMEOUT_AFTER_N",
@@ -64,6 +65,11 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
       {.name = "CRASH", .type = AC_ARGTYPE_BOOLFLAG, .target = &crash},
       // optional arg for TIMEOUT_AFTER_N
       {.name = "INTERNAL_ONLY", .type = AC_ARGTYPE_BOOLFLAG, .target = &internal_only},
+      // pause after N results
+      {.name = "PAUSE_AFTER_N",
+       .type = AC_ARGTYPE_SUBARGS_N,
+       .target = &pauseArgs,
+       .slicelen = 1},
       {NULL}};
 
   ACArgSpec *errSpec = NULL;
@@ -126,7 +132,15 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
     }
   }
 
-  // Handle out of memory
+  // Handle pause
+  if (AC_IsInitialized(&pauseArgs)) {
+    unsigned long long results_count = -1;
+    if (AC_GetUnsignedLongLong(&pauseArgs, &results_count, AC_F_GE0) != AC_OK) {
+      QueryError_SetError(status, QUERY_EPARSEARGS, "Invalid PAUSE_AFTER_N count");
+      return REDISMODULE_ERR;
+    }
+    PipelineAddPauseAfterCount(&debug_req->r, results_count);
+  }
 
 
   return REDISMODULE_OK;
--- a/src/debug_commands.c
+++ b/src/debug_commands.c
@@ -1830,6 +1830,68 @@ DEBUG_COMMAND(YieldCounter) {
   return RedisModule_ReplyWithLongLong(ctx, g_yieldCallCounter);
 }
 
+/**
+ * FT.DEBUG QUERY_CONTROLLER SET_PAUSE_RP_RESUME
+ */
+DEBUG_COMMAND(setPauseRPResume) {
+  if (!debugCommandsEnabled(ctx)) {
+    return RedisModule_ReplyWithError(ctx, NODEBUG_ERR);
+  }
+  if (argc != 2) {
+    return RedisModule_WrongArity(ctx);
+  }
+
+  globalDebugCtx.query.pause = false;
+
+  return RedisModule_ReplyWithSimpleString(ctx, "OK");
+}
+
+/**
+ * FT.DEBUG QUERY_CONTROLLER GET_IS_RP_PAUSED
+ */
+DEBUG_COMMAND(getIsRPPaused) {
+  if (!debugCommandsEnabled(ctx)) {
+    return RedisModule_ReplyWithError(ctx, NODEBUG_ERR);
+  }
+  if (argc != 2) {
+    return RedisModule_WrongArity(ctx);
+  }
+
+  // Check availability of the debug RP
+  if (globalDebugCtx.query.debugRP == NULL) {
+    return RedisModule_ReplyWithError(ctx, "Result processor is not available");
+  }
+
+  // Verify the type of the debug RP
+  if (globalDebugCtx.query.debugRP->type != RP_PAUSE) {
+    return RedisModule_ReplyWithError(ctx, "Result processor is not of type PAUSE");
+  }
+
+  return RedisModule_ReplyWithLongLong(ctx, globalDebugCtx.query.pause);
+}
+
+/**
+ * FT.DEBUG QUERY_CONTROLLER <command> [options]
+ */
+DEBUG_COMMAND(queryController) {
+  if (!debugCommandsEnabled(ctx)) {
+    return RedisModule_ReplyWithError(ctx, NODEBUG_ERR);
+  }
+  if (argc < 3) {
+    return RedisModule_WrongArity(ctx);
+  }
+  const char *op = RedisModule_StringPtrLen(argv[2], NULL);
+
+  // Check here all background indexing possible commands
+  if (!strcmp("SET_PAUSE_RP_RESUME", op)) {
+    return setPauseRPResume(ctx, argv + 1, argc - 1);
+  }
+  if (!strcmp("GET_IS_RP_PAUSED", op)) {
+    return getIsRPPaused(ctx, argv + 1, argc - 1);
+  }
+  return RedisModule_ReplyWithError(ctx, "Invalid command for 'QUERY_CONTROLLER'");
+}
+
 DebugCommandType commands[] = {{"DUMP_INVIDX", DumpInvertedIndex}, // Print all the inverted index entries.
                                {"DUMP_NUMIDX", DumpNumericIndex}, // Print all the headers (optional) + entries of the numeric tree.
                                {"DUMP_NUMIDXTREE", DumpNumericIndexTree}, // Print tree general info, all leaves + nodes + stats
@@ -1866,6 +1928,7 @@ DebugCommandType commands[] = {{"DUMP_INVIDX", DumpInvertedIndex}, // Print all 
                                {"INFO", IndexObfuscatedInfo},
                                {"GET_HIDE_USER_DATA_FROM_LOGS", getHideUserDataFromLogs},
                                {"YIELDS_ON_LOAD_COUNTER", YieldCounter},
+                               {"QUERY_CONTROLLER", queryController},
                                /**
                                 * The following commands are for debugging distributed search/aggregation.
                                 */
--- a/src/debug_commands.h
+++ b/src/debug_commands.h
@@ -33,11 +33,20 @@ typedef struct BgIndexingDebugCtx {
 
 } BgIndexingDebugCtx;
 
+// Struct used for debugging queries
+// Note: unrelated to timeout debugging
+typedef struct QueryDebugCtx {
+  volatile atomic_bool pause; // Volatile atomic bool to wait for the resume command
+  bool pauseOnOOM; // Whether to pause on OOM
+} QueryDebugCtx;
+
 // General debug context
 typedef struct DebugCTX {
   bool debugMode; // Indicates whether debug mode is enabled
   BgIndexingDebugCtx bgIndexing; // Background indexing debug context
+  QueryDebugCtx query; // Query debug context
 } DebugCTX;
 
 // Should be called after each debug command that changes the debugCtx
+// Exception for QueryDebugCtx
 void validateDebugMode(DebugCTX *debugCtx);
--- a/src/debug_commands.h
+++ b/src/debug_commands.h
@@ -11,6 +11,7 @@
 #include "redismodule.h"
 #include  <stdbool.h>
 #include <stdatomic.h>
+#include "result_processor.h"
 
 #define RS_DEBUG_FLAGS 0, 0, 0
 #define DEBUG_COMMAND(name) static int name(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
@@ -38,6 +39,7 @@ typedef struct BgIndexingDebugCtx {
 typedef struct QueryDebugCtx {
   volatile atomic_bool pause; // Volatile atomic bool to wait for the resume command
   bool pauseOnOOM; // Whether to pause on OOM
+  ResultProcessor *debugRP; // Result processor for debugging, supports debugging one query at a time
 } QueryDebugCtx;
 
 // General debug context
--- a/src/result_processor.c
+++ b/src/result_processor.c
@@ -1178,6 +1178,115 @@ ResultProcessor *RPCounter_New() {
   return &ret->base;
 }
 
+ /*******************************************************************************************************************
+   *  Max Score Normalizer Result Processor
+   *
+   * This result processor normalizes the scores of search results using division by
+   * the max score. It gathers all results from the upstream processor, finds the
+   * maximum score, and divides each score by the maximum. This ensures that all scores
+   * fall within the range [0, 1].
+   *
+   * The processor works in two phases:
+   * 1. Accumulation: Gather all results from upstream and find the max score.
+   * 2. Yield: Normalize each result’s score by division with the max score, then pass
+   *    it downstream.
+  *******************************************************************************************************************/
+ typedef struct {
+   ResultProcessor base;
+   // Stores the max value found (if needed in the future)
+   double maxValue;
+   const RLookupKey *scoreKey;
+   SearchResult *pooledResult;
+   arrayof(SearchResult *) pool;
+   bool timedOut;
+ } RPMaxScoreNormalizer;
+
+
+ static void RPMaxScoreNormalizer_Free(ResultProcessor *base) {
+   RPMaxScoreNormalizer *self = (RPMaxScoreNormalizer *)base;
+   array_free_ex(self->pool, srDtor(*(char **)ptr));
+   srDtor(self->pooledResult);
+   rm_free(self);
+ }
+
+ static int RPMaxScoreNormalizer_Yield(ResultProcessor *rp, SearchResult *r){
+   RPMaxScoreNormalizer* self = (RPMaxScoreNormalizer*)rp;
+   size_t length = array_len(self->pool);
+   if (length == 0) {
+    // We've already yielded all results, return EOF
+    int ret = self->timedOut ? RS_RESULT_TIMEDOUT : RS_RESULT_EOF;
+    self->timedOut = false;
+    return ret;
+   }
+  SearchResult *poppedResult = array_pop(self->pool);
+  SearchResult_Override(r, poppedResult);
+  rm_free(poppedResult);
+  double oldScore = r->score;
+  if (self->maxValue != 0) {
+    r->score /= self->maxValue;
+  }
+  if (self->scoreKey) {
+    RLookup_WriteOwnKey(self->scoreKey, &r->rowdata, RS_NumVal(r->score));
+  }
+  EXPLAIN(r->scoreExplain,
+        "Final BM25STD.NORM: %.2f = Original Score: %.2f / Max Score: %.2f",
+        r->score, oldScore, self->maxValue);
+  return RS_RESULT_OK;
+ }
+
+static int RPMaxScoreNormalizerNext_innerLoop(ResultProcessor *rp, SearchResult *r) {
+  RPMaxScoreNormalizer *self = (RPMaxScoreNormalizer *)rp;
+  // get the next result from upstream. `self->pooledResult` is expected to be empty and allocated.
+  int rc = rp->upstream->Next(rp->upstream, self->pooledResult);
+  // if our upstream has finished - just change the state to not accumulating, and yield
+  if (rc == RS_RESULT_EOF) {
+    rp->Next = RPMaxScoreNormalizer_Yield;
+    return rp->Next(rp, r);
+  } else if (rc == RS_RESULT_TIMEDOUT && (rp->parent->timeoutPolicy == TimeoutPolicy_Return)) {
+    self->timedOut = true;
+    rp->Next = RPMaxScoreNormalizer_Yield;
+    return rp->Next(rp, r);
+  } else if (rc != RS_RESULT_OK) {
+    return rc;
+  }
+
+  self->maxValue = MAX(self->maxValue, self->pooledResult->score);
+  // copy the index result to make it thread safe - but only if it is pushed to the heap
+  self->pooledResult->indexResult = NULL;
+  array_ensure_append_1(self->pool, self->pooledResult);
+
+  // we need to allocate a new result for the next iteration
+  self->pooledResult = rm_calloc(1, sizeof(*self->pooledResult));
+  return RESULT_QUEUED;
+}
+
+static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
+  RPMaxScoreNormalizer *self = (RPMaxScoreNormalizer *)rp;
+  uint32_t chunkLimit = rp->parent->resultLimit;
+  rp->parent->resultLimit = UINT32_MAX; // we want to accumulate all results
+  int rc;
+  while ((rc = RPMaxScoreNormalizerNext_innerLoop(rp, r)) == RESULT_QUEUED) {};
+  rp->parent->resultLimit = chunkLimit; // restore the limit
+  return rc;
+}
+
+ /* Create a new Max Collector processor */
+ ResultProcessor *RPMaxScoreNormalizer_New(const RLookupKey *rlk) {
+  RPMaxScoreNormalizer *ret = rm_calloc(1, sizeof(*ret));
+  ret->pooledResult = rm_calloc(1, sizeof(*ret->pooledResult));
+  ret->pool = array_new(SearchResult*, 0);
+  ret->base.Next = RPMaxScoreNormalizer_Accum;
+  ret->base.Free = RPMaxScoreNormalizer_Free;
+  ret->base.type = RP_MAX_SCORE_NORMALIZER;
+  ret->scoreKey = rlk;
+  return &ret->base;
+}
+
+/*******************************************************************************************************************
+ *  Debug only result processors
+ *
+ * *******************************************************************************************************************/
+
 /*******************************************************************************************************************
  *  Timeout Processor - DEBUG ONLY
  *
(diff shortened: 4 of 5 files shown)


src/aggregate/aggregate_debug.c
```diff
@@ -115,12 +142,40 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
...
+      QueryError_SetWithoutUserDataFmt(status, QUERY_EPARSEARGS, "Invalid %s count", invalidStr);
+      return REDISMODULE_ERR;
+    }
+    const char *rp_type_str = NULL;
+    if (AC_GetString(pauseArgs, &rp_type_str, NULL, 0) != AC_OK) {
+      QueryError_SetWithoutUserDataFmt(status, QUERY_EPARSEARGS, "Invalid %s RP type", invalidStr);
       return REDISMODULE_ERR;
     }
+    ResultProcessorType rp_type = StringToRPType(rp_type_str);
+    if (rp_type == RP_MAX) {
```
meiravgri, line 171: Since RP_MAX counter doesn't include debugger RPs.
Please add a test to verify we get an error if trying to use one of the debug RPs
```suggestion
    if (rp_type >= RP_MAX) {
```

src/result_processor.c
```diff
@@ -1071,6 +1072,15 @@ const char *RPTypeToString(ResultProcessorType type) {
   return RPTypeLookup[type];
 }
 
+ResultProcessorType StringToRPType(const char *str) {
+  for (int i = 0; i < RP_MAX; i++) {
+    if (!strcmp(str, RPTypeLookup[i])) {
+      return i;
```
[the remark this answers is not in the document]
lerman25: How would that cause issues?

src/result_processor.c
```diff
@@ -1303,106 +1506,73 @@ void PipelineAddCrash(struct AREQ *r) {
...
-    return rp->Next(rp, r);
-  } else if (rc != RS_RESULT_OK) {
-    return rc;
+static void RPPauseAfterCount_Pause(RPPauseAfterCount *self) {
+
+  globalDebugCtx.query.pause = true;
+  self->paused = true;
+  while (globalDebugCtx.query.pause) { // volatile variable
+    usleep(1000);
   }
```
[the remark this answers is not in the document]
lerman25: It's non production, I think it's better to keep it simple

src/result_processor.c
```diff
@@ -1303,106 +1506,73 @@ void PipelineAddCrash(struct AREQ *r) {
...
-  ret->scoreKey = rlk;
+ResultProcessor *RPPauseAfterCount_New(size_t count) {
+  RPPauseAfterCount *ret = rm_calloc(1, sizeof(RPPauseAfterCount));
+  ret->count = count;
+  ret->remaining = count;
+  ret->base.type = RP_PAUSE;
+  ret->base.Next = RPPauseAfterCount_Next;
+  ret->base.Free = RPPauseAfterCount_Free;
+
+  globalDebugCtx.query.debugRP = &ret->base;
```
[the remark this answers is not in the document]
lerman25: It's non production, I think it's better to keep it simple

src/debug_commands.h
```diff
@@ -33,11 +34,21 @@ typedef struct BgIndexingDebugCtx {
 
 } BgIndexingDebugCtx;
 
+// Struct used for debugging queries
+// Note: unrelated to timeout debugging
+typedef struct QueryDebugCtx {
+  volatile atomic_bool pause; // Volatile atomic bool to wait for the resume command
```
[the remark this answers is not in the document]
lerman25: In the fear of making existing tests flaky, I'd rather not (I don't have much knowledge here).
@meiravgri WDYT?

src/result_processor.h
```diff
@@ -283,6 +285,22 @@ void Profile_AddRPs(QueryProcessingCtx *qiter);
 // Return string for RPType
 const char *RPTypeToString(ResultProcessorType type);
 
+// Return RPType for string
+ResultProcessorType StringToRPType(const char *str);
```
meiravgri: Why isn't this under the DEBUG section?

src/result_processor.h
```diff
@@ -299,13 +317,15 @@ void PipelineAddTimeoutAfterCount(struct AREQ *r, size_t results_count);
...
- ResultProcessor *RPMaxScoreNormalizer_New(const RLookupKey *rlk);
+/*******************************************************************************************************************
+ *  Pause Processor - DEBUG ONLY
+ *
+ * Pauses the query after N results, N >= 0.
+ *******************************************************************************************************************/
+ResultProcessor *RPPauseAfterCount_New(size_t count);
+void PipelineAddPauseAfterCount(struct AREQ *r, size_t results_count);
+void PipelineAddPauseRPcount(struct AREQ *r, size_t results_count, bool before, ResultProcessorType rp_type);
+
```
meiravgri: what's the difference? Add a short description

src/result_processor.c
```diff
@@ -1303,106 +1506,73 @@ void PipelineAddCrash(struct AREQ *r) {
...
-   // Stores the max value found (if needed in the future)
-   double maxValue;
-   const RLookupKey *scoreKey;
-   SearchResult *pooledResult;
-   arrayof(SearchResult *) pool;
-   bool timedOut;
- } RPMaxScoreNormalizer;
+/*******************************************************************************************************************
+ *  Pause Processor - DEBUG ONLY
+ * TBD
```
meiravgri: ? :)

src/result_processor.c
```diff
@@ -1303,106 +1506,73 @@ void PipelineAddCrash(struct AREQ *r) {
...
-    RLookup_WriteOwnKey(self->scoreKey, &r->rowdata, RS_NumVal(r->score));
+void PipelineAddPauseAfterCount(AREQ *r, size_t results_count) {
+  ResultProcessor *RPPauseAfterCount = RPPauseAfterCount_New(results_count);
+  addResultProcessor(r, RPPauseAfterCount);
+}
+
+void PipelineAddPauseRPcount(AREQ *r, size_t results_count, bool before, ResultProcessorType rp_type) {
+  ResultProcessor *RPPauseAfterCount = RPPauseAfterCount_New(results_count);
+  if (before) {
+    addResultProcessorBeforeType(r, RPPauseAfterCount, rp_type);
```
meiravgri: check return code and propagate (change this function's return type) if there's an error (rp not found)
add tests to verify the query fails in this case

src/result_processor.c
```diff
@@ -1212,6 +1318,103 @@ static void addResultProcessor(AREQ *r, ResultProcessor *rp) {
...
+      cur = cur->upstream;
+    }
+    // Update the endProc to the new head in case it was changed
+    r->qiter.endProc = dummyHead.upstream;
+    return true;
+  }
+
+  // Target type not found
+  return false;
+}
```
meiravgri:
```suggestion
    }
  return found;
}
```

src/result_processor.c
```diff
@@ -1212,6 +1318,103 @@ static void addResultProcessor(AREQ *r, ResultProcessor *rp) {
...
+  ResultProcessor *downstream = &dummyHead;
+
+  bool found = false;
+
+  // Search for the target result processor type
+  while (cur) {
+    // Change downstream -> cur(type) -> cur->upstream
+    // To: downstream -> rp -> cur(type) -> cur->upstream
+    if (cur->type == target_type) {
+      rp->parent = &r->qiter;
```
meiravgri: where are you using `&r->qiter`?
If needed, i suggest to move to the rp intiliztion as its not related to what this function does.

src/result_processor.c
```diff
@@ -1212,6 +1318,103 @@ static void addResultProcessor(AREQ *r, ResultProcessor *rp) {
...
+  bool found = false;
+
+  // Search for the target result processor type
+  while (cur) {
+    // Change downstream -> cur(type) -> cur->upstream
+    // To: downstream -> cur(type) -> rp-> cur->upstream
+    if (cur->type == target_type) {
+      rp->upstream = cur->upstream;
+      cur->upstream = rp;
+      rp->parent = &r->qiter;
```
meiravgri: same

src/result_processor.c
```diff
@@ -1212,6 +1318,103 @@ static void addResultProcessor(AREQ *r, ResultProcessor *rp) {
...
+
+  // Target type not found
+  return false;
+}
+
+// Insert the result processor after the first occurrence of a specific RP type in the upstream
+static bool addResultProcessorAfterType(AREQ *r, ResultProcessor *rp, ResultProcessorType target_type) {
+  ResultProcessor *cur = r->qiter.endProc;
+  ResultProcessor dummyHead = { .upstream = cur };
+  ResultProcessor *downstream = &dummyHead;
```
meiravgri: is this in use?

src/result_processor.c
```diff
@@ -1212,6 +1318,103 @@ static void addResultProcessor(AREQ *r, ResultProcessor *rp) {
...
+  if (found) {
+    while (cur) {
+      if (!cur->upstream) {
+        break;
+      }
+      downstream = cur;
+      cur = cur->upstream;
+    }
+    // Update the endProc to the new head in case it was changed
+    r->qiter.endProc = dummyHead.upstream;
```
meiravgri: is it possible?

src/result_processor.c
```diff
@@ -1212,6 +1318,103 @@ static void addResultProcessor(AREQ *r, ResultProcessor *rp) {
...
+    return true;
+  }
+
+  // Target type not found
+  return false;
+}
+
+// Insert the result processor after the first occurrence of a specific RP type in the upstream
+static bool addResultProcessorAfterType(AREQ *r, ResultProcessor *rp, ResultProcessorType target_type) {
```
meiravgri: What would happen if `target_type == RP_INDEX` ?
Note that `RP_INDEX` doesn't call upstream->Next()

src/result_processor.c
```diff
@@ -1212,6 +1318,103 @@ static void addResultProcessor(AREQ *r, ResultProcessor *rp) {
...
+    }
+    // Update the endProc to the new head in case it was changed
+    r->qiter.endProc = dummyHead.upstream;
+    return true;
+  }
+
+  // Target type not found
+  return false;
+}
+
```
meiravgri:
```suggestion
  }

  return found;
}

```

src/result_processor.c
```diff
@@ -1303,106 +1506,73 @@ void PipelineAddCrash(struct AREQ *r) {
...
-   const RLookupKey *scoreKey;
-   SearchResult *pooledResult;
-   arrayof(SearchResult *) pool;
-   bool timedOut;
- } RPMaxScoreNormalizer;
+/*******************************************************************************************************************
+ *  Pause Processor - DEBUG ONLY
+ * TBD
+ *******************************************************************************************************************/
+extern DebugCTX globalDebugCtx;
```
meiravgri, line 1819: reduntant empty line

src/aggregate/aggregate_debug.c
```diff
@@ -54,6 +54,9 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
   ArgsCursor timeoutArgs = {0};
   int crash = 0;
   int internal_only = 0;
+  ArgsCursor pauseArgs = {0};
```
meiravgri: please add new options documentation to `aggregate_debug.h`

src/aggregate/aggregate_debug.c
```diff
@@ -115,12 +142,40 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
...
+      return REDISMODULE_ERR;
+    }
+    PipelineAddPauseAfterCount(&debug_req->r, results_count);
+  }
+
+  // Handle pause before/after RP after N (contains the same logic)
+  if (AC_IsInitialized(&pauseAfterArgs) || AC_IsInitialized(&pauseBeforeArgs)) {
+    bool before = AC_IsInitialized(&pauseBeforeArgs);
+    ArgsCursor *pauseArgs = before ? &pauseBeforeArgs : &pauseAfterArgs;
+    const char * invalidStr = before ? "PAUSE_BEFORE_RP_N" : "PAUSE_AFTER_RP_N";
```
meiravgri:
```suggestion
    const char *invalidStr = before ? "PAUSE_BEFORE_RP_N" : "PAUSE_AFTER_RP_N";
```

src/aggregate/aggregate_debug.c
```diff
@@ -115,12 +142,40 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
...
+  // Handle pause after N
+  if (AC_IsInitialized(&pauseArgs)) {
+    unsigned long long results_count = -1;
+    if (AC_GetUnsignedLongLong(&pauseArgs, &results_count, AC_F_GE0) != AC_OK) {
+      QueryError_SetError(status, QUERY_EPARSEARGS, "Invalid PAUSE_AFTER_N count");
+      return REDISMODULE_ERR;
+    }
+    PipelineAddPauseAfterCount(&debug_req->r, results_count);
+  }
+
```
meiravgri: is this only a syntactic sugar for `PAUSE_BEFORE_RP_N N RP_INDEX`?
If so, IMO this is redundant.
If you diagree is suggest to clarify (by documentation or/and renaming the query arg) where exactly we expect to be paused.

src/aggregate/aggregate_debug.c
```diff
@@ -115,12 +142,40 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
...
+  // Handle pause before/after RP after N (contains the same logic)
+  if (AC_IsInitialized(&pauseAfterArgs) || AC_IsInitialized(&pauseBeforeArgs)) {
```
meiravgri, line 144: Should be documntated that the args order matters (first count then RP)
btw to me intuitively according to the name, it should be  RP then count 
but your call :)

src/result_processor.h
```diff
@@ -299,13 +317,15 @@ void PipelineAddTimeoutAfterCount(struct AREQ *r, size_t results_count);
...
 
- /*******************************************************************************************************************
-  *  Normalizer Result Processor
-  *
-  * Normalizes search result scores to [0, 1] range by dividing each score by the maximum score.
-  * First accumulates all results from the upstream, then normalizes and yields them.
-  *******************************************************************************************************************/
- ResultProcessor *RPMaxScoreNormalizer_New(const RLookupKey *rlk);
+/*******************************************************************************************************************
+ *  Pause Processor - DEBUG ONLY
```
meiravgri, line 400: Is it supportted in cluster mode?
If not - 1. why? 2. is it being verified and tested? 3. i suggest to have INTERNAL_ONLY as a mandatory input, so in case in the future it will be enabled on cluster the behaviour won't chnage/ we won't need to expose another flag

src/aggregate/aggregate_debug.c
```diff
@@ -115,12 +142,40 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
...
       PipelineAddTimeoutAfterCount(&debug_req->r, results_count);
     }
-  } else {
-    if (internal_only) {
-      QueryError_SetError(status, QUERY_EPARSEARGS,
-                          "INTERNAL_ONLY must be used with TIMEOUT_AFTER_N");
+  }
+
+  // Handle pause after N
+  if (AC_IsInitialized(&pauseArgs)) {
```
meiravgri: should also verify workers > 0, no?

src/result_processor.h
```diff
@@ -299,13 +317,15 @@ void PipelineAddTimeoutAfterCount(struct AREQ *r, size_t results_count);
...
+ *
+ * Pauses the query after N results, N >= 0.
```
meiravgri, line 402: what does it do if N == 0?

tests/pytests/common.py
```diff
@@ -847,6 +853,30 @@ def runDebugQueryCommandAndCrash(env, query_cmd):
     debug_params = ['CRASH']
     return env.expect(debug_cmd(), *query_cmd, *debug_params, 'DEBUG_PARAMS_COUNT', len(debug_params)).error()
 
+def parseDebugCommandPauseAfterN(query_cmd, pause_after_n):
```
meiravgri: in use?

src/debug_commands.c
```diff
@@ -1834,6 +1834,99 @@ DEBUG_COMMAND(YieldCounter) {
...
+ */
+DEBUG_COMMAND(setPauseRPResume) {
+  if (!debugCommandsEnabled(ctx)) {
+    return RedisModule_ReplyWithError(ctx, NODEBUG_ERR);
+  }
+  if (argc != 2) {
+    return RedisModule_WrongArity(ctx);
+  }
+
+  globalDebugCtx.query.pause = false;
```
meiravgri: should this function be called if globalDebugCtx.query.pause is already false?

src/debug_commands.h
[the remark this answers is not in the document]
meiravgri: i say keep the volatile

src/debug_commands.c
```diff
@@ -1834,6 +1834,99 @@ DEBUG_COMMAND(YieldCounter) {
...
+ */
+DEBUG_COMMAND(getIsRPPaused) {
+  if (!debugCommandsEnabled(ctx)) {
+    return RedisModule_ReplyWithError(ctx, NODEBUG_ERR);
+  }
+  if (argc != 2) {
+    return RedisModule_WrongArity(ctx);
+  }
+
+  return RedisModule_ReplyWithLongLong(ctx, globalDebugCtx.query.pause);
```
meiravgri: consider encapsulating `globalDebugCtx`  and introducng set/get APIs

src/aggregate/aggregate_debug.c
lerman25: `StringToRPType` doesn't apply to the debug RPs as I didn't want to add debug options to  `RPTypeLookup` as it is used in production. I can change to `>=` but it doesn't make much sense since it can't be returned by StringToRPType.

src/result_processor.h
lerman25: Even tough it's only used by debug, it's a complementary function to `RPTypeToString`, so I think it makes sense for them to be adjacent.

src/debug_commands.c
```diff
@@ -1834,6 +1834,99 @@ DEBUG_COMMAND(YieldCounter) {
...
+  }
+  RedisModule_ReplySetArrayLength(ctx, resultSize);
+
+  return REDISMODULE_OK;
+}
+
+/**
+ * FT.DEBUG QUERY_CONTROLLER <command> [options]
+ */
+DEBUG_COMMAND(queryController) {
```
meiravgri, line 1976: i dont get how is that diffrenet from the seperated commands? and how are they not ambigiuos?

meiravgri (commented):
Great PR and feature!
Please add proper documentation that describes the capabilities, expected input, supported environments, and limitations.
In addition, elaborate tests to include edge cases like insert as first, trying(and failing) to insert as first, unexpected inputs etc.

src/result_processor.c
lerman25: It's aligned with `addResultProcessor`, `buildDistRPChain`. Do you think it should be different?
Moving the assignment of parent to the initialization is not aligned with the current implementations of `RP.._New()`

src/aggregate/aggregate_debug.c
meiravgri: This is why is should be an error...

src/result_processor.c
lerman25: TBH, since it's a debugging tool, my though was that if someone wants to try it, let him.
I can add guard there if you think we shouldn't allow that

src/aggregate/aggregate_debug.c
meiravgri: ohh ok sorry got it

src/result_processor.c
```diff
@@ -1303,106 +1506,73 @@ void PipelineAddCrash(struct AREQ *r) {
...
-  while ((rc = RPMaxScoreNormalizerNext_innerLoop(rp, r)) == RESULT_QUEUED) {};
-  rp->parent->resultLimit = chunkLimit; // restore the limit
-  return rc;
+static void RPPauseAfterCount_Free(ResultProcessor *base) {
+  rm_free(base);
+  // Clear the debugRP pointer if it points to us
+  if (globalDebugCtx.query.debugRP == base) {
+    globalDebugCtx.query.debugRP = NULL;
+  }
 }
```
meiravgri: We have here a race condition between parallel queries, no?

src/result_processor.c
meiravgri: i think it woudlnt work...
Even if its not used a production, we should define the expected behaviour to allow others to utilize this cool tool in the future, and inform them if they abuse it.

src/result_processor.c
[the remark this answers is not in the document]
meiravgri: what if we have multiple queries in parallel?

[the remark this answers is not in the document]
lerman25: That is not supported and not supposed to be supported

src/result_processor.h
lerman25: Added cluster testing

src/result_processor.h
lerman25: Pause at the first result (before the first was processed)

src/debug_commands.c
lerman25: It's not different, it's for better code structure , less duplication and not exploding the debug commands list.
This function calls the related sub function, I've done something similar with `BG_SCAN_CONTROLLER`

src/result_processor.c
lerman25: Parallel queries are not allowed

src/result_processor.c
lerman25: No, removed

src/result_processor.c
lerman25: OK, added a guard

src/debug_commands.c
lerman25: I think we should wait for it to be a little more complicated before that

src/debug_commands.c
lerman25: Basically, no reason to call that function if it's false, but nothing will happen if called

src/debug_commands.c
meiravgri: why?

src/debug_commands.c
meiravgri: Oh now i see you didn't include the subcommands in the list.
I was misled since their structure is following the debug commands format

src/debug_commands.c
meiravgri: If this behaviour is not expected, consider returning an error

src/debug_commands.c
lerman25: it's still really basic right now, only the pause, which is atomic, is used. 
I think that trying to guess a future need might over-complicate things that are used now.
If you think that an API is needed , I will add

src/result_processor.c
meiravgri: It is not part of RP.._New() bbecause it’s set when an RP is added to the pipeline, which doesn’t apply to this RP. Ideally we’d introduce a BaseRP constructor. In the meantime, this is an acceptable compromise.

src/debug_commands.c
meiravgri: I suggest exposing an API only for the current capabilities for example
GlobalDebug_setQueryState(state)
GlobalDebug_getQueryState()
GlobalDebug_freeDebugRP()
etc...

Also, no need to declare extern globalDebugCtx, as it should be only accessed by its owner file.

tests/pytests/test_debug_commands.py
```diff
@@ -1035,3 +1038,194 @@ def test_yield_counter(env):
...
+    # Set workers to 1 to make sure the query can be paused
+    env.expect('FT.CONFIG', 'SET', 'WORKERS', 1).ok()
+
+    # Create 1 docs
+    env.expect('HSET', 'doc1', 'name', 'name1').equal(1)
+    env.expect('FT.CREATE', 'idx', 'SCHEMA', 'name', 'TEXT').ok()
+
+    # Wait for index to finish scan
+    waitForIndex(env, 'idx')
+
```
meiravgri: can we just index after the index was created?
```suggestion
   
    env.expect('FT.CREATE', 'idx', 'SCHEMA', 'name', 'TEXT').ok()
    env.expect('HSET', 'doc1', 'name', 'name1').equal(1)
```

src/debug_commands.c
```diff
@@ -1869,6 +1869,99 @@ DEBUG_COMMAND(IndexerSleepBeforeYieldMicros) {
...
+DEBUG_COMMAND(printRPStream) {
+  if (!debugCommandsEnabled(ctx)) {
+    return RedisModule_ReplyWithError(ctx, NODEBUG_ERR);
+  }
+  if (argc != 2) {
+    return RedisModule_WrongArity(ctx);
+  }
+
+  if (!globalDebugCtx.query.debugRP) {
+    return RedisModule_ReplyWithNull(ctx);
```
meiravgri: 1. consider returning an error or any kind of indication that this shouldn't be called when no query is currently running
(preferably replacing `!globalDebugCtx.query.debugRP` with the new encapsualted API of `isQueryPaused` :)))
2. add coverage

tests/pytests/test_debug_commands.py
```diff
@@ -1035,3 +1038,194 @@ def test_yield_counter(env):
...
+            env.assertEqual(rp_stream, ['Threadsafe-Loader','Sorter','DEBUG_RP','Scorer','Index'])
+
+        # Resume the query
+        setPauseRPResume(env)
+        t_query.join()
+
+@skip(cluster=False)
+def test_cluster_query_controller_pause_and_resume(env):
+    # Set workers to 1 on all shards to make sure queries can be paused
+    verify_command_OK_on_all_shards(env, '_FT.CONFIG SET WORKERS 1')
```
meiravgri:
```suggestion
def test_cluster_query_controller_pause_and_resume():
    # Set workers to 1 on all shards to make sure queries can be paused
    env = Env(moduleArgs='WORKERS 1')
```

src/result_processor.c
```diff
@@ -1212,6 +1318,81 @@ static void addResultProcessor(AREQ *r, ResultProcessor *rp) {
...
+      }
+      rp->upstream = cur;
+      // If we inserted before the current endProc,
+      if (cur == r->qiter.endProc) {
+        r->qiter.endProc = rp;
+      }
+      return true;
+    }
+
+    downstream = cur;
```
meiravgri, line 2063: i think its clearer to emphasize both "ifs" handle the same case. wdyt?
```suggestion
    if (cur->type == target_type) {
      rp->parent = &r->qiter;
      rp->upstream = cur;
      // Checking edge case: we are the first RP in the stream
      if (cur == r->qiter.endProc) {
        r->qiter.endProc = rp;
      } else {
        downstream->upstream = rp;
      }
      
      return true;
    }

    downstream = cur;
```

src/aggregate/aggregate_debug.c
```diff
@@ -115,10 +136,41 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
...
+      QueryError_SetWithoutUserDataFmt(status, QUERY_EPARSEARGS, "Invalid %s RP type", invalidStr);
+      return REDISMODULE_ERR;
+    }
+
+    if (AC_GetUnsignedLongLong(pauseArgs, &results_count, AC_F_GE0) != AC_OK) {
+      QueryError_SetWithoutUserDataFmt(status, QUERY_EPARSEARGS, "Invalid %s count", invalidStr);
+      return REDISMODULE_ERR;
+    }
+
+    if (!PipelineAddPauseRPcount(&debug_req->r, results_count, before, rp_type)) {
```
meiravgri: :rocket:
Add test please also for cluster (trying to add after `Network`):)

tests/pytests/test_debug_commands.py
```diff
@@ -1035,3 +1038,194 @@ def test_yield_counter(env):
...
+    for i in range(n_docs):
+        res = conn.execute_command('HSET', f'doc{i}', 't', f'text{i}')
+        env.assertEqual(res, 1)
+
+    # Create index
+    res = conn.execute_command('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT')
+    env.assertEqual(res, 'OK')
+
+    # Wait until index scan completes to prevent contention on the spec lock
+    allShards_waitForIndexFinishScan(env, 'idx')
```
meiravgri: Lets create the index before the hset to skip this wait

tests/pytests/test_debug_commands.py
```diff
@@ -1035,3 +1038,194 @@ def test_yield_counter(env):
...
+
+    # Helper to call a function and push its return value into a list
+    def _call_and_store(fn, args, out_list):
+        out_list.append(fn(*args))
+
+    queries_completed = 0
+
+    for query_type in ['FT.SEARCH']:
```
meiravgri: Aggregate not supported?

src/aggregate/aggregate_debug.h
```diff
@@ -41,6 +41,12 @@
  *         - Only applicable in FT.AGGREGATE cluster mode.
  *         - If specified, the timeout applies solely to internal shard queries,
  *           without affecting the coordinator pipeline.
+ *       - **`PAUSE_AFTER_RP_N <RP_TYPE> <N>`**:
+ *         - Inserts a pause RP **after** the first occurrence of `<RP_TYPE>`; pauses after `<N>` results
+ *           flow past that RP. Fails if `<RP_TYPE>` is invalid or not present.
```
meiravgri: or its the last RP

src/aggregate/aggregate_debug.h
```diff
@@ -51,11 +57,13 @@
  *   ```
  *   _FT.DEBUG FT.SEARCH idx "*" TIMEOUT_AFTER_N 100 DEBUG_PARAMS_COUNT 2
  *   ```
+ *     ```
```
meiravgri:
```suggestion
```

src/aggregate/aggregate_debug.h
```diff
@@ -151,6 +159,17 @@
...
  *      - Total results = 12, no timeout warning.
  *
+ * #### Pause Simulation:
+ * Allows pausing query execution
+ *
+ * - **`PAUSE_AFTER_RP_N <RP_TYPE> <N>`**, **`PAUSE_BEFORE_RP_N <RP_TYPE> <N>`**:
+ *   - Inserts a pause RP after/before the first occurrence of `<RP_TYPE>`.
+ *   - Fails fast on invalid RP type or if the type is not found in the stream.
+ *
+ * **Notes (Pause):**
```
meiravgri, line 182: Please explain how to use in cluster (build pipeline to pause via coordinator debug, but resume via shard commands)
Also, since we pause the shards' pipeline, i suggest to force passing INTERNAL_ONLY (error if not passed when in cluster mode).
It allows to elaborate this API in the future to work also on the coordinator pipeline, without breaking it (i.e avoiding changing the normal behaviour when not passing INTERNAL_ONLY)

tests/pytests/test_debug_commands.py
```diff
@@ -1035,3 +1038,194 @@ def test_yield_counter(env):
...
+        # Start the query and the pause-check in parallel
+        t_query.start()
+
+        # Wait for any shard to be paused
+        while False in allShards_getIsRPPaused(env):
+            time.sleep(0.1)
+
+        # If we are here, at least one query is paused
+        # Verify that we have active queries across the cluster
+        active_queries = env.cmd('INFO', 'MODULES')['search_total_active_queries']
```
meiravgri: should be also checked across all shards

tests/pytests/test_debug_commands.py
```diff
@@ -1035,3 +1038,194 @@ def test_yield_counter(env):
...
+        env.assertEqual(active_queries, 1)
+
+        # Resume all shards
+        allShards_setPauseRPResume(env)
+
+        t_query.join()
+
+        queries_completed += 1
```
meiravgri: in use?

tests/pytests/test_debug_commands.py
```diff
@@ -1035,3 +1038,194 @@ def test_yield_counter(env):
...
+
+        # Verify the query returned results (should be at least 1 result per shard)
+        env.assertTrue(query_result[0][0] >= 1, message=f"Expected at least 1 result, got {query_result[0][0]}")
```
meiravgri: can't we predict the exact amount of results?

src/aggregate/aggregate_debug.c
```diff
@@ -115,10 +136,41 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
...
+  // Args order: RP_TYPE, N
+  if (AC_IsInitialized(&pauseAfterArgs) || AC_IsInitialized(&pauseBeforeArgs)) {
+
+    bool before = AC_IsInitialized(&pauseBeforeArgs);
+    ArgsCursor *pauseArgs = before ? &pauseBeforeArgs : &pauseAfterArgs;
+    const char *invalidStr = before ? "PAUSE_BEFORE_RP_N" : "PAUSE_AFTER_RP_N";
+    const char *rp_type_str = NULL;
+
+    if (!(debug_req->r.reqflags & QEXEC_F_RUN_IN_BACKGROUND)) {
+      QueryError_SetWithoutUserDataFmt(status, QUERY_EPARSEARGS, "Query %s is only supported in with WORKERS", invalidStr);
```
meiravgri:
```suggestion
      QueryError_SetWithoutUserDataFmt(status, QUERY_EPARSEARGS, "Query %s is only supported with WORKERS", invalidStr);
```

src/aggregate/aggregate_debug.c
```diff
@@ -115,10 +136,41 @@ int parseAndCompileDebug(AREQ_Debug *debug_req, QueryError *status) {
...
+
+    if (AC_GetString(pauseArgs, &rp_type_str, NULL, 0) != AC_OK) {
+      QueryError_SetWithoutUserDataFmt(status, QUERY_EPARSEARGS, "Invalid %s RP type", invalidStr);
+      return REDISMODULE_ERR;
+    }
+    unsigned long long results_count = -1;
+    ResultProcessorType rp_type = StringToRPType(rp_type_str);
+    // Verify the RP type is valid, not a debug RP type
+    if (rp_type == RP_MAX) {
+      QueryError_SetWithoutUserDataFmt(status, QUERY_EPARSEARGS, "Invalid %s RP type", invalidStr);
```
meiravgri: can we add to the message also the rp_type_str?

src/result_processor.c
```diff
@@ -1303,106 +1484,76 @@ void PipelineAddCrash(struct AREQ *r) {
...
-    rp->Next = RPMaxScoreNormalizer_Yield;
-    return rp->Next(rp, r);
-  } else if (rc != RS_RESULT_OK) {
-    return rc;
+}
+
+static void RPPauseAfterCount_Pause(RPPauseAfterCount *self) {
+
+  globalDebugCtx.query.pause = true;
+  self->paused = true;
```
meiravgri: wherre are we using `self->paused`?

src/aggregate/aggregate_debug.h
```diff
@@ -41,6 +41,12 @@
  *         - Only applicable in FT.AGGREGATE cluster mode.
  *         - If specified, the timeout applies solely to internal shard queries,
  *           without affecting the coordinator pipeline.
+ *       - **`PAUSE_AFTER_RP_N <RP_TYPE> <N>`**:
+ *         - Inserts a pause RP **after** the first occurrence of `<RP_TYPE>`; pauses after `<N>` results
```
meiravgri, line 45: 1. consider mentioning that after resume, and will continue the query. Wan't clear to me at first :)
2. how are we handling timeout? maybe we should also increase timeout by the time the query was paused?

src/result_processor.c
meiravgri: please add validations

meiravgri (commented):
Some more comments :)
1. make sure all edge scenarios are covered: all possible errors, trying to call getPasue/resume when no query is running, trying to pause a query whike another query is already paused.. etc :)
2. what about FT.AGGREGATE in cluster? if not supported 
 - why?
 - add proper validations and test coverage

src/result_processor.c
lerman25: I don't think we should "treat" this RP different in initialization from other RPs, even the debug ones (Timeout, Crash) don't initialize `rp->parent` and set the pointer at `addResultProcessor`.

tests/pytests/test_debug_commands.py
lerman25: Changed to num_docs

src/debug_commands.c
```diff
@@ -32,6 +32,35 @@
 
 DebugCTX globalDebugCtx = {0};
 
+// QueryDebugCtx API implementations
```
meiravgri, line 36: 🚀 so clean

meiravgri (dismissed):

meiravgri (dismissed):

src/debug_commands.c
```diff
@@ -33,6 +33,27 @@
...
+bool QueryDebugCtx_IsPaused(void) {
+  return globalDebugCtx.query.pause;
+}
+
+void QueryDebugCtx_SetPause(bool pause) {
+  globalDebugCtx.query.pause = pause;
+}
```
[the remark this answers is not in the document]
lerman25: fixed

src/result_processor.h
```diff
@@ -287,9 +289,104 @@ uint64_t RPProfile_GetCount(ResultProcessor *rp);
...
+ * First accumulates all results from the upstream, then normalizes and yields them.
+ *******************************************************************************************************************/
+ResultProcessor *RPMaxScoreNormalizer_New(const RLookupKey *rlk);
+/*******************************************************************************************************************
+ *  Normalizer Result Processor
+ *
+ * Normalizes search result scores to [0, 1] range by dividing each score by the maximum score.
+ * First accumulates all results from the upstream, then normalizes and yields them.
+ *******************************************************************************************************************/
+ResultProcessor *RPMaxScoreNormalizer_New(const RLookupKey *rlk);
```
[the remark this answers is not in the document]
lerman25: fixed

meiravgri (dismissed):

meiravgri (approved):

PR #6753, opened 2025-09-01
[MOD-11099] Have a wrapper to track the field masks inside the Rust inverted index impl

chesedo:
This wrapper matches the equivalent logic for the `fieldMask` union field in the C implementation. This also changes the C code to update this field inside the write function rather than some parent caller (after it called the write function anyway)

#### Mark if applicable


src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -322,6 +322,50 @@ impl<E: Encoder> InvertedIndex<E> {
     }
 }
 
+pub struct FieldMaskTrackingIndex<E> {
```
gdesmott, line 338: Need `///`

chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6753/commits/6f0d8f5c63807841d6cf7b2b850bf3698fca8e2e

gdesmott (approved):

PR #6758, opened 2025-09-02
[MOD-11220] move core logic of FGC_applyInvertedIndex to II

GlenDC:
small stepping stone in [the MOD-10702 quest](https://redislabs.atlassian.net/browse/MOD-10702)

Closes: https://redislabs.atlassian.net/browse/MOD-11220

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -25,11 +25,12 @@ COV=${COV:-0}    # Coverage mode (for building and testing)
 SVS_PRE_COMPILED_LIB=${SVS_PRE_COMPILED_LIB:-0} # Use SVS pre-compiled library
 
 # Test configuration (0=disabled, 1=enabled)
-BUILD_TESTS=0    # Build test binaries
-RUN_UNIT_TESTS=0 # Run C/C++ unit tests
-RUN_RUST_TESTS=0 # Run Rust tests
-RUN_PYTEST=0     # Run Python tests
-RUN_ALL_TESTS=0  # Run all test types
+BUILD_TESTS=0          # Build test binaries
+RUN_UNIT_TESTS=0       # Run C/C++ unit tests
+RUN_RUST_TESTS=0       # Run Rust tests
+RUN_RUST_VALGRIND=0    # Run Valgrind Rust tests
+RUN_PYTEST=0           # Run Python tests
+RUN_ALL_TESTS=0        # Run all test types
 RUN_MICRO_BENCHMARKS=0 # Run micro-benchmarks
 
 # Rust configuration
@@ -69,6 +70,9 @@ parse_arguments() {
         ;;
       RUN_RUST_TESTS|run_rust_tests)
         RUN_RUST_TESTS=1
+        ;;
+      RUN_RUST_VALGRIND|run_rust_valgrind)
+        RUN_RUST_VALGRIND=1
         ;;
       RUN_MICRO_BENCHMARKS|run_micro_benchmarks|RUN_MICROBENCHMARKS|run_microbenchmarks)
         RUN_MICRO_BENCHMARKS=1
@@ -129,7 +133,7 @@ parse_arguments() {
 #-----------------------------------------------------------------------------
 setup_test_configuration() {
   # If any tests will be run, ensure BUILD_TESTS is enabled
-  if [[ "$RUN_ALL_TESTS" == "1" || "$RUN_UNIT_TESTS" == "1" || "$RUN_RUST_TESTS" == "1" || "$RUN_PYTEST" == "1" || "$RUN_MICRO_BENCHMARKS" == "1" ]]; then
+  if [[ "$RUN_ALL_TESTS" == "1" || "$RUN_UNIT_TESTS" == "1" || "$RUN_RUST_TESTS" == "1" || "$RUN_RUST_VALGRIND" == "1" || "$RUN_PYTEST" == "1" || "$RUN_MICRO_BENCHMARKS" == "1" ]]; then
     if [[ "$BUILD_TESTS" != "1" ]]; then
       echo "Test execution requested, enabling test build automatically"
       BUILD_TESTS="1"
@@ -546,7 +550,7 @@ run_rust_tests() {
 
   # Set Rust test environment
   RUST_DIR="$ROOT/src/redisearch_rs"
-  
+
   # Set up RUSTFLAGS for warnings
   if [[ "$RUST_DENY_WARNS" == "1" ]]; then
     export RUSTFLAGS="${RUSTFLAGS:+${RUSTFLAGS} }-D warnings"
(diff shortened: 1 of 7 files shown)


src/fork_gc.c
```diff
@@ -682,84 +682,32 @@ static void checkLastBlock(ForkGC *gc, InvIdxBuffers *idxData, MSG_IndexInfo *in
...
-     */
-
-    if (!info->lastblkDocsRemoved) {
-      /**
-       * Last block was unmodified-- let's prefer the last block's pointer
-       * over our own (which may be stale).
-       * If the last block was repaired, this is handled above in checkLastBlock()
-       */
-      idxData->newBlocklist[idxData->newBlocklistSize - 1] = InvertedIndex_Block(idx, info->nblocksOrig - 1);
-    }
```
chesedo: These comments carry a lot of information. Let's keep more of it at the new comment on line 686

src/inverted_index/inverted_index.c
```diff
@@ -1468,3 +1370,173 @@ void IndexReader_SwapIndex(IndexReader *ir, const InvertedIndex *newIdx) {
 InvertedIndex *IndexReader_II(const IndexReader *ir) {
   return (InvertedIndex *)ir->idx;
 }
+
+// ----- GC related API
+
+size_t IndexBlock_Repair(IndexBlock *blk, DocTable *dt, IndexFlags flags, IndexRepairParams *params) {
```
chesedo, line 1380: Let's restore the original (or improved) function comment here

src/inverted_index/inverted_index.c
```diff
@@ -1468,3 +1370,173 @@ void IndexReader_SwapIndex(IndexReader *ir, const InvertedIndex *newIdx) {
...
+        memcpy(d->new_blocklist + d->new_blocklist_size,
+                InvertedIndex_BlockRef(idx, nblocks_orig),
+                newAddedLen * sizeof(*d->new_blocklist));
+      }
+    }
+    InvertedIndex_SetBlocks(idx, d->new_blocklist, totalLen);
+    // ownership of new_blocklist has moved into idx
+    d->new_blocklist = NULL;
+  } else if (d->deleted_len) {
+    // The child deleted all blocks it had seen
```
chesedo: Can we restore a bit of the original comments here?

chesedo (commented):
A few nitpicks about comments

src/inverted_index/inverted_index.c
GlenDC: The comment is still there but I moved it to header file, which seems like a more logical place for these function comments? What is the right thing to do there, hopefully not both header and impl file?

chesedo: Nvm then. I missed this :+1:

chesedo (dismissed):

GlenDC:
@chesedo did I make a mistake here? You have a clue why CI is failing (retried it already)

- [Pull Request Flow / pr-validation (pull_request)](https://github.com/RediSearch/RediSearch/actions/runs/17426413185/job/49499295269?pr=6758)Failing after 4s
- [Pull Request Flow / sanitize / Test ubuntu-latest, Redis 8.2.1 (pull_request)](https://github.com/RediSearch/RediSearch/actions/runs/17426413185/job/49496553433?pr=6758)Failing after 32m

Or is that because of code-cov?

GlenDC:
> @chesedo did I make a mistake here? You have a clue why CI is failing (retried it already)
> 
> * [Pull Request Flow / pr-validation (pull_request)](https://github.com/RediSearch/RediSearch/actions/runs/17426413185/job/49499295269?pr=6758)Failing after 4s
> * [Pull Request Flow / sanitize / Test ubuntu-latest, Redis 8.2.1 (pull_request)](https://github.com/RediSearch/RediSearch/actions/runs/17426413185/job/49496553433?pr=6758)Failing after 32m
> 
> Or is that because of code-cov?

nvm @chesedo turns out to be memory leak. Or at least that's how it's reported. Will investigate.

chesedo (dismissed):

src/fork_gc.c
```diff
@@ -682,84 +682,37 @@ static void checkLastBlock(ForkGC *gc, InvIdxBuffers *idxData, MSG_IndexInfo *in
...
+  }
+
+  InvertedIndexGcDelta delta = {
+    .new_blocklist = idxData->newBlocklist,
+    .new_blocklist_size = idxData->newBlocklistSize,
+    .deleted = (void *)idxData->delBlocks,
+    .deleted_len = idxData->numDelBlocks,
+    .repaired = (void *)idxData->changedBlocks,
+    .repaired_len = info->nblocksRepaired,
+    .last_block_ignored = idxData->lastBlockIgnored,
```
BenGoldberger: I dont see where we use this boolean

src/fork_gc.c
```diff
@@ -682,84 +682,37 @@ static void checkLastBlock(ForkGC *gc, InvIdxBuffers *idxData, MSG_IndexInfo *in
...
-    MSG_RepairedBlock *blockModified = idxData->changedBlocks + i;
-    InvertedIndex_SetBlock(idx, blockModified->newix, blockModified->blk);
-  }
-  // Consume changed blocks array
-  rm_free(idxData->changedBlocks);
+  // consume block arrays from InvIdxBuffers
+  // (in future, when InvIdxBuffers stops existing this will be cleaner)
   idxData->changedBlocks = NULL;
+  idxData->delBlocks = NULL;
+  idxData->newBlocklist = NULL;
```
BenGoldberger, line 712: all those pointers are set to null inside the function, I dont think we need to do it twice

src/fork_gc.c
GlenDC: Correct, we need this info in a later PR but here it is not needed for now yet and perhaps it might also be replaced by something else when we get to it. I'll remove it for now.

src/fork_gc.c
GlenDC: In a future PR there won't be two structs and in fact fork_gc won't even be aware of these inner pointers. However for now we are copying pointers around and so the pointers being set to NULL are two different pointers to the same data. I can remove it as anyway the data is deleted so not like something should anyway be touching it still, but it would mean that `idxData` contains dangling pointers.

GlenDC:
@BenGoldberger review addressed.

GlenDC:
master merged into this branch (was merge conflict due to other PR that was just now merged)

BenGoldberger (approved):

BenGoldberger (approved):

build.sh
```diff
@@ -587,6 +591,63 @@ run_rust_tests() {
...
+
+  # Set up RUSTFLAGS for warnings
+  if [[ "$RUST_DENY_WARNS" == "1" ]]; then
+    export RUSTFLAGS="${RUSTFLAGS:+${RUSTFLAGS} }-D warnings"
+  fi
+
+  # Pin a specific working version of nightly to prevent breaking the CI because
+  # regressions in a nightly build.
+  # Make sure to synchronize updates across all modules: Redis and RedisJSON.
+  NIGHTLY_VERSION="nightly-2025-07-30"
```
gdesmott, line 616: This should not be duplicated. The whole point of this var was to have the nightly version defined in a single place.

build.sh
```diff
@@ -587,6 +591,63 @@ run_rust_tests() {
...
+
+  # Add Rust test extensions
+  if [[ -n "$SAN" ]]; then
+    RUST_EXTENSIONS="+$NIGHTLY_VERSION miri"
```
gdesmott, line 620: I don't think there is much point running both miri and valgrind at the same time.

build.sh
GlenDC: I'll update it in next FGC PR

build.sh
GlenDC: ok I'll update

PR #6759, opened 2025-09-02
[MOD-11194] Port `NumericFilter` to Rust

chesedo:
The inverted index reader might need to filter records based on the numeric filter. For the Rust implementation to do the same, the `NumericFilter` struct needs to be accessible by Rust. There this PR ports the struct over to Rust for easy direct access.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -97,16 +97,8 @@ impl NumericFilter {
 
     /// Check if the given value is in the range specified by this filter
     pub fn value_in_range(&self, value: f64) -> bool {
-        let min_ok = if self.min_inclusive {
-            value >= self.min
-        } else {
-            value > self.min
-        };
-        let max_ok = if self.max_inclusive {
-            value <= self.max
-        } else {
-            value < self.max
-        };
+        let min_ok = value > self.min || (self.min_inclusive && value == self.min);
+        let max_ok = value < self.max || (self.max_inclusive && value == self.max);
 
         min_ok && max_ok
     }
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -62,6 +62,7 @@ impl IdDelta for u32 {
 /// cbindgen:rename-all=CamelCase
 #[repr(C)]
 pub struct NumericFilter {
+    /// The field specification which this filter is acting on
     field_spec: *const FieldSpec,
 
     /// Beginning of the range
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -89,6 +89,13 @@ pub struct NumericFilter {
     offset: usize,
 }
 
+impl NumericFilter {
+    /// Check if this is a numeric filter (and not a geo filter)
+    pub fn is_numeric_filter(&self) -> bool {
+        self.geo_filter.is_null()
+    }
+}
+
 /// Encoder to write a record into an index
 pub trait Encoder {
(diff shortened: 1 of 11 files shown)


src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -54,6 +58,60 @@ impl IdDelta for u32 {
...
+
+impl NumericFilter {
+    /// Check if this is a numeric filter (and not a geo filter)
+    pub fn is_numeric_filter(&self) -> bool {
+        self.geo_filter.is_null()
+    }
+
+    /// Check if the given value is in the range specified by this filter
+    pub fn value_in_range(&self, value: f64) -> bool {
+        let min_ok = if self.min_inclusive {
```
GlenDC: any reason not to do these as `min_ok = value > self.min || (self.min_inclusive && value == self.min)`? Same for max. As I suppose it's more likely that you are greater or lesser, compared to the single possibility that you are equal

GlenDC (approved):
Looks like a clean 1:1 port, 1 nitpick comment but looks good

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -54,6 +58,60 @@ impl IdDelta for u32 {
     }
 }
 
+/// Filter details to apply to numeric values
+/// cbindgen:rename-all=CamelCase
+#[repr(C)]
+pub struct NumericFilter {
+    field_spec: *const FieldSpec,
```
gdesmott, line 66: undocumented.

Also why all these extra `\n` ? :D

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: No reason. I did not think of this approach but think it is worth making.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Haha, I have no idea what this field does. And could not steal the documentation from the original declaration :sweat_smile: . I'll figure it out and document appropriately.

For the extra lines, haven't we always had a new line between fields and the comment of the next field?

gdesmott: > For the extra lines, haven't we always had a new line between fields and the comment of the next field?

Do we? I find it a waste of space but can live with it. :)

chesedo: I think the little comments make it look weird in this case

https://github.com/RediSearch/RediSearch/blob/e10a8e7eb83d92c6e5fe6645d0bb200280c52250/src/redisearch_rs/inverted_index/src/index_result.rs#L627-L652

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6759/commits/d9341605b0da695453f543a41b57da5ff43a123f

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: I added the missing comment in https://github.com/RediSearch/RediSearch/pull/6759/commits/7c82e603f611c52f4f7b07f6c51eea1ba3ac96b7

gdesmott (approved):

alonre24 (approved):
LGTM
Coverage decrease is expected here?

chesedo:
> LGTM Coverage decrease is expected here?

Yeah, this is being covered by some C tests. But the coverage is not correctly detecting the Rust coverage for calls from C.

The next few PRs (#6766) will have some Rust tests which exercise this code

PR #6763, opened 2025-09-02
[MOD-11011] Fix deadlock while RDB loading and RM_Yield

JoanFM:
For now we aim to reproduce the issue with a flow test


src/spec.c
```diff
@@ -3341,6 +3347,12 @@ int IndexSpec_UpdateDoc(IndexSpec *spec, RedisModuleCtx *ctx, RedisModuleString
     return REDISMODULE_ERR;
   }
 
+  unsigned int baseYield = RSGlobalConfig.indexerYieldEveryOpsWhileLoading;
+  unsigned int yieldEveryOps = doc.numFields > 0
+                                 ? baseYield / doc.numFields
+                                 : baseYield;
+
+  IndexerYieldWhileLoading(ctx, yieldEveryOps, REDISMODULE_YIELD_FLAG_CLIENTS);
```
BenGoldberger: this implementation could lead to some wierd behaviour if the numFileds is defferent between docs. perhaps it would be better to send to the function the numFields and increase the counter by that argument. 

also why do we want to pass the flags as an argument?

src/indexer.c
```diff
@@ -394,14 +392,16 @@ bool g_isLoading = false;
...
+ * @param yieldEveryOps The number of operations to perform before yielding
+ * @param flags The flags to pass to RedisModule_Yield
  */
-static void IndexerYieldWhileLoading(RedisModuleCtx *ctx) {
+void IndexerYieldWhileLoading(RedisModuleCtx *ctx, unsigned int yieldEveryOps, int flags) {
   static size_t opCounter = 0;
 
   // If server is loading, Yield to Redis every RSGlobalConfig.indexerYieldEveryOps operations
-  if (g_isLoading && ++opCounter >= RSGlobalConfig.indexerYieldEveryOpsWhileLoading) {
+  if (g_isLoading && ++opCounter >= yieldEveryOps) { // RSGlobalConfig.indexerYieldEveryOpsWhileLoading) {
```
BenGoldberger: remove the comment of the old code

src/spec.c
JoanFM: what weird behavior? this is already being considered for each document.

Passing the flag is a leftover from a test I wanted to do, but it may make sense that we would eventually want to Yield with a different flag so I kept it

JoanFM: I see ur point, may be interesting indeed.

alonre24 (commented):
Fix LGTM, can add a small flow test that reproduce the issue? consider adding to ft.debug a mechanism that inject waiting in certain points of the code (similar to `ft.debug BG_SCAN_CONTROLLER`)

JoanFM:
> Fix LGTM, can add a small flow test that reproduce the issue? consider adding to ft.debug a mechanism that inject waiting in certain points of the code (similar to `ft.debug BG_SCAN_CONTROLLER`)

@alonre24 with the team we discussed to add this test as an RL Automation test, does this sound good? 

Maybe involving the addition of some other parameter it may be easier to reproduce and we can have both,

alonre24:
> > Fix LGTM, can add a small flow test that reproduce the issue? consider adding to ft.debug a mechanism that inject waiting in certain points of the code (similar to `ft.debug BG_SCAN_CONTROLLER`)
> 
> @alonre24 with the team we discussed to add this test as an RL Automation test, does this sound good?
> 
> Maybe involving the addition of some other parameter it may be easier to reproduce and we can have both,

As I see it, here with RLTest we should have a small test that mimics the exact bad behavior. In RS Automation, we should also add an E2E test that loads from a large RDB while info commands are sent (to be done by @galilev). I see these as complementary tests, not one on the expanse of the other

JoanFM:
> > Fix LGTM, can add a small flow test that reproduce the issue? consider adding to ft.debug a mechanism that inject waiting in certain points of the code (similar to `ft.debug BG_SCAN_CONTROLLER`)

Done, added a DEBUG Command to set some sleep time to ease the reproduction of the issue, and added a flow test currently passing

BenGoldberger (approved):

PR #6766, opened 2025-09-03
[MOD-11171] Rust filter wrapper around inverted index reader

chesedo:
This ports over the last filters which will be needed by the inverted index reader in Rust. They are equivalent to these lines in C

https://github.com/RediSearch/RediSearch/blob/74035a27a248489bb797bd49ac080ce6a6a51de4/src/inverted_index/inverted_index.c#L837-L849

The Rust version splits this into two separate wrappers. One for filtering the numeric values, and another for filtering the geo radius.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -711,6 +711,43 @@ impl<'index, I: Iterator<Item = RSIndexResult<'index>>> Iterator for FilterMaskR
     }
 }
 
+/// A reader that filters out records that do not match a given numeric filter. It is used to
+/// filter records in an index based on their numeric value, allowing only those that match the
+/// specified filter to be returned.
+///
+/// This should only be wrapped around readers that return numeric records.
+pub struct FilterNumericReader<I> {
+    /// The numeric filter that is used to filter the records.
+    filter: NumericFilter,
+
+    /// The inner reader that will be used to read the records from the index.
+    inner: I,
+}
+
+impl<'index, I: Iterator<Item = RSIndexResult<'index>>> FilterNumericReader<I> {
+    /// Create a new filter numeric reader with the given filter and inner iterator.
+    pub fn new(filter: NumericFilter, inner: I) -> Self {
+        Self { filter, inner }
+    }
+}
+
+impl<'index, I: Iterator<Item = RSIndexResult<'index>>> Iterator for FilterNumericReader<I> {
+    type Item = RSIndexResult<'index>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            let next = self.inner.next()?;
+            let value = next.as_numeric()?;
+
+            if !self.filter.value_in_range(value) {
+                continue;
+            }
+
+            return Some(next);
+        }
+    }
+}
+
 /// A reader that filters out records that do not match a given geo filter. It is used to
 /// filter records in an index based on their geo location, allowing only those that match the
 /// specified geo filter to be returned.
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -14,8 +14,8 @@ use std::{
 
 use debug::{BlockSummary, Summary};
 use ffi::{
-    FieldSpec, IndexFlags, IndexFlags_Index_HasMultiValue, IndexFlags_Index_StoreFieldFlags,
-    IndexFlags_Index_StoreNumeric,
+    FieldSpec, GeoFilter, IndexFlags, IndexFlags_Index_HasMultiValue,
+    IndexFlags_Index_StoreFieldFlags, IndexFlags_Index_StoreNumeric,
 };
 pub use ffi::{t_docId, t_fieldMask};
 pub use index_result::{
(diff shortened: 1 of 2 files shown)


src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -581,5 +592,80 @@ impl<'index, I: Iterator<Item = RSIndexResult<'index>>> Iterator for FilterMaskR
     }
 }
 
+/// A reader that filters out records that do not match a given numeric filter. It is used to
+/// filter records in an index based on their numeric value, allowing only those that match the
+/// specified filter to be returned.
+///
+/// This should only be wrapped around readers that return numeric records.
```
gdesmott, line 716: Could we enforce this within the type system somehow?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -581,5 +592,84 @@ impl<'index, I: Iterator<Item = RSIndexResult<'index>>> Iterator for FilterMaskR
...
+
+impl<'index, I: Iterator<Item = RSIndexResult<'index>>> Iterator for FilterNumericReader<I> {
+    type Item = RSIndexResult<'index>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            let next = self.inner.next()?;
+            let value = next.as_numeric()?;
+
+            if !self.filter.value_in_range(value) {
```
GlenDC: I would write this as 

```
if true { return Some ... }
```

instead of the current

```
if !true { continue } return Some....
```

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -581,5 +592,84 @@ impl<'index, I: Iterator<Item = RSIndexResult<'index>>> Iterator for FilterMaskR
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            let mut next = self.inner.next()?;
+            let value = next.as_numeric_mut()?;
+
+            // SAFETY: we know the filter is not a null pointer
+            let filtered = unsafe { isWithinRadius(self.filter, *value, value) };
+
+            if !filtered {
```
GlenDC: same here

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -581,5 +592,84 @@ impl<'index, I: Iterator<Item = RSIndexResult<'index>>> Iterator for FilterMaskR
...
+    for FilterGeoReader<'filter, I>
+{
+    type Item = RSIndexResult<'index>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            let mut next = self.inner.next()?;
+            let value = next.as_numeric_mut()?;
+
+            // SAFETY: we know the filter is not a null pointer
```
GlenDC: why do we know that?

GlenDC: resolved right after I commented but before I submitted the review

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -34,6 +34,17 @@ pub mod raw_doc_ids_only;
 #[doc(hidden)]
 pub mod test_utils;
 
+// Manually define some C functions, because we'll create a circular dependency if we use the FFI
+// crate to make them automatically.
+unsafe extern "C" {
+    /// Checks if a value (distance) is within the given geo filter.
+    ///
+    /// # Safety
+    /// The [`GeoFilter`] should not be null and a valid instance
```
gdesmott, line 47: Could we use `NonNull` then?

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Interesting idea :thinking: 

It's not possible to restrict it on the `RSIndexResult`. But rather than having these as an iterator wrapper, they can just be predicates instead. But maybe there is another way I'm not aware of?

gdesmott: Just brainstorming, I did not really thing about it. :D 

But yeah as long as we have this opaque `RSIndexResult` type I don't think it's possible.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6766/commits/84b00022b92d4e4ecad643d6b23ef777653a8057

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6766/commits/84b00022b92d4e4ecad643d6b23ef777653a8057

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: I just need something simple to match this

https://github.com/RediSearch/RediSearch/blob/aa1241c5da5c08be956f8bdc90c5b2a4f348d0be/src/geo_index.h#L88

Because adding `geo_index.h` to the ffi crate opens a whole new can of worms

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: For now reducing this to predicates won't be possible because the index reader will need to return the numeric filter

https://github.com/RediSearch/RediSearch/blob/aa1241c5da5c08be956f8bdc90c5b2a4f348d0be/src/inverted_index/inverted_index.h#L237-L240

gdesmott (approved):

PR #6767, opened 2025-09-03
[MOD-10716] refactor(rlookup): move C-accessible public fields into *Header structs

JonasKruckenberg:
In order to get the C header file to generate nicely, this change splits up `RLookupKey` and `RLookup` into a C-accessible header structs and a Rust-private structs.

This is not ideal, but given the limitations of cbindgen the cleanest solution since it allowed removing the `CBCow` and `CBOption` as well.

### DIFF
--- a/src/redisearch_rs/rlookup/src/lookup.rs
+++ b/src/redisearch_rs/rlookup/src/lookup.rs
@@ -12,7 +12,7 @@ use crate::bindings::{
 use enumflags2::{BitFlags, bitflags, make_bitflags};
 use pin_project::pin_project;
 use std::{
-    borrow::Borrow,
+    borrow::Cow,
     cell::UnsafeCell,
     ffi::{CStr, c_char},
     mem,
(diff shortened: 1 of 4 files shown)


src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -265,10 +140,28 @@ impl<T> Default for CBOption<T> {
 /// This is used for data generated on the fly, or for data not stored within
 /// the sorting vector.
 /// ```
+///
+/// cbindgen:no-export
 #[pin_project(!Unpin)]
 #[derive(Debug)]
 #[repr(C)]
 pub struct RLookupKey<'a> {
```
chesedo, line 148: Does this still need `repr(C` if it is not exported?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -265,10 +140,28 @@ impl<T> Default for CBOption<T> {
...
+
+    // Private Rust fields
+    /// The actual "owning" strings, we need to hold onto these
+    /// so the pointers above stay valid. Note that you
+    /// MUST NEVER MOVE THESE BEFORE THE name AND path FIELDS UNLESS
+    /// YOU WANT TO POTENTIALLY RISK UB
+    #[pin]
+    _name: Cow<'a, CStr>,
+    #[pin]
+    _path: Option<Cow<'a, CStr>>,
```
chesedo: Which `name` and `path`? Should it be `header` now?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -295,25 +188,16 @@ pub struct RLookupKey<'a> {
...
-    _path: CBOption<CBCow<'a, CStr>>,
 }
 
 /// An append-only list of [`RLookupKey`]s.
 ///
 /// This type maintains a mapping from string names to [`RLookupKey`]s.
+/// cbindgen:no-export
 #[derive(Debug)]
 #[repr(C)]
 pub struct RLookup<'a> {
```
chesedo, line 200: Again, is `repr(C)` still needed

chesedo (commented):
Minor comments

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: Actually yes, because we must guarantee that `RLookupKey` has the same layout as `RLookupKeyHeader` since we effectively transmute between the two at the FFI boundary (we tell C code that `RLookupKeyHeader` is the same as `RLookupKey` through the rename)

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: updated the comment 👍🏻

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: see abvove

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -265,10 +140,28 @@ impl<T> Default for CBOption<T> {
...
+    header: RLookupKeyHeader<'a>,
+
+    /// The actual "owning" strings, we need to hold onto these
+    /// so the pointers in the above header stay valid. Note that you
+    /// MUST NEVER MOVE THESE BEFORE THE name AND path FIELDS UNLESS
+    /// YOU WANT TO POTENTIALLY RISK UB
+    #[pin]
+    _name: Cow<'a, CStr>,
+    #[pin]
+    _path: Option<Cow<'a, CStr>>,
```
chesedo: The `name` and `path` wording still does not make sense here

JonasKruckenberg: good point yeah

chesedo (dismissed):

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -265,10 +140,27 @@ impl<T> Default for CBOption<T> {
...
+///
+/// cbindgen:no-export
 #[pin_project(!Unpin)]
 #[derive(Debug)]
 #[repr(C)]
 pub struct RLookupKey<'a> {
+    // THIS MUST BE THE FIRST FIELD
+    header: RLookupKeyHeader<'a>,
+
+    // The actual "owning" strings, we need to hold onto these
```
kei-nan, line 154: Can we have a rust test to verify this?
Might be better to enforce via a test over a comment.

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -295,25 +187,17 @@ pub struct RLookupKey<'a> {
...
 /// An append-only list of [`RLookupKey`]s.
 ///
 /// This type maintains a mapping from string names to [`RLookupKey`]s.
+/// cbindgen:no-export
 #[derive(Debug)]
 #[repr(C)]
 pub struct RLookup<'a> {
-    keys: KeyList<'a>,
+    // THIS MUST BE THE FIRST FIELD
+    header: RLookupHeader<'a>,
```
kei-nan, line 204: can we have a test to verify this?
Also would be nice to know why.

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: [I added const-assertion tests](https://github.com/RediSearch/RediSearch/pull/6767/commits/b518fd89f40024ff37b499e0edff79ccfe19bf72) to ensure this as well as updated the comment to be more helpful 👍🏻

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: See above

kei-nan (dismissed):

PR #6776, opened 2025-09-04
MOD-11144: Fix FT.HYBRID GROUPBY bug

Itzikvaknin:
The `aggregate_request:parseGroupby` code was assuming the argc cursor to be of type 3 (sds), and therefore that its objects could be directly cast to `const char **`.

This PR updates the logic to build the array of properties by unpacking strings from the ArgCursor using `AC_GetString`.

With this change, the helper array is owned by the group step (but not its contents) and will be freed by `groupStepFree`.

### DIFF
--- a/tests/pytests/test_hybrid_groupby.py
+++ b/tests/pytests/test_hybrid_groupby.py
@@ -16,6 +16,9 @@ doc:1, doc:2 and doc:3 are at distance 1 from the query vector [0,0]
 doc:4, doc:5 and doc:6 are at distance 2 from the query vector [0,0]
 doc:7, doc:8 and doc:9 are at distance 3 from the query vector [0,0]
 
+IMPORTANT: to save calculations, redis stores only the squared distance in the vector index,
+therefore we square the radius and numpy l2 norm to get the squared distance
+
 This allows controlling result set sizes with RADIUS:
 - RADIUS 1**2 → 3 results (distance 1)
 - RADIUS 2**2 → 6 results (distances 1-2)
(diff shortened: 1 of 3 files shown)


src/aggregate/aggregate_plan.h
```diff
@@ -115,7 +116,7 @@ typedef struct {
   PLN_BaseStep base;
   RLookup lookup;
 
-  const char **properties;
+  StrongRef properties_ref;  // StrongRef to properties array
   size_t nproperties;
```
kei-nan: if you are holding an array then do you need nproperties?
Can you use array_len instead?

tests/pytests/test_hybrid_groupby.py
```diff
@@ -0,0 +1,144 @@
...
+
+VECTOR SPACE LAYOUT:
+===================
+Query vector at origin [0, 0]
+
+doc:1, doc:2 and doc:3 are at distance 1 from the query vector [0,0]
+doc:4, doc:5 and doc:6 are at distance 2 from the query vector [0,0]
+doc:7, doc:8 and doc:9 are at distance 3 from the query vector [0,0]
+
+This allows controlling result set sizes with RADIUS:
```
nafraf, line 22: Since it's not documented now, can we add a note that we take L2distance^2 and not L2distance?

src/aggregate/aggregate_plan.h
Itzikvaknin: Right, removed it

kei-nan (approved):

nafraf (commented):
What about `test_knn_groupby()` in `test_hybrid.py`? We need to uncommented it or remove it if it's now unneeded.

kei-nan (approved):

nafraf (approved):

PR #6779, opened 2025-09-04
[MOD-11191] fix: multiply optional weight instead of apply

JoanFM:

Changes the way Optional node weight is considered, multiply the value of the child instead of applying its weight

### DIFF
--- a/src/iterators/optional_iterator.c
+++ b/src/iterators/optional_iterator.c
@@ -71,7 +71,7 @@ static IteratorStatus OI_SkipTo_Optimized(QueryIterator *base, t_docId docId) {
   if (docId == oi->child->lastDocId) {
     // Has a real hit on the child iterator
     oi->base.current = oi->child->current;
-    oi->base.current->weight = oi->weight;
+    oi->base.current->weight *= oi->weight;
   } else {
     oi->virt->docId = oi->wcii->lastDocId;
     oi->base.current = oi->virt;
@@ -111,7 +111,7 @@ static IteratorStatus OI_Read_Optimized(QueryIterator *base) {
     oi->base.current = oi->virt;
   } else {
     oi->base.current = oi->child->current;
-    oi->base.current->weight = oi->weight;
+    oi->base.current->weight *= oi->weight;
   }
 
   base->lastDocId = oi->base.current->docId = oi->wcii->lastDocId;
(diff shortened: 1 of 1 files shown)


src/iterators/optional_iterator.c
```diff
@@ -290,7 +290,7 @@ QueryIterator *NewOptionalIterator(QueryIterator *it, QueryEvalCtx *q, double we
   oi->virt = NewVirtualResult(0, RS_FIELDMASK_ALL);
   oi->virt->freq = 1;
   oi->maxDocId = q->docTable->maxDocId;
-  oi->weight = weight;
+  oi->weight = weight * it->current->weight;
```
kei-nan, line 296: consider adding a bit of documentation explaining the wanted behaviour and how this code achieves that.

alonre24 (commented):
LGTM, consider adding a flow test as well (you can use the scenario from the ticket)

JoanFM:
I am waiting from @adrianoamaral input to make sure this is the desired behavior

JoanFM:
> LGTM, consider adding a flow test as well (you can use the scenario from the ticket)

Flow test added

src/iterators/optional_iterator.c
JoanFM: done

kei-nan (dismissed):

src/iterators/optional_iterator.c
```diff
@@ -290,7 +290,8 @@ QueryIterator *NewOptionalIterator(QueryIterator *it, QueryEvalCtx *q, double we
   oi->virt = NewVirtualResult(0, RS_FIELDMASK_ALL);
   oi->virt->freq = 1;
   oi->maxDocId = q->docTable->maxDocId;
-  oi->weight = weight;
+  // The weight of the result should be the multiplication of the child's weight and the node's weight.
+  oi->weight = weight * it->current->weight;
```
GuyAv46, line 296: Consider asserting that `it->current` is set. `it` should be valid and not an empty iterator (we can assert that as well), so it should be safe

GuyAv46, line 296: Also, we miss the multiplication in the `OptionalIteratorReducer`, in case the child is a wildcard

GuyAv46 (approved):

PR #6780, opened 2025-09-04
[MOD-11242] FFI crate for the Rust inverted index

chesedo:
This adds an ffi crate and header file generation for the incoming Rust inverted index implementation. The generated header file is located in `headers/future` so that the C code does not include in yet. A future PR will move the header file to `headers` to activate the Rust implementation.

More functions will also be added to this FFI in future PRs.

#### Mark if applicable


src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+) -> *mut InvertedIndex {
+    let ii = match (
+        flags & INDEX_STORAGE_MASK,
+        raw_doc_id_encoding,
+        compress_floats,
+    ) {
+        (FULL_MASK, _, _) => {
+            let ii = FieldMaskTrackingIndex::new(flags, Full);
+
+            *mem_size = ii.memory_usage();
```
gdesmott: All those can be moved out of the `match`.

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+/// should be freed using [`InvertedIndex_Free`] when no longer needed.
+///
+/// # Safety
+///
+/// The following invariant must be upheld when calling this function:
+/// - `mem_size` must be a valid pointer to a `usize`.
+#[unsafe(no_mangle)]
+pub extern "C" fn NewInvertedIndex_Ex(
+    flags: IndexFlags,
+    raw_doc_id_encoding: bool,
```
gdesmott, line 145: Probably worth documenting those args.

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+    compress_floats: bool,
+    mem_size: &mut usize,
+) -> *mut InvertedIndex {
```
gdesmott, line 148: I'd also document that the returned ptr should be freed with `InvertedIndex_Free`.

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+    // Needs to track the entries count because it has the `StoreNumeric` flag set
+    Numeric(EntriesTrackingIndex<Numeric>),
+}
+
+impl InvertedIndex {
+    // Add a record to the inverted index. Returns by how many bytes the memory grew by.
+    fn add_record(&mut self, record: &RSIndexResult) -> std::io::Result<usize> {
+        use InvertedIndex::*;
+
+        match self {
```
gdesmott: Could be worth having an helper returning `ii` to save the duplication with `InvertedIndex_MemUsage`.

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6780/commits/d1345c5325143050c56e54390cecb20e900a47c1

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6780/commits/d1345c5325143050c56e54390cecb20e900a47c1

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6780/commits/86a3bc3e8342077b7fcf7f971bd6817108c3cd08

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
chesedo: This function already has it documented. But I added it to the helper in https://github.com/RediSearch/RediSearch/pull/6780/commits/86a3bc3e8342077b7fcf7f971bd6817108c3cd08

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+const FREQS_OFFSETS_MASK: IndexFlags =
+    IndexFlags_Index_StoreFreqs | IndexFlags_Index_StoreTermOffsets;
+const DOC_IDS_ONLY_MASK: IndexFlags = IndexFlags_Index_DocIdsOnly;
+const NUMERIC_MASK: IndexFlags = IndexFlags_Index_StoreNumeric;
+
+/// Create a new inverted index instance based on the provided flags and options. `raw_doc_encoding`
+/// controls whether document IDs only encoding should use raw encoding (true) or varint encoding
+/// (false). `compress_floats` controls whether numeric encoding should have its floating point
+/// numbers compressed (true) or not (false). Compressing floating point numbers saves memory
+/// but losses some precision.
```
LukeMathWalker:
```suggestion
/// but lowers precision.
```

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+        _ => panic!("Unsupported index flags: {flags:?}"),
+    };
+
+    *mem_size = ii_dispatch!(&ii, memory_usage);
+
+    let ii_boxed = Box::new(ii);
+    Box::into_raw(ii_boxed)
+}
+
+/// Free the memory associated with the inverted index instance created using [`NewInvertedIndex_Ex`].
```
LukeMathWalker:
```suggestion
/// Free the memory associated with an inverted index instance created using [`NewInvertedIndex_Ex`].
```

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+///
+/// # Safety
+/// The following invariant must be upheld when calling this function:
+/// - `ii` must be a valid pointer to an `InvertedIndex` instance created using
+///   [`NewInvertedIndex_Ex`] or `NewInvertedIndex`.
+/// - `ii` must not be NULL.
```
LukeMathWalker: Isn't the second invariant redundant? The constructors will never return a NULL pointer, no?

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+/// Get the memory usage of the inverted index instance in bytes.
+///
+/// # Safety
+/// The following invariant must be upheld when calling this function:
+/// - `ii` must be a valid pointer to an `InvertedIndex` instance and must not be NULL.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn InvertedIndex_MemUsage(ii: *const InvertedIndex) -> usize {
+    // SAFETY: The caller must ensure that `ii` is a valid pointer to an `InvertedIndex`
```
LukeMathWalker, line 230: Every time we assume the pointer must not be NULL, we should have a debug assertions that checks it.

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+    full::{Full, FullWide},
+    numeric::Numeric,
+    offsets_only::OffsetsOnly,
+    raw_doc_ids_only::RawDocIdsOnly,
+};
+
+/// An opaque inverted index structure. The actual implementation is determined at runtime based on
+/// the index flags provided when creating the index. However, C does not support generics, so we
+/// need to use an enum to represent the different implementations.
+pub enum InvertedIndex {
```
LukeMathWalker: While it is true that C doesn't support generics, we would have the same problem in Rust—whatever collection we use to store the inverted indexes, we would still need to normalize them to the same type via enum dispatch.

So I'd rephrase this slightly.

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
chesedo: Updated in https://github.com/RediSearch/RediSearch/pull/6780/commits/35f0f83b7ab3c9845253d6bb0a3a91f7563cea73

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6780/commits/35f0f83b7ab3c9845253d6bb0a3a91f7563cea73

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6780/commits/8ce67f2c030e3d7bfbc0a02a0788674c3652217e

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -0,0 +1,262 @@
+/*
```
GlenDC, line 1: I suppose Redis is ok with but in general I would frown upon the usage of the `ii` / `II` usage and instead just declare it as `inverted_index` and `InvertedIndex`. For a newcomers to the codebase II is for sure not gonna click. This is different then common knowledge such as `http` or `css`. II is however very specific to this codebase and not sure we need to safe on character count.

I have a feeling you'll keep it as is though, so feel free to just close the comment, wanted to either way voice my opinion on it.

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+            InvertedIndex::RawDocumentIdOnly(II::new(flags, RawDocIdsOnly))
+        }
+        (NUMERIC_MASK, _, false) => {
+            InvertedIndex::Numeric(EntriesTrackingIndex::new(flags, Numeric::new()))
+        }
+        (NUMERIC_MASK, _, true) => InvertedIndex::Numeric(EntriesTrackingIndex::new(
+            flags,
+            Numeric::new().with_float_compression(),
+        )),
+        _ => panic!("Unsupported index flags: {flags:?}"),
```
GlenDC, line 198: Is there a good reason why our only option is a panic? And if so why not go for something like `unreachable` again. Regardless I think it warrants mention  in the docs under "Panics" as that is a very extreme measure. For which I'm still not convinced it is needed, but even if it is, we need ti documented in the function docs.

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
chesedo: I removed the struct alias in https://github.com/RediSearch/RediSearch/pull/6780/commits/ea61c0cf862606e0e3fd2877012963252f873707.

I left the short variable name since that is always close to the context which makes it easy to see it means "inverted index"

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
chesedo: I documented when it can panic in https://github.com/RediSearch/RediSearch/pull/6780/commits/a8ffdb792c0fd4dfcf77e956bd3004e15aa82089

GlenDC (approved):

PR #6786, opened 2025-09-05
[8.0] [MOD-11011] Fix deadlock while RDB loading and RM_Yield (#6763)

JoanFM:


alonre24:
We are not maintaining 8.0, this one can be closed

PR #6796, opened 2025-09-08
[MOD-11174] Port NewInvIndIterator_NumericFull to Rust

gdesmott:
#### Mark if applicable


src/redisearch_rs/rqe_iterators/src/inverted_index_it.rs
```diff
@@ -0,0 +1,128 @@
...
+/// An iterator over numeric inverted index entries.
+///
+/// This iterator provides full index scan to all document IDs in a numeric inverted index.
+/// It is not suitable for queries.
+///
+/// # Type Parameters
+///
+/// * `'index` - The lifetime of the index being iterated over.
+/// * `E` - The encoder used to encode the inverted index.
+pub struct NumericFull<'index, E: DecodedBy> {
```
chesedo: Since this is for numeric inverted indices, is the generic `E` needed? Why not set it to use the `Numeric` decoder directly?

src/redisearch_rs/rqe_iterators/src/inverted_index_it.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    fn update_result(&mut self, record: RSIndexResult<'_>) -> Result<(), RQEIteratorError> {
+        self.last_doc_id = record.doc_id;
+        self.result.doc_id = record.doc_id;
+        let record_num = record
+            .as_numeric()
+            .ok_or(RQEIteratorError::WrongRecordType)?;
+        // SAFETY: we created this record as numeric
+        *self.result.as_numeric_mut().unwrap() = record_num;
```
chesedo: Is this done to not touch the other fields (metrics, dmd, etc) of the `IndexResult`?

src/redisearch_rs/rqe_iterators/src/inverted_index_it.rs
gdesmott: You're totally right, I over generalized for no reason. Fixed.

src/redisearch_rs/rqe_iterators/src/inverted_index_it.rs
gdesmott: No, as those are not set for numeric results.

This is because we had to change the iterator API to return a borrow rather than an owned result because of performance, see https://github.com/RediSearch/RediSearch/pull/6706#issuecomment-3228122396

src/redisearch_rs/rqe_iterators/src/inverted_index_it.rs
gdesmott: Wondering if I can then  remove `RQEIteratorError::WrongRecordType` and assert that the decoder will return numeric results.

What do you think?

gdesmott: I think that's the right move here so I changed it.

src/redisearch_rs/rqe_iterators_bencher/src/ffi.rs
```diff
@@ -77,6 +79,11 @@ impl QueryIterator {
         Self(unsafe { bindings::NewWildcardIterator_NonOptimized(max_id, num_docs, 1f64) })
     }
 
+    #[inline(always)]
+    pub fn new_numeric_full(ii: *mut bindings::InvertedIndex) -> Self {
+        Self(unsafe { bindings::NewInvIndIterator_NumericFull(ii) })
```
GlenDC, line 87: What's the value of having this function? Seems to just turn an unsafe function into a safe function without actually adding guarantees on top or using sane defaults. It's basically just a direct unsafe function call now appearing as "safe". Or am I missing something here?

src/redisearch_rs/rqe_iterators_bencher/src/ffi.rs
```diff
@@ -158,12 +165,46 @@ impl QueryIterator {
...
+    }
+
+    #[inline(always)]
+    pub fn iterator_numeric_full(&self) -> QueryIterator {
+        QueryIterator::new_numeric_full(self.0)
+    }
+}
+
+impl Drop for InvertedIndex {
+    fn drop(&mut self) {
```
GlenDC, line 198: Why is this one not inlined?

src/redisearch_rs/rqe_iterators_bencher/src/lib.rs
```diff
@@ -33,3 +44,52 @@ pub extern "C" fn ResultMetrics_Free(metrics: *mut ::ffi::RSYieldableMetric) {
...
+            pub extern "C" fn $fn_name() {
+                panic!(concat!(stringify!($fn_name), " should not be called by any of the benchmarks"));
+            }
+        )*
+    };
+}
+
+stub_c_fn! {
+    DocTable_Exists,
+    IndexSpec_GetFormattedKey,
```
GlenDC, line 72: How did you come up with this list btw? or was this just trial and error until it worked?

GlenDC: Would be nice to document more or less the reasoning behind the compilation of this list though, to make future reviewers and maintainers their life easier to know what is expected and how to evolve it safely.

src/redisearch_rs/rqe_iterators/tests/inverted_index.rs
```diff
@@ -0,0 +1,145 @@
...
+        let res = it.skip_to(self.doc_ids.last().unwrap() + 1);
+        assert!(matches!(res, Ok(None)));
+        // we just rewound
+        assert_eq!(it.last_doc_id(), 0);
+        assert!(it.at_eof());
+    }
+}
+
+#[test]
+/// test reading from NumericFull iterator
```
GlenDC, line 182: I feel the comments of these tests do not really tell much that the name doesn't tell. Besides that I wonder, shouldn't you do some more in depth testing? E.g. no validation of post-state is done, to say one thing.

src/redisearch_rs/rqe_iterators/tests/inverted_index.rs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        Self { doc_ids, ii }
+    }
+
+    /// test read functionality for a given iterator.
+    fn read<I: RQEIterator>(&self, mut it: I) {
+        let mut i = 0;
+
+        loop {
```
GlenDC: might be safer to add a high value and make it into a `for ..` range loop, so do not ever halt the test and can throw a more clear error if this ever loops (a lot) more than expected

src/redisearch_rs/rqe_iterators/tests/inverted_index.rs
```diff
@@ -0,0 +1,145 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use core::panic;
```
GlenDC: this is in the prelude normally, there should be no need to having to import this?

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -24,11 +28,15 @@ pub enum SkipToOutcome<'iterator, 'index> {
     NotFound(&'iterator RSIndexResult<'index>),
 }
 
-#[derive(Debug)]
+#[derive(Debug, Error)]
 /// An iterator failure indications
 pub enum RQEIteratorError {
     /// The iterator has reached the time limit for execution.
```
GlenDC, line 33: might a well drop the comments as the error descriptions and the comments are pretty much the same at this point

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -9,12 +9,16 @@
 
 pub mod empty;
 pub mod id_list;
+// suffixed with '_it' to prevent nameclash with the inverted_index crate
+pub mod inverted_index_it;
 pub mod wildcard;
 
 use ffi::t_docId;
```
GlenDC: nitpick: but it's a bit unconventional IMHO to declare modules above import statements. Usually import statements go above module definitions. But not sure if we care or if that's universally true.

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -9,12 +9,16 @@
 
 pub mod empty;
 pub mod id_list;
+// suffixed with '_it' to prevent nameclash with the inverted_index crate
+pub mod inverted_index_it;
```
GlenDC: That's very unconventional to suffix it for preventing nameclases, I think

what I see more often is

- import extern via `::` prefix, e.g. `use ::{inverted_index::RSIndexResult, thiserror::Error}`;
- use `self::inverted_index` to refer to the child module

src/redisearch_rs/rqe_iterators/src/inverted_index_it.rs
```diff
@@ -0,0 +1,112 @@
...
+            self.at_eos = true;
+            Ok(None)
+        }
+    }
+
+    fn skip_to(
+        &mut self,
+        doc_id: t_docId,
+    ) -> Result<Option<SkipToOutcome<'_, '_>>, RQEIteratorError> {
+        // skip_to pre-condition
```
GlenDC: pointless comment, a debug_assert is _always_ about a pre-condition, that said I do think these should be documented as on why we assume this as I feel the codebase (also the C one) is very light on this tribal knowledge

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        if !self.reader.skip_to(doc_id) {
+            self.at_eos = true;
+            return Ok(None);
+        }
+
+        if !self.reader.seek_record(doc_id, &mut self.result)? {
+            // reached end of iterator
+            self.at_eos = true;
```
GlenDC, line 88: C code seems to call `InvIndIterator_Read_Default` at this point?

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
```diff
@@ -0,0 +1,112 @@
...
+        if self.reader.next_record(&mut self.result)? {
+            self.last_doc_id = self.result.doc_id;
+            Ok(Some(&self.result))
+        } else {
+            self.at_eos = true;
+            Ok(None)
+        }
+    }
+
+    fn skip_to(
```
GlenDC, line 74: What about

- `InvIndIterator_SkipTo_withSeeker_CheckExpiration`
- `InvIndIterator_SkipTo_Default`
- `InvIndIterator_SkipTo_CheckExpiration`

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
```diff
@@ -0,0 +1,112 @@
...
+            reader,
+            at_eos: false,
+            last_doc_id: 0,
+            result: RSIndexResult::numeric(0.0),
+        }
+    }
+}
+
+impl<'index> RQEIterator for NumericFull<'index> {
+    fn read<'it>(&mut self) -> Result<Option<&RSIndexResult<'_>>, RQEIteratorError> {
```
GlenDC: What about

* `InvIndIterator_Read_SkipMulti`
* `InvIndIterator_Read_CheckExpiration`
* `InvIndIterator_Read_SkipMulti_CheckExpiration`

src/redisearch_rs/rqe_iterators_bencher/src/ffi.rs
gdesmott: It's just for convenience, so the bencher does not have to write any `unsafe` code. It also translate from the CI API naming (`NewInvIndIterator_NumericFull`) to the Rust one (`new_numeric_full`) so user code is more homogeneous.

src/redisearch_rs/rqe_iterators_bencher/src/ffi.rs
gdesmott: Good point, it probably should be. Fixed.

src/redisearch_rs/rqe_iterators/tests/inverted_index.rs
gdesmott: I ported the existing C++ tests.

The implementation of validation, and its extended test suites, will be part of https://github.com/RediSearch/RediSearch/pull/6821

src/redisearch_rs/rqe_iterators/tests/inverted_index.rs
gdesmott: Indeed, don't know why `vscode` added this. Removed.

src/redisearch_rs/rqe_iterators_bencher/src/lib.rs
gdesmott: > How did you come up with this list btw? or was this just trial and error until it worked?

Yes, automated using `claude` because life's too short.

Added a comment.

I agree it's a bit ugly, but this is just a temporary step as all this code will go away once we switch to the Rust iterators and remove the C ones.

src/redisearch_rs/rqe_iterators/tests/inverted_index.rs
gdesmott: done.

src/redisearch_rs/rqe_iterators/src/lib.rs
gdesmott: I rather keep the doc comment so it appears in the doc generated by `rustdoc` as well.

src/redisearch_rs/rqe_iterators/src/lib.rs
gdesmott: Good point, I changed it. Thanks for the suggestion.

src/redisearch_rs/rqe_iterators/src/lib.rs
gdesmott: Changed.

src/redisearch_rs/rqe_iterators/src/inverted_index_it.rs
gdesmott: Improved comment.

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
gdesmott: Those are not used for this specifc iterator. They will be implemented with the following ones.

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
gdesmott: Those are not used for this specifc iterator. They will be implemented with the following ones.

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
gdesmott: We implemented with @chesedo seek support on all Rust decoders,  while it was only implemented on some specific C ones. So we can just rely on that here saving us the need to sequentially read manually here.

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
GlenDC: ok I feel this might still need to. be documented perhaps as a NOTE or TODO, with some pointers and step-like explanation

gdesmott: I added a couple of  `TODO`.

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
```diff
@@ -0,0 +1,125 @@
...
+///
+/// This iterator provides full index scan to all document IDs in a numeric inverted index.
+/// It is not suitable for queries.
+///
+/// # Type Parameters
+///
+/// * `'index` - The lifetime of the index being iterated over.
+pub struct NumericFull<'index> {
+    /// The reader used to iterate over the inverted index.
+    reader: IndexReaderCore<'index, Numeric, Numeric>,
```
chesedo: Hmm, won't this need the potential numeric filters?

src/redisearch_rs/rqe_iterators/tests/inverted_index.rs
```diff
@@ -0,0 +1,143 @@
...
+        assert!(it.at_eof());
+        assert_eq!(it.num_estimated(), self.doc_ids.len());
+        assert_eq!(it.num_estimated(), self.ii.unique_docs());
+    }
+
+    /// test skip_to functionality for a given iterator.
+    fn skip_to<'index>(&self, it: &mut NumericFull<'index>) {
+        // Test skipping to any id between 1 and the last id
+        let mut i = 1;
+        for id in self.doc_ids.iter().copied() {
```
chesedo, line 99: Why is the `copied()` needed?

src/redisearch_rs/rqe_iterators/tests/inverted_index.rs
```diff
@@ -0,0 +1,143 @@
...
+            while i < id {
+                it.rewind();
+                let res = it.skip_to(i);
+                let Ok(Some(SkipToOutcome::NotFound(record))) = res else {
+                    panic!("skip_to {i} should succeed with NotFound: {res:?}");
+                };
```
chesedo, line 110: I'm confused. Why would this not find the `i` if we just reminded?

src/redisearch_rs/rqe_iterators_bencher/src/benchers/inverted_index.rs
```diff
@@ -0,0 +1,237 @@
...
+    fn skip_to_sparse(&self, c: &mut Criterion) {
+        let mut group = benchmark_group(c, "NumericFull", "SkipTo Sparse");
+        self.c_skip_to_sparse(&mut group);
+        self.rust_skip_to_sparse(&mut group);
+        group.finish();
+    }
+
+    fn c_index(sparse: bool) -> ffi::InvertedIndex {
+        let ii = ffi::InvertedIndex::new(ffi::IndexFlags_Index_StoreNumeric);
+        let delta = if sparse { SPARSE_DELTA } else { 1 };
```
chesedo: Why not take `delta` as an arg instead?

src/redisearch_rs/rqe_iterators_bencher/src/lib.rs
```diff
@@ -18,11 +18,22 @@ use std::ffi::c_void;
...
+pub use triemap_ffi::TRIEMAP_NOTFOUND;
 pub use types_ffi::NewVirtualResult;
+pub use varint_ffi::WriteVarint;
 
 redis_mock::bind_redis_alloc_symbols_to_mock_impl!();
 
+// symbols required by the C code we need to redefine
+#[unsafe(no_mangle)]
+#[allow(non_upper_case_globals)]
+pub static mut RSGlobalConfig: *const c_void = std::ptr::null();
```
chesedo, line 31: This is the reason for the flaky issue in #6848. We should set the `numericCompress` so that it does not get a random value and therefore result in random results in your tests.

chesedo: This might actually not be needed since none of your tests have floats that can be compressed. Just be aware of it if you were to ever update the tests.

src/redisearch_rs/rqe_iterators/tests/inverted_index.rs
gdesmott: To convert from `&u64`  to `u64`, saving me a `let id = *id;`

src/redisearch_rs/rqe_iterators/tests/inverted_index.rs
gdesmott: Because here we are skipping to an even id and the index only contains odd ones.

The test was really not that clear (I naively copied its logic from the C++ version). I made it clearer and added comments.

src/redisearch_rs/rqe_iterators_bencher/src/benchers/inverted_index.rs
gdesmott: Why indeed. Fixed.

src/redisearch_rs/rqe_iterators_bencher/src/lib.rs
gdesmott: That's good to know.

Do we already have code to create and set an actual `RSGlobalConfig` from Rust?
It's set to `null` in `inverted_index_bencher/src/lib.rs` as well.

chesedo: Not yet. It is that `null` setting in `inverted_index_bencher/src/lib.rs` which is the source of that flaky test

gdesmott: Ok, I'll keep this for now then as I don't use any floats atm and will update it with your solution once it's ready.

src/redisearch_rs/rqe_iterators/tests/inverted_index.rs
```diff
@@ -0,0 +1,153 @@
...
+impl<E: Encoder> BaseTest<E> {
+    fn new(enc: E) -> Self {
+        let n_records = E::RECOMMENDED_BLOCK_ENTRIES;
+        // Generate a set of odd document IDs for testing, starting from 1.
+        // The numeric record has a value of `doc_id * 2.0`.
+        let doc_ids = (0..n_records)
+            .map(|i| (2 * i + 1) as t_docId)
+            .collect::<Vec<_>>();
+
+        let mut ii = InvertedIndex::new(IndexFlags_Index_DocIdsOnly, enc);
```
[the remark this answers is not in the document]
gdesmott: Fixed.

src/redisearch_rs/rqe_iterators_bencher/src/lib.rs
chesedo: :+1: I have no solution for setting this. The switch to the Rust II also removes that flake test so it will no longer be an issue there. If you ever add floas tests then we will have to figure this out however.

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
gdesmott: Fixed using the marker traits we discussed yesterday.

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
```diff
@@ -0,0 +1,200 @@
...
+/// This iterator provides full index scan to all document IDs in an inverted index.
+/// It is not suitable for queries.
+///
+/// # Type Parameters
+///
+/// * `'index` - The lifetime of the index being iterated over.
+/// * `R` - The reader type used to read the inverted index.
+struct FullIterator<'index, R>
+where
+    R: IndexReader<'index>,
```
chesedo: Most Rust code doesn't qualify generics on struct definitions
```suggestion
struct FullIterator<'index, R>
```

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
```diff
@@ -0,0 +1,200 @@
...
+///
+/// This iterator provides full index scan to all document IDs in a numeric inverted index.
+/// It is not suitable for queries.
+///
+/// # Type Parameters
+///
+/// * `'index` - The lifetime of the index being iterated over.
+pub struct NumericFull<'index, R>
+where
+    R: NumericReader<'index>,
```
chesedo:
```suggestion
pub struct NumericFull<'index, R>
```

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
gdesmott: Indeed, fixed.

chesedo (dismissed):

src/redisearch_rs/rqe_iterators_bencher/src/lib.rs
```diff
@@ -18,11 +18,22 @@ use std::ffi::c_void;
...
+
+#[unsafe(no_mangle)]
+#[allow(non_upper_case_globals)]
+pub static mut RSDummyContext: *const c_void = std::ptr::null();
```
GuyAv46, line 35: Did we start using the Rust Redis module API?

src/tag_index.h
```diff
@@ -23,7 +23,7 @@ struct InvertedIndex;
 extern "C" {
 #endif
 
-/**
```
GuyAv46, line 26: Is there a motivation for this change?

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -894,6 +894,9 @@ impl<'index, I: Iterator<Item = RSIndexResult<'index>>> IndexReader<'index> for
     }
 }
 
+// Claim iterators are numeric readers so they can be used in tests.
+impl<'index, I: Iterator<Item = RSIndexResult<'index>>> NumericReader<'index> for I {}
```
GuyAv46, line 898: Temporary until we have other types?

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
```diff
@@ -0,0 +1,190 @@
...
+        if !self.reader.skip_to(doc_id) {
+            self.at_eos = true;
+            return Ok(None);
+        }
+
+        if !self.reader.seek_record(doc_id, &mut self.result)? {
+            // reached end of iterator
+            self.at_eos = true;
+            return Ok(None);
+        }
```
GuyAv46, line 90: Can we cover this case with a test?

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
```diff
@@ -0,0 +1,190 @@
...
+            self.at_eos = true;
+            Ok(None)
+        }
+    }
+
+    // TODO: this a port of InvIndIterator_SkipTo_Default, the simplest skip_to version.
+    // The more complex ones will be implemented as part of the next iterators:
+    // - InvIndIterator_SkipTo_withSeeker_CheckExpiration
+    // - InvIndIterator_SkipTo_withSeeker
+    // - InvIndIterator_SkipTo_CheckExpiration
```
GuyAv46: Just for the record, since the Rust reader has a seeker "built-in", we don't really need to implement 4 variants here, and this implementation is closer to `InvIndIterator_SkipTo_withSeeker` than to `InvIndIterator_SkipTo_Default`

src/tag_index.h
gdesmott: There is, see the commit message of https://github.com/RediSearch/RediSearch/pull/6796/commits/1550b9e2ee3685c2e5bf168604d8f982d5be2380

src/redisearch_rs/rqe_iterators_bencher/src/lib.rs
gdesmott: We do use it already, see https://github.com/RediSearch/RediSearch/blob/master/src/redisearch_rs/Cargo.toml#L110

Anyway looks like it's no longer needed to re-define this symbol so I removed it.

src/redisearch_rs/inverted_index/src/tests.rs
gdesmott: This only implemented for this specific tests module as a convenience as it makes the tests (which are using numerics) easier to write.

The implementation won't leak in the rest of the crate.

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
gdesmott: Good point. I updated the comment.

src/redisearch_rs/rqe_iterators/src/inverted_index.rs
gdesmott: Actually `seek_record()` calls `skip_to` as well, so we can simplify this code. Good catch!

GuyAv46 (dismissed):

PR #6800, opened 2025-09-08
MOD-11221: Delta construction helpers

GlenDC:
Add helpers in inverted_index.c to build and free GC deltas, replace direct InvIdxBuffers handling in fork_gc.c with these calls, and only pass an opaque delta handle to InvertedIndex_ApplyGcDelta.

Closes https://redislabs.atlassian.net/browse/MOD-11221

### DIFF
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -694,22 +694,30 @@ static void FGC_applyInvertedIndex(ForkGC *gc, InvIdxBuffers *idxData, MSG_Index
         InvertedIndex_Block(idx, info->nblocksOrig - 1);
   }
 
-  InvertedIndexGcDelta delta = {
-    .new_blocklist = idxData->newBlocklist,
-    .new_blocklist_size = idxData->newBlocklistSize,
-    .deleted = (void *)idxData->delBlocks,
-    .deleted_len = idxData->numDelBlocks,
-    .repaired = (void *)idxData->changedBlocks,
-    .repaired_len = info->nblocksRepaired,
-  };
+  // safety: ensure our shims match the wire structs (compile-time)
+  // NOTE: in later phase we get rid of MSG entirely
+  _Static_assert(sizeof(InvertedIndex_RepairedInput) == sizeof(MSG_RepairedBlock), "repaired layout mismatch");
+  _Static_assert(sizeof(InvertedIndex_DeletedInput)  == sizeof(MSG_DeletedBlock), "deleted layout mismatch");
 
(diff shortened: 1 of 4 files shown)


chesedo (dismissed):

GlenDC:
codecov is neglect-able. Code didn't change in functionality. Just small refactor as a next step.

src/fork_gc.c
```diff
@@ -694,22 +694,30 @@ static void FGC_applyInvertedIndex(ForkGC *gc, InvIdxBuffers *idxData, MSG_Index
...
-    .repaired = (void *)idxData->changedBlocks,
-    .repaired_len = info->nblocksRepaired,
-  };
+  // safety: ensure our shims match the wire structs (compile-time)
+  // NOTE: in later phase we get rid of MSG entirely
+  _Static_assert(sizeof(InvertedIndex_RepairedInput) == sizeof(MSG_RepairedBlock), "repaired layout mismatch");
+  _Static_assert(sizeof(InvertedIndex_DeletedInput)  == sizeof(MSG_DeletedBlock), "deleted layout mismatch");
 
-  InvertedIndex_ApplyGcDelta(idx, &delta, info->nblocksOrig, &info->nbytesAdded);
+  InvertedIndexGcDelta *delta = InvertedIndex_GcDelta_New();
```
BenGoldberger, line 702: why do we need this InvertedIndex_GcDelta_New function?

src/fork_gc.c
```diff
@@ -694,22 +694,30 @@ static void FGC_applyInvertedIndex(ForkGC *gc, InvIdxBuffers *idxData, MSG_Index
...
+
+  InvertedIndex_GcDelta_SetDeleted(delta,
+    (InvertedIndex_DeletedInput *)idxData->delBlocks, idxData->numDelBlocks);
+  idxData->delBlocks = NULL;
+  idxData->numDelBlocks = 0;
+
+  InvertedIndex_GcDelta_SetRepaired(delta,
+    (InvertedIndex_RepairedInput *)idxData->changedBlocks, info->nblocksRepaired);
+  idxData->changedBlocks = NULL;
+  info->nblocksRepaired = 0;
```
BenGoldberger, line 716: why split it into 3 set functions, and not just pass the idxData?

src/fork_gc.c
GlenDC: Because gc info doesn't know about the internals of the delta any longer, so no way it would be able to allocate it. Do you have a different idea in mind?

src/fork_gc.c
GlenDC: Because I want to get rid of the idxData. I think we might be able to get also rid of all these setters later, but need to see later.

PR #6801, opened 2025-09-09
MOD-11276, MOD-11043: Reconstruct RlookupRow when merging

Itzikvaknin:
## Implement RLookup Reconstruction for Hybrid Search

This PR replaces the old `RLookup` sync mechanism based on cloning with a reconstruction-based approach.  
It implements the new reconstruction mechanism and integrates it into the hybrid search flow for better schema handling and field management.  

### Key Changes
- **Added** `RLookup_AddKeysFrom()` — add keys from a source lookup into a destination  
- **Added** `RLookupRow_WriteFieldsFrom()` — write fields data of source row to dest row (with different schemas)  
- **Added** `merge_rlookuprow()` — merge field data from multiple hybrid search sources  
- **Removed** `RLookupKey_Clone()` and `RLookup_CloneInto()` — no longer needed  
- **Updated** `mergeSearchResults()` to use the new reconstruction-based lookup context and row merging for hybrid results  
- **Integrated** `HybridLookupContext` across the `FT.HYBRID` flow for consistent field management  
- **enabled** `YIELD_DISTANCE_AS` to test `mergeSearchResults()` in out of sync lookup indexes case.


tests/cpptests/test_cpp_rlookup.cpp
```diff
@@ -72,344 +72,668 @@ TEST_F(RLookupTest, testRow) {
...
+                           const std::vector<const char*>& fieldNames,
+                           const std::vector<double>& expectedValues) {
+  EXPECT_EQ(fieldNames.size(), expectedValues.size()) << "Field count must match expected value count";
 
-  // Test that the value can be read correctly
-  double num_val = 0;
-  int result = RSValue_ToNumber(retrieved, &num_val);
-  ASSERT_EQ(1, result);  // Should succeed
-  ASSERT_EQ(123.0, num_val);
+  for (size_t i = 0; i < fieldNames.size() && i < expectedValues.size(); i++) {
```
nafraf: Do we need to check both conditions? We have the assertion above which confirmed that `fieldNames.size() == expectedValues.size()`

src/hybrid/hybrid_request.c
```diff
@@ -11,6 +13,7 @@
 #include "info/info_redis/block_client.h"
 #include "query_error.h"
 #include "spec.h"
+#include "ext/default.h"
```
nafraf: leftover?

src/result_processor.c
```diff
@@ -20,6 +20,7 @@
 #include "util/references.h"
 #include "hybrid/hybrid_scoring.h"
 #include "hybrid/hybrid_search_result.h"
+#include "hybrid/hybrid_lookup_context.h"
```
nafraf: needed? We already included in `result_processor.h`

tests/pytests/test_rrf.py
```diff
@@ -37,8 +39,8 @@ def run_test_scenario(env, no_tag_search_query, with_tag_search_query):
     results_with_tag = get_results_from_hybrid_response(hybrid_res_with_tag)
     shared_keys = results_no_tag.keys() & results_with_tag.keys()
     for key in shared_keys:
-        score_no_tag = results_no_tag[key]
-        score_with_tag = results_with_tag[key]
+        score_no_tag = float(results_no_tag[key][SCORE_FIELD])
+        score_with_tag = float(results_with_tag[key][SCORE_FIELD])
```
nafraf, line 41: We are not using `score_with_tag`. 
The next assert is wrong. `score_no_tag == score_no_tag`

tests/cpptests/test_cpp_rlookup.cpp
Itzikvaknin: No we don't, removed it

src/hybrid/hybrid_request.c
Itzikvaknin: yes

src/result_processor.c
Itzikvaknin: No

tests/pytests/test_rrf.py
Itzikvaknin: Changed to 
```env.assertAlmostEqual(score_with_tag, score_no_tag, delta=1E-6)```

File: src/rlookup.h [hunk 5]
```diff
@@ -431,6 +417,29 @@ int jsonIterToValue(RedisModuleCtx *ctx, JSONResultsIterator iter, unsigned int
...
+ * If not exists, create new key in dest with unique dstidx.
+ * Handle existing keys based on flags (skip by default, override with RLOOKUP_F_OVERRIDE).
+ *
+ * Flag handling:
+ * - Preserves persistent source key properties (F_SVSRC, F_HIDDEN, F_EXPLICITRETURN, etc.)
+ * - Filters out transient flags from source keys (F_OVERRIDE, F_FORCE_LOAD)
+ * - Respects caller's control flags for behavior (F_OVERRIDE, F_FORCE_LOAD, etc.)
+ * - Final flags = caller_flags | (source_flags & ~RLOOKUP_TRANSIENT_FLAGS)
+ */
+void RLookup_AddKeysFrom(RLookup *dest, const RLookup *src, uint32_t flags);
```
oshadmi: Can put the src arg first?

oshadmi: Does `flags` overwrite already existing flags or are they being restored to before calling this function?

src/rlookup.h
```diff
@@ -431,6 +417,29 @@ int jsonIterToValue(RedisModuleCtx *ctx, JSONResultsIterator iter, unsigned int
...
+ * Add keys from source lookup into destination lookup.
+ * For each key in src, check if it already exists in dest by name.
+ * If not exists, create new key in dest with unique dstidx.
+ * Handle existing keys based on flags (skip by default, override with RLOOKUP_F_OVERRIDE).
+ *
+ * Flag handling:
+ * - Preserves persistent source key properties (F_SVSRC, F_HIDDEN, F_EXPLICITRETURN, etc.)
+ * - Filters out transient flags from source keys (F_OVERRIDE, F_FORCE_LOAD)
+ * - Respects caller's control flags for behavior (F_OVERRIDE, F_FORCE_LOAD, etc.)
+ * - Final flags = caller_flags | (source_flags & ~RLOOKUP_TRANSIENT_FLAGS)
```
oshadmi: Final flags can be named `target_flags`?

src/rlookup.h
```diff
@@ -431,6 +417,29 @@ int jsonIterToValue(RedisModuleCtx *ctx, JSONResultsIterator iter, unsigned int
...
+ */
+void RLookup_AddKeysFrom(RLookup *dest, const RLookup *src, uint32_t flags);
+
+/**
+ * Write field data from source row to destination row with different schemas.
+ * Iterate through source lookup keys, find corresponding keys in destination by name,
+ * and write it to destination row using RLookup_WriteOwnKey().
+ * Assumes all source keys exist in destination (enforce with ASSERT).
+ */
+void RLookupRow_WriteFieldsFrom(RLookupRow *srcRow, const RLookup *srcLookup,
```
oshadmi: Can `srcRow` be const?

src/rlookup.h
Itzikvaknin:
```
// Flags that are allowed to be passed to GetKey
#define RLOOKUP_GET_KEY_FLAGS (RLOOKUP_F_NAMEALLOC | RLOOKUP_F_OVERRIDE | RLOOKUP_F_HIDDEN | RLOOKUP_F_EXPLICITRETURN | \
                               RLOOKUP_F_FORCE_LOAD)
// Flags do not persist to the key, they are just options to GetKey()
#define RLOOKUP_TRANSIENT_FLAGS (RLOOKUP_F_OVERRIDE | RLOOKUP_F_FORCE_LOAD)
``` 

According to the documentation, these flags are transient (options to GetKey) and don’t persist on the key. 
Masking them out is effectively a no-op, but kept for readability. The intent is to let the caller decide how to handle key-conflict behavior.

src/hybrid/hybrid_search_result.c
```diff
@@ -147,10 +170,21 @@ SearchResult* mergeSearchResults(HybridSearchResult *hybridResult, HybridScoring
       mergeFlags(&primary->flags, &hybridResult->searchResults[i]->flags);
     }
   }
-
+  // Merge field data into primary result's rowdata if lookup context is provided
+  if (lookupCtx) {
```
ofiryanai: it's always true, HybridMerger_New always constructs lookupCtx so I think this check is redundant

tests/pytests/test_hybrid_distance.py
ofiryanai, line 1: Not sure how this is related to the PR

src/hybrid/hybrid_search_result.c
```diff
@@ -147,10 +170,21 @@ SearchResult* mergeSearchResults(HybridSearchResult *hybridResult, HybridScoring
...
+    // Create temporary row for merging (avoids modifying primary while reading from it)
+    RLookupRow tempRow = {0};  // Stack allocation, zero-initialized
+    merge_rlookuprow(hybridResult, lookupCtx, &tempRow);
+
+    // Prepare primary row and move merged data from temporary row
+    RLookupRow_Wipe(&primary->rowdata);  // Clear primary row
+    RLookupRow_Move(lookupCtx->tailLookup, &tempRow, &primary->rowdata);  // Move temp → primary
+    RLookupRow_Cleanup(&tempRow);
```
ofiryanai: I think some of these cleanups/wipes are redundant, but not sure

src/rlookup.c
```diff
@@ -988,69 +988,54 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
...
-    dst->flags = src->flags;  // Don't set NAMEALLOC if no strings
+    // Combine caller's control flags with source key's persistent properties
+    // Only preserve non-transient flags from source (F_SVSRC, F_HIDDEN, etc.)
+    // while respecting caller's control flags (F_OVERRIDE, F_FORCE_LOAD, etc.)
+    uint32_t combined_flags = flags | (src_key->flags & ~RLOOKUP_TRANSIENT_FLAGS);
+    RLookupKey *dest_key = RLookup_GetKey_Write(dest, src_key->name, combined_flags);
+    if (dest_key) {
+      // Copy properties from source (flags are already handled by RLookup_GetKey_Write)
+      dest_key->svidx = src_key->svidx;
+    }
```
ofiryanai: couldn't this use same index for different source keys? I thought needs to find a ne widx if there's a collision

src/rlookup.c
```diff
@@ -988,69 +988,54 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
...
-  } else {
-    //source has no name
-    //dst->name = NULL;
-    //dst->path = NULL;
-    dst->flags = src->flags;  // Don't set NAMEALLOC if no strings
+    // Combine caller's control flags with source key's persistent properties
+    // Only preserve non-transient flags from source (F_SVSRC, F_HIDDEN, etc.)
+    // while respecting caller's control flags (F_OVERRIDE, F_FORCE_LOAD, etc.)
+    uint32_t combined_flags = flags | (src_key->flags & ~RLOOKUP_TRANSIENT_FLAGS);
+    RLookupKey *dest_key = RLookup_GetKey_Write(dest, src_key->name, combined_flags);
```
ofiryanai, line 1006: this method finds the key if exist. maybe it could be enough to write all keys and if they already exist it will override them, then no need to wipe anything

tests/pytests/test_hybrid_distance.py
Itzikvaknin: This tests one of the main motivations for the PR (score in RLookup). 
The PR has two parts: updating the RLOOKUP mechanism and enabling vector aliasing.
They were combined into a single PR due to time pressure and stacked PRs.

src/rlookup.c
Itzikvaknin: Each `RLookupRow` has its own sorting vector, so the index is local to the `RLookup`.
It makes sense to let `RLookup_GetKey_Write` handle its management, since we read according to the source lookup and write according to the destination, this will work.
I removed the code, nice catch

src/rlookup.c
Itzikvaknin: We only perform a wipe in `mergeSearchResults` and `merge_rlookuprow`.
In `merge_rlookuprow`, it is done as a defensive precaution to ensure the refcounts of the destination row are handled correctly before writing to it.

In `mergeSearchResults`, `tempRow` gets wiped twice (once in `RLookupRow_Move` and once in `RLookupRow_Cleanup`). This is redundant, but I preferred to rely on the existing API for such a core data structure rather than its internals.

In any case, this code will likely be optimized later (or the solution might even change), so I’m comfortable leaving it as is for now.

src/hybrid/hybrid_search_result.c
Itzikvaknin: We only perform a wipe in `mergeSearchResults` and `merge_rlookuprow`.
In `merge_rlookuprow`, it is done as a defensive precaution to ensure the refcounts of the destination row are handled correctly before writing to it.

In `mergeSearchResults`, `tempRow` gets wiped twice (once in `RLookupRow_Move` and once in `RLookupRow_Cleanup`). This is redundant, but I preferred to rely on the existing API for such a core data structure rather than its internals.

In any case, this code will likely be optimized later (or the solution might even change), so I’m comfortable leaving it as is for now.

src/rlookup.c
Itzikvaknin: I left this decision to the caller, they can choose to override it using the flags parameter (`RLOOKUP_F_OVERRIDE`).

tests/pytests/test_hybrid_distance.py
ofiryanai: please state it in the PR description :)
I now see its one of the linked tickets but except MOD-11043 all title and description are related to the rlookup reconstruction

src/hybrid/hybrid_search_result.c
ofiryanai: ok. keep in mind that this is the hot path (happens for each search result per query, possible thousands of times per second) so I don't think we can allow ourselves to be over defensive

ofiryanai: ok. keep in mind that this is the hot path (happens for each search result per query, possible thousands of times per second) so I don't think we can allow ourselves to be over defensive

src/hybrid/hybrid_search_result.c
Itzikvaknin: Removed

tests/pytests/test_hybrid_distance.py
Itzikvaknin: Done

src/rlookup.h
```diff
@@ -431,6 +417,29 @@ int jsonIterToValue(RedisModuleCtx *ctx, JSONResultsIterator iter, unsigned int
  */
 const FieldSpec *findFieldInSpecCache(const RLookup *lookup, const char *name);
 
+/**
+ * Add keys from source lookup into destination lookup.
+ * For each key in src, check if it already exists in dest by name.
```
oshadmi, line 422: It is adding only non-overridden keys, right? Can be clearer, such as in this commnet or in the function name

src/rlookup.c
oshadmi: We cannot easily check that all additions were successful, right? (`RLookup_GetKey_Write` may return `NULL` also upon success)
Nitpick: So if we keep this function `void` - no need for the local `dest_key` variable

src/rlookup.c
```diff
@@ -988,69 +988,50 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
...
+    RSValue *value = RLookup_GetItem(src_key, srcRow);
+    if (!value) {
+      // No data for this key in source row
+      continue;
     }
 
-    prev_dst_key = dst_key;
-    src_key = src_key->next;
+    // Find corresponding key in destination lookup
+    RLookupKey *dest_key = RLookup_FindKey(destLookup, src_key->name, src_key->name_len);
```
oshadmi, line 1030: If needed - a potential performance improvement could be to support constant time lookup since now it is called here and also on all source fields from `RLookup_AddKeysFrom`

src/result_processor.c
```diff
@@ -1866,6 +1866,12 @@ static inline bool RPHybridMerger_Error(const RPHybridMerger *self) {
    // Free the upstreams array, the upstreams themselves are freed by the pipeline(e.g as a result of AREQ_Free)
    array_free(self->upstreams);
 
+   // Free lookup context if it exists
+   if (self->lookupCtx) {
```
oshadmi, line 1870: Nitpick: Should `HybridLookupContext` have a dedicated free function?

src/hybrid/hybrid_request.c
```diff
@@ -55,17 +57,48 @@ arrayof(ResultProcessor*) HybridRequest_BuildDepletionPipeline(HybridRequest *re
...
+    }
+
+    // Build lookup context for field merging
+    HybridLookupContext *lookupCtx = rm_calloc(1, sizeof(HybridLookupContext));
+    lookupCtx->tailLookup = tailLookup;
+    lookupCtx->sourceLookups = array_newlen(const RLookup*, nrequests);
+
+    // Collect source lookups
+    for (size_t i = 0; i < nrequests; i++) {
+        lookupCtx->sourceLookups[i] = AGPLN_GetLookup(AREQ_AGGPlan(requests[i]), NULL, AGPLN_GETLOOKUP_FIRST);
```
oshadmi: Could the array have a NULL `RLookup*` entry?
If `srcLookup` above is NULL we skip it. Here we don't.
Do we need the `RsourceLookups` array to be aligned with the `requests` array?

src/rlookup.h
Itzikvaknin: Edited the comment 
```
/**
 * Add non-overridden keys from source lookup into destination lookup (overriden keys are skipped).
 * For each key in src, check if it already exists in dest by name.
 * If doesn't exists, create new key in dest.
 * Handle existing keys based on flags (skip with RLOOKUP_F_NOFLAGS, override with RLOOKUP_F_OVERRIDE).
 */
```

src/rlookup.c
Itzikvaknin: Sure, we can have `src_key -> dest_key` mapping after one iteration and use it to save lookup time,
I think this is the right approach, moving as much as we can to "construction time"

src/result_processor.c
Itzikvaknin: It is used as a wrapper and is doesn't own anything
I wanted to avoid creating a `.c`  for hybrid_lookup_context.h

src/hybrid/hybrid_request.c
Itzikvaknin: No, I added assertions to ensure there are no null pointers used as lookups at any point.
I also updated the docs in `InitializeHybridLookupContext` and `HybridRequest_BuildMergePipeline`.

This assumption is valid because `HybridRequest_BuildMergePipeline` calls `InitializeHybridLookupContext`, and it is only invoked after `HybridRequest_BuildDepletionPipeline`, which in turn calls `AREQ_BuildPipeline → Pipeline_BuildQueryPart → RLookup_Init` (all unconditionally).

We need to ensure this assumption holds whenever `HybridRequest_BuildMergePipeline` is called, which I expect will only be in the coordinator.
I also added a note in the [ticket](https://redislabs.atlassian.net/browse/MOD-10039) to verify this, although the assertion should immediately catch any violation.

`sourceLookups` must be aligned with the `requests` array, since we rely on position for `SEARCH/VSIM` resolution.

tests/cpptests/test_cpp_rlookup.cpp
```diff
@@ -72,344 +72,714 @@ TEST_F(RLookupTest, testRow) {
...
+}
 
-  // Test writing with original and reading with cloned (same row, same lookup)
-  RLookupRow rr = {0};
-  RSValue *value = RS_Int64Val(123);
+// Helper: Create test values with distinct integers starting from baseValue
+std::vector<RSValue*> create_test_values(const std::vector<int>& values) {
+  std::vector<RSValue*> rsValues;
+  for (int val : values) {
+    rsValues.push_back(RS_Int64Val(val));
```
oshadmi, line 101: For later, consider testing with more value types such as RS_StringVal

tests/cpptests/test_cpp_rlookup.cpp
```diff
@@ -72,344 +72,714 @@ TEST_F(RLookupTest, testRow) {
...
+    RSValue_Decref(values[i]);
+  }
+  RLookupRow_Cleanup(&srcRow);
+  RLookupRow_Cleanup(&destRow);
   RLookup_Cleanup(&source);
   RLookup_Cleanup(&dest);
 }
+
+
+// Tests hybrid search with distinct field sets from each source
```
oshadmi: Need to update the comment describing the test

tests/cpptests/test_cpp_rlookup.cpp
```diff
@@ -72,344 +72,714 @@ TEST_F(RLookupTest, testRow) {
...
+  cleanup_values(src2Values);
+  RLookupRow_Cleanup(&src1Row);
+  RLookupRow_Cleanup(&src2Row);
+  RLookupRow_Cleanup(&destRow);
+  RLookup_Cleanup(&src1);
+  RLookup_Cleanup(&src2);
+  RLookup_Cleanup(&dest);
+}
+
+// Tests hybrid search with overlapping field names (last write wins)
```
oshadmi: Need to update the comment describing the test

tests/cpptests/test_cpp_rlookup.cpp
```diff
@@ -72,344 +72,714 @@ TEST_F(RLookupTest, testRow) {
...
-  ASSERT_TRUE(override_src_key);
-  ASSERT_NE(src_key, override_src_key);  // Should be different keys
+  // Create rows and add data with distinct values
+  RLookupRow srcRow = {0}, destRow = {0};
 
-  // Verify the overridden field structure is cloned properly
-  RLookupKey *dest_key = RLookup_GetKey_Read(&dest, "test_field", RLOOKUP_F_NOFLAGS);
-  ASSERT_TRUE(dest_key);
+  // Create test values with distinct data
+  RSValue *values[] = {RS_Int64Val(111), RS_Int64Val(222), RS_Int64Val(333)};
```
oshadmi: Can use `create_test_values` ?

tests/cpptests/test_cpp_rlookup.cpp
```diff
@@ -72,344 +72,714 @@ TEST_F(RLookupTest, testRow) {
...
+
+  // Add keys (first source wins for key creation, but last write wins for data)
+  RLookup_AddKeysFrom(&src1, &dest, RLOOKUP_F_NOFLAGS);
+  RLookup_AddKeysFrom(&src2, &dest, RLOOKUP_F_NOFLAGS);
+
+  // Create rows with conflicting data for "field2"
+  RLookupRow src1Row = {0}, src2Row = {0}, destRow = {0};
+
+  // Create src1 values: field1=1, field2=100, field3=3
+  RSValue *s1_vals[] = {RS_Int64Val(1), RS_Int64Val(100), RS_Int64Val(3)};
```
oshadmi: Can use `create_test_values` ?

src/rlookup.c
```diff
@@ -988,69 +988,50 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
...
+    RS_ASSERT(dest_key != NULL);  // Assumption: all source keys exist in destination
```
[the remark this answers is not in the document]
oshadmi: Assuming all keys were already added to the destination (detailed in the function declaration comments) - so `RLookup_FindKey` will find them.

oshadmi (approved):

PR #6804, opened 2025-09-09
[MOD-9916] Swap `Counter` result processor to rust

enricozb:
Swaps to the Rust implementation of the counter result processor.

1. Deletes C code defining `RPCounter` and associated methods.
2. Adds `result_processor_ffi` to the `c_entrypoint` deps.
3. Exposes a new C function `RPProfile_IncrementCount` to rust through the FFI.
    - This is because the original C implementation of `rpcountNext` accessed a private field of an `RPProfile` struct, and performing the correct pointer math on the rust side was difficult as the struct size was variable under debug builds, resulting in OOB memory writes.
4. Modifies `Chain` in the result processing testing utils to also mock a `QueryProcessingCtx` parent structure.

1. `RPCounter`

#### Mark if applicable

---

> [!NOTE]
> Migrates `RPCounter` to a Rust `Counter`, adds an FFI profiling count increment, and updates context/plumbing and tests accordingly.
> 
> - **Result Processing (Core)**:
>   - Replace C `RPCounter` with Rust `Counter` in `result_processor` crate; remove C `RPCounter` code from `src/result_processor.c` and declaration from `src/result_processor.h`.
>   - Profiling: when upstream is `RP_PROFILE`, Rust `Counter` manually bumps profile count via new C API `RPProfile_IncrementCount`.
>   - Introduce Rust `ffi::QueryProcessingCtx` and plumb parent pointers; change `Header.parent` to `*const ffi::QueryProcessingCtx`; expose `Context::parent()` and `Upstream::ty()`.
>   - Test utils: `Chain` now owns a `QueryProcessingCtx` and appends processors to it; add `MockResultProcessor` and a profiling count test.
> - **FFI/Entrypoint**:
>   - Add `result_processor_ffi` dependency and re-export (`redisearch_rs` entrypoint); fix cbindgen forward-decl to `typedef struct ResultProcessor ResultProcessor;` and include `result_processor_rs.h` from C.
> - **C API**:
>   - Add `RPProfile_IncrementCount` implementation and declaration; keep `RPProfile_*` getters; leave `Profile_AddRPs` behavior unchanged.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0a2a3f7cd605c0f534913dca64db6ac06316f0ad. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.codespell/.codespellrc
+++ b/.codespell/.codespellrc
@@ -1,6 +1,6 @@
 [codespell]
 # Ignore certain files and directories.
-skip = .git,./deps/*,*.csv,./srcutil/*,./bin/*,./sbin/*,*/parser.c,*/parser.h,*/parser.out,*/lexer.c,*/Makefile,src/redisearch_rs/trie_bencher/data/*,src/redisearch_rs/wildcard/tests/integration/*,src/redisearch_rs/wildcard/benches/*
+skip = .git,./deps/*,*.csv,./srcutil/*,./bin/*,./sbin/*,*/parser.c,*/parser.h,*/parser.out,*/lexer.c,*/Makefile,src/redisearch_rs/trie_bencher/data/*,src/redisearch_rs/wildcard/tests/integration/*,src/redisearch_rs/wildcard/benches/*,tests/cpptests/test_cpp_trie.cpp
 # Ignore words.
 ignore-words = .codespell/ignore_wordlist.txt
 
--- a/.github/workflows/event-merge-to-queue.yml
+++ b/.github/workflows/event-merge-to-queue.yml
@@ -48,6 +48,13 @@ jobs:
       rejson-branch: master
       test-config: '' # run all tests
 
+  run-on-intel:
+    needs: get-latest-redis-tag
+    secrets: inherit
+    uses: ./.github/workflows/flow-intel.yml
+    with:
+      get-redis: ${{ needs.get-latest-redis-tag.outputs.tag }}
+
   sanitize:
     needs: get-latest-redis-tag
     secrets: inherit
@@ -65,6 +72,7 @@ jobs:
       - test-linux
       - test-macos
       - coverage
+      - run-on-intel
       - sanitize
     runs-on: ${{ vars.RUNS_ON || 'ubuntu-latest' }}
     if: ${{ !cancelled() }}
--- a/.github/workflows/event-nightly.yml
+++ b/.github/workflows/event-nightly.yml
@@ -33,6 +33,12 @@ jobs:
       test-config: QUICK=1
       get-redis: unstable
 
+  run-on-intel:
+    secrets: inherit
+    uses: ./.github/workflows/flow-intel.yml
+    with:
+      get-redis: unstable
+
   sanitize:
     secrets: inherit
     uses: ./.github/workflows/task-test.yml
@@ -48,7 +54,7 @@ jobs:
 
 
   notify-failure:
-    needs: [test-linux, test-macos, coverage, sanitize, micro-benchmarks]
+    needs: [test-linux, test-macos, coverage, run-on-intel, sanitize, micro-benchmarks]
     if: failure()
     runs-on: ${{ vars.RUNS_ON || 'ubuntu-latest' }}
     steps:
--- a/.github/workflows/event-pull_request.yml
+++ b/.github/workflows/event-pull_request.yml
@@ -33,6 +33,7 @@ jobs:
     if: >
       (
         needs.check-what-changed.outputs.CODE_CHANGED == 'true' ||
+        needs.check-what-changed.outputs.TESTS_CHANGED == 'true' ||
         contains(github.event.pull_request.labels.*.name, 'enforce:test')
       )
     uses: ./.github/workflows/task-test.yml
@@ -47,6 +48,7 @@ jobs:
     if: >
       vars.ENABLE_CODE_COVERAGE != 'false' && (
           (!github.event.pull_request.draft && needs.check-what-changed.outputs.CODE_CHANGED == 'true') ||
+          (!github.event.pull_request.draft && needs.check-what-changed.outputs.TESTS_CHANGED == 'true') ||
           contains(github.event.pull_request.labels.*.name, 'enforce:coverage')
       )
     uses: ./.github/workflows/task-test.yml
@@ -62,6 +64,7 @@ jobs:
     if: >
       (
         (!github.event.pull_request.draft && needs.check-what-changed.outputs.CODE_CHANGED == 'true') ||
+        (!github.event.pull_request.draft && needs.check-what-changed.outputs.TESTS_CHANGED == 'true') ||
         contains(github.event.pull_request.labels.*.name, 'enforce:sanitize')
       )
     secrets: inherit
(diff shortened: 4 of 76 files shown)


src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -123,6 +123,18 @@ impl Context<'_> {
...
+    pub fn last_processor(&mut self) -> *mut ffi::ResultProcessor {
+        // Safety: We trust that this result processor's pointer is valid.
+        let query_processing_context_ptr = unsafe { self.ptr.as_ref() }.parent;
+        // Safety: We trust that the pointer to the owning (parent) structure is valid.
+        let query_processing_context_ref = unsafe { query_processing_context_ptr.as_ref() };
+
+        query_processing_context_ref
+            .expect("result processor `parent` was null")
+            .endProc
+    }
```
JonasKruckenberg: We should split this apart into a `parent` method that returns something like `Option<&ffi::QueryProcessingCtx>` so that in the future we can access more fields of that context when needed.

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -38,6 +38,15 @@ impl ResultProcessor for Counter {
...
         }
 
+        // This logic was present in the original C code, with the following comment:
+        // > Since this never returns RM_OK, in profile mode, count should be increased to compensate for EOF.
+        if upstream.ty() == ffi::ResultProcessorType_RP_PROFILE {
+            // Safety: If the upstream processor is a Profile, then the original C code currently
+            // expects the last processor (`base->parent->endProc`) to be a `RPProfile*`, and
+            // increments its count. This is replicating that behavior.
+            unsafe { ffi::RPProfile_IncrementCount(cx.last_processor()) }
+        }
```
JonasKruckenberg: as a general style rule, we try to write comments "as if the C code doesn't exist". The C code will be gone soon and these comment would be referencing non-existent code.

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -133,6 +145,12 @@ pub struct Upstream<'a> {
 }
 
 impl Upstream<'_> {
+    pub fn ty(&self) -> ffi::ResultProcessorType {
+        // Safety: We have to trust that the upstream pointer set by our QueryInterator parent
+        // is correct
```
JonasKruckenberg:
```suggestion
        // Safety: We have to trust that the upstream pointer set by our QueryProcessingCtx parent
        // is correct
```

JonasKruckenberg (changes requested):
Looks pretty solid, just a few minor style nitpicks.

JonasKruckenberg:
RE the failing tests: You will need to add a mock implementation of `RPProfile_IncrementCount` (that ideally should update a counter field in some mock RP that we can then assert in tests).

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -38,6 +38,26 @@ impl ResultProcessor for Counter {
             }
         }
 
+        // In profiling mode, RPProfile is interleaved into the result processor chain: A chain of
+        // processors A -> B -> C becomes A -> RPProfile -> B -> RPProfile -> C -> RPProfile, to
+        // profile each of the individual result processors.
+        //
+        // Because the Counter result processor returns Ok(None), this is equivalent to returning
+        // ffi::RPStatus_RS_RESULT_EOF (see ResultProcessorWrapper::result_processor_next). This
+        // apparently (in a way enricozb cannot figure out) prevents from the very last RPProfile
```
enricozb:
```suggestion
        // apparently (in a way enricozb cannot figure out) prevents the very last RPProfile
```

src/redisearch_rs/result_processor/src/test_utils.rs
```diff
@@ -98,13 +142,24 @@ impl Chain {
 
         if let Some(upstream) = self.result_processors.last() {
             result_processor.header.upstream = upstream.as_ptr();
+            result_processor.header.parent = &mut self.query_processing_context;
```
JonasKruckenberg: We should set the parent pointer regardless of the presence of an upstream

src/redisearch_rs/c_entrypoint/result_processor_ffi/tests/counter.rs
```diff
@@ -37,6 +37,12 @@ unsafe extern "C" fn SearchResult_Clear(r: *mut ffi::SearchResult) {
     //   DMD_Return(r->dmd);
 }
 
+/// Mock implementation of `RPProfile_IncrementCount` for tests
+///
+// FIXME: replace with `Profile::increment_count` once the profile result processor is ported.
+#[unsafe(no_mangle)]
+unsafe extern "C" fn RPProfile_IncrementCount(_r: *mut ffi::ResultProcessor) {}
```
JonasKruckenberg: We do not expect to call this here since it only exists for linking right? We should turn a potentially silent bug into a panic in that case

```suggestion
/// Mock implementation of `RPProfile_IncrementCount` for tests
///
// FIXME: replace with `Profile::increment_count` once the profile result processor is ported.
#[unsafe(no_mangle)]
unsafe extern "C" fn RPProfile_IncrementCount(_r: *mut ffi::ResultProcessor) {
   unreachable!()
}
```

src/redisearch_rs/result_processor/src/test_utils.rs
```diff
@@ -95,16 +135,29 @@ impl Chain {
         P: ResultProcessor + 'static,
     {
         let mut result_processor = ResultProcessorWrapper::new(result_processor);
+        let query_processing_context_ptr: *mut ffi::QueryProcessingCtx =
+            &*self.query_processing_context as *const _ as *mut _;
```
JonasKruckenberg: lets use as few `as` pointer casts as possible (we should esp never use *inferring* (e.g. `as *mut _` or `as *const _`) casts)
```suggestion
        let query_processing_context_ptr: *mut ffi::QueryProcessingCtx =
        self.query_processing_context.as_ptr().cast_mut();
```

enricozb: changed to `&raw mut *self.query_processing_context`.

JonasKruckenberg (approved):

src/redisearch_rs/result_processor/src/test_utils.rs
```diff
@@ -95,16 +135,28 @@ impl Chain {
         P: ResultProcessor + 'static,
     {
         let mut result_processor = ResultProcessorWrapper::new(result_processor);
+        let query_processing_context_ptr = &raw mut *self.query_processing_context;
 
         if let Some(upstream) = self.result_processors.last() {
             result_processor.header.upstream = upstream.as_ptr();
+            result_processor.header.parent = query_processing_context_ptr;
```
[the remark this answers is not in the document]
enricozb: Fixed

ofiryanai (dismissed):
I understand the code, didn't find any issue. Seems like it only adds RPProfile case. LGTM

ofiryanai (approved):

PR #6805, opened 2025-09-09
[MOD-9811] Rust FFI for inverted index

chesedo:
This adds the remaining FFI for the inverted index Rust implementation. The C code won't call this yet since the generated code is in `headers/future`. However, this PR should cover all the functions on an inverted index the C code will call once the switch happens.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/headers/future/inverted_index.h
+++ b/src/redisearch_rs/headers/future/inverted_index.h
@@ -173,6 +173,16 @@ void InvertedIndex_BlocksSummaryFree(IIBlockSummary *blocks,
 t_fieldMask InvertedIndex_FieldMask(const struct InvertedIndex *ii);
 
 /**
+ * Get the number of entries in the inverted index. This is only valid for numeric indexes created
+ * with the `StoreNumeric` flag. For other index types, this function will return 0.
+ *
+ * # Safety
+ * The following invariant must be upheld when calling this function:
+ * - `ii` must be a valid pointer to an `InvertedIndex` instance and cannot be NULL.
+ */
+uintptr_t InvertedIndex_NumEntries(const struct InvertedIndex *ii);
+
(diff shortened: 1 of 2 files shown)


gdesmott (approved):

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -576,10 +612,45 @@ impl<E: Encoder> FieldMaskTrackingIndex<E> {
...
+
+    /// Returns the number of unique documents in the index.
+    pub fn unique_docs(&self) -> usize {
+        self.index.unique_docs()
+    }
+
+    /// Returns the flags of this index.
+    pub fn flags(&self) -> IndexFlags {
+        self.index.flags()
+    }
```
BenGoldberger, line 628: dont we want these functions to be part of some trait?

src/redisearch_rs/headers/future/inverted_index.h
BenGoldberger, line 1: can you explain whats the purpose of headers/future ?

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Maybe eventually, but for now we don't need this to be abstract

src/redisearch_rs/headers/future/inverted_index.h
chesedo: To make a file which matches `src/inverted_index/inverted_index.h` over multiple PRs which are easy to review. At the same time I don't want the C code to link to this file yet, hence why it is under `/future`. This will also keep the last PR which will switch from the C inverted index to the Rust inverted index small since this file will only move up one directory

BenGoldberger (approved):

PR #6809, opened 2025-09-10
MOD-10398 rlookup: port SearchResult

JonasKruckenberg:
This change implements the `SearchResult` type in Rust.

stacked on top off #6808

---

> [!NOTE]
> Introduce a new `search_result` crate implementing `SearchResult` and `DocumentMetadata` bindings, wire it into the workspace, extend FFI headers, and tweak `RLookupRow`.
> 
> - **search_result (new crate)**:
>   - Implement `SearchResult` with `doc_id`, `score`, `score_explain`, `document_metadata`, `index_result`, `row_data`, and `flags`, including clear/reset lifecycle and accessors.
>   - Add `bindings::DocumentMetadata` ref-counted wrapper over `ffi::RSDocumentMetadata` with `Clone`/`Drop`.
> - **FFI/bindgen**:
>   - Add headers `src/doc_table.h`, `src/score_explain.h`, `src/rlookup.h` to bindings generation.
> - **rlookup**:
>   - `RLookupRow`: derive `Clone`; make `new()` a `const fn`; minor doc updates.
> - **Workspace/Cargo**:
>   - Add `search_result` to workspace members and dependencies.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f2351c59074e1cc93ba2021a6fbd351e6b2df8b1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #6808, opened 2025-09-10
  MOD-10399 refactor: encapsulate SearchResult

JonasKruckenberg:
This change refactors the `SearchResult` interface, replacing all direct field accesses with `static inline` getter & setter functions. This will enable the porting to Rust later as we are now free to change the exact layout and shape of `SearchResult`.

To prevent accidental use the fields, I renamed each `__<old name>`.

---

> [!NOTE]
> Encapsulates `SearchResult` behind a new accessor-based API, moves its impl to dedicated files, and updates processors, serialization, bindings, and tests accordingly with const-correctness fixes.
> 
> - **Core API**:
>   - Introduces `search_result.h/.c` encapsulating `SearchResult` with inline getters/setters and helpers (`SearchResult_*`), migrating cleanup/override logic from `result_processor.c`.
>   - Replaces direct field access across the codebase with accessor calls; updates includes to use the new header.
>   - Makes `SEReply` take `const RSScoreExplain*`; makes `IndexResult_GetMatchedTerms` `const`-correct.
> - **Processors/Execution**:
>   - Updates aggregate, hybrid, highlight, group-by, sorting, metrics, loaders, depleter, and normalizers to use `SearchResult_*` accessors and `RLookupRow` getters/mutators.
>   - Hybrid merger now merges flags via `SearchResult_MergeFlags` and uses accessors for score/docId/rows.
>   - Minor serialization changes to use accessors consistently.
> - **Build/Bindings**:
>   - Adds `search_result.h` to Rust FFI bindgen headers.
> - **Tests**:
>   - Adjusts C++ tests to use new `SearchResult` API and updated behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 527ab4770c2e971836f1b562058374437e5f098c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #6810, opened 2025-09-10
rlookup: SearchResult FFI

JonasKruckenberg:
This change implements the C FFI interface for the `SearchResult` type ported in #6809

stacks on top off #6808, #6809


src/redisearch_rs/c_entrypoint/search_result_ffi/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+use search_result::{SearchResultFlags, bindings::DocumentMetadata};
+use std::ptr::{self, NonNull};
+
+/// cbindgen:ignore
+pub type SearchResult = search_result::SearchResult<'static>;
+
+/// Construct a new [`SearchResult`].
+#[unsafe(no_mangle)]
+pub extern "C" fn SearchResult_New() -> mimic::Size72Align8 {
+    SearchResult::new().into_mimic()
```
kei-nan, line 23: what does this code do?
From looking online it looks like rust's mimic library is more meant for testing?
But here it seems to be production code.

src/redisearch_rs/c_entrypoint/search_result_ffi/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+///
+/// 1. `res` must be a [valid], non-null pointer to a [`SearchResult`].
+/// 2. `res` **must not** be used again after this function is called.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn SearchResult_Destroy(res: Option<NonNull<SearchResult>>) {
+    // Safety: ensured by caller (1.,2.)
+    unsafe { res.unwrap().drop_in_place() };
+}
```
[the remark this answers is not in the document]
kei-nan: @JonasKruckenberg FYI

src/redisearch_rs/Cargo.lock
```diff
@@ -988,6 +988,11 @@ version = "2.7.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32a282da65faaf38286cf3be983213fcf1d2e2a58700e808f83f4ea9a4804bc0"
 
+[[package]]
+name = "mimic"
+version = "0.1.0"
+source = "git+https://github.com/JonasKruckenberg/mimic?branch=jonas%2Fcbindgen-compat#119ed3804fd180af7e5211f2e43d4673d958f000"
```
kei-nan: are we okay pointing the source to such a fork under your username?
Maybe there is more official url we can use? wiith a tag/release?

src/redisearch_rs/Cargo.toml
```diff
@@ -95,6 +95,7 @@ ureq = "3.0.10"
 wildcard_cloudflare = { package = "wildcard", version = "0.3.0" }
 pin-project = "1.1.10"
 strum = {version = "0.27.2",features = ["derive"]}
+mimic = { git = "https://github.com/JonasKruckenberg/mimic", branch = "jonas/cbindgen-compat" }
```
kei-nan: not a fan of using a non official library/fork,.
Maybe there is a more official version which we can use?

JonasKruckenberg:
`mimic` is a crate we designed and developed (specifically for use in redisearch but has wider applicability). The problem it solves is not being able to stack-allocate opaque types. Being able to pass opaque types from Rust to C is great bc that means we don't need to disclose the internals of Rust code to C. We're also free to use types and shapes that we cannot communicate to C. The problem of course is that opaque types have _unknown size and alignment_ and therefore cannot be allocated on the stack. This means we quite often would have to move types that were stack allocated in C to be heap allocated in Rust with the obvious perf problems that entails.
`mimic` works around this by giving us "mimic types" that have the same size and alignment as the Rust type but none of the fields (just an array of bytes). As such the current GH link is _the_ original source for the crate and not some fork.

I agree though we should not commit it as is, we should publish this first 👍🏻

src/redisearch_rs/c_entrypoint/search_result_ffi/src/lib.rs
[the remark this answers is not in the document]
JonasKruckenberg: Given that only a single place in code calls the `SearchResult_AllocateMove` function I introduced a special destructor `SearchResult_DeallocateDestroy` to be called in this situation. A more rigorous runtime tracking of allocation status seems overblown here.

JonasKruckenberg:
Update: in the mean time we published the crate(s) as `unsafe-tools` on crates.io

PR #6811, opened 2025-09-10
rlookup: SearchResult swap out C with Rust implementation

JonasKruckenberg:
This change swaps out the C `SearchResult` implementation for the Rust one.

stacked on top off #6808, #6809, #6810


src/redisearch_rs/c_entrypoint/search_result_ffi/src/lib.rs
```diff
@@ -84,7 +84,7 @@ pub unsafe extern "C" fn SearchResult_Destroy(res: Option<NonNull<SearchResult>>
 ///
 /// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
 #[unsafe(no_mangle)]
-pub const unsafe extern "C" fn SearchResult_GetDocId(res: *const SearchResult) -> ffi::t_docId {
```
kei-nan, line 87: why is this no longer const?
what does const means in this context?

src/redisearch_rs/result_processor/src/counter.rs
```diff
@@ -21,7 +22,7 @@ impl ResultProcessor for Counter {
     fn next(
         &mut self,
         mut cx: crate::Context,
-        res: &mut ffi::SearchResult,
+        res: &mut SearchResult<'_>,
```
kei-nan, line 25: what does this lifetime marking mean?

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -474,7 +475,7 @@ pub(crate) mod test {
 
             let rp = unsafe { chain.last_raw() };
             let found =
-                unsafe { (rp.as_mut().next.unwrap())(rp.as_ptr(), &mut default_search_result()) };
+                unsafe { (rp.as_mut().next.unwrap())(rp.as_ptr(), &mut SearchResult::new()) };
```
kei-nan, line 478: this pattern seems to repeat.
maybe this should move to a function?

src/redisearch_rs/result_processor/src/lib.rs
```diff
@@ -599,12 +598,8 @@ pub(crate) mod test {
...
             const TYPE: ffi::ResultProcessorType = ffi::ResultProcessorType_RP_MAX;
 
-            fn next(
-                &mut self,
-                _cx: Context,
-                res: &mut ffi::SearchResult,
-            ) -> Result<Option<()>, Error> {
-                res.score = 42.0;
+            fn next(&mut self, _cx: Context, res: &mut SearchResult) -> Result<Option<()>, Error> {
+                res.set_score(42.0);
```
kei-nan, line 602: this is production code?
why do we set score to be 42?

src/result_processor.c
```diff
@@ -449,6 +452,7 @@ static int rpsortNext_innerLoop(ResultProcessor *rp, SearchResult *r) {
     }
     // we need to allocate a new result for the next iteration
     self->pooledResult = rm_calloc(1, sizeof(*self->pooledResult));
+    *self->pooledResult = SearchResult_New();
```
kei-nan, line 455: is there a heap allocation for a rust search result and not stack one?
do we really benefit from these two construction steps?

src/result_processor.c
```diff
@@ -546,6 +550,7 @@ ResultProcessor *RPSorter_NewByFields(size_t maxresults, const RLookupKey **keys
 
   ret->pq = mmh_init_with_size(maxresults, ret->cmp, ret->cmpCtx, srDtor);
   ret->pooledResult = rm_calloc(1, sizeof(*ret->pooledResult));
+  *ret->pooledResult = SearchResult_New();
```
kei-nan, line 553: same here regarding two construction steps.

src/result_processor.c
```diff
@@ -793,6 +798,7 @@ static SearchResult *GetResultsBlock(RPSafeLoader *self, size_t idx) {
   // If the block is not allocated, allocate it
   if (!*ret) {
     *ret = array_new(SearchResult, DEFAULT_BUFFER_BLOCK_SIZE);
+    array_foreach(*ret, res, res = SearchResult_New());
```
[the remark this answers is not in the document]
kei-nan: @JonasKruckenberg if array length is indeed 0 then the foreach statement is not needed.

src/result_processor.c
```diff
@@ -1267,6 +1273,7 @@ static int RPMaxScoreNormalizerNext_innerLoop(ResultProcessor *rp, SearchResult
 
   // we need to allocate a new result for the next iteration
   self->pooledResult = rm_calloc(1, sizeof(*self->pooledResult));
+  *self->pooledResult = SearchResult_New();
```
kei-nan, line 1274: same here regarding the heap allocation
there are additional placed in the file.

src/redisearch_rs/c_entrypoint/search_result_ffi/src/lib.rs
JonasKruckenberg: `const` functions in Rust are functions that may be used in initializers items whose initializers are evaluated at compile time ([constants](https://doc.rust-lang.org/std/keyword.const.html) and [statics](https://doc.rust-lang.org/std/keyword.static.html)). 

The reason why I had to remove the `const`-ness from this function is that `SearchResult` actually consists of this type `SearchResult(Canary<SearchResultInner>)` where `Canary` is a special type we created that will check for incorrectly initialized values on every access. It does so by providing a custom implementation of the `Deref` trait that first checks a magic value to determine initialized-ness and then provides a reference to the `SearchResultInner`. Since traits in Rust currently cannot be marked as `const` we also cannot mark a function accessing the inner value of `SearchResult(Canary<SearchResultInner>)` as const, which means we cannot mark _this_ function as const.

Note though that `const` has to implications on inlining or other optimizations. Theses still happen as usual regardless of `const`-ness.

src/result_processor.c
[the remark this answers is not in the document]
JonasKruckenberg: yeah fair enough 👍🏻 double checked and it seems tests are fine without too

src/redisearch_rs/result_processor/src/counter.rs
JonasKruckenberg: `'_` is the special syntax for an _elided_ lifetime. We're basically telling the compiler "you infer the correct lifetime here" (just like type inference). When using `'_` in an argument as done here the compiler has to be maximally conservative and basically choose a concrete lifetime that is only valid for the scope of the `next` function it is an argument to. It's a good idea in situations like this (callbacks and traits etc) to only give downstream consumers the smallest lifetime possible (so they can't do any funny business with it). Most importantly this restricted lifetime means impls cant return data out of the scope of this method.

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: this is all test code (tests start line 431)

src/redisearch_rs/result_processor/src/lib.rs
JonasKruckenberg: we could yes, but since its tests anyways I didn't consider it super necessary (I'd rather have people better understand what is going on)

src/result_processor.c
JonasKruckenberg: `SearchResult_New` _does not_ heap allocate on its own. It returns a `SearchResult` which is a sized-opaque type. We therefore need to heap allocate it manually if needed.

src/result_processor.c
JonasKruckenberg: See above.

src/result_processor.c
JonasKruckenberg: See above.

src/redisearch_rs/result_processor/src/counter.rs
kei-nan: and if we have a recursive chain of next calls will the topmost caller still be able to use the search result?
does the lifetime of the first caller trickles down to the last caller?

JonasKruckenberg: yes because the topmost caller (inside aggregate) has allocated the search result (this just ensures any RPs in that chain cant hold references to the data that outlives their local scope)

PR #6812, opened 2025-09-10
fix: Rust build-utils

JonasKruckenberg:
This change updates the `build-utils` crate to the newer Rust/Cargo version that uses `cargo::` instead of `cargo:`.

This change also adds a `println!("cargo::rerun-if-changed=cbindgen.toml");` to each `build.rs` file to correctly regenerate the header files when the cbindgen config changes.

### DIFF
--- a/src/redisearch_rs/build_utils/src/lib.rs
+++ b/src/redisearch_rs/build_utils/src/lib.rs
@@ -38,7 +38,7 @@ fn rerun_if_changes(dir: &Path, extensions: &[&str]) -> std::io::Result<()> {
         } else if let Some(extension) = path.extension().and_then(|ext| ext.to_str())
             && extensions.contains(&extension)
         {
-            println!("cargo:rerun-if-changed={}", path.display());
+            println!("cargo::rerun-if-changed={}", path.display());
         }
     }
     Ok(())
@@ -60,11 +60,15 @@ fn rerun_if_rust_changes(dir: &Path) -> std::io::Result<()> {
     rerun_if_changes(dir, &["rs"])
 }
 
-/// Emit granular `rerun-if-changed` statements for each files which are
-/// part of the crates listed in `config.parse.include`.
-/// This ensures cbindgen output is regenerated when the underlying Rust code
-/// is updated.
-pub fn rerun_cbinden(config: &cbindgen::Config) -> Result<(), Box<dyn std::error::Error>> {
+/// Generate a C header file via `cbindgen` for the calling crate.
+/// It'll read `cbindgen` configuration from the `cbindgen.toml` file at the crate root
+/// and output the header file to `header_path`.
+pub fn run_cbinden(header_path: impl AsRef<Path>) -> Result<(), Box<dyn std::error::Error>> {
+    let config =
+        cbindgen::Config::from_file("cbindgen.toml").expect("Failed to find cbindgen config");
+    println!("cargo::rerun-if-changed=cbindgen.toml");
+
+    // emit `rerun-if-changed` for all the headers files referenced by the config as well
     if let Some(include) = &config.parse.include {
         for included_crate in include.iter() {
             let path = git_root()?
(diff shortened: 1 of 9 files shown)


src/redisearch_rs/build_utils/src/lib.rs
```diff
@@ -38,7 +38,7 @@ fn rerun_if_changes(dir: &Path, extensions: &[&str]) -> std::io::Result<()> {
         } else if let Some(extension) = path.extension().and_then(|ext| ext.to_str())
             && extensions.contains(&extension)
         {
-            println!("cargo:rerun-if-changed={}", path.display());
+            println!("cargo::rerun-if-changed={}", path.display());
```
gdesmott, line 41: Looks like this syntax has been introduced in `1.77`. Should we define it as MSRV? We do not have `rust-version` yet in the workspace `Cargo.toml`.

src/redisearch_rs/c_entrypoint/inverted_index_ffi/build.rs
```diff
@@ -17,6 +17,7 @@ fn main() {
 
     let config = Config::from_file("cbindgen.toml").expect("Failed to find cbindgen config");
     let _ = rerun_cbinden(&config);
+    println!("cargo::rerun-if-changed=cbindgen.toml");
```
gdesmott: Would it make sense to add this to `rerun_cbinden()` ?

src/redisearch_rs/build_utils/src/lib.rs
LukeMathWalker: We're not a library, so the effective MSRV is what we use in our `rust-toolchain.toml` file.

src/redisearch_rs/c_entrypoint/inverted_index_ffi/build.rs
JonasKruckenberg: Hmmm, at first I thought no because you can't access the config file path from the `&config` param, but I guess there is no harm in passing it as a string 🤔

gdesmott: Can't we just hardcode it as `cbindgen.toml`?  This file has always this name doesn't it?

JonasKruckenberg: @gdesmott I updated the change quite a bit, realizing that all out cbindgen invocations are actually the same. Its quite a bit simpler now, keeping the actually important paths explicit the in the build scripts.

src/redisearch_rs/build_utils/src/lib.rs
```diff
@@ -77,6 +83,14 @@ pub fn rerun_cbinden(config: &cbindgen::Config) -> Result<(), Box<dyn std::error
         }
     }
 
+    let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+
+    cbindgen::Builder::new()
+        .with_crate(crate_dir)
+        .with_config(config)
+        .generate()?
+        .write_to_file(header_path);
```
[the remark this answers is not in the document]
JonasKruckenberg: this is not true, `write_to_file` only returns wether a file was updated or not, failure is propagated through panics.

src/redisearch_rs/build_utils/src/lib.rs
```diff
@@ -60,11 +60,18 @@ fn rerun_if_rust_changes(dir: &Path) -> std::io::Result<()> {
     rerun_if_changes(dir, &["rs"])
 }
 
-/// Emit granular `rerun-if-changed` statements for each files which are
-/// part of the crates listed in `config.parse.include`.
-/// This ensures cbindgen output is regenerated when the underlying Rust code
-/// is updated.
-pub fn rerun_cbinden(config: &cbindgen::Config) -> Result<(), Box<dyn std::error::Error>> {
+/// Run cbindgen using for the calling crate. Will read from the provided config file and output to the
+/// provided output path.
```
LukeMathWalker:
```suggestion
/// Generate a C header file via `cbindgen` for the calling crate.
/// It'll read `cbindgen` configuration from the provided `config_path` and output the header file to `header_path`.
```

src/redisearch_rs/build_utils/src/lib.rs
```diff
@@ -60,11 +60,18 @@ fn rerun_if_rust_changes(dir: &Path) -> std::io::Result<()> {
...
+pub fn run_cbinden(
+    config_path: impl AsRef<Path>,
```
LukeMathWalker: I think it'd be OK to canonically expect the file to be located at `cbindgen.toml` in the current dir.

JonasKruckenberg: fair enough 👍🏻

LukeMathWalker (approved):

PR #6817, opened 2025-09-11
MOD-11190: Add IndexSpec RDB Comparison Tests - Fix Index RDB field initialization

kei-nan:
We noticed some fields do not get initialized correctly when the index spec is initialized from an RDB.
Added mock rdb support for tests and added tests to compare an index to itself after it stored to an RDB and loaded again.

1. Current: Missing index spec struct members initialization
2. Change: Create one function to initialize all regular members and call it from both the new function and when loading from an RDB.
3. Outcome: More correct index initialization when loading an index from an RDB.

1. Index Spec

#### Mark if applicable

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -2057,6 +2057,9 @@ static void initializeIndexSpec(IndexSpec *sp, const HiddenString *name) {
   memset(&sp->stats, 0, sizeof(sp->stats));
   sp->stats.indexError = IndexError_Init();
 
+  sp->fieldIdToIndex = array_new(t_fieldIndex, 0);
+  sp->terms = NewTrie(NULL, Trie_Sort_Lex);
+
   IndexSpec_InitLock(sp);
 }
 
@@ -2065,8 +2068,6 @@ IndexSpec *NewIndexSpec(const HiddenString *name) {
   initializeIndexSpec(sp, name);
   sp->fields = rm_calloc(sizeof(FieldSpec), SPEC_MAX_FIELDS);
   sp->stopwords = DefaultStopWordList();
-  sp->terms = NewTrie(NULL, Trie_Sort_Lex);
-  sp->fieldIdToIndex = array_new(t_fieldIndex, 0);
   return sp;
 }
 
@@ -2919,18 +2920,13 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
   IndexSpec_InitLock(sp);
   StrongRef spec_ref = StrongRef_New(sp, (RefManager_Free)IndexSpec_Free);
   sp->own_ref = spec_ref;
+  initializeIndexSpec(sp, specName);
 
-  // `indexError` must be initialized before attempting to free the spec
-  sp->stats.indexError = IndexError_Init();
+  // Note: indexError, fieldIdToIndex, docs, specName, obfuscatedName, terms, and monitor flags are already initialized in initializeIndexSpec
 
   IndexSpec_MakeKeyless(sp);
-  sp->fieldIdToIndex = array_new(t_fieldIndex, 0);
-  sp->docs = DocTable_New(INITIAL_DOC_TABLE_SIZE);
-  sp->specName = specName;
-  sp->obfuscatedName = IndexSpec_FormatObfuscatedName(sp->specName);
   sp->flags = (IndexFlags)LoadUnsigned_IOError(rdb, goto cleanup);
-  sp->monitorDocumentExpiration = true;
-  sp->monitorFieldExpiration = RedisModule_HashFieldMinExpire != NULL;
+  // Note: monitorDocumentExpiration and monitorFieldExpiration are already set in initializeIndexSpec
   if (encver < INDEX_MIN_NOFREQ_VERSION) {
     sp->flags |= Index_StoreFreqs;
   }
@@ -2972,14 +2968,6 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
     goto cleanup;
   }
 
-  sp->terms = NewTrie(NULL, Trie_Sort_Lex);
-  /* For version 3 or up - load the generic trie */
-  //  if (encver >= 3) {
-  //    sp->terms = TrieType_GenericLoad(rdb, 0);
-  //  } else {
-  //    sp->terms = NewTrie(NULL);
-  //  }
-
   if (sp->flags & Index_HasCustomStopwords) {
     sp->stopwords = StopWordList_RdbLoad(rdb, encver);
     if (sp->stopwords == NULL)
@@ -3009,7 +2997,6 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
     }
   }
 
-  sp->scan_in_progress = false;
   return sp;
 
 cleanup:
--- a/src/spec.c
+++ b/src/spec.c
@@ -2034,6 +2034,14 @@ static void IndexSpec_InitLock(IndexSpec *sp) {
   pthread_rwlock_init(&sp->rwlock, &attr);
 }
 
+// Helper function for initializing a field spec
+static void initializeFieldSpec(FieldSpec *fs, t_fieldIndex index) {
+  fs->index = index;
+  fs->indexError = IndexError_Init();
+}
+
+// Helper function for initializing an index spec
+// Solves issues where a field is initialized in index creation but not when loading from RDB
 static void initializeIndexSpec(IndexSpec *sp, const HiddenString *name, IndexFlags flags,
                                 int16_t numFields) {
   sp->flags = flags;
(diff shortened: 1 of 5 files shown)


src/spec.c
```diff
@@ -2887,24 +2920,13 @@ int IndexSpec_CreateFromRdb(RedisModuleCtx *ctx, RedisModuleIO *rdb, int encver,
   IndexSpec_InitLock(sp);
```
JoanFM: remove this InitLock here if this is part of general initialization

JoanFM (approved):

tests/cpptests/test_cpp_rdb.cpp
```diff
@@ -0,0 +1,223 @@
...
+    EXPECT_EQ(spec->timerId, loadedSpec->timerId);
+    EXPECT_EQ(spec->monitorDocumentExpiration, loadedSpec->monitorDocumentExpiration);
+    EXPECT_EQ(spec->monitorFieldExpiration, loadedSpec->monitorFieldExpiration);
+    EXPECT_EQ(spec->isDuplicate, loadedSpec->isDuplicate);
+    EXPECT_EQ(spec->scan_in_progress, loadedSpec->scan_in_progress);
+    EXPECT_EQ(spec->scan_failed_OOM, loadedSpec->scan_failed_OOM);
+    EXPECT_EQ(spec->used_dialects, loadedSpec->used_dialects);
+    EXPECT_EQ(spec->counter, loadedSpec->counter);
+    EXPECT_EQ(spec->activeCursors, loadedSpec->activeCursors);
+    int sameInitialLockState = memcmp((const void*)&spec->rwlock, (const void *)&loadedSpec->rwlock, sizeof(pthread_rwlock_t));
```
GuyAv46: `sameInitialLockState` not used

tests/cpptests/test_cpp_rdb.cpp
```diff
@@ -0,0 +1,223 @@
...
+    EXPECT_EQ(0, lockResult);
+    if (lockResult == 0) {
+        pthread_rwlock_unlock(&loadedSpec->rwlock);
+    }
+
+    // Verify field specifications are preserved
+    for (int i = 0; i < loadedSpec->numFields; i++) {
+        FieldSpec *field = &spec->fields[i];
+        FieldSpec *loadedField = &loadedSpec->fields[i];
+        EXPECT_TRUE(loadedField->types != 0);
```
GuyAv46: nitpick - Consider replacing boolean checks with comparisons, so we know what value we get if something fails
```suggestion
        EXPECT_NE(loadedField->types, 0);
```

tests/cpptests/test_cpp_rdb.cpp
```diff
@@ -0,0 +1,223 @@
...
+    if (lock_result == 0) {
+        pthread_rwlock_unlock(&spec->rwlock);
+        return true;  // Lock is properly initialized and unlocked
+    }
+    return false;  // Lock failed - either not initialized or locked
+}
+
+// Second function - IndexSpec RDB serialization test
+TEST_F(RdbMockTest, testIndexSpecRdbSerialization) {
+    printf("Starting testIndexSpecRdbSerialization\n");
```
GuyAv46: Let's avoid littering with `printf` in tests

tests/cpptests/test_cpp_rdb.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+    // Create an IndexSpec
+    const char *args[] = {"SCHEMA", "title", "TEXT", "WEIGHT", "2.0", "body", "TEXT", "price", "NUMERIC"};
+    QueryError err = {QUERY_OK};
+
+    printf("Calling IndexSpec_ParseC\n");
+    StrongRef original_spec_ref = IndexSpec_ParseC("test_rdb_idx", args, sizeof(args) / sizeof(const char *), &err);
+    printf("IndexSpec_ParseC returned, checking for errors\n");
+    ASSERT_FALSE(QueryError_HasError(&err)) << QueryError_GetUserError(&err);
```
GuyAv46: Let's avoid `printf` in tests. Use the `<<` operator to add log messages if some assertion fails
```suggestion
    StrongRef original_spec_ref = IndexSpec_ParseC("test_rdb_idx", args, sizeof(args) / sizeof(const char *), &err);
    ASSERT_FALSE(QueryError_HasError(&err)) << QueryError_GetUserError(&err);
```

src/spec.c
```diff
@@ -2064,8 +2057,17 @@ IndexSpec *NewIndexSpec(const HiddenString *name) {
...
+  sp->fieldIdToIndex = array_new(t_fieldIndex, 0);
+  sp->terms = NewTrie(NULL, Trie_Sort_Lex);
+
   IndexSpec_InitLock(sp);
+}
 
+IndexSpec *NewIndexSpec(const HiddenString *name) {
+  IndexSpec *sp = rm_calloc(1, sizeof(IndexSpec));
+  initializeIndexSpec(sp, name);
+  sp->fields = rm_calloc(sizeof(FieldSpec), SPEC_MAX_FIELDS);
```
GuyAv46: Any reason not to do this in `initializeIndexSpec` as well, plus the field-spec indexing error initialization? We had a handful of bugs caused by late initializations of the error structs. This looks like a good opportunity to make it safer for the future.

kei-nan: rdb initializes the array differently.
It only creates an array based on the number of fields it needs.
```
sp->fields = rm_calloc(sp->numFields, sizeof(FieldSpec));
```
Which arguably conserves a bit in the memory.
Didn't want to do a lot of logical changes in the flow.

JoanFM: is sp->numFields initialized somehow? Can the num fields be an argument?

GuyAv46: I think we should make it an argument and pass the max if we don’t know the size yet. Also allocating the max size is probably a huge waste

JoanFM: I do not think is a huge waste, is something not allocated so many times. Trying to be smarter here would look overkilling?

kei-nan:
```
sp->numFields = LoadUnsigned_IOError(rdb, goto cleanup);
```
The problem with making it an argument it forces the initialize to be called after we load the num of fields.
And if someone needs a field we initialized before that is can cause issues.

GuyAv46: However, before the number of fields, we only have the spec name and flags, and we have already parsed the name before initialization...

kei-nan: I'll change the call sites and see how risky it looks I guess.

src/spec.c
```diff
@@ -2034,23 +2034,19 @@ static void IndexSpec_InitLock(IndexSpec *sp) {
...
-
-IndexSpec *NewIndexSpec(const HiddenString *name) {
-  IndexSpec *sp = rm_calloc(1, sizeof(IndexSpec));
-  sp->fields = rm_calloc(sizeof(FieldSpec), SPEC_MAX_FIELDS);
-  sp->flags = INDEX_DEFAULT_FLAGS;
+static void initializeIndexSpec(IndexSpec *sp, const HiddenString *name, IndexFlags flags,
+                                int16_t numFields) {
+  sp->flags = flags;
+  sp->numFields = numFields;
+  sp->fields = rm_calloc(sizeof(FieldSpec), numFields);
```
GuyAv46: Can we also add the index error initialization loop here?

kei-nan: When creating an index it would mean doing an init 1024 times.
Seems a bit excessive, no?

GuyAv46: Not if we attempt to free all of the 1024 later... how are we avoiding crashes there at the moment? We change the `sp->numFields` later on to the actual number?

GuyAv46: Allocating 1024 fields is excessive on its own; maybe we can avoid that completely?

kei-nan: I'll try passing 0 and see if the unit tests pass , don't want to cause an issue elsewhere in the code but I guess it would be better behaviour.
I think we do a realloc when creating the first field anyhow - so the duration in the code we would be without a field is small. 
Wondering if a schema with zero fields is even possible.

src/spec.c
```diff
@@ -2876,41 +2880,61 @@ void Indexes_ScanAndReindex() {
...
+  IndexFlags flags = (IndexFlags)LoadUnsigned_IOError(rdb, goto cleanup);
+  // Note: monitorDocumentExpiration and monitorFieldExpiration are already set in initializeIndexSpec
+  if (encver < INDEX_MIN_NOFREQ_VERSION) {
+    flags |= Index_StoreFreqs;
   }
+  int16_t numFields = LoadUnsigned_IOError(rdb, goto cleanup);
 
-  // `indexError` must be initialized before attempting to free the spec
-  sp->stats.indexError = IndexError_Init();
+  initializeIndexSpec(sp, specName, flags, numFields);
```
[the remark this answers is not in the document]
kei-nan: It should get freed since it is a ref counted object and `IndexSpec_FreeUnlinkedData` should get called.

JoanFM (approved):

PR #6819, opened 2025-09-11
[MOD-11137] Add module.conf file

alonre24:

Split Redis.full.conf to modules - include search configs
Copy from redis.full.conf, except for:
1. added the new search-enable-unstable-features, search-bm25std-tanh-factor and  search-indexer-yield-every-ops configs
2. Fix the default on-timeout to reflect the reality (from fail to return)

#### Mark if applicable

### DIFF
--- a/module.conf
+++ b/module.conf
@@ -1,0 +1,233 @@
+############################## QUERY ENGINE CONFIG ############################
+
+# Keep numeric ranges in numeric tree parent nodes of leafs for `x` generations.
+# numeric, valid range: [0, 2], default: 0
+#
+# search-_numeric-ranges-parents 0
+
+# The number of iterations to run while performing background indexing
+# before we call usleep(1) (sleep for 1 micro-second) and make sure that we
+# allow redis to process other commands.
+# numeric, valid range: [1, UINT32_MAX], default: 100
+#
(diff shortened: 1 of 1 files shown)


module.conf
```diff
@@ -0,0 +1,233 @@
...
+# any pending requests will be processed, even if the topology is not fully connected.
+# numeric, valid range: [0, LLONG_MAX], default: 30000
+#
+# search-topology-validation-timeout 30000
+
+
+# Enable unstable new features for early adopters. Use at your own risk.
+#
+# bool, default: no
+# search-enable-unstable-features false
```
oshadmi: Why not keep the `search-enable-unstable-features` out of the conf file? Not relevant for most users,
and if needed they can add `search-enable-unstable-features yes` or set it using `CONFIG SET`

alonre24: I don't have a strong opinion here; many of the configurations here are niche. I thought we would include all of them (except for ones that are pure debug configs)

oshadmi: I am inclining towards making this specific config less noticeable (excluded from conf file)
If you prefer to keep it, change the default from `false` to `no`

alonre24: ok removed

PR #6820, opened 2025-09-11
MOD-11048, MOD-11157: Add default normalizers to vector search

Itzikvaknin:
### Summary
Implements vector distance normalization for linear hybrid scoring.

### Changes
- **Added `PLN_T_VECTOR_NORMALIZER`**: aggplan step to mark when vector distance normalization is needed in the processing chain
- **Added `RPVectorNormalizer` result processor**: Applies metric-specific normalization formulas to normalise vector scores
- **Pipeline resolution logic**: `pipeline_construction.c` selects appropriate normalization function based on vector field's distance metric (L2, IP, Cosine)
- **Added `isInternal` flag to `MetricRequest`**: Indicates whether the metric should be returned as part of the response

### Normalization Formulas
Follows the [weighted and hybrid scoring design](https://redislabs.atlassian.net/wiki/spaces/DX/pages/3904012298/Weighted+and+Hybrid+Scoring+-+design#Establishing-vector-distance-as-document-score):
- **L2 Distance**: `1 / (1 + distance)`  
- **Inner Product**: `(1 + dot_product) / 2`
- **Cosine Distance**: `(1 + (1 - cosine_distance)) / 2`

### Flow
1. **Parse hybrid request**  
   Parser stores the vector field name and vector score key name in `PLN_T_VECTOR_NORMALIZER`.

2. **Pipeline construction**  
   Creates an `RPVectorNormalizer` with the right normalization function (based on metric) and the `RLookupKey` of the vector score key.

3. **Yielding results**  
   The processor fetches the raw score, normalizes it, and writes it back to the `RLookupRow` and `SearchResult.score`.

### DIFF
--- a/src/result_processor.c
+++ b/src/result_processor.c
@@ -1466,9 +1466,9 @@ static int RPVectorNormalizer_Next(ResultProcessor *rp, SearchResult *r) {
 
   // Apply normalization to the score
   double normalizedScore = 0.0;
-  double originalScore = 0.0;
   RSValue *scoreValue = RLookup_GetItem(self->scoreKey, &r->rowdata);
   if (scoreValue) {
+    double originalScore = 0.0;
     if (RSValue_ToNumber(scoreValue, &originalScore)) {
       normalizedScore = self->normFunc(originalScore);
     }


src/query_node.h
```diff
@@ -158,6 +158,7 @@ typedef enum {
   QueryNode_IsTag = 0x10,
   // Marks this as the main vector node in a hybrid vector subquery
   QueryNode_HybridVectorSubqueryNode = 0x20,
+  QueryNode_HideDistanceField = 0x40,
```
alonre24:
```suggestion
  QueryNode_HideVectorDistanceField = 0x40,
```

src/hybrid/vector_query_utils.h
```diff
@@ -25,10 +25,12 @@ extern "C" {
...
-  const char *fieldName;          // Field name for later resolution (NOT owned - points to args)
-  QueryAttribute *attributes;     // Non-vector-specific attributes like YIELD_DISTANCE_AS (OWNED)
-  bool isParameter;               // true if vector data is a parameter
-  bool hasExplicitK;              // Flag to track if K was explicitly set in KNN query
+  const char *fieldName;           // Field name for later resolution (NOT owned - points to args)
+  QueryAttribute *attributes;      // Non-vector-specific attributes like YIELD_DISTANCE_AS (OWNED)
+  bool isParameter;                // true if vector data is a parameter
+  bool hasExplicitK;               // Flag to track if K was explicitly set in KNN query
+  bool hasExplicitYieldDistanceAs; // Flag to track if YIELD_DISTANCE_AS was explicitly set
+  char *distanceFieldAlias;        //
```
alonre24: missing description

src/aggregate/aggregate_plan.c
```diff
@@ -132,6 +134,26 @@ void loadDtor(PLN_BaseStep *bstp) {
   rm_free(lstp);
 }
 
+static void vectorNormalizerDtor(PLN_BaseStep *bstp) {
+  PLN_VectorNormalizerStep *vnStep = (PLN_VectorNormalizerStep *)bstp;
+  // vectorFieldName is not owned (points to parser tokens), so don't free it
+  if (vnStep->distanceFieldAlias) {
```
alonre24, line 141: in aggregate_plan.h It says that `distanceFieldAlias` is not owned as well

src/vector_index.h
```diff
@@ -131,6 +134,43 @@ const char *VecSimType_ToString(VecSimType type);
...
+
+  VecSimAlgo field_algo = vec_params.algo;
+  AlgoParams algo_params = vec_params.algoParams;
+
+  if (field_algo == VecSimAlgo_TIERED) {
+    VecSimParams *primary_params = algo_params.tieredParams.primaryIndexParams;
+    if (primary_params->algo == VecSimAlgo_HNSWLIB) {
+      HNSWParams hnsw_params = primary_params->algoParams.hnswParams;
+      return hnsw_params.metric;
+    } else if (primary_params->algo == VecSimAlgo_BF) {
```
alonre24: `VecSimAlgo_BF` cannot be the primary index for `VecSimAlgo_TIERED`

src/vector_index.h
```diff
@@ -131,6 +134,43 @@ const char *VecSimType_ToString(VecSimType type);
...
+      BFParams bf_params = primary_params->algoParams.bfParams;
+      return bf_params.metric;
+    } else {
```
alonre24: The case of `svsParams` is missing under tiered

src/hybrid/parse_hybrid.c
```diff
@@ -161,16 +158,16 @@ static int parseKNNClause(ArgsCursor *ac, VectorQuery *vq, ParsedVectorData *pvd
         .vallen = strlen(value)
       };
       pvd->attributes = array_ensure_append_1(pvd->attributes, attr);
-      hasYieldDistanceAs = true;
-
+      pvd->hasExplicitYieldDistanceAs = true;
+      pvd->distanceFieldAlias = rm_strdup(value);
```
alonre24: Why is duplication required here?

tests/cpptests/test_cpp_parsed_hybrid_pipeline.cpp
```diff
@@ -466,15 +466,6 @@ TEST_F(HybridRequestParseTest, testKeyCorrespondenceBetweenSearchAndTailPipeline
       ASSERT_TRUE(tailKey != NULL)
         << "Key '" << upstreamKey->name << "' from upstream request " << reqIdx << " not found in tail pipeline";
 
-      // Verify that the keys point to the same indices
-      EXPECT_EQ(upstreamKey->dstidx, tailKey->dstidx)
```
alonre24, line 470: Why removing these checks? Should they be replaced rather than removed?

tests/pytests/test_hybrid_vector_normalizer.py
```diff
@@ -0,0 +1,110 @@
...
+    def VectorNorm_IP(dot_product):
+        return (1.0 + dot_product) / 2.0
+
+    vec1 = np.frombuffer(vec1_bytes, dtype=np.float32)
+    vec2 = np.frombuffer(vec2_bytes, dtype=np.float32)
+    # IP distance is 1 - dot_product
+    return VectorNorm_IP(1.0 - np.dot(vec1, vec2))
+
+# Distance calculation function mapping
+DISTANCE_CALCULATORS = {
```
alonre24: Let's have this sanity test also over all vector indexes (flat, HNSW, SVS-VAMANA) and all supported data types (note that SVS-VAMANA only supported over FP32 and FP16).
Also, consider testing over a query that has both range and KNN query (if that is supported today in ft.hybrid), and test over JSON with multi-value as well

tests/pytests/test_rrf.py
```diff
@@ -41,7 +41,7 @@ def run_test_scenario(env, no_tag_search_query, with_tag_search_query):
     for key in shared_keys:
         score_no_tag = float(results_no_tag[key][SCORE_FIELD])
         score_with_tag = float(results_with_tag[key][SCORE_FIELD])
-        env.assertAlmostEqual(score_no_tag, score_no_tag, delta=1E-6)
+        env.assertAlmostEqual(score_with_tag, score_no_tag, delta=1E-6)
```
alonre24: Is the change related to the PR?

src/aggregate/aggregate_plan.c
Itzikvaknin: Ownership transferred from pvd to this step, documented in both as owned, protected freeing by null checking

tests/pytests/test_rrf.py
Itzikvaknin: No

tests/cpptests/test_cpp_parsed_hybrid_pipeline.cpp
Itzikvaknin: The previous approach assumed that all lookups (search, VSIM, and tail) shared the same keys mapped to the same indices
Now, we rebuild the RLookupRow during each merge, so this check is no longer needed

src/hybrid/parse_hybrid.c
Itzikvaknin: Removed `pvd->hasExplicitYieldDistanceAs `

src/vector_index.h
Itzikvaknin: Since SVS was added after branching from master, we cannot test it at this time.
Ticket [MOD-11352](https://redislabs.atlassian.net/browse/MOD-11352) was created for SVS support and tests, but it is blocked until the feature branch is merged into master.

tests/cpptests/test_cpp_rlookup.cpp
```diff
@@ -118,7 +118,7 @@ void verify_values_by_names(RLookup* lookup, RLookupRow* row,
                            const std::vector<double>& expectedValues) {
   EXPECT_EQ(fieldNames.size(), expectedValues.size()) << "Field count must match expected value count";
 
-  for (size_t i = 0; i < fieldNames.size() && i < expectedValues.size(); i++) {
+  for (size_t i = 0; i < fieldNames.size(); i++) {
```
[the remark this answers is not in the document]
Itzikvaknin: ```EXPECT_EQ(fieldNames.size(), expectedValues.size()) ```

tests/pytests/requirements.txt
```diff
@@ -9,3 +9,4 @@ scipy <= 1.15.2
 faker <= 37.1.0
 distro <= 1.9.0
 orderly-set <= 5.3.0 # Update pin once Python 3.8 is not used in CI
+ml_dtypes <= 0.5.3
```
Itzikvaknin: Looks like a good option for extending NumPy datatypes https://snyk.io/advisor/python/ml-dtypes

tests/pytests/test_hybrid_vector_normalizer.py
Itzikvaknin: Queries combining both Range and KNN are not supported.
We’ll test JSON separately, not as part of this PR, since the plan is to parameterize all tests to cover both setups.

Also, I’ve expanded test_hybrid_vector_normalizer to the following matrix:
```
data_types = ['FLOAT32', 'FLOAT64', 'FLOAT16', 'BFLOAT16']
metrics = ['L2', 'COSINE', 'IP']
algorithms = ['FLAT', 'HNSW']
```

src/vector_index.c
```diff
@@ -574,3 +574,31 @@ int VecSim_CallTieredIndexesGC(WeakRef spRef) {
...
+
+  switch (field_algo) {
+    case VecSimAlgo_TIERED: {
+      VecSimParams *primary_params = algo_params.tieredParams.primaryIndexParams;
+      if (primary_params->algo == VecSimAlgo_HNSWLIB) {
+        HNSWParams hnsw_params = primary_params->algoParams.hnswParams;
+        return hnsw_params.metric;
+      } else {
+        // TODO: Add SVS support here after FT.HYBRID is merged to master
+        // Unknown primary algorithm in tiered index
```
alonre24, line 593: I would strongly recommend updating the feature branch now and add the support to SVS now rather than later

tests/pytests/test_hybrid_vector_normalizer.py
```diff
@@ -52,59 +38,96 @@ def calculate_l2_distance_normalized(vec1_bytes, vec2_bytes):
...
+
 # TODO: remove once FT.HYBRID for cluster is implemented
 @skip(cluster=True)
 def test_hybrid_vector_normalizer():
     """Test that yield_distance_as value equals the expected distance for all distance metrics"""
 
+    # Test all supported vector types
+    data_types = ['FLOAT32', 'FLOAT64', 'FLOAT16', 'BFLOAT16']
+    metrics = ['L2', 'COSINE', 'IP']
+    algorithms = ['FLAT', 'HNSW'] # TODO: Add SVS support here after FT.HYBRID is merged to master
```
alonre24, line 109: Same as above - bettter to have it now

tests/pytests/test_hybrid_vector_normalizer.py
alonre24: and what about INT8 ?

Itzikvaknin: Added `['INT8', 'UINT8']` in https://github.com/RediSearch/RediSearch/pull/6820/commits/9139d4893968af5e2993bfc1a9c808b77c3dd976

src/pipeline/pipeline_construction.c
```diff
@@ -533,6 +538,34 @@ int Pipeline_BuildAggregationPart(Pipeline *pipeline, const AggregationPipelineP
...
+        PLN_VectorNormalizerStep *vnStep = (PLN_VectorNormalizerStep *)stp;
+
+        // Resolve vector field to get distance metric
+        const FieldSpec *vectorField = IndexSpec_GetFieldWithLength(params->common.sctx->spec,
+                                                                     vnStep->vectorFieldName,
+                                                                     strlen(vnStep->vectorFieldName));
+        if (!vectorField || !FIELD_IS(vectorField, INDEXFLD_T_VECTOR)) {
+          QueryError_SetError(status, QUERY_ESYNTAX, "Invalid vector field for normalization");
+          goto error;
+        }
```
alonre24, line 552: Do we have coverage for this case?

tests/pytests/test_hybrid_vector_normalizer.py
```diff
@@ -0,0 +1,133 @@
...
+        for algorithm in algorithms:
+            for data_type in data_types:
+                env = Env()
+                test_data = setup_index(env, metric, algorithm, data_type)
+                query_vector = np.array([0.5, 0.5], dtype=data_type.lower()).tobytes()
+
+                for vector_query in [['KNN', '4', 'K', '10'], ['RANGE', '4', 'RADIUS', '10']]:
+                    response = env.cmd('FT.HYBRID', 'idx', 'SEARCH', 'green', 'VSIM', '@embedding', query_vector,
+                                        *vector_query, 'YIELD_DISTANCE_AS', 'vector_distance')
+                    results = get_results_from_hybrid_response(response)
```
alonre24: If I understand correctly, we are getting the normalized vector score under the `vector_distance` field due to passing the `YIELD_DISTANCE_AS` argument. The notion of "distance" here is bad in my opinion, since the distance was converted into a score. Ideally, we would have the two options of `YIELD_DISTANCE_AS` and `YIELD_SCORE_AS`. However, if exposing both metrics is not in the scope of this PR, I would recommend just changing "distance" to "score".

src/hybrid/parse_hybrid.c
```diff
@@ -642,6 +642,17 @@ int parseHybridCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc,
...
   }
 
+  const char *distanceFieldAlias = NULL;
+  if (vectorRequest->parsedVectorData->distanceFieldAlias != NULL) {
+    distanceFieldAlias = vectorRequest->parsedVectorData->distanceFieldAlias;
+    vectorRequest->parsedVectorData->distanceFieldAlias = NULL;
+  } else {
+    distanceFieldAlias = VectorQuery_GetDefaultScoreFieldName(vectorRequest->parsedVectorData->fieldName, strlen(vectorRequest->parsedVectorData->fieldName));
+    vectorRequest->parsedVectorData->queryNodeFlags |= QueryNode_HideVectorDistanceField;
+  }
```
alonre24: Consider adding some explanations here - non-so-trivial code that includes conditional moving ownership...

src/pipeline/pipeline_construction.c
Itzikvaknin: This is called in HybridRequest_BuildPipeline → ... → AREQ_BuildPipeline → Pipeline_BuildAggregationPart
In parseHybridCommand, we call AREQ_ApplyContext → applyVectorQuery, which performs the following check:

```
  // Resolve field spec
  const FieldSpec *vectorField = IndexSpec_GetFieldWithLength(sctx->spec, fieldName, strlen(fieldName));
  if (!vectorField) {
    QueryError_SetWithUserDataFmt(status, QUERY_ESYNTAX, "Unknown field", " `%s`", fieldName);
    return REDISMODULE_ERR;
  } else if (!FIELD_IS(vectorField, INDEXFLD_T_VECTOR)) {
    QueryError_SetWithUserDataFmt(status, QUERY_ESYNTAX, "Expected a " SPEC_VECTOR_STR " field", " `%s`", fieldName);
    return REDISMODULE_ERR;
  }
  ```
  
Since `parseHybridCommand` is called before `HybridRequest_BuildPipeline`, we currently have no way to test this check.

I think it’s good to leave it there for future code

File: src/hybrid/parse_hybrid.c (hunk 16 above)
Comment (Itzikvaknin) [reply src/hybrid/parse_hybrid.c]: Added:
```
  // If YIELD_DISTANCE_AS was specified, use its string (pass ownership from pvd to vnStep),
  // otherwise, store the distance in a default key.
  
  ...
  
  // Store the key string so it could fetch the distance from the RlookupRow
    
 ```

tests/pytests/test_hybrid_vector_normalizer.py
Itzikvaknin: Renamed `YIELD_DISTANCE_AS` to `YIELD_SCORE_AS`
In the future we will have both, and `YIELD_SCORE_AS` would be used for calculations

alonre24 (dismissed):

src/hybrid/parse_hybrid.c
```diff
@@ -142,14 +140,14 @@ static int parseKNNClause(ArgsCursor *ac, VectorQuery *vq, ParsedVectorData *pvd
...
-        QueryError_SetError(status, QUERY_EDUPPARAM, "Duplicate YIELD_DISTANCE_AS parameter");
+    } else if (AC_AdvanceIfMatch(ac, "YIELD_SCORE_AS")) {
+      if (pvd->vectorScoreFieldAlias != NULL) {
+        QueryError_SetError(status, QUERY_EDUPPARAM, "Duplicate YIELD_SCORE_AS parameter");
         return REDISMODULE_ERR;
       }
       const char *value;
       if (AC_GetString(ac, &value, NULL, 0) != AC_OK) {
-        QueryError_SetError(status, QUERY_ESYNTAX, "Missing distance field name");
+        QueryError_SetError(status, QUERY_ESYNTAX, "Missing vector score field name");
```
oshadmi: Should use same message also in parseRangeClause

src/pipeline/pipeline_construction.c
```diff
@@ -533,6 +538,34 @@ int Pipeline_BuildAggregationPart(Pipeline *pipeline, const AggregationPipelineP
...
+
+        // Extract distance metric from vector field
+        VecSimMetric metric = GetVecSimMetricFromVectorField(vectorField);
```
oshadmi:
```suggestion
        VecSimMetric metric = getVecSimMetricFromVectorField(vectorField);
```

src/pipeline/pipeline_construction.c
```diff
@@ -533,6 +538,34 @@ int Pipeline_BuildAggregationPart(Pipeline *pipeline, const AggregationPipelineP
...
+
+        // Get appropriate normalization function
+        VectorNormFunction normFunc = GetVectorNormalizationFunction(metric);
+
+        // Get score key for writing normalized scores
+        RLookup *curLookup = AGPLN_GetLookup(pln, stp, AGPLN_GETLOOKUP_PREV);
```
oshadmi, line 561: Add assert for `curLookup`?

src/query.h
```diff
@@ -32,6 +32,7 @@ extern "C" {
 typedef struct MetricRequest{
   const char *metric_name;
   RLookupKey **key_ptr;
+  bool isInternal;
```
oshadmi: Add description

src/vector_index.c
```diff
@@ -566,3 +574,31 @@ int VecSim_CallTieredIndexesGC(WeakRef spRef) {
...
+        RS_ABORT("Unknown primary algorithm in tiered index");
```
oshadmi: Use `VecSimAlgo_ToString` instead of "tiered"

src/vector_index.c
```diff
@@ -566,3 +574,31 @@ int VecSim_CallTieredIndexesGC(WeakRef spRef) {
...
+      }
+      break;
+    }
+    case VecSimAlgo_BF:
+      return algo_params.bfParams.metric;
+    default:
+      // Unknown algorithm type
+      RS_ABORT("Unknown VecSimAlgo in GetVecSimMetricFromVectorField");
```
oshadmi:
```suggestion
      RS_ABORT("Unknown algorithm in vector index");
```

src/result_processor.c
```diff
@@ -1441,6 +1441,62 @@ static int RPMaxScoreNormalizer_Accum(ResultProcessor *rp, SearchResult *r) {
...
+  if (rc != RS_RESULT_OK) {
+    return rc;
+  }
+
+  // Apply normalization to the score
+  double originalScore = 0.0; // Default value
+  RSValue_ToNumber(RLookup_GetItem(self->scoreKey, &r->rowdata), &originalScore);
+  // originalScore will be 0.0 if conversion failed, or the actual value if successful
+  double normalizedScore = self->normFunc(originalScore);
+  r->score = normalizedScore;
```
oshadmi, line 1476: In case of missing score in row data - we should normalize to zero.

tests/cpptests/test_cpp_parse_hybrid.cpp
```diff
@@ -781,11 +785,10 @@ TEST_F(ParseHybridTest, testVsimKNNDuplicateEFRuntime) {
...
 
 
-// TODO: Enable this once we support YIELD_DISTANCE_AS (not part of phase 1)
-TEST_F(ParseHybridTest, testKNNDuplicateYieldDistanceAs) {
-  // Test KNN with duplicate YIELD_DISTANCE_AS parameters
-  RMCK::ArgvList args(ctx, "FT.HYBRID", index_name.c_str(), "SEARCH", "hello", "VSIM", "@vector", "$BLOB", "KNN", "6", "K", "10", "YIELD_DISTANCE_AS", "dist1", "YIELD_DISTANCE_AS", "dist2", "PARAMS", "2", "BLOB", TEST_BLOB_DATA);
-  testErrorCode(args, QUERY_EDUPPARAM, "Duplicate YIELD_DISTANCE_AS parameter");
+TEST_F(ParseHybridTest, testKNNDuplicateYieldScoreAs) {
+  // Test KNN with duplicate YIELD_SCORE_AS parameters
+  RMCK::ArgvList args(ctx, "FT.HYBRID", index_name.c_str(), "SEARCH", "hello", "VSIM", "@vector", "$BLOB", "KNN", "6", "K", "10", "YIELD_SCORE_AS", "dist1", "YIELD_SCORE_AS", "dist2", "PARAMS", "2", "BLOB", TEST_BLOB_DATA);
```
oshadmi, line 790: Is there a test for missing `YIELD_SCORE_AS` parameter? (for KNN and for RANGE)

tests/pytests/test_hybrid_vector_normalizer.py
oshadmi: The default normalizer is always applied in this PR, but it is mandatory only for LINEAR,
and not for RRF, where we can keep the distance as-is (and not necessarily make it a score)
As an optimization for later, we can normalize in RRF only if `YIELD_SCORE_AS`  is specified

tests/cpptests/test_cpp_parse_hybrid.cpp
Itzikvaknin: Since `YIELD_SCORE_AS` is optional, it cannot be considered missing.
We ensure that the number of parameters is even, so if this parameter is present, it will always have a value.
The error message has been updated to:
```
 QueryError_SetError(status, QUERY_EBADVAL, "Invalid vector score field name");
```

Itzikvaknin: Added in https://github.com/RediSearch/RediSearch/pull/6826

src/result_processor.c
```diff
@@ -1465,10 +1465,14 @@ static int RPVectorNormalizer_Next(ResultProcessor *rp, SearchResult *r) {
...
   // Apply normalization to the score
-  double originalScore = 0.0; // Default value
-  RSValue_ToNumber(RLookup_GetItem(self->scoreKey, &r->rowdata), &originalScore);
-  // originalScore will be 0.0 if conversion failed, or the actual value if successful
-  double normalizedScore = self->normFunc(originalScore);
+  double normalizedScore = 0.0;
+  double originalScore = 0.0;
+  RSValue *scoreValue = RLookup_GetItem(self->scoreKey, &r->rowdata);
+  if (scoreValue) {
+    if (RSValue_ToNumber(scoreValue, &originalScore)) {
```
Itzikvaknin, line 1472: Move `double originalScore = 0.0;` here

PR #6827, opened 2025-09-15
MOD 11233: Fix Incorrect Words Getting Highlighted

kei-nan:
```
FT.CREATE product_index ON HASH MAXTEXTFIELDS STOPWORDS 0 SCHEMA id NUMERIC product_id NUMERIC sku TEXT NOSTEM category TEXT NOSTEM brand TEXT NOSTEM description TEXT NOSTEM price NUMERIC SORTABLE created_date NUMERIC SORTABLE status TEXT NOSTEM created_by TEXT NOSTEM modified_by TEXT NOSTEM tags TEXT NOSTEM

HSET product:12345 id 12345 product_id 67890 sku PROD_12345 category Product brand GenericBrand description "This Product is available for purchase and comes with warranty coverage." price 99.99 created_date 1756447303000 status active created_by admin modified_by "" tags "electronics gadget"

FT.SEARCH product_index Product HIGHLIGHT 
```
Would result in Product not being correctly highlighted in the description field.

1. Current: `RSByteOffset_Iterate` doesn't account for field going out of bounds - if word in field doesn't exist in offsets then we will reach end of stream.
e.g we look for product in field status which has active value the token for "active" will be higher than token "product" in field description, so we will reach end of stream.
2. Change: What is the change - Check if we even have a match in the offset and return EOF if not, also more clear logic in `RSByteOffsetIterator_Next` function, add a few flow tests for the highlight feature.
3. Outcome: Better highlight behaviour.

1. Byte Offests Iterator

#### Mark if applicable

### DIFF
--- a/src/byte_offsets.c
+++ b/src/byte_offsets.c
@@ -105,10 +105,10 @@ int RSByteOffset_Iterate(const RSByteOffsets *offsets, uint32_t fieldId,
   iter->rdr = NewBufferReader(&iter->buf);
   iter->endPos = offField->lastTokPos;
 
-  iter->lastValue = 0;
-
-  for (iter->curPos = 1; iter->curPos < offField->firstTokPos && !BufferReader_AtEnd(&iter->rdr); ++iter->curPos) {
-    iter->lastValue = ReadVarint(&iter->rdr) + iter->lastValue;
+  for (iter->curPos = 0; iter->curPos < offField->firstTokPos && !BufferReader_AtEnd(&iter->rdr); ++iter->curPos) {
+    // simply override, we only care about the initial offset for this field
+    iter->lastValue = ReadVarint(&iter->rdr);
+    RedisModule_Log(NULL, "notice", "lastValue: %u, curPos: %u, firstTokPos: %u, lastTokPos: %u", iter->lastValue, iter->curPos, offField->firstTokPos, offField->lastTokPos);
   }
 
   // If we reached the end of the stream before we reached the first token position, return an error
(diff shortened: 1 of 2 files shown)


src/byte_offsets.c
```diff
@@ -103,25 +103,40 @@ int RSByteOffset_Iterate(const RSByteOffsets *offsets, uint32_t fieldId,
   iter->buf.data = (char *) offsets_data;
   iter->buf.offset = offsets_len;
   iter->rdr = NewBufferReader(&iter->buf);
-  iter->curPos = 1;
   iter->endPos = offField->lastTokPos;
 
   iter->lastValue = 0;
 
-  while (iter->curPos < offField->firstTokPos && !BufferReader_AtEnd(&iter->rdr)) {
+  for (iter->curPos = 1; iter->curPos < offField->firstTokPos && !BufferReader_AtEnd(&iter->rdr); ++iter->curPos) {
```
JoanFM: should this be encapsulated in a method with proper naming taking Iterator as input?

src/byte_offsets.c
```diff
@@ -103,25 +103,40 @@ int RSByteOffset_Iterate(const RSByteOffsets *offsets, uint32_t fieldId,
...
+  // If we're at the end of the stream, or we've reached the end of the field, return EOF
+  if (iter->curPos >= iter->endPos) {
     return RSBYTEOFFSET_EOF;
   }
 
-  iter->lastValue = ReadVarint(&iter->rdr) + iter->lastValue;
-  return iter->lastValue;
+  // it is possible we are at end of buffer but still need to return the last result
+  const uint32_t result = iter->lastValue;
+  ++iter->curPos;
```
JoanFM, line 133: should this be moved to after line 138?

src/byte_offsets.c
```diff
@@ -103,25 +103,40 @@ int RSByteOffset_Iterate(const RSByteOffsets *offsets, uint32_t fieldId,
...
   }
 
-  iter->curPos--;
+  // If we reached the end of the stream before we reached the first token position, return an error
+  if (iter->curPos < offField->firstTokPos) {
+    return REDISMODULE_ERR;
+  }
+
+  // if range is [1, 1] we want curPos to be 0 so RSByteOffsetIterator_Next will return the first value
+  --iter->curPos;
```
JoanFM, line 121: why don't we put it explicitly at 0 instead of -1 then? Not sure I see the relationship of the code with the comment.

JoanFM (changes requested):
some comments (not very familiar with the code)

PR #6769, opened 2025-09-03
[MOD-11161] Add ON_OOM config

lerman25:
This PR introduces a new search config, ON_OOM. 
It's identical to ON_TIMEOUT config, but controls the behavior when query has reached OOM stats.

PR #6828, opened 2025-09-15
[MOD-11398] Add "ignore" option to ON_OOM config

lerman25:
Add ignore as option to ON_OOM config.

Because on_oom and on_timeout do not share policy options, the changes from #6769 are reverted and the policy enums are seperated. 

Because of the revert, I suggest, for an easier review, to review by commit and not the PR as is, the revert commit was done automaticly.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1555,14 +1555,9 @@ int RSConfig_SetOption(RSConfig *config, RSConfigOptions *options, const char *n
 }
 
 const char *TimeoutPolicy_ToString(RSTimeoutPolicy policy) {
-  switch (policy) {
-    case TimeoutPolicy_Return:
-      return on_timeout_vals[TimeoutPolicy_Return];
-    case TimeoutPolicy_Fail:
-      return on_timeout_vals[TimeoutPolicy_Fail];
-    default:
-      return "invalid";
-  }
+  // Assert policy is valid
+  RS_ASSERT(policy < TimeoutPolicy_Invalid);
+  return on_timeout_vals[policy];
 }
 
 RSTimeoutPolicy TimeoutPolicy_Parse(const char *s, size_t n) {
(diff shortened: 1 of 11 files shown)


src/config.c
```diff
@@ -1554,26 +1554,52 @@ int RSConfig_SetOption(RSConfig *config, RSConfigOptions *options, const char *n
...
 
-const char *FailurePolicy_ToString(RSFailurePolicy policy) {
+const char *TimeoutPolicy_ToString(RSTimeoutPolicy policy) {
+  switch (policy) {
+    case TimeoutPolicy_Return:
+      return on_timeout_vals[TimeoutPolicy_Return];
+    case TimeoutPolicy_Fail:
+      return on_timeout_vals[TimeoutPolicy_Fail];
+    default:
+      return "invalid";
```
GuyAv46: Consider asserting if it's unreachable

src/config.c
```diff
@@ -1554,26 +1554,52 @@ int RSConfig_SetOption(RSConfig *config, RSConfigOptions *options, const char *n
...
-    case FailurePolicy_Return:
-      return on_failure_vals[FailurePolicy_Return];
-    case FailurePolicy_Fail:
-      return on_failure_vals[FailurePolicy_Fail];
+    case OomPolicy_Return:
+      return on_oom_vals[OomPolicy_Return];
+    case OomPolicy_Fail:
+      return on_oom_vals[OomPolicy_Fail];
+    case OomPolicy_Ignore:
+      return on_oom_vals[OomPolicy_Ignore];
```
GuyAv46: Consider asserting the offset and then `return on_oom_vals[policy];`

src/config.h
```diff
@@ -15,28 +15,49 @@
...
+static const char *on_timeout_vals[2] = {
   "return",
   "fail"
 };
 
+typedef enum {
+  OomPolicy_Return,       // Return what we have on OOM
+  OomPolicy_Fail,         // Just fail without returning anything
+  OomPolicy_Ignore,       // Ignore OOM and continue
+  OomPolicy_Invalid       // Not a real value
```
GuyAv46, line 36: Why do we need the `Invalid` value? (regarding both enums)

lerman25: In timeout, it is used in SET `ON_TIMEOUT` calling `TimeoutPolicy_Parse`, if a user gave a wrong string as option it returns `TimeoutPolicy_Invalid`. 
I followed suit.
You think it's better if we remove it and use a different return value for parse?

GuyAv46: No, up to you. I think it's OK, especially if it's aligned with past configs. I don't think it adds anything, and we don't store invalid configs, so it should be unreachable

GuyAv46 (approved):

PR #6573, opened 2025-07-30
MOD-10577: Dynamically use Disk API If Available

kei-nan:
## Describe the changes in the pull request

A clear and concise description of what the PR is solving, including:
1. Current: Search primarily operates on ram
2. Change: Conditionally use a disk API to index and search data on disk
3. Outcome: Allow disk API to be used in certain cases.

#### Main objects this PR modified
1. IndexSpec
2. SearchFlow

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #6841, opened 2025-09-16
[MOD-10577] Dynamically Using Search API

JoanFM:
It is the continuation of #6573 but fixing the conflicts with master, and aligning with new Iterator API

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -1261,12 +1261,12 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx) {
     return REDISMODULE_ERR;
   }
 
-  bool disk_initialized = SearchDisk_Initialize(ctx);
-  if (disk_initialized && isFlex) {
-    RedisModule_Log(ctx, "notice", "Search Disk is initialized");
-  } else if (!disk_initialized && isFlex) {
-    RedisModule_Log(ctx, "error", "Search Disk is enabled but could not be initialized");
-    return REDISMODULE_ERR;
+  if (isFlex) {
+    bool disk_initialized = SearchDisk_Initialize(ctx);
+    if (!disk_initialized) {
+      RedisModule_Log(ctx, "error", "Search Disk is enabled but could not be initialized");
+      return REDISMODULE_ERR;
+    }
   }
 
   // register trie-dictionary type


src/module.c
```diff
@@ -3585,18 +3593,25 @@ static int initSearchCluster(RedisModuleCtx *ctx, RedisModuleString **argv, int
...
+    if (clusterConfig.connPerShard) {
+      num_connections_per_shard = clusterConfig.connPerShard;
+    } else {
+      // default
+      num_connections_per_shard = RSGlobalConfig.numWorkerThreads + 1;
+    }
 
-  size_t num_io_threads = clusterConfig.coordinatorIOThreads;
-  size_t conn_pool_size = CEIL_DIV(num_connections_per_shard, num_io_threads);
+    size_t num_io_threads = clusterConfig.coordinatorIOThreads;
```
kei-nan: indentation

src/spec.h
```diff
@@ -189,6 +190,7 @@ typedef enum {
 
   Index_HasNonEmpty = 0x80000,  // Index has at least one field that does not indexes empty values
 
+  Index_StoreInRAM = 0x100000, // Index is stored on RAM (force it)
```
kei-nan: do we want to keep this flag?

JoanFM: I think it is confortable for testing and benchmarking (perhaps Raz wants to have this code only one "uncomment" away?

src/module.c
JoanFM: It has the same indentation as MR_Init and is due to the nested `if`. Do not think there is any difference here.

kei-nan: I do see this comment:
`// Temporary hack to get single-shard logic without sending first `search-CLUSTERSET` command`
I think this is something Raz added.
Not sure if we want this part in master.

src/spec.h
kei-nan: Not a fan of keeping it inside master.
If Raz needs it he can create his own branch and use that.

src/param.c
```diff
@@ -8,6 +8,7 @@
 */
 #include "param.h"
 #include "rmalloc.h"
+#include "module.h"
```
JoanFM: remove this import

src/module.c
```diff
@@ -3570,7 +3578,7 @@ int SetClusterCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
 
 /* Perform basic configurations and init all threads and global structures */
 static int initSearchCluster(RedisModuleCtx *ctx, RedisModuleString **argv, int argc, bool isClusterEnabled) {
-  RedisModule_Log(ctx, "notice",
+  RedisModule_Log(ctx, "warning",
```
kei-nan: maybe keep it as info?

src/result_processor.c
```diff
@@ -128,15 +130,28 @@ static int rpQueryItNext(ResultProcessor *base, SearchResult *res) {
     }
 
 validate_current:
-
-    if (it->current->dmd) {
-      dmd = it->current->dmd;
+    IndexSpec* spec = RP_SPEC(base);
+    if (spec->diskSpec) {
+      RSDocumentMetadata* diskDmd = (RSDocumentMetadata *)rm_calloc(1, sizeof(RSDocumentMetadata));
```
kei-nan: maybe this snippet of code can move to a small function?

src/module.c
JoanFM: yes u are right

src/result_processor.c
```diff
@@ -634,9 +649,17 @@ typedef struct {
 static void rpLoader_loadDocument(RPLoader *self, SearchResult *r) {
   // If the document was modified or deleted, we don't load it, and we need to mark
   // the result as expired.
-  if ((r->dmd->flags & Document_FailedToOpen) || (r->dmd->flags & Document_Deleted)) {
-    r->flags |= Result_ExpiredDoc;
-    return;
+  if (self->base.parent->sctx->spec->diskSpec) {
```
kei-nan: this snippet can probably move to a function and wrap the logic of finding out if a doc was deleted/expired.

src/module.c
```diff
@@ -1127,6 +1128,11 @@ static void GetRedisVersion(RedisModuleCtx *ctx) {
     RedisModule_FreeCallReply(reply);
   }
 
+  char *isFlexStr = getRedisConfigValue(ctx, "bigredis-enabled");
```
kei-nan: Maybe this check can move to search_disk.c.
It will encapsulate the logic of choosing if we need to be on disk or not.
Easier to maintain.

CMakeLists.txt
```diff
@@ -292,7 +292,29 @@ set(FINAL_OBJECTS
...
+# Default behavior: SHARED when main project, STATIC when subdirectory
+if(NOT DEFINED REDISEARCH_BUILD_SHARED)
+    if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+        set(REDISEARCH_BUILD_SHARED ON)
+    else()
+        set(REDISEARCH_BUILD_SHARED OFF)
+    endif()
+endif()
+
+option(REDISEARCH_BUILD_SHARED "Build RediSearch as shared library" OFF)
```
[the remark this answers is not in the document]
JoanFM: having this optuon before the logic, seemed to take precedence

src/search_disk.c
```diff
@@ -0,0 +1,91 @@
...
+    return disk->docTable.getDocumentMetadata(handle, docId, dmd, &sdsnewlen);
+}
+
+bool SearchDisk_DocIdDeleted(RedisSearchDiskIndexSpec *handle, t_docId docId) {
+    RS_ASSERT(disk && handle);
+    return disk->docTable.isDocIdDeleted(handle, docId);
+}
+
+bool SearchDisk_IsEnabled(RedisModuleCtx *ctx) {
+  bool isFlex = false;
```
kei-nan, line 84: So I think this entire code can move to the `SearchDisk_Initialize` function.
Then the caller doesn't need to know about flex and we encapsulate things a bit better.

src/result_processor.c
```diff
@@ -86,6 +88,44 @@ typedef struct {
...
+ * If the document is not deleted or expired, return true.
+ * If the document is not deleted or expired, and dmd is not NULL, set *dmd to the document metadata.
+ * @param spec The index spec
+ * @param docs The document table
+ * @param sctx The search context
+ * @param it The query iterator
+ * @param dmd The document metadata pointer to set
+ * @return true if the document is not deleted or expired, false otherwise.
+ */
+static bool Get_DocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx *sctx, const QueryIterator *it, const RSDocumentMetadata **dmd) {
```
kei-nan: The function name is a bit odd.

```suggestion
static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx *sctx, const QueryIterator *it, const RSDocumentMetadata **dmd) {
```

src/search_disk.c
JoanFM: I kind of prefer separating the `Initialization`with the `allowance `or  `enablement`.

kei-nan: But right now we still initialize even if it is disabled... which doesn't make sense
```cpp
  bool disk_initialized = SearchDisk_Initialize(ctx);
  if (disk_initialized && isFlex) {
    RedisModule_Log(ctx, "notice", "Search Disk is initialized");
  } else if (!disk_initialized && isFlex) {
    return REDISMODULE_ERR;
  }
```
And we still assign the enable result into isFlex which breaks the encapsulation IMHO.

JoanFM: but because right now enablement implies  `isFlex`. The logic could change if requirement change.

I will skip initialization if not needed is right

kei-nan (approved):

PR #6842, opened 2025-09-16
build.sh support for python tests with external redis-server instances

DarthB:

Add two arguments to `build.sh` to control the pytests handling of redis-server, allowing it to connect to an external redis-server on a specified IP (default 127.0.0.1) and port (default 6379).

This allows to start a redis-server instance attached with a debugger from an IDE, e.g. Visual Code, and let the python test use that redis-server as backend.

#### Mark if applicable

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -84,6 +84,20 @@ parse_arguments() {
       RUN_PYTEST|run_pytest)
         RUN_PYTEST=1
         ;;
+      EXT=*|ext=*)
+        EXT="${arg#*=}"
+        ;;
+      EXT_HOST=*|ext_host=*)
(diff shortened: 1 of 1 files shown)


ofiryanai (approved):
Looking good. 
A question- when are you planning to use it? When running tests locally? Isn't the redis-server gets teardown after the tests finish? Interested in the workflow that requires that

DarthB:
> Looking good. A question- when are you planning to use it? When running tests locally? Isn't the redis-server gets teardown after the tests finish? Interested in the workflow that requires that

I use `VS Code` to debug redis-server with the extensions `RediSearch` and `REJSON` loaded via `Run->Start Debugging`. 

So if a `pytest` crashes the server or asserts, I can use the debugger, add breakpoints, and step through the code.

*Addition*:
If you give them an "external" instance, the tear-down does not happen. This can lead to side effects if you run multiple tests. Best use `build.sh DEBUG RUN_PYTEST TEST=YourTest`

PR #6846, opened 2025-09-17
MOD-10082 Rust Wildcard Iterator

ofiryanai:

Implement Wildcard Iterator in Rust.
Changed the benchmarks C integration to call direct implementation instead of ffi call on every function to avoid overhead and allow C code compiler optimizations.
Results:
```
Iterator - Wildcard - Read Large Range/C                       time:   [291.63 µs 292.65 µs 293.63 µs]
Iterator - Wildcard - Read Large Range/Rust                    time:   [286.83 µs 288.00 µs 289.16 µs]
Iterator - Wildcard - Read Medium Range/C                      time:   [147.38 µs 147.92 µs 148.41 µs]
Iterator - Wildcard - Read Medium Range/Rust                   time:   [143.62 µs 144.69 µs 146.21 µs]
Iterator - Wildcard - SkipTo Large Range/C                     time:   [8.0376 µs 8.0798 µs 8.1230 µs]
Iterator - Wildcard - SkipTo Large Range/Rust                  time:   [3.8148 µs 3.8224 µs 3.8328 µs]
Iterator - Wildcard - SkipTo Medium Range/C                    time:   [4.0508 µs 4.0666 µs 4.0816 µs]
Iterator - Wildcard - SkipTo Medium Range/Rust                 time:   [1.9079 µs 1.9105 µs 1.9139 µs]
```

#### Mark if applicable


src/redisearch_rs/build_utils/src/lib.rs
```diff
@@ -156,6 +156,7 @@ pub fn generate_c_bindings(
     let includes = [
         root.join("deps").join("RedisModulesSDK"),
         root.join("src"),
+        root.join("src").join("inverted_index"),  // Add missing inverted_index include
```
chesedo: This adds inverted index for everything that uses this util. We will only want to add it for the iterator benchmarks instead.

src/redisearch_rs/rqe_iterators/src/wildcard.rs
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! Empty iterator implementation
```
chesedo: Just update the module comment

src/redisearch_rs/rqe_iterators/src/wildcard.rs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        self.current_id += 1;
+        self.result.doc_id = self.current_id;
+        Ok(Some(&self.result))
+    }
+
+    fn skip_to(
+        &mut self,
+        _doc_id: t_docId,
```
chesedo:
```suggestion
        doc_id: t_docId,
```
The underscore prefix is a convention used to indicate a variable is not being used. Since you are using the variable it should not be needed.

src/redisearch_rs/rqe_iterators/src/wildcard.rs
```diff
@@ -0,0 +1,84 @@
...
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! Empty iterator implementation
+use ffi::t_docId;
+use inverted_index::RSIndexResult;
+
+use crate::{RQEIterator, RQEIteratorError, SkipToOutcome};
+
+/// An iterator that yields all ids within a given range, from 0 to max id in an index.
```
chesedo:
```suggestion
/// An iterator that yields all ids within a given range, from 0 to max id (but not including) in an index.
```

src/redisearch_rs/rqe_iterators_bencher/src/benchers/wildcard.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        group.finish();
+    }
+
+    fn skip_patterns(&self, c: &mut Criterion) {
+        let mut group = self.benchmark_group(c, "Iterator - Wildcard - Skip Patterns");
+
+        // Variable step sizes
+        group.bench_function("Variable Steps", |b| {
+            let steps = [1, 10, 100, 1000];
```
chesedo: You will rather want a bench function for each step like this

https://github.com/RediSearch/RediSearch/blob/a4c9723b2f8afb63fe2efbb458458ddf0b09d4fa/src/redisearch_rs/inverted_index_bencher/src/benchers/numeric.rs#L355-L374

src/redisearch_rs/rqe_iterators_bencher/src/benchers/wildcard.rs
```diff
@@ -0,0 +1,335 @@
...
+                        }
+                    }
+                },
+                criterion::BatchSize::SmallInput,
+            );
+        });
+
+        // Random access pattern
+        group.bench_function("Random Access", |b| {
+            let targets = [1, 50000, 25000, 75000, 10000, 90000, 5000, 95000];
```
chesedo: This can also have a bench function for each target

src/redisearch_rs/rqe_iterators_bencher/src/direct_benchmarks.c
chesedo, line 1: I'm interested in how much this approach changed the final benchmarking numbers. It's nice that there is a way to avoid the FFI overhead. I'm not sure if the improvement is worth the creation of a 130-line temporary file though.

src/redisearch_rs/build_utils/src/lib.rs
```diff
@@ -156,6 +156,7 @@ pub fn generate_c_bindings(
     let includes = [
         root.join("deps").join("RedisModulesSDK"),
         root.join("src"),
+        root.join("src").join("inverted_index"), // Add missing inverted_index include
```
chesedo: This adds `inverted_index` to everything which uses this utils crate. We won't want that, but will only want the iterators bencher to have this include.

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+mod c_mocks;
+
+// Test cases with different top_id values to cover various scenarios
+static TEST_CASES: &[(u64, &str)] = &[
+    (100, "standard range"),
+    (1, "single document"),
+    (10, "small range"),
+    (1000, "large range"),
+    (42, "arbitrary value"),
+];
```
chesedo: I would not have these, but will rather keep the tests simple by only testing a single API contract / behaviour. Currently, none of these test the logic in a unique way (ie they all run exactly the same code paths)

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+        );
+    }
+}
+
+#[test]
+fn skip_to_valid_targets() {
+    let top_id = 100u64;
+    let skip_targets = [5, 10, 20, 50, 75, 100];
+
+    for &target in &skip_targets {
```
chesedo: Let's remove the loop for hard-coded values instead.

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+    assert!(result.is_ok(), "skip_to({}) should succeed", skip_target);
+
+    let outcome = result.unwrap();
+    assert!(
+        outcome.is_some(),
+        "skip_to({}) should return Some",
+        skip_target
+    );
+
+    if let SkipToOutcome::Found(doc) = outcome.unwrap() {
```
chesedo: This test should also check that `Found` is always returned

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+        assert_eq!(
+            it.last_doc_id(),
+            skip_target,
+            "last_doc_id should be {} after skip",
+            skip_target
+        );
+    }
+
+    // Continue reading sequentially from skip_target + 1
+    for expected_id in (skip_target + 1)..=top_id {
```
chesedo: This should not need a loop. Only one or two reads will be enough

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+                description,
+                valid_target
+            );
+        }
+    }
+}
+
+#[test]
+fn zero_documents() {
+    let mut it = Wildcard::new(0);
```
chesedo, line 188: Does anything ever construct the wildcard iterator with `0`? Why should it be possible - ie is it not the same as the empty iterator?

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+    assert!(result.is_ok(), "second skip_to({}) should succeed", target);
+
+    let outcome = result.unwrap();
+    assert!(
+        outcome.is_some(),
+        "second skip_to({}) should return Some",
+        target
+    );
+
+    if let SkipToOutcome::Found(doc) = outcome.unwrap() {
```
chesedo: This should also check that `Found` is always returned

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+            "skip_to({}) should find exact target",
+            second_target
+        );
+        assert_eq!(
+            it.last_doc_id(),
+            second_target,
+            "last_doc_id should be {} after backward skip",
+            second_target
+        );
+    }
```
chesedo: Hmm, this should never hit right?

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+}
+
+#[test]
+fn mixed_read_and_skip_operations() {
```
chesedo: I would say the skip to test already does the same and this extra test is not needed.

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+    assert!(result.is_ok(), "read() after EOF should succeed");
+    assert!(
+        result.unwrap().is_none(),
+        "read() after EOF should return None"
+    );
+}
+
+#[test]
+fn edge_case_single_document() {
+    let mut it = Wildcard::new(1);
```
chesedo: Why is this an edge case? I don't see it triggering any special behaviour

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+        assert_eq!(it.last_doc_id(), 1, "last_doc_id should be 1 after skip");
+        assert!(
+            it.at_eof(),
+            "should be at EOF after skipping to last document"
+        );
+    }
+}
+
+#[test]
+fn stress_test_large_ranges() {
```
chesedo: What's the purpose of this test? I don't see what you can change in the code that will make all the other tests pass but will cause this test to fail

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+                doc.doc_id, i,
+                "Case '{}': read({}) wrong doc_id",
+                description, i
+            );
+        }
+    }
+}
+
+#[test]
+fn property_based_skip_to_invariants() {
```
chesedo: I would say this is the same as the skip to test and can be removed

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+                    target,
+                    target + 1
+                );
+            }
+        }
+    }
+}
+
+#[test]
+fn memory_efficiency_validation() {
```
chesedo: Why is this test needed?

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,809 @@
...
+            result.unwrap().is_none(),
+            "Case '{}': should still return None after EOF",
+            description
+        );
+    }
+}
+
+#[test]
+fn skip_to_valid_targets() {
+    let top_id = 100u64;
```
chesedo: This file has the `u64` suffices at a bunch of places. You should not need most of them which will make the code easier to read.

chesedo (commented):
Looks good. I mostly have comments on making things simpler.

src/redisearch_rs/rqe_iterators/src/wildcard.rs
chesedo: Oops, this suggestion is wrong. It should be from 1 to max id (inclusive)

src/redisearch_rs/rqe_iterators_bencher/src/direct_benchmarks.c
ofiryanai: The bench for rust read were X6 faster than the equivalent C. When I switched to this method they were roughly equivalent. I wouldn't add so much code otherwise

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: I agree, I'll simplify to one happy path test instead of all these.

src/redisearch_rs/rqe_iterators_bencher/src/direct_benchmarks.c
chesedo: Oh, wow. Yeah, let's keep it then

src/redisearch_rs/rqe_iterators/src/wildcard.rs
ofiryanai: correct! fixed

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: removed

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: This is actually not supported I think we never skip backwards. also we don't validate this

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: added

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: I think this scenario can happen if a query with wildcard iterator is executed on an empty index. It does not trigger a different behavior but I think it's a fine contract to show with a test

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: 👍  added `else { panic! }` statement to the if let unpacking

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: you're correct, added debug_assert! to enforce that.

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: it doesn't test anything special. a bit too defensive I agree. removed 👍

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: removed 👍

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: removed

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: removed

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,241 @@
...
+    let result = it.skip_to(5);
+    assert!(result.is_ok());
+    assert_eq!(it.last_doc_id(), 5);
+
+    // only skipping to position greater than last_doc_id is possible, should panic
+    std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        let _ = it.skip_to(5);
+    }))
+    .expect_err("Expected panic when skipping to same position");
+}
```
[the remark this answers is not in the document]
ofiryanai: added `#[cfg(debug_assertions)]` to address that

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,241 @@
...
```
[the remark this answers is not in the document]
ofiryanai: added `#[cfg(debug_assertions)]` to address that

src/redisearch_rs/rqe_iterators_bencher/src/benchers/wildcard.rs
ofiryanai: I split the method to variable_step and random_access and provided steps and targets as inputs, hope it addresses your comment

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,241 @@
...
+
+    let outcome = result.unwrap();
+    assert!(outcome.is_some());
+
+    if let Some(SkipToOutcome::Found(doc)) = outcome {
+        assert_eq!(doc.doc_id, target_doc_id);
+    } else {
+        panic!("Expected Found outcome, got {:?}", outcome);
+    }
+}
```
chesedo: You will want to use a proc macro instead for a helper in tests else a failed test will reference these lines. And this will make it harder to find the actual test line which failed

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,241 @@
...
+    let mut it = Wildcard::new(100);
+
+    let _ = it.skip_to(75);
+
+    // Try to skip backwards to position 25, should panic
+    std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        let _ = it.skip_to(25);
+    }))
+    .expect_err("Expected panic when skipping backwards");
+}
```
chesedo, line 222:
```suggestion
#[test]
#[should_panic(expected = "The expected panic message")]
fn skip_to_backwards() {
    let mut it = Wildcard::new(100);

    let _ = it.skip_to(75);

    // Try to skip backwards to position 25, should panic
    let _ = it.skip_to(25);
}
```
The idiomatic way to test something panics with a given message

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
```diff
@@ -0,0 +1,241 @@
...
+
+#[test]
+#[cfg(debug_assertions)]
+fn skip_to_same_position() {
```
chesedo, line 227: This can use the same update as the `should_panic` suggestion

chesedo (commented):
Only small suggestions

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: Thanks, changed to  `#[should_panic]`, looking much better. I don't like asserting the specific message, too detailed expectation IMO

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
ofiryanai: cool, changed it. didn't macro yet in any language thats a first

src/redisearch_rs/rqe_iterators/tests/wildcard.rs
chesedo: Using a specific panic message will just prevent the test from accidentally passing on another panic. Especially helpful if you are the one controlling the panic message.

chesedo (dismissed):

src/redisearch_rs/rqe_iterators/src/wildcard.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// A reusable result object to avoid allocations on each `read` call.
+    result: RSIndexResult<'static>,
+}
+
+impl Wildcard {
+    pub fn new(top_id: t_docId) -> Self {
+        Wildcard {
+            top_id,
+            current_id: 0,
+            result: RSIndexResult::virt(),
```
GuyAv46: `Freq` is 0 by default now, right? We should set it to 1, no?

src/redisearch_rs/rqe_iterators_bencher/build.rs
```diff
@@ -16,13 +16,25 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    let headers = ["iterator_api.h", "empty_iterator.h", "idlist_iterator.h"]
-        .iter()
-        .map(|h| root.join("src").join("iterators").join(h))
-        .collect::<Vec<_>>();
-    generate_c_bindings(headers, ".*/iterators/.*.h")?;
+    cc::Build::new()
+        .file("src/benchers/c/wildcard.c")
+        .include(root.join("src").join("wildcard"))
+        .opt_level(3)
+        .compile("wildcard_iterator_benchmark");
```
GuyAv46, line 25: Should we do that for every iterator?

src/redisearch_rs/rqe_iterators/src/wildcard.rs
ofiryanai: set to 1. btw, in id list there's no need to set freq to 1?

src/redisearch_rs/rqe_iterators_bencher/build.rs
ofiryanai: I'm not sure, my initial benchmark got X6 faster rust because of FFI calls to the C code that probably didn't allow optimizations. If it weren't producing weird results, I wouldn't go down that path. but writing a direct C code fixed that and now rust looks roughly the same performance as C as expected

LukeMathWalker, line 25: From a build system perspective, wouldn't it preferable to compile the C code via cmake and then limit ourselves to linking here? 
We've generally tried to avoid compiling C from Rust build scripts, to make sure we don't diverge in options with the overall build.

ofiryanai: I'm not sure where exactly would it be fit to add to existing cmake. The single standalone file compiled here is highly coupled with the `rqe_iterators_bencher` which is rust based, converted to C just to create even grounds for the benchmark.
Correct my if I'm wrong, regarding your general effort you're talking about the "original" C impl of redisearch, right? Where there's already cmake that compiling it. When the porting to Rust will finish the cmake will kind of evaporate as well. I think this is a separate story here

src/redisearch_rs/rqe_iterators/src/wildcard.rs
```diff
@@ -0,0 +1,85 @@
...
+/// An iterator that yields all ids within a given range, from 1 to max id (inclusive) in an index.
+#[derive(Default)]
+pub struct Wildcard {
+    // Supposed to be the max id in the index
+    top_id: t_docId,
+
+    current_id: t_docId,
+
+    /// A reusable result object to avoid allocations on each `read` call.
+    result: RSIndexResult<'static>,
```
GuyAv46: Shouldn't the result be bound to the iterator's lifetime?

ofiryanai: added lifetime `'a` to Wildcard iterator and its result

GuyAv46 (approved):

PR #6853, opened 2025-09-18
[MOD-11403] Add `revalidate` to the Rust inverted index reader

chesedo:
This allows the reader to communicate to iterators it needs a reset because the underlying index was garbage collected. All the autogenerated FFI code is in `headers/future` so the C code is not using them yet. The switch will happen in another PR.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/headers/future/inverted_index.h
+++ b/src/redisearch_rs/headers/future/inverted_index.h
@@ -229,7 +229,7 @@ uintptr_t InvertedIndex_GcMarker(const struct InvertedIndex *ii);
  * The following invariant must be upheld when calling this function:
  * - `ii` must be a valid, non NULL, pointer to an `InvertedIndex` instance.
  */
-void InvertedIndex_GcMarkerInc(const struct InvertedIndex *ii);
+void InvertedIndex_GcMarkerInc(struct InvertedIndex *ii);
 
 /**
  * Create a new inverted index reader for the given inverted index and filter. The returned pointer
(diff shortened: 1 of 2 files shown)


GlenDC (approved):

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -733,6 +775,7 @@ impl<'index, E: DecodedBy<Decoder = D>, D: Decoder> IndexReader<'index, E, D> {
             current_buffer: Cursor::new(&first_block.buffer),
             current_block_idx: 0,
             last_doc_id: first_block.first_doc_id,
+            gc_marker: ii.gc_marker.load(atomic::Ordering::Relaxed),
```
LukeMathWalker, line 778: We might as well use the method we defined, no?

chesedo: That method is only defined for some C tests. I was thinking we remove it after cleaning up the tests.

src/redisearch_rs/headers/future/inverted_index.h
```diff
@@ -211,6 +211,26 @@ const struct IndexBlock *InvertedIndex_BlockRef(const struct InvertedIndex *ii,
...
+
+/**
+ * Increment the garbage collector marker of the inverted index. This is used by some C tests.
+ *
+ * # Safety
+ *
+ * The following invariant must be upheld when calling this function:
+ * - `ii` must be a valid, non NULL, pointer to an `InvertedIndex` instance.
+ */
+void InvertedIndex_GcMarkerInc(const struct InvertedIndex *ii);
```
meiravgri: why const if the value is about to be modified?

chesedo: Nice catch! Updated in https://github.com/RediSearch/RediSearch/pull/6853/commits/f347bf6004ca627f3a1ecaa46cc458510825c010

PR #6854, opened 2025-09-18
ci: adapt redisearch_rs build to allow RediSearch as dependency of 3rd party

JoanFM:

Edit the build system so that redisearch_rs compilation is based in CMakeLists.txt (this allows RediSearch to be added as a dependency from another project)

Related / extracted from #6573 #6841

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -358,10 +358,8 @@ prepare_cmake_arguments() {
     export RUSTFLAGS
   fi
 
-  # Forward RUSTFLAGS to CMake if set
-  if [[ "$RUSTFLAGS" != "" ]]; then
-    CMAKE_BASIC_ARGS="$CMAKE_BASIC_ARGS -DRUSTFLAGS=\"$RUSTFLAGS\""
-  fi
+  # RUSTFLAGS will be passed as environment variable to avoid quoting issues
+  # This prevents CMake argument parsing from truncating complex flag values
 
   if [[ "$RUST_PROFILE" != "" ]]; then
     CMAKE_BASIC_ARGS="$CMAKE_BASIC_ARGS -DRUST_PROFILE=$RUST_PROFILE"
@@ -403,9 +401,9 @@ run_cmake() {
     # If verbose, dump all CMake variables before and after configuration
     if [[ "$VERBOSE" == "1" ]]; then
       echo "Running CMake with verbose output..."
-      $CMAKE_CMD --trace-expand
+      RUSTFLAGS="$RUSTFLAGS" $CMAKE_CMD --trace-expand
     else
-      $CMAKE_CMD
+      RUSTFLAGS="$RUSTFLAGS" $CMAKE_CMD
     fi
   fi
 }
--- a/src/redisearch_rs/CMakeLists.txt
+++ b/src/redisearch_rs/CMakeLists.txt
@@ -12,12 +12,9 @@ if(NOT DEFINED RUST_PROFILE)
     set(RUST_PROFILE "release")
 endif()
 
-# Set RUSTFLAGS (use CMake variable if provided, otherwise use environment variable)
-if(DEFINED RUSTFLAGS)
-    set(RUST_FLAGS "${RUSTFLAGS}")
-else()
-    set(RUST_FLAGS "$ENV{RUSTFLAGS}")
-endif()
+# Set RUSTFLAGS from environment variable
+# This avoids CMake argument parsing issues with complex flag values
+set(RUST_FLAGS "$ENV{RUSTFLAGS}")
 
 # Map Rust profile names to their corresponding artifact directory names
 if(RUST_PROFILE STREQUAL "dev")


src/redisearch_rs/CMakeLists.txt
```diff
@@ -0,0 +1,103 @@
...
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug" OR DEFINED SAN OR DEFINED COV)
+        set(RUST_PROFILE "dev")
+    else()
+        set(RUST_PROFILE "optimised_test")
+    endif()
+else()
+    # For non-test builds, only use dev profile for Debug builds (not SAN/COV)
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+        set(RUST_PROFILE "dev")
+    endif()
```
LukeMathWalker: This should follow the same pattern of `RUST_DYN_CRT`: allow the invoker to set the profile, but have fallback logic if it isn't set.
We should also remove the corresponding profile selection logic from `build.sh` and just forward `RUST_PROFILE` if it was set by the user.

LukeMathWalker: Thinking about it again: we can't remove that logic from `build.sh` since we need it for test execution. But duplicating it here isn't nice either 🤔

JoanFM: I see, but is there much of an alternative?

LukeMathWalker: It depends on the requirements of the projects that are pulling this in as a dependency.
Do they care about test execution? That seems a concern that's relevant only to the repository where the project is developed.
If that's the case, we could remove all this logic here and simply default to `release` if a `RUST_PROFILE` hasn't explicitly been set by the caller.

JoanFM: no, for that project is not relevant

LukeMathWalker: Then let's keep the profile selection logic in `build.sh` and simplify this down to "either propagate or fallback to `release`".

src/redisearch_rs/CMakeLists.txt
```diff
@@ -0,0 +1,105 @@
...
+# Check if we're being used as a subdirectory to avoid target conflicts
+if(NOT TARGET redisearch_rs_build)
+    # Create the custom target for building Rust code
+    add_custom_target(redisearch_rs_build
+        COMMAND ${CMAKE_COMMAND} -E env
+            "RUSTFLAGS=${RUST_FLAGS}"
+            "CARGO_TARGET_DIR=${RUST_TARGET_DIR}"
+            ${CARGO_EXECUTABLE} build
+            --workspace
+            ${EXCLUDE_ARGS}
```
LukeMathWalker: Based on the conversation above, you don't want to build the whole workspace.
You only want to build the crate that generates the static library you need, as well as its deps.
```suggestion
            ${CARGO_EXECUTABLE} build
            --package redisearch_rs
```
That should allow you to remove `EXCLUDE_ARGS` entirely.

src/redisearch_rs/CMakeLists.txt
```diff
@@ -0,0 +1,105 @@
...
+        endif()
+    endif()
+endif()
+
+# Set up RUSTFLAGS - respect existing environment variable and default to -D warnings
+if(DEFINED ENV{RUSTFLAGS})
+    set(RUST_FLAGS "$ENV{RUSTFLAGS}")
+else()
+    set(RUST_FLAGS "-D warnings")
+endif()
```
LukeMathWalker: We should default to empty. `-D warnings` only makes sense for project developers.

src/redisearch_rs/CMakeLists.txt
```diff
@@ -0,0 +1,105 @@
...
+        else()
+            set(RUST_PROFILE "optimised_test")
+        endif()
+    else()
+        # For non-test builds, only use dev profile for Debug builds (not SAN/COV)
+        if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+            set(RUST_PROFILE "dev")
+        endif()
+    endif()
+endif()
```
LukeMathWalker, line 13:
```suggestion
if(NOT DEFINED RUST_PROFILE)
    if(CMAKE_BUILD_TYPE STREQUAL "Debug" OR DEFINED SAN OR DEFINED COV)
        set(RUST_PROFILE "dev")
    else()
         set(RUST_PROFILE "release")
    endif()
endif()
```
This is what I meant before.

src/redisearch_rs/CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
+cmake_minimum_required(VERSION 3.13)
+
+# Find Rust and Cargo
```
LukeMathWalker:
```suggestion
# Find Cargo
```

build.sh
```diff
@@ -181,16 +181,22 @@ setup_build_environment() {
     FLAVOR="release"
   fi
 
-  # If unset, determine the Rust build profile
-  if [[ "$RUST_PROFILE" == "" ]]; then
+  # Determine Rust profile for tests (keep this logic for rust_tests functionality)
+  # This is separate from the CMake RUST_PROFILE forwarding
+  # Only set RUST_PROFILE if it wasn't already set by the user
```
LukeMathWalker:
```suggestion
  # Determine the correct Rust profile for both build and tests
  # Only set RUST_PROFILE if it wasn't already set by the user
```

src/redisearch_rs/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+endif()
+
+# Add dynamic C runtime flag if requested (crucial for Alpine Linux)
+if(DEFINED RUST_DYN_CRT AND RUST_DYN_CRT)
+    set(RUST_FLAGS "${RUST_FLAGS} -C target-feature=-crt-static")
+endif()
+
+# Exclude benchmarking crates that link C code
+set(EXCLUDE_ARGS "--exclude" "inverted_index_bencher" "--exclude" "rqe_iterators_bencher")
+
```
LukeMathWalker:
```suggestion
```
No longer needed.

build.sh
```diff
@@ -340,6 +345,25 @@ prepare_cmake_arguments() {
...
+
+  # Handle RUST_DYN_CRT flag for Alpine Linux compatibility
+  if [[ "$RUST_DYN_CRT" == "1" ]]; then
+    # Add the dynamic C runtime flag to RUSTFLAGS
+    if [[ "$RUSTFLAGS" == "" ]]; then
+      RUSTFLAGS="-C target-feature=-crt-static"
+    else
+      RUSTFLAGS="$RUSTFLAGS -C target-feature=-crt-static"
+    fi
+  fi
```
LukeMathWalker, line 359: To match the previous logic, we need to `export` the resulting `RUSTFLAGS`.

LukeMathWalker (approved):

PR #6856, opened 2025-09-18
MOD-11172: Add Argument Parser Struct To Improve Parsing Experience

kei-nan:
The parsing logic of Hybrid is intertwined with that of AGGREGATE and SEARCH.
Some keywords shouldn't be supported in HYBRID but will silently get accepted because we use the same parsing functions.
e.g handleCommonArgs.

This PR is meant to provide an infrastructure to separate the parsing flow of Hybrid and improve the ease in which we parse commands.

1. Current: No simple parsing framework exists
2. Change: Add an ArgsParser struct
3. Outcome: Provide an infrastructure for parsing commands.

1. ArgsParser

#### Mark if applicable

### DIFF
--- a/src/util/arg_parser.c
+++ b/src/util/arg_parser.c
@@ -321,27 +321,13 @@ static int parse_single_arg(ArgParser *parser, ArgDefinition *def) {
         }
 
         case ARG_TYPE_SUBARGS: {
-            // For SubArgs, we need to determine how many arguments to consume
-            int args_to_consume = def->options.subargs.max_args;
-
-            // If max_args is -1, it means variable length - try to consume remaining args
-            if (args_to_consume == -1) {
-                args_to_consume = AC_NumRemaining(parser->cursor);
+            if (def->options.subargs.max_args && def->options.subargs.min_args == def->options.subargs.max_args) {
+                // Single argument slice
+                rv = AC_GetSlice(parser->cursor, (ArgsCursor*)def->target, def->options.subargs.max_args);
+            } else {
+                // Variable arguments
+                rv = AC_GetVarArgs(parser->cursor, (ArgsCursor*)def->target);
             }
-
-            // Ensure we don't consume more than available or less than minimum
-            int remaining = AC_NumRemaining(parser->cursor);
-            if (args_to_consume > remaining) {
-                args_to_consume = remaining;
-            }
-
-            // Check minimum requirement
-            if (args_to_consume < def->options.subargs.min_args) {
-                set_error(parser, "Not enough arguments provided", def->name);
-                return AC_ERR_NOARG;
-            }
-
-            rv = AC_GetSlice(parser->cursor, (ArgsCursor*)def->target, args_to_consume);
             break;
         }
 
--- a/src/util/arg_parser.c
+++ b/src/util/arg_parser.c
@@ -125,18 +125,18 @@ ArgParser *ArgParser_AddInt(ArgParser *parser, const char *name, const char *des
     return parser;
 }
 
-ArgParser *ArgParser_AddLong(ArgParser *parser, const char *name, const char *description,
+ArgParser *ArgParser_AddLongLong(ArgParser *parser, const char *name, const char *description,
                             long long *target) {
-    ArgDefinition *def = add_definition(parser, name, description, ARG_TYPE_LONG, target);
+    ArgDefinition *def = add_definition(parser, name, description, ARG_TYPE_LONG_LONG, target);
     if (def && target) {
         *target = 0;  // Initialize to 0
     }
     return parser;
 }
 
-ArgParser *ArgParser_AddULong(ArgParser *parser, const char *name, const char *description,
+ArgParser *ArgParser_AddULongLong(ArgParser *parser, const char *name, const char *description,
                              unsigned long long *target) {
-    ArgDefinition *def = add_definition(parser, name, description, ARG_TYPE_ULONG, target);
+    ArgDefinition *def = add_definition(parser, name, description, ARG_TYPE_ULONG_LONG, target);
     if (def && target) {
         *target = 0;  // Initialize to 0
     }
@@ -279,7 +279,7 @@ static int parse_single_arg(ArgParser *parser, ArgDefinition *def) {
             break;
         }
 
-        case ARG_TYPE_LONG: {
+        case ARG_TYPE_LONG_LONG: {
             long long long_val;
             rv = AC_GetLongLong(parser->cursor, &long_val, 0);
             if (rv == AC_OK && def->target) {
@@ -297,7 +297,7 @@ static int parse_single_arg(ArgParser *parser, ArgDefinition *def) {
             break;
         }
 
-        case ARG_TYPE_ULONG: {
+        case ARG_TYPE_ULONG_LONG: {
             unsigned long long ulong_val;
             rv = AC_GetUnsignedLongLong(parser->cursor, &ulong_val, 0);
             if (rv == AC_OK && def->target) {
@@ -619,7 +619,7 @@ static void apply_defaults(ArgParser *parser) {
             case ARG_TYPE_INT:
                 *(int*)def->target = (int)def->defaults.int_default;
                 break;
-            case ARG_TYPE_LONG:
+            case ARG_TYPE_LONG_LONG:
                 *(long long*)def->target = def->defaults.int_default;
                 break;
             case ARG_TYPE_DOUBLE:
@@ -662,8 +662,8 @@ static void apply_variadic_options(ArgParser *parser, va_list args) {
             }
 
             case ARG_OPT_RANGE:
-                if (def->type == ARG_TYPE_INT || def->type == ARG_TYPE_LONG ||
-                    def->type == ARG_TYPE_ULONG) {  // Removed ARG_TYPE_UINT (doesn't exist)
+                if (def->type == ARG_TYPE_INT || def->type == ARG_TYPE_LONG_LONG ||
+                    def->type == ARG_TYPE_ULONG_LONG) {
                     def->options.numeric.min_val = va_arg(args, long long);
                     def->options.numeric.max_val = va_arg(args, long long);
                     def->options.numeric.has_min = true;
@@ -685,7 +685,7 @@ static void apply_variadic_options(ArgParser *parser, va_list args) {
                 break;
 
             case ARG_OPT_DEFAULT_INT:
-                if (def->type == ARG_TYPE_INT || def->type == ARG_TYPE_LONG) {
+                if (def->type == ARG_TYPE_INT || def->type == ARG_TYPE_LONG_LONG) {
                     def->defaults.int_default = va_arg(args, long long);
                     def->has_default = true;
                 }
@@ -775,9 +775,9 @@ ArgParser *ArgParser_AddIntV(ArgParser *parser, const char *name, const char *de
     return parser;
 }
 
-ArgParser *ArgParser_AddLongV(ArgParser *parser, const char *name, const char *description,
+ArgParser *ArgParser_AddLongLongV(ArgParser *parser, const char *name, const char *description,
                              long long *target, ...) {
-    ArgParser_AddLong(parser, name, description, target);
+    ArgParser_AddLongLong(parser, name, description, target);
 
     va_list args;
     va_start(args, target);
@@ -787,9 +787,9 @@ ArgParser *ArgParser_AddLongV(ArgParser *parser, const char *name, const char *d
     return parser;
 }
 
-ArgParser *ArgParser_AddULongV(ArgParser *parser, const char *name, const char *description,
+ArgParser *ArgParser_AddULongLongV(ArgParser *parser, const char *name, const char *description,
                               unsigned long long *target, ...) {
-    ArgParser_AddULong(parser, name, description, target);
+    ArgParser_AddULongLong(parser, name, description, target);
 
     va_list args;
     va_start(args, target);
--- a/src/util/arg_parser.c
+++ b/src/util/arg_parser.c
@@ -1,8 +1,11 @@
 /*
- * Copyright Redis Ltd. 2016 - present
- * Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
- * the Server Side Public License v1 (SSPLv1).
- */
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
 
 #include "arg_parser.h"
 #include "rmalloc.h"
--- a/src/util/arg_parser.c
+++ b/src/util/arg_parser.c
@@ -1,0 +1,838 @@
+/*
+ * Copyright Redis Ltd. 2016 - present
+ * Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
+ * the Server Side Public License v1 (SSPLv1).
+ */
+
+#include "arg_parser.h"
+#include "rmalloc.h"
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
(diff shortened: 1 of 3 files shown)


src/util/ARG_PARSER_README.md
```diff
@@ -0,0 +1,369 @@
...
+- **Maintainability**: Adding new arguments requires minimal code changes
+- **Reliability**: Built-in type validation and comprehensive error reporting
+- **Flexibility**: Supports named arguments, positional arguments, bitflags, and custom validators
+- **Performance**: Efficient bitflag operations for setting multiple flags
+
+## Files
+
+- `src/util/arg_parser.h` - Header file with API definitions
+- `src/util/arg_parser.c` - Implementation
+- `src/hybrid/parse/hybrid_optional_args.c` - Real-world usage examples
```
JoanFM: does this really exist? is it in another branch?

src/util/ARG_PARSER_README.md
```diff
@@ -0,0 +1,369 @@
...
+#include "util/arg_parser.h"
+
+// Define variables to hold parsed values
+bool verbose = false;
+long long timeout = 5000;
+const char *format = NULL;
+int flags = 0;  // For bitflag operations
+
+// Create parser
+ArgParser *parser = ArgParser_New(&cursor, "MYCOMMAND");
```
JoanFM: what would `cursor` be here?

JoanFM: Would be nice to see (if possible) how to go from string to cursor, to arguments with its types

src/util/arg_parser.h
```diff
@@ -0,0 +1,206 @@
...
+typedef struct {
+    const char *name;           // Argument name (e.g., "LIMIT", "TIMEOUT")
+    const char *description;    // Help description
+    ArgType type;               // Argument type
+    void *target;               // Pointer to store parsed value
+    bool required;              // Whether argument is required
+    bool repeatable;            // Whether argument can appear multiple times
+
+    // Positional constraint (1-based). If set, this argument is parsed by position instead of name.
+    int position;               // 1 = first argument after command, 2 = second, etc.
```
JoanFM: should it be unsigned? or does negative position make sense?

src/util/arg_parser.c
```diff
@@ -0,0 +1,737 @@
...
+    }
+
+    rm_free(parser->definitions);
+    rm_free(parser->parsed_flags);
+    rm_free((void*)parser->command_name);
+    rm_free(parser);
+}
+
+static int resize_definitions(ArgParser *parser) {
+    size_t new_capacity = parser->def_count * 2;
```
JoanFM: Cursor is right, 

you are always reallocating here regardless of the current capacity. (Should keep capacity and used) info to track the array

src/util/arg_parser.h
```diff
@@ -0,0 +1,206 @@
...
+    const char *error_message;
+    const char *error_arg;      // Which argument caused the error
+    int error_position;         // Position in argument list where error occurred
+};
+
+// Main parser structure
+struct ArgParser {
+    ArgsCursor *cursor;         // Underlying cursor
+    ArgDefinition *definitions; // Array of argument definitions
+    size_t def_count;          // Number of definitions
```
JoanFM: we need extra member to keep track of capacity of definitions

JoanFM (changes requested):

JoanFM (changes requested):
I miss the presence of unit tests to prove/document the usage

src/util/arg_parser.c
```diff
@@ -0,0 +1,737 @@
...
+}
+
+// Internal helper to access the last added definition (used by variadic option applier)
+static ArgDefinition *get_last_definition(ArgParser *parser) {
+    if (parser->def_count == 0) return NULL;
+    return &parser->definitions[parser->def_count - 1];
+}
+
+static ArgDefinition *find_definition(ArgParser *parser, const char *name) {
+    for (size_t i = 0; i < parser->def_count; i++) {
```
JoanFM: can we maybe build a better struct to optimize this find operation? seems quite a common operation

src/util/arg_parser.c
```diff
@@ -0,0 +1,737 @@
...
+        }
+
+        case ARG_TYPE_STRING: {
+            const char *str_val;
+            rv = AC_GetString(parser->cursor, &str_val, NULL, 0);
+            if (rv == AC_OK && def->target) {
+                // Validate against allowed values if specified
+                if (def->options.string.allowed_values) {
+                    bool found = false;
+                    for (const char **allowed = def->options.string.allowed_values; *allowed; allowed++) {
```
JoanFM, line 251: how are we guaranteeing that `allowed_values` is indeed NULL terminated?

src/util/ARG_PARSER_README.md
kei-nan: yep, in a future PR. will remove it from documentation for now.

src/util/ARG_PARSER_README.md
kei-nan: will show how ArgsCursor is initialized from an argv and argc.

src/util/arg_parser.h
kei-nan: changed to uint16_t

src/util/arg_parser.c
kei-nan: moved to using arrayof api.

src/util/arg_parser.h
kei-nan: moved to using arrayof api.

src/util/arg_parser.c
kei-nan: I think because we will mostly have few definitions it should be fine.
Don't see a real need for a hash table or trie in this specific case.

src/util/arg_parser.c
kei-nan: I think no matter how we try to enforce it there will always be a level of trust.
Even if a count will be provided we have to trust the count the developer provided is correct.
I can maybe ask for an arrayof(const char*)
but that would mean further allocations for the string array.
Let me know what you prefer.

src/util/arg_parser.c
JoanFM: perfect

src/util/arg_parser.c
JoanFM: Keep it as it is. Thanks for the explanation

src/util/arg_parser.c
```diff
@@ -0,0 +1,845 @@
...
+#include <stdio.h>
+#include <stdarg.h>
+#include <assert.h>
+
+#define INITIAL_DEF_CAPACITY 16
+#define MAX_ERROR_MSG_LEN 512
+
+// Internal helper functions
+static ArgDefinition *find_definition(ArgParser *parser, const char *name);
+static ArgDefinition *find_positional_definition(ArgParser *parser, int position);
```
JoanFM: now position should also be `uint16_t`.

JoanFM (changes requested):
one last comment

src/util/arg_parser.c
```diff
@@ -14,10 +14,11 @@
 
 #define INITIAL_DEF_CAPACITY 16
 #define MAX_ERROR_MSG_LEN 512
+#define MAX_POSITIONAL_ARGS 20 // reasonable limit for number of expected positional arguments
```
JoanFM, line 20: a query itself is an argument? because otherwise queries may contain many many fields

src/util/arg_parser.c
```diff
@@ -0,0 +1,822 @@
+/*
```
nafraf, line 1: Is this the right license?

src/util/arg_parser.h
```diff
@@ -0,0 +1,199 @@
+/*
```
nafraf, line 1: outdated license?

src/util/arg_parser.c
```diff
@@ -0,0 +1,822 @@
...
+                    set_error(parser, "Value above maximum", def->name);
+                    return AC_ERR_ELIMIT;
+                }
+                *(int*)def->target = int_val;
+            }
+            break;
+        }
+
+        case ARG_TYPE_LONG: {
+            long long long_val;
```
nafraf: The type is named `ARG_TYPE_LONG` but it supports `long long`, it's a bit confusing.

src/util/arg_parser.c
```diff
@@ -0,0 +1,822 @@
...
+                if (def->type == ARG_TYPE_INT || def->type == ARG_TYPE_LONG ||
+                    def->type == ARG_TYPE_ULONG) {  // Removed ARG_TYPE_UINT (doesn't exist)
+                    def->options.numeric.min_val = va_arg(args, long long);
+                    def->options.numeric.max_val = va_arg(args, long long);
+                    def->options.numeric.has_min = true;
+                    def->options.numeric.has_max = true;
+                }
+                break;
+
+            case ARG_OPT_ALLOWED_VALUES:
```
nafraf, line 676: Please consider adding a test for this option: `ARG_OPT_ALLOWED_VALUES`

src/util/arg_parser.c
kei-nan: yes, a long query is wrapped in quotes or double quotes.

src/util/arg_parser.c
kei-nan: nice catch, update the license.

src/util/arg_parser.c
kei-nan: ah, true
missed that
will fix

src/util/arg_parser.c
```diff
@@ -0,0 +1,824 @@
...
+        }
+
+        // Advance the cursor to the argument value
+        rv = AC_Advance(parser->cursor);
+        if (rv != AC_OK) {
+            set_error(parser, "Failed to parse past", pos_def->name);
+            break;
+        }
+        // Parse the positional argument value directly (no name expected)
+        rv = parse_single_arg(parser, pos_def);
```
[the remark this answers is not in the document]
kei-nan: I don't think this is a bug.
if we have "COMBINE" and it is a positional argument then we find it using the "COMBINE" string.
We want to advance past that string in case of a match.

src/util/arg_parser.c
nafraf: I see you updated the license in `arg_parser.h`, but in `arg_parser.c` it seems outdated

nafraf (dismissed):

src/util/arg_parser.c
```diff
@@ -0,0 +1,838 @@
...
```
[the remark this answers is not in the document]
JoanFM: what about this report @kei-nan ?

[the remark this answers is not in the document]
JoanFM: I checked the code and could not find what it means, but maybe you have some comment about it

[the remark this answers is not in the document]
JoanFM: Augment came up with this test to prove a potential bug:

```cpp
TEST_F(ArgParserTest, PositionalArgumentBugDemo) {
    // This test demonstrates the bug where positional arguments get skipped values
    // due to incorrect cursor advancement
    SetupCustomArgs({"COMMAND", "first_value", "second_value", "third_value"});

    const char *first_arg = nullptr;
    const char *second_arg = nullptr;
    const char *third_arg = nullptr;

    // Add three positional arguments
    ArgParser_AddStringV(parser, "FIRST", "First positional argument", &first_arg,
                         ARG_OPT_REQUIRED,
                         ARG_OPT_POSITION, 1,
                         ARG_OPT_END);

    ArgParser_AddStringV(parser, "SECOND", "Second positional argument", &second_arg,
                         ARG_OPT_REQUIRED,
                         ARG_OPT_POSITION, 2,
                         ARG_OPT_END);

    ArgParser_AddStringV(parser, "THIRD", "Third positional argument", &third_arg,
                         ARG_OPT_REQUIRED,
                         ARG_OPT_POSITION, 3,
                         ARG_OPT_END);

    ArgParseResult result = ArgParser_Parse(parser);
    ASSERT_TRUE(result.success) << "Parse failed: " << ArgParser_GetErrorString(parser);

    // BUG: Due to incorrect cursor advancement, these assertions will fail
    // Expected: first_arg="first_value", second_arg="second_value", third_arg="third_value"
    // Actual: first_arg="second_value", second_arg="third_value", third_arg=nullptr (or error)
    ASSERT_STREQ(first_arg, "first_value") << "First argument should be 'first_value'";
    ASSERT_STREQ(second_arg, "second_value") << "Second argument should be 'second_value'";
    ASSERT_STREQ(third_arg, "third_value") << "Third argument should be 'third_value'";
}

```
With this explanatioN:


This test will fail because:

1. The cursor starts at "first_value"
2. AC_Advance() moves it to "second_value"
3. parse_single_arg() reads "second_value" for the first positional argument
4. The cursor advances to "third_value"
5. AC_Advance() moves it past "third_value"
6. parse_single_arg() tries to read beyond the end or gets garbage

The test expects each positional argument to get its corresponding value, but due to the bug, each argument gets the next value in the sequence.

JoanFM (changes requested):
I would like to clarify the potential bug found by Cursor reviewer

src/util/arg_parser.c
```diff
@@ -0,0 +1,824 @@
...
+            if (pos_def) {
+                // Advance past the argument name
+                rv = AC_Advance(parser->cursor);
+                if (rv != AC_OK) {
+                    set_error(parser, "Failed to parse past", pos_def->name);
+                    break;
+                }
+                // Parse as positional argument
+                size_t def_index = pos_def - parser->definitions;
+                rv = parse_single_arg(parser, pos_def);
```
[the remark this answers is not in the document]
kei-nan: When the code was written by augment it assumed positional arguments and their values are different.
This is reflected in the comment by cursor.

if I define an argument COMBINE at position 1.
And my 1st argument is hello
the value hello would have gone into the COMBINE argument.

I changed it so the first string has to be COMBINE and the second one is the value.
so instead of "helllo"
it should be "COMBINE hello:

because we expect the fist argument to match the positional argument name we need to advance the argument cursor.
Which is what cursor commented on.

Hope that clarifies things

kei-nan:
> I would like to clarify the potential bug found by Cursor reviewer

@JoanFM replied to the cursor comment in:
https://github.com/RediSearch/RediSearch/pull/6856#discussion_r2371398343

JoanFM (dismissed):

src/util/arg_parser.c
```diff
@@ -0,0 +1,826 @@
...
+        ArgDefinition *def = &parser->definitions[i];
+        if (def->has_position && def->position == position && (name == NULL || strcasecmp(def->name, name) == 0)) {
+            return def;
+        }
+    }
+    return NULL;
+}
+
+
+
```
Itzikvaknin, line 204:
```suggestion

```

Itzikvaknin: 3 empty lines -> 1 empty line

tests/cpptests/test_cpp_arg_parser.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+    ArgParser_AddStringV(parser, "FORMAT", "Output format", &format,
+                         ARG_OPT_OPTIONAL,
+                         ARG_OPT_ALLOWED_VALUES, allowed_formats,
+                         ARG_OPT_END);
+
+    ArgParseResult result = ArgParser_Parse(parser);
+    ASSERT_FALSE(result.success) << "Parse should fail for invalid value";
+    ASSERT_STREQ(result.error_arg, "FORMAT");
+}
```
Itzikvaknin, line 473: newline EOF

src/util/arg_parser.c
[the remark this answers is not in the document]
Itzikvaknin: Resolved? @kei-nan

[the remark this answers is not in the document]
kei-nan: The test is incorrect it should be 
    SetupCustomArgs({"COMMAND", "FIRST", "first_value", "SECOND", "second_value", "THIRD", "third_value"});
The meaning of a positional arguments is a bit different than the regular positional arguments.
In hybrid command we have the COMBINE keyword and if it appears it must be first.
So we needed to support this scenario where it would give an error it it appeared anywhere else but the start.

PR #6857, opened 2025-09-18
MOD-11172: Refactor Hybrid Parsing

kei-nan:
This PR aims to refactor the hybrid parsing by using the newly added ArgsParser.
It also aims to separate the hybrid parsing flow from the aggregation or search flows.

1. Current: Parsing of HYBRID is accepting keywords it shouldn't from the Aggregate and Search worlds.
2. Change: Separate the hybrid parsing and use ArgsParser to make our lives a bit easier.
3. Outcome: Better and more maintainable parsing code which does not depend on the aggregation/search keywords. 

1. Hybrid Parsing

#### Mark if applicable


src/hybrid/parse_hybrid.c
```diff
@@ -755,29 +609,10 @@ int parseHybridCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc,
   AGPLN_AddStep(&vectorRequest->pipeline.ap, &vnStep->base);
 
   // Save the current position to determine remaining arguments for the merge part
```
nafraf: Please consider updating this comment or removing it

src/hybrid/parse/hybrid_optional_args.h
```diff
@@ -0,0 +1,67 @@
...
+    SPECIFIED_ARG_LIMIT,
+    SPECIFIED_ARG_SORTBY,
+    SPECIFIED_ARG_WITHCURSOR,
+    SPECIFIED_ARG_PARAMS,
+    SPECIFIED_ARG_DIALECT,
+    SPECIFIED_ARG_FORMAT,
+    SPECIFIED_ARG_WITHSCORES,
+    SPECIFIED_ARG_EXPLAINSCORE,
+    SPECIFIED_ARG_COMBINE,
+} SpecifiedArg;
```
nafraf, line 37: To use it as bitmask, you need to specify the values, don't you?

src/hybrid/parse/hybrid_optional_args.h
```diff
@@ -0,0 +1,67 @@
...
+ * 
+ * This function handles arguments like:
+ * - LIMIT offset count
+ * - SORTBY field [ASC|DESC] [field [ASC|DESC] ...]
+ * - WITHCURSOR [COUNT count] [MAXIDLE maxidle]
+ * - PARAMS param value [param value ...]
+ * - TIMEOUT timeout
+ * - DIALECT dialect
+ * - FORMAT format
+ * - WITHSCORES
```
nafraf, line 68: Please review WITHSCORES support. `FT.AGGREGATE -> ADDSCORES`, `FT.SEARCH -> WITHSCORES`

src/hybrid/parse/hybrid_optional_args.h
```diff
@@ -0,0 +1,67 @@
...
+ * - EXPLAINSCORE  
+ * - COMBINE [RRF [K k] [WINDOW window]] | [LINEAR weight1 weight2 ...]
```
nafraf, line 70: Please consider updating the description. For LINEAR syntax, we now supports `LINEAR 4 ALPHA 0.7 BETA 0.3` as done in MOD-11258.

src/hybrid/parse/hybrid_optional_args.c
```diff
@@ -0,0 +1,130 @@
...
+    static const char *allowedFormats[] = {"STRING", NULL};
+    ArgParser_AddStringV(parser, "FORMAT", "Output format",
+                         &formatTarget, 1, 1,
+                         ARG_OPT_OPTIONAL,
+                         ARG_OPT_ALLOWED_VALUES, allowedFormats,
+                         ARG_OPT_CALLBACK, handleFormat, ctx,
+                         ARG_OPT_END);
+
+    // WITHSCORES flag - sets QEXEC_F_SEND_SCORES
+    ArgParser_AddBitflagV(parser, "WITHSCORES", "Include scores in results",
```
nafraf: Check this, it seems `WITHSCORES` is only supported by `FT.SEARCH`

src/hybrid/parse/hybrid_callbacks.c
```diff
@@ -0,0 +1,214 @@
+#include "hybrid_callbacks.h"
```
nafraf, line 9: license?

src/hybrid/parse/hybrid_combine.c
```diff
@@ -0,0 +1,144 @@
+#include "hybrid_callbacks.h"
```
nafraf, line 9: license?

src/util/arg_parser.c
```diff
@@ -295,9 +295,9 @@ static int parse_single_arg(ArgParser *parser, ArgDefinition *def) {
         }
 
         case ARG_TYPE_SUBARGS: {
-            if (def->options.subargs.max_args == 1) {
+            if (def->options.subargs.max_args == 1 || def->options.subargs.min_args == def->options.subargs.max_args) {
```
nafraf: Please review the license of this file, it seems it was added with the old license.

src/hybrid/parse/hybrid_callbacks.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
```
nafraf, line 9: license?

src/hybrid/parse/hybrid_optional_args.c
```diff
@@ -0,0 +1,130 @@
+#include "hybrid/parse/hybrid_optional_args.h"
```
nafraf, line 9: license?

src/hybrid/parse/hybrid_optional_args.h
```diff
@@ -0,0 +1,67 @@
+#pragma once
```
nafraf, line 9: license?

src/hybrid/parse/hybrid_optional_args.c
```diff
@@ -0,0 +1,130 @@
...
+                         ARG_OPT_OPTIONAL,
+                         ARG_OPT_CALLBACK, handleParams, ctx,
+                         ARG_OPT_END);
+
+    // TIMEOUT timeout - query timeout in milliseconds
+    // TIMEOUT timeout - query timeout in milliseconds
+    // Note: We need to use a temporary int and copy to long long later
+    // since ArgParser_AddIntV expects int*, not long long*
+    static int tempTimeout = 0;
+    ArgParser_AddIntV(parser, "TIMEOUT", "Query timeout in milliseconds",
```
nafraf: Can't we use `ArgParser_AddLong` ?

kei-nan: `ctx->reqConfig->queryTimeoutMS`
is long long.
Added a callback to assign the int into the `ctx->reqConfig->queryTimeoutMS` member.

src/hybrid/parse/hybrid_optional_args.c
kei-nan: Yep, we need the scores to be sent in hybrid.
So we want to support this keyword.
Originally was meant to be added in the next PR:
https://github.com/RediSearch/RediSearch/pull/6858
But the change wound up here.
If you prefer I can move it to the next PR.

src/hybrid/parse/hybrid_optional_args.h
kei-nan: yep, you are correct.
will fix

src/hybrid/parse/hybrid_optional_args.h
kei-nan: We need the scores to be sent in the hybrid cursor.
WITHSCORES made the most sense at the time.
I can change it so it only gets parsed for the internal command, that way we can change it in the future if we want to.
Let me know what you prefer.

nafraf: If it doesn't require too much effort, I'd like to allow `WITHSCORES`only for the internal command.

src/hybrid/parse/hybrid_optional_args.h
```diff
@@ -0,0 +1,79 @@
...
+    SpecifiedArg specifiedArgs;             // Bitmask of specified arguments
+    HybridScoringContext *hybridScoringCtx; // Hybrid scoring context for COMBINE
+    size_t numSubqueries;                   // Number of subqueries for weight validation
+
+    AGGPlan *plan;                          // Aggregate plan for LIMIT/SORTBY
+    RSSearchOptions *searchopts;            // Search options for PARAMS
+    CursorConfig *cursorConfig;             // Cursor configuration
+    RequestConfig *reqConfig;               // Request configuration for DIALECT/TIMEOUT
+    QEFlags *reqFlags;                      // Request flags
+    size_t *maxResults;                     // Maximum results
```
JoanFM, line 54: I think it would be better to have simply a `size_t` and not a pointer

src/hybrid/parse/hybrid_optional_args.c
```diff
@@ -0,0 +1,177 @@
...
+    ArgParser_Free(parser);
+
+    // Handle dialect-specific validation (replicated from original)
+    if (ctx->specifiedArgs & SPECIFIED_ARG_DIALECT && ctx->reqConfig->dialectVersion < APIVERSION_RETURN_MULTI_CMP_FIRST &&
+        (*(ctx->reqFlags) & QEXEC_F_SEND_SCOREEXPLAIN)) {
+        QueryError_SetError(status, QUERY_EPARSEARGS, "EXPLAINSCORE is not supported in this dialect version");
+        return REDISMODULE_ERR;
+    }
+
+      if (!(*ctx->reqFlags & QEXEC_F_SEND_HIGHLIGHT) &&
```
JoanFM: Can we give names to the booleans to better understand what activates this option? perhaps a small static function with relevant name can help with readibility

src/hybrid/parse/hybrid_combine.c
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  // Store the parsed values
+  linearCtx->linearWeights[0] = alphaValue;
+  linearCtx->linearWeights[1] = betaValue;
+
+  ArgParser_Free(parser);
+}
+
+int parseRRFArgs(ArgsCursor *ac, double *constant, int *window, bool *hasExplicitWindow, QueryError *status) {
```
JoanFM: should it be static?

src/hybrid/parse/hybrid_combine.c
```diff
@@ -0,0 +1,185 @@
...
+  
+  // Store the parsed values
+  rrfCtx->constant = constantValue;
+  rrfCtx->window = windowValue;
+  rrfCtx->hasExplicitWindow = hasExplicitWindow;
+
+}
+
+// COMBINE callback - implements exact ParseCombine behavior from hybrid_args.c
+void handleCombine(ArgParser *parser, const void *value, void *user_data) {
```
JoanFM, line 155: this function has different indentation

JoanFM (changes requested):

src/hybrid/parse/hybrid_callbacks.c
```diff
@@ -0,0 +1,404 @@
...
+    ArgsCursor *ac = (ArgsCursor*)value;
+    QueryError *status = ctx->status;
+    ctx->specifiedArgs |= SPECIFIED_ARG_LIMIT;
+
+    if (AC_NumRemaining(ac) < 2) {
+        QueryError_SetError(status, QUERY_EPARSEARGS, "LIMIT requires 2 arguments");
+        return;
+    }
+
+    long long offset, num;
```
JoanFM: shouldn't offset be unsigned? is there such a way to limit?

src/hybrid/parse/hybrid_callbacks.c
```diff
@@ -0,0 +1,404 @@
...
+            return;
+        }
+
+        // Remove '@' prefix if present (same logic as parseSortby)
+        if (*field == '@') {
+            field++;  // Skip the '@' prefix
+        }
+
+        // Default to ascending
+        int ascending = 1;
```
JoanFM: it may be nice to have some definition "DEFAULT_SORTBY_BEHAVIOR" and ASCENDING as 1 and DESC as 0.

src/hybrid/parse/hybrid_callbacks.c
```diff
@@ -0,0 +1,404 @@
...
+    HybridParseContext *ctx = (HybridParseContext*)user_data;
+    ctx->specifiedArgs |= SPECIFIED_ARG_GROUPBY;
+    ArgsCursor *ac = (ArgsCursor*)value;
+    QueryError *status = ctx->status;
+
+    if (!ensureExtendedMode(ctx->reqFlags, "GROUPBY", status)) {
+        return;
+    }
+
+    const long long nproperties = AC_NumRemaining(ac);
```
JoanFM, line 253: would it be better to have unsigned? Should we protect against potential overflows?

src/hybrid/parse/hybrid_callbacks.c
```diff
@@ -0,0 +1,404 @@
...
+    size_t exprLen = strlen(expr);
+
+    HiddenString* expression = NewHiddenString(expr, exprLen, false);
+    PLN_MapFilterStep *stp = PLNMapFilterStep_New(expression, PLN_T_FILTER);
+    HiddenString_Free(expression, false);
+    AGPLN_AddStep(ctx->plan, &stp->base);
+}
+
+// TIMEOUT callback - implements EXACT original logic from handleTimeout
+void handleTimeout(ArgParser *parser, const void *value, void *user_data) {
```
JoanFM, line 394: why are there no actual implementation here? not yet implemented? is it a known limitation or debt?

src/hybrid/parse/hybrid_callbacks.c
kei-nan: true, changed to uint64_t
I think we later limit the number of results but type safety is important

src/hybrid/parse/hybrid_callbacks.c
kei-nan: I changed the variable to be bool.
Added ASC_BY_DEFAULT macro in the c file to be true.

src/hybrid/parse/hybrid_callbacks.c
kei-nan: if we make it unsigned and have an overflow we will probably crash trying to allocate the array or during the loop..
I'll check the number of properties is positive.

kei-nan: Added:
```
    const long long nproperties = AC_NumRemaining(ac);
    if (nproperties <= 0) {
        QueryError_SetWithoutUserDataFmt(status, QUERY_EPARSEARGS, "Bad arguments for GROUPBY: Expected at least one property", ", got: %d", nproperties);
        return;
    }
```

src/hybrid/parse/hybrid_callbacks.c
kei-nan: we use the callback functions here only for signalling which arguments were specified.
The assignment of the values already took place by the parser.

src/hybrid/parse/hybrid_optional_args.c
kei-nan: moved to `applyRichResultsOptimization` function

src/hybrid/parse/hybrid_optional_args.h
kei-nan: Then the caller would have to do a copy from this maxResult to the actual maxResults.
Giving the level of indirection simplifies the life of the user.

src/hybrid/parse/hybrid_optional_args.h
kei-nan: np, added a bool internal, if it is true then adding the withscores argument so it can be parsed.
changed the code paths to pass the bool the parse hybrid function

src/hybrid/parse/hybrid_combine.c
```diff
@@ -0,0 +1,185 @@
...
+    return AC_ERR_PARSE;
+  }
+
+  double defaultConstant = HYBRID_DEFAULT_RRF_CONSTANT;
+  // Define the optional arguments with validation
+  ArgParser_AddDoubleV(parser, "CONSTANT", "RRF constant value (must be positive)", 
+                       constant, ARG_OPT_OPTIONAL,
+                       ARG_OPT_DEFAULT_DOUBLE, defaultConstant,
+                       ARG_OPT_RANGE, 0.0, LLONG_MAX,
+                       ARG_OPT_END);
```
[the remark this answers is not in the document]
kei-nan: TBH, Understood from an internal discussion we don't have a real reason to cap the value with a range.
Will remove the range validation

src/hybrid/parse/hybrid_optional_args.h
JoanFM: I think it pollutes all the code as it should check for NuLl right? plus eventually u keep copying pointer addresses around? Anyway, lets keep it

src/util/arg_parser.c
```diff
@@ -321,12 +321,23 @@ static int parse_single_arg(ArgParser *parser, ArgDefinition *def) {
         }
 
         case ARG_TYPE_SUBARGS: {
+            unsigned int count = 0;
+            const char *notEnoughArgMessage = NULL;
             if (def->options.subargs.max_args && def->options.subargs.min_args == def->options.subargs.max_args) {
-                // Single argument slice
-                rv = AC_GetSlice(parser->cursor, (ArgsCursor*)def->target, def->options.subargs.max_args);
+                count = def->options.subargs.max_args;
+                notEnoughArgMessage = "Not enough arguments were provided";
```
JoanFM, line 331: we could add info about how many are expected? and perhaps a hint about the config imposing that?

src/util/arg_parser.c
```diff
@@ -321,12 +321,23 @@ static int parse_single_arg(ArgParser *parser, ArgDefinition *def) {
...
+            } else {
+                rv = AC_GetUnsigned(parser->cursor, &count, 0);
+                notEnoughArgMessage = "Not enough arguments were provided based on argument count";
```
JoanFM, line 334: Some more details here perhaps?

src/util/arg_parser.c
kei-nan: it would require allocating a string in memory, maybe we can do it in a different PR?
it would require arg_parser to allocate deallocate error message strings which can complicate its logic a bit.

src/util/arg_parser.c
kei-nan: What did you have in mind?
if we want a dynamic string then maybe it would be better for that to go in a different PR.

src/util/arg_parser.c
JoanFM: we can do in different PR sure

src/util/arg_parser.c
JoanFM: different PR is good

JoanFM (approved):

nafraf (approved):

Issue #6084, opened 2025-05-05
[Feature Request] Exclude keys from FT.Search


Aloune92:
Would it be possible to implement a `NOTINKEYS` option for FT.Search to exclude keys from search ?

RushabhMehta2005:
I've reviewed this feature request for a NOTINKEYS option in FT.SEARCH and I would really like to work on implementing it.
Is this feature still desired? If so, and if no one else is currently working on it, please feel free to assign this issue to me.

RushabhMehta2005:
@oshadmi Could you please take a look at this? Sorry to ping.

Aloune92:
@RushabhMehta2005 it would be really appreciated :)

RushabhMehta2005:
Thanks for the reply, I would surely want to implement this myself, however I need to confirm this first with a maintainer. If they do not think this feature is worth including, my efforts may be wasted.

RushabhMehta2005:
You assigned this issue to another user with the name adrianomaral, are they not working on this anymore? Just clarifying.

Aloune92:
I didn't manually assign to anyone. A lot of feature requests are stale in this repo without any answer but it looks like adrianomaral is still active.

RushabhMehta2005:
Alright, I will look into this then open a new PR I think.

PR #6864, opened 2025-09-20
feat(search): Add NOTINKEYS option to FT.SEARCH

RushabhMehta2005:
## Basic Idea

Currently, RedisSearch provides an INKEYS option to restrict a search to a specific list of documents. However, there is no efficient way to perform the inverse operation: searching an entire index while excluding a small, specific subset of documents. The current workaround is to fetch all keys, filter them in the client which is highly impractical and inefficient.
This proposal introduces a NOTINKEYS option to solve this problem directly and performantly.

I am only looking for advice on whether the community thinks this problem is worth solving, as I would like to implement it if that is the case.

1. [#6084](https://github.com/RediSearch/RediSearch/issues/6084)


RushabhMehta2005:
@gdesmott Hey, this PR has been sitting for a while and I was wondering if it is inline with RedisSearch's priority at this time. Please take a moment to consider it when your time permits. Thank you.

PR #6834, opened 2025-09-16
[MOD-11407] Lower cross-language performance penalty for `RSIndexResult`

LukeMathWalker:
## Describe the changes in the pull request

Fix the performance regression caused by https://github.com/RediSearch/RediSearch/pull/6492.
In particular, avoid redundant checks across languages and minimize C->Rust FFI hops.

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #6872, opened 2025-09-22
Have Rust decode to a mut reference `RSIndexResult`

chesedo:
The iterators set some values on the `RSIndexResult` which is expected to persist between decoded records. The old approach (while being more idiomatic Rust) caused these values to get lost when a new record was decoded since each decode made a new record instance.

This PR updates the decoding to rather receive a mut ref `RSIndexResult` to decode unto and thus keeping these values between decodes.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
+++ b/src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
@@ -20,7 +20,7 @@ use ffi::{
 
 use inverted_index::{
     EntriesTrackingIndex, FieldMaskTrackingIndex, FilterGeoReader, FilterMaskReader,
-    FilterNumericReader, IndexBlock, NumericFilter, RSIndexResult, ReadFilter,
+    FilterNumericReader, IndexBlock, IndexReader as _, NumericFilter, RSIndexResult, ReadFilter,
     debug::{BlockSummary, Summary},
     doc_ids_only::DocIdsOnly,
     fields_offsets::{FieldsOffsets, FieldsOffsetsWide},
@@ -768,7 +768,11 @@ pub unsafe extern "C" fn IndexReader_HasSeeker(_ir: *const IndexReader) -> bool 
 pub unsafe extern "C" fn IndexReader_Next<'index, 'filter>(
     ir: *mut IndexReader<'index, 'filter>,
     res: *mut RSIndexResult<'index>,
-) -> bool {
+) -> bool
+where
+    'index: 'filter,
+    'filter: 'index,
+{
     debug_assert!(!ir.is_null(), "ir must not be null");
     debug_assert!(!res.is_null(), "res must not be null");
 
@@ -778,14 +782,7 @@ pub unsafe extern "C" fn IndexReader_Next<'index, 'filter>(
     // SAFETY: The caller must ensure that `ir` is a valid pointer to an `IndexReader`
     let res = unsafe { &mut *res };
 
-    match ir_dispatch!(ir, next) {
-        Some(new_res) => {
-            *res = new_res;
-
-            true
-        }
-        None => false,
-    }
+    ir_dispatch!(ir, next_record, res).unwrap_or_default()
 }
 
 /// Skip the internal block of the inverted index reader to the block that may contain the given
@@ -822,7 +819,11 @@ pub unsafe extern "C" fn IndexReader_Seek<'index, 'filter>(
     ir: *mut IndexReader<'index, 'filter>,
     doc_id: t_docId,
     res: *mut RSIndexResult<'index>,
-) -> bool {
+) -> bool
+where
+    'index: 'filter,
+    'filter: 'index,
+{
     debug_assert!(!ir.is_null(), "ir must not be null");
     debug_assert!(!res.is_null(), "res must not be null");
 
@@ -832,15 +833,7 @@ pub unsafe extern "C" fn IndexReader_Seek<'index, 'filter>(
     // SAFETY: The caller must ensure that `ir` is a valid pointer to an `IndexReader`
     let res = unsafe { &mut *res };
 
-    match ir_dispatch!(ir, seek_record, doc_id) {
-        Ok(Some(new_res)) => {
-            *res = new_res;
-
-            true
-        }
-        Ok(None) => false,
-        Err(_) => false,
-    }
+    ir_dispatch!(ir, seek_record, doc_id, res).unwrap_or_default()
 }
 
 /// Check if the index reader can return multiple entries for the same document ID.
--- a/src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
+++ b/src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
@@ -515,33 +515,37 @@ pub unsafe extern "C" fn InvertedIndex_LastId(ii: *const InvertedIndex) -> t_doc
 /// single interface for all index reader types while still being able to optimize the storage
 /// and performance for each index reader type.
 pub enum IndexReader<'index, 'filter> {
-    Full(FilterMaskReader<inverted_index::IndexReader<'index, Full, Full>>),
-    FullWide(FilterMaskReader<inverted_index::IndexReader<'index, FullWide, FullWide>>),
-    FreqsFields(FilterMaskReader<inverted_index::IndexReader<'index, FreqsFields, FreqsFields>>),
-    FreqsFieldsWide(
-        FilterMaskReader<inverted_index::IndexReader<'index, FreqsFieldsWide, FreqsFieldsWide>>,
+    Full(FilterMaskReader<inverted_index::IndexReaderCore<'index, Full, Full>>),
+    FullWide(FilterMaskReader<inverted_index::IndexReaderCore<'index, FullWide, FullWide>>),
+    FreqsFields(
+        FilterMaskReader<inverted_index::IndexReaderCore<'index, FreqsFields, FreqsFields>>,
     ),
-    FreqsOnly(inverted_index::IndexReader<'index, FreqsOnly, FreqsOnly>),
-    FieldsOnly(FilterMaskReader<inverted_index::IndexReader<'index, FieldsOnly, FieldsOnly>>),
+    FreqsFieldsWide(
+        FilterMaskReader<inverted_index::IndexReaderCore<'index, FreqsFieldsWide, FreqsFieldsWide>>,
+    ),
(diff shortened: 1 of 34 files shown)


src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -515,33 +515,37 @@ pub unsafe extern "C" fn InvertedIndex_LastId(ii: *const InvertedIndex) -> t_doc
 /// single interface for all index reader types while still being able to optimize the storage
 /// and performance for each index reader type.
 pub enum IndexReader<'index, 'filter> {
```
LukeMathWalker: I think it'd be simpler (and equally suitable) to boil this down to a single `'index` lifetime parameter.

GlenDC: in that case might want to rename it to just `'a` or so

src/redisearch_rs/inverted_index/src/freqs_only.rs
```diff
@@ -46,13 +46,13 @@ impl Decoder for FreqsOnly {
...
+        result: &mut RSIndexResult<'index>,
+    ) -> std::io::Result<()> {
         let (decoded_values, _bytes_consumed) = qint_decode::<2, _>(cursor)?;
         let [delta, freq] = decoded_values;
 
-        let record = RSIndexResult::virt()
-            .doc_id(base + delta as t_docId)
-            .frequency(freq);
-        Ok(record)
+        result.doc_id = base + delta as t_docId;
```
GlenDC, line 54: This comments applies to the overal PR, but this seems like a good location as any to ask.

It feels dangerous to use this approach, assigning public properties to mutable references like this. I wonder if there will never a point in time (if not already now) where it is possible to introduce impossible variants like this. Unless you typesystem-wise have a solid set of properties that are impossible to get wrong coordination wise.

That's the beauty of returning a result is that you have full control over the value without having to worry about possible invariants due to partial assignments (previous state)...

src/redisearch_rs/inverted_index/src/index_result.rs
```diff
@@ -800,6 +883,35 @@ impl<'index> RSIndexResult<'index> {
...
+        match &mut self.data {
+            RSResultData::Term(term) => term,
+            RSResultData::Union(_)
+            | RSResultData::Intersection(_)
+            | RSResultData::Virtual
+            | RSResultData::Numeric(_)
+            | RSResultData::Metric(_)
+            | RSResultData::HybridMetric(_) => {
+                // SAFETY: unreachable because of safety condition 1
+                unsafe { std::hint::unreachable_unchecked() }
```
GlenDC, line 934: TIL: about `std::hint::unreachable_unchecked`

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6872/commits/963d9072160950f5b4446a809ec57c320a209375

src/redisearch_rs/inverted_index/src/index_result.rs
chesedo: Lol, I also learned it from #6834

src/redisearch_rs/inverted_index/src/freqs_only.rs
chesedo: I agree. We will want this to be covered by the type system instead. But that is not possible because the iterators are still in C.

Regarding partial assignments, that is exactly what the C code is currently expecting (and why this change is needed). Eg. this iterator is expecting the term, weight, etc to persist between decoded records.

https://github.com/RediSearch/RediSearch/blob/c54dce4a971dbf546ba45a8215bde56e1544e5b0/src/iterators/inverted_index_iterator.c#L555-L566

GlenDC (dismissed):

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/lib.rs
```diff
@@ -761,27 +791,20 @@ pub unsafe extern "C" fn IndexReader_HasSeeker(_ir: *const IndexReader) -> bool
...
 
-    match ir_dispatch!(ir, next) {
-        Some(new_res) => {
-            *res = new_res;
-
-            true
-        }
-        None => false,
-    }
+    ir_dispatch!(ir, next_record, res).unwrap_or_default()
```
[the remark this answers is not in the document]
chesedo: This observation is correct. And we also tried to design a way to make this safer but won't be able to it until the iterators are in Rust too

src/redisearch_rs/inverted_index/src/fields_offsets.rs
```diff
@@ -108,9 +117,16 @@ impl Decoder for FieldsOffsets {
...
+        decode_term_record_offsets(
+            cursor,
+            base,
+            0,
+            field_mask as t_fieldMask,
+            1,
+            offsets_sz,
+            result,
+        )?;
+        Ok(true)
```
GuyAv46, line 128: Can it be simplified to return the result of `decode_term_record_offsets`?

Edit: No, `decode_term_record_offsets` returns `std::io::Result<()>`

src/redisearch_rs/inverted_index/src/fields_offsets.rs
```diff
@@ -162,32 +178,41 @@ impl Decoder for FieldsOffsetsWide {
...
             cursor,
             base,
             delta,
             field_mask as t_fieldMask,
             1,
             offsets_sz,
+            result,
         )?;
-        Ok(record)
+        Ok(())
```
GuyAv46: Simplify return
```suggestion
)
```

src/redisearch_rs/inverted_index/src/freqs_offsets.rs
```diff
@@ -61,24 +61,32 @@ impl Decoder for FreqsOffsets {
...
-    ) -> std::io::Result<RSIndexResult<'index>> {
+        result: &mut RSIndexResult<'index>,
+    ) -> std::io::Result<()> {
         let (decoded_values, _bytes_consumed) = qint_decode::<3, _>(cursor)?;
         let [delta, freq, offsets_sz] = decoded_values;
 
-        let record = decode_term_record_offsets(cursor, base, delta, 0, freq, offsets_sz)?;
-        Ok(record)
+        decode_term_record_offsets(cursor, base, delta, 0, freq, offsets_sz, result)?;
+        Ok(())
```
GuyAv46: consider simplifying
```suggestion
        decode_term_record_offsets(cursor, base, delta, 0, freq, offsets_sz, result)
```

src/redisearch_rs/inverted_index/src/index_result.rs
```diff
@@ -251,6 +251,19 @@ impl<'index> RSTermRecord<'index> {
...
+
+    /// Set the offsets of this term record, replacing any existing offsets.
+    pub fn set_offsets(&mut self, offsets: RSOffsetVector<'index>) {
+        match self {
+            RSTermRecord::Borrowed { offsets: o, .. } => {
+                *o = offsets;
+            }
+            RSTermRecord::Owned { offsets: o, .. } => {
+                o.free_data();
+                *o = offsets.to_owned();
```
GuyAv46, line 263: I guess there is no case where we should change the ownership tag?

src/redisearch_rs/inverted_index/src/fields_offsets.rs
chesedo: Yeah, the return types are different and it does not make sense to return a `bool` from `decode_term_record_offsets`

src/redisearch_rs/inverted_index/src/fields_offsets.rs
chesedo: Since this returns a `Result` type it needs to have an explicit `Ok` or `Err` at the end

chesedo: Nvm, this will work...

src/redisearch_rs/inverted_index/src/index_result.rs
chesedo: No, there is no case since we are either setting borrowed offsets to the borrowed type, or making an owned copy when setting it to the owned type.

src/redisearch_rs/inverted_index/src/fields_offsets.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6872/commits/f49ab7acc01e39199cadaad0669fa157342997f0

src/redisearch_rs/inverted_index/src/freqs_offsets.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6872/commits/f49ab7acc01e39199cadaad0669fa157342997f0

src/redisearch_rs/inverted_index/src/numeric.rs
```diff
@@ -466,9 +473,18 @@ impl Decoder for Numeric {
...
+        // SAFETY: Caller must ensure `result` is numeric
+        unsafe {
+            *result.as_numeric_unchecked_mut() = num;
+        }
+
+        Ok(())
+    }
+
+    fn base_result<'index>() -> RSIndexResult<'index> {
+        RSIndexResult::numeric(0.0)
```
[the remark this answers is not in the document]
chesedo: Correct, these can only become compile time checks once the iterators are also in Rust.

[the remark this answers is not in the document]
chesedo: This initial Rust behaviour matches the C code though

src/redisearch_rs/inverted_index_bencher/src/benchers/fields_offsets.rs
```diff
@@ -211,10 +211,15 @@ impl Bencher {
...
                         let result = if self.wide {
-                            FieldsOffsetsWide::default().decode(buffer, 100).unwrap()
+                            FieldsOffsetsWide::default()
+                                .decode(buffer, 100, &mut record)
+                                .unwrap()
                         } else {
-                            FieldsOffsets::default().decode(buffer, 100).unwrap()
+                            FieldsOffsets::default()
+                                .decode(buffer, 100, &mut record)
+                                .unwrap()
```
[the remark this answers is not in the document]
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6872/commits/3e607ece4b776291516f1b8b08398c76983301c3

PR #6875, opened 2025-09-23
refactor: simplify `RSValue` mocking and testing code

JonasKruckenberg:
This change removes duplicate test implementations of the `RSValueTrait` trait in favor of one centralized type `RSValueMock` in the `value` crate.

### DIFF
--- a/src/redisearch_rs/value/src/lib.rs
+++ b/src/redisearch_rs/value/src/lib.rs
@@ -10,6 +10,11 @@
 //! Ports part of the RediSearch RSValue type to Rust. This is a temporary solution until we have a proper
 //! Rust port of the RSValue type.
 
+#[cfg(feature = "test_utils")]
+mod test_utils;
+#[cfg(feature = "test_utils")]
+pub use test_utils::RSValueMock;
+
 use std::{ffi::c_char, ptr::NonNull};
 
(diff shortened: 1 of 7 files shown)


src/redisearch_rs/value/src/lib.rs
```diff
@@ -204,3 +187,82 @@ impl RSValueTrait for RSValueFFI {
         std::mem::size_of::<ffi::RSValue>()
     }
 }
+
+/// Mock implementation of `RSValue` for testing purposes.
+#[derive(Clone, Debug, PartialEq)]
+pub struct RSValueMock(Arc<RSValueMockInner>);
```
LukeMathWalker: Should we put this behind a feature flag (e.g. `test_utils`)?

JonasKruckenberg: I thought about it but given that this will just all go away with Henks work stream I figured its not worth the complexity.

LukeMathWalker: Let's add it anyway, just a few lines here and there making our intent a bit cleaner.

JonasKruckenberg: done 👍🏻

PR #6877, opened 2025-09-23
[MOD-11400]  Add memory guardrail before AREQ_BuildPipeline

lerman25:
This PR adds a OOM guardrail before the query's pipeline is built, enforcing memory limits. 
The guardrail currently checks if memory consumption is over 100% of the limit, but this can be changed to estimate the queries consumption with an heuristics .
Continuing the work from #6730 , #6750 , #6769, #6828. 
This check is, temporarily, limited to standalone settings.

---

> [!NOTE]
> Adds a standalone-only OOM guardrail before building/executing queries, returning QUERY_EOOM when memory exceeds limits, with tests and a temporary cluster bypass.
> 
> - **Core/Execution**:
>   - Add OOM guardrail in `execCommandCommon` using `RedisMemory_GetUsedMemoryRatioUnified` to abort early with `QUERY_EOOM` when over memory limit; applies only in standalone (temporary cluster bypass via `isClusterEnv_TempOOM_DoNotUse`).
>   - Introduce `QUERY_EOOM` error code and message; log notice on OOM.
>   - Include headers `util/redis_mem_info.h`, `module.h`.
> - **Tests**:
>   - Add `tests/pytests/test_query_oom.py` covering OOM policies (`ignore`, `fail`, `return`) and behavior in standalone vs cluster environments for `FT.SEARCH` and `FT.AGGREGATE`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7a269bb8e54b166703dd65a22278039618896c1b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -27,6 +27,7 @@
 #include "pipeline/pipeline.h"
 #include "util/units.h"
 #include "hybrid/hybrid_request.h"
+#include "util/redis_mem_info.h"
 
 typedef enum {
   EXEC_NO_FLAGS = 0x00,
@@ -42,7 +43,26 @@ typedef struct {
   WeakRef spec_ref;
 } blockedClientReqCtx;
 
+// OOM guardrail with heuristics
+// TODO: add heuristics
+static bool estimateOOM(RedisModuleCtx *ctx, bool lockGIL) {
+  if (lockGIL) {
+    RedisModule_ThreadSafeContextLock(ctx);
+  }
+  bool oom = RedisMemory_GetUsedMemoryRatioUnified(ctx) > 1;
+  if (lockGIL) {
+    RedisModule_ThreadSafeContextUnlock(ctx);
+  }
+  return oom;
+}
+
 static void runCursor(RedisModule_Reply *reply, Cursor *cursor, size_t num);
+
+// Temporary function to check if we are in a cluster environment
+// TODO : Remove this function once it's no longer needed
+static bool isClusterEnv() {
+  return GetNumShards_UnSafe() > 1;
+}
 
 /**
  * Get the sorting key of the result. This will be the sorting key of the last
@@ -790,6 +810,18 @@ int prepareExecutionPlan(AREQ *req, QueryError *status) {
     req->profileParseTime = rs_wall_clock_diff_ns(&req->initClock, &parseClock);
   }
 
+  // Currently supporting OOM policy only in standalone env
+  if (req->reqConfig.oomPolicy != OomPolicy_Ignore && !isClusterEnv())
+  {
+    // OOM guardrail
+    // Hold the GIL if necessary before check OOM
+    bool background = req->reqflags & QEXEC_F_RUN_IN_BACKGROUND;
+    if (estimateOOM(sctx->redisCtx, background)) {
+      RedisModule_Log(sctx->redisCtx, "notice", "Not enough memory available to execute the query");
+      QueryError_SetError(status, QUERY_EOOM, NULL);
+      return REDISMODULE_ERR;
+    }
+  }
   rc = AREQ_BuildPipeline(req, status);
 
   if (is_profile) {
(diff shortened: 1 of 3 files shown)


src/aggregate/aggregate_exec.c
```diff
@@ -42,8 +44,27 @@ typedef struct {
...
+    RedisModule_ThreadSafeContextUnlock(ctx);
+  }
+  return oom;
+}
+
 static void runCursor(RedisModule_Reply *reply, Cursor *cursor, size_t num);
 
+// Temporary function to check if we are in a cluster environment
+// TODO : Remove this function once it's no longer needed
+static bool isClusterEnv() {
```
GuyAv46: Let's name this function in a way nobody tries to use it while it's in the code (you can use your own name)
```suggestion
static bool isClusterEnv_TempOOM_DoNotUse() {
```

src/aggregate/aggregate_exec.c
```diff
@@ -42,8 +44,27 @@ typedef struct {
   WeakRef spec_ref;
 } blockedClientReqCtx;
 
+// OOM guardrail with heuristics
+// TODO: add heuristics
```
GuyAv46, line 48: Is this still planned?

src/aggregate/aggregate_exec.c
GuyAv46: There is also a check to see if a command is internal or not. You can try using it and only check OOM if it's not

src/aggregate/aggregate_exec.c
```diff
@@ -790,6 +811,18 @@ int prepareExecutionPlan(AREQ *req, QueryError *status) {
...
+  {
+    // OOM guardrail
+    // Hold the GIL if necessary before check OOM
+    bool background = req->reqflags & QEXEC_F_RUN_IN_BACKGROUND;
+    if (estimateOOM(sctx->redisCtx, background)) {
+      RedisModule_Log(sctx->redisCtx, "notice", "Not enough memory available to execute the query");
+      QueryError_SetError(status, QUERY_EOOM, NULL);
+      return REDISMODULE_ERR;
+    }
+  }
```
GuyAv46: I think we can do the check much earlier, and then we can also avoid checking this from the background

src/aggregate/aggregate_exec.c
```diff
@@ -790,6 +811,18 @@ int prepareExecutionPlan(AREQ *req, QueryError *status) {
...
 
+  // Currently supporting OOM policy only in standalone env
+  if (req->reqConfig.oomPolicy != OomPolicy_Ignore && !isClusterEnv())
+  {
+    // OOM guardrail
+    // Hold the GIL if necessary before check OOM
+    bool background = req->reqflags & QEXEC_F_RUN_IN_BACKGROUND;
+    if (estimateOOM(sctx->redisCtx, background)) {
+      RedisModule_Log(sctx->redisCtx, "notice", "Not enough memory available to execute the query");
+      QueryError_SetError(status, QUERY_EOOM, NULL);
```
GuyAv46: Use the code only API
```suggestion
      QueryError_SetCode(status, QUERY_EOOM);
```

src/aggregate/aggregate_exec.c
lerman25: TBD, will remain as TODO for now

src/aggregate/aggregate_exec.c
```diff
@@ -960,9 +975,21 @@ static int execCommandCommon(RedisModuleCtx *ctx, RedisModuleString **argv, int
     return RedisModule_WrongArity(ctx);
   }
 
-  AREQ *r = AREQ_New();
   QueryError status = {0};
 
+  // Currently supporting OOM policy only in standalone env
+  if (RSGlobalConfig.requestConfigParams.oomPolicy != OomPolicy_Ignore && !isClusterEnv_TempOOM_DoNotUse())
+  {
```
GuyAv46, line 982:
```suggestion
  if (RSGlobalConfig.requestConfigParams.oomPolicy != OomPolicy_Ignore && !isClusterEnv_TempOOM_DoNotUse()) {
```

lerman25: Will be fix in a coming PR

PR #6844, opened 2025-09-16
mov repairInvIndex logic and friends from gc to II

GlenDC:
## Describe the changes in the pull request

Continuation of work done for https://redislabs.atlassian.net/browse/MOD-10702 (Implement ForkGc for inverted index in Rust).

Move scan/apply logic specific to II or its internals to II code, in a continued
attempt to keep the II surface touched by fork_gc as abstract and high level as possible.

Closes:

* https://redislabs.atlassian.net/browse/MOD-11222 (Move rules and cleanup; e.g. Shift last-block preservation and array cleanup into inverted_index.c, so fork_gc.c no longer manipulates block pointers directly.)
* https://redislabs.atlassian.net/browse/MOD-11223 (Remove custom types from fork_gc; Teach the inverted index to serialize/deserialize GC deltas, drop MSG_RepairedBlock, MSG_DeletedBlock, and InvIdxBuffers, leaving fork_gc.c to only orchestrate scan → delta → apply)

#### Mark if applicable

- [x] This PR introduces API changes (**NOTE: internal to II / GC only**)
- [ ] This PR introduces serialization changes

PR #6878, opened 2025-09-25
[MOD-11542] Implement Inverted Index GC logic in Rust

chesedo:
This implements the scan and apply logic needed for the inverted index garbage collection in Rust. Another PR will focus on the serializing/deserializing needed for the scan results.

Nothing uses this Rust logic until the inverted index is switched to Rust.

#### Mark if applicable

### DIFF
--- a/src/redisearch_rs/inverted_index/src/doc_ids_only.rs
+++ b/src/redisearch_rs/inverted_index/src/doc_ids_only.rs
@@ -16,7 +16,7 @@ use crate::{DecodedBy, Decoder, Encoder, RSIndexResult};
 
 /// Encode and decode only the delta document ID of a record, without any other data.
 /// The delta is encoded using [varint encoding](varint).
-#[derive(Default)]
+#[derive(Clone, Default)]
 pub struct DocIdsOnly;
 
 impl Encoder for DocIdsOnly {
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -597,26 +597,39 @@ impl<E: Encoder + DecodedBy> InvertedIndex<E> {
         }
     }
 
-    /// Apply the results of a garbage collection scan to the index. This will modify the index
+    /// Apply the deltas of a garbage collection scan to the index. This will modify the index
     /// by deleting or repairing blocks as needed.
-    pub fn apply_gc(&mut self, results: Vec<BlockGcScanResult>) {
+    pub fn apply_gc(&mut self, delta: GcScanDelta) {
+        let GcScanDelta {
+            last_block_idx,
+            last_block_num_entries,
+            deltas,
+        } = delta;
+
         // We apply the repairs in reverse order so that the indices of the blocks to be repaired
         // remain valid as we modify the blocks vector.
-        for result in results.into_iter().rev() {
-            match result.repair {
+        for delta in deltas.into_iter().rev() {
+            // Skip if the last block has changed since the scan
+            if delta.index == last_block_idx
+                && self.blocks[delta.index].num_entries != last_block_num_entries
+            {
+                continue;
+            }
+
+            match delta.repair {
                 RepairType::Delete => {
-                    let block = self.blocks.remove(result.index);
+                    let block = self.blocks.remove(delta.index);
                     self.n_unique_docs -= block.num_entries;
                 }
                 RepairType::Split { mut blocks } => {
-                    let old_block = self.blocks.remove(result.index);
+                    let old_block = self.blocks.remove(delta.index);
                     self.n_unique_docs -= old_block.num_entries;
 
                     blocks.reverse();
 
                     for block in blocks {
                         self.n_unique_docs += block.num_entries;
-                        self.blocks.insert(result.index, block);
+                        self.blocks.insert(delta.index, block);
                     }
                 }
             }
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -296,17 +296,24 @@ impl IndexBlock {
     ) -> Option<RepairType> {
         let mut cursor: Cursor<&'index [u8]> = Cursor::new(&self.buffer);
         let last_doc_id = self.first_doc_id;
+        let mut num_deletes = 0;
 
         while !cursor.fill_buf().unwrap().is_empty() {
             let base = D::base_id(self, last_doc_id);
             let result = decoder.decode(&mut cursor, base).unwrap();
 
-            if doc_exist_cb(result.doc_id) {
-                todo!()
+            if !doc_exist_cb(result.doc_id) {
+                num_deletes += 1;
             }
         }
 
-        Some(RepairType::Delete)
+        if num_deletes == self.num_entries {
+            Some(RepairType::Delete)
+        } else if num_deletes == 0 {
+            None
+        } else {
+            todo!("What did we repair")
+        }
     }
 }
 
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -573,6 +573,32 @@ impl<E: Encoder + DecodedBy> InvertedIndex<E> {
 
         Ok(results)
     }
+
+    /// Apply the results of a garbage collection scan to the index. This will modify the index
+    /// by deleting or repairing blocks as needed.
+    pub fn apply_gc(&mut self, results: Vec<BlockGcScanResult>) {
+        // We apply the repairs in reverse order so that the indices of the blocks to be repaired
+        // remain valid as we modify the blocks vector.
+        for result in results.into_iter().rev() {
+            match result.repair {
+                RepairType::Delete => {
+                    let block = self.blocks.remove(result.index);
+                    self.n_unique_docs -= block.num_entries;
+                }
+                RepairType::Split { mut blocks } => {
+                    let old_block = self.blocks.remove(result.index);
+                    self.n_unique_docs -= old_block.num_entries;
+
+                    blocks.reverse();
+
+                    for block in blocks {
+                        self.n_unique_docs += block.num_entries;
+                        self.blocks.insert(result.index, block);
+                    }
+                }
+            }
+        }
+    }
 }
 
 /// A wrapper around the inverted index to track the total number of entries in the index.
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -289,21 +289,44 @@ impl IndexBlock {
         buffer
     }
 
-    fn repair<'index, D: Decoder>(
+    fn repair<'index, E: Encoder + DecodedBy<Decoder = D>, D: Decoder>(
         &'index self,
         doc_exist_cb: fn(doc_id: t_docId) -> bool,
-        decoder: D,
+        encoder: E,
     ) -> Option<RepairType> {
         let mut cursor: Cursor<&'index [u8]> = Cursor::new(&self.buffer);
-        let last_doc_id = self.first_doc_id;
+        let mut last_doc_id = self.first_doc_id;
         let mut num_deletes = 0;
+        let mut new_buffer = Vec::with_capacity(self.buffer.len());
+        let mut new_cursor = Cursor::new(&mut new_buffer);
+        let decoder = E::decoder();
+
+        let mut new_block = None;
 
         while !cursor.fill_buf().unwrap().is_empty() {
             let base = D::base_id(self, last_doc_id);
             let result = decoder.decode(&mut cursor, base).unwrap();
+            last_doc_id = result.doc_id;
 
             if !doc_exist_cb(result.doc_id) {
                 num_deletes += 1;
+            } else {
+                if new_block.is_none() {
+                    new_block = Some(IndexBlock {
+                        first_doc_id: result.doc_id,
+                        last_doc_id: result.doc_id,
+                        num_entries: 0,
+                        buffer: Vec::with_capacity(self.buffer.len()),
+                    });
+                }
+                let new_block = new_block.as_mut().unwrap();
+
+                let delta_base = E::delta_base(new_block);
+                let delta = result.doc_id.wrapping_sub(delta_base);
+                let delta = E::Delta::from_u64(delta).unwrap();
+
+                encoder.encode(&mut new_cursor, delta, &result).unwrap();
+                new_block.num_entries += 1;
             }
         }
 
@@ -312,7 +335,12 @@ impl IndexBlock {
         } else if num_deletes == 0 {
             None
         } else {
-            todo!("What did we repair")
+            Some(RepairType::Rebuild {
+                first_doc_id: new_block.as_ref().unwrap().first_doc_id,
+                last_doc_id: new_block.as_ref().unwrap().last_doc_id,
+                num_entries: new_block.as_ref().unwrap().num_entries,
+                buffer: new_buffer,
+            })
         }
     }
 }
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -526,6 +526,7 @@ impl<E: Encoder> InvertedIndex<E> {
 }
 
 /// Result of scanning a block for garbage collection
+#[derive(Debug, Eq, PartialEq)]
 pub struct BlockGcScanResult {
     /// The index of the block in the inverted index
     index: usize,
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -525,6 +525,21 @@ impl<E: Encoder> InvertedIndex<E> {
     }
 }
 
+/// Result of scanning the index for garbage collection
+#[derive(Debug, Eq, PartialEq)]
+pub struct GcScanDelta {
+    /// The index of the last block in the index at the time of the scan. This is used to ensure
+    /// that the index has not changed since the scan was performed.
+    last_block_idx: usize,
+
+    /// The number of entries in the last block at the time of the scan. This is used to ensure
+    /// that the index has not changed since the scan was performed.
+    last_block_num_entries: usize,
+
+    /// The results of the scan for each block that needs to be repaired or deleted.
+    deltas: Vec<BlockGcScanResult>,
+}
+
 /// Result of scanning a block for garbage collection
 #[derive(Debug, Eq, PartialEq)]
 pub struct BlockGcScanResult {
@@ -546,11 +561,11 @@ impl<E: Encoder + DecodedBy> InvertedIndex<E> {
     /// callback is used to check if a document exists. It should return `true` if the document
     /// exists and `false` otherwise.
     ///
-    /// This function returns a vector of all the blocks which should be repaired or deleted.
+    /// This function returns a delta if GC is needed, or `None` if no GC is needed.
     pub fn scan_gc(
         &self,
         doc_exist_cb: fn(doc_id: t_docId) -> bool,
-    ) -> std::io::Result<Vec<BlockGcScanResult>> {
+    ) -> std::io::Result<Option<GcScanDelta>> {
         let mut results = Vec::new();
 
         for (i, block) in self.blocks.iter().enumerate() {
@@ -571,7 +586,15 @@ impl<E: Encoder + DecodedBy> InvertedIndex<E> {
             }
         }
 
-        Ok(results)
+        if results.is_empty() {
+            Ok(None)
+        } else {
+            Ok(Some(GcScanDelta {
+                last_block_idx: self.blocks.len() - 1,
+                last_block_num_entries: self.blocks.last().map(|b| b.num_entries).unwrap_or(0),
+                deltas: results,
+            }))
+        }
     }
 
     /// Apply the results of a garbage collection scan to the index. This will modify the index
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -308,22 +308,19 @@ impl IndexBlock {
             last_doc_id = result.doc_id;
 
             if doc_exist_cb(result.doc_id) {
-                if new_block.is_none() {
-                    new_block = Some(IndexBlock {
-                        first_doc_id: result.doc_id,
-                        last_doc_id: result.doc_id,
-                        num_entries: 0,
-                        buffer: Vec::with_capacity(self.buffer.len()),
-                    });
-                }
-                let new_block = new_block.as_mut().unwrap();
+                let tmp_block = new_block.get_or_insert(IndexBlock {
+                    first_doc_id: result.doc_id,
+                    last_doc_id: result.doc_id,
+                    num_entries: 0,
+                    buffer: Vec::with_capacity(self.buffer.len()),
+                });
 
-                let delta_base = E::delta_base(new_block);
+                let delta_base = E::delta_base(tmp_block);
                 let delta = result.doc_id.wrapping_sub(delta_base);
                 let delta = E::Delta::from_u64(delta).unwrap();
 
                 encoder.encode(&mut new_cursor, delta, &result).unwrap();
-                new_block.num_entries += 1;
+                tmp_block.num_entries += 1;
             }
         }
 
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -296,7 +296,6 @@ impl IndexBlock {
     ) -> Option<RepairType> {
         let mut cursor: Cursor<&'index [u8]> = Cursor::new(&self.buffer);
         let mut last_doc_id = self.first_doc_id;
-        let mut num_deletes = 0;
         let mut new_buffer = Vec::with_capacity(self.buffer.len());
         let mut new_cursor = Cursor::new(&mut new_buffer);
         let decoder = E::decoder();
@@ -308,9 +307,7 @@ impl IndexBlock {
             let result = decoder.decode(&mut cursor, base).unwrap();
             last_doc_id = result.doc_id;
 
-            if !doc_exist_cb(result.doc_id) {
-                num_deletes += 1;
-            } else {
+            if doc_exist_cb(result.doc_id) {
                 if new_block.is_none() {
                     new_block = Some(IndexBlock {
                         first_doc_id: result.doc_id,
@@ -330,17 +327,15 @@ impl IndexBlock {
             }
         }
 
-        if num_deletes == self.num_entries {
-            Some(RepairType::Delete)
-        } else if num_deletes == 0 {
-            None
-        } else {
-            Some(RepairType::Rebuild {
-                first_doc_id: new_block.as_ref().unwrap().first_doc_id,
-                last_doc_id: new_block.as_ref().unwrap().last_doc_id,
-                num_entries: new_block.as_ref().unwrap().num_entries,
+        match new_block {
+            Some(block) if block.num_entries == self.num_entries => None,
+            Some(block) => Some(RepairType::Rebuild {
+                first_doc_id: block.first_doc_id,
+                last_doc_id: block.last_doc_id,
+                num_entries: block.num_entries,
                 buffer: new_buffer,
-            })
+            }),
+            None => Some(RepairType::Delete),
         }
     }
 }
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -288,33 +288,33 @@ impl IndexBlock {
         &'index self,
         doc_exist_cb: fn(doc_id: t_docId) -> bool,
         encoder: E,
-    ) -> Option<RepairType> {
+    ) -> std::io::Result<Option<RepairType>> {
         let mut cursor: Cursor<&'index [u8]> = Cursor::new(&self.buffer);
         let mut last_doc_id = self.first_doc_id;
         let decoder = E::decoder();
 
         let mut tmp_inverted_index = InvertedIndex::new(IndexFlags_Index_DocIdsOnly, encoder);
 
-        while !cursor.fill_buf().unwrap().is_empty() {
+        while !cursor.fill_buf()?.is_empty() {
             let base = D::base_id(self, last_doc_id);
-            let result = decoder.decode(&mut cursor, base).unwrap();
+            let result = decoder.decode(&mut cursor, base)?;
             last_doc_id = result.doc_id;
 
             if doc_exist_cb(result.doc_id) {
-                tmp_inverted_index.add_record(&result).unwrap();
+                tmp_inverted_index.add_record(&result)?;
             }
         }
 
         if tmp_inverted_index.blocks.is_empty() {
-            Some(RepairType::Delete)
+            Ok(Some(RepairType::Delete))
         } else if tmp_inverted_index.blocks.len() == 1
             && tmp_inverted_index.blocks[0].num_entries == self.num_entries
         {
-            None
+            Ok(None)
         } else {
-            Some(RepairType::Split {
+            Ok(Some(RepairType::Split {
                 blocks: tmp_inverted_index.blocks,
-            })
+            }))
         }
     }
 }
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -250,9 +250,13 @@ pub struct IndexBlock {
     buffer: Vec<u8>,
 }
 
+/// The type of repair needed for a block after a garbage collection scan.
 #[derive(Debug, Eq, PartialEq)]
 enum RepairType {
+    /// This block can be deleted completely.
     Delete,
+
+    /// The block contains GCed entries, and might need to be split into the following blocks.
     Split { blocks: Vec<IndexBlock> },
 }
 
@@ -284,6 +288,8 @@ impl IndexBlock {
         buffer
     }
 
+    /// Repair a block by removing records which no longer exists according to `doc_exists_cb`. If
+    /// there is nothing to repair in this block then `None` is returned.
     fn repair<'index, E: Encoder + DecodedBy<Decoder = D>, D: Decoder>(
         &'index self,
         doc_exist_cb: fn(doc_id: t_docId) -> bool,
@@ -519,8 +525,12 @@ impl<E: Encoder> InvertedIndex<E> {
     }
 }
 
+/// Result of scanning a block for garbage collection
 pub struct BlockGcScanResult {
+    /// The index of the block in the inverted index
     index: usize,
+
+    /// The type of repair needed for this block
     repair: RepairType,
 }
 
@@ -530,6 +540,12 @@ impl<E: Encoder + DecodedBy> InvertedIndex<E> {
         IndexReader::new(self)
     }
 
+    /// Scan the index for blocks that can be garbage collected. A block can be garbage collected
+    /// if any of its records point to documents that no longer exist. The `doc_exist_cb`
+    /// callback is used to check if a document exists. It should return `true` if the document
+    /// exists and `false` otherwise.
+    ///
+    /// This function returns a vector of all the blocks which should be repaired or deleted.
     pub fn scan_gc(
         &self,
         doc_exist_cb: fn(doc_id: t_docId) -> bool,
--- a/src/redisearch_rs/inverted_index/src/lib.rs
+++ b/src/redisearch_rs/inverted_index/src/lib.rs
@@ -14,7 +14,7 @@ use std::{
 
 use debug::{BlockSummary, Summary};
 use ffi::{
-    FieldSpec, GeoFilter, IndexFlags, IndexFlags_Index_HasMultiValue,
+    FieldSpec, GeoFilter, IndexFlags, IndexFlags_Index_DocIdsOnly, IndexFlags_Index_HasMultiValue,
     IndexFlags_Index_StoreFieldFlags, IndexFlags_Index_StoreNumeric,
 };
 pub use ffi::{t_docId, t_fieldMask};
@@ -253,12 +253,7 @@ pub struct IndexBlock {
 #[derive(Debug, Eq, PartialEq)]
 enum RepairType {
     Delete,
-    Rebuild {
-        first_doc_id: t_docId,
-        last_doc_id: t_docId,
-        num_entries: usize,
-        buffer: Vec<u8>,
-    },
+    Split { blocks: Vec<IndexBlock> },
 }
 
 impl IndexBlock {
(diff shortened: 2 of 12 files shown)


GlenDC (dismissed):

GlenDC:
I think you have same issue here @chesedo as I have in https://github.com/RediSearch/RediSearch/pull/6844, once mine is fixed and merged in main, a rebase on latest master probably fixes yours. I'll leave a follow up comment later.

GlenDC:
> I think you have same issue here @chesedo as I have in #6844, once mine is fixed and merged in main, a rebase on latest master probably fixes yours. I'll leave a follow up comment later.

Yes, seems to have worked. It's in merge queue.

src/redisearch_rs/inverted_index/src/doc_ids_only.rs
```diff
@@ -16,7 +16,7 @@ use crate::{DecodedBy, Decoder, Encoder, RSIndexResult};
 
 /// Encode and decode only the delta document ID of a record, without any other data.
 /// The delta is encoded using [varint encoding](varint).
-#[derive(Default)]
+#[derive(Clone, Default)]
```
LukeMathWalker: Copy, while we are at it?

LukeMathWalker: Here and in other unit struct cases.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -255,6 +255,16 @@ pub struct IndexBlock {
...
 static TOTAL_BLOCKS: AtomicUsize = AtomicUsize::new(0);
 
+/// The type of repair needed for a block after a garbage collection scan.
+#[derive(Debug, Eq, PartialEq)]
+enum RepairType {
+    /// This block can be deleted completely.
+    Delete,
+
+    /// The block contains GCed entries, and might need to be split into the following blocks.
+    Split { blocks: Vec<IndexBlock> },
```
LukeMathWalker: Do we know how big this vector is likely to be in practice?
Given that we are splitting a single block, I imagine we won't fragment it into too many smaller blocks?

If that's the case, it might make sense to use `smallvec`.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -289,6 +299,42 @@ impl IndexBlock {
     pub fn total_blocks() -> usize {
         TOTAL_BLOCKS.load(atomic::Ordering::Relaxed)
     }
+
+    /// Repair a block by removing records which no longer exists according to `doc_exists_cb`. If
+    /// there is nothing to repair in this block then `None` is returned.
+    fn repair<'index, E: Encoder + DecodedBy<Decoder = D>, D: Decoder>(
+        &'index self,
+        doc_exist_cb: fn(doc_id: t_docId) -> bool,
```
```suggestion
        doc_exist: fn(doc_id: t_docId) -> bool,
```

No need for polish naming conventions.

src/redisearch_rs/inverted_index/src/lib.rs
LukeMathWalker: Looking at the repair logic below, it does indeed seem to be the case that `smallvec` would be a good fit.
I would suggest renaming from `Split` to `Replace`, since the vector is likely to contain a single block in most cases.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -497,11 +543,147 @@ impl<E: Encoder> InvertedIndex<E> {
...
+    /// if any of its records point to documents that no longer exist. The `doc_exist_cb`
+    /// callback is used to check if a document exists. It should return `true` if the document
+    /// exists and `false` otherwise.
+    ///
+    /// This function returns a delta if GC is needed, or `None` if no GC is needed.
+    pub fn scan_gc(
+        &self,
+        doc_exist_cb: fn(doc_id: t_docId) -> bool,
+    ) -> std::io::Result<Option<GcScanDelta>> {
+        let mut results = Vec::new();
```
LukeMathWalker, line 683:
```suggestion
        let mut results = Vec::with_capacity(self.blocks.len());
```

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -497,11 +543,147 @@ impl<E: Encoder> InvertedIndex<E> {
...
+            if let Some(repair) = repair {
+                results.push(BlockGcScanResult { index: i, repair });
+            }
+        }
+
+        if results.is_empty() {
+            Ok(None)
+        } else {
+            Ok(Some(GcScanDelta {
+                last_block_idx: self.blocks.len() - 1,
```
LukeMathWalker, line 699: Is an inverted index guaranteed to always have at least one block inside?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -497,11 +543,147 @@ impl<E: Encoder> InvertedIndex<E> {
...
+        let mut info = GcApplyInfo {
+            bytes_freed: 0,
+            bytes_allocated: 0,
+            entries_removed: 0,
+            blocks_ignored: 0,
+        };
+
+        // We apply the repairs in reverse order so that the indices of the blocks to be repaired
+        // remain valid as we modify the blocks vector.
+        for delta in deltas.into_iter().rev() {
```
LukeMathWalker, line 652: It may be more efficient to use `Vec::retain` rather than doing N `Vec::remove` invocations—less shifting.

LukeMathWalker: It feels like we are doing a lot of shifting here, for both removals and replacements.
I wonder if it'd be faster to have an auxiliary `Vec` with the same capacity of the old one, drain the old one and push the right entries into the new one directly.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Won't this waste memory especially if there might be thousands of blocks but only a few need to be repaired?

src/redisearch_rs/inverted_index/src/doc_ids_only.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6878/commits/356ce90a0a16c5d356962ef10c02b024cc8cbaf2

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6878/commits/9f6fa7a06a6e363052dc3d5e3d316d1fafe1437b

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6878/commits/198c49e078c7034c56b01c305f2bd01010746408

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: No, the blocks start of as empty. But it is also impossible to hit this line if an inverted index has no blocks since that will mean the `results` (on the `if` block) will be empty too.

src/redisearch_rs/inverted_index/src/lib.rs
LukeMathWalker: You are indeed right.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6878/commits/2100d715e2a9a6feaad7f1fcd7a7205b86c65250

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -311,6 +331,62 @@ impl IndexBlock {
...
+        encoder: E,
+    ) -> std::io::Result<Option<RepairType>> {
+        let mut cursor: Cursor<&'index [u8]> = Cursor::new(&self.buffer);
+        let mut last_doc_id = None;
+        let decoder = E::decoder();
+        let mut result = D::base_result();
+        let mut unique_read = 0;
+        let mut unique_write = 0;
+
+        let mut tmp_inverted_index = InvertedIndex::new(IndexFlags_Index_DocIdsOnly, encoder);
```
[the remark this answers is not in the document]
chesedo: The flag is not used for encoding and the correct encoder is also passed at the second argument. Since there is no flag variant to indicate a `None` flag, this is fine.

src/redisearch_rs/inverted_index/Cargo.toml
```diff
@@ -10,6 +10,7 @@ enumflags2.workspace = true
 ffi.workspace = true
 low_memory_thin_vec.workspace = true
 qint.workspace = true
+smallvec.workspace = true
```
LukeMathWalker: We want to enable the `union` feature flag to reduce the size further.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -277,6 +278,25 @@ pub struct IndexBlock {
...
+        /// Number of unique records this will remove
+        n_unique_docs_removed: usize,
+    },
+
+    /// The block contains GCed entries, and should be replaced with the following blocks.
+    Replace {
+        /// The new blocks to replace this block with
+        blocks: SmallVec<[IndexBlock; 3]>,
+
+        /// The unique documents that were removed from this block
```
```suggestion
        /// How many unique documents were removed from the block being replaced.
```

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -1203,3 +1212,668 @@ fn blocks_summary_store_numeric() {
...
+
+#[test]
+fn index_block_repair_delete() {
+    // Make a block with three entries (two duplicates) which will be deleted
+    let block = IndexBlock {
+        buffer: vec![0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0],
+        num_entries: 3,
+        first_doc_id: 10,
+        last_doc_id: 11,
+    };
```
LukeMathWalker, line 1242: I find this hard to read, since I don't actually get to see the entries that are being encoded.
Wouldn't it be more readable to build the blocks from index result entries?

src/redisearch_rs/inverted_index/Cargo.toml
GlenDC: How are you even aware of that union flag @LukeMathWalker and what does it do?

chesedo: @GlenDC it replaces the internal enum with a `union` instead

https://docs.rs/smallvec/latest/smallvec/#union

LukeMathWalker: I read the manual 😁

GlenDC: ouch,shame on me. Gotta read that manual!

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6878/commits/2e3104d6dfc6330deeb64c53a54c13d3fdd3c2f4

src/redisearch_rs/inverted_index/Cargo.toml
LukeMathWalker: Feature flag documentation is usually horrible, but the `smallvec` folks did a great job there.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -531,11 +607,170 @@ impl<E: Encoder> InvertedIndex<E> {
...
+                        RepairType::Replace {
+                            blocks,
+                            n_unique_docs_removed,
+                        } => {
+                            info.entries_removed += block.num_entries;
+                            info.bytes_freed += IndexBlock::SIZE + block.buffer.capacity();
+                            self.n_unique_docs -= n_unique_docs_removed;
+
+                            for block in blocks {
+                                info.entries_removed -= block.num_entries;
```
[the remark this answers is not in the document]
chesedo: This is actually correct. It is done in this way to offset the addition of the original block removal a few lines higher.

LukeMathWalker (approved):

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -641,6 +876,32 @@ impl<E: Encoder + DecodedBy> EntriesTrackingIndex<E> {
...
+    /// callback is used to check if a document exists. It should return `true` if the document
+    /// exists and `false` otherwise.
+    ///
+    /// If a doc does exist, then `repair` is called with it to run any repair calculations needed.
+    ///
+    /// This function returns a delta if GC is needed, or `None` if no GC is needed.
+    pub fn scan_gc<'index>(
+        &'index self,
+        doc_exist: impl Fn(t_docId) -> bool,
+        repair: impl FnMut(&RSIndexResult<'index>, &IndexBlock),
```
meiravgri, line 924: should it be aligned with `self.index.scan_gc` args type?
```suggestion
        mut repair: impl FnMut(&RSIndexResult<'index>, &IndexBlock),
```

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -531,11 +607,170 @@ impl<E: Encoder> InvertedIndex<E> {
...
+    /// exists and `false` otherwise.
+    ///
+    /// If a doc does exist, then `repair` is called with it to run any repair calculations needed.
+    ///
+    /// This function returns a delta if GC is needed, or `None` if no GC is needed.
+    pub fn scan_gc<'index>(
+        &'index self,
+        doc_exist: impl Fn(t_docId) -> bool,
+        mut repair: impl FnMut(&RSIndexResult<'index>, &IndexBlock),
+    ) -> std::io::Result<Option<GcScanDelta>> {
```
meiravgri, line 682: Did we intentionally drop this code block?
https://github.com/RediSearch/RediSearch/blob/2e3104d6dfc6330deeb64c53a54c13d3fdd3c2f4/src/inverted_index/inverted_index.c#L1751-L1760

If i understood correctly, it was addressed by allocating more blocks instead of just skipping blocks with big deltas?
Was it a bug in the C code?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -531,11 +607,170 @@ impl<E: Encoder> InvertedIndex<E> {
...
+        for (i, block) in self.blocks.iter().enumerate() {
+            if block.num_entries == 0 {
+                results.push(BlockGcScanResult {
+                    index: i,
+                    repair: RepairType::Delete {
+                        n_unique_docs_removed: 0,
+                    },
+                });
+                continue;
+            }
```
meiravgri: what scenario does this cover?
As far as i know, num_entries is only updated in `block.repair`

meiravgri: After some thinking, the only scenario i can think about is that we have 
1. inserted some docs to the index (otherwise, we shoudnkt get here, as we lazy-initialize the indexes)
2. deleted them all (if only delete some blocks, after the GC we shoudln't have empty blocks)
3. run gc (resulting in an index containing one empty block)
4. not adding new entries
5. run gc again

Is there another scenario where this code is reached?
If not, can we skip calling `scan_gc` entirly if the index is empty?

src/redisearch_rs/inverted_index/src/lib.rs
meiravgri: Was this changed with the transition to rust?
AFAIK, in the c implementation we always start with one empty block

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -1207,3 +1216,742 @@ fn blocks_summary_store_numeric() {
...
+    // Create 5 blocks:
+    // - One which is empty
+    // - One which will be completely deleted
+    // - One which will be partially deleted
+    // - Two which will be unchanged
+    let encoder = Dummy;
+    let blocks = vec![
+        IndexBlock {
+            buffer: vec![],
+            num_entries: 0,
```
meiravgri: is it possible that to have an empty block in a non-empty index?
and what is the meaning of `first_doc_id` = `last_doc_id` = 5 when it's empty?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -331,6 +351,62 @@ impl IndexBlock {
...
+    ///
+    /// `None` is returned when there is nothing to repair in this block.
+    fn repair<'index, E: Encoder + DecodedBy<Decoder = D>, D: Decoder>(
+        &'index self,
+        doc_exist: impl Fn(t_docId) -> bool,
+        mut repair: impl FnMut(&RSIndexResult<'index>, &IndexBlock),
+        encoder: E,
+    ) -> std::io::Result<Option<RepairType>> {
+        let mut cursor: Cursor<&'index [u8]> = Cursor::new(&self.buffer);
+        let mut last_doc_id = None;
```
meiravgri:
```suggestion
        let mut last_read_doc_id = None;
```

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -331,6 +351,62 @@ impl IndexBlock {
...
+            last_doc_id = Some(result.doc_id);
+        }
+
+        if tmp_inverted_index.blocks.is_empty() {
+            Ok(Some(RepairType::Delete {
+                n_unique_docs_removed: unique_read,
+            }))
+        } else if tmp_inverted_index.blocks.len() == 1
+            && tmp_inverted_index.blocks[0].num_entries == self.num_entries
+        {
```
meiravgri: woudnlt it be simpler to introduce a `is_changed` flag and turn it on if we enter 
[if doc_exist(result.doc_id) {](https://github.com/RediSearch/RediSearch/blob/7453edbe70cfad2f36e0c77faa892ad0e047d328/src/redisearch_rs/inverted_index/src/lib.rs#L378-L379)

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -543,11 +619,170 @@ impl<E: Encoder> InvertedIndex<E> {
...
+    /// Apply the deltas of a garbage collection scan to the index. This will modify the index
+    /// by deleting or repairing blocks as needed.
+    pub fn apply_gc(&mut self, delta: GcScanDelta) -> GcApplyInfo {
+        let GcScanDelta {
+            last_block_idx,
+            last_block_num_entries,
+            deltas,
+        } = delta;
+
+        let mut info = GcApplyInfo {
```
meiravgri, line 715: Didn't know where is the most appropriate place for this comment - 
Since now the new index might include more blocks comparing to the original, it should be mentioned that the bytes collected might be actually **smaller** than added, to be carefully applied when we calculate the total collected bytes, to prevent overflow of unsigned statistics.
(Intuitively it would be `total collected = freed - allocated`)
I don't think an overflow is possible today (in the C imp)

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -543,11 +619,170 @@ impl<E: Encoder> InvertedIndex<E> {
...
+
+        let mut deltas = deltas.into_iter().peekable();
+
+        for (block_index, block) in tmp_blocks.into_iter().enumerate() {
+            match deltas.peek() {
+                Some(delta)
+                    if delta.index == block_index
+                        && delta.index == last_block_idx
+                        && block.num_entries != last_block_num_entries =>
+                {
```
meiravgri: Do we have to check it EVERY iteration? Can't it be moved outside the for loop?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -543,11 +619,170 @@ impl<E: Encoder> InvertedIndex<E> {
...
+                    // The last block has changed since the scan, so we ignore this delta
+                    info.blocks_ignored += 1;
+                    self.blocks.push(block);
+                }
+                Some(delta) if delta.index == block_index => {
+                    // This block needs to be repaired
+                    let delta = deltas.next().expect("we just peeked it");
```
meiravgri: Does `expect` can lead to a crash at runtime?
If so, is there a more graceful way to handle unexpected behaviours?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -543,11 +619,170 @@ impl<E: Encoder> InvertedIndex<E> {
...
+            entries_removed: 0,
+            blocks_ignored: 0,
+        };
+
+        let mut tmp_blocks = Vec::with_capacity(self.blocks.len());
+        std::mem::swap(&mut self.blocks, &mut tmp_blocks);
+
+        let mut deltas = deltas.into_iter().peekable();
+
+        for (block_index, block) in tmp_blocks.into_iter().enumerate() {
```
meiravgri, line 756: So in this implantation, unlike the C code, we anyway iterate the entire ii blocks and repair them one by one, rather than using the re-written blocks array. Did i get it correctly?
Can it significancy harm the gc performance?

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -277,6 +278,25 @@ pub struct IndexBlock {
 
 static TOTAL_BLOCKS: AtomicUsize = AtomicUsize::new(0);
 
+/// The type of repair needed for a block after a garbage collection scan.
+#[derive(Debug, Eq, PartialEq)]
+enum RepairType {
+    /// This block can be deleted completely.
+    Delete {
+        /// Number of unique records this will remove
+        n_unique_docs_removed: usize,
```
meiravgri, line 287: why are we tracking this? (here and in `Replace)`

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -1207,3 +1216,742 @@ fn blocks_summary_store_numeric() {
...
+        })
+    );
+}
+
+#[test]
+fn index_block_repair_delta_too_big() {
+    #[derive(Clone)]
+    struct SmallDeltaDummy;
+
+    struct U5Delta(u32);
```
meiravgri: Taking the opportunity for some rust questions :blush: 
Why not to use `u8` here?

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -1207,3 +1216,742 @@ fn blocks_summary_store_numeric() {
         ]
     );
 }
+
+/// Helper macro to encode a series of doc IDs using the provided encoder. The first ID is encoded
+/// as a delta from 0, and each subsequent ID is encoded as a delta from the previous ID.
+macro_rules! encode_ids {
```
meiravgri, line 1216: why not to use production encoders/decoders?

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -1207,3 +1216,742 @@ fn blocks_summary_store_numeric() {
...
+
+        fn encode<W: std::io::Write + std::io::Seek>(
+            &self,
+            mut writer: W,
+            delta: Self::Delta,
+            _record: &RSIndexResult,
+        ) -> std::io::Result<usize> {
+            writer.write_all(&delta.0.to_be_bytes())?;
+
+            Ok(8)
```
meiravgri: why 8?

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -1207,3 +1216,742 @@ fn blocks_summary_store_numeric() {
...
+                    last_doc_id: 21,
+                }],
+                n_unique_docs_removed: 2,
+            },
+        },
+    ];
+
+    let delta = GcScanDelta {
+        last_block_idx: 1,
+        last_block_num_entries: 2,
```
meiravgri, line 1772: ?
```suggestion
        last_block_num_entries: 3,
```

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -1207,3 +1216,742 @@ fn blocks_summary_store_numeric() {
...
+            // The third and fifth block was split making 168 new bytes
+            bytes_allocated: 168,
+            entries_removed: 5,
+            blocks_ignored: 0
+        }
+    );
+}
+
+#[test]
+fn ii_apply_gc_last_block_updated() {
```
meiravgri, line 1721: consider adding:
1. a test that skips last block and doesn't change the number of blocks (first block was repaired or untouched)
2. a test that skips last block + one of the blocks became 2 because of a too-big delta

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -1207,3 +1216,742 @@ fn blocks_summary_store_numeric() {
...
+                num_entries: 1,
+                buffer: encode_ids!(Dummy, 11),
+            }],
+            n_unique_docs_removed: 2
+        })
+    );
+}
+
+#[test]
+fn index_block_repair_delta_too_big() {
```
meiravgri, line 1316: I believe this test should also include the apply_gc part, and make sure we handle well a scenario where blocks number actually increase. Especially considering this was not the behaviour until today

meiravgri (commented):
Rust code is so clean comparing to C its amazing.
Had some comments.
And a general comment about the tests - should also add tests that does both scan and apply?
I suggest to add to the PR description the fundmental chnages this PR introduces in the GC behaviour
This chnages should be also benchmarked the ensure no regression in performance(bullet 1), or memory consumption (bullet 2)
1. No new blocks array, we alawys iterate the index blocks on the parent side
2. Blocks count can actually grow
If i missed additional fundmental changes, please add them too :) 
Thanks!

Edit:
And also this change?
https://github.com/RediSearch/RediSearch/pull/6878#discussion_r2404358104

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -543,11 +619,170 @@ impl<E: Encoder> InvertedIndex<E> {
...
+                last_block_idx: self.blocks.len() - 1,
+                last_block_num_entries: self.blocks.last().map(|b| b.num_entries).unwrap_or(0),
+                deltas: results,
+            }))
+        }
+    }
+
+    /// Apply the deltas of a garbage collection scan to the index. This will modify the index
+    /// by deleting or repairing blocks as needed.
+    pub fn apply_gc(&mut self, delta: GcScanDelta) -> GcApplyInfo {
```
gdesmott, line 708: This function taking a `&mut self` will make it not convenient to use.

For example, I'm trying to re-implement the test from https://github.com/RediSearch/RediSearch/pull/6958/files#r2410064213 with my Rust II iterator and it's impossible to call `apply_gc` while having an iterator around as the iterator already owns an immutable ref on the II.

This may not be a blocker for this PR but something to keep in mind.
I think we'll have to come up with a smarter way to handle references between iterators and the II as they have their own internal logic to handle concurrency access.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -543,11 +619,170 @@ impl<E: Encoder> InvertedIndex<E> {
...
+
+    /// The type of repair needed for this block
+    repair: RepairType,
+}
+
+/// Information about the result of applying a garbage collection scan to the index
+#[derive(Debug, Eq, PartialEq)]
+pub struct GcApplyInfo {
+    /// The number of bytes that were freed
+    bytes_freed: usize,
```
gdesmott: I think all those should be `pub` so the caller, or at least tests, can check what has been done by the GC.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: @meiravgri yes, this is a change with the Rust implementation. The C had an extra field to keep track of the last doc ID in the index. I did not want this extra field in Rust, but rather just get it from the last block in the index. This means I cannot start with an empty block as that will cause the last doc ID for the index to be wrong.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: I do the alignment at the FFI layer (see #6933). At this level I want to keep things as idiomatic Rust as possible.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Yes, I'm intentionally not implementing that code block. I consider it a bug to not collect an entry just because it will result in the delta being too big. Instead, I can correctly collect the entry by splitting the block into multiple blocks.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: I have no idea what this was for :sweat_smile: Good catch! I'll remove it

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: It will be equivalent, but I'm not sure it will be significantly simpler

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Good point, this should be okay since we only use `totalCollected` via this `FGC_updateStats` in #6933 

https://github.com/RediSearch/RediSearch/blob/745a33f3585903be9487c522b05c845afa6c19d2/src/fork_gc.c#L58-L66

It correctly adds the collected before subtracting the allocation like your suggestion.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: This `expect` will never panic because we just peeked the value. Hence why that is documented in the expect.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: To correctly update the `n_unique_docs` of the inverted index here

https://github.com/RediSearch/RediSearch/blob/807cb6c3a5c8e4808a3291704c45cd40fa61d00a/src/redisearch_rs/inverted_index/src/lib.rs#L741-L763

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: No reason really. I just did not think of a `u8` :sweat_smile: I'll make an update

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: Because I want simpler encoders that make testing easier. Sometimes I also want an encoder / decoder to trigger a specific test case like handling a delta that is too big.

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: Are there any edge-cases why we would want these tests? Btw, skipping a too-big delta is not possible

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: The apply test already does that here

https://github.com/RediSearch/RediSearch/blob/807cb6c3a5c8e4808a3291704c45cd40fa61d00a/src/redisearch_rs/inverted_index/src/tests.rs#L1635-L1654

I will not want to bloat this test with that too. Then it will turn into an integration test.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6878/commits/036747f94cc35eea511a2ddc26acfb3535d2ca9f

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: Correct, empty blocks are no longer possible. I removed these from the tests in https://github.com/RediSearch/RediSearch/pull/6878/commits/4b1f5a27b910af59f0a8c0ed8e078157ca927cfe

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6878/commits/871fd758884cecec58df7d04364bf7311536fee2

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: I moved it outside the loop in https://github.com/RediSearch/RediSearch/pull/6878/commits/f1f5dd8243a23ac669839ef74025885d8e15664a

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: I did not see a need for re-writting all the blocks. But we certainly don't have to loop over all of them if nothing changed. So I made this update in https://github.com/RediSearch/RediSearch/pull/6878/commits/be2f83a865ec523bb5421eceee4a9df097267590

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6878/commits/1e81444a2dba2defba87ad8aa5299f7aa0154534

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: Updated in https://github.com/RediSearch/RediSearch/pull/6878/commits/1e81444a2dba2defba87ad8aa5299f7aa0154534 (a `u8` will write one byte)

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: I documented the odd number in https://github.com/RediSearch/RediSearch/pull/6878/commits/d2a47b33e44406a76561ab286b1bf1e03d08cf56

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6878/commits/d62ea8cef89f2cdb5d2b0b1e45dfd07f06750f42

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -814,9 +1107,8 @@ impl<'index, E: DecodedBy<Decoder = D>, D: Decoder> IndexReader<'index>
     for IndexReaderCore<'index, E, D>
 {
     fn next_record(&mut self, result: &mut RSIndexResult<'index>) -> std::io::Result<bool> {
-        // Check if the current buffer is empty. The GC might clean out a block so we have to
-        // continue checking until we find a block with data.
-        while self.current_buffer.fill_buf()?.is_empty() {
+        // Check if the current buffer is empty
+        if self.current_buffer.fill_buf()?.is_empty() {
```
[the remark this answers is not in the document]
chesedo: Empty blocks can't exist hence why this change is correct

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: I can write a Rust benchmark for this

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: We should add two python tests for when all the documents are collected from an index to ensure this won't crash. One test for numeric and another for text data.

chesedo: The numeric test should also cover the "merge logic" that will happen during GC

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: The `is_changed` will be simpler so will add  the change

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: The `totalCollected` should be changed to a signed integer

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: I will just research if this will affect performance

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: I'll add an integration test with fuller behaviours to cover both the scan and apply parts

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -277,6 +278,25 @@ pub struct IndexBlock {
...
+    /// This block can be deleted completely.
+    Delete {
+        /// Number of unique records this will remove
+        n_unique_docs_removed: usize,
+    },
+
+    /// The block contains GCed entries, and should be replaced with the following blocks.
+    Replace {
+        /// The new blocks to replace this block with
+        blocks: SmallVec<[IndexBlock; 3]>,
```
meiravgri, line 293: why 3?

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: This benchmark and other common patterns are in https://github.com/RediSearch/RediSearch/pull/6878/commits/11c6f5fbe605f018abbd100545d9c252ce693287

The PR description is updated with the benchmark results

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/6878/commits/875028c5314cf765d7fe949e3c6e6d791be4bfe7

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: The `unwrap()` or `expect()` is indeed the most common way of doing this. It adds one extra step in release compiles for the happy path. I still made an update in https://github.com/RediSearch/RediSearch/pull/6878/commits/4c5ca510fb2829fd6817ef10248509233ef14671 which is slightly quicker according to the benchmarks.

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: Full suite of integration tests are added here https://github.com/RediSearch/RediSearch/pull/6878/commits/077a9145ba9a564fbf200b8ec6cd48702ec70b25

They test scan, apply, read, and seek on a single index.

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: The growing test was added here https://github.com/RediSearch/RediSearch/pull/6878/commits/077a9145ba9a564fbf200b8ec6cd48702ec70b25. I'll have to add the other one in python in a separate PR

meiravgri (approved):

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: This is done in #7088

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: The extra tests are in #7088

src/redisearch_rs/inverted_index/src/tests.rs
chesedo: This python test is almost what we want to test the number of blocks does not change.

https://github.com/RediSearch/RediSearch/blob/152dd9a7b62b62e26e92447004c160408a0d7ffd/tests/pytests/test_gc.py#L8-L31

However, I don't know how to pause the GC to add a record to have in not clean the last block.

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Oops, missed this. It just seemed rational that most splits won't make more than 3 blocks. `3` here means using the stack to store the elements if there are less than `3`. So if the split makes more than 3 blocks then they will be stored on the heap instead (slower)

PR #6881, opened 2025-09-25
MOD-11581 Implement `RLookupRow::write_fields_from` porting `RLookupR…

JonasKruckenberg:
…ow_WriteFieldsFrom`

This change adds the Rust method `RLookupRow::write_fields_from` porting the C function `RLookupRow_WriteFieldsFrom`

#### Mark if applicable


src/redisearch_rs/rlookup/src/row.rs
```diff
@@ -144,4 +143,35 @@ impl<'a, T: RSValueTrait> RLookupRow<'a, T> {
...
+    //  * Assumes all source keys exist in destination (enforce with ASSERT).
+    //  */
+    // void RLookupRow_WriteFieldsFrom(const RLookupRow *srcRow, const RLookup *srcLookup,
+    //                                RLookupRow *destRow, const RLookup *destLookup);
+    pub fn copy_fields_from(
+        dst_row: &mut Self,
+        dst_lookup: &RLookup,
+        src_row: &Self,
+        src_lookup: &RLookup,
+    ) {
```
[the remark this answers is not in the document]
DarthB: Use member function. The ffi module will handle this

src/redisearch_rs/rlookup/src/row.rs
```diff
@@ -164,4 +164,51 @@ impl<'a, T: RSValueTrait> RLookupRow<'a, T> {
...
+    /// - `src_row`: The source row from which to copy values.
+    /// - `src_lookup`: The source lookup containing the schema of the source row.
+    ///
+    /// # Safety
+    ///
+    /// This function is marked as unsafe because it assumes that the provided lookups and rows are
+    /// valid and compatible.
+    ///
+    /// 1. The caller must ensure that `dst_lookup` is initialized and is compatible with `self`.
+    /// 2. The caller must ensure that `src_row` is initialized and is compatible with `src_lookup`.
```
JonasKruckenberg: use `associated` instead of `compatible` to use terminology established here https://github.com/RediSearch/RediSearch/pull/6884

src/redisearch_rs/rlookup/src/row.rs
```diff
@@ -164,4 +164,51 @@ impl<'a, T: RSValueTrait> RLookupRow<'a, T> {
...
+        let dst_row = self;
+
+        // NB: the `Iterator` impl for `Cursor` will automatically skip overridden keys
+        for src_key in src_lookup.cursor() {
+            // Get value from source row
+            if let Some(value) = src_row.get(src_key) {
+                // Find corresponding key in destination lookup
+                let dst_key = dst_lookup
+                    .find_key_by_name(src_key.name())
+                    .expect("we expect all source keys exist in destination")
```
kei-nan:
```suggestion
                    .expect("we expect all source keys to exist in destination")
```

src/redisearch_rs/rlookup/src/row.rs
```diff
@@ -164,4 +164,51 @@ impl<'a, T: RSValueTrait> RLookupRow<'a, T> {
...
+                    .into_current()
+                    .unwrap();
+
+                // Write fields to destination
+                dst_row.write_key(dst_key, value.to_owned());
```
kei-nan: do we still have ref count on the value or do we actually copy the value here?

src/redisearch_rs/rlookup/tests/row.rs
```diff
@@ -493,6 +493,338 @@ fn test_write_multiple_different_keys() {
...
+    let dst_key1 = dst_cursor1.into_current().unwrap();
+    let dst_cursor2 = dst_lookup.find_key_by_name(&key2_name).unwrap();
+    let dst_key2 = dst_cursor2.into_current().unwrap();
+
+    assert!(dst_row.get(dst_key1).is_none());
+    assert!(dst_row.get(dst_key2).is_none());
+}
+
+#[test]
+fn rlookuprow_write_fields_different_mapping() {
```
kei-nan: do we really need the rlookuprow prefix?
seems odd we are in the row file/context and still specify the object in the function name.

src/redisearch_rs/rlookup/tests/row.rs
```diff
@@ -493,6 +493,338 @@ fn test_write_multiple_different_keys() {
...
+    src_row.write_key_by_name(&mut src_lookup, key1_name.to_owned(), value1.clone());
+    src_row.write_key_by_name(&mut src_lookup, key2_name.to_owned(), value2.clone());
+    src_row.write_key_by_name(&mut src_lookup, key3_name.to_owned(), value3.clone());
+
+    // Create some dest keys first to ensure different indices
+    let other_key_name = CString::new("other_field").unwrap();
+    dst_lookup.get_key_write(other_key_name, RLookupKeyFlags::empty());
+
+    // Add source keys to destination (they'll have different dstidx values)
+    dst_lookup.get_key_write(key1_name.to_owned(), RLookupKeyFlags::empty());
```
kei-nan: maybe open the keys in a different order than the source?
not sure how the keys are stored inside but for our case we needed to handle it in the c code.
or maybe the second key can only exist in the destination and not the source?

src/redisearch_rs/rlookup/src/row.rs
```diff
@@ -164,4 +164,51 @@ impl<'a, T: RSValueTrait> RLookupRow<'a, T> {
...
+    ///
+    /// If a source key is not found in the destination lookup the function will panic (same as C behavior).
+    ///
+    /// If a source key has no value in the source row, it is skipped.
+    ///
+    /// # Arguments
+    ///
+    /// - `dst_lookup`: The destination lookup containing the schema of this row.
+    /// - `src_row`: The source row from which to copy values.
+    /// - `src_lookup`: The source lookup containing the schema of the source row.
```
kei-nan: maybe add something about src_lookup is expected to be different than dst_lookup?
or maybe we can add a debug assert to the code?
or does rust already ensures it using the `mut` mechanism?

kei-nan (changes requested):
Left some comments

src/redisearch_rs/rlookup/src/row.rs
JonasKruckenberg: This will increase the refcount the inner value is never copied 👍🏻 I changed the code to call the more accurate method.

src/redisearch_rs/rlookup/tests/row.rs
JonasKruckenberg: good call, went through all the tests in that module and updated their names accordingly 👍🏻

src/redisearch_rs/rlookup/src/row.rs
JonasKruckenberg: that is already ensured by the `&mut`, you can ever only have one mutable reference or many immutable ones but never a mix.

src/redisearch_rs/rlookup/tests/row.rs
JonasKruckenberg: it shouldn't make a difference but just to be sure I scrambled the order in which they are inserted 👍🏻

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -233,13 +233,15 @@ where
...
 {
-    let new_required = lookup.find_by_name(field_cstr).is_none();
+    let new_required = lookup.find_key_by_name(field_cstr).is_none();
     if new_required {
         let name = Cow::from(field_cstr.to_owned());
         lookup.get_key_write(name, RLookupKeyFlags::empty());
     }
 
-    let cursor = lookup.find_by_name(field_cstr).expect("key must exist now");
+    let cursor = lookup
```
kei-nan, line 242: why do we need to perform another lookup?
we can store the result from the first `find_key_by_name`
then use map so if it empty we create a new one, no?

kei-nan (approved):
Left one optional comment, since it isn't directly related to your code I'm approving the PR regardless.

src/redisearch_rs/rlookup/src/load_document/hash.rs
kei-nan: I understood you opened a ticket to improve this down the line
marking as resolved

kei-nan: I understood you opened a ticket to improve this down the line
marking as resolved

PR #6883, opened 2025-09-25
MOD-11582 Port `RLookupRow_Move` as `RLookupRow::move_fields_from`

JonasKruckenberg:
This change ports the C function `RLookupRow_Move`. This function is implemented as an FFI-only function since we expect this can be replaced by Rust builtins.

#### Mark if applicable

---

> [!NOTE]
> Introduces FFI-backed row field move (`RLookupRow_MoveFieldsFrom`) with Rust implementation, updates C/headers/call sites, adds tests, and refactors cursor usage to explicit traversal.
> 
> - **C API and Callers**:
>   - Rename and implement row move: `RLookupRow_Move` -> `RLookupRow_MoveFieldsFrom` in `rlookup.c`/`rlookup.h` and use it in `src/hybrid/hybrid_search_result.c`.
>   - Update generated header `rlookup_rs.h` signatures to use `RLookupRow *` and declare `RLookupRow_MoveFieldsFrom`.
> - **Rust FFI (rlookup_ffi)**:
>   - Export `row` module; add `type RLookupRow` alias.
>   - Add externs: `RLookup_WriteKey`, `RLookup_WriteOwnKey`, `RLookupRow_Wipe`, `RLookupRow_Reset`, and new `RLookupRow_MoveFieldsFrom` (bridges to Rust `move_fields_from`).
>   - New tests `c_entrypoint/rlookup_ffi/tests/row.rs` exercising `RLookupRow_MoveFieldsFrom` with mocked C functions.
> - **Rust rlookup**:
>   - Add `RLookupRow::move_fields_from` and `assert_valid`; adjust `copy_fields_from` to manual cursor traversal.
>   - Make `RLookup::id` and `RLookupKey::rlookup_id` public in debug builds.
>   - Remove `Iterator` impls for `Cursor`/`CursorMut`; switch to explicit `move_next` traversal.
> - **Value layer**:
>   - Extend `RSValueTrait` with `refcount()`; implement for `RSValueFFI` (reads C `_refcount`) and test mock.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d2dfed0317cdad7edf67eae6eeacbb51d81a0a87. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rlookup/src/lookup.rs
+++ b/src/redisearch_rs/rlookup/src/lookup.rs
@@ -455,7 +455,6 @@ impl<'a> RLookupKey<'a> {
         unsafe { Pin::new_unchecked(b) }
     }
 
-    #[cfg(any(debug_assertions, test))]
     pub fn is_overridden(&self) -> bool {
         let is_overridden = self.name.is_null();
 
--- a/src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
+++ b/src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
@@ -120,7 +120,7 @@ pub unsafe extern "C-unwind" fn RLookupRow_MoveFieldsFrom(
     let src = unsafe { src.expect("`src` must not be null").as_mut() };
     // Safety: ensured by caller (3.)
     let dst = unsafe { dst.expect("`dst` must not be null").as_mut() };
-    debug_assert!(dst.is_empty(), "expected `dst` to be pre-cleared");
+
     #[cfg(debug_assertions)]
     {
         assert_eq!(
--- a/src/redisearch_rs/rlookup/src/row.rs
+++ b/src/redisearch_rs/rlookup/src/row.rs
@@ -311,7 +311,7 @@ impl<'a, T: RSValueTrait> RLookupRow<'a, T> {
         }
     }
 
-    /// Move data from the source row into this row. The source row is cleared.
+    /// Move data from the source row to the destination row. The source row is cleared.
     pub fn move_fields_from(&mut self, src: &mut Self, lookup: &RLookup) {
         let mut c = lookup.cursor();
         while let Some(key) = c.current() {


src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -244,7 +244,7 @@ pub struct Cursor<'list, 'a> {
     current: Option<NonNull<RLookupKey<'a>>>,
 }
 
-/// A cursor over an [`RLookup`]s key list with editing operations.
+/// A cursor over a an [`RLookup`]s key list with editing operations.
```
Itzikvaknin: Previous was fine

src/redisearch_rs/c_entrypoint/rlookup_ffi/tests/row.rs
```diff
@@ -0,0 +1,218 @@
...
+    assert!(src.num_dyn_values() == 0);
+    let key = lookup
+        .get_key_read(c"foo", RLookupKeyFlags::empty())
+        .unwrap();
+    assert!(dst.get(&key).is_some());
+}
+
+#[test]
+#[cfg_attr(debug_assertions, should_panic)]
+fn rlookuprow_move_panics_on_non_empty_dst() {
```
Itzikvaknin: This isn't the C behavior
```
void RLookupRow_Move(const RLookup *lk, RLookupRow *src, RLookupRow *dst) {
  for (const RLookupKey *kk = lk->head; kk; kk = kk->next) {
    RSValue *vv = RLookup_GetItem(kk, src);
    if (vv) {
      RLookup_WriteKey(kk, dst, vv);
    }
  }
  RLookupRow_Wipe(src);
}
```

Overall i think its a good change, but need to check carefully if this invariant holds across the codebase

JonasKruckenberg: The doc comment for the C function says "The destination row should be pre-cleared (though its cache may still exist)." which I took to mean the row must be empty. This assertion is just codifying it.

But now I'm actually unsure because the only place this function is used https://github.com/RediSearch/RediSearch/blob/a5721cef0e36a8d65fd8bec8d55898ff2e2db3eb/src/hybrid/hybrid_search_result.c#L105-L120 moves into the destination without ever clearing it. So I don't really know what's going on.

The above function also doesn't seem to be covered by tests in a way that would trigger the assert.

PR #6884, opened 2025-09-25
chore: add debug-only `RLookup` branding to `RLookupKey` and `RLookupRow`

JonasKruckenberg:
This change adds a unique, debug-only ID to `RLookup` which gets tracked by all `RLookupKey`s created by the lookup and all `RLookupRow`s associated with the lookup.

Currently `RLookupKey`s are scoped to their "parent" `RLookup` and cannot be transferred. Likewise `RLookupRow`s are scoped to their associated `RLookup` because the `dstidx`/`svidx` values are only meaningful when tracked by the rlookup.

Note though, that multiple `RLookupRow`s can be associated with a single `RLookup`.

This change tracks this relation explicitly helping us to catch bugs.

#### Mark if applicable

---

> [!NOTE]
> Introduces a debug-only RLookupId branded through RLookup, RLookupKey, and RLookupRow (with parent assertions), updates constructors and call sites accordingly, excludes the ID from C bindings, and adds targeted tests.
> 
> - **rlookup (core)**:
>   - Add debug-only `RLookupId` and store it in `RLookup` (with `id()`), `RLookupKey`, and `RLookupRow`.
>   - Change constructors: `RLookupKey::new(parent, name, flags)` and `RLookupRow::new(&lookup)`; propagate branding and assert same-parent on `RLookupRow::write_key` (debug).
>   - Honor `NameAlloc` by cloning `name`/`path`; adjust `from_parts`; minor tweak to `get_key_load` cache lookup chaining.
> - **FFI**:
>   - Exclude `RLookupId` from `cbindgen` exports.
> - **Tests**:
>   - Update all sites to pass the parent `RLookup`; add branding and `NameAlloc` behavior tests; extend load/document tests accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8b7c08d7a3f532ee139f6903a06995f71b949937. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rlookup/tests/row.rs
+++ b/src/redisearch_rs/rlookup/tests/row.rs
@@ -552,16 +552,17 @@ fn row_panics_on_foreign_lookup() {
 
 /// Mock implementation of `IndexSpecCache_Decref` from spec.h for testing purposes
 #[unsafe(no_mangle)]
-extern "C" fn IndexSpecCache_Decref(s: Option<NonNull<ffi::IndexSpecCache>>) {
-    let s = s.unwrap();
+extern "C" fn IndexSpecCache_Decref(spcache: Option<NonNull<ffi::IndexSpecCache>>) {
+    let spcache = spcache.expect("`spcache` must not be null");
     let refcount = unsafe {
-        s.byte_add(offset_of!(ffi::IndexSpecCache, refcount))
+        spcache
+            .byte_add(offset_of!(ffi::IndexSpecCache, refcount))
             .cast::<usize>()
     };
 
     let refcount = unsafe { AtomicUsize::from_ptr(refcount.as_ptr()) };
 
     if refcount.fetch_sub(1, Ordering::Relaxed) == 1 {
-        drop(unsafe { Box::from_raw(s.as_ptr()) });
+        drop(unsafe { Box::from_raw(spcache.as_ptr()) });
     }
 }
--- a/src/redisearch_rs/rlookup/src/load_document/mod.rs
+++ b/src/redisearch_rs/rlookup/src/load_document/mod.rs
@@ -721,7 +721,7 @@ mod tests {
                 .build()?;
 
                 let mut lookup = RLookup::new();
-                let mut row = RLookupRow::new();
+                let mut row = RLookupRow::new(&lookup);
                 let res = load_document_int(&mut lookup, &mut row, &opt, &ctx);
                 if res != case.expected_error {
                     println!(
--- a/src/redisearch_rs/rlookup/src/rlookup_id.rs
+++ b/src/redisearch_rs/rlookup/src/rlookup_id.rs
@@ -10,6 +10,9 @@
 use core::fmt;
 use std::sync::atomic::{AtomicU64, Ordering};
 
+/// Unique identifier for a RLookup instance, encapsulated a u64.
+///
+/// This struct provides a method to generate the next unique identifier which is thread-safe.
 #[derive(Clone, Copy, Debug, Hash, Eq, PartialEq)]
 pub struct RLookupId(u64);
 


src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -290,6 +302,8 @@ impl<'a> RLookupKey<'a> {
             },
             _name: name,
             _path: None,
+            #[cfg(debug_assertions)]
+            rlookup_id,
         }
     }
```
enricozb, line 310: I wonder if it would result in less code if we did something like:

```rust
    pub fn new(
        name: &'a CStr,
        flags: RLookupKeyFlags,
    ) -> Self {
        let name = if flags.contains(RLookupKeyFlag::NameAlloc) {
            Cow::Owned(name.to_owned())
        } else {
        // ...
        
            _name: name,
            _path: None,
            #[cfg(debug_assertions)]
            RLookupId::stub(),
        }
    }
```

And a custom constructor that was conditionally compiled (e.g. `RLookupKey::with_id`) in places where the id should be set.

enricozb (dismissed):

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: yeahhh I wasn't super happy about it either, changed it so that the `new` method returns the placeholder and added an `with_rlookup_id` method to set the id 👍🏻

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -1407,7 +1459,14 @@ mod tests {
     #[test]
     fn key_from_parts_only_name() {
         let name = Cow::Borrowed(c"foo");
-        let key = RLookupKey::from_parts(name, Option::None, 0, RLookupKeyFlags::empty());
+        let key = RLookupKey::from_parts(
+            name,
+            Option::None,
```
enricozb: Do we need the `Option::` prefix here?

JonasKruckenberg: good catch, leftover from when it was `CBOption`, I thought I had removed them all but apparently not. should be fixed now

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -1108,7 +1160,11 @@ impl<'a> RLookup<'a> {
             .keys
             .find_by_name_mut(name)
             .expect("key should have been created above");
-        let key = if let Some(fs) = self.index_spec_cache.as_ref()?.find_field(name) {
+        let key = if let Some(fs) = self
```
JonasKruckenberg, line 1232: This change actually fixes a bug that we haven't caught yet. The C `findFieldInSpecCache` treats "no spcache" the same as "no such field in spcache". This early return (`?`) in the Rust version wasn't correct. 

I added a test further below to assert this correct behavior going forward.

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -341,19 +355,25 @@ impl<'a> RLookupKey<'a> {
...
+        self.rlookup_id
+    }
+
     /// Construct an `RLookupKey` from its main parts. Prefer Self::new if you are unsure which to use.
     fn from_parts(
         name: Cow<'a, CStr>,
         path: Option<Cow<'a, CStr>>,
         dstidx: u16,
         flags: RLookupKeyFlags,
+        #[cfg(debug_assertions)] rlookup_id: RLookupId,
```
JonasKruckenberg, line 370: I left the `cfg` argument here because this is supposed to be a raw internal constructor

kei-nan (approved):

src/redisearch_rs/rlookup/tests/row.rs
```diff
@@ -493,29 +520,48 @@ fn test_write_multiple_different_keys() {
...
+    // This should panic, key has a different RLookupId that Row!
+    row.write_key(&key, RSValueMock::create_num(42.0));
+}
+
 /// Mock implementation of `IndexSpecCache_Decref` from spec.h for testing purposes
 #[unsafe(no_mangle)]
-extern "C" fn IndexSpecCache_Decref(spcache: Option<NonNull<ffi::IndexSpecCache>>) {
-    let spcache = spcache.expect("`spcache` must not be null");
+extern "C" fn IndexSpecCache_Decref(s: Option<NonNull<ffi::IndexSpecCache>>) {
+    let s = s.unwrap();
```
enricozb: I feel like this piece of code was better before with `spcache` as the variable name and the `expect()` with an error message

DarthB: d0e5bed12ca3e9d46e5bb2f2a625f8e878c15c26

enricozb (approved):

enricozb (approved):

JonasKruckenberg:
Seems to be this commit here that breaks it https://github.com/RediSearch/RediSearch/commit/1727027fd2aa33d155ff7bd1b527693fd4d61707

DarthB:
> Seems to be this commit here that breaks it [1727027](https://github.com/RediSearch/RediSearch/commit/1727027fd2aa33d155ff7bd1b527693fd4d61707)

I think you misread a warning as an error. Let's hope the 6 prior pull requests don't require another rebase.

PR #6888, opened 2025-09-25
Fix test_hybrid_multithread flakiness

Itzikvaknin:
Add 
```
    # Drain the thread pool to make sure all jobs are done.
    env.expect(debug_cmd(), 'WORKERS', 'DRAIN').ok()
```
Before validating the threadpool stats to make sure they are stable 

Add `enableDebugCommand=True` to the environment creation since this test uses debug commands

### DIFF
--- a/tests/pytests/test_hybrid_multithread.py
+++ b/tests/pytests/test_hybrid_multithread.py
@@ -50,7 +50,7 @@ def setup_basic_index(env):
 # TODO: remove once FT.HYBRID for cluster is implemented
 @skip(cluster=True)
 def test_hybrid_multithread():
-    env = Env(moduleArgs='WORKERS 2 DEFAULT_DIALECT 2')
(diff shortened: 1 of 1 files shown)


tests/pytests/test_hybrid_multithread.py
```diff
@@ -50,7 +50,7 @@ def setup_basic_index(env):
 # TODO: remove once FT.HYBRID for cluster is implemented
 @skip(cluster=True)
 def test_hybrid_multithread():
-    env = Env(moduleArgs='WORKERS 2 DEFAULT_DIALECT 2')
+    env = Env(moduleArgs='WORKERS 2 DEFAULT_DIALECT 2', enableDebugCommand=True)
```
nafraf, line 53: Do we need `enableDebugCommand` ?

Itzikvaknin: Failed locally because of that, and we are using debug command there
Its a matter of tests global config Vs test local config
I prefer the later 
WDYT?

PR #6895, opened 2025-09-26
MOD-11004: Respect timeout in HYBRID

oshadmi:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

### DIFF
--- a/src/result_processor.c
+++ b/src/result_processor.c
@@ -1503,14 +1503,17 @@ typedef struct {
   RedisSearchCtx *nextThreadCtx;       // Downstream search context - used by the thread calling Next
   arrayof(SearchResult *) results;     // Array of pointers to SearchResult, filled by the depleting thread
   bool done_depleting;                 // Set to `true` when depleting is finished (under lock)
-  uint cur_idx;                        // Current index for yielding results
+  size_t cur_idx;                      // Current index for yielding results
   RPStatus last_rc;                    // Last return code from upstream
   bool first_call;                     // Whether the first call to Next has been made
   StrongRef sync_ref;                  // Reference to shared synchronization object (DepleterSync)
+
+  RedisSearchCtx *sctx;                // Search context (contains timeout info)
 } RPDepleter;
 
 /*
  * Shared synchronization object for all RPDepleter instances of a pipeline.
+ *
  * We have two main synchronization fronts:
  * 1. The pipeline thread should wake up once ANY depleter finishes depleting.
  *    For this, we have the shared condition variable `cond` and mutex `mutex`.
@@ -1524,10 +1527,12 @@ typedef struct {
 typedef struct {
   pthread_cond_t cond;
   pthread_mutex_t mutex;
-  uint num_depleters;      // Number of depleters to sync
+  unsigned int num_depleters;    // Number of depleters to sync
   atomic_int num_locked;   // Number of depleters that have locked the index
   bool index_released;     // Whether or not the index-spec has been released by the pipeline thread yet
   bool take_index_lock;    // Whether or not the depleter should take the index lock
+
+
 } DepleterSync;
 
 // Free function for DepleterSync
@@ -1539,7 +1544,7 @@ static void DepleterSync_Free(void *obj) {
 }
 
 // Create a new shared sync object for a pipeline
-StrongRef DepleterSync_New(uint num_depleters, bool take_index_lock) {
+StrongRef DepleterSync_New(unsigned int num_depleters, bool take_index_lock) {
   DepleterSync *sync = rm_calloc(1, sizeof(DepleterSync));
   pthread_cond_init(&sync->cond, NULL);
   pthread_mutex_init(&sync->mutex, NULL);
@@ -1560,6 +1565,8 @@ static void RPDepleter_Free(ResultProcessor *base) {
 
 /**
  * Background thread function: consumes all results from upstream and stores them in the results array.
+ *
+ * Checks for timeout before starting execution and relies on upstream timeout detection during processing.
  * Signals completion by setting done_depleting to `true` and broadcasting to condition variable.
  */
 static void RPDepleter_Deplete(void *arg) {
@@ -1567,6 +1574,12 @@ static void RPDepleter_Deplete(void *arg) {
   RPStatus rc;
 
   DepleterSync *sync = (DepleterSync *)StrongRef_Get(self->sync_ref);
+
+  // Check if timeout was exceeded before starting execution
+  if (self->sctx && TimedOut(&self->sctx->time.timeout) == TIMED_OUT) {
+    self->last_rc = RS_RESULT_TIMEDOUT;
+    goto signal_completion;
+  }
 
   if (sync->take_index_lock) {
     // Lock the index for read
@@ -1582,15 +1595,18 @@ static void RPDepleter_Deplete(void *arg) {
     r = rm_calloc(1, sizeof(*r));
   }
   rm_free(r);
+
+
+
   // Save the last return code from the upstream.
   self->last_rc = rc;
 
-  // Verify the index is unlocked (in case the pipeline did not release the lock,
-  // e.g., limit + no Loader)
+  // Unlock the index if we locked it
   if (sync->take_index_lock) {
     RedisSearchCtx_UnlockSpec(self->depletingThreadCtx);
   }
 
+signal_completion:
   // Signal completion under mutex protection
   pthread_mutex_lock(&sync->mutex);
   self->done_depleting = true;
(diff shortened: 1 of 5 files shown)


src/result_processor.c
```diff
@@ -1524,10 +1527,12 @@ typedef struct {
 typedef struct {
   pthread_cond_t cond;
   pthread_mutex_t mutex;
-  uint num_depleters;      // Number of depleters to sync
+  unsigned int num_depleters;    // Number of depleters to sync
   atomic_int num_locked;   // Number of depleters that have locked the index
   bool index_released;     // Whether or not the index-spec has been released by the pipeline thread yet
   bool take_index_lock;    // Whether or not the depleter should take the index lock
+
```
kei-nan: remove reduandnt newlines

src/result_processor.c
```diff
@@ -1772,7 +1821,17 @@ int RPDepleter_DepleteAll(arrayof(ResultProcessor*) depleters) {
   for (size_t i = 0; i < count; i++) {
     RPDepleter* depleter = (RPDepleter*)depleters[i];
     depleter->first_call = false;
-    RPDepleter_StartDepletionThread(depleter);
+
+    // Check timeout before attempting to start thread
+    if (depleter->sctx && TimedOut(&depleter->sctx->time.timeout) == TIMED_OUT) {
```
kei-nan: we potentially already started other deplete threads, you can't simply return without waiting for them.
This is very risky, I suggest you don't check for timeout inside the loop, only before it.

src/result_processor.c
```diff
@@ -1503,14 +1503,17 @@ typedef struct {
   RedisSearchCtx *nextThreadCtx;       // Downstream search context - used by the thread calling Next
   arrayof(SearchResult *) results;     // Array of pointers to SearchResult, filled by the depleting thread
   bool done_depleting;                 // Set to `true` when depleting is finished (under lock)
-  uint cur_idx;                        // Current index for yielding results
+  size_t cur_idx;                      // Current index for yielding results
   RPStatus last_rc;                    // Last return code from upstream
   bool first_call;                     // Whether the first call to Next has been made
   StrongRef sync_ref;                  // Reference to shared synchronization object (DepleterSync)
+
+  RedisSearchCtx *sctx;                // Search context (contains timeout info)
```
kei-nan: You alread have nextThreadCtx and depletingThreadCtx.
Please remove sctx and use the right search context for your purpose.

src/result_processor.c
```diff
@@ -1568,6 +1575,12 @@ static void RPDepleter_Deplete(void *arg) {
 
   DepleterSync *sync = (DepleterSync *)StrongRef_Get(self->sync_ref);
 
+  // Check if timeout was exceeded before starting execution
+  if (self->sctx && TimedOut(&self->sctx->time.timeout) == TIMED_OUT) {
```
kei-nan: you probably need to use the depleter search context here since you are in the depleter thread.

src/result_processor.c
```diff
@@ -1582,15 +1595,18 @@ static void RPDepleter_Deplete(void *arg) {
     r = rm_calloc(1, sizeof(*r));
   }
   rm_free(r);
+
```
kei-nan, line 1454: reduce amount of newlines

src/result_processor.c
```diff
@@ -1607,6 +1623,10 @@ static int RPDepleter_Next_Yield(ResultProcessor *base, SearchResult *r) {
   // Depleting thread is done, it's safe to return the results.
   if (self->cur_idx == array_len(self->results)) {
     // We've reached the end of the array, return the last code from the upstream.
+    // Handle timeout policy: if we timed out but have TimeoutPolicy_Return, return EOF to yield partial results
+    if (self->last_rc == RS_RESULT_TIMEDOUT && base->parent && base->parent->timeoutPolicy == TimeoutPolicy_Return) {
+      return RS_RESULT_EOF;  // Return EOF to indicate we're done yielding partial results
```
kei-nan: maybe you should also set the last_rc, not sure what it is being used for but it might be a logic loop hole.

src/result_processor.c
```diff
@@ -1719,8 +1761,15 @@ ResultProcessor *RPDepleter_New(StrongRef sync_ref, RedisSearchCtx *depletingThr
   ret->sync_ref = sync_ref;
   ret->depletingThreadCtx = depletingThreadCtx;
   ret->nextThreadCtx = nextThreadCtx;
+
+  // Initialize timeout fields
+  ret->sctx = nextThreadCtx;  // Store search context for timeout access
+
   // Make sure the sync reference is valid
   RS_LOG_ASSERT(StrongRef_Get(sync_ref), "Invalid sync reference");
+
```
kei-nan, line 1657: too many newlines

src/result_processor.c
```diff
@@ -1688,7 +1720,17 @@ static int RPDepleter_Next_Dispatch(ResultProcessor *base, SearchResult *r) {
...
     self->first_call = false;
-    RPDepleter_StartDepletionThread(self);
+
+    // Check timeout before attempting to start thread
+    if (self->sctx && TimedOut(&self->sctx->time.timeout) == TIMED_OUT) {
+      return RS_RESULT_TIMEDOUT;
+    }
+
+    // Try to start the depletion thread
+    if (!RPDepleter_StartDepletionThread(self)) {
```
kei-nan, line 1611: consider adding a test which will check what will happen in this case.
If it can't happen then maybe the code needs to go back to the way it was to make it easier to follow.

src/result_processor.c
```diff
@@ -1719,8 +1761,15 @@ ResultProcessor *RPDepleter_New(StrongRef sync_ref, RedisSearchCtx *depletingThr
   ret->sync_ref = sync_ref;
   ret->depletingThreadCtx = depletingThreadCtx;
   ret->nextThreadCtx = nextThreadCtx;
+
+  // Initialize timeout fields
+  ret->sctx = nextThreadCtx;  // Store search context for timeout access
```
kei-nan: I don't think you need this, also it is probably the wrong ctx when in the context of the depleter thread.

src/result_processor.c
```diff
@@ -1961,11 +2037,35 @@ static inline bool RPHybridMerger_Error(const RPHybridMerger *self) {
       // Store the final return code for this upstream
       self->upstreamReturnCodes[i] = rc;
 
-      // Currently continues processing other upstreams.
-      // TODO: Update logic to stop processing further results — we want to return immediately on timeout or error : MOD-11004
-      // Note: This processor might have rp_depleter as an upstream, which currently lacks a mechanism to stop its spawned thread before completion.
-      consumed[i] = true;
-      numConsumed++;
+      // Early termination on timeout or error for faster response
+      if (rc == RS_RESULT_TIMEDOUT || rc == RS_RESULT_ERROR) {
```
kei-nan: It looks like:
```
consumed[i] = true;
numConsumed++;
made_progress = true;
```
Also happens in the else flow, consider moving it before the if.

src/result_processor.c
```diff
@@ -1961,11 +2037,35 @@ static inline bool RPHybridMerger_Error(const RPHybridMerger *self) {
...
+        consumed[i] = true;
+        numConsumed++;
+        made_progress = true;
+
+        // If timeout policy is fail, we can terminate early
+        if (rc == RS_RESULT_TIMEDOUT && rp->parent->timeoutPolicy == TimeoutPolicy_Fail) {
```
kei-nan: Why are we exiting the loop without waiting for all the depleter threads.
We must wait for all the threads to complete.

This is very risky, I think we should rethink this flow.
Also the code is becoming very complex, lets try and keep it simple.
I don't see a point in marking everybody as timed out, just do another check in the end to see if someone timed out and put the relevant logic there.(but don't think you should modify the upstream return code)

src/result_processor.c
oshadmi: Replaced it with an assert at the end of the loop, but it seems too harsh.
Need to fail the command if one or more threads could not be added to the threadpool queue.
I will try to fail gracefully - need to avoid waiting for this unborn thread.

src/result_processor.c
oshadmi: Removed from here. Should not lose the information that we timed out.

src/result_processor.c
oshadmi: Simplified

src/result_processor.c
```diff
@@ -1397,26 +1398,33 @@ StrongRef DepleterSync_New(uint num_depleters, bool take_index_lock) {
...
+  self->cur_idx = 0;
+}
+
 /**
  * Destructor
  */
 static void RPDepleter_Free(ResultProcessor *base) {
   RPDepleter *self = (RPDepleter *)base;
-  array_free_ex(self->results, srDtor(*(char **)ptr));
+  // array_free_ex(self->results, srDtor(*(char **)ptr));
```
nafraf, line 1428: leftover?

oshadmi:
Replaced by #7035

PR #6898, opened 2025-09-27
Support multiple slot-ranges per shard - [MOD-11613]

GuyAv46:
### Implements the support for multiple slot ranges per shard.

Today, we almost don't care for the slots of the shard (most commands are not slot-bounded, only the deprecated `FT.ADD`, `FT.GET`, and `FT.DEL`), but we do need to know the sharding topology and each node's role.
Until now, the module has not supported multi-slot ranges per shard, although it was possible on OSS clusters. Our topology struct could only hold a single slot range per shard, and for every slot range, it assumes there is a different shard.

This PR refactors this structure, allowing multiple slot ranges per shard. It also updates all the places that use this struct, and most importantly, the `SEARCH.CLUSTER*` commands that create/serialize this struct.

Most of the code changes are around parsing the topology info into the struct. No API changes to the `SEARCH.CLUSTERSET` command (Enterprise), only serializing the info into the new structs. For OSS `SEARCH.CLUSTERREFRESH`, we now use internally `CLUSTER SHARDS` instead of `CLUSTER SLOTS` to get the topology information by shards, which is more convenient for storing it this way. We are also now aware of shards with no slots, and can send `FT.CREATE` and other schema commands to these shards even before they are assigned documents.

### Note
`SEARCH.CLUSTERSET` still doesn't support multiple slot ranges per shard, and assumes each node is unique.

1. `MRClusterShard` now holds an array of slot ranges, instead of a single start and end
2. `SEARCH.CLUSTERINFO` now returns an array of ranges per shard. The length of the reply still represents the number of shards and not the number of slot ranges

#### Mark if applicable

- [X] This PR introduces API changes

---

> [!NOTE]
> Refactors cluster topology and routing to handle multiple slot ranges per shard, switches command routing to shard indices, updates parsing to CLUSTER SHARDS, and adjusts CLUSTERINFO plus tests accordingly.
> 
> - **Topology & Routing**:
>   - Refactor `MRClusterShard` to hold multiple slot ranges (`ranges[]`), replacing `startSlot/endSlot`.
>   - Add `MRClusterShard_AddRange` and clone/free support for ranges.
>   - Switch command targeting from `targetSlot` to `targetShard` across iterator and routing paths.
>   - Update shard selection to respect `targetShard` or map by slot via new ranges lookup.
>   - Ensure master node is first in `nodes`; simplify master selection.
> - **Topology Discovery**:
>   - Parse OSS topology via `CLUSTER SHARDS` (new parsers), preferring `tls-port` and reordering master first.
> - **APIs & Replies**:
>   - Change `SEARCH.CLUSTERINFO` output from `slots` (start/end) to `shards` with an array of slot ranges and nodes; RESP2/3 adjusted.
> - **Misc**:
>   - Use `RedisModule_HoldString` in dict key dup.
>   - Logging/assertion updates and minor fixes.
> - **Tests**:
>   - Add tests for multiple slot ranges per shard and shardless nodes; update existing tests to new `CLUSTERINFO`/routing format.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 522e52311361333bd6751d2e60682bf50aed1b7a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/coord/rmr/cluster_topology.c
+++ b/src/coord/rmr/cluster_topology.c
@@ -16,14 +16,14 @@ void MRClusterShard_AddNode(MRClusterShard *sh, MRClusterNode *n) {
     sh->capNodes += 1;
     sh->nodes = rm_realloc(sh->nodes, sh->capNodes * sizeof(MRClusterNode));
   }
-  sh->nodes[sh->numNodes] = *n;
-  if (n->flags & MRNode_Master) {
+
+  if (n->flags & MRNode_Master && sh->numNodes > 0) {
     // Ensure that master nodes are always at the front of the list
-    if (sh->numNodes) {
-      MRClusterNode tmp = sh->nodes[sh->numNodes];
-      sh->nodes[sh->numNodes] = sh->nodes[0];
-      sh->nodes[0] = tmp;
-    }
+    sh->nodes[sh->numNodes] = sh->nodes[0];
+    sh->nodes[0] = *n;
+  } else {
+    // Otherwise, just append
+    sh->nodes[sh->numNodes] = *n;
   }
   sh->numNodes++;
 }


src/coord/rmr/cluster_topology.c
```diff
@@ -16,15 +16,36 @@ void MRClusterShard_AddNode(MRClusterShard *sh, MRClusterNode *n) {
...
+  if (sh->capRanges == sh->numRanges) {
+    sh->capRanges++;
+    sh->ranges = rm_realloc(sh->ranges, sh->capRanges * sizeof(mr_slot_range_t));
+  }
+  sh->ranges[sh->numRanges].start = start;
+  sh->ranges[sh->numRanges].end = end;
+  sh->numRanges++;
+}
+
+MRClusterShard MR_NewClusterShard(uint32_t capNodes) {
```
JoanFM, line 42: let's add capRanges here also. otherwise we are almost guaranteed to reallocate after

src/coord/rmr/cluster_topology.c
```diff
@@ -55,7 +76,12 @@ MRClusterTopology *MRClusterTopology_Clone(MRClusterTopology *t) {
   MRClusterTopology *topo = MR_NewTopology(t->numShards, t->numSlots, t->hashFunc);
   for (int s = 0; s < t->numShards; s++) {
     MRClusterShard *original_shard = &t->shards[s];
-    MRClusterShard new_shard = MR_NewClusterShard(original_shard->startSlot, original_shard->endSlot, original_shard->numNodes);
+    MRClusterShard new_shard = MR_NewClusterShard(original_shard->numNodes);
+    new_shard.numRanges = original_shard->numRanges;
+    new_shard.capRanges = original_shard->numRanges;
+    new_shard.ranges = rm_realloc(new_shard.ranges, new_shard.capRanges * sizeof(mr_slot_range_t));
```
JoanFM, line 82: can't we pass numRanges in constructor so that we already allocate them?

src/coord/rmr/command.c
```diff
@@ -58,7 +58,7 @@ static void MRCommand_Init(MRCommand *cmd, size_t len) {
   cmd->num = len;
   cmd->strs = rm_malloc(sizeof(*cmd->strs) * len);
   cmd->lens = rm_malloc(sizeof(*cmd->lens) * len);
-  cmd->targetSlot = -1;
+  cmd->targetShard = -1;
```
JoanFM: can we have a #define INVALID_SHARD -1 and use it?

src/coord/rmr/cluster.c
```diff
@@ -118,12 +111,19 @@ MRConn* MRCluster_GetConn(IORuntimeCtx *ioRuntime, bool mastersOnly, MRCommand *
 
   if (!ioRuntime->topo) return NULL;
 
-  /* Get the cluster slot from the sharder */
-  unsigned slot = getSlotByCmd(cmd, ioRuntime->topo);
+  MRClusterShard *sh;
+  if (cmd->targetShard != -1 && cmd->targetShard < ioRuntime->topo->numShards) {
```
JoanFM: use here the INVALID_SHARD (or UNINITIALIZED_SHARD naming)

JoanFM (changes requested):

src/coord/rmr/redise.c
```diff
@@ -20,16 +20,32 @@ typedef struct {
...
+        // New node in the same shard
+        MRClusterShard_AddNode(&t->shards[i], &sh->node);
+        return;
+      }
+    }
+    if (!(sh->node.flags & MRNode_Master)) {
+      continue; // Only master nodes can cause shard merging
+    }
+    // Check if the master node already exists in the shard, in which case we can just
+    for (int n = 0; n < t->shards[i].numNodes; n++) {
```
JoanFM: uint16_t?

src/coord/rmr/redise.c
```diff
@@ -20,16 +20,32 @@ typedef struct {
...
+        // Consume the node, we don't need to add it again
+        MRClusterNode_Free(&sh->node);
+        return;
+      }
     }
   }
 
   // New shard
-  MRClusterShard csh = MR_NewClusterShard(sh->startSlot, sh->endSlot, 2);
+  MRClusterShard csh = MR_NewClusterShard(2);
```
JoanFM, line 32: I know it ws here, but can we give a meaning to this magic number with a #define?

src/coord/rmr/rmr.c
```diff
@@ -490,17 +498,24 @@ void MR_ReplyClusterInfo(RedisModuleCtx *ctx, MRClusterTopology *topo) {
         MRClusterShard *sh = &topo->shards[i];
         RedisModule_Reply_Array(reply); // >shards
 
-        RedisModule_Reply_LongLong(reply, sh->startSlot);
-        RedisModule_Reply_LongLong(reply, sh->endSlot);
+        RedisModule_Reply_Array(reply); // >>>slots
+        for (int r = 0; r < sh->numRanges; r++) {
```
JoanFM: uint16_t

src/coord/rmr/rmr.c
```diff
@@ -490,17 +498,24 @@ void MR_ReplyClusterInfo(RedisModuleCtx *ctx, MRClusterTopology *topo) {
...
+          long long start = sh->ranges[r].start;
```
JoanFM: why don't we use unsigned here and also unsigned to respond?

src/coord/rmr/rmr.c
```diff
@@ -435,15 +435,23 @@ void MR_ReplyClusterInfo(RedisModuleCtx *ctx, MRClusterTopology *topo) {
     RedisModule_ReplyKV_LongLong(reply, "num_slots", topo ? (long long)topo->numSlots : 0);
 
     if (!topo) {
```
JoanFM, line 437: just commenting here, this I assume is a very well agreed breaking change right? We should make sure the replica node information goes away before release, so we do not have to keep it there. Will do

src/coord/rmr/redis_cluster.c
```diff
@@ -13,110 +13,174 @@
...
+      }
+    } else if (STR_EQ(key_str, key_len, "tls-port")) {
+      n->endpoint.port = (int)RedisModule_CallReplyInteger(val); // Prefer tls-port if available
+    } else if (STR_EQ(key_str, key_len, "port") && n->endpoint.port == -1) {
+      n->endpoint.port = (int)RedisModule_CallReplyInteger(val); // Only set if tls-port wasn't set
+    }
+  }
+}
+
+static bool hasNoSlots(const RedisModuleCallReply *shard) {
```
JoanFM: I tend to prefer positive logic (isEmpty())? or hasSlots and ! when u check the result?

src/coord/rmr/redis_cluster.c
```diff
@@ -13,110 +13,174 @@
...
+          6# "role" => "master"
+      2)  1# "id" => "6daa25c08025a0c7e4cc0d1ab255949ce6cee902"
+          2# "port" => (integer) 30005
+          3# "tls-port" => (integer) 40005
+          4# "ip" => "127.0.0.1"
+          5# "endpoint" => "localhost"
+          6# "role" => "replica"
+  */
+
+  size_t numShards = RedisModule_CallReplyLength(cluster_shards);
```
JoanFM, line 137: if numShards is 1, do we need to get topology from the start?

src/coord/rmr/redis_cluster.c
```diff
@@ -13,110 +13,174 @@
...
+  if (numShards == 1 && hasNoSlots(RedisModule_CallReplyArrayElement(cluster_shards, 0))) {
```
JoanFM: is this even possible?

src/coord/rmr/cluster_topology.c
GuyAv46: 1. It's not too common to have more than one slot range per shard
2. I'm going to remove the slot info collection in the following PR (#7020), so this array will be removed as well. I'd rather not reiterate on these parts until this, #7020, and another PR to collect only masters' info are merged

src/coord/rmr/cluster_topology.c
GuyAv46: https://github.com/RediSearch/RediSearch/pull/6898#discussion_r2413459877

src/coord/rmr/redise.c
GuyAv46: https://github.com/RediSearch/RediSearch/pull/6898#discussion_r2413459877
After the third PR is merged, we also won't need this capacity

src/coord/rmr/rmr.c
GuyAv46: Just didn't mean to change it. For the expected value range, it shouldn't matter.

src/coord/rmr/cluster.c
oshadmi: Should we log this anomaly?

src/coord/rmr/redis_cluster.c
oshadmi: Is zero length response still possible?

src/coord/rmr/redis_cluster.c
```diff
@@ -13,110 +13,174 @@
...
-    // parse the nodes
-    for (size_t n = 0; n < numNodes; n++) {
-      RedisModuleCallReply *nd = RedisModule_CallReplyArrayElement(e, n + 2);
-      // the array node must be 3 elements (future versions may add more...)
-      if (RedisModule_CallReplyLength(nd) < 3) {
-        goto err;
+  for (size_t i = 0; i < numShards; i++) {
+    RedisModuleCallReply *currShard = RedisModule_CallReplyArrayElement(cluster_shards, i);
+    RS_ASSERT(RedisModule_CallReplyType(currShard) == REDISMODULE_REPLY_ARRAY);
+    RS_ASSERT(RedisModule_CallReplyLength(currShard) == 4);
```
oshadmi: Comment why 4 ? (2 pairs of key & value)

src/coord/rmr/redis_cluster.c
```diff
@@ -13,110 +13,174 @@
...
+      const char *val_str = RedisModule_CallReplyStringPtr(val, &val_len);
+      if (STR_EQ(val_str, val_len, "master")) {
+        n->flags |= MRNode_Master;
+      }
+    } else if (STR_EQ(key_str, key_len, "tls-port")) {
+      n->endpoint.port = (int)RedisModule_CallReplyInteger(val); // Prefer tls-port if available
+    } else if (STR_EQ(key_str, key_len, "port") && n->endpoint.port == -1) {
+      n->endpoint.port = (int)RedisModule_CallReplyInteger(val); // Only set if tls-port wasn't set
+    }
+  }
```
oshadmi, line 68: Consider assert for missing mandatory fields (id, endpoint, , port)

src/coord/rmr/redise.c
```diff
@@ -20,16 +20,32 @@ typedef struct {
...
+        MRClusterShard_AddNode(&t->shards[i], &sh->node);
+        return;
+      }
+    }
+    if (!(sh->node.flags & MRNode_Master)) {
+      continue; // Only master nodes can cause shard merging
+    }
+    // Check if the master node already exists in the shard, in which case we can just
+    for (int n = 0; n < t->shards[i].numNodes; n++) {
+      if (!strcmp(sh->node.id, t->shards[i].nodes[n].id)) {
```
oshadmi: Should skip if `t->shards[i].role` is not master?

src/coord/rmr/rmr.c
```diff
@@ -453,17 +461,17 @@ void MR_ReplyClusterInfo(RedisModuleCtx *ctx, MRClusterTopology *topo) {
           REPLY_KVSTR_SAFE("id", node->id);
           REPLY_KVSTR_SAFE("host", node->endpoint.host);
           RedisModule_ReplyKV_LongLong(reply, "port", node->endpoint.port);
-          RedisModule_ReplyKV_SimpleStringf(reply, "role", "%s%s",                        // TODO: move the space to "self"
-                                      node->flags & MRNode_Master ? "master " : "slave ", // "master" : "slave",
-                                      node->flags & MRNode_Self ? "self" : "");           // " self" : ""
+          RedisModule_ReplyKV_SimpleStringf(reply, "role", "%s%s",
+                                      node->flags & MRNode_Master ? "master" : "slave",
```
oshadmi: Consider changing slave to replica

src/coord/rmr/redis_cluster.c
GuyAv46: A bit philosophical 😆 
If we have a valid single shard, we can avoid parsing its info (it's us, obviously), but I don't see any harm in storing it.
If we have an invalid single shard, it's a configuration issue (similar to an empty array in `CLUSTER SLOTS`, but `CLUSTER SHARDS` always returns at least one shard), and we need to report this.
Currently, if we return `NULL`, it indicates an error, and the cluster is uninitialized. I'd rather parse a single valid shard and return it to demonstrate we really have only a single shard. Then, the single shards optimizations will kick in, and we won't actually use the topology struct

src/coord/rmr/redis_cluster.c
GuyAv46: 1. YES - if the cluster is not set correctly. While `CLUSTER SLOTS` returns no slots in this case, `CLUSTER SHARDS` always returns some info about the self.
2. I don't think it's possible, but I'll make this condition fail in this case as well

src/coord/rmr/redise.c
GuyAv46: As discussed, I removed the support for multiple slot ranges per shard with `SEARCH.CLUSTERSET`, so this validation is not needed.

src/coord/rmr/cluster.c
```diff
@@ -118,12 +111,22 @@ MRConn* MRCluster_GetConn(IORuntimeCtx *ioRuntime, bool mastersOnly, MRCommand *
...
-  unsigned slot = getSlotByCmd(cmd, ioRuntime->topo);
+  MRClusterShard *sh;
+  if (cmd->targetShard != INVALID_SHARD && cmd->targetShard < ioRuntime->topo->numShards) {
+    /* Get the shard directly by the targetShard field */
+    sh = &ioRuntime->topo->shards[cmd->targetShard];
 
-  /* Get the shard from the slot map */
-  MRClusterShard *sh = _MRCluster_FindShard(ioRuntime->topo, slot);
-  if (!sh) return NULL;
+  } else {
```
JoanFM, line 119: can this else actually happen?

PR #6903, opened 2025-09-28
add wait for gc

BenGoldberger:
fix the gc oom flaky test


nafraf:
If the test is only flaky on Mac, maybe it's better to skip the FDs check only on Mac

BenGoldberger:
> If the test is only flaky on Mac, maybe it's better to skip the FDs check only on Mac

I think the test is problematic, because we have no control on the FDs redis opens, so it has potential to be flaky on every OS

PR #6914, opened 2025-09-28
check if alpine

meiravgri:
Currently, svs intel optimizations are not supported on alpine due to lack of glibc on this platform.

When running test_vecsim:test_svs_vamana_info_with_compression on intel and build defined to enable svs optimizations, we expect to succeed creating an index with optimizaed compression flavors such as LVQ, LeanVec etc...
But since those are not supported on alpine, we fallback to GlobalSQ.

This PR fixes the SVS info test to that into consideration.
If intel's optimizations support on alpine is implmneted in the future, we should remove this restriction from the test.

### DIFF
--- a/tests/pytests/test_vecsim.py
+++ b/tests/pytests/test_vecsim.py
@@ -2541,6 +2541,18 @@ def test_svs_vamana_info_with_compression():
         except (IOError, FileNotFoundError):
             return False
 
+    def is_alpine():
+        """Parse /etc/os-release into a dictionary (Linux only)"""
+        try:
+            with open("/etc/os-release") as f:
+                for line in f:
+                    if '=' in line:
+                        k, v = line.strip().split('=', 1)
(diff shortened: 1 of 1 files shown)


tests/pytests/test_vecsim.py
```diff
@@ -2532,6 +2541,18 @@ def is_intel_cpu():
         except (IOError, FileNotFoundError):
             return False
 
+    def is_alpine():
```
nafraf, line 2545: In `testParseTime()` we are also checking for distro
https://github.com/RediSearch/RediSearch/blob/a69c3b414e47c7ed3a4bda7ea840d3682918d14a/tests/pytests/test_aggregate.py#L152-L153

Can we unify the validation method?

meiravgri: checking! Thanks

nafraf (approved):

PR #6924, opened 2025-09-29
[MOD-11188] apply window parameter in Hybrid linear commands

JoanFM:
Allow Linear combination to apply Window.


src/result_processor.c
```diff
@@ -1957,7 +1956,7 @@ ResultProcessor *RPHybridMerger_New(HybridScoringContext *hybridScoringCtx,
      maximalSize = hybridScoringCtx->rrfCtx.window * numUpstreams;
    } else {
      // For LINEAR scoring, use a reasonable default for dictionary pre-sizing
-     maximalSize = 1000; // Conservative estimate for dictionary sizing
+     maximalSize = hybridScoringCtx->linearCtx.window * numUpstreams;
```
Itzikvaknin, line 1957: Can remove the comment above

tests/cpptests/test_cpp_hybrid_defaults.cpp
```diff
@@ -304,3 +291,82 @@ TEST_F(HybridDefaultsTest, testKAlreadyWithinWindow) {
...
+                      "LIMIT", "0", "50");
+
+  parseCommand(args);
+  ASSERT_EQ(parseCtx.hybridParams->scoringCtx->scoringType, HYBRID_SCORING_LINEAR);
+  // For LINEAR, WINDOW should use LIMIT fallback when not explicitly set
+  ASSERT_EQ(HYBRID_DEFAULT_WINDOW, parseCtx.hybridParams->scoringCtx->linearCtx.window)
+      << "Expected LINEAR window=" << HYBRID_DEFAULT_WINDOW << " (should use default, not LIMIT fallback), got " << parseCtx.hybridParams->scoringCtx->linearCtx.window;
+}
+
+// Test LINEAR explicit WINDOW overrides LIMIT
```
Itzikvaknin: Why do we need this test? To verify that `WINDOW` is set correctly when an explicit `LIMIT` is present?
If so lets change the name

tests/cpptests/test_cpp_hybrid_defaults.cpp
```diff
@@ -304,3 +291,82 @@ TEST_F(HybridDefaultsTest, testKAlreadyWithinWindow) {
...
+
+  QueryError status = {QueryErrorCode(0)};
+
+  ParseHybridCommandCtx cmd = {0};
+  cmd.search = result->requests[0];
+  cmd.vector = result->requests[1];
+  cmd.tailPlan = &result->tailPipeline->ap;
+  cmd.hybridParams = &hybridParams;
+  cmd.reqConfig = &result->reqConfig;
+  cmd.cursorConfig = &result->cursorConfig;
```
Itzikvaknin: Do we need this? why not following the test error structure?

JoanFM:
> Nice, Left some comments

Handled

tests/cpptests/test_cpp_hybrid_defaults.cpp
```diff
@@ -304,3 +291,59 @@ TEST_F(HybridDefaultsTest, testKAlreadyWithinWindow) {
...
+  RMCK::ArgvList args(ctx, "FT.HYBRID", index_name.c_str(),
+                      "SEARCH", "hello", "VSIM", "@vector", TEST_BLOB_DATA,
+                      "COMBINE", "LINEAR", "6", "ALPHA", "0.6", "BETA", "0.4", "WINDOW", "30");
+
+  parseCommand(args);
+  ASSERT_EQ(parseCtx.hybridParams->scoringCtx->scoringType, HYBRID_SCORING_LINEAR);
+  ASSERT_EQ(30, parseCtx.hybridParams->scoringCtx->linearCtx.window)
+      << "Expected LINEAR window=30, got " << parseCtx.hybridParams->scoringCtx->linearCtx.window;
+
+  // K should remain at default since LINEAR doesn't apply K≤WINDOW constraint
```
oshadmi, line 305: Is there a difference between RRF and LINEAR? Function `applyLimitKNNFallback` doesn't differntiate?

JoanFM: this will be removed because this limit ,K relation is being removed in #6930

oshadmi: What about relation between K and WINDOW - not the same for RRF and Linear?

JoanFM: I am unaware of such relationship. What is the desired behavior?

oshadmi: The comment says
>  // K should remain at default since LINEAR doesn't apply K≤WINDOW constraint

The relation is
KNN K <= WINDOW
But why should it be different between RRFand LINEAR?

Itzikvaknin: It shouldnt, its an optimization that doesnt affect the result set at all 
We can apply it to all combine methods

JoanFM: I did the change that applies to all

src/hybrid/parse_hybrid.c
```diff
@@ -486,9 +462,13 @@ static void applyLimitParameterFallbacks(AGGPlan *tailPlan, HybridPipelineParams
 static void applyKNNTopKWindowConstraint(ParsedVectorData *pvd,
                                  HybridPipelineParams *hybridParams) {
   // Apply K ≤ WINDOW constraint for RRF scoring to prevent wasteful computation
```
Itzikvaknin: Lets change the comment to 
```
// Apply K = min(K, WINDOW)  prevent wasteful computation
```

JoanFM:
```suggestion
  // Apply K = min(K, WINDOW)  prevents wasteful computation
```

tests/cpptests/test_cpp_hybrid_defaults.cpp
```diff
@@ -266,30 +266,16 @@ TEST_F(HybridDefaultsTest, testKFromLimitCappedAtExplicitWindow) {
   ASSERT_EQ(12, parseCtx.hybridParams->scoringCtx->rrfCtx.window);
 }
 
-// Test K ≤ WINDOW constraint: explicit K > WINDOW from LIMIT fallback should cap K to WINDOW
-TEST_F(HybridDefaultsTest, testExplicitKCappedAtWindowFromLimit) {
+// Test that Linear scoring is affected by K ≤ WINDOW constraint
```
Itzikvaknin:
```suggestion
// Test K = min{ K, WINDOW} optimization is used in LINEAR
```

Itzikvaknin (approved):

PR #6937, opened 2025-09-30
Sync tail RLookup with upstream RLookup after consuming results

Itzikvaknin:
In the coordinator setup, `RPNet->lookup` is constructed dynamically during execution, with each shard result contributing its own field keys to the schema.

Because of this, we cannot synchronize the RLookup instances in advance (at build time). Instead, we must perform the synchronization during execution.

Calling `RLookup_AddKeysFrom(fullyConsumedUpstreamLookup, tailLookup)` ensures that all keys written during RPNet processing are included in the upstream RLookup (the source lookup).

### DIFF
--- a/src/hybrid/hybrid_lookup_context.h
+++ b/src/hybrid/hybrid_lookup_context.h
@@ -27,7 +27,7 @@ extern "C" {
  */
 typedef struct {
   arrayof(const RLookup*) sourceLookups;  // Source lookups from each request
-  const RLookup *tailLookup;              // Unified destination lookup
+  RLookup *tailLookup;              // Unified destination lookup
 } HybridLookupContext;
 
 #ifdef __cplusplus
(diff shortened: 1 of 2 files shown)


src/hybrid/hybrid_lookup_context.h
```diff
@@ -27,7 +27,7 @@ extern "C" {
  */
 typedef struct {
   arrayof(const RLookup*) sourceLookups;  // Source lookups from each request
-  const RLookup *tailLookup;              // Unified destination lookup
+  RLookup *tailLookup;              // Unified destination lookup
```
nafraf: Fix comment alignment ? :)

nafraf (dismissed):

src/result_processor.c
```diff
@@ -1856,7 +1856,7 @@ static inline bool RPHybridMerger_Error(const RPHybridMerger *self) {
 
       // Store the final return code for this upstream
       self->upstreamReturnCodes[i] = rc;
-
+      RLookup_AddKeysFrom(self->lookupCtx->sourceLookups[i], self->lookupCtx->tailLookup, RLOOKUP_F_NOFLAGS);
```
oshadmi: Consider avoiding this on TIMEOUT on FAIL

PR #6938, opened 2025-09-30
[draft] - highlight issue

JoanFM:


tests/pytests/test_highlight.py
```diff
@@ -0,0 +1,143 @@
+# -*- coding: utf-8 -*-
+
+from includes import *
```
kei-nan, line 3: usually, we don't have the mod in the file name,, maybe as a function in the file itself.
I don't think we have a highlight test file.

src/forward_index.c
```diff
@@ -223,9 +223,15 @@ static void ForwardIndex_HandleToken(ForwardIndex *idx, const char *tok, size_t
 
 }
 
-int forwardIndexTokenFunc(void *ctx, const Token *tokInfo) {
+int forwardIndexTokenFunc(void *ctx, const Token *tokInfo, bool handle_only_highlighting_offsets) {
 #define SYNONYM_BUFF_LEN 100
   const ForwardIndexTokenizerCtx *tokCtx = ctx;
+  if (tokCtx->allOffsets && tokCtx->allOffsets->vw) {
```
kei-nan: This means that empty strings costs us additional memory, is there a way to avoid this?
I think the issue requires a bit more explanation/documentation.
The one comment above is not enough IMHO.

src/document.c
```diff
@@ -433,12 +433,9 @@ FIELD_PREPROCESSOR(fulltextPreprocessor) {
...
       Token tok = {0};
       while (0 != aCtx->tokenizer->Next(aCtx->tokenizer, &tok)) {
-        if (!indexesEmpty && tok.tokLen == 0) {
-          // Skip empty values if the field should not index them
-          // Empty tokens are returned only if the original value was empty
-          continue;
-        }
-        forwardIndexTokenFunc(&tokCtx, &tok);
+        // Decide whether tokenization needs to add empty tokens to forward index or they only need to handle the byte offsets needed for highlighting
+        bool handle_only_highlighting_offsets = (tok.tokLen == 0 && !indexesEmpty);
```
kei-nan: here I think the variable name doesn't fit the naming convention in this area of the code.

src/forward_index.c
JoanFM: Empty tokens cost us a VarInt more in the document metadata. but they are not indexed in the forward index.

I beieve you would somehow need to store this bit of information.

And given the actual Highlighting algorithm with the Offsets and Token count, it seems the safest way to go. (Considering that now I see a big risk of accessing memory out of bounds and the complexity of a potential rewrite).

will describe the investigation better

JoanFM: check the explanation added in the PR

src/document.c
JoanFM: changed in 7d438180c10a524d7c4a14049b36c1af9df65696

tests/pytests/test_highlight.py
JoanFM: changed in https://github.com/RediSearch/RediSearch/commit/7d438180c10a524d7c4a14049b36c1af9df65696

src/document.c
```diff
@@ -433,12 +433,9 @@ FIELD_PREPROCESSOR(fulltextPreprocessor) {
...
       Token tok = {0};
       while (0 != aCtx->tokenizer->Next(aCtx->tokenizer, &tok)) {
-        if (!indexesEmpty && tok.tokLen == 0) {
-          // Skip empty values if the field should not index them
-          // Empty tokens are returned only if the original value was empty
-          continue;
-        }
-        forwardIndexTokenFunc(&tokCtx, &tok);
+        // Decide whether tokenization needs to add empty tokens to forward index or they only need to handle the byte offsets needed for highlighting
+        bool handleOnlyOffsets = (tok.tokLen == 0 && !indexesEmpty);
```
kei-nan: maybe this logic can be moved inside `forwardIndexTokenFunc`?
Seems internal to the function behaviour.
Then you could add the comment to the `forwardIndexTokenFunc` documentation

JoanFM: I think then I would need to pass an `IndexesEmpty` flag, which makes `forward_index`aware of a feature that should belong to a higher abstraction. However, offsets handling seems to be a tokenizer/forward_Index responsability (this could be separated, but it is so)

kei-nan: I think the `forwardIndexTokenFunc` should see it was called with an empty token and handle it.
So on one end we keep the old code of:
```
if (!indexesEmpty && tok.tokLen == 0) {
```
At the caller level
But on the other hand we have the function check if the token was an empty string and handle it accordingly.

src/forward_index.c
kei-nan: PR description is much better,
But wonder if `forwardIndexTokenFunc` should have a bit of documentation for future generations.

JoanFM: 100% would have loved that

tests/pytests/test_highlight.py
```diff
@@ -0,0 +1,143 @@
...
+    conn.hset('doc:4153814', mapping={
+        'seattle': '',
+        'lisbon': 'my Dog sleeps',
+    })
+
+    result = env.cmd('FT.SEARCH', 'seattle_lisbon', 'Dog', 'LIMIT', '0', '1', 'SORTBY', 'lisbon', 'DESC', 'HIGHLIGHT')
+    verify_word_is_highlighted(env, result, '<b>Dog</b>')
+
+
+def test_highlight_empty_field_index_empty(env):
```
kei-nan, line 129: consider adding a test where we try and highlight an empty string

JoanFM: test exists on test_empty but will add a new one

tests/pytests/test_highlight.py
```diff
@@ -0,0 +1,189 @@
...
+              't', 'TEXT').ok()
+
+    waitForIndex(env, 't_idx')
+
+    conn.hset('doc:4153814', mapping={
+        't': '',
+    })
+
+    try:
+        _ = env.cmd('FT.SEARCH', 't_idx', '@t:("")', 'HIGHLIGHT', 'FIELDS', '1', 't')
```
kei-nan: I think we have expect().error() or something like that to test for failures

src/document.c
```diff
@@ -433,6 +439,8 @@ FIELD_PREPROCESSOR(fulltextPreprocessor) {
 
       Token tok = {0};
       while (0 != aCtx->tokenizer->Next(aCtx->tokenizer, &tok)) {
+        // Decide whether tokenization needs to add empty tokens to forward index or they only need to handle the byte offsets needed for highlighting
```
kei-nan: I think the comment need to be updated, here is an optional phrasing.

```suggestion
        // We always want to write the byte offset, even when string is empty since it is global across all fields
```

kei-nan (approved):

PR #6966, opened 2025-10-03
MOD-11660: Store KNN K in parsedVectorData

Itzikvaknin:
This is a preparation branch for the coordinator:

In the coordinator, a sorter is created for each subquery, and the heap size is defined as:
```
search: window
vsim: min(window, knn_k)
```
Our goal is to make it easier to access the `KNN K` data.

The alternative would be to navigate through:
`vsim_areq->ast->root->vn->vq->type` and
`vsim_areq->ast->root->vn->vq->knn->k`


JoanFM (commented):
cant we just have a macro accessing it?

Having duplicated info increases the risk of discrepancies IMO

Itzikvaknin:
Won't do

PR #6975, opened 2025-10-05
Avoid mock code in regular build

oshadmi:
Code branching on `RS_IsMock` is not required in production.
Adding a preprocessor flag `RS_UNIT_TESTS` for building unit-tests (when `TESTS=1`) that can be used to wrap mock code so it can be avoided in production.
 In mock code, still testing for `RS_IsMock` so it could also run against real Redis.

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

---

> [!NOTE]
> Wrap RS_IsMock-dependent paths with RS_UNIT_TESTS and add CMake define so mock logic is excluded in production but enabled for unit tests.
> 
> - **Build/Config**:
>   - Add `RS_UNIT_TESTS` compile definition when `BUILD_SEARCH_UNIT_TESTS` is enabled in `CMakeLists.txt`.
> - **Core (conditionally compiled under tests only)**:
>   - Guard `RS_IsMock` logic with `#ifdef RS_UNIT_TESTS` in `src/redisearch.h`, `src/gc.c`, `src/redis_index.c`, `src/spec.c`, and `src/util/timeout.h` to avoid mock behavior in production builds.
>   - Preserve mock-specific operations (e.g., GC timer bypass, mock GC stop, timeout short-circuit) only when unit tests are built.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d0bb8c225b19bbc806087047e784b52f71351c9e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -1999,11 +1999,9 @@ StrongRef IndexSpec_LoadUnsafeEx(IndexLoadOptions *options) {
     IndexSpec_IncreasCounter(sp);
   }
 
-  #ifndef RS_UNIT_TESTS
-  if ((sp->flags & Index_Temporary) && !(options->flags & INDEXSPEC_LOAD_NOTIMERUPDATE)) {
+  if (!RS_IsMock && (sp->flags & Index_Temporary) && !(options->flags & INDEXSPEC_LOAD_NOTIMERUPDATE)) {
     IndexSpec_SetTimeoutTimer(sp, StrongRef_Demote(spec_ref));
   }
-  #endif // RS_UNIT_TESTS
   return spec_ref;
 }
 
--- a/src/spec.c
+++ b/src/spec.c
@@ -361,16 +361,22 @@ static void IndexSpec_TimedOutProc(RedisModuleCtx *ctx, WeakRef w_ref) {
   RedisModule_Log(RSDummyContext, REDISMODULE_LOGLEVEL_VERBOSE, "Freeing index %s by timer", name);
 
   sp->isTimerSet = false;
+  #ifdef RS_UNIT_TESTS
   if (RS_IsMock) {
     IndexSpec_Free(sp);
   } else {
+    // We want the tests to be able to also run against real Redis
+  #endif // RS_UNIT_TESTS
     // called on master shard for temporary indexes and deletes all documents by defaults
     // pass FT.DROPINDEX with "DD" flag to self.
     RedisModuleCallReply *rep = RedisModule_Call(RSDummyContext, RS_DROP_INDEX_CMD, "cc!", HiddenString_GetUnsafe(sp->specName, NULL), "DD");
     if (rep) {
       RedisModule_FreeCallReply(rep);
     }
+  #ifdef RS_UNIT_TESTS
+  // Close the else block
   }
+  #endif // RS_UNIT_TESTS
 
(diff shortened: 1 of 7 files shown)


src/spec.c
```diff
@@ -361,16 +361,22 @@ static void IndexSpec_TimedOutProc(RedisModuleCtx *ctx, WeakRef w_ref) {
...
     // called on master shard for temporary indexes and deletes all documents by defaults
     // pass FT.DROPINDEX with "DD" flag to self.
     RedisModuleCallReply *rep = RedisModule_Call(RSDummyContext, RS_DROP_INDEX_CMD, "cc!", HiddenString_GetUnsafe(sp->specName, NULL), "DD");
     if (rep) {
       RedisModule_FreeCallReply(rep);
     }
+  #ifdef RS_UNIT_TESTS
+  // Close the else block
   }
+  #endif // RS_UNIT_TESTS
```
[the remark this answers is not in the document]
oshadmi: In unit tests both if and else parts can be done
In regular build only the else part is done.

src/spec.c
```diff
@@ -1991,9 +1999,11 @@ StrongRef IndexSpec_LoadUnsafeEx(IndexLoadOptions *options) {
     IndexSpec_IncreasCounter(sp);
   }
 
-  if (!RS_IsMock && (sp->flags & Index_Temporary) && !(options->flags & INDEXSPEC_LOAD_NOTIMERUPDATE)) {
+  #ifndef RS_UNIT_TESTS
+  if ((sp->flags & Index_Temporary) && !(options->flags & INDEXSPEC_LOAD_NOTIMERUPDATE)) {
     IndexSpec_SetTimeoutTimer(sp, StrongRef_Demote(spec_ref));
   }
+  #endif // RS_UNIT_TESTS
```
raz-mon: What about compiling with the tests (`RS_UNIT_TESTS=1`), but then running against Redis?
We will not have this code, that's a problem

raz-mon: Same for other `#ifndef RS_UNIT_TESTS`

raz-mon (commented):
IMHO it would be better to remove `RS_IsMock` and add it if/when we need it, but it's not critical so your call on that.
See the other comment, that I think is problematic

src/spec.c
oshadmi: It never ran against Redis, only against mock, so now it will not even be compiled - same effect, no?
It was doing
```if (!RS_IsMock && ...)```
And now it will be preprocessed out

oshadmi: Same for other `#ifndef RS_UNIT_TESTS`

raz-mon: Look at our CI for instance.
We compile the code with `make build TESTS=1`, such that `RS_UNIT_TESTS=1` as well.
We then run our regular pytests against this artifact --> We run against Redis with an artifact that doesn't have this code.
You're right that our production artifacts probably don't have this code, but still it doesn't make sense to remove this code under this condition, since you may want to run an artifact that you wanted to run unit-tests on, against Redis (i.e., load it to a redis server)

oshadmi: You're right - if we use the same built artifact for both the unit tests as well as for the pytests then the check must be done in runtime and cannot be done in compile time.

raz-mon (approved):
LGTM (I would remove the redundant ismock condition but not critical).

PR #6987, opened 2025-10-06
Fix command registration

GuyAv46:

Correctly register `FT.ADD`, `FT.GET`, and `FT.DEL` with their key location, removing the need to manually redirect them. Similar to the fix in #5184.

This fix also allows us to delete all slot calculation code, copied from the redis source.

### TODO:
- [ ] Don't store slots in the topology data structures
- [ ] Fix `SEARCH.CLUSTERINFO` reply accordingly

#### Mark if applicable


src/coord/rmr/redis_cluster.c
```diff
@@ -154,7 +141,7 @@ static MRClusterTopology *RedisCluster_GetTopology(RedisModuleCtx *ctx) {
 
     // Handle slots
     ASSERT_KEY(currShard, 0, "slots");
-    parseSlots(RedisModule_CallReplyArrayElement(currShard, 1), &topo->shards[i]);
+    // We don't actually use the slots, as we don't handle slot-level routing ourselves
```
alonre24, line 179: What was the original behavior? Why were the slots required previously?

src/coord/rmr/redise.c
```diff
@@ -73,7 +64,6 @@ MRClusterTopology *RedisEnterprise_ParseTopology(RedisModuleCtx *ctx, RedisModul
   const char *myID = NULL;                 // Mandatory. No default.
   size_t numShards = 0;                    // Mandatory. No default.
   size_t numSlots = 16384;                 // Default.
```
alonre24, line 63: Can `numSlots` be removed as well?

src/coord/rmr/redise.c
```diff
@@ -18,20 +18,11 @@ typedef struct {
...
-
-  for (size_t i = 0; i < t->numShards; i++) {
-    RS_ASSERT(t->shards[i].numRanges == 1); // We only support single-range shards using this method
-    if (sh->startSlot == t->shards[i].ranges[0].start && sh->endSlot == t->shards[i].ranges[0].end) {
-      // New node in the same shard
-      MRClusterShard_AddNode(&t->shards[i], &sh->node);
-      return;
-    }
-  }
+  RS_ASSERT(sh->node.flags & MRNode_Master);
```
alonre24: In redise.c you wrote that we don't mind for replicas using this command, but here in particular, we are asserting that the node is master. Seems inconsistent

src/coord/rmr/redise.c
```diff
@@ -170,6 +156,12 @@ MRClusterTopology *RedisEnterprise_ParseTopology(RedisModuleCtx *ctx, RedisModul
       ERROR_BADVAL("ADDR", addr);
       goto error;
     }
+    /* Optional MASTER */
+    if (!AC_AdvanceIfMatch(&ac, "MASTER")) {
+      // We don't care for replicas using this command anymore
+      MREndpoint_Free(&sh.node.endpoint);
```
alonre24, line 159: Why don't we need the endpoint in case of a replica?

src/coord/rmr/cluster.c
```diff
@@ -53,80 +38,17 @@ MRClusterNode *_MRClusterShard_SelectNode(MRClusterShard *sh, bool mastersOnly)
...
-    unsigned slot = getSlotByCmd(cmd, ioRuntime->topo);
-
-    /* Get the shard from the slot map */
-    sh = _MRCluster_FindShard(ioRuntime->topo, slot);
-    if (!sh) return NULL;
+  // No target shard, or out of range.
+  if (cmd->targetShard >= ioRuntime->topo->numShards) {
+    RedisModule_Log(RSDummyContext, "warning", "Command targetShard %d is out of bounds (numShards=%zu)", cmd->targetShard, ioRuntime->topo->numShards);
   }
+  if (cmd->targetShard == INVALID_SHARD || cmd->targetShard >= ioRuntime->topo->numShards) return NULL;
```
alonre24: Should we log the warning in other cases where we return NULL, rather than only when `cmd->targetShard >= ioRuntime->topo->numShards`?

tests/pytests/test.py
```diff
@@ -4109,7 +4109,7 @@ def cluster_set_test(env: Env):
     def verify_address(addr):
         try:
             with TimeLimit(10, f'Failed waiting cluster set command to be updated with the new IP address `{addr}`'):
-                while env.cmd('SEARCH.CLUSTERINFO')[9][1][1] != addr:
+                while env.cmd('SEARCH.CLUSTERINFO')[9][0][1] != addr:
```
alonre24: This is an internal command, correct? Let's make sure that we update relevant clients/docs if needed about this breaking change

src/coord/rmr/redis_cluster.c
GuyAv46: It was all part of the manual partitioning, and was left for command redirect.

src/coord/rmr/redise.c
GuyAv46: Done. This variable was left to keep the commands behavior, but no longer stored afterward

src/coord/rmr/redise.c
GuyAv46: We don't keep the replica info anymore.
We need to free the endpoint we just allocated before continuing.

src/coord/rmr/redise.c
```diff
@@ -170,19 +155,17 @@ MRClusterTopology *RedisEnterprise_ParseTopology(RedisModuleCtx *ctx, RedisModul
...
+      continue;
+    }
```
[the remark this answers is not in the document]
GuyAv46: This is dead code today, as the DMC doesn't send any replica info

tests/pytests/test.py
GuyAv46: It is undocumented, designed for the DMC, and has no interesting info except for our tests.

src/coord/rmr/cluster.c
GuyAv46: - `cmd->targetShard == INVALID_SHARD` can become an assertion now, we don't expect it anymore, and don't have how to handle it here.
- No topology - shouldn't be possible because the IO won't reach here if it doesn't have a topology (but maybe there is an edge I can't think about, so I won't assert it now)
- It's also possible that there is no available connection, but we don't need to log this here (we have logs when the connection is dropped)

PR #7001, opened 2025-10-07
[MOD-11535] Define FFI crate and type definitions for rust port of `RSValue`

hdoordt:
Define an initial type layout for Rust `RsValue` and `SharedRsValue`, and create `value_ffi` crate as the FFI specification for `value`.

- Move `RSValueTrait` and `RSValueFFI` in the `value` crate
   to a separate module, and put them behind a feature gate
   as the module depends on the `ffi` crate which conflicts
   with cbindgen when it attempts to generate bindings for
   the `value` crate. This feature is to be removed
   when `RsValueFFI` is eliminated in a later stage of the
   porting process.
 - Define `RsValue` and `SharedRsValue` types, both still
   incomplete and unoptimized, so that we can generate bindings
   referencing these types. `RsValue` is a stack-based type,
   whereas `SharedRsValue` is heap-allocated and reference
   counted.
 - Define `RsValueMap`, `RsValueTrio`, as well as `RsValueInternal`
      to represent the internals of `RsValue` and `SharedRsValue`.
- Create `value_ffi` to mimic the C `value.h` interface.
    `RSValue_NewWithType`, `RSValue_WithType`, and
    `RSValue_NewStringWithType` are omitted intentionally,
    and are to be replaced with specialized constructors
    that fully initialized the produced value.


src/redisearch_rs/c_entrypoint/value_ffi/src/shared.rs
```diff
@@ -0,0 +1,201 @@
...
+use redis_module::RedisModuleString;
+use value::{RsValueInternal, map::RsValueMap, shared::SharedRsValue};
+
+/// Creates a heap-allocated `RsValue` wrapping a string.
+/// Doesn't duplicate the string. Use strdup if the value needs to be detached.
+/// @param str The string to wrap (ownership is transferred)
+/// @param len The length of the string
+/// @return A pointer to a heap-allocated RsValue
+#[unsafe(no_mangle)]
+pub extern "C" fn SharedRsValue_NewString(str: Option<NonNull<c_char>>, len: u32) -> SharedRsValue {
```
enricozb, line 25: What's the benefit of Option<NonNull<c_char>> over *const c_char? It looks like some of these funcs use `*const c_char` and others use `Option<NonNull<c_char>>`.

src/redisearch_rs/c_entrypoint/value_ffi/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+edition.workspace = true
+license-file.workspace = true
+publish.workspace = true
+
+[lints]
+workspace = true
+
+[build-dependencies]
+cbindgen.workspace = true
+build_utils = { path = "../../build_utils" }
```
enricozb, line 13: should this be a workspace dependency?

src/redisearch_rs/c_entrypoint/value_ffi/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+/// # Safety
+/// - The passed string pointer must point to a valid C string that
+///   was allocated using `rm_malloc`
+/// - The passed length must match the length to the string.
+///
+/// @param str The malloc'd string to wrap (ownership is transferred)
+/// @param len The length of the string
+/// @return A stack-allocated `RsValue` of type `RsValueType_String` with `RSString_Malloc` subtype
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RsValue_String(str: Option<NonNull<c_char>>, len: u32) -> RsValue {
```
enricozb, line 54: I'm not sure if `Option<NonNull<c_char>>` has a stable ABI.

src/redisearch_rs/value/src/shared.rs
```diff
@@ -0,0 +1,163 @@
...
+/// A heap-allocated and refcounted RedisSearch dynamic value.
+/// This type is backed by [`Arc<RsValueInternal>`], but uses
+/// the NULL pointer to encode an undefined value, and is FFI safe.
+///
+/// # Invariants
+/// - If this pointer is non-NULL, it was obtained from `Arc::into_raw`.
+/// - If it is NULL, it represents an undefined value.
+/// - A non-null pointer represents one clone of said `Arc`, and as such, as
+///   long as the [`SharedRsValue] lives and holds a non-null pointer, the Arc
+///   is still valid.
```
enricozb: nit: Can we put attributes after (doc) comments?

src/redisearch_rs/c_entrypoint/value_ffi/src/shared.rs
hdoordt: `Option<NonNull<c_char>>` forces you to check whether the pointer was null, and is therefore less error-prone. `NonNull` essentially being `* mut`, it can't be used to replace `* const` unfortunately

src/redisearch_rs/c_entrypoint/value_ffi/src/lib.rs
hdoordt: According to the [Nomicon](https://doc.rust-lang.org/nomicon/ffi.html#the-nullable-pointer-optimization), the null-pointer optimization makes this valid

src/redisearch_rs/c_entrypoint/value_ffi/Cargo.toml
hdoordt: I agree that's probably better. However, the other FFI-crates would need to do this as well. I'll open a separate PR for that.

PR #7007, opened 2025-10-08
[MOD-11541] Create `OwnedSds`, a rusty wrapper around `hiredis/sds.h`

hdoordt:
Create `OwnedSds`, a rusty wrapper around `hiredis/sds.h`

- Create `sds_rs` crate, that defines the `OwnedSds` type
  which provides an idiomatic Rust wrapper around SDS
- Add `Sds` variant to `RsValueInternal` in `value` crate,
  holding an `OwnedSds`.
- Update `ffi` crate to generate bindings for `hiredis/sds.h`

Based on #7001 

---

> [!NOTE]
> Introduces `sds_rs::OwnedSds` and `RsValueInternal::Sds`, generates `hiredis/sds.h` bindings, updates headers/FFI, and adjusts build/coverage/linking.
> 
> - **Value/Types**:
>   - Add `Sds` variant to `RsValueInternal` holding `sds_rs::OwnedSds`.
>   - Update C header `headers/value.h`: include `hiredis/sds.h`, add `RsValueType_Sds`, `OwnedSds` typedef, and union field.
> - **New crate**:
>   - Introduce `sds_rs` (Rust wrapper around `hiredis` SDS): defines `OwnedSds`, `Clone`/`Drop`/`Send`/`Sync`; conditional linking via `build.rs` (`unittest` feature).
>   - Wire into workspace and `value` crate deps (incl. dev `unittest` feature).
> - **FFI/Bindings**:
>   - `ffi/build.rs`: add include path and header for `deps/hiredis/sds.h`.
>   - `c_entrypoint/value_ffi/cbindgen.toml`: include `hiredis/sds.h`; parse `sds_rs`.
> - **Build system**:
>   - `build.sh` (coverage): append `RUSTFLAGS` linker arg `-C link-arg=--coverage` for gcov runtime.
>   - `build_utils`: refactor to `bin_root()` and `target_os()` helpers; reuse in `link_static_libraries`.
>   - `build/hiredis/hiredis.cmake`: remove debug postfix for `hiredis`/`hiredis_ssl` artifacts.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4c46976e84ba719862951a4714e676aaae4a8df6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -581,6 +581,13 @@ run_rust_tests() {
       --ignore-filename-regex="varint_bencher/*,trie_bencher/*,inverted_index_bencher/*"
       --output-path=$BINROOT/rust_cov.info
     "
+    # C libraries (like hiredis) are built with gcov coverage (--coverage flag) while Rust uses LLVM coverage.
+    # We need to link with the gcov runtime to resolve __gcov_* symbols from C libraries.
+    # Use RUSTFLAGS to pass linker arguments that will work across platforms:
+    # - Linux with GCC: --coverage expands to -lgcov
+    # - macOS with Clang: --coverage links the profile runtime
+    # - Both: --coverage is portable and matches what CMake uses for C code
+    export RUSTFLAGS="${RUSTFLAGS:+${RUSTFLAGS} }-C link-arg=--coverage"
   elif [[ -n "$SAN" || "$RUN_MIRI" == "1" ]]; then # using `elif` as we shouldn't run with both
     RUST_EXTENSIONS="+$NIGHTLY_VERSION miri"
   fi
--- a/src/redisearch_rs/sds_rs/src/lib.rs
+++ b/src/redisearch_rs/sds_rs/src/lib.rs
@@ -107,3 +107,8 @@ unsafe impl Send for OwnedSds {}
 /// Safety: `OwnedSds` is safe to share across threads because it owns its
 /// backing memory buffer and ensures that no other references exist to it.
 unsafe impl Sync for OwnedSds {}
+
+#[cfg(all(test, not(feature = "unittest")))]
+compile_error!(
+    "When building tests, enable the `unittest` feature to ensure C dependencies get linked"
+);
--- a/build/hiredis/hiredis.cmake
+++ b/build/hiredis/hiredis.cmake
@@ -5,6 +5,11 @@ option(DISABLE_TESTS "If tests should be compiled or not" ON)
 
 add_subdirectory(${root}/deps/hiredis hiredis)
 
+# Override hiredis debug postfix to prevent 'd' suffix on debug builds
+# since we already separate debug and release artifacts into different directories
+set_target_properties(hiredis PROPERTIES DEBUG_POSTFIX "")
+set_target_properties(hiredis_ssl PROPERTIES DEBUG_POSTFIX "")
+
 if(APPLE)
     include_directories(/usr/local/opt/openssl/include)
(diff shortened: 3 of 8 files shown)


build.sh
```diff
@@ -581,6 +581,13 @@ run_rust_tests() {
       --ignore-filename-regex="varint_bencher/*,trie_bencher/*,inverted_index_bencher/*"
       --output-path=$BINROOT/rust_cov.info
     "
+    # C libraries (like hiredis) are built with gcov coverage (--coverage flag) while Rust uses LLVM coverage.
+    # We need to link with the gcov runtime to resolve __gcov_* symbols from C libraries.
+    # Use RUSTFLAGS to pass linker arguments that will work across platforms:
+    # - Linux with GCC: --coverage expands to -lgcov
+    # - macOS with Clang: --coverage links the profile runtime
+    # - Both: --coverage is portable and matches what CMake uses for C code
+    export RUSTFLAGS="${RUSTFLAGS:+${RUSTFLAGS} }-C link-arg=--coverage"
```
kei-nan, line 590: can't we control how we want to build hiredis?

src/redisearch_rs/sds_rs/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+///   backing memory buffer.
+/// - The wrapped [`sds`] points to a null-terminated C string.
+/// - The wrapped [`sds`] is valid for reads of bytes at least until and including
+///   its null terminator.
+/// - The null terminator is within `isize::MAX` from the address the wrapped [`sds`]
+///   points to, i.e. the start of the actual C string.
+#[repr(transparent)]
+pub struct OwnedSds(sds);
+
+impl OwnedSds {
```
kei-nan, line 36: Instead of relying on hiredis, can't we implement our own sds?
Why do we need to have an sds object, can't we just implement the sds api?
I guess I'm not sure why we need a special type to represent a string.

build.sh
kei-nan: and not sure why we need hiredis in the first place.

kei-nan (changes requested):
I'm not sure why the internal rust type needs to interact with hiredis in the first place.
Anyhow left some initial comments

build.sh
hdoordt: We're wrapping `sds`, which is in `hiredis`. Are you suggesting to just building `sds.c`?

src/redisearch_rs/sds_rs/src/lib.rs
hdoordt: The reason is that while a Rust port of `sds` would be nice, we can never get rid of the C implementation as it lives in `hiredis`. As such, we'd be stuck with multiple implementations of which the internals may diverge

> I guess I'm not sure why we need a special type to represent a string.

The reason is that an `sds` is not just a string. Not only does it have the variable length header, it needs to be freed using `sdsfree

kei-nan: I see we have 22 places where sds.h is included.
If we remove the include and provide the header file rust generates then can't that work?
Meaning implement `sdsfree` ffi binding or something like that?
Then we could hide the sds implementation in rust entirely.

hdoordt:
Closed in favor of #7183

PR #6921, opened 2025-09-29
[MOD-11395] Handle shard returning OOM in Coord

lerman25:
This PR builds on the work introduced in #6877 by extending OOM (Out of Memory) handling beyond the standalone (SA) setup to also support cluster deployments.

OOM checks are applied to `FT.SEARCH`, `FT.AGGREGATE`, and `FT.HYBRID` commands. In a cluster environment, the behavior on OOM can be configured to one of three modes: `IGNORE`, `FAIL`, or `RETURN`.

FAIL: The cluster returns an OOM error if either the cluster as a whole or any shard runs out of memory.

RETURN: If the cluster is in OOM, it returns an error. Otherwise, it returns results from any shards that are not in OOM.

Note - for FT.HYBRID, only SA behavior is implemented, so there is only shard-level OOM guardrail

---

> [!NOTE]
> Add OOM policy support across standalone and cluster paths (SEARCH/AGGREGATE/HYBRID), propagate shard OOMs with policy-aware errors or warnings, and surface them in replies and profiles.
> 
> - **Core execution**:
>   - Introduce `CommonPipelineCtx { timeoutPolicy, timeout, oomPolicy }` and update `startPipelineCommon(...)` to aggregate when `TimeoutPolicy_Fail` or `OomPolicy_Fail`.
>   - Move OOM heuristic `estimateOOM` to `module.c` and apply guardrail before executing `FT.SEARCH`/`FT.AGGREGATE`/`FT.HYBRID` (incl. coord entry points).
> - **Coordinator**:
>   - Add `extractQueryErrorFromReply` and `should_return_error` to classify shard errors (timeout/OOM/generic) and decide early bail-out vs continue.
>   - On shard OOM with return policy, continue and set warning; with fail, return error. Support RESP2/RESP3 and profile collection (clone error replies for profile).
>   - Extend reducers to track `queryOOM` in `searchRequestCtx` and print `QUERY_WOOM_CLUSTER` in results/profile.
> - **Replies & profiling**:
>   - Add OOM warning plumbing: `QUERY_WOOM_CLUSTER`, `QueryError._queryOOM` + helpers, and `ProfilePrinterCtx.queryOOM`. Surface warnings in RESP2/3 results and profile maps.
> - **Aggregate/Hybrid**:
>   - Wire OOM policy through aggregate and hybrid pipelines; update profile context setup.
> - **RMR**:
>   - Add `MRReply_Clone` for string/error replies.
> - **Tests/Utils**:
[...]

PR #7017, opened 2025-10-09
[MOD-10774] - test: add test validating crash from different timeout policies

JoanFM:

Tests that #6921 fixes MOD-10774

---

> [!NOTE]
> Adds a regression test that configures mixed shard timeout policies and verifies FT.AGGREGATE does not crash the coordinator.
> 
> - **Tests**:
>   - **New test** `tests/pytests/test_cluster_aggregate_timeout.py`:
>     - Configures shard 1 with `search-on-timeout=return`; others with `search-on-timeout=fail` and `search-timeout=1`.
>     - Creates `idx` schema, bulk-loads ~20k docs via pipelined `HSET`.
>     - Runs `FT.AGGREGATE` grouping by `@category` with `COUNT` and `AVG @price` and asserts results are returned (no coordinator crash).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f4eec47f7aa72a27b710ea6322f332d5078709e7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_cluster_aggregate_timeout.py
+++ b/tests/pytests/test_cluster_aggregate_timeout.py
@@ -67,7 +67,6 @@ def test_cluster_aggregate_with_shards_timeout(env):
         # Execute pipeline every batch_size docs to avoid memory issues
         if (i + 1) % batch_size == 0:
             pipeline.execute()
-            pipeline = conn.pipeline(transaction=False)
 
     # Execute remaining docs
     pipeline.execute()


tests/pytests/test_cluster_aggregate_timeout.py
```diff
@@ -0,0 +1,69 @@
...
+            conn.execute_command('CONFIG', 'SET', 'search-on-timeout', 'return')
+        else:
+            # All other shards use FAIL policy
+            conn.execute_command('CONFIG', 'SET', 'search-on-timeout', 'fail')
+            conn.execute_command('CONFIG', 'SET', 'search-timeout', '1')
+
+@skip(cluster=False)
+def test_cluster_aggregate_with_shards_timeout(env):
+    if not env.isCluster():
+        env.skip()
```
GuyAv46: redundant

tests/pytests/test_cluster_aggregate_timeout.py
```diff
@@ -0,0 +1,69 @@
...
+                           'title', title,
+                           'price', price,
+                           'category', category,
+                           'brand', brand,
+                           'rating', rating,
+                           'stock', stock,
+                           'tags', tags,
+                           'created_date', created_date)
+
+    waitForIndex(env, 'idx')
```
GuyAv46: redundant (we add documents after creating the schema, so the indexing is complete when the `HSET` command returns

tests/pytests/test_cluster_aggregate_timeout.py
```diff
@@ -0,0 +1,69 @@
...
+
+    info = index_info(env, 'idx')
+    env.assertGreater(int(info['num_docs']), 0)
```
GuyAv46: Consider summing over all the shards?

tests/pytests/test_cluster_aggregate_timeout.py
```diff
@@ -0,0 +1,69 @@
...
+
+    result = env.cmd('FT.AGGREGATE', 'idx', '*',
+                    'GROUPBY', '1', '@category',
+                    'REDUCE', 'COUNT', '0', 'AS', 'count',
+                    'REDUCE', 'AVG', '1', '@price', 'AS', 'avg_price')
+
+    env.assertGreater(len(result), 1)  # Should have results, at least the coordinator does not timeout
```
GuyAv46: If we can't have a better check, let's at least have the result as part of the failure message
```suggestion
    env.assertGreater(len(result), 1, message=result)  # Should have results, at least the coordinator does not timeout
```

tests/pytests/test_cluster_aggregate_timeout.py
```diff
@@ -0,0 +1,62 @@
...
+    for shard_id in range(1, env.shardsCount + 1):
+        conn = env.getConnection(shard_id)
+
+        if shard_id == 1:
+            # First shard uses RETURN policy
+            conn.execute_command('CONFIG', 'SET', 'search-on-timeout', 'return')
+        else:
+            # All other shards use FAIL policy
+            conn.execute_command('CONFIG', 'SET', 'search-on-timeout', 'fail')
+            conn.execute_command('CONFIG', 'SET', 'search-timeout', '1')
```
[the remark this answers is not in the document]
JoanFM: this is the point of the test

tests/pytests/test_cluster_aggregate_timeout.py
```diff
@@ -0,0 +1,62 @@
...
+               'created_date', 'NUMERIC').ok()
+
+    waitForIndex(env, 'idx')
+
+    conn = getConnectionByEnv(env)
+    num_docs = 100000
+
+    brands = ['Apple', 'Samsung', 'Sony', 'LG', 'Dell', 'HP', 'Nike', 'Adidas', 'Zara', 'H&M']
+
+    for i in range(num_docs):
```
kei-nan, line 46: consider using transactions for quicker hsets
with this amount of docs I think it is important for this test to run as fast as possible.

If you want to simulate timeouts then maybe you can use the `TIMEOUT_AFTER_N` keyword in the FT.DEBUG command, instead of using a large dataset.

JoanFM: I am not convinced.

I do not want the timeout to happen in every shard and with this TIMEOUT_AFTER_N this seems to be applied to every shard,and then the code path does not go where the crash happened, so it is useless test

kei-nan: okay

tests/pytests/test_cluster_aggregate_timeout.py
```diff
@@ -0,0 +1,80 @@
...
+                                'brand', brand,
+                                'rating', rating,
+                                'stock', stock,
+                                'tags', tags,
+                                'created_date', created_date)
+
+        # Execute pipeline every batch_size docs to avoid memory issues
+        if (i + 1) % batch_size == 0:
+            pipeline.execute()
+            pipeline = conn.pipeline(transaction=False)
```
GuyAv46: No need to recreate the pipeline after each execution
```suggestion
```

PR #7021, opened 2025-10-09
[MOD-11664] Handle _INDEX_PREFIXES in hybrid distributed queries

JoanFM:
> [!NOTE]
> Adds `_INDEX_PREFIXES` propagation in coordinator MR commands and validates query/index prefix coherence across shards, with parser support and tests.
> 
> - **Hybrid/Coordinator**:
>   - `HybridRequest_buildMRCommand`: append `_INDEX_PREFIXES <n> <prefix...>` when `IndexSpec` has prefixes.
>   - `parse_hybrid.c`: collect prefixes from optional args and validate against `spec->rule->prefixes` via `IsIndexCoherentWithQuery`; fail with `QUERY_EMISSMATCH` on mismatch.
>   - Optional args/parsing: add `_INDEX_PREFIXES` support (`handleIndexPrefixes`) and plumb through `HybridParseContext`.
> - **Aggregate**:
>   - Init `AREQ` with `req->prefixesOffset = 0`; parse `_INDEX_PREFIXES` to record offset for later index-coherence check.
> - **Tests**:
>   - Add Python tests `test_hybrid_prefixes.py` for cross-shard prefix (in)compatibility.
>   - Update C++ coord tests to expect `_INDEX_PREFIXES` and to build specs with prefixes; add `index_utils.{h,cpp}` changes and include in `CMakeLists.txt`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 438e8e1008673de4662b8abd1c9c774f212efcf2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/cpptests/coord_tests/test_cpp_hybrid_build_mr_cmd.cpp
+++ b/tests/cpptests/coord_tests/test_cpp_hybrid_build_mr_cmd.cpp
@@ -66,7 +66,7 @@ protected:
         // Verify transformation: FT.HYBRID -> _FT.HYBRID
         EXPECT_STREQ(xcmd.strs[0], "_FT.HYBRID");
 
-        // Verify all other original args are preserved (except first)
+        // Verify all other original args are preserved (except first). Attention: This is not true if TIMEOUT is not at the end before DIALECT
         for (size_t i = 1; i < inputArgs.size(); i++) {
             EXPECT_STREQ(xcmd.strs[i], inputArgs[i]) << "Argument at index " << i << " should be preserved";
         }
@@ -75,8 +75,6 @@ protected:
         EXPECT_STREQ(xcmd.strs[xcmd.num - 3], "WITHCURSOR") << "WITHCURSOR should be fifth to last";
         EXPECT_STREQ(xcmd.strs[xcmd.num - 2], "WITHSCORES") << "WITHSCORES should be fourth to last";
         EXPECT_STREQ(xcmd.strs[xcmd.num - 1], "_NUM_SSTRING") << "_NUM_SSTRING should be third to last";
-        printArgvList(args, args.size());
-        printMRCommand(&xcmd);
 
         MRCommand_Free(&xcmd);
     }
@@ -96,18 +94,17 @@ protected:
       ASSERT_NE(sp, nullptr) << "IndexSpec should be accessible from RefManager";
       ASSERT_NE(sp->rule, nullptr) << "IndexSpec should have a rule";
       ASSERT_NE(sp->rule->prefixes, nullptr) << "IndexSpec rule should have prefixes";
+      ASSERT_EQ(array_len(sp->rule->prefixes), 2) << "IndexSpec rule should have 2 prefixes";
 
       // Build MR command
       MRCommand xcmd;
-      HybridRequest_buildMRCommand(args, args.size(), &xcmd, NULL, nullptr, &hybridParams);
+      HybridRequest_buildMRCommand(args, args.size(), &xcmd, NULL, sp, &hybridParams);
       // Verify transformation: FT.HYBRID -> _FT.HYBRID
       EXPECT_STREQ(xcmd.strs[0], "_FT.HYBRID");
-
-      // Verify all other original args are preserved (except first)
+        // Verify all other original args are preserved (except first). Attention: This is not true if TIMEOUT is not at the end before DIALECT
       for (size_t i = 1; i < inputArgs.size(); i++) {
           EXPECT_STREQ(xcmd.strs[i], inputArgs[i]) << "Argument at index " << i << " should be preserved";
       }
-
       // Verify WITHCURSOR, WITHSCORES, _NUM_SSTRING, _INDEX_PREFIXES, and prefix count are added at the end
       EXPECT_STREQ(xcmd.strs[xcmd.num - 7], "WITHCURSOR") << "WITHCURSOR should be seventh to last";
       EXPECT_STREQ(xcmd.strs[xcmd.num - 6], "WITHSCORES") << "WITHSCORES should be sixth to last";
@@ -116,8 +113,6 @@ protected:
       EXPECT_STREQ(xcmd.strs[xcmd.num - 3], "2") << "Prefix count should be third to last";
       EXPECT_STREQ(xcmd.strs[xcmd.num - 2], "prefix1") << "First prefix should be second to last";
       EXPECT_STREQ(xcmd.strs[xcmd.num - 1], "prefix2") << "Second prefix should be last";
-      printArgvList(args, args.size());
-      printMRCommand(&xcmd);
 
       // Clean up
       MRCommand_Free(&xcmd);
@@ -212,8 +207,8 @@ TEST_F(HybridBuildMRCommandTest, testComplexCommandParamsAfterTimeout) {
         "FT.HYBRID", "test_idx", "SEARCH", "@title:($param1)",
         "VSIM", "@vector_field", "$BLOB",
         "COMBINE", "LINEAR", "4", "ALPHA", "0.7", "BETA", "0.3",
-        "TIMEOUT", "3000",
         "PARAMS", "4", "param1", "hello", "BLOB", TEST_BLOB_DATA,
+        "TIMEOUT", "3000",
         "DIALECT", "2"
     });
     testCommandTransformationWithIndexSpec({


src/coord/hybrid/dist_hybrid.c
```diff
@@ -90,6 +90,17 @@ void HybridRequest_buildMRCommand(RedisModuleString **argv, int argc,
   // Numeric responses are encoded as simple strings.
   MRCommand_Append(xcmd, "_NUM_SSTRING", strlen("_NUM_SSTRING"));
 
+  MRCommand_Append(xcmd, "_INDEX_PREFIXES", strlen("_NUM_SSTRING"));
```
JoanFM: will try to abstract as common code with aggregate

src/hybrid/parse/hybrid_optional_args.c
```diff
@@ -124,6 +119,11 @@ int HybridParseOptionalArgs(HybridParseContext *ctx, ArgsCursor *ac, bool intern
                           ARG_OPT_OPTIONAL, ARG_OPT_END);
 
         //TODO : add INDEX_PREFIXES
```
JoanFM:
```suggestion
```

src/hybrid/parse/hybrid_callbacks.h
```diff
@@ -108,6 +108,12 @@ void handleWithScores(ArgParser *parser, const void *value, void *user_data);
  */
 void handleExplainScore(ArgParser *parser, const void *value, void *user_data);
 
+/**
+ * _INDEX_PREFIXES callback - handles _INDEX_PREFIXES prefix [prefix ...]
+ * Validates and sets index prefixes for hybrid request
```
JoanFM:
```suggestion
 * sets index prefix offset for later validation if needed
```

src/coord/hybrid/dist_hybrid.c
JoanFM: this is not easy as Aggregate does nbot use the MRCOmmand Append but an array, and I do not want to touch that

src/hybrid/parse/hybrid_callbacks.c
```diff
@@ -405,3 +405,11 @@ void handleNumSString(ArgParser *parser, const void *value, void *user_data) {
     ctx->specifiedArgs |= SPECIFIED_ARG_NUM_SSTRING;
 }
 
+// _INDEX_PREFIXES callback - implements EXACT original logic from handleIndexPrefixes
+void handleIndexPrefixes(ArgParser *parser, const void *value, void *user_data) {
+  HybridParseContext *ctx = (HybridParseContext*)user_data;
+  ArgsCursor *paramsArgs = (ArgsCursor*)value;
+  QueryError *status = ctx->status;
+  size_t numPrefixes = paramsArgs->argc;
+  ctx->prefixesOffset = parser->cursor->offset - numPrefixes - 2; // one is for the offset moving and the other for the number of prefixes
```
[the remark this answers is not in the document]
JoanFM: test proves otherwise

src/hybrid/parse_hybrid.c
```diff
@@ -505,6 +505,35 @@ static PLN_LoadStep *createImplicitLoadStep(void) {
...
 }
 
+static bool isIndexCoherentWithQuery(HybridParseContext *ctx, ArgsCursor *ac, IndexSpec *spec)  {
+  if (ctx->prefixesOffset == 0) {
+    // No prefixes in the query --> No validation needed.
+    return true;
+  }
+
+  sds *args = (sds *)ac->objs;
+  long long n_prefixes = strtol(args[ctx->prefixesOffset + 1], NULL, 10);
```
Itzikvaknin: whats this 10 ?

tests/pytests/test_hybrid_prefixes.py
```diff
@@ -0,0 +1,49 @@
+from common import *
+
+@skip(cluster=False, min_shards=2)
```
Itzikvaknin, line 3: ```@skip(cluster=False, min_shards=2)``` this is actually dormant,
have you @JoanFM  tested it manually?

src/hybrid/parse_hybrid.c
oshadmi: The base

src/hybrid/parse_hybrid.c
```diff
@@ -505,6 +505,35 @@ static PLN_LoadStep *createImplicitLoadStep(void) {
...
+  uint base_idx = ctx->prefixesOffset + 2;
+  for (uint i = 0; i < n_prefixes; i++) {
+    sds arg = args[base_idx + i];
+    if (HiddenUnicodeString_CompareC(spec_prefixes[i], arg) != 0) {
+      // Unmatching prefixes
+      return false;
+    }
+  }
+
+  return true;
```
oshadmi: Can this code be common with `IsIndexCoherent(AREQ *req)` ?
A common function can receive an `IndexSpec*` and `sds*` and do the same?

src/coord/hybrid/dist_hybrid.c
```diff
@@ -90,6 +90,22 @@ void HybridRequest_buildMRCommand(RedisModuleString **argv, int argc,
   // Numeric responses are encoded as simple strings.
   MRCommand_Append(xcmd, "_NUM_SSTRING", strlen("_NUM_SSTRING"));
 
+  MRCommand_Append(xcmd, "_INDEX_PREFIXES", strlen("_INDEX_PREFIXES"));
```
nafraf: Why do we need to send `_INDEX_PREFIXES 0`, can't we avoid to add it if there is no indexes?

tests/pytests/test_hybrid_prefixes.py
JoanFM: what do u mean dormant? I did test it manually but will double check

src/hybrid/parse_hybrid.c
JoanFM: yes will try this, first will make every test pass

src/coord/hybrid/dist_hybrid.c
JoanFM: I will try. Not sure if the Optional args will complain. But will check

tests/pytests/test_hybrid_prefixes.py
Itzikvaknin: My bad, thought it skips cluster

src/hybrid/parse_hybrid.c
```diff
@@ -505,6 +505,35 @@ static PLN_LoadStep *createImplicitLoadStep(void) {
     return implicitLoadStep;
 }
 
+static bool isIndexCoherentWithQuery(HybridParseContext *ctx, ArgsCursor *ac, IndexSpec *spec)  {
+  if (ctx->prefixesOffset == 0) {
+    // No prefixes in the query --> No validation needed.
+    return true;
+  }
+
+  sds *args = (sds *)ac->objs;
```
kei-nan: Maybe use AC_GetLongLong instead?
You can use ArgCursor_Slice or something like that.
try avoiding doing direct casts on ac->objs.

tests/pytests/test_hybrid_prefixes.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # Query via the cluster connection, such that we will get the mismatch error
+    commands = [
+        ['FT.HYBRID', index_name, 'SEARCH', 'text', 'VSIM', '@embedding', '$BLOB', 'PARAMS', '2', 'BLOB', '0']
+    ]
+
+    # Run commands on second shard (different index prefixes -> error)
+    for command in commands:
+        try:
+            second_conn.execute_command(*command)
```
kei-nan, line 36: maybe use expect() and error() methods instead of the exception catching?

JoanFM: ok, copied from the test that tests for search command (in test.py)

JoanFM: I see is not possible, because we wan to force sending to the second shard for which u use the connection that does not have this API

src/hybrid/parse_hybrid.c
JoanFM: After @kei-nan suggestions and the proper use of ArgCursor to handle the iteration, this does not seem so easy to do without quite some refactor in aggregate_request which should be left out of the scope now.

So I think we should not do

src/hybrid/parse_hybrid.c
JoanFM: Good, this helped me understand better the ArgCursor

src/coord/hybrid/dist_hybrid.c
JoanFM: done

src/coord/hybrid/dist_hybrid.c
```diff
@@ -90,6 +90,20 @@ void HybridRequest_buildMRCommand(RedisModuleString **argv, int argc,
   // Numeric responses are encoded as simple strings.
   MRCommand_Append(xcmd, "_NUM_SSTRING", strlen("_NUM_SSTRING"));
 
+  if (sp && sp->rule && sp->rule->prefixes) {
```
oshadmi: Also check that `array_len(sp->rule->prefixes)` is > 0 ?
So we don't just add `_INDEX_PREFIXES 0`

src/hybrid/parse/hybrid_optional_args.h
```diff
@@ -55,6 +55,7 @@ typedef struct {
     RequestConfig *reqConfig;               // Request configuration for DIALECT/TIMEOUT
     QEFlags *reqFlags;                      // Request flags
     size_t *maxResults;                     // Maximum results
+    arrayof(const char*) prefixes; // Prefixes for the index
```
oshadmi: Indent the comment

src/coord/hybrid/dist_hybrid.c
nafraf: Done.

src/hybrid/parse/hybrid_optional_args.h
nafraf: Fixed.

src/hybrid/parse/hybrid_optional_args.c
```diff
@@ -123,7 +123,11 @@ int HybridParseOptionalArgs(HybridParseContext *ctx, ArgsCursor *ac, bool intern
                           ARG_OPT_CALLBACK, handleNumSString, ctx,
                           ARG_OPT_OPTIONAL, ARG_OPT_END);
 
-        //TODO : add INDEX_PREFIXES
+        ArgParser_AddSubArgsV(parser, "_INDEX_PREFIXES", "Index prefixes",
+                             &subArgs, 1, -1,
+                             ARG_OPT_OPTIONAL,
```
kei-nan, line 128: consider removing `ARG_OPT_OPTIONAL`

src/hybrid/hybrid_request.c
```diff
@@ -343,6 +343,9 @@ HybridRequest *MakeDefaultHybridRequest(RedisSearchCtx *sctx) {
   extern size_t NumShards;  // Declared in module.c
   AREQ *search = AREQ_New();
   AREQ *vector = AREQ_New();
+  // Prefixes validation will be done w.r.t Hybrid Request
+  search->prefixesOffset = 0;
```
kei-nan: maybe this can move into `AREQ_New`?

src/hybrid/parse_hybrid.c
```diff
@@ -505,6 +505,38 @@ static PLN_LoadStep *createImplicitLoadStep(void) {
     return implicitLoadStep;
 }
 
+// This cannot be easily merged with IsIndexCoherent from aggregate_request.c since aggregate does not use ArgsCursor
+static bool IsIndexCoherentWithQuery(arrayof(const char*) prefixes, IndexSpec *spec)  {
```
kei-nan, line 510: consider seeing the comment is still true, maybe we can avoid code duplication.

src/hybrid/parse_hybrid.c
```diff
@@ -551,6 +583,9 @@ int parseHybridCommand(RedisModuleCtx *ctx, ArgsCursor *ac,
   searchRequest->ast.validationFlags |= QAST_NO_VECTOR;
   vectorRequest->ast.validationFlags |= QAST_NO_WEIGHT | QAST_NO_VECTOR;
 
+  // Prefixes for the index
+  arrayof(const char*) prefixes = array_new(const char*, 0);
```
kei-nan: consider removing this variable and just using the variable in `hybridParseCtx`

src/hybrid/parse_hybrid.c
```diff
@@ -675,6 +713,14 @@ int parseHybridCommand(RedisModuleCtx *ctx, ArgsCursor *ac,
   // Apply KNN K ≤ WINDOW constraint after all argument resolution is complete
   applyKNNTopKWindowConstraint(vectorRequest->parsedVectorData, hybridParams);
 
+  if (!IsIndexCoherentWithQuery(hybridParseCtx.prefixes, parsedCmdCtx->search->sctx->spec)) {
+    QueryError_SetError(status, QUERY_EMISSMATCH, NULL);
```
kei-nan, line 716: we need an error here

src/hybrid/parse_hybrid.c
JoanFM: still true

src/hybrid/parse_hybrid.c
JoanFM: the same as for Regular queries, QUERY_EMISSMATCH

JoanFM: what do u mean here?

src/hybrid/parse_hybrid.c
```diff
@@ -575,10 +610,14 @@ int parseHybridCommand(RedisModuleCtx *ctx, ArgsCursor *ac,
       .cursorConfig = parsedCmdCtx->cursorConfig,
       .reqConfig = parsedCmdCtx->reqConfig,
       .maxResults = &maxHybridResults,
+      .prefixes = prefixes,
```
kei-nan:
```suggestion
      .prefixes = &prefixes,
```

src/hybrid/parse_hybrid.c
```diff
@@ -575,10 +610,14 @@ int parseHybridCommand(RedisModuleCtx *ctx, ArgsCursor *ac,
...
   };
+  // may change prefixes in internal array_ensure_append_1
   if (HybridParseOptionalArgs(&hybridParseCtx, ac, internal) != REDISMODULE_OK) {
+    prefixes = hybridParseCtx.prefixes;
```
kei-nan:
```suggestion
```

src/hybrid/parse/hybrid_optional_args.c
JoanFM: it is optional right? why would  we remove it? what should we use instead?

src/hybrid/parse_hybrid.c
```diff
@@ -575,10 +610,14 @@ int parseHybridCommand(RedisModuleCtx *ctx, ArgsCursor *ac,
...
     goto error;
   }
+  prefixes = hybridParseCtx.prefixes;
```
kei-nan:
```suggestion
```

src/hybrid/parse/hybrid_optional_args.h
```diff
@@ -55,6 +55,7 @@ typedef struct {
     RequestConfig *reqConfig;               // Request configuration for DIALECT/TIMEOUT
     QEFlags *reqFlags;                      // Request flags
     size_t *maxResults;                     // Maximum results
+    arrayof(const char*) prefixes;          // Prefixes for the index
```
kei-nan:
```suggestion
    arrayof(const char*) *prefixes;          // Prefixes for the index
```

tests/cpptests/index_utils.h
```diff
@@ -31,7 +31,7 @@ InvertedIndex *createPopulateTermsInvIndex(int size, int idStep, int start_with=
  * To get the spec object (not safe), call get_spec(ism);
  * To free the spec and its resources, call freeSpec;
  */
-RefManager *createSpec(RedisModuleCtx *ctx);
+RefManager *createSpec(RedisModuleCtx *ctx, const char **prefixes = nullptr, int nprefixes = 0);
```
kei-nan: maybe use arrayof to make nprefixes redundant?
or use std::vector since we are in cpp world?

tests/cpptests/test_cpp_forkgc.cpp
```diff
@@ -88,7 +89,8 @@ class FGCTest : public ::testing::Test {
   pthread_t thread;
 
   void SetUp() override {
-    ism = createSpec(ctx);
+    Initialize_KeyspaceNotifications();
```
kei-nan, line 92: do we know why we need this inside a test?

src/hybrid/parse/hybrid_optional_args.c
Itzikvaknin: It is not optional, required for ACL

src/hybrid/parse_hybrid.c
kei-nan: But I don't see any usage of ArgCursor in either functions.

src/hybrid/parse_hybrid.c
kei-nan: I guess it is a bit odd to see but okay.
Usually used to seeing an error.

tests/cpptests/test_cpp_forkgc.cpp
JoanFM: because the test uses the path of FT.ADD which calls HSET internally

JoanFM: this was added to master in another PR

src/hybrid/parse_hybrid.c
JoanFM: oh fuck, this was changed yes. Sorry I lost track of the changes

src/hybrid/parse/hybrid_optional_args.c
JoanFM: but what if there are no prefixes, then we do not pass it

src/hybrid/parse_hybrid.c
JoanFM: I like that it is a constant and not a magic String

src/hybrid/parse_hybrid.c
JoanFM: I think is still complex to change it (would need refactor on aggregate side, I will add a TODO comment perhaps)

tests/cpptests/index_utils.h
JoanFM: done, thx

src/hybrid/parse/hybrid_optional_args.c
kei-nan: According to @raz-mon it should always be provided.

src/hybrid/parse_hybrid.c
```diff
@@ -705,6 +749,8 @@ int parseHybridCommand(RedisModuleCtx *ctx, ArgsCursor *ac,
   return REDISMODULE_OK;
 
 error:
+  array_free(prefixes);
```
kei-nan, line 752: shouldn't you check if prefixes is null?
since there is a possible flow where we release them after using them in `IsIndexCoherentWithQuery`.

src/hybrid/parse/hybrid_optional_args.c
JoanFM: I do not see why to be hones. if the sp doe snot have prefixes, we do not add it. 

So if we remove this we need to force _INDEX_PREFIXES "0" which seems like a waster and noisy

src/hybrid/parse_hybrid.c
JoanFM: array_free checks against NULL already, I can add the double check though

JoanFM:
I am waiting for the TimeOut fix to be able to add a test of the happy path of the feature

kei-nan (approved):

kei-nan (approved):

src/hybrid/parse_hybrid.c
```diff
@@ -505,6 +505,39 @@ static PLN_LoadStep *createImplicitLoadStep(void) {
...
 }
 
+// This cannot be easily merged with IsIndexCoherent from aggregate_request.c since aggregate request parses prefixes differently.
+// Unifying would require some refactor on the aggregate flow.
+static bool IsIndexCoherentWithQuery(arrayof(const char*) prefixes, IndexSpec *spec)  {
+
+  size_t n_prefixes = array_len(prefixes);
+  if (n_prefixes == 0) {
+    // No prefixes in the query --> No validation needed.
+    return true;
```
oshadmi, line 515: Should we return true only if the spec prefix list is empty as well?
Otherwise, the query has no prefixes meaning it can read all keys, and spec may have some prefixes meaning it can only read a subset of the keys, and this should not be allowed?

JoanFM: so u mean, if the spec has no prefixes it should also return true?

oshadmi: The same check is done in `IsIndexCoherent` - so I suppose my comment was wrong.

PR #7006, opened 2025-10-08
Hybrid jk hybrid rlookup fixes

kei-nan:
## Describe the changes in the pull request

A clear and concise description of what the PR is solving, including:
1. Current: The current state briefly
2. Change: What is the change
3. Outcome: Adding the outcome

#### Which additional issues this PR fixes
1. MOD-...
2. #...

#### Main objects this PR modified
1. ...

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #7028, opened 2025-10-10
Fix hybrid timeout

nafraf:
This PR includes several improvements and fixes:

- Proper initialization of the time field in the search context for hybrid request flows.

- Explicit setup of command properties in `HybridRequest_prepareForExecution`.

- Updated `iterStartCb` private data to use a ref-counted data instead of a raw void*.

- Fixed several memory leaks.

---

> [!NOTE]
> Initialize hybrid query timeouts before pipeline build, set correct command flags, move iterator private data and cursor mappings to Weak/StrongRef handling, and create/free search contexts appropriately for cluster vs standalone; plus minor test tweaks.
> 
> - **Hybrid execution**:
>   - Initialize timeouts for all subqueries and tail context before building pipelines; set `MRCommand` flags (`forCursor`, `rootCommand = C_READ`).
>   - Store cursor mappings in `RPNet` as `StrongRef` clones and release originals; adjust cleanup paths.
>   - Create thread‑safe search contexts based on `NumShards` (no detached ctx in cluster); free contexts differently in `HybridRequest_Free`.
> - **RMR/RPNet infrastructure**:
>   - Change `MR_IterateWithPrivateData` to accept `StrongRef*`; `IteratorData` now holds a `WeakRef` and `iterCursorMappingCb` promotes to `StrongRef`, handling freed mappings and cleaning up correctly.
>   - In `rpnet`, build `_FT.CURSOR READ` with `rootCommand = C_READ` and pass mappings by reference to the iterator.
> - **Debug/Timeouts**:
>   - Apply tail timeout using `hreq->sctx` instead of a subquery context.
> - **Tests**:
>   - Add codespell ignores and improve assertions/messages in timeout and other tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 55ed74281f115dad877d74fc1c05776601bdf861. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/coord/hybrid/dist_hybrid.c
+++ b/src/coord/hybrid/dist_hybrid.c
@@ -266,7 +266,7 @@ static int HybridRequest_prepareForExecution(HybridRequest *hreq, RedisModuleCtx
     // Construct the command string
     MRCommand xcmd;
     HybridRequest_buildMRCommand(argv, argc, &us, &xcmd, sp, &hybridParams);
-    xcmd.protocol = 3;
+    xcmd.protocol = is_resp3(ctx) ? 3 : 2;
     xcmd.forCursor = hreq->reqflags & QEXEC_F_IS_CURSOR;
     xcmd.forProfiling = false;  // No profiling support for hybrid yet
     xcmd.rootCommand = C_READ;


src/hybrid/hybrid_request.c
```diff
@@ -316,8 +333,8 @@ AREQ **MakeDefaultHybridUpstreams(RedisSearchCtx *sctx) {
   initializeAREQ(search);
   initializeAREQ(vector);
   const char *indexName = HiddenString_GetUnsafe(sctx->spec->specName, NULL);
-  search->sctx = createDetachedSearchContext(sctx->redisCtx, indexName);
-  vector->sctx = createDetachedSearchContext(sctx->redisCtx, indexName);
+  search->sctx = createSearchContext(sctx->redisCtx, indexName);
```
Itzikvaknin: The name `createSearchContext` doesn’t clearly reflect its behavior (it creates a detached context when needed).
Let’s add `NumShards` as an argument and rename it to something like `createThreadSafeSearchContext`

src/coord/hybrid/dist_hybrid.c
```diff
@@ -261,19 +268,12 @@ static int HybridRequest_prepareForExecution(HybridRequest *hreq, RedisModuleCtx
     xcmd.protocol = is_resp3(ctx) ? 3 : 2;
     // xcmd.forCursor = hreq->reqflags & QEXEC_F_IS_CURSOR;
     // xcmd.forProfiling = false;  // No profiling support for hybrid yet
-    // xcmd.rootCommand = C_AGG;   // Response is equivalent to a `CURSOR READ` response
+    xcmd.rootCommand = C_AGG;   // Response is equivalent to a `CURSOR READ` response
```
Itzikvaknin: The comment is odd, since we have a specific callback to handle the `_FT.HYBRID` request format.
Lets make sure we now where and why it is needed.

Itzikvaknin: rpnetNext_StartWithMappings dispatches the command using netCursorCallback as its processor.
In aggregate, `netCursorCallback` works as follows:
- It first dispatches an FT.AGGREGATE command.
- Then, netCursorCallback invokes getCursorCommand.
- Inside getCursorCommand, the code takes the if (cmd->rootCommand == C_AGG) branch.
- If a valid cursor is returned (which is always the case in FT.HYBRID, right @kei-nan?), it sets the root command as follows: 
```     
newCmd = MR_NewCommand(4, "_FT.CURSOR", "READ", idx, buf);
newCmd.rootCommand = C_READ;
```
      
Therefore, I think we should assign it to `C_READ`

In coverage tests, we got the following crash:
```
             at /home/runner/work/RediSearch/RediSearch/redis/src/debug.c:2516:5
   5: _serverAssert
             at /home/runner/work/RediSearch/RediSearch/redis/src/debug.c:1162:25
   6: getCursorCommand
             at /home/runner/work/RediSearch/RediSearch/src/hybrid/dist_utils.c:154:5
   7: netCursorCallback
             at /home/runner/work/RediSearch/RediSearch/src/hybrid/dist_utils.c:106:8
   8: mrIteratorRedisCB
             at /home/runner/work/RediSearch/RediSearch/src/coord/rmr/rmr.c:557:5
   9: __redisRunCallback
```
which is 
```    
// The previous command was a _FT.CURSOR READ command, so we may not need to change anything.
RS_LOG_ASSERT(cmd->rootCommand == C_READ, "calling `getCursorCommand` after a DEL command");
```

I'll include the fix in https://github.com/RediSearch/RediSearch/pull/7030

src/hybrid/hybrid_request.c
nafraf: OK. Done.

src/coord/hybrid/dist_hybrid.c
kei-nan: the mapping should already arrive before `netCursorCallback`.
So I guess we can go directly to C_READ

src/coord/rpnet.c
```diff
@@ -176,10 +176,14 @@ int rpnetNext_StartWithMappings(ResultProcessor *rp, SearchResult *r) {
 
     // Create cursor read command using the copied index name
     nc->cmd = MR_NewCommand(3, "_FT.CURSOR", "READ", idx_copy);
+    nc->cmd.rootCommand = C_READ;
     nc->cmd.protocol = 3;
     rm_free(idx_copy);
 
-    nc->it = MR_IterateWithPrivateData(&nc->cmd, netCursorCallback, NULL, iterCursorMappingCb, vsimOrSearch);
+    StrongRef *mappingsRef = rm_malloc(sizeof(StrongRef));
```
kei-nan: not sure about this.
I don't think this pattern makes sense
The lifetime of the iterator is tied to our lifetime.
Is it possible we will get deleted and the iterator stays alive somehow?

nafraf: It was added to solve a memory leak detected during timeout tests.
This was the problematic flow:
1.  `HybridRequest_executePlan` created cursor mappings and stored them in RPNet via `StrongRef`
2.  `rpnetNext_StartWithMappings` called `MR_IterateWithPrivateData` with `iterCursorMappingCb`
3.  Critical bug: A raw pointer (`CursorMappings*`) was passed to the async callback instead of a protected `StrongRef`
4.  `HybridRequest_Free` was called immediately after `sendChunk_hybrid`, freeing the mappings
5.  When `iterCursorMappingCb` finally executed asynchronously, it accessed freed memory

Itzikvaknin: It looks like `MR_IterateWithPrivateData` isn’t blocking, so the execution order isn’t guaranteed.
That means the scenario described is valid, and reference counting at this level seems necessary.
WDYT @kei-nan ?

kei-nan: How I see it is that rpnet result processor should hold a strong ref, not the async callback.
When rpnet result processor is destroyed it should wait/canel the iterator and the async callback should not be executed.
Then it should release the strong ref he is holding.
That is what I think the correct behaviour should be.

src/coord/rmr/rmr.c
```diff
@@ -674,9 +674,17 @@ void iterStartCb(void *p) {
...
+  StrongRef mappingsRef = WeakRef_Promote(data->privateDataRef);
+  CursorMappings *vsimOrSearch = StrongRef_Get(mappingsRef);
+  if (!vsimOrSearch) {
+    // Cursor mappings have been freed - cannot proceed with command dispatch.
+    // Release the iterator to decrement its reference count and trigger cleanup.
+    // This handles the case where we abort before sending commands to any shards.
+    MRIterator_Release(it);
+    rm_free(data);
+    return;
+  }
```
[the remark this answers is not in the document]
Itzikvaknin:
```
static StrongRef _Ref_GetStrong(RefManager *rm) {
  StrongRef s_ref = {0};
  if (RefManager_TryGetStrongReference(rm)) {
    // a strong reference also holds a weak reference (reference to the RefManager),
    // so it won't be freed before the object it manages.
    RefManager_GetWeakReference(rm);
    s_ref.rm = rm;
  }
  return s_ref;
}
```


```
inline void *__RefManager_Get_Object(RefManager *rm) {
  return rm ? rm->obj : NULL;
}
```

If vsimOrSearch is NULL, no strong ref was taken → no release needed.

src/hybrid/hybrid_request.c
```diff
@@ -51,6 +51,7 @@ int HybridRequest_BuildDepletionPipeline(HybridRequest *req, const HybridPipelin
         // Create a depleter processor to extract results from this pipeline
         // The depleter will feed results to the hybrid merger
         RedisSearchCtx *nextThread = params->aggregationParams.common.sctx; // We will use the context provided in the params
+        SearchCtx_UpdateTime(nextThread, areq->reqConfig.queryTimeoutMS);
```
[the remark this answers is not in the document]
Itzikvaknin: I think `hreq->context` should be initialized next to the AREQs sctx initialization 
What meaning does it have to sample the time here?

src/coord/hybrid/dist_hybrid.c
```diff
@@ -258,22 +265,15 @@ static int HybridRequest_prepareForExecution(HybridRequest *hreq, RedisModuleCtx
     // Construct the command string
     MRCommand xcmd;
     HybridRequest_buildMRCommand(argv, argc, &us, &xcmd, sp, &hybridParams);
-    xcmd.protocol = is_resp3(ctx) ? 3 : 2;
-    // xcmd.forCursor = hreq->reqflags & QEXEC_F_IS_CURSOR;
-    // xcmd.forProfiling = false;  // No profiling support for hybrid yet
-    // xcmd.rootCommand = C_AGG;   // Response is equivalent to a `CURSOR READ` response
+    xcmd.protocol = 3;
```
Itzikvaknin: Lets bring back the macro

nafraf: We will replace it soon by `HYBRID_RESP_PROTOCOL_VERSION`

Itzikvaknin: Handled in https://github.com/RediSearch/RediSearch/pull/7006

PR #7032, opened 2025-10-12
Serialize and Restore Schema command - [MOD-11803]

GuyAv46:
### Serialization & Deserialization
Implement `IndexSpec_Serialize` and `IndexSpec_Deserialize` functions to generate a `RedisModuleString` from a schema and a schema (`IndexSpec`) from a serialization string.
The implementations rely on the type `rdb_load` and `rdb_save` implementations, so we now re-expose these methods to redis. This is okay, as we don't actually save the schemas as part of the key-space, so they won't be called by Redis if we don't call the manual implementations.

Testing is focused on the new logic, as the RDB serialization/deserialization is already well tested

### _FT._RESTOREIFNX Command
A new command for manual `RESTORE` logic for non-key module-type objects of the module. Currently, there is a single "SCHEMA" sub-command to restore indexes.
An additional `FT.DEBUG DUMP_SCHEMA` command was added to generate a schema dump for the restore command.

Testing is again focused on the behavior of the restore command. The new debug command and the ACL registration are also tested

1. Schema's type (`ft_index0`) RDB load and save implementations
2. Implement manual dump and restore functions for `IndexSpec`
3. Implement a new internal restore command
4. Implement a debug dump command

#### Mark if applicable

---

> [!NOTE]
> Introduce IndexSpec serialization/deserialization, an internal schema restore command, and a debug command to dump schemas, with supporting RDB refactors and tests.
> 
> - **Core / RDB**:
>   - Add `IndexSpec_Serialize` and `IndexSpec_Deserialize` to convert schemas to/from RDB-encoded `RedisModuleString`.
>   - Expose type rdb methods via `IndexSpec_RdbLoad_Logic` and wrapper for `IndexSpec_RdbSave`; split store logic into `IndexSpec_StoreAfterRdbLoad`.
>   - Refactor `IndexSpec_StartGC` to remove `RedisModuleCtx*`; switch GC logs to `RSDummyContext`.
> - **Commands**:
>   - New internal command `'_FT._RESTOREIFNX SCHEMA <encver> <blob>'` (macro `RS_RESTORE_IF_NX`) to restore schema; register subcommand.
>   - New debug command `FT.DEBUG DUMP_SCHEMA <index>` returning serialized schema and encoding version; register in debug help.
> - **Module wiring**:
>   - Register restore command and subcommand; adjust command flags/ACL categories.
> - **Mocks / Tests**:
>   - Redismock: add `SaveDataTypeToString`/`LoadDataTypeFromStringEncver` helpers.
>   - C++ tests for RDB roundtrip and string serialize/deserialize.
>   - Python tests for debug help, `DUMP_SCHEMA`, and end-to-end restore (`_FT._RESTOREIFNX SCHEMA`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 17b4fe3dd07d162de79d16c2b3b9c90f307804a5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -1108,7 +1108,7 @@ int RestoreSchema(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
 int RegisterRestoreIfNxCommands(RedisModuleCommand *restoreCmd) {
   int rc;
 
-  const char *schema_flags = IsEnterprise() ? "write "CMD_PROXY_FILTERED : "write internal";
+  const char *schema_flags = IsEnterprise() ? "write "CMD_PROXY_FILTERED : "write "CMD_INTERNAL;
   rc = RedisModule_CreateSubcommand(restoreCmd, "SCHEMA", RestoreSchema, schema_flags, 0, 0, 0);
   if (rc != REDISMODULE_OK) return rc;
 
@@ -1411,7 +1411,7 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx) {
          "write deny-oom", INDEX_ONLY_CMD_ARGS, "", !IsEnterprise()))
 
   RM_TRY(RMCreateSearchCommand(ctx, RS_RESTORE_IF_NX, NULL,
-         IsEnterprise() ? "write " CMD_PROXY_FILTERED : "write", 0, 0, 0, "", true))
+         "write", INDEX_ONLY_CMD_ARGS, "", true))
   RM_TRY_F(RegisterRestoreIfNxCommands, RedisModule_GetCommand(ctx, RS_RESTORE_IF_NX))
 
   // Special cases: Register drop commands which write to arbitrary keys
--- a/tests/pytests/test_index.py
+++ b/tests/pytests/test_index.py
@@ -147,7 +147,7 @@ def test_lazy_index_creation_info_modules(env):
 def test_restore_schema(env: Env):
 
     # Test that the command is not exposed to normal users
-    env.expect('_FT._RESTOREIFNX', 'SCHEMA').error().contains('unknown command')
+    env.expect('_FT._RESTOREIFNX', 'SCHEMA').error().contains('unknown subcommand')
     # Mark the client as internal for the rest of the test
     env.cmd('DEBUG', 'MARK-INTERNAL-CLIENT')
 
--- a/tests/cpptests/test_cpp_rdb.cpp
+++ b/tests/cpptests/test_cpp_rdb.cpp
@@ -253,4 +253,8 @@ TEST_F(RdbMockTest, testIndexSpecStringSerialize) {
     ASSERT_STREQ(HiddenString_GetUnsafe(spec->fields[0].fieldName, NULL), "title");
     ASSERT_STREQ(HiddenString_GetUnsafe(spec->fields[1].fieldName, NULL), "body");
     ASSERT_STREQ(HiddenString_GetUnsafe(spec->fields[2].fieldName, NULL), "price");
+
+    // Clean up
+    IndexSpec_RemoveFromGlobals(loaded_spec_ref, false);
+    RedisModule_FreeString(NULL, serialized);
 }
--- a/tests/pytests/test_acl.py
+++ b/tests/pytests/test_acl.py
@@ -26,7 +26,7 @@ def test_acl_search_commands(env):
         'FT._ALIASDELIFX', 'FT._CREATEIFNX', 'FT._ALIASADDIFNX', 'FT._ALTERIFNX',
         'FT._DROPINDEXIFX', 'FT.DROPINDEX', 'FT.TAGVALS', 'FT._DROPIFX',
         'FT.DROP', 'FT.GET', 'FT.SYNADD', 'FT.ADD', 'FT.MGET', 'FT.DEL',
-        '_FT.CONFIG', '_FT.DEBUG', 'FT.SAFEADD', '_FT._RESTOREIFNX',
+        '_FT.CONFIG', '_FT.DEBUG', 'FT.SAFEADD'
     ]
     if not env.isCluster():
         commands.append('FT.CONFIG')
--- a/tests/pytests/test_debug_commands.py
+++ b/tests/pytests/test_debug_commands.py
@@ -191,9 +191,9 @@ class TestDebugCommands(object):
         self.env.expect(debug_cmd(), 'dump_terms', 'idx1').error()
 
     def testDumpSchema(self):
-        self.env.expect(debug_cmd(), 'dump_schema').error()
-        self.env.expect(debug_cmd(), 'dump_schema', 'idx1').error()
-        self.env.expect(debug_cmd(), 'dump_schema', 'idx').noError()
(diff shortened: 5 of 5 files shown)


src/module.c
```diff
@@ -1083,6 +1083,41 @@ int IndexList(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+
+int RegisterRestoreIfNxCommands(RedisModuleCommand *restoreCmd) {
+
+  RedisModule_Log(RSDummyContext, "warning", "Registering internal command %s: %p", RS_RESTORE_IF_NX, restoreCmd);
+  int rc;
+
+  const char *schema_flags = IsEnterprise() ? "write "CMD_PROXY_FILTERED : "write internal";
+  rc = RedisModule_CreateSubcommand(restoreCmd, "SCHEMA", RestoreSchema, schema_flags, 0, 0, 0);
+  RedisModule_Log(RSDummyContext, "warning", "Registering internal command %s SCHEMA: %s", RS_RESTORE_IF_NX, rc == REDISMODULE_OK ? "OK" : rc == REDISMODULE_ERR ? "ERR" : "UNKNOWN");
+  if (rc != REDISMODULE_OK) return rc;
```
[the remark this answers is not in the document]
GuyAv46: Oops. Leftovers

src/module.c
```diff
@@ -1083,6 +1083,38 @@ int IndexList(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+    return RedisModule_ReplyWithError(ctx, "ERRBADVAL Failed to deserialize schema");
+  }
+
+  return RedisModule_ReplyWithSimpleString(ctx, "OK");
+}
+
+int RegisterRestoreIfNxCommands(RedisModuleCommand *restoreCmd) {
+  int rc;
+
+  const char *schema_flags = IsEnterprise() ? "write "CMD_PROXY_FILTERED : "write internal";
```
alonre24: Should we add the `allow-loading` flag here as well? 
@raz-mon, can you please also review the new command's registration logic?

src/spec.c
```diff
@@ -3341,11 +3384,15 @@ static void LoadingProgressCallback(RedisModuleCtx *ctx, RedisModuleEvent eid, u
   workersThreadPool_Drain(ctx, 100);
 }
 
+static void IndexSpec_RdbSave_Wrapper(RedisModuleIO *rdb, void *value) {
```
alonre24, line 3388: Why do we need this wrapper?

src/spec.c
```diff
@@ -3341,11 +3384,15 @@ static void LoadingProgressCallback(RedisModuleCtx *ctx, RedisModuleEvent eid, u
...
+  IndexSpec_RdbSave(rdb, value);
+}
+
 int IndexSpec_RegisterType(RedisModuleCtx *ctx) {
   RedisModuleTypeMethods tm = {
       .version = REDISMODULE_TYPE_METHOD_VERSION,
-      .rdb_load = IndexSpec_LegacyRdbLoad,
-      .rdb_save = IndexSpec_LegacyRdbSave,
+      .rdb_load = IndexSpec_RdbLoad_Logic,    // We don't store the index spec in the key space,
+      .rdb_save = IndexSpec_RdbSave_Wrapper,  // but these are useful for serialization/deserialization (and legacy loading)
```
alonre24, line 3396: Can you remind me of the motivation for this change?
Is the reason for not using simply using `Indexes_RdbLoad` upon handling the new ASM event of adding a new node (next PR) supporting legacy loading?

tests/pytests/test_index.py
```diff
@@ -142,3 +142,91 @@ def test_lazy_index_creation_info_modules(env):
...
+            'algorithm', 'FLAT',
+            'data_type', 'FLOAT32',
+            'dim', 2,
+            'distance_metric', 'L2',
+        ]
+    ]
+    env.assertEqual(index_info(env)['attributes'], expected)
+
+    # Test that synonyms were also restored correctly
+    env.expect('FT.SYNDUMP', 'idx').equal(['cat', ['meow'], 'dog', ['bark']])
```
alonre24, line 232: Nice
Consider adding also `rdb_compatbility` test for sanity

src/spec.c
GuyAv46: Just to silence a warning about implicit casting (the API expects a `RedisModuleIO` and a `void` pointers).
This is safer than casting, as if the inner function signature were to change, the cast will still work, while in this way, it won't compile

src/spec.c
GuyAv46: We can't serialize all the schemas together with `aux_save`, first and foremost, because there is no API to do so. It's also simpler to handle the schemas one by one, as some may exist already, while others may not.

This change gives back the existing load and save functionalities of single schema serialization and deserialization, allowing us to use `RedisModule_LoadDataTypeFromStringEncver` and `RedisModule_SaveDataTypeToString`, and maybe other APIs in the future.
We still don't actually create any key of the schema type, so these should not be called automatically by redis while saving to RDB (only by the `aux_save` callback)

tests/pytests/test_index.py
GuyAv46: There is already `redisearch_1.6.13_with_synonyms.rdb` case

src/module.c
GuyAv46: Ozan says it's not needed for the ASM feature. But I don't see any harm in adding it to this command (I will avoid adding it to `FT.DICTADD`, for example)

src/spec.c
```diff
@@ -3273,6 +3277,45 @@ void Indexes_RdbSave2(RedisModuleIO *rdb, int when) {
...
+  }
+}
+
+/**
+ * Convert an IndexSpec to its RDB serialized form, by calling the `IndexSpecType` rdb_save function.
+ * Note that the returned RedisModuleString* must be freed by the caller
+ * using RedisModule_FreeString
+*/
+RedisModuleString * IndexSpec_Serialize(IndexSpec *sp) {
+  return RedisModule_SaveDataTypeToString(NULL, sp, IndexSpecType);
```
raz-mon, line 3304: Is the dummy context better to give here?

raz-mon (commented):
LGTM overall
One question

raz-mon (approved):
LGTM
Regarding the ASM design (upcoming PRs to use this), I think it is not a great design (for our module at least), since replication of the index should occur upon shard addition rather than upon slot migration. It's quite wasteful to send all the aux data for every slot migration.
Nonetheless this functionality is good for us and nicely done! 💪

src/spec.c
```diff
@@ -3273,6 +3277,45 @@ void Indexes_RdbSave2(RedisModuleIO *rdb, int when) {
   }
 }
 
+void *IndexSpec_RdbLoad_Logic(RedisModuleIO *rdb, int encver) {
```
JoanFM, line 3280: what does this Logic stand for?

GuyAv46: I tried to separate between a legacy load that may occur while loading a key from a legacy RDB, and the new load logic that we only expect to get to if we manually triggered the load, and not from actual keys Redis handles.

PR #7040, opened 2025-10-13
Allow RSIndexResult in RQEIterator be mutated when wrapping iterators

GlenDC:

Some RQEIterator implementations wrap one or more other RQEIterator’s. As part of their read implementation they might however for some code paths mutate the result by for example
applying a weight to it or metrics.

This was prior to this commit not possible due to lifetime constrains and the fact that a shared reference is returned.

This PR makes these references mutable which is similar to how the C implementation was doing it
(reference instead of pointer, and returned instead of indirectly accessed, but other than that similar).

Originally we went for a Cow-inspired structure to make this less error-prone, but this had a high impact
on the benchmarks.

> (!) In future we'll switch this to owned return values completely,
> but for now that is not possible due to the inner FFI call. Once this is fully in Rust,
> we should be able to do this less error prone by returning `RSIndexResult` instead of `&mut RSIndexResult`
> but for now it is what it is.

This PR unblocks iterators which wrap other iterators and needs to mutate the inner results,
e.g. the optional iterator that applies a weight to these results.

1. MOD-11768

Make it possible to mutate iterator results where required,
while preserving the reuse of results when possible.

#### Mark if applicable

- [x] This PR introduces API changes (but just in the still unused Rust iterators AFAIK, so should be ok?)

---

> [!NOTE]
> Make RQEIterator return mutable RSIndexResult with explicit lifetimes and update Empty, IdList, Wildcard implementations and tests accordingly.
> 
> - **API**:
>   - `RQEIterator` now uses lifetimes `<'iterator, 'index>` and methods take `&'iterator mut self`.
>   - `read` and `skip_to` return mutable `RSIndexResult` (`&'iterator mut RSIndexResult<'index>`), and `SkipToOutcome::{Found,NotFound}` carry mutable refs.
> - **Implementations**:
>   - Update `empty`, `id_list`, and `wildcard` to conform to new trait signatures.
>   - `IdList` becomes `IdList<'index>`; internal `result` uses `RSIndexResult<'index>`; return values now mutable.
>   - `Wildcard` holds `RSIndexResult<'index>` and returns mutable results; minor internal adjustments.
> - **Tests**:
>   - Adjust tests for new signatures and add a test verifying post-read mutation support.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d5ff0ddf9e800cc4c0f4fca9f3278acddb583742. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/redisearch_rs/rqe_iterators/tests/c_mocks/mod.rs
```diff
@@ -40,6 +40,13 @@ pub extern "C" fn IndexResult_ConcatMetrics(
     // Do nothing since the code will call this
 }
 
+#[unsafe(no_mangle)]
+pub extern "C" fn RSYieldableMetrics_Clone(
+    _src: *const RSYieldableMetric,
+) -> *const RSYieldableMetric {
+    std::ptr::null()
```
gdesmott: Should not this be `unimplemented!()` if it's just to make the linker happy?

src/redisearch_rs/inverted_index/src/index_result.rs
```diff
@@ -699,6 +699,73 @@ pub struct RSIndexResult<'index> {
...
+pub enum CowRSIndexResult<'index> {
+    Borrowed(&'index RSIndexResult<'index>),
+    Owned(RSIndexResult<'static>),
+}
+
+impl Clone for CowRSIndexResult<'_> {
+    fn clone(&self) -> Self {
+        match self {
+            Self::Borrowed(arg0) => Self::Borrowed(arg0.to_owned()),
+            Self::Owned(arg0) => Self::Owned(arg0.to_owned()),
```
gdesmott:
```suggestion
            Self::Owned(result) => Self::Owned(result.to_owned()),
```

src/redisearch_rs/inverted_index/src/index_result.rs
```diff
@@ -699,6 +699,73 @@ pub struct RSIndexResult<'index> {
     pub weight: f64,
 }
 
+/// Copy-On-Write version of [`RSIndexResult`].
+///
+/// Used for places where you may wish to share a [`RSIndexResult`]
+/// most of the time until you do not. At which point you would make a copy to write it.
```
gdesmott: I'd document here how one can make this copy.

LukeMathWalker:
> Sounds like a good idea, but would like @LukeMathWalker 's input on the overall design.

At a glance, it looks like it'd be simpler (and more efficient) to have iterators return a `&mut RsIndexResult` to enable the usecase in https://github.com/RediSearch/RediSearch/pull/7046. What'd be the blocker to that kind of design?

GlenDC:
> > Sounds like a good idea, but would like @LukeMathWalker 's input on the overall design.
> 
> At a glance, it looks like it'd be simpler (and more efficient) to have iterators return a `&mut RsIndexResult` to enable the usecase in #7046. What'd be the blocker to that kind of design?

- simpler: not sure, seems more or less same
- efficient: sure, but at what cost

Having this kind of state be mutated all over the place seems to me like a recipe for disaster, but that's perhaps just my opinion. If that is what you want @LukeMathWalker , I'll replace this PR with doing that. Just let me know.

The use case is btw not only https://github.com/RediSearch/RediSearch/pull/7046 but also an open PR of Ben to allow metrics to be added. And I believe there are more of such use cases as we continue to port iterators which wrap other iterators.

src/redisearch_rs/inverted_index/src/index_result.rs
GlenDC: Document as in saying one would call `to_mut` at which point a copy would be made if still in a borrowed state, or what do you have in mind @gdesmott ?

src/redisearch_rs/rqe_iterators/tests/c_mocks/mod.rs
GlenDC: I can do, if that's preferred I will go for it

GlenDC:
> > > Sounds like a good idea, but would like @LukeMathWalker 's input on the overall design.
> > 
> > 
> > At a glance, it looks like it'd be simpler (and more efficient) to have iterators return a `&mut RsIndexResult` to enable the usecase in #7046. What'd be the blocker to that kind of design?
> 
> * simpler: not sure, seems more or less same
> * efficient: sure, but at what cost
> 
> Having this kind of state be mutated all over the place seems to me like a recipe for disaster, but that's perhaps just my opinion. If that is what you want @LukeMathWalker , I'll replace this PR with doing that. Just let me know.
> 
> The use case is btw not only #7046 but also an open PR of Ben to allow metrics to be added. And I believe there are more of such use cases as we continue to port iterators which wrap other iterators.

We had an offline discussion with @LukeMathWalker and @gdesmott .
Given the design how it is currently and given we want to (for now) stay as close
as possible to the C version I'll change this PR to drop the COW-like design
and replace it with a mutable reference.

In a future version we should be replace that `&mut` reference with an owned object,
as dropping should no longer have a measurable impact once the FFI call is removed.
Until than, we will live with the `&mut` reference and need to ensure that the base layers reset its results,
in-between calls.

src/redisearch_rs/inverted_index/src/index_result.rs
GlenDC: No more applicable, this will be dropped.

src/redisearch_rs/inverted_index/src/index_result.rs
GlenDC: No more applicable, this will be dropped.

src/redisearch_rs/rqe_iterators/tests/c_mocks/mod.rs
GlenDC: No more needed with the use of `&mut`. Dropping this mock.

src/redisearch_rs/inverted_index/src/index_result.rs
```diff
@@ -719,6 +719,18 @@ impl<'index> RSIndexResult<'index> {
         }
     }
 
+    pub fn reset_virt_metadata(&mut self) -> &mut Self {
```
gdesmott: Can you document this API?

src/redisearch_rs/inverted_index/src/index_result.rs
```diff
@@ -719,6 +719,18 @@ impl<'index> RSIndexResult<'index> {
...
+        debug_assert_eq!(self.data, RSResultData::Virtual);
+        debug_assert!(self.dmd.is_null());
+
+        self.weight = 0.;
+        self.metrics = std::ptr::null_mut();
```
gdesmott: Won't this leak `metrics`? See `impl Drop for RSIndexResult<'_> `.

src/redisearch_rs/rqe_iterators/src/id_list.rs
```diff
@@ -42,27 +42,31 @@ impl IdList {
...
+        &'iterator mut self,
+    ) -> Result<Option<&'iterator mut RSIndexResult<'index>>, RQEIteratorError> {
         let Some(doc_id) = self.get_current() else {
             return Ok(None);
         };
         self.offset += 1;
-        self.result.doc_id = doc_id;
-        Ok(Some(&self.result))
+
+        self.result.reset_virt_metadata().set_doc_id(doc_id);
```
gdesmott: Can you run the existing benchmarks to check the affect of those changes on performances?

src/redisearch_rs/inverted_index/src/index_result.rs
GlenDC: Good question, let me check it out!

src/redisearch_rs/rqe_iterators/src/id_list.rs
GlenDC: Can you tell me how and what to do to show you the report, I'll add it then as a PR comment

src/redisearch_rs/inverted_index/src/index_result.rs
GlenDC: yes, I think it's a ugly hack though, only thing I know for 99% certain is that we want to somehow reset if we choose to reuse, but still feels wonky, so not sure this is the right approach

src/redisearch_rs/rqe_iterators/src/id_list.rs
gdesmott: you can just run `cargo bench` in `src/redisearch_rs/inverted_index_bencher` before and after the change. It should tell you if there is significant changes. You can also them compare the  performance of the C vs Rust implementations.

src/redisearch_rs/inverted_index/src/index_result.rs
gdesmott: Yeah, I'm not big on it either.

And the current code could reset `freq` before actually setting it to its actual value right after. But maybe that's something the compiler will be smart enough to optimize?

src/redisearch_rs/rqe_iterators/src/id_list.rs
GlenDC: Running that takes very long, and seems like a very manual process to compare. Is that what you have been using so far? Seems very labour-intensive and error prone? Or are there essentially file outputs (before/after) that I can easily compare

gdesmott: > you can just run cargo bench in src/redisearch_rs/inverted_index_bencher

Sorry I meant `src/redisearch_rs/rqe_iterators_bencher`, those are the II benches.

>  seems like a very manual process to compare.

Yes, micro benchmarks are not really automated at this point. We rely on those to:
- compare the performance of the C and Rust versions of the same code, so we can spot big perf hit when rewriting.
- find regressions when doing significant changes, like now. This is done semi-automatically by `criterion` as it will tell you performance changes across runs. So you can just run it twice, without and with your changes, and check its output.

src/redisearch_rs/inverted_index/src/index_result.rs
```diff
@@ -719,12 +719,28 @@ impl<'index> RSIndexResult<'index> {
         }
     }
 
+    /// Resets the metadata attached to this [`RSIndexResult`].
+    ///
+    /// This [`RSIndexResult`] is assumed to be [`Virtual`] with no
+    /// [`RSDocumentMetadata`] embedded in it.
+    ///
+    /// [`Virtual`]: RSResultData::Virtual
```
gdesmott: What is this line for?

src/redisearch_rs/rqe_iterators/src/id_list.rs
LukeMathWalker: You can use `critcmp` for comparisons across branches.

LukeMathWalker: https://crates.io/crates/critcmp

LukeMathWalker:
```
git checkout master && cargo bench -- --save-baseline before
git checkout your-branch && cargo bench -- --save-baseline change
critcmp before change
```

GlenDC:
```
group                                             after                                   before
-----                                             -----                                   ------
Iterator - Empty - Read/C                         1.09      1.8±0.02ns        ? ?/sec     1.00      1.6±0.06ns        ? ?/sec
Iterator - Empty - Read/Rust                      1.07      0.3±0.00ns        ? ?/sec     1.00      0.2±0.02ns        ? ?/sec
Iterator - Empty - SkipTo/Rust                    1.10      0.3±0.00ns        ? ?/sec     1.00      0.2±0.00ns        ? ?/sec
Iterator - Empty - SkipTo/c                       1.07      1.8±0.03ns        ? ?/sec     1.00      1.7±0.05ns        ? ?/sec
Iterator - IdList - Read Dense/C                  1.12      2.1±0.48ms        ? ?/sec     1.00  1830.6±36.47µs        ? ?/sec
Iterator - IdList - Read Dense/Rust               2.12  857.8±125.08µs        ? ?/sec     1.00    405.3±6.23µs        ? ?/sec
Iterator - IdList - Read Sparse/C                 1.20      2.3±0.64ms        ? ?/sec     1.00  1874.2±49.11µs        ? ?/sec
Iterator - IdList - Read Sparse/Rust              2.63  1092.8±580.63µs        ? ?/sec    1.00   415.0±11.18µs        ? ?/sec
Iterator - IdList - SkipTo Dense/C                1.03   600.8±42.90µs        ? ?/sec     1.00  584.4±134.48µs        ? ?/sec
Iterator - IdList - SkipTo Dense/Rust             1.12   442.5±11.00µs        ? ?/sec     1.00   394.8±24.10µs        ? ?/sec
Iterator - IdList - SkipTo Sparse/C               1.21     21.9±5.78ms        ? ?/sec     1.00     18.0±0.72ms        ? ?/sec
Iterator - IdList - SkipTo Sparse/Rust            1.39     20.1±7.16ms        ? ?/sec     1.00     14.5±0.44ms        ? ?/sec
Iterator - Wildcard - Read Large Range/C          1.08   314.5±16.00µs        ? ?/sec     1.00    291.7±6.54µs        ? ?/sec
Iterator - Wildcard - Read Large Range/Rust       2.99  854.6±137.41µs        ? ?/sec     1.00    285.8±4.43µs        ? ?/sec
Iterator - Wildcard - Read Medium Range/C         1.06    156.0±4.44µs        ? ?/sec     1.00    147.8±3.64µs        ? ?/sec
Iterator - Wildcard - Read Medium Range/Rust      2.59    382.4±2.00µs        ? ?/sec     1.00    147.7±6.11µs        ? ?/sec
Iterator - Wildcard - SkipTo Large Range/C        1.12      9.2±1.91µs        ? ?/sec     1.00      8.2±0.35µs        ? ?/sec
Iterator - Wildcard - SkipTo Large Range/Rust     6.13     23.7±8.34µs        ? ?/sec     1.00      3.9±0.11µs        ? ?/sec
Iterator - Wildcard - SkipTo Medium Range/C       1.18      4.8±1.26µs        ? ?/sec     1.00      4.1±0.14µs        ? ?/sec
Iterator - Wildcard - SkipTo Medium Range/Rust    5.40     10.5±2.37µs        ? ?/sec     1.00  1937.5±63.41ns        ? ?/sec

```

Seems a significant difference. Shall I just drop the reset functionality for now?
C code is also not resetting anyway 🤷

src/redisearch_rs/inverted_index/src/index_result.rs
GlenDC: Rust doc comments are markdown.
This is defining the link for [`Virtual`] given that symbol is not in scope.

src/redisearch_rs/rqe_iterators/src/id_list.rs
gdesmott: Does forcing inline on `reset_virt_metadata()` help?

If not yeah, let's remove it then.

GlenDC: The difference is a bit less, but still not good.
So I'll remove it.

Here it is:

```
group                                             after                                  before
-----                                             -----                                  ------
Iterator - Empty - Read/C                         1.00      1.6±0.04ns        ? ?/sec    1.01      1.6±0.04ns        ? ?/sec
Iterator - Empty - Read/Rust                      1.00      0.2±0.01ns        ? ?/sec    1.01      0.2±0.01ns        ? ?/sec
Iterator - Empty - SkipTo/Rust                    1.00      0.2±0.01ns        ? ?/sec    1.01      0.2±0.01ns        ? ?/sec
Iterator - Empty - SkipTo/c                       1.00      1.6±0.05ns        ? ?/sec    1.02      1.6±0.05ns        ? ?/sec
Iterator - IdList - Read Dense/C                  1.01  1673.0±43.16µs        ? ?/sec    1.00  1656.7±31.06µs        ? ?/sec
Iterator - IdList - Read Dense/Rust               1.88    708.1±7.34µs        ? ?/sec    1.00   377.5±10.30µs        ? ?/sec
Iterator - IdList - Read Sparse/C                 1.00  1682.3±33.82µs        ? ?/sec    1.01  1694.1±39.78µs        ? ?/sec
Iterator - IdList - Read Sparse/Rust              2.01   768.8±11.40µs        ? ?/sec    1.00    382.7±6.55µs        ? ?/sec
Iterator - IdList - SkipTo Dense/C                1.00   494.9±60.29µs        ? ?/sec    1.07   531.6±54.32µs        ? ?/sec
Iterator - IdList - SkipTo Dense/Rust             1.03   388.2±14.84µs        ? ?/sec    1.00   375.9±18.69µs        ? ?/sec
Iterator - IdList - SkipTo Sparse/C               1.00     16.8±0.27ms        ? ?/sec    1.03     17.3±0.52ms        ? ?/sec
Iterator - IdList - SkipTo Sparse/Rust            1.11     14.7±0.10ms        ? ?/sec    1.00     13.3±0.14ms        ? ?/sec
Iterator - Wildcard - Read Large Range/C          1.00    270.7±5.74µs        ? ?/sec    1.00    271.2±4.62µs        ? ?/sec
Iterator - Wildcard - Read Large Range/Rust       2.55   693.9±15.42µs        ? ?/sec    1.00    272.3±6.75µs        ? ?/sec
Iterator - Wildcard - Read Medium Range/C         1.00    134.0±5.85µs        ? ?/sec    1.00    133.3±8.54µs        ? ?/sec
Iterator - Wildcard - Read Medium Range/Rust      2.49    347.7±5.19µs        ? ?/sec    1.00    139.6±5.38µs        ? ?/sec
Iterator - Wildcard - SkipTo Large Range/C        1.00      7.6±0.21µs        ? ?/sec    1.01      7.7±0.34µs        ? ?/sec
Iterator - Wildcard - SkipTo Large Range/Rust     4.55     16.9±0.90µs        ? ?/sec    1.00      3.7±0.12µs        ? ?/sec
Iterator - Wildcard - SkipTo Medium Range/C       1.00      3.8±0.25µs        ? ?/sec    1.01      3.9±0.14µs        ? ?/sec
Iterator - Wildcard - SkipTo Medium Range/Rust    4.71      8.7±0.24µs        ? ?/sec    1.00  1855.0±146.65ns        ? ?/sec
```

GlenDC:
All feedback is addressed AFAIK @gdesmott

PR #7042, opened 2025-10-13
allow to run clippy on bencher crates

gdesmott:
`clippy` is not run by the CI on the bencher crates as they require C code to be build first. 

Still, it may be worth it to run it manually, to find potential issues or just for the sake of being able to run manually `cargo clippy` directly from the workspace.

#### Mark if applicable

---

> [!NOTE]
> Refactors bencher crates to satisfy clippy by using unit-style encoders/decoders, removing unnecessary mutability, adjusting FFI decoder ctx usage, and adding targeted clippy allows.
> 
> - **Inverted Index Bencher**:
>   - Replace `::default()` with unit-style types (`DocIdsOnly`, `FieldsOnly`, `FieldsOffsets`, `FreqsOnly`, `FreqsFields`, `FreqsOffsets`, `OffsetsOnly`, `Full`, and `Wide` variants) in encode/decode paths.
>   - Remove unnecessary `mut` in closures and buffer params; pass buffers directly where possible.
>   - Add `Default` for certain `Bencher` structs (`freqs_only`, `numeric`).
>   - FFI: make decoder context (`IndexDecoderCtx`) immutable and pass by reference; add clippy allows (`ptr_offset_with_cast`, `useless_transmute`), and tweak calls accordingly.
>   - Crate root: add clippy allow attributes and minor extern fn attribute tweaks.
> - **RQE Iterators Bencher**:
>   - Use `Empty` (unit) instead of `Empty::default()`; minor `skip_to`/`read` usage cleanup.
>   - Add clippy allows in FFI module and crate root.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 28476c1fbf62849b2d5fb26f1133458e3c3597a4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rqe_iterators_bencher/src/ffi.rs
+++ b/src/redisearch_rs/rqe_iterators_bencher/src/ffi.rs
@@ -14,6 +14,8 @@ mod bindings {
     #![allow(unsafe_op_in_unsafe_fn)]
     #![allow(improper_ctypes)]
     #![allow(dead_code)]
+    #![allow(clippy::ptr_offset_with_cast)]
+    #![allow(clippy::useless_transmute)]
 
     use ffi::{NumericFilter, t_fieldMask};
     use inverted_index::t_docId;
--- a/src/redisearch_rs/rqe_iterators_bencher/src/lib.rs
+++ b/src/redisearch_rs/rqe_iterators_bencher/src/lib.rs
@@ -7,6 +7,12 @@
  * GNU Affero General Public License v3 (AGPLv3).
 */
 
+#![allow(
+    clippy::undocumented_unsafe_blocks,
+    clippy::missing_safety_doc,
+    clippy::multiple_unsafe_ops_per_block
+)]
+
 use std::ffi::c_void;
 
 pub mod benchers;
@@ -18,6 +24,7 @@ pub use types_ffi::NewVirtualResult;
 redis_mock::bind_redis_alloc_symbols_to_mock_impl!();
 
 #[unsafe(no_mangle)]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
 pub extern "C" fn ResultMetrics_Free(metrics: *mut ::ffi::RSYieldableMetric) {
     if metrics.is_null() {
         return;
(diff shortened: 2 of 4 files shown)


src/redisearch_rs/rqe_iterators_bencher/src/lib.rs
```diff
@@ -7,6 +7,12 @@
  * GNU Affero General Public License v3 (AGPLv3).
 */
 
+#![allow(
```
GlenDC, line 10: FWIW you can put these in cargo.toml these days, which I think is a better thing to do, especially as you can also share them over the entire workspace, including with overrides, all nicely from the cargo.toml

src/redisearch_rs/rqe_iterators_bencher/src/ffi.rs
```diff
@@ -14,6 +14,8 @@ mod bindings {
     #![allow(unsafe_op_in_unsafe_fn)]
     #![allow(improper_ctypes)]
     #![allow(dead_code)]
+    #![allow(clippy::ptr_offset_with_cast)]
```
GlenDC, line 17: FWIW you can put these in cargo.toml these days, which I think is a better thing to do, especially as you can also share them over the entire workspace, including with overrides, all nicely from the cargo.toml

src/redisearch_rs/rqe_iterators_bencher/src/lib.rs
```diff
@@ -18,6 +24,7 @@ pub use types_ffi::NewVirtualResult;
 redis_mock::bind_redis_alloc_symbols_to_mock_impl!();
 
 #[unsafe(no_mangle)]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
```
GlenDC, line 27: might be nice to document such overrides here with a one line comment on why that is fine

gdesmott: They are loads of those in the `bindgen` generated code, not something we can document ourselves.

src/redisearch_rs/rqe_iterators_bencher/src/lib.rs
gdesmott: We already have

```toml
[lints]
workspace = true
```

I did not find a way to keep inheriting those and overrides/disable only a few specific ones. Is there a way to do that?

chesedo (approved):

PR #7059, opened 2025-10-15
MOD-11754: Enforce rlookup keys are prefixed correctly in FT.HYBRID

kei-nan:
In FT.HYBRID we would like to force @ before rlookup keys
Up until now we silently ate the @ if it existed and if not then carried on.

1. Current: Allow fields without @ to be passed in load step and reduce step
2. Change: Require @ before fields or $ in the case of load
3. Outcome: More strict behaviour for FT.HYBRID

1. Pipeline

#### Mark if applicable

---

> [!NOTE]
> Enforces strict field prefixing (@ for fields, $ for JSON) across FT.HYBRID parsing, LOAD, GROUPBY/REDUCE, and distribution; adds validation helper, updates plans/reducers, and adjusts tests.
> 
> - **Core validation**
>   - Introduce `ExtractKeyName` in `util/misc.{c,h}` to parse and validate `@`/`$` prefixes; emit errors on missing/invalid prefixes.
> - **Planning/Parsing**
>   - Add `strictPrefix` to `PLN_LoadStep` and `PLN_GroupStep`; copy in clone and pass through distribution.
>   - Change `PLNGroupStep_New` signature to accept `strictPrefix`; set `true` in hybrid parsing, `false` in aggregate parsing.
>   - Hybrid callbacks: require prefixes in `SORTBY`, `LOAD`, `GROUPBY`; implicit key fields now include `@` in `HYBRID_IMPLICIT_KEY_FIELDS`.
> - **Reducers**
>   - Extend `ReducerOptions` with `strictPrefix`; `ReducerOpts_GetKey` now uses `ExtractKeyName` and validates prefixes.
> - **Pipeline/Execution**
>   - `processLoadStep` and related code use `ExtractKeyName` to normalize/validate paths; include misc util.
>   - Distributed pipeline serialization: `LOAD` args now serialize with `@` for names, preserve `$` JSON paths; propagate `strictPrefix` in group steps.
> - **Tests**
>   - Update existing tests to use `@__key`, `@__score`, etc.
>   - Add `test_hybrid_field_validation.py` covering required prefixes in `LOAD`, `FILTER`, `APPLY`, `SORTBY`, `GROUPBY/REDUCE` and JSON `$` paths.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1415cdbe9f6f162a7b1cc8c4c5142079a012e309. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_hybrid_field_validation.py
+++ b/tests/pytests/test_hybrid_field_validation.py
@@ -159,7 +159,7 @@ def test_hybrid_sortby_requires_at_prefix(env):
         'PARAMS', '2', 'query_vec', query_vector,
         'LOAD', '1', '@price',
         'SORTBY', '2', 'price', 'DESC'  # Missing @ prefix
-    ).error().contains('"Missing prefix: name requires \'@\' prefix, JSON path require \'$\' prefix, got: price in SORTBY"')
+    ).error().contains('Missing prefix: name requires \'@\' prefix, JSON path require \'$\' prefix, got: price in SORTBY')
 
     # Test SORTBY with @ prefix - should succeed
     env.expect(


src/hybrid/hybrid_search_result.c
```diff
@@ -100,26 +100,21 @@ double calculateHybridScore(HybridSearchResult *hybridResult, HybridScoringConte
...
   for (size_t i = 0; i < hybridResult->numSources; i++) {
     if (hybridResult->hasResults[i]) {
       SearchResult *sourceResult = hybridResult->searchResults[i];
       RS_ASSERT(sourceResult);
 
-      // Write fields from source row to destination row
-      RLookupRow_WriteFieldsFrom(SearchResult_GetRowData(sourceResult), lookupCtx->sourceLookups[i],
-                                destination, lookupCtx->tailLookup);
+      // mocw fields from source row to destination row
+      RLookupRow_Move(lookupCtx->tailLookup, SearchResult_GetRowDataMut(sourceResult), SearchResult_GetRowDataMut(targetResult));
```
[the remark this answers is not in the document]
kei-nan: code was updated

tests/pytests/test_hybrid_field_validation.py
```diff
@@ -0,0 +1,335 @@
...
+    """Setup JSON index with test data for JSON path validation tests"""
+    conn = env.getClusterConnectionIfNeeded()
+    env.expect('FT.CREATE json_idx ON JSON SCHEMA $.description AS description TEXT $.embedding AS embedding VECTOR FLAT 6 TYPE FLOAT32 DIM 2 DISTANCE_METRIC L2 $.category AS category TAG $.price AS price NUMERIC').noError()
+    
+
+LOAD_ERROR_MSG = 'Missing field symbol prefix for field name in LOAD'
+
+
+# TODO: remove skip once FT.HYBRID for cluster is implemented
+@skip(cluster=True)
```
raz-mon: Can remove?

src/aggregate/aggregate_plan.h
```diff
@@ -111,6 +111,7 @@ typedef struct {
   ArgsCursor args;
   const RLookupKey **keys;
   size_t nkeys;
+  bool strict; // Whether we should fail if a field is not prefixed with an @ or $ sign
```
raz-mon:
```suggestion
  bool strict_prefix; // Whether we should fail if a field is not prefixed with an @ or $ sign
```

src/pipeline/pipeline_construction.c
```diff
@@ -268,6 +268,9 @@ static int processLoadStepArgs(PLN_LoadStep *loadStep, RLookup *lookup, uint32_t
     if (*path == '@') {
       path++;
       name_len--;
+    } else if (loadStep->strict && *path != '$') {
+        QueryError_SetWithUserDataFmt(status, QUERY_EPARSEARGS, "Missing field symbol prefix for field name in LOAD", ", perhaps you meant to use @%s?", path);
```
raz-mon: Consider having the generic error string defined in `query_error.h` instead of the duplications.

raz-mon: Regarding the enforcement of the `$` - let's make sure if it's needed always - CC @oshadmi

tests/pytests/test_hybrid_field_validation.py
```diff
@@ -0,0 +1,335 @@
...
+    env.expect(
+        'FT.HYBRID', 'idx',
+        'SEARCH', 'shoes',
+        'VSIM', '@embedding', '$query_vec',
+        'PARAMS', '2', 'query_vec', query_vector,
+        'LOAD', '2', '@description', '@category'  # Both with @ prefix
+    ).noError()
+
+# TODO: remove skip once FT.HYBRID for cluster is implemented
+@skip(cluster=True)
```
raz-mon: Can remove?

src/pipeline/pipeline_construction.c
kei-nan: regarding the error message, if I move it to `query_error.h` then don't think I could fill out the user data about the problematic field.

kei-nan: @oshadmi can you look at the code from JSON perspective?
Is checking the $ good enough?
Can the user give any other json paths that are valid?

src/aggregate/reducer.c
```diff
@@ -70,6 +70,9 @@ int ReducerOpts_GetKey(const ReducerOptions *options, const RLookupKey **out) {
   // Get the input key..
   if (*s == '@') {
     s++;
+  } else if (options->strict) {
```
kei-nan: @oshadmi should we verify we don't get a json path here?
Seems unlikely but not sure how strict we want to be here.

src/aggregate/reducer.h
```diff
@@ -124,13 +124,16 @@ typedef struct {
    * NOT NULL, then the error information will be set here.
    */
   QueryError *status;
+
+  // Whether to enforce strict parsing of arguments
+  bool strict;
```
raz-mon:
```suggestion
  bool strict_prefix;
```

src/aggregate/reducer.c
oshadmi: I think JSON path needs to start with a `$` or a `.`

src/pipeline/pipeline_construction.c
oshadmi: I think JSON path needs to start with a `$` or a `.`
See API `japi->pathParse` ?

src/aggregate/reducer.c
kei-nan: @raz-mon regarding your question I see we require the field to exist in either the document or pipeline, otherwise we output:
QueryError_SetWithUserDataFmt(options->status, QUERY_ENOPROPKEY, "Property is not present in document or pipeline", ": `%s`", s);
Will add support for $ and .

kei-nan: looks like . is for legacy json get set but not for load.
Don't think we need to support '.'.

src/aggregate/reducer.h
kei-nan: went with strictPrefix

src/aggregate/aggregate_plan.h
kei-nan: went with strictPrefix

src/pipeline/pipeline_construction.c
kei-nan: looks like . is only for JSON GET/SET and is legacy.
$ is the standardized prefix from what I could gather.

oshadmi: But actually also JSON fields have an alias in addition to their JSONPath so the alias should be referenced with a `@` as well, no? Do we allow to refer to the JSONPath - I don't think so (can also be ambiguous)
```
FT.CREATE idx ON JSON SCHEMA $.name as name TEXT $.age as age NUMERIC
JSON.SET doc:1 $ '{"name":"Homer", "age":42}'
JSON.SET doc:2 $ '{"name":"Bart", "age":13}'
```

This is OK:
```
FT.AGGREGATE idx * LOAD 1 @name APPLY "upper(@name)" as fake
```

This is not OK:
```
FT.AGGREGATE idx * LOAD 1 $.name APPLY "upper($.name)" as fake
(error) Unknown symbol 'name'
```

src/aggregate/reducer.c
raz-mon: Validated @kei-nan?

src/rlookup.c
```diff
@@ -982,6 +982,20 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
...
 }
 
+const char *RLookup_ExtractKeyName(const char *s, size_t *len, QueryError *status, bool strictPrefix, const char *context) {
+  if (*s == '@') {
+    --*len;
+    return s + 1;
+  } else if (*s == '$') {
+    return s;
+  } else if (strictPrefix) {
+    QueryError_SetWithUserDataFmt(status, QUERY_EPARSEARGS, "Missing field symbol prefix for field name", ": %s in %s", s, context);
```
raz-mon: This isn't a great error string for the case of `$`

src/coord/hybrid/dist_hybrid_plan.cpp
```diff
@@ -62,11 +62,17 @@ int HybridRequest_BuildDistributedDepletionPipeline(HybridRequest *req, const Hy
...
     array_append(*target, rm_strndup("LOAD", 4));
-    char *ldsze;
-    rm_asprintf(&ldsze, "%lu", (unsigned long)keys.size());
-    array_append(*target, ldsze);
+    char *buffer;
+    rm_asprintf(&buffer, "%lu", (unsigned long)keys.size());
+    array_append(*target, buffer);
     for (auto kk : keys) {
-      array_append(*target, rm_strndup(kk->name, kk->name_len));
+      // json paths should probably be serialized as is
```
raz-mon: Probably?

tests/pytests/test_hybrid_field_validation.py
```diff
@@ -0,0 +1,325 @@
...
+        'FT.HYBRID', 'idx',
+        'SEARCH', 'shoes',
+        'VSIM', '@embedding', '$query_vec',
+        'PARAMS', '2', 'query_vec', query_vector,
+        'LOAD', '*'  # Load all fields
+    ).noError()
+
+
+def test_hybrid_special_fields_work(env):
+    """Test that FT.HYBRID works with special fields like __key and __score"""
```
raz-mon: Wrong according to the test

src/rlookup.c
```diff
@@ -982,6 +982,20 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
   return rv;
 }
 
+const char *RLookup_ExtractKeyName(const char *s, size_t *len, QueryError *status, bool strictPrefix, const char *context) {
```
raz-mon: Why is this under RLookup?

kei-nan: couldn't find a better place for it TBH
We eventually use the string to open an RLookup key so there is some validity in having the logic there.

tests/pytests/test_hybrid_field_validation.py
kei-nan: Changed to:
Test that FT.HYBRID field prefix validation works with special fields like __key and __score

src/coord/hybrid/dist_hybrid_plan.cpp
kei-nan: it makes it more human friendly when we see it in monitor command
but it would still work if someone did @$.foo since we would remove the @ silently.
removed the word probably

src/rlookup.c
kei-nan: We don't know which field the user meant, it could be an alias, a json field
we have no way of knowing in this current context.
Do you have a suggestion for a better error message?

raz-mon: Yeah..
You could always be super explicit: `"Hash fields are required to be prefixed with `@` and JSON documents with `$`"`
But your call I can live with this as well, hope it won't confuse users

src/rlookup.c
raz-mon: Let's find a better place and name for this function

kei-nan: moved to misc.h under util

src/rlookup.c
kei-nan: will ask chatbot for suggestions

src/util/misc.c
```diff
@@ -21,3 +21,17 @@ int GetRedisErrorCodeLength(const char* error) {
...
 }
+
+const char *ExtractKeyName(const char *s, size_t *len, QueryError *status, bool strictPrefix, const char *context) {
+  if (*s == '@') {
+    --*len;
+    return s + 1;
+  } else if (*s == '$') {
+    return s;
+  } else if (strictPrefix) {
+    QueryError_SetWithUserDataFmt(status, QUERY_EPARSEARGS, "Field name requires prefix: use '@' for regular fields or '$' for JSON paths", ", field: %s in %s", s, context);
```
oshadmi: Ideally we would know if index is on JSON or not and allow `$` for JSON.

oshadmi: Slightly improved phrasing could be using `field alias` or `field name` and `field path`, e.g., 
"name requires prefix: use '@' for aliases or '$' for JSON paths"

oshadmi (approved):

src/aggregate/aggregate_plan.h
```diff
@@ -135,14 +136,17 @@ typedef struct {
...
   } * reducers;
   int idx;
+  // Whether we should fail if a key is not prefixed with an @ sign
+  bool strictPrefix;
 } PLN_GroupStep;
 
  /**
   * Allocates and initializes a new group step.
   * @param properties_ref StrongRef referencing the properties array (must be cloned by caller)
+  * @param strict Whether we should fail if a key is not prefixed with an @ sign
```
nafraf:
```suggestion
  * @param strictPrefix Whether we should fail if a key is not prefixed with an @ sign
```

PR #7065, opened 2025-10-15
[MOD-10037] add default scorer in configuration

JoanFM:

Add DEFAULT_SCORER to configuration to be applied for every type of query

---

> [!NOTE]
> Introduce a validated DEFAULT_SCORER config (via FT.CONFIG and Redis CONFIG) and apply it across SEARCH/AGGREGATE/HYBRID when no scorer is specified.
> 
> - **Configuration**:
>   - Add `DEFAULT_SCORER` (FT.CONFIG) and `search-default-scorer` (Redis CONFIG) with validation against registered scorers and exposure in `INFO` and config dumps.
>   - Store in `RSGlobalConfig.defaultScorer`; getters/setters, load-time and runtime support; deprecation messages preserved.
> - **Query Execution**:
>   - Use global default scorer when `SCORER` not provided in `FT.SEARCH`/`FT.AGGREGATE`/`FT.HYBRID` (`AREQ_ApplyContext`, pipeline scorer, LLAPI iterator).
>   - Optimizer treats unspecified or `BM25STD` equivalently for scorer type.
> - **Extensions/Core**:
>   - Add `Extensions_InitDone()`; validate configured default scorer after loading extensions; log/abort on invalid; free on unload.
> - **Info/Diagnostics**:
>   - Include `default_scorer` in runtime info and formatted config string.
> - **Tests/Mocks**:
>   - New pytests for config and behavior (`test_default_scorer.py`); updates to existing tests and redismock to init/cleanup `defaultScorer`; minor hybrid tests adjusted to expect default scorer name.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9110b7bb3e804749a919d1933a7c09d039184b98. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/module-init/module-init.c
+++ b/src/module-init/module-init.c
@@ -198,7 +198,7 @@ int RediSearch_Init(RedisModuleCtx *ctx, int mode) {
   RS_ASSERT(RSGlobalConfig.defaultScorer);
   ExtScoringFunctionCtx *scoreCtx = Extensions_GetScoringFunction(NULL, RSGlobalConfig.defaultScorer);
   if (scoreCtx == NULL) {
-    DO_LOG("warning", "default scorer '%s' is not a valid scorer", RSGlobalConfig.defaultScorer);
+    DO_LOG("warning", "The scorer '%s' specified in the configuration for the default scorer is not a valid scorer", RSGlobalConfig.defaultScorer);
     return REDISMODULE_ERR;
   }
 


src/config.c
```diff
@@ -643,6 +676,38 @@ RedisModuleString * get_friso_ini(const char *name, void *privdata) {
...
+  const char *scorerName;
+  int acrc = AC_GetString(ac, &scorerName, NULL, 0);
+  if (acrc == AC_OK) {
+    // Validate scorer name against registered scorers
+    ExtScoringFunctionCtx *scoreCtx = Extensions_GetScoringFunction(NULL, scorerName);
+    if (scoreCtx == NULL) {
+      QueryError_SetError(status, QUERY_EBADVAL, "Invalid default scorer value");
+      return REDISMODULE_ERR;
+    }
+    config->defaultScorer = rm_strdup(scorerName);
```
kei-nan, line 706: what frees defaultScorer?
if this setter is called twice will this lead to a memory leak?
Edit: Saw that cursor also saw this potential issue.

JoanFM: handled as per comment from cursor

src/config.c
```diff
@@ -193,6 +195,37 @@ int set_immutable_string_config(const char *name, RedisModuleString *val, void *
...
 
+int set_default_scorer_config(const char *name, RedisModuleString *val, void *privdata, RedisModuleString **err) {
+    REDISMODULE_NOT_USED(name);
+
+    // Get the scorer name from the Redis module string
+    size_t len;
+    const char *newScorerName = RedisModule_StringPtrLen(val, &len);
+
+    // Allocate temporary storage for the new scorer name
+    char *tempScorer = rm_strndup(newScorerName, len);
```
kei-nan: can't we do the allocation only if `Extensions_GetScoringFunction` was successful?
might clean up the code a bit

src/config.c
```diff
@@ -1594,6 +1674,7 @@ void iteratorsConfig_init(IteratorsConfig *config) {
 
 
 int RegisterModuleConfig(RedisModuleCtx *ctx) {
+  RSGlobalConfig.defaultScorer = rm_strdup(DEFAULT_SCORER_NAME);
```
kei-nan: this is not the correct place to do this.
Look at how we initialize our config.

JoanFM: how can we do it? in the config.h it does not allow because it is a static thing.

src/config.c
JoanFM: Good, thanks

src/config.c
JoanFM: I do not like the idea of having it be NULL and have the setters and getters need to handle that. This is the way to make sure this is the default from the start of the module

src/config.c
```diff
@@ -643,6 +674,45 @@ RedisModuleString * get_friso_ini(const char *name, void *privdata) {
...
+      return REDISMODULE_ERR;
+    }
+    // Free the old scorer name before assigning the new one
+    if (config->defaultScorer) {
+      rm_free((void *)config->defaultScorer);
+    }
+    config->defaultScorer = rm_strdup(scorerName);
+  }
+  RETURN_STATUS(acrc);
+}
```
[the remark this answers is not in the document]
JoanFM: this is due to the fact that setDefaultScorer is guaranteed to be called with InitDone, I will do an assert there

tests/pytests/test_config.py
```diff
@@ -1871,3 +1873,35 @@ def test_on_oom(env):
...
     env.expect(config_cmd(), 'SET', 'ON_OOM', 'invalid').error().contains('Invalid ON_OOM value')
+
+@skip(cluster=True)
+def testDefaultScorerConfig(env):
+    """Test DEFAULT_SCORER configuration via FT.CONFIG and CONFIG commands"""
+    env.expect(config_cmd(), 'GET', 'DEFAULT_SCORER').equal([['DEFAULT_SCORER', 'BM25STD']])
+    env.expect('CONFIG', 'GET', 'search-default-scorer').equal(['search-default-scorer', 'BM25STD'])
+
+    env.expect(config_cmd(), 'SET', 'DEFAULT_SCORER', 'TFIDF').equal('OK')
+    env.expect(config_cmd(), 'GET', 'DEFAULT_SCORER').equal([['DEFAULT_SCORER', 'TFIDF']])
```
nafraf: This seems to be redundant, we have 'TFIDF' in the list `valid_scores`

tests/pytests/test_config.py
```diff
@@ -1871,3 +1873,35 @@ def test_on_oom(env):
...
+    env.expect('CONFIG', 'GET', 'search-default-scorer').equal(['search-default-scorer', 'TFIDF'])
+
+    env.expect('CONFIG', 'SET', 'search-default-scorer', 'BM25').equal('OK')
```
nafraf: can't we add 'BM25' to `valid_scores` liste?

tests/pytests/test_config.py
JoanFM: I do not understand, what do u mean?

tests/cpptests/test_cpp_forkgc.cpp
```diff
@@ -90,6 +90,8 @@ class FGCTest : public ::testing::Test {
 
   void SetUp() override {
     Initialize_KeyspaceNotifications();
+    // Initialize default scorer for search operations
+    RSGlobalConfig.defaultScorer = DEFAULT_SCORER_NAME;
```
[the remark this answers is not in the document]
JoanFM: but there is no module cleanup in this one

src/config.h
```diff
@@ -318,6 +322,7 @@ char *getRedisConfigValue(RedisModuleCtx *ctx, const char* confName);
 #define RS_DEFAULT_CONFIG {                                                    \
     .extLoad = NULL,                                                           \
     .frisoIni = NULL,                                                          \
+    .defaultScorer = NULL,                                                     \
```
kei-nan, line 325: I think we need to provide here the default scorer name here, no?

src/config.c
```diff
@@ -193,6 +195,38 @@ int set_immutable_string_config(const char *name, RedisModuleString *val, void *
   return REDISMODULE_OK;
 }
 
+int set_default_scorer_config(const char *name, RedisModuleString *val, void *privdata, RedisModuleString **err) {
+    REDISMODULE_NOT_USED(name);
+    if (RSGlobalConfig.defaultScorer == NULL) {
```
kei-nan, line 200: I don't think we should ever allow the default score to be null - it is an invariant we must maintain since all queries rely on it.

JoanFM: this is what I am doing here

src/config.h
JoanFM: I would need to turn defaultScorer into a char[]. Doing that started a series of compilation issues, related to static vars and so on

src/config.c
JoanFM: when the Config is initialized, this is called for first time and it sets to not null

src/config.h
JoanFM: If I had to add a value, it would be directly the DEFAULT_SCORER_NAME, but then it would fail in memory because it would be static when I try to free when I change it.

And I cannot do rm_strdup(DEFAULT_SCORER_NAME) because it fails to compilew with:
 error: initializer element is not a compile-time constant

src/config.c
kei-nan: I think it will only get called if the config for it exists in the config file.
Maybe it would be better to initialize it in the default config macro?
Meaning: `RS_DEFAULT_CONFIG`

src/config.c
```diff
@@ -643,6 +677,48 @@ RedisModuleString * get_friso_ini(const char *name, void *privdata) {
   return config_friso_ini;
 }
 
+RedisModuleString *get_default_scorer_config(const char *name, void *privdata) {
+  char *str = *(char **)privdata;
+  if (str == NULL) {
```
kei-nan: if we ensure it is never null then we could remove this if.

src/config.c
JoanFM: I tried and found the problems described in that answer.

I added a test that shows that the scorer is set even if it is not in the config file from start.

src/config.h
kei-nan: how about:

```suggestion
    .defaultScorer = rm_strndup(DEFAULT_SCORER, strlen(DEFAULT_SCORER),                               \
```

src/config.c
JoanFM: I sent the limitations. I believe this is unreachable code, but still prefer to double check

src/config.h
kei-nan: Maybe this can be an enum instead of a string?
Will remove the need for all the memory management stuff

JoanFM: this would be ideal for me. But we would need to refactor all of this and have breaking changes w.r.t extensions

src/config.c
```diff
@@ -193,6 +195,39 @@ int set_immutable_string_config(const char *name, RedisModuleString *val, void *
...
+          }
+          return REDISMODULE_ERR;
+      }
+    }
+
+    // Validation passed, now allocate and apply it to RSGlobalConfig
+    char *tempScorer = rm_strndup(newScorerName, len);
+    char **ptr = (char **)privdata;
+    if (*ptr) {
+        rm_free(*ptr);
```
Itzikvaknin: Can you please document what do we free here?

src/config.c
Itzikvaknin: Consider an assertion instead

src/extension.c
```diff
@@ -31,6 +31,10 @@ void Extensions_Init() {
   }
 }
 
+bool Extensions_InitDone() {
+  return scorers_g != NULL;
```
Itzikvaknin, line 37: Is that thread safe?

src/module-init/module-init.c
```diff
@@ -195,6 +195,12 @@ int RediSearch_Init(RedisModuleCtx *ctx, int mode) {
     return REDISMODULE_ERR;
   }
 
+  ExtScoringFunctionCtx *scoreCtx = Extensions_GetScoringFunction(NULL, RSGlobalConfig.defaultScorer);
+  if (scoreCtx == NULL) {
+    DO_LOG("warning", "default scorer %s not a valid scorer", RSGlobalConfig.defaultScorer);
+    return REDISMODULE_ERR;
+  }
```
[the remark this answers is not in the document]
Itzikvaknin: @JoanFM can you reply to it?

tests/pytests/test_default_scorer.py
```diff
@@ -0,0 +1,237 @@
...
+    conn = getConnectionByEnv(env)
+    conn.execute_command('HSET', 'doc1', 'title', 'hello world', 'content', 'this is a test document', 'vector', b'\x00\x00\x80\x3f\x00\x00\x00\x40\x00\x00\x40\x40\x00\x00\x80\x40')
+    conn.execute_command('HSET', 'doc2', 'title', 'hello universe', 'content', 'another test document', 'vector', b'\x00\x00\x00\x40\x00\x00\x40\x40\x00\x00\x80\x40\x00\x00\xa0\x40')
+    conn.execute_command('HSET', 'doc3', 'title', 'world peace', 'content', 'final test document', 'vector', b'\x00\x00\x40\x40\x00\x00\x80\x40\x00\x00\xa0\x40\x00\x00\xc0\x40')
+
+    # Wait for indexing
+    waitForIndex(env, 'idx')
+    search_default_bm25std_without_config_set = env.cmd('FT.SEARCH', 'idx', 'hello', 'WITHSCORES', 'NOCONTENT') # Default scorer is BM25STD from start
+
+    default_scorer = env.cmd('FT.CONFIG', 'GET', 'DEFAULT_SCORER')
```
Itzikvaknin:
```suggestion
    default_scorer = env.cmd('CONFIG', 'GET', 'search-default-scorer')
```

src/module-init/module-init.c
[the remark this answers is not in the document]
JoanFM: this is proven in tests. This is to validate that the setter works proper and that the value in config file is valid

src/pipeline/pipeline_construction.c
```diff
@@ -225,8 +225,9 @@ static ResultProcessor *getArrangeRP(Pipeline *pipeline, const AggregationPipeli
 static ResultProcessor *getScorerRP(Pipeline *pipeline, RLookup *rl, const RLookupKey *scoreKey, const QueryPipelineParams *params) {
   const char *scorer = params->scorerName;
   if (!scorer) {
-    scorer = DEFAULT_SCORER_NAME;
+    scorer = RSGlobalConfig.defaultScorer;
   }
+  RS_LOG_ASSERT(scorer, "No scorer name");
```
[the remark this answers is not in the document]
JoanFM: this comes from the nuances of our initialization, but it is proven to work with the tests.

src/config.c
JoanFM: showed the problematic.

src/extension.c
JoanFM: initialization is done at the startup in the main thread. Will add a note that this is for assertion only so I would not put so much attention either

tests/cpptests/test_cpp_forkgc.cpp
```diff
@@ -90,6 +90,7 @@ class FGCTest : public ::testing::Test {
 
   void SetUp() override {
     Initialize_KeyspaceNotifications();
+    // Initialize default scorer for search operations
```
JoanFM:
```suggestion
```

src/module-init/module-init.c
```diff
@@ -195,6 +195,15 @@ int RediSearch_Init(RedisModuleCtx *ctx, int mode) {
...
   }
 
+  ExtScoringFunctionCtx *scoreCtx = Extensions_GetScoringFunction(NULL, RSGlobalConfig.defaultScorer);
+  if (scoreCtx == NULL) {
+    DO_LOG("warning", "default scorer '%s' is not a valid scorer", RSGlobalConfig.defaultScorer);
+    // Free the invalid scorer name, in tests OnUnload is not called
+    rm_free((void *)RSGlobalConfig.defaultScorer);
+    RSGlobalConfig.defaultScorer = NULL;
+    return REDISMODULE_ERR;
+  }
```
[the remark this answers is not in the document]
JoanFM: assert added

src/module-init/module-init.c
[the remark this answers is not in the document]
JoanFM: checked in tests, the order is different

src/module-init/module-init.c
```diff
@@ -195,6 +195,12 @@ int RediSearch_Init(RedisModuleCtx *ctx, int mode) {
     return REDISMODULE_ERR;
   }
 
+  ExtScoringFunctionCtx *scoreCtx = Extensions_GetScoringFunction(NULL, RSGlobalConfig.defaultScorer);
+  if (scoreCtx == NULL) {
+    DO_LOG("warning", "default scorer %s not a valid scorer", RSGlobalConfig.defaultScorer);
+    return REDISMODULE_ERR;
+  }
+
```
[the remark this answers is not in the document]
JoanFM: tests prove this

src/module-init/module-init.c
```diff
@@ -195,6 +195,13 @@ int RediSearch_Init(RedisModuleCtx *ctx, int mode) {
     return REDISMODULE_ERR;
   }
 
+  RS_ASSERT(RSGlobalConfig.defaultScorer);
+  ExtScoringFunctionCtx *scoreCtx = Extensions_GetScoringFunction(NULL, RSGlobalConfig.defaultScorer);
+  if (scoreCtx == NULL) {
+    DO_LOG("warning", "default scorer '%s' is not a valid scorer", RSGlobalConfig.defaultScorer);
```
kei-nan:
```suggestion
    DO_LOG("warning", "The scorer '%s' specified in the configuration for the default scorer is not a valid scorer", RSGlobalConfig.defaultScorer);
```

tests/cpptests/common.cpp
```diff
@@ -20,11 +20,11 @@
 extern "C" {
 
 static int my_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
-
   if (RedisModule_Init(ctx, REDISEARCH_MODULE_NAME, REDISEARCH_MODULE_VERSION,
                        REDISMODULE_APIVER_1) == REDISMODULE_ERR) {
     return REDISMODULE_ERR;
   }
+  RSGlobalConfig.defaultScorer = rm_strdup(DEFAULT_SCORER_NAME);
```
kei-nan, line 27: Seeing this after calling `RedisModule_Init` makes me think this initialization should be moved into `RedisModule_Init`.

src/config.c
```diff
@@ -193,6 +195,37 @@ int set_immutable_string_config(const char *name, RedisModuleString *val, void *
...
+    }
+
+    // Get the scorer name from the Redis module string
+    size_t len;
+    const char *newScorerName = RedisModule_StringPtrLen(val, &len);
+
+    // If Extension is not yet initialized, we will validate the defaultScorer after initialization for validation
+    if (Extensions_InitDone()) {
+      // Validate the scorer name against registered scorers only when the extension system is initialized
+      ExtScoringFunctionCtx *scoreCtx = Extensions_GetScoringFunction(NULL, newScorerName);
```
kei-nan, line 211: it looks like `Extensions_GetScoringFunction` is case sensitive.
Maybe we should move the input to be uppercase or inform the user about this requirement?

tests/cpptests/common.cpp
JoanFM: I agree I also had the same impression that this would require some refactoring, but I do not think is the correct scope right now

src/config.c
JoanFM: I will ask

JoanFM: This works exactly as the ones in the query, so I think it is coherent feature. If we want to change the behavior we can change it with all the context.

PR #7064, opened 2025-10-15
MOD-11867: Fix ACL test to validate index creation

raz-mon:
Fixes an ACL test to send the correct argument, and validate the index creation.

---

> [!NOTE]
> Refactors ACL tests to create vector-enabled indexes, add FT.HYBRID coverage, adjust permission/error assertions, and simplify command execution with alias and cluster-safe sug keys.
> 
> - **Tests (ACL)**:
>   - **Index setup**: Replace inline `FT.CREATE` with `create_index` helper adding `embedding VECTOR FLAT` schema; support optional `PREFIX`; add `index_to_drop` variable.
>   - **Permission checks**:
>     - Add `FT.HYBRID` to denied `index_commands` and use it with `VSIM @embedding`.
>     - Relax error assertions to generic "User does not have the required permissions"; when allowed, execute via `env.cmd` and only allow specific non-permission errors (e.g., missing cursor, duplicate doc).
>   - **Non-index commands**: Use `env.expect(...).noError()`; add `FT.ALIASADD` before `ALIASDEL`; use `{3}` hash tags for `FT.SUG*` keys to work in cluster.
>   - **Cleanup**: Remove final default-user verification block.
> - **Constants**: Update `READ_SEARCH_COMMANDS` to include `FT.HYBRID` and dedupe `FT.CURSOR`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6777b2a1d7a6de5843db701e33954e321ed1b0e4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #7072, opened 2025-10-16
MOD-11867: Fix ACL test

raz-mon:
Fixes an ACL test, and adds a test for `FT.HYBRID`

---

> [!NOTE]
> Updates ACL tests to use a vector index helper, broaden permission checks, add FT.HYBRID to read commands and coverage, and streamline non-index command handling.
> 
> - **Tests (ACL)**:
>   - **Key-permissions test (`test_acl_key_permissions_validation`)**:
>     - Add `create_index` helper to create indexes with vector field (`embedding` VECTOR FLAT) and optional `PREFIX`.
>     - Introduce `no_perm_index_to_drop` and use it in `FT.DROPINDEX` checks.
>     - Add permission check for `FT.HYBRID` and generalize expected error text to "User does not have the required permissions".
>     - Replace manual try/except with `env.expect(...).noError()` for non-index commands; use `env.cmd` for allowed index ops and assert only non-permission errors.
>     - Add `FT.ALIASADD` (to `perm_index`) so `FT.ALIASDEL` can be tested; use cluster-safe keys (`{3}`) for `FT.SUG*` commands.
>     - Remove final default-user access verification block.
>   - **Constants**: Extend `READ_SEARCH_COMMANDS` to include `FT.HYBRID` and clean up duplicates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e4b8398c69d7cd0ae653530542f67e0027018228. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/aggregate/aggregate.h
+++ b/src/aggregate/aggregate.h
@@ -41,7 +41,7 @@ struct QOptimizer;
 
 /*
  * A query can be of one type. So QEXEC_F_IS_AGGREGATE, QEXEC_F_IS_SEARCH, QEXEC_F_IS_HYBRID_TAIL,
- * QEXEC_F_IS_HYBRID_SEARCH_SUBQUERY, QEXEC_F_IS_HYBRID_VECTOR_AGGREGATE_SUBQUERY, and QEXEC_F_IS_HYBRID_COORDINATOR_SUBQUERY are mutually exclusive (Only one can be set).
+ * QEXEC_F_IS_HYBRID_SEARCH_SUBQUERY, QEXEC_F_IS_HYBRID_VECTOR_AGGREGATE_SUBQUERY are mutually exclusive (Only one can be set).
  */
 typedef enum {
   QEXEC_F_IS_AGGREGATE = 0x01,    // Is an aggregate command
@@ -114,12 +114,9 @@ typedef enum {
   // The query is a Vector Subquery of a Hybrid Request (aggregate equivalent)
   QEXEC_F_IS_HYBRID_VECTOR_AGGREGATE_SUBQUERY = 0x2000000,
 
-  // The query is a Coordinator Subquery of a Hybrid Request
-  QEXEC_F_IS_HYBRID_COORDINATOR_SUBQUERY = 0x4000000,
-
   // The query has an explicit SORT BY 0 step - no sorting at all
   // Currently only used in when QEXEC_F_IS_HYBRID_TAIL is set - i.e this is the tail part
-  QEXEC_F_NO_SORT = 0x8000000,
+  QEXEC_F_NO_SORT = 0x4000000,
 
   // The query is for debugging. Note that this is the last bit of uint32_t
   QEXEC_F_DEBUG = 0x80000000,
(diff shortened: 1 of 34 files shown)


tests/pytests/test_acl.py
```diff
@@ -233,37 +252,36 @@ def test_acl_key_permissions_validation(env):
...
+            env.cmd(*command)
         except Exception as e:
-            env.assertNotContains("User does not have the required permissions", str(e))
+            # Assert that the error is one of the expected errors
+            possible_errors = [
+                "Cursor not found",
+                "Cursor does not exist",
+                "Document already exists",
+            ]
+            env.assertTrue(any([err in str(e) for err in possible_errors]))
```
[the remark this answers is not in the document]
raz-mon: If it fails on permissions, we get an error that is not one of the expected ones, and thus we fail.

tests/pytests/test_acl.py
```diff
@@ -219,10 +234,14 @@ def test_acl_key_permissions_validation(env):
         ['FT.ADD', no_perm_index, 'h:doc1', '1.0', 'FIELDS', 'n', '5'],
         ['FT.GET', no_perm_index, 'h:doc1'],
         ['FT.DEL', no_perm_index, 'h:doc1'],
-        ['FT.DROPINDEX', 'index_to_drop'],
+        ['FT.DROPINDEX', no_perm_index_to_drop],
     ]
     for command in index_commands:
-        env.expect(*command).error().contains("User does not have the required permissions to query the index")
+        try:
```
kei-nan: maybe use expect().error()?

tests/pytests/test_acl.py
```diff
@@ -188,11 +188,24 @@ def test_acl_key_permissions_validation(env):
...
+            env.expect(
+                'FT.CREATE', index_name, 'PREFIX', len(prefixes), *prefixes,\
+                'SCHEMA', 'n', 'NUMERIC', 'embedding', 'VECTOR', 'FLAT', 6,\
+                'TYPE', 'FLOAT32', 'DIM', 2, 'DISTANCE_METRIC', 'L2'
+            ).ok()
+        else:
+            env.expect('FT.CREATE', index_name, 'SCHEMA', 'n', 'NUMERIC',\
+                'embedding', 'VECTOR', 'FLAT', 6, 'TYPE', 'FLOAT32', 'DIM', 2,\
+                'DISTANCE_METRIC', 'L2'
+            ).ok()
```
kei-nan:
```suggestion
        prefix_clause = ['PREFIX', len(prefixes), *prefixes] if prefixes else []
        env.expect(
            'FT.CREATE', index_name, *prefix_clause,\
            'SCHEMA', 'n', 'NUMERIC', 'embedding', 'VECTOR', 'FLAT', 6,\
            'TYPE', 'FLOAT32', 'DIM', 2, 'DISTANCE_METRIC', 'L2'
        ).ok()
```

tests/pytests/test_acl.py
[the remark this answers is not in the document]
kei-nan: what happens if command succeeds and no errors are received, shouldn't the test fail?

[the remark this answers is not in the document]
raz-mon: No, it should not.

tests/pytests/test_acl.py
raz-mon: Yeah I changed to this syntax since I thought I needed to use `conn.execute_command` for cluster support, but at the end didn't and didn't retrieve it. Reverting back.

tests/pytests/test_acl.py
```diff
@@ -233,37 +243,37 @@ def test_acl_key_permissions_validation(env):
...
-        except Exception as e:
-            env.assertNotContains("User does not have the required permissions", str(e))
-
-    # For completeness, we verify that the default user, which has permissions
-    # to access all keys, can access the index
-    env.expect('AUTH', 'default', 'nopass').true()
-    for command in index_commands + non_index_commands:
-        try:
-            env.execute_command(*command)
+            env.cmd(*command)
```
nafraf, line 270: Can we be more specific with the tests that will fail? Something like:
```suggestion
            env.cmd(*command)
            if [command[0], command[1]] in [
                ['FT.CURSOR', 'READ'],
                ['FT.CURSOR', 'DEL'],
                ['FT.ADD', perm_index],
            ]:
                env.assertTrue(False) # We expect a different error for these commands
            else:
                env.assertTrue(True) # No error expected
```

raz-mon: I think it's quite clear from the comment that all listed possible errors except the permissions one are fine.. No matter which command gets them.
I can make it more explicit or remove the errors by doing some operations before the test if you think it's important or can cause problems later.

nafraf:
Replaced by #7064

PR #7078, opened 2025-10-16
fix: fix Index prefixes

JoanFM:
Fixes the case where n_prefixes is 0 but we return true, this would make the success of the query dependant on which shard acts as coordinator.

---

> [!NOTE]
> Ensure index-prefix coherence (handling zero-prefix indexes), add a check flag to AREQ_ApplyContext, enforce only for internal/hybrid paths, and update config/spec/rules and tests accordingly.
> 
> - **Core API**:
>   - `AREQ_ApplyContext` gains `checkIndexCoherence` flag; callers updated (e.g., use `QEXEC_F_INTERNAL`).
>   - Index coherence checks fixed to correctly handle zero prefixes (`IsIndexCoherent`, `IsIndexCoherentWithQuery`), use `strtoll`, and compare against spec prefixes.
> - **Hybrid parsing/execution**:
>   - Coherence validation applied only for internal commands; sub-requests call `AREQ_ApplyContext(..., false)`.
>   - Enforce `_INDEX_PREFIXES` validation for internal flows.
> - **Config/Spec/Rules**:
>   - No-prefix indexes now stored as `nprefixes=0`, `prefixes=NULL` (instead of empty-string prefix).
>   - Creation and RDB load/save paths updated to support 0 prefixes; rules create prefix arrays only when `nprefixes>0`.
> - **Tests & Utilities**:
>   - Add `ArgvList.append` helper; tests append `_INDEX_PREFIXES` from spec when needed.
>   - Update/extend C++ and Python tests for internal hybrid flows and prefix coherence (including zero-prefix and multi-shard cases).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c2adeb772d11fae2addd4dbdaafa982d2fd3f7df. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -921,9 +921,8 @@ CONFIG_SETTER(setUpgradeIndex) {
       rule->prefixes[i] = rm_strdup(RedisModule_StringPtrLen(rule_prefixes.objs[i], NULL));
     }
   } else {
-    rule->nprefixes = 1;
-    rule->prefixes = rm_malloc(sizeof(char *));
-    rule->prefixes[0] = rm_strdup("");
+    rule->nprefixes = 0;
+    rule->prefixes = NULL;
   }
 
   // duplicate all rule arguments so it will leave after this function finish
--- a/src/rules.c
+++ b/src/rules.c
@@ -118,10 +118,12 @@ SchemaRule *SchemaRule_Create(SchemaRuleArgs *args, StrongRef ref, QueryError *s
     rule->lang_default = DEFAULT_LANGUAGE;
   }
 
-  rule->prefixes = array_new(HiddenUnicodeString*, args->nprefixes);
-  for (int i = 0; i < args->nprefixes; ++i) {
-    HiddenUnicodeString* p = NewHiddenUnicodeString(args->prefixes[i]);
-    array_append(rule->prefixes, p);
+  if (args->nprefixes > 0) {
+    rule->prefixes = array_new(HiddenUnicodeString*, args->nprefixes);
+    for (int i = 0; i < args->nprefixes; ++i) {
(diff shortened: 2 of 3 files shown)


oshadmi (approved):

JoanFM:
> can we have a flow test to it?

done

tests/pytests/test_hybrid_prefixes.py
```diff
@@ -49,14 +49,48 @@ def modify_index(conn, index_name, prefixes):
...
+
+    # Query via the cluster connection, such that we will get the mismatch error
+    commands = [
+        ['FT.HYBRID', index_name, 'SEARCH', 'text', 'VSIM', '@embedding', '$BLOB', 'PARAMS', '2', 'BLOB', '0']
+    ]
+
+    # Run commands on second shard (different index prefixes -> error)
+    for command in commands:
+        try:
+            second_conn.execute_command(*command)
```
Itzikvaknin, line 84: `first_conn.execute_command(*command)` is failing too, with the same error, how can we show the cluster/shard dependency here?

JoanFM: yes both fail. The feature is that they should fail irrespective of the shard u target

JoanFM:
As discussed with @oshadmi this PR is removed. We may have some tech debt to remove this.

Right now thanks to the "bug" we are trying to solve, the user is able to send internal (_FT.) queries in OSS without needing to pass _INDEX_PREFIXES 1 "" along (because the Spec has prefixes [""]) by default.

So the condition of n_prefixes == 0 => return True is only available if an intenral command is sent directly (not related to the regular coordinator-shard communication)

PR #7088, opened 2025-10-17
Extra Python GC tests and edits

chesedo:
This adds extra python tests to ensure we are confident with the #6878 implementation when switching the inverted index over to Rust. It also makes `totalCollected` signed to prevent an underflow when we switch.

#### Mark if applicable

---

> [!NOTE]
> Adds several GC robustness tests and changes collected-bytes stats to signed types to handle over-allocation cases.
> 
> - **GC Stats**:
>   - Change `ForkGCStats.totalCollected` and `InfoGCStats.totalCollectedBytes` from `size_t` to `ssize_t` to support cases where GC allocates more than it frees.
> - **Tests (Python)**:
>   - Numeric: `testNumericCompleteGCAndRepopulation`, `testNumericMergesTrees`.
>   - Tag: `testTagCompleteGCAndRepopulation`.
>   - Concurrency/Robustness: `testConcurrentFTInfoDuringIndexDeletion` (continuous `FT.INFO` during deletions/GC).
>   - OOM handling: `test_gc_oom` (bytes collected under tight/unlimited memory).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 60e32c8eba869ba5b66714352f828f1ec308474e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/fork_gc.h
+++ b/src/fork_gc.h
@@ -21,6 +21,8 @@ extern "C" {
 
 typedef struct {
   // total bytes collected by the GC
+  // This is signed because block splitting (when deltas are too big) can cause more bytes to be
+  // allocated by the GC than the number of bytes collected.
   ssize_t totalCollected;
   // number of cycle ran
   size_t numCycles;
@@ -119,7 +121,10 @@ void FGC_ForkAndWaitBeforeApply(ForkGC *gc);
 void FGC_Apply(ForkGC *gc);
 
 typedef struct InfoGCStats {
-  size_t totalCollectedBytes; // Total bytes collected by the GCs
+  // Total bytes collected by the GCs
+  // This is signed because block splitting (when deltas are too big) can cause more bytes to be
+  // allocated by a GC than the number of bytes collected.
+  ssize_t totalCollectedBytes;
   size_t totalCycles;         // Total number of cycles ran
   size_t totalTime;           // In ms
 } InfoGCStats;


src/fork_gc.h
```diff
@@ -21,7 +21,7 @@ extern "C" {
 
 typedef struct {
   // total bytes collected by the GC
-  size_t totalCollected;
+  ssize_t totalCollected;
```
JoanFM, line 26: typo?

JoanFM: I did not even know it existed? Why do  we want a signed version of it? Why would we have negative?
Also we should take a look at how is returned in INFO commands

meiravgri: https://github.com/RediSearch/RediSearch/blob/a4541c3e43d77e603b1b47341a56615c2942dc7c/src/fork_gc.h#L122

chesedo: The new GC coming with the Rust code will correctly split the blocks in an inverted index (unlike the C implementation) which might cause the number of collected bytes for a GC run to be negative. This is expected to be rare, but this code change prevents an underflow in the event it does happen.

chesedo: > I did not even know it existed?

I had to research to find it :sweat_smile:

tests/pytests/test_gc.py
```diff
@@ -72,6 +72,96 @@ def testNumericGCIntensive(env):
...
+    env.assertEqual(res, [[]])
+
+    # Verify search returns no results
+    search_res = env.cmd('ft.search', 'idx', '@id:[0 99]')
+    env.assertEqual(search_res[0], 0)
+
+    # Phase 3: Re-add documents with the same numeric values
+    NewDocs = 50
+    for i in range(NewDocs):
+        env.assertOk(env.cmd('ft.add', 'idx', 'newdoc%d' % i, 1.0, 'fields', 'id', str(i)))
```
meiravgri: please dont ever use ft.add :)
```suggestion
        env.assertOk(env.cmd('hset', 'newdoc%d' % i, 'id', str(i)))
```

to delete:
```
        env.assertOk(env.cmd('del', 'newdoc%d' % i)
```

tests/pytests/test_gc.py
```diff
@@ -72,6 +72,96 @@ def testNumericGCIntensive(env):
...
+    waitForIndex(env, 'idx')
+
+    # Phase 1: Add initial documents
+    InitialDocs = 255
+    for i in range(InitialDocs):
+        env.assertOk(env.cmd('ft.add', 'idx', 'doc%d' % i, 1.0, 'fields', 'id', str(i)))
+
+    # Verify initial state
+    res = env.cmd(debug_cmd(), 'DUMP_NUMIDX', 'idx', 'id')
+    env.assertEqual(len(res), 4)
```
meiravgri, line 139: why 4?

src/fork_gc.h
meiravgri: is it possible to add a test that will cover negative `totalCollected`?

tests/pytests/test_gc.py
```diff
@@ -72,6 +72,96 @@ def testNumericGCIntensive(env):
...
+    env.assertEqual(len(all_doc_ids), NewDocs)
+
+    # Verify search works
+    search_res = env.cmd('ft.search', 'idx', '@id:[0 49]')
+    env.assertEqual(search_res[0], NewDocs)
+
+@skip(cluster=True)
+def testNumericMergesTrees(env):
+    """Test to check the numeric index trees merges when half or more of all the trees are empty"""
+    env.expect(config_cmd(), 'set', 'FORK_GC_CLEAN_THRESHOLD', 0).equal('OK')
```
meiravgri, line 128: let's also set a very larg interval to make sure we control its trigger rather than the index
```
env.expect(config_cmd(), 'SET', 'FORK_GC_RUN_INTERVAL', '30000').ok()
```

tests/pytests/test_gc.py
chesedo: Done in https://github.com/RediSearch/RediSearch/pull/7088/commits/952b2dbd00ac0a1d090978c2403a033e45631bc8

src/fork_gc.h
chesedo: We can't test this as discussed. But I added the comments here https://github.com/RediSearch/RediSearch/pull/7088/commits/60e32c8eba869ba5b66714352f828f1ec308474e

tests/pytests/test_gc.py
chesedo: Not needed as discussed since the auto run GC will cause the system to be in the same state

src/fork_gc.h
JoanFM: We need to make sure that when this INFO is reported in the INFO command, it is using the proper type.

chesedo: @JoanFM I double checked this with @meiravgri . It uses the double type which should be correct.

JoanFM: Ok, will double check I am not affecting on my PR

JoanFM (approved):

PR #7101, opened 2025-10-20
[MOD-11532] Create `query_error_ffi` crate

enricozb:
Creates a `query_error_ffi` crate which generates the C header that will replace the existing `src/query_error.h` header file.

This exports a (renamed) `OpaqueQueryError` which has the same size and alignment of the Rust `QueryError` struct. Take a close look at `query_error_ffi/src/opaque.rs`.

Unfortunately there are a few C functions which have to remain in C:
1. `QueryError_SetWithUserDataFmt` and `QueryError_SetWithoutUserDataFmt` are variadic, and so they will be implemented in C and call `QueryError_SetDetail` or `QueryError_SetError`.
2. `QueryError_ReplyAndClear` is a macro that references another dependency which isn't ported yet.
3. `QueryError_FmtUnknownArg` references `ArgsCursor` which is a dependency which has not been ported yet.

This does not touch any C code.

Note that the new exported variants of `QueryErrorCode` take the form `QUERY_ERROR_CODE_*`, which is incompatible with the existing `QUERY_E*`-style identifiers. Unfortunately, without going against Rust naming conventions, there is no cbindgen config that can export the same identifiers as the C version.

This next PR renames these identifiers: https://github.com/RediSearch/RediSearch/pull/7102

---

> [!NOTE]
> Introduces `c_entrypoint/query_error_ffi` to expose `QueryError` over a C ABI and generate `headers/query_error.h`, updating workspace deps accordingly.
> 
> - **FFI**:
>   - Add `c_entrypoint/query_error_ffi` crate exposing `QueryError` via C ABI (`src/lib.rs`).
>   - Implement opaque wrapper `OpaqueQueryError` with size/alignment checks (`src/opaque.rs`).
>   - Generate header via `build.rs` using `cbindgen`.
> - **Headers**:
>   - Add generated `headers/query_error.h` exporting `QueryError`, `QueryErrorCode`, helper macros, and declarations for variadic C functions (left for C implementation).
>   - Add `cbindgen.toml` configuring includes, macros, trailer, and type renames.
> - **Workspace**:
>   - Update `Cargo.toml` to include `query_error` in workspace dependencies; `Cargo.lock` adds `query_error_ffi`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 42e91177340f9bf5a15053055f883cb44e953e7c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/query_error_ffi/src/opaque.rs
+++ b/src/redisearch_rs/c_entrypoint/query_error_ffi/src/opaque.rs
@@ -97,7 +97,7 @@ impl QueryErrorExt for QueryError {
     }
 
     fn as_opaque_ptr(&self) -> *const OpaqueQueryError {
-        std::ptr::from_ref(&self).cast()
+        std::ptr::from_ref(self).cast()
     }
 
     fn as_opaque_mut_ptr(&mut self) -> *mut OpaqueQueryError {


src/redisearch_rs/c_entrypoint/query_error_ffi/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+use std::os::raw::c_char;
+
+pub use crate::opaque::{OpaqueQueryError, QueryErrorExt};
+pub use query_error::QueryErrorCode;
+
+/// Returns the default [`QueryError`].
+///
+/// # Safety
+///
+/// `OpaqueQueryError` and `QueryError` must have the same size and alignment.
```
hdoordt: This is already covered by the const assertion in opaque.rs, right?

src/redisearch_rs/c_entrypoint/query_error_ffi/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+    unsafe { !QueryError_IsOk(query_error) }
+}
+
+/// Returns a human-readable string representing the provided [`QueryErrorCode`].
+///
+/// # Safety
+///
+/// This function should always return without a panic for any value provided.
+/// It is unique among the `QueryError_*` API as the only function which allows
+/// an invalid [`QueryErrorCode`] to be provided.
```
hdoordt, line 54: This is not really a memory safety concern relevant for the caller of this function, right? Just for the implementor. If so, let's not mark it unsafe and remove the `# Safety` header

src/redisearch_rs/c_entrypoint/query_error_ffi/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+}
+
+/// Sets the [`QueryErrorCode`] for a [`QueryError`].
+///
+/// This does not mutate `query_error` if it already has an error set.
+///
+/// # Safety
+///
+/// - `query_error` must have been created by [`QueryError_Default`].
+/// - `code` must be a valid variant of [`QueryErrorCode`].
```
hdoordt: A panic occurs (line 113) if that code is invalid, not UB. Let's put it under a `# Panics` header instead

src/redisearch_rs/c_entrypoint/query_error_ffi/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+}
+
+/// Clones the `src` [`QueryError`] into `dest`.
+///
+/// This does nothing if `dest` already has an error set.
+///
+/// # Safety
+///
+/// - `src` must have been created by [`QueryError_Default`].
+/// - `dest` must have been created by [`QueryError_Default`].
```
hdoordt, line 150: Can we support cloning into an uninitialized `dest`? Or is it just set up this way so that it mimics C behaviour?

src/redisearch_rs/c_entrypoint/query_error_ffi/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+        .unwrap_or_else(|| query_error.code().to_c_str())
+        .as_ptr()
+}
+
+/// Returns an message of a [`QueryError`].
+///
+/// This preferentially returns the private message if any, or the public
+/// message if any, lastly defaulting to the error code's string error.
+///
+/// If `obfuscate` is set, the private message is not returned.
```
hdoordt:
```suggestion
/// If `obfuscate` is set, the private message is not returned,
/// but the public message is returned instead.
```

src/redisearch_rs/c_entrypoint/query_error_ffi/src/lib.rs
enricozb: Not really, because this function has to check whether `dest` has an error code set or not.

In order to do that, `dest` must be initialized. I'm not sure how we would check whether it was uninitialized.

src/redisearch_rs/c_entrypoint/query_error_ffi/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+edition.workspace = true
+license-file.workspace = true
+publish.workspace = true
+
+[dependencies]
+query_error.workspace = true
+
+[build-dependencies]
+cbindgen.workspace = true
+build_utils = { path = "../../build_utils" }
```
enricozb, line 13: `build_utils` should be a workspace dep

raz-mon (approved):
LGTM - Nice!

PR #7107, opened 2025-10-20
MOD-11970: Fix memory leak 

nafraf:
## Describe the changes in the pull request

A clear and concise description of what the PR is solving, including:
1. Current: The current state briefly
2. Change: What is the change
3. Outcome: Adding the outcome

#### Which additional issues this PR fixes
1. MOD-...
2. #...

#### Main objects this PR modified
1. ...

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

---

> [!NOTE]
> Reworks metric propagation to store keys in MetricRequest and access them safely from iterators, updates hybrid/idlist iterators and pipeline accordingly, and enables two hybrid vector range tests.
> 
> - **Core/Query**:
>   - Replace `MetricRequest.key_ptr` with direct `key` storage and add `MetricRequest_GetKey(...)` helper for safe access via pointer-to-pointer array.
>   - Update `addMetricRequest` and vector node evaluation to pass iterator `metricRequestIdx` and `metricRequestsP` instead of raw key pointers.
> - **Iterators**:
>   - `HybridIterator`/`MetricIterator`: add `metricRequestIdx` and `metricRequestsP`; use `MetricRequest_GetKey(...)` when adding results; init/clear on create/free; remove `ownKey` usage.
>   - `hybrid_reader.c` and `idlist_iterator.c`: include `query.h`, update heap insert/yield paths to fetch key via `MetricRequest_GetKey`.
> - **Pipeline**:
>   - `getAdditionalMetricsRP`: set `requests[i].key` directly when creating `RLookupKey`.
> - **Headers**:
>   - Add forward declarations and new fields in `hybrid_reader.h`/`idlist_iterator.h`; adjust includes.
> - **Tests**:
>   - Unskip cluster for `test_hybrid_vector_direct_blob_range` and `..._with_filter`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 84f922b5ed7d8f6d4cf3390e428cec7a04febb71. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #7108, opened 2025-10-20
MOD-11970: avoid use MetricRequest in pipeline construction after free in intersection optimization

oshadmi:
Avoid use `MetricRequest` in pipeline construction after free in intersection optimization in `IntersectionIteratorReducer`
Add some simple "smart" pointer to detect that pointer was invalidated (by the intersection iterator freeing child vector iterator) 

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

---

> [!NOTE]
> Introduce a handle-based reference for metric RLookupKeys to prevent use-after-free, update iterators/pipeline accordingly, and enable hybrid vector tests on cluster.
> 
> - **Core (Query/AST)**:
>   - Introduce `RLookupKeyHandle` smart handle; `MetricRequest` now holds `key_handle` (not `key_ptr`).
>   - Update `addMetricRequest(...)` and vector node eval to create/set handles and back-references; free handles in `QAST_Destroy`.
> - **Iterators**:
>   - `HybridIterator` and `MetricIterator` add `keyHandle` back-reference; set on creation and invalidate on `Free`.
>   - Include `query.h` where needed.
> - **Pipeline**:
>   - `getAdditionalMetricsRP` writes to iterator keys via `key_handle` only if `is_valid`.
>   - Add necessary includes for iterators.
> - **Tests**:
>   - Remove `@skip(cluster=True)` from hybrid vector tests to run on cluster.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3c27cdd73014d31873fdcf0b0ee3d4305cba67a0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


oshadmi:
Replaces #7107

src/query.c
```diff
@@ -1002,17 +1004,28 @@ static QueryIterator *Query_EvalVectorNode(QueryEvalCtx *q, QueryNode *qn) {
...
+      metricIt->keyHandle = handle; // Set up back-reference
     } else {
       // Only reason to get an iterator of type different than HYBRID or METRIC
       // is if the entire iterator was optimized away
-      RS_ASSERT(it->type == EMPTY_ITERATOR);
+      rm_free(handle);
+      handle = NULL;
     }
+
+    request->key_handle = handle;
```
nafraf, line 1025: Are we assigning the `key_handle` even for EMPTY_ITERATOR?

src/query.c
```diff
@@ -1002,17 +1004,28 @@ static QueryIterator *Query_EvalVectorNode(QueryEvalCtx *q, QueryNode *qn) {
   // If iterator was created successfully, and we have a metric to yield, update the
   // relevant position in the metricRequests ptr array to the iterator's RLookup key ptr.
   if (it && qn->vn.vq->scoreField) {
+    MetricRequest *request = array_ensure_at(q->metricRequestsP, idx, MetricRequest);
```
nafraf, line 1008: Do we need this for `EMPTY_ITERATOR` as well?

src/query.c
oshadmi: The else block sets it to `NULL`
But I will make it cleaner

src/query.c
oshadmi: No.
Will fix this.

nafraf (dismissed):

src/query.h
```diff
@@ -27,10 +27,16 @@
 extern "C" {
 #endif
 
+// Smart pointer handle for RLookupKey that can be invalidated when iterator is freed
+typedef struct RLookupKeyHandle {
```
JoanFM, line 31: Should be made it clear that is not thread safe. So that we use with care.

It seems the current usage happens in the same thread.

JoanFM:
Would also work to use all our StrongRef, WeakRef constructs?

JoanFM (dismissed):

src/query.h
oshadmi: Seems it is a very dedicated and specific struct for this purpose only and should not be reused in other use cases.

PR #7121, opened 2025-10-22
fix lifetimes in RQEIterator rust trait

GlenDC:

MOD-11768 unlocked the mutation of the RSIndexResult return values of the read and skip_to methods, part of the RQEIterator interface.

As part of this change the lifetimes of RQEIterator was increased from one (index) to two (iterator == self + index). That sub change turned out to be overly restricted as it forbids borrowing one property while also borrowing another property exclusively already.

This commit rectifies this situation by once again only having a single lifetime (index) for RQEIterator while still allowying the same result-mutation
capabiities unlocked since MOD_11768

None are tracked in jira, it was originally solved as part of
MOD-10081 (non-optimized optional iterator) but it is
best already merged prior to as to unlock other team members
such as @BenGoldberger and @gdesmott.

`RQEIterator` defined in `src/redisearch_rs/rqe_iterators/src/lib.rs`
and all Rust iterators implementing it. Users of the object should in most cases feel no effect,
but use-cases where the trait was used as a trait bound would also need to be updated (once again).

#### Mark if applicable

- [x] This PR introduces API changes (in a Rust trait not yet used in production)

---

> [!NOTE]
> Simplifies `RQEIterator` to use only the `'index` lifetime, updating method signatures and the `Empty`, `IdList`, and `Wildcard` implementations accordingly.
> 
> - **API**:
>   - **`RQEIterator<'index>`**: Remove the extra `'iterator` lifetime; methods now take `&mut self` and return `&mut RSIndexResult<'index>`.
>   - **Signatures**: `read`, `skip_to`, `revalidate`, and trait property methods updated; `SkipToOutcome` and `RQEValidateStatus` usages now employ elided lifetimes (`'_'`).
> - **Iterators**:
>   - **`empty.rs`**, **`id_list.rs`**, **`wildcard.rs`**: Update implementations to the new trait signatures; logic unchanged.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 998bc4626aa10d18362ba273c0969cd5de942b6f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/redisearch_rs/rqe_iterators/src/empty.rs
```diff
@@ -20,17 +20,15 @@ use crate::{RQEIterator, RQEIteratorError, SkipToOutcome};
...
+    fn read(&mut self) -> Result<Option<&mut RSIndexResult<'index>>, RQEIteratorError> {
         Ok(None)
     }
 
     fn skip_to(
-        &'iterator mut self,
+        &mut self,
         _doc_id: t_docId,
-    ) -> Result<Option<SkipToOutcome<'iterator, 'index>>, RQEIteratorError> {
+    ) -> Result<Option<SkipToOutcome<'_, 'index>>, RQEIteratorError> {
```
gdesmott, line 31: Would it not be clearer to keep naming this first lifetime here?

GlenDC: this is afaik std convention to keep this `'_`, I did not invent this pattern.  If you do not do this we force all implementations to write `fn name<'iterator>(&'iterator mut self, ...) -> Result<...<...'iterator, 'index>>, ...> { ... }` which anyway boils down the the same as not specifying it at all.

For the same reason clippy also warns against defining something like `impl<'a> struct Foo<'a>` when not requiring the use of it and instead define it as `impl struct Foo<'_>`.

That said rust community conventions are also just that and in case we have a policy within the Redisearch code to do it differently do let me know. Or in case you believe I misinterpret this convention do correct me. Otherwise if you do agree you can resolve this comment.

gdesmott: Fair enough, let's keep it as this then.

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -60,31 +58,31 @@ pub trait RQEIterator<'iterator, 'index> {
     ///
     /// Return `Ok(SkipToOutcome::Found)` if the iterator has found a record with the `docId` and `Ok(SkipToOutcome::NotFound)`
     /// if the iterator found a result greater than `docId`. 'None" will be returned if the iterator has reached the end of the index.
-    fn skip_to(
+    fn skip_to<'iterator>(
         &'iterator mut self,
         doc_id: t_docId,
     ) -> Result<Option<SkipToOutcome<'iterator, 'index>>, RQEIteratorError>;
```
BenGoldberger: why does the lifetime is still in the skip_to?

src/redisearch_rs/fnv/src/lib.rs
```diff
@@ -49,7 +49,7 @@ impl Fnv32 {
     /// [offset basis]: http://www.isthe.com/chongo/tech/comp/fnv/#FNV-param
     #[inline]
     #[must_use]
-    pub fn with_offset_basis(offset_basis: u32) -> Fnv32 {
+    pub const fn with_offset_basis(offset_basis: u32) -> Fnv32 {
```
gdesmott, line 52: We should either merge this PR right away or move this to another PR and merge it now as CI is currently broken.

(I still don't understand how this regression passed the CI and managed to be merged).

gdesmott: it's now in https://github.com/RediSearch/RediSearch/pull/7124

src/redisearch_rs/rqe_iterators/src/lib.rs
GlenDC: It is not explicitly needed @BenGoldberger, and it was already removed in the `impl` blocks, seems I forgot to remove it from the trait. But just so you understand, that lifetime is _always_ there.

THe signature is now:

```
fn skip_to(
        &mut self,
        doc_id: t_docId,
    ) -> Result<Option<SkipToOutcome<'_, 'index>>, RQEIteratorError>;
```

However in Rust _every_ reference HAS a lifetime. `'_` is a reserved lifetime notation that can only be used in some locations. When Rust was still in its early days you had to explicitly define lifetimes for pretty much any reference and situation. Similar to how you still define types in function signatures. The same was true for lifetimes.

Over time, the rust developers improved the algorithms and detections within the rust compiler and now in a lot of situations you do not really have to define lifetimes at all. To the point that most rust beginners might have got very far in their rust journey and even made projects without ever seeing a lifetime to begin with. Outside of perhaps errors, but certainly not defining it themselves.

This is called Lifetime elision and you can learn more about it at https://doc.rust-lang.org/nomicon/lifetime-elision.html ; The nomicon is a pretty good reference FWIW once you start to feel comfortable in Rust.

To get back to the situation at hand, the signature is really in the end really this (if we desugar the "sugar"):

```
fn skip_to<'1>(
        &'1 mut self,
        doc_id: t_docId,
    ) -> Result<Option<SkipToOutcome<'1, 'index>>, RQEIteratorError>;
```

That's the reason why the `impl` blocks already didn't define it, despite the trait still defining it and it still working. The reason this worked is because they were the same already.

Hope that makes sense to you. Or perhaps you knew this already and I'm being overly verbose in my feedback, in which my appologies. Hope it clears it up in which case this can be merged now? Thanks for catching it though, as even though it wouldn't have caused any issues, it's good to keep it clean and lean anyway.

BenGoldberger: thanks, that clearify it

src/redisearch_rs/fnv/src/lib.rs
```diff
@@ -104,7 +104,7 @@ impl Fnv64 {
     /// [offset basis]: http://www.isthe.com/chongo/tech/comp/fnv/#FNV-param
     #[inline]
     #[must_use]
-    pub fn with_offset_basis(offset_basis: u64) -> Fnv64 {
+    pub const fn with_offset_basis(offset_basis: u64) -> Fnv64 {
```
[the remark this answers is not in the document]
GlenDC: nah it's all good

PR #7122, opened 2025-10-22
Rust inverted index optimizations

chesedo:
Some optimizations to improve the performance of the Rust inverted index implementation

#### Mark if applicable

---

> [!NOTE]
> Aggressively inlines hot decode/reader paths, switches buffer-end checks to position-based logic, and streamlines numeric decoding to reduce copies.
> 
> - **Performance/Inlining**:
>   - Add `#[inline(always)]` to `decode`/`next_record`/`seek_record` across decoders (`DocIdsOnly`, `FieldsOnly(_Wide)`, `Freqs*`, `OffsetsOnly`, `Full(_Wide)`, `Numeric`) and readers (`IndexReaderCore`, `Filter*Reader`).
>   - Inline helper fns in numeric module and `qint::qint_decode`.
> - **Reader logic**:
>   - Replace `BufRead::fill_buf()` emptiness checks with `Cursor::position() >= len` in block iteration (`IndexBlock::repair`, `IndexReaderCore::next_record`).
>   - Drop unused `BufRead` import.
> - **Numeric decoding**:
>   - Optimize read helpers to avoid extra copies using new `FromSlice` trait (`u64`, `f32`, `f64`) and slice-based conversions.
>   - Mark read helpers as inline and simplify parsing paths.
> - **Misc**:
>   - Mark `NumericFilter::value_in_range` as inline.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e0bd9e26b8d34d9ecc423c383e80fc7cae2c5c12. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/inverted_index/src/numeric.rs
+++ b/src/redisearch_rs/inverted_index/src/numeric.rs
@@ -511,13 +511,8 @@ fn read_u64_and_u64<R: Read>(
     // Use one read since it is faster
     reader.read_exact(&mut buffer[..total_bytes])?;
 
-    let mut first = [0; 8];
-    first[..first_bytes].copy_from_slice(&buffer[..first_bytes]);
-    let first = u64::from_le_bytes(first);
-
-    let mut second = [0; 8];
-    second[..second_bytes].copy_from_slice(&buffer[first_bytes..first_bytes + second_bytes]);
-    let second = u64::from_le_bytes(second);
+    let first = u64::from_slice(&buffer[..first_bytes]);
+    let second = u64::from_slice(&buffer[first_bytes..first_bytes + second_bytes]);
 
     Ok((first, second))
 }
@@ -530,13 +525,8 @@ fn read_u64_and_f32<R: Read>(reader: &mut R, first_bytes: usize) -> std::io::Res
     // Use one read since it is faster
     reader.read_exact(&mut buffer[..total_bytes])?;
 
-    let mut first = [0; 8];
-    first[..first_bytes].copy_from_slice(&buffer[..first_bytes]);
-    let first = u64::from_le_bytes(first);
-
-    let mut second = [0; 4];
-    second.copy_from_slice(&buffer[first_bytes..first_bytes + 4]);
-    let second = f32::from_le_bytes(second);
+    let first = u64::from_slice(&buffer[..first_bytes]);
+    let second = f32::from_slice(&buffer[first_bytes..first_bytes + 4]);
 
     Ok((first, second))
 }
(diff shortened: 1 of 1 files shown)


src/redisearch_rs/inverted_index/src/numeric.rs
```diff
@@ -512,15 +515,14 @@ fn read_u64_and_u64<R: Read>(
     first[..first_bytes].copy_from_slice(&buffer[..first_bytes]);
     let first = u64::from_le_bytes(first);
 
-    let second = u64::from_le_bytes(
-        buffer[first_bytes..first_bytes + 8]
-            .try_into()
-            .expect("to convert slice into array"),
-    );
+    let mut second = [0; 8];
+    second[..second_bytes].copy_from_slice(&buffer[first_bytes..first_bytes + second_bytes]);
```
GlenDC: are you saying that this manually written code is faster than `try_into`???? As in the end that alos just turns it into an array, which is also what crates like `byteorder` do. Are we testing this in optimized release builds? I would be very surprised if we have to manually write this code to get this behaviour as I would expect this in an optimized release build to be also pretty much this code. Is it not???

GlenDC: Ah... https://godbolt.org/z/dGToYdcbE
I suppose the reason is that you avoid the code

```
 test    al, 1
        jne     .LBB12_1
        jmp     .LBB12_4
```
which is basically the unwrap which either goes to panic w/ unwind code.
Branching code can be indeed be something you feel in a hot path....

Damn... Bit odd there is no other unsafe path you can do here.
I guess manual code it is for now. Which is also safer anyhow,
even though it can hide a logic bug as copy from slice deals perhaps
a bit too gracefully with short/long buffers..

GlenDC: Given how often you do this trick for all kind of numerics you might want to introduce an inline function or extension trait for this. As the problem here is not the use of a function or trait. The problem is the `unwrap` call required when using `try_into`. So might still be safer to introduce a module where you have code to deal with this....

I would introduce a module with functions (not traits), mark them `always(inline)` and add some debug_assert to do check bounds as copy_from_slice doesn't do this. To ensure you are giving a long enough sub-slice...

e..g something like `f32_from_slice(&[u8], offset: usize)`, or something like that and that for all the ones you want... This way we can add some tests and do not have to do it manually each time like a caveman.

chesedo: Yip, this avoids the `unwrap()` overhead. The extra method / function sounds good. I'll add that

chesedo: Done in https://github.com/RediSearch/RediSearch/pull/7122/commits/e0bd9e26b8d34d9ecc423c383e80fc7cae2c5c12

GlenDC: Ah sorry @chesedo added my comments on the commit instead of this PR, silly github didn't realise there was a difference: https://github.com/RediSearch/RediSearch/commit/e0bd9e26b8d34d9ecc423c383e80fc7cae2c5c12#r168571959

PR #7127, opened 2025-10-22
ci: run lint in merge-to-queue workflow

gdesmott:
Should prevent CI regression when linter rules are updated.

#### Mark if applicable

---

> [!NOTE]
> Runs lint in the merge-to-queue workflow and requires it for PR validation.
> 
> - **CI/Workflow (`.github/workflows/event-merge-to-queue.yml`)**:
>   - Add `lint` job using `./.github/workflows/task-lint.yml` with `secrets: inherit`.
>   - Update `pr-validation` to depend on `lint`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e3c6da99ec27c5495defe5452fb2d6635786cffc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/event-merge-to-queue.yml
+++ b/.github/workflows/event-merge-to-queue.yml
@@ -21,6 +21,10 @@ jobs:
     with:
       repo: redis/redis
 
+  lint:
+    uses: ./.github/workflows/task-lint.yml
+    secrets: inherit
+
   test-linux:
     needs: get-latest-redis-tag
     uses: ./.github/workflows/flow-linux-platforms.yml
@@ -69,6 +73,7 @@ jobs:
   pr-validation:
     needs:
       - get-latest-redis-tag
+      - lint
       - test-linux
       - test-macos
       # - coverage


.github/workflows/event-merge-to-queue.yml
```diff
@@ -21,6 +21,11 @@ jobs:
     with:
       repo: redis/redis
 
+  lint:
+    needs: get-latest-redis-tag
```
GuyAv46: Why is it needed?

gdesmott: I don't know, I just copied the same logic as the other jobs. :-D

Should I remove it then?

GuyAv46: It means this job will wait for `get-latest-redis-tag` to complete successfully before running. If we don't need its output, we can avoid this dependency

gdesmott: I guess we can remove it then.

PR #7077, opened 2025-10-16
MOD-11846: Add FT.HYBRID to commands.json file

kei-nan:
Add the FT.HYBRID command syntax to the commands.json file.
Used augment to generate the right section with some light editing to remove unsupported keywords like withcursor/dialect and withscores.

A clear and concise description of what the PR is solving, including:
1. Current: commands.json is not up to date
2. Change: update the commands.json with the FT.HYBRID syntax.
3. Outcome: Expose the syntax to interested parties

#### Main objects this PR modified
1. commands.json file

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

---

> [!NOTE]
> Adds `FT.HYBRID` command definition enabling hybrid text and vector similarity search with configurable combining, sorting, loading, grouping, and apply/filter options.
> 
> - **Search**:
>   - **New Command**: Add `FT.HYBRID` to `commands.json` (since `8.4.0`).
>     - Combines text `SEARCH` with vector `VSIM` (`KNN`/`RANGE`, optional `FILTER`).
>     - Optional `COMBINE` methods: `RRF` and `LINEAR` (with tuning params and score aliasing).
>     - Result controls: `LIMIT`, `SORTBY`/`NOSORT`, `PARAMS`, `TIMEOUT`, `FORMAT`, `LOAD` (`*` or fields).
>     - Aggregation-like ops: `GROUPBY` with multiple `REDUCE` functions.
>     - Post-processing: `APPLY` expressions (string, math, time, geo utilities) and `FILTER` expression.
>     - Group: `search`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e576e9e8987c9810d83bd80e07337f63b43d6b59. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #7128, opened 2025-10-22
MOD-11846: Add FT.HYBRID to commands.json file - With Updated LOAD

kei-nan:
See Original PR:
https://github.com/RediSearch/RediSearch/pull/7077

Had to change LOAD clause to fit FT.AGGREGATE syntax
Updated Alpha Beta Syntax so they are considered optional as one block

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -2092,14 +2092,21 @@
                     "type": "integer"
                   },
                   {
-                    "name": "alpha",
-                    "type": "double",
-                    "token": "ALPHA"
-                  },
-                  {
-                    "name": "beta",
-                    "type": "double",
-                    "token": "BETA"
+                    "name": "weights",
+                    "type": "block",
+                    "optional": true,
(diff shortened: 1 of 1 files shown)


commands.json
```diff
@@ -1898,5 +1898,845 @@
...
+            "name": "field",
+            "type": "string",
+            "multiple": true
+          }
+        ]
+      },
+      {
+        "name": "loadall",
+        "type": "pure-token",
+        "token": "LOAD *",
```
nafraf, line 2252: We support multiple `LOAD` clauses, then it's valid to have `LOAD n fields .... LOAD *`, we don't need to put the LOADs in a block `oneof`, right?

kei-nan: I did what FT.AGGREGATE did, that was the request we got in order not to break existing components.

oshadmi (approved):

PR #7150, opened 2025-10-24
[RED-165934] [RED-165935] add memory object loading and dumping to API

JoanFM:

Aim is to work in pair with https://github.com/redislabsdev/RediSearchDisk/pull/11

---

> [!NOTE]
> Adds SearchDisk memory-object support (RDB aux load/save), registers a new module type, updates Disk API to pass mem-obj, and reopens disk after loading.
> 
> - **Disk API & Implementation**:
>   - Extend `BasicDiskAPI.open` to accept a `RedisSearchDiskMemObject *memObj`; add `MemObjectAPI` (`fromRDB`, `toRDB`, `free`).
>   - Manage `disk_mem_obj` in `search_disk.c`; pass it to `open`; free on close.
>   - Implement RDB aux persistence for the mem-object (`SearchDisk_AuxLoadFromRDB`, `SearchDisk_AuxSaveToRDB`).
>   - Add `SearchDisk_ReOpen(ctx)` to re-open the DB with the loaded mem-object.
> - **Module Integration**:
>   - Register new data type `ft_search_disk0` via `SearchDisk_RegisterType(ctx)` when `isFlex`.
>   - On loading ended event, call `SearchDisk_ReOpen(ctx)` (Flex only).
> - **Spec loading (Flex/disk)**:
>   - When spec is on disk, load `sp->terms` with `TrieType_GenericLoad(rdb, 0)` during RDB load.
> - **Headers**:
>   - Update `search_disk.h`/`search_disk_api.h` with mem-object types/APIs and new functions (`ReOpen`, aux load/save, register type).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b7ffaeaebe73a01c04dfd79f595f2fd59161fe77. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -161,11 +161,24 @@ static inline bool checkEnterpriseACL(RedisModuleCtx *ctx, IndexSpec *sp) {
   return !IsEnterprise() || ACLUserMayAccessIndex(ctx, sp);
 }
 
-// OOM guardrail with heuristics
+// OOM check with heuristics
 // TODO: add heuristics
 // Assumes the GIL is held by the caller
-bool estimateOOM(RedisModuleCtx *ctx) {
+static inline bool estimateOOM(RedisModuleCtx *ctx) {
   return RedisMemory_GetUsedMemoryRatioUnified(ctx) > 1;
+}
+
+// OOM guardrail for queries function
+// Such as DistSearchCommand/DistAggregateCommand and hybridCommandHandler
+// Assumes the GIL is held by the caller
+// Returns true if the query should be aborted due to OOM
+bool QueryMemoryGuard(RedisModuleCtx *ctx) {
+  // Check OOM if OOM policy is not ignore
+  if (RSGlobalConfig.requestConfigParams.oomPolicy != OomPolicy_Ignore) {
+    // No need to hold the GIL since we are not in a background thread
+    return estimateOOM(ctx);
+  }
+  return false;
 }
 
 // Returns true if the current context has permission to execute debug commands
@@ -1127,8 +1140,8 @@ int RegisterRestoreIfNxCommands(RedisModuleCommand *restoreCmd) {
 
 Version supportedVersion = {
     .majorVersion = 8,
-    .minorVersion = 0,
-    .patchVersion = 0,
+    .minorVersion = 3,
+    .patchVersion = 200,
 };
 
 static void GetRedisVersion(RedisModuleCtx *ctx) {
@@ -3210,6 +3223,12 @@ int DistAggregateCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
     return RedisModule_WrongArity(ctx);
   }
 
+  // Memory guardrail
+  if (QueryMemoryGuard(ctx)) {
+    RedisModule_Log(ctx, "notice", "Not enough memory available to execute the query");
+    return RedisModule_ReplyWithError(ctx, QueryError_Strerror(QUERY_EOOM));
+  }
+
   // Coord callback
   ConcurrentCmdHandler dist_callback = RSExecDistAggregate;
 
@@ -3230,14 +3249,6 @@ int DistAggregateCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
     return RedisModule_ReplyWithErrorFormat(ctx, "No such index %s", idx);
   }
 
-  // Check OOM if OOM policy is not ignore
-  if (RSGlobalConfig.requestConfigParams.oomPolicy != OomPolicy_Ignore) {
-    // No need to hold the GIL since we are not in a background thread
-    if (estimateOOM(ctx)) {
-      RedisModule_Log(ctx, "notice", "Not enough memory available to execute the query");
-      return RedisModule_ReplyWithError(ctx, QueryError_Strerror(QUERY_EOOM));
-    }
-  }
 
   bool isProfile = (RMUtil_ArgIndex("FT.PROFILE", argv, 1) != -1);
   // Check the ACL key permissions of the user w.r.t the queried index (only if
@@ -3266,6 +3277,12 @@ int DistHybridCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
     return RedisModule_ReplyWithError(ctx, CLUSTERDOWN_ERR);
   } else if (argc < 3) {
     return RedisModule_WrongArity(ctx);
+  }
+
+  // Memory guardrail
+  if (QueryMemoryGuard(ctx)) {
+    RedisModule_Log(ctx, "notice", "Not enough memory available to execute the query");
+    return RedisModule_ReplyWithError(ctx, QueryError_Strerror(QUERY_EOOM));
   }
 
   // Coord callback
@@ -3597,6 +3614,12 @@ int DistSearchCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
     return RedisModule_WrongArity(ctx);
   }
 
+  // Memory guardrail
+  if (QueryMemoryGuard(ctx)) {
+    RedisModule_Log(ctx, "notice", "Not enough memory available to execute the query");
+    return RedisModule_ReplyWithError(ctx, QueryError_Strerror(QUERY_EOOM));
+  }
+
   // Coord callback
   void (*dist_callback)(void *) = DistSearchCommandHandler;
 
@@ -3605,15 +3628,6 @@ int DistSearchCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
     argv++;
     argc--;
     dist_callback = DEBUG_DistSearchCommandHandler;
-  }
-
-  // Check OOM if OOM policy is not ignore
-  if (RSGlobalConfig.requestConfigParams.oomPolicy != OomPolicy_Ignore) {
-    // No need to hold the GIL since we are not in a background thread
-    if (estimateOOM(ctx)) {
-      RedisModule_Log(ctx, "notice", "Not enough memory available to execute the query");
-      return RedisModule_ReplyWithError(ctx, QueryError_Strerror(QUERY_EOOM));
-    }
   }
 
   // Prepare spec ref for the background thread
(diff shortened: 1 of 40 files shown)


src/search_disk.c
```diff
@@ -27,15 +31,50 @@ bool SearchDisk_Initialize(RedisModuleCtx *ctx) {
...
+void SearchDisk_SaveToRDB(RedisModuleIO *rdb, void *value) {
+  if (!disk) return;
+  disk->memObject.toRDB((RedisSearchDiskMemObject*)value, rdb);
+}
+
+int SearchDisk_AuxLoadFromRDB(RedisModuleIO *rdb, int encver, int when) {
+  if (!disk) return REDISMODULE_ERR;
+  disk_mem_obj = disk->memObject.fromRDB(rdb);
+  return REDISMODULE_OK;
+}
```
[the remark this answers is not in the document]
JoanFM: rdb_load and rdb_save are added more as a preparation than an intention to be used for now

src/search_disk.c
```diff
@@ -89,3 +128,22 @@ bool SearchDisk_IsEnabled(RedisModuleCtx *ctx) {
...
+
+int SearchDisk_RegisterType(RedisModuleCtx *ctx) {
+  RS_ASSERT(disk);
+  // rdb_load and aux_load are equivalent because disk_db is a singleton, so no multiple instances of this type are created
+  RedisModuleTypeMethods tm = {.version = REDISMODULE_TYPE_METHOD_VERSION,
+    .rdb_load = SearchDisk_LoadFromRDB,
+    .rdb_save = SearchDisk_SaveToRDB,
+    .aux_load = SearchDisk_AuxLoadFromRDB,
+    .aux_save2 = SearchDisk_AuxSaveToRDB,
+    .free = SearchDisk_Close,
```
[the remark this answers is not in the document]
JoanFM: We properly protect I think

src/module.c
```diff
@@ -1356,6 +1356,9 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx) {
 
   RM_TRY_F(RegisterLegacyTypes, ctx);
 
+  if (isFlex) {
+    RM_TRY_F(SearchDisk_RegisterType, ctx);
+  }
```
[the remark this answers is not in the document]
JoanFM: if Initialization fails, this will not happen, as it would have already failed. However, I will add it for completion

src/module.c
```diff
@@ -1356,6 +1356,9 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx) {
 
   RM_TRY_F(RegisterLegacyTypes, ctx);
 
+  if (isFlex && SearchDisk_Initialize(ctx)) {
```
raz-mon: We call `SearchDisk_Initialize` above already

src/search_disk.c
```diff
@@ -27,15 +35,55 @@ bool SearchDisk_Initialize(RedisModuleCtx *ctx) {
...
-  disk_db = disk->basic.open(ctx, "redisearch");
+  disk_db = disk->basic.open(ctx, "redisearch", disk_mem_obj); // Memory Object is empty by now
   return disk_db != NULL;
 }
 
+void SearchDisk_ReOpen(RedisModuleCtx *ctx) {
+  RS_ASSERT(disk);
+  if (disk_db) {
+    disk->basic.close(disk_db);
+    disk_db = NULL;
```
raz-mon, line 46: Unnecessary nullification

src/notifications.c
```diff
@@ -583,6 +584,9 @@ void RDB_LoadingEvent(RedisModuleCtx *ctx, RedisModuleEvent eid, uint64_t subeve
     break;
   case REDISMODULE_SUBEVENT_LOADING_ENDED:
     Indexes_EndRDBLoadingEvent(ctx);
+    if (isFlex) {
```
raz-mon, line 587: Why on the end instead of start event?

src/search_disk.c
```diff
@@ -27,15 +35,55 @@ bool SearchDisk_Initialize(RedisModuleCtx *ctx) {
...
   RedisModule_Log(ctx, "warning", "RediSearch disk API enabled");
 
-  disk_db = disk->basic.open(ctx, "redisearch");
+  disk_db = disk->basic.open(ctx, "redisearch", disk_mem_obj); // Memory Object is empty by now
   return disk_db != NULL;
 }
 
+void SearchDisk_ReOpen(RedisModuleCtx *ctx) {
+  RS_ASSERT(disk);
+  if (disk_db) {
```
raz-mon, line 44: Perhaps we need to assert that `!disk_db`

src/spec.c
```diff
@@ -2955,6 +2955,11 @@ void IndexSpec_RdbSave(RedisModuleIO *rdb, IndexSpec *sp) {
     SynonymMap_RdbSave(rdb, sp->smap);
   }
 
+  if (isSpecOnDisk(sp)) {
+    // The spec will not be built from keyspace, so we need to save the terms trie to enable rich queries
+    TrieType_GenericSave(rdb, sp->terms, 0);
```
raz-mon: 100% right, but we'll do this in phase 2, so for the MVP we won't need it.

raz-mon (commented):
Partial review - to be continued shortly.

src/module.c
JoanFM: yes but I feel like this the code is more clear and is a cheap check at start time. Also like this the code is not so dependant on some spefific order of operations  (Plus it silents cursor)

src/notifications.c
JoanFM: because then we are sure we have loaded the  disk mem object

src/search_disk.c
JoanFM: I think it would still make sense to reopen an already existing one

src/spec.c
JoanFM: okey

src/search_disk.c
JoanFM: I know it does not happen multiple times in the flow, but for the sake of completeness, I prefer to have this to avoid double closures and problems.

src/module.c
raz-mon: But we don't want to initialize twice, and under the same condition

JoanFM: ok I can remove this

src/search_disk.c
raz-mon: We set it in the next line.. Not sure I follow

src/search_disk.c
```diff
@@ -27,15 +35,55 @@ bool SearchDisk_Initialize(RedisModuleCtx *ctx) {
...
+  return disk_mem_obj;
+}
+
+void SearchDisk_SaveToRDB(RedisModuleIO *rdb, void *value) {
+  if (!disk) return;
+  disk->memObject.toRDB((RedisSearchDiskMemObject*)value, rdb);
+}
+
+int SearchDisk_AuxLoadFromRDB(RedisModuleIO *rdb, int encver, int when) {
+  if (!disk) return REDISMODULE_ERR;
```
raz-mon: Assert?

src/search_disk.c
JoanFM: Is just a pattern I feel more comfortable with. But can remove if needed

src/search_disk.c
```diff
@@ -89,3 +137,22 @@ bool SearchDisk_IsEnabled(RedisModuleCtx *ctx) {
   rm_free(isFlexStr);
   return isFlex;
 }
+
+int SearchDisk_RegisterType(RedisModuleCtx *ctx) {
+  RS_ASSERT(disk);
+  // rdb_load and aux_load are equivalent because disk_db is a singleton, so no multiple instances of this type are created
```
raz-mon, line 132: We don't create any keys of this type, so practically the `rdb_save/load` are not needed.
We can still keep them, but worth noting that we don't use them currently.

src/search_disk.c
raz-mon: So we create the db, close it, and create it again?

JoanFM: yes but most likely you would have created an empty one.

JoanFM:
> Wouldn't it be better to have the index disk-related data serialized and deserialized with it? Instead of serializing/deserializing the index data again. Is there a reason not to do that?

do you mean with the IndexSpec? I believe that since we do not yet have a direct link with it, this is the way to go. If we change that later then I would agree to also change this.

JoanFM:
> > Wouldn't it be better to have the index disk-related data serialized and deserialized with it? Instead of serializing/deserializing the index data again. Is there a reason not to do that?
> 
> do you mean with the IndexSpec? I believe that since we do not yet have a direct link with it, this is the way to go. If we change that later then I would agree to also change this.

Also I think that given the API as it is now where DB is its own object, it makes sense to associate a type to it. Like this an RDB can also more cleanly separate different concerns

src/search_disk.c
raz-mon: OK I don't mind that much 🤷‍♂️

raz-mon:
Since the index name is unique that can be the link between the index-spec and the disk counterpart.
Not critical just seems better to me, and more aligned with what we have today, a continuous section per-index.
Your call, other than that looks good to me

src/notifications.c
raz-mon: Can we not create the disk counterparts on the fly as we read from the rdb?

raz-mon: That's why we have `SearchDisk_AuxLoadFromRDB`
Something doesn't add up to me here

src/search_disk.c
JoanFM: done

src/search_disk.c
JoanFM: done

src/notifications.c
JoanFM: actually there will be another event we are waiting to have that will say that the SST transmission finished. But for now, at this time, everything is new, so we reset. Maybe we can have a softer reset API

PR #7171, opened 2025-10-28
[MOD-12192] Replace full II iterator by IndexReader

gdesmott:
#### Mark if applicable

---

> [!NOTE]
> Replaces usages of full inverted-index QueryIterator with IndexReader across debug commands, fork GC, and numeric index, adding a helper and updating iteration/decoding logic.
> 
> - **Debug commands**:
>   - Replace `NewInvIndIterator_*` iterations with `IndexReader` in `DumpInvertedIndex`, `DumpTagIndex`, `InfoTagIndex`, and inverted index numeric dumps.
>   - Add `ReplyReaderResultsIDs(reader, res, ctx)` helper to stream doc IDs from an `IndexReader`.
> - **Fork GC**:
>   - In `resetCardinality`, switch numeric index traversal from QueryIterator to `IndexReader` with `SkipTo`/`Next`.
> - **Numeric index**:
>   - Refactor median calculation: `NumericRange_GetMedian` now takes `IndexReader`/`RSIndexResult` and uses `IndexReader_*` APIs; `NumericRangeNode_Split` updated accordingly.
> - **General**:
>   - Remove legacy full inverted-index iterators in favor of `IndexReader` with appropriate `IndexDecoderCtx` initialization and result allocation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 362b2fdc4b5831bcea2a66831e9dda5975db87de. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/debug_commands.c
+++ b/src/debug_commands.c
@@ -600,8 +600,12 @@ DEBUG_COMMAND(DumpTagIndex) {
   while (TrieMapIterator_Next(iter, &tag, &len, (void **)&iv)) {
     RedisModule_ReplyWithArray(sctx->redisCtx, 2);
     RedisModule_ReplyWithStringBuffer(sctx->redisCtx, tag, len);
-    QueryIterator *iterator = NewInvIndIterator_TagFull(iv, tagIndex);
-    ReplyIteratorResultsIDs(iterator, sctx->redisCtx);
+    IndexDecoderCtx decoderCtx = {.field_mask_tag = IndexDecoderCtx_FieldMask, .field_mask = RS_FIELDMASK_ALL};
+    IndexReader *reader = NewIndexReader(iv, decoderCtx);
+    RSIndexResult *res = NewTokenRecord(NULL, 1);
+    res->freq = 1;
+    res->fieldMask = RS_FIELDMASK_ALL;
+    ReplyReaderResultsIDs(reader, res, sctx->redisCtx);
     ++resultSize;
   }
   RedisModule_ReplySetArrayLength(sctx->redisCtx, resultSize);
@@ -1167,8 +1171,12 @@ DEBUG_COMMAND(InfoTagIndex) {
 
     if (options.dumpIdEntries) {
       RedisModule_ReplyWithLiteral(ctx, "entries");
-      QueryIterator *reader = NewInvIndIterator_TagFull(iv, idx);
-      ReplyIteratorResultsIDs(reader, sctx->redisCtx);
+      IndexDecoderCtx decoderCtx = {.field_mask_tag = IndexDecoderCtx_FieldMask, .field_mask = RS_FIELDMASK_ALL};
+      IndexReader *reader = NewIndexReader(iv, decoderCtx);
+      RSIndexResult *res = NewTokenRecord(NULL, 1);
+      res->freq = 1;
+      res->fieldMask = RS_FIELDMASK_ALL;
+      ReplyReaderResultsIDs(reader, res, sctx->redisCtx);
     }
 
     RedisModule_ReplySetArrayLength(ctx, nsubelem);


src/debug_commands.c
```diff
@@ -230,8 +242,12 @@ DEBUG_COMMAND(DumpInvertedIndex) {
     RedisModule_ReplyWithError(sctx->redisCtx, "Can not find the inverted index");
     goto end;
   }
-  QueryIterator *iter = NewInvIndIterator_TermFull(invidx);
-  ReplyIteratorResultsIDs(iter, sctx->redisCtx);
```
chesedo, line 234: Can the `ReplyIteratorResultsIDs` C definition now be removed?

chesedo (commented):
Can all the `*Full` definitions also be removed with this PR?

src/fork_gc.c
```diff
@@ -526,17 +526,25 @@ static void resetCardinality(NumGcInfo *info, NumericRange *range, size_t blocks
...
+  bool reading = IndexReader_SkipTo(reader, startId) && IndexReader_Next(reader, res);
 
   // Continue reading the rest
-  while (status == ITERATOR_OK) {
-    double value = IndexResult_NumValue(iter->current);
+  while (reading) {
+    if (res->docId < startId)
+      // Skip to the starting ID
+      // TODO: remove when switching to IndexReader_Seek
+      continue;
```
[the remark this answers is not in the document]
chesedo: Yeah, a next seems to be missing before this `continue`

src/debug_commands.c
gdesmott: Not, it's still used in `DumpNumericIndex` which uses `NewInvIndIterator_NumericQuery`.

src/fork_gc.c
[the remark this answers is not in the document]
gdesmott: Fixed.

gdesmott:
> Can all the `*Full` definitions also be removed with this PR?

Not yet because it's still used in a lot of C/C++ tests and benchmarks.

Not sure if it's worth porting all those to the reader or just wait until we delete them once the iterator Rust port is done.

I'm also not sure how to handle https://github.com/RediSearch/RediSearch/blob/master/tests/cpptests/benchmark_vecsim_hybrid_queries.cpp#L75 for example where it uses an union iterator with full iterators as children. Could we use a query iterator with a wide range instead?

@GuyAv46 : what do you think?

src/fork_gc.c
```diff
@@ -526,17 +526,25 @@ static void resetCardinality(NumGcInfo *info, NumericRange *range, size_t blocks
...
   t_docId startId = IndexBlock_FirstId(startBlock);
-  QueryIterator *iter = NewInvIndIterator_NumericFull(range->entries);
-  // Skip to the starting ID
-  IteratorStatus status = iter->SkipTo(iter, startId);
+  IndexDecoderCtx decoderCtx = {.tag = IndexDecoderCtx_None};
+  IndexReader *reader = NewIndexReader(range->entries, decoderCtx);
+  RSIndexResult *res = NewNumericResult();
+  // TODO: use IndexReader_Seek once we switched to the Rust implementation
+  // as the new version has a seek implementation for all decoders.
+  bool reading = IndexReader_SkipTo(reader, startId) && IndexReader_Next(reader, res);
```
GuyAv46: Why do we need `&& IndexReader_Next(reader, res)`? We should handle a single result if there is one, no? Or am I missing something in the API

src/fork_gc.c
```diff
@@ -526,17 +526,25 @@ static void resetCardinality(NumGcInfo *info, NumericRange *range, size_t blocks
...
 
   // Continue reading the rest
-  while (status == ITERATOR_OK) {
-    double value = IndexResult_NumValue(iter->current);
-    hll_add(&range->hll, &value, sizeof(value));
-    status = iter->Read(iter);
+  while (reading) {
+    // Skip to the starting ID
+    // TODO: remove when switching to IndexReader_Seek
+    if (res->docId >= startId) {
```
GuyAv46: Why do we need this check if we skipped to `startId` before the loop?

src/numeric_index.c
```diff
@@ -79,20 +79,21 @@ static size_t NumericRange_Add(NumericRange *n, t_docId docId, double value) {
  * but if we see performance issues in the future, we can consider using another algorithm
  * like QuickSelect or an approximation algorithm for the median.
  */
-static double NumericRange_GetMedian(QueryIterator *iter, size_t num_entries) {
+static double NumericRange_GetMedian(IndexReader *reader, size_t num_entries) {
   size_t median_idx = num_entries / 2;
   double_heap_t *low_half = double_heap_new(median_idx);
+  RSIndexResult *res = NewNumericResult();
```
GuyAv46: We already allocated a result before calling this function. Consider using it

src/numeric_index.c
```diff
@@ -79,20 +79,21 @@ static size_t NumericRange_Add(NumericRange *n, t_docId docId, double value) {
  * but if we see performance issues in the future, we can consider using another algorithm
  * like QuickSelect or an approximation algorithm for the median.
  */
-static double NumericRange_GetMedian(QueryIterator *iter, size_t num_entries) {
+static double NumericRange_GetMedian(IndexReader *reader, size_t num_entries) {
   size_t median_idx = num_entries / 2;
```
GuyAv46: Is it possible to get the number of entries from the reader? It wasn't possible with a generic iterator, but a reader always has an index that it reads. I think it makes the API clearer
```suggestion
static double NumericRange_GetMedian(IndexReader *reader) {
  size_t median_idx = InvertedIndex_NumEntries(reader->idx) / 2;
```

gdesmott: Indeed, that's easier.

src/fork_gc.c
gdesmott: I'm using `IndexReader_SkipTo` which skips to the **block** that may contain the ID, but it may actually not seek to the record. For that we need `IndexReader_Seek` but I can't rely on it for now as seek is not implemented on all the C decoders.

@chesedo : or did I miss understood something here?

src/fork_gc.c
gdesmott: See my reply below, `IndexReader_SkipTo` skip to the block, not to the record.

src/fork_gc.c
GuyAv46: In this scenario, we need to iterate over the entire blocks starting at index `startIdx` (the block `startBlock`). We seek the first ID in the first block because that's the API, but we expect all the IDs in this and the following blocks to be `res->docId >= startId`

src/fork_gc.c
GuyAv46: Since we know there is a non-empty block with `startId`, I think we can avoid the second check.
Also, `bool reading` will always be `true` initially for the same reason, but for the sake of readability, we can keep it set to the `IndexReader_SkipTo` return value

gdesmott: Got it.
We still need to call `IndexReader_Next()` before entering the `while` loop, so what about this? I think it's clearer to init `reading` with the `Next` return value then.

GuyAv46 (approved):

src/fork_gc.c
GuyAv46: I agree. This is also what I did previously :)

PR #7200, opened 2025-10-30
Implement slots tracker for ASM - [MOD-12167]

GuyAv46:
## Implement slots state machine as part of the ASM effort

Implements a C API for a state machine, as described in the design document.

The implementation has 3 levels:
1. `SlotSet` - Implement set operations sor a set of slots, represented as an array of ranges, represented by the start and end slot of each range. This is to be memory efficient (under unfragmanted topology), and to seamlessly work with `RedisModuleSlotRangeArray`, which is expected to be supplied by the notifications we are going to use it in.
2. `SlotsTracker` - Implements the state machine as described in the design document, with functionalities according to the changes we want to track.
Internally, it uses 3 `SlotSet`s to track the different states of slots in the shard while the topology changes.
3. the C API - A wrapper around a static `SlotsTracker` instance. It gives an unsafe access to the state machine, which is safe under the assumption only the main thread calls these functions, except `slots_tracker_get_version` to atomically fetch the current version of the `SlotsTracker`, which can be called from any thread. The implementation is safe with `thread_local!` instanciation (so no thread can access the state machine managed by the main-thread), and the API remember the thread-ID that first called the API and panics if any other thread calls it, so we should never have more than a single state machine

1. `SlotSet` - track slots (values `[0-16383]`) using vector of continuous ranges, represented by the start and end slot of the range, and implements Set operations
3. `SlotsTracker` - Implements the state machine as described in the design document
4. The C API - wrapper around a single instance of `SlotsTracker` with minimal synchronization overhead

#### Mark if applicable

---

> [!NOTE]
> Introduces a Rust slots tracker and C FFI to manage cluster slot ranges with versioned state, plus an atomic version counter in result processing for future slot-based filtering.
> 
> - **Core (Rust)**:
>   - **New crate** `slots_tracker`: implements `SlotSet` (range-based set ops) and `SlotsTracker` (local/fully/partially-available sets, wraparound `Version`, availability checks).
>   - Extensive unit tests for set operations, versioning, and state transitions.
> - **FFI**:
>   - **New crate** `c_entrypoint/slots_tracker_ffi`: thread-local singleton tracker with owner-thread enforcement; exports C functions: `slots_tracker_set_local_slots_internal`, `slots_tracker_mark_partially_available_slots_internal`, `slots_tracker_promote_to_local_slots`, `slots_tracker_mark_fully_available_slots`, `slots_tracker_remove_deleted_slots`, `slots_tracker_has_fully_available_overlap`, `slots_tracker_check_availability`.
>   - **Generated header** `headers/slots_tracker.h` via `cbindgen` with inline wrappers managing an atomic version and `OptionSlotTrackerVersion`.
> - **Integration**:
>   - Adds `key_space_version` (`atomic_uint`) declaration/definition in `result_processor.c`.
>   - Updates workspace `Cargo.toml` and dependencies to include `slots_tracker` and FFI crate; build script for header generation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 88a4827e25759a7da3d583413a3e52d39dd2112b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/slots_tracker_ffi/src/lib.rs
+++ b/src/redisearch_rs/c_entrypoint/slots_tracker_ffi/src/lib.rs
@@ -194,8 +194,10 @@ pub unsafe extern "C" fn slots_tracker_set_local_slots_internal(
 /// This function updates the "partially available slots" set by adding the provided ranges.
 /// It also removes the given slots from "local slots" and "fully available slots", and
 /// increments the version counter.
+/// DO NOT call this function directly, use `slots_tracker_mark_partially_available_slots` in the C header instead.
 ///
-/// Returns the current version after the operation.
+/// Returns the current version after the operation, used by `slots_tracker_mark_partially_available_slots`
+/// in the C header for atomic version management.
 ///
 /// # Safety
 ///
@@ -323,7 +325,7 @@ pub unsafe extern "C" fn slots_tracker_has_fully_available_overlap(
 ///
 /// Return values (via OptionSlotTrackerVersion):
 /// - `is_some = false`: Required slots are not available. Query should be rejected.
-/// - `is_some = true`: Slots available; Store the returned `version` and compare it with the version returned by other functions to detect changes.
+/// - `is_some = true`: Slots available; Store the returned `version` and compare it (equality check) with the tracker's version.
 ///
 /// # Safety
 ///
--- a/src/redisearch_rs/headers/slots_tracker.h
+++ b/src/redisearch_rs/headers/slots_tracker.h
@@ -107,8 +107,10 @@ uint32_t slots_tracker_set_local_slots_internal(const RedisModuleSlotRangeArray 
  * This function updates the "partially available slots" set by adding the provided ranges.
  * It also removes the given slots from "local slots" and "fully available slots", and
  * increments the version counter.
+ * DO NOT call this function directly, use `slots_tracker_mark_partially_available_slots` in the C header instead.
  *
- * Returns the current version after the operation.
+ * Returns the current version after the operation, used by `slots_tracker_mark_partially_available_slots`
+ * in the C header for atomic version management.
  *
  * # Safety
  *
@@ -190,7 +192,7 @@ bool slots_tracker_has_fully_available_overlap(const RedisModuleSlotRangeArray *
  *
  * Return values (via OptionSlotTrackerVersion):
  * - `is_some = false`: Required slots are not available. Query should be rejected.
- * - `is_some = true`: Slots available; Store the returned `version` and compare it with the version returned by other functions to detect changes.
+ * - `is_some = true`: Slots available; Store the returned `version` and compare it (equality check) with the tracker's version.
  *
  * # Safety
  *


src/redisearch_rs/slots_tracker/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+    let current = STATE.version.load(Ordering::Relaxed);
+    debug_assert!(
+        current <= MAX_VALID_VERSION,
+        "Version counter out of valid range"
+    );
+    let next = if current < MAX_VALID_VERSION {
+        current + 1
+    } else {
+        0 // Wrap around to 0
+    };
```
JoanFM: should we simply use `current + 1` modulus MAX_VALID_VERSION?

src/redisearch_rs/slots_tracker/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+/// Increments the version counter, skipping reserved values.
+///
+/// If the current version is `MAX_VALID_VERSION` (u32::MAX - 2), wraps to 0
+/// instead of incrementing to reserved values (u32::MAX - 1 or u32::MAX).
+///
+/// # Safety
+///
+/// This function assumes single-threaded access (main thread only).
+fn increment_version() {
+    let current = STATE.version.load(Ordering::Relaxed);
```
JoanFM: Can we attach the method to the struct? Like can we do implement increment_version for State? and then use these entry to call the static one? This would help in the testing right?

src/redisearch_rs/slots_tracker/src/lib.rs
```diff
@@ -0,0 +1,445 @@
...
+/// # Panics
+///
+/// Panics if the pointer is null or num_ranges is less than 1.
+///
+/// # Safety
+///
+/// The caller must ensure the pointer is valid and points to a properly initialized
+/// SlotRangeArray with at least `num_ranges` elements in the flexible array.
+unsafe fn parse_slot_ranges(ranges: *const SlotRangeArray) -> &'static [SlotRange] {
+    assert!(!ranges.is_null(), "SlotRangeArray pointer is null");
```
JoanFM: why do we need to return a static lifetime?

src/redisearch_rs/slots_tracker/src/lib.rs
GuyAv46: We could, but I don't think it will make it more readable

src/redisearch_rs/slots_tracker/src/lib.rs
GuyAv46: Good idea. The C got the better of me... I implemented a state machine struct and made the C API a wrapper around a single static instance of it

src/redisearch_rs/slots_tracker/src/lib.rs
GuyAv46: We don't. Good catch!
Rust requires to specify a lifetime when you return a reference (slice in this case) to something with no clear lifetime (like to the array behind the pointer in this case). I wasn't sure how to do it so I made it static, but the right thing to do is to bound the lifetime to the caller, like `std::slice::from_raw_parts` does. Fixed.

src/redisearch_rs/slots_tracker/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+
+    // SAFETY: Caller guarantees valid pointer
+    let ranges = unsafe { &*ranges };
+
+    assert!(
+        ranges.num_ranges >= 1,
+        "num_ranges must be at least 1, got {}",
+        ranges.num_ranges
+    );
```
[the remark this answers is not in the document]
GuyAv46: Not sure if there is a valid scenario where we get an empty input, but since `std::slice::from_raw_parts` can handle 0 length (as long as the pointer is not null, which we assert), all we need to do is update this assertion to allow 0 ranges

src/redisearch_rs/slots_tracker/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+
+    assert!(
+        ranges.num_ranges >= 0,
+        "num_ranges must be at least 0, got {}",
+        ranges.num_ranges
+    );
+
+    // SAFETY: Caller guarantees the flexible array has num_ranges elements
+    unsafe { std::slice::from_raw_parts(ranges.ranges.as_ptr(), ranges.num_ranges as usize) }
+}
```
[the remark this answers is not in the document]
GuyAv46: It does ties it (unsafely) in the best way possible, given this is an FFI API

src/redisearch_rs/slots_tracker/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+/// If the ranges are identical to the current configuration, no changes are made.
+///
+/// # Safety
+///
+/// This function must be called from the main thread only.
+/// The `ranges` pointer must be valid and point to a properly initialized RedisModuleSlotRangeArray.
+/// The ranges array must contain `num_ranges` valid elements.
+/// All ranges must be sorted and have start <= end, with values in [0, 16383].
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn slots_tracker_set_local_slots(ranges: *const SlotRangeArray) {
```
chesedo: You'll want the C -> Rust API layer to be in the `c_entrypoint` directory. This means making another crate there with the API you want. You can just follow the structure of one of the existing crates in that directory. And will want the `cbindgen` to generate the C headers for you.

With that change, this crate should have no `unsafe` code and should be a pure Rust library.

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,781 @@
...
+    ///
+    /// **Assumes input is already sorted and normalized**.
+    /// In debug builds, validates this assumption.
+    /// Used by `slots_tracker_set_partially_available_slots` and `slots_tracker_set_fully_available_slots`.
+    pub(crate) fn add_ranges(&mut self, ranges: &[SlotRange]) {
+        debug_assert_valid_normalized_input(ranges);
+
+        // Simply add all ranges and normalize once at the end
+        self.ranges.extend_from_slice(ranges);
+        self.normalize();
```
JoanFM, line 90: what is the difference between the normalization done in the method w.r.t the normalization asserted at input level?

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,781 @@
...
+///
+/// **Invariant**: The ranges vector is always kept sorted and normalized:
+/// - Sorted by start slot in ascending order
+/// - No overlapping ranges
+/// - No adjacent ranges (they are merged)
+/// - All ranges are valid (start <= end, values in [0, 16383])
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub(crate) struct SlotSet {
+    /// Vector of slot ranges (start, end), kept sorted and normalized.
+    ranges: Vec<SlotRange>,
```
JoanFM, line 61: isn't there a need to specify mutability?

JoanFM: ah no, it seems in the methods you do

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,781 @@
...
+
+            // Apply all overlapping remove ranges to current
+            while let Some(&&remove) = remove_iter.peek() {
+                if remove.start > current.end {
+                    break; // No more overlaps for current
+                }
+
+                // Handle overlap cases
+                match (remove.start <= current.start, remove.end >= current.end) {
+                    (true, true) => {
```
JoanFM, line 114: can't we implement some comparison API for SlotRange with a more meaningful syntax? so that we do not rely on the comments, but that some comparison results make automatic sense

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,781 @@
...
+}
+
+impl PartialEq<SlotSet> for &[SlotRange] {
+    fn eq(&self, other: &SlotSet) -> bool {
+        *self == other.ranges
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
JoanFM, line 308: is there a way to parametrize these tests? so that we can run the same test with different inputs (like contiguous ranges, previously normalized or not, etc ...?)

JoanFM (changes requested):

src/redisearch_rs/slots_tracker/src/slot_set.rs
GuyAv46: A member mutability is as the entire struct mutability, unless you use some kind of [`Cell`](https://doc.rust-lang.org/std/cell/struct.Cell.html)

src/redisearch_rs/slots_tracker/src/slot_set.rs
GuyAv46: I think there are external crates (like mentioned [here](https://stackoverflow.com/questions/34662713/how-can-i-create-parameterized-tests-in-rust)), but I'd prefer to avoid using it. Also, most test has a different outcome for each input

src/redisearch_rs/slots_tracker/src/slot_set.rs
GuyAv46: The assert only verifies the input is normalized (in debug mode), while `self.normalize()` normalize `self` if it's not in normal form. Checking if a slot ranges array is normal is simple, while normalizing is a bit more complex.

In this case, we assert the input is valid, and then we append `ranges` (which may invalidate the normal form), and then normalize self again.
There are all kind of edge cases when adding ranges - partial coverage, full coverage, bridging gaps between our ranges and so on. We handle then in `normalize`. We could reuse it in other functionalities in the future.

src/redisearch_rs/slots_tracker/src/slot_set.rs
GuyAv46: I guess we can, but note that in this case we only handle some of the possible relations between ranges, as we have some knowledge here (there is some overlap), and the 4 cases here are exactly what we need to handle in this case

JoanFM: you can still cover the others and do nothing on those cases. It would help readbility I think but may be an overkill

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,1040 @@
...
+
+impl SlotSet {
+    /// Creates a new empty `SlotSet`.
+    pub(crate) const fn new() -> Self {
+        Self { ranges: Vec::new() }
+    }
+
+    // ========================================================================
+    // Public API methods required for the C FFI:
+    // ========================================================================
```
chesedo: This crate should have no knowledge of the FFI

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,1040 @@
...
+fn debug_assert_normalized_ranges(ranges: &[SlotRange]) {
+    debug_assert!(ranges.windows(2).all(|w| w[0].end + 1 < w[1].start));
+}
+
+/// Validates that ranges are valid, sorted, and normalized.
+#[inline]
+fn debug_assert_valid_normalized_input(ranges: &[SlotRange]) {
+    debug_assert_valid_ranges(ranges);
+    debug_assert_normalized_ranges(ranges);
+}
```
chesedo, line 46: Why have this? It feels like rather repeating the code at the call sites will make it clearer what is being asserted

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,1040 @@
...
+
+        // Input is already normalized, just replace our vector
+        self.ranges = ranges.to_vec();
+    }
+
+    /// Adds/merges ranges into the set (union operation).
+    ///
+    /// **Assumes input is already sorted and normalized**.
+    /// In debug builds, validates this assumption.
+    /// Used by `slots_tracker_set_partially_available_slots` and `slots_tracker_set_fully_available_slots`.
```
chesedo:
```suggestion
    /// Adds/merges ranges into the set (union operation).
    ///
    /// The input is required to be sorted and normalized.
```
There are two changes here that can apply to all functions in this crate:
1. The caller does not need to know what happens in debug builds. Instead the caller should assume these assertions are always validated.
2. This code should function as an isolated library and should not know what is calling it where.

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,1040 @@
...
+    /// **Assumes input is already sorted and normalized**.
+    /// In debug builds, validates this assumption.
+    /// May split existing ranges or remove them entirely.
+    ///
+    /// **Optimized**: Takes advantage of sorted input to avoid re-scanning from the beginning.
+    pub(crate) fn remove_ranges(&mut self, ranges: &[SlotRange]) {
+        debug_assert_valid_normalized_input(ranges);
+
+        let old_ranges = std::mem::take(&mut self.ranges);
+        let mut remove_iter = ranges.iter().peekable();
```
chesedo, line 102:
```suggestion
        let mut remove_iter = ranges.into_iter().peekable();
```
This should allow you to not need the extra borrow everywhere.

chesedo: This only works because you need to consume the `ranges` once

src/redisearch_rs/slots_tracker/src/slot_set.rs
chesedo: You can look at [`PartialOrd`](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html)

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,1040 @@
...
+        debug_assert_valid_normalized_input(ranges);
+
+        let mut our_iter = self.ranges.iter();
+        let mut their_iter = ranges.iter().peekable();
+
+        for &our_range in our_iter.by_ref() {
+            // Skip their ranges that end before our current range starts
+            while their_iter.peek().is_some_and(|&&r| r.end < our_range.start) {
+                their_iter.next();
+            }
```
chesedo: These feel like a [`skip_while`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip_while) operation

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,1040 @@
...
+
+            // Check if next their range overlaps with our current range
+            if let Some(&&their_range) = their_iter.peek() {
+                if their_range.start <= our_range.end {
+                    return true; // Overlap found
+                }
+            } else {
+                break; // No more their ranges to check
+            }
```
chesedo: This feels like an [`any`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any) operation

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,1040 @@
...
+        }
+
+        false
+    }
+
+    /// Returns true if this set contains no ranges.
+    pub(crate) const fn is_empty(&self) -> bool {
+        self.ranges.is_empty()
+    }
```
chesedo, line 171: Why `pub(crate)`? Can't the `pub` be removed entirely?

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,1040 @@
...
+            // Find the first of our ranges that does not end before their range starts
+            let current_our = loop {
+                match our_iter.peek() {
+                    Some(&current) if current.end < their_range.start => {
+                        has_extra = true; // Skipped an entire our range
+                        our_iter.next();
+                    }
+                    Some(&current) => break current,
+                    None => return CoverageRelation::NoMatch,
+                }
```
chesedo: This is a [`find`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find) operation

src/redisearch_rs/slots_tracker/src/slot_set.rs
```diff
@@ -0,0 +1,1040 @@
...
+    ///
+    /// **Assumes input is already sorted and normalized** (no overlaps, no adjacent ranges).
+    /// In debug builds, validates this assumption.
+    /// Used only by `slots_tracker_set_local_slots`.
+    pub(crate) fn set_from_ranges(&mut self, ranges: &[SlotRange]) {
+        debug_assert_valid_normalized_input(ranges);
+
+        // Input is already normalized, just replace our vector
+        self.ranges = ranges.to_vec();
+    }
```
chesedo:
```suggestion
    pub(crate) fn from_ranges(ranges: &[SlotRange]) -> Self {
        debug_assert_valid_normalized_input(ranges);

        // Input is already normalized, just replace our vector
        Self { ranges:ranges.to_vec() }
    }
```

This overrides everything so can just be a constructor instead

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
```diff
@@ -0,0 +1,701 @@
...
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! Safe implementation of the slots tracker state.
+//!
+//! This module provides a safe, non-thread-safe implementation of the slots tracker.
+//! All methods take `&mut self`, making the borrowing rules enforce single-threaded access.
```
chesedo:
```suggestion
//! This module provides a single-thread-safe implementation of the slots tracker.
//! All methods take `&mut self`, making the borrowing rules enforce single-threaded access.
```

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
```diff
@@ -0,0 +1,701 @@
...
+    /// Sets the partially available slot ranges.
+    ///
+    /// This function updates the `partially_available` set by adding the provided ranges.
+    /// It also removes the given slots from `local` and `fully_available`, and
+    /// increments the `version` counter (fully available or not available slots are now partially available).
+    ///
+    /// # Arguments
+    ///
+    /// * `ranges` - Slice of slot ranges. Must be sorted and normalized (no overlaps, no adjacent ranges).
+    pub fn set_partially_available_slots(&mut self, ranges: &[SlotRange]) {
```
chesedo:
```suggestion
    /// Sets the partially available slot ranges.
    ///
    /// This function updates the `partially_available` set by adding the provided ranges.
    /// It also removes the given slots from `local` and `fully_available`, and
    /// increments the `version` counter (fully available or not available slots are now partially available).
    ///
    /// # Arguments
    ///
    /// * `ranges` - Slice of slot ranges. Must be sorted and normalized (no overlaps, no adjacent ranges).
    pub fn add_partially_available_slots(&mut self, ranges: &[SlotRange]) {
```
The other `set_*` function does a replace so using add instead avoids a confusion.

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
```diff
@@ -0,0 +1,701 @@
...
+    /// (the version was already bumped when slots became partially available, and while partially
+    /// available slots exist, `check_availability` returns unstable/unavailable anyway).
+    ///
+    /// # Arguments
+    ///
+    /// * `ranges` - Slice of slot ranges. Must be sorted and normalized (no overlaps, no adjacent ranges).
+    pub fn promote_to_local_slots(&mut self, ranges: &[SlotRange]) {
+        self.local.add_ranges(ranges);
+        self.partially_available.remove_ranges(ranges);
+    }
```
chesedo, line 186: Should this have a debug assert that `partially_available` fully contains `ranges`? To prevent ranges from being added that were not `partially_available`.

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
```diff
@@ -0,0 +1,701 @@
...
+    /// Sets the fully available non-owned slot ranges.
+    ///
+    /// This function updates the `fully_available` set by adding the provided ranges.
+    /// It removes the given slots from `local`.
+    /// Version is NOT incremented by this operation (slots availability is unchanged).
+    ///
+    /// # Arguments
+    ///
+    /// * `ranges` - Slice of slot ranges. Must be sorted and normalized (no overlaps, no adjacent ranges).
+    pub fn set_fully_available_slots(&mut self, ranges: &[SlotRange]) {
```
chesedo:
```suggestion
    /// Sets the fully available non-owned slot ranges.
    ///
    /// This function updates the `fully_available` set by adding the provided ranges.
    /// It removes the given slots from `local`.
    /// Version is NOT incremented by this operation (slots availability is unchanged).
    ///
    /// # Arguments
    ///
    /// * `ranges` - Slice of slot ranges. Must be sorted and normalized (no overlaps, no adjacent ranges).
    pub fn add_fully_available_slots(&mut self, ranges: &[SlotRange]) {
```

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
```diff
@@ -0,0 +1,701 @@
...
+    /// # Arguments
+    ///
+    /// * `ranges` - Slice of slot ranges to check. Must be sorted and normalized.
+    ///
+    /// # Returns
+    ///
+    /// - Current version (0..u32::MAX-2): Slots match exactly and are stable
+    /// - `SLOTS_TRACKER_UNAVAILABLE` (u32::MAX): Required slots are not available
+    /// - `SLOTS_TRACKER_UNSTABLE_VERSION` (u32::MAX-1): Slots available but partial/inexact match
+    pub fn check_availability(&self, ranges: &[SlotRange]) -> u32 {
```
chesedo: Why not return an `enum` instead?

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
```diff
@@ -0,0 +1,701 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn slot_ranges_slice(ranges: &[(u16, u16)]) -> &[SlotRange] {
+        unsafe { std::slice::from_raw_parts(ranges.as_ptr() as *const SlotRange, ranges.len()) }
+    }
```
chesedo: Rather implement the `From<(u16, u16)>` trait

chesedo: Using a `From<Range>` might also improve the DX

[Range](https://doc.rust-lang.org/std/ops/struct.Range.html)

src/redisearch_rs/slots_tracker/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+///
+/// The version field is atomic and can be safely read from any thread.
+struct SlotsTrackerState {
+    /// The safe slots tracker implementation wrapped in UnsafeCell for static access.
+    tracker: UnsafeCell<SlotsTracker>,
+    /// Atomic version counter that mirrors the tracker's internal version.
+    /// This allows thread-safe reads of the version without accessing the tracker.
+    version: AtomicU32,
+}
+
```
chesedo: Why is the wrapper needed? It seems to be possible to expose/use the `SlotTracker` directly

src/redisearch_rs/slots_tracker/src/slot_set.rs
GuyAv46: I don't like reapiting code, and I find `debug_assert_valid_normalized_input` clearer than `ranges.windows(2).all(|w| w[0].end + 1 < w[1].start)`. Also if we ever change the definition of normal we could fix the assertion once.
But if it's more idiomatic I'll put the asserts directly in each function

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/cbindgen.toml
```diff
@@ -0,0 +1,21 @@
...
+parse_deps = true
+include = ["slots_tracker_ffi"]
+
+[export]
+# Export the slots tracker types and constants
+include = ["SLOTS_TRACKER_UNAVAILABLE"]
+
+[export.rename]
+"SlotRange" = "RedisModuleSlotRange"
+"SlotRangeArray" = "RedisModuleSlotRangeArray"
```
chesedo, line 68: This seems perfect while redismodule-rs is missing the type :partying_face:

GuyAv46: I still had to define `SlotRange` and `SlotRangeArray` in the `slots_tracker` module and not in the FFI module, so it's available internally (`SlotRange` specifically) and that `RedisModuleSlotRange` and `RedisModuleSlotRangeArray` are not re-defined in the `slots_tracker.h` header.
Is there a better way to do it?

src/redisearch_rs/slots_tracker/src/slot_set.rs
GuyAv46: Maybe I'm missing something, but I don't manage to use `skip_while` with a peelable iterator

src/redisearch_rs/slots_tracker/src/slot_set.rs
GuyAv46: GT, LT, and EQ are not enough for ranges, and 3x3 for all the combinations of each end of the range, and this is without considering the special cases of adjacent ranges, feels like an overkill when we have only 4 distinct cases to handle here

src/redisearch_rs/slots_tracker/src/slot_set.rs
GuyAv46: We use it in `slots_tracker` to avoid heavy computations in the common cases, so no

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
GuyAv46: I agree, but `add_*` is also not exactly what we do because it may be used to demote slots from fully available or local to partially available.
How about `mark_*`?

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
GuyAv46: OK. Under the expected usage this is correct.
I wonder if there are some edge cases were this will be incorrect. I intentionally made the state machine "error tolerable" because the topology may change suddenly from `set_local_slots` if non-ASM migration is used, or some other unexpected server event (replica promoted to master due to server crash for example). But in our test suite it should be correct

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
GuyAv46: We have a versioning mechanism, so normally we would want to return the current version `u32`. In theory, if migration events happen right after another, we may have multiple valid version in parallel until all the trimming is done.

Internally, we have 2 special values - `SLOTS_TRACKER_UNAVAILABLE` and `SLOTS_TRACKER_UNSTABLE_VERSION`. We guarantee (internally) that the tracker version is never equal to these values (so comparing them to `get_version` result is always false).
Externally, `SLOTS_TRACKER_UNSTABLE_VERSION` is not a special value, but if we get `SLOTS_TRACKER_UNAVAILABLE` back, we should reject the query. We can refactor this function to return `Option<u32>` or a custom enum, but I think we should wait until we don't call it from C code, as it adds complexity for no good reason (unless it's simple and I'm overreacting)

src/redisearch_rs/slots_tracker/src/lib.rs
GuyAv46: For two reasons:
1. To implement `Sync` for the static instance, with no overhead (as only the main thread is expected to interact with the `SlotsTracker`)
2. To group the atomic counter that mirrors the internal version with the tracker instance

src/redisearch_rs/slots_tracker/src/slot_set.rs
chesedo: Fair, if you find it reads easier :+1:

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
chesedo: `mark_*` sounds good

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
chesedo: If `partially_available` not containing all the ranges is considered acceptable then the assert is not needed. I just assumed that is a precondition based on the name of the variables (ie it needs to be partially available first before being promoted)

src/redisearch_rs/slots_tracker/src/slot_set.rs
GuyAv46: I still have issues with `peekable`. I tried
```
            // If we are about to skip some of our ranges, turn on has_extra
            has_extra |= our_iter.peek().is_some_and(|&r| r.end < their_range.start);

            // Find the first of our ranges that does not end before their range starts
            let Some(current_our) = our_iter.by_ref().find(|&r| r.end >= their_range.start) else {
                return CoverageRelation::NoMatch;
            };
```
But `find` seems to advance the iterator even if the current element is what we look for

src/redisearch_rs/slots_tracker/src/slot_set.rs
GuyAv46: `Clippy` refuses, as `ranges` is a slice and we don't consume it anyways. It warns that `iter` and `into_iter` are equivalent in this case. I left it as `iter`

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
chesedo: I was thinking of this kind of enum

```rust
enum Availability {
    Version(u32),
    Unstable = u32::MAX - 1,
    Unavailable = u32::MAX,
}
```

Then it can encapsulate the wrapping logic.

src/redisearch_rs/slots_tracker/src/lib.rs
chesedo: Why do you need `Sync` if only one thread is accessing it anyway?

For context, I want to use the `SlotsTracker` directly to get rid of the mirrored version here

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/cbindgen.toml
chesedo: No, that sounds correct and how I would have done it too

src/redisearch_rs/slots_tracker/src/slot_set.rs
chesedo: Hmmm, `find` should stop at the found element. With my other suggestions you might not need peekable at all

src/redisearch_rs/slots_tracker/src/slot_set.rs
chesedo: Can you remove the peekable by only using the `Iter` interface?

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
GuyAv46: I will add the debug assert, but definitely not as `assert!`

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
GuyAv46: I implemented `From<(u16, u16)>`. For testing perpuses, I think seeing the numbers is the clearest way to understand the tests in this case

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
chesedo: Haha, yes a debug only one please :smile:

src/redisearch_rs/slots_tracker/src/slot_set.rs
GuyAv46: It's not so clean to give up the `peekable`, as we iterate with nested loops over two iterators in parallel. I found `next_if` of `Peekable` that has a documented example for a similar situation:
```
while their_iter.next_if(|&&r| r.end < our_range.start).is_some() {}
```

src/redisearch_rs/slots_tracker/src/slot_set.rs
GuyAv46: end up using `while their_iter.next_if(|&&r| r.end < our_range.start).is_some()`

src/redisearch_rs/slots_tracker/src/slot_set.rs
JoanFM: Doesn't Rust provide a Set type?

GuyAv46: I’m sure there is one, but it won’t be efficient in our scenario. If I had to choose an alternative approach I’d use a bitmap before choosing a set

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
JoanFM: this would make sense

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/src/lib.rs
```diff
@@ -0,0 +1,335 @@
...
+///
+/// This allows thread-safe reads of the version from any thread.
+static VERSION: AtomicU32 = AtomicU32::new(0);
+
+// Thread-local slots tracker instance.
+//
+// Only accessible from the thread that first calls any FFI function.
+thread_local! {
+    static TRACKER: RefCell<SlotsTracker> = const { RefCell::new(SlotsTracker::new()) };
+}
```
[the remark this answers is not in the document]
GuyAv46: Only the main thread should call this API, and in Redis it never terminates while the server is up

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/cbindgen.toml
```diff
@@ -0,0 +1,16 @@
+includes = ["redismodule.h"]
+language = "C"
+autogen_warning = "/* Warning, this file is autogenerated by cbindgen from `src/redisearch_rs/c_entrypoint/slots_tracker_ffi/build.rs. Don't modify it manually. */"
+pragma_once = true
+
+[parse]
+parse_deps = true
+include = ["slots_tracker_ffi"]
```
chesedo:
```suggestion
[parse]
parse_deps = true
```
This should not be needed since this is `slots_tracker_ffi`. You would only add external dependencies here

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+/// # Panics
+///
+/// Panics if the pointer is null or num_ranges is less than 0.
+///
+/// # Safety
+///
+/// The caller must ensure the pointer is valid and points to a properly initialized
+/// SlotRangeArray with at least `num_ranges` elements in the flexible array.
+unsafe fn parse_slot_ranges<'a>(ranges: *const SlotRangeArray) -> &'a [SlotRange] {
+    assert!(!ranges.is_null(), "SlotRangeArray pointer is null");
```
chesedo: Currently all the FFI layers has these as `debug_assert!`

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+
+    // SAFETY: Caller guarantees the flexible array has num_ranges elements
+    unsafe { std::slice::from_raw_parts(ranges.ranges.as_ptr(), ranges.num_ranges as usize) }
+}
+
+// ============================================================================
+// Main API Functions
+// ============================================================================
+
+/// Sets the local responsibility slot ranges.
```
chesedo:
```suggestion
/// Sets the local responsible slot ranges.
```

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+/// # Safety
+///
+/// This function must be called from the main thread only.
+/// The `ranges` pointer must be valid and point to a properly initialized RedisModuleSlotRangeArray.
+/// The ranges array must contain `num_ranges` valid elements.
+/// All ranges must be sorted and have start <= end, with values in [0, 16383].
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn slots_tracker_check_availability(
+    ranges: *const SlotRangeArray,
+) -> OptionSlotTrackerVersion {
```
chesedo, line 339: The comment just needs updating

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/cbindgen.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[export]
+# Export the slots tracker types and constants
+include = ["SLOTS_TRACKER_UNAVAILABLE"]
```
chesedo: This no longer exists right?

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+                is_some: true,
+                value: 0,
+            },
+            None => Self {
+                is_some: false,
+                value: 0,
+            },
+        }
+    }
+}
```
chesedo, line 59: Nice one :star_struck:

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+use std::cell::RefCell;
+use std::sync::OnceLock;
+use std::sync::atomic::{AtomicU32, Ordering::Relaxed};
+use std::thread::ThreadId;
+
+#[repr(C)]
+pub struct OptionSlotTrackerVersion {
+    is_some: bool,
+    value: u32,
+}
```
chesedo, line 39: You may want to comment all the field combinations on this.

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
```diff
@@ -0,0 +1,931 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Version {
+    /// Stable version values: 1..=u32::MAX
+    /// Each value represents a specific version counter.
+    Stable(NonZeroU32),
+    /// Unstable state marker
+    /// Slots are available but configuration is changing.
+    Unstable,
+}
+const _: () = assert!(std::mem::size_of::<Version>() == 4);
```
chesedo: Add a comment why this exists

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
```diff
@@ -0,0 +1,931 @@
...
+        match self {
+            Self::Stable(v) => {
+                if v.get() < u32::MAX {
+                    // SAFETY: 1 < v + 1 <= u32::MAX, so not zero
+                    Self::Stable(unsafe { NonZeroU32::new_unchecked(v.get() + 1) })
+                } else {
+                    // Wrap around to 1 from u32::MAX
+                    // SAFETY: value is not zero
+                    Self::Stable(unsafe { NonZeroU32::new_unchecked(1) })
+                }
```
chesedo, line 66: SInce you are doing all this checking anyway, why not use a normal `u32` for the `Stable` variant? Mostly to remove the need for this `unsafe` code

chesedo (commented):
Just minor comments. This is looking nice!

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+/// The thread ID that owns the tracker instance.
+///
+/// Set once when the first FFI function is called. All subsequent calls
+/// must come from the same thread.
+static OWNER_THREAD: OnceLock<ThreadId> = OnceLock::new();
+
+/// Atomic version counter that mirrors the tracker's internal version.
+///
+/// This allows thread-safe reads of the version from any thread.
+static VERSION: AtomicU32 = AtomicU32::new(0);
```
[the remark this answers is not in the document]
GuyAv46: The initial value doesn't matter before the state is initialized with some actual topology, so we can treat it as the uninitialized state of `VERSION`, which later shouldn't be 0 anymore (set from a NonZeroU32)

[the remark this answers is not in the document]
GuyAv46: The initial value doesn't matter before the state is initialized with some actual topology, so we can treat it as the uninitialized state of VERSION, which later shouldn't be 0 anymore (set from a NonZeroU32)

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
GuyAv46: It makes the `Version` size 64bit.
I can avoid the unsafe new with `NonZeroU32::new(1).unwrap()`.
Or make Stable use u16

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/src/lib.rs
```diff
@@ -0,0 +1,363 @@
...
```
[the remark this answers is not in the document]
JoanFM: @GuyAv46 to check

src/redisearch_rs/slots_tracker/src/slot_set.rs
JoanFM: why wouldn't it be? because of the search algorithm in sets? Isn't there a way to use a set and updating its methods?

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
JoanFM: But is this complication worth for 32 bits? How many slot tracker instance will we have? One right?

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
```diff
@@ -0,0 +1,932 @@
...
+/// for manipulating the three slot sets and version counter.
+///
+/// All methods take `&mut self`, ensuring compile-time enforcement of exclusive access.
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct SlotsTracker {
+    /// Local responsibility slots - owned by this Redis instance in the cluster topology.
+    local: SlotSet,
+    /// Fully available non-owned slots - locally available but not owned by this instance.
+    fully_available: SlotSet,
+    /// Partially available non-owned slots - partially available and not owned by this instance.
```
JoanFM, line 86: clarify partially available meaning here?

It means that not all keys of the range slot are in this instance?

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
```diff
@@ -0,0 +1,932 @@
...
+    /// It also removes the given slots from `fully_available` and `partially_available`.
+    /// If the new ranges are identical to the current local slots, no change occurs and version
+    /// is not incremented.
+    ///
+    /// # Arguments
+    ///
+    /// * `ranges` - Slice of slot ranges. Must be sorted and normalized (no overlaps, no adjacent ranges).
+    pub fn set_local_slots(&mut self, ranges: &[SlotRange]) {
+        // Check if the ranges are already equal (no change needed)
+        if self.local == ranges {
```
JoanFM, line 140: Are we sure if self.local=ranges, we do not still need to remove from other sets and increment version?

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
```diff
@@ -0,0 +1,932 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct SlotsTracker {
+    /// Local responsibility slots - owned by this Redis instance in the cluster topology.
+    local: SlotSet,
+    /// Fully available non-owned slots - locally available but not owned by this instance.
+    fully_available: SlotSet,
+    /// Partially available non-owned slots - partially available and not owned by this instance.
+    partially_available: SlotSet,
+    /// Version counter for tracking changes to the slots configuration.
+    /// Incremented whenever the slot configuration changes. Wraps around safely using wrapping arithmetic.
```
JoanFM, line 90: Can we clarify what events increment it, as seen in the API not all of them increment it. Can we make this logic exhaustive.

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
```diff
@@ -0,0 +1,932 @@
...
+    /// This function adds the provided ranges to `local` and removes them from `partially_available`.
+    /// Does NOT modify `fully_available` and does NOT increment the version counter
+    /// (the version was already bumped when slots became partially available, and while partially
+    /// available slots exist, `check_availability` returns unstable/unavailable anyway).
+    ///
+    /// # Arguments
+    ///
+    /// * `ranges` - Slice of slot ranges. Must be sorted and normalized (no overlaps, no adjacent ranges).
+    pub fn promote_to_local_slots(&mut self, ranges: &[SlotRange]) {
+        debug_assert!(matches!(
```
JoanFM, line 180: can we assert then that fully_available does not contain them? This would give more certainty that the dependency call (version counter increase) is satisfied.

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
JoanFM: maybe with the API naming. Or clarify which of the sets is considered for the state to be in another version

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
GuyAv46: One instance, but each query (and cursor) will store a version as well.
It’s also less about the memory footprint and more about the atomic read we will do for each result

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
GuyAv46: Yes, as far as the module is concerned. In practice it means we cannot rely on these slots for queries, but also we cannot ignore their existence

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
GuyAv46: The tracker’s sets should never overlap. This is why we attempt to remove the ranges from the other sets when we do detect a change by this API.

src/redisearch_rs/slots_tracker/src/slots_tracker.rs
GuyAv46: > Can we make this logic exhaustive

WDYM?

JoanFM: I mean, who and how and on which considerations the version changes

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/src/lib.rs
```diff
@@ -0,0 +1,342 @@
...
+    fn from(version: Option<Version>) -> Self {
+        match version {
+            Some(Version::Stable(v)) => Self {
+                is_some: true,
+                version: v.get(),
+            },
+            Some(Version::Unstable) => Self {
+                is_some: true,
+                version: 0,
+            },
```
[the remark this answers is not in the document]
GuyAv46: a stabe version is non-zero

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/cbindgen.toml
```diff
@@ -0,0 +1,68 @@
...
+ * The ranges array must contain `num_ranges` valid elements.
+ * All ranges must be sorted and have start <= end, with values in [0, 16383].
+ */
+static void slots_tracker_set_local_slots(const RedisModuleSlotRangeArray *ranges) {
+  uint32_t version_before = atomic_load_explicit(&key_space_version, memory_order_relaxed);
+  uint32_t version_after = slots_tracker_set_local_slots_internal(ranges);
+  if (version_after != version_before) {
+    atomic_store_explicit(&key_space_version, version_after, memory_order_relaxed);
+  }
+}
```
chesedo, line 43: I assume this means you can't define this function in the Rust FFI layer?

GuyAv46: The goal is to avoid FFI overhead in the query execution path, which only requires reading the atomic variable.
With this trick, we can leave the atomic variable on the C side, giving very low overhead (everything is done with relaxed memory order), while still guarantee its maintenance by the FFI library (using static definition in the H file).
Any attempt to maintain the atomic variable from the FFI will give some overhead. This should be simplified once the pipeline execution is moved to Rust.
Alternativly, we could pass a pointer to the atomic variable to the FFI, and only have overhead when updating it (which is fine, as it's not the hot path). But it's an additional unsafety and will also require the caller to pass the right atomic variable, while with this solution the caller soen't have to explicitly know what's the right variable

src/redisearch_rs/headers/slots_tracker.h
```diff
@@ -48,14 +108,16 @@ void slots_tracker_set_local_slots(const RedisModuleSlotRangeArray *ranges);
...
+ *
  * # Safety
  *
  * This function must be called from the main thread only.
  * The `ranges` pointer must be valid and point to a properly initialized RedisModuleSlotRangeArray.
  * The ranges array must contain `num_ranges` valid elements.
  * All ranges must be sorted and have start <= end, with values in [0, 16383].
  */
-void slots_tracker_mark_partially_available_slots(const RedisModuleSlotRangeArray *ranges);
+uint32_t slots_tracker_mark_partially_available_slots_internal(const RedisModuleSlotRangeArray *ranges);
```
JoanFM, line 122: add in the docstring explanation of the return and the parameter

src/redisearch_rs/headers/slots_tracker.h
```diff
@@ -138,15 +200,3 @@ bool slots_tracker_has_fully_available_overlap(const RedisModuleSlotRangeArray *
  * All ranges must be sorted and have start <= end, with values in [0, 16383].
  */
 struct OptionSlotTrackerVersion slots_tracker_check_availability(const RedisModuleSlotRangeArray *ranges);
```
JoanFM, line 204: same here

src/redisearch_rs/c_entrypoint/slots_tracker_ffi/cbindgen.toml
chesedo: Okay, how many places in the code is expected to call this? If it is more than one then we'll have to rethink this (what will happen when one of the call sides are ported to Rust while the others are not)

JoanFM: @GuyAv46 ?

GuyAv46: Each function will be called from a single component (the first from the coordinator, the second from a notification).
In any case, the atomic variable should be linked to the query result processor (where the access performance is critical). I believe we can handle any in-between state while we migrate

chesedo: Okay, sounds good then

PR #7209, opened 2025-11-03
MOD-12017: FT.HYBRID - Fix Index Prefix Checks

kei-nan:
Unicode prefixes were not correctly handled during the FT.HYBRID prefix checks

1. Current: We use const char* while the hidden unicode string was expecting an sds string.
2. Change: Initialize and pass an sds string
3. Outcome: Proper support of unicode, also avoid segfault since we won't access an sds header on a const char*

1. Prefix checking

#### Mark if applicable

---

> [!NOTE]
> Centralizes index prefix validation via IndexSpec_IsCoherent and switches FT.HYBRID prefix parsing to sds for correct unicode handling and safe memory management.
> 
> - **Core**:
>   - Add `IndexSpec_IsCoherent(spec, sds* prefixes, size_t n)` in `spec.c` and declare in `spec.h` to validate index prefixes against schema.
> - **Aggregate**:
>   - Simplify `IsIndexCoherent` in `aggregate_request.c` to delegate to `IndexSpec_IsCoherent` using `sds` args.
> - **Hybrid parsing**:
>   - `_INDEX_PREFIXES`: parse with length (`AC_GetString`), create `sds` per prefix, store as `arrayof(sds)`; free via `array_free_ex(..., sdsfree(...))`.
>   - Remove local prefix coherence function; call `IndexSpec_IsCoherent` with collected prefixes before applying contexts.
>   - Update types and memory management in `hybrid_optional_args.h`, `hybrid_callbacks.c`, and `parse_hybrid.c` to use `sds` throughout.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1d4c4fc673dbcda173180364fea708de7f29e767. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/hybrid/parse_hybrid.c
+++ b/src/hybrid/parse_hybrid.c
@@ -696,7 +696,7 @@ int parseHybridCommand(RedisModuleCtx *ctx, ArgsCursor *ac,
     QueryError_SetError(status, QUERY_ERROR_CODE_MISMATCH, NULL);
     goto error;
   }
-  array_free_ex(prefixes, sdsfree);
+  array_free_ex(prefixes, sdsfree(*(sds *)ptr));
   prefixes = NULL;
 
   // Apply context to each request
@@ -729,7 +729,7 @@ int parseHybridCommand(RedisModuleCtx *ctx, ArgsCursor *ac,
   return REDISMODULE_OK;
 
 error:
-  array_free(prefixes);
+  array_free_ex(prefixes, sdsfree(*(sds *)ptr));
   prefixes = NULL;
   if (mergeSearchopts.params) {
     Param_DictFree(mergeSearchopts.params);


src/spec.c
```diff
@@ -1524,6 +1524,28 @@ int IndexSpec_AddFields(StrongRef spec_ref, IndexSpec *sp, RedisModuleCtx *ctx,
   return rc;
 }
 
+bool IndexSpec_IsCoherent(IndexSpec *spec, sds* prefixes, size_t n_prefixes) {
```
Itzikvaknin, line 1527: Consider having an inner function to compare between arrays of stings
Its easier to unit-test that way

src/hybrid/parse/hybrid_callbacks.c
```diff
@@ -434,10 +434,12 @@ void handleIndexPrefixes(ArgParser *parser, const void *value, void *user_data)
...
   while (!AC_IsAtEnd(paramsArgs)) {
     const char *prefix;
-    if (AC_GetString(paramsArgs, &prefix, NULL, 0) != AC_OK) {
+    size_t prefixLen;
+    if (AC_GetString(paramsArgs, &prefix, &prefixLen, 0) != AC_OK) {
       QueryError_SetError(status, QUERY_ERROR_CODE_PARSE_ARGS, "Bad arguments for _INDEX_PREFIXES");
       return;
     }
-    array_ensure_append_1(*ctx->prefixes, prefix);
+    sds prefixSds = sdsnewlen(prefix, prefixLen);
```
Itzikvaknin, line 442: Dont we convert the main `ArgCursor` to sds when creating the requests?
`HybridRequest_InitArgsCursor`
any reason for that `sds` creation?

kei-nan: I think you know what happens when we rely on the internal string inside ArgsCursor :)
Not a fan on relying how the strings are stored inside ArgsCursor.

src/spec.c
kei-nan: I think at that point we are effectively testing `HiddenUnicodeString_CompareC` in a loop.
I think we already have a test for `HiddenUnicodeString_CompareC` so I think it is redundant.

Itzikvaknin: Your call :)

Itzikvaknin (approved):
⚡

PR #7212, opened 2025-11-04
MOD-12174: FT.HYBRID - Support Running In Single Thread

kei-nan:
We found a bug,
When running with WORKERS=0 we still used depleters that created threads to run the subqueries.
The issue is that the loader they used was plain loader since the background flag was not turned on.
Initially we tried turning the flag on but it led to a deadlock since the main thread was waiting in sendchunk and holding the global lock.

The chosen solution is to not add the depleters when WORKERS is zero.

1. Current: We always add depleters even when WORKERS is Zero.
2. Change: Support not having depleters in the shard subquery pipeline.
3. Outcome: No deadlocks or crashes when running with WORKERS==0.

#### Mark if applicable

---

> [!NOTE]
> Adds a background-depletion flag to hybrid pipelines, allowing single-threaded execution without depleters and using direct upstreams in the merger.
> 
> - **Hybrid pipeline execution**
>   - Introduce `depleteInBackground`/`backgroundDepletion` flags across `HybridRequest_Build{Depletion,Pipeline}()`, `buildPipelineAndExecute()`, and `HybridRequest_StartCursors()` to control depleter usage.
>   - Foreground (single-thread) path builds pipelines without depleters; background path retains depleters and pre-depletion.
>   - `HybridRequest_BuildMergePipeline()` now consumes generic upstream `endProc`s instead of requiring `RP_DEPLETER`s.
> - **Resource/context management**
>   - Always create detached thread-safe `RedisModuleCtx` for subqueries; simplify cleanup path (remove cluster-mode branching).
> - **Command/debug flow**
>   - `hybrid_exec`: wire flag through foreground/background paths; only call `RPDepleter_DepleteAll()` when background.
>   - `hybrid_debug`: build pipeline with background depletion disabled.
> - **Tests**
>   - C++ tests pass new flag when building pipelines.
>   - Python tests use `_FT.CURSOR READ` for internal cursors.
> - **Misc**
>   - Minor typo fix in `RPDepleter_DepleteAll` return statement.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 98e6787de2a7470bb0c152edbcfd3c15ebfa5d8f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/coord/hybrid/dist_hybrid_plan.cpp
+++ b/src/coord/hybrid/dist_hybrid_plan.cpp
@@ -28,7 +28,6 @@ int HybridRequest_BuildDistributedDepletionPipeline(HybridRequest *req, const Hy
   for (size_t i = 0; i < req->nrequests; i++) {
       AREQ *areq = req->requests[i];
 
-      // areq->rootiter = QAST_Iterate(&areq->ast, &areq->searchopts, AREQ_SearchCtx(areq), areq->reqflags, &req->errors[i]);
       AREQ_AddRequestFlags(areq, QEXEC_F_BUILDPIPELINE_NO_ROOT);
 
       int rc = AREQ_BuildPipeline(areq, &req->errors[i]);
--- a/src/hybrid/hybrid_exec.c
+++ b/src/hybrid/hybrid_exec.c
@@ -370,20 +370,21 @@ static inline void replyWithCursors(RedisModuleCtx *replyCtx, arrayof(Cursor*) c
     RedisModule_EndReply(reply);
 }
 
-int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, QueryError *status, bool async) {
+int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, QueryError *status, bool backgroundDepletion) {
     HybridRequest *req = StrongRef_Get(hybrid_ref);
     if (req->nrequests == 0) {
       QueryError_SetError(&req->tailPipelineError, QUERY_ERROR_CODE_GENERIC, "No subqueries in hybrid request");
       return REDISMODULE_ERR;
     }
+    // helper array to collect depleters so in async we can deplete them all at once before returning the cursors
     arrayof(ResultProcessor*) depleters = NULL; 
-    if (async) {
+    if (backgroundDepletion) {
       depleters = array_new(ResultProcessor *, req->nrequests);
     }
     arrayof(Cursor*) cursors = array_new(Cursor*, req->nrequests);
     for (size_t i = 0; i < req->nrequests; i++) {
       AREQ *areq = req->requests[i];
-      if (async) {
+      if (backgroundDepletion) {
         if (areq->pipeline.qctx.endProc->type != RP_DEPLETER) {
           break;
         }
@@ -410,7 +411,7 @@ int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, Q
       return REDISMODULE_ERR;
     }
 
-    if (async) {
+    if (backgroundDepletion) {
       int rc = RPDepleter_DepleteAll(depleters);
       array_free(depleters);
       if (rc != RS_RESULT_OK) {
@@ -437,11 +438,12 @@ int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, Q
  * @param sctx Redis search context
  * @param status Output parameter for error reporting
  * @param internal Whether the request is internal (not exposed to the user)
+ * @param depleteInBackground Whether the pipeline should be built for asynchronous depletion
  * @return REDISMODULE_OK on success, REDISMODULE_ERR on error
 */
 static int buildPipelineAndExecute(StrongRef hybrid_ref, HybridPipelineParams *hybridParams,
                                    RedisModuleCtx *ctx, RedisSearchCtx *sctx, QueryError *status,
-                                   bool internal, bool async) {
+                                   bool internal, bool depleteInBackground) {
   // Build the pipeline and execute
   HybridRequest *hreq = StrongRef_Get(hybrid_ref);
   hreq->reqflags = hybridParams->aggregationParams.common.reqflags;
@@ -450,16 +452,16 @@ static int buildPipelineAndExecute(StrongRef hybrid_ref, HybridPipelineParams *h
   if (internal) {
     RS_LOG_ASSERT(isCursor, "Internal hybrid command must be a cursor request from a coordinator");
     isCursor = true;
-    if (HybridRequest_BuildDepletionPipeline(hreq, hybridParams, async) != REDISMODULE_OK) {
+    if (HybridRequest_BuildDepletionPipeline(hreq, hybridParams, depleteInBackground) != REDISMODULE_OK) {
       return REDISMODULE_ERR;
     }
-  } else if (HybridRequest_BuildPipeline(hreq, hybridParams, async) != REDISMODULE_OK) {
+  } else if (HybridRequest_BuildPipeline(hreq, hybridParams, depleteInBackground) != REDISMODULE_OK) {
     return REDISMODULE_ERR;
   }
 
   if (!isCursor) {
     HybridRequest_Execute(hreq, ctx, sctx);
-  } else if (HybridRequest_StartCursors(hybrid_ref, ctx, status, async) != REDISMODULE_OK) {
+  } else if (HybridRequest_StartCursors(hybrid_ref, ctx, status, depleteInBackground) != REDISMODULE_OK) {
     return REDISMODULE_ERR;
   }
 
--- a/src/hybrid/hybrid_request.c
+++ b/src/hybrid/hybrid_request.c
@@ -21,12 +21,12 @@
 extern "C" {
 #endif
 
-int HybridRequest_BuildDepletionPipeline(HybridRequest *req, const HybridPipelineParams *params, bool async) {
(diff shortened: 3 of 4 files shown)


src/hybrid/hybrid_request.c
```diff
@@ -21,10 +21,13 @@
...
 
-int HybridRequest_BuildDepletionPipeline(HybridRequest *req, const HybridPipelineParams *params) {
+int HybridRequest_BuildDepletionPipeline(HybridRequest *req, const HybridPipelineParams *params, bool async) {
     // Create synchronization context for coordinating depleter processors
     // This ensures thread-safe access when multiple depleters read from their pipelines
-    StrongRef sync_ref = DepleterSync_New(req->nrequests, params->synchronize_read_locks);
+    StrongRef sync_ref = {0};
+    if (async) {
+      sync_ref = DepleterSync_New(req->nrequests, params->synchronize_read_locks);
+    }
```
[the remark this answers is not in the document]
kei-nan: I think this is no longer the case

src/hybrid/hybrid_request.c
```diff
@@ -308,28 +308,21 @@ void HybridRequest_ClearErrors(HybridRequest *req) {
...
-  } else {
-    // Standalone mode: create detached context for thread safety
-    RedisModuleCtx *detachedCtx = RedisModule_GetDetachedThreadSafeContext(ctx);
-    RedisModule_SelectDb(detachedCtx, RedisModule_GetSelectedDb(ctx));
-    return NewSearchCtxC(detachedCtx, indexname, true);
-  }
+static RedisSearchCtx* createThreadSafeSearchContext(RedisModuleCtx *ctx, const char *indexname) {
+  RedisModuleCtx *detachedCtx = RedisModule_GetDetachedThreadSafeContext(ctx);
+  RedisModule_SelectDb(detachedCtx, RedisModule_GetSelectedDb(ctx));
+  return NewSearchCtxC(detachedCtx, indexname, true);
```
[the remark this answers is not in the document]
kei-nan: this was fixed.

src/coord/hybrid/dist_hybrid_plan.cpp
```diff
@@ -29,7 +29,7 @@ int HybridRequest_BuildDistributedDepletionPipeline(HybridRequest *req, const Hy
       AREQ *areq = req->requests[i];
 
       // areq->rootiter = QAST_Iterate(&areq->ast, &areq->searchopts, AREQ_SearchCtx(areq), areq->reqflags, &req->errors[i]);
```
nafraf: Not related to this PR. But can we remove this commented line to clean up the code?

src/hybrid/hybrid_exec.c
```diff
@@ -432,7 +441,7 @@ int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, Q
 */
 static int buildPipelineAndExecute(StrongRef hybrid_ref, HybridPipelineParams *hybridParams,
                                    RedisModuleCtx *ctx, RedisSearchCtx *sctx, QueryError *status,
-                                   bool internal) {
+                                   bool internal, bool async) {
```
nafraf: Consider adding `async` to function `@params` description.

src/hybrid/hybrid_exec.c
```diff
@@ -397,22 +402,26 @@ int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, Q
...
-    int rc = RPDepleter_DepleteAll(depleters);
-    array_free(depleters);
-    if (rc != RS_RESULT_OK) {
-      array_free_ex(cursors, Cursor_Free(*(Cursor**)ptr));
-      if (rc == RS_RESULT_TIMEDOUT) {
-        QueryError_SetWithoutUserDataFmt(status, QUERY_ERROR_CODE_TIMED_OUT, "Depleting timed out");
-      } else {
-        QueryError_SetWithoutUserDataFmt(status, QUERY_ERROR_CODE_GENERIC, "Failed to deplete set of results, rc=%d", rc);
+    if (async) {
+      int rc = RPDepleter_DepleteAll(depleters);
```
Itzikvaknin: I dont think the name async is correct here
If `async` -> call `RPDepleter_DepleteAll`, which is a blocking call
Consider change it to `background` or something else

src/hybrid/hybrid_debug.c
```diff
@@ -220,7 +220,7 @@ static HybridRequest_Debug* HybridRequest_Debug_New(RedisModuleCtx *ctx, RedisMo
 
   // Set request flags from hybridParams
   hreq->reqflags = hybridParams.aggregationParams.common.reqflags;
-  if (HybridRequest_BuildPipeline(hreq, &hybridParams) != REDISMODULE_OK) {
+  if (HybridRequest_BuildPipeline(hreq, &hybridParams, false) != REDISMODULE_OK) {
```
Itzikvaknin, line 223: Is this also the case in Aggregate/Search? Can we use debug commands with `WORKERS != 0`?

src/hybrid/hybrid_exec.c
```diff
@@ -370,20 +370,22 @@ static inline void replyWithCursors(RedisModuleCtx *replyCtx, arrayof(Cursor*) c
...
     arrayof(Cursor*) cursors = array_new(Cursor*, req->nrequests);
     for (size_t i = 0; i < req->nrequests; i++) {
       AREQ *areq = req->requests[i];
-      if (areq->pipeline.qctx.endProc->type != RP_DEPLETER) {
-         break;
+      if (async) {
+        if (areq->pipeline.qctx.endProc->type != RP_DEPLETER) {
+          break;
+        }
+        array_ensure_append_1(depleters, areq->pipeline.qctx.endProc);
```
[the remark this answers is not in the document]
Itzikvaknin: Three lines above:
```
    if (async) {
      depleters = array_new(ResultProcessor *, req->nrequests);
    }
```

src/hybrid/hybrid_exec.c
```diff
@@ -370,20 +370,25 @@ static inline void replyWithCursors(RedisModuleCtx *replyCtx, arrayof(Cursor*) c
...
-int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, QueryError *status) {
+int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, QueryError *status, bool async) {
     HybridRequest *req = StrongRef_Get(hybrid_ref);
     if (req->nrequests == 0) {
       QueryError_SetError(&req->tailPipelineError, QUERY_ERROR_CODE_GENERIC, "No subqueries in hybrid request");
       return REDISMODULE_ERR;
     }
-    arrayof(ResultProcessor*) depleters = array_new(ResultProcessor*, req->nrequests);
+    arrayof(ResultProcessor*) depleters = NULL; 
+    if (async) {
```
Itzikvaknin: Can you add a comment that this is a helper array?

Itzikvaknin (changes requested):
Nice catch!
Left a few comments.
We need to think about how we can improve the coverage here, specifically in the optimized case. I think most of our tests at the moment are running with `WORKERS = 0`

src/hybrid/hybrid_debug.c
oshadmi: Can be done in another PR - to support debug hybrid command to run in the workers thread pool when it exists

oshadmi: (MOD-12318)

src/hybrid/hybrid_exec.c
kei-nan: went with:
    // helper array to collect depleters so in async we can deplete them all at once before returning the cursors

src/hybrid/hybrid_exec.c
kei-nan: changed to `depleteInBackground` and `backgroundDepletion`

Itzikvaknin (approved):
Not blocking this PR, but we need to make sure tests are added
I think this will be covered in test_hybrid_multithreaded.py once it’s enabled for the cluster, but we shouldn’t let it slip through

tests/pytests/test_hybrid_internal.py
```diff
@@ -84,7 +84,7 @@ def read_cursor_completely_resp2(env, index_name, cursor_id, batch_callback=None
     current_cursor = cursor_id
 
     while current_cursor != 0:
-        cursor_response = env.cmd('FT.CURSOR', 'READ', index_name, current_cursor)
+        cursor_response = env.cmd('_FT.CURSOR', 'READ', index_name, current_cursor)
```
oshadmi, line 87: Can we keep both?

kei-nan: since we use _FT.HYBRID it doesn't make sense to keep both.
The file is supposed to test the internal flow of hybrid commands.
So we expect to always operate against the shard.

tests/pytests/test_hybrid_internal.py
```diff
@@ -43,7 +43,7 @@ def read_cursor_completely_resp3(env, index_name, cursor_id, batch_callback=None
...
```
kei-nan, line 46: the cursor id belongs to the shard.
it was created with WORKERS 0 - > unsafe loader
"FT.CURSOR" goes to the coordinator which assumes it is a coordinator cursor id.
it moves the cursor read to be done in the background.
Since the pipeline uses a non safe loader we get a crash because we are reading from the cursor in a coordinator background thread.
by using "_FT.CURSOR" we go directly to the shard and avoid any background thread issues.

oshadmi (approved):

PR #7330, opened 2025-11-12
[DEBUG] Cleanup SHARD_CONNECTION_STATES debug command

GuyAv46:
### Cleanup debug command for master

1. Use array for the statuses list (for simplicity and auto-managed memory)
2. Use block client callbacks
3. Call `RedisModule_BlockedClientMeasureTimeEnd` from the main thread

#### Mark if applicable

---

> [!NOTE]
> Switch connection-state aggregation to dynamic arrays and move reply/cleanup to block-client callbacks with proper thread coordination.
> 
> - **Debug command: `SHARD_CONNECTION_STATES`**
>   - Use dynamic arrays for per-host connection states (`array_new/append/len`) instead of NULL-terminated lists.
>   - Build keys with `rm_asprintf`; fetch/update entries via `dictAddOrFind` and `dictSetVal`.
>   - Reply generation now iterates via `array_len(stateList)`.
>   - Refactor threading: workers only fill shared `replyDict` and decrement `pending_threads`; last worker calls `RedisModule_UnblockClient`.
>   - Add `connectionPoolStateReply` (main-thread reply) and `freeConnectionPoolStateCtx` (cleanup); call `RedisModule_BlockedClientMeasureTimeEnd` in reply.
>   - `MR_GetConnectionPoolState` now uses `RedisModule_BlockClient` with reply/free callbacks.
> - **Utilities**
>   - Update `stringsListValDestructor` to free dynamic arrays with `array_free_ex`.
>   - Initialize new dict entry value (`RS_dictAddOrFind`) to zero to distinguish new vs. existing entries.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9a4d16671661b60a36eb1e81cc860a93fc8dcde2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/coord/rmr/conn.c
```diff
@@ -151,17 +151,11 @@ void MRConnManager_ReplyState(dict *stateDict, RedisModuleCtx *ctx) {
     RedisModule_ReplyWithCString(ctx, key);
 
     // Get the value (array of connection state strings)
-    char **stateList = dictGetVal(entry);
```
JoanFM, line 154: can we avoid adding unrelated changes? we can add another PR to do these other cleanups

src/coord/rmr/rmr.c
```diff
@@ -327,31 +327,41 @@ static void uvGetConnectionPoolState(void *p) {
   IORuntimeCtx *ioRuntime = reducedConnPoolStateCtx->ioRuntime;
   struct MultiThreadedRedisBlockedCtx *mt_bc = reducedConnPoolStateCtx->mt_ctx;
   RedisModuleBlockedClient *bc = mt_bc->bc;
-  RedisModuleCtx *ctx = RedisModule_GetThreadSafeContext(bc);
```
JoanFM, line 330: why are we not getting the safe context? What is the safe context's purpose?

src/coord/rmr/rmr.c
```diff
@@ -327,31 +327,41 @@ static void uvGetConnectionPoolState(void *p) {
...
-  }
+  size_t pending_threads = --mt_bc->pending_threads;
+  pthread_mutex_unlock(&mt_bc->lock);
 
+  if (pending_threads == 0) {
+    // We are the last ones to reply, so we can now send the response (from the unblock callback)
+    RedisModule_UnblockClient(bc, mt_bc);
+  }
+  // Request is complete for this ioRuntime
+  IORuntimeCtx_RequestCompleted(ioRuntime);
```
JoanFM, line 341: this is the only part of the fix? Is there a test that reproduces the issue?

JoanFM (changes requested):

JoanFM:
Let's please separate the resolution of the issue from the extra cleanups.

src/coord/rmr/rmr.c
GuyAv46: will add in the new PR

src/coord/rmr/rmr.c
GuyAv46: We only use it for the reply, so we don't need to create it here anymore

src/util/dict.c
```diff
@@ -128,15 +128,8 @@ void* redisStringsKeyDup(void *privdata, const void *key){
 }
 
 static void stringsListValDestructor(void *privdata, void *val) {
-  if (!val) return;
```
JoanFM, line 131: let's assert val then?

GuyAv46: `array_free` (and the `array_free_ex` variant) can handle `NULL`s

src/util/dict.c
```diff
@@ -481,6 +474,9 @@ int RS_dictReplace(dict *d, void *key, void *val)
 dictEntry *RS_dictAddOrFind(dict *d, void *key) {
     dictEntry *entry, *existing;
     entry = RS_dictAddRaw(d,key,&existing);
+    // If we have a new entry, initialize the value union to zero,
+    // otherwise we don't have a way to know it's a new entry or existing one.
+    if (entry) memset(&entry->v, 0, sizeof(entry->v));
```
GuyAv46, line 479: `dictAddOrFind` wasn't used anywhere until now

JoanFM (dismissed):

PR #7336, opened 2025-11-12
MOD-12212: Fix high temporary memory consumption when loading search data from RDB.

Itzikvaknin:
Manually backported to 2.10, 2.8: 
- 2.10: https://github.com/RediSearch/RediSearch/pull/7360
- 2.8: https://github.com/RediSearch/RediSearch/pull/7366

---

> [!NOTE]
> Defers GC startup and registration to non-duplicate indexes when loading from RDB, and adds a test verifying only one index is stored when duplicates are present.
> 
> - **Backend (RDB load)**:
>   - Move `IndexSpec_StartGC` to run only in the non-duplicate path in `IndexSpec_StoreAfterRdbLoad`, preventing GC startup for duplicate indexes.
> - **Tests**:
>   - Add `testDuplicateIndexRdbLoad` to write the same index 30 times to RDB and verify only one is loaded/stored.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c9b15e113615cd8711a8c01ff01c38a2f3fe945e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/cpptests/test_cpp_rdb.cpp
+++ b/tests/cpptests/test_cpp_rdb.cpp
@@ -287,6 +287,10 @@ TEST_F(RdbMockTest, testDuplicateIndexRdbLoad) {
     }
     EXPECT_EQ(0, RMCK_IsIOError(io));
 
+    // Remove the original spec from globals before loading from RDB
+    IndexSpec_RemoveFromGlobals(spec_ref, false);
+    ASSERT_TRUE(IndexSpec_LoadUnsafe("test_duplicate_idx").rm == NULL);
+
     // Reset read position to load from RDB
     io->read_pos = 0;
 


pack/ramp-enterprise.yml
```diff
@@ -6,7 +6,7 @@ description: High performance search index on top of Redis (with clustering)
 homepage: 'http://redisearch.io'
 license: Redis Source Available License 2.0 (RSALv2) or the Server Side Public License v1 (SSPLv1) or the GNU Affero General Public License version 3 (AGPLv3)
 command_line_args: ""
-compatible_redis_version: "99.99"
+compatible_redis_version: "8.2"
```
JoanFM: can we put this into another PR to isolate changes?

JoanFM: this would clarify the context and socpe, and make backporting easier

Itzikvaknin: This is a temporary fix, so i could upload it and test it on master cluster

src/spec.c
```diff
@@ -3107,7 +3107,6 @@ static int IndexSpec_StoreAfterRdbLoad(IndexSpec *sp) {
     RedisModule_Log(RSDummyContext, "notice", "Loading an already existing index, will just ignore.");
   }
 
-  IndexSpec_StartGC(spec_ref, sp);
   Cursors_initSpec(sp);
```
GuyAv46, line 3110: Let's verify this is not causing similar issues on previous versions (even if it's only setting a number on `master`)

tests/cpptests/test_cpp_rdb.cpp
```diff
@@ -258,3 +258,55 @@ TEST_F(RdbMockTest, testIndexSpecStringSerialize) {
...
+    EXPECT_EQ(REDISMODULE_OK, result);
+    EXPECT_EQ(0, RMCK_IsIOError(io));
+
+
+    // Verify the loaded index exists and has the correct name
+    StrongRef loaded_spec_ref = IndexSpec_LoadUnsafe("test_duplicate_idx");
+    IndexSpec *loaded_spec = (IndexSpec *)StrongRef_Get(loaded_spec_ref);
+    ASSERT_TRUE(loaded_spec != nullptr);
+    ASSERT_STREQ(HiddenString_GetUnsafe(loaded_spec->specName, NULL), "test_duplicate_idx");
+    ASSERT_EQ(loaded_spec->numFields, 1);
```
GuyAv46, line 308: If possible, check that we only created a single GC when loading

Itzikvaknin: Looks like the GC context is owned by the spec object.
The duplication path was: 
- Create a GC context (add the job to the GC pool)
- Free the spec
- GC wakes up, it has weak ref to null spec, free and finish the job 
So in this case, the GC context object is kind of orphan, and hard to test in a non flaky way as far as I know

src/spec.c
Itzikvaknin: It doesn't, verified on 2.8 and 2.10

oshadmi (approved):

PR #7344, opened 2025-11-12
RED-176382 Fix FT.CREATE failure with LeanVec parameters on non-Intel architectures

meiravgri:
`VecSim_IsLeanVecCompressionType()` was checking both the compression type AND Intel architecture support, causing `FT.CREATE` commands with valid `LeanVec` parameters to fail on non-Intel systems during parameter validation.

### Solution
- Simplified `VecSim_IsLeanVecCompressionType()` to only check if the compression type is LeanVec (removed Intel CPU check)
- Exposed `isLVQSupported()` function to check Intel architecture support separately
- Moved the Intel architecture gating to `FT.INFO` when displaying `reduced_dim` field

### DIFF
--- a/tests/pytests/test_vecsim_svs.py
+++ b/tests/pytests/test_vecsim_svs.py
@@ -162,6 +162,8 @@ def test_svs_vamana_info():
                     'DIM', dim, 'DISTANCE_METRIC', 'L2']
         if compression_type != 'NO_COMPRESSION':
             cmd_params.extend(['COMPRESSION', compression_type])
+            if compression_type == 'LeanVec4x8':
+                cmd_params.extend(['REDUCE', dim // 2])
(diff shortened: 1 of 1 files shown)


tests/pytests/test_vecsim_svs.py
```diff
@@ -156,7 +156,7 @@ def test_svs_vamana_info():
     data_type = random.choice(VECSIM_SVS_DATA_TYPES)
 
     # Create SVS VAMANA index with all compression flavors (except for global SQ8).
-    compression_types = SVS_COMPRESSION_TYPES if is_intel_opt_enabled() and EXTENDED_PYTESTS else ['NO_COMPRESSION', 'LVQ8']
+    compression_types = SVS_COMPRESSION_TYPES if is_intel_opt_enabled() and EXTENDED_PYTESTS else ['NO_COMPRESSION', 'LVQ8', 'LeanVec4x8']
```
alonre24, line 159: Let's add `REDUCE` to `cmd_params` if we're on `LeanVec4x8`.

alonre24 (approved):

PR #7345, opened 2025-11-12
[MOD-12320] Ensure full profile output on timeout with RETURN policy

meiravgri:
### Overview

Previously, shard profile data was omitted on timeout because it's only sent in the final `CURSOR READ` response. This PR ensures that profile data from **all shards** is collected and returned even when a timeout occurs during profiling with `RETURN` policy. 

**Implementation:** Introduces a new internal `FT.CURSOR PROFILE` command that is sent to shards (instead of `CURSOR READ`) when a timeout occurs with RETURN policy during profiling, allowing the coordinator to retrieve profile data from all shards.

### Execution Flow

**Normal profiling flow (no timeout):**
1. Coordinator sends `FT.AGGREGATE` with PROFILE to shards
2. Shards return partial results via `CURSOR READ` responses
3. Final `CURSOR READ` (cursor_id=0) includes profile data
4. Coordinator collects and prints all shard profiles

**New timeout handling flow:**

1. **Shard execution** (`sendChunk`):
   - Detects timeout during execution
   - Sets `req->has_timedout = true` to **preserve timeout state for later `FT.CURSOR PROFILE` call**
   - Returns partial results with cursor_id (not 0)
   
2. **Coordinator thread** (`rpnetNext` → `sendChunk`):
   - Detects timeout (via coordinator timer or shard warning in RESP3)
   - Returns `RS_RESULT_TIMEDOUT`
   - Sets `MRIteratorCallback_SetTimedOut()` on iterator context **only if detected by coordinator timer**
   
3. **UV thread** (`netCursorCallback`):
   - For RESP3 + profiling: may also detect timeout warning in shard reply and set `MRIteratorCallback_SetTimedOut()` 
   - Calls `getCursorCommand()` to determine next command
   
4. **UV thread** (`getCursorCommand`):
   - Checks timeout flag
   - If timeout + profiling: **switches command** from `CURSOR READ` → `CURSOR PROFILE`
   - Sends `FT.CURSOR PROFILE {index} {CID}` to shard
   
5. **Shard execution** (`RSCursorCommand` with PROFILE):
   - Retrieves cursor with preserved `has_timedout` state
   - Calls `sendChunk_ReplyOnly_EmptyResults()` which uses `createProfilePrinterCtx()`
   - Returns an empty results containing the profile data with timeout information
   - Purges cursor
   
6. **Coordinator thread** (`printAggProfile`):
   - **Drains remaining replies** from channel (the `CURSOR PROFILE` responses)

---

> [!NOTE]
> Ensures complete profile output on timeout by adding FT.CURSOR PROFILE and wiring timeout propagation across shards, coordinator, and reply handling.
> 
> - **Profile/Timeout handling**:
>   - Add `AREQ::has_timedout` and set it during execution; use new `createProfilePrinterCtx()` everywhere profile is printed (incl. empty-results replies).
>   - RESP3 warning parsing marks iterator as timed out; coordinator now preserves and uses timeout state.
> - **New command**:
>   - Implement `FT.CURSOR PROFILE {index} {CID}` to fetch profile after timeout and purge cursor; RESP2/RESP3 supported.
> - **Coordinator changes**:
>   - On profiling timeout, switch follow-up shard command from `CURSOR READ` to `CURSOR PROFILE` (`MRRootCommand` adds `C_PROFILE`).
>   - `printAggProfile` drains pending replies and logs shard/profile count; uses iterator helpers `MRIterator_GetChannelSize()` and `MRIterator_GetNumShards()`.
>   - Expose `getNextReply(RPNet*)` for draining.
> - **RMR/MRIterator**:
>   - Add iterator helpers `MRIterator_GetChannelSize`, `MRIterator_GetNumShards` and export them; minor command mutation/reset fixes when switching to PROFILE.
> - **Tests**:
>   - Add cursor profile validations and coordinator profile timeout tests; assert shard profiles count and errors for invalid PROFILE usage.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f8845f0bb33df8970df2bd1337e7644c30fbd73c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -1402,9 +1402,11 @@ int RSCursorCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
       // The index was dropped while the cursor was idle.
       // Notify the client that the query was aborted.
       RedisModule_Reply_Error(reply, "The index was dropped while the cursor was idle");
+    } else {
+      sendChunk_ReplyOnly_EmptyResults(reply, req);
+      IndexSpecRef_Release(execution_ref);
     }
-    sendChunk_ReplyOnly_EmptyResults(reply, req);
-    IndexSpecRef_Release(execution_ref);
+
     // Free the cursor
     Cursor_Free(cursor);
   } else if (strcasecmp(cmd, "DEL") == 0) {


src/coord/rpnet.c
```diff
@@ -108,7 +108,7 @@ static int getNextReply(RPNet *nc) {
   if (nc->cmd.forProfiling) {
     // if the cursor id is 0, this is the last reply from this shard, and it has the profile data
     if (CURSOR_EOF == MRReply_Integer(MRReply_ArrayElement(root, 1))) {
-      MRReply *profile_data;
+    MRReply *profile_data;
```
GuyAv46:
```suggestion
      MRReply *profile_data;
```

src/aggregate/aggregate.h
```diff
@@ -261,6 +261,10 @@ typedef struct AREQ {
 
   // The offset of the prefixes in the command
   size_t prefixesOffset;
+
+  // Indicates whether the query has timed out.
+  // Useful for query with cursor and RETURN policy
+  bool has_timedout;
```
GuyAv46, line 267: Today, we have this in the error structure of the request. Can't we reuse it instead?

src/aggregate/aggregate_exec.c
```diff
@@ -721,12 +735,8 @@ void sendChunk(AREQ *req, RedisModule_Reply *reply, size_t limit) {
...
 
     if (IsProfile(req)) {
-
-      ProfilePrinterCtx profileCtx = {0};
-      profileCtx.req = req;
-      profileCtx.queryOOM = QueryError_HasQueryOOMWarning(AREQ_QueryProcessingCtx(req)->err);
-
-      RedisModule_Reply_MapEnd(reply); // >Results
+      ProfilePrinterCtx profileCtx = createProfilePrinterCtx(req);
+      RedisModule_Reply_MapEnd(reply);  // >Results
```
GuyAv46, line 739: Can we close the map first, and add a comment on why we do it?

src/aggregate/aggregate_exec.c
```diff
@@ -1367,6 +1376,26 @@ int RSCursorCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+    if (!IsProfile(req)) {
+      RedisModule_ReplyWithErrorFormat(ctx, "cursor request is not profile, id: %d", cid);
+      RedisModule_EndReply(reply);
+      return REDISMODULE_OK;
+    }
+
+    sendChunk_ReplyOnly_EmptyResults(reply, req);
+
+    // Free the cursor
+    Cursors_Purge(GetGlobalCursor(cid), cid);
```
GuyAv46: Use `Cursors_Free` when we already have it on hand. `Cursors_Purge` is for requesting to delete an unheld cursor.
```suggestion
    Cursors_Free(cursor);
```

src/aggregate/aggregate_exec.c
```diff
@@ -1367,6 +1376,26 @@ int RSCursorCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
       cursorRead(reply, cursor, count, false);
     }
+  } else if (strcasecmp(cmd, "PROFILE") == 0) {
+    // Return profile
+    Cursor *cursor = Cursors_TakeForExecution(GetGlobalCursor(cid), cid);
+    if (cursor == NULL) {
+      RedisModule_ReplyWithErrorFormat(ctx, "Cursor not found, id: %d", cid);
+      RedisModule_EndReply(reply);
+      return REDISMODULE_OK;
+    }
```
GuyAv46, line 1387: Let's cover this scenario

src/aggregate/aggregate_exec.c
```diff
@@ -1367,6 +1376,26 @@ int RSCursorCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+
+    AREQ *req = cursor->execState;
+    if (!IsProfile(req)) {
+      RedisModule_ReplyWithErrorFormat(ctx, "cursor request is not profile, id: %d", cid);
+      RedisModule_EndReply(reply);
+      return REDISMODULE_OK;
```
GuyAv46, line 1393: Let's cover this scenario. We leak the cursor here (we should pause it again)

src/coord/dist_aggregate.c
```diff
@@ -228,11 +228,29 @@ void printAggProfile(RedisModule_Reply *reply, void *ctx) {
...
+
+  // Therefore, we loop to drain all remaining replies from the channel.
+  // Pending might be zero, but there might still be replies in the channel to read.
+  // We may have pulled all the replies from the channel and arrived here due to a timeout,
+  // and now we're waiting for the profile results.
+  while (MRIterator_GetPending(rpnet->it) || MRIterator_GetChannelSize(rpnet->it)) {
+    getNextReply(rpnet);
+  }
+
+  size_t num_shards = GetNumShards_UnSafe();  // can we make this safe
```
GuyAv46: Can we get this from the iterator? It will also make more sense to take the number of commands we sent and not the current number of shards

src/coord/dist_aggregate.c
```diff
@@ -228,11 +228,29 @@ void printAggProfile(RedisModule_Reply *reply, void *ctx) {
   // profileRP replace netRP as end PR
   ProfilePrinterCtx *cCtx = ctx;
   RPNet *rpnet = (RPNet *)AREQ_QueryProcessingCtx(cCtx->req)->rootProc;
+
+  // Therefore, we loop to drain all remaining replies from the channel.
+  // Pending might be zero, but there might still be replies in the channel to read.
+  // We may have pulled all the replies from the channel and arrived here due to a timeout,
+  // and now we're waiting for the profile results.
+  while (MRIterator_GetPending(rpnet->it) || MRIterator_GetChannelSize(rpnet->it)) {
+    getNextReply(rpnet);
```
GuyAv46: Don't we leak the replies here? Shouldn't we free them?

src/coord/dist_aggregate.c
```diff
@@ -228,11 +228,29 @@ void printAggProfile(RedisModule_Reply *reply, void *ctx) {
   // profileRP replace netRP as end PR
   ProfilePrinterCtx *cCtx = ctx;
   RPNet *rpnet = (RPNet *)AREQ_QueryProcessingCtx(cCtx->req)->rootProc;
+
+  // Therefore, we loop to drain all remaining replies from the channel.
+  // Pending might be zero, but there might still be replies in the channel to read.
+  // We may have pulled all the replies from the channel and arrived here due to a timeout,
+  // and now we're waiting for the profile results.
+  while (MRIterator_GetPending(rpnet->it) || MRIterator_GetChannelSize(rpnet->it)) {
```
GuyAv46: Consider checking that `getNextReply` didn't return `NULL` instead (this should cover everything, as we don't support profile in cursor mode), and maybe just as a precaution, also check `MRIterator_GetPending`

src/coord/hybrid/dist_utils.c
```diff
@@ -33,6 +33,23 @@ void netCursorCallback(MRIteratorCallbackCtx *ctx, MRReply *rep) {
     return;
   }
 
+  if (cmd->forProfiling && cmd->protocol == 3) {
+    MRReply *rows = NULL, *meta = NULL;
+    meta = MRReply_ArrayElement(rep, 0);
+    meta = MRReply_MapElement(meta, "results");  // profile has an extra level
```
GuyAv46: It would be safer to put this after the assertions block (so we are guaranteed the structure is valid)

src/coord/hybrid/dist_utils.c
```diff
@@ -33,6 +33,23 @@ void netCursorCallback(MRIteratorCallbackCtx *ctx, MRReply *rep) {
...
+
+    // Check if we got timeout
+    MRReply *warning = MRReply_MapElement(meta, "warning");
+    if (MRReply_Length(warning) > 0) {
+      const char *warning_str = MRReply_String(MRReply_ArrayElement(warning, 0), NULL);
+      // Set an error to be later picked up by `getCursorCommand`
+      if (!strcmp(warning_str, QueryError_Strerror(QUERY_ERROR_CODE_TIMED_OUT))) {
+        // When a shard returns timeout on RETURN policy, the profile is not returned.
+        // We set the timeout here so in the next getCursorCommand we will send CURSOR PROFILE
+        MRIteratorCallback_SetTimedOut(MRIteratorCallback_GetCtx(ctx));
```
GuyAv46: Shouldn't we do it from the profile reply? Or we assume profile means no cursor, so we don't need to retain the remote cursor? (let's assert and comment that if so)

src/coord/hybrid/dist_utils.c
```diff
@@ -157,11 +178,23 @@ bool getCursorCommand(long long cursorId, MRCommand *cmd, MRIteratorCtx *ctx) {
...
-    // command instead of a READ command (here we know it has more results)
-    if (timedout && !cmd->forCursor) {
-      MRCommand_ReplaceArg(cmd, 1, "DEL", 3);
-      cmd->rootCommand = C_DEL;
+    if (timedout) {
+      // We are going to modify the command, so we need to free the cached sds
+      if (cmd->cmd) {
+        sdsfree(cmd->cmd);
+        cmd->cmd = NULL;
+      }
```
GuyAv46, line 188: I'm going to fix this soon, but I guess it's a must for this PR to work, right?

GuyAv46 (commented):
NICE

src/aggregate/aggregate.h
meiravgri: isn't it cleaned up?

src/aggregate/aggregate_exec.c
meiravgri: I don't understand your suggestion

src/aggregate/aggregate_exec.c
meiravgri: mm any idea how?

src/coord/dist_aggregate.c
meiravgri: i had a comment explaining exactly why getNextReply alone is not enough i dont know where its gone

src/coord/hybrid/dist_utils.c
meiravgri: yes
Thanks!

src/coord/dist_aggregate.c
meiravgri: we probably should

src/coord/dist_aggregate.c
meiravgri: you mean it->len?

src/aggregate/aggregate_exec.c
GuyAv46: Unit test, or a Python standalone and try calling on a normal cursor

src/coord/hybrid/dist_utils.c
meiravgri: i dont understand you question

src/aggregate/aggregate_exec.c
GuyAv46: Close the map call before defining the profiling context, as they are not related.
And the comment to explain why we close the map (not related to your PR, not mandatory)

src/aggregate/aggregate_exec.c
meiravgri: aw right

src/aggregate/aggregate.h
GuyAv46: Ahh I see

src/coord/dist_aggregate.c
meiravgri: replaced with an if conditions

src/coord/hybrid/dist_utils.c
meiravgri: added assert on cmd->forCursor

src/aggregate/aggregate_exec.c
meiravgri: done

GuyAv46 (approved):

src/aggregate/aggregate_exec.c
```diff
@@ -1367,6 +1377,36 @@ int RSCursorCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+    StrongRef execution_ref = IndexSpecRef_Promote(cursor->spec_ref);
+    if (!StrongRef_Get(execution_ref)) {
+      // The index was dropped while the cursor was idle.
+      // Notify the client that the query was aborted.
+      RedisModule_Reply_Error(reply, "The index was dropped while the cursor was idle");
+    }
+    sendChunk_ReplyOnly_EmptyResults(reply, req);
+    IndexSpecRef_Release(execution_ref);
+    // Free the cursor
+    Cursor_Free(cursor);
```
GuyAv46, line 1411: only reply the profile and release the reference if we got it (you can also switch the order if you prefer)
```suggestion
    if (!StrongRef_Get(execution_ref)) {
      // The index was dropped while the cursor was idle.
      // Notify the client that the query was aborted.
      RedisModule_Reply_Error(reply, "The index was dropped while the cursor was idle");
    } else {
      sendChunk_ReplyOnly_EmptyResults(reply, req);
      IndexSpecRef_Release(execution_ref);
    }

    // Free the cursor
    Cursor_Free(cursor);
```

PR #7355, opened 2025-11-13
[MOD-12378] Fix `FLOAT16` overflow in `test_index_multi_value_json`

meiravgri:
The test became flaky after increasing `n` to 250 for FLOAT16 because L2 distance calculations produced values exceeding `FLOAT16's` maximum (65,504), resulting in infinity. 
Fixed by scaling all vector values by 1/8, which keeps the maximum L2 distance at ~3,906 (well within FLOAT16 range) while maintaining test coverage.

### DIFF
--- a/tests/pytests/test_vecsim.py
+++ b/tests/pytests/test_vecsim.py
@@ -1883,6 +1883,10 @@ def test_index_multi_value_json():
         n = 100
         conn.flushall()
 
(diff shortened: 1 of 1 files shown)


meiravgri:
not backporting to 8.2 since multi svs isnt supported in 8.2, and this test doesn't include the change that made it unstable

PR #7296, opened 2025-11-11
[MOD-12390] Swap idlist and metric iterator with the Rust version

LukeMathWalker:
## Describe the changes in the pull request

The PR shows a potential approach to integrate the newly written Rust iterators into the existing C codebase one at a time, without having to wait for all iterators to be ported over.
Most of the heavy-lifting is done in a generic wrapper that can accommodate all "primitive" iterators—i.e. iterators that don't, internally, wrap around one or more other iterators. 

## Unsorted variants

The debug assertions in the Rust version of `IdList` and `Metric` highlighted that the sortedness assumption made in `IdList` was violated by RFF in a few cases.
This PR tries to resolve the issue by introducing unsorted versions of both iterators, utilising a `const` generic parameter with no runtime overhead. Unfortunately, there are cases of RFF trying to skip over an unsorted list of document ids.
To unblock merging, we gate our checks behind a feature flag `disable_sort_checks_in_idlist`. The feature flag should be removed once the bug in RFF is fixed.

## Future extensions

To accommodate composite iterators, we will also need a shim in the other direction: a way to consume a C iterator as if it were a Rust iterator. It should be possible, using a pretty similar approach, but I haven't fleshed it out in this branch. `MOD-12375`

## Performance

The wrapper introduces a performance penalty, which would disappear once the migration is complete and the wrapper is removed. But it'd allow us to validate our work incrementally rather than pushing all integration issues at the very end of the epic.
There is also a performance penalty when converting C arrays for document ids and metric data into Rust vectors. It is left as a follow-up PR to devise a more efficient scheme to pass data over the FFI boundary.

## Follow-up work

- Use different `IteratorType` variants for the sorted/unsorted versions of `IdList` and `MetricIterator`: `MOD-12373`
- Use a more efficient mechanism to pass ids and metric data over the FFI boundary: `MOD-12374`
- Fix the bug in RFF and remove the `disable_sort_checks_in_idlist` feature flag: `MOD-12370` and `MOD-12372`

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

---

> [!NOTE]
> Replaces C `IdList`/`Metric` iterators with Rust-backed implementations, updates all call sites (including vector, profiling, benches, and tests), and adds sorted/unsorted iterator variants.
> 
> - **Iterators (C → Rust)**:
>   - Remove legacy C `idlist_iterator.{c,h}` and switch includes to `iterators_rs.h`.
[...]

PR #7364, opened 2025-11-13
[MOD-12389] add ffi compat layer so Rust iterators can be used from C

gdesmott:
Compat layer enabling Rust iterators to be used from C.

Only the `idlist` and `metric` iterators are using it for now.

See https://github.com/RediSearch/RediSearch/pull/7296 for the full context.

#### Mark if applicable

---

> [!NOTE]
> Introduces `iterators_ffi` bridging Rust iterators to C, exposing id list and metric iterator APIs, integrating into `redisearch_rs`, and updating build exclusions.
> 
> - **FFI Layer**:
>   - Add `src/redisearch_rs/c_entrypoint/iterators_ffi` crate implementing a C-compatible wrapper for Rust `RQEIterator` with `Read/SkipTo/Revalidate/Rewind` callbacks.
>   - Generate C header `headers/iterators_rs.h` via cbindgen.
> - **New C API (iterators)**:
>   - `NewSortedIdListIterator`, `NewUnsortedIdListIterator` for ID lists.
>   - `NewMetricIteratorSortedById`, `NewMetricIteratorSortedByScore`, `SetMetricRLookupHandle`, `GetMetricOwnKeyRef`, `GetMetricType` for metric iterators.
> - **Integration**:
>   - Add dependency on `iterators_ffi` in `redisearch_rs/Cargo.toml` and re-export as `redisearch_rs::iterators`.
>   - Update `Cargo.lock` to include `iterators_ffi` and reference it from `redisearch_rs`.
> - **Build**:
>   - Update `build.sh` to exclude `iterators_ffi` from Rust bench/coverage crates to avoid early C linking.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5ec7453c27121bc76dbb7a652513d7de0f8f80fd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/redisearch_rs/c_entrypoint/iterators_ffi/src/id_list.rs
```diff
@@ -0,0 +1,85 @@
...
+/// 2. The caller must ensure that `ids` is not null unless `num` is zero.
+/// 3. The memory pointed to by `ids` will be freed using `RedisModule_Free`,
+///    so the caller must ensure that the pointer was allocated in a compatible manner.
+unsafe fn new_id_list_iterator<const SORTED: bool>(
+    ids: *const t_docId,
+    num: u64,
+    weight: f64,
+) -> *mut QueryIterator {
+    // TODO: Figure out a mechanism to avoid re-allocating ids on the Rust side.
+    let mut vec = Vec::with_capacity(num as usize);
```
alonre24, line 65: Probably a newbie question, but how do we guarantee that Rust allocations are using the Redis allocator?

src/redisearch_rs/c_entrypoint/redisearch_rs/src/lib.rs
```diff
@@ -19,6 +19,7 @@ static REDIS_MODULE_ALLOCATOR: redis_module::alloc::RedisAlloc = redis_module::a
 
 pub use fnv_ffi as fnv;
 pub use inverted_index_ffi as inverted_index;
+pub use iterators_ffi as iterators;
```
alonre24, line 22: Do we still need `inverted_index_ffi` ?

src/redisearch_rs/c_entrypoint/iterators_ffi/src/wrapper.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+extern "C" fn read<'index, I: RQEIterator<'index> + 'index>(
+    base: *mut QueryIterator,
+) -> IteratorStatus {
+    // SAFETY: Guaranteed by invariant 1. in [`RQEIteratorWrapper`].
+    let wrapper = unsafe { RQEIteratorWrapper::<I>::from_header(base) };
+    match wrapper.inner.read() {
+        Ok(Some(result)) => {
+            wrapper.header.current = result as *mut RSIndexResult as *mut ffi::RSIndexResult;
```
alonre24, line 91: Why is this double casting needed here?

src/redisearch_rs/c_entrypoint/iterators_ffi/src/wrapper.rs
```diff
@@ -0,0 +1,172 @@
...
+    let wrapper = unsafe { RQEIteratorWrapper::<I>::from_header(base) };
+    wrapper.inner.num_estimated()
+}
+
+extern "C" fn free_iterator<'index, I: RQEIterator<'index> + 'index>(base: *mut QueryIterator) {
+    if !base.is_null() {
+        // SAFETY: Callbacks are guaranteed to get a header pointer created by
+        //  [`RQEIteratorWrapper::new`], which (internally) uses `Box::into_raw` to
+        //  return a raw header pointer.
+        let _ = unsafe { Box::from_raw(base as *mut RQEIteratorWrapper<I>) };
```
alonre24, line 182: Is this deallocating the wrapper?

src/redisearch_rs/c_entrypoint/iterators_ffi/src/id_list.rs
```diff
@@ -0,0 +1,85 @@
...
+        debug_assert_eq!(
+            num, 0,
+            "The pointer to the array of IDs is null, but the number of IDs is non-zero."
+        );
+    }
+    RQEIteratorWrapper::boxed_new(
+        IteratorType_ID_LIST_ITERATOR,
+        IdList::<SORTED>::with_result(vec, RSIndexResult::virt().weight(weight)),
+    )
+}
```
alonre24, line 85: Should we add unit tests here and in the metric iterator? (in addition to the cpp tests in #7296)

src/redisearch_rs/c_entrypoint/iterators_ffi/src/wrapper.rs
```diff
@@ -0,0 +1,172 @@
...
+                Read: Some(read::<I>),
+                SkipTo: Some(skip_to::<I>),
+                Revalidate: Some(revalidate::<I>),
+                Free: Some(free_iterator::<I>),
+                Rewind: Some(rewind::<I>),
+            },
+            inner,
+        });
+        Box::into_raw(wrapper) as *mut QueryIterator
+    }
```
alonre24, line 61: This is quite a magic, super nice :)
No need to have the bound callbacks in `headers_rs.h` as well?

src/redisearch_rs/c_entrypoint/iterators_ffi/src/id_list.rs
gdesmott: Be defining a custom global allocator here: https://github.com/RediSearch/RediSearch/blob/master/src/redisearch_rs/c_entrypoint/redisearch_rs/src/lib.rs#L18

src/redisearch_rs/c_entrypoint/redisearch_rs/src/lib.rs
gdesmott: Yes, that's all the code implementing the C API calling into Rust, we need it.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/wrapper.rs
gdesmott: `&mut T` can be casted to `*mut T` but not to `*mut T2`. While pointers (`*mut`) can be casted to other types.

`RSIndexResult` is a different type than `ffi:RSIndexResult` so we need to cast to a pointer first.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/wrapper.rs
gdesmott: Yes, `Box::from_raw()` takes ownership of the `base` memory, see https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from_raw

We drop the `Box` right away, calling its `Drop` implementation which then frees `base`. It's a common pattern to free memory in ffi code.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/wrapper.rs
gdesmott: No, it's all transparent to the C code. Quite magic indeed. :)

src/redisearch_rs/c_entrypoint/iterators_ffi/src/id_list.rs
gdesmott: I don't think so. All this binding code will be tested extensively once doing the swap (https://github.com/RediSearch/RediSearch/pull/7296) with the existing cpp tests.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/wrapper.rs
```diff
@@ -0,0 +1,172 @@
...
+/// A wrapper around a Rust iterator—i.e. an implementer of the [`RQEIterator`] trait.
+///
+/// It allows existing C code to invoke the Rust iterator
+/// as if it were a C iterator, conforming to the existing C iterator conventions.
+///
+/// # Invariants
+///
+/// 1. It is always safe to cast a raw [`QueryIterator`] pointer returned by [`RQEIteratorWrapper::new`]
+///    to an [`RQEIteratorWrapper`] pointer when invoking one of the callbacks stored in the header.
+pub struct RQEIteratorWrapper<E> {
```
chesedo, line 30: I also need this wrapper for my search on disk work. Would it be possible to have it in a crate where I won't have to pull in a whole lot of other FFI symbols?

gdesmott: Aren't most symbols needed to define the iterator API?

Which ones don't you need here?

alonre24 (dismissed):

src/redisearch_rs/c_entrypoint/iterators_ffi/src/wrapper.rs
chesedo: I don't need any of the iterator C symbols - eg. I don't the id list C symbols since the code is pure Rust.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/metric.rs
```diff
@@ -0,0 +1,164 @@
...
+/// 4. The memory pointed to by `ids` and `metric_list` will be freed using `RedisModule_Free`,
+///    so the caller must ensure that these pointers were allocated in a compatible manner.
+pub unsafe extern "C" fn NewMetricIteratorSortedById(
+    ids: *const t_docId,
+    metric_list: *const f64,
+    num: usize,
+    type_: MetricType,
+) -> *mut QueryIterator {
+    // SAFETY: All safety preconditions are guaranteed by the caller.
+    unsafe { new_metric_iterator::<true>(ids, metric_list, num, type_) }
```
GlenDC, line 35: Not specific to this PR or function, but I was wondering... Would it not hold more value for these safety assumptions if we also couple them with `debug_asserts` the the like where possible? I think most can be done that way and for others we can probably still have checks but we possible phantom and opaque wrapper types embedded.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/wrapper.rs
gdesmott: But what's the downside of having those symbols around in this crate even if you don't use them then?

src/redisearch_rs/c_entrypoint/iterators_ffi/src/metric.rs
gdesmott: How can you check that an input pointer is actually valid and point to the right type of data?

src/redisearch_rs/c_entrypoint/iterators_ffi/src/wrapper.rs
chesedo: I was thinking they will clash with the C symbols... but thinking about it now I not sure how that will happen :sweat_smile: . I can use this as is

src/redisearch_rs/c_entrypoint/iterators_ffi/src/metric.rs
LukeMathWalker: You can check for alignment and it being non-null, which can be useful.
Other unsafe functions here would benefit from checking the iterator type prior to casting, but that requires us to have distinct iterator types for sorted/unsorted before moving forward.

gdesmott: > You can check for alignment and it being non-null, which can be useful.

I added those.

> Other unsafe functions here would benefit from checking the iterator type prior to casting, but that requires us to have distinct iterator types for sorted/unsorted before moving forward.

Let's do that later once `MOD-12373` is done then.

GlenDC (approved):

PR #7377, opened 2025-11-15
Validate CLUSTER SHARDS Reply - [MOD-12432]

GuyAv46:

As we moved to use `CLUSTER SHARDS` instead of `CLUSTER SLOTS`, we discovered a few bugs in redis and that the command may return invalid nodes in some edge cases.

In this PR, we start validating the node's info we get, and filter out badly formatted ones. We now know only of valid shards, which for us means:
1. It has some slots (This was the previous behaviour, and if ASM is used to move slots to the new shard, it will be synced with our schemas and auxiliary data)
2. It has a valid master node:
   1. It has a valid node ID
   2. It has a valid port (positive value)
   3. It has a valid endpoint (not NULL or a reserved bad value)

This requires reverting some previously modified tests.

#### Also fixed in this PR:
Adding slots info back to `SEARCH.CLUSTERINFO`

#### Mark if applicable

---

> [!NOTE]
> Validate and filter CLUSTER SHARDS data (id/endpoint/port), build topology only from valid masters with slots, include `slots` per shard in `SEARCH.CLUSTERINFO`, and update tests accordingly.
> 
> - **Coordinator/Topology**:
>   - Make `parseNode`, `parseMasterNode`, and `parseSlots` return `bool`; validate node `id` (length), non-empty `endpoint`, and positive `port` (prefer `tls-port`).
>   - Build topology only from shards with slots and a valid master; skip invalid shards; handle empty/invalid cases without abort; sort shards by `node.endpoint.port`.
>   - Improve logging for invalid shards; free partial allocations; drop cached local slots after topology update (`Slots_DropCachedLocalSlots`).
> - **API: `SEARCH.CLUSTERINFO`**:
>   - Add `slots` array (slot ranges) per shard to match `CLUSTER SHARDS` syntax.
> - **Tests**:
>   - Update expectations and indices for `SEARCH.CLUSTERINFO` structure (now includes `slots`).
>   - Adjust cluster partition expectations: new shard without slots not counted until assigned.
>   - Add explicit `SEARCH.CLUSTERREFRESH` where needed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5be8b406e37c707e32f991472010ba122ec2d12d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/coord/rmr/redis_cluster.c
+++ b/src/coord/rmr/redis_cluster.c
@@ -51,14 +51,14 @@ static bool parseNode(RedisModuleCallReply *node, MRClusterNode *n) {
         host = rm_strndup(val_str, val_len);
       }
     } else if (STR_EQ(key_str, key_len, "tls-port")) {
-      int val = (int)RedisModule_CallReplyInteger(val);
-      if (val > 0) {
-        port = val; // Prefer tls-port if available (but only if valid)
+      int port_val = (int)RedisModule_CallReplyInteger(val);
+      if (port_val > 0) {
+        port = port_val; // Prefer tls-port if available (but only if valid)
       }
     } else if (STR_EQ(key_str, key_len, "port") && port == 0) {
-      int val = (int)RedisModule_CallReplyInteger(val);
-      if (val > 0) {
-        port = val; // Only set if tls-port wasn't set and port is valid
+      int port_val = (int)RedisModule_CallReplyInteger(val);
+      if (port_val > 0) {
+        port = port_val; // Only set if tls-port wasn't set and port is valid
       }
     }
   }


tests/pytests/test_issues.py
```diff
@@ -889,9 +889,10 @@ def mod5778_add_new_shard_to_cluster(env: Env):
     env.addShardToClusterIfExists()
     new_shard_conn = env.getConnection(shardId=initial_shards_count+1)
     verify_shard_init(new_shard_conn)
-    # Expect that the cluster will be aware of the new shard.
+    # Expect that the cluster will be aware of the new shard, but for redisearch coordinator, the new shard isn't
+    # considered part of the partition yet as it does not contain any slots.
     env.assertEqual(int(new_shard_conn.execute_command("cluster info")['cluster_known_nodes']), initial_shards_count+1)
-    env.assertEqual(new_shard_conn.execute_command("search.clusterinfo")[:2], ['num_partitions', int(initial_shards_count) + 1])
+    env.assertEqual(new_shard_conn.execute_command("search.clusterinfo")[:2], ['num_partitions', int(initial_shards_count)])
```
alonre24, line 895: So until now, we were always getting the new shard in a bad format? Isn't it time-dependent (flaky)?

src/coord/rmr/redis_cluster.c
```diff
@@ -40,23 +42,41 @@ static void parseNode(RedisModuleCallReply *node, MRClusterNode *n) {
...
     if (STR_EQ(key_str, key_len, "id")) {
       const char *val_str = RedisModule_CallReplyStringPtr(val, &val_len);
-      n->id = rm_strndup(val_str, val_len);
+      if (val_str && val_len == REDISMODULE_NODE_ID_LEN) {
+        id = rm_strndup(val_str, REDISMODULE_NODE_ID_LEN);
+      }
     } else if (STR_EQ(key_str, key_len, "endpoint")) {
       const char *val_str = RedisModule_CallReplyStringPtr(val, &val_len);
-      n->endpoint.host = rm_strndup(val_str, val_len);
+      if (val_str && !STR_EQ(val_str, val_len, "") && !STR_EQ(val_str, val_len, "?")) {
```
oshadmi, line 50: What about?
```suggestion
      if (val_str && val_len > 0 && !STR_EQ(val_str, val_len, "?")) {
```

tests/pytests/test_issues.py
GuyAv46: No, this was a change when we branched 8.4. We have now reverted this behavior and aligned again with previous versions.

src/coord/rmr/redis_cluster.c
GuyAv46: I think for readability,
```
(val_str && !STR_EQ(val_str, val_len, "") && !STR_EQ(val_str, val_len, "?")
```
is clearer - it shows exactly what values we call "invalid": `NULL`, `""`, and `"?"`.

The performance should remain the same (GCC and Clang optimize away `strlen("")` calls, and `STR_EQ` checks the length before the data comparison)

src/coord/rmr/redis_cluster.c
```diff
@@ -190,17 +218,29 @@ static MRClusterTopology *RedisCluster_GetTopology(RedisModuleCtx *ctx) {
...
     RedisModuleCallReply *nodes = RedisModule_CallReplyArrayElement(currShard, 3);
     RS_ASSERT(RedisModule_CallReplyType(nodes) == REDISMODULE_REPLY_ARRAY);
     // parse and store the master
-    parseMasterNode(nodes, &topo->shards[i].node);
+    if (!parseMasterNode(nodes, &topo->shards[topo->numShards].node)) {
+      rm_free(topo->shards[topo->numShards].slotRanges);
+      continue;
+    }
+    // Successfully parsed this shard
+    topo->numShards++;
```
oshadmi, line 233: Just confirming:
`topo->numShards` may be less than the actual allocated number of shards now, but allocated `shards` will get cleaned up in `MRClusterTopology_Free` just fine.

src/coord/rmr/rmr.c
```diff
@@ -417,14 +417,22 @@ void MR_ReplyClusterInfo(RedisModuleCtx *ctx, MRClusterTopology *topo) {
...
     RedisModule_ReplyKV_Array(reply, "shards"); // >shards
     for (int i = 0; i < topo->numShards; i++) {
-      MRClusterNode *node = &topo->shards[i].node;
-      RedisModule_Reply_Map(reply); // >>(node)
+      RedisModule_Reply_Map(reply); // >>(shard)
+
+      // Same syntax as in CLUSTER SHARDS
+      RedisModule_ReplyKV_Array(reply, "slots"); // >>>slots
+      for (int r = 0; r < topo->shards[i].slotRanges->num_ranges; r++) {
+        RedisModule_Reply_LongLong(reply, topo->shards[i].slotRanges->ranges[r].start);
```
oshadmi, line 425: No "start" and "end" keys?
As in https://github.com/RediSearch/RediSearch/blob/2.10/coord/src/rmr/rmr.c#L358

GuyAv46: I prefer the simple dump, I don't think it helps to see all the `start` and `end` when you have more than a single range

GuyAv46: And, as the comment states, this is the syntax of `CLUSTER SHARDS` https://redis.io/docs/latest/commands/cluster-shards/

PR #7378, opened 2025-11-16
MOD-12444 run on cached containers

meiravgri:
### **Motivation**
Reduce flaky CI failures caused by unreliable dependency installations (apt-get, pip, cargo) by pre-building Docker containers with all dependencies installed.
  
- **Cache shared across:** Sequential runs of the same container + architecture (e.g., `ubuntu:noble` on x86)

### **When We Rebuild**
- **Cache hit:** When `inputs.container` is specified AND none of these files changed:
  - `.install/**` (installation scripts)
  - `Dockerfile`
  - `pyproject.toml` / `uv.lock` (Python dependencies)
  - `build.sh` (Rust nightly version)
- Any of the above files changed
- First run (no cache exists yet)
- Different architecture (x86 vs ARM have separate caches)

### **Why Different Cache Refs Per Architecture?**
x86 and ARM produce different binaries (different SHA256 hashes). Sharing the same cache ref causes cache thrashing - the second build overwrites the first's cache, forcing a full rebuild. Separate cache refs ensure both architectures can reuse their own cached layers.

### **Fallback Setup**
When `build-image` job fails or is skipped (no `inputs.container`):
- Falls back to traditional setup: runs `.install/install_script.sh` directly on the runner
- Ensures tests can still run even if Docker build fails

### **Fix HOME Directory**
GitHub Actions [overrides `$HOME` in containers](https://github.com/actions/runner/issues/863), breaking Rust/Python tool paths. We detect and fix this by reading the actual home directory from `/etc/passwd` and setting `$HOME` correctly before running tests.

### **Re-enable Rocky9**
🤞 It will be stable

---

> [!NOTE]
> Builds and caches per-container/arch Docker images for CI and uses them in tests with fallback to legacy setup; splits Rust/Python test deps, tweaks installers, and tightens docker context.
> 
> - **CI Workflow (`.github/workflows/task-test.yml`)**:
>   - Add `build-image` job to build and push cached Docker images to GHCR (per container, sanitizer, env, arch); use as `container` for tests.
>   - Fallback to on-runner installation when image build is skipped/fails; conditionally run setup steps.
>   - Fix `$HOME` inside containers to expose Rust/Python toolchains; add debug for image outputs.
>   - Separate setup of Rust and Python test deps; minor quoting/format fixes.
> - **Docker**:
>   - New `Dockerfile` building from dynamic `BASE_IMAGE`, installs project deps via `.install/*`, sets PATH for Rust/Python.
> - **Install scripts (`.install/*`)**:
>   - Split test deps into `install_rust_deps.sh` and `install_python_deps.sh`; `common_installations.sh` delegates to them.
>   - Add required packages/utilities across distros (e.g., bash, tar/gzip, gcompat libs on Alpine); improve `uv` install PATH handling; add rustup/cargo debug output.
> - **.dockerignore**:
>   - Ignore everything except installation scripts and essentials to minimize build context; exclude `.install/boost/*`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b138fb9dc891ed909c119fef485aecd54ae575b4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,6 +4,7 @@
 !.install
 !.install/**
 !build.sh
+!.github/workflows/task-get-config.yml
 
 # Make sure we don't include the boost directory in the context.
 .install/boost/*


.dockerignore
```diff
@@ -1,16 +1,9 @@
...
-/site/
-/deps/RedisJSON/
-/tests/pytests/logs/
-/1/
+# Ignore everything, except the scripts required to install dependencies
+# into our CI containers.
+*
+!.install
+!.install/**
+!build.sh
```
LukeMathWalker, line 6: One improvement I wanted to make: extract the nightly version into its own file, thus avoiding cache invalidation if unrelated things change in build.sh

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+# Build argument to select the base image dynamically.
+# Examples:
+#   docker build --build-arg BASE_IMAGE=ubuntu:24.04 .
+#   docker build --build-arg BASE_IMAGE=rockylinux:9 .
+#   docker build --build-arg BASE_IMAGE=alpine:3 .
+# ---------------------------------------------------------
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+
+ENV HOME=/root
```
LukeMathWalker: I do wonder if this is necessary after having changed the workflow to "reset" $HOME for containers. My gut feeling is that it should work with the default home for the underlying base image (which is often going to be /root anyway though).

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+WORKDIR /project/.install
+RUN bash -l -eo pipefail install_script.sh
+WORKDIR /project
+RUN bash -l -eo pipefail .install/test_deps/install_rust_deps.sh
+# Expose newly-installed Rust and Python tools via PATH
+ENV PATH="/root/.cargo/bin:/root/.local/bin:${PATH}"
+
+WORKDIR /project
+RUN which uv
```
LukeMathWalker: This was a debugging statement, we can probably remove it.

LukeMathWalker:
Looking good! 
Left a couple of comments on minor things.

We should also consider the following:
- Set `RUSTUP_AUTO_INSTALL` to `0` in our Docker container to ensure we don't end up installing new toolchain versions later ([reference](https://rust-lang.github.io/rustup/environment-variables.html))
- Add the [`rust-cache`](https://github.com/Swatinem/rust-cache) as a step to our test workflow to cache Rust packages on top of caching Rust toolchains
- Copy [`rust-toolchain.toml`](https://github.com/RediSearch/RediSearch/blob/master/src/redisearch_rs/rust-toolchain.toml) into our Docker container, changing it from `stable` to a pinned stable version (1.91.1) to ensure we bust the cache when we upgrade the Rust stable toolchain. We can/should move that file to the top-level directory.
- In the Dockerfile, we should run `uv sync --locked --no-install-workspace` to download and install all third-party Python dependencies. This'll require copying over `uv.lock` and _maybe_ `pyproject.toml`.

Peter-Sh:
Hi @LukeMathWalker!

I was asked by @alonre24 to add some feedback.

Great work, by the way! 

One thought so far:

Maybe consider building the images on a schedule instead of during each run?
This could save some time and - more importantly - simplify the `common-flow` [conditions](https://github.com/RediSearch/RediSearch/pull/7378/files#diff-e572bdb8510685b825e2671ebe24897798eb1fd8e42bdf4ecd3690c97154c0a2R114).

In my opinion, the combination of `!cancelled()` and other expressions is clever, but it might be a bit hard to maintain and somewhat error-prone. If I understand correctly, the `common-flow` will still attempt to run even if the `build-image` job fails due to this condition.

Building a set of images in a schedule (e.g. nightly) using the same `task-get-linux-configurations.yml` reusable workflow - and simply relying on existing images in GHCR - could make things simpler.

Just my two cents! :)

LukeMathWalker:
> Hi @LukeMathWalker!
> 
> I was asked by @alonre24 to add some feedback.
> 
> Great work, by the way! 

Thanks!
>

> One thought so far:
> 
> Maybe consider building the images on a schedule instead of during each run?

I considered the option, but it would make cache invalidation/freshness more cumbersome to manage.
In particular, PRs may touch files that in turn would require a new Docker build (e.g. change a Python dependency, update Rust version, etc). Relying on GHA docker caching makes "building every time" fast in the happy case (~40s) and correct if indeed rebuilding is required.

> This could save some time and - more importantly - simplify the `common-flow` [conditions](https://github.com/RediSearch/RediSearch/pull/7378/files#diff-e572bdb8510685b825e2671ebe24897798eb1fd8e42bdf4ecd3690c97154c0a2R114).
> 
> In my opinion, the combination of `!cancelled()` and other expressions is clever, but it might be a bit hard to maintain and somewhat error-prone. If I understand correctly, the `common-flow` will still attempt to run even if the `build-image` job fails due to this condition.

I agree, supporting both container and container-less builds in the same flow adds a fair amount of trickiness.

> Just my two cents! :)

Thanks for having a look in the first place!

Peter-Sh:
> I considered the option, but it would make cache invalidation/freshness more cumbersome to manage. In particular, PRs may touch files that in turn would require a new Docker build (e.g. change a Python dependency, update Rust version, etc). Relying on GHA docker caching makes "building every time" fast in the happy case (~40s) and correct if indeed rebuilding is required.

Got it, makes sense. We used to resolve the similar problem by calculating hash of all the installation scripts and use it as a part of the image TAG, but I don't think it's worth it and it definitely won't make things simpler.

meiravgri:
@Peter-Sh 

> In my opinion, the combination of !cancelled() and other expressions is clever, but it might be a bit hard to maintain and somewhat error-prone. If I understand correctly, the common-flow will still attempt to run even if the build-image job fails due to this condition.

It is intentional - we don't want to fail the entire flow, but have a fallback

Ass for nightly builds - not sure I understand how it simplifies the workflow considering that we need to cover cases where `task-test.yml` runs in a non-container env.

.dockerignore
meiravgri: We change `build.sh` very rarely so we can be strict about it

Dockerfile
meiravgri: trying
https://github.com/RediSearch/RediSearch/actions/runs/19437696106

meiravgri: worked

meiravgri:
@Peter-Sh @LukeMathWalker 
are we good to merge this one?

Peter-Sh:
> It is intentional - we don't want to fail the entire flow, but have a fallback

Yes, if `build-image` job is skipped then `common-flow` should work for non-docker tests. But if `build-image` is started and it fails the `common-flow` job will still try to run (because of `!cancelled()`) and the `common-flow` job will fail because of missing image.
Not that this is a problem by itself, but it's a sign that this place having somewhat unexpected behavior could be potential source of problems in the future when making modifications.

> Ass for nightly builds - not sure I understand how it simplifies the workflow considering that we need to cover cases where task-test.yml runs in a non-container env.

Regarding the nightly and schedule @LukeMathWalker  explained me that it's not the way for you. Another idea it simplifying container vs non-container flows with the help of reusable actions.

This allows to avoid tricky conditions involving `cancelled()` or `always()` and IMO makes the dependencies more clear, but would require an additional work to move common-flow to reusable action (not reusable workflow).
```

jobs:
  build-image:
   if: inputs.container
   .... build the image ...
  
  # split container and non-container jobs
  common-flow-container:
    needs: build-image
    if: inputs.container && (inputs.standalone || inputs.coordinator)
    container:
      image: ${{ needs..... }}
    steps:
      - name: Run tests
        uses: .github/actions/run-tests
        with:
          standalone: ${{ inputs.standalone }}
          ....

  common-flow-host:
    if: inputs.container == '' && (inputs.standalone || inputs.coordinator)
    steps:
      - name: Run tests
        uses: .github/actions/run-tests
        with:
          standalone: ${{ inputs.standalone }}
          ....

```

@meiravgri 
>  are we good to merge this one?

I don't see any problems, the above is just a side view that you may (or may not) consider useful, as you have your own approach and much deeper domain knowledge.

LukeMathWalker:
No objections from me.

.github/workflows/task-test.yml
```diff
@@ -29,39 +29,112 @@ on:
...
+            exit 0
+          fi
+          echo "HOME=$HOME was broken. Setting it to $h"
+          ls -ld $HOME
+          ls -ld $h
+          echo "USER: $USER"
+          echo "id: $(id)"
+          echo "HOME=$h" >> $GITHUB_ENV
       # Split to alpine and non-alpine due to different default shells, once the dependency installation is done, we can use the same shell in the rest of the flow
       - name: Pre-steps Dependencies (Alpine)
```
alonre24: Can we skip that as well if `build-image.result == 'success'` ?

meiravgri: i guess we can
i think this pr will be kicked out of the queue any due to conflicts with Johnson PR 😅 let's wait

meiravgri:
@LukeMathWalker @Peter-Sh 
i encountered this error:
https://github.com/RediSearch/RediSearch/actions/runs/19491146538/job/55783408377
Copilot says something about docker hub rate limits that can be pypassed by using docker hub token
Does it make sense?
@Peter-Sh  Is there a company dockerhub token i can use?

```
This job is failing due to a Docker authentication error:

Code
ERROR: Error response from daemon: unauthorized: authentication required
This occurs when trying to pull the Docker image moby/buildkit:buildx-stable-1 during the booting process of BuildKit. The job definition and logs indicate that the workflow attempts to use Docker Buildx without logging in to Docker Hub or the registry. This wasn't required before because public pulls from Docker Hub normally work, but Docker occasionally enforces rate limits or changes access requirements, causing previously reliable jobs to fail unexpectedly ("flaky" behavior).

Why is it flaky / why now?

Docker Hub enforces rate limits for unauthenticated users and may intermittently require authentication due to increased usage or recent policy changes.
If the runner’s IP is shared (like on GitHub-hosted runners), quota is reached faster, causing random failures.
This is not a permanent failure—sometimes it works if the rate limit isn't reached, which is why it appears flaky.
Solution: To resolve this, add a step to log in to Docker Hub before any buildx operations. Use the docker/login-action and store your Docker credentials as secrets in the repository:

YAML
- name: Login to DockerHub
  uses: docker/login-action@v3
  with:
    username: ${{ secrets.DOCKERHUB_USERNAME }}
    password: ${{ secrets.DOCKERHUB_TOKEN }}
Add this step early in your workflow file (.github/workflows/flow-linux-platforms.yml, ref: 32c83ced28fc643d555aaf34237ddeab2bf6041a), ideally before booting BuildKit or running any Docker commands.

Summary:
This flakiness is due to Docker Hub's authentication/rate limit policy changes. Adding an authentication step to your GitHub Actions workflow should make the job reliable again.

```

LukeMathWalker:
> @LukeMathWalker @Peter-Sh 
> i encountered this error:
> https://github.com/RediSearch/RediSearch/actions/runs/19491146538/job/55783408377
> Copilot says something about docker hub rate limits that can be pypassed by using docker hub token
> Does it make sense?
> @Peter-Sh  Is there a company dockerhub token i can use?
> 
> ```
> This job is failing due to a Docker authentication error:
> 
> Code
> ERROR: Error response from daemon: unauthorized: authentication required
> This occurs when trying to pull the Docker image moby/buildkit:buildx-stable-1 during the booting process of BuildKit. The job definition and logs indicate that the workflow attempts to use Docker Buildx without logging in to Docker Hub or the registry. This wasn't required before because public pulls from Docker Hub normally work, but Docker occasionally enforces rate limits or changes access requirements, causing previously reliable jobs to fail unexpectedly ("flaky" behavior).
> 
> Why is it flaky / why now?
> 
> Docker Hub enforces rate limits for unauthenticated users and may intermittently require authentication due to increased usage or recent policy changes.
> If the runner’s IP is shared (like on GitHub-hosted runners), quota is reached faster, causing random failures.
> This is not a permanent failure—sometimes it works if the rate limit isn't reached, which is why it appears flaky.
> Solution: To resolve this, add a step to log in to Docker Hub before any buildx operations. Use the docker/login-action and store your Docker credentials as secrets in the repository:
> 
> YAML
> - name: Login to DockerHub
>   uses: docker/login-action@v3
>   with:
>     username: ${{ secrets.DOCKERHUB_USERNAME }}
>     password: ${{ secrets.DOCKERHUB_TOKEN }}
> Add this step early in your workflow file (.github/workflows/flow-linux-platforms.yml, ref: 32c83ced28fc643d555aaf34237ddeab2bf6041a), ideally before booting BuildKit or running any Docker commands.
> 
> Summary:
> This flakiness is due to Docker Hub's authentication/rate limit policy changes. Adding an authentication step to your GitHub Actions workflow should make the job reliable again.
> 
> ```

Yes, it makes sense, it was introduced a few years ago. 
The simplest workaround is a Docker Hub token. Alternatively, I think we can publish the base images we need to our GitHub registry with a cronjob and thus bypass DockerHub entirely in all the other workflows.

meiravgri:
@LukeMathWalker @alonre24 had conflicts with master, need re approval 🙏

LukeMathWalker (approved):

PR #7379, opened 2025-11-16
[CI] Update RLTest version

GuyAv46:
> [!NOTE]
> Pin redis and bump RLTest for pytests; refactor tests to use env.waitCluster() and adjust TLS-related skips.
> 
> - **Dependencies**:
>   - Pin `redis` to `>=5.2.1,<7.0.0` and bump `RLTest` to `>=0.7.18,<0.8.0` in `tests/pytests/pyproject.toml`.
> - **Tests**:
>   - Replace manual cluster stabilization loops and `SEARCH.CLUSTERREFRESH` calls with `env.waitCluster()` (e.g., after topology changes, TLS config updates, and shard additions) in `tests/pytests/test.py`, `tests/pytests/test_asm.py`, and `tests/pytests/test_issues.py`.
>   - Update skips to disable certain TLS tests on macOS/ASAN.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 02a1ec04c054661f33776ac2ff99368249c3db88. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test.py
+++ b/tests/pytests/test.py
@@ -4520,7 +4520,8 @@ def test_with_tls():
 
     common_with_auth(env)
 
-@skip(cluster=False)
+# TODO: enable macos+san once https://redislabs.atlassian.net/browse/RED-176581 is fixed
+@skip(cluster=False, macos=True, asan=True)
 def test_with_tls_and_non_tls_ports():
     """Tests that the coordinator-shard connections are using the correct
     protocol (TLS vs. non-TLS) according to the redis `tls-cluster` configuration."""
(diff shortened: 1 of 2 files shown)


tests/pytests/pyproject.toml
```diff
@@ -7,7 +7,7 @@ dependencies = [
     "packaging<=24.2",
     "deepdiff<=8.3.0",
     "redis>=5.2.1",
-    "RLTest<=0.7.16",
```
alonre24, line 10: Why not `RLTest==0.7.18` ? what if someone breaks something in 0.7.19 ?

GuyAv46: Because then we will need to update all the branches with every active fix. We already missed 0.7.17 completely.
We shouldn't break patch releases. Did it ever happen?

alonre24: Not intentionally, obviously, but I believe it happened in the past. 
Essentially, you are right, but we decided to be conservative and pin the versions at the expense of auto updates. We can claim that in RLTest, we should take a different appraoch since we are maintaining it (though it is shared across all modules).

GuyAv46: Actually, don't we need to run `uv lock` to refresh the version? Even if we allow higher versions, UV has chosen 0.7.18 specifically

alonre24: Good point. I don't know, but if so, all those restrictions are redundant

alonre24 (dismissed):

alonre24 (dismissed):

PR #7386, opened 2025-11-17
[MOD-12416] Create basis for tracking Query Errors and track syntax and args errors

lerman25:
First PR from a series of PRs, aiming to extend the current info metric tracked in **RediSearch**.
This PR creates the basis for tracking specific query error, starting with `QUERY_ERROR_CODE_ADD_ARGS` and `QUERY_ERROR_CODE_PARSE_ARGS`.
It introduces `QueryErrorsGlobalStats` struct, which keep counters for each error we want to track, and
introduces `QueryErrorsGlobalStats_UpdateError` function, which receives `QueryErrorCode`, and update the relevant counter (if the code is tracked). 
In cluster setting,  some args error might be detected in the **coord**, while  others in the **shards**, so this PR creates individual counter for coord and SA/shards. This separation might  also help debug which specific error case was triggered.

---

> [!NOTE]
> Adds global query error counters (syntax/args) split by shard vs coordinator, wires them into SEARCH/AGGREGATE/HYBRID error paths, and surfaces them in INFO MODULES with tests.
> 
> - **Stats/Infra**:
>   - Introduce `QueryErrorsGlobalStats` and `QueryErrorsGlobalStats_UpdateError(...)` to count query errors by type (`syntax`, `parse_args`) and origin (shard vs coordinator).
>   - Extend `QueriesGlobalStats`/`TotalGlobalStats_GetQueryStats()` to include these counters.
> - **Error accounting hooks**:
>   - Update error paths in `aggregate_exec.c`, `dist_aggregate.c`, `dist_hybrid.c`, `hybrid_exec.c`, and `module.c` to increment the appropriate shard/coordinator counters before replying.
> - **INFO exposure**:
>   - Add new `INFO MODULES` fields under `warnings_and_errors` and `coordinator_warnings_and_errors`:
>     - `shard_total_query_errors_syntax`, `shard_total_query_errors_arguments`
>     - `coord_total_query_errors_syntax`, `coord_total_query_errors_arguments`
> - **Tests**:
>   - Add comprehensive SA/cluster tests verifying counters for `FT.SEARCH`, `FT.AGGREGATE`, and `FT.HYBRID` across syntax/args errors and that successful queries don’t change metrics.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f998a91cb4e63c244eef3d3599aa2577fa3df25. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_info_modules.py
+++ b/tests/pytests/test_info_modules.py
@@ -620,6 +620,15 @@ SEARCH_COORD_PREFIX = 'search_coord_'
 SYNTAX_ERROR_COORD_METRIC = f"{SEARCH_COORD_PREFIX}total_query_errors_syntax"
 ARGS_ERROR_COORD_METRIC = f"{SEARCH_COORD_PREFIX}total_query_errors_arguments"
 
+# Expect env and conn so we can assert
+def _verify_metrics_not_changed(env, conn, prev_info_dict: dict, ignored_metrics : list):
+  info_dict = info_modules_to_dict(conn)
+  for section in [WARN_ERR_SECTION, COORD_WARN_ERR_SECTION]:
+    for metric in info_dict[section]:
+      if metric in ignored_metrics:
+        continue
+      env.assertEqual(info_dict[section][metric], prev_info_dict[section][metric], message = f"Metric {metric} changed")
+
 def _common_warnings_errors_test_scenario(env):
   """Common setup for warnings and errors tests"""
   # Create index
@@ -640,7 +649,7 @@ class testWarningsAndErrorsStandalone:
     _common_warnings_errors_test_scenario(self.env)
     self.prev_info_dict = info_modules_to_dict(self.env)
 
-  def setup(self):
+  def setUp(self):
     self.prev_info_dict = info_modules_to_dict(self.env)
 
   def test_syntax_errors_SA(self):
@@ -650,21 +659,24 @@ class testWarningsAndErrorsStandalone:
     self.env.expect('FT.SEARCH', 'idx', 'hello world:').error().contains('Syntax error at offset')
     # Test counter
     info_dict = info_modules_to_dict(self.env)
-    prev_syntax_error_count = int(self.prev_info_dict[WARN_ERR_SECTION][SYNTAX_ERROR_SHARD_METRIC])
-    syntax_error_count = int(info_dict[WARN_ERR_SECTION][SYNTAX_ERROR_SHARD_METRIC])
-    self.env.assertEqual(syntax_error_count, prev_syntax_error_count + 1)
+    syntax_error_count = info_dict[WARN_ERR_SECTION][SYNTAX_ERROR_SHARD_METRIC]
+    self.env.assertEqual(syntax_error_count, '1')
     # Test syntax errors in aggregate
     self.env.expect('FT.AGGREGATE', 'idx', 'hello world:').error().contains('Syntax error at offset')
     # Test counter
     info_dict = info_modules_to_dict(self.env)
-    syntax_error_count = int(info_dict[WARN_ERR_SECTION][SYNTAX_ERROR_SHARD_METRIC])
-    self.env.assertEqual(syntax_error_count, prev_syntax_error_count + 2)
+    syntax_error_count = info_dict[WARN_ERR_SECTION][SYNTAX_ERROR_SHARD_METRIC]
+    self.env.assertEqual(syntax_error_count, '2')
     # Test syntax errors in hybrid
     self.env.expect('FT.HYBRID', 'idx_vec', 'SEARCH', 'hello world:', 'VSIM', '@vector', '0').error().contains('Syntax error at offset')
     # Test counter
     info_dict = info_modules_to_dict(self.env)
-    syntax_error_count = int(info_dict[WARN_ERR_SECTION][SYNTAX_ERROR_SHARD_METRIC])
-    self.env.assertEqual(syntax_error_count, prev_syntax_error_count + 3)
+    syntax_error_count = info_dict[WARN_ERR_SECTION][SYNTAX_ERROR_SHARD_METRIC]
+    self.env.assertEqual(syntax_error_count, '3')
+
+    # Test other metrics not changed
+    tested_in_this_test = [SYNTAX_ERROR_SHARD_METRIC]
+    _verify_metrics_not_changed(self.env, self.env, self.prev_info_dict, tested_in_this_test)
 
   def test_args_errors_SA(self):
     # Standalone shards are considered as shards in the info metrics
@@ -673,21 +685,24 @@ class testWarningsAndErrorsStandalone:
     self.env.expect('FT.SEARCH', 'idx', 'hello world', 'LIMIT', 0, 0, 'MEOW').error().contains('Unknown argument')
     # Test counter
     info_dict = info_modules_to_dict(self.env)
-    prev_args_error_count = int(self.prev_info_dict[WARN_ERR_SECTION][ARGS_ERROR_SHARD_METRIC])
-    args_error_count = int(info_dict[WARN_ERR_SECTION][ARGS_ERROR_SHARD_METRIC])
-    self.env.assertEqual(args_error_count, prev_args_error_count + 1)
+    args_error_count = info_dict[WARN_ERR_SECTION][ARGS_ERROR_SHARD_METRIC]
+    self.env.assertEqual(args_error_count, '1')
     # Test args errors in aggregate
     self.env.expect('FT.AGGREGATE', 'idx', 'hello world', 'LIMIT', 0, 0, 'MEOW').error().contains('Unknown argument')
     # Test counter
     info_dict = info_modules_to_dict(self.env)
-    args_error_count = int(info_dict[WARN_ERR_SECTION][ARGS_ERROR_SHARD_METRIC])
-    self.env.assertEqual(args_error_count, prev_args_error_count + 2)
+    args_error_count = info_dict[WARN_ERR_SECTION][ARGS_ERROR_SHARD_METRIC]
+    self.env.assertEqual(args_error_count, '2')
     # Test args errors in hybrid
     self.env.expect('FT.HYBRID', 'idx_vec', 'SEARCH', 'hello world', 'VSIM', '@vector', '0', 'LIMIT', 0, 0, 'MEOW').error().contains('Unknown argument')
     # Test counter
     info_dict = info_modules_to_dict(self.env)
-    args_error_count = int(info_dict[WARN_ERR_SECTION][ARGS_ERROR_SHARD_METRIC])
-    self.env.assertEqual(args_error_count, prev_args_error_count + 3)
+    args_error_count = info_dict[WARN_ERR_SECTION][ARGS_ERROR_SHARD_METRIC]
+    self.env.assertEqual(args_error_count, '3')
+
+    # Test other metrics not changed
+    tested_in_this_test = [ARGS_ERROR_SHARD_METRIC]
+    _verify_metrics_not_changed(self.env, self.env, self.prev_info_dict, tested_in_this_test)
 
   def test_no_error_queries_SA(self):
     # Standalone shards are considered as coordinator in the info metrics
@@ -731,21 +746,28 @@ class testWarningsAndErrorsCluster:
 
   def __init__(self):
     skipTest(cluster=False)
-    self.env = Env(shardsCount=3)
+    self.env = Env()
     _common_warnings_errors_cluster_test_scenario(self.env)
+    self.shards_prev_info_dict = {}
+    self.coord_prev_info_dict = info_modules_to_dict(self.env)
     # Init all shards
-    for i in range(self.env.shardsCount):
+    for i in range(1, self.env.shardsCount + 1):
       verify_shard_init(self.env.getConnection(i))
+      self.shards_prev_info_dict[i] = info_modules_to_dict(self.env.getConnection(i))
+
+  def setUp(self):
+    self.coord_prev_info_dict = info_modules_to_dict(self.env)
+    # Init all shards
+    for i in range(1, self.env.shardsCount + 1):
+      self.shards_prev_info_dict[i] = info_modules_to_dict(self.env.getConnection(i))
+
+  def _verify_metrics_not_changes_all_shards(self, ignored_metrics : list):
+    # Verify shards (coord is one of the shards as well)
+    for shardId in range(1, self.env.shardsCount + 1):
+      _verify_metrics_not_changed(self.env, self.env.getConnection(shardId), self.shards_prev_info_dict[shardId], ignored_metrics)
 
   def test_syntax_errors_cluster(self):
     # In cluster mode, syntax errors are only tracked at shard level
-
-    # Check syntax error metric before adding any errors on each shard
-    for shardId in range(1, self.env.shardsCount + 1):
-      shard_conn = self.env.getConnection(shardId)
-      info_dict = info_modules_to_dict(shard_conn)
-      syntax_error_count = info_dict[WARN_ERR_SECTION][SYNTAX_ERROR_SHARD_METRIC]
-      self.env.assertEqual(syntax_error_count, '0', message=f"Shard {shardId} has wrong initial syntax error count")
 
     # Test syntax errors for shard level syntax error
     self.env.expect('FT.SEARCH', 'idx', 'hello world:').error().contains('Syntax error at offset')
@@ -789,6 +811,10 @@ class testWarningsAndErrorsCluster:
     coord_syntax_error_count = info_dict[COORD_WARN_ERR_SECTION][SYNTAX_ERROR_COORD_METRIC]
     self.env.assertEqual(coord_syntax_error_count, '1')
 
+    # Test other metrics not changed
+    tested_in_this_test = [SYNTAX_ERROR_SHARD_METRIC, SYNTAX_ERROR_COORD_METRIC]
+    self._verify_metrics_not_changes_all_shards(tested_in_this_test)
+
   def test_args_errors_cluster(self):
 
     # Check args error metric before adding any errors on each shard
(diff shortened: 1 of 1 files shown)


src/aggregate/aggregate_exec.c
```diff
@@ -1098,9 +1098,14 @@ static int execCommandCommon(RedisModuleCtx *ctx, RedisModuleString **argv, int
   return REDISMODULE_OK;
 
 error:
+  // Update global query errors statistics before freeing the request.
+  // Both SA and internal are considered shards.
```
meiravgri: What's the difference between SA and internal? we don't get here if we have more than 1 shard

src/hybrid/hybrid_exec.c
```diff
@@ -562,6 +562,8 @@ static inline void DefaultCleanup(StrongRef hybrid_ref) {
 static inline int CleanupAndReplyStatus(RedisModuleCtx *ctx, StrongRef hybrid_ref, HybridPipelineParams *hybridParams, QueryError *status) {
     freeHybridParams(hybridParams);
     DefaultCleanup(hybrid_ref);
+    // Update global query errors, this path is only used for SA and internal, both are considered shards.
```
meiravgri: same question regarding SA and internal

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +602,282 @@ def test_indexing_metrics(env: Env):
...
+def _common_warnings_errors_test_scenario(env):
+  """Common setup for warnings and errors tests"""
+  # Create index
+  env.expect('FT.CREATE', 'idx', 'SCHEMA', 'text', 'TEXT').ok()
+  # Create doc
+  env.expect('HSET', 'doc:1', 'text', 'hello world').equal(1)
+  # Create vector index for hybrid
+  env.expect('FT.CREATE', 'idx_vec', 'PREFIX', '1', 'vec:', 'SCHEMA', 'vector', 'VECTOR', 'FLAT', '6', 'TYPE', 'FLOAT32', 'DIM', '2', 'DISTANCE_METRIC', 'L2').ok()
+  # Create doc for hybrid
+  env.expect('HSET', 'vec:1', 'vector', '0,0').equal(1)
```
meiravgri: this is not a vector of dim =2

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +602,282 @@ def test_indexing_metrics(env: Env):
...
+    self.prev_info_dict = info_modules_to_dict(self.env)
+
+  def setup(self):
+    self.prev_info_dict = info_modules_to_dict(self.env)
+
+  def test_syntax_errors_SA(self):
+    # Standalone shards are considered as shards in the info metrics
+
+    # Test syntax errors
+    self.env.expect('FT.SEARCH', 'idx', 'hello world:').error().contains('Syntax error at offset')
```
meiravgri, line 659: why is it a syntax error :sweat:

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +602,282 @@ def test_indexing_metrics(env: Env):
...
+  env.expect('HSET', 'vec:1', 'vector', '0,0').equal(1)
+
+class testWarningsAndErrorsStandalone:
+  """Test class for warnings and errors metrics in standalone mode"""
+
+  def __init__(self):
+    skipTest(cluster=True)
+    self.env = Env()
+    _common_warnings_errors_test_scenario(self.env)
+    self.prev_info_dict = info_modules_to_dict(self.env)
```
meiravgri, line 650: lets verify all fields are initialized

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +602,282 @@ def test_indexing_metrics(env: Env):
...
+    # Test counter
+    info_dict = info_modules_to_dict(self.env)
+    syntax_error_count = int(info_dict[WARN_ERR_SECTION][SYNTAX_ERROR_SHARD_METRIC])
+    self.env.assertEqual(syntax_error_count, prev_syntax_error_count + 3)
+
+  def test_args_errors_SA(self):
+    # Standalone shards are considered as shards in the info metrics
+
+    # Test args errors
+    self.env.expect('FT.SEARCH', 'idx', 'hello world', 'LIMIT', 0, 0, 'MEOW').error().contains('Unknown argument')
```
meiravgri, line 685: i am proud

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +602,282 @@ def test_indexing_metrics(env: Env):
...
+class testWarningsAndErrorsCluster:
+  """Test class for warnings and errors metrics in cluster mode with RESP2"""
+
+  def __init__(self):
+    skipTest(cluster=False)
+    self.env = Env(shardsCount=3)
+    _common_warnings_errors_cluster_test_scenario(self.env)
+    # Init all shards
+    for i in range(self.env.shardsCount):
+      verify_shard_init(self.env.getConnection(i))
```
meiravgri: why is this needed?

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +602,282 @@ def test_indexing_metrics(env: Env):
...
+    for shardId in range(1, self.env.shardsCount + 1):
+      shard_conn = self.env.getConnection(shardId)
+      info_dict = info_modules_to_dict(shard_conn)
+      syntax_error_count = info_dict[WARN_ERR_SECTION][SYNTAX_ERROR_SHARD_METRIC]
+      self.env.assertEqual(syntax_error_count, '1', message=f"Shard {shardId} has wrong syntax error count")
+    # Check coord metric unchanged
+    info_dict = info_modules_to_dict(self.env)
+    coord_syntax_error_count = info_dict[COORD_WARN_ERR_SECTION][SYNTAX_ERROR_COORD_METRIC]
+    self.env.assertEqual(coord_syntax_error_count, '0')
+    # Syntax error in FT.SEARCH are not checked on the coordinator
```
meiravgri: move comment before the assertion

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +602,282 @@ def test_indexing_metrics(env: Env):
...
+      self.env.assertEqual(syntax_error_count, '2', message=f"Shard {shardId} has wrong syntax error count")
+    # Check coord metric unchanged
+    info_dict = info_modules_to_dict(self.env)
+    coord_syntax_error_count = info_dict[COORD_WARN_ERR_SECTION][SYNTAX_ERROR_COORD_METRIC]
+    self.env.assertEqual(coord_syntax_error_count, '0')
+    # Syntax error in FT.AGGREGATE are not checked on the coordinator
+
+
+    # Test syntax errors in hybrid
+    # Syntax errors in the hybrid command are only counted on the coordinator.
```
meiravgri, line 801: is it a bug or a feature

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +602,282 @@ def test_indexing_metrics(env: Env):
...
+  """Common setup for warnings and errors cluster tests"""
+  # Create index
+  env.expect('FT.CREATE', 'idx', 'SCHEMA', 'text', 'TEXT').ok()
+  # Create doc
+  conn = getConnectionByEnv(env)
+  conn.execute_command('HSET', 'doc:1', 'text', 'hello world')
+  # Create vector index for hybrid
+  env.expect('FT.CREATE', 'idx_vec', 'PREFIX', '1', 'vec:', 'SCHEMA', 'vector', 'VECTOR', 'FLAT', '6', 'TYPE', 'FLOAT32', 'DIM', '2', 'DISTANCE_METRIC', 'L2').ok()
+  # Create doc for hybrid
+  conn.execute_command('HSET', 'vec:1', 'vector', '0,0')
```
meiravgri: same

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +602,282 @@ def test_indexing_metrics(env: Env):
...
+  env.expect('FT.CREATE', 'idx_vec', 'PREFIX', '1', 'vec:', 'SCHEMA', 'vector', 'VECTOR', 'FLAT', '6', 'TYPE', 'FLOAT32', 'DIM', '2', 'DISTANCE_METRIC', 'L2').ok()
+  # Create doc for hybrid
+  env.expect('HSET', 'vec:1', 'vector', '0,0').equal(1)
+
+class testWarningsAndErrorsStandalone:
+  """Test class for warnings and errors metrics in standalone mode"""
+
+  def __init__(self):
+    skipTest(cluster=True)
+    self.env = Env()
```
meiravgri, line 648: 1. do we need to setup env here? isnt it done automatically?
2. don't we need dialect 2 for hybrid?

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +602,282 @@ def test_indexing_metrics(env: Env):
...
+      shard_conn = self.env.getConnection(shardId)
+      info_dict = info_modules_to_dict(shard_conn)
+      args_error_count = info_dict[WARN_ERR_SECTION][ARGS_ERROR_SHARD_METRIC]
+      self.env.assertEqual(args_error_count, '1', message=f"Shard {shardId} has wrong args error count")
+    # Check coord metric
+    info_dict = info_modules_to_dict(self.env)
+    coord_args_error_count = info_dict[COORD_WARN_ERR_SECTION][ARGS_ERROR_COORD_METRIC]
+    self.env.assertEqual(coord_args_error_count, '2')
+
+  def test_no_error_queries_SA(self):
```
meiravgri:
```suggestion
  def test_no_error_queries(self):
```

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +602,282 @@ def test_indexing_metrics(env: Env):
...
+
+  def test_args_errors_cluster(self):
+    # In cluster mode, args errors are only tracked at shard level
+
+    # Check args error metric before adding any errors on each shard
+    for shardId in range(1, self.env.shardsCount + 1):
+      shard_conn = self.env.getConnection(shardId)
+      info_dict = info_modules_to_dict(shard_conn)
+      args_error_count = info_dict[WARN_ERR_SECTION][ARGS_ERROR_SHARD_METRIC]
+      self.env.assertEqual(args_error_count, '0', message=f"Shard {shardId} has wrong initial args error count")
```
meiravgri, line 825: consider moving both ARGS_ERROR_SHARD_METRIC and SYNTAX_ERROR_SHARD_METRIC initialization to the __init__ function under the same loop

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +602,282 @@ def test_indexing_metrics(env: Env):
...
+      shard_conn = self.env.getConnection(shardId)
+      info_dict = info_modules_to_dict(shard_conn)
+      syntax_error_count = info_dict[WARN_ERR_SECTION][SYNTAX_ERROR_SHARD_METRIC]
+      self.env.assertEqual(syntax_error_count, '2', message=f"Shard {shardId} has wrong syntax error count")
+    # Check coord metric
+    info_dict = info_modules_to_dict(self.env)
+    coord_syntax_error_count = info_dict[COORD_WARN_ERR_SECTION][SYNTAX_ERROR_COORD_METRIC]
+    self.env.assertEqual(coord_syntax_error_count, '1')
+
+  def test_args_errors_cluster(self):
```
meiravgri, line 818: we should cover here also params error coming from the coordinator, for example invalid limit, for all query types (i think that the fact that FT.SEARCH with invalid limit arg is not a coord error is a bug. you can use a negative number for testing coordintor error)

src/info/global_stats.h
```diff
@@ -39,10 +43,19 @@ typedef struct {
   size_t numVectorFieldsSvsVamanaCompressed;
 } FieldsGlobalStats;
 
+typedef struct {
+  size_t syntax; // Number of syntax errors
```
meiravgri, line 47: in which scenario do we get syntax error from the coordinator?

tests/pytests/test_info_modules.py
meiravgri: also let's not limit the env to s specific number of shards if possible

tests/pytests/test_info_modules.py
lerman25: Usage of ":" without "[...]"

tests/pytests/test_info_modules.py
lerman25: What fields should be initialized?

tests/pytests/test_info_modules.py
lerman25: Some weird lazy initialization that is needed sometimes, without it I had error in a previous PR relating to OOM + Early bailout (maybe here it's not necessary at the moment, but will be)

tests/pytests/test_info_modules.py
lerman25: Design choice I guess, so feature

tests/pytests/test_info_modules.py
lerman25: 1. Every test class that I've seen does that, so I think it's necessary 
2. Is it not the default?

tests/pytests/test_info_modules.py
lerman25: Good catch, to be aligned with other tests, I changed to `_cluster`

src/info/global_stats.h
lerman25: Hybrid sadly :-(

tests/pytests/test_info_modules.py
lerman25: This check might be redundant, but if we keep it, I think this is the correct place for this check.
We want to make sure that the other tests in the class didn't add a error to the count, by mistake (Like counting an arg error in the metric because of a syntax error).

tests/pytests/test_info_modules.py
meiravgri: i meant error info fields (initialized to 0) :)

tests/pytests/test_info_modules.py
meiravgri: @Itzikvaknin can you confirm?

tests/pytests/test_info_modules.py
meiravgri: 2. not that im aware of
https://github.com/RediSearch/RediSearch/blob/a3a216334097ecd5d8d52589bfd5272be5d81f97/src/config.h#L284

tests/pytests/test_info_modules.py
meiravgri: Why not to align the SA tests and check it was increased by 1? also, If you want to test that the args error doesn't increase syntax error (which i agree we should), it should be part of this test, and not the other way around

tests/pytests/test_info_modules.py
Itzikvaknin: We should align it with `SEARCH/AGGREGATE` approach
Is it counted on the shards as well?
Anyway, its on our side, [MOD-12101](https://redislabs.atlassian.net/browse/MOD-12101)
You can add this test expecting the current status and we will fix that as part of 👆

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +603,305 @@ def test_indexing_metrics(env: Env):
...
+      info_dict = info_modules_to_dict(shard_conn)
+      args_error_count = info_dict[WARN_ERR_SECTION][ARGS_ERROR_SHARD_METRIC]
+      self.env.assertEqual(args_error_count, '1', message=f"Shard {shardId} has wrong args error count")
+    # Check coord metric unchanged
+    info_dict = info_modules_to_dict(self.env)
+    coord_args_error_count = info_dict[COORD_WARN_ERR_SECTION][ARGS_ERROR_COORD_METRIC]
+    self.env.assertEqual(coord_args_error_count, '0')
+
+    #### Commented out due to a bug (MOD-12465), when fixed, please enable
+    #Note that all the next coord_args_error_count checks should be updated as well
```
meiravgri:
```suggestion
    # Note that all the next coord_args_error_count checks should be updated as well
```

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +603,305 @@ def test_indexing_metrics(env: Env):
...
+    if False:
+      # Test args errors that are counted in the coord
```
meiravgri: maybe instead of comment out compare to the current result so once it is fixed it will fail

meiravgri: also, arnt there other param errors we can test?

tests/pytests/test_info_modules.py
```diff
@@ -602,3 +603,305 @@ def test_indexing_metrics(env: Env):
...
+        info_dict = info_modules_to_dict(shard_conn)
+        args_error_count = info_dict[WARN_ERR_SECTION][ARGS_ERROR_SHARD_METRIC]
+        self.env.assertEqual(args_error_count, '1', message=f"Shard {shardId} has wrong args error count")
+      # Check coord metric unchanged
+      info_dict = info_modules_to_dict(self.env)
+      coord_args_error_count = info_dict[COORD_WARN_ERR_SECTION][ARGS_ERROR_COORD_METRIC]
+      self.env.assertEqual(coord_args_error_count, '1')
+
+    # Test args errors in aggregate
+    # All args errors in FT.AGGREGATE are counted on the coordinator
```
meiravgri: are you sure?
https://github.com/RediSearch/RediSearch/blob/a5584f48327db38313b89ae35acb71215356c11d/src/coord/dist_plan.cpp#L211

tests/pytests/test_info_modules.py
lerman25: self. env must be set to some Env
Removed the limit on shardsCount

tests/pytests/test_info_modules.py
lerman25: > also, arnt there other param errors we can test?

Used missing DIALECT param, invokes args error on coord

tests/pytests/test_info_modules.py
lerman25: I tried reaching this line with `ft.aggregate idx hello_world groupby 1 @text reduce count 1 as count` 
but it reaches an error before that (in the coord)

lerman25: The code is really not that clear in regards what is only caught by the shard,
so I changed the comment to 
`# All args errors in FT.AGGREGATE should be (de facto) counted on the coordinator`
If we can find one that can, I will add it

tests/pytests/test_info_modules.py
lerman25: Added `_verify_metrics_not_changed`, that checks all field are not changed

PR #7398, opened 2025-11-18
[MOD-12375] Consume C iterators from Rust

LukeMathWalker:

Allow Rust to consume iterators written in C.

Stacked on #7364 

#### Mark if applicable

---

> [!NOTE]
> Introduce a C-to-Rust iterator shim and update interop wrappers (mut/const accessors, moved-state handling), adjusting metric iterator FFI accordingly.
> 
> - **FFI/Interop**:
>   - **C→Rust shim**: Add `c_entrypoint/iterators_ffi/src/c2rust.rs` with `CRQEIterator` to consume C `QueryIterator`s via the `RQEIterator` trait.
>   - **Rust→C wrapper**: Add `iterators_ffi/src/wrapper.rs` and enhance `rqe_iterators_interop` with `mut_ref_from_header_ptr`/`ref_from_header_ptr` (replacing `from_header`) for safer header casting.
>   - **Validation**: Update `revalidate` to set `header.current`, `lastDocId`, and `atEOF` on `Moved`.
> - **Metric iterator FFI**:
>   - Switch to new wrapper accessors in `SetMetricRLookupHandle`, `GetMetricOwnKeyRef`, and `GetMetricType`.
> - **Module wiring**:
>   - Expose `mod c2rust` in `iterators_ffi/src/lib.rs`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9bea2281a9d5807b2791cf85b725b027889eb6cd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
+++ b/src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
@@ -53,7 +53,8 @@ pub(crate) struct CRQEIterator {
     ///    and can be converted to a reference.
     /// 2. [`Self::header`] is an owning pointer, in the same way `Box` owns the
     ///    allocated heap data.
-    /// 3. All callbacks are defined (i.e. the function pointers are not NULL)
+    /// 3. All callbacks are defined (i.e. the function pointers are not NULL),
+    ///    with the exception of `SkipTo`, which is optional.
     /// 4. All callbacks can be safely called, when the right aliasing conditions are
     ///    in place
     header: NonNull<QueryIterator>,
@@ -129,7 +130,8 @@ impl CRQEIterator {
     ///    and can be converted to a reference.
     /// 2. `header` is an owning pointer, in the same way `Box` owns the
     ///    allocated heap data.
-    /// 3. All callbacks are defined (i.e. the function pointers are not NULL)
+    /// 3. All callbacks are defined (i.e. the function pointers are not NULL),
+    ///    with the exception of `SkipTo`, which is optional.
     /// 4. All callbacks can be safely called, when the right aliasing conditions are
     ///    in place
     #[allow(unused)]
@@ -143,10 +145,6 @@ impl CRQEIterator {
         debug_assert!(
             self_.Read.is_some(),
             "The `Read` callback is a NULL function pointer"
-        );
-        debug_assert!(
-            self_.SkipTo.is_some(),
-            "The `SkipTo` callback is a NULL function pointer"
         );
         debug_assert!(
             self_.Revalidate.is_some(),
@@ -212,8 +210,9 @@ impl<'index> RQEIterator<'index> for CRQEIterator {
         &mut self,
         doc_id: t_docId,
     ) -> Result<Option<SkipToOutcome<'_, 'index>>, RQEIteratorError> {
-        // SAFETY: Safe thanks to invariant 3. of [`CRQEIterator::header`].
-        let callback = unsafe { self.SkipTo.unwrap_unchecked() };
+        let callback = self
+            .SkipTo
+            .expect("The `SkipTo` callback is a NULL function pointer");
         // SAFETY:
         // - We have a unique handle over this iterator.
         // - The C code must guarantee, by constructor, that callbacks


src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
```diff
@@ -0,0 +1,190 @@
...
+}
+
+impl CRQEIterator {
+    /// Transmute a non-null pointer to a C query iterator into a mutable reference
+    /// to its Rust wrapper.
+    ///
+    /// # Safety
+    ///
+    /// 1. It must be safe to cast `NonNull<QueryIterator>` to `&mut QueryIterator`.
+    ///    In particular, the pointer cannot be aliased.
```
gdesmott: How is aliasing breaking it?

LukeMathWalker: You're casting a raw pointer to a `&mut` reference. If you were to call it twice using the same raw pointer, you'd have two `&mut` references to the same location, thus undefined behaviour.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
```diff
@@ -0,0 +1,309 @@
...
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+// TODO: remove once we have compound iterators written in Rust that leverage
+//   this shim.
+#![allow(unused)]
```
GlenDC:
```suggestion
#![expect(
    unused,
    reason = "already here for once we have compound iterators written in Rust that leverage this shim",
)]
```

in worst case this will mean splitting up in 2 `ffi:{..}` import groups between the ones used and not used. But it does make it at least clear which ones are already used and which not.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
```diff
@@ -0,0 +1,309 @@
...
+    ///
+    /// 1. [`Self::header`] is a valid pointer to a [`QueryIterator`] instance,
+    ///    and can be converted to a reference.
+    /// 2. [`Self::header`] is an owning pointer, in the same way `Box` owns the
+    ///    allocated heap data.
+    header: NonNull<QueryIterator>,
+}
+
+impl AsRef<QueryIterator> for CRQEIterator {
+    fn as_ref(&self) -> &QueryIterator {
```
GlenDC, line 64: why do we need to be able to (de)reference (mut) from the wrapper to the C one? Is that not a detail specific to `CRQEIterator` given we just want to use it as if it was a Rust impl of `RQEIterator` which is afaik usually just used via its interface once created?

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
```diff
@@ -0,0 +1,309 @@
...
+            .expect("The `SkipTo` callback is a NULL function pointer");
+        // SAFETY:
+        // - We have a unique handle over this iterator.
+        // - The C code must guarantee, by constructor, that callbacks
+        //   can be called on types that implement its C iterator API.
+        let status = unsafe { callback(self.header.as_ptr(), doc_id) };
+        #[allow(non_upper_case_globals)]
+        match status {
+            IteratorStatus_ITERATOR_EOF => {
+                self.atEOF = true;
```
GlenDC: why do we need to set this value? That is something the C QueryIterator would already set if desired, no?

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
```diff
@@ -0,0 +1,309 @@
...
+                // SAFETY:
+                // - We have a unique handle over this iterator.
+                let data = unsafe {
+                    data.as_mut()
+                        .expect("`current` is a NULL pointer after `SkipTo` returned NOT_FOUND")
+                };
+                Ok(Some(SkipToOutcome::NotFound(data)))
+            }
+            _ => {
+                unreachable!("`SkipTo` returned an unexpected iterator status, {status}")
```
GlenDC, line 246: `unreachable` is normally used for something that you know fur sure cannot happen, but this seems to be a bit of a stretch given that it is just an integer and coming from C, which dunno how certain you can be that it doesn't happen...

I would honestly at production time return an error here, be it requiring to add some kind of ffi or other error variant to the enum error we already have. With perhaps a `debug-only` panic.

Just my 2 cents though.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
```diff
@@ -0,0 +1,309 @@
...
+            }
+        }
+    }
+
+    fn rewind(&mut self) {
+        let callback = self
+            .Rewind
+            .expect("The `Rewind` callback is a NULL function pointer");
```
GlenDC: IMHO this to me shows that rewind should probably be fallible as it's not unreasonable to think that something can go wrong during rewinding 🤷

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
```diff
@@ -0,0 +1,309 @@
...
+        //   can be called on types that implement its C iterator API.
+        unsafe { callback(self.header.as_ptr()) };
+    }
+
+    fn revalidate(
+        &mut self,
+    ) -> Result<rqe_iterators::RQEValidateStatus<'_, 'index>, RQEIteratorError> {
+        let callback = self
+            .Revalidate
+            .expect("The `Revalidate` callback is a NULL function pointer");
```
GlenDC: we already have an error... this should IMHO just be an error...

GlenDC (dismissed):
Looks good, added some comments and thoughts, but afaik nothing blocking and I have a feeling that some you'll just close with a good explanation. All good for me.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
```diff
@@ -0,0 +1,309 @@
...
+        /// at the same time.
+        unsafe {
+            self.header.as_mut()
+        }
+    }
+}
+
+impl Drop for CRQEIterator {
+    fn drop(&mut self) {
+        if let Some(free) = self.Free {
```
gdesmott: Can this ever be `None`? C code seems to unconditionally call `Free()`.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
LukeMathWalker: We need to be able to return a `*mut QueryIterator`/`*const QueryIterator` back to C in some circumstances, without giving up ownership (where we'd use `into_raw`).
You can see it in action in #7464 for the `Get*` method on the FFI interface for the optional iterator.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
LukeMathWalker: Excess of zeal on my side, removed.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
LukeMathWalker: The current error enum on the C side doesn't allow us to express an "internal error", so panicking is the only option at this point.
But we should definitely refactor it to allow for the equivalent of "something unexpected went wrong", as that's going to be required for I/O errors and other failure modes.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
LukeMathWalker: Same as above, you can't make this an error.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
LukeMathWalker: This can't fail unless the underlying callback fails, therefore we should evaluate if making things fallible makes sense when porting specific iterators.
As for the `expect`, that's an assumption of well-formedness we need to be able to make.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
LukeMathWalker: The `unused` here referred to our `pub(crate)` methods. I moved it to the specific items that need it, rather than making it a blanket exemption.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
LukeMathWalker: It shouldn't, switched it up to a panic.
But I added asserts early on in the constructor to verify that all callbacks are set.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
```diff
@@ -0,0 +1,323 @@
...
+        debug_assert!(
+            self_.Free.is_some(),
+            "The `Free` callback is a NULL function pointer"
+        );
+        debug_assert!(
+            self_.Read.is_some(),
+            "The `Read` callback is a NULL function pointer"
+        );
+        debug_assert!(
+            self_.SkipTo.is_some(),
```
GuyAv46: Some iterators don't implement `SkipTo` (hybrid reader and optimizer), as there is no way to do it, and they should only be at the root of the iterator tree. I guess for these iterators, we don't expect to use this shim.
What about the unsorted ID list?

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
```diff
@@ -0,0 +1,323 @@
...
+        let self_ = ManuallyDrop::new(self);
+        self_.header
+    }
+}
+
+impl<'index> RQEIterator<'index> for CRQEIterator {
+    fn read(&mut self) -> Result<Option<&mut RSIndexResult<'index>>, RQEIteratorError> {
+        let callback = self
+            .Read
+            .expect("The `Read` callback is a NULL function pointer");
```
GuyAv46: Can we unwrap unchecked? We validate that in the constructor.
Should we unwrap unchecked?

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
```diff
@@ -0,0 +1,323 @@
...
+        }
+    }
+
+    fn skip_to(
+        &mut self,
+        doc_id: t_docId,
+    ) -> Result<Option<SkipToOutcome<'_, 'index>>, RQEIteratorError> {
+        let callback = self
+            .SkipTo
+            .expect("The `SkipTo` callback is a NULL function pointer");
```
GuyAv46, line 215: Can we unwrap unchecked? We validate that in the constructor.
Should we unwrap unchecked?

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
LukeMathWalker: This only checks that there is a `SkipTo` callback—e.g. unsorted ID list does have a `SkipTo` callback, which panics.
If the expectation is that `SkipTo` can actually be NULL, then we should remove this debug assertion.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
LukeMathWalker: We can, changed it in ca926edff

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
LukeMathWalker: We can, changed it in ca926edff

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
LukeMathWalker: OK, it's actually `NULL` in the cases you mentioned:
```c
  ri->SkipTo = NULL;            // The iterator is always on top and and Read() is called
```
I'll amend accordingly.

LukeMathWalker: Changed in c2fd756ac

src/redisearch_rs/c_entrypoint/iterators_ffi/src/c2rust.rs
GuyAv46: I see you left this one, so we do it unchecked for everything besides `SkipTo` 👍🏼

LukeMathWalker: Yeah, since `NULL` is valid we need to check it at runtime in that case.

GlenDC (approved):

PR #7281, opened 2025-11-10
[MOD-11717] (3/3) Implement string variants of Rust RsValue

hdoordt:
### Implement string variants of RsValue and SharedRsValue
Define types `ConstString`, `OwnedRedisString`, `OwnedRmAllocString`,
`RedisStringRef`, and `RsValueString`, which are the different kinds
of string `RsValue` and `SharedRsValue` will have to hold.

Add variants to `RsValueInteral` for these kinds of strings, and
add corresponding constructors to the `Value` trait.

### Implement RSValue string constructors in `value_ffi`
Implement the various constructors for string variants
of `RsValue` and `SharedRsValue` that are in use by the
C codebase.

Stacked on #7280 

---

> [!NOTE]
> Implements string-backed RsValue/SharedRsValue variants with full FFI constructors/APIs and updates headers, mocks, and deps accordingly.
> 
> - **Value (core)**:
>   - Add string types `OwnedRmAllocString`, `ConstString`, `OwnedRedisString`, `RedisStringRef`, `RsValueString` and error `CreateRsValueStringError` in `value::strings`.
>   - Extend `RsValueInternal` with string variants; add corresponding `Value` constructors (`take_rm_alloc_string`, `copy_rm_alloc_string`, `const_string`, `borrowed_redis_string`, `retain_owned_redis_string`, `take_owned_redis_string`, `string`).
>   - Minor: `RsValue::null_const`, `get_number`, test macro hook for mock allocator.
> - **FFI (value_ffi)**:
>   - Implement `RsValue_String` and `RsValue_Type`.
>   - Implement `SharedRsValue_*` constructors for strings: `NewString`, `NewCString`, `NewConstString`, `NewBorrowedRedisString`, `NewOwnedRedisString`, `NewStolenRedisString`, `NewCopiedString`, plus helpers: `NewParsedNumber`, `NewNumber`, `NewNumberFromInt64`, `NewStringArray`, `NewConstStringArray`, `NewTrio`, and `Number_Get`.
>   - Add `RsValueType` enum cases for all string kinds and map them from internals.
> - **C Headers (`headers/value.h`)**:
>   - Expose new `RsValueType_*` string variants and declare new/updated FFI functions (arg types normalized to `u32`, borrowed Redis string now non-const, add `SharedRsValue_Number_Get`).
> - **Redis mock and bench/test glue**:
>   - Provide `RSDummyContext` and allocator symbols via macro; export `ffi`; add helper `redis_module_ctx()`.
[...]

PR #7403, opened 2025-11-18
[MOD-11718] (1/?): Introduce dynamic RsValue types, implement type getters

hdoordt:
This PR makes the following changes:

### Introduce dynamic RsValue types
This PR restructures `value_ffi` such that it no longer
exposes `SharedRsValue` and `RsValue` separately, but instead
exposes enum types that represent either. The reasoning is twofold:

- `RSValue_Dereference` may either return an exclusive or a shared RSValue,
  and `value_ffi` needs to support that
- More importantly, this change makes the API `value_ffi` defines much more similar
  to the current C API and will therefore make the migration significantly
  easier.

In particular, this PR introduces the following types:

- `value::DynRsValueRef`: this enum either holds a shared reference to an
  `RsValue` or owns a `SharedRsValue`. It is the return type of `Value::deep_deref`,
  which forms the basis of `RsValue_Dereference`;
- `value_ffi::DynRsValue`: this enum either owns an `RsValue` or a `SharedRsValue`.
  It is exposed as an opaque sized type to C, allowing `DynRsValue` to be placed
  on the C stack. The constructors `value_ffi` exposess each return this value,
  and mutating methods take `Option<NonNull<DynRsValue>>` as parameter.
- `value_ffi::DynRsValuePtr`: this enum either holds a const pointer to an
  `RsValue`, or a const pointer to the `Arc` wrapped by a `SharedRsValue`.
  It provides a means for Rust to disambiguate whether the value is shared or
  exclusive, and requires only a single dereference for each case, whereas a
  `*const DynRsValue` would require two to obtain the inner value of the
  shared variant. `DynRsValuePtr` is exposed to C as an opaque sized type,
  and is a parameter to methods that read the value, but don't mutate it.

Note that `DynRsValue` and `DynRsValuePtr` are solely there to create an FFI
that resembles the C RSValue API. Rust code that operates on `RsValue`s, should
use `RsValue` and `SharedRsValue` for explicitness about their behaviour.

### Add `RsValue_Number_Get`, `RsValue_IntoNumber`, and `SharedRsValue_IntoNumber`
Add method `to_number` to `Value` that converts a value to a number in-place,
use it to implement `RsValue_IntoNumber`.

### Implement type getters

While implementing `RsValue_Dereference`, which may either return a shared value or an exclusive one, the decision was made to unify the FFI module for RSValue such that C does not get exposed to there being separate RsValue types. It would be more neat to create separate PRs for the introduction of `DynRsValue` and the implementation of the getters, but restructuring history is cumbersome in this case and the getters are not very complicated. For these reasons, I kept the combined PR and will stack the implementation of other getters and utilities on top of this one.

Stacked on #7281 

---

> [!NOTE]
> Adds dynamic `RsValue` and pointer types to FFI, migrates C API to unified opaque types, and implements type predicates, number ops, string access, dereference, replace, and pointer conversion.
> 
> - **FFI (`value_ffi`)**:
>   - **Dynamic types**: Add `dynamic::DynRsValue` (exclusive/shared enum) and `DynRsValuePtr` with opaque C projections (`RsValue`, `RsValuePtr`).
>   - **API unification**: Replace `SharedRsValue_*` C APIs with unified `RsValue_*` constructors and helpers; `RsValue_NullStatic` now returns `RsValuePtr`.
>   - **Type predicates**: Implement `RsValue_Type` and `RsValue_Is*` helpers (number/string/array/ref/trio/null/any-string/redis-string/owned-redis-string).
>   - **Utilities**: Add `RsValue_Number_Get`, `RsValue_IntoNumber`, `RsValue_StringPtrLen`, `RsValue_Dereference`, `RsValue_DynPtr`, `RsValue_Replace`, `RsValue_Free`, and `RsValue_ToShared`.
>   - **cbindgen**: Export renames updated to map `OpaqueDynRsValue`→`RsValue`, `OpaqueDynRsValuePtr`→`RsValuePtr`.
>   - **Tests**: New unit tests for creation, conversion, replace, and free across exclusive/shared variants.
> - **Core value crate (`value`)**:
>   - Add `dynamic::DynRsValueRef` and integrate with `Value` trait (`to_dyn_ref`, `to_shared`, `deep_deref`, `to_number`, `string_as_bytes`).
>   - Enhance `SharedRsValue` with `into_raw`, `as_raw`, `from_raw`; implement `Debug` changes.
>   - Add `RedisStringRef::as_bytes`.
> - **C FFI utils**:
>   - Extend `opaque!` macro to handle generics/lifetimes and add `as_opaque_non_null`/`from_opaque_non_null`.
> - **Build/Test**:
>   - Add `redis_mock` as a dev-dependency; lockfile updated.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f51cf8e0a72eb3c7145a9f6bbcb3b15cd3d1c281. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/Cargo.lock
+++ b/src/redisearch_rs/Cargo.lock
@@ -2185,6 +2185,7 @@ dependencies = [
  "cbindgen",
  "ffi",
  "libc",
+ "redis_mock",
  "value",
 ]
 
--- a/src/redisearch_rs/c_entrypoint/c_ffi_utils/src/opaque.rs
+++ b/src/redisearch_rs/c_entrypoint/c_ffi_utils/src/opaque.rs
@@ -7,6 +7,8 @@
  * GNU Affero General Public License v3 (AGPLv3).
 */
 
+use std::ptr::NonNull;
+
 /// A type with size `N`.
 #[repr(transparent)]
 pub struct Size<const N: usize>(std::mem::MaybeUninit<[u8; N]>);
@@ -38,6 +40,10 @@ pub trait IntoOpaque: Sized {
     /// `*mut Self::Opaque`.
     fn as_opaque_mut_ptr(&mut self) -> *mut Self::Opaque;
 
+    /// Converts [`Self`] mutable reference into a
+    /// [`NonNull<Self::Opaque>`](NonNull).
+    fn as_opaque_non_null(&mut self) -> NonNull<Self::Opaque>;
+
     /// Converts an [`Self::Opaque`] back to [`Self`].
     ///
     /// # Safety
@@ -66,6 +72,17 @@ pub trait IntoOpaque: Sized {
     /// pointed to by `opaque` must also be an alignment-compatible address for
     /// a [`Self`].
     unsafe fn from_opaque_mut_ptr<'a>(opaque: *mut Self::Opaque) -> Option<&'a mut Self>;
+
+    /// Converts a [`NonNull<Self::Opaque>`](NonNull) to a mutable
+    /// reference to a [`Self`].
+    ///
+    /// # Safety
+    ///
+    /// The pointer itself must have been created via
+    /// [`IntoOpaque::as_opaque_non_null`], as the alignment of the value
+    /// pointed to by `opaque` must also be an alignment-compatible address for
+    /// a [`Self`].
+    unsafe fn from_opaque_non_null<'a>(opaque: NonNull<Self::Opaque>) -> &'a mut Self;
 }
 
 /// Implements [`IntoOpaque`] for the passed type,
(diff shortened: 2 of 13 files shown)


JonasKruckenberg (commented):
Code looks fine as always, just a few high level comments
- Needs crate-level documentation to help consumers understand the difference between the RsValue types (when do I use which type?, etc.) and understand how they relate
- Needs documentation on how types behave when cloned bc now the runtime perf profile is less obvious (might copy in unexpected places)
- We should be clearer about which types exists for backwards compatibility with C and which should be used for new Rust code going forward. Maybe we can move the C FFI types into its own crate or into the `c_entrypoint/value_ffi` crate?

PR #7404, opened 2025-11-18
MOD-12489: FT.HYBRID - Coordinator Deadlock - Avoid Taking Index Read Lock 

kei-nan:
A deadlock was found since the read locks for the index were held by the depleters.

1. Current: Depleters in coordinator always take index lock
2. Change: Depleters in coordinator won't take index lock
3. Outcome: Avoid Deadlocks in coordinator that are related to hybrid.

1. Depleter Index Lock Flow

#### Mark if applicable

---

> [!NOTE]
> Stop taking index read locks in distributed hybrid depleters to prevent deadlocks; remove the synchronize_read_locks parameter and set locking explicitly where needed.
> 
> - **Coordinator/Distributed Depletion**:
>   - In `src/coord/hybrid/dist_hybrid_plan.cpp`, create depleter sync with `false` to avoid index read locks during distributed depletion.
> - **Standalone/Background Depletion**:
>   - In `src/hybrid/hybrid_request.c`, when `depleteInBackground` is true, create depleter sync with `true` (explicitly enabling coordination), decoupled from params.
> - **API Cleanup**:
>   - Remove `HybridPipelineParams.synchronize_read_locks` from `src/pipeline/pipeline.h` and its usage.
>   - Drop initialization of `hybridParams->synchronize_read_locks` in `src/hybrid/parse_hybrid.c`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fe37ea555a2bf2a172753a237ea5e06d0a484f42. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/coord/hybrid/dist_hybrid_plan.cpp
+++ b/src/coord/hybrid/dist_hybrid_plan.cpp
@@ -21,8 +21,8 @@ static void pushDepleter(QueryProcessingCtx *qctx, ResultProcessor *depleter) {
 // should make sure the product of AREQ_BuildPipeline(areq, &req->errors[i]) would result in rpSorter only (can set up the aggplan to be a sorter only)
 int HybridRequest_BuildDistributedDepletionPipeline(HybridRequest *req, const HybridPipelineParams *params) {
   // Create synchronization context for coordinating depleter processors
-  // We avoid taking the index lock since we are not direclyt accessing the index at all
-  // This avoids nasty deadlocks with main thread while it is trying to access the index
+  // We avoid taking the index lock since we are not directly accessing the index at all
+  // This avoids deadlocks with main thread while it is trying to access the index
   StrongRef sync_ref = DepleterSync_New(req->nrequests, false);
 
   // Build individual pipelines for each search request


src/coord/hybrid/dist_hybrid_plan.cpp
```diff
@@ -21,8 +21,9 @@ static void pushDepleter(QueryProcessingCtx *qctx, ResultProcessor *depleter) {
 // should make sure the product of AREQ_BuildPipeline(areq, &req->errors[i]) would result in rpSorter only (can set up the aggplan to be a sorter only)
 int HybridRequest_BuildDistributedDepletionPipeline(HybridRequest *req, const HybridPipelineParams *params) {
   // Create synchronization context for coordinating depleter processors
-  // This ensures thread-safe access when multiple depleters read from their pipelines
-  StrongRef sync_ref = DepleterSync_New(req->nrequests, params->synchronize_read_locks);
+  // We avoid taking the index lock since we are not direclyt accessing the index at all
+  // This avoids nasty deadlocks with main thread while it is trying to access the index
```
Itzikvaknin:
```suggestion
  // We avoid taking the index lock since we are not directly accessing the index at all
  // This avoids deadlocks with main thread while it is trying to access the index
```

src/hybrid/hybrid_request.c
```diff
@@ -27,7 +27,7 @@ int HybridRequest_BuildDepletionPipeline(HybridRequest *req, const HybridPipelin
     StrongRef sync_ref = {0};
     int rc = REDISMODULE_OK;
     if (depleteInBackground) {
-      sync_ref = DepleterSync_New(req->nrequests, params->synchronize_read_locks);
+      sync_ref = DepleterSync_New(req->nrequests, true);
```
Itzikvaknin, line 30: So now the decision wether to take the read lock is dependent on which function we use to build the pipeline?
 `HybridRequest_BuildDepletionPipeline` \ `HybridRequest_BuildDepletionPipeline` ?
 Honestly I dont like it, I prefer the parameter approach 
 Anyway, its not crucial and leaving it to you to decide

kei-nan: when we operate at the shard level we need to take the index lock
when we operate at the coord level we don't
we have the internal boolean in the parse code but it doesn't tell us if we are in coordinator or shard.
because in standalone we also need to take the lock and internal would be false in such a case.
for now will leave as is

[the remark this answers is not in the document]
kei-nan: hybrid_exec.c is standalone and shard context, not coordinator context.
The coordinator calls `HybridRequest_BuildDistributedDepletionPipeline` where we pass false.
The execution logic for the coordinator is in dist_hybrid.c

PR #7407, opened 2025-11-18
MOD-12377: Add search and vsim subqueries scores to the output in case of test failure

Itzikvaknin:
Replaced the old-not-working rank output in case of failure ,with scores output if `YIELD_SCORE_AS` is used in the hybrid query

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 5ae3ad283b10dd278ce0fe0cce6151df2c42f9f0. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_hybrid.py
+++ b/tests/pytests/test_hybrid.py
@@ -108,7 +108,8 @@ class testHybridSearch:
     def test_knn_wildcard_search(self):
         """Test hybrid search using KNN + wildcard search scenario"""
         scenario = {
-            "hybrid_query": "SEARCH * VSIM @vector $BLOB",
+            "hybrid_query": "SEARCH * YIELD_SCORE_AS search_score \
+                VSIM @vector $BLOB YIELD_SCORE_AS vector_score",
             "search_equivalent": "*",
             "vector_equivalent": "*=>[KNN 10 @vector $BLOB AS vector_distance]"
         }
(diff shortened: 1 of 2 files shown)


tests/pytests/test_hybrid.py
```diff
@@ -108,7 +108,8 @@ def test_knn_single_token_search(self):
     def test_knn_wildcard_search(self):
         """Test hybrid search using KNN + wildcard search scenario"""
         scenario = {
-            "hybrid_query": "SEARCH * VSIM @vector $BLOB",
+            "hybrid_query": "SEARCH * YIELD_SCORE_AS search_score \
```
nafraf, line 111: Do we have another test with the `KNN` defaults?
If we are removing this, can we add an equivalent test in `test_hybrid_search`?

Itzikvaknin: Removing what? we dont remove/change a functionality in the query

Regarding the second comment, yes

nafraf: After adding `YIELD_SCORE_AS` in this tests, we don't have a test of simple `KNN` like `FT.HYBRID SEARCH * VSIM @vector $BLOB`
Do you think we don't need it?

Itzikvaknin: Yes, In the end it doesnt do much, just turning on a flag for writing the score

kei-nan (approved):

PR #7410, opened 2025-11-18
[MOD-12252] [MOD-11658] reduce number of worker threads asynchronously

JoanFM:

This PR aims to solve the deadlock while reducing number of threads and waiting for them to finish.

Changes:

- Reduction of number of threads in the worker threadpool are done asynchronously, the effects are not inmediate.

---

> [!NOTE]
> Adds an async thread-reduction mechanism and reliable thread start signaling to the threadpool, switches workers to non-blocking downsizing, and updates tests and tooling accordingly.
> 
> - **Threadpool (deps/thpool)**:
>   - Add `redisearch_thpool_schedule_config_reduce_threads_job(...)` to asynchronously reduce threads (supports terminate-when-empty); mark pool UNINITIALIZED when removing all.
>   - Refactor thread startup: `thread_init(thpool, bool *started)` and `thread_do(void *)` with start signaling; wait deterministically for new threads.
>   - Extract `redisearch_thpool_push_chain(...)`; add `redisearch_thpool_add_n_work_not_verify_init(...)` for enqueueing without init verification.
>   - Improve `redisearch_thpool_add_threads(...)` to wait on per-thread start flags; logging/RS_ASSERT updates.
>   - Remove synchronous APIs (`redisearch_thpool_remove_threads`, `redisearch_thpool_terminate_when_empty`) from headers; adjust internal uses.
> - **Workers (src/util/workers.c)**:
>   - Replace blocking downsizing with async `redisearch_thpool_schedule_config_reduce_threads_job(...)` for both partial and full reduction; keep activation/deactivation hooks; add drain logging.
> - **Config/Debug/Notifications**:
>   - Keep coordinator update trigger after WORKERS change; warn that `FT.DEBUG WORKERS DRAIN` blocks; drain workers during loading progress.
> - **Tests**:
>   - Update to async reduction semantics; add stress tests for reducing WORKERS under load (avoid deadlock), gradual scaling up/down, and various thread lifecycle scenarios.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit aa11f690b2416361b7aa9537ff68bd4a5d4bd2f7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/deps/thpool/thpool.c
+++ b/deps/thpool/thpool.c
@@ -1064,8 +1064,8 @@ void redisearch_thpool_schedule_config_reduce_threads_job(redisearch_thpool_t *t
   size_t n_threads = thpool_p->n_threads;
   LOG_IF_EXISTS("verbose", "Scheduling from main thread to remove %zu threads", n_threads_to_remove);
   RS_LOG_ASSERT((!terminate_when_empty || n_threads_to_remove == n_threads), "If remove_all is set, n_threads_to_remove must be equal to n_threads");
-  RS_ASSERT(thpool_p->n_threads >= n_threads_to_remove && "Number of threads can't be negative");
-  RS_ASSERT(thpool_p->jobqueues.state == JOBQ_RUNNING && "Can't remove threads while jobq is paused");
+  RS_LOG_ASSERT(thpool_p->n_threads >= n_threads_to_remove, "Number of threads can't be negative");
+  RS_LOG_ASSERT(thpool_p->jobqueues.state == JOBQ_RUNNING, "Can't remove threads while jobq is paused");
 
   ThreadState new_state = terminate_when_empty ? THREAD_TERMINATE_WHEN_EMPTY : THREAD_TERMINATE_ASAP;
   thpool_p->n_threads -= n_threads_to_remove;


deps/thpool/thpool.c
```diff
@@ -1024,3 +1065,83 @@ static void redisearch_thpool_broadcast_new_state(redisearch_thpool_t *thpool,
...
+  redisearch_thpool_t *thpool = job_arg->thpool;
+  size_t n_threads_to_remove = job_arg->n_threads_to_remove;
+  /* Create jobs and their args. */
+  redisearch_thpool_work_t jobs[n_threads_to_remove];
+  /* Create a barrier. */
+  pthread_barrier_t *barrier = rm_malloc(sizeof(pthread_barrier_t));
+  pthread_barrier_init(barrier, NULL, n_threads_to_remove);
+  struct SignalThreadCtxWithBarrier *signal_job_arg = rm_malloc(sizeof(struct SignalThreadCtxWithBarrier));
+  signal_job_arg->barrier = barrier;
+  signal_job_arg->new_state = THREAD_TERMINATE_ASAP;
```
JoanFM: pass this as argument, as if workers is set to 0, it should terminate when Empty

deps/thpool/thpool.c
```diff
@@ -903,6 +933,17 @@ static priorityJobCtx priority_queue_pull(priorityJobqueue *priority_queue_p) {
...
 }
 static inline priorityJobCtx priority_queue_pull_from_queues_unsafe(priorityJobqueue *priority_queue_p) {
+  if (priority_queue_p->last_config_reduce_threads_job) {
+    priority_queue_p->num_jobs_in_progress++;
+    priorityJobCtx job_ctx = {
+        .job = priority_queue_p->last_config_reduce_threads_job,
+        .is_admin = false,
+        .has_priority_ticket = false,
+    };
+    priority_queue_p->last_config_reduce_threads_job = NULL;
```
JoanFM: copy and set to NULLz this is now stupid

deps/thpool/thpool.c
```diff
@@ -373,9 +370,9 @@ size_t redisearch_thpool_add_threads(redisearch_thpool_t *thpool_p,
     usleep(1);
   }
 
-  LOG_IF_EXISTS("verbose", "Thread pool size increased to %zu successfully", n_threads)
+  pthread_mutex_unlock(&thpool_p->jobqueues.lock);
```
JoanFM: remove leftover

deps/thpool/thpool.c
```diff
@@ -114,6 +114,14 @@ typedef struct redisearch_thpool_t {
                                                 null byte) to leave room for
                                                 '-<thread id>'*/
 } redisearch_thpool_t;
+
+struct SignalThreadCtxWithBarrier {
```
JoanFM: move down

deps/thpool/thpool.c
```diff
@@ -1024,3 +1025,79 @@ static void redisearch_thpool_broadcast_new_state(redisearch_thpool_t *thpool,
...
+  /* n_threads is only configured and read by the main thread (protected by the GIL). */
+  /** THPOOL_UNINITIALIZED means either:
+   * 1. thpool->n_threads > 0, and there are no threads alive
+   * 2. There are threads alive in TERMINATE_WHEN_EMPTY state.
+   * In any case we cannot remove more threads. */
+  if (thpool_p->state == THPOOL_UNINITIALIZED || thpool_p->n_threads == 0)
+    return;
+
+  size_t n_threads = thpool_p->n_threads;
+  size_t jobs_count = priority_queue_len_unsafe(&thpool_p->jobqueues);
```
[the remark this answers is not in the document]
JoanFM: true, leftover from when that code was protected

JoanFM:
```suggestion
  size_t jobs_count = priority_queue_len(&thpool_p->jobqueues);
```

deps/thpool/thpool.c
```diff
@@ -664,6 +664,7 @@ static void *thread_do(redisearch_thpool_t *thpool_p) {
 
   /* Mark thread as alive (initialized) */
   thpool_p->num_threads_alive += 1;
+  *args->started = true;
```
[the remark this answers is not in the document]
JoanFM: @JoanFM to take a look. We can free when the thread terminates

deps/thpool/thpool.c
```diff
@@ -308,74 +319,34 @@ static void redisearch_thpool_verify_init(struct redisearch_thpool_t *thpool_p)
...
-  size_t n_threads = thpool_p->n_threads;
-
-  /** THPOOL_UNINITIALIZED means either:
-   * 1. thpool->n_threads > 0, and there are no threads alive
-   * 2. There are threads alive in terminate_when_empty state.
-   * In both cases only calling `verify_init` will add/remove threads to adjust
-   * `num_threads_alive` to `n_threads`
-   * @note if thpool->n_threads was decreased to 0 (thpool_remove_threads), the thpool is in INITIALIZED
-   * state. */
-  if (thpool_p->state == THPOOL_UNINITIALIZED)
```
JoanFM, line 363: this only complicates the logic and adds the chance of having barrier waits in the main thread in the verify_init path. I think lazy loading of threads can be useful at the setup of the engine (which already happens), but when the user actually aims to add new threads explicitly, there is no point on doing this lazily

deps/thpool/thpool.c
```diff
@@ -1024,3 +1025,79 @@ static void redisearch_thpool_broadcast_new_state(redisearch_thpool_t *thpool,
...
 }
+
+
+/* ========================== CONFIGURATION THREAD REDUCTION ============================ */
+
+struct SignalThreadCtxWithBarrier {
+  ThreadState new_state;
+  pthread_barrier_t *barrier;
+  atomic_int num_threads_to_wait;
+  redisearch_thpool_t *thpool;
```
JoanFM: remove, it was only there for logging

deps/thpool/thpool.c
```diff
@@ -1024,3 +1017,75 @@ static void redisearch_thpool_broadcast_new_state(redisearch_thpool_t *thpool,
...
+
+  /* Wait all threads to get the barrier */
+  pthread_barrier_wait(signal_struct->barrier);
+  int num_threads_to_wait_before = atomic_fetch_sub(&signal_struct->num_threads_to_wait, 1);
+  if (num_threads_to_wait_before - 1 == 0) {
+    pthread_barrier_destroy(signal_struct->barrier);
+    rm_free(signal_struct->barrier);
+    rm_free(signal_struct);
+  }
+}
```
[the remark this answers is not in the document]
JoanFM: The atomic counter is initialized to the same amount of threads as the barrier, so when it hits 0, it means nobody else will need to use/access the struct or the barrier

deps/thpool/thpool.c
```diff
@@ -293,89 +296,61 @@ static void redisearch_thpool_verify_init(struct redisearch_thpool_t *thpool_p)
...
+    size_t n_started = 0;
+    while (n_started < n_new_threads) {
+      for (size_t n = 0; n < n_new_threads; n++) {
+        if (started[n]) {
+          ++n_started;
+          started[n] = false;
+        }
+      }
+      usleep(1);
+    }
```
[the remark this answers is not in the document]
JoanFM: tests would catch these compile optimizations, but eventually they would be read.

deps/thpool/thpool.c
```diff
@@ -1024,3 +1020,76 @@ static void redisearch_thpool_broadcast_new_state(redisearch_thpool_t *thpool,
...
+    jobs[i].function_p = admin_job_change_state_last_destroys_barrier;
+  }
+  // I do not need to verify init since we are actually putting in priority queue, and I do not want to wait on another barrier
+  // As per the input, I know i am Initialized, I do not need to verify it (and avoid waiting)
+  redisearch_thpool_add_n_work_not_verify_init(thpool_p, jobs, n_threads_to_remove, THPOOL_PRIORITY_ADMIN);
+  if (remove_all) {
+    assert(thpool_p->n_threads == 0 && "If remove_all is set, n_threads_to_remove must be equal to n_threads");
+    thpool_p->state = THPOOL_UNINITIALIZED;
+  }
+}
```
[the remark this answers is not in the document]
JoanFM: I will add the assertion, but this would be guaranteed by the regular working of the mechanism (with lazy init, if it is INITIALIZED the num_threads_alive match at least the num_threads

deps/thpool/thpool.c
```diff
@@ -1024,3 +1022,76 @@ static void redisearch_thpool_broadcast_new_state(redisearch_thpool_t *thpool,
...
+  job_arg->thread_ctx->thread_state = new_state;
+
+  /* Wait all threads to get the barrier */
+  pthread_barrier_wait(signal_struct->barrier);
+  int num_threads_to_wait_before = atomic_fetch_sub(&signal_struct->num_threads_to_wait, 1);
+  if (num_threads_to_wait_before - 1 == 0) {
+    pthread_barrier_destroy(signal_struct->barrier);
+    rm_free(signal_struct->barrier);
+    rm_free(signal_struct);
+  }
```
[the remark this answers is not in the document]
JoanFM: with the counter it is guaranteed noone needs it when destroying and freeing the structures

deps/thpool/thpool.c
```diff
@@ -1012,3 +1020,75 @@ static void redisearch_thpool_broadcast_new_state(redisearch_thpool_t *thpool,
...
+  for (size_t i = 0; i < n_threads_to_remove; i++) {
+    jobs[i].arg_p = signal_job_arg;
+    jobs[i].function_p = admin_job_change_state_last_destroys_barrier;
+  }
+  // I do not need to verify init since we are actually putting in priority queue, and I do not want to wait on another barrier
+  // As per the input, I know i am Initialized, I do not need to verify it (and avoid waiting)
+  redisearch_thpool_add_n_work_not_verify_init(thpool_p, jobs, n_threads_to_remove, THPOOL_PRIORITY_ADMIN);
+  if (remove_all) {
+    thpool_p->state = THPOOL_UNINITIALIZED;
+  }
```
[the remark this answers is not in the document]
JoanFM: well yes we do not check memory allocation failures throughout the code

deps/thpool/thpool.c
```diff
@@ -293,89 +296,61 @@ static void redisearch_thpool_verify_init(struct redisearch_thpool_t *thpool_p)
   }
 
   /* Add new threads if needed */
-  for (size_t n = 0; n < n_new_threads; n++) {
-    thread_init(thpool_p);
-  }
+  if (n_new_threads > 0) {
+    bool started[n_new_threads];
```
JoanFM: @JoanFM review the lifetime of the started flag sent to thread. It is safe now, but better not to have dangling variables in the thread stack

JoanFM: If I have a heap-allocated stack, I need to have a better coordination and more complex mechanism to free it, or I have to free at the end of the thread, which means unnecessary memory is used.

What have done, is to keep it as it is, mark started as volatile in the thread argument, and make sure the pointer is set to NULL so that no potential threat involving access to stack can see the dangling pointer.

deps/thpool/thpool.c
```diff
@@ -293,89 +296,61 @@ static void redisearch_thpool_verify_init(struct redisearch_thpool_t *thpool_p)
...
+  size_t n_threads = thpool_p->n_threads + n_threads_to_add;
+  thpool_p->n_threads = n_threads;
+  if (thpool_p->state == THPOOL_UNINITIALIZED) {
+    // If the thpool is not initialized, we just set the n_threads and return, so that when workers are added
+    // they will be initialized with the new n_threads.
     return n_threads;
-
+  }
   /* Add new threads */
+  bool started[n_threads_to_add];
```
alonre24: Should this be atomic since it is accessed in the threads?

deps/thpool/thpool.c
alonre24: I agree. But eventually, you do not propose changing it now, correct?

src/debug_commands.c
```diff
@@ -1478,6 +1478,7 @@ DEBUG_COMMAND(WorkerThreadsSwitch) {
     // Log that we're waiting for the workers to finish.
     RedisModule_Log(RSDummyContext, "notice", "Debug workers drain");
 
+    //TODO(Joan): Should we remove this debug (or should we add documentation about the danger?)
```
alonre24: I don't think we should remove it for debug, I would go with documenting the danger

deps/thpool/thpool.c
```diff
@@ -1012,3 +1020,76 @@ static void redisearch_thpool_broadcast_new_state(redisearch_thpool_t *thpool,
...
+void redisearch_thpool_schedule_config_reduce_threads_job(redisearch_thpool_t *thpool_p, size_t n_threads_to_remove, bool remove_all) {
+  /* n_threads is only configured and read by the main thread (protected by the GIL). */
+  /** THPOOL_UNINITIALIZED means either:
+   * 1. thpool->n_threads > 0, and there are no threads alive
+   * 2. There are threads alive in TERMINATE_WHEN_EMPTY state.
+   * In any case we cannot remove more threads. */
+  if (thpool_p->state == THPOOL_UNINITIALIZED || thpool_p->n_threads == 0) {
+    if (thpool_p->n_threads > 0) {
+      // If is UNINITIALIZED, at least it would lazily initialize less threads
+      assert(thpool_p->n_threads >= n_threads_to_remove && "Number of threads can't be negative");
```
alonre24: Use `RS_ASSERT` instead of `assert` everywhere

deps/thpool/thpool.c
```diff
@@ -1012,3 +1020,76 @@ static void redisearch_thpool_broadcast_new_state(redisearch_thpool_t *thpool,
...
+  LOG_IF_EXISTS("verbose", "Scheduling from main thread to remove %zu threads", n_threads_to_remove);
+  assert((!remove_all || n_threads_to_remove == n_threads) && "If remove_all is set, n_threads_to_remove must be equal to n_threads");
+  assert(thpool_p->n_threads >= n_threads_to_remove && "Number of threads can't be negative");
+  assert(thpool_p->jobqueues.state == JOBQ_RUNNING && "Can't remove threads while jobq is paused");
+  assert(thpool_p->num_threads_alive >= n_threads_to_remove && "Not enough alive threads to remove");
+
+  ThreadState new_state = remove_all ? THREAD_TERMINATE_WHEN_EMPTY : THREAD_TERMINATE_ASAP;
+
+  // If config_reduce_threads_job is not there but the ADMIN jobs are running, if TERMINATE_ASAP we are safe because the admin jobs would be taken by the other threads.
+  // If the old config is TERMINATE_WHEN_EMPTY it means that we were removing all threads, and we should not enter here again
```
alonre24: I didn't understand this comment

deps/thpool/thpool.h
```diff
@@ -296,6 +275,8 @@ thpool_stats redisearch_thpool_get_stats(redisearch_thpool_t *);
 
 size_t redisearch_thpool_get_num_threads(redisearch_thpool_t *);
 
+void redisearch_thpool_schedule_config_reduce_threads_job(redisearch_thpool_t *thpool_p, size_t n_threads_to_remove, bool remove_all);
```
alonre24: Consider adding documentation about this function and its arguments

tests/cpptests/test_cpp_thpool.cpp
```diff
@@ -355,8 +371,15 @@ TEST_P(PriorityThpoolTestRuntimeConfig, TestRemoveThreads) {
     // Wait for the jobs to be pulled.
     jobs_pull_wait(this->pool, n_threads);
 
-    // Remove rest of the threads while the threads are waiting for the admin jobs to be pushed to the queue
-    ASSERT_EQ(redisearch_thpool_remove_threads(this->pool, n_threads), 0);
+    // Remove rest of the threads while the threads are waiting for the admin jobs to be pushed to the queue,
+    //redisearch_thpool_schedule_config_reduce_threads_job(this->pool, n_threads, true);
+    redisearch_thpool_schedule_config_reduce_threads_job(this->pool, n_threads, false);
+    ASSERT_EQ(redisearch_thpool_get_num_threads(this->pool), 0);
```
alonre24: WHy is this commented out? And how come we are removing all the threads with `remove_all=false` ?

tests/cpptests/test_cpp_thpool.cpp
```diff
@@ -566,18 +599,19 @@ TEST_P(PriorityThpoolTestRuntimeConfig, TestAddThreadsToEmptyPool) {
     redisearch_thpool_wait(this->pool);
 
     // Remove all threads.
-    ASSERT_EQ(redisearch_thpool_remove_threads(this->pool, RUNTIME_CONFIG_N_THREADS), 0);
-    // Threadpool is still initialized, but no threads are alive.
+    redisearch_thpool_schedule_config_reduce_threads_job(this->pool, RUNTIME_CONFIG_N_THREADS, false);
+    ASSERT_EQ(redisearch_thpool_get_num_threads(this->pool), 0);
+    // Threadpool is not initialized,
```
alonre24: Threadpool is *not* initialized?

tests/pytests/test_multithread.py
```diff
@@ -503,47 +506,64 @@ def check_threads(expected_num_threads_alive, expected_n_threads):
...
     # Query should be executed by the threadpool
-    env.assertEqual(getWorkersThpoolStats(env)['totalJobsDone'], 1)
+    env.assertEqual(getWorkersThpoolStats(env)['totalJobsDone'], num_query_threads + 1)
 
     # Add threads to a running pool
     env.expect(config_cmd(), 'SET', 'WORKERS', '2').ok()
     check_threads(expected_num_threads_alive=2, expected_n_threads=2)
     # Remove threads from a running pool
     env.expect(config_cmd(), 'SET', 'WORKERS', '1').ok()
+    time.sleep(1)
```
alonre24: This would be flaky...

tests/pytests/test_multithread.py
```diff
@@ -503,47 +506,64 @@ def check_threads(expected_num_threads_alive, expected_n_threads):
...
+    num_query_threads = 100
+    query_threads = []
+
+    for i in range(num_query_threads):
+        t = threading.Thread(target=send_query, name=f'QueryThread-{i}')
+        t.start()
+        query_threads.append(t)
+
+    for t in query_threads:
+        t.join()
```
alonre24, line 524: I'm not sure I understand what we are testing here... why send the query from `query_threads`?

tests/pytests/test_avoid_deadlock_while_changing_num_workers.py
```diff
@@ -0,0 +1,279 @@
...
+        t.start()
+        query_threads.append(t)
+
+    # Let queries run for a bit to establish load
+    # Increase time to ensure all threads are actively querying
+    time.sleep(3)
+
+    # Verify queries are running successfully
+    initial_success = query_success_count[0]
+    env.assertTrue(initial_success > 0, message="Queries should be running successfully before config change")
```
alonre24: This check is time-dependent - could be flaky

tests/pytests/test_avoid_deadlock_while_changing_num_workers.py
```diff
@@ -0,0 +1,279 @@
...
+    final_ping = env.cmd('PING')
+    env.assertTrue(final_ping in ['PONG', True], message="Redis should still respond to PING at end of test")
+
+    final_search = env.cmd('FT.SEARCH', 'idx', '*', 'LIMIT', '0', '5')
+    env.assertTrue(final_search[0] > 0, message="Search should return results at end of test")
+
+    check_threads(env, 0, 0)
+
+
+def test_workers_reduction_sequence():
```
alonre24: Do we really need those tests, or do we have enough coverage for that in the existing tests in test_multithread?

tests/pytests/test_avoid_deadlock_while_changing_num_workers.py
```diff
@@ -0,0 +1,279 @@
...
+from common import *
+import threading
+import time
+import random
+
+def check_threads(env, expected_num_threads_alive, expected_n_threads):
+    env.assertEqual(getWorkersThpoolStats(env)['numThreadsAlive'], expected_num_threads_alive, depth=1, message='numThreadsAlive should match num_threads_alive')
+    env.assertEqual(getWorkersThpoolNumThreads(env), expected_n_threads, depth=1, message='n_threads should match WORKERS')
+
+def test_workers_reduction_under_load():
```
alonre24: This test is good, but we try to avoid long stress tests in our flow tests, since we don't want to inflate the CI tests' time. Consider shortening it and reducing sleeps where possible.

tests/pytests/test_avoid_deadlock_while_changing_num_workers.py
JoanFM: maybe for these ones we have enough coverage, maybe we should only add the redcuction one under test_issues.py. Will do so

src/debug_commands.c
JoanFM: how should we go about documenting this danger?

deps/thpool/thpool.c
JoanFM: I do not think is needed. It does not matter if the main thread takes another loop to discover the flag being set.

tests/cpptests/test_cpp_thpool.cpp
JoanFM: leftover from a previous change. Will cleanup these messages

tests/pytests/test_multithread.py
JoanFM: I think it is more complete to test while queries are being sent

deps/thpool/thpool.c
GuyAv46: Shouldn't this variable also be set as volatile? So the waiting thread actually checks the value, and it is not optimized away?

GuyAv46: Also, is there a reason to use a flag array instead of an atomic counter?

deps/thpool/thpool.c
```diff
@@ -293,89 +296,61 @@ static void redisearch_thpool_verify_init(struct redisearch_thpool_t *thpool_p)
...
-
-  /** THPOOL_UNINITIALIZED means either:
-   * 1. thpool->n_threads > 0, and there are no threads alive
-   * 2. There are threads alive in terminate_when_empty state.
-   * In both cases only calling `verify_init` will add/remove threads to adjust
-   * `num_threads_alive` to `n_threads`
-   * @note if thpool->n_threads was decreased to 0 (thpool_remove_threads), the thpool is in INITIALIZED
-   * state. */
-  if (thpool_p->state == THPOOL_UNINITIALIZED)
+  assert(n_threads_to_add > 0 && "Number of threads to add must be greater than 0");
```
GuyAv46: We should use `RS_ASSERT` (throughout this file)

deps/thpool/thpool.c
GuyAv46: Consider using an atomic counter

deps/thpool/thpool.c
```diff
@@ -439,6 +438,15 @@ static void redisearch_thpool_push_chain_verify_init_threads(
   redisearch_thpool_verify_init(thpool_p);
 }
 
+static void redisearch_thpool_push_chain(
```
GuyAv46, line 429: Consider reusing this in `redisearch_thpool_push_chain_verify_init_threads` as well

deps/thpool/thpool.c
```diff
@@ -427,6 +402,30 @@ int redisearch_thpool_add_n_work(redisearch_thpool_t * thpool_p,
   return 0;
 }
 
+/* Add n work to the thread pool */
+int redisearch_thpool_add_n_work_not_verify_init(redisearch_thpool_t * thpool_p,
```
GuyAv46: Can be `static`, no?

deps/thpool/thpool.c
```diff
@@ -427,6 +402,30 @@ int redisearch_thpool_add_n_work(redisearch_thpool_t * thpool_p,
...
+
+  job *first_newjob = jobs_chain.first_job;
+  job *last_newjob = jobs_chain.last_job;
+  if (!jobs_chain.first_job || !jobs_chain.last_job) {
+    LOG_IF_EXISTS("warning",
+    "redisearch_thpool_add_n_work_not_verify_init(): Could not allocate memory "
+    "for %zu new jobs",
+    n_jobs);
+    return -1;
+  }
```
GuyAv46, line 421: This can only happen if an `rm_malloc` call fails? Because it never fails

deps/thpool/thpool.c
```diff
@@ -1012,3 +1023,76 @@ static void redisearch_thpool_broadcast_new_state(redisearch_thpool_t *thpool,
...
+static void admin_job_change_state_last_destroys_barrier(void *job_arg_) {
+  adminJobArg *job_arg = job_arg_;
+  struct SignalThreadCtxWithBarrier *signal_struct = job_arg->arg;
+  ThreadState new_state = signal_struct->new_state;
+  job_arg->thread_ctx->thread_state = new_state;
+
+  /* Wait all threads to get the barrier */
+  pthread_barrier_wait(signal_struct->barrier);
+  int num_threads_to_wait_before = atomic_fetch_sub(&signal_struct->num_threads_to_wait, 1);
+  if (num_threads_to_wait_before - 1 == 0) {
```
GuyAv46:
```suggestion
  int num_threads_to_wait_before = atomic_sub_fetch(&signal_struct->num_threads_to_wait, 1);
  if (num_threads_to_wait_before == 0) {
```

deps/thpool/thpool.c
JoanFM: will us RedisModule_Assert

deps/thpool/thpool.c
JoanFM: this was a leftover from an old implementation, is not needed now

tests/cpptests/test_cpp_thpool.cpp
JoanFM: This comes from a confusion from the `remove_all`argument. Maybe I should rename it. 

The `remove_all` wants to mean that REMOVE_WHEN_EMPTY should be used, but this is a proxy. I will change the name of the argument and will make more sense

tests/pytests/test_avoid_deadlock_while_changing_num_workers.py
JoanFM: if we sleep 3 seconds and no success is seen, something strange would have happened.

tests/pytests/test_avoid_deadlock_while_changing_num_workers.py
JoanFM: I am reducing some of the sleeping time

deps/thpool/thpool.c
```diff
@@ -1012,3 +1023,76 @@ static void redisearch_thpool_broadcast_new_state(redisearch_thpool_t *thpool,
...
+    pthread_barrier_destroy(signal_struct->barrier);
+    rm_free(signal_struct->barrier);
+    rm_free(signal_struct);
+  }
+}
+
+void redisearch_thpool_schedule_config_reduce_threads_job(redisearch_thpool_t *thpool_p, size_t n_threads_to_remove, bool remove_all) {
```
GuyAv46: Can't we deduce the `remove_all` from the number of threads we remove?

tests/pytests/test_multithread.py
```diff
@@ -503,47 +506,64 @@ def check_threads(expected_num_threads_alive, expected_n_threads):
...
+    time.sleep(1)
     check_threads(expected_num_threads_alive=0, expected_n_threads=0)
 
     # Query should not be executed by the threadpool
-    env.expect('FT.CREATE', 'idx', 'SCHEMA', 't', 'text').ok()
+
     env.expect('ft.search', 'idx', '*').equal([0])
     check_threads(expected_num_threads_alive=0, expected_n_threads=0)
-    env.assertEqual(getWorkersThpoolStats(env)['totalJobsDone'], 0)
+    env.assertEqual(getWorkersThpoolStats(env)['totalJobsDone'], num_query_threads)
```
GuyAv46, line 537: Why has this changed? We previously wanted admin jobs not to affect this number

tests/pytests/test_multithread.py
JoanFM: Changed to use TimeLimit and checking when the condition wanted is checked

tests/pytests/test_avoid_deadlock_while_changing_num_workers.py
JoanFM: Changed to use TimeLimit while waiting for the condition

tests/pytests/test_avoid_deadlock_while_changing_num_workers.py
JoanFM: I still think some more coverage is nice, so keeping this in test_multithread.py

tests/pytests/test_avoid_deadlock_while_changing_num_workers.py
JoanFM: Also no sleeping time, but TimeLimit and waiting for some condition to reduce the real time while making it less flaky

deps/thpool/thpool.c
JoanFM: Changed to RedisModule_Assert

deps/thpool/thpool.c
JoanFM: changed the naming, it was wrong, the signal is about terminating_when_empty

tests/pytests/test_multithread.py
JoanFM: because we have added query threads. Notice the jobs are based on query threads, not on ADMIN jobs

deps/thpool/thpool.c
JoanFM: will check about volatile. I personally like this individual signaling better, but would be equivalent.

deps/thpool/thpool.c
JoanFM: Personally prefer this way, but can change

deps/thpool/thpool.c
JoanFM: yes Could not allocate memory. I just see this pattern used in this file, but I agree this is the only place in the code where we properly report if we fail to allocate memory. Can remove if u want

deps/thpool/thpool.c
JoanFM: For the volatile, I was worried about optimizing away the NULL setting, but maybe this is dangerous as well (but I thought this would be found in tests)

deps/thpool/thpool.c
alonre24: No, `RedisModule_Assert` will crash in production. We rather using `RS_ASSERT` which asserts in debug and in our tests in CI only.

deps/thpool/thpool.c
JoanFM: atomic_sub_fetch seems not to compile for me

JoanFM: Will keep a variant of this, prefer to use the stdatomic ones.

alonre24 (approved):

deps/thpool/thpool.c
```diff
@@ -447,6 +435,15 @@ static void redisearch_thpool_push_chain(
   redisearch_thpool_unlock(thpool_p);
 }
 
+static void redisearch_thpool_push_chain_verify_init_threads(
+  redisearch_thpool_t *thpool_p, job *f_newjob_p, job *l_newjob_p, size_t n,
+  thpool_priority priority) {
+redisearch_thpool_push_chain(thpool_p, f_newjob_p, l_newjob_p, n, priority);
+
+/* Initialize threads if needed */
+redisearch_thpool_verify_init(thpool_p);
```
GuyAv46:
```suggestion
  redisearch_thpool_push_chain(thpool_p, f_newjob_p, l_newjob_p, n, priority);

  /* Initialize threads if needed */
  redisearch_thpool_verify_init(thpool_p);
```

deps/thpool/thpool.c
```diff
@@ -1058,23 +1055,19 @@ void redisearch_thpool_schedule_config_reduce_threads_job(redisearch_thpool_t *t
...
 
   size_t n_threads = thpool_p->n_threads;
   LOG_IF_EXISTS("verbose", "Scheduling from main thread to remove %zu threads", n_threads_to_remove);
-  assert((!remove_all || n_threads_to_remove == n_threads) && "If remove_all is set, n_threads_to_remove must be equal to n_threads");
-  assert(thpool_p->n_threads >= n_threads_to_remove && "Number of threads can't be negative");
-  assert(thpool_p->jobqueues.state == JOBQ_RUNNING && "Can't remove threads while jobq is paused");
-  assert(thpool_p->num_threads_alive >= n_threads_to_remove && "Not enough alive threads to remove");
-
-  ThreadState new_state = remove_all ? THREAD_TERMINATE_WHEN_EMPTY : THREAD_TERMINATE_ASAP;
+  RS_ASSERT((!terminate_when_empty || n_threads_to_remove == n_threads) && "If remove_all is set, n_threads_to_remove must be equal to n_threads");
```
GuyAv46:
```suggestion
  RS_LOG_ASSERT((!terminate_when_empty || n_threads_to_remove == n_threads), "If remove_all is set, n_threads_to_remove must be equal to n_threads");
```

deps/thpool/thpool.c
GuyAv46: Individual flags require aggregation at the waiting thread, so a counter seems simpler. But not mandatory

deps/thpool/thpool.c
GuyAv46: Because this file was copied from some external implementation, and `malloc` may fail. But `rm_malloc` [cannot fail](https://redis.io/docs/latest/develop/reference/modules/modules-api-ref/#redismodule_alloc). Up to you

deps/thpool/thpool.c
GuyAv46: I see, there is `__atomic_fetch_sub` of Clang and GCC. But it's cosmetic mostly

deps/thpool/thpool.c
GuyAv46: We can also deduce `terminating_when_empty` by the number (if we reduce to 0), no?

tests/pytests/test_multithread.py
```diff
@@ -503,47 +506,64 @@ def check_threads(expected_num_threads_alive, expected_n_threads):
...
+        t.start()
+        query_threads.append(t)
+
+    for t in query_threads:
+        t.join()
+
+    check_threads(expected_num_threads_alive=1, expected_n_threads=1)
     # Set it to 0
     env.expect(config_cmd(), 'SET', 'WORKERS', '0').ok()
+    time.sleep(1)
```
GuyAv46: Can we use `with TimeLimit` on some condition instead? Seems like a pending flakiness

JoanFM: Actually done, check the most recent change

deps/thpool/thpool.c
```diff
@@ -1058,23 +1055,19 @@ void redisearch_thpool_schedule_config_reduce_threads_job(redisearch_thpool_t *t
...
   LOG_IF_EXISTS("verbose", "Scheduling from main thread to remove %zu threads", n_threads_to_remove);
-  assert((!remove_all || n_threads_to_remove == n_threads) && "If remove_all is set, n_threads_to_remove must be equal to n_threads");
-  assert(thpool_p->n_threads >= n_threads_to_remove && "Number of threads can't be negative");
-  assert(thpool_p->jobqueues.state == JOBQ_RUNNING && "Can't remove threads while jobq is paused");
-  assert(thpool_p->num_threads_alive >= n_threads_to_remove && "Not enough alive threads to remove");
-
-  ThreadState new_state = remove_all ? THREAD_TERMINATE_WHEN_EMPTY : THREAD_TERMINATE_ASAP;
+  RS_LOG_ASSERT((!terminate_when_empty || n_threads_to_remove == n_threads), "If remove_all is set, n_threads_to_remove must be equal to n_threads");
+  RS_ASSERT(thpool_p->n_threads >= n_threads_to_remove && "Number of threads can't be negative");
+  RS_ASSERT(thpool_p->jobqueues.state == JOBQ_RUNNING && "Can't remove threads while jobq is paused");
```
GuyAv46: can we do it for these as well?
```suggestion
  RS_LOG_ASSERT(thpool_p->n_threads >= n_threads_to_remove, "Number of threads can't be negative");
  RS_LOG_ASSERT(thpool_p->jobqueues.state == JOBQ_RUNNING, "Can't remove threads while jobq is paused");
```

GuyAv46 (dismissed):

PR #7420, opened 2025-11-19
fix: remove draining from Flush callback

JoanFM:

Remove Draining from onFlush callback

---

> [!NOTE]
> Stop draining the workers thread pool in `onFlush` and add a test that exercises querying during `FLUSHALL` and after creating a new index.
> 
> - **Flush handling**:
>   - Remove call to `workersThreadPool_Drain(ctx, 0)` from `onFlush` in `src/spec.c`.
> - **Tests**:
>   - Add `tests/pytests/test_query_while_flush.py` to continuously query an index before/after `FLUSHALL`, verify expected errors post-flush, and validate new index queries succeed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3b10983dab87b82c791db38ea7ca638359d83074. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_query_while_flush.py
+++ b/tests/pytests/test_query_while_flush.py
@@ -2,7 +2,6 @@ import threading
 import time
 from common import *
 
-@skip(cluster=True)
 def test_query_while_flush():
     """
     Test scenario:
@@ -20,7 +19,7 @@ def test_query_while_flush():
 
     # Add 100 documents to index1
     for i in range(100):
-        env.cmd('HSET', f'doc:{i}', 'text', f'hello world document {i}')
+        env.getClusterConnectionIfNeeded().execute_command('HSET', f'doc:{i}', 'text', f'hello world document {i}')
 
     # Wait for indexing to complete
     waitForIndex(env, 'index1')
@@ -110,7 +109,7 @@ def test_query_while_flush():
 
     # Add some documents to index2
     for i in range(10):
-        env.cmd('HSET', f'newdoc:{i}', 'text', f'new document {i}')
+        env.getClusterConnectionIfNeeded().execute_command('HSET', f'newdoc:{i}', 'text', f'new document {i}')
 
     # Wait for indexing to complete
     waitForIndex(env, 'index2')


alonre24 (commented):
What would happen if we call `flushall` while jobs are pending (due to not draining), and then we immediately create a new index and push new jobs? Should we test such a scenario?

JoanFM:
> What would happen if we call `flushall` while jobs are pending (due to not draining), and then we immediately create a new index and push new jobs? Should we test such a scenario?

Okey, but I guess the expected behavior is that the queries that were queued before and during FLUSHALL will fail and the subsequent ones should succeed right?

JoanFM:
> What would happen if we call `flushall` while jobs are pending (due to not draining), and then we immediately create a new index and push new jobs? Should we test such a scenario?

What seems weird to me is that we are draining but after we have freed spectdict_g, so isn't there a chance that one of the workers could try to access an invalid struct?

tests/pytests/test_query_while_flush.py
```diff
@@ -0,0 +1,133 @@
...
+        thread.start()
+
+    # Let queries run for a bit to accumulate some successes
+    time.sleep(0.5)
+
+    # Verify we have some successes before flush
+    stats['flushall_called'] = True
+    time.sleep(0.5)
+    env.assertGreater(stats['before_flush_successes'], 0)
+    env.assertEqual(stats['before_flush_errors'], 0)
```
GuyAv46, line 93: Use `TimeLimit` instead of sleeping

alonre24:
> > What would happen if we call `flushall` while jobs are pending (due to not draining), and then we immediately create a new index and push new jobs? Should we test such a scenario?
> 
> Okey, but I guess the expected behavior is that the queries that were queued before and during FLUSHALL will fail and the subsequent ones should succeed right?

yes.

alonre24:
> > What would happen if we call `flushall` while jobs are pending (due to not draining), and then we immediately create a new index and push new jobs? Should we test such a scenario?
> 
> What seems weird to me is that we are draining but after we have freed spectdict_g, so isn't there a chance that one of the workers could try to access an invalid struct?

Only the main thread is allowed to access the `specdict_g` struct AFAIK

tests/pytests/test_query_while_flush.py
```diff
@@ -0,0 +1,133 @@
+import threading
+import time
+from common import *
+
+@skip(cluster=True)
+def test_query_while_flush(env):
```
alonre24: We should create an `env` with workers, no?
Also consider using `getWorkersThpoolStats` for validation

tests/pytests/test_query_while_flush.py
```diff
@@ -0,0 +1,133 @@
...
+        'stop_queries': False,
+        'flush_completed': False,
+        'flushall_called': False
+    }
+
+    def query_worker(stats):
+        """Worker thread that continuously queries the index"""
+        local_conn = env.getClusterConnectionIfNeeded()
+
+        while not stats['stop_queries']:
```
alonre24, line 48: Does Python guarantee to synchronize `stats` between threads natively?

tests/pytests/test_query_while_flush.py
JoanFM: oh sure, will create with WORKERS

tests/pytests/test_query_while_flush.py
JoanFM: well is not atomic, will try to make it more robust

JoanFM: so stats are not properly synchronized, but I only care about "0" and not "0", for this we can rely on the stats. 

The only thing I will check to make more robust is the flushall_called

tests/pytests/test_query_while_flush.py
JoanFM: but here I want to sleep, I do not want to do anything with a limit, I think it is different purpose

PR #7423, opened 2025-11-19
Fix Fork GC potential double-free on error path - [MOD-12521]

GuyAv46:
## Fix polluting out-pointers on `FGC_recvBuffer` when error occurs

Currently, if `FGC_recvBuffer`, it may have set the out pointer even if it fails. This makes the memory management unclear, as `FGC_recvBuffer` frees the buffer it allocates, but if the caller sees that the out pointer is not `NULL`, it may attempt to free it as well

Solution: don't set the out parameters unless the call succeeds.

This PR also improves the log printed when polling from the child fails/times out.

#### Mark if applicable

---

> [!NOTE]
> Strengthens Fork GC read/recv error paths with clearer logging and safer out-params, initializes pollfd, and adds tests that simulate pipe failures during GC/apply.
> 
> - **Fork GC core**:
>   - Improve `FGC_recvFixed` polling/error handling and add detailed log (incl. `revents`).
>   - Rework `FGC_recvBuffer` to use a temp length and set out-params only on success; fix potential double-free.
>   - Initialize `gc->pollfd_read` after pipe creation; minor safety cleanups (e.g., init `fieldName`).
> - **Tests**:
>   - Add `testPipeErrorDuringGC` and `testPipeErrorDuringApply` to simulate pipe closure/errors (including threaded timing) and verify graceful handling without crashes or double-frees.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cef56a7758f4417332a9eaf958ff18f206813715. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/cpptests/test_cpp_forkgc.cpp
+++ b/tests/cpptests/test_cpp_forkgc.cpp
@@ -29,6 +29,8 @@ extern "C" {
 #include <set>
 #include <random>
 #include <unordered_set>
+#include <thread>
+
 /**
  * The following tests purpose is to make sure the garbage collection is working properly,
  * without causing any data corruption or loss.


src/fork_gc.c
```diff
@@ -98,48 +98,53 @@ static void FGC_sendTerminator(ForkGC *fgc) {
...
     }
     if (len == 0) {
       return REDISMODULE_OK;
     }
   }
-  RedisModule_Log(fgc->ctx, "warning", "ForkGC - got timeout while reading from pipe (%s)", strerror(errno));
+  short revents = fgc->pollfd_read[0].revents;
+  const char *what = (poll_rc == 0) ? "timeout" : "error";
+  RedisModule_Log(fgc->ctx, "warning", "ForkGC - got %s while reading from pipe. errno: %s, revents: 0x%x (POLLIN=%x POLLERR=%x POLLHUP=%x POLLNVAL=%x)",
+                  what, strerror(errno), revents, (revents & POLLIN), (revents & POLLERR), (revents & POLLHUP), (revents & POLLNVAL));
```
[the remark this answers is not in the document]
GuyAv46: Better to see the value than not

src/fork_gc.c
```diff
@@ -98,48 +98,53 @@ static void FGC_sendTerminator(ForkGC *fgc) {
 
 static int __attribute__((warn_unused_result)) FGC_recvFixed(ForkGC *fgc, void *buf, size_t len) {
   // poll the pipe, so that we don't block while read, with timeout of 3 minutes
-  while (poll(fgc->pollfd_read, 1, 180000) == 1) {
+  int poll_rc;
```
JoanFM, line 101: could we move these changes to a different PR for clarification?

src/fork_gc.c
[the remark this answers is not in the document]
JoanFM: I do not think this is minor, it says that in this code path, these messages do not provide value. I think misleading information is worse than no information

src/fork_gc.c
GuyAv46: I think it's relevant. @alonre24 WDYT?

src/fork_gc.c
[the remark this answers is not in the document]
GuyAv46: 1. The AI is wrong - this value is always set
2. We now have accurate context - if it says timeout had occurred, it's a timeout (and the errno value is meaningless). Otherwise, it's an error (and all the codes are relevant)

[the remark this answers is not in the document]
JoanFM: ok if AI is wrong no prob

src/fork_gc.c
```diff
@@ -744,7 +749,7 @@ static FGCError FGC_parentHandleNumeric(ForkGC *gc) {
 
 static FGCError FGC_parentHandleTags(ForkGC *gc) {
   size_t fieldNameLen;
-  char *fieldName;
+  char *fieldName = NULL;
```
alonre24, line 752: Is this where we had the invalid free? https://github.com/RediSearch/RediSearch/pull/7423/files#diff-fdfd94118f1dfc4f48a70f8e1cf6dc2174d164a47054794e6076e89e146ffc22R845

src/fork_gc.c
```diff
@@ -111,35 +112,40 @@ static int __attribute__((warn_unused_result)) FGC_recvFixed(ForkGC *fgc, void *
...
 FGC_recvBuffer(ForkGC *fgc, void **buf, size_t *len) {
-  TRY_RECV_FIXED(fgc, len, sizeof *len);
-  if (*len == SIZE_MAX) {
+  size_t temp_len;
+  if (FGC_recvFixed(fgc, &temp_len, sizeof temp_len) != REDISMODULE_OK) {
+    return REDISMODULE_ERR;
+  }
+  if (temp_len == SIZE_MAX) {
+    *len = temp_len;
     *buf = RECV_BUFFER_EMPTY;
```
eyalrund, line 132: why is that?

src/fork_gc.c
alonre24: Quite relevant, we can leave it here

src/fork_gc.c
```diff
@@ -98,48 +98,53 @@ static void FGC_sendTerminator(ForkGC *fgc) {
 
 static int __attribute__((warn_unused_result)) FGC_recvFixed(ForkGC *fgc, void *buf, size_t len) {
   // poll the pipe, so that we don't block while read, with timeout of 3 minutes
-  while (poll(fgc->pollfd_read, 1, 180000) == 1) {
+  int poll_rc;
+  while ((poll_rc = poll(fgc->pollfd_read, 1, 180000)) == 1) {
```
eyalrund, line 102: why hard code 3 min here? cant we take it from somewhere?

alonre24, line 102: Just to confirm, the first time we call `poll` here from the parent, should it return only after the child has finished all of his work? Or can it run in parallel?

src/fork_gc.c
GuyAv46: One of them, potentially

src/fork_gc.c
GuyAv46: It could be in parallel. We are only blocked until we read the number of bytes we wanted

GuyAv46: We can do whatever we want.
This was meant to be a safety net, so we are not blocked forever if the child is not responding. Under good behavior, we should never reach the timeout.
Should I make it configurable?

eyalrund: I wouled move it to a const

PR #7441, opened 2025-11-20
Fix GC regression - [MOD-12538]

GuyAv46:

We added `RedisModule_SendChildHeartbeat` calls #5503, after **every inverted index scanned** (even if it wasn't changed).
This was recently proved to slow the process down significantly - sub-second tasks take minutes to complete. 

The goal was to get a print for the CoW info before we exit, so as we expect GC runs to be short, we will now call the heartbeat API once, right before finishing.
We add another terminal marker to the protocol, so the parent will wait until this info is logged before attempting to kill the child.

#### Mark if applicable

---

> [!NOTE]
> Consolidates ForkGC child heartbeats to a single final heartbeat and adds a terminal marker so the parent waits before termination, removing progress tracking.
> 
> - **ForkGC child flow**:
>   - Remove per-scan progress heartbeats and related calls.
>   - Send a single final `RedisModule_SendChildHeartbeat(1.0)` after all scans complete.
>   - Emit an extra final `FGC_sendTerminator` to mark post-processing completion.
> - **Parent/child protocol**:
>   - Parent now waits for and verifies the final terminator (`SIZE_MAX`) before finishing, ensuring the heartbeat is logged.
> - **Cleanup**:
>   - Delete progress reporting helpers and the `ForkGC.progress` field.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 42b809066c685a4dedb9cc6c9384a5377cc692a8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -167,15 +167,6 @@ static void sendHeaderString(void* ptrCtx) {
   FGC_sendBuffer(ctx->gc, iov->iov_base, iov->iov_len);
 }
 
-static void FGC_reportProgress(ForkGC *gc) {
-  RedisModule_SendChildHeartbeat(gc->progress);
-}
-
-static void FGC_setProgress(ForkGC *gc, float progress) {
-  gc->progress = progress;
-  FGC_reportProgress(gc);
-}
-
 static void FGC_childCollectTerms(ForkGC *gc, RedisSearchCtx *sctx) {
   TrieIterator *iter = Trie_Iterate(sctx->spec->terms, "", 0, 0, 1);
   rune *rstr = NULL;
@@ -198,8 +189,6 @@ static void FGC_childCollectTerms(ForkGC *gc, RedisSearchCtx *sctx) {
           &wr, sctx, idx,
           &cb, NULL
       );
-
-      FGC_reportProgress(gc);
     }
     rm_free(term);
   }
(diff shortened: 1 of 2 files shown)


src/fork_gc.c
```diff
@@ -457,18 +441,15 @@ static void FGC_childScanIndexes(ForkGC *gc, IndexSpec *spec) {
...
-  FGC_setProgress(gc, 0.2);
   FGC_childCollectNumeric(gc, &sctx);
-  FGC_setProgress(gc, 0.4);
   FGC_childCollectTags(gc, &sctx);
-  FGC_setProgress(gc, 0.6);
   FGC_childCollectMissingDocs(gc, &sctx);
-  FGC_setProgress(gc, 0.8);
   FGC_childCollectExistingDocs(gc, &sctx);
-  FGC_setProgress(gc, 1);
+  RedisModule_SendChildHeartbeat(1.0); // final heartbeat
```
alonre24, line 449: Should we keep these 5 times of `FGC_setProgress`  and call `RedisModule_SendChildHeartbeat` rather than removing it completely?

alonre24 (commented):
Can we add your micro benchmark to this PR?

src/fork_gc.c
GuyAv46: I tried that, but redis tries to optimize the calls, and we get only the first CoW message (with 0MB always).
The whole point is to see how much we consumed, so if we call it less than 100 times, we'd better call it once at the end

PR #7492, opened 2025-11-24
[MOD-12647] fix: handle the case in Coordinator when SCORE is sent alone without extra fields.

JoanFM:

In case of Loading Document fails because key is expired or trimmed, Hybrid Result serializer is sending scores even when documents are not loaded.

This breaks the invairants in the Coordinator leading to a potential crash.

---

> [!NOTE]
> Gracefully handle RESP3 rows with scores but no fields in the coordinator, update hybrid response parsing for RESP3, and add FT.HYBRID expiration tests for RESP2/RESP3.
> 
> - **Coordinator (`src/coord/rpnet.c`)**:
>   - Handle cases where RESP3 results may include `score` without `extra_attributes` (e.g., expired docs): add `has_fields` guard and skip field writes when absent.
>   - Relax/adjust assertions for RESP2/RESP3 to validate fields only when present; keep score type checks.
> - **Tests**:
>   - Update `get_results_from_hybrid_response` to parse RESP3 dict replies and return results with `total_results`.
>   - Add FT.HYBRID expiration tests (`test_expire_ft_hybrid_resp2`/`resp3`) verifying only non-expired docs are returned with expected attributes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 451305b76e915abfe139d6adf97ac0b7d00edce5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/coord/rpnet.c
+++ b/src/coord/rpnet.c
@@ -347,7 +347,7 @@ int rpnetNext(ResultProcessor *self, SearchResult *r) {
 
   MRReply *score = NULL;
   MRReply *fields = MRReply_ArrayElement(rows, nc->curIdx++);
-  bool has_fields = true;
+  bool has_fields = false;
   if (resp3) {
     RS_LOG_ASSERT(fields && MRReply_Type(fields) == MR_REPLY_MAP, "invalid result record");
     // extract score if it exists, WITHSCORES was specified
@@ -357,8 +357,8 @@ int rpnetNext(ResultProcessor *self, SearchResult *r) {
     // we do not have keys to return.
     has_fields = fields && MRReply_Type(fields) == MR_REPLY_MAP;
   } else {
-    RS_LOG_ASSERT(fields && MRReply_Type(fields) == MR_REPLY_ARRAY, "invalid result record");
-    RS_LOG_ASSERT(MRReply_Length(fields) % 2 == 0, "invalid fields record");
+    has_fields = fields && MRReply_Type(fields) == MR_REPLY_ARRAY;
+    RS_LOG_ASSERT(!has_fields || has_fields && MRReply_Length(fields) % 2 == 0, "invalid fields record");
   }
 
   // The score is optional, in hybrid we need the score for the sorter and hybrid merger
--- a/tests/pytests/common.py
+++ b/tests/pytests/common.py
@@ -1000,7 +1000,16 @@ def get_results_from_hybrid_response(response) -> Dict[str, Dict[str, any]]:
         Dict mapping key -> dict of all fields from the results list
         Example: {'doc:1': {'__score': '0.5', 'vector_distance': '0.3'}}
     """
-    # return dict mapping key -> all fields from the results list
+    # Handle RESP3 format (dict)
+    if isinstance(response, dict):
+        results = {}
+        for result in response.get('results', []):
+            if '__key' in result:
+                key = result['__key']
+                results[key] = result
+        total_results = response.get('total_results', 0)
+        return results, total_results
+
     res_results_index = recursive_index(response, 'results')
     res_count_index = recursive_index(response, 'total_results')
     res_results_index[-1] += 1
--- a/tests/pytests/test_expire.py
+++ b/tests/pytests/test_expire.py
@@ -254,8 +254,9 @@ def test_expire_aggregate(env):
     env.assertEqual(res, [1, ['t', 'arr'], ['t', 'bar']])
 
 
-def test_expire_ft_hybrid(env):
-    # Use "lazy" expire (expire only when key is accessed) on all shards
+def expire_ft_hybrid_test(protocol):
+    env = Env(protocol=protocol)
+        # Use "lazy" expire (expire only when key is accessed) on all shards
     env.cmd('DEBUG', 'SET-ACTIVE-EXPIRE', '0')
 
     # Create index with text, vector, and numeric fields
@@ -311,6 +312,12 @@ def test_expire_ft_hybrid(env):
         env.assertEqual(actual_results_dict[doc_key]['t'], f'text{doc_num}')
         env.assertEqual(actual_results_dict[doc_key]['n'], str(doc_num))
         env.assertTrue(float(actual_results_dict[doc_key]['__score']) >= 0)
+
+def test_expire_ft_hybrid_resp2():
+    expire_ft_hybrid_test(protocol=2)
+
+def test_expire_ft_hybrid_resp3():
+    expire_ft_hybrid_test(protocol=3)
 
 def createTextualSchema(field_to_additional_schema_keywords):
     schema = []


src/hybrid/hybrid_exec.c
```diff
@@ -102,67 +102,65 @@ static void serializeResult_hybrid(HybridRequest *hreq, RedisModule_Reply *reply
...
+            // Single
+            v = RSValue_Trio_GetLeft(v);
           }
+        } else {
+          // EXPAND
+          v = RSValue_Trio_GetRight(v);
         }
-        RedisModule_Reply_RSValue(reply, v, flags);
       }
+      RedisModule_Reply_RSValue(reply, v, flags);
```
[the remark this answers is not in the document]
JoanFM: NOFIELD right now for Hybrid is only active when LIMIT 0 0 is set, and therefore limit is 0.

src/hybrid/hybrid_exec.c
```diff
@@ -98,74 +98,69 @@ static void serializeResult_hybrid(HybridRequest *hreq, RedisModule_Reply *reply
...
+        RedisModule_Reply_ArrayEnd(reply);
+      }
     }
-  }
-
-  if (!(options & QEXEC_F_SEND_NOFIELDS)) {
     const RLookup *lk = cv->lastLookup;
 
-    if (SearchResult_GetFlags(r) & Result_ExpiredDoc) {
-      RedisModule_Reply_Null(reply);
```
kei-nan, line 123: I think not replying null is problematic from a count perspective.
I remember there were issues with that, maybe it was only in resp2 but worth checking.
Why continue in this function if the doc was expired?

tests/pytests/test_expire.py
```diff
@@ -253,6 +253,69 @@ def test_expire_aggregate(env):
     # The result count is not accurate in aggregation, for now we compare res to the expected results with the wrong count
     env.assertEqual(res, [1, ['t', 'arr'], ['t', 'bar']])
 
+
+def test_expire_ft_hybrid(env):
+    # Use "lazy" expire (expire only when key is accessed) on all shards
+    if env.isCluster():
+        run_command_on_all_shards(env, 'DEBUG', 'SET-ACTIVE-EXPIRE', '0')
```
kei-nan: can't this run even if we aren't a cluster?
do we need this if?

tests/pytests/test_expire.py
```diff
@@ -253,6 +253,69 @@ def test_expire_aggregate(env):
...
+    else:
+        env.cmd('DEBUG', 'SET-ACTIVE-EXPIRE', '0')
+
+    # Create index with text, vector, and numeric fields
+    env.expect('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT', 'n', 'NUMERIC', 'v', 'VECTOR', 'FLAT', '6', 'TYPE', 'FLOAT32', 'DIM', '2', 'DISTANCE_METRIC', 'L2').ok()
+    env.cmd(debug_cmd(), 'SET_MONITOR_EXPIRATION', 'idx', 'not-documents')
```
kei-nan: why are you disabling expiration monitoring?

kei-nan: why are you disabling expiration monitoring?

src/hybrid/hybrid_exec.c
JoanFM: But I am just not adding a result when it is expired. What do u mean by continuing?

tests/pytests/test_expire.py
JoanFM: I can remove, I guess AI took from expiration_aggregate test

JoanFM: removed

JoanFM: removed.

src/hybrid/hybrid_exec.c
kei-nan: I think in resp2 we set the result array length for example in advance.
if you are sending fewer results then the count needs to be corrected.
This is something which is facing the cli so the client could choose to use resp2.

So we need to ensure the array count is set correctly even when we avoid sending a result.

tests/pytests/test_expire.py
```diff
@@ -253,6 +253,65 @@ def test_expire_aggregate(env):
...
+                conn.execute_command('PEXPIRE', doc_key, 1)
+
+    # Ensure expiration before query
+    time.sleep(0.01)
+
+    # Test FT.HYBRID requesting 1000 results but expecting only 10 (non-expired documents)
+    hybrid_query = ['FT.HYBRID', 'idx', 'SEARCH', '*', 'VSIM', '@v', query_vector, 'LIMIT', '0', '1000', 'COMBINE', 'RRF', '2', 'CONSTANT', '60', 'LOAD', '4', '@__key', '@__score', '@t', '@n']
+
+    # Execute query using cluster-aware command to get expected results
+    expected_res = env.cmd(*hybrid_query)
```
kei-nan: is this the expected or the actual result?
consider renaming the expected variables to be "actual_"

JoanFM: sorry leftover to when I was comparing between different shards

src/coord/rpnet.c
```diff
@@ -347,12 +347,15 @@ int rpnetNext(ResultProcessor *self, SearchResult *r) {
...
+  bool has_fields = true;
   if (resp3) {
     RS_LOG_ASSERT(fields && MRReply_Type(fields) == MR_REPLY_MAP, "invalid result record");
     // extract score if it exists, WITHSCORES was specified
     score = MRReply_MapElement(fields, "score");
     fields = MRReply_MapElement(fields, "extra_attributes");
-    RS_LOG_ASSERT(fields && MRReply_Type(fields) == MR_REPLY_MAP, "invalid fields record");
+    // It could happen if Result_ExpiredDoc is set by the Loader on the shard, that no extra attributes is returned. In that case
+    // we do not have keys to return.
+    has_fields = fields && MRReply_Type(fields) == MR_REPLY_MAP;
```
kei-nan, line 358: can't the same thing happen for FT.AGGREGATE in resp2?
should we also fill out `has_fields` in resp2 case?

tests/pytests/test_expire.py
```diff
@@ -253,6 +253,65 @@ def test_expire_aggregate(env):
     # The result count is not accurate in aggregation, for now we compare res to the expected results with the wrong count
     env.assertEqual(res, [1, ['t', 'arr'], ['t', 'bar']])
 
+
+def test_expire_ft_hybrid(env):
```
kei-nan: in the end did you get this test to fail without your changes?

JoanFM: no I did not manage. But still see it as relevant missing test

oshadmi: If the fix is relevant for RESP3, shouldn't the test also run using RESP3?

JoanFM: isnt it the default here? how should I enforce this?

JoanFM: Added tests for both

src/coord/rpnet.c
JoanFM: Handled also for resp2 makes sense

src/coord/rpnet.c
```diff
@@ -347,15 +347,18 @@ int rpnetNext(ResultProcessor *self, SearchResult *r) {
 
   MRReply *score = NULL;
   MRReply *fields = MRReply_ArrayElement(rows, nc->curIdx++);
+  bool has_fields = true;
```
oshadmi: Nitpick: should be `false` or just declared? (set later for both RESP2 and RESP3)

tests/pytests/common.py
```diff
@@ -1000,7 +1000,16 @@ def get_results_from_hybrid_response(response) -> Dict[str, Dict[str, any]]:
...
+    # Handle RESP3 format (dict)
+    if isinstance(response, dict):
+        results = {}
+        for result in response.get('results', []):
+            if '__key' in result:
+                key = result['__key']
+                results[key] = result
+        total_results = response.get('total_results', 0)
+        return results, total_results
+
```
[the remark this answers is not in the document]
JoanFM: the signature did not change, it returns results, total_results in both cases.

src/coord/rpnet.c
JoanFM: setting to false, I like my variables initialized

src/coord/rpnet.c
```diff
@@ -366,12 +369,14 @@ int rpnetNext(ResultProcessor *self, SearchResult *r) {
...
-    size_t len;
-    const char *field = MRReply_String(MRReply_ArrayElement(fields, i), &len);
-    MRReply *val = MRReply_ArrayElement(fields, i + 1);
-    RSValue *v = MRReply_ToValue(val);
-    RLookup_WriteOwnKeyByName(nc->lookup, field, len, SearchResult_GetRowDataMut(r), v);
+  if (has_fields) {
+    for (size_t i = 0; i < MRReply_Length(fields); i += 2) {
+      size_t len;
+      const char *field = MRReply_String(MRReply_ArrayElement(fields, i), &len);
+      MRReply *val = MRReply_ArrayElement(fields, i + 1);
```
oshadmi, line 376: Calling `MRReply_ArrayElement` on a Map element is currently working since the underlying storage is an array, but it could be broken later on?

JoanFM: I did not analyze this. This has not been the purpose of this PR. The problem this PR aims to solve is the assertions that were not true in some cases.

oshadmi: Seems OK
https://github.com/redis/hiredis/blob/master/README.md#resp3

PR #7521, opened 2025-11-26
[MOD-12699] fix freeing searchRequestCtx on error

JoanFM:
> [!NOTE]
> Ensure `searchRequestCtx` is freed if the index spec is unavailable while preparing the distributed search command to prevent a memory leak.
> 
> - **Memory Management**
>   - Free `searchRequestCtx` in `prepareCommand` when `IndexSpec` is unavailable (dropped in background), preventing a leak during distributed search preparation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 60cb64643cfa1474d955d88390ca41da54aad6df. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -3493,13 +3493,10 @@ void sendRequiredFields(searchRequestCtx *req, MRCommand *cmd) {
   }
 }
 
-static void bailOut(RedisModuleBlockedClient *bc, QueryError *status, searchRequestCtx *req) {
+static void bailOut(RedisModuleBlockedClient *bc, QueryError *status) {
   RedisModuleCtx* clientCtx = RedisModule_GetThreadSafeContext(bc);
   QueryErrorsGlobalStats_UpdateError(QueryError_GetCode(status), 1, COORD_ERR_WARN);
   QueryError_ReplyAndClear(clientCtx, status);
-  if (req) {
-    searchRequestCtx_Free(req);
-  }
   RedisModule_BlockedClientMeasureTimeEnd(bc);
   RedisModule_UnblockClient(bc, NULL);
   RedisModule_FreeThreadSafeContext(clientCtx);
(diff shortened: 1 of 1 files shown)


src/module.c
```diff
@@ -3615,6 +3615,7 @@ int FlatSearchCommandHandler(RedisModuleBlockedClient *bc, int protocol,
   MRCommand cmd = MR_NewCommandFromRedisStrings(argc, argv);
   int rc = prepareCommand(&cmd, req, bc, protocol, argv, argc, spec_ref, &status);
   if (!(rc == REDISMODULE_OK)) {
+    SearchRequestCtx_Free(req);
```
meiravgri: consider adding to `bailOut`

src/module.c
```diff
@@ -3561,8 +3564,7 @@ static int prepareCommand(MRCommand *cmd, searchRequestCtx *req, RedisModuleBloc
   if (!sp) {
     MRCommand_Free(cmd);
     QueryError_SetCode(status, QUERY_ERROR_CODE_DROPPED_BACKGROUND);
-
-    bailOut(bc, status);
+    bailOut(bc, status, req);
```
GuyAv46: IMO it doesn't make sense to pass it to `bailOut`, if usually there is no `req` to pass. Also, why pass the `req` and not the `cmd` as well?

GuyAv46: Consider dropping the `req` next to the `cmd` instead of in `bailOut`.

GuyAv46: @meiravgri WDYT?

meiravgri: Not sure where is "next to cmd" ? 

I think i was wrong, and the previous location, after returning from `prepareCommand` was better
just need to make sure we set everything inside to NULL to avoid double frees

GuyAv46: Two lines above seems like the best place

JoanFM: done

PR #7534, opened 2025-11-26
[8.2] [MOD-12252] [MOD-11658] reduce number of worker threads asynchronously

JoanFM:
Backport #7410 to 8.2

---

> [!NOTE]
> Introduce asynchronous thread reduction in the thread pool with robust initialization signaling, update worker management to use it, and add tests to prevent deadlocks when reducing WORKERS under load.
> 
> - **Threadpool (`deps/thpool`)**:
>   - Add `redisearch_thpool_schedule_config_reduce_threads_job(...)` to reduce threads asynchronously via ADMIN jobs and barriers; set pool UNINITIALIZED on terminate-when-empty.
>   - Refactor init and enqueue paths: split `redisearch_thpool_push_chain(...)`, use per-thread `started` flags (via `thread_do_args`) to reliably wait for new threads; `thread_do` signature changed.
>   - Strengthen checks/logging (`RS_ASSERT`, improved messages); remove synchronous APIs from header (`redisearch_thpool_remove_threads`, `redisearch_thpool_terminate_when_empty`).
> - **Workers (`src/util/workers.c`)**:
>   - Switch thread count changes to new async reduction API; keep adds synchronous with start-wait; add notices and assertions; improve drain logging.
> - **Config/Debug**:
>   - Keep `WORKERS` changes triggering pool reconfig; add warning note for `FT.DEBUG WORKERS DRAIN`; minor assert macro swap.
> - **Tests**:
>   - Update C++ thpool tests to use async reduction and new init semantics.
>   - Add/expand Python tests to validate worker scaling (increase/decrease, zero↔nonzero), gradual reductions, and MOD-11658: avoid deadlock when setting `WORKERS` to 0 under load.
>   - Minor test robustness tweaks (thread joins, timing).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bf344b2a5401547885b16d77ee2c4395792bb671. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_issues.py
+++ b/tests/pytests/test_issues.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from common import *
 import random
+import threading
 
 def test_1282(env):
   conn = getConnectionByEnv(env)


tests/pytests/test_multithread.py
```diff
@@ -467,65 +467,216 @@ def expected(conns):
...
+def check_threads(env, expected_num_threads_alive, expected_n_threads):
+    env.assertEqual(getWorkersThpoolStats(env)['numThreadsAlive'], expected_num_threads_alive, depth=1, message='numThreadsAlive should match num_threads_alive')
+    env.assertEqual(getWorkersThpoolNumThreads(env), expected_n_threads, depth=1, message='n_threads should match WORKERS')
 
-    def check_threads(expected_num_threads_alive, expected_n_threads):
-        env.assertEqual(getWorkersThpoolStats(env)['numThreadsAlive'], expected_num_threads_alive, depth=1, message='numThreadsAlive should match num_threads_alive')
-        env.assertEqual(getWorkersThpoolNumThreads(env), expected_n_threads, depth=1, message='n_threads should match WORKERS')
+def test_change_workers_number():
+    def send_query():
+        env.expect('ft.search', 'idx', '*').equal([0])
```
GuyAv46, line 477: Is it OK to define this with `env` before we define `env` itself?

JoanFM: it is indeed a weird pattern, but I guess when it is called it is int he scope.

JoanFM: will fix it though, so that a potential change in Python version would not affect.

Will however apply this change to every branch indepndently, so let's continue with merging

PR #7547, opened 2025-11-27
Don't access `NULL` GC delta

chesedo:
The GC delta was freed from memory when it was applied a few lines before the `resetCardinality` call. This changes gets the last block ID in the delta before this apply call which frees it, and then passes this last block to `resetCardinality` instead.

This is for this redisearch `master` build error log -> https://github.com/redis/redis/issues/14570#issuecomment-3581813475

#### Mark if applicable

---

> [!NOTE]
> Prevents NULL access by computing numeric start position before applying GC deltas, corrects HLL reset to include ignored block, and standardizes delta ownership by nulling after apply and unconditionally freeing across handlers.
> 
> - **Numeric index/HLL**:
>   - Compute `blocksSinceFork` before `InvertedIndex_ApplyGcDelta` and pass to `resetCardinality`.
>   - In `resetCardinality`, include ignored block (`blocksSinceFork++`) and derive `startIdx` via `InvertedIndex_NumBlocks(...) - blocksSinceFork`.
> - **Delta ownership/lifecycle**:
>   - After `InvertedIndex_ApplyGcDelta(...)`, set `delta = NULL`/`ninfo->delta = NULL` and unconditionally `InvertedIndex_GcDelta_Free(delta)` in cleanup.
>   - Remove `shouldFreeDeltas` branches in term/tag/missing/existing docs handlers, standardizing freeing behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d42191025d8f0fdf59a7990638ac80bc307c34ba. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -498,7 +498,7 @@ error:
   return FGC_CHILD_ERROR;
 }
 
-static void resetCardinality(NumGcInfo *info, NumericRange *range, size_t blocksSinceFork, size_t gc_delta_last_block_idx) {
+static void resetCardinality(NumGcInfo *info, NumericRange *range, size_t blocksSinceFork) {
   if (info->info.blocks_ignored == 0) {
     hll_set_registers(&range->hll, info->registersWithLastBlock, NR_REG_SIZE);
     if (blocksSinceFork == 0) {
@@ -506,9 +506,10 @@ static void resetCardinality(NumGcInfo *info, NumericRange *range, size_t blocks
     }
   } else {
     hll_set_registers(&range->hll, info->registersWithoutLastBlock, NR_REG_SIZE);
+    blocksSinceFork++; // Count the ignored block as well
   }
   // Add the entries that were added since the fork to the HLL
-  size_t startIdx = gc_delta_last_block_idx;
+  size_t startIdx = InvertedIndex_NumBlocks(range->entries) - blocksSinceFork; // Here `blocksSinceFork` > 0
   const IndexBlock *startBlock = InvertedIndex_BlockRef(range->entries, startIdx);
   t_docId startId = IndexBlock_FirstId(startBlock);
   IndexDecoderCtx decoderCtx = {.tag = IndexDecoderCtx_None};
@@ -531,8 +532,7 @@ static void applyNumIdx(ForkGC *gc, RedisSearchCtx *sctx, NumGcInfo *ninfo) {
   NumericRangeNode *currNode = ninfo->node;
   InvertedIndexGcDelta *delta = ninfo->delta;
   II_GCScanStats *info = &ninfo->info;
-  size_t gc_delta_last_block_idx = GcScanDelta_LastBlockIdx(delta);
-  size_t blocksSinceFork = InvertedIndex_NumBlocks(currNode->range->entries) - gc_delta_last_block_idx - 1; // record before applying changes
+  size_t blocksSinceFork = InvertedIndex_NumBlocks(currNode->range->entries) - GcScanDelta_LastBlockIdx(delta) - 1; // record before applying changes
   InvertedIndex_ApplyGcDelta(currNode->range->entries, delta, info);
   ninfo->delta = NULL;
   currNode->range->invertedIndexSize += info->bytes_allocated;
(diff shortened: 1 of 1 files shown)


LukeMathWalker:
Can we get a regression test?

src/fork_gc.c
```diff
@@ -498,7 +498,7 @@ static FGCError recvNumIdx(ForkGC *gc, NumGcInfo *ninfo) {
   return FGC_CHILD_ERROR;
 }
 
-static void resetCardinality(NumGcInfo *info, NumericRange *range, size_t blocksSinceFork) {
+static void resetCardinality(NumGcInfo *info, NumericRange *range, size_t blocksSinceFork, size_t gc_delta_last_block_idx) {
```
GuyAv46: We don't need both `blocksSinceFork` and `gc_delta_last_block_idx`; this is a bit redundant.
We need them to find the `startIdx`, or to detect that we don't need to manually add the tail to the HLL.
Consider keeping only one of them, and use `InvertedIndex_NumBlocks(range->entries)`

src/fork_gc.c
```diff
@@ -508,7 +508,7 @@ static void resetCardinality(NumGcInfo *info, NumericRange *range, size_t blocks
     hll_set_registers(&range->hll, info->registersWithoutLastBlock, NR_REG_SIZE);
   }
   // Add the entries that were added since the fork to the HLL
-  size_t startIdx = GcScanDelta_LastBlockIdx(info->delta);
+  size_t startIdx = gc_delta_last_block_idx;
```
GuyAv46: This is a bug.
- We have 3 blocks before the fork (2 full, one partial). Last idx is 2
- GC fork deletes the 2 full blocks, modifies the last
- The main process modifies the last block
- Now `gc_delta_last_block_idx` is 2, `blocksSinceFork` is 0
- We apply the delta, so now we have 1 block
- We ignore the last block, so we need to add it manually
- `startIdx` should be 0! but we set it to 2

This is why we used to calculate the `startIdx` according to `blocksSinceFork` - the tail (unknown to the fork) stays the same, but we don't track the changes in the beginning of the index

GuyAv46: We can track the changing part, but it's more complicated. If we start merging or splitting blocks by the fork, we will need to track these changes as well.

The simple and robust calculation is to take the difference between the observed number of blocks before the fork and the number of blocks after (before applying the delta). Then maybe add 1 if we ignore the last block, and this is the number of trailing blocks we need to add to the HLL manually

chesedo:
> Can we get a regression test?

The `FGCTestNumeric.testNumericBlocksSinceFork` test already covered this code path. But it just read that freed memory without failing :sweat_smile: . I updated the code to explicitly set that pointer to `NULL` after it is freed. With this that test not fails trying to access a `NULL` pointer. This can be seen by running this test on the commit right before the fix.

```bash
./build.sh RUN_UNIT_TESTS TEST="FGCTestNumeric.testNumericBlocksSinceFork"
```

I've also updated the code to use the `NULL` state to correctly free that data.

src/fork_gc.c
chesedo: I don't see a way in using only one of them.

Suppose I only use `gc_delta_last_block_idx`, then I'll have to calculate `blocksSinceFork` in `resetCardinality`. But calling `InvertedIndex_NumBlocks(range->entries)` here to make that calculation can give incorrect results since the GC apply step might have reduced the number of blocks. That's why this variable is calculated before the apply step.

Now suppose I pass `blocksSinceFork` down and remove `gc_delta_last_block_idx` which I'll then calculate in `resetCardinality`. I won't be able to call `InvertedIndex_NumBlocks(range->entries)` to get the original last block idx since the GC apply might have changed the number of inverted index blocks on this path too.

So both cases of using only one of the variables results in possible errors when calculating the other one.

GuyAv46: But you don’t need to original number of blocks, you need the trailing blocks since the fork from the current number of blocks

src/fork_gc.c
chesedo: Ohh dang, I get what you are saying :sweat_smile: I also can't recall why I made this change. So reverted it back in https://github.com/RediSearch/RediSearch/pull/7547/commits/d42191025d8f0fdf59a7990638ac80bc307c34ba

src/fork_gc.c
chesedo: The other update caused me to only use one variable

src/fork_gc.c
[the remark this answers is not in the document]
chesedo: Yip, fixed in https://github.com/RediSearch/RediSearch/pull/7547/commits/d42191025d8f0fdf59a7990638ac80bc307c34ba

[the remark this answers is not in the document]
GuyAv46: Did the AI stole my reasoning?

PR #7548, opened 2025-11-27
feat: `tracing-redismodule`

JonasKruckenberg:
This change implements a subscriber for the Rust `tracing` ecosystem that emits traces and logs to the redismodule logging system.
Group 6

In its current state kept intentionally simple, it allows filtering the produced log messages via the standard `RUST_LOG` environment variable and enabling/disabling colored output via the `RUST_LOG_STYLE` (with automatic best guessing).

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

---

> [!NOTE]
> Introduces a Rust `tracing` subscriber that logs to Redis via a new FFI, initialized on module load, with env-based filtering/color and accompanying docs/tests.
> 
> - **Logging / Tracing**:
>   - Add `tracing_redismodule` crate and FFI `tracing_redismodule_ffi` exposing `TracingRedisModule_Init` with C header `src/redisearch_rs/headers/tracing_redismodule.h`.
>   - Integrate into C module: include `tracing_redismodule.h` and call `TracingRedisModule_Init(ctx)` in `RedisModule_OnLoad`.
>   - Configure via `RUST_LOG` (filtering) and `RUST_LOG_STYLE` (color); document usage and level mapping in `developer.md`.
>   - Add Python tests `tests/pytests/test_tracing.py` validating default level, `RUST_LOG=trace`, and crate-specific filtering.
> - **Build/Workspace**:
>   - Update workspace to include new crates; add `tracing`, `tracing-core`, `tracing-subscriber`, `anstyle-query` deps and export FFI in `c_entrypoint`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 55d0219a7af56341dead3a5480733305d9330faa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/Cargo.toml
+++ b/src/redisearch_rs/Cargo.toml
@@ -29,8 +29,7 @@ members = [
     "varint",
     "varint_bencher",
     "wildcard",
-    "workspace_hack",
-    "tracing_redismodule"
+    "workspace_hack"
 ]
 
 resolver = "3"
@@ -90,7 +89,6 @@ value = { path = "./value" }
 varint = { path = "./varint" }
 wildcard = { path = "./wildcard" }
 workspace_hack = { path = "./workspace_hack" }
-tracing_redismodule = { path = "./tracing_redismodule" }
 
 anstyle-query = "1.1.5"
 cbindgen = "0.29"
--- a/src/redisearch_rs/tracing_redismodule/Cargo.toml
+++ b/src/redisearch_rs/tracing_redismodule/Cargo.toml
@@ -11,6 +11,7 @@ ffi.workspace = true
 tracing.workspace = true
 tracing-core.workspace = true
 tracing-subscriber.workspace = true
+workspace_hack.workspace = true
 
 [lints]
 workspace = true
--- a/src/redisearch_rs/Cargo.lock
+++ b/src/redisearch_rs/Cargo.lock
@@ -2195,6 +2195,7 @@ dependencies = [
  "tracing",
  "tracing-core",
  "tracing-subscriber",
+ "workspace_hack",
 ]
 
 [[package]]
@@ -2205,6 +2206,7 @@ dependencies = [
  "cbindgen",
  "ffi",
  "tracing_redismodule",
+ "workspace_hack",
 ]
 
 [[package]]
@@ -2644,13 +2646,13 @@ name = "workspace_hack"
 version = "0.1.0"
 dependencies = [
  "bindgen 0.66.1",
- "bindgen 0.72.0",
- "bitflags 2.9.3",
+ "bindgen 0.72.1",
+ "bitflags 2.10.0",
  "clang-sys",
  "clap",
  "clap_builder",
  "either",
- "getrandom 0.3.3",
+ "getrandom 0.3.4",
  "libc",
  "memchr",
  "miniz_oxide",
@@ -2658,7 +2660,7 @@ dependencies = [
  "once_cell",
  "ppv-lite86",
  "proc-macro2",
- "quote 1.0.40",
+ "quote 1.0.42",
  "rand",
  "rand_chacha",
  "redis-module",
@@ -2668,8 +2670,9 @@ dependencies = [
  "serde",
  "serde_core",
  "syn 1.0.109",
- "syn 2.0.106",
+ "syn 2.0.111",
  "thiserror",
+ "zerocopy",
 ]
 
 [[package]]
--- a/src/redisearch_rs/c_entrypoint/tracing_redismodule_ffi/Cargo.toml
+++ b/src/redisearch_rs/c_entrypoint/tracing_redismodule_ffi/Cargo.toml
@@ -8,6 +8,7 @@ publish.workspace = true
 [dependencies]
 tracing_redismodule.workspace = true
 ffi.workspace = true
+workspace_hack.workspace = true
 
 [build-dependencies]
 cbindgen.workspace = true
--- a/src/redisearch_rs/workspace_hack/Cargo.toml
+++ b/src/redisearch_rs/workspace_hack/Cargo.toml
@@ -28,12 +28,13 @@ quote = { version = "1" }
 rand = { version = "0.9" }
 rand_chacha = { version = "0.9", default-features = false, features = ["std"] }
 regex = { version = "1" }
-regex-automata = { version = "0.4", default-features = false, features = ["dfa-onepass", "hybrid", "meta", "nfa-backtrack", "perf-inline", "perf-literal", "std", "unicode"] }
+regex-automata = { version = "0.4", default-features = false, features = ["dfa-build", "dfa-onepass", "hybrid", "meta", "nfa-backtrack", "perf-inline", "perf-literal", "std", "unicode"] }
 regex-syntax = { version = "0.8" }
 serde = { version = "1", features = ["alloc", "derive"] }
-serde_core = { version = "1", default-features = false, features = ["alloc", "result", "std"] }
+serde_core = { version = "1", features = ["alloc"] }
 syn-f595c2ba2a3f28df = { package = "syn", version = "2", features = ["extra-traits", "fold", "full", "visit", "visit-mut"] }
 thiserror = { version = "2" }
+zerocopy = { version = "0.8", default-features = false, features = ["derive", "simd"] }
 
 [build-dependencies]
 clap = { version = "4" }
@@ -49,13 +50,14 @@ quote = { version = "1" }
 rand = { version = "0.9" }
 rand_chacha = { version = "0.9", default-features = false, features = ["std"] }
 regex = { version = "1" }
-regex-automata = { version = "0.4", default-features = false, features = ["dfa-onepass", "hybrid", "meta", "nfa-backtrack", "perf-inline", "perf-literal", "std", "unicode"] }
+regex-automata = { version = "0.4", default-features = false, features = ["dfa-build", "dfa-onepass", "hybrid", "meta", "nfa-backtrack", "perf-inline", "perf-literal", "std", "unicode"] }
 regex-syntax = { version = "0.8" }
 serde = { version = "1", features = ["alloc", "derive"] }
-serde_core = { version = "1", default-features = false, features = ["alloc", "result", "std"] }
+serde_core = { version = "1", features = ["alloc"] }
 syn-dff4ba8e3ae991db = { package = "syn", version = "1", features = ["extra-traits", "full", "visit-mut"] }
 syn-f595c2ba2a3f28df = { package = "syn", version = "2", features = ["extra-traits", "fold", "full", "visit", "visit-mut"] }
 thiserror = { version = "2" }
+zerocopy = { version = "0.8", default-features = false, features = ["derive", "simd"] }
 
 [target.x86_64-unknown-linux-gnu.dependencies]
 bindgen-76a6958a3b5fde4 = { package = "bindgen", version = "0.72", default-features = false, features = ["runtime"] }


src/redisearch_rs/Cargo.toml
```diff
@@ -26,7 +26,7 @@ members = [
     "value",
     "varint",
     "varint_bencher",
-    "wildcard",
+    "wildcard", "tracing_redismodule",
```
enricozb: nit: should be on a newline

src/redisearch_rs/tracing_redismodule/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+//! consists of several parts:
+//!
+//! ```text
+//! target[span{field=value}]=level
+//! ```
+//!
+//! Each component (`target`, `span`, `field`, `value`, and `level`) will be covered in turn.
+//!
+//! - `target` matches the event or span's target. In general, this is the module path and/or crate name.
+//!   Examples of targets `h2`, `tokio::net`, or `tide::server`. For more information on targets,
```
GlenDC, line 27: should be not make these examples more in context of redis as I don't think we use any of these in the project here, or am I wrong?

src/redisearch_rs/tracing_redismodule/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+
+            let bytes_written = buf.write(input)?;
+
+            // NB: tracing subscriber always adds a trailing newline. We don't need this as the redismodule
+            // logging system already adds one for us as well. BUT we can use this to our advantage and change it
+            // for a NULL byte thereby making it a valid C string.
+            debug_assert_eq!(buf[bytes_written - 1], '\n' as u8);
+            buf[bytes_written - 1] = 0;
+
+            // Safety: Rust strings do not contain interior null bytes AND we just pushed the trailing zero.
```
enricozb: Is this true? This works on the playground for me

```rust
    let s = "Hello,\0world!".to_string();
    assert_eq!(s.chars().nth(6), Some(0 as char));
```

src/redisearch_rs/tracing_redismodule/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+            let cstr = unsafe { CStr::from_bytes_with_nul_unchecked(buf.as_mut()) };
+
+            // <https://redis.io/docs/latest/develop/reference/modules/modules-api-ref/#redismodule_log>
+            // Safety: The documentation explicitly allows ctx to be a nullptr, notice is a valid level
+            // and we ensured the C string is valid above.
+            unsafe {
+                ffi::RedisModule_Log.unwrap()(
+                    self.ctx.map_or(ptr::null_mut(), |ctx| ctx.as_ptr()),
+                    c"notice".as_ptr(), // one of "debug" "verbose" "notice" "warning"
```
enricozb: Should this be hard-coded?

src/redisearch_rs/tracing_redismodule/src/lib.rs
enricozb: I guess I'm not sure what the byte representation of a null UTF-8 codepoint is.

src/redisearch_rs/tracing_redismodule/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+/// Returns an Error if the initialization was unsuccessful, likely because
+/// a global subscriber was already installed by another call to `try_init`.
+pub fn try_init(
+    ctx: Option<NonNull<ffi::RedisModuleCtx>>,
+) -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
+    let env_filter = tracing_subscriber::filter::Builder::default()
+        .with_default_directive(LevelFilter::INFO.into())
+        .with_env_var("RUST_LOG")
+        .from_env()
+        .unwrap();
```
GlenDC: should these really unwrap if you already return a result anyway?

src/redisearch_rs/Cargo.toml
```diff
@@ -83,6 +83,12 @@ trie_rs = { path = "./trie_rs" }
 value = { path = "./value" }
 varint = { path = "./varint" }
 wildcard = { path = "./wildcard" }
+qint = { path = "./qint" }
+rlookup = { path = "./rlookup" }
+rqe_iterators = { path = "./rqe_iterators" }
+search_result = { path = "./search_result" }
+field = { path = "./field" }
```
enricozb: Why are these added here?

src/redisearch_rs/Cargo.toml
GlenDC: seems the convention in the file is to have it on its own line?

src/redisearch_rs/c_entrypoint/tracing_redismodule_ffi/cbindgen.toml
```diff
@@ -0,0 +1,7 @@
+language = "C"
+autogen_warning = "/* Warning, this file is autogenerated by cbindgen from `src/redisearch_rs/c_entrypoint/tracing_redismodule_ffi/build.rs. Don't modify it manually. */"
+cpp_compat = true
+pragma_once = true
+
+# [parse]
+# parse_deps = true
```
chesedo: We can remove this

src/redisearch_rs/Cargo.toml
JonasKruckenberg: mm weird merge thingy, resolved 👍🏻

src/redisearch_rs/tracing_redismodule/src/lib.rs
```diff
@@ -0,0 +1,238 @@
...
+//! consists of several parts:
+//!
+//! ```text
+//! target[span{field=value}]=level
+//! ```
+//!
+//! Each component (`target`, `span`, `field`, `value`, and `level`) will be covered in turn.
+//!
+//! - `target` matches the event or span's target. In general, this is the module path and/or crate name.
+//!   Examples of targets `h2`, `tokio::net`, or `tide::server`. For more information on targets,
```
chesedo: Can't this instead link to the `env_logger` documentation for setting up the filter?

src/redisearch_rs/tracing_redismodule/src/lib.rs
```diff
@@ -0,0 +1,238 @@
...
+///
+/// # Errors
+///
+/// Returns an Error if the initialization was unsuccessful, likely because
+/// a global subscriber was already installed by another call to `try_init`.
+pub fn try_init(
+    ctx: Option<NonNull<ffi::RedisModuleCtx>>,
+) -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
+    let env_filter = EnvFilter::builder()
+        .with_default_directive(LevelFilter::INFO.into())
```
chesedo, line 95: Will it be possible to get the `loglevel` Redis setting here instead for the default?

src/redisearch_rs/tracing_redismodule/src/lib.rs
chesedo: We might want to try an approach that allows us to correctly get the level here

src/redisearch_rs/tracing_redismodule/src/lib.rs
JonasKruckenberg: good catch, changed for the checked version

src/redisearch_rs/tracing_redismodule/src/lib.rs
JonasKruckenberg: Is that a setting that actually exists? I can't find this anywhere

src/redisearch_rs/tracing_redismodule/src/lib.rs
JonasKruckenberg: Its not actually `env_logger` syntax because this includes tracing specific extensions, the tracing docs aren't actually that helpful for ppl not already familiar with with logging so I chose to copy and adapt the docs here. But no hard preference either way, I'm fine with removing as well

JonasKruckenberg:
> I think the Writer might not be the approach we want here. 

Can you explain you reasoning for why? Building a custom layer is significantly more effort (esp if you want all the niceties of formatting) for not that much value (given that ppl will not be composing this probably ever) so the convenience outweighs the downsides quite dramatically IMO.

> Also is this just a first version or do we want to this complete based on our current understanding?

Not quite sure I understand what you mean 😂

src/redisearch_rs/tracing_redismodule/src/lib.rs
JonasKruckenberg: Changing it to emit a more proper level is trivial but my problem was that its unclear which level should map to what. We have 4 redis log levels and 5 Rust ones, the error level doesn't have a redis equivalent. Should we just map warn and error both to warning?

src/redisearch_rs/tracing_redismodule/src/lib.rs
chesedo: Sorry, I meant just linking to this instead after the small example -> https://docs.rs/tracing-subscriber/0.3.20/tracing_subscriber/filter/struct.EnvFilter.html

src/redisearch_rs/tracing_redismodule/src/lib.rs
chesedo: Yes, you can set it in the config (I've been using it with search on disk). Although I can find any documentation on this. A quick search just gives this

https://linuxcommandlibrary.com/man/redis-server

chesedo:
> > I think the Writer might not be the approach we want here.
> 
> Can you explain you reasoning for why? Building a custom layer is significantly more effort (esp if you want all the niceties of formatting) for not that much value (given that ppl will not be composing this probably ever) so the convenience outweighs the downsides quite dramatically IMO.

Only so that we can set the log level correctly which the `Writer` does not have access to

> 
> > Also is this just a first version or do we want to this complete based on our current understanding?
> 
> Not quite sure I understand what you mean 😂

I mean maybe we can merge this as is and figure out the log level stuff later? Or do we want this to be "perfect"?

src/redisearch_rs/tracing_redismodule/src/lib.rs
JonasKruckenberg: huh 🤔 good to know but yeah I'm really not sure this is exposed to the modules

JonasKruckenberg:
> Only so that we can set the log level correctly which the Writer does not have access to

You underestimate the prescience of the tracing authors 👀 pushed a change that now automatically sets the "correct" redismodule level based on the tracing level

src/redisearch_rs/tracing_redismodule/src/lib.rs
JonasKruckenberg: resolved now

src/redisearch_rs/tracing_redismodule/src/lib.rs
JonasKruckenberg: updated the comment 👍🏻

src/redisearch_rs/tracing_redismodule/src/lib.rs
JonasKruckenberg: updated the comment 👍🏻

src/redisearch_rs/Cargo.toml
```diff
@@ -110,6 +112,10 @@ thiserror = "2.0.12"
 unsafe-tools = "0.1.1"
 ureq = "3.0.10"
 wildcard_cloudflare = { package = "wildcard", version = "0.3.0" }
+tracing = "0.1.42"
+tracing-core = "0.1"
+tracing-subscriber = { version = "0.3.20", default-features = false, features = ["std", "fmt", "ansi", "env-filter"] }
+anstyle-query = "1.1.5"
```
chesedo: Just fix the sort at some point again

src/redisearch_rs/tracing_redismodule/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+edition.workspace = true
+license-file.workspace = true
+publish.workspace = true
+
+[dependencies]
+tracing-core.workspace = true
+tracing-subscriber.workspace = true
+ffi.workspace = true
+tracing.workspace = true
+anstyle-query.workspace = true
```
chesedo: Sorting here too

src/redisearch_rs/tracing_redismodule/src/lib.rs
chesedo: Since we now log with the correct levels redis filters the logs based on this config right? But we still spend the CPU cycles to generate the logs in the Rust code?

JonasKruckenberg: yeah it is essentially unnecessary overhead but:
1. I expect that redis probably wants this to be disabled for release builds anyway (an no harm in spending a few extra cycles in debug mode).
2. We can always later optimize this if it turns out that there is a way to access this config setting from redis

src/redisearch_rs/tracing_redismodule/src/lib.rs
```diff
@@ -0,0 +1,254 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! A subscriber for the Rust `tracing` ecosystem that emits traces and logs to the redismodule logging system.
```
kei-nan, line 10: I think an actual example of how to use this would really help.
Either here or somewhere else.
Feels like a lot of what the documentation talks about requires some rust knowledge in advance.

src/redisearch_rs/tracing_redismodule/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+        .with_thread_names(true)
+        .with_thread_ids(true)
+        .with_span_events(FmtSpan::FULL)
+        .with_env_filter(env_filter)
+        .with_ansi(should_print_colors())
+        .without_time() // redis already prints timestamps
+        .with_writer(MakeRedisModuleWriter {
+            ctx: ctx.and_then(|ctx| {
+                // Safety: We assume this static will not be written to after it has been initialized
+                let detach_ctx = unsafe { ffi::RedisModule_GetDetachedThreadSafeContext.unwrap() };
```
kei-nan, line 110: will this leak?
are we supposed to free it at some point?

src/redisearch_rs/tracing_redismodule/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+        thread_local! {
+            // per-CPU "cached" allocation for formatting log messages into
+            // reusing it means we dont allocate for every message which would be prohibitive
+            static BUF: RefCell<Vec<u8>> = const { RefCell::new(Vec::new()) };
+        }
+
+        BUF.with(|buf| {
+            // NB: variable declarations to extend the lifetime to the entire scope
+            let borrow = buf.try_borrow_mut();
+            let mut a;
```
kei-nan, line 211: can we maybe give a and b more meaningful names?
do we really need to declare them up front?

src/redisearch_rs/tracing_redismodule/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+struct RedisModuleWriter {
+    ctx: Option<NonNull<ffi::RedisModuleCtx>>,
+    level: &'static CStr,
+    log: LogFunc,
+}
+
+impl io::Write for RedisModuleWriter {
+    fn write(&mut self, input: &[u8]) -> io::Result<usize> {
```
kei-nan, line 196: is there a way to test logging?
maybe we you can add a flow test that will check the behaviour of this feature?

kei-nan (changes requested):
left some comments

src/redisearch_rs/tracing_redismodule/src/lib.rs
JonasKruckenberg: Once initialized, you actually cannot "de-initialize" a subscriber. It lives until the end of the program, so there is no way to forget to free.

src/redisearch_rs/tracing_redismodule/src/lib.rs
JonasKruckenberg: The whole idea here is that we have a thread-locally cached buffer (so we dont need to malloc on every log message) so far so easy. But, in signal handlers or other abnormal situations we could have reentrant logging (where we attempt to log while the same thread is already logging) and just panicking or not logging anything isn't an option (these weird cases are where you especially want logs after all) we need to gracefully handle this and just allocate a new buffer (a bit slower is fine for these rare, abnormal situations).

So in code, in one arm of the if we want to borrow the thread-local buffer mutably and in the other arm (if that borrowing wasnt allowed) we want to newly allocate. If we just naively borrow or allocate though:

```rust
if cond {
   buf.borrow_mut(); // returns a RAII type that "unborrows" the buffer when dropped
} else {
   Vec::new(); // vec will free its memory when dropped
}
```

the lifetime of either the borrow or vec end at the end of the if statement. Not what we want. We would usually fix this by returning the type by value from the if:

```rust
if cond {
   buf.borrow_mut() // returns a RAII type that "unborrows" the buffer when dropped
} else {
   Vec::new() // vec will free its memory when dropped
}
```
(notice the lacking semicolon)

buuut the problem here is `borrow_mut` and `Vec::new` return totally different types! so we actually cannot return these from the if statement.

Which is where `a` and `b` come in: We essentially declare two variables (for the two different types) that life for the entire scope (not just the if statement) and and conditionally initialize either or the other. We can then (to get one singular type that we can use going forward) borrow from eithet the mutable borrow or newly allocated vec and get a `&mut [u8]`.

JonasKruckenberg: TLDR: yes we do need to declare them upfront. As for naming, i guess we could name them more descriptively but since they are not used at all i didn't consider it necessary. But happy to change if you feel strongly.

src/redisearch_rs/tracing_redismodule/src/lib.rs
JonasKruckenberg: I added documentation in `developer.md` 👍🏻

tests/pytests/test_tracing.py
```diff
@@ -0,0 +1,84 @@
...
+# Assert that by default "init message" is hidden
+def test_default_level(env):
+    logDir = env.cmd("config", "get", "dir")[1]
+    logFileName = env.cmd("CONFIG", "GET", "logfile")[1]
+    logFilePath = os.path.join(logDir, logFileName)
+    matchCount = _grep_file_count(logFilePath, "Tracing Subscriber Initialized!")
+    env.assertEqual(
+        matchCount, 0, message="tracing subscriber message present for default level"
+    )
+    env.stop()
```
kei-nan: no need to stop it manually

tests/pytests/test_tracing.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+# Assert that we can set the `RUST_LOG` env var to enable the "init message"s level
+def test_trace_level():
+    prev_lvl = os.environ.get("RUST_LOG")
+
+    os.environ["RUST_LOG"] = "trace"
```
kei-nan: Suggestion for temporarly modifying environment variables
Add this struct:
```python
class EnvContextManager:
    def __init__(self, **kwargs):
        self.env_vars = kwargs
    
    def __enter__(self):
        self.old_values = {key: os.environ.get(key) for key in self.env_vars.keys()}
        os.environ.update(self.env_vars)
    
    def __exit__(self, exc_type, exc_value, exc_traceback):
        os.environ.update(self.old_values)
```

then you can do:
```python
with EnvContextManager(RUST_LOG='trace'):
    env = Env()
    ....
```
See here for more details:
https://ingosview.medium.com/playing-around-with-pythons-context-managers-to-change-env-variables-2678afeb1e6

tests/pytests/test_tracing.py
```diff
@@ -0,0 +1,84 @@
...
+    logDir = env.cmd("config", "get", "dir")[1]
+    logFileName = env.cmd("CONFIG", "GET", "logfile")[1]
+    logFilePath = os.path.join(logDir, logFileName)
+    matchCount = _grep_file_count(logFilePath, "Tracing Subscriber Initialized!")
+    env.assertEqual(
+        matchCount,
+        0,
+        message="tracing subscriber message even though source was disabled",
+    )
+    env.stop()
```
kei-nan: no need to stop the env manually

kei-nan (approved):

PR #7571, opened 2025-11-29
Register Cursor Sub-Commands as such - [MOD-12807, MOD-12808]

GuyAv46:
## Register cursor sub-commands using sub-commands API

The motivation is mostly for statistical information for each sub-command, and also for code simplicity

#### Issues this PR fixes:
1. MOD-12807
2. MOD-12808

#### Mark if applicable

---

> [!NOTE]
> Register FT.CURSOR subcommands (READ/PROFILE/DEL/GC) via the subcommands API, refactor cursor handlers to per-subcommand functions, adjust reply plumbing, and update tests/errors/commandstats accordingly.
> 
> - **Commands/Registration**:
>   - Register `FT.CURSOR` subcommands using Redis subcommands API: `READ`, `PROFILE`, `DEL`, `GC` (both external `FT.CURSOR` and internal `_FT.CURSOR`).
>   - Split former monolithic `RSCursorCommand` into `RSCursorReadCommand`, `RSCursorProfileCommand`, `RSCursorDelCommand`, `RSCursorGCCommand`.
>   - Add coordinator wrappers/macros (`CURSOR_SUBCOMMAND(...)`) to dispatch subcommands in cluster mode.
> - **Execution/Replies**:
>   - Change `sendChunk_ReplyOnly_EmptyResults` signature to use `RedisModuleCtx *` and create/end replies internally; update all callers.
>   - Refactor cursor read path to accept `RedisModuleCtx *` (build reply locally), improve error when index dropped while idle, and ensure replies are properly ended.
>   - Implement standalone handlers for `PROFILE`, `DEL`, and `GC` with explicit argument validation and return codes; `GC` aggregates local/coord results.
> - **Module wiring**:
>   - Update registration in both single-shard and coordinator init paths to create `FT.CURSOR` with subcommands and proper key specs/ACL flags.
> - **Tests**:
>   - Adjust tests to new subcommand errors/messages and commandstats keys (`cmdstat__FT.CURSOR|READ/DEL`).
>   - Add tests for `_FT.CURSOR PROFILE`/READ after index drop, argument validation, and `GC` edge cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 19020b258c2b4ed41b056f978d030a11293f5b9a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -3408,7 +3408,6 @@ int Cursor##name##Command(RedisModuleCtx *ctx, RedisModuleString **argv, int arg
 CURSOR_SUBCOMMAND(Read)
 CURSOR_SUBCOMMAND(Del)
 CURSOR_SUBCOMMAND(GC)
-CURSOR_SUBCOMMAND(Profile)
 
 int TagValsCommandHandler(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
   if (argc < 3) {
@@ -4030,18 +4029,13 @@ RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
   {
     int firstkey, lastkey, keystep;
     if (clusterConfig.type == ClusterType_RedisLabs) {
-      firstkey = 3;
-      lastkey = 1;
-      keystep = -3;
+      firstkey = 3; lastkey = 1; keystep = -3;
     } else {
-      firstkey = 0;
-      lastkey = 0;
-      keystep = -1;
+      firstkey = 0; lastkey = 0; keystep = -1;
     }
     RM_TRY(RMCreateSearchCommand(ctx, "FT.CURSOR", NULL, "readonly", firstkey, lastkey, keystep, "read", false))
     RedisModuleCommand *cursorCmd = RedisModule_GetCommand(ctx, "FT.CURSOR");
     RM_TRY(RedisModule_CreateSubcommand(cursorCmd, "READ", SafeCmd(CursorReadCommand), "readonly", firstkey, lastkey, keystep))
-    RM_TRY(RedisModule_CreateSubcommand(cursorCmd, "PROFILE", SafeCmd(CursorProfileCommand), "readonly", firstkey, lastkey, keystep))
     RM_TRY(RedisModule_CreateSubcommand(cursorCmd, "DEL", SafeCmd(CursorDelCommand), "readonly", firstkey, lastkey, keystep))
     RM_TRY(RedisModule_CreateSubcommand(cursorCmd, "GC", SafeCmd(CursorGCCommand), "readonly", firstkey, lastkey, keystep))
   }
--- a/tests/pytests/test_cursors.py
+++ b/tests/pytests/test_cursors.py
@@ -525,19 +525,21 @@ def testCursorDepletionStrictTimeoutPolicy():
         'FT.AGGREGATE', 'idx', '*', 'LOAD', '1', '@t', 'GROUPBY', '1', '@t', 'WITHCURSOR', 'COUNT', str(num_docs), 'TIMEOUT', '1'
     ).error().contains('Timeout limit was reached')
 
-def test_cursor_profile(env):
+@skip(cluster=True)
+def test_cursor_profile(env: Env):
     env.expect('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT').ok()
-    conn = getConnectionByEnv(env)
+    env.cmd('DEBUG', 'MARK-INTERNAL-CLIENT')
 
-    conn.execute_command('HSET', f'doc1', 't', str(1))
-    conn.execute_command('HSET', f'doc2', 't', str(2))
+    env.cmd('HSET', f'doc1', 't', str(1))
+    env.cmd('HSET', f'doc2', 't', str(2))
 
-    env.expect('FT.CURSOR', 'PROFILE', 'idx', '123').error().contains('Cursor not found')
+    env.expect('_FT.CURSOR', 'PROFILE', 'idx', '123').error().contains('Cursor not found')
 
     # create a cursor
-    res, cursor = env.cmd('FT.AGGREGATE', 'idx', '*', 'WITHCURSOR', 'COUNT', '1')
+
+    _, cursor = env.cmd('FT.AGGREGATE', 'idx', '*', 'WITHCURSOR', 'COUNT', '1')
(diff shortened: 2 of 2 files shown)


src/aggregate/aggregate_exec.c
```diff
@@ -1299,7 +1301,8 @@ static void cursorRead(RedisModule_Reply *reply, Cursor *cursor, size_t count, b
     if (!StrongRef_Get(execution_ref)) {
       // The index was dropped while the cursor was idle.
       // Notify the client that the query was aborted.
-      RedisModule_Reply_Error(reply, "The index was dropped while the cursor was idle");
+      Cursor_Free(cursor);
```
kei-nan, line 1304: Why switch the ordering? any particular reason?

src/aggregate/aggregate_exec.c
```diff
@@ -1353,115 +1356,129 @@ static void cursorRead_ctx(CursorReadCtx *cr_ctx) {
...
+
+  // Return profile
+  Cursor *cursor = Cursors_TakeForExecution(GetGlobalCursor(cid), cid);
+  if (cursor == NULL) {
+    return RedisModule_ReplyWithErrorFormat(ctx, "Cursor not found, id: %d", cid);
+  }
+
+  AREQ *req = cursor->execState;
+  if (!IsProfile(req)) {
+    Cursor_Pause(cursor); // Pause the cursor again since we are not going to use it, but it's still valid.
```
kei-nan, line 1424: This line is a bugfix?
didn't see a pause in previous code.
Do we have a test that covers this line?

src/aggregate/aggregate_exec.c
```diff
@@ -1353,115 +1356,129 @@ static void cursorRead_ctx(CursorReadCtx *cr_ctx) {
...
+
+/**
+ * FT.CURSOR GC {index}
+ */
+int RSCursorGCCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+  if (argc < 3) {
+    return RedisModule_WrongArity(ctx);
+  }
+
+  int rc = Cursors_CollectIdle(&g_CursorsList);
```
kei-nan: It looks like -1 can be returned from one call and 1 from the other.
That means the user can get 0.
Seems like a bug.
If you prefer to open a task instead of fixing it in this PR, then it also works.

src/module.c
```diff
@@ -1480,8 +1479,16 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx) {
          "readonly", 1, 1, 1, "read", false))
 
   // Do not force cross slot validation since coordinator will handle it.
-  RM_TRY(RMCreateSearchCommand(ctx, RS_CURSOR_CMD, RSCursorCommand, "readonly",
+  RM_TRY(RMCreateSearchCommand(ctx, RS_CURSOR_CMD, NULL, "readonly",
```
kei-nan, line 1482: This is really going to conflict with the set command info PR that itzik is working on.
@Itzikvaknin FYI

src/module.c
```diff
@@ -3385,15 +3388,28 @@ static int CursorCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
...
     return ReplyBlockDeny(ctx, argv[0]);
   }
 
-  return ConcurrentSearch_HandleRedisCommandEx(DIST_THREADPOOL, CursorCommandInternal, ctx, argv, argc,
+  return ConcurrentSearch_HandleRedisCommandEx(DIST_THREADPOOL, dist_callback, ctx, argv, argc,
                                                (WeakRef){0});
 }
 
+#define CURSOR_SUBCOMMAND(name) \
+static void Cursor##name##CommandInternal(RedisModuleCtx *ctx, RedisModuleString **argv, int argc, struct ConcurrentCmdCtx *cmdCtx) { \
```
kei-nan, line 3401: are we really benefiting from these macros?
couldn't we use an enum of some kind and have a single internal function which calls the RS function based on the enum value?

src/aggregate/aggregate_exec.c
GuyAv46: There is no order switch - this is a cursor leak.
Previously, it was taken for execution but then never released or put back, so we never freed it (and sanitizer doesn't report it as we have a valid pointer to it).

src/aggregate/aggregate_exec.c
GuyAv46: Yes, this was a cursor leak. It is covered by a test now.
I also opened a tech-debt to verify no cursor is logically leaked in sanitizer runs (check that there are no non-idle cursors after stopping all threads on shutdown)

src/aggregate/aggregate_exec.c
GuyAv46: Hmm, good point.
Let's open a bug for that. I'm not sure what the desired return value will be in this case.

src/module.c
GuyAv46: Yes, we are aligned on that. This PR is going to be BP, so we agreed to merge it first, and @Itzikvaknin will align his PR

src/module.c
GuyAv46: This is what this macro does.
We must have a function per sub-command, and this macro defines it and calls the generic (inlined) function with the right callbacks

kei-nan: I understand the logic and what they do but feels like we could have used the context or some user data and avoided using macros.
At least at for the internal function we could have had one internal function which receives the right enum value.

src/aggregate/aggregate_exec.c
kei-nan: feel free to resolve once you open the task.

src/module.c
GuyAv46: There are endless ways to implement this.
Why an enum is better than a function pointer? (Which gets inlined so there is no overhead) I’d prefer to keep this layout.
I can avoid the macro and define the functions manually if you prefer…

src/aggregate/aggregate_exec.c
GuyAv46: Fixed it here eventually. Will open a ticket for tracking

src/module.c
kei-nan: I guess it doesn't matter so much.
prefer to avoid macros when possible.

kei-nan (dismissed):

PR #7590, opened 2025-12-01
[MOD-12409]: Port DocumentType enum to Rust

hdoordt:
Port DocumentType enum to Rust as `doc_type::DocumentType`, removing it from `redisearch.h`. Use `doc_type_DocumentType` in `rlookup` instead of `rlookup::bindings::DocumentType`.

---

> [!NOTE]
> Ports `DocumentType` to a new Rust `document` crate, exposes it via `document_ffi` to C, updates `redisearch.h` to include the generated header, and switches `rlookup` to use the shared enum.
> 
> - **Core/FFI**:
>   - **New Rust crate `document`**: Defines `DocumentType` enum with string/repr conversions and tests.
>   - **New crate `c_entrypoint/document_ffi`**: Generates C header `headers/document_rs.h` via cbindgen, exporting `DocumentType`.
>   - **C Header changes**: `src/redisearch.h` now includes `document_rs.h` and removes the inline `DocumentType` enum; existing macros continue using `DocumentType_*`.
> - **rlookup**:
>   - Removes local `DocumentType` from `bindings.rs` and imports `document::DocumentType`.
>   - `load_document` and options updated to use shared `DocumentType`.
> - **Workspace**:
>   - Adds `document` and `document_ffi` to workspace members/dependencies; updates Cargo files accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a035fb6d7eb4d467107d3a383d47d93ebff636f2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/redisearch_rs/c_entrypoint/doc_type_ffi/cbindgen.toml
```diff
@@ -0,0 +1,15 @@
...
+no_includes = true
+
+[parse]
+parse_deps = true
+include = ["doc_type", "c_ffi_utils"]
+
+[export]
+include = ["DocumentType"]
+
+[export.rename]
```
chesedo: This section can be removed

src/redisearch_rs/doc_type/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "doc_type"
```
chesedo: Maybe this crate should be `document` instead? Or something a little broader which is likely to "own" the doc type definition.

src/redisearch_rs/doc_type/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    use crate::{ConvertDocumentTypeError, DocumentType};
+
+    #[test]
+    fn test_serialize_deserialize_from_ffi() {
+        #[rustfmt::skip]
+        let doc_types = [
+            (DocumentType::Hash,        "hash",         0),
+            (DocumentType::Json,        "json",         1),
+            (DocumentType::Unsupported, "unsupported",  3),
```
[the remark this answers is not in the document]
chesedo: This AI comment seems correct. But also raises a question: why panic on the integer conversion? Maybe all unrecognized enums should resolve to the `Unsupported` variant?

src/redisearch_rs/test.json
chesedo, line 1: Accidental commit?

src/redisearch_rs/doc_type/src/lib.rs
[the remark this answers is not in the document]
hdoordt: It's indeed not quite clear what the behavior should be, but taking a closer look it seems as if the `Unsupported` variant encodes the error case. I'll update the code to reflect that.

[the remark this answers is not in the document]
hdoordt: Done in 241c8e059

src/redisearch_rs/c_entrypoint/doc_type_ffi/cbindgen.toml
hdoordt: Fixed in 241c8e059

src/redisearch_rs/doc_type/Cargo.toml
hdoordt: Done in 241c8e059

src/redisearch_rs/test.json
hdoordt: Removed in 241c8e059

src/redisearch.h
```diff
@@ -14,6 +14,7 @@
 #include <limits.h>
 #include <stdbool.h>
 #include <time.h>
+#include "document_rs.h"
```
kei-nan, line 17: is having the "_rs" suffix considered to be the convention for rust header files?

Maybe if it was inside a folder, it would remove the need for such a suffix.

src/redisearch_rs/headers/document_rs.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+/* Warning, this file is autogenerated by cbindgen from `src/redisearch_rs/c_entrypoint/doc_type_ffi/build.rs. Don't modify it manually. */
+
+typedef enum DocumentType {
```
kei-nan, line 9: Maybe add a bit of documentation on what this type is used for?

src/redisearch.h
hdoordt: In this case the reason is that we would get a name collision with the C `document` module. Indeed, a path prefix would remove this need and would also make it more clear that the module is implemented in Rust. Still, I deem this out of scope for this issue.

Now, why not pick a different name entirely? The reason is that when we _do_ end up porting the `document` C module, its contents fit nicely along with the `DocumentType` enum, and the name of the ported module should still be `document`.

src/redisearch_rs/headers/document_rs.h
hdoordt: Done in a080865a9

chesedo (approved):

PR #7593, opened 2025-12-01
[2.10] [MOD-11987] [MOD-12816] Backport report GIL time in FT.PROFILE and initialize GILTime properly for RPSafeLoader

lerman25:
Continuing #6168, backport to 2.10

Piggybacking this PR for MOD-12806:
Currently, GILTime field in struct ResultProcessor is not initialized properly and is not updated. 
Current use for the field is only for struct RPSafeLoader in rpSafeLoaderNext_Accumulate, but it is not updated in this function. 
Resulting in profile output of:
GILTime 0 

This PR fixes that, plus changes the field name to avoid confusion.
More details in [ticket](https://redislabs.atlassian.net/browse/MOD-12816) 

---

> [!NOTE]
> Adds Total GIL time to FT.PROFILE (RESP2/3) and per-loader GIL-Time, with execution tracking to accumulate these timings; updates structs and tests accordingly.
> 
> - **Profile output**:
>   - Add `Total GIL time` to FT.PROFILE (RESP2/RESP3).
>   - For `RP_SAFE_LOADER`, include per-RP `GIL-Time` in profile.
> - **Execution tracking**:
>   - Track query start (`qiter.initTime`) and accumulate `qiter.queryGILTime` when running in background and during safe loader critical section.
>   - Set `qiter.isProfile` to gate timing.
> - **Struct/API updates**:
>   - `QueryIterator`: add `initTime`, `queryGILTime`, `isProfile`.
>   - `ResultProcessor`: add `rpGILTime`.
>   - Add `printProfileGILTime` helper.
> - **Tests/Utils**:
>   - Adjust profile array indexes for new GIL time entry; add `recursive_index`/`access_nested_list` helpers.
>   - New `testProfileGILTime` validates GIL times; numerous expectations updated to include the new fields.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 293200250b66396bd9e863b5691ac11c58fb5b8d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_profile.py
+++ b/tests/pytests/test_profile.py
@@ -818,11 +818,12 @@ def testProfileGILTime():
   env = Env(moduleArgs='WORKERS 1')
   conn = getConnectionByEnv(env)
 
+  env.expect('ft.create', 'idx', 'SCHEMA', 'f', 'TEXT').ok()
+
   # Populate db
   for i in range(10):
     res = conn.execute_command('hset', f'doc{i}', 'f', 'hello world',)
 
-  env.cmd('ft.create', 'idx', 'SCHEMA', 'f', 'TEXT')
   res = env.cmd('FT.PROFILE', 'idx', 'AGGREGATE', 'query', 'hello' ,'SORTBY', '1', '@f')
 
   # Record structure:


src/profile.h
```diff
@@ -9,7 +9,7 @@
 #include "aggregate/aggregate.h"
 #include "util/timeout.h"
 
-#define printProfileGILTime(vtime) RedisModule_ReplyKV_Double(reply, "GIL-Time", (rs_timer_ms(&(vtime))))
+#define printProfileGILTime(vtime) RedisModule_ReplyKV_Double(reply, "GIL-Time", (vtime))
```
Itzikvaknin, line 12: Are you sure its ok?

lerman25: Yes
The call to this function is `printProfileGILTime(rs_wall_clock_convert_ns_to_ms_d(rp->rpGILTime));`
where `rpGILTime` is `rs_wall_clock_ns_t`

Itzikvaknin (approved):

src/profile.c
```diff
@@ -108,84 +112,118 @@ void Profile_Print(RedisModule_Reply *reply, ProfilePrinterCtx *ctx) {
...
+    // Print whether a warning was raised throughout command execution
+    if (ctx->bgScanOOM) {
+      RedisModule_ReplyKV_SimpleString(reply, "Warning", QUERY_WINDEXING_FAILURE);
+    }
+    if (ctx->timedout) {
+      RedisModule_ReplyKV_SimpleString(reply, "Warning",
+                                       QueryError_Strerror(QUERY_ETIMEDOUT));
+    } else if (ctx->reachedMaxPrefixExpansions) {
+      RedisModule_ReplyKV_SimpleString(reply, "Warning", QUERY_WMAXPREFIXEXPANSIONS);
+    } else if (!ctx->bgScanOOM) {
```
alonre24: why adding this?

tests/pytests/test_profile.py
```diff
@@ -807,3 +810,42 @@ def verify_search_mode(query_type, query, params, expected_mode, expected_iterat
...
+  res = env.cmd('FT.PROFILE', 'idx', 'AGGREGATE', 'query', 'hello' ,'SORTBY', '1', '@f')
+
+  # Record structure:
+  # ['Type', 'Threadsafe-Loader', 'GIL-Time', ANY , 'Time', ANY, 'Counter', 100]
+  # ['Total GIL time', ANY]
+
+  try:
+    # env.assertTrue(recursive_contains(res, 'Threadsafe-Loader'), message=f"res: {res}")
+    # env.assertTrue(recursive_contains(res, 'Total GIL time'), message=f"res: {res}")
+
```
alonre24: leftover?

src/profile.c
lerman25: mistake, removing

tests/pytests/test_profile.py
lerman25: I guess left as is from the backport,
@Itzikvaknin is there a reason not to remove this?

tests/pytests/test_profile.py
```diff
@@ -807,3 +810,42 @@ def verify_search_mode(query_type, query, params, expected_mode, expected_iterat
...
+    rp_GIL_time = rp_record[rp_record.index('GIL-Time') + 1]
+
+    # extract the total GIL time
+    total_GIL_index = recursive_index(res, 'Total GIL time')
+    total_GIL_index[-1] += 1
+    total_GIL_time = access_nested_list(res, total_GIL_index)
+
+    env.assertGreaterEqual(float(total_GIL_time), 0)
+    env.assertGreaterEqual(float(rp_GIL_time), 0)
+    env.assertGreaterEqual(float(total_GIL_time), float(rp_GIL_time))
```
alonre24: @Itzikvaknin what is the purpose of this assert? `assertGreaterEqual` will always be true isn't it? Is there a way to test that we waited for the GIL in a non-flaky way?

lerman25: @alonre24 kind of a hack, but what if we always add epsilon to the counter each time we wait for load?

tests/pytests/test_profile.py
Itzikvaknin: We can enable it, there was an issue that if a timeout occurred we dont get the full response format, causing these assertions to fail 

I think Guy fixed it, need to make sure it was backported and if so we can enable it

lerman25: The same fields are used later in the test so I don't think it's needed to re-add this assertions

PR #7594, opened 2025-12-01
[MOD-12806] [MOD-12816] Change run in background check in FT.PROFILE and initialize GILTime properly for RPSafeLoader

lerman25:
Currently, in RediSearch >= 8.0, when we print Total GIL Time, we check if the query is background by using RunInThread(), but this function checks the current global config instead of the specific query status, instead we should check if QEXEC_F_RUN_IN_BACKGROUND is set. 
This PR fixes that, plus fixes the indentation in that function  

Piggybacking this PR for MOD-12806:
Currently, GILTime field in struct ResultProcessor is not initialized properly and is not updated.
Current use for the field is only for struct RPSafeLoader in rpSafeLoaderNext_Accumulate, but it is not updated in this function.
Resulting in profile output of:
GILTime 0

This PR fixes that, plus changes the field name to avoid confusion.
More details in [ticket](https://redislabs.atlassian.net/browse/MOD-12816)

---

> [!NOTE]
> Fixes FT.PROFILE GIL-time reporting by using the per-query background flag and properly accumulating/printing query and RP SafeLoader GIL time; updates C/Rust structs and profile output accordingly.
> 
> - **Profile output/logic**:
>   - Use `QEXEC_F_RUN_IN_BACKGROUND` (per-request) to determine background execution in `Profile_Print` instead of `RunInThread()`.
>   - Report "Total GIL time" from `qctx->queryGILTime` when background; otherwise from `qctx->initTime`.
>   - For `RP_SAFE_LOADER`, print GIL time via `rp->rpGILTime` (converted to ms); adjust `printProfileGILTime` macro to accept a double.
> - **GIL-time accounting**:
>   - Rename/shift per-query accumulator to `QueryProcessingCtx::queryGILTime`; update uses and initialization in `aggregate_exec.c`.
>   - In `rpSafeLoaderNext_Accumulate`, accumulate elapsed GIL time into both `parent->queryGILTime` and per-RP `rp->rpGILTime`.
> - **Result processor structs (C/Rust FFI)**:
>   - Replace `ResultProcessor::GILTime` timespec with `rpGILTime` (`rs_wall_clock_ns_t`) and update field layout assertions.
>   - Update Rust `Header` to use `rp_gil_time: rs_wall_clock_ns_t` and keep FFI offsets in sync.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 200e5d2a92f77c4597038ba7864bd6c1d4bbfa5b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/result_processor/src/lib.rs
+++ b/src/redisearch_rs/result_processor/src/lib.rs
@@ -21,7 +21,7 @@ pub mod counter;
 #[cfg(test)]
 mod test_utils;
 
-use libc::{c_int};
+use libc::c_int;
 use pin_project::pin_project;
 #[cfg(debug_assertions)]
 use std::any::{TypeId, type_name};


src/result_processor.h
```diff
@@ -159,7 +159,8 @@ typedef struct ResultProcessor {
   // Type of result processor
   ResultProcessorType type;
 
-  struct timespec GILTime;
+  rs_wall_clock_ns_t rpGILTime; // Accumulated GIL time of the ResultProcessor, if applicable (e.g. RP_SAFE_LOADER)
```
GuyAv46, line 162: So we decided not to move it to the loader only?

src/profile.c
```diff
@@ -108,7 +108,7 @@ static double _recursiveProfilePrint(RedisModule_Reply *reply, ResultProcessor *
 
       case RP_SAFE_LOADER:
         printProfileType(RPTypeToString(rp->type));
-        printProfileGILTime(rp->GILTime);
+        printProfileGILTime(rs_wall_clock_convert_ns_to_ms_d(rp->rpGILTime));
```
GuyAv46, line 111: Regarding [previous comment](https://github.com/RediSearch/RediSearch/pull/7594/files#r2581403206)
This is the only place we access `rpGILTime`, and we know the `rp` type here.

src/result_processor.h
lerman25: Since `RPSafeLoader` is not exposed (declared in `result_processor.c`, and I can see why it's this way, I think we should be keeping the field in `ResultProcessor`. Maybe with the move to rust we will be able to make this change.

PR #7556, opened 2025-11-27
[MOD-12069] Add `*_pending_jobs` metrics

meiravgri:
[MOD-12069] Multi threaded section
[MOD-12696] number of queries waiting in the shards queue
[MOD-12697] number of  pending jobs in “low priority queue”

Extends the `multi_threading` section in `INFO MODULES` by adding `workers_low_priority_pending_jobs` cand `workers_high_priority_pending_jobs` metrics to track pending jobs in the worker thread pool queues.

* `workers_low_priority_pending_jobs` currently refers to background indexing jobs
* `workers_high_priority_pending_jobs` refers to queries

Continues #7552, #7538, and #7546 which added `search_multi_threading` section with `search_active_io_threads`, `search_active_worker_threads`, and `search_active_coord_threads` fields.

### INFO MODULES Output

```
# search_multi_threading
...
workers_low_priority_pending_jobs:0
workers_high_priority_pending_jobs:0
```

### Additional Changes and Improvements

- **Thread Pool APIs**: Add `redisearch_thpool_high_priority_pending_jobs()` and `redisearch_thpool_low_priority_pending_jobs()` to expose queue lengths from `thpool` (`deps/thpool/thpool.{h,c}`).
- **Workers Utility**: Add `workersThreadPool_LowPriorityPendingJobsCount()` and `workersThreadPool_HighPriorityPendingJobsCount()` wrapper APIs (`src/util/workers.{h,c}`).
- **Global Stats**: Populate `low_priority_pending_jobs` and `high_priority_pending_jobs` in `GlobalStats_GetMultiThreadingStats()` using new worker APIs (`src/info/global_stats.c`).
- **Test Infrastructure**: Add `wait_for_condition()` helper for polling with timeout and state reporting (`tests/pytests/common.py`).

---

> [!NOTE]
> Expose `workers_low_priority_pending_jobs` and `workers_high_priority_pending_jobs` in `INFO MODULES` via new thpool/worker APIs, plus tests and a wait helper.
> 
> - **Info/Stats**:
>   - Extend `MultiThreadingStats` with `workers_low_priority_pending_jobs` and `workers_high_priority_pending_jobs` (`src/info/global_stats.h`).
>   - Populate new fields in `GlobalStats_GetMultiThreadingStats()` using worker APIs (`src/info/global_stats.c`).
>   - Export metrics in `INFO MODULES` under `multi_threading` section (`src/info/info_redis/info_redis.c`).
> - **Threadpool (`deps/thpool/`)**:
[...]

PR #7609, opened 2025-12-02
[MOD-12831] `test_info_modules:test_pending_jobs_metrics_search` Extend output on timeout

meiravgri:
A flaky failure of this test was encounterd in cluster env on macos intel.
The test was introduced in #7556

This PR extends the output log in case of timeout.
Also, the previous mechnism to check if the query errored was wrong.
I moved the queries distribtion to a general callback in `common.py` that also detects better failures.

---

> [!NOTE]
> Adds a threading helper to run repeated commands with exception monitoring and updates pending-jobs metrics tests to use it and include worker stats for better timeout diagnostics.
> 
> - **Tests (pending jobs metrics)**:
>   - Replace ad-hoc threading with `launch_cmds_in_bg_with_exception_check` to run `FT.SEARCH`/`FT.AGGREGATE` in background and detect early failures.
>   - Enrich `wait_for_condition` state with `WORKERS stats` in checks for indexing pending jobs, query pending jobs, and metrics reset.
>   - On helper-detected failure, `WORKERS RESUME` and exit early to avoid deadlock.
> - **Common utilities**:
>   - Add `launch_cmds_in_bg_with_exception_check(env, command, num_triggers, exception_timeout)` to spawn threads, monitor exceptions, and assert on failures.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8f8fb49397310814b15aeb70cbc5804068229f87. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/common.py
+++ b/tests/pytests/common.py
@@ -1104,25 +1104,37 @@ def call_and_store(fn, args, out_list):
     """
     out_list.append(fn(*args))
 
-def run_cmds_in_bg(env, command, num_commands):
+def launch_cmds_in_bg_with_exception_check(env, command, num_triggers, exception_timeout=1):
+    """
+    Launch the same Redis command multiple times in background threads with exception monitoring.
+    
+    Args:
+        env: Redis test environment for executing commands.
+        command: A list containing the Redis command to execute (e.g., ['FT.SEARCH', 'idx', 'query']).
+        num_triggers: Number of background threads to spawn, each executing the same command.
+        exception_timeout: Seconds to wait for exception detection (default: 1).
+    
+    Returns:
+        list[Thread]: Started thread objects if no exceptions occur, None if any thread fails.
+    """
     threads = []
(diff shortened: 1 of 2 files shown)


tests/pytests/common.py
```diff
@@ -1104,6 +1104,31 @@ def call_and_store(fn, args, out_list):
...
+            exceptions.append(e)
+            exception_event.set()
+
+    for i in range(num_commands):
+        t = threading.Thread(target=run_query)
+        threads.append(t)
+        t.start()
+
+    # Check for exceptions before proceeding
+    if exception_event.wait(timeout=1):
```
JoanFM: I think this can be error prone. What if exceptions simply work.

I would simply return the threads.

JoanFM (approved):

PR #7596, opened 2025-12-01
Pending jobs in coordinator queue

meiravgri:
[MOD-12069] Multi threaded section
[MOD-12788] Number of pending jobs in coordinator queue

Extends the `multi_threading` section in `INFO MODULES` by adding `coord_high_priority_pending_jobs` metric to track pending jobs in the coordinator thread pool queue.

* `coord_high_priority_pending_jobs` tracks jobs waiting in the coordinator queue (cluster mode only)
* Applies to `FT.SEARCH`, `FT.AGGREGATE`, `FT.CURSOR *`, and `FT.HYBRID` commands

### INFO MODULES Output

```
# search_multi_threading
...
coord_high_priority_pending_jobs:0
```

### Additional Changes and Improvements

- **Coordinator Thread Pool APIs**: Add `ConcurrentSearchPool_HighPriorityPendingJobsCount()` to expose coordinator queue length from `thpool` (`src/concurrent_ctx.{h,c}`).

- **Test Infrastructure**: Add comprehensive tests for `coord_high_priority_pending_jobs` metric using pause/resume capabilities to ensure deterministic and stable verification across `FT.SEARCH`, `FT.AGGREGATE`, `FT.CURSOR READ`, and `FT.HYBRID` commands (`tests/pytests/test_info_modules.py`).

### New Debug Commands for Corrdinator Threads

- **`FT.DEBUG COORD_THREADS PAUSE`**: Pauses the coordinator thread pool.
- **`FT.DEBUG COORD_THREADS RESUME`**: Resumes the coordinator thread pool.
- **Debug Infrastructure**: Add `ConcurrentSearch_isPaused()`, `ConcurrentSearch_pause()`, and `ConcurrentSearch_resume()` for coordinator thread pool control (`src/concurrent_ctx.{h,c}`).

---

> [!NOTE]
> Adds coord_high_priority_pending_jobs metric to INFO MODULES and introduces FT.DEBUG COORD_THREADS pause/resume for the coordinator thread pool, wiring through new thpool stats and tests.
> 
> - **Stats/Info**:
>   - Add `coord_high_priority_pending_jobs` to `multi_threading` (`GlobalStats_GetMultiThreadingStats`, `INFO MODULES`).
>   - Expose coordinator queue size via `ConcurrentSearchPool_HighPriorityPendingJobsCount()`.
> - **Coordinator Thread Pool**:
[...]

PR #7641, opened 2025-12-03
Fix: Add teardown to resume coordinator threads in test cleanup

meiravgri:
The `TestCoordHighPriorityPendingJobs` test class introduced in #7596 had no cleanup mechanism. If a test raised an error before calling `COORD_THREADS RESUME`, the coordinator thread pool remained paused, causing subsequent tests to hang or fail.

**Solution:** 
- Added `tearDown()` method to resume coordinator threads if paused
- Added `FT.DEBUG COORD_THREADS is_paused` command to check pause state

### DIFF
--- a/tests/pytests/test_info_modules.py
+++ b/tests/pytests/test_info_modules.py
@@ -1909,6 +1909,9 @@ class TestCoordHighPriorityPendingJobs(object):
     info_dict = info_modules_to_dict(self.env)
     self.env.assertEqual(info_dict[MULTI_THREADING_SECTION][COORD_HIGH_PRIORITY_PENDING_JOBS_METRIC], '0')
 
+  def tearDown(self):
+    if self.env.cmd(debug_cmd(), 'COORD_THREADS', 'is_paused'):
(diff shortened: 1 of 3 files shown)


tests/pytests/test_info_modules.py
```diff
@@ -1909,6 +1909,9 @@ def __init__(self):
     info_dict = info_modules_to_dict(self.env)
     self.env.assertEqual(info_dict[MULTI_THREADING_SECTION][COORD_HIGH_PRIORITY_PENDING_JOBS_METRIC], '0')
 
+  def tearDown(self):
+    if self.env.cmd(debug_cmd(), 'COORD_THREADS', 'is_paused'):
+      self.env.expect(debug_cmd(), 'COORD_THREADS', 'RESUME').ok()
```
JoanFM, line 1914: why don't we just send RESUME and ignore any errors?

JoanFM (approved):
I would remove `is_paused` and simply send RESUME and ignore result

PR #7494, opened 2025-11-24
document all iterators modules

gdesmott:
Some modules were missing top level documentation.

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

---

> [!NOTE]
> Add module-level docs across iterator modules and re-export key iterator types from `lib.rs`.
> 
> - **Docs**:
>   - Add module-level documentation to `empty.rs`, `id_list.rs`, `inverted_index.rs`, `maybe_empty.rs`, `metric.rs`, and `wildcard.rs`.
>   - Add brief trait description in `lib.rs` for `RQEIterator`.
> - **API (re-exports)**:
>   - Re-export iterator types in `lib.rs`: `Empty`, `IdList`, `NumericFull`, `TermFull`, `MetricIterator`, `Wildcard`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eb9269c409dfce6b3d4db945111b619e090c44d5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #7046, opened 2025-10-13
[MOD-10081] add Rust version of Non-Optimized Optional Iterator

GlenDC:
## Describe the changes in the pull request

add Rust version of Non-Optimized Optional Iterator

based on original C iterator code
as found on 'src/iterators/optional_iterator.c'
but adapting it and implementing it in the
rust-native style as established by prior art in rqe_iterators

also ported over the original C++ tests
to verify we preserve the OG business logic
... to achieve this we also introduce a util MockIterator,
which itself is also ported from those original C++ tests

optional iterator benchmarker results:

```
Iterator - Optional - Read Dense/C
                        time:   [814.79 µs 821.48 µs 827.78 µs]
Found 9 outliers among 100 measurements (9.00%)
  9 (9.00%) low mild
Iterator - Optional - Read Dense/Rust
                        time:   [1.1685 ms 1.1705 ms 1.1726 ms]
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

Iterator - Optional - Read Sparse/C
                        time:   [745.14 µs 746.46 µs 747.70 µs]
Found 14 outliers among 100 measurements (14.00%)
  6 (6.00%) low severe
  3 (3.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe
Iterator - Optional - Read Sparse/Rust
                        time:   [530.10 µs 531.49 µs 532.76 µs]
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) low severe
  3 (3.00%) low mild

Iterator - Optional - SkipTo Dense/C
                        time:   [18.818 µs 18.885 µs 18.949 µs]
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) low mild
  3 (3.00%) high mild
Iterator - Optional - SkipTo Dense/Rust
                        time:   [12.348 µs 12.382 µs 12.422 µs]
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  1 (1.00%) high mild
  3 (3.00%) high severe

Iterator - Optional - SkipTo Sparse/C
                        time:   [18.843 µs 18.911 µs 18.974 µs]
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) low mild
  1 (1.00%) high mild
Iterator - Optional - SkipTo Sparse/Rust
                        time:   [5.8449 µs 5.8686 µs 5.8931 µs]

[...]

PR #7663, opened 2025-12-07
[MOD-10079] Port Intersection iterator (no slop, no order)

dor-forer:

Current: The RediSearch query engine iterators are implemented in C++. The Intersection iterator (AND logic) needs to be ported to Rust as part of the iterator migration effort.
Change: Port the Intersection iterator to Rust with:
Core  Intersection struct implementing merge-based AND algorithm
 RQEIterator trait extensions: is_empty() and is_wildcard() methods
Box<dyn RQEIterator> implementation for dynamic dispatch
Comprehensive test suite (25 tests) ported from C++ tests
Outcome: Rust-native Intersection iterator ready for integration, with full test coverage.
Note: max_slop and in_order features are not included in this first version (tracked in MOD-12717).

Tests NOT implemented (intentionally):
Slop, InOrder, SlopAndOrder - These C++ tests are for max_slop and in_order features which are explicitly noted as "not yet implemented in the Rust port" in the file header.
NullChildren - Not applicable in Rust (Rust doesn't have null pointers in the same way).

1. MOD-10079

 1. src/redisearch_rs/rqe_iterators/src/intersection.rs - New Intersection iterator implementation
 2. src/redisearch_rs/rqe_iterators/src/lib.rs - Trait extensions and exports
 3. src/redisearch_rs/rqe_iterators/src/empty.rs - Added is_empty() impl
 4. src/redisearch_rs/rqe_iterators/src/wildcard.rs - Added is_wildcard() impl
 5. src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs - Test suite

#### Mark if applicable

---

> [!NOTE]
> Ports the Intersection (AND) iterator to Rust, adds Box<dyn RQEIterator> support, extensive integration tests, and intersection benchmarks with necessary FFI/build updates.
> 
> - **Iterator Core**:
>   - Add `intersection::Intersection` implementing merge-based AND across children; sorts children by `num_estimated` and builds aggregate results (uses targeted `unsafe` for result aggregation).
>   - Export `Intersection` in `rqe_iterators::lib`.
>   - Implement `RQEIterator` for `Box<dyn RQEIterator>` to enable dynamic dispatch.
> - **Tests**:
>   - New integration suite `tests/integration/intersection.rs` covering read/skip_to/rewind/edge cases and revalidate flows; updates `tests/integration/main.rs` to include module.
> - **Benchmarks & FFI**:
>   - Add `benchers::intersection` and hook into Criterion benches; enable C vs Rust comparisons.
>   - Extend bencher FFI to construct C intersection iterators; expose required symbols and constants.
>   - Update bencher `build.rs` to generate bindings for `intersection_iterator.h` and `headers/iterators_rs.h`; adjust `Cargo.toml` deps.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 70ae99163b29d1cb68c3d68d6eba218059357879. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rqe_iterators/src/intersection.rs
+++ b/src/redisearch_rs/rqe_iterators/src/intersection.rs
@@ -56,6 +56,7 @@ where
     #[must_use]
     pub fn new(mut children: Vec<I>) -> Self {
         children.sort_by_cached_key(|c| c.num_estimated());
+
         let Some(num_expected) = children.first().map(|c| c.num_estimated()) else {
             return Self {
                 children,
@@ -65,13 +66,13 @@ where
                 result: RSIndexResult::intersect(0),
             };
         };
-
+        let num_children = children.len();
         Self {
             children,
             last_doc_id: 0,
             num_expected,
             is_eof: false,
-            result: RSIndexResult::intersect(children.len()),
+            result: RSIndexResult::intersect(num_children),
         }
     }
 


src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,427 @@
...
+    let mut non_wildcards = Vec::with_capacity(children.len());
+    let mut last_wildcard: Option<I> = None;
+
+    for child in children {
+        if child.is_wildcard() {
+            last_wildcard = Some(child);
+        } else {
+            non_wildcards.push(child);
+        }
+    }
```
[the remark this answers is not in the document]
dor-forer: The current behavior matches the original C implementation
In real usage, all wildcards in an intersection would have the same top_id
Returning the last vs. finding the minimum is an optimization trade-off (avoiding an extra comparison loop)

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,427 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! Intersection iterator - finds documents appearing in ALL child iterators.
```
gdesmott: We unified those module doc: https://github.com/RediSearch/RediSearch/pull/7494/files

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,427 @@
...
+                children,
+                last_doc_id: 0,
+                num_expected: 0,
+                is_eof: true,
+                result: RSIndexResult::intersect(0),
+            };
+        }
+
+        children.sort_by_cached_key(|c| c.num_estimated());
+        let num_expected = children.first().map(|c| c.num_estimated()).unwrap_or(0);
```
gdesmott:
```suggestion
        let num_expected = children.first().map(|c| c.num_estimated()).expect("should have at least one children);
```

Should be safe with the check above, right?

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,427 @@
...
+    children: Vec<I>,
+    /// Last doc_id successfully found in ALL children (returned by `last_doc_id()`).
+    last_doc_id: t_docId,
+    num_expected: usize,
+    is_eof: bool,
+    /// Aggregate result combining children's results, reused to avoid allocations.
+    result: RSIndexResult<'index>,
+}
+
+enum AgreeResult {
```
gdesmott, line 41: Let's document this one.

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,427 @@
...
+    ///
+    /// - Returns `None` if the iterator is at EOF.
+    /// - Returns `Some(&mut result)` otherwise.
+    ///
+    /// # Note
+    ///
+    /// Before the first call to [`read()`](RQEIterator::read) or
+    /// [`skip_to()`](RQEIterator::skip_to), this returns a reference to the
+    /// uninitialized result buffer (with `doc_id = 0`). Callers should only
+    /// rely on `current()` after a successful read operation.
```
gdesmott: This is actually a potential issue for all iterators.

I'm wondering if we should handling this better, using a debug assertion at least.

@GlenDC : what do you think?

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,427 @@
...
+        match self.skip_to(max_child_doc_id)? {
+            Some(_) => Ok(RQEValidateStatus::Moved {
+                current: Some(&mut self.result),
+            }),
+            None => Ok(RQEValidateStatus::Moved { current: None }),
+        }
+    }
+}
+
+/// Result of reducing an intersection's children.
```
gdesmott: We need to come up with a good design for the various reducer code: `MOD-12355`

I'd leave it out for now.

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -110,4 +112,69 @@ pub trait RQEIterator<'index> {
...
+    /// Wildcard iterators match all documents and can be removed from intersections
+    /// without changing the result (since intersection with "all" is identity).
+    /// Default implementation returns `false`.
+    #[inline(always)]
+    fn is_wildcard(&self) -> bool {
+        false
+    }
+}
+
+// Implement RQEIterator for Box<dyn RQEIterator> to support dynamic dispatch
```
gdesmott, line 123: @GlenDC : isn't there a way to have this automatically?

src/redisearch_rs/rqe_iterators/src/intersection.rs
dor-forer: I agree, but now considering your other comment about leaving out the reducers, it won't reduce it to an empty iterator so we should allow 0 children, no?

gdesmott: The 0 children iterator should be `eof` directly and so this code is not meant to be called at all, if I understand it correctly.

src/redisearch_rs/rqe_iterators/tests/integration/utils/mock_iterator.rs
```diff
@@ -0,0 +1,323 @@
+/*
```
gdesmott, line 1: @GlenDC may be we should merge this already as it's being used here and will be in the not iterator as well I think.

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,427 @@
...
+        for child in &mut self.children {
+            if let Some(child_result) = child.current() {
+                let child_ptr: *const RSIndexResult<'index> = child_result;
+                // SAFETY: result_ptr points to self.result which is valid for this method's
+                // duration. `children` and `result` are disjoint fields.
+                let result_ref = unsafe { &mut *result_ptr };
+                // SAFETY: child_ptr points to child's result containing data with 'index
+                // lifetime. Children are owned by self, so their results remain valid.
+                let child_ref = unsafe { &*child_ptr };
+                result_ref.push_borrowed(child_ref);
```
gdesmott:
```suggestion
                self.result.push_borrowed(child_ref);
```

Seems to work. Or am I missing something?

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,427 @@
...
+                AgreeResult::Agreed => return Ok(Some(curr_target)),
+                AgreeResult::Ahead(new_target) => curr_target = new_target,
+                AgreeResult::Eof => return Ok(None),
+            }
+        }
+    }
+
+    /// Builds the aggregate result from all children's current results.
+    ///
+    /// # Why `unsafe` is used here
```
gdesmott, line 127: @GlenDC  @chesedo @LukeMathWalker : Here is a fun one.

I'd be interested to brainstorm ideas about how to properly fix this.

src/redisearch_rs/rqe_iterators/src/intersection.rs
dor-forer: Yes, you're right. I must've been carried with the unsafe wave.

src/redisearch_rs/rqe_iterators/src/intersection.rs
chesedo: Hmm, does the original C do a borrowed or owned push?

src/redisearch_rs/rqe_iterators/src/intersection.rs
dor-forer: You're right, when answering I probably missed the part of the if num_children == 0

src/redisearch_rs/rqe_iterators/src/intersection.rs
gdesmott: It calls [`AggregateResult_AddChild`](https://github.com/RediSearch/RediSearch/blob/master/src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs#L492) which is internally calling either `push_boxed()` or `push_borrowed()`.

chesedo: This iterator makes it own result. Therefore it is owned... and you can get rid of the lifetime :partying_face: 

https://github.com/RediSearch/RediSearch/pull/7663/files/aec8cfc7ffd5004b95ca560ab15db4868f8af1e2#diff-1037a974fa7d9e2e51d992c6a853550746510836d4166e0f75a339511f9fe6a2R62

You can therefore also use the owned push

chesedo: ... well maybe :sweat_smile:

src/redisearch_rs/rqe_iterators/src/empty.rs
```diff
@@ -61,4 +61,9 @@ impl<'index> RQEIterator<'index> for Empty {
     fn revalidate(&mut self) -> Result<RQEValidateStatus<'_, 'index>, RQEIteratorError> {
         Ok(RQEValidateStatus::Ok)
     }
+
+    #[inline(always)]
+    fn is_empty(&self) -> bool {
```
GlenDC: @gdesmott this feels like the wrong tool for the job. We are adding a runtime check for something already known at compile time. Why not use a marker trait instead, for example `trait RQEIteratorEmpty: RQEIterator {}`?

I also do not think this method adds a benefit even in context of wrapper types, since wrappers explicitly no longer are an empty result set even when wrapping an Empty iterator.

src/redisearch_rs/rqe_iterators/tests/integration/utils/mock_iterator.rs
GlenDC: I would prefer to just merge the optional iterator once the last review is resolved as to not keep making PR after PR. But up to you, if that's what you want I'll do it 🤷

src/redisearch_rs/rqe_iterators/src/lib.rs
GlenDC: Nope. You can write a macro that you can then apply to all the ones where you just need to dereference. You also have the same issue with having a trait for T also be implemented for `&T`, or `Arc<T>`, you need those impl blocks. There are good reasons for that and why it is not automatically done.

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -110,4 +112,69 @@ pub trait RQEIterator<'index> {
...
+    /// Returns `true` if this iterator is known to yield no results.
+    ///
+    /// Used by compound iterators (like Intersection) to optimize away empty children.
+    /// Default implementation returns `false`.
+    #[inline(always)]
+    fn is_empty(&self) -> bool {
+        false
+    }
+
+    /// Returns `true` if this iterator is a wildcard iterator.
```
GlenDC: Same as for `is_empty` I think we are reaching for the wrong tool for the job. Is anything besides the Wildcard iterator even a wildcard iterator. Take for example the `Not` or `Optional` wrappers, are those sometimes wildcard and sometimes not? 

I imagine not as this is all about a static property rather than the current state which can for those iterators flip all the time. If so then I think it means we do not even need to make a public trait for it. The intersection trait could just make private marker traits and change behaviour depending on that... feels to me like a local problem for which we make it here globally for no good reason?

src/redisearch_rs/rqe_iterators/src/intersection.rs
dor-forer: Removed

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,426 @@
...
+    }
+}
+
+/// Delegates a method call to the inner variant of `ReducedIntersection`.
+///
+/// This macro implements static dispatch for the enum, forwarding each trait method
+/// to the underlying variant (`Empty`, `Single`, or `Intersection`). This avoids
+/// the overhead of dynamic dispatch (vtable) while keeping the code DRY.
+///
+/// Example: `delegate!(self, skip_to, doc_id)` expands to:
```
GlenDC: This is a private macro and not sure the example really  provides any additional info, it basically is the same as can be seen the lines below given the macro is not even a complex one. So the only ones seeing this are the authors reading the code file anyway.

GlenDC: FWIW I would prefer even a macro-less version here, the name is ambigious and you anyway just write the code once. But that's not a strong opinion and more a matter of taste so you can neglect my last piece of feedback if you wish.

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,426 @@
...
+/// - Multiple children → Intersection
+pub enum ReducedIntersection<'index, I> {
+    Empty(Empty),
+    Single(I),
+    Intersection(Intersection<'index, I>),
+}
+
+/// Reduces children by removing wildcards and optimizing edge cases.
+#[must_use]
+pub fn reduce<'index, I>(children: Vec<I>) -> ReducedIntersection<'index, I>
```
GlenDC: In order to give feedback on this function can you please provide me detailed context on the call site of this function. What produces this container of iterators?

I feel we are porting over an API from a language and add all kind of things to facilitate that API despite the fact that AFAIK the callees are not based on some dynamic input but actually known at compile time, meaning I think it might be saner to just have the callee decide what kind of reduce interestector to make?

src/redisearch_rs/rqe_iterators/src/intersection.rs
GlenDC: The issue as usually is that you are working with &mut instead of just owned values though... 

As long as we do that it's always a matter of work arounds.

Due to how the life times are structured for this trait I don't think you will find a non-unsafe approach here, and honestly not convinced that any safety comment here makes it really actually safe.

For a much simpler case I was able to work around it in optional iterator by using the `result` method to get last result, but that was for just 1 child, here we are with a series.... It's a mess

src/redisearch_rs/rqe_iterators/src/intersection.rs
GlenDC: Can be workaround by adding to `self.is_eof` also the check if `doc_id == 0`,  no? Or perhaps a debug assert But yes such cases are indeed an issue in all iterators and so you would basically have to put that in all of them....

But once again this method is anyway a hack for the fact that we are so damn stuborn on returning mut references instead of owned values though. Because IMHO this method shouldn't even have to exist

src/redisearch_rs/rqe_iterators/src/intersection.rs
dor-forer: I removed the entire reducer, but I will keep it in mind

src/redisearch_rs/rqe_iterators/src/intersection.rs
dor-forer: I removed the entire reducer

src/redisearch_rs/rqe_iterators/src/empty.rs
dor-forer: Removed it - was part of the reducer

src/redisearch_rs/rqe_iterators/src/lib.rs
dor-forer: Removed it, it was part of the reducer.
I made a few design errors, so we should rethink about all of it again.

src/redisearch_rs/rqe_iterators/tests/integration/utils/mock_iterator.rs
dor-forer: I will wait for the optional iterator to be merged and then rebase again to get your changes.
Just copied it for now to test my code.

src/redisearch_rs/rqe_iterators/src/empty.rs
gdesmott: > this feels like the wrong tool for the job.

I agree. Let's remove that for now until we have a proper design for reducers.

src/redisearch_rs/rqe_iterators/tests/integration/utils/mock_iterator.rs
gdesmott: > I would prefer to just merge the optional iterator once the last review is resolved

Me too.

If it's finally merged this week then all good, if not let's at least get the mock iterator in so other branches are not blocked forever.

dor-forer: @gdesmott So, should be my next step? Should I rebase from the optional iterator? Or is it better for me to wait?

src/redisearch_rs/rqe_iterators/src/lib.rs
dor-forer: @gdesmott @GlenDC So just leave as is?

src/redisearch_rs/rqe_iterators/src/intersection.rs
dor-forer: Do I have anything else to do here?

src/redisearch_rs/rqe_iterators/src/intersection.rs
dor-forer: Should I look for alternatives?

src/redisearch_rs/rqe_iterators/tests/integration/utils/mock_iterator.rs
gdesmott: If @GlenDC is confident that https://github.com/RediSearch/RediSearch/pull/7046 will be merged this week then yes, I'd say rebase your branch on top of it.

src/redisearch_rs/rqe_iterators/src/lib.rs
gdesmott: Yeah.

gdesmott:
An easy next step is to add micro benchmarks comparing the performance of the C and Rust versions.
You can look at https://github.com/RediSearch/RediSearch/pull/7046/files demonstrating how those work.

You can run them by calling `cargo bench` in `src/redisearch_rs/rqe_iterators_bencher`.

We can also have a call together so I show you how all this work.

src/redisearch_rs/rqe_iterators/src/intersection.rs
LukeMathWalker: The mutability part is not really the issue here. 
The real challenge is that we're trying to build a self-referential object: `self` contains an aggregate result that borrows from another field within `self` itself (the children).
Within the current design constraints, the only viable solution would be to move the "cached" `RSIndexResult` instances out of the corresponding iterator, into what would most likely look like an arena.
It's viable, and I'm happy to sketch what it'd look like, but I wouldn't bundle it with this PR.

gdesmott: >  It's viable, and I'm happy to sketch what it'd look like, but I wouldn't bundle it with this PR.

I agree. Let's keep this `unsafe` for now then and experiment with alternatives on the side.

src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs
```diff
@@ -0,0 +1,1115 @@
...
+        210, 220, 230, 240, 250,
+    ],
+];
+
+// =============================================================================
+// Read tests - equivalent to TEST_P(IntersectionIteratorCommonTest, Read)
+// =============================================================================
+
+#[test]
+#[cfg_attr(miri, ignore)]
```
LukeMathWalker, line 83: Why are we ignoring miri here?

src/redisearch_rs/rqe_iterators/src/intersection.rs
gdesmott: I'd document this directly in `RQEIterator::current()`.

Something like: `Calling this method before the first call to read or skip_to, or directly after rewinding the iterator will return a default result without meaningful fields.`

src/redisearch_rs/rqe_iterators/src/intersection.rs
LukeMathWalker: Once we're in a position to swap, I'd suggest doing a benchmark run where we invoke `.to_owned` on the child result prior to adding them to the aggregate. I'd be curious to see what the overhead is.

src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs
LukeMathWalker: I see, we're running into `RSYieldableMetric_Concat(&mut self.metrics, child.metrics);` being an FFI call 🤔

src/redisearch_rs/rqe_iterators/src/intersection.rs
gdesmott: Good idea.

@dor-forer : can you please add a `TODO` about this in the code so we'll keep it in mind?

dor-forer:
@gdesmott Added the benchamrks

src/redisearch_rs/rqe_iterators_bencher/src/benchers/intersection.rs
```diff
@@ -0,0 +1,192 @@
...
+        })
+        .collect()
+}
+
+/// Convert ID vectors to Rust SortedIdList iterators.
+fn ids_to_rust_children(ids: &[Vec<u64>]) -> Vec<SortedIdList<'static>> {
+    ids.iter()
+        .map(|id_vec| SortedIdList::new(id_vec.clone()))
+        .collect()
+}
```
gdesmott:
```suggestion
fn ids_to_rust_children(ids: Vec<Vec<u64>>) -> Vec<SortedIdList<'static>> {
    ids.into_iter()
        .map(|id_vec| SortedIdList::new(id_vec))
        .collect()
}
```

You can save cloning the `Vec` by passing them by values instead of borrowing.

gdesmott:
The mock iterator has been merged ( https://github.com/RediSearch/RediSearch/pull/7745 ) so you can now rebase your branch on top of `master`.

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,308 @@
...
+        if moved_to_eof {
+            self.is_eof = true;
+            return Ok(RQEValidateStatus::Moved { current: None });
+        }
+
+        match self.skip_to(max_child_doc_id)? {
+            Some(_) => Ok(RQEValidateStatus::Moved {
+                current: Some(&mut self.result),
+            }),
+            None => Ok(RQEValidateStatus::Moved { current: None }),
```
gdesmott, line 276: Can we have a test covering this?
https://app.codecov.io/gh/RediSearch/RediSearch/pull/7663?dropdown=coverage&src=pr&el=h1#d2ced3bd43f9a6c6a40f7bcea9c15f99-R305

src/redisearch_rs/rqe_iterators/tests/integration/utils/mock_iterator.rs
gdesmott: mock iterator has been merged in https://github.com/RediSearch/RediSearch/pull/7745

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,308 @@
...
+    /// - Returns `None` if the iterator is at EOF.
+    /// - Returns `Some(&mut result)` otherwise.
+    ///
+    /// # Note
+    ///
+    /// Before the first call to [`read()`](RQEIterator::read) or
+    /// [`skip_to()`](RQEIterator::skip_to), this returns a reference to the
+    /// uninitialized result buffer (with `doc_id = 0`). Callers should only
+    /// rely on `current()` after a successful read operation.
+    #[inline]
```
gdesmott, line 172:
```suggestion
{
    #[inline]
```

I think we can remove all that as it does not bring any useful info which is not already part of the trait documentation.

src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs
```diff
@@ -0,0 +1,1115 @@
...
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! Integration tests for the Intersection iterator.
+//!
+//! These tests are ported from the C++ tests in
```
gdesmott: We prefix comments referring to the original C/C++ code with `C-Code:` as they will eventually be removed. See `MOD-12457`

gdesmott (commented):
Great work, the implementation is very clear and well-written! 👍

dor-forer:
@GlenDC Waiting for your approval on the requested changes.

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,295 @@
...
+}
+
+impl<'index, I> Intersection<'index, I>
+where
+    I: RQEIterator<'index>,
+{
+    /// Creates a new intersection iterator. Children are sorted by estimated count. If `children`
+    /// is empty, returns an iterator immediately at EOF.
+    #[must_use]
+    pub fn new(mut children: Vec<I>) -> Self {
```
GlenDC, line 57: I would implement this constructor as follows, it avoids the expect and still communicates the same intent:

```rust
#[must_use]
    pub fn new(mut children: Vec<I>) -> Self {
        children.sort_by_cached_key(|c| c.num_estimated());
        let Some(num_expected) = children.first().map(|c| c.num_estimated()) else {
            return Self {
                children,
                last_doc_id: 0,
                num_expected: 0,
                is_eof: true,
                result: RSIndexResult::intersect(0),
            };
        };

        let result = RSIndexResult::intersect(children.len());
        Self {
            children,
            last_doc_id: 0,
            num_expected,
            is_eof: false,
            result,
        }
    }
```

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,295 @@
...
+            last_doc_id: 0,
+            num_expected,
+            is_eof: false,
+            result: RSIndexResult::intersect(num_children),
+        }
+    }
+
+    /// Reads from the first child. Returns the doc_id or None if EOF.
+    fn read_from_first_child(&mut self) -> Result<Option<t_docId>, RQEIteratorError> {
+        debug_assert!(!self.children.is_empty());
```
GlenDC: this debug_assert is not serving any purpose here,
in debug builds this will panic, but in release build we will anyway panic in the next statement if it's empty due to `self.children[0].read()?`, so I think you can just drop the debug_assert all together. It's either way a critical bug in your static logic should it happen, should be easily caught by the set of tests if it ever happens to regress.

If we do ever want to make this impossible to compile we should probably enforce this split between an empty intersect and the regular one by the typesystem and if you wanna go even further by introducing types such as a kind of NonEmptyVec.

For now however I think you ccan just drop the debug assert call

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,295 @@
...
+                self.is_eof = true;
+                Ok(None)
+            }
+        }
+    }
+
+    /// Tries to get all children to agree on the target doc_id.
+    fn agree_on_doc_id(&mut self, target: t_docId) -> Result<AgreeResult, RQEIteratorError> {
+        for child in &mut self.children {
+            // Use child's cached last_doc_id instead of maintaining a parallel array
```
GlenDC, line 93: I think all our iterators do that so far, so not sure we need to document that, it mostly is about a C-difference, which in general we do not document (@gdesmott WDYT?)

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,295 @@
...
+                }
+            }
+        }
+        Ok(AgreeResult::Agreed)
+    }
+
+    /// Loops until all children agree on a doc_id, or EOF is reached.
+    fn find_consensus(
+        &mut self,
+        initial_target: t_docId,
```
GlenDC: there is not much value in introducing a new variable here as you anyway do not use it, so I would just declare this as 

```
mut target: t_docId
```

and use that instead of `curr_target`, but up to you, strong nit pick

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,295 @@
...
+                AgreeResult::Eof => return Ok(None),
+            }
+        }
+    }
+
+    /// Builds the aggregate result from all children's current results.
+    ///
+    /// # Why `unsafe` is used here
+    ///
+    /// We attempted to solve this with safe split borrows (standalone function taking
```
GlenDC, line 129: (You can resolve this as-is, just leaving my feedback on this one)

Very nice comment ,greatly documented on the different points of views provided. Appologies for derailing that conversation slightly earlier in the review process.

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+}
+
+impl<'index, I> RQEIterator<'index> for Intersection<'index, I>
+where
+    I: RQEIterator<'index>,
+{
+    #[inline]
+    fn current(&mut self) -> Option<&mut RSIndexResult<'index>> {
+        if self.is_eof {
```
GlenDC: You can write this shorter as

```rust
self.is_eof.then_some(&mut sell.result)
```

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,295 @@
...
+            Some(&mut self.result)
+        }
+    }
+
+    fn read(&mut self) -> Result<Option<&mut RSIndexResult<'index>>, RQEIteratorError> {
+        if self.is_eof {
+            return Ok(None);
+        }
+
+        let target = match self.read_from_first_child()? {
```
GlenDC: Since recent rust versions you can do for cases like this:

```rust
let Some(target) = self.read_from_first_child()? else {
      return Ok(None)
};
```

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,295 @@
...
+                            max_child_doc_id = max_child_doc_id.max(result.doc_id);
+                        }
+                        None => moved_to_eof = true,
+                    }
+                }
+                RQEValidateStatus::Ok => {}
+            }
+        }
+
+        if !any_child_moved || self.is_eof {
```
GlenDC, line 263: Does it not matter if moved_to_eof is true while self.is_eof is still false?

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -69,6 +71,8 @@ pub trait RQEIterator<'index> {
     /// not have an immediate use for the actual result, and would instead want to keep it aside
     /// for later in time. The child iterator already has that result anyway,
     /// and it is this method which provides the ability to expose it (for later use).
+    /// Calling this method before the first `read` or `skip_to`, or directly after `rewind` will return a default result
```
GlenDC:
```suggestion
    /// 
    ///  # Usage
    /// 
    /// Calling this method before the first `read` or `skip_to`, or directly after `rewind` will return a default result
```

GlenDC (changes requested):
Your PR looks great, left a couple of remarks, but none which will change this PR in any drastical way, so after this it can probably be merged if your team and @gdesmott agree as well.

src/redisearch_rs/rqe_iterators/src/intersection.rs
dor-forer: I changed it to:
`(!self.is_eof).then_some(&mut self.result)`

src/redisearch_rs/rqe_iterators/src/intersection.rs
dor-forer: (!any_child_moved || self.is_eof) handles two early-exit cases: 
1. No child moved at all, or: 
2. The intersection was already at EOF before this revalidate call.

When `moved_to_eof = true` and `self.is_eof = false`, we fall through to line 268, where moved_to_eof is checked, and we properly set self.is_eof = true and return Moved { current: None }.

GlenDC (dismissed):

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,279 @@
...
+            return Self {
+                children,
+                last_doc_id: 0,
+                num_expected: 0,
+                is_eof: true,
+                result: RSIndexResult::intersect(0),
+            };
+        };
+
+        let result = RSIndexResult::intersect(children.len());
```
memark: For consistency with the `return Self { }` above we might want to inline the value of the `result` variable in the below expression.

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -0,0 +1,279 @@
...
+    Eof,
+}
+
+impl<'index, I> Intersection<'index, I>
+where
+    I: RQEIterator<'index>,
+{
+    /// Creates a new intersection iterator. Children are sorted by estimated count. If `children`
+    /// is empty, returns an iterator immediately at EOF.
+    #[must_use]
```
memark, line 56: What's the rationale behind adding this attribute?

memark (commented):
Good code! Some minor comments.

src/redisearch_rs/rqe_iterators/src/intersection.rs
dor-forer: If someone calls Intersection::new(children) without using the result, it's almost certainly a bug - they meant to assign it to a variable.
From what I read it is common in constructors

dor-forer:
@GlenDC @gdesmott Had merge conflicts, so had to rebase manually. Please reapprove.

PR #7745, opened 2025-12-11
add mock rust iterator, ported from the C++ version

GlenDC:

Originally part of Non-Optimized Optional iterator, but pulled out as to unblock several other PRs which would love to make use of the Mock iterator

It allows to test several cases for iterators which make use of child iterators without having to write your own mock iterator from scratch. It's also what allows you to on purpose trigger errors such as a timeout error and verify the expected behaviour as a result of it now and later

N/A

N/A

#### Mark if applicable

---

> [!NOTE]
> Introduce a configurable mock `RQEIterator` and wire a `utils` test module to support iterator-focused integration tests.
> 
> - **Tests**:
>   - **Mock iterator**: Add `utils/mock_iterator.rs` implementing `RQEIterator` over a fixed doc-id sequence with `read`, `skip_to`, `revalidate`, `rewind`, `num_estimated`, `last_doc_id`, and `at_eof`.
>     - Supports configurable outcomes via `MockData` (set `revalidate` result, inject end-of-input error, inspect call counts).
>     - Provides `MockIteratorError` and `MockRevalidateResult` enums.
>   - **Wiring**: Add `tests/integration/utils` module and re-exports; include `utils` in `tests/integration/main.rs` (with expect unused).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 931b6195696dc41778b5bd442056350d02c55c9a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rqe_iterators/tests/integration/utils/mock_iterator.rs
+++ b/src/redisearch_rs/rqe_iterators/tests/integration/utils/mock_iterator.rs
@@ -1,0 +1,324 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use std::{cell::RefCell, rc::Rc};
+
+use ffi::{RS_FIELDMASK_ALL, t_docId};
+use inverted_index::RSIndexResult;
+use rqe_iterators::RQEIterator;
+
+/// Test iterator used in unit tests that expect an [`RQEIterator`]
+/// child which produces a fixed sequence of document identifiers.
+///
+/// `Mock` simulates a very small posting list:
+///
+/// * It owns a fixed array of document ids that must be sorted in
+///   increasing order.
+/// * Calls to [`RQEIterator::read`] walk that array from left to right
+///   and copy the current id into a reusable [`RSIndexResult`] that is
+///   stored inside the iterator.
+/// * Calls to [`RQEIterator::skip_to`] advance `next_index` until it
+///   reaches the requested id or the first id that is greater than it.
+///
+/// The iterator is intentionally simple and deterministic so that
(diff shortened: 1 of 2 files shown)


src/redisearch_rs/rqe_iterators/tests/integration/utils/mock_iterator.rs
```diff
@@ -0,0 +1,324 @@
...
+    ///
+    /// This helper keeps the mock specific error type private to the
+    /// test utilities while still surfacing the correct error value to
+    /// production code and test assertions.
+    fn as_rqe_iterator_error(self) -> rqe_iterators::RQEIteratorError {
+        match self {
+            Self::TimeoutError => rqe_iterators::RQEIteratorError::TimedOut,
+        }
+    }
+}
```
LukeMathWalker, line 83: Why not a `From` impl?

GlenDC: That's public, and I consider this a private detail...

LukeMathWalker: This is in a test crate 🤷🏻

GlenDC: How does that matter? Scopes have different levels, and From still exposes a detail over the entire test crate where it doesn't have any utility... And once it's public it is defacto your API... Yes it's a test crate and yes you can just refactor, but why go for it at all...

It doesn't come with any benefits and it is only to be used within the mock iterator, so not sure why your strong desire for `From` here? I would agree if you expose it as a public function, in that case you want to indeed follow common practices and implement the traits and API conventions that people expect, but for a private detail, I do not follow...

LukeMathWalker: I don't see why it'd be so undesirable to give an extra level of affordance to the rest of the testing crate when working with this mock error type, but it's obviously not a dealbreaker. Feel free to move forward as you prefer.

LukeMathWalker (approved):

PR #7757, opened 2025-12-11
[CI] make manual test trigger API human readable

GuyAv46:
https://github.com/RediSearch/RediSearch/actions/runs/20147867437/job/57833194303

---

> [!NOTE]
> Introduces explicit unit-test control and new inputs in CI workflows, updates options to include unit-tests, and gates C/Rust tests and job behavior accordingly.
> 
> - **Workflows**:
>   - **flow-test.yml**:
>     - Add inputs `test-config`, `unit-tests`, `standalone`, `cluster`, `rejson`, `fail-fast` for manual dispatch; set default options to include `unit-tests`.
>     - Derive flags from `options` or boolean inputs; simplify `test-config` resolution (fallback to `inputs.test-config`).
>     - Update matrix `fail-fast` logic and job name expression; keep coverage/sanitize handling.
>   - **task-test.yml**:
>     - Add `unit-tests` input and include it in run conditions for C/C++ and Rust tests.
>     - Expand common-flow run condition to `(standalone || coordinator || unit-tests)`.
>   - **event-merge-to-queue.yml / event-nightly.yml**:
>     - Update `options` to include `unit-tests` alongside existing `coordinator`, `standalone`, `rejson`, `coverage`, `sanitize`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d87efc466105bf5bfc411d003775dfb0c4872add. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/flow-temp.yml
+++ b/.github/workflows/flow-temp.yml
@@ -9,15 +9,10 @@ name: temporary testing
 
 on:
   push:
-    # branches-ignore: ['**'] # ignore all branches. Comment this line to run your workflow below on every push.
+    branches-ignore: ['**'] # ignore all branches. Comment this line to run your workflow below on every push.
 
 jobs:
 
   merge-queue:
-    uses: ./.github/workflows/flow-test.yml
+    uses: ./.github/workflows/event-merge-to-queue.yml
     secrets: inherit
-    with:
-      platform: ubuntu:jammy
-      architecture: x86_64
-      options: "standalone,rejson"
-      job-test-config: ""


JoanFM (changes requested):
We should change job-test-config to test-config?

What is inputs.opptions?

.github/workflows/flow-test.yml
```diff
@@ -56,30 +56,33 @@ on:
...
-        default: '{"test": "QUICK=1", "coverage": "QUICK=1", "sanitize": "QUICK=1"}'
+        default: QUICK=1
       redis-ref:
         description: 'Redis version to use (e.g. "7.2.3", "unstable"). Defaults to "unstable"'
         type: string
       rejson-branch:
         type: string
         default: master
         description: 'Branch to use when building RedisJSON for tests (default: master)'
-      options:
```
kei-nan, line 70: I think flow-nightly.yml uses this file,
you should probably modify it as well to not use options anymore.

.github/workflows/flow-test.yml
```diff
@@ -56,30 +56,33 @@ on:
...
-          - coordinator,standalone,rejson,fail-fast,coverage
-          - coordinator,standalone,rejson,fail-fast,sanitize
-          - coordinator,standalone,rejson,fail-fast,coverage,sanitize
-        description: 'Comma-separated list of options to enable'
-        default: coordinator,standalone,rejson
+      standalone:
+        description: 'Whether to run standalone tests'
+        type: boolean
+        default: true
+      cluster:
```
kei-nan, line 78: Making sure this is intended.
You don't prefer to keep it as coordinator?

kei-nan (changes requested):
Left two comments

.github/workflows/flow-test.yml
```diff
@@ -97,18 +100,18 @@ jobs:
...
       get-redis: ${{ inputs.redis-ref }}
-      rejson: ${{ contains(inputs.options, 'rejson') }}
+      rejson: ${{ contains(inputs.options, 'rejson') || inputs.RedisJSON }}
       rejson-branch: ${{ inputs.rejson-branch }}
-      coordinator: ${{ contains(inputs.options, 'coordinator') }}
-      standalone: ${{ contains(inputs.options, 'standalone') }}
-      test-config: ${{ fromJson(inputs.job-test-config)[(matrix.job-type || 'test')] }}
+      coordinator: ${{ contains(inputs.options, 'coordinator') || inputs.cluster }}
+      standalone: ${{ contains(inputs.options, 'standalone') || inputs.standalone }}
+      test-config: ${{ fromJson(inputs.job-test-config || '{"test":""}')[(matrix.job-type || 'test')] || inputs.test-config }}
```
kei-nan: Not a fan of this one line logic.
Is there a way to simplify this somehow?
Maybe we can provide defaults to the inputs?
Or use env variables?

kei-nan: Also the input name is no longer `job-test-config`.

.github/workflows/flow-test.yml
GuyAv46: It was intentional
I think since 8.0 we don’t have a “coordinator” mode anymore, we always have it but only use it in cluster mode with multiple shards. I think cluster mode makes more sense paired with standalone, while the coordinator should be paired with the “core module”

.github/workflows/flow-test.yml
GuyAv46: It is still exists.
I have to handle the different inputs on `workflow_call` and `workflow_dispatch`.
This is the recommended way to use variables with fallbacks. I’m sorry you don’t like this syntax

.github/workflows/flow-test.yml
GuyAv46: We use it still on `workflow_call`, I only changed the `workflow_dispatch` inputs, which allowed to be different

.github/workflows/flow-test.yml
GuyAv46: Also, not sure how moving this definition to an env variable will solve the syntax. It will look exactly the same in a different line

.github/workflows/flow-test.yml
kei-nan: ah, okay

.github/workflows/flow-test.yml
kei-nan: first env variable: inputs.job-test-config || '{"test":""}'
second env variable: matrix.job-type || 'test'
I think that would simplify the line

Also, I think the matrix always have a job type, no?
Can we get to a place where we try and access a key in the dict which doesn't exist?

.github/workflows/flow-test.yml
```diff
@@ -56,30 +56,33 @@ on:
...
+        type: boolean
+        default: true
+      cluster:
+        description: 'Whether to run tests in cluster mode'
+        type: boolean
+        default: true
+      RedisJSON:
+        description: 'Whether to include RedisJSON in the tests'
+        type: boolean
+        default: true
```
kei-nan, line 85: if the workflow is not triggered via the workflow dispatch, then will we still get default value for these variables?

.github/workflows/flow-test.yml
GuyAv46: https://github.com/RediSearch/RediSearch/actions/runs/20206107527
This is mandatory because you added the job type access. `[(matrix.job-type || 'test')]`
I have to give a fallback that will return an empty string, so the external fallback will be taken.

.github/workflows/flow-test.yml
GuyAv46: No, it will be undefined if the triggering event is not `workflow_dispatch`

.github/workflows/flow-test.yml
```diff
@@ -97,18 +100,18 @@ jobs:
...
       get-redis: ${{ inputs.redis-ref }}
-      rejson: ${{ contains(inputs.options, 'rejson') }}
+      rejson: ${{ contains(inputs.options, 'rejson') || inputs.RedisJSON }}
       rejson-branch: ${{ inputs.rejson-branch }}
-      coordinator: ${{ contains(inputs.options, 'coordinator') }}
-      standalone: ${{ contains(inputs.options, 'standalone') }}
-      test-config: ${{ fromJson(inputs.job-test-config)[(matrix.job-type || 'test')] }}
+      coordinator: ${{ contains(inputs.options, 'coordinator') || inputs.cluster }}
+      standalone: ${{ contains(inputs.options, 'standalone') || inputs.standalone }}
+      test-config: ${{ fromJson(inputs.job-test-config)[(matrix.job-type || 'test')] || inputs.test-config }}
```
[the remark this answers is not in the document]
GuyAv46: @kei-nan

.github/workflows/flow-test.yml
GuyAv46: https://github.com/RediSearch/RediSearch/actions/runs/20206256414/job/58005297715?pr=7757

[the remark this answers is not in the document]
kei-nan: @GuyAv46 maybe it would make sense to normalize the input variable name,

.github/workflows/flow-test.yml
kei-nan: maybe just add a line of documentation after simplifying it.

.github/workflows/flow-test.yml
```diff
@@ -56,30 +56,33 @@ on:
...
+        type: boolean
+        default: true
+      RedisJSON:
+        description: 'Whether to include RedisJSON in the tests'
+        type: boolean
+        default: true
+      fail-fast:
+        description: 'Whether to fail fast on test failures'
+        type: boolean
+        default: true
```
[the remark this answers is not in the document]
GuyAv46: irrelevant and intentional

.github/workflows/flow-test.yml
```diff
@@ -97,18 +100,18 @@ jobs:
...
       get-redis: ${{ inputs.redis-ref }}
-      rejson: ${{ contains(inputs.options, 'rejson') }}
+      rejson: ${{ contains(inputs.options, 'rejson') || inputs.RedisJSON == true }}
       rejson-branch: ${{ inputs.rejson-branch }}
-      coordinator: ${{ contains(inputs.options, 'coordinator') }}
-      standalone: ${{ contains(inputs.options, 'standalone') }}
-      test-config: ${{ fromJson(inputs.job-test-config)[(matrix.job-type || 'test')] }}
+      coordinator: ${{ contains(inputs.options, 'coordinator') || inputs.cluster == true }}
+      standalone: ${{ contains(inputs.options, 'standalone') || inputs.standalone == true }}
+      test-config: ${{ fromJson(inputs.job-test-config || '{"test":""}')[(matrix.job-type || 'test')] || inputs.test-config }}
```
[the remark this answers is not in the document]
GuyAv46: Only one of `inputs.test-config` and `inputs.job-test-config` will be defined.
When any of them is defined as nothing, the outcome will be nothing (empty)

.github/workflows/flow-test.yml
```diff
@@ -92,23 +95,23 @@ jobs:
...
-      test-config: ${{ fromJson(inputs.job-test-config)[(matrix.job-type || 'test')] }}
-      coverage: ${{ (matrix.job-type || 'test') == 'coverage' }}
-      san: ${{ (matrix.job-type || 'test') == 'sanitize' && 'address' || '' }}
-      fail-fast: ${{ contains(inputs.options, 'fail-fast') }}
+      coordinator: ${{ contains(inputs.options, 'coordinator') || inputs.cluster == true }}
+      standalone: ${{ contains(inputs.options, 'standalone') || inputs.standalone == true }}
+      test-config: ${{ fromJson(inputs.job-test-config || '{}')[matrix.job-type] || inputs.test-config }}
+      coverage: ${{ matrix.job-type == 'coverage' }}
+      san: ${{ matrix.job-type == 'sanitize' && 'address' || '' }}
+      fail-fast: ${{ contains(inputs.options, 'fail-fast') || inputs.fail-fast == true }}
```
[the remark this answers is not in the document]
GuyAv46: wrong, `inputs.options` is available on `workflow_call`

.github/workflows/flow-test.yml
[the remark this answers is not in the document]
GuyAv46: Agree. I thought this variable name was showing when filling the form, but it's not, so no reason to give it a nice name

PR #7765, opened 2025-12-12
MOD-13010: Fix cmp_strings() - correctly handle binary data with embedded NULLs

nafraf:

1. Current: 
  Since `cmp_strings()` uses `strcmp()` the comparison ends when the first NULL character is found, and this causes wrong results when `cmp_strings()` is used for the comparison of byte strings.
2. Change:
 `cmp_strings()` is modified to use `memcmp()` instead of `strcmp()`
3. Outcome: 
  Right comparison of byte strings and text.

---

> [!NOTE]
> Replace strncmp with memcmp in string comparison to support embedded NULs and add a test ensuring aggregate results are coherent with binary strings.
> 
> - **Core**:
>   - Update `cmp_strings` in `src/value/value.c` to use `memcmp` (instead of `strncmp`) for length-aware comparison supporting binary data with embedded `\0`.
> - **Tests**:
>   - Add `test_mod_13010` in `tests/pytests/test_issues.py` verifying coherence between `FT.AGGREGATE` queries with and without `GROUPBY` when values contain embedded NULs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 89baa585d470d0f8a95bd5cf63fe90fe7db94c30. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/value/value.c
+++ b/src/value/value.c
@@ -48,7 +48,7 @@ static inline mempool_t *getPool() {
 
 static inline int cmp_strings(const char *s1, const char *s2, size_t l1, size_t l2) {
   // Use memcmp instead of strncmp to correctly handle binary data with embedded NULLs
-  int cmp = strncmp(s1, s2, MIN(l1, l2));
+  int cmp = memcmp(s1, s2, MIN(l1, l2));
   if (l1 == l2) {
     // if the strings are the same length, just return the result of memcmp
     return cmp;


src/value/value.c
```diff
@@ -47,9 +47,10 @@ static inline mempool_t *getPool() {
 ///////////////////////////////////////////////////////////////
 
 static inline int cmp_strings(const char *s1, const char *s2, size_t l1, size_t l2) {
-  int cmp = strncmp(s1, s2, MIN(l1, l2));
+  // Use memcmp instead of strncmp to correctly handle binary data with embedded NULLs
+  int cmp = memcmp(s1, s2, MIN(l1, l2));
```
oshadmi, line 51: Seems OK - `memcmp` is also comparing bytes, similar to `strncmp` - only supporting embedded nulls.
When comparing null-terminated strings - will behave the same as before - and no extra comparisons will be done.
When comparing slices, which can contain embedded nulls, will behave as expected.

oshadmi (approved):

JoanFM:
Should we aim to have some RSValue unit tests? or porting soon to Rust?

PR #7795, opened 2025-12-15
[MOD-13085] fix: refresh buffer pointers on revalidate

JoanFM:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

---

> [!NOTE]
> When revalidating, readers now refresh buffer pointers if no GC occurred, and `IndexReader_Revalidate` takes a mutable reader; adds a test and plumbs a new `refresh_buffer_pointers` method through readers.
> 
> - **FFI / C-API**
>   - Change `IndexReader_Revalidate` signature to accept `struct IndexReader *` (was `const`), in both Rust FFI and `headers/inverted_index.h`.
>   - On revalidate, if `needs_revalidation` is false, call `ir_dispatch!(..., refresh_buffer_pointers)` before returning.
> - **Core library (`inverted_index`)**
>   - Add `IndexReader::refresh_buffer_pointers(&mut self)` to the trait and implement in `IndexReaderCore` to rebind the internal `Cursor` to potentially reallocated block buffers.
>   - Propagate `refresh_buffer_pointers` through `FilterMaskReader`, `FilterNumericReader`, and `FilterGeoReader`.
> - **Tests**
>   - Add `test_refresh_buffer_pointers_after_reallocation` to verify correct reading after buffer growth/reallocation (skipped under `miri`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0f6c619d15b0451159339d3dca78c8d6f38c5e66. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/inverted_index/src/tests.rs
+++ b/src/redisearch_rs/inverted_index/src/tests.rs
@@ -2009,7 +2009,7 @@ fn ii_apply_gc_entries_tracking_index() {
     );
 }
 // the memory hack below raises error in miri
-#[cfg(not(miri))] 
+#[cfg(not(miri))]
 #[test]
 fn test_refresh_buffer_pointers_after_reallocation() {
     let mut ii = InvertedIndex::<Dummy>::new(IndexFlags_Index_DocIdsOnly);


src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -2004,3 +2008,54 @@ fn ii_apply_gc_entries_tracking_index() {
         }
     );
 }
+
+#[test]
+fn test_refresh_buffer_pointers_after_reallocation() {
+    let mut ii = InvertedIndex::<Dummy>::new(IndexFlags_Index_DocIdsOnly);
```
gdesmott:
```suggestion
// the memory hack below raises error in miri
#[cfg(not(miri))] 
#[test]
fn test_refresh_buffer_pointers_after_reallocation() {
    let mut ii = InvertedIndex::<Dummy>::new(IndexFlags_Index_DocIdsOnly);
```

This should fix the CI.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -1188,6 +1188,9 @@ pub trait IndexReader<'index> {
     /// Check if the underlying index has been modified since the last time this reader read from it.
     /// If it has, then the reader should be reset before reading from it again.
     fn needs_revalidation(&self) -> bool;
+
+    /// Refresh buffer pointers in case blocks were reallocated without GC changes
+    fn refresh_buffer_pointers(&mut self);
```
gdesmott, line 1193: @chesedo : Maybe this should be done directly in `needs_revalidation()` like it uses to be it in the C version?

Adding an extra function to the trait does not seem to bring much.

JoanFM: I think it clarifies the API right? one thing is needing revalidation and another actually perhaps revalidating w.r.t the buffer pointer.

in C, it was done inside revaldate, not in the check of the gc Marker.

Moreover, may be clearer to remove in the future when with more safe Rust there I guess should not be needed?

gdesmott: I think the current API a bit confusing actually. 

This would add an extra API call when using the Rust API directly instead of the C wrapper, that can easily be forgotten.

It also feels like we are leaking implementation details (buffer pointers) inside the `IndexReader` trait.

So I'd like @chesedo 's input wondering it wouldn't make more sense to bundle them together, maybe renamed to something like `check_validation()`.

JoanFM: since it is only on master, can we merge as is, and consider a refactor later on. It is blocking an important work stream

JoanFM: considering everything green of course

gdesmott: Sure, let's merge this for now if it's urgent.

I think you just need to run `cargo fmt` to fix the style and the CI.

JoanFM: yes

gdesmott: Let's keep this thread opened so we can keep discussing the API with @chesedo once he's back.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -1314,6 +1317,16 @@ impl<'index, E: DecodedBy<Decoder = D>, D: Decoder> IndexReader<'index>
     fn needs_revalidation(&self) -> bool {
         self.gc_marker != self.ii.gc_marker.load(atomic::Ordering::Relaxed)
     }
+
+    fn refresh_buffer_pointers(&mut self) {
+        if !self.ii.blocks.is_empty() && self.current_block_idx < self.ii.blocks.len() {
```
GuyAv46, line 1322: Not really expected, but what should we do if that's not the case? Should we set it to EOF? Reset the cursor anyway to some allocation-safe pointer (null? dangling?)

JoanFM: lets keep this in mind for the review. Will create a ticket for this.

Since it is only in master we can wait and let ASM continue

src/redisearch_rs/inverted_index/src/lib.rs
chesedo: Hmm, I also prefer the more explicit case with the new method. But we should give this API another visit to see if there are ways in which it can be improved once we are done with the iterators.

src/redisearch_rs/inverted_index/src/tests.rs
```diff
@@ -2004,3 +2008,55 @@ fn ii_apply_gc_entries_tracking_index() {
...
+    ii.add_record(&RSIndexResult::default().doc_id(10)).unwrap();
+    ii.add_record(&RSIndexResult::default().doc_id(11)).unwrap();
+
+    // SAFETY: We need to bypass Rust's borrowing rules to simulate the real-world
+    // scenario where buffer reallocation happens while a reader is active.
+    // This is safe because:
+    // 1. We're not accessing the reader during the mutation
+    // 2. The InvertedIndex structure remains valid
+    // 3. We call refresh_buffer_pointers before using the reader again
+    let ii_ptr = &mut ii as *mut InvertedIndex<Dummy>;
```
LukeMathWalker, line 2027: This is undefined behaviour, which is why `miri` flags it. The fact that we are going through a raw pointer + `unsafe` to "work around" the issue doesn't make it OK.

The real fix will come by modelling revalidation properly, and we should amend the comment above accordingly.

PR #7818, opened 2025-12-17
Consolidate Intel tests into flow-test.yml

meiravgri:
### Summary
This PR integrates Intel EC2-based testing into the unified `flow-test.yml` workflow, eliminating the separate `flow-intel.yml` workflow and consolidating all platform testing into a single matrix-based approach.

### Changes

**`task-get-config.yml`**
- Added `ec2_image_id` and `ec2_instance_type` outputs for self-hosted EC2 runner configuration

**`task-test.yml`**
- Added `start-runner` job that provisions EC2 runner when `ec2_image_id` is set in config
- Added `stop-runner` job for cleanup (runs on `always()` to ensure instance termination)
- `common-flow` now uses EC2 runner label when available, falls back to regular runners otherwise

**`generate-matrix.yml`**
- Added Intel (`intel`, `x86_64`) to `all_combinations`

**`flow-test.yml`**
- Added `intel` to the platform dropdown options

**`event-merge-to-queue.yml`**
- Added `intel` to the platform list
- Removed separate `run-on-intel` job

**`event-nightly.yml`**
- Removed separate `run-on-intel` job (Intel now runs as part of `platform: all`)

### How It Works
1. When `platform: intel` (or `all`) is requested, the matrix includes Intel
2. `task-get-config.yml` returns EC2 config for Intel (AMI + instance type)
3. `start-runner` job detects non-empty `ec2_image_id` and provisions EC2 instance
4. `common-flow` runs tests on the EC2 runner
5. `stop-runner` always terminates the EC2 instance after tests complete

---

> [!NOTE]
> Integrates Intel x86_64 EC2 self-hosted runner into the unified `flow-test.yml` matrix and removes separate Intel workflow calls.
> 
> - **CI workflows**:
>   - Add `intel` to `platform` in `.github/workflows/event-merge-to-queue.yml`; drop separate `run-on-intel` job/needs.
>   - In `.github/workflows/event-nightly.yml`, remove `run-on-intel` and adjust `notify-failure` needs.
> - **Matrix generation**:
>   - Add `intel` to `workflow_dispatch` options in `flow-test.yml` and include `intel` in matrix via `.github/workflows/generate-matrix.yml`.
> - **Configuration outputs**:
>   - In `.github/workflows/task-get-config.yml`, add `ec2_image_id` and `ec2_instance_type` outputs; define `intel` platform config with AMI `ami-09fabd03bb09b3704` and instance `c7i.xlarge`.
> - **Test workflow integration** (`.github/workflows/task-test.yml`):
>   - Add `start-runner` job to launch EC2 via `machulav/ec2-github-runner@v2` and `stop-runner` for cleanup.
>   - Route `common-flow` to run on the EC2 runner label when `intel` is selected; update `needs` accordingly.
>   - Keep container build path unchanged; no behavioral changes to test steps aside from runner orchestration.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 45c8ee5b1dea8a29fc1b10c13ad49636e99ba092. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


.github/workflows/flow-test.yml
```diff
@@ -87,6 +87,10 @@ on:
         description: 'Fail fast on test failures'
         type: boolean
         default: true
+      intel:
+        description: 'Run Intel tests on self-hosted EC2 runner'
+        type: boolean
+        default: true
```
GuyAv46: Consider making `intel` a new platform option to choose, instead of a checkbox (it doesn’t make sense to pick alpine+ARM, for example, and then get both this and an intel jobs)

meiravgri: But it also means that it will be included in "all" and can't be explicitly disabled like sanitizer or codecov
Is this what we want (i do 😄 )?

GuyAv46: I removed the option to manually trigger coverage or sanitizer (IMO it doesn’t make sense to run these without a PR context)

From the `workflow_dispatch` API, I think we should treat it as any other platform (and may or may not include it in “all”)

In the `workflow_call` API, we can have it as a special option like coverage and sanitizer, if you prefer. I think it make sense

.github/workflows/event-nightly.yml
```diff
@@ -19,19 +19,13 @@ jobs:
       job-test-config: '{"test": "QUICK=1", "coverage": "QUICK=1", "sanitize": "QUICK=1"}'
       options: ${{ vars.ENABLE_CODE_COVERAGE != 'false' && 'unit-tests,coordinator,standalone,rejson,coverage,sanitize' || 'unit-tests,coordinator,standalone,rejson,sanitize' }}
 
-  run-on-intel:
```
GuyAv46, line 22: So not running on nightly anymore

.github/workflows/task-test.yml
```diff
@@ -85,6 +89,46 @@ jobs:
...
+      AWS_INSTANCE_TYPE: c7i.xlarge
+      TAGS: | # Make sure there is no trailing comma!
+        [
+        {"Key": "Name",         "Value": "redisearch-ci-runner"},
+        {"Key": "Environment",  "Value": "CI"},
+        {"Key": "Run ID",       "Value": "${{ github.run_id }}"},
+        {"Key": "PR",           "Value": "${{ github.event.number }}"},
+        {"Key": "Owner",        "Value": "${{ github.actor }}"},
+        {"Key": "Project",      "Value": "${{ github.repository }}"}
+        ]
```
GuyAv46, line 101: For better string safety:
```suggestion
        [
          {"Key": "Name",         "Value": "redisearch-ci-runner"},
          {"Key": "Environment",  "Value": "CI"},
          {"Key": "Run ID",       "Value": ${{ toJson(github.run_id) }}},
          {"Key": "PR",           "Value": "${{ github.event.number }}"},
          {"Key": "Owner",        "Value": ${{ toJson(github.actor) }}},
          {"Key": "Project",      "Value": ${{ toJson(github.repository) }}}
        ]
```

.github/workflows/task-test.yml
```diff
@@ -85,6 +89,46 @@ jobs:
...
       architecture: ${{ inputs.architecture }}
+
+  # Start EC2 runner for self-hosted platforms
+  start-runner:
+    name: Start self-hosted EC2 runner
+    if: ${{ inputs.self-hosted }}
+    runs-on: ubuntu-latest
+    env:
+      AWS_IMAGE_ID: ami-09fabd03bb09b3704
+      AWS_INSTANCE_TYPE: c7i.xlarge
```
GuyAv46: These should come from the config job when asking for `intel`. It won't always be the right AMI and type.

Also, consider relying on whether this value is not empty to run this job, instead of the `self-hosted` input

.github/workflows/task-test.yml
```diff
@@ -152,14 +196,18 @@ jobs:
...
   common-flow:
-    needs: [get-config, build-image]
+    needs: [get-config, build-image, start-runner]
     name: Test ${{ needs.get-config.outputs.name }}, Redis ${{ inputs.get-redis || 'unstable' }}
     # Run if get-config succeeded and at least one test type is enabled
+    # Note: start-runner may be skipped for non-self-hosted platforms
+    # always() is required for the if condition to be evaluated when start-runner is skipped
     if: |
+      always() &&
       !cancelled() &&
```
GuyAv46, line 195: Isn't the `!cancelled()` good enough for that?

.github/workflows/task-test.yml
```diff
@@ -546,3 +594,24 @@ jobs:
           flags: unit
           fail_ci_if_error: true # Fail on upload errors
           token: ${{ secrets.CODECOV_TOKEN }}
+
+  # Stop EC2 runner for self-hosted platforms (always runs to ensure cleanup)
```
GuyAv46:
```suggestion
  # Stop EC2 runner for self-hosted platforms (always runs if `start-runner` succeeds, to ensure cleanup)
```

.github/workflows/task-get-config.yml
```diff
@@ -30,6 +30,12 @@ on:
       name:
         description: "Configuration name"
         value: ${{ jobs.get-config.outputs.name }}
+      ec2_image_id:
```
kei-nan, line 33: do we really need them as input variables?
how is the file using them? it looks like we are the ones generating them.

.github/workflows/task-test.yml
```diff
@@ -152,14 +191,14 @@ jobs:
...
-    needs: [get-config, build-image]
+    needs: [get-config, build-image, start-runner]
     name: Test ${{ needs.get-config.outputs.name }}, Redis ${{ inputs.get-redis || 'unstable' }}
     # Run if get-config succeeded and at least one test type is enabled
     if: |
       !cancelled() &&
       needs.get-config.result == 'success' &&
       (inputs.standalone || inputs.coordinator || inputs.unit-tests)
-    runs-on: ${{ inputs.custom-env || needs.get-config.outputs.env || vars.RUNS_ON || 'ubuntu-latest' }}
+    runs-on: ${{ needs.get-config.outputs.ec2_image_id && needs.start-runner.outputs.label || inputs.custom-env || needs.get-config.outputs.env || vars.RUNS_ON || 'ubuntu-latest' }}
```
kei-nan: I think custom-env is no longer needed, I think we had it only for intel
can probably be removed.

.github/workflows/task-get-config.yml
meiravgri: Those are outputs used by the start-runner job

.github/workflows/task-test.yml
meiravgri: done

.github/workflows/task-test.yml
```diff
@@ -152,14 +191,14 @@ jobs:
...
     container: ${{
```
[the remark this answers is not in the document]
kei-nan: @meiravgri does this flow works for non intel environments?
according to the comment the entire test will be skipped.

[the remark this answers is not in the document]
meiravgri: yes. https://github.com/RediSearch/RediSearch/actions/runs/20308034100/job/58331061552
`build-image` is also part of the `needs` array and skipped if we are not running inside a container

PR #7822, opened 2025-12-17
[MOD-13115] RSValue - Cleanup C interface usage

daxhuiberts:
This PR is the amalgamation of a lot of small changes to `value.h` and code which uses it so it is in a cleaner state to ease the migration to Rust.

---

> [!NOTE]
> Refactors RSValue’s C API and updates all call sites for consistency and safety.
> 
> - Replace legacy setters (`RSValue_IntoNumber/IntoNull/IntoUndefined`) with `RSValue_SetNumber/SetNull` and `RSValue_Clear`; propagate across expression evaluation and functions
> - Unify string handling: remove `RSValueType_OwnRstring` and borrowed/owned/stolen constructors; add `RSValue_NewRedisString` with explicit `RedisModule_RetainString` where needed; update reply, rlookup, JSON paths, highlighting
> - Narrow string checks from `RSValue_IsAnyString` to `RSValue_IsString`; update geo/string functions, highlighter, reducers
> - Rework random-sample reducer to store `RSValue**` with proper refcounting and build the result array at finalize
> - Introduce `RSValueSize`, adjust memory sizing in sortable; reorder RSValueType enum values (affects error codes), and update C++/Python tests accordingly
> - Misc: use `RSValue_DecrRef` instead of `RSValue_Free`, minor includes/whitespace fixes
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 40c5b67f419a92a580282ae27e00e26dd5345eea. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/value/value.c
+++ b/src/value/value.c
@@ -259,16 +259,8 @@ bool RSValue_IsNumber(const RSValue *v) {
   return (v && v->_t == RSValueType_Number);
 }
 
-bool RSValue_IsString(const RSValue *v) {
-  return (v && v->_t == RSValueType_String);
-}
-
 bool RSValue_IsArray(const RSValue *v) {
   return (v && v->_t == RSValueType_Array);
-}
-
-bool RSValue_IsRedisString(const RSValue *v) {
-  return (v && v->_t == RSValueType_RedisString);
 }
 
 bool RSValue_IsTrio(const RSValue *v) {
@@ -502,7 +494,7 @@ const char *RSValue_ConvertStringPtrLen(const RSValue *value, size_t *lenp, char
                                         size_t buflen) {
   value = RSValue_Dereference(value);
 
-  if (RSValue_IsAnyString(value)) {
+  if (RSValue_IsString(value)) {
     return RSValue_StringPtrLen(value, lenp);
   } else if (value->_t == RSValueType_Number) {
     size_t n = snprintf(buf, buflen, "%f", value->_numval);
--- a/src/value/value.h
+++ b/src/value/value.h
@@ -283,25 +283,11 @@ bool RSValue_IsReference(const RSValue *v);
 bool RSValue_IsNumber(const RSValue *v);
 
 /**
- * Check if the RSValue is a string type.
- * @param v The value to check
- * @return true if the value is of type RSValueType_String, false otherwise
- */
-bool RSValue_IsString(const RSValue *v);
-
-/**
  * Check if the RSValue is an array type.
  * @param v The value to check
  * @return true if the value is of type RSValueType_Array, false otherwise
  */
 bool RSValue_IsArray(const RSValue *v);
-
-/**
- * Check if the RSValue is a Redis string type.
- * @param v The value to check
- * @return true if the value is of type RSValueType_RedisString, false otherwise
- */
-bool RSValue_IsRedisString(const RSValue *v);
 
 /**
  * Check whether the RSValue is of type RSValueType_Trio.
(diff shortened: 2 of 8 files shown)


src/value/value.h
```diff
@@ -41,8 +41,6 @@ typedef enum {
   RSValueType_RedisString = 5,
   // An array of values, that can be of any type
   RSValueType_Array = 6,
-  // A redis string, but we own a refcount to it; tied to RSDummy
-  RSValueType_OwnRstring = 7,
   // Reference to another value
   RSValueType_Reference = 8,
```
raz-mon:
```suggestion
  RSValueType_Reference = 7,
```
And the rest below..

src/value/value.c
```diff
@@ -243,37 +220,6 @@ RSValue *RSValue_NewMap(RSValueMap map) {
...
-
-RSValue *RSValue_NewConstStringArray(char **strs, uint32_t sz) {
-  RSValue **arr = RSValue_AllocateArray(sz);
-
-  for (uint32_t i = 0; i < sz; i++) {
-    arr[i] = RSValue_NewConstString(strs[i], strlen(strs[i]));
-  }
-  return RSValue_NewArray(arr, sz);
-}
-
```
raz-mon, line 276: All unused?

src/value/value.c
```diff
@@ -484,24 +419,14 @@ uint16_t RSValue_Refcount(const RSValue *v) {
 // Other Functions (utility, comparison, conversion, etc.)
 ///////////////////////////////////////////////////////////////
 
-/* Free a value's internal value. It only does anything in the case of a string, and doesn't free
- * the actual value object */
-void RSValue_Free(RSValue *v) {
```
raz-mon, line 489: Why was this removed?

src/aggregate/reducers/random_sample.c
```diff
@@ -36,27 +36,39 @@ static int sampleAdd(Reducer *rbase, void *ctx, const RLookupRow *srcrow) {
...
-  return RSValue_IncrRef(sc->samplesArray); // return a reference to the array
+
+  uint32_t len = MIN(sc->seen, r->len);
+  RSValue **array_ptr = rm_calloc(sizeof(RSValue *), len);
+
+  for (uint32_t i = 0; i < len; i++) {
+    array_ptr[i] = RSValue_IncrRef(sc->samplesArray[i]);
+  }
+
+  return RSValue_NewArray(array_ptr, len);
```
raz-mon, line 61: Why was this added? Please add some documentation to the function.

raz-mon (commented):
Overall LGTM, but the reason for some of the changes is not clear to me, some docs would greatly help.

src/aggregate/reducers/random_sample.c
daxhuiberts: The RSValue array was first mutated in place which will not be possible in the Rust version. So this is changed to build up a version in memory before being passed to `RSValue_NewArray`. This now behaves the same to other areas where `RsValue_NewArray` is used.

src/value/value.c
daxhuiberts: Yes, nowhere in the code is this used. Only in a unit test, but the buildup is being done there directly.

src/value/value.c
daxhuiberts: The logic of this free function is added to `RSValue_DecrRef`.
`RSValue_Free` should never be called directly.

src/value/value.h
daxhuiberts: I thought since `2` is also already missing it might be preferred to not change the numbering?

daxhuiberts:
@raz-mon Added two more changes (see two last commits)

daxhuiberts:
> Overall LGTM, but the reason for some of the changes is not clear to me, some docs would greatly help.

I kept all separate changes to individual commits. That should make it clearer. Not sure where docs would help, except perhaps in the `random_sample.c` change.

src/value/value.h
raz-mon: Why would it be preferred? I don't see a reason for it.

raz-mon (approved):
LGTM
But, please do fix the numbering of the enum as I see no good reason for the holes

raz-mon (approved):

PR #7833, opened 2025-12-18
MOD-13497 feat: RedisJSON bindings

JonasKruckenberg:

This change adds Rust bindings for RedisJSON so later work can call that modules APIs.

1. MOD-...
2. #...

1. ...

#### Mark if applicable

---

> [!NOTE]
> Adds a new `redis_json_api` crate exposing safe, version-gated wrappers around the RedisJSON C API.
> 
> - Defines `RedisJsonApi` handle plus `JsonValue`/`JsonValueRef`, `JsonPath`, `ResultsIter`, `KeyValuesIterator`, and `JsonType` with accessors and serialization helpers
> - Supports opening keys (`openKey*`), JSONPath parsing/checks, iterating query results and object key-values, and retrieving scalar/container data (`getInt`, `getDouble`, `getString`, `getAt`, `getLen`, etc.)
> - Constants for API versioning and root path: `MIN_API_VERSION`, `LATEST_API_VERSION`, `JSON_ROOT`
> - Updates workspace to include `redis_json_api` and adds `src/json.h` to FFI bindgen headers in `ffi/build.rs`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 47bdebe33af134c4dd832fb9f07e206a0a2f66e2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/redis_json_api/src/results.rs
+++ b/src/redisearch_rs/redis_json_api/src/results.rs
@@ -17,7 +17,6 @@ use std::ptr::NonNull;
 ///
 /// This iterator is returned by [`JsonValueRef::get`] and yields
 /// all values matching a JSON path expression.
-// TODO should be lending iterator
 pub struct ResultsIter<'a> {
     ptr: NonNull<c_void>,
     next: unsafe extern "C" fn(ffi::JSONResultsIterator) -> ffi::RedisJSON,


src/redisearch_rs/redis_json_api/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// 1. `ctx` must be a valid Redis module context.
+    pub unsafe fn open_key_with_flags(
+        &self,
+        ctx: *mut ffi::RedisModuleCtx,
+        key_name: &RedisString,
+        flags: i32,
+    ) -> Option<JsonValueRef<'_>> {
+        let vtable = self.vtable();
+        let open_key_with_flags = vtable
+            .openKeyWithFlags
```
oshadmi, line 156: Should we add a safety requirement that caller must ensure the require API version is available, e.g., `openKeyWithFlags` is available since API V5.

src/redisearch_rs/redis_json_api/src/results.rs
```diff
@@ -0,0 +1,135 @@
...
+use crate::{JsonValueRef, SerializeError};
+use redis_module::RedisString;
+use std::os::raw::c_void;
+use std::ptr::NonNull;
+
+/// An iterator over JSON query results.
+///
+/// This iterator is returned by [`JsonValueRef::get`] and yields
+/// all values matching a JSON path expression.
+// TODO should be lending iterator
```
oshadmi: The values returned from the iterator are actually owned by the JSON document and not by the iterator? So no need to lend?

src/redisearch_rs/redis_json_api/src/lib.rs
JonasKruckenberg: If the function is not available it will crash with a panic, which isn't great but no _unsafe_ in the Rust sense. Adding a safety comment therefore doesn't make much sense BUT documenting the minimum required versions for these functions actually is a good idea 👍

src/redisearch_rs/redis_json_api/src/results.rs
JonasKruckenberg: Ah yes good catch, this was leftover from a previous version

src/redisearch_rs/redis_json_api/src/value.rs
```diff
@@ -0,0 +1,352 @@
...
+use crate::SerializeError;
+use core::slice;
+use std::ffi::CStr;
+use std::ffi::c_char;
+use std::ffi::c_void;
+use std::ptr::NonNull;
+
+/// The type of a JSON value.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum JsonType {
```
oshadmi, line 26: We have now 3 definitions, one in RedisJSON, one in RediSearch C code (rejson_api.h) and this one in RediSearch Rust code.
What about using an opaque `JSONType` type as C code is using? (`ffi::JSONType`)

JonasKruckenberg: `JSONType` is just an alias for `c_uint` which doesn't afford us the nice pattern matching benefits of Rust. I think is worth the slight duplication cost (after all how often is this going to change?)

JonasKruckenberg:
added the version number hints, rebased onto master and fixed a new lint

src/redisearch_rs/redis_json_api/src/value.rs
oshadmi: OK. Could be worthwhile to add comments referencing the corresponding definitions.

JonasKruckenberg: added a comment to the type 👍🏻

PR #7837, opened 2025-12-18
[CI] Add missing setup step

GuyAv46:

Add a missing step to install Bash on Alpine, in `task-test.yml`.
This is usually not required if the cached Docker container job succeeds, but if it fails, the Alpine flow will fail as well

https://github.com/RediSearch/RediSearch/actions/runs/20341974069

#### Mark if applicable

---

> [!NOTE]
> Adds a conditional step to install Bash via apk when running `alpine:3.22` container jobs in `.github/workflows/task-test.yml`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4910c4a7642f62c0c0fe5e6bd4aa65d72ae8634e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/task-test.yml
+++ b/.github/workflows/task-test.yml
@@ -209,6 +209,11 @@ jobs:
       run:
         shell: bash -l -eo pipefail {0}
     steps:
+      - name: Install bash
+        if: needs.get-config.outputs.container == 'alpine:3.22'
+        shell: sh -l -eo pipefail {0}
+        run: ${{ format(env.SETUP_RETRY_LOOP, 'apk add bash') }}
+
       - name: Free up disk space (container)
         # For container jobs - delete mounted host directories
         if: ${{ needs.get-config.outputs.container }}


.github/workflows/task-test.yml
```diff
@@ -209,6 +209,11 @@ jobs:
       run:
         shell: bash -l -eo pipefail {0}
     steps:
+      - name: Install bash
```
meiravgri, line 212: should we skip if build-image succeeded?

GuyAv46: We can, but if `bash` is already installed, this step is a no-op.
I'd rather keep it aligned with `task-build-artifacts.yml` and the `8.2` version

meiravgri: 8.2 also have cached containers now 😄

PR #7841, opened 2025-12-19
[MOD-13156] Rewrite schema rule wrapper

memark:
Rewrite the SchemaRuleWrapper into a simpler type, more appropriate for what its callers need. Update all call sites and tests accordingly.

---

> [!NOTE]
> **Rewrite schema rule wrapper**
> 
> - Introduces `SchemaRule` (transparent wrapper over `ffi::SchemaRule`) replacing `SchemaRuleWrapper`; adds `from_raw`, `lang_field/score_field/payload_field`, and simplifies `is_special_key` with `maybe_cstr_from_ptr`.
> - Changes FFI: `RLookup_GetLength` now takes `const SchemaRule*` and handles null; header `rlookup_rs.h` updated accordingly.
> - Updates `row` APIs to take `Option<&SchemaRule>` and adjusts internal logic to filter special keys.
> - Cleans up exports/imports (`pub use schema_rule::SchemaRule`; remove `SchemaRuleWrapper`), and adds `SchemaRule` to cbindgen `exclude` list.
> - Refactors tests to construct `ffi::SchemaRule` directly and use `SchemaRule::from_raw`; removes wrapper-specific allocation/free logic.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit db5b92195792602160d78a5f68a2ca2a22a71725. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rlookup/src/schema_rule.rs
+++ b/src/redisearch_rs/rlookup/src/schema_rule.rs
@@ -11,7 +11,7 @@ use std::ffi::{CStr, c_char};
 
 use crate::RLookupKey;
 
-/// A safe wrapper around a pointer to a `SchemaRule`, the underlying pointer is [NonNull].
+/// A safe wrapper around a pointer to a `SchemaRule`, the underlying pointer is non-null.
 ///
 /// # Safety
 ///


JonasKruckenberg:
> It can now also contain a null pointer.

I don't think that right now? We passing a `&` reference which can never be null

memark:
> > It can now also contain a null pointer.
> 
> I don't think that right now? We passing a `&` reference which can never be null

You are correct. I read to quickly that we removed the NonNull type, and wanted to not leave the description too empty... 🤣 But we still unwrap() on the ref ofc. Updated description!

memark:
@JonasKruckenberg After rebasing, the ffi::SchemaRule is no longer Clone/Copy, so I removed those derives from the wrapper as well. I don't think it has any adverse effects.

PR #7864, opened 2025-12-22
[MOD-13212] Make sure Flex explicitly only supports TERM indices (fail if other indices used/define)

JoanFM:
> [!NOTE]
> Strengthens Flex schema validation to only allow `TEXT` fields and surfaces a clear error for unsupported types.
> 
> - Adds `QUERY_ERROR_CODE_FLEX_UNSUPPORTED_FIELD` to C/Rust `QueryErrorCode` with message "Unsupported field for Flex index"
> - Introduces `SearchDisk_MarkUnsupportedFieldIfDiskEnabled` and integrates it into `parseFieldSpec` to fail `TAG`, `GEO`, `GEOSHAPE`, `NUMERIC`, and `VECTOR` when Flex validation is on
> - Keeps existing limit check via `SearchDisk_CheckLimitNumberOfIndexes`
> - Tests: `test_flex_max_index_limit` (unchanged behavior) and new `test_invalid_field_type` asserting specific error messages for disallowed field types
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7dc081697e8c85cc059dd25ace7d860fda971925. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/search_disk_utils.c
+++ b/src/search_disk_utils.c
@@ -17,7 +17,7 @@ bool SearchDisk_CheckLimitNumberOfIndexes(size_t nIndexes) {
   return nIndexes <= FLEX_MAX_INDEX_COUNT;
 }
 
-bool SearchDisk_MarkUnsupportedField(const char *fieldTypeStr, const FieldSpec *fs, QueryError *status) {
+bool SearchDisk_MarkUnsupportedFieldIfDiskEnabled(const char *fieldTypeStr, const FieldSpec *fs, QueryError *status) {
   if (SearchDisk_IsEnabledForValidation()) {
     QueryError_SetWithoutUserDataFmt(status, QUERY_ERROR_CODE_FLEX_UNSUPPORTED_FIELD, "%s fields are not supported in Flex indexes", fieldTypeStr);
     return false;
--- a/src/search_disk_utils.h
+++ b/src/search_disk_utils.h
@@ -31,4 +31,4 @@ bool SearchDisk_CheckLimitNumberOfIndexes(size_t nIndexes);
  * @param status Query error status
  * @return true if the field type is supported, false otherwise
  */
-bool SearchDisk_MarkUnsupportedField(const char *fieldTypeStr, const FieldSpec *fs, QueryError *status);
+bool SearchDisk_MarkUnsupportedFieldIfDiskEnabled(const char *fieldTypeStr, const FieldSpec *fs, QueryError *status);
--- a/src/spec.c
+++ b/src/spec.c
@@ -1262,27 +1262,27 @@ static int parseFieldSpec(ArgsCursor *ac, IndexSpec *sp, StrongRef sp_ref, Field
       sp->flags |= Index_HasNonEmpty;
     }
   } else if (AC_AdvanceIfMatch(ac, SPEC_TAG_STR)) {  // tag field
-    if (!SearchDisk_MarkUnsupportedField(SPEC_TAG_STR, fs, status)) goto error;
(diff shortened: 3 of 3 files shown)


alonre24 (approved):

src/spec.c
```diff
@@ -1248,11 +1248,18 @@ static int parseGeometryField(IndexSpec *sp, FieldSpec *fs, ArgsCursor *ac, Quer
   return 1;
 }
 
+static bool checkFlexFieldSupport(const char *fieldTypeStr, const FieldSpec *fs, QueryError *status) {
```
kei-nan: I feel like we are going to have many of these types of functions.
maybe worth having them in one dedicated header/c files?

src/redisearch_rs/query_error/src/lib.rs
```diff
@@ -153,6 +154,7 @@ impl QueryErrorCode {
             Self::OutOfMemory => c"Not enough memory available to execute the query",
             Self::UnavailableSlots => c"Query requires unavailable slots",
             Self::FlexLimitNumberOfIndexes => c"Flex index limit was reached",
+            Self::FlexInvalidField => c"Invalid field for Flex index",
```
kei-nan:
```suggestion
            Self::FlexUnsupportedField => c"Unsupported field for Flex index",
```

src/search_disk_utils.c
```diff
@@ -10,14 +10,17 @@
...
- */
 bool SearchDisk_CheckLimitNumberOfIndexes(size_t nIndexes) {
   if (!SearchDisk_IsEnabledForValidation()) {
     return true;
   }
   return nIndexes <= FLEX_MAX_INDEX_COUNT;
 }
+
+bool SearchDisk_CheckFieldSupport(const char *fieldTypeStr, const FieldSpec *fs, QueryError *status) {
+  if (SearchDisk_IsEnabledForValidation()) {
```
kei-nan, line 21: shouldn't there be actual logic here checking if the field type is supported?

JoanFM: yes, changed the name of the method, since in order to check it should have Parsing capabilites which I do not want, just changed the name of the function

src/redisearch_rs/query_error/src/lib.rs
JoanFM: done

src/search_disk_utils.h
```diff
@@ -20,3 +22,13 @@
...
  */
 bool SearchDisk_CheckLimitNumberOfIndexes(size_t nIndexes);
+
+/**
+ * @brief Mark a field as unsupported in Flex indexes
+ *
+ * @param fieldTypeStr Field type string
+ * @param fs Field specification
+ * @param status Query error status
+ * @return true if the field type is supported, false otherwise
```
kei-nan, line 32: the return type doesn't make a whole lot of sense
on the one hand the function is now called mark which seems to indicate the decision if a field is supported or not happens in the caller.
on the other hand, it returns true or false to indicate if the field is supported.
Maybe it should be void and you can add `SearchDisk_IsEnabledForValidation` to the if in the spec.c file.

kei-nan (commented):
I think the `SearchDisk_MarkUnsupportedField` function should change a bit.
I think the return type is conflicting with the function name.

src/search_disk_utils.h
JoanFM: remaned it to be more explicit, would not change anything else

kei-nan (approved):
Okay, approving

PR #7869, opened 2025-12-23
[MOD-12899] Add warning when Cursor may give inaccurate results because of ASM

JoanFM:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

---

> [!NOTE]
> Introduces end-to-end handling for ASM-induced result inaccuracy during slot migration, including reply warnings, cursor state propagation, and global observability.
> 
> - Adds `QEXEC_S_ASM_TRIMMING_DELAY_TIMEOUT` to `QEStateFlags`; sets it when ASM max delay elapses (via `CursorList_MarkASMInaccuracy()` from `notifications.c`) and when shards return the new warning (`rpnet.c`)
> - Emits warnings in replies: RESP2 increments stats; RESP3 appends `QUERY_ASM_INACCURATE_RESULTS` to `warning` and updates global stats in `aggregate_exec.c`
> - Tracks new stats: `QUERY_WARNING_CODE_ASM_INACCURATE_RESULTS` and `QUERY_ERROR_CODE_UNAVAILABLE_SLOTS`; exposes in `INFO MODULES` (coord/shard counters) via `global_stats.[ch]` and `info_redis.c`
> - Extends Rust/C FFI with new warning codes and strings in `query_error` headers and lib
> - Adds cursor API to mark all active cursors as inaccurate (`CursorList_MarkASMInaccuracy`) and propagates the flag in replies; initializes `cursor_reads`/`stateflags`
> - Updates coordinator pipeline to parse/propagate ASM warnings from shards (`rpnet.c`)
> - Expands tests (`tests/pytests/test_asm.py`) to validate RESP2/RESP3 cursor behavior, warnings text, and INFO counters
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a67aa8d3af6a5896cf8c05ce8b2c955963e1aa2a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_asm.py
+++ b/tests/pytests/test_asm.py
@@ -801,17 +801,30 @@ def _test_ft_cursors_trimmed(env: Env, protocol: int):
       env.assertNotEqual(results_set, expected_set)
       env.assertGreater(len(expected_set), len(results_set))
 
-      shard_num_warnings = 0
-      coord_num_warnings = 0
-      for shard in env.getOSSMasterNodesConnectionList():
+      shard_total_num_warnings = 0
+      coord_total_num_warnings = 0
+      for shard_id, shard in enumerate(env.getOSSMasterNodesConnectionList(), 1):
+        shard_num_warnings = 0
+        coord_num_warnings = 0
         info_dict = info_modules_to_dict(shard)
         if 'search_warnings_and_errors' in info_dict and 'search_shard_total_query_warnings_asm_inaccurate_results' in info_dict['search_warnings_and_errors']:
-            shard_num_warnings += int(info_dict['search_warnings_and_errors']['search_shard_total_query_warnings_asm_inaccurate_results'])
+            shard_num_warnings = int(info_dict['search_warnings_and_errors']['search_shard_total_query_warnings_asm_inaccurate_results'])
         if 'search_coordinator_warnings_and_errors' in info_dict and 'search_coord_total_query_warnings_asm_inaccurate_results' in info_dict['search_coordinator_warnings_and_errors']:
-            coord_num_warnings += int(info_dict['search_coordinator_warnings_and_errors']['search_coord_total_query_warnings_asm_inaccurate_results'])
-      env.assertGreater(shard_num_warnings + coord_num_warnings, 0)
-      if protocol == 3:
-        env.assertEqual(num_warnings, coord_num_warnings)
+            coord_num_warnings = int(info_dict['search_coordinator_warnings_and_errors']['search_coord_total_query_warnings_asm_inaccurate_results'])
+        if shard_id == 1:
+            env.assertEqual(shard_num_warnings, 0)
+            env.assertGreater(coord_num_warnings, 0)
+            if protocol == 3:
+              # ShardID 1 is the coordinator so it gets the warnings as nonInternal and are the ones seen in the replies
+              env.assertEqual(num_warnings, coord_num_warnings)
+        elif (shard_id == 2):
+            # ShardID 2 is the one where trimming happens (source shard), so it puts its warnings in the shard
+            env.assertGreater(shard_num_warnings, 0)
+            env.assertEqual(coord_num_warnings, 0)
+        else:
+            # Other shards don't have any warnings
+            env.assertEqual(shard_num_warnings, 0)
+            env.assertEqual(coord_num_warnings, 0)
 
 @skip(cluster=False, min_shards=2)
 def test_ft_cursors_trimmed_protocol_2():


src/redisearch_rs/headers/query_error.h
```diff
@@ -34,6 +34,7 @@
 #define QUERY_WINDEXING_FAILURE "Index contains partial data due to an indexing failure caused by insufficient memory"
 #define QUERY_WOOM_SHARD "One or more shards failed to execute the query due to insufficient memory"
 #define QUERY_WOOM_COORD "Coordinator failed to execute the query due to insufficient memory"
+#define QUERY_ASM_INACCURATE_RESULTS "Query execution exceeded maximum delay for RediSearch to delay key trimming. Results may be incomplete or have duplications due to Atomic Slot Migration."
```
alonre24: Duplications are possible? I thought that only missing results are possible in that case

src/cursor.c
```diff
@@ -204,6 +204,23 @@ static uint64_t CursorList_GenerateId(CursorList *curlist) {
...
+
+void CursorList_MarkASMInaccuracy() {
+  CursorList *cl = getCursorList(true);
+  CursorList_Lock(cl);
+  Cursors_ForEach(cl, cursorMarkASMInaccuracyCb, NULL);
+  CursorList_Unlock(cl);
+  cl = getCursorList(false);
+  CursorList_Lock(cl);
+  Cursors_ForEach(cl, cursorMarkASMInaccuracyCb, NULL);
+  CursorList_Unlock(cl);
```
alonre24, line 221: Did we test the case where we set the warning in internal cursors?

src/redisearch_rs/query_error/src/lib.rs
```diff
@@ -255,6 +257,10 @@ impl QueryWarningCode {
             Self::OutOfMemoryCoord => {
                 c"One or more shards failed to execute the query due to insufficient memory"
             }
+            Self::UnavailableSlots => c"Query requires unavailable slots",
+            Self::AsmInaccurateResults => {
+                c"Query execution exceeded maximum delay for RediSearch to delay key trimming. Results may be incomplete or have duplications due to Atomic Slot Migration."
```
alonre24: Where are we using these?

src/info/global_stats.c
```diff
@@ -102,13 +102,18 @@ QueriesGlobalStats TotalGlobalStats_GetQueryStats() {
   stats.coord_errors.timeout = READ(RSGlobalStats.totalStats.queries.coord_errors.timeout);
   stats.shard_errors.oom = READ(RSGlobalStats.totalStats.queries.shard_errors.oom);
   stats.coord_errors.oom = READ(RSGlobalStats.totalStats.queries.coord_errors.oom);
+  stats.shard_errors.unavailableSlots = READ(RSGlobalStats.totalStats.queries.shard_errors.unavailableSlots);
+  stats.coord_errors.unavailableSlots = READ(RSGlobalStats.totalStats.queries.coord_errors.unavailableSlots);
```
alonre24, line 106: Where is that set?

tests/pytests/test_asm.py
```diff
@@ -738,29 +755,87 @@ def _test_ft_cursors_trimmed(env: Env):
...
+
+      shard_num_warnings = 0
+      coord_num_warnings = 0
+      for shard in env.getOSSMasterNodesConnectionList():
+        info_dict = info_modules_to_dict(shard)
+        if 'search_warnings_and_errors' in info_dict and 'search_shard_total_query_warnings_asm_inaccurate_results' in info_dict['search_warnings_and_errors']:
+            shard_num_warnings += int(info_dict['search_warnings_and_errors']['search_shard_total_query_warnings_asm_inaccurate_results'])
+        if 'search_coordinator_warnings_and_errors' in info_dict and 'search_coord_total_query_warnings_asm_inaccurate_results' in info_dict['search_coordinator_warnings_and_errors']:
+            coord_num_warnings += int(info_dict['search_coordinator_warnings_and_errors']['search_coord_total_query_warnings_asm_inaccurate_results'])
+      env.assertGreater(shard_num_warnings + coord_num_warnings, 0)
```
alonre24: Consider verifying each shard seperatly + coord for better granularity

src/redisearch_rs/headers/query_error.h
JoanFM: I think u are right, will correct

src/cursor.c
JoanFM: I think is very hard to test that, and is very hard we would have that. But I think for completion is good to mark everything we have there.

src/redisearch_rs/query_error/src/lib.rs
JoanFM: These are internals of redisearch_rs, to have complete headers., and to get stats and error conversion from string to error.

src/info/global_stats.c
JoanFM: Automatically mapping from the error name to the Query error code.

tests/pytests/test_asm.py
JoanFM: I think is hard to know which shard is giving the error, too much deep inside how cursors work with the 1000 results and so on

I think is best to have this error just to show the warnings are there and the stats show them

JoanFM:
> Nice! Few questions Also, what do you think increase the max trimming default from 5 seconds to ~60 seconds, to be on the safe side with long run queries? @GuyAv46 @JoanFM

I do not have a strong opinion, but this should anyway be handled in another PR.

I think delaying too much can be more harmful to the normal operations of the engine than giving some inaccurate results on some set of queries

src/aggregate/aggregate.h
```diff
@@ -286,6 +286,9 @@ typedef struct AREQ {
   // Number of cursor reads: 1 for the initial FT.AGGREGATE WITHCURSOR,
   // plus 1 for each subsequent FT.CURSOR READ call.
   size_t cursor_reads;
+
+  /** Flag indicating that the cursor might be return bad results because ASM trimming started while Cursor is alive */
+  bool asm_potential_inaccuracy;
```
JoanFM: check using flags

tests/pytests/test_asm.py
JoanFM: done

src/cursor.c
alonre24: There is a pause mechanism introduced via FT.DEBUG command that may allow testing it more easily, consider using it (nota a must)

JoanFM: feels like an overkill

GuyAv46 (approved):

GuyAv46:
> What do you think about increasing the max trimming default from 5 seconds to ~60 seconds, to be on the safe side with long run queries?

@alonre24 I think it's too high for a default. This delay causes both memory spikes and blocks additional migrations. Because it also affects cursors, and their default idle time is 5 minutes, in a user-cursor use-cases we are likely to wait the full delay time.
Maybe we can detect the use case and give more time if the user doesn't have user (coord) cursors?

alonre24:
> > What do you think about increasing the max trimming default from 5 seconds to ~60 seconds, to be on the safe side with long run queries?
> 
> @alonre24 I think it's too high for a default. This delay causes both memory spikes and blocks additional migrations. Because it also affects cursors, and their default idle time is 5 minutes, in a user-cursor use-cases we are likely to wait the full delay time. Maybe we can detect the use case and give more time if the user doesn't have user (coord) cursors?

You mean give more time if the user *does* have user (coord) cursors? I think we can go with than compromise. @JoanFM WDYT? Should we add that in this PR?

JoanFM:
> > > What do you think about increasing the max trimming default from 5 seconds to ~60 seconds, to be on the safe side with long run queries?
> > 
> > 
> > @alonre24 I think it's too high for a default. This delay causes both memory spikes and blocks additional migrations. Because it also affects cursors, and their default idle time is 5 minutes, in a user-cursor use-cases we are likely to wait the full delay time. Maybe we can detect the use case and give more time if the user doesn't have user (coord) cursors?
> 
> You mean give more time if the user _does_ have user (coord) cursors? I think we can go with than compromise. @JoanFM WDYT? Should we add that in this PR?

I think this should go into another PR. I also would vote for simplicity and keep as is now and potentially improve in the future. What would be the case if they have user cursors? wait 60s in that case?

GuyAv46:
The other way around - if there are **_no_** user cursors, extend the max delay.
But I’m not sure it’s a simple check, we need to know if a local internal cursor is for a user cursor on another shard. We need to add a marker to the internal command or something. 
let’s think about it and do it in another PR

JoanFM:
> The other way around - if there are no user cursors, extend the max delay. But I’m not sure it’s a simple check, we need to know if a local internal cursor is for a user cursor on another shard. We need to add a marker to the internal command or something. let’s think about it and do it in another PR

Then I do not think complexity is justified at this moment

PR #7892, opened 2025-12-25
MOD-13252 Support multiple warnings in reply

meiravgri:
### Problem
Currently, if a query triggers more than one warning (e.g., both timeout and max prefix expansions reached), only one warning is exposed to the client while the rest are silently dropped.

### Solution
Modified warning handling to collect and return all relevant warnings from query execution.

#### FT.PROFILE (⚠️ breaking change)
Per-shard warning field changes from **string** to **array of strings**:

**Location:** `res['Profile']['Shards'][i]['Warning']`

| Before | After |
|--------|-------|
| `"Timeout limit was reached"` | `["Timeout limit was reached", "Max prefix expansions limit was reached"]` |

#### FT.SEARCH / FT.AGGREGATE (RESP3)
`res['Results']['warning']` was already an array but only ever contained one warning. Now it can contain multiple:

| Before | After |
|--------|-------|
| `["Timeout limit was reached"]` | `["Timeout limit was reached", "Max prefix expansions limit was reached"]` |

#### Coordinator warnings (cluster mode, RESP3)
The warnings field remains an array, but we now aggregate warnings from all shards instead of just one.

---

> [!NOTE]
> Introduces multi-warning support and unifies warning formatting.
> 
> - Refactors RESP3 warning emission into `_replyWarnings` and ensures Results `warning` accumulates multiple entries
> - Coordinator (`rpnet.c`) iterates all warning strings from shards, setting timeout/max-prefix/OOM/indexing-failure/ASM flags; handles empty replies and returns timed out when present
> - Changes FT.PROFILE `Warning` from string to array for shards and coordinator; prints multiple warnings or `['None']`
> - Updates tests to expect arrays and multiple warnings; adds `get_shards_profile` helper in `common.py` and adjusts RESP2 expectations where profile packs warnings in arrays
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c996a3c94d0a5aa1e1d085554b7d7be3f74d51d9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_resp3.py
+++ b/tests/pytests/test_resp3.py
@@ -1611,6 +1611,24 @@ def test_multiple_warnings():
   env.assertContains('Timeout limit was reached', res['warning'])
   env.assertContains('Max prefix expansions limit was reached', res['warning'])
 
+  # Query with wildcard that exceeds the limit and force timeout
+  query = ['FT.PROFILE', 'idx', 'AGGREGATE', 'QUERY', 'prefix*']
+  timeout_after_n = 1
+  res = runDebugQueryCommandTimeoutAfterN(env, query, timeout_after_n, internal_only=True)
+
+  shards_profile = get_shards_profile(env, res)
+  env.assertEqual(len(shards_profile), env.shardsCount, message=f"unexpected shard count: {res}")
+  for shard_profile in shards_profile:
+    env.assertContains('Timeout limit was reached', shard_profile['Warning'])
+    if env.isCluster():
+        # MOD-12984: In cluster mode, warnings are not persisted across cursor reads,
+        # so only the timeout warning is present.
+        # Once MOD-12984 is fixed, remove this branch and keep only the assertContains below.
+        env.assertEqual(len(shard_profile['Warning']), 1, message=f"full reply output: {res}")
+    else:
+        # In standalone mode, both warnings should be present
+        env.assertContains('Max prefix expansions limit was reached', shard_profile['Warning'])
+
 # TODO: `total_results` is currently not  on cluster - to be fixed in MOD-9094
 @skip(cluster=True)
 def test_totalResults_aggregate():


src/aggregate/aggregate_exec.c
```diff
@@ -631,7 +631,8 @@ static void sendChunk_Resp3(AREQ *req, RedisModule_Reply *reply, size_t limit,
     } else if (rc == RS_RESULT_ERROR) {
       // Non-fatal error
       RedisModule_Reply_SimpleString(reply, QueryError_GetUserError(qctx->err));
-    } else if (QueryError_HasReachedMaxPrefixExpansionsWarning(qctx->err)) {
+    }
```
lerman25: Why do we still keep it in  `if else` (the `RS_RESULT_ERROR`)?

lerman25: Also - consider moving the entire section of warning reply to an helper function, the code get's messy

image

src/coord/rpnet.c
```diff
@@ -229,18 +229,22 @@ static bool shardResponseBarrier_HandleError(RPNet *nc) {
...
-      const char *warning_str = MRReply_String(MRReply_ArrayElement(warning, 0), NULL);
+    size_t num_warnings = MRReply_Length(warning);
+    // Iterate over all warnings in the array
+    for (size_t i = 0; i < num_warnings; i++) {
+      const char *warning_str = MRReply_String(MRReply_ArrayElement(warning, i), NULL);
       // Set an error to be later picked up and sent as a warning
       if (!strcmp(warning_str, QueryError_Strerror(QUERY_ERROR_CODE_TIMED_OUT))) {
         timed_out = true;
-      } else if (!strcmp(warning_str, QUERY_WMAXPREFIXEXPANSIONS)) {
+      }
```
lerman25: The else - if from before is still correct though 
Maybe it's even better.
But I think a refactor for this code might be the better approach (smarter string compare)

tests/pytests/test_resp3.py
```diff
@@ -1582,10 +1582,35 @@ def test_warning_maxprefixexpansions():
   env.assertEqual(res['Results']['warning'], ['Max prefix expansions limit was reached'])
   n_warnings = 0
   for i, shard in enumerate(res['Profile']['Shards']):
-    if shard['Warning']== 'Max prefix expansions limit was reached':
+    if 'Max prefix expansions limit was reached' in shard['Warning']:
          n_warnings += 1
   env.assertEqual(n_warnings, 1)
 
+def test_multiple_warnings():
```
lerman25, line 1589: Can't we test the profile also here?

src/aggregate/aggregate_exec.c
meiravgri: the rc code can't get more than one value

lerman25: So it shouldn't be only an else to timeout right?

src/coord/rpnet.c
meiravgri: right :)

lerman25 (approved):

PR #7829, opened 2025-12-18
[8.4][MOD-12171] [MOD-12169] Implement ASM-aware search flow (#7343)

JoanFM:
Backport #7343 to 8.4

---

> [!NOTE]
> Introduces ASM-aware execution and slot validation across query flows with versioned keyspace tracking and controlled trimming.
> 
> - Replace `slotsVersion`/`SharedSlotRangeArray` with `keySpaceVersion` + `querySlots` across `AREQ`, pipeline params, and `RPQueryIterator`; remove local slots cache and related APIs
> - Validate `_SLOTS_INFO` via `slots_tracker_check_availability`; set `keySpaceVersion`, track query counts with `ASM_KeySpaceVersionTracker_*`; return `QUERY_EUNAVAILABLE_SLOTS` on unavailable slots
> - Filter results by requested `querySlots` when `key_space_version` changes; add `DOCUMENT_OPEN_KEY_QUERY_FLAGS` `OPEN_KEY_ACCESS_TRIMMED`
> - Add ASM state machine (`asm_state_machine.h`) for slot updates, migration/trim events, and per-version query tracking; wire into module init/end and coordinator (`MR_UpdateTopology`), plus immediate topology reschedule helper
> - Implement trimming coordination with timers and new configs: `_MIN_TRIM_DELAY_MS`, `_MAX_TRIM_DELAY_MS`, `_TRIMMING_STATE_CHECK_DELAY_MS` (with validation). New error `QUERY__TRIM_DELAY_CONFIG_INVALID`
> - Update hybrid/coordinator paths to drop local-slots passing; adjust parsing/build to use `keySpaceVersion`
> - Add `SlotRangeArray_ContainsSlot`, expose `slots_tracker_reset`, and new error code `QUERY_EUNAVAILABLE_SLOTS`
> - CI: temporarily pin Redis ref to `8.4`; Build: switch includes to `deps/RedisModulesSDK/8.4`
> - Extensive test updates for ASM flows, trimming, configs, and slot-range helpers
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b4407360c91306d48bfa8d99a6007e149a0d234a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #7903, opened 2025-12-28
MOD-13143: Solve Hybrid ASM Memory Leak

kei-nan:

1. Current: There is a memory leak in 8.4 in regards to FT.HYBRID and the use of GetLocalSlots
2. Change: Take the slots in the main thread and pass them in the concurrent ctx struct, clone them to each subquery so they will be destroyed correctly
3. Outcome: No memory leaks from the asm feature in regard to hybrid.

#### Mark if applicable

---

> [!NOTE]
> Addresses leaks and thread-safety around cluster slot ranges in query execution.
> 
> - Change `AREQ_ApplyContext` to accept `localSlots`; store on request and pass to pipeline; update all call sites
> - Capture `Slots_GetLocalSlots()` on main thread in concurrent context, expose via `ConcurrentCmdCtx_GetSlotsMutable`, and free after worker completion
> - Add `Slots_Clone` and ensure slots are cloned per hybrid subquery and freed in cleanup paths (hybrid exec/debug/parse)
> - Update FT.HYBRID flow to thread slot ranges through `parseHybridCommand`, `HybridRequest_prepareForExecution`, and execution cleanup
> - Adjust tests to pass and free slot ranges
> - Command info: change `FILTER` arg to a block with `COUNT`, `POLICY` (ADHOC|BATCHES), optional `BATCH_SIZE`; bump `since` to 8.4.4
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 669f47fd5a02a11e49367589626efdd4e5d5c3de. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/command_info/command_info.c
+++ b/src/command_info/command_info.c
@@ -3116,13 +3116,49 @@ int SetFtHybridInfo(RedisModuleCommand *cmd) {
       {
         .name = "filter",
         .token = "FILTER",
-        .type = REDISMODULE_ARG_TYPE_STRING,
+        .type = REDISMODULE_ARG_TYPE_BLOCK,
         .flags = REDISMODULE_CMD_ARG_OPTIONAL,
+        .subargs = (RedisModuleCommandArg[]){
+          {
+            .name = "count",
+            .type = REDISMODULE_ARG_TYPE_INTEGER,
+          },
+          {
+            .name = "filter_expression",
+            .type = REDISMODULE_ARG_TYPE_STRING,
+          },
+          {
+            .name = "policy",
+            .token = "POLICY",
+            .type = REDISMODULE_ARG_TYPE_ONEOF,
+            .flags = REDISMODULE_CMD_ARG_OPTIONAL,
+            .subargs = (RedisModuleCommandArg[]){
+              {
+                .name = "adhoc",
+                .token = "ADHOC",
+                .type = REDISMODULE_ARG_TYPE_PURE_TOKEN,
+              },
+              {
+                .name = "batches",
+                .token = "BATCHES",
+                .type = REDISMODULE_ARG_TYPE_PURE_TOKEN,
+              },
+              {0}
+            },
+          },
+          {
+            .name = "batch_size_value",
+            .token = "BATCH_SIZE",
+            .type = REDISMODULE_ARG_TYPE_INTEGER,
+            .flags = REDISMODULE_CMD_ARG_OPTIONAL,
+          },
+          {0}
+        },
       },
       {0}
     },
     .arity = -4,
-    .since = "8.4.0",
+    .since = "8.4.4",
   };
   return RedisModule_SetCommandInfo(cmd, &info);
 }


JoanFM:
The implementation is ablut to change drastically, so I am not sure it is worth applying this change now and making the real ASM backport more complex

src/command_info/command_info.c
```diff
@@ -3116,13 +3116,49 @@ int SetFtHybridInfo(RedisModuleCommand *cmd) {
       {
         .name = "filter",
         .token = "FILTER",
-        .type = REDISMODULE_ARG_TYPE_STRING,
+        .type = REDISMODULE_ARG_TYPE_BLOCK,
         .flags = REDISMODULE_CMD_ARG_OPTIONAL,
+        .subargs = (RedisModuleCommandArg[]){
```
GuyAv46: This is unrelated to ASM, right? Not sure what's going on here

src/concurrent_ctx.c
```diff
@@ -94,6 +98,10 @@ WeakRef ConcurrentCmdCtx_GetWeakRef(ConcurrentCmdCtx *cctx) {
   return cctx->spec_ref;
 }
 
+const SharedSlotRangeArray *ConcurrentCmdCtx_GetSlots(ConcurrentCmdCtx *cctx) {
+  return cctx->slotRanges;
```
GuyAv46, line 102: Should we return a clone, or move ownership here? Passing the pointer is not enough.
Consider setting `cctx->slotRanges` to `NULL` before returning

src/concurrent_ctx.h
```diff
@@ -60,6 +60,9 @@ void ConcurrentCmdCtx_KeepRedisCtx(struct ConcurrentCmdCtx *ctx);
 // Returns the WeakRef held in the context.
 WeakRef ConcurrentCmdCtx_GetWeakRef(struct ConcurrentCmdCtx *cctx);
 
+// Returns the local slots held in the context.
```
GuyAv46: Ownership?

src/hybrid/parse_hybrid.c
```diff
@@ -816,11 +817,11 @@ int parseHybridCommand(RedisModuleCtx *ctx, ArgsCursor *ac,
   prefixes = NULL;
 
   // Apply context to each request
-  if (AREQ_ApplyContext(searchRequest, searchRequest->sctx, status) != REDISMODULE_OK) {
+  if (AREQ_ApplyContext(searchRequest, searchRequest->sctx, status, Slots_Clone(parsedCmdCtx->localSlots)) != REDISMODULE_OK) {
```
GuyAv46, line 820: Where do we clean up `parsedCmdCtx->localSlots` on error? Who owns the reference in the context?

src/command_info/command_info.c
kei-nan: yep, I think this is because Itzik backported the json command info generation to 8.4
@Itzikvaknin FYI

src/hybrid/parse_hybrid.c
kei-nan: if called from coordinator background thread the concurrent ctx will clean it up.
if called from main thread then will be cleaned up in `CleanupAndReplyStatus` function
if called from FT.DEBUG it releases it right after the parse function call.

src/concurrent_ctx.c
kei-nan: The ownership belongs to the concurrent ctx.
if the caller decides to use it then he will need to clone it as appropriate.
This is available to be used during the lifetime of the concurrent function call.

src/concurrent_ctx.h
kei-nan: updated the comment

GuyAv46 (approved):

GuyAv46:
Should be reverted in #7829. Let's add the revert commit once this PR is merged, so we don't forget to do so.
Plus, there is the command info stuff that I guess should not be reverted. @kei-nan can you handle the revert commit?

src/slot_ranges.c
```diff
@@ -25,7 +25,7 @@ struct SharedSlotRangeArray {
 // Dropped when Slots_DropCachedLocalSlots is called (when we know local slots have changed)
 static SharedSlotRangeArray *localSlots = NULL;
 
-const SharedSlotRangeArray *Slots_GetLocalSlots(void) {
+SharedSlotRangeArray *Slots_GetLocalSlotsMutable() {
```
GuyAv46: Unexpected, but this PR is temporary, so I'll approve this.
But for the record, this is a shared object, so it should never be mutable (apart from internal reference counting and freeing)

src/slot_ranges.h
```diff
@@ -21,6 +21,11 @@ extern "C" {
 
 typedef struct SharedSlotRangeArray SharedSlotRangeArray;
 
+/// @brief Get mutable slot ranges for the local node. The returned value must be freed using FreeLocalSlots
```
GuyAv46: "mutable"

GuyAv46 (approved):

src/slot_ranges.c
kei-nan: I needed it for the ref counting, didn't want to cast away the const in order to increment the ref count.

GuyAv46: I agree it's not the best, but I prefer to cast away const internally when it's clearly needed (and done safely) instead of exposing a mutable reference.
In Rust, we could do internal mutability instead :)

src/command_info/command_info.c
```diff
@@ -3116,13 +3116,49 @@ int SetFtHybridInfo(RedisModuleCommand *cmd) {
...
+            .flags = REDISMODULE_CMD_ARG_OPTIONAL,
+          },
+          {0}
+        },
       },
       {0}
     },
     .arity = -4,
-    .since = "8.4.0",
+    .since = "8.4.4",
```
oshadmi: Not 8.4.3?

src/slot_ranges.c
```diff
@@ -41,6 +41,17 @@ const SharedSlotRangeArray *Slots_GetLocalSlots(void) {
   return localSlots;
 }
 
+const SharedSlotRangeArray *Slots_Clone(const SharedSlotRangeArray *src) {
+  // remove constness - needed in order to manage the refcount
+  SharedSlotRangeArray *slots = (SharedSlotRangeArray *)src;
+  if (!slots) {
+    return NULL;
+  }
+  uint32_t refcount = atomic_fetch_add_explicit(&slots->refcount, 1, memory_order_acquire);
```
oshadmi: For incrementing the ref count, can `memory_order_relaxed` suffice?

src/command_info/command_info.c
kei-nan: that is what Itzik set in the commands.json.

src/slot_ranges.c
kei-nan: I don't think we can because of the decrement section in the free function.
the memory ordering here will affect the logic there.

src/command_info/command_info.c
Itzikvaknin: I wasn't sure, I think I used this PR to also update the VSIM FILTER in commands.json

src/slot_ranges.c
oshadmi: Consider what Cursor suggested:
* `memory_order_relaxed` when incrementing (since we already have a valid reference to increment from, there's nothing we need to synchronize with - relaxed is optimal)
* `memory_order_acq_rel` when releasing (e.g., in `Slots_FreeLocalSlots`)
(need **release** semantics on the decrement to ensure writes to the object are visible to other threads. acquire only gives us visibility of others' writes, not the other way around)

oshadmi: Also suggested a slightly more efficient option (but a bit more code)
```
// Increment
atomic_fetch_add(&ref_count, 1, memory_order_relaxed);

// Decrement
if (atomic_fetch_sub(&ref_count, 1, memory_order_release) == 1) {
    atomic_thread_fence(memory_order_acquire);
    destroy();
}
```

(more efficient than `memory_order_acq_rel` because we only pay for the acquire barrier when actually destructing, not on every decrement)

GuyAv46: I think we can use `relaxed` because we know the count won't reach 0 while we are here, as the caller has a valid reference.
Anyway, `acquire` is also good

kei-nan: will change to relaxed

src/slot_ranges.c
kei-nan: used a cast in the end

src/slot_ranges.h
kei-nan: removed the code, so resolving

src/command_info/command_info.c
oshadmi: RediSearch 8.4.3 is the coming release.
Anyway, this also requires a change in commands.json which is the source from which commands_info.c is generated, no?

oshadmi: I see it is aligned with the commands.json. Let's keep 8.4.4.

kei-nan: okay, 
tried reverting it
will bring it back

oshadmi (approved):

kei-nan:
https://github.com/RediSearch/RediSearch/pull/7903#issuecomment-3694813168
The PR already removed part of the slots code
going to be a bit tricky to revert

GuyAv46:
@kei-nan right.
Maybe branch from 8.4, revert this PR, and then CP the rest of #7829.

kei-nan:
> @kei-nan right. Maybe branch from 8.4, revert this PR, and then CP the rest of #7829.

I the end opened a PR in 8.4 which reverts this PR after talking to Joan.
we should merge that one before we merge Joan's PR.

PR #7913, opened 2025-12-29
[MOD-12899] add asm warning in Profile

JoanFM:
Handle the warning from ASM in profile

---

> [!NOTE]
> Highlights ASM-related inaccuracy to users and tooling during Atomic Slot Migration.
> 
> - Add `PROFILE_WARNING_TYPE_ASM_INACCURATE_RESULTS` to `profile.h`; include it in `Profile_Print` warnings and bitset checks
> - In `aggregate_exec.c`, when `QEXEC_S_ASM_TRIMMING_DELAY_TIMEOUT` is set, update global stats, add `PROFILE_WARNING_TYPE_ASM_INACCURATE_RESULTS`, and reply `QUERY_ASM_INACCURATE_RESULTS` across RESP2, RESP3, and empty-results paths (including cursor flows and `_replyWarnings`)
> - Extend tests in `tests/pytests/test_asm.py` to verify `FT.CURSOR PROFILE` returns the ASM warning during migration (both single-threaded and BG variants)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9c702cf0041b4c82f8aaac9463aa1f7ac4a8a725. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -766,17 +766,18 @@ void sendChunk(AREQ *req, RedisModule_Reply *reply, size_t limit) {
       ProfileWarnings_Add(&req->profileCtx.warnings, PROFILE_WARNING_TYPE_QUERY_OOM);
     }
 
-    // Add BG_SCAN_OOM warning to profile context if applicable
-    RedisSearchCtx *sctx = AREQ_SearchCtx(req);
-    if (sctx && sctx->spec && sctx->spec->scan_failed_OOM) {
-      ProfileWarnings_Add(&req->profileCtx.warnings, PROFILE_WARNING_TYPE_BG_SCAN_OOM);
-    }
     if (req->stateflags & QEXEC_S_ASM_TRIMMING_DELAY_TIMEOUT) {
       QueryWarningsGlobalStats_UpdateWarning(QUERY_WARNING_CODE_ASM_INACCURATE_RESULTS, 1, !IsInternal(req));
       ProfileWarnings_Add(&req->profileCtx.warnings, PROFILE_WARNING_TYPE_ASM_INACCURATE_RESULTS);
       RedisModule_Reply_SimpleString(reply, QUERY_ASM_INACCURATE_RESULTS);
     }
     RedisModule_Reply_ArrayEnd(reply);
+
+    // Add BG_SCAN_OOM warning to profile context if applicable
+    RedisSearchCtx *sctx = AREQ_SearchCtx(req);
+    if (sctx && sctx->spec && sctx->spec->scan_failed_OOM) {
+      ProfileWarnings_Add(&req->profileCtx.warnings, PROFILE_WARNING_TYPE_BG_SCAN_OOM);
+    }
 
     if (IsProfile(req)) {
       RedisModule_Reply_MapEnd(reply);  // >Results


src/cursor.c
```diff
@@ -207,6 +207,7 @@ static uint64_t CursorList_GenerateId(CursorList *curlist) {
 static void cursorMarkASMInaccuracyCb(CursorList *cl, Cursor *cur, void *arg) {
   if (cur->execState) {
     cur->execState->stateflags |= QEXEC_S_ASM_TRIMMING_DELAY_TIMEOUT;
+    ProfileWarnings_Add(AREQ_ProfilePrinterCtx(cur->execState), PROFILE_WARNING_TYPE_ASM_INACCURATE_RESULTS);
```
meiravgri: We already updated profile warnings in sendChunk, should we do it here as well?

src/profile.c
```diff
@@ -54,10 +54,12 @@ void printInvIdxIt(RedisModule_Reply *reply, QueryIterator *root, ProfileCounter
       decodeGeo(it->profileCtx.numeric.rangeMax, nw);
       RedisModule_Reply_SimpleStringf(reply, "%g,%g - %g,%g", se[0], se[1], nw[0], nw[1]);
     }
-  } else {
-    printProfileType("TEXT");
+  } else if (root && root->current) {
```
meiravgri: Is this change related to this PR?

src/profile.h
```diff
@@ -62,6 +62,7 @@ typedef enum {
   PROFILE_WARNING_TYPE_MAX_PREFIX_EXPANSIONS = 1 << 1,
   PROFILE_WARNING_TYPE_QUERY_OOM = 1 << 2,
   PROFILE_WARNING_TYPE_BG_SCAN_OOM = 1 << 3,
+  PROFILE_WARNING_TYPE_ASM_INACCURATE_RESULTS = 1 << 4,
```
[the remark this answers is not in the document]
meiravgri: please address :) either change order or the assert

tests/pytests/test_asm.py
```diff
@@ -772,17 +772,38 @@ def info_modules_to_dict(conn):
           info[section_name][data[0]] = data[1]
     return info
 
+def _test_ft_cursors_trimmed_profile_warning(env: Env):
+    for shard in env.getOSSMasterNodesConnectionList():
+        shard.execute_command('CONFIG', 'SET', 'search-_max-trim-delay-ms', 2500)
```
meiravgri: consider using `run_command_on_all_shards`

tests/pytests/test_asm.py
```diff
@@ -772,17 +772,38 @@ def info_modules_to_dict(conn):
...
+
+    n_docs = 2**14
```
meiravgri, line 778: 🤯 Do we really need this amount of docs?

tests/pytests/test_asm.py
```diff
@@ -772,17 +772,38 @@ def info_modules_to_dict(conn):
...
+    create_and_populate_index(env, 'idx', n_docs)
+
+    shard1, shard2 = env.getConnection(1), env.getConnection(2)
+    query = ('_FT.PROFILE', 'idx', 'AGGREGATE', 'QUERY', '@n:[1 999999]', 'LOAD', 1, 'n', 'WITHCURSOR', '_SLOTS_INFO', generate_slots(range(int(2**14/env.shardsCount) + 1, 2 * int(2**14/env.shardsCount) + 2)))
+    shard2.execute_command('DEBUG', 'MARK-INTERNAL-CLIENT')
+    _, cursor_id = shard2.execute_command(*query)
+    wait_for_migration_complete(env, shard1, shard2)
+    time.sleep(5)
+    num_warnings = 0
+    while cursor_id != 0:
```
meiravgri: '_FT.CURSOR PROFILE', always returns cursor_id == 0

tests/pytests/test_asm.py
```diff
@@ -772,17 +772,38 @@ def info_modules_to_dict(conn):
...
+        res, cursor_id = shard2.execute_command('_FT.CURSOR', 'PROFILE', 'idx', cursor_id, 'COUNT', 1)
+        if 'Warning' in res['Profile']['Shards'][0] and len(res['Profile']['Shards'][0]['Warning']) > 0:
```
meiravgri: The warning field always exists. It may be an array containing only 'None'

meiravgri:
Please add a short description and a MOD reference

tests/pytests/test_asm.py
JoanFM: it is ASM, having at least one doc per slot makes sense

src/profile.c
JoanFM: I have another PR open for a crash found with this change, u can review that one separately

JoanFM: https://github.com/RediSearch/RediSearch/pull/7956

src/cursor.c
JoanFM: solved.

tests/pytests/test_asm.py
```diff
@@ -772,17 +772,32 @@ def info_modules_to_dict(conn):
...
     for shard in env.getOSSMasterNodesConnectionList():
         shard.execute_command('CONFIG', 'SET', 'search-_max-trim-delay-ms', 2500)
     n_docs = 2**14
     create_and_populate_index(env, 'idx', n_docs)
 
     shard1, shard2 = env.getConnection(1), env.getConnection(2)
-    query = ('FT.AGGREGATE', 'idx', '@n:[1 999999]', 'LOAD', 1, 'n', 'WITHCURSOR')
 
+    query = ('FT.AGGREGATE', 'idx', '@n:[1 999999]', 'LOAD', 1, 'n', 'WITHCURSOR')
+    env.expect('DEBUG', 'MARK-INTERNAL-CLIENT').ok()
```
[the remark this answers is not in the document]
JoanFM: does not fail as tests pass.

src/aggregate/aggregate_exec.c
```diff
@@ -770,6 +772,10 @@ void sendChunk(AREQ *req, RedisModule_Reply *reply, size_t limit) {
     if (sctx && sctx->spec && sctx->spec->scan_failed_OOM) {
       ProfileWarnings_Add(&req->profileCtx.warnings, PROFILE_WARNING_TYPE_BG_SCAN_OOM);
     }
+    if (req->stateflags & QEXEC_S_ASM_TRIMMING_DELAY_TIMEOUT) {
+      QueryWarningsGlobalStats_UpdateWarning(QUERY_WARNING_CODE_ASM_INACCURATE_RESULTS, 1, !IsInternal(req));
```
meiravgri, line 770: How do we know we didn't already log this event to the metric counter in sendChunk?

JoanFM: I do not quite get it, this is a different code path. Same as other warnings? These are totally different return paths so why would they be called them together? If the concern is about coord/shard separation, there is the bool coord to differentiate the level.

meiravgri: not completely different paths - this function is also called from `_FT.CURSOR PFOILE` , which might be triggered after the previous cursor read already passed sendChunck path.
Unlike query oom which is triggered first thing first, and the code goes directly here.

Also for sendChunk itself, if we have multiple cursor reads, is it correct to log the same event to the metric counter each time?

src/aggregate/aggregate_exec.c
```diff
@@ -763,13 +765,18 @@ void sendChunk(AREQ *req, RedisModule_Reply *reply, size_t limit) {
...
       ProfileWarnings_Add(&req->profileCtx.warnings, PROFILE_WARNING_TYPE_QUERY_OOM);
     }
-    RedisModule_Reply_ArrayEnd(reply);
 
     // Add BG_SCAN_OOM warning to profile context if applicable
     RedisSearchCtx *sctx = AREQ_SearchCtx(req);
     if (sctx && sctx->spec && sctx->spec->scan_failed_OOM) {
       ProfileWarnings_Add(&req->profileCtx.warnings, PROFILE_WARNING_TYPE_BG_SCAN_OOM);
     }
+    if (req->stateflags & QEXEC_S_ASM_TRIMMING_DELAY_TIMEOUT) {
```
meiravgri, line 769: Consider placing it before checking `scan_failed_OOM` so it will not break the order of populating the warnings array

src/aggregate/aggregate_exec.c
JoanFM: to me it is ok to log for every ft cursor read one warning event yes.

src/aggregate/aggregate_exec.c
JoanFM: done

src/aggregate/aggregate_exec.c
meiravgri: ok if it makes sense...

PR #7918, opened 2025-12-29
MOD-13134: Save and load disk-related index data to RDB

raz-mon:
Adds support for disk API to save and load the disk-related index state to/from the RDB.
Adds two debug commands to retrieve the max-doc_id and dump the deleted-ids set.
In a future PR we will add the distinction between SST-based persistence operations to "regular" ones, such that the value of  `load_from_sst` will change accordingly.

---

> [!NOTE]
> Implements RDB persistence for disk-backed indexes and exposes doc-id introspection/debugging.
> 
> - Save/load disk index state via `SearchDisk_IndexSpecRdbSave/Load`; bump `INDEX_CURRENT_VERSION` to `26` and introduce `INDEX_DISK_VERSION`.
> - During spec create/load, open disk index when applicable, load its RDB section (even for duplicates), and surface failures with new `QUERY_ERROR_CODE_DISK_CREATION`.
> - Add `DocTable_GetMaxDocId` and disk wrappers: `SearchDisk_GetMaxDocId`, `Get/CountDeletedIds`, and `SearchDisk_DeleteDocument`; extend disk API vtables accordingly (`indexSpecRdbSave/Load`, `deleteDocument`, getters).
> - Route `IndexSpec_DeleteDoc` to disk deletion when `diskSpec` is present; maintain existing RAM path otherwise.
> - New debug commands: `FT.DEBUG GET_MAX_DOC_ID` and `FT.DEBUG DUMP_DELETED_IDS`; add tests validating outputs and help list.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 44eef035d8559c4d9d9f6a41f303c62ed962d750. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/query_error/src/lib.rs
+++ b/src/redisearch_rs/query_error/src/lib.rs
@@ -75,6 +75,7 @@ pub enum QueryErrorCode {
     FlexLimitNumberOfIndexes,
     FlexUnsupportedField,
     FlexUnsupportedFTCreateArgument,
+    DiskCreation,
 }
 
 impl Debug for QueryErrorCode {
@@ -159,6 +160,7 @@ impl QueryErrorCode {
             Self::FlexUnsupportedFTCreateArgument => {
                 c"Unsupported FT.CREATE argument for Flex index"
             }
+            Self::DiskCreation => c"Could not create disk index",
         }
     }
 }


src/spec.c
```diff
@@ -3059,6 +3064,24 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
...
+  }
+
+  bool load_from_sst = true; // Fix according to the below
+  if (encver >= INDEX_DISK_VERSION && isSpecOnDisk(sp) && load_from_sst) {
+    // TODO: Load the disk-related data only in case the `REDISMODULE_CTX_FLAGS_SST_RDB`
+    // context flag is set, as we wrote this data only in that case.
+    if (SearchDisk_IndexSpecRdbLoad(rdb, sp->diskSpec) != REDISMODULE_OK) {
+      goto cleanup;
+    }
+  }
```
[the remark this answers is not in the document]
raz-mon: I don't think there is a problem here. Look deeper and let me know.

src/debug_commands.c
```diff
@@ -2155,6 +2156,77 @@ DEBUG_COMMAND(VecSimMockTimeout) {
...
+
+  if (sctx->spec->diskSpec) {
+    // Disk-based index
+    uint64_t count = SearchDisk_GetDeletedIdsCount(sctx->spec->diskSpec);
+
+    if (count == 0) {
+      RedisModule_ReplyWithEmptyArray(sctx->redisCtx);
+    } else {
+      t_docId *buffer = rm_malloc(count * sizeof(t_docId));
+      if (!buffer) {
```
JoanFM, line 2199: I believe we never check Out of Memory in this way. Somehow rm_malloc I believe would just panic, so no point on having these. 

I personally would add this everywhere, but not the pattern

src/spec.c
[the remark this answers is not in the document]
JoanFM: I do not think, but still think is better to have the same API used to be more coherent.

src/spec.c
```diff
@@ -3503,22 +3519,31 @@ void IndexSpec_DeleteDoc_Unsafe(IndexSpec *spec, RedisModuleCtx *ctx, RedisModul
...
-  RedisSearchCtx_UnlockSpec(&sctx);
+  if (sctx.spec->diskSpec) {
+    IndexSpec_IncrActiveWrites(spec);
+    size_t len;
+    const char *keyStr = RedisModule_StringPtrLen(key, &len);
+    SearchDisk_DeleteDocument(sctx.spec->diskSpec, keyStr, len);
+    IndexSpec_DecrActiveWrites(spec);
+  } else {
+    // TODO: is this necessary?
+    RedisSearchCtx_LockSpecRead(&sctx);
```
JoanFM, line 3680: are we sure we do not need to have the Lock in DiskSpec?

src/spec.c
```diff
@@ -3503,22 +3519,31 @@ void IndexSpec_DeleteDoc_Unsafe(IndexSpec *spec, RedisModuleCtx *ctx, RedisModul
...
-  // TODO: is this necessary?
-  RedisSearchCtx_LockSpecRead(&sctx);
-  // Get the doc ID
-  t_docId id = DocTable_GetIdR(&spec->docs, key);
-  RedisSearchCtx_UnlockSpec(&sctx);
+  if (sctx.spec->diskSpec) {
+    IndexSpec_IncrActiveWrites(spec);
+    size_t len;
+    const char *keyStr = RedisModule_StringPtrLen(key, &len);
+    SearchDisk_DeleteDocument(sctx.spec->diskSpec, keyStr, len);
```
JoanFM, line 3675: is the multithreading handled inside the API?

src/spec.h
```diff
@@ -220,7 +220,8 @@ typedef uint16_t FieldSpecDedupeArray[SPEC_MAX_FIELDS];
   (Index_StoreFreqs | Index_StoreFieldFlags | Index_StoreTermOffsets | Index_StoreNumeric | \
    Index_WideSchema)
 
-#define INDEX_CURRENT_VERSION 25
+#define INDEX_CURRENT_VERSION 26
```
JoanFM, line 223: why?

src/search_disk_api.h
```diff
@@ -103,6 +105,44 @@ typedef struct DocTableDiskAPI {
...
 
   bool (*getDocumentMetadata)(RedisSearchDiskIndexSpec* handle, t_docId docId, RSDocumentMetadata* dmd, AllocateKeyCallback allocateKey);
+
+  /**
+   * @brief Gets the maximum document ID assigned in the index
+   *
+   * @param handle Handle to the document table
+   * @return The maximum document ID, or 0 if the index is empty
+   */
+  t_docId (*getMaxDocId)(RedisSearchDiskIndexSpec* handle);
```
JoanFM, line 124: Should we comment that these are only for Debugging? I also think, should we have instead a debug command for each element, just have a debug command that aims to extract all the IN-memory info of the DB? This way we can evolve the disk implementation without changing the API?

raz-mon: Will add a comment, sure 👍 
Regarding unifying the debug commands, I like the separation better since it's more confined to what you actually want so it's easier to deal with, rather than indexing into the reply structure and it exploding in time..
Also, we are not tied to the debug commands' API, i.e., we can change them as we wish

src/spec.h
raz-mon: Since we change the RDB serialization format.
We need to mark that with some version id. This is registered together with the `IndexSpec` type such that upon serialization, this is serialized as well and passed to our deserialization callback.

src/spec.c
raz-mon: Basically we should not be locking the index for this operation (and many others!), since our indexes and doc-table can now be concurrently updated.
Nevertheless, we need to be careful w.r.t order of some of the operations and mainly **consistency**.
I think it's wise to take a good look at multithreading in a dedicated effort, so I can at this point add write-locking for safety, and we can revisit this when the multithreading effort is done.

src/spec.c
raz-mon: Basically yes, but as said above consistency must be thoroughly looked at, so I can add locking for this PR while we revisit and optimize later on.

src/debug_commands.c
raz-mon: Yeah, that's right, but this has big potential to be large so I think it's a proper place to check.
Plus, it's a debug command

src/spec.c
JoanFM: yes I would be conservative now and have a deep look at it in a later stage

src/debug_commands.c
kei-nan: maybe block it with a constant number.
e.g 1000 or something like that.
only dump the first 1000 documents.
I don't know the point in dumping out more then that.

src/search_disk.c
```diff
@@ -64,6 +64,16 @@ void SearchDisk_CloseIndex(RedisSearchDiskIndexSpec *index) {
     disk->basic.closeIndexSpec(index);
 }
 
+void SearchDisk_IndexSpecRdbSave(RedisModuleIO *rdb, RedisSearchDiskIndexSpec *index) {
```
kei-nan, line 67:
```suggestion
void SearchDisk_IndexSpecRdbSave(RedisModuleIO *rdb, const RedisSearchDiskIndexSpec *index) {
```

src/search_disk.c
```diff
@@ -64,6 +64,16 @@ void SearchDisk_CloseIndex(RedisSearchDiskIndexSpec *index) {
...
+  RS_ASSERT(disk && index);
+  disk->basic.indexSpecRdbSave(rdb, index);
+}
+
+int SearchDisk_IndexSpecRdbLoad(RedisModuleIO *rdb, RedisSearchDiskIndexSpec *index) {
```
kei-nan, line 72:
```suggestion
int SearchDisk_IndexSpecRdbLoad(RedisModuleIO *rdb, const RedisSearchDiskIndexSpec *index) {
```

src/search_disk.c
```diff
@@ -95,6 +105,26 @@ bool SearchDisk_DocIdDeleted(RedisSearchDiskIndexSpec *handle, t_docId docId) {
     return disk->docTable.isDocIdDeleted(handle, docId);
 }
 
+t_docId SearchDisk_GetMaxDocId(RedisSearchDiskIndexSpec *handle) {
```
kei-nan, line 108:
```suggestion
t_docId SearchDisk_GetMaxDocId( const RedisSearchDiskIndexSpec *handle) {
```

src/search_disk.c
```diff
@@ -95,6 +105,26 @@ bool SearchDisk_DocIdDeleted(RedisSearchDiskIndexSpec *handle, t_docId docId) {
...
+    RS_ASSERT(disk && handle);
+    return disk->docTable.getMaxDocId(handle);
+}
+
+uint64_t SearchDisk_GetDeletedIdsCount(RedisSearchDiskIndexSpec *handle) {
```
kei-nan, line 113:
```suggestion
uint64_t SearchDisk_GetDeletedIdsCount(const RedisSearchDiskIndexSpec *handle) {
```

src/search_disk.c
```diff
@@ -95,6 +105,26 @@ bool SearchDisk_DocIdDeleted(RedisSearchDiskIndexSpec *handle, t_docId docId) {
...
+    RS_ASSERT(disk && handle);
+    return disk->docTable.getDeletedIdsCount(handle);
+}
+
+size_t SearchDisk_GetDeletedIds(RedisSearchDiskIndexSpec *handle, t_docId *buffer, size_t buffer_size) {
```
kei-nan, line 118:
```suggestion
size_t SearchDisk_GetDeletedIds(const RedisSearchDiskIndexSpec *handle, t_docId *buffer, size_t buffer_size) {
```

src/search_disk.c
```diff
@@ -95,6 +105,26 @@ bool SearchDisk_DocIdDeleted(RedisSearchDiskIndexSpec *handle, t_docId docId) {
...
+    RS_ASSERT(disk && handle);
+    return disk->docTable.getDeletedIds(handle, buffer, buffer_size);
+}
+
+void SearchDisk_DeleteDocument(RedisSearchDiskIndexSpec *handle, const char *key, size_t keyLen) {
```
kei-nan, line 123:
```suggestion
void SearchDisk_DeleteDocument(const RedisSearchDiskIndexSpec *handle, const char *key, size_t keyLen) {
```

src/search_disk_api.h
```diff
@@ -27,9 +27,11 @@ typedef char* (*AllocateKeyCallback)(const void*, size_t len);
 
 typedef struct BasicDiskAPI {
   RedisSearchDisk *(*open)(RedisModuleCtx *ctx, const char *path);
+  void (*close)(RedisSearchDisk *disk);
   RedisSearchDiskIndexSpec *(*openIndexSpec)(RedisSearchDisk *disk, const char *indexName, size_t indexNameLen, DocumentType type);
   void (*closeIndexSpec)(RedisSearchDiskIndexSpec *index);
-  void (*close)(RedisSearchDisk *disk);
+  void (*indexSpecRdbSave)(RedisModuleIO *rdb, RedisSearchDiskIndexSpec *index);
```
kei-nan, line 33:
```suggestion
  void (*indexSpecRdbSave)(RedisModuleIO *rdb, const RedisSearchDiskIndexSpec *index);
```

src/search_disk_api.h
```diff
@@ -27,9 +27,11 @@ typedef char* (*AllocateKeyCallback)(const void*, size_t len);
...
+  u_int32_t (*indexSpecRdbLoad)(RedisModuleIO *rdb, RedisSearchDiskIndexSpec *index);
```
kei-nan, line 34:
```suggestion
  u_int32_t (*indexSpecRdbLoad)(RedisModuleIO *rdb, const RedisSearchDiskIndexSpec *index);
```

src/spec.c
```diff
@@ -1652,7 +1652,9 @@ StrongRef IndexSpec_Parse(const HiddenString *name, const char **argv, int argc,
     size_t len;
     const char* name = HiddenString_GetUnsafe(spec->specName, &len);
     spec->diskSpec = SearchDisk_OpenIndex(name, len, spec->rule->type);
-    RS_LOG_ASSERT(spec->diskSpec, "Failed to open disk spec")
+    if (!spec->diskSpec) {
+      goto failure;
```
kei-nan: maybe set the query error with a error string so we can know it failed.

src/spec.c
kei-nan, line 3208: this was moved here specifically to avoid opening an index with same name multiple times.
this is an edge case I don't think we want to reach.

src/spec.c
```diff
@@ -3232,13 +3229,17 @@ void *IndexSpec_LegacyRdbLoad(RedisModuleIO *rdb, int encver) {
...
-  if (SearchDisk_IsEnabled(NULL)) {
+  if (SearchDisk_IsEnabled()) {
     // TODO: Change to `if (isFlex && !(sp->flags & Index_StoreInRAM)) {` once
     // we add the `Index_StoreInRAM` flag to the rdb file.
     RS_ASSERT(disk_db);
     size_t len;
     const char* name = HiddenString_GetUnsafe(sp->specName, &len);
     sp->diskSpec = SearchDisk_OpenIndex(name, len, sp->rule->type);
+    if (!sp->diskSpec) {
+      StrongRef_Release(spec_ref);
```
kei-nan: maybe set the query error to provide more information to the caller.

src/spec.h
```diff
@@ -220,7 +220,8 @@ typedef uint16_t FieldSpecDedupeArray[SPEC_MAX_FIELDS];
...
+#define INDEX_DISK_VERSION 26
```
kei-nan, line 224: as we discussed, I don't think disk version should be here.

raz-mon: In general it should be, since the serialization format changed and we need to know that upon loading from RDB in a version with disk indexes (i.e., that we are loading an index from a previous version that did not hold disk-index data on the RDB).
The only way it may not be necessary, is if the SST-based saving/loading context flag can tell us if this RDB was saved in an SST-based save operation, which may be ok but I think this is a good way to prevent sticky situations in that regard.

src/search_disk.c
raz-mon: Shouldn't be const here.

src/search_disk_api.h
raz-mon: Not const

src/spec.c
raz-mon: I don't see the code you point at here 🤔 
Anyways, I changed the handling of the duplicate indexes not to open and close the disk db.

src/spec.c
raz-mon: Which query error?
I added a log.

src/debug_commands.c
kei-nan: @raz-mon didn't see a response here, tagging you so you can provide your input

src/spec.c
```diff
@@ -3189,6 +3199,29 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
     }
   }
 
+  // Open the index on disk only if we are on Flex, and this is not a duplicate.
+  if (isSpecOnDisk(sp) && !sp->isDuplicate) {
+    RS_ASSERT(disk_db);
+    size_t len;
+    const char* name = HiddenString_GetUnsafe(sp->specName, &len);
+    sp->diskSpec = SearchDisk_OpenIndex(name, len, sp->rule->type);
+    IndexSpec_PopulateVectorDiskParams(sp);
```
kei-nan, line 3208: what is this?
I don't think this code belongs in this PR

src/spec.c
```diff
@@ -3189,6 +3199,29 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
     }
   }
 
+  // Open the index on disk only if we are on Flex, and this is not a duplicate.
+  if (isSpecOnDisk(sp) && !sp->isDuplicate) {
+    RS_ASSERT(disk_db);
+    size_t len;
+    const char* name = HiddenString_GetUnsafe(sp->specName, &len);
+    sp->diskSpec = SearchDisk_OpenIndex(name, len, sp->rule->type);
```
kei-nan, line 3207: I am really really concerned about opening another pointer to the same underlying db.
This is an opening to a world of troubles.
If you are insisting on going down this route then lets bring it up for discussion internally.

src/spec.c
```diff
@@ -3199,43 +3232,33 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
...
 
   StrongRef spec_ref = sp->own_ref;
 
+  Cursors_initSpec(sp);
+
   // setting isDuplicate to true will make sure index will not be removed from aliases container.
-  const RefManager *oldSpec = dictFetchValue(specDict_g, sp->specName);
-  sp->isDuplicate = oldSpec != NULL;
+  // It may have already been set.
+  if (!sp->isDuplicate && dictFetchValue(specDict_g, sp->specName) != NULL) {
```
kei-nan, line 3246: where is it already being set?

src/spec.c
```diff
@@ -3368,6 +3380,23 @@ void *IndexSpec_LegacyRdbLoad(RedisModuleIO *rdb, int encver) {
...
+    // We do not call `SearchDisk_IndexSpecRdbLoad` since there cannot be disk-related
+    // data in this version of RDB (encver).
+    if (!sp->diskSpec) {
+      RedisModule_LogIOError(rdb, "warning",
+        "Could not open disk index");
+        StrongRef_Release(spec_ref);
+        return NULL;
+      }
+    // Populate diskParams for vector fields now that diskSpec is available
+    IndexSpec_PopulateVectorDiskParams(sp);
```
kei-nan, line 3397: I think this code should be in a different PR, it is out of scope.

src/spec.c
raz-mon: In `IndexSpec_RdbLoad`, we now set it early on there.

src/spec.c
raz-mon: I don't understand your comment.
We are NOT opening the db if this is a duplicate.

raz-mon:
@kei-nan I don't think you understood what we're doing in the duplicate case, at least that's how it sounds from your comments.

src/spec.c
raz-mon: It was moved from another place, please see the full diff.

src/spec.c
raz-mon: It was moved from another place, please see the full diff.

src/debug_commands.c
raz-mon: >I don't know the point in dumping out more then that.

The point would be to check if an id is there.
If you block it you may cut out important information.
Since this is a debug command I would like it to be as informative as it can, and if we see it blows up when used in practice we can block it or add an argument for that.

src/search_disk.c
```diff
@@ -95,6 +105,26 @@ bool SearchDisk_DocIdDeleted(RedisSearchDiskIndexSpec *handle, t_docId docId) {
...
+
+uint64_t SearchDisk_GetDeletedIdsCount(RedisSearchDiskIndexSpec *handle) {
+    RS_ASSERT(disk && handle);
+    return disk->docTable.getDeletedIdsCount(handle);
+}
+
+size_t SearchDisk_GetDeletedIds(RedisSearchDiskIndexSpec *handle, t_docId *buffer, size_t buffer_size) {
+    RS_ASSERT(disk && handle);
+    return disk->docTable.getDeletedIds(handle, buffer, buffer_size);
+}
```
chesedo, line 121: Are both really needed? Can't `buffer_size` be used for the count?

raz-mon: We deduce `buffer_size` from the call to `SearchDisk_GetDeletedIdsCount`, to maintain allocations/deallocations on the C side.

chesedo: I meant can it be made an output argument instead (like `buffer`) so that the count path is not needed?

raz-mon: But then we need to do the allocations on Rust and deallocations on C, I preferred doing them on C thus the API.
Otherwise I need to pass ownership, didn't want to get into that but if you think it's easy and will help here I'm open to it.

src/debug_commands.c
kei-nan: maybe instead we need a debug command to check if a specific id/group of ids exist inside the deleted ids?
It would simplify the logic and usage of this code/function.

src/search_disk.c
kei-nan: Looks like you plan on using it to check if a specific id is deleted, if so then maybe instead of getting all deleted ids you can instead get as an argument ids and check which ones are deleted.

src/spec.h
kei-nan: okay,
but future changes in disk rdb serialization will be internal to disk or will they require incrementing this version again?

src/debug_commands.c
raz-mon: I think this is good, allows maximal flexibility.
Can be extended in the future if needed/wanted.

src/search_disk.c
raz-mon: I prefer printing all of them, as it extends the usage - at a given point in time, these are all the deleted-ids.
In your suggestion you can't get a point-in-time snapshot of the deleted-ids.
We can add more debug APIs if needed/wanted in the future.

src/spec.h
raz-mon: Internally in disk.

src/search_disk.c
raz-mon: > I meant can it be made an output argument instead (like buffer) so that the count path is not needed?

but then how do you know how much to allocate for the buffer @chesedo?
Again this is only to avoid passing ownership between Rust and C.. If that's easy and safe I'm open for it, but in general assume it's not the recommended.

chesedo: Ohh, I did not see about the C buffer allocation. This is the best option then.

src/redisearch_rs/query_error/src/lib.rs
```diff
@@ -75,6 +75,7 @@ pub enum QueryErrorCode {
     FlexLimitNumberOfIndexes,
     FlexUnsupportedField,
     FlexUnsupportedFTCreateArgument,
+    DiskCreation,
```
oshadmi, line 78: Nitpick: consider renaming later to something like `DiskIndexCreation` (matching the actual error "Could not create disk index")

src/spec.c
```diff
@@ -3189,6 +3199,29 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
...
 
+  // Open the index on disk only if we are on Flex, and this is not a duplicate.
+  if (isSpecOnDisk(sp) && !sp->isDuplicate) {
+    RS_ASSERT(disk_db);
+    size_t len;
+    const char* name = HiddenString_GetUnsafe(sp->specName, &len);
+    sp->diskSpec = SearchDisk_OpenIndex(name, len, sp->rule->type);
+    IndexSpec_PopulateVectorDiskParams(sp);
+    if (!sp->diskSpec) {
+      goto cleanup;
```
oshadmi, line 3210: In `IndexSpec_LegacyRdbLoad` a warning is logged. Should also be logged here?

PR #7920, opened 2025-12-29
[MOD-13213] Limit the FT CREATE allowed args for Search on Disk

JoanFM:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

---

> [!NOTE]
> Imposes Flex-mode validation on `FT.CREATE` and adds explicit error reporting.
> 
> - Adds `QUERY_ERROR_CODE_FLEX_UNSUPPORTED_FT_CREATE_ARGUMENT` (C/Rust) with message mapping
> - In `spec.c`, when `SearchDisk_IsEnabledForValidation()` is true:
>   - Parses only a limited set of FT.CREATE args (`ON`, `PREFIX`, `FILTER`, `LANGUAGE[_FIELD]`, `SCORE[_FIELD]`, `STOPWORDS`)
>   - Rejects other args (e.g., `NOOFFSETS`, `NOHL`, `NOFIELDS`, `NOFREQS`, `ASYNC`, `SKIPINITIALSCAN`, `MAXTEXTFIELDS`) with the new error
>   - Enforces `ON HASH` only; `ON JSON` errors with a specific message
>   - Provides `handleBadArguments` to distinguish unknown vs. Flex-unsupported args
> - Corrects `SearchDisk_IsEnabled(NULL)` to `SearchDisk_IsEnabled()` in RDB load path
> - Adds pytest coverage for valid TEXT schemas, supported/unsupported Flex args, `ON JSON` rejection, and default `ON HASH` behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c1c019295d86b745afe15d840ce225830a0c6298. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -1561,6 +1561,26 @@ inline static bool isSpecOnDiskForValidation(const IndexSpec *sp) {
   return SearchDisk_IsEnabledForValidation();
 }
 
+void handleBadArguments(IndexSpec *spec, const char *badarg, QueryError *status, ACArgSpec *non_flex_argopts) {
+  if (isSpecOnDiskForValidation(spec)) {
+    bool isKnownArg = false;
+    for (int i = 0; non_flex_argopts[i].name; i++) {
+      if (strcasecmp(badarg, non_flex_argopts[i].name) == 0) {
+        isKnownArg = true;
+        break;
+      }
+    }
+    if (isKnownArg) {
+      QueryError_SetWithUserDataFmt(status, QUERY_ERROR_CODE_FLEX_UNSUPPORTED_FT_CREATE_ARGUMENT,
+        "Unsupported argument for Flex index:", " `%s`", badarg);
+    } else {
+      QueryError_SetWithUserDataFmt(status, QUERY_ERROR_CODE_PARSE_ARGS, "Unknown argument", " `%s`", badarg);
+    }
+  } else {
+    QueryError_SetWithUserDataFmt(status, QUERY_ERROR_CODE_PARSE_ARGS, "Unknown argument", " `%s`", badarg);
+  }
+}
+
 /* The format currently is FT.CREATE {index} [NOOFFSETS] [NOFIELDS]
     SCHEMA {field} [TEXT [WEIGHT {weight}]] | [NUMERIC]
   */
@@ -1656,23 +1676,7 @@ StrongRef IndexSpec_Parse(const HiddenString *name, const char **argv, int argc,
   if (!AC_AdvanceIfMatch(&ac, SPEC_SCHEMA_STR)) {
     if (AC_NumRemaining(&ac)) {
       const char *badarg = AC_GetStringNC(&ac, NULL);
-      if (isSpecOnDiskForValidation(spec)) {
-        bool isKnownArg = false;
-        for (int i = 0; non_flex_argopts[i].name; i++) {
-          if (strcasecmp(badarg, non_flex_argopts[i].name) == 0) {
-            isKnownArg = true;
-            break;
-          }
-        }
-        if (isKnownArg) {
-          QueryError_SetWithUserDataFmt(status, QUERY_ERROR_CODE_FLEX_UNSUPPORTED_FT_CREATE_ARGUMENT,
-            "Unsupported argument for Flex index:", " `%s`", badarg);
-        } else {
-          QueryError_SetWithUserDataFmt(status, QUERY_ERROR_CODE_PARSE_ARGS, "Unknown argument", " `%s`", badarg);
-        }
-      } else {
-        QueryError_SetWithUserDataFmt(status, QUERY_ERROR_CODE_PARSE_ARGS, "Unknown argument", " `%s`", badarg);
-      }
+      handleBadArguments(spec, badarg, status, non_flex_argopts);
     } else {
       QueryError_SetError(status, QUERY_ERROR_CODE_PARSE_ARGS, "No schema found");
     }


src/spec.c
```diff
@@ -1577,10 +1580,30 @@ StrongRef IndexSpec_Parse(const HiddenString *name, const char **argv, int argc,
   size_t dummy2;
   SchemaRuleArgs rule_args = {0};
   ArgsCursor rule_prefixes = {0};
-
-  ACArgSpec argopts[] = {
+  int rc = AC_OK;
+  ACArgSpec *errarg = NULL;
+  if (isSpecOnDiskForValidation(spec)) {
```
kei-nan: maybe it will be cleaner to have a function which returns an arrayof(ACArgSpec) and internalize the logic of which keywords we should parse.

src/spec.c
```diff
@@ -1577,10 +1580,30 @@ StrongRef IndexSpec_Parse(const HiddenString *name, const char **argv, int argc,
...
+      {.name = "FILTER", .target = &rule_args.filter_exp_str, .len = &dummy2, .type = AC_ARGTYPE_STRING},
+      {.name = "LANGUAGE", .target = &rule_args.lang_default, .len = &dummy2, .type = AC_ARGTYPE_STRING},
+      {.name = "LANGUAGE_FIELD", .target = &rule_args.lang_field, .len = &dummy2, .type = AC_ARGTYPE_STRING},
+      {.name = "SCORE", .target = &rule_args.score_default, .len = &dummy2, .type = AC_ARGTYPE_STRING},
+      {.name = "SCORE_FIELD", .target = &rule_args.score_field, .len = &dummy2, .type = AC_ARGTYPE_STRING},
+      {.name = SPEC_STOPWORDS_STR, .target = &acStopwords, .type = AC_ARGTYPE_SUBARGS},
+      {.name = NULL}
+    };
+    rc = AC_ParseArgSpec(&ac, argopts, &errarg);
+    if (strcasecmp(rule_args.type, RULE_TYPE_HASH) != 0) {
```
kei-nan: what happens if no "ON" is provided?

kei-nan: also you aren't checking the rc before calling this code which can hide other errors.

src/spec.c
```diff
@@ -1635,7 +1658,12 @@ StrongRef IndexSpec_Parse(const HiddenString *name, const char **argv, int argc,
   if (!AC_AdvanceIfMatch(&ac, SPEC_SCHEMA_STR)) {
     if (AC_NumRemaining(&ac)) {
       const char *badarg = AC_GetStringNC(&ac, NULL);
-      QueryError_SetWithUserDataFmt(status, QUERY_ERROR_CODE_PARSE_ARGS, "Unknown argument", " `%s`", badarg);
+      if (isSpecOnDiskForValidation(spec)) {
```
kei-nan: I think we should have a single error here and not the option for two.

src/spec.c
JoanFM: I thought about not losing the capacity to have it in the stack with very little effort. 

Also since the logic of parsing is quite scattered, I think the benefit would not be so crucial.

For instance, Schema is handled in another part

src/spec.c
```diff
@@ -1557,6 +1557,9 @@ inline static bool isSpecOnDisk(const IndexSpec *sp) {
   return SearchDisk_IsEnabled();
 }
 
+inline static bool isSpecOnDiskForValidation(const IndexSpec *sp) {
```
kei-nan, line 1560: I don't think we need another wrapper around this function.
even if we are getting an index spec
don't think we are going to support one index in flex and one not in flex.

src/spec.c
JoanFM: I think is more informative to tell the user that this option is not supported in Flex, could help them migrate in a smoother way. But I can unify do not care much

tests/pytests/test_flex_validation.py
```diff
@@ -35,3 +35,103 @@ def test_invalid_field_type(env):
...
+    env.expect('FT.CREATE', 'idx', 'ON', 'HASH', 'ASYNC', 'SCHEMA', 'field', 'TEXT') \
+        .error().contains('Unsupported argument for Flex index: `ASYNC`')
+
+@skip(cluster=True)
+def test_invalid_on_json(env):
+    """Test that ON JSON fails when search-_simulate-in-flex is true"""
+    # Set the simulate-in-flex configuration to true
+    env.expect('CONFIG', 'SET', 'search-_simulate-in-flex', 'yes').ok()
+
+    env.expect('FT.CREATE', 'idx', 'ON', 'JSON', 'SCHEMA', 'field', 'TEXT') \
```
kei-nan, line 143: why don't we want to support json?

src/spec.c
JoanFM: I think it lays out the ground for this to happen, and if we want to remove this, I think it should be done in another PR. So I would keep with these semantics where we ask the spec and not directly the global struct

tests/pytests/test_flex_validation.py
JoanFM: It is a requirement statet in MVP

src/spec.c
JoanFM: will check

JoanFM:
@kei-nan I acted on reviews.

I keep 2 arrays in the stack so that I can validate later if the bad argument is a totally unknown argument, or is just something not valid in Flex mode

src/spec.c
```diff
@@ -1577,33 +1580,51 @@ StrongRef IndexSpec_Parse(const HiddenString *name, const char **argv, int argc,
...
+    {.name = "NOSCOREIDX", .target = &dummy, .type = AC_ARGTYPE_BOOLFLAG},
+    {.name = "ON", .target = &rule_args.type, .len = &dummy2, .type = AC_ARGTYPE_STRING},
+    SPEC_FOLLOW_HASH_ARGS_DEF(&rule_args)
+    {.name = SPEC_TEMPORARY_STR, .target = &timeout, .type = AC_ARGTYPE_LLONG},
+    {.name = SPEC_STOPWORDS_STR, .target = &acStopwords, .type = AC_ARGTYPE_SUBARGS},
+    {.name = NULL}
+  };
+  ACArgSpec *argopts = isSpecOnDiskForValidation(spec) ? flex_argopts : non_flex_argopts;
+  rc = AC_ParseArgSpec(&ac, argopts, &errarg);
+  invalid_flex_on_type = isSpecOnDiskForValidation(spec) && rule_args.type && (strcasecmp(rule_args.type, RULE_TYPE_HASH) != 0);
```
kei-nan, line 1637:
```suggestion
  invalidFlexOnType = isSpecOnDiskForValidation(spec) && rule_args.type && (strcasecmp(rule_args.type, RULE_TYPE_HASH) != 0);
```

src/spec.c
```diff
@@ -1577,33 +1580,51 @@ StrongRef IndexSpec_Parse(const HiddenString *name, const char **argv, int argc,
...
-      SPEC_FOLLOW_HASH_ARGS_DEF(&rule_args)
-      {.name = SPEC_TEMPORARY_STR, .target = &timeout, .type = AC_ARGTYPE_LLONG},
-      {.name = SPEC_STOPWORDS_STR, .target = &acStopwords, .type = AC_ARGTYPE_SUBARGS},
-      {.name = NULL}};
-
+  int rc = AC_OK;
   ACArgSpec *errarg = NULL;
-  int rc = AC_ParseArgSpec(&ac, argopts, &errarg);
+  bool invalid_flex_on_type = false;
+  ACArgSpec flex_argopts[] = {
```
kei-nan, line 1606: consider moving the code to a different function.

src/spec.c
```diff
@@ -1635,7 +1656,23 @@ StrongRef IndexSpec_Parse(const HiddenString *name, const char **argv, int argc,
   if (!AC_AdvanceIfMatch(&ac, SPEC_SCHEMA_STR)) {
     if (AC_NumRemaining(&ac)) {
       const char *badarg = AC_GetStringNC(&ac, NULL);
-      QueryError_SetWithUserDataFmt(status, QUERY_ERROR_CODE_PARSE_ARGS, "Unknown argument", " `%s`", badarg);
+      if (isSpecOnDiskForValidation(spec)) {
```
kei-nan: I think this code should be in its own function.

src/spec.c
kei-nan: after the changes it looks a bit better since you now check if it is really an unknown arg.

src/spec.c
JoanFM: I think I would need to return around pointers to the different Arg options, I do not think it would make this any cleaner

kei-nan: okay

PR #7943, opened 2025-12-30
MOD-13165 MOD-13166 Add disk-based vector index support infrastructure

ofiryanai:
**Current**: Vector indexes are created only in RAM via VecSimIndex_New
**Change**: Add disk-based vector index creation path through SearchDisk API

---

> [!NOTE]
> Introduces disk-backed vector indexing and plumbs it across parsing, lifecycle, and query paths.
> 
> - Add `VectorDiskAPI` to `search_disk_api.h` and wrappers `SearchDisk_CreateVectorIndex`/`SearchDisk_FreeVectorIndex`
> - Extend `FieldSpec.vectorOpts` with `VecSimHNSWDiskParams` and free `indexName`; in `FieldSpec_Cleanup` free via `SearchDisk` when on disk
> - Populate `diskParams` for HNSW vector fields during `FT.CREATE` (when `sp->diskSpec` exists) and after RDB/legacy loads via `IndexSpec_PopulateVectorDiskParams`
> - Allow `TEXT` and `VECTOR` fields on disk indexes; keep other Flex restrictions and validations
> - Update `openVectorIndex` to create either RAM (`VecSimIndex_New`) or disk-backed (`SearchDisk_CreateVectorIndex`) indexes based on `diskParams.storage`
> - Open `spec->diskSpec` earlier in parse flow so vector disk params can be set
> - Tests: adjust Flex validation (no longer error on `VECTOR`) and add `IS_DISK` expectations in `VECSIM_INFO`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5f24f34ae84a78a4fd47f1defac33316173290bf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/search_disk.c
+++ b/src/search_disk.c
@@ -116,9 +116,7 @@ bool SearchDisk_IsEnabledForValidation() {
 // Vector API wrappers
 void* SearchDisk_CreateVectorIndex(RedisSearchDiskIndexSpec *index, const struct VecSimHNSWDiskParams *params) {
     RS_ASSERT(disk && index && params);
-    if (!disk->vector.createVectorIndex) {
-        return NULL;  // Vector disk API not implemented
-    }
+    RS_ASSERT(disk->vector.createVectorIndex);
     return disk->vector.createVectorIndex(index, params);
 }
 
@@ -127,7 +125,5 @@ void SearchDisk_FreeVectorIndex(void *vecIndex) {
     // Assert that if vecIndex is not NULL, the free function must be set
     // to avoid silent memory leaks from partially implemented API
     RS_ASSERT(!vecIndex || disk->vector.freeVectorIndex);
-    if (vecIndex && disk->vector.freeVectorIndex) {
-        disk->vector.freeVectorIndex(vecIndex);
-    }
+    disk->vector.freeVectorIndex(vecIndex);
 }


src/field_spec.c
```diff
@@ -31,7 +32,11 @@ void FieldSpec_Cleanup(FieldSpec* fs) {
   if (FIELD_IS(fs, INDEXFLD_T_VECTOR)) {
     VecSimParams_Cleanup(&fs->vectorOpts.vecSimParams);
     if (fs->vectorOpts.vecSimIndex) {
-      VecSimIndex_Free(fs->vectorOpts.vecSimIndex);
+      if (fs->vectorOpts.diskParams.storage) {
+        SearchDisk_FreeVectorIndex(fs->vectorOpts.vecSimIndex);
```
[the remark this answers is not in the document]
GuyAv46: Cursor is right.
@ofiryanai, do we still need the field name?
Let's not include it if not, or move its freeing to after we drop the index

src/search_disk.c
```diff
@@ -112,3 +112,19 @@ bool SearchDisk_IsEnabled() {
...
+    RS_ASSERT(disk && index && params);
+    if (!disk->vector.createVectorIndex) {
+        return NULL;  // Vector disk API not implemented
+    }
+    return disk->vector.createVectorIndex(index, params);
+}
+
+void SearchDisk_FreeVectorIndex(void *vecIndex) {
+    RS_ASSERT(disk);
+    if (vecIndex && disk->vector.freeVectorIndex) {
```
GuyAv46: Consider asserting that `!vecIndex || disk->vector.freeVectorIndex` (if `vecIndex` is not `NULL`, the free function is set)

src/spec.c
```diff
@@ -1185,6 +1186,23 @@ static int parseVectorField(IndexSpec *sp, StrongRef sp_ref, FieldSpec *fs, Args
     // Point to the same logCtx as the external wrapping VecSimParams object, which is the owner.
     params->logCtx = logCtx;
     result = parseVectorField_hnsw(fs, params, ac, status);
+    // Build disk params if disk mode is enabled
+    if (result && sp->diskSpec) {
```
GuyAv46, line 1190: Good enough for a start 👍🏼

src/vector_index.c
```diff
@@ -43,14 +44,20 @@ bool isLVQSupported() {
   return false; // In which case we know that LVQ not supported.
 }
 
-VecSimIndex *openVectorIndex(FieldSpec *spec, bool create_if_missing) {
-  RS_ASSERT(FIELD_IS(spec, INDEXFLD_T_VECTOR));
+VecSimIndex *openVectorIndex(FieldSpec *fieldSpec, bool create_if_missing) {
```
GuyAv46, line 47: We can also pass the `IndexSpec` to this function if we need to test the "isDisk" flag

GuyAv46 (commented):
Nice 👍🏼

src/vector_index.c
ofiryanai: I actually tried to avoid that to keep `openVectorIndex` clean, that's why I copied the flag into fieldSpec instead of adding a parameter

src/field_spec.c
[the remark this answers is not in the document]
ofiryanai: it's used for the column family name (I've changed it just now from `{index_name}:{field_name}:vector_index` to `{index_name}:{field_name}`)
I'll check if theres a leak / more correct free order

src/spec.c
```diff
@@ -1185,6 +1186,23 @@ static int parseVectorField(IndexSpec *sp, StrongRef sp_ref, FieldSpec *fs, Args
...
+        .metric = hnsw->metric,
+        .M = hnsw->M,
+        .efConstruction = hnsw->efConstruction,
+        .efRuntime = hnsw->efRuntime,
+        .blockSize = hnsw->blockSize,
+        .multi = hnsw->multi,
+        .storage = sp->diskSpec,
+        .logCtx = logCtx,
+      };
+      fs->vectorOpts.diskParams.indexName = HiddenString_GetUnsafe(fs->fieldName, &fs->vectorOpts.diskParams.indexNameLen);
```
alonre24: Who is the owner of this string? Let's document it

ofiryanai: It was copied instantly in creation flow in C++ and not used anymore, and owned by redisearch. but I added copy because it was smelly

src/search_disk.c
```diff
@@ -112,3 +112,19 @@ bool SearchDisk_IsEnabled() {
...
+        disk->vector.freeVectorIndex(vecIndex);
+    }
```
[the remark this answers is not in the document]
ofiryanai: added assertion

src/search_disk.c
ofiryanai: added

src/spec.c
```diff
@@ -1271,7 +1292,6 @@ static int parseFieldSpec(ArgsCursor *ac, IndexSpec *sp, StrongRef sp_ref, Field
     if (!SearchDisk_MarkUnsupportedFieldIfDiskEnabled(SPEC_GEOMETRY_STR, fs, status)) goto error;
     if (!parseGeometryField(sp, fs, ac, status)) goto error;
   } else if (AC_AdvanceIfMatch(ac, SPEC_VECTOR_STR)) {  // vector field
-    if (!SearchDisk_MarkUnsupportedFieldIfDiskEnabled(SPEC_VECTOR_STR, fs, status)) goto error;
```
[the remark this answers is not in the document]
ofiryanai: I'm not interested in allowing `FT.ALTER` in this PR

src/vector_index.c
GuyAv46: I only recently removed the `spec` from this function (see it's still there for numeric fields). If we need a general `spec` info, I think it's considered clean to add this parameter, especially if the alternative is to copy this info into the field spec

src/vector_index.c
```diff
@@ -43,14 +44,20 @@ bool isLVQSupported() {
...
+      // Disk path - create disk-based HNSW index
+      fieldSpec->vectorOpts.vecSimIndex = SearchDisk_CreateVectorIndex(
+        fieldSpec->vectorOpts.diskParams.storage, &fieldSpec->vectorOpts.diskParams);
+    } else {
+      // RAM path - use standard VectorSimilarity
+      fieldSpec->vectorOpts.vecSimIndex = VecSimIndex_New(&fieldSpec->vectorOpts.vecSimParams);
+    }
   }
-  return spec->vectorOpts.vecSimIndex;
+  return fieldSpec->vectorOpts.vecSimIndex;
```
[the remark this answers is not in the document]
ofiryanai: will be handled in another ticket

src/search_disk.c
```diff
@@ -112,3 +112,22 @@ bool SearchDisk_IsEnabled() {
 bool SearchDisk_IsEnabledForValidation() {
   return isFlex || RSGlobalConfig.simulateInFlex;
 }
+
+// Vector API wrappers
+void* SearchDisk_CreateVectorIndex(RedisSearchDiskIndexSpec *index, const struct VecSimHNSWDiskParams *params) {
+    RS_ASSERT(disk && index && params);
+    if (!disk->vector.createVectorIndex) {
+        return NULL;  // Vector disk API not implemented
+    }
```
raz-mon: We should not call the API in this case, consider removing or at least asserting.

src/search_disk.c
```diff
@@ -112,3 +112,22 @@ bool SearchDisk_IsEnabled() {
...
+    return disk->vector.createVectorIndex(index, params);
+}
+
+void SearchDisk_FreeVectorIndex(void *vecIndex) {
+    RS_ASSERT(disk);
+    // Assert that if vecIndex is not NULL, the free function must be set
+    // to avoid silent memory leaks from partially implemented API
+    RS_ASSERT(!vecIndex || disk->vector.freeVectorIndex);
+    if (vecIndex && disk->vector.freeVectorIndex) {
```
raz-mon: Isn't this redundant? Consider leaving the assertion only

src/search_disk_api.h
```diff
@@ -105,10 +105,35 @@ typedef struct DocTableDiskAPI {
...
+   * @brief Creates a disk-based vector index.
+   *
+   * The returned handle is a VecSimIndex* that can be used with all standard
+   * VecSimIndex_* functions (AddVector, TopKQuery, etc.) due to polymorphism.
+   *
+   * @param index Pointer to the index spec (provides storage context)
+   * @param params Vector index parameters
+   * @return VecSimIndex* handle, or NULL on error
+   */
+  void* (*createVectorIndex)(RedisSearchDiskIndexSpec* index, const struct VecSimHNSWDiskParams* params);
```
raz-mon, line 122: We should align the APIs for text and vector, currently it's quite awkward, e.g., we open the text CFs in database opening and later add the vector.
Can be done in a later PR since it involves changing the text side as well.

ofiryanai: I think they'll be changed anyways when we will access speedb via redismodules api. for now we got to start working somehow so I hope that's good enough. The APIs will probably stay separate, as vector indexes from search indexes in OSS today, but we'll see

src/search_disk.c
ofiryanai: removed

src/search_disk.c
ofiryanai: aligned with other behavior, replaced with assert 👍

src/search_disk_api.h
raz-mon: I agree they should be separate, we should also separate the text CF creation from the index opening.
Not in this PR, I opened a ticket for it.

PR #7971, opened 2026-01-04
MOD-13332: Avoid Using Negative Key Position Values During Command Registration

kei-nan:
Avoid using negative values in command registration for key step and positions.
Override the positions in the ramp process.

Related to #7978 (same fix already done specifically in 8.4 branch)

---

> [!NOTE]
> Simplifies command registration by removing negative key positions and delegating key-spec/dispatch policy to the proxy via RAMP.
> 
> - Register most coordinator commands (read/write/cluster/deprecated) with `DEFINE_COMMAND_KEYS(0, 0, 0)` instead of negative key positions
> - Update `FT.CURSOR` and its subcommands to use no key-spec; clarify that key-spec comes from RAMP
> - Expand `pack/ramp-enterprise.yml` `overide_command` with detailed step policy docs, add `"module"` flags, include `FT.HYBRID`, adjust `FT.CURSOR` `command_arity` and step values
> - Consolidate command setups (e.g., `FT.SEARCH`, `FT.AGGREGATE`, `FT.PROFILE`, `FT.SPELLCHECK`, `FT.INFO`, `FT.HYBRID`, deprecated/tag commands) to use `noKeyArgs`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d9ae24252896557284748c9aa07cccabbc0b63c0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pack/ramp-enterprise.yml
+++ b/pack/ramp-enterprise.yml
@@ -61,15 +61,12 @@ overide_command:
     # -1 : requesting::random_shard
     # -2 : requesting::random_shard, replying::retrieve_cursor
     # -3 : requesting::by_cursor, replying::retrieve_cursor
+    # We only specify a custom policy if the command could cause a bottleneck and should be spread across the cluster
+    # If a policy is not specified the same shard would be preferred
     # Command Arity: required by the proxy in order to validate commands before sending them to the module
-    # It is mandatory because during the parsing of the module.json file 
+    # It is mandatory because during the parsing of the module.json file
     # it is used in a python script as a direct key and get with default value is not used
-    - {"command_arity": -1, "command_name": "FT.AGGREGATE", "first_key": 0, "flags": ["readonly" ], "last_key": 1, "step": -2}
-    - {"command_arity": -2, "command_name": "FT.CURSOR", "first_key": 3, "flags": ["readonly"], "last_key": 1, "step": -3}
-    - {"command_arity": -1, "command_name": "FT.HYBRID", "first_key": 0, "flags": ["readonly"], "last_key": 1, "step": -2}
-    - {"command_arity": -1, "command_name": "FT.INFO", "first_key": 0, "flags": ["readonly"], "last_key": 0, "step": -1}
-    - {"command_arity": -1, "command_name": "FT.MGET", "first_key": 0, "flags": ["readonly"], "last_key": 0, "step": -1}
-    - {"command_arity": -1, "command_name": "FT.PROFILE", "first_key": 0, "flags": ["readonly"], "last_key": 0, "step": -1}
-    - {"command_arity": -1, "command_name": "FT.SEARCH", "first_key": 0, "flags": ["readonly"], "last_key": 0, "step": -1}
-    - {"command_arity": -1, "command_name": "FT.SPELLCHECK", "first_key": 0, "flags": ["readonly"], "last_key": 0, "step": -1}
-    - {"command_arity": -1, "command_name": "FT.TAGVALS", "first_key": 0, "flags": ["readonly"], "last_key": 0, "step": -1}
+    - {"command_arity": -1, "command_name": "FT.AGGREGATE", "first_key": 0, "flags": ["module", "readonly" ], "last_key": 1, "step": -2}
+    - {"command_arity": -2, "command_name": "FT.CURSOR", "first_key": 3, "flags": ["module", "readonly"], "last_key": 1, "step": -3}
+    - {"command_arity": -1, "command_name": "FT.HYBRID", "first_key": 0, "flags": ["module", "readonly"], "last_key": 1, "step": -1}
+    - {"command_arity": -1, "command_name": "FT.SEARCH", "first_key": 0, "flags": ["module", "readonly"], "last_key": 0, "step": -1}


pack/ramp-enterprise.yml
```diff
@@ -57,6 +57,16 @@ exclude_commands:
     - search.CLUSTERREFRESH
     - search.CLUSTERINFO
 overide_command:
+    # The step field in modules is used to encodes the custom policy's requesting/replying policy:
+    # -1 : requesting::random_shard
+    # -2 : requesting::random_shard, replying::retrieve_cursor
+    # -3 : requesting::by_cursor, replying::retrieve_cursor
     - {"command_arity": -1, "command_name": "FT.AGGREGATE", "first_key": 0, "flags": ["readonly" ], "last_key": 1, "step": -2}
```
GuyAv46: Do we need to specify `"command_arity": -1` on each command? Maybe we should just remove this value?

kei-nan: The override is done in a simplistic way of:
```python
        metadata["commands"][override_index[0]] = override
```
I don't think we can do without the command_arity.

GuyAv46: This is bad on it's own, it should use [`update`](https://docs.python.org/3/library/stdtypes.html#dict.update)

GuyAv46: But if there is no default arity, I guess not setting it is equivalent to setting it to `-1`.
Consider verifying that, and remove it if not needed, to avoid more magic stuff we don't understand in the code base

kei-nan: All I can see if a file getting generated in the end.
How it is parsed is beyond the scope of the RAMP code I have access to.
```python
    packname = output.format(**metadata)
    if packname_file:
        with open(packname_file, 'w') as file:
            file.write(packname)
```
Where should I look to see how default arity is read from the file?

src/module.c
```diff
@@ -4166,30 +4168,19 @@ RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
-    const CommandKeys aggregateKeys = enterpriseCmdArgs;
-    const CommandKeys hybridKeys = enterpriseCmdArgs;
-    const CommandKeys cursorKeys = {3, 1, -3};
-    aggregate.position = aggregateKeys;
-    hybrid.position = hybridKeys;
-    cursor.position = cursorKeys;
-  }
+  // Commands that don't operate on Redis keys use (0, 0, 0).
+  // The proxy gets key-spec from the RAMP file (pack/ramp-enterprise.yml).
+  const CommandKeys noKeyArgs = DEFINE_COMMAND_KEYS(0, 0, 0);
```
oshadmi, line 4173: Command with no keys => 0, 0, 0 - according to https://redis.io/docs/latest/develop/reference/modules/modules-api-ref/

pack/ramp-enterprise.yml
kei-nan: Looked at the documentation, found this:
```md
arity: Number of arguments, including the command name itself. A positive number specifies an exact number of arguments and a negative number specifies a minimum number of arguments, so use -N to say >= N. Redis validates a call before passing it to a module, so this can replace an arity check inside the module command implementation. A value of 0 (or an omitted arity field) is equivalent to -2 if the command has sub commands and -1 otherwise.
```
So I think we have to specify the arity.

kei-nan: maybe we should put -2 for the cursor command

kei-nan: changed cursor to use -2.
it looked like arity is required in the resulting file.

pack/ramp-enterprise.yml
```diff
@@ -57,6 +57,16 @@ exclude_commands:
...
-    - {"command_arity": -1, "command_name": "FT.CURSOR", "first_key": 3, "flags": ["readonly"], "last_key": 1, "step": -3}
+    - {"command_arity": -2, "command_name": "FT.CURSOR", "first_key": 3, "flags": ["readonly"], "last_key": 1, "step": -3}
+    - {"command_arity": -1, "command_name": "FT.HYBRID", "first_key": 0, "flags": ["readonly"], "last_key": 1, "step": -2}
+    - {"command_arity": -1, "command_name": "FT.INFO", "first_key": 0, "flags": ["readonly"], "last_key": 0, "step": -1}
+    - {"command_arity": -1, "command_name": "FT.MGET", "first_key": 0, "flags": ["readonly"], "last_key": 0, "step": -1}
+    - {"command_arity": -1, "command_name": "FT.PROFILE", "first_key": 0, "flags": ["readonly"], "last_key": 0, "step": -1}
     - {"command_arity": -1, "command_name": "FT.SEARCH", "first_key": 0, "flags": ["readonly"], "last_key": 0, "step": -1}
+    - {"command_arity": -1, "command_name": "FT.SPELLCHECK", "first_key": 0, "flags": ["readonly"], "last_key": 0, "step": -1}
+    - {"command_arity": -1, "command_name": "FT.TAGVALS", "first_key": 0, "flags": ["readonly"], "last_key": 0, "step": -1}
```
GuyAv46: Consider commenting on why we have `command_arity` set here.
Also, consider using better values

oshadmi (approved):

GuyAv46 (approved):

pack/ramp-enterprise.yml
```diff
@@ -57,6 +57,16 @@ exclude_commands:
...
+    # -2 : requesting::random_shard, replying::retrieve_cursor
+    # -3 : requesting::by_cursor, replying::retrieve_cursor
+    # We only specify a custom policy if the command could cause a bottleneck and should be spread across the cluster
+    # If a policy is not specified the same shard would be preferred
+    # Command Arity: required by the proxy in order to validate commands before sending them to the module
+    # It is mandatory because during the parsing of the module.json file
+    # it is used in a python script as a direct key and get with default value is not used
+    - {"command_arity": -1, "command_name": "FT.AGGREGATE", "first_key": 0, "flags": ["module", "readonly" ], "last_key": 1, "step": -2}
+    - {"command_arity": -2, "command_name": "FT.CURSOR", "first_key": 3, "flags": ["module", "readonly"], "last_key": 1, "step": -3}
+    - {"command_arity": -1, "command_name": "FT.HYBRID", "first_key": 0, "flags": ["module", "readonly"], "last_key": 1, "step": -1}
```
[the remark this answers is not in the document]
kei-nan: the withcursor we support is internal, right now from the proxy perspective there is no cursor support in hybrid.

Itzikvaknin (approved):

PR #7976, opened 2026-01-05
[MOD-13399] Create static libs so Rust code can create indexes

gdesmott:
My branch porting the II iterators to Rust needs to be able to properly create indexes, spec, etc.
All this code and its deps is currently in C requiring to build them as static libs so the Rust tests can link to it.

I used the current pattern of creating one lib per C module (with some exceptions for example `
search_disk.c` and `search_disk_utils.c` are bundled into the same lib) and moved the files to a sub directory matching the lib name.
This was very cumbersome and a pain when rebasing as `git`/`jj` does not handle conflicts in moved files very well.

So I think we should consider alternatives here:
- Stop moving files to at least reduce conflicts?
- Create bigger bundles in static libs?
- Or even create a single static lib with all the C code so the Rust side can just use that? The tricky part here is to prevent circular dependencies with the C code calling Rust one.

#### Mark if applicable


gdesmott:
Closing this PR has we are going to use https://github.com/RediSearch/RediSearch/pull/8003 instead.

PR #7506, opened 2025-11-25
[MOD-10080] Port Not iterator - Non-optimized

lerman25:
This PR adds a Rust version of the NOT iterator, based on the existing C implementation.
It keeps the same behavior, but makes small adjustments for Rust iterator and EOF semantics.
The small changes are needed because Rust iterators behave a bit differently than the C ones:
- Rust marks the child as EOF right after the last doc, so we adjust the logic to still exclude that last child doc from NOT results.
- We make  skip_to clearly return  Found / NotFound(next) and stop at  max_doc_id.
```

Iterator - Not - Read Empty Child/C    time:   [2.7981 ms 2.8490 ms 2.9007 ms]
Iterator - Not - Read Empty Child/Rust time:   [360.91 µs 365.47 µs 370.64 µs]

Iterator - Not - Read Sparse Child/C    time:   [4.9985 ms 5.0180 ms 5.0387 ms]
Iterator - Not - Read Sparse Child/Rust time:   [1.4887 ms 1.5186 ms 1.5515 ms]

Iterator - Not - Read Dense Child/C    time:   [4.9582 ms 5.0522 ms 5.1539 ms]
Iterator - Not - Read Dense Child/Rust time:   [2.7560 ms 2.8196 ms 2.8876 ms]


Iterator - Not - SkipTo Empty Child/C    time:   [40.948 µs 41.753 µs 42.696 µs]
Iterator - Not - SkipTo Empty Child/Rust time:   [12.361 µs 12.621 µs 12.928 µs]

Iterator - Not - SkipTo Sparse Child/C    time:   [202.51 µs 204.61 µs 207.05 µs]
Iterator - Not - SkipTo Sparse Child/Rust time:   [206.68 µs 207.94 µs 209.37 µs]

Iterator - Not - SkipTo Dense Child/C    time:   [1.2097 ms 1.2869 ms 1.3690 ms]
Iterator - Not - SkipTo Dense Child/Rust time:   [461.83 µs 493.66 µs 529.92 µs]

```

---

> [!NOTE]
> Introduce a Rust `Not` iterator that yields the complement of a child iterator up to `max_doc_id`, with comprehensive tests and new benchmarks (including C non-optimized comparisons).
> 
> - **Iterators (core)**:
[...]

PR #7981, opened 2026-01-05
MOD-13400: support gran timeout in Not Iterator (Rust)

GlenDC:

support gran timeout in Not Iterator (Rust)

In order to easily test it, and ensure it is not
confused with a propagated child timeout,
this commit also introduced the capability to the
Mock iterator to timeout at any index, so well
before an EOF.

Benchmark results for Not iterator in this branch:

```
Iterator - Not - Read Empty Child/Rust
                        time:   [648.05 µs 654.52 µs 665.87 µs]
Iterator - Not - Read Empty Child/C
                        time:   [775.79 µs 782.66 µs 790.06 µs]

Iterator - Not - Read Sparse Child/Rust
                        time:   [1.8177 ms 1.8272 ms 1.8383 ms]
Iterator - Not - Read Sparse Child/C
                        time:   [1.7285 ms 1.7343 ms 1.7406 ms]

Iterator - Not - Read Dense Child/Rust
                        time:   [2.0101 ms 2.0211 ms 2.0394 ms]
Iterator - Not - Read Dense Child/C
                        time:   [1.2150 ms 1.2223 ms 1.2293 ms]

Iterator - Not - SkipTo Empty Child/Rust
                        time:   [19.042 µs 19.095 µs 19.175 µs]
Iterator - Not - SkipTo Empty Child/C
                        time:   [20.288 µs 20.412 µs 20.557 µs]

Iterator - Not - SkipTo Sparse Child/Rust
                        time:   [67.798 µs 67.914 µs 68.034 µs]
Iterator - Not - SkipTo Sparse Child/C
                        time:   [75.605 µs 75.894 µs 76.278 µs]

Iterator - Not - SkipTo Dense Child/Rust
                        time:   [245.17 µs 247.54 µs 250.19 µs]
Iterator - Not - SkipTo Dense Child/C
                        time:   [292.38 µs 295.39 µs 298.41 µs]
```

AFAIK, these are in line with the results
as seen in the original Not iterator port PR: https://github.com/RediSearch/RediSearch/pull/7506

1. MOD-13400

1. Not Iterator (`src/redisearch_rs/rqe_iterators/src/not_iterator.rs`)
2. Mock Iterator (`src/redisearch_rs/rqe_iterators/tests/integration/utils/mock_iterator.rs`)

#### Mark if applicable

- [x] This PR introduces API changes (rust `cfg(test`) only: `Mock` iterator)

---

> [!NOTE]
> Introduces execution timeout handling to `Not` iterator with an amortized check and updates surrounding infra.
> 
> - New `util::TimeoutContext` performs periodic deadline checks; `Not::new` now accepts a `Duration` and uses it to enforce timeouts, setting a `forced_eof` and checking in `read`/`skip_to`; `rewind` clears EOF but timeout deadline persists
> - Renames module `not_iterator` to `not`; updates exports and all references
> - Enhances test `Mock` iterator with per-index delay injection and timeout error configuration to validate timeouts pre-EOF
> - Adds integration tests covering timeout propagation and timeouts triggered by `TimeoutContext`
> - Updates benchmarks to use `not` module and provide a large timeout duration
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit af793a6173b4476939e37695bdee0c3134d1cd5a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rqe_iterators_bencher/src/benchers/not.rs
+++ b/src/redisearch_rs/rqe_iterators_bencher/src/benchers/not.rs
@@ -16,15 +16,18 @@ use rqe_iterators::{RQEIterator, empty::Empty, id_list::SortedIdList, not::Not};
 
 use crate::ffi::{IteratorStatus_ITERATOR_OK, QueryIterator};
 
-static WEIGHT: f64 = 1.0;
-static MAX_DOC_ID: u64 = 1_000_000;
-
 #[derive(Default)]
 pub struct Bencher;
 
 impl Bencher {
     const MEASUREMENT_TIME: Duration = Duration::from_millis(500);
     const WARMUP_TIME: Duration = Duration::from_millis(200);
+
+    const WEIGHT: f64 = 1.0;
+    const MAX_DOC_ID: u64 = 1_000_000;
+
+    /// Duration chosen to be big enough such that it will not be reached.
+    const NOT_ITERATOR_LARGE_TIMEOUT: Duration = Duration::from_secs(300);
 
     fn benchmark_group<'a>(
         &self,
@@ -53,7 +56,14 @@ impl Bencher {
         // Rust implementation
         group.bench_function("Rust", |b| {
             b.iter_batched_ref(
-                || Not::new(Empty, MAX_DOC_ID, 1.0, Duration::from_secs(300)),
+                || {
+                    Not::new(
+                        Empty,
+                        Self::MAX_DOC_ID,
+                        1.0,
+                        Self::NOT_ITERATOR_LARGE_TIMEOUT,
+                    )
+                },
                 |it| {
                     while let Ok(Some(current)) = it.read() {
                         criterion::black_box(current);
@@ -68,7 +78,7 @@ impl Bencher {
             b.iter_batched_ref(
                 || {
                     let child = QueryIterator::new_empty();
-                    QueryIterator::new_not_non_optimized(child, MAX_DOC_ID, WEIGHT)
+                    QueryIterator::new_not_non_optimized(child, Self::MAX_DOC_ID, Self::WEIGHT)
                 },
                 |it| {
                     while it.read() == IteratorStatus_ITERATOR_OK {
@@ -91,12 +101,12 @@ impl Bencher {
             b.iter_batched_ref(
                 || {
                     // Child has 1% of docs (every 100th doc)
-                    let data = (1..MAX_DOC_ID).step_by(100).collect();
+                    let data = (1..Self::MAX_DOC_ID).step_by(100).collect();
                     Not::new(
                         SortedIdList::new(data),
-                        MAX_DOC_ID,
+                        Self::MAX_DOC_ID,
                         1.0,
-                        Duration::from_secs(300),
+                        Self::NOT_ITERATOR_LARGE_TIMEOUT,
                     )
                 },
                 |it| {
@@ -112,9 +122,9 @@ impl Bencher {
         group.bench_function("C", |b| {
             b.iter_batched_ref(
                 || {
-                    let data = (1..MAX_DOC_ID).step_by(100).collect();
+                    let data = (1..Self::MAX_DOC_ID).step_by(100).collect();
                     let child = QueryIterator::new_id_list(data);
-                    QueryIterator::new_not_non_optimized(child, MAX_DOC_ID, WEIGHT)
+                    QueryIterator::new_not_non_optimized(child, Self::MAX_DOC_ID, Self::WEIGHT)
                 },
                 |it| {
                     while it.read() == IteratorStatus_ITERATOR_OK {
@@ -137,12 +147,12 @@ impl Bencher {
             b.iter_batched_ref(
                 || {
                     // Child has 99% of docs (all except every 100th doc)
-                    let data = (1..MAX_DOC_ID).filter(|x| x % 100 != 0).collect();
+                    let data = (1..Self::MAX_DOC_ID).filter(|x| x % 100 != 0).collect();
                     Not::new(
                         SortedIdList::new(data),
-                        MAX_DOC_ID,
+                        Self::MAX_DOC_ID,
                         1.0,
-                        Duration::from_secs(300),
+                        Self::NOT_ITERATOR_LARGE_TIMEOUT,
                     )
                 },
                 |it| {
@@ -158,9 +168,9 @@ impl Bencher {
         group.bench_function("C", |b| {
             b.iter_batched_ref(
                 || {
-                    let data = (1..MAX_DOC_ID).filter(|x| x % 100 != 0).collect();
+                    let data = (1..Self::MAX_DOC_ID).filter(|x| x % 100 != 0).collect();
                     let child = QueryIterator::new_id_list(data);
-                    QueryIterator::new_not_non_optimized(child, MAX_DOC_ID, WEIGHT)
+                    QueryIterator::new_not_non_optimized(child, Self::MAX_DOC_ID, Self::WEIGHT)
                 },
                 |it| {
                     while it.read() == IteratorStatus_ITERATOR_OK {
@@ -182,7 +192,14 @@ impl Bencher {
         // Rust implementation
         group.bench_function("Rust", |b| {
             b.iter_batched_ref(
-                || Not::new(Empty, MAX_DOC_ID, 1.0, Duration::from_secs(300)),
+                || {
+                    Not::new(
+                        Empty,
+                        Self::MAX_DOC_ID,
+                        1.0,
+                        Self::NOT_ITERATOR_LARGE_TIMEOUT,
+                    )
+                },
                 |it| {
                     while let Ok(Some(current)) = it.skip_to(it.last_doc_id() + step) {
                         criterion::black_box(current);
@@ -197,7 +214,7 @@ impl Bencher {
             b.iter_batched_ref(
                 || {
                     let child = QueryIterator::new_empty();
-                    QueryIterator::new_not_non_optimized(child, MAX_DOC_ID, WEIGHT)
+                    QueryIterator::new_not_non_optimized(child, Self::MAX_DOC_ID, Self::WEIGHT)
                 },
                 |it| {
                     while it.skip_to(it.last_doc_id() + step) == IteratorStatus_ITERATOR_OK {
@@ -220,12 +237,12 @@ impl Bencher {
         group.bench_function("Rust", |b| {
             b.iter_batched_ref(
                 || {
-                    let data = (1..MAX_DOC_ID).step_by(100).collect();
+                    let data = (1..Self::MAX_DOC_ID).step_by(100).collect();
                     Not::new(
                         SortedIdList::new(data),
-                        MAX_DOC_ID,
+                        Self::MAX_DOC_ID,
                         1.0,
-                        Duration::from_secs(300),
+                        Self::NOT_ITERATOR_LARGE_TIMEOUT,
                     )
                 },
                 |it| {
@@ -241,9 +258,9 @@ impl Bencher {
         group.bench_function("C", |b| {
             b.iter_batched_ref(
                 || {
-                    let data = (1..MAX_DOC_ID).step_by(100).collect();
+                    let data = (1..Self::MAX_DOC_ID).step_by(100).collect();
                     let child = QueryIterator::new_id_list(data);
-                    QueryIterator::new_not_non_optimized(child, MAX_DOC_ID, WEIGHT)
+                    QueryIterator::new_not_non_optimized(child, Self::MAX_DOC_ID, Self::WEIGHT)
                 },
                 |it| {
                     while it.skip_to(it.last_doc_id() + step) == IteratorStatus_ITERATOR_OK {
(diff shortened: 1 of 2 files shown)


src/redisearch_rs/rqe_iterators/src/not.rs
```diff
@@ -29,20 +30,25 @@ pub struct Not<'index, I> {
     max_doc_id: t_docId,
     /// A reusable result object to avoid allocations on each `read` call.
     result: RSIndexResult<'index>,
-    // TODO: Timeout
+    /// Tracks the execution deadline for this iterator.
+    ///
+    /// Uses an amortized check to minimize overhead in hot paths. The timeout
+    /// is absolute for the iterator's lifetime and does not reset upon rewinding.
+    timeout_ctx: TimeoutContext,
```
gdesmott, line 41: We should re run benchmarks to check how this affect performances.

src/redisearch_rs/rqe_iterators/src/not_iterator.rs
```diff
@@ -64,17 +70,27 @@ where
 
             match self.result.doc_id.cmp(&self.child.last_doc_id()) {
                 Ordering::Less => {
+                    if self.timeout_ctx.check_timeout() {
```
gdesmott: Maybe this function should return a `Result` so we can use `?` here ?

src/redisearch_rs/rqe_iterators/src/not.rs
```diff
@@ -29,20 +30,25 @@ pub struct Not<'index, I> {
...
 {
-    pub const fn new(child: I, max_doc_id: t_docId, weight: f64) -> Self {
+    pub fn new(child: I, max_doc_id: t_docId, weight: f64, timeout: Duration) -> Self {
         Self {
             child: MaybeEmpty::new(child),
             max_doc_id,
             result: RSIndexResult::virt()
                 .weight(weight)
                 .field_mask(RS_FIELDMASK_ALL),
+            timeout_ctx: TimeoutContext::new(timeout, 5_000),
```
gdesmott, line 60: let's use a documented constant here instead of magic number.

gdesmott: And should not this be `100`?


`#define TIMEOUT_COUNTER_LIMIT 100`

gdesmott (commented):
Need to fix the CI.

src/redisearch_rs/rqe_iterators/src/not_iterator.rs
GlenDC: Great suggestion. Thx!

src/redisearch_rs/rqe_iterators/src/not.rs
GlenDC: As discussed offline:

- that limit is just the default in C, it is not actually ever the one used for the not iterator,
  and on top of that we use a different design pattern in Rust, as to avoid mistakes regarding typos.

Will document it though, focussing on the semantics, not the reason why this constant was picked (As that's just the C one).

src/redisearch_rs/rqe_iterators/src/not_iterator.rs
```diff
@@ -83,6 +99,10 @@ where
                             "child read backwards without rewind"
                         );
                     } else {
+                        if self.timeout_ctx.check_timeout() {
```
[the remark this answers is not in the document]
GlenDC: Fair. Fixed it.

GlenDC:
All feedback has been resolved.
Ready for another review @gdesmott .

src/redisearch_rs/rqe_iterators/src/not.rs
gdesmott: The `Read Dense Child` bench seems to be significantly slower than the C version. Is this because of this change?

src/redisearch_rs/rqe_iterators/src/not.rs
gdesmott: I meant have something like that defined at the start of the file:

```rust
/// meaning of this value
const TIMEOUT_LIMIT_CHECK: u32 = 5_000;
```

src/redisearch_rs/rqe_iterators/src/util/timeout.rs
```diff
@@ -0,0 +1,61 @@
...
+            deadline: Instant::now() + duration,
+            counter: 0,
+            limit,
+        }
+    }
+
+    /// Increments the internal counter and, if the `limit` is reached, checks if
+    /// the current time has passed the `deadline`.
+    ///
+    /// Returns `true` if the deadline has been reached or exceeded.
```
gdesmott: doc needs to be updated

GlenDC: Addressed this.

gdesmott (approved):

src/redisearch_rs/rqe_iterators/src/not.rs
GlenDC: As discussed and agreed offline, this falls inline with the original benchmarks of the Not iterator port. If it does turn out an issue we can revisit later.

src/redisearch_rs/rqe_iterators/src/not.rs
GlenDC: For now it is sufficient as-is, as agreed offline. For codebase consistency- or similar reasons we can revisit this later if required.

oshadmi: Consider making the timeout counter a parameter as well, so for `Not` iterator it can be 5000 or some value, and for other iterators it could be 100 (as in C)
allowing to control the frequency per iteration type

src/redisearch_rs/rqe_iterators/src/util/timeout.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// the current time has passed the `deadline`.
+    ///
+    /// Returns error [`RQEIteratorError::TimedOut`] if the deadline has been reached or exceeded.
+    #[inline(always)]
+    pub fn check_timeout(&mut self) -> Result<(), RQEIteratorError> {
+        self.counter += 1;
+        if self.counter >= self.limit {
+            self.counter = 0;
+            if Instant::now() >= self.deadline {
+                return Err(RQEIteratorError::TimedOut);
```
oshadmi, line 54: `check_timeout` is not considering `RSTimeoutPolicy`: If it is `RSTimeoutPolicy_TimeoutPolicy_Fail` we should return a timeout error, but if it is `RSTimeoutPolicy_TimeoutPolicy_Return` we should consider this as EOF and not an error.

lerman25 (commented):
Agree with @oshadmi comment. 
Another question is - 
In the c version, the entire repo uses `timeout.h` file, while here it's associated only with iterators. Planning ahead - will this file move out of this scope or will we create a different timeout file for every scope?

src/redisearch_rs/rqe_iterators_bencher/src/benchers/not.rs
```diff
@@ -252,7 +267,12 @@ impl Bencher {
             b.iter_batched_ref(
                 || {
                     let data = (1..MAX_DOC_ID).filter(|x| x % 100 != 0).collect();
-                    Not::new(SortedIdList::new(data), MAX_DOC_ID, 1.0)
+                    Not::new(
+                        SortedIdList::new(data),
+                        MAX_DOC_ID,
+                        1.0,
+                        Duration::from_secs(300),
```
LukeMathWalker: Worth moving these two out into constants, since we're copy-pasting them everywhere?

GlenDC:
> Agree with @oshadmi comment. Another question is - In the c version, the entire repo uses `timeout.h` file, while here it's associated only with iterators. Planning ahead - will this file move out of this scope or will we create a different timeout file for every scope?

For now it is in the rqe_iterators crate because it's the only use-case. And neither do we have some generic kind of utils crate. Refactoring is Rust's greatest superpower so I would suggest to move it to another crate only once you start needing it for non-rqe-iterator use cases.

That said, if you think of a good location already and want to do it already, do let me know.
CC @LukeMathWalker

src/redisearch_rs/rqe_iterators/src/util/timeout.rs
GlenDC: I requested some guidance by you @oshadmi and @lerman25 offline, as I am missing some context here.

src/redisearch_rs/rqe_iterators/src/util/timeout.rs
```diff
@@ -0,0 +1,60 @@
...
+            }
+        }
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
GlenDC: Yes that is indeed what @oshadmi noted. But as I requested offline as well as in https://github.com/RediSearch/RediSearch/pull/7981#discussion_r2679562964 I will need more context. Because I somehow seem to not understand what I would need to do differently in the NOT (rust) Iterator in order for that policy to be respected. I was so far under the impression that that policy was more about how an iterator user would handle timeout errors?

src/redisearch_rs/rqe_iterators/src/util/timeout.rs
GlenDC: The current code is correct. See my comment below.

GlenDC:
Following some great sync discussions with @lerman25 offline, I wanted to summarize our alignment on how we are handling iterator timeouts. There was a slight initial misalignment regarding how Rust’s `Result` type interacts with our `FAIL` vs. `RETURN` policies, and I think it serves as a great "teachable moment" for our transition from C to Rust. @oshadmi tagging you as well so that you can weight in as well in case you do think there's a design misalignment here. @LukeMathWalker and @gdesmott looping you in as well as it seems that our current iterator design in Rust is not fully understood yet by the Redis team so we might need to address that further over next days and weeks where appropriate.

## Returning Status Vs Returning Result

In the C implementation, we return `ITERATOR_TIMEOUT`. The concern was that returning an `Err(RQEIteratorError::TimedOut)` in Rust might "force" an error path, preventing the coordinator from returning partial results (the `RETURN` policy).

In Rust, a `Result` is simply an **enum**, not an exception or a forced crash (in case of error). Unlike status codes in C which can be accidentally ignored, or Exceptions in other languages which force a jump in control flow, Rust's `Result` packs the outcome into a type-safe container.

* **C approach:** Returns a status code (integer). The caller must remember to check this integer against a global policy and there is no easy way to bubble it up other than to check if it's OK or something else.
* **Rust approach:** Returns `Result<Option<T>, E>`. The type system ensures the caller acknowledges that a timeout (or any other error) *could*  have happened.

## `match` vs. the `?` Operator

A common misconception when starting with Rust is that `Err` must always be "bubbled up" using the `?` operator. While `?` is a convenient shortcut for "exit early on error," it is entirely optional.

For our Result Processors (RP), we don't always want to exit early. We can use pattern matching to handle a timeout as a valid state for partial data retrieval, as for example in this theoretical code:

```rust
// The caller (e.g., RPNet) handles the result based on the policy
match iterator.next() {
    Ok(Some(result)) => {
        // Process next result...
    },
    Ok(None) => {
        // Standard EOF logic
    },
    Err(RQEIteratorError::TimedOut) => {
        if policy == Policy::RETURN {
            // Even though we received an Err, we choose to treat it 
            // as a successful stop and return the data we have so far.
            return_partial_results()
        } else {
            // FAIL policy: now we use the error as an error.
            return_error_to_client(RQEIteratorError::TimedOut)
        }
    }
}

```

## Conclusion

By using `Result`, we maintain the exact same intent as the C version but with better type safety.

1. The iterator signals **why** it stopped (Timeout).
2. The `atEOF` state is implicitly handled (if it's an `Err` or `None`, there's no more data).
3. The caller (the RP) retains full control over whether that timeout is "fatal" or "acceptable" based on the user's query profile.

This approach keeps the iterators clean and idiomatic while ensuring our pipeline logic remains as flexible as the original C implementation.

gdesmott (commented):
Could we add a test checking this timeout/eof change?

gdesmott (approved):

src/redisearch_rs/rqe_iterators_bencher/src/benchers/not.rs
```diff
@@ -53,7 +53,7 @@ impl Bencher {
         // Rust implementation
         group.bench_function("Rust", |b| {
             b.iter_batched_ref(
-                || Not::new(Empty, MAX_DOC_ID, 1.0),
+                || Not::new(Empty, MAX_DOC_ID, 1.0, Duration::from_secs(300)),
```
lerman25: Would it make sense to move this to a shared constant like NOT_ITERATOR_LARGE_TIMEOUT here as well?

lerman25 (commented):
Thanks from the great summary about the timeout return type choice! 
A little nitpick -

src/redisearch_rs/rqe_iterators_bencher/src/benchers/not.rs
GlenDC: Sure, that works for me. I do not have a strong opinion about it, and there are good reasons to prefer that approach, so I used the constant.

While I was at it I also made a few related changes.

* I changed the statics to constants since there was no reason for them to be statics. A static lives in the data section of your program or library and therefore has a static lifetime. A const is simply a named literal. In most cases you want a const, especially for primitive values like durations or other numeric values.

* I reduced the scope of these constants. It does not really matter in this case and you can keep them as module level private constants if you prefer, just as they were when they were statics. That works fine too. But since these private constants only exist to support these specific bench functions, it makes sense to keep them grouped with the type because it clarifies the intent.

lerman25 (approved):

PR #7987, opened 2026-01-07
Make build arguments more flexible

memark:
Allow `build.sh` to accept arguments like `DEBUG=1` (previously only `DEBUG`) to decrease developer mistakes and frustration.

---

> [!NOTE]
> Improves CLI ergonomics and compatibility for `build.sh`.
> 
> - Enables `extglob` and switches boolean flags to `FLAG?(=1)` patterns (e.g., `DEBUG`, `PROFILE`, `RUN_*`, `TESTS`) to accept both `FLAG` and `FLAG=1`
> - Normalizes argument keys to uppercase for parsing (Bash 3.2 compatible), avoiding duplicate upper/lower cases
> - Preserves explicit value parsing (e.g., `COORD=*`, `COV=*`, `EXT=*`, `EXT_PORT=*`) and keeps `EXT_HOST` IPv4 validation
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eae2fcdb033512ccc7e02348a28e4e3bb4fe520d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -56,7 +56,9 @@ NIGHTLY_VERSION=$(cat ${ROOT}/.rust-nightly)
 #-----------------------------------------------------------------------------
 parse_arguments() {
   for arg in "$@"; do
-    case ${arg^^} in
+    # MacOS only has bash 3.2 built-in, which doesn't support the more modern ${arg^^} syntax.
+    upper_arg=$(printf '%s' "$arg" | tr '[:lower:]' '[:upper:]')
+    case $upper_arg in
       COORD=*)
         COORD="${arg#*=}"
         ;;


build.sh
```diff
@@ -51,55 +52,56 @@ NIGHTLY_VERSION=$(cat ${ROOT}/.rust-nightly)
...
-      RUN_RUST_TESTS|run_rust_tests)
+      RUN_RUST_TESTS?(=1))
         RUN_RUST_TESTS=1
         ;;
-      RUN_RUST_VALGRIND|run_rust_valgrind)
+      RUN_RUST_VALGRIND?(=1))
         RUN_RUST_VALGRIND=1
         ;;
-      RUN_MICRO_BENCHMARKS|run_micro_benchmarks|RUN_MICROBENCHMARKS|run_microbenchmarks)
+      RUN_MICRO_BENCHMARKS?(=1))
```
LukeMathWalker, line 86: This also allowed different placing of `_` beyond capitalization. We should keep it.

memark: True. I did a search in the whole repo, and nowhere was the removed version used. But if you think _people_ use it, I can put it back in.

LukeMathWalker: @JoanFM?

build.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -e
+shopt -s extglob
```
GlenDC, line 3: Regarding Luca's proposal you can also add which is eaiser than having to add so much duplicate checks but does have side -effects.

```
shopt -s nocasematch
```

But that would need to be evaluated, should be ok though but do check.

memark: The script actually already handles case insensitivity through
`case ${arg^^} in`
if that's what you mean?

build.sh
JoanFM: I think is not critical, it was a nice to have but nothing big

LukeMathWalker (approved):

build.sh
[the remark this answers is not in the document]
memark: Deemed not necessary.

memark:
I had to switch to older syntax due to old built-in bash version on MacOS (one of the runners).

PR #7988, opened 2026-01-07
Add an option to run coordinator tests as a separate job to increase CI parallelism.

LukeMathWalker:

Coordinator tests take the lion share of CI time. Alone, they take longer than all other testing steps combined.

This PR adds support for running coordinator tests in parallel to all other testing steps, in a separate job.
This should reduce our end-to-end CI latency, thus speeding up PR feedback and merge queue.

We retain a single testing matrix in both the merge queue and the PR flow.

#### Mark if applicable

---

> [!NOTE]
> Improves CI parallelism by optionally splitting coordinator tests from other tests.
> 
> - Adds `separate-coordinator-job` input to workflows and propagates `test-mode` (`standalone`, `coordinator`, `all`) through matrices and job names
> - Updates `generate-matrix.yml` and `generate-basic-matrix.yml` to emit per-mode entries when enabled
> - Adjusts `flow-test.yml` and PR/merge workflows to set `coordinator`, `standalone`, and `unit-tests` based on `test-mode`; enables split in both flows
> - Refines `task-test.yml`: gate MIRI on `unit-tests` and upload Codecov either combined or per-mode (standalone/coordinator)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 504377590fd027a5fc068b78c9069869195a7b01. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/task-test.yml
+++ b/.github/workflows/task-test.yml
@@ -605,8 +605,17 @@ jobs:
       - name: Import Codecov GPG public key
         if: inputs.coverage
         run: gpg --import .github/codecov_gpg.pub
+      - name: Upload flow coverage (combined)
+        if: inputs.coverage && inputs.standalone && inputs.coordinator
+        uses: codecov/codecov-action@v5
+        with:
+          files: bin/flow_standalone.info,bin/flow_coordinator.info
+          disable_search: true
+          flags: flow
+          fail_ci_if_error: true # Fail on upload errors
+          token: ${{ secrets.CODECOV_TOKEN }}
       - name: Upload flow coverage (standalone)
-        if: inputs.coverage && inputs.standalone
+        if: inputs.coverage && inputs.standalone && !inputs.coordinator
         uses: codecov/codecov-action@v5
         with:
           files: bin/flow_standalone.info
@@ -615,13 +624,13 @@ jobs:
           fail_ci_if_error: true # Fail on upload errors
           token: ${{ secrets.CODECOV_TOKEN }}
       - name: Upload flow coverage (coordinator)
-        if: inputs.coverage && inputs.coordinator
+        if: inputs.coverage && !inputs.standalone && inputs.coordinator
         uses: codecov/codecov-action@v5
         with:
           files: bin/flow_coordinator.info
           disable_search: true
           flags: flow
-          fail_ci_if_error: true
+          fail_ci_if_error: true # Fail on upload errors
           token: ${{ secrets.CODECOV_TOKEN }}
       - name: Upload unit coverage
         if: inputs.coverage && inputs.unit-tests
@@ -630,7 +639,7 @@ jobs:
           files: bin/unit.info,bin/rust_cov.info
           disable_search: true
           flags: unit
-          fail_ci_if_error: true
+          fail_ci_if_error: true # Fail on upload errors
           token: ${{ secrets.CODECOV_TOKEN }}
 
   # Stop EC2 runner for self-hosted platforms (always runs if `start-runner` succeeds, to ensure cleanup)


.github/workflows/task-test.yml
```diff
@@ -603,23 +603,32 @@ jobs:
           echo "Coordinator: ${{ steps.coordinator_tests.outcome }}"
           exit 1
 
-      - name: Upload flow coverage
-        if: inputs.coverage
+      - name: Upload flow coverage (standalone)
```
GuyAv46, line 617: Is this required?
1. Codecov recommends having a 1:1 ratio of upload steps and flags
2. I assume the action won't fail if a file is missing, or at least there is a config to not fail on that
3. If 2. is wrong, consider formatting the `files` input according to the tests we ran

.github/workflows/task-test.yml
```diff
@@ -603,23 +603,32 @@ jobs:
...
+          token: ${{ secrets.CODECOV_TOKEN }}
+      - name: Upload flow coverage (coordinator)
+        if: inputs.coverage && inputs.coordinator
         uses: codecov/codecov-action@v5
         with:
-          files: bin/flow_standalone.info,bin/flow_coordinator.info
+          files: bin/flow_coordinator.info
           disable_search: true
           flags: flow
-          fail_ci_if_error: true # Fail on upload errors
```
GuyAv46, line 615: Consider keeping the comment

.github/workflows/event-pull_request.yml
```diff
@@ -41,9 +41,17 @@ jobs:
...
   test-matrix:
-    needs: [check-what-changed, get-latest-redis-tag, generate-basic-matrix, spellcheck]
+    name: ${{ matrix.job-type }}${{ matrix.test-mode == 'standalone' && ' (Standalone)' || (matrix.test-mode == 'coordinator' && ' (Coordinator-only)' || '') }}
+    needs:
+      [
+        check-what-changed,
+        get-latest-redis-tag,
+        generate-basic-matrix,
+        spellcheck,
+      ]
```
GuyAv46:
```suggestion
    needs:
      - check-what-changed
      - get-latest-redis-tag
      - generate-basic-matrix
      - spellcheck
```

.github/workflows/task-test.yml
LukeMathWalker: We are forced to upload flow coverage in two separate batches, since we are running them in separate jobs. We could use more specific flags (i.e. `flow-standalone` and `flow-coordinator`) and then configure them to be unified on the codecov side.

LukeMathWalker: Looking at their docs, though, they seem to discourage specifying multiple flags for a single upload. I don't see anything with respect to merging multiple reports under a single flag (and it does seem to work just fine in this PR).

LukeMathWalker: Specifically looking at https://docs.codecov.com/docs/flags#one-to-one-relationship-of-flags-to-uploads

GuyAv46: Right, two jobs will yield two files.
But when we have a unified run, we can use a single step. I think formatting the file list is cleaner than having two jobs that could be run in a single step, or one will be skipped.

I understand that if a file is missing, the step fails?

LukeMathWalker: We can make it not fail if a file is missing, but then it becomes harder to spot issues if filenames were to change in the future.

I am fine with either approach.

.github/workflows/task-test.yml
```diff
@@ -199,11 +199,11 @@ jobs:
     runs-on: ${{ needs.get-config.outputs.ec2_image_id && needs.start-runner.outputs.label || needs.get-config.outputs.env || vars.RUNS_ON || 'ubuntu-latest' }}
     container: ${{
       needs.get-config.outputs.container
-        && fromJSON(
-          format('{{"image":"{0}","options":"--user root -v /usr/share/dotnet:/host_usr/dotnet -v /usr/local/lib/android:/host_usr/android -v /opt/ghc:/host_opt/ghc -v /opt/hostedtoolcache:/host_opt/hostedtoolcache {1}"}}',
-            needs.build-image.outputs.succeeded == 'true' && needs.build-image.outputs.image || needs.get-config.outputs.container,
-            needs.get-config.outputs.container == 'alpine:3.22' && '-v /opt:/opt:rw,rshared -v /opt:/__e/node20:ro,rshared' || '')
-        )
```
GuyAv46, line 206: The nesting here is for the reader to understand the hierarchy. Why flatten it?

.github/workflows/task-test.yml
LukeMathWalker: I introduced a combined upload step in 504377590. Does that work @GuyAv46?

.github/workflows/task-test.yml
LukeMathWalker: Stray change, reverted.

PR #7996, opened 2026-01-07
MOD-13320: Condition disk-side RDB save/load on context flag

raz-mon:
We condition the disk-side saving and loading API-calls on the `REDISMODULE_CTX_FLAGS_SST_RDB` context flag.
the new flag is temporarily added it to `spec.c` until we use the closed-source module-header file for disk.

Depends on https://github.com/RediSearch/RediSearch/pull/7918.

---

> [!NOTE]
> Conditions disk index serialization in `spec.c` on the `REDISMODULE_CTX_FLAGS_SST_RDB` context flag and aligns mock RDB helpers accordingly.
> 
> - Save: only call `SearchDisk_IndexSpecRdbSave` when `sp->diskSpec` exists and SST-RDB flag is set
> - Load: only call `SearchDisk_IndexSpecRdbLoad` when on-disk index and SST-RDB flag is set; add symmetry comments
> - Tests/mock: switch `redismock` RDB helpers to allocate/free `RedisModuleIO` via `RMCK_CreateRdbIO`/`RMCK_FreeRdbIO`, avoiding stack IO and ensuring context association
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e05f62c534a19a1fe692c1a18730cb8ccccd8781. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/cpptests/redismock/redismock.cpp
+++ b/tests/cpptests/redismock/redismock.cpp
@@ -756,24 +756,25 @@ int RMCK_IsIOError(RedisModuleIO *io) {
 void *RMCK_LoadDataTypeFromStringEncver(const RedisModuleString *str,
                                         const RedisModuleType *mt,
                                         int encver) {
-  RedisModuleIO io{};
-  io.buffer.insert(io.buffer.end(), str->c_str(), str->c_str() + str->size());
-  void *ret = mt->typemeths.rdb_load(&io, encver);
-  RMCK_FreeRdbIO(&io);
+  RedisModuleIO *io = RMCK_CreateRdbIO();
+  io->buffer.insert(io->buffer.end(), str->c_str(), str->c_str() + str->size());
+  void *ret = mt->typemeths.rdb_load(io, encver);
+  RMCK_FreeRdbIO(io);
   return ret;
 }
 
 RedisModuleString *RMCK_SaveDataTypeToString(RedisModuleCtx *ctx,
                                              void *data,
                                              const RedisModuleType *mt) {
-  RedisModuleIO io{};
-  mt->typemeths.rdb_save(&io, data);
-  RMCK_FreeRdbIO(&io);
-  if (io.error_flag) {
+  RedisModuleIO *io = RMCK_CreateRdbIO();
+  mt->typemeths.rdb_save(io, data);
+  if (io->error_flag) {
+    RMCK_FreeRdbIO(io);
     return nullptr;
   }
-  RedisModuleString *rms = new RedisModuleString(std::string(io.buffer.begin(), io.buffer.end()));
+  RedisModuleString *rms = new RedisModuleString(std::string(io->buffer.begin(), io->buffer.end()));
   if (ctx) ctx->addPointer(rms);
+  RMCK_FreeRdbIO(io);
   return rms;
 }
 


src/spec.c
```diff
@@ -3211,12 +3216,17 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
...
+
+  // Load the disk-related index data if we are on disk and the save flow used
+  // sst-files, even if this is a duplicate.
   // In the case of a duplicate, `sp->diskSpec=NULL` thus handled appropriately
-  // On the disk side.
-  if (encver >= INDEX_DISK_VERSION && isSpecOnDisk(sp)) {
-    // TODO: Load the disk-related data only in case the `REDISMODULE_CTX_FLAGS_SST_RDB`
-    // context flag is set, as we wrote this data only in that case.
+  // On the disk side (RDB is depleted, without updating index fields).
+  bool use_sst = RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_SST_RDB;
```
kei-nan: maybe we can add a C++ unit test for this one?
we have mock for rdb, maybe we can add a mock for disk?

kei-nan:
```suggestion
  const bool useSst = RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_SST_RDB;
```

src/spec.c
```diff
@@ -3113,7 +3113,12 @@ void IndexSpec_RdbSave(RedisModuleIO *rdb, IndexSpec *sp) {
   }
 
   // Disk index
-  if (sp->diskSpec) {
+  // Check if we are using SST files with this RDB.
+  // We assume symmetry w.r.t this context flag. I.e., If it is not set, we
+  // assume it was not set in when the RDB will be loaded as well
+  RedisModuleCtx *ctx = RedisModule_GetContextFromIO(rdb);
+  bool use_sst = RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_SST_RDB;
```
kei-nan:
```suggestion
  const bool useSst = RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_SST_RDB;
```

src/spec.c
```diff
@@ -3113,7 +3113,12 @@ void IndexSpec_RdbSave(RedisModuleIO *rdb, IndexSpec *sp) {
   }
 
   // Disk index
-  if (sp->diskSpec) {
+  // Check if we are using SST files with this RDB.
```
kei-nan: I think it is worth adding some documentation on why we use RDB when sst is available and why we don't when it isn't.

src/spec.c
raz-mon: Given we test that the disk data is correctly saved and loaded on the disk side (flow-test), and that the current save and load flows on OSS work fine - would that add anything?
What would you want that to test?

src/spec.c
raz-mon: I don't understand your comment. You probably meant something else, since we always use the RDB.

raz-mon: The question is whether or not we use the SST files.
I can elaborate on that a bit.
Added.

JoanFM (approved):

src/spec.c
oshadmi: Note: Prefered to avoid nested `if` and call `RedisModule_GetContextFlags` also when not `sp->diskSpec` for readability (with hardly any other impact )

src/spec.c
```diff
@@ -3211,12 +3216,17 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
...
+  // Load the disk-related index data if we are on disk and the save flow used
+  // sst-files, even if this is a duplicate.
   // In the case of a duplicate, `sp->diskSpec=NULL` thus handled appropriately
-  // On the disk side.
-  if (encver >= INDEX_DISK_VERSION && isSpecOnDisk(sp)) {
-    // TODO: Load the disk-related data only in case the `REDISMODULE_CTX_FLAGS_SST_RDB`
-    // context flag is set, as we wrote this data only in that case.
+  // On the disk side (RDB is depleted, without updating index fields).
+  bool use_sst = RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_SST_RDB;
+  if (encver >= INDEX_DISK_VERSION && isSpecOnDisk(sp) && use_sst) {
```
oshadmi: Wouldn't this change be more readable if the additional API calls are moved into the `if` block?

oshadmi (approved):

src/spec.c
raz-mon: I like a single condition if possible, such that as a reader you don't need to remember conditions in your stack.
It can be more readable to split if the condition gets sticky and hard, but here it's as simple as it gets, easy conjunction, so I like it better like this

PR #8001, opened 2026-01-08
build: add LTO option enabling C/Rust LTO

gdesmott:
This should improve performance at the ffi boundary.

It requires to build the C code using a `clang` version matching LLVM version used by `rustc`.

#### Mark if applicable

---

> [!NOTE]
> Introduces an optional cross-language LTO build mode to optimize FFI boundaries.
> 
> - Adds `LTO` flag support (`LTO=1`) in `Makefile` and `build.sh`; updates help text
> - Enforces Linux-only usage and verifies matching LLVM major versions between `rustc` and `clang`
> - Configures CMake to use `clang/clang++` and `lld`, enables IPO, and sets Rust `RUSTFLAGS` (`-C linker-plugin-lto -C linker=clang -C link-arg=-fuse-ld=lld`)
> - No functional code changes beyond build system adjustments
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 37c76f33c8807ff5eff89f207266022817d6f8ed. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -23,6 +23,10 @@ VERBOSE=0        # Verbose output flag
 QUICK=${QUICK:-0} # Quick test mode (subset of tests)
 COV=${COV:-0}    # Coverage mode (for building and testing)
 BUILD_INTEL_SVS_OPT=${BUILD_INTEL_SVS_OPT:-0} # Use SVS pre-compiled library
+# Enable Rust/C LTO. Requires Clang and lld (Linux only).
+# Clang needs to have the same version as the LLVM version used by Rust.
+# Check using `clang --version` and `rustc --version --verbose`.
+LTO=0
 
 # Test configuration (0=disabled, 1=enabled)
 BUILD_TESTS=0          # Build test binaries
@@ -137,6 +141,9 @@ parse_arguments() {
         ;;
       BUILD_INTEL_SVS_OPT=*)
         BUILD_INTEL_SVS_OPT="${arg#*=}"
+        ;;
+      LTO|lto)
+        LTO=1
         ;;
       *)
         # Pass all other arguments directly to CMake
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -336,6 +343,35 @@ prepare_cmake_arguments() {
   # Initialize with base arguments
   CMAKE_BASIC_ARGS="-DCOORD_TYPE=$COORD"
 
+  if [[ "$LTO" == "1" ]]; then
+    # Enable Rust/C LTO by using clang and lld
```
LukeMathWalker, line 354: Here we should check that we are indeed on Linux and either crash or emit a warning if that's not the case (skipping the remaining lines).

build.sh
```diff
@@ -336,6 +343,35 @@ prepare_cmake_arguments() {
...
+    # Check LLVM version compatibility between Rust and Clang
+    RUSTC_LLVM_VERSION=$(rustc --version --verbose | grep "LLVM version" | awk '{print $3}' | cut -d. -f1)
+    CLANG_LLVM_VERSION=$(clang --version | head -n1 | grep -oP 'version \K[0-9]+' | head -n1)
```
LukeMathWalker, line 357: What happens here if `clang` isn't installed?

build.sh
gdesmott: Good point, added that.

build.sh
gdesmott:
```
./build.sh: line 357: clang: command not found
Error: Could not detect LLVM versions for Rust and Clang
Cross-language LTO requires matching LLVM major versions.
Rust LLVM version: 21
Clang LLVM version: 
```

I think that's good enough?

build.sh
```diff
@@ -138,6 +142,9 @@ parse_arguments() {
       BUILD_INTEL_SVS_OPT=*)
         BUILD_INTEL_SVS_OPT="${arg#*=}"
         ;;
+      LTO|lto)
```
memark, line 145: You might want to convert this to the syntax used above for the other arguments (this was recently changed in https://github.com/RediSearch/RediSearch/pull/7987).

gdesmott: Thanks for the info. I changed it in my follow up PR.

PR #8003, opened 2026-01-08
[MOD-13399] One C static library to rule them all

LukeMathWalker:

Create a single static library that bundles together RediSearch's own C code as well as its C/C++ dependencies.
This dramatically simplifies the setup for our Rust tests and benchmarks whenever we need to invoke C symbols: we now need to link a single library rather than creating (many) ad-hoc static libraries on a case-by-case basis and going through the pain of figuring out which symbols need to be stubbed.

In a follow-up PR I'll remove all the intermediate static libraries we created for testing purposes over the past few months.
 
1. ...

#### Mark if applicable

---

> [!NOTE]
> Consolidates native code linkage and simplifies Rust test/bench setups.
> 
> - New `src/CMakeLists.txt` builds all C components/deps, exports `REDISEARCH_C_FINAL_OBJECTS`, creates `redisearch_c` and a unified `libredisearch_all.a`
> - Top-level `CMakeLists.txt` delegates C build to `src/`, builds Rust (`src/redisearch_rs`), and links final `redisearch` against `redisearch_c` + `redisearch_rs`
> - Add `build_utils::bind_foreign_c_symbols()` to link `libredisearch_all.a` and C++ stdlib, with strict unresolved-symbol handling
> - `redisearch_rs` `build.rs` scans `*_ffi` crates and emits `link_guard.rs` to keep `extern "C"` symbols from being stripped; included in the crate
> - Replace `bind_redis_alloc_symbols_to_mock_impl!` with `mock_or_stub_missing_redis_c_symbols!`, adding OpenSSL-related stubs; update tests/benchers to depend on `redisearch_rs` and macro
> - Update Cargo manifests (add `cc`, `quote`, `syn`, `toml`, `walkdir`, etc.) and add Rust docs in `CONTRIBUTING.md` on invoking foreign C symbols
> - Minor: comment fix, include dirs clarified
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4ac87c23ba56ccc3b4a4a8aa2b3495ca5a275215. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -139,146 +139,145 @@ add_dependencies(redisearch_c generate_command_info)
 # by RediSearch or by one of its dependencies (either direct or transitive).
 # This static library will be linked by Rust tests and benchmarks whenever
 # they have to invoke a foreign symbol.
-if(BUILD_SEARCH_UNIT_TESTS)
-    # A helper function to recursively collect all static library dependencies from a target.
-    # Accumulates results in _COLLECT_LIBS (library paths) and _COLLECT_TARGETS (target names).
-    function(_collect_static_libs_recurse target)
-        # Skip if already visited or not a valid target
-        if(target IN_LIST _COLLECT_TARGETS OR NOT TARGET ${target})
-            return()
-        endif()
 
-        # Mark as visited immediately to avoid cycles
-        list(APPEND _COLLECT_TARGETS ${target})
-        set(_COLLECT_TARGETS ${_COLLECT_TARGETS} PARENT_SCOPE)
+# A helper function to recursively collect all static library dependencies from a target.
+# Accumulates results in _COLLECT_LIBS (library paths) and _COLLECT_TARGETS (target names).
+function(_collect_static_libs_recurse target)
+    # Skip if already visited or not a valid target
+    if(target IN_LIST _COLLECT_TARGETS OR NOT TARGET ${target})
+        return()
+    endif()
 
-        # Get target properties
-        get_target_property(target_type ${target} TYPE)
-        get_target_property(is_imported ${target} IMPORTED)
+    # Mark as visited immediately to avoid cycles
+    list(APPEND _COLLECT_TARGETS ${target})
+    set(_COLLECT_TARGETS ${_COLLECT_TARGETS} PARENT_SCOPE)
 
-        # Try to get library file path for static libraries
-        set(lib_path "")
-        if(target_type STREQUAL "STATIC_LIBRARY")
-            if(is_imported)
-                # Try various IMPORTED_LOCATION properties
-                foreach(loc_prop IMPORTED_LOCATION IMPORTED_LOCATION_RELEASE IMPORTED_LOCATION_NOCONFIG)
-                    get_target_property(lib_path ${target} ${loc_prop})
-                    if(lib_path)
-                        break()
-                    endif()
-                endforeach()
-            else()
-                set(lib_path $<TARGET_FILE:${target}>)
-            endif()
-        elseif(target_type STREQUAL "UNKNOWN_LIBRARY")
-            # IMPORTED target with unknown type - check if it's a static library
-            foreach(loc_prop IMPORTED_LOCATION IMPORTED_LOCATION_RELEASE)
+    # Get target properties
+    get_target_property(target_type ${target} TYPE)
+    get_target_property(is_imported ${target} IMPORTED)
+
+    # Try to get library file path for static libraries
+    set(lib_path "")
+    if(target_type STREQUAL "STATIC_LIBRARY")
+        if(is_imported)
+            # Try various IMPORTED_LOCATION properties
+            foreach(loc_prop IMPORTED_LOCATION IMPORTED_LOCATION_RELEASE IMPORTED_LOCATION_NOCONFIG)
                 get_target_property(lib_path ${target} ${loc_prop})
                 if(lib_path)
                     break()
                 endif()
             endforeach()
+        else()
+            set(lib_path $<TARGET_FILE:${target}>)
         endif()
-
-        # Add to collection if we found a static library
-        if(lib_path AND (lib_path MATCHES "\\.(a|lib)$" OR NOT is_imported))
-            list(APPEND _COLLECT_LIBS ${lib_path})
-            set(_COLLECT_LIBS ${_COLLECT_LIBS} PARENT_SCOPE)
-        endif()
-
-        # Recurse into link dependencies
-        foreach(prop LINK_LIBRARIES INTERFACE_LINK_LIBRARIES)
-            get_target_property(deps ${target} ${prop})
-            if(NOT deps)
-                continue()
+    elseif(target_type STREQUAL "UNKNOWN_LIBRARY")
+        # IMPORTED target with unknown type - check if it's a static library
+        foreach(loc_prop IMPORTED_LOCATION IMPORTED_LOCATION_RELEASE)
+            get_target_property(lib_path ${target} ${loc_prop})
+            if(lib_path)
+                break()
             endif()
-            foreach(dep IN LISTS deps)
-                # Skip generator expressions
-                if(dep MATCHES "^\\$<")
-                    continue()
-                endif()
-                # Recurse if it's a valid target
-                if(TARGET ${dep})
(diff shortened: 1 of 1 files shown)


src/redisearch_rs/build_utils/src/lib.rs
```diff
@@ -96,33 +96,29 @@ pub fn run_cbinden(header_path: impl AsRef<Path>) -> Result<(), Box<dyn std::err
     Ok(())
 }
 
-/// Links static libraries
+/// Link all the relevant C dependencies to allow Rust (testing) code to invoke
```
gdesmott:
```suggestion
/// Link all the relevant C dependencies to allow Rust (testing and benchmarking) code to invoke
```

src/redisearch_rs/build_utils/src/lib.rs
```diff
@@ -138,14 +134,30 @@ pub fn link_static_libraries(libs: &[(&str, &str)]) {
             Some("x86_64") | None => "x64".to_owned(),
             Some(a) => a.to_owned(),
         };
+        let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap_or_else(|_| "linux".to_string());
```
gdesmott, line 137: Did you check if it also works when linking to a `DEBUG` build?
We often have to when debugging.

src/redisearch_rs/build_utils/src/lib.rs
```diff
@@ -138,14 +134,30 @@ pub fn link_static_libraries(libs: &[(&str, &str)]) {
...
-        link_static_lib(&bin_root, lib_subdir, lib_name).unwrap();
-    }
+    link_static_lib(&bin_root, "src", "redisearch_all").unwrap();
+}
+
+/// Link the C++ standard library using the platform's default.
+///
+/// This is needed for VectorSimilarity and other C++ code that RediSearch depends on.
+/// We compile a dummy C++ file which causes cc to emit the appropriate link flags,
+/// using the same approach as the `link-c-plusplus` crate.
```
gdesmott, line 150: We cannot just use this crate?

src/redisearch_rs/c_entrypoint/redisearch_rs/build.rs
```diff
@@ -0,0 +1,206 @@
...
+/// since the Rust-provided symbols it needs are missing.
+///
+/// # Obvious Solutions That Don't Work
+///
+/// - **`extern crate`**: Adding `extern crate fnv_ffi;` ensures the crate is linked, but
+///   doesn't prevent the linker from stripping unused symbols within that crate.
+///
+/// - **`#[used]` on functions**: This would be the ideal solution, but `#[used]` is only
+///   stable for `static` items. Using it on functions requires the unstable
+///   `#![feature(used_linker)]`.
```
gdesmott: We could link the ticket about the stabilization of this feature, so one can reconsider at some point when/if it reaches stable.

src/redisearch_rs/c_entrypoint/redisearch_rs/build.rs
```diff
@@ -0,0 +1,206 @@
...
+        })
+        .unwrap_or_default();
+    if ffi_crates.is_empty() {
+        panic!("No *_ffi crates found in Cargo.toml dependencies");
+    }
+
+    let mut symbols: Vec<(String, String)> = Vec::new();
+
+    for (_, crate_path) in &ffi_crates {
+        println!("cargo:rerun-if-changed={}", crate_path.display());
```
gdesmott, line 97: That's enough to regenerate if the rs files change? We don't need it for each source file?

src/redisearch_rs/build_utils/src/lib.rs
LukeMathWalker: I couldn't get it to work reliably, and it becomes one more thing that each crate that needs to consume C code has to add to its Cargo.toml. This way, instead, we centralize the knowledge in one place, behind a single function invocation.

src/redisearch_rs/c_entrypoint/redisearch_rs/build.rs
LukeMathWalker: Yes, if the path points to a directory scans each file inside of it for changes automatically.

src/redisearch_rs/build_utils/src/lib.rs
LukeMathWalker: It doesn't allow linking to `DEBUG` builds, but that was the case beforehand as well.
We can tackle that in a follow-up change.

src/redisearch_rs/c_entrypoint/redisearch_rs/build.rs
LukeMathWalker: Added the tracking issue!

LukeMathWalker:
I triggered a build on all platforms [here](https://github.com/RediSearch/RediSearch/actions/runs/20950339048), and it's all ✅

src/redisearch_rs/workspace_hack/Cargo.toml
```diff
@@ -35,6 +35,7 @@ serde = { version = "1", features = ["alloc", "derive"] }
 serde_core = { version = "1", features = ["alloc"] }
 syn-f595c2ba2a3f28df = { package = "syn", version = "2", features = ["extra-traits", "fold", "full", "visit", "visit-mut"] }
 thiserror = { version = "2" }
+toml = { version = "0.9" }
```
gdesmott, line 38: how was it failing?

LukeMathWalker: The hack crate wasn't up to date after having rebased on top of master.

gdesmott (dismissed):

src/CMakeLists.txt
```diff
@@ -0,0 +1,284 @@
...
+
+add_dependencies(redisearch_c VectorSimilarity)
+add_dependencies(redisearch_c generate_command_info)
+
+#----------------------------------------------------------------------------------------------
+# Build a static library that combines all symbols defined in C/C++,
+# by RediSearch or by one of its dependencies (either direct or transitive).
+# This static library will be linked by Rust tests and benchmarks whenever
+# they have to invoke a foreign symbol.
+if(BUILD_SEARCH_UNIT_TESTS)
```
gdesmott: Working on branches based on yours, I'm wondering if it would not make everyone's life easier to always build this.

When switching branches I generally do `./build.sh FORCE` and then manually run the tests I'm working on. Problem is I then end up with an outdated lib leading to linking issues.

The solution is obviously to use `./build.sh RUN_RUST_TESTS FORCE` but it's trickier and I suspect more people may be confused by this as well.

Enabling this is just creating an extra lib from existing object files right? So the increase in compile time should be trivial.

LukeMathWalker: The build time overhead is minimal, I concur. 
I'd be OK with building it unconditionally to minimize wasted time during development. Wdyt @alonre24?

alonre24: no problem

src/redisearch_rs/value/src/lib.rs
```diff
@@ -164,4 +164,9 @@ pub trait Value: Sized {
 }
 
 #[cfg(test)]
-redis_mock::bind_redis_alloc_symbols_to_mock_impl!();
+redis_mock::mock_or_stub_missing_redis_c_symbols!();
+#[cfg(test)]
+#[allow(non_upper_case_globals)]
+#[unsafe(no_mangle)]
+pub static mut RSDummyContext: *mut redis_mock::ffi::RedisModuleCtx =
+    redis_mock::globals::redis_module_ctx();
```
alonre24, line 172: Why do we need these additions only here?

alonre24 (approved):
One small comment. 
I believe I got the full picture. Awkward, but hope it will make our lives easier.

src/redisearch_rs/value/src/lib.rs
LukeMathWalker: The `value` test suite isn't linking to C code at the moment, but rather using a mocked implementation of the relevant symbols. @daxhuiberts is removing the whole mock in its RSValue working branch, so I avoided bundling that change in here.

PR #8008, opened 2026-01-08
[MOD-13586] RSValue: Type checks

daxhuiberts:
This adds the RsValue ffi type checks (`RSValue_Type` and `RSValue_IsXXX`)

This also adds a `expect_value` util helper. This will be used many upcoming calls in value ffi to convert an `RsValue` pointer to an `RsValue` reference. Since the rsvalue functions are invoked very often we don't want to pay the runtime check of `expect()`, but we do in a debug build.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> Introduces FFI type inspection for `RsValue` and a utility for safe pointer-to-ref conversion.
> 
> - Adds `RSValue_Type`, `RSValue_IsReference`, `RSValue_IsNumber`, `RSValue_IsString`, `RSValue_IsArray`, `RSValue_IsTrio`, and `RSValue_IsNull` in `value_ffi` with C declarations in `headers/value.h`
> - Defines new `RsValueType` enum layout and values; maps string variants to `String` and `Ref` to `Reference`
> - Adds `util::expect_value` for debug null checks and release `unwrap_unchecked`
> - Implements `RsValue::fully_dereferenced()` to resolve nested `Ref` chains for null checks
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dd9aef100048f320b6aef3e969b02f670f20b7bb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/value_ffi/src/value_type.rs
+++ b/src/redisearch_rs/c_entrypoint/value_ffi/src/value_type.rs
@@ -141,11 +141,14 @@ pub const unsafe extern "C" fn RSValue_IsTrio(value: *const RsValue) -> bool {
 ///
 /// 1. If `value` is non-null, it must point to a valid [`RsValue`] obtained from an `RSValue_*` function.
 #[unsafe(no_mangle)]
-pub const unsafe extern "C" fn RSValue_IsNull(value: *const RsValue) -> bool {
+pub unsafe extern "C" fn RSValue_IsNull(value: *const RsValue) -> bool {
     // Safety: ensured by caller (1.)
     let Some(value) = (unsafe { value.as_ref() }) else {
         return true;
     };
+
+    // C implementation does a recursive check on reference types.
+    let value = value.fully_dereferenced();
 
     matches!(value, RsValue::Null)
 }
--- a/src/redisearch_rs/value/src/lib.rs
+++ b/src/redisearch_rs/value/src/lib.rs
@@ -58,6 +58,16 @@ pub enum RsValue {
     Map(RsValueMap),
 }
 
+impl RsValue {
+    pub fn fully_dereferenced(&self) -> &Self {
+        if let RsValue::Ref(ref_value) = self {
+            ref_value.value().fully_dereferenced()
+        } else {
+            self
+        }
+    }
+}
+
 #[cfg(test)]
 redis_mock::mock_or_stub_missing_redis_c_symbols!();
 #[cfg(test)]


src/redisearch_rs/c_entrypoint/value_ffi/src/value_type.rs
```diff
@@ -7,52 +7,116 @@
...
+    Reference = 6,
+    Trio = 7,
+    Map = 8,
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_Type(value: *const RsValue) -> RsValueType {
+    debug_assert!(!value.is_null());
+
+    let value = unsafe { &*value };
```
daxhuiberts: I have to do this quite often (in upcoming PRs) that it might warrant something more abstract. I was thinking about a macro like:
```rust
macro_rules! cast_unchecked {
    ($var:ident) => {{
        debug_assert!(!$var.is_null(), concat!(stringify!($var), " must not be NULL"));
        &*$var
    }};
}
```
And then use it like:
```rust
let value = unsafe { cast_unchecked!(value) };
```
Is this preferrable? Or maybe there is a alternative solution?

src/redisearch_rs/c_entrypoint/value_ffi/src/value_type.rs
```diff
@@ -7,52 +7,116 @@
...
+    }
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_IsReference(value: *const RsValue) -> bool {
+    if value.is_null() {
+        return false;
+    }
+
+    let value = unsafe { &*value };
```
daxhuiberts: Also here, I have to do this in quite a few places as well (also in upcoming PRs) that it also might warrant something more abstract, like this:
```rust
macro_rules! cast_or_return {
    ($var:ident) => {{ cast_or_return!($var, Default::default()) }};
    ($var:ident, $return_val:expr) => {{
        if $var.is_null() {
            return $return_val;
        } else {
            &*$var
        }
    }};
}
```
And then use it like:
```rust
let value = unsafe { cast_or_return!(value) };
```
The implicit return is less ideal here, but the macro name should convey as such.
Same question: Is this preferrable? Or maybe there is a alternative solution?

src/redisearch_rs/c_entrypoint/value_ffi/src/value_type.rs
memark: If the safety docs require the pointer to be non-null, I think it's ok to enforce it in prod as well, and therefore enough to do
`.as_ref().expect("x must not be null")`

I personally don't think a macro is warranted in this case.

src/redisearch_rs/c_entrypoint/value_ffi/src/value_type.rs
memark: My first question is again, is a null pointer really? I'm not that familiar with RSValue, but it seems liberal to accept. :)

If it _is_, then I would prefer regular Rust code as well (not that much longer):
`let Some(value) = value.as_ref() else { return false; }`

memark:
On a more general note, what would macros gain us in these cases, compared to regular `#[inline]` functions?

src/redisearch_rs/c_entrypoint/value_ffi/src/value_type.rs
daxhuiberts: Yes, the existing C implementation does the null check and without it stuff crashes.

daxhuiberts:
> On a more general note, what would macros gain us in these cases, compared to regular `#[inline]` functions?

I was working back from a larger change, code already using `expect_unchecked` and some other operating working with `Arc::from_raw`. Thus I created this slimmed down PR to first get some fundamentals right and your suggestions are extremely helpful.

src/redisearch_rs/c_entrypoint/value_ffi/src/value_type.rs
LukeMathWalker: The whole point of making it a safety requirement, though, it's to avoid having to check it at runtime. I'm not against checking it in production, but we need to measure the overhead.

If we are looking for terseness, we could just as easily use a helper function though, I don't think a macro is needed in this particular case.

src/redisearch_rs/c_entrypoint/value_ffi/src/value_type.rs
LukeMathWalker: Rather than using a macro, I'd create a `matches_type` helper function that takes a raw pointer and a kind as argument. Then delegate to it in each entry point.

src/redisearch_rs/c_entrypoint/value_ffi/src/value_type.rs
daxhuiberts: Thanks for the feedback. I'll change it to a helper function which does a `debug_assert`.
I might like to do a benchmark run to use `expect` after I get all of the `RsValue` stuff implemented to see whether this might give a performance impact.

Lastly regarding the macro: The only thing it provides over a helper function is that the var name is automatically used in the debug message. But a stack trace with line info gives us enough information already.
I got this idea from `expect_unchecked`, but that macro might also be overkill then.

src/redisearch_rs/c_entrypoint/value_ffi/src/value_type.rs
daxhuiberts: Well, each of those `IsX` functions does a `matches!(value, <pattern>)` check. You can't provide `<pattern>` as an argument to a `matches_type` helper function.

I'll change it to a `let Some(value) = value.as_ref() else { return false; }`. That little bit of identical logic in each of those 6 functions is fine imo.

LukeMathWalker: > Well, each of those IsX functions does a matches!(value, <pattern>) check. You can't provide <pattern> as an argument to a matches_type helper function.

You only need a proper pattern for the string case, though, no? All the others can be a simple `value.type_() == expected_type`.

LukeMathWalker: (It doesn't matter anyway, let's be clear)

src/redisearch_rs/c_entrypoint/value_ffi/src/value_type.rs
```diff
@@ -7,52 +7,161 @@
...
+///
+/// 1. If `value` is non-null, it must point to a valid [`RsValue`] obtained from an `RSValue_*` function.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_IsNull(value: *const RsValue) -> bool {
+    // Safety: ensured by caller (1.)
+    let Some(value) = (unsafe { value.as_ref() }) else {
+        return true;
+    };
+
+    let value = value.fully_dereferenced();
```
LukeMathWalker, line 151: This is a bit of a departure from the other `RsValue_Is*` functions—why the difference?

daxhuiberts: Compatibility with the C version. See here: https://github.com/RediSearch/RediSearch/blob/master/src/value/value.c#L270-L274

I'll add it as a comment in the code.

daxhuiberts: Also, removing it will fail some test.

kei-nan (approved):

PR #8017, opened 2026-01-10
[MOD-13457] Let version suffix in snapshot contains git commit sha

alonre24:

Having the git commit sha part of the version suffix in the snapshot artifact file name instead of the workflow id. 

This should allow us to narrow down the list of files to take into consideration from the snapshot dir upon moving the artifacts to the release dir (without doing so, we are exhausting the storage of the GitHub runner - see https://github.com/RediSearch/RediSearch/actions/runs/20824549408)

Followup task - adjust the QA to use the updated artifact name

#### Mark if applicable

---

> [!NOTE]
> Switches snapshot artifact naming to include the short git SHA and updates release selection to match it, improving determinism and logging around included/excluded artifacts.
> 
> - In `flow-build-artifacts.yml`, version suffix now uses `.{TIMESTAMP}.{SHORT_SHA}` derived from `set-sha` instead of workflow run number
> - In `event-release.yml`, S3 suffix matching updated to expect `.{SOURCE}.{YYYYMMDD}.{HHMMSS}.{SHORT_SHA}.zip`; derives `expected_short_sha` from `EXPECTED_SHA`
> - Filters artifacts by exact `EXPECTED_SHA`, logs `Excluded Files` and `Unexpected SHAs`, warns when mismatches found; copies only matching artifacts and then logs `New Files`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0f53f7e4a359f5b70b304b75d53103665add9063. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/event-release.yml
+++ b/.github/workflows/event-release.yml
@@ -257,7 +257,8 @@ jobs:
           sha_list = list(sha_list)
           
           # Include only files that match the expected SHA
-          include_list = [f for f, sha in zip(files, sha_list) if sha == expected_sha]
+          exclude_list = [(f, sha) for f, sha in zip(files, sha_list) if sha != expected_sha]
+          include_list = [f for f in files if f not in [x for x, _ in exclude_list]]
 
           if not include_list:
               raise Exception(f"::error title=No artifacts found with expected SHA {expected_sha}!")
@@ -281,9 +282,17 @@ jobs:
           dest_files = [get_target_name(f) for f in include_list]
 
           # Log files
+          group_print("Excluded Files", exclude_list)
           group_print("Included Files", include_list)
-          group_print("New Files", dest_files)
-
+          group_print("Unexpected SHAs", set([sha for _, sha in exclude_list]))
+          
           # Copy included files to new location
           for src, dst in zip(include_list, dest_files):
               client.copy_object(Bucket=bucket, Key=dst, CopySource={"Bucket": bucket, "Key": src}, ACL="public-read")
+
+          group_print("New Files", dest_files)
+
+          if len(exclude_list) > 0:
+              print("::warning title=Unexpected Files::The workflow has encountered files that do not match the "
+                    "expected git sha. These files will not be included in the release artifacts. Look for the "
+                    "`Excluded Files` section above for more details.")


alonre24:
Succecfull run on a dummy version branch 6.4 - https://github.com/RediSearch/RediSearch/actions/runs/20880732731/job/59996635306

BenGoldberger:
Looks good 
make sure you backport to all branches

BenGoldberger (approved):

.github/workflows/event-release.yml
```diff
@@ -196,12 +196,14 @@ jobs:
           oss_dir = "redisearch-oss"
           ent_dir = "redisearch"
 
-          # Suffix pattern: .{SOURCE}.{TIMESTAMP}.{WORKFLOW_NUM}.zip
-          # Example: .2.10.20231228.123456.123.zip
-          # Pattern matches: .+.{SOURCE}. followed by timestamp (YYYYMMDD.HHMMSS) and workflow number
-          suffix_pattern = re.compile(rf"(.+)\.{re.escape(os.environ['SOURCE'])}\.\d{{8}}\.\d{{6}}\.\d+\.zip$")
-          new_suffix = rf"\1.{os.environ['CUR_VERSION']}.zip"
           expected_sha = os.environ["EXPECTED_SHA"]
+          expected_short_sha = expected_sha[:7]
```
GuyAv46, line 200: Any reason not to use the full SHA?

GuyAv46 (dismissed):

.github/workflows/event-release.yml
alonre24: not really, do you think we should?

GuyAv46 (commented):
Consider adding back the "unexpected SHAs" and excluded files, now that we don't expect a huge list of candidates, and we can add the extra validation that the zip suffix matches the internal SHA from the module.json file (we now expect to never see such candidates)

.github/workflows/event-release.yml
GuyAv46: Not really, but also no harm in doing so.
In any case, we should keep validating that the sha in the module.json matches

GuyAv46 (approved):

PR #8018, opened 2026-01-10
MOD-12736: Fix FT.HYBRID with LOAD *

nafraf:

1. Current:
`FT.HYBRID` + `LOAD *` was crashing.
The original code assumed all keys would exist in the destination lookup before `RLookupRow_WriteFieldsFrom` was called, but with `LOAD *`, lookup keys are created dynamically during document loading.

2. Change:
- Modify `RLookupRow_WriteFieldsFrom()` to create on-demand the missing keys only when `LOAD *` is used.
- Skip initial sync for LOAD * (it's useless - keys don't exist yet)

3. Outcome:
`FT.HYBRID` + `LOAD *`  should work.

1. MOD-...
2. #...

1. ...

#### Mark if applicable

---

> [!NOTE]
> Fixes crashes with `FT.HYBRID` + `LOAD *` by deferring schema sync and creating missing lookup keys on demand.
> 
> - Add `createMissingKeys` flow: `HybridLookupContext` gains a flag; `RLookupRow_WriteFieldsFrom(...)` now accepts `createMissingKeys` and creates destination keys if absent (LOAD * case)
> - Propagate LOAD * state: set `hreq->reqflags` from parsed params; pass flag to merger via `HybridLookupContext_New`; `hybridMergerStoreUpstreamResult` uses new API
> - Skip upfront key synchronization when `QEXEC_AGG_LOAD_ALL` is set; perform it only otherwise (both standalone and coordinator paths)
> - Minor: open score key as before; small comment/format tweaks
> - Tests: update C++ rlookup tests to new signature; add new tests for on-demand key creation; add Python tests validating `FT.HYBRID ... LOAD *` across field combinations and with yielded scores
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit de1733fa51964e8c759b2adb0bef7cfbbe4eead3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/rlookup.c
+++ b/src/rlookup.c
@@ -987,14 +987,12 @@ void RLookupRow_WriteFieldsFrom(const RLookupRow *srcRow, const RLookup *srcLook
     RLookupKey *dest_key = RLookup_FindKey(destLookup, src_key->name, src_key->name_len);
     if (!createMissingKeys) {
       RS_ASSERT(dest_key != NULL);  // Assumption: all source keys exist in destination
-    } else {
-      if (!dest_key) {
+    } else if (!dest_key) {
         // Key doesn't exist in destination - create it on demand.
         // This can happen with LOAD * where keys are created dynamically.
         // Inherit non-transient flags from source.
         uint32_t flags = src_key->flags & ~RLOOKUP_TRANSIENT_FLAGS;
         dest_key = RLookup_GetKey_WriteEx(destLookup, src_key->name, src_key->name_len, flags);
-      }
     }
     // Write fields to destination (increments refcount, shares ownership)
     RLookup_WriteKey(dest_key, destRow, value);


src/rlookup.c
```diff
@@ -984,7 +984,13 @@ void RLookupRow_WriteFieldsFrom(const RLookupRow *srcRow, const RLookup *srcLook
 
     // Find corresponding key in destination lookup
     RLookupKey *dest_key = RLookup_FindKey(destLookup, src_key->name, src_key->name_len);
-    RS_ASSERT(dest_key != NULL);  // Assumption: all source keys exist in destination
+    if (!dest_key) {
+      // Key doesn't exist in destination - create it on demand.
+      // This can happen with LOAD * where keys are created dynamically.
+      // Inherit non-transient flags from source.
+      uint32_t flags = src_key->flags & ~RLOOKUP_TRANSIENT_FLAGS;
+      dest_key = RLookup_GetKey_Write(destLookup, src_key->name, flags);
```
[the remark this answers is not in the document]
nafraf: Fixed. [2f68227](https://github.com/RediSearch/RediSearch/pull/8018/commits/2f682270889b2027fd1a472429b188c26204e29d)

src/rlookup.c
```diff
@@ -984,7 +984,13 @@ void RLookupRow_WriteFieldsFrom(const RLookupRow *srcRow, const RLookup *srcLook
 
     // Find corresponding key in destination lookup
     RLookupKey *dest_key = RLookup_FindKey(destLookup, src_key->name, src_key->name_len);
-    RS_ASSERT(dest_key != NULL);  // Assumption: all source keys exist in destination
+    if (!dest_key) {
+      // Key doesn't exist in destination - create it on demand.
+      // This can happen with LOAD * where keys are created dynamically.
+      // Inherit non-transient flags from source.
+      uint32_t flags = src_key->flags & ~RLOOKUP_TRANSIENT_FLAGS;
```
kei-nan: how does this work when we have 3 lookup tables?
2 for subqueries and 1 for hybrid merger
Are we sure everything works correctly?
I remember this being super sensitive

src/rlookup.c
```diff
@@ -984,7 +984,13 @@ void RLookupRow_WriteFieldsFrom(const RLookupRow *srcRow, const RLookup *srcLook
 
     // Find corresponding key in destination lookup
     RLookupKey *dest_key = RLookup_FindKey(destLookup, src_key->name, src_key->name_len);
-    RS_ASSERT(dest_key != NULL);  // Assumption: all source keys exist in destination
+    if (!dest_key) {
```
Itzikvaknin: In the .h file
```
/**
 * Write field data from source row to destination row with different schemas.
 * Iterate through source lookup keys, find corresponding keys in destination by name,
 * and write it to destination row using RLookup_WriteOwnKey().
 * Assumes all source keys exist in destination (enforce with ASSERT).
 */
void RLookupRow_WriteFieldsFrom(const RLookupRow *srcRow, const RLookup *srcLookup,
                               RLookupRow *destRow, RLookup *destLookup);
```

This is no longer valid:
` * Assumes all source keys exist in destination (enforce with ASSERT).`

Itzikvaknin:
I haven’t reviewed it yet, but regarding "Modify `RLookupRow_WriteFieldsFrom()` to create missing keys on demand" -
We originally separated `RLookupRow_WriteFieldsFrom` from `RLookup_AddKeysFrom` specifically to keep row operations and schema operations independent.

This change may break that separation. If we decide to go in that direction, we should clean up the code accordingly. Instead of maintaining a mixed state - where we both synchronize the `RLookups` and dynamically add fields at runtime - we can move fully to a dynamic approach.

src/rlookup.c
Itzikvaknin: I think we should use a boolean parameter instead of checking `if (!dest_key)`
We already know whether the expected schema is known in advance or not (`LOAD *`)
I’d prefer the code to make this explicit by having the caller of `RLookupRow_WriteFieldsFrom` pass a flag that indicates this

src/rlookup.c
nafraf: Unit tests added to confirm the behavior when the destination is empty and when there is partial overlap.

src/rlookup.c
nafraf: OK. Flag added and comments updated.  Thanks.

src/rlookup.c
```diff
@@ -984,7 +985,17 @@ void RLookupRow_WriteFieldsFrom(const RLookupRow *srcRow, const RLookup *srcLook
 
     // Find corresponding key in destination lookup
     RLookupKey *dest_key = RLookup_FindKey(destLookup, src_key->name, src_key->name_len);
-    RS_ASSERT(dest_key != NULL);  // Assumption: all source keys exist in destination
+    if (!createMissingKeys) {
+      RS_ASSERT(dest_key != NULL);  // Assumption: all source keys exist in destination
+    } else {
```
kei-nan: maybe use else if here in one line?

src/coord/hybrid/dist_hybrid.c
```diff
@@ -402,8 +402,8 @@ static int HybridRequest_prepareForExecution(HybridRequest *hreq, RedisModuleCtx
     }
     SearchCtx_UpdateTime(hreq->sctx, hreq->reqConfig.queryTimeoutMS);
 
-    // // Set request flags from hybridParams
-    // hreq->reqflags = hybridParams.aggregationParams.common.reqflags;
+    // Set request flags from hybridParams
+    hreq->reqflags = hybridParams.aggregationParams.common.reqflags;
```
kei-nan, line 406: why do we need this change?
are we sure it isn't done somewhere else?

src/rlookup.c
nafraf: Yes, it's better.  Done. [de1733f](https://github.com/RediSearch/RediSearch/pull/8018/commits/de1733fa51964e8c759b2adb0bef7cfbbe4eead3)

src/coord/hybrid/dist_hybrid.c
nafraf: We need it, because the cluster test was failing, the value of the flag was lost.

It's done in `buildPipelineAndExecute()` but it's executed in standalone or in cluster only for the internal command.
It's weird that the code was commented out, probably because we only were using `regflags` to check `QEXEC_F_IS_CURSOR` but we don't use it in `FT.HYBRID`

PR #8019, opened 2026-01-11
MOD-13102 MOD-13364: Add disk-based metrics to `INFO`

raz-mon:
This PR adds disk related info fields to the `INFO` command response.
Note: The disk memory is accounted for in the general (existing) memory metrics we report.

---

> [!NOTE]
> Introduces disk metrics reporting and integrates disk usage into existing memory stats.
> 
> - Defines `DiskColumnFamilyMetrics` and `MetricsDiskAPI` in `search_disk_api.h`; adds wrappers `SearchDisk_CollectDocTableMetrics` and `SearchDisk_CollectTextInvertedIndexMetrics` in `search_disk.[ch]`
> - Aggregates per-index disk metrics in `IndexesInfo_TotalInfo()` via `AccumulateDiskMetrics`, updating `total_mem`, `min_mem`, and `max_mem` (memtables, table readers, SST sizes)
> - Emits new `disk` section in INFO with `disk_doc_table` and `disk_text_inverted_index` dicts containing memtable/compaction/size/version/memory counters
> - Minor includes and plumbing to enable metrics only when disk API is enabled
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 24d4bece56ff97b025916a9e0b7ec82678079d48. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/search_disk_api.h
+++ b/src/search_disk_api.h
@@ -10,6 +10,7 @@
 
 #include "redismodule.h"
 #include "redisearch.h"
+#include "VecSim/vec_sim_common.h"
 
 #ifdef __cplusplus
 extern "C" {


src/info/indexes_info.c
```diff
@@ -10,8 +10,57 @@
...
+  // Memory usage
+  dest->estimate_table_readers_mem += src->estimate_table_readers_mem;
+}
+
+// Returns the total memory used by the disk components of the index.
+// We currently take into account:
+//  1. The mem-tables' size
+//  2. The estimate of the tables' readers memory
+//  3. SST files' size.
+// TODO: Add memory used for the deleted-ids set (relevant for doc-table only).
```
JoanFM, line 57: is this tracked in some ticket?

src/info/indexes_info.c
```diff
@@ -10,8 +10,57 @@
...
+static inline size_t updateTotalMemFromDiskMetrics(TotalIndexesInfo *info, DiskColumnFamilyMetrics *diskMetrics) {
```
JoanFM: why is info passed as input? I think the naming of the function is missleading, is not any update, is some kind of getter right?

src/info/info_redis/info_redis.c
```diff
@@ -84,6 +85,11 @@ void RS_moduleInfoFunc(RedisModuleInfoCtx *ctx, int for_crash_report) {
   // Run time configuration
   AddToInfo_RSConfig(ctx);
 
+  // Disk metrics, on Flex only.
+  if (isFlex) {
```
JoanFM: can u use the SearchDisk_IsEnabled family of APIs? so we can unify all the checks and calls to disk API (and we can change easily the condition if needed)

src/info/indexes_info.c
raz-mon: No, if I won't be able to add it in this PR I'll make one.

src/info/indexes_info.c
raz-mon: Yeah from my previous implementation, where it was setting it as well.
I understood it's better it not to set anything so we set from the main function.
Fixed, thanks

src/info/info_redis/info_redis.c
raz-mon: Sure, changed.

src/info/indexes_info.h
```diff
@@ -33,6 +34,7 @@ typedef struct {
   // GC
   InfoGCStats gc_stats;  // Garbage collection statistics
 
+  // Field stats
   TotalIndexesFieldsInfo fields_stats;  // Aggregated Fields statistics
```
kei-nan, line 38: how is this going to be filled out?

src/info/indexes_info.h
```diff
@@ -33,6 +34,7 @@ typedef struct {
   // GC
   InfoGCStats gc_stats;  // Garbage collection statistics
```
kei-nan, line 35: how is this going to be filled out?

src/info/indexes_info.h
kei-nan, line 27: how is this going to be filled out?

src/info/indexes_info.h
```diff
@@ -44,9 +46,13 @@ typedef struct {
   size_t num_active_indexes_querying;  // Number of active read indexes
   size_t num_active_indexes_indexing;  // Number of active write indexes
   size_t total_active_write_threads;   // Total number of active writes (proportional to the number
-                                       // of threads)
+  // of threads)
   size_t total_num_docs_in_indexes;      // Total number of documents in all indexes
```
kei-nan, line 50: how is this going to be filled out?

src/search_disk_api.h
```diff
@@ -169,11 +171,72 @@ typedef struct VectorDiskAPI {
   void (*freeVectorIndex)(void* vecIndex);
 } VectorDiskAPI;
 
+/**
+ * @brief Column family metrics for RocksDB/SpeeDB
+ *
+ * All metrics are non-string (integer) properties that can be queried efficiently.
+ * These metrics are specific to individual column families (doc_table or inverted_index).
+ */
+typedef struct DiskColumnFamilyMetrics {
```
kei-nan, line 181: I think this exposes too much of the implementation.
we shouldn't expose these fields here.
disk should be agnostic in this side of the fence

src/info/indexes_info.h
```diff
@@ -44,9 +46,13 @@ typedef struct {
...
   size_t total_active_queries;         // Total number of active queries (reads)
+
+  // Disk metrics
+  DiskColumnFamilyMetrics disk_doc_table;      // Aggregated doc_table metrics
```
kei-nan, line 54: This breaks the barrier we tried to keep up until now,
I think we should come up with a way where the logic and code for metrics will only be on the implementation side.

src/info/indexes_info.h
raz-mon: It's already filled out properly as far as I saw.

src/info/indexes_info.h
raz-mon: Once we implement our GC, another API call will populate a new struct for the GC info when needed.

src/info/indexes_info.h
raz-mon: When? Where?
This is currently filled by the disk usage (RAM + storage), and later on in hybrid indexes will contain the RAM indexes as well.

src/info/indexes_info.h
raz-mon: This is filled in the OSS side, and can (and should in my eyes) be done similarly for disk.
Requires a fix for the disk paths probably, I will take a look but it should be done in a separate PR, even so.

src/search_disk_api.h
raz-mon: While we do expose the metrics here, I'm not sure it's too much, and this is why I held the meeting to sync on this.
Whether this is too much can be discussed, but the aggregation happens on the OSS side so this is much more natural.
If we had a closed-source repo life would have been easier, but given we don't I think this does not pass the line of implementation exposure.

src/info/indexes_info.h
raz-mon: I held a meeting about this where we decided this is not too much to expose.
We can revisit whether it's too much or not (CC @alonre24), but I think it's fine if it's only the metrics we expose.
We also wanted something fast that does not require further migrations, we can perhaps more easily hide disk code when the migration is done later on.

raz-mon:
@kei-nan This was agreed upon in a meeting you missed. I don't think we can really 100% avoid exposure while we have this constellation. We can revisit if felt necessary, but it's most natural and other ways will require potential migrations/creativity that will take time (part of the considerations here, SKO).
@alonre24 WDYT?

src/info/indexes_info.h
kei-nan: I mean this specific field.

src/search_disk_api.h
kei-nan: I think an alternate approach can be to move the total info struct into rust, fill it out initially on this end of the fence.
move the struct to a callback and have the disk output the information from his end into a reply that will be sent out the client.

src/info/indexes_info.h
kei-nan: I think this should be reconsidered then.
I think creating a rust binding from the total info struct will be cleaner in the long run.

Even if we think the rust bindings is too much then a simple callback for disk to output its metrics and pass the reply pointer in the callback will be enough

src/info/indexes_info.h
raz-mon: Yes, I got that.

src/search_disk_api.h
raz-mon: If we can avoid populating the response from multiple places I think that's better.
Also note that some of the disk info affects the general info, e.g., `total_mem`.
If this exposure is acceptable I think calling disk API from here is fine, do we agree on that?

src/info/indexes_info.h
kei-nan: it seems to me like eventually the disk will have to add its information to this struct,
It sounds like if we port the struct into and fill it out on implementation end we would benefit from that long term.
when adding new field to info statistics it could be filled out by the disk side very fast.

src/info/indexes_info.h
raz-mon: Resolving as reconsidered with the same result.

src/info/indexes_info.h
raz-mon: Resolving as this is the same issue as in other threads.

src/info/indexes_info.h
raz-mon: Resolving as this is the same issue as in other closed threads.

raz-mon:
After discussion we decided to keep this approach.

src/info/indexes_info.c
raz-mon: Making a separate ticket for this, so we can separate and simplify.

src/search_disk_api.h
```diff
@@ -10,12 +10,15 @@
...
 #include "redisearch.h"
-#include "iterators/iterator_api.h"
+#include "VecSim/vec_sim_common.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+// Forward declarations to avoid circular dependencies
+typedef struct QueryIterator QueryIterator;
```
kei-nan, line 20: I don't understand why we need this here.
the type shouldn't be used in this header file.
it shouldn't be here.
maybe you meant to put it somewere else
anyhow please remove it from this file.

kei-nan (commented):
added one comment

src/search_disk_api.h
raz-mon: No, it should be here.
It's used in the file below, and we cannot include the header as it is cyclic.

PR #8026, opened 2026-01-11
MOD-12373: split IdList/Metric iterator types in ctx of sorting

GlenDC:
## Describe the changes in the pull request

Split IdList/Metric iterator types in ctx of sorting.

Rust FFI Constructors apply the correct types already.
Buggy corresponding C code is yet to be resolved but is out of scope.
Where these iterator types were used in C we do already handle that.

#### Which additional issues this PR fixes
1. MOD-12373

#### Main objects this PR modified

1. IteratorType:
   * ID_LIST => SORTED_ID_LIST || UNSORTED_ID_LIST
   * METRIC => METRIC_SORTED_BY_ID || METRIC_SORTED_BY_SCORE

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

#### Release Notes

- [ ] This PR requires release notes
- [x] This PR does not require release notes

If a release note is required (bug fix / new feature / enhancement), describe the **user impact** of this PR in the title.

---

> [!NOTE]
> **Medium Risk**
> Touches cross-language iterator plumbing (C/C++/Rust FFI) and iterator-type dispatch used by profiling and vector metric handling; mistakes could cause mis-typed iterators or memory ownership bugs.
> 
> **Overview**
> **User-visible change:** `FT.PROFILE` now reports distinct iterator types for ID list and vector range metric iterators (e.g. `ID-LIST-SORTED`/`ID-LIST-UNSORTED` and `METRIC SORTED BY ID`/`METRIC SORTED BY SCORE`), improving clarity when diagnosing query plans.
> 
> Internally, iterator type enums are split and threaded through query evaluation, geometry iterators, and the profile printer; Rust FFI constructors stop copying ID/metric arrays by taking ownership via a new `OwnedSlice`, reducing allocations and making memory ownership explicit.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b9496aa7a66cfb0a0366ec21d7e970375e3a8b45. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #8027, opened 2026-01-11
MOD-12374: re-use id-list/metrics C Alloc mem in Rust iterators

GlenDC:

Re-use id-list/metrics C Alloc mem in Rust iterators.

PR on top of https://github.com/RediSearch/RediSearch/pull/8026, so:
- either we merge it in that PR
- or we first merge the other one and rebase this on master afterwards.

1. MOD-12374

1. IdList
2. Metric

#### Mark if applicable

- [x] This PR introduces API changes (minor, but incase you were passing ids/metric-values by `Into` it will fail as we now have these constructors take an `Into<OwnedList>` which won't work in that case. Same for things like Collect. Easy enough to fix though and compiler will guide you. As such already fixed in this PR.

---

> [!NOTE]
> Introduces zero-copy reuse of C-allocated buffers for iterators.
> 
> - Adds `utils::OwnedSlice` (and `RedisSlice`) to wrap C-allocated arrays and free them via `RedisModule_Free` on drop
> - Updates `IdList` and `Metric` to store `OwnedSlice` instead of `Vec`, and constructors now accept `Into<OwnedSlice<...>>`
> - Removes redundant Rust-side allocations/copies and explicit frees in FFI layers
> - Changes FFI signatures to take mutable pointers: `NewSortedIdListIterator`, `NewUnsortedIdListIterator`, `NewMetricIteratorSortedById`, `NewMetricIteratorSortedByScore`; header `iterators_rs.h` regenerated accordingly
> - Exposes `utils` module in `rqe_iterators`; minor bench fixes to use explicit `Vec<_>` collections
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 500dd94df52aece9a727efe3c1b633cf426c7aad. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rqe_iterators/tests/integration/not.rs
+++ b/src/redisearch_rs/rqe_iterators/tests/integration/not.rs
@@ -538,7 +538,7 @@ fn read_timeout_via_timeout_ctx() {
             .doc_id,
         "rewind should have allowed us to start reading again from start, despites earlier timeout"
     )
-    // that said... internal timeout context is _not_ resetted,
+    // that said... internal timeout context is _not_ reset,
     // so it is bound to timeout once you make the required amount of read/skip_to calls...
 }
 
@@ -589,6 +589,6 @@ fn skip_to_timeout_via_timeout_ctx() {
             .doc_id,
         "rewind should have allowed us to start reading again from start, despites earlier timeout"
     )
-    // that said... internal timeout context is _not_ resetted,
+    // that said... internal timeout context is _not_ reset,
     // so it is bound to timeout once you make the required amount of read/skip_to calls...
 }


src/redisearch_rs/rqe_iterators/src/utils/owned_slice.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+impl<T> OwnedSlice<T> {
+    /// # Safety
+    ///
+    /// ptr must be non-null and point to `len` initialized elements
+    /// allocated via `RedisModule_Alloc`.
+    #[inline(always)]
+    pub const unsafe fn from_c(ptr: *const T, len: usize) -> Self {
```
gdesmott: Maybe this could be a `ptr::NonNull`?

src/redisearch_rs/rqe_iterators/src/utils/owned_slice.rs
```diff
@@ -0,0 +1,109 @@
...
+            kind: SliceKind::Rust(Vec::default()),
+        }
+    }
+}
+
+impl<T> OwnedSlice<T> {
+    /// # Safety
+    ///
+    /// ptr must be non-null and point to `len` initialized elements
+    /// allocated via `RedisModule_Alloc`.
```
gdesmott: I'd document here that this function take ownership of `ptr` and so should no longer be freed by the caller.

src/redisearch_rs/rqe_iterators/src/utils/owned_slice.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn deref(&self) -> &[T] {
+        match &self.kind {
+            SliceKind::C(s) => s,
+            SliceKind::Rust(v) => v,
+        }
+    }
+}
+
+enum SliceKind<T> {
+    C(RedisSlice<T>),
```
LukeMathWalker, line 63: I think we could reuse the `Buffer` type we have in the `buffer` crate here. What do you think?

src/redisearch_rs/rqe_iterators/src/utils/owned_slice.rs
```diff
@@ -0,0 +1,109 @@
...
+
+impl<T> OwnedSlice<T> {
+    /// # Safety
+    ///
+    /// ptr must be non-null and point to `len` initialized elements
+    /// allocated via `RedisModule_Alloc`. [`OwnedSlice`] takes
+    /// ownership from ptr and should therefore no longer be freed by callee.
+    #[inline(always)]
+    pub const unsafe fn from_c(ptr: *mut T, len: usize) -> Self {
+        Self {
```
gdesmott, line 32:
```suggestion
    pub const unsafe fn from_c(ptr: *mut T, len: usize) -> Self {
        debug_assert!(!ptr.is_null());
        Self {
```

GlenDC: Moved it to Redislice but, yes agreed and applied.

src/redisearch_rs/rqe_iterators/src/utils/owned_slice.rs
GlenDC: @LukeMathWalker it is similar in use except:

1. we do not need growth capabilities
2. more important... we are dealing with `T` slices instead of `u8` slices

Of course we could interpret `T` as a bunch of `u8` but dunno... 
To me it would make the code less easy to understand, without really adding much benefit. We're not dealing with a lot of code as it is right now either.

WDYT?

LukeMathWalker: I agree, let's go forward with this implementation.

BenGoldberger (approved):

PR #8033, opened 2026-01-13
[MOD-12284] Minor C adjustments

memark:
Some minor adjustment on the C side for `LoadRuleFields` code path:

- `RedisSearchCtx.spec` is not being used for this call path and was removed to simplify Rust port.
- `RLookupLoadOptions.keys` is never altered and was made also inner `const` to tighten up immutability.
- Comments on array type was added to _rules.h_ to document some knowledge which took me some time to figure out.

---

> [!NOTE]
> Focuses on minor C cleanups for the `LoadRuleFields` path and related types.
> 
> - Initialize `RedisSearchCtx` with only `redisCtx` in `RLookup_LoadRuleFields`; drop unused `.spec`
> - Make `RLookupLoadOptions.keys` `const`-correct (`const RLookupKey *const *`) to prevent mutation of the pointed-to keys and the array
> - Update `SchemaRule.filter_fields` type to `arrayof(char*)` to align with array utilities
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 23022000d0a232101bd5b22a0b8a2bcf6efd0a77. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/rules.h
+++ b/src/rules.h
@@ -53,9 +53,7 @@ typedef struct SchemaRule {
   arrayof(HiddenUnicodeString*) prefixes;
   HiddenString *filter_exp_str;
   struct RSExpr *filter_exp;
-  // array_t type array
-  char **filter_fields;
-  // regular array
+  arrayof(char*) filter_fields;
   int *filter_fields_index;
   char *lang_field;
   char *score_field;


src/rlookup.c
```diff
@@ -928,7 +928,7 @@ int RLookup_LoadRuleFields(RedisModuleCtx *ctx, RLookup *it, RLookupRow *dst, In
   }
 
   // load
-  RedisSearchCtx sctx = {.redisCtx = ctx, .spec = spec };
```
kei-nan, line 931: I don't know about this line, I think dropping the initialization means it will be initialized to mangled bytes.
I'm not sure it will be initialized to 0 in all compilers.
Maybe you intended to remove if from the struct altogether?
Not sure what was your intent here.

src/rules.h
```diff
@@ -53,7 +53,9 @@ typedef struct SchemaRule {
   arrayof(HiddenUnicodeString*) prefixes;
   HiddenString *filter_exp_str;
   struct RSExpr *filter_exp;
+  // array_t type array
   char **filter_fields;
+  // regular array
```
kei-nan: consider instead of the commants using arrayof() notation.
e.g
```
arrayof(int)
arrayof(char*)
```
how the arrays are allocated don't have to change.
but it documents the fields are array using a pseudo type notation.

src/rlookup.c
memark: I looked at the code path starting here (basically this function and the call into `loadIndividualKeys()`) and noted that this property is never being read. At the same time, its mutability causes problems on the Rust side, which I thought we could avoid using it in this case.

It's my impression that `RedisSearchCtx` is used in many different places, and certain properties are only used for some purposes. In this case `.spec` is not relevant. It is being used in other places, so we cannot remove it.

As far as I have been able to interpret the C standard, all omitted properties will be zero-initialized. But my case here is actually that it doesn't matter, since it will never be read.

memark: I could of course just not set it in the Rust version of the code, but I thought it made sense to reflect this on the C side as well, and to get your opinions about it.

src/rules.h
memark: I'm happy to change to any other syntax you prefer!

Could you help me formulate the two comments in this new syntax? What I want to highlight is the difference between the two _types_ of array (one "array_t" and one regular), not the _contents_ of the arrays.

I.e. on the first one I can/have to use `array_len_func`, but this doesn't work on the second one.

src/rlookup.c
kei-nan: Looks like it will be initialized to null according to ISO/IEC 9899, §6.7.9, ¶21 in the c standard.

However after looking at how sctx is used here I'm wondering if it might be better to modify `RLookupLoadOptions` so it won't have an sctx pointer but only a ctx pointer.
I think that will document what you are saying better in the code and will ensure no one tries to use the spec in the future from the rlookup options

src/rules.h
kei-nan: then maybe one can be arrayof() and the other remain as is.
arrayof is defined in arr.h so it implies array functions should get used.
I prefer types to document the type and less so comments.

memark: Agree. Done!

src/rlookup.c
memark: I like this idea, but I'm not sure it's possible. The `sctx->spec` is still used in other places, here for example:

(result_processor.c:653)
image

kei-nan: I don't know if it is too much but if that is the only place it is used maybe it can move into RPLoader so it still has sctx just not in the loadopts.

memark: Yes, I think that's out of scope. This was only the first error I found. Grepping for `sctx->spec` gives 257 hits. 😮‍💨

memark: No, wait. That's not relevant. Doing another search.

memark: I'm not able to do any meaningful searching I'm afraid... But yes, the `RLookupLoadOptions.sctx` is being used in more places.

memark: Could we do the proposed change now, and maybe add a ticket about doing the restructuring you suggest later on?

PR #8035, opened 2026-01-13
[MOD-12246] Add support for thin vectors with different capacity types (u8, u32 and u64)

LukeMathWalker:

Add a new generic parameter to `LowMemoryThinVec`, specifying its desired maximum capacity.
New unit tests have been added to cover the new capacity types. The handling of capacity has been refactored to be as type safe as possible, pushing errors (i.e. capacity overflows) ahead of allocation attempts.

In a follow-up PR, I'll rename the crate to `thin_vec`, since it no longer provides "just" the low-memory variant. 

1. `low_memory_thin_vec` crate

#### Mark if applicable

---

> [!NOTE]
> Introduces a generic capacity type for thin vectors and aligns FFI and headers with the new representation.
> 
> - Add generic `LowMemoryThinVec<T, S>` (default `S = u16`) with new `VecCapacity` trait and generic `Header<S>`; implement for `u8`, `u16`, `u32`, `u64`
> - Refactor allocation/layout logic (`allocation_layout`, `header_field_padding`, `reallocate`) to be size-type-aware; add per-size empty header singletons
> - Change several methods to non-`const` where needed (e.g., `len`, `is_empty`, `capacity`, `pop`) due to trait-based conversions
> - Update C FFI: expose `Header_u16` and alias `Header` to it; adjust generated headers (`low_memory_thin_vec.h`, `triemap.h`, `types_rs.h`) and cbindgen configs to exclude `Header_u16`
> - In inverted index, update `RSAggregateResult` accessors (`len`, `is_empty`, `capacity`) to non-`const`
> - Add extensive tests for all size types, capacity limits/overflows, iterator/clone/drop behavior, and memory usage
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 67038df36b22d6e034197eaaee13e077f6366b6b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/low_memory_thin_vec/tests/tests.rs
+++ b/src/redisearch_rs/low_memory_thin_vec/tests/tests.rs
@@ -239,6 +239,16 @@ impl Drop for DropCounter<'_> {
     }
 }
 
+/// A drop tracker that increments a shared counter when dropped.
+/// Safe to use in parallel tests unlike `static mut`.
+struct DropTracker(std::rc::Rc<std::cell::Cell<u32>>);
+
+impl Drop for DropTracker {
+    fn drop(&mut self) {
+        self.0.set(self.0.get() + 1);
+    }
+}
+
 #[test]
 fn test_small_vec_struct() {
     assert!(size_of::<LowMemoryThinVec<u8>>() == size_of::<usize>());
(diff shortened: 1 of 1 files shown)


src/redisearch_rs/inverted_index/src/index_result.rs
```diff
@@ -401,23 +401,23 @@ impl<'index> RSAggregateResult<'index> {
     }
 
     /// The number of results in this aggregate result
-    pub const fn len(&self) -> usize {
```
GuyAv46, line 404: How is this related, and what are the implications?

src/redisearch_rs/low_memory_thin_vec/src/layout.rs
```diff
@@ -1,16 +1,29 @@
...
+
+/// Gets the layout of the allocated memory for a `LowMemoryThinVec<T, S>` with the given `usize` capacity.
+///
+/// # Implementation details
+///
+/// This function doesn't check that the capacity is lower or equal to `S::MAX`.
+/// We keep this function around for performance reasons: `S::to_usize` can't be made a `const` function
+/// since it's a trait method, but we really don't want to pay a runtime cost when computing
+/// the layout alignment in [`allocation_alignment`].
+/// By having this (private) helper function, we can accommodate both without duplicating logic around.
```
GuyAv46, line 18: 🤩

src/redisearch_rs/inverted_index/src/index_result.rs
LukeMathWalker: This invokes the `len` method on `LowMemoryThinVec`, which is no longer const because it needs to invoke `to_usize` which is a trait method, and trait methods can't be const in stable Rust (yet).
There are no real perf implications, since the call would only be const for an `RSIndexResult` constructed in a `const` context, which is:

- Something we never do
- And it'd be empty, since you can't allocate in a non-const context

LukeMathWalker: Same reasoning for the other methods that I had to de-constify in this PR.

PR #8037, opened 2026-01-13
[MOD-13432] Fix FULLTEXT field metric count

lerman25:
Changed search_total_indexing_ops_text_fields metric to count unique terms indexed instead of documents.

Moved the metric update call from fulltextPreprocessor() to Indexer_Add(), after writeCurEntries() completes successfully. This ensures the metric accurately reflects the number of unique terms added to the inverted index, persisting even after document deletion.

---

> [!NOTE]
> Adjusts field indexing metrics to be accurate and consistent across types.
> 
> - FULLTEXT: move metric update from `fulltextPreprocessor` to `writeCurEntries()` and increment by new unique terms (`spec->stats.numTerms` delta)
> - Stats API: change `FieldsGlobalStats_UpdateFieldDocsIndexed` to accept `FieldType` (was `FieldSpec*`); update all call sites
> - INFO metrics: clarify semantics as total indexing operations; for TEXT count unique terms, others counted per-field per-doc; ensure JSON NULLs don’t increment; multi-value JSON counts once per doc
> - Tests: remove TEXT from doc-count tests, add `test_total_terms_indexed_text_fields`, update overlapping-index and same-type-field cases, and add JSON NULL field coverage
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 29d1782ac3bae9b202aa4294b8613fe54095ec94. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_info_modules.py
+++ b/tests/pytests/test_info_modules.py
@@ -2307,12 +2307,12 @@ def test_total_terms_indexed_text_fields(env):
   env.assertEqual(get_text_metric(), prev_metric,
                   message="Empty text field should not add any terms")
 
-  # Test 8: Document with only some TEXT fields populated
+  # Test 8: Document with only some TEXT fields that match the index schema
   prev_metric = get_text_metric()
-  # Only t1 is populated with 2 new unique terms, t2 is missing
-  conn.execute_command('HSET', 'multi:2', 't1', 'delta epsilon')
+  # Only t1 is populated with 2 new unique terms, t2 is missing, t3 is not indexed
+  conn.execute_command('HSET', 'multi:2', 't1', 'delta epsilon', 't3', 'gamma delta')
   env.assertEqual(get_text_metric(), prev_metric + 2,
-                  message="Partial doc should only count unique terms from populated fields")
+                  message="Only populated fields that match the index schema should be counted")
 
 
 # Test the 'total_indexing_ops_<field_type>_fields' INFO MODULES metrics with multi-value JSON.


src/indexer.c
```diff
@@ -388,6 +389,11 @@ static void Indexer_Process(RSAddDocumentCtx *aCtx) {
   // Handle FULLTEXT indexes
   if ((aCtx->fwIdx && (aCtx->stateFlags & ACTX_F_ERRORED) == 0)) {
     writeCurEntries(aCtx, &ctx);
+
+    // Update global statistics for fulltext fields after successful indexing
+    if (aCtx->numTextFields > 0) {
+      FieldsGlobalStats_UpdateTextFieldDocsIndexed(aCtx->numTextFields);
```
raz-mon: Any reason not to just put this in the `while` loop in `writeCurEntries`?
What is the exact metric this counter is counting? If it's the amount of entries we write for the text field, I think it can be there

src/info/global_stats.c
```diff
@@ -252,3 +252,12 @@ void FieldsGlobalStats_UpdateFieldDocsIndexed(const FieldSpec *fs, int toAdd) {
       break;
   }
 }
+
+void FieldsGlobalStats_UpdateTextFieldDocsIndexed(int toAdd) {
+  // Wrapper for updating fulltext field statistics without needing a FieldSpec.
+  // Creates a mock FieldSpec with just the necessary fields set.
+  FieldSpec mockTextFieldSpec = {0};
+  mockTextFieldSpec.types = INDEXFLD_T_FULLTEXT;
+  mockTextFieldSpec.options = 0;  // Indexable by default (FieldSpec_NotIndexable is not set)
```
raz-mon: This is redundant since you set it to `0` above already, perhaps a comment will suffice.
In general I would prefer removing the `is_indexable` check from the main updating function and just send a field-type as an argument to it, such that we also don't need this wrapper. I think we won't get to the indexing path that increments it if the field is not indexable, so we should be fine no?

src/indexer.c
lerman25: IIUC, the loop is for terms rather than for each field? We want to count docs per field

src/info/global_stats.c
lerman25: Yes, we can remove those lines

raz-mon: I see you left the wrapper function and the `is_indexable` check..
Your call it's not critical, but it's probably unnecessary

src/indexer.c
raz-mon: I thought the meaning of this metric is how many times we indexed documents for text fields..
`// Total number of documents indexed by each field type`
Actually by this description I would think it's something else - one increment per indexed doc, no matter the field count. I know that's not the meaning since you increment in bulkAdd, but it's a bit confusing.
Anyways, we can go with what you defined it to be, if it's the number of fields then you're right. If you wish to change to the number of index entries that would probably make a bit more sense IMHO.

src/info/global_stats.c
lerman25: Left by mistake, removed

src/indexer.c
```diff
@@ -124,6 +128,9 @@ static void writeCurEntries(RSAddDocumentCtx *aCtx, RedisSearchCtx *ctx) {
 
     entry = ForwardIndexIterator_Next(&it);
   }
+
+  // Update the number of terms added for metrics
+  FieldsGlobalStats_UpdateFieldDocsIndexed(INDEXFLD_T_FULLTEXT, spec->stats.numTerms - prevNumTerms);
```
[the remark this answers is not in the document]
lerman25: exactly

src/indexer.c
```diff
@@ -125,6 +128,9 @@ static void writeCurEntries(RSAddDocumentCtx *aCtx, RedisSearchCtx *ctx) {
...
```
raz-mon, line 133: Notice that this metric won't be correct for disk like this.
If you condition it on `isNew` in the loop that should solve that (other than the fact that we're not doing the necessary check on disk.. We should, not now of-course)

lerman25: Then `FT.INFO` num_terms is also incorrect for disk?

raz-mon: Yes..
I'll open a ticket for that on the disk side

lerman25: I think we can continue with this, and the metric will be "fixed" when num_terms will be fixed instead of changing the logic and then reverting later on

raz-mon: Why would we do that?
We will not revert anything, just fix the other metric

lerman25: I don't understand then, 
If the metric is only based on num_terms, and num_terms is correct for both disk and non-disk, why should we change the logic?
I think it's better than introducing a different logic just because a temporary bug exists

raz-mon: That's the way I would go, but I'm ok with this approach as well

tests/pytests/test_info_modules.py
```diff
@@ -2205,50 +2195,124 @@ def get_field_metrics():
...
+  prev_metric = get_text_metric()
+  conn.execute_command('HSET', 'doc:4', 't', '')
+  env.assertEqual(get_text_metric(), prev_metric,
+                  message="Empty text field should not add any terms")
+
+  # Test 8: Document with only some TEXT fields that match the index schema
+  prev_metric = get_text_metric()
+  # Only t1 is populated with 2 new unique terms, t2 is missing, t3 is not indexed
+  conn.execute_command('HSET', 'multi:2', 't1', 'delta epsilon', 't3', 'gamma delta')
+  env.assertEqual(get_text_metric(), prev_metric + 2,
```
meiravgri, line 2314: what about indexing the same term in two different fields (that exist in the schema) in the same doc?

tests/pytests/test_info_modules.py
```diff
@@ -2102,6 +2102,7 @@ def get_total_docs_indexed():
 
 # Test the 'total_indexing_ops_<field_type>_fields' INFO MODULES metrics.
 # These metrics count how many times each field type has indexed a document.
+# Note: TEXT fields are excluded from this test as they count terms, not documents.
```
meiravgri, line 2105: shoudlnt we at least test it behvaes as expected when combined with other fields? specifically tags?

PR #8041, opened 2026-01-14
[MOD-12244] Reduce the size of inverted indexes.

LukeMathWalker:
Align with the original C implementation, using a thin vector with u32 capacity rather than a Vec with usize capacity.
In particular, this reduces the size of inverted indexes with no blocks by 16 bytes.

---

> [!NOTE]
> Reduces inverted index memory usage and aligns accounting with the C implementation.
> 
> - Replace `Vec<IndexBlock>` with `LowMemoryThinVec<IndexBlock, u32>` (`BlockCapacity`) in `InvertedIndex`, cutting empty index from 40 to 24 bytes and accounting for an 8-byte thin-vec header
> - New/updated sizing: `InvertedIndex::memory_usage()`, `IndexBlock::STACK_SIZE`, `IndexBlock::mem_usage()`, `add_block()` returns heap growth; GC `apply_gc()` now accounts for header via `Header::<u32>::size_with_padding::<IndexBlock>()` and trims capacity
> - API/behavior tweaks: `IndexBlock::new(doc_id)` no longer returns size; `number_of_blocks()` no longer `const`; test-only `from_blocks` now takes `LowMemoryThinVec`; GC repair uses `SmallVec::from_iter`
> - Update C++/Rust/Python tests and constants to new sizes and header overhead
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9aded9587cee4597ac71e8ecf220ecc1b03ae003. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/cpptests/test_cpp_forkgc.cpp
+++ b/tests/cpptests/test_cpp_forkgc.cpp
@@ -403,8 +403,8 @@ TEST_F(FGCTestTag, testRemoveAllBlocksWhileUpdateLast) {
   ASSERT_EQ(1, sctx.spec->stats.numDocuments);
   // But the last block deletion was skipped.
   ASSERT_EQ(2, sctx.spec->stats.numRecords);
-  // 40 bytes is the base size of an inverted index
-  ASSERT_EQ(lastBlockMemory + 40, sctx.spec->stats.invertedSize);
+  // 24 bytes is the base size of an inverted index, 8 is the header of the block vector
+  ASSERT_EQ(lastBlockMemory + 24 + 8, sctx.spec->stats.invertedSize);
   ASSERT_EQ(1, TotalIIBlocks() - startValue);
 }
 
--- a/tests/cpptests/test_cpp_index.cpp
+++ b/tests/cpptests/test_cpp_index.cpp
@@ -168,15 +168,15 @@ TEST_P(IndexFlagsTest, testRWFlags) {
   ASSERT_EQ(exp_t_fieldMask_memsize, t_fiedlMask_memsize);
 
   // Details of the memory occupied by InvertedIndex in bytes (64-bit system):
-  // Vec<IndexBlock> blocks    24
-  // u32 n_uniqe_blocks         4
-  // flags IndexFlags           4
-  // u32 gc_marker              4
-  // ----------------------------
-  // Total                     36
-  // After padding             40
+  // LowMemoryThinVec<IndexBlock, u32> blocks    8
+  // u32 n_uniqe_blocks                          4
+  // flags IndexFlags                            4
+  // u32 gc_marker                               4
+  // ---------------------------------------------
+  // Total                                      20
+  // After padding                              24
 
-  size_t exp_idx_no_block_memsize = 40;
+  size_t exp_idx_no_block_memsize = 24;
 
   if (useFieldMask) {
     exp_idx_no_block_memsize += t_fiedlMask_memsize;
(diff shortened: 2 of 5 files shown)


src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -608,23 +603,34 @@ impl<E: Encoder> InvertedIndex<E> {
...
+        let mut mem_growth = 0;
         if self.blocks.len() == self.blocks.capacity() {
             self.blocks.reserve_exact(1);
+
+            mem_growth = if had_allocated {
+                IndexBlock::STACK_SIZE
+            } else {
+                // Nothing is allocated until the first block is added.
+                // When that happens, the heap allocation has to grow by the size of the block
+                // as well as the size of the thin vector head (i.e. length and capacity).
```
gdesmott, line 634: Should we not account for the header size as well then in the `had_allocated` block if we just did the first allocation?

LukeMathWalker: If we had allocated before, the header is already there, so it shouldn't be included in the memory growth for the block we are adding here.

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -870,7 +874,18 @@ impl<E: Encoder + DecodedBy> InvertedIndex<E> {
...
-        self.blocks.shrink_to_fit();
+        // Remove excess capacity from the blocks vector.
+        {
+            let had_allocated = self.blocks.has_allocated();
+            self.blocks.shrink_to_fit();
+            // If we got rid of the heap block buffer entirely, we have also freed the memory occupied
+            // by the thin vec header (4 bytes to track length, 4 bytes to track capacity).
+            // That hasn't been accounted for yet, so we add it to the bytes freed now.
+            if !self.blocks.has_allocated() && had_allocated {
+                info.bytes_freed += 4 + 4;
```
gdesmott: Would it make sense to retrieve the header size from a `const` method of the vector? Less magic numbers around and easier to update if we ever need to.

LukeMathWalker: Good point, done in 537a4c9d9

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -482,7 +487,7 @@ impl<E: Encoder> InvertedIndex<E> {
     /// Create a new inverted index from the given blocks and encoder. The blocks are expected to not
     /// contain duplicate entries and be ordered by document ID.
     #[cfg(test)]
-    fn from_blocks(flags: IndexFlags, blocks: Vec<IndexBlock>) -> Self {
+    fn from_blocks(flags: IndexFlags, blocks: LowMemoryThinVec<IndexBlock, u32>) -> Self {
```
GuyAv46:
```suggestion
    fn from_blocks(flags: IndexFlags, blocks: LowMemoryThinVec<IndexBlock, BlockCapacity>) -> Self {
```

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -608,23 +611,34 @@ impl<E: Encoder> InvertedIndex<E> {
...
+
+            mem_growth = if had_allocated {
+                IndexBlock::STACK_SIZE
+            } else {
+                // Nothing is allocated until the first block is added.
+                // When that happens, the heap allocation has to grow by the size of the block
+                // as well as the size of the thin vector head (i.e. length and capacity).
+                self.blocks.mem_usage()
+            };
         }
```
GuyAv46: consider this style
```suggestion
        let mem_growth = if self.blocks.len() == self.blocks.capacity() {
self.blocks.reserve_exact(1);

            if had_allocated {
                IndexBlock::STACK_SIZE
            } else {
                // Nothing is allocated until the first block is added.
                // When that happens, the heap allocation has to grow by the size of the block
                // as well as the size of the thin vector head (i.e. length and capacity).
                self.blocks.mem_usage()
            }
         } else {
           0
         };
```

PR #8045, opened 2026-01-14
MOD-13306: Support disk deletion flow

raz-mon:
Fixes the disk deletion flow to update the statistics as well.
Practically adds the disk deletion flow to the `IndexSpec_DeleteDoc_Unsafe` function, which is now branched between RAM and disk based indexes.

---

> [!NOTE]
> Introduces a unified deletion path that supports disk-backed indexes and ensures accurate statistics.
> 
> - Refactors `IndexSpec_DeleteDoc_Unsafe` to handle both RAM and disk: looks up/deletes by key on disk, captures deleted `docLen` and `docId`, updates `numDocuments`/`totalDocsLen`, triggers GC, and cleans VecSim/Geometry; `IndexSpec_DeleteDoc` simplified to always lock and call the unsafe variant.
> - Extends disk API: `SearchDisk_PutDocument` now outputs `oldLen`; `SearchDisk_DeleteDocument` now outputs `oldLen` and deleted `docId`; headers and wrappers updated (`search_disk.[ch]`, `search_disk_api.h`).
> - Updates indexer: on disk `put`, decrements stats for replaced docs via `oldLen`, and increments with the new doc’s length; sets `docId` accordingly.
> - Adjusts error cleanup in `document.c` to call the new `IndexSpec_DeleteDoc_Unsafe` signature.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a7086cbd50756706c09438b4e4466738e87c00d4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/search_disk.h
+++ b/src/search_disk.h
@@ -160,17 +160,6 @@ t_docId SearchDisk_PutDocument(RedisSearchDiskIndexSpec *handle, const char *key
 bool SearchDisk_GetDocumentMetadata(RedisSearchDiskIndexSpec *handle, t_docId docId, RSDocumentMetadata *dmd);
 
 /**
- * @brief Get the document ID for a given key
- *
- * @param handle Handle to the disk index spec.
- * @param key The key to get the document ID for.
- * @param len The length of the key.
- * @return the doc-id corresponding to `key` if one exists, 0 (invalid doc-id)
- * otherwise.
- */
-t_docId SearchDisk_GetId(RedisSearchDiskIndexSpec *handle, const char *key, size_t len);
-
-/**
  * @brief Check if a document ID is deleted
  *
  * @param handle Handle to the document table
--- a/src/search_disk_api.h
+++ b/src/search_disk_api.h
@@ -119,17 +119,6 @@ typedef struct DocTableDiskAPI {
   bool (*getDocumentMetadata)(RedisSearchDiskIndexSpec* handle, t_docId docId, RSDocumentMetadata* dmd, AllocateKeyCallback allocateKey);
 
   /**
-   * @brief Get the document ID for a given key
-   *
-   * @param handle Handle to the disk index spec.
-   * @param key The key to get the document ID for.
-   * @param len The length of the key.
-   * @return the doc-id corresponding to `key` if one exists, 0 (invalid doc-id)
-   * otherwise.
-   */
-  t_docId (*getId)(RedisSearchDiskIndexSpec *handle, const char *key, size_t len);
-
-  /**
    * @brief Gets the maximum document ID assigned in the index
    *
    * @param handle Handle to the document table
--- a/src/spec.c
+++ b/src/spec.c
@@ -3650,14 +3650,14 @@ int IndexSpec_UpdateDoc(IndexSpec *spec, RedisModuleCtx *ctx, RedisModuleString 
 
 void IndexSpec_DeleteDoc_Unsafe(IndexSpec *spec, RedisModuleCtx *ctx, RedisModuleString *key) {
   t_docId id = 0;
-  size_t docLen = 0;
+  uint32_t docLen = 0;
   if (isSpecOnDisk(spec)) {
     RS_LOG_ASSERT(spec->diskSpec, "disk handle is unexpectedly NULL");
     size_t len;
     const char *keyStr = RedisModule_StringPtrLen(key, &len);
 
     // Delete the document
-    SearchDisk_DeleteDocument(spec->diskSpec, keyStr, len, &docLen, &id); // TODO: Add the id. We need it for the below logic (vector deletion)
+    SearchDisk_DeleteDocument(spec->diskSpec, keyStr, len, &docLen, &id);
 
     if (id == 0) {
       // Nothing to delete
--- a/src/search_disk.c
+++ b/src/search_disk.c
@@ -100,11 +100,6 @@ bool SearchDisk_GetDocumentMetadata(RedisSearchDiskIndexSpec *handle, t_docId do
     return disk->docTable.getDocumentMetadata(handle, docId, dmd, &sdsnewlen);
 }
 
-t_docId SearchDisk_GetId(RedisSearchDiskIndexSpec *handle, const char *key, size_t len) {
-  RS_ASSERT(disk && handle);
-  return disk->docTable.getId(handle, key, len);
-}
-
 bool SearchDisk_DocIdDeleted(RedisSearchDiskIndexSpec *handle, t_docId docId) {
     RS_ASSERT(disk && handle);
     return disk->docTable.isDocIdDeleted(handle, docId);


src/search_disk.h
```diff
@@ -156,6 +156,17 @@ t_docId SearchDisk_PutDocument(RedisSearchDiskIndexSpec *handle, const char *key
...
+/**
+ * @brief Get the document ID for a given key
+ *
+ * @param handle Handle to the disk index spec.
+ * @param key The key to get the document ID for.
+ * @param len The length of the key.
+ * @return the doc-id corresponding to `key` if one exists, 0 (invalid doc-id)
+ * otherwise.
+ */
+t_docId SearchDisk_GetId(RedisSearchDiskIndexSpec *handle, const char *key, size_t len);
```
kei-nan: I don't think we need this function.

src/spec.c
```diff
@@ -3648,19 +3648,42 @@ int IndexSpec_UpdateDoc(IndexSpec *spec, RedisModuleCtx *ctx, RedisModuleString
...
-    if (spec->gc) {
-      GCContext_OnDelete(spec->gc);
+void IndexSpec_DeleteDoc_Unsafe(IndexSpec *spec, RedisModuleCtx *ctx, RedisModuleString *key) {
+  RSDocumentMetadata *md = NULL;
+  if (isSpecOnDisk(spec)) {
+    // TODO: Remove once metadata API is used - MOD-13508
+    // Get the id
+    size_t len;
+    const char *keyStr = RedisModule_StringPtrLen(key, &len);
+    t_docId id = SearchDisk_GetId(spec->diskSpec, key, len);
```
kei-nan: I don't understand why we need this additional flow.
Can't we just call `SearchDisk_DeleteDocument` using the key?
It can return true or false to indicate if it deleted the key.

src/search_disk_api.h
```diff
@@ -115,6 +115,17 @@ typedef struct DocTableDiskAPI {
...
+  /**
+   * @brief Get the document ID for a given key
+   *
+   * @param handle Handle to the disk index spec.
+   * @param key The key to get the document ID for.
+   * @param len The length of the key.
+   * @return the doc-id corresponding to `key` if one exists, 0 (invalid doc-id)
+   * otherwise.
+   */
+  t_docId (*getId)(RedisSearchDiskIndexSpec *handle, const char *key, size_t len);
```
kei-nan: Don't think this is needed.

src/spec.c
```diff
@@ -3648,19 +3648,42 @@ int IndexSpec_UpdateDoc(IndexSpec *spec, RedisModuleCtx *ctx, RedisModuleString
...
-    RS_LOG_ASSERT(spec->stats.totalDocsLen >= md->docLen, "totalDocsLen is smaller than md->docLen");
-    spec->stats.totalDocsLen -= md->docLen;
-    DMD_Return(md);
-
-    // Increment the index's garbage collector's scanning frequency after document deletions
-    if (spec->gc) {
-      GCContext_OnDelete(spec->gc);
+void IndexSpec_DeleteDoc_Unsafe(IndexSpec *spec, RedisModuleCtx *ctx, RedisModuleString *key) {
+  RSDocumentMetadata *md = NULL;
+  if (isSpecOnDisk(spec)) {
```
kei-nan, line 3654: maybe worth checking if disk handle is valid.
`isSpecOnDisk` doesn't actually check anything in index object.

src/document.c
```diff
@@ -851,9 +851,7 @@ int Document_AddToIndexes(RSAddDocumentCtx *aCtx, RedisSearchCtx *sctx) {
   if (ourRv != REDISMODULE_OK) {
     // if a document did not load properly, it is deleted
```
JoanFM, line 852: should this comment be moved inside DeleteDoc?

src/spec.c
```diff
@@ -3669,47 +3692,27 @@ void IndexSpec_DeleteDoc_Unsafe(IndexSpec *spec, RedisModuleCtx *ctx, RedisModul
       if (spec->fields[i].types == INDEXFLD_T_VECTOR) {
         VecSimIndex *vecsim = openVectorIndex(spec->fields + i, DONT_CREATE_INDEX);
         if(!vecsim)
-          continue;
-        VecSimIndex_DeleteVector(vecsim, id);
+        continue;
```
JoanFM: review this, maybe the continue can go in the same line to if?

src/spec.c
```diff
@@ -3648,19 +3648,42 @@ int IndexSpec_UpdateDoc(IndexSpec *spec, RedisModuleCtx *ctx, RedisModuleString
...
+  RSDocumentMetadata *md = NULL;
+  if (isSpecOnDisk(spec)) {
+    // TODO: Remove once metadata API is used - MOD-13508
+    // Get the id
+    size_t len;
+    const char *keyStr = RedisModule_StringPtrLen(key, &len);
+    t_docId id = SearchDisk_GetId(spec->diskSpec, key, len);
+
+    // Get the dmd, if exists
+    RSDocumentMetadata* diskDmd = (RSDocumentMetadata *)rm_calloc(1, sizeof(RSDocumentMetadata));
```
JoanFM: does this need to be on the heap?

src/search_disk.h
raz-mon: It's in order to easily replace when we incorporate the metadata set/get API

src/spec.c
raz-mon: No, we need the dmd for the stats update, which is needed for scoring.

src/search_disk_api.h
raz-mon: It's in order to easily replace when we incorporate the metadata set/get API.
I can add more comments to remove it when it's incorporated if you think it's necessary.

src/spec.c
raz-mon: I didn't want to change delete, but for the sake of performance for the SKO I can alter the delete_document API

src/spec.c
raz-mon: I can add an assertion for that, but we should have a single uniform check we use for this in production.

src/spec.c
raz-mon: Just so we can later free with `DmdReturn`

src/spec.c
raz-mon: Sure, missed this tnx

src/document.c
raz-mon: No..

src/spec.c
raz-mon: Added.

src/search_disk.h
raz-mon: Removed to save the IO on delete (not the hot path but ok..)

src/search_disk_api.h
raz-mon: Removed

src/spec.c
raz-mon: All this was removed in order to optimize the flow

PR #8050, opened 2026-01-15
ci: Structure periodicity of benchmark runs

JoanFM:
Sets the following structurte for benchmark runs:

- Nightly Non Extended e2e benchmarks
- Periodic (every 3 days) extended e2e benchmarks
- Extra: Remove weeklfy follow and move link check to nightly flow


.github/workflows/event-nightly-e2e-benchmarks.yml
```diff
@@ -0,0 +1,13 @@
+name: Nightly E2E Benchmarks
```
alonre24, line 1: Should we add a new event or use the current nightly? 
Also, let's remove the benchmark flow from the workflow that is running after every merge to a version branch if we intent to replace that.
And if we do so, I would suggest running this flow for every version branch (we should probably create a matrix and run those in parallel)

PR #8051, opened 2026-01-15
MOD-7574: Output Current Thread IndexSpec Information On Crash

kei-nan:
Right now we only output the index name which is fine but it would be good to have index spec specific statistics for more helpful debugging of issues in the future.

---

> [!NOTE]
> Adds resilient crash reporting for the current thread and refactors INFO emitters to be safe in signal/lock-sensitive contexts.
> 
> - Output `current_thread` section with `run_time_ns` and FT.INFO-like index details via `IndexSpec_AddToInfo(ctx, spec, obfuscate, skip_unsafe_ops)`; skips unsafe ops and supports obfuscation (new `Obfuscate_Prefix`) and field/path obfuscation
> - Make cursor stats INFO-safe by introducing `CursorList_TryLock` and returning `cursor_stats{status=locked}` if locks can’t be acquired
> - Always expose GC and stopwords stats for INFO (remove `FTINFO_FOR_INFO_MODULES` guards); wire `GCContext_RenderStatsForInfo` and `ForkGC` `statsForInfoCb`
> - Enhance index INFO: guard vector/tag/text memory computations and prefix/field formatting under `skip_unsafe_ops`; switch indexing failures metric to `indexError.error_count`
> - Track per-thread index runtime (`SpecInfo.runningTime`), initialize on `CurrentThread_SetIndexSpec`, and print in crash report
> - Minor: adjust `module.c` to set current thread spec only when relevant; small iterator API change in stopwords
> - Tests: replace brittle log substring checks with ordered extraction/validation of crash report fields, including Rust panic path
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7e66d9c4eeb78b3abc1fb76a8d81dd468b6f9de2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -2990,12 +2990,6 @@ void IndexSpec_AddToInfo(RedisModuleInfoCtx *ctx, IndexSpec *sp, bool obfuscate,
   RedisModule_InfoBeginDictField(ctx, "index_failures");
   RedisModule_InfoAddFieldLongLong(ctx, "hash_indexing_failures", sp->stats.indexError.error_count);
   RedisModule_InfoAddFieldLongLong(ctx, "indexing", !!global_spec_scanner || sp->scan_in_progress);
-  if (!skip_unsafe_ops) {
-    // Skip when unsafe - calls RedisModule_DbSize which is not async-signal-safe
-    IndexesScanner *scanner = global_spec_scanner ? global_spec_scanner : sp->scanner;
-    double percent_indexed = IndexesScanner_IndexedPercent(ctx, scanner, sp);
-    RedisModule_InfoAddFieldDouble(ctx, "percent_indexed", percent_indexed);
-  }
   RedisModule_InfoEndDictField(ctx);
 
   // Garbage collector - safe to call, just reads struct fields


src/info/info_command.c
```diff
@@ -334,6 +334,37 @@ void fillReplyWithIndexInfo(RedisSearchCtx* sctx, RedisModule_Reply *reply, bool
...
 
+/**
+ * Crash-safe version of fillReplyWithIndexInfo.
+ * Outputs basic index info without allocations or locks.
+ */
+void fillReplyWithIndexInfoCrashSafe(const IndexSpec *sp, RedisModuleInfoCtx *info_ctx, bool obfuscate) {
+  if (!sp) return;
+
+  // Index name - use existing strings, no allocation
+  const char* specName = obfuscate ? sp->obfuscatedName : sp->specName;
```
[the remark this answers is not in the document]
JoanFM: @kei-nan this seems important.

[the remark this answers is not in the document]
kei-nan: yep, will fix

src/info/info_command.c
```diff
@@ -334,6 +334,37 @@ void fillReplyWithIndexInfo(RedisSearchCtx* sctx, RedisModule_Reply *reply, bool
   RedisModule_Reply_MapEnd(reply); // top
 }
 
+/**
+ * Crash-safe version of fillReplyWithIndexInfo.
+ * Outputs basic index info without allocations or locks.
+ */
+void IndexInfoCrashSafe(const IndexSpec *sp, RedisModuleInfoCtx *info_ctx, bool obfuscate) {
```
raz-mon:
```suggestion
void fillReplyWithIndexInfoCrash(const IndexSpec *sp, RedisModuleInfoCtx *info_ctx, bool obfuscate) {
```

src/info/info_command.c
```diff
@@ -334,6 +334,37 @@ void fillReplyWithIndexInfo(RedisSearchCtx* sctx, RedisModule_Reply *reply, bool
...
+  if (!sp) return;
+
+  // Index name - use existing strings, no allocation
+  const char* indexName = obfuscate ? sp->obfuscatedName : HiddenString_GetUnsafe(sp->specName, NULL);
+  if (indexName) {
+    RedisModule_InfoAddFieldCString(info_ctx, "index_name", (char*)indexName);
+  }
+
```
raz-mon: Do we want to print some info of the schema fields?

src/info/info_command.c
```diff
@@ -334,6 +334,37 @@ void fillReplyWithIndexInfo(RedisSearchCtx* sctx, RedisModule_Reply *reply, bool
...
+  // Basic statistics - just reading integers
+  RedisModule_InfoAddFieldLongLong(info_ctx, "num_docs", sp->stats.numDocuments);
+  RedisModule_InfoAddFieldLongLong(info_ctx, "max_doc_id", sp->docs.maxDocId);
+  RedisModule_InfoAddFieldLongLong(info_ctx, "num_terms", sp->stats.numTerms);
+  RedisModule_InfoAddFieldLongLong(info_ctx, "num_records", sp->stats.numRecords);
+
+  // Memory stats
+  RedisModule_InfoAddFieldDouble(info_ctx, "inverted_sz_mb", sp->stats.invertedSize / (float)0x100000);
+  RedisModule_InfoAddFieldDouble(info_ctx, "doc_table_size_mb", sp->docs.memsize / (float)0x100000);
+  RedisModule_InfoAddFieldDouble(info_ctx, "sortable_values_size_mb", sp->docs.sortablesSize / (float)0x100000);
```
raz-mon: Some fields are missing here, e.g., `vector_index_sz_mb`.
Is that on purpose?

tests/pytests/test_crash.py
```diff
@@ -42,7 +42,17 @@ def expect_query_crash_output(env, expected_fragments, crash_in_rust=False):
...
+        expect_query_crash_output(env, [
+            "search_current_thread",
+            "search_index_name:idx",
+            "search_num_docs:",
+            "search_max_doc_id:",
+            "search_num_terms:",
+            "search_inverted_sz_mb:",
+            "search_indexing:",
+            "search_num_fields:",
+            "search_run_time_ns:",
```
raz-mon: What are we testing here?

tests/pytests/test_crash.py
```diff
@@ -58,7 +68,10 @@ def test_query_thread_crash_with_rust_panic():
                 # The panic message
                 'A panic occurred in the Rust code panic.payload="Crash in Rust code"',
                 "search_current_thread",
-                "search_index:idx",
+                "search_index_name:idx",
+                "search_num_docs:",
+                "search_inverted_sz_mb:",
+                "search_run_time_ns:",
```
raz-mon: Why are there no values in the expected output?

src/info/info_command.c
kei-nan: as long as we don't allocate or lock stuff, what did you have in mind?

src/info/info_command.c
kei-nan: no, just tried to have a bare minimum of things
We can add stuff but at some point would like to stop
Don't want this PR to blow up.

tests/pytests/test_crash.py
kei-nan: that these strings appear in the crash report.

src/info/info_command.c
kei-nan: I prefer the current name

tests/pytests/test_crash.py
kei-nan: We are checking the existence of the fields, not their values.
I think verifying the values is going to really complicate the test, both to in logic and maintainability.

raz-mon: We should at least check that they make sense, e.g., `> 0`

tests/pytests/test_crash.py
raz-mon: We should at least check that they make sense, e.g., `> 0`

src/info/info_command.c
raz-mon: There are some more useful fields that won't blow it up, let's add them

src/info/info_command.c
raz-mon: Perhaps field types, whatever can help and doesn't require forbidden operations

src/info/info_command.c
raz-mon: OK about the name, but I also think this isn't the right place for this function. It should be in `redis_info.c`

kei-nan: will move it

kei-nan: in the end found an alternate function I could use, removed this one.

src/info/info_command.c
kei-nan: going to use an existing index spec function in the end.

src/cursor.c
```diff
@@ -362,10 +366,27 @@ void Cursors_RenderStats(CursorList *cl, CursorList *cl_coord, const IndexSpec *
   CursorList_Unlock(cl);
 }
 
-#ifdef FTINFO_FOR_INFO_MODULES
 void Cursors_RenderStatsForInfo(CursorList *cl, CursorList *cl_coord, const IndexSpec *spec, RedisModuleInfoCtx *ctx) {
-  CursorList_Lock(cl);
+  // pthread_mutex_trylock returns 0 on success, non-zero on failure
+  int lock_result = CursorList_TryLock(cl);
```
raz-mon, line 371: Is this ok to call when after a crash?
You mention specifically "no locks"

src/info/info_redis/info_redis.c
```diff
@@ -381,9 +381,15 @@ void AddToInfo_CurrentThread(RedisModuleInfoCtx *ctx) {
     if (!spec) {
       RedisModule_InfoAddFieldCString(ctx, "index", specInfo->specName ? specInfo->specName : "n/a");
     } else {
-      RedisModule_InfoAddFieldCString(ctx, "index", IndexSpec_FormatName(spec, RSGlobalConfig.hideUserDataFromLog));
-      // output FT.INFO
+      // Output FT.INFO in a crash-safe manner (no allocations, no locks)
+      // This includes the index name, so no need to output it separately
+      IndexSpec_AddToInfo(ctx, spec, RSGlobalConfig.hideUserDataFromLog, true);
     }
+    // Give us a sense of how long this thread was active on this index before we crashed.
```
raz-mon:
```suggestion
    // Gives us a sense of how long this thread was active on this index before we crashed.
```

src/spec.c
```diff
@@ -2886,16 +2889,24 @@ void IndexSpec_AddToInfo(RedisModuleInfoCtx *ctx, IndexSpec *sp) {
...
-      prefixes = array_ensure_append_n(prefixes, "\",", 2);
+  if (num_prefixes) {
+    const char *first_prefix = HiddenUnicodeString_GetUnsafe(rule->prefixes[0], NULL);
+    if (first_prefix && first_prefix[0] != '\0') {
+      if (!skip_unsafe_ops) {
+        // Skip when unsafe operations should be avoided (e.g., in signal handler) due to memory allocations
+        arrayof(char) prefixes = array_new(char, 512);
+        for (int i = 0; i < num_prefixes; ++i) {
+          const char *prefix = HiddenUnicodeString_GetUnsafe(rule->prefixes[i], NULL);
+          const char *prefix_to_use = obfuscate ? Obfuscate_Text(prefix) : prefix;
```
raz-mon: Perhaps let's use a different obfuscation API here, that will just return "***" or `*redacted*` like Redis uses.

Out of scope - was there a reason we didn't do that in general for our all our obfuscations?

src/spec.h
```diff
@@ -549,12 +549,11 @@ void IndexesScanner_Cancel(struct IndexesScanner *scanner);
 void IndexesScanner_ResetProgression(struct IndexesScanner *scanner);
 
 void IndexSpec_ScanAndReindex(RedisModuleCtx *ctx, StrongRef ref);
-#ifdef FTINFO_FOR_INFO_MODULES
 /**
  * Exposing all the fields of the index to INFO command.
+ * @param skip_unsafe_ops - if true, skips operations unsafe in signal handler context (allocations, locks)
```
raz-mon, line 557: If you're adding this param's doc, add the rest as well

tests/pytests/test_crash.py
```diff
@@ -15,53 +15,141 @@ def passStopProcess(self, *args, **kwargs):
...
+        "search_run_time_ns:",
+    ])
+
+    # Verify all fragments were found
+    for fragment, value in results.items():
+        env.assertIsNotNone(value, message=f"Fragment '{fragment}' not found in crash log")
+
+    # Verify specific values
+    # Empty index should have 0 documents
+    env.assertEqual(results["search_number_of_docs:"], "0")
```
raz-mon: Let's add a test with a populated index, or populate this one

src/cursor.c
kei-nan: the no locking is to avoid a deadlock,
if we only try to lock and handle the failure then deadlock is not an option.

src/spec.c
kei-nan: I think there was a long discussion on how to obfuscate our data.
Points were raised that we might want to have the same obfuscated string for the same piece of data - e.g like we do for index name - so we can still identify the index when it appears in the logs.
The same can be said on field names.

Beyond that we talked about strings general and that we might want to identify the string type, e.g if it was a number or a text string.

Anyhow we can maybe change the logic of Obfuscate_Text to do what you wanted but it is out of scope for this PR IMHO.

src/cursor.c
```diff
@@ -362,10 +366,27 @@ void Cursors_RenderStats(CursorList *cl, CursorList *cl_coord, const IndexSpec *
...
+  int lock_coord_result = CursorList_TryLock(cl_coord);
```
GuyAv46, line 372: Why not simply hold both? Why do we want to maybe fail rendering the stats?

src/module.c
```diff
@@ -814,11 +814,11 @@ int SynDumpCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
     return RedisModule_ReplyWithError(ctx, NOPERM_ERR);
   }
 
-  CurrentThread_SetIndexSpec(ref);
-
   if (!sp->smap) {
     return RedisModule_ReplyWithMap(ctx, 0);
   }
+  
+  CurrentThread_SetIndexSpec(ref);
```
GuyAv46, line 821: Is this a bug fix?

src/spec.c
```diff
@@ -2872,8 +2869,14 @@ void IndexSpec_AddToInfo(RedisModuleInfoCtx *ctx, IndexSpec *sp) {
...
   SchemaRule *rule = sp->rule;
   RedisModule_InfoAddFieldCString(ctx, "type", (char*)DocumentType_ToString(rule->type));
-  if (rule->filter_exp_str)
-    RedisModule_InfoAddFieldCString(ctx, "filter", rule->filter_exp_str);
+  if (rule->filter_exp_str) {
+    const char *filter = HiddenString_GetUnsafe(rule->filter_exp_str, NULL);
+    if (obfuscate) {
+      RedisModule_InfoAddFieldCString(ctx, "filter", (char*)Obfuscate_Text(filter));
+    } else {
+      RedisModule_InfoAddFieldCString(ctx, "filter", (char*)filter);
```
GuyAv46: Remove unnecessary cast
```suggestion
    if (obfuscate) {
      RedisModule_InfoAddFieldCString(ctx, "filter", Obfuscate_Text(filter));
    } else {
      RedisModule_InfoAddFieldCString(ctx, "filter", filter);
```

src/spec.h
GuyAv46, line 557: For all the protected parts, should we give some simpler but some data?

src/info/info_redis/info_redis.c
kei-nan: will also move it to be before the index info.

src/cursor.c
raz-mon: IIUC this is not a function we should call in a crash, but this is best effort anyways so I'm fine with it

src/spec.c
raz-mon: I just think "Text" is quite weird to print here. This is not even a Text field.
I don't think it's critical, just weird, so up to you if you want to print something else here, e.g., "Prefix".

The more general thoughts of the obfuscation API behavior are of course not for this PR, just asking since it's weird to me and is not aligned with Redis.

src/cursor.c
kei-nan: because this function can be called while we are crashing and we don't want to potentially deadlock when outputting the cursor stats.

src/module.c
kei-nan: yes

src/spec.h
kei-nan: Can you give a suggestion for such a case?
What is a simpler form of data for inverted index size for tags for example?

src/spec.c
kei-nan: I agree it is a bit weird to print Text.
Added `Obfuscate_Prefix` to output "Prefix" instead, I guess that is slightly better.

src/spec.c
```diff
@@ -2906,8 +2917,8 @@ void IndexSpec_AddToInfo(RedisModuleInfoCtx *ctx, IndexSpec *sp) {
     sprintf(title, "%s_%d", "field", (i+1));
     RedisModule_InfoBeginDictField(ctx, title);
 
-    RedisModule_InfoAddFieldCString(ctx, "identifier", fs->path);
-    RedisModule_InfoAddFieldCString(ctx, "attribute", fs->name);
+    RedisModule_InfoAddFieldCString(ctx, "identifier", FieldSpec_FormatPath(fs, obfuscate));
+    RedisModule_InfoAddFieldCString(ctx, "attribute", FieldSpec_FormatName(fs, obfuscate));
```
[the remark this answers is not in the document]
kei-nan: will switch to using `Obfuscate_FieldPath`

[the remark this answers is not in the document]
kei-nan: will switch to using `Obfuscate_Field`

src/cursor.c
GuyAv46: OK, but don't we use it elsewhere? It may fail if some thread holds the cursor lists at any time, and then it won't show any meaningful data

src/module.c
GuyAv46: Do you think we should split it from this PR then?

src/spec.h
GuyAv46: I wonder if some of these sections could be partially filled or estimated, instead of not having any of them

src/cursor.c
kei-nan: You can see it was previously wrapped with `FTINFO_FOR_INFO_MODULES` macro.
So up until now it was not used.in production.

src/module.c
kei-nan: okay, will create a different PR to fix it

src/spec.h
kei-nan: I feel like that is too big of a change to be done in this PR.
So I think it is out of scope.
We can create a followup task if you want.

src/cursor.c
GuyAv46: I see, so it's more of a typo than a bug until now 👍🏼

GuyAv46 (approved):

PR #8053, opened 2026-01-15
[MOD-12284] Implement RLookup load_rule_fields()

memark:
Implement `RLookup` `load_rule_fields()` (+ FFI function).

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches unsafe FFI boundaries and pointer/lifetime handling while wiring through Redis loading (`loadIndividualKeys`), so mis-specified lifetimes or pointer assumptions could cause crashes despite added tests.
> 
> **Overview**
> Implements `RLookup::load_rule_fields()` and exposes it via a new C API `RLookup_LoadRuleFields`, enabling populating an `RLookupRow` by loading the schema rule’s filter fields (using `IndexSpec`/`FieldSpec` metadata when available) from Redis.
> 
> Refactors `HiddenStringRef`/`FieldSpec` APIs to be lifetime-aware (`HiddenStringRef<'a>`) and to consume the wrapper when extracting the underlying `CStr`, and adjusts `KeyList::push`’s return lifetime to unblock pushing keys created from the spec; adds unit coverage for key creation from `IndexSpec`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c5d83bf84fd74ca7578e2a653a290d4f831759a5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rlookup/src/lookup.rs
+++ b/src/redisearch_rs/rlookup/src/lookup.rs
@@ -524,8 +524,11 @@ mod tests {
 
     use crate::mock::{array_free, array_new};
 
-    use std::ptr::{self, NonNull};
     use std::{ffi::CString, mem::MaybeUninit};
+    use std::{
+        ffi::c_char,
+        ptr::{self, NonNull},
+    };
 
     use enumflags2::make_bitflags;
     use pretty_assertions::assert_eq;
@@ -1457,7 +1460,7 @@ mod tests {
         }
     }
 
-    fn filter_fields_array(filter_fields: &[&CStr]) -> *mut *mut i8 {
+    fn filter_fields_array(filter_fields: &[&CStr]) -> *mut *mut c_char {
         let temp = filter_fields
             .iter()
             .map(|ff| ff.as_ptr().cast_mut())


src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
```diff
@@ -437,3 +439,47 @@ pub unsafe extern "C" fn RLookup_Cleanup(lookup: Option<NonNull<RLookup<'_>>>) {
...
+///     1. The entire memory range of this `CStr` must be contained within a single allocation!
+///     2. `key` must be non-null even for a zero-length cstr.
+/// 7. The nul terminator must be within `isize::MAX` from `key`
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RLookup_LoadRuleFields(
+    module_ctx: Option<NonNull<ffi::RedisModuleCtx>>,
+    lookup: Option<NonNull<RLookup<'_>>>,
+    dst_row: Option<NonNull<RLookupRow>>,
```
JonasKruckenberg: these should have the same lifetime, so both should be `'_`

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
```diff
@@ -437,3 +439,47 @@ pub unsafe extern "C" fn RLookup_Cleanup(lookup: Option<NonNull<RLookup<'_>>>) {
     // Safety: ensured by caller (1.,2.)
     unsafe { lookup.unwrap().drop_in_place() };
 }
+
+/// Initialize the lookup with fields from hash.
```
JonasKruckenberg:
```suggestion
/// Initialize the lookup with fields from a redis hash.
```

src/redisearch_rs/rlookup/src/lookup/key_list.rs
```diff
@@ -56,7 +56,10 @@ impl<'a> KeyList<'a> {
     /// Insert a `RLookupKey` into this `KeyList` and return a mutable reference to it.
     ///
     /// The key will be owned by the list and freed when dropping the list.
-    pub(crate) fn push(&mut self, mut key: RLookupKey<'a>) -> Pin<&mut RLookupKey<'a>> {
+    pub(crate) fn push<'b>(&mut self, mut key: RLookupKey<'a>) -> Pin<&'b mut RLookupKey<'a>>
```
JonasKruckenberg, line 61: yeah this ain't great 😅 can't wait to get rid of this.

JonasKruckenberg: actually, could you add a TODO-style comment that calls out this for being weird and that we should remove it in the refactor?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -411,18 +418,135 @@ impl<'a> RLookup<'a> {
...
+        key: &CStr,
+    ) -> i32 {
+        let keys = create_keys_from_spec(self, index_spec);
+        let pushed_keys = keys.into_iter().map(|k| self.keys.push(k)).collect();
+        load_many_keys(self, module_ctx, dst_row, index_spec, key, pushed_keys)
+    }
+}
+
+fn create_keys_from_spec<'a>(
+    lookup: &mut RLookup,
```
JonasKruckenberg: should also be `<'a>` right?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -411,18 +418,135 @@ impl<'a> RLookup<'a> {
...
+        .iter()
+        .zip(rule.filter_fields())
+        .map(|(&index, filter_field)| {
+            create_key_from_data(lookup, index, filter_field, field_specs)
+        })
+        .collect::<Vec<_>>()
+}
+
+fn create_key_from_data<'a>(
+    lookup: &mut RLookup,
```
JonasKruckenberg: same here should be tied to the same lifetime

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -411,18 +418,135 @@ impl<'a> RLookup<'a> {
...
+        let field_spec = &field_specs[index];
+        let field_name = field_spec.field_name().get_secret_value();
+        let path = field_spec.field_path().get_secret_value();
+        let path = Cow::Borrowed(path);
+
+        RLookupKey::new_with_path(lookup, field_name, path, RLookupKeyFlags::empty())
+    }
+}
+
+fn load_many_keys(
```
JonasKruckenberg: do we not need explicit lifetimes here?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -411,18 +418,135 @@ impl<'a> RLookup<'a> {
...
+            let k = unsafe { Pin::into_inner_unchecked(k.into_ref()) };
+            ptr::from_ref(k).cast::<ffi::RLookupKey>()
+        })
+        .collect::<Vec<_>>();
+
+    let mut sctx = create_redis_search_ctx(module_ctx);
+
+    // TODO: Add comment explaining why "for real" temp.
+    // Safety: ...
+    let mut status = unsafe { mem::transmute(QueryError::default()) };
```
JonasKruckenberg: don't forget about this before declaring this as ready 😉

src/redisearch_rs/rlookup/src/lookup.rs
memark: In this case actually no. But I can add it for consistency if you prefer?

src/redisearch_rs/rlookup/src/lookup.rs
memark: Thanks, I won't. Got a couple more I need to write before it's ready.

src/redisearch_rs/rlookup/src/lookup.rs
JonasKruckenberg: yeah lets add them too

src/redisearch_rs/rlookup/src/lookup/key_list.rs
```diff
@@ -56,7 +56,13 @@ impl<'a> KeyList<'a> {
...
     ///
     /// The key will be owned by the list and freed when dropping the list.
-    pub(crate) fn push(&mut self, mut key: RLookupKey<'a>) -> Pin<&mut RLookupKey<'a>> {
+    //
+    // The different 'a and 'b lifetimes are really a borrow-checker hack
+    // and should be removed as soon as we refactor this code.
+    pub(crate) fn push<'b>(&mut self, mut key: RLookupKey<'a>) -> Pin<&'b mut RLookupKey<'a>>
+    where
+        'a: 'b,
+    {
```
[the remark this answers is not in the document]
memark: This is by design and good enough for now.

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,18 +423,130 @@ impl<'a> RLookup<'a> {
...
+    index: i32,
+    filter_field: &'a CStr,
+    field_specs: &'a [FieldSpec],
+) -> RLookupKey<'a> {
+    const NO_MATCH: i32 = -1;
+    if NO_MATCH == index {
+        RLookupKey::new(lookup, filter_field, RLookupKeyFlags::empty())
+    } else {
+        let index = usize::try_from(index).expect("index must be positive and fit into usize");
+        let field_spec = &field_specs[index];
```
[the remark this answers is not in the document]
memark: This is by design.

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,18 +423,130 @@ impl<'a> RLookup<'a> {
...
+    key: &CStr,
+    keys: Vec<Pin<&mut RLookupKey>>,
+) -> i32 {
+    let lookup = ptr::from_mut(lookup).cast::<ffi::RLookup>();
+    let dst_row = ptr::from_mut(dst_row).cast::<ffi::RLookupRow>();
+
+    let mut keys = keys
+        .into_iter()
+        .map(|k| {
+            // Safety: ...
```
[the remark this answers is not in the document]
memark: Added safety docs.

src/redisearch_rs/rlookup/src/hidden_string_ref.rs
```diff
@@ -31,7 +31,9 @@ impl HiddenStringRef {
     ///
     /// This is safe **only if** the C function returns a pointer that stays valid
     /// for at least the lifetime of `&self`, and the memory contains a NUL at `len`.
-    pub fn get_secret_value(&self) -> &CStr {
+    ///
+    /// This consumes the `HiddenStringRef` and can only be called once.
+    pub fn into_secret_value<'a>(self) -> &'a CStr {
```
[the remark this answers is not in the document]
memark: Not sure what to do here. Any ideas?

src/redisearch_rs/rlookup/src/lookup/key_list.rs
```diff
@@ -56,7 +56,13 @@ impl<'a> KeyList<'a> {
     /// Insert a `RLookupKey` into this `KeyList` and return a mutable reference to it.
     ///
     /// The key will be owned by the list and freed when dropping the list.
-    pub(crate) fn push(&mut self, mut key: RLookupKey<'a>) -> Pin<&mut RLookupKey<'a>> {
+    //
+    // The different 'a and 'b lifetimes are really a borrow-checker hack
+    // and should be removed as soon as we refactor this code.
```
JonasKruckenberg:
```suggestion
    // TODO remove the 'a and 'b lifetimes borrow-checker hack when we refactor this code.
```

lets have a TODO here so its easy to grep for

src/redisearch_rs/rlookup/src/hidden_string_ref.rs
JonasKruckenberg:
```suggestion
pub struct HiddenStringRef<'a>(NonNull<ffi::HiddenString>, PhantomData<&'a ffi::HiddenString>);
```

to make this correctly behave as a reference and fix the lifetime issue below

src/redisearch_rs/rlookup/src/hidden_string_ref.rs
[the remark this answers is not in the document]
JonasKruckenberg: this needs the lifetime of the ref to be tracked and inherited by this `&CStr` see above

src/redisearch_rs/rlookup/src/hidden_string_ref.rs
JonasKruckenberg:
```suggestion
#[derive(Debug, Clone, Copy)]
```

since this is an immutable ref you can also safely derive `Clone` and `Copy` for it (you can then also keep the `get_secret_value` call below because self is copy)

JonasKruckenberg, line 26:
```suggestion
    ///    This also applies to any of its subfields.
    /// 2. The pointed to `ffi::HiddenString` must not be mutated for the entire lifetime of the returned `HiddenStringRef`.
```

or something like this

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,18 +423,130 @@ impl<'a> RLookup<'a> {
...
+            ptr::from_ref(k).cast::<ffi::RLookupKey>()
+        })
+        .collect::<Vec<_>>();
+
+    let mut sctx = create_redis_search_ctx(module_ctx);
+
+    // Safety: The size and alignment of the Rust source struct matches the target C struct exactly. This is the only way we can initialize such a struct.
+    let mut status = unsafe {
+        mem::transmute::<query_error::QueryError, ffi::QueryError>(QueryError::default())
+    };
```
JonasKruckenberg: Did you check with redis this was meant to be actually handled? If in theory yes, but out of scope for this PR (which I think would be reasonable) we should keep the `TODO` comment here

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,18 +423,130 @@ impl<'a> RLookup<'a> {
...
+
+    let mut options = ffi::RLookupLoadOptions {
+        keys: keys.as_mut_ptr(),
+        nkeys: keys.len(),
+        sctx: &mut sctx,
```
JonasKruckenberg:
```suggestion
        sctx: ptr::from_mut(&mut sctx),
```

nitpick: this field is a pointer right? personally dislike the mutref -> pointer coercion so would prefer to see a `ptr::from_mut` here.

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,18 +423,130 @@ impl<'a> RLookup<'a> {
...
+        keyPtr: key.as_ptr(),
+        type_: index_spec.rule().type_(),
+        status: &mut status,
```
JonasKruckenberg:
```suggestion
        status: ptr::from_mut(&mut status),
```

same here

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,18 +423,130 @@ impl<'a> RLookup<'a> {
...
+    rule.filter_fields_index()
+        .iter()
+        .zip(rule.filter_fields())
+        .map(|(&index, filter_field)| {
+            create_key_from_data(lookup, index, filter_field, field_specs)
+        })
+        .collect::<Vec<_>>()
+}
+
+fn create_key_from_data<'a>(
```
JonasKruckenberg, line 463: what does `data` refer to here? the fields specs? If so, we should just name it that

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,18 +423,130 @@ impl<'a> RLookup<'a> {
...
+        let index = usize::try_from(index).expect("index must be positive and fit into usize");
+        let field_spec = &field_specs[index];
+        let field_name = field_spec.field_name().into_secret_value();
+        let path = field_spec.field_path().into_secret_value();
+
+        RLookupKey::new_with_path(lookup, field_name, path, RLookupKeyFlags::empty())
+    }
+}
+
+fn load_many_keys<'a>(
```
JonasKruckenberg: nitpick: in the porting PR I named this `load_specific_keys`. Don't have a particular preference for any name but we should ideally agree on one

src/redisearch_rs/rlookup/src/lookup.rs
memark: Yes, these struct instances are usually neither released nor reused.

What should we put in the TODO?

src/redisearch_rs/rlookup/src/lookup.rs
memark: It is. I don't mind, I'll accept your edit. (Is this something to put in the project guidelines?)

src/redisearch_rs/rlookup/src/lookup.rs
memark: `data` is indeed vague, and refers to all function arguments except the lookup. depending on the `index` value, the function uses the information from either `filter_field` or `field_specs`.

Would you prefer `create_key_from_filter_field_or_field_specs`?

src/redisearch_rs/rlookup/src/lookup.rs
memark: That sounds familiar. I don't have a strong opinion, I can change it.

memark: Nitpicks are good.

src/redisearch_rs/rlookup/src/hidden_string_ref.rs
memark: Done.

src/redisearch_rs/rlookup/src/lookup.rs
memark: This refers to the comment on the C side saying "TODO: handle errors".
Take a pointer, cast it etc.

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,18 +423,130 @@ impl<'a> RLookup<'a> {
...
+        type_: index_spec.rule().type_(),
+        status: &mut status,
+        forceLoad: true,
+        mode: ffi::RLookupLoadFlags_RLOOKUP_LOAD_KEYLIST,
+        dmd: ptr::null(),
+        forceString: false,
+    };
+
+    // Safety: All pointers passed to this function are non-null and properly aligned since we created them above in this function.
+    unsafe { ffi::loadIndividualKeys(lookup, dst_row, &mut options) }
```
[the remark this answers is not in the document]
memark: We'll remove the transmute and this becomes a non-issue.

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,19 +421,136 @@ impl<'a> RLookup<'a> {
...
+        let pushed_keys = keys.into_iter().map(|k| self.keys.push(k)).collect();
+        load_specific_keys(
+            self,
+            module_ctx,
+            dst_row,
+            index_spec,
+            key,
+            pushed_keys,
+            status,
+        )
```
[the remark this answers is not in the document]
memark: @JonasKruckenberg any thoughts?

(Btw, I love how new issues come up when just pushing the same code.)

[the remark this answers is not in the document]
kei-nan: should running miri catch this sort of thing?
is there a test we can add to prevent future code from causing the same lifetime issue?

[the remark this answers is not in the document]
JonasKruckenberg: This is not quite right though, yes the `KeyList::push` is very easy to abuse but its not unsound in the way we use it (because every key is its own pinned heap allocation). The bot is also wrong about Rusts UB model regarding mutable aliasing. Yes it is UB to _use_ or _produce_ overlapping mutable references but it's not UB to have code that _could produce_ overlapping mutable references.

This is fine in it's current form and @kei-nan clippy exists exactly for this reason, the fact that is doesn't complain is a good sign. 

Yet one more sign to remove the intrusive list asap...

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
```diff
@@ -448,6 +450,57 @@ pub unsafe extern "C" fn RLookup_Cleanup(lookup: Option<NonNull<RLookup<'_>>>) {
...
+/// 8. `status` must be a [valid], non-null pointer to an `ffi::QueryError` that is properly initialized.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RLookup_LoadRuleFields<'a>(
+    module_ctx: Option<NonNull<ffi::RedisModuleCtx>>,
+    lookup: Option<NonNull<RLookup<'a>>>,
+    dst_row: Option<NonNull<RLookupRow<'a>>>,
+    index_spec: Option<NonNull<ffi::IndexSpec>>,
+    key: *const c_char,
```
kei-nan, line 478: maybe we should pass the length of the key as well?
in case we don't want to rely on null delimiter?

src/redisearch_rs/rlookup/src/hidden_string_ref.rs
```diff
@@ -7,31 +7,40 @@
...
 
-use std::{ffi::CStr, ptr::NonNull};
+use std::{ffi::CStr, marker::PhantomData, ptr::NonNull};
 
 /// A safe wrapper around a non-null `ffi::HiddenString` reference.
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
 #[repr(transparent)]
-pub struct HiddenStringRef(NonNull<ffi::HiddenString>);
+pub struct HiddenStringRef<'a>(
```
kei-nan, line 15: why were these changes needed?
- are we okay we okay with copying and cloning this struct?
- why phantom data is needed?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,19 +421,136 @@ impl<'a> RLookup<'a> {
...
+    pub fn load_rule_fields(
+        &mut self,
+        module_ctx: &mut ffi::RedisModuleCtx,
+        dst_row: &mut RLookupRow<'a, value::RSValueFFI>,
+        index_spec: &'a IndexSpec,
+        key: &CStr,
+        status: &mut ffi::QueryError,
+    ) -> i32 {
+        let keys = create_keys_from_spec(self, index_spec);
+        let pushed_keys = keys.into_iter().map(|k| self.keys.push(k)).collect();
```
kei-nan, line 434: TBH not sure what is going on here.
`create_keys_from_spec` is a self function and we later also modify `self.keys`
seems a bit redundant, no?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,19 +421,136 @@ impl<'a> RLookup<'a> {
...
+    lookup: &mut RLookup<'a>,
+    index_spec: &'a IndexSpec,
+) -> Vec<RLookupKey<'a>> {
+    let rule = index_spec.rule();
+    let field_specs = index_spec.field_specs();
+    rule.filter_fields_index()
+        .iter()
+        .zip(rule.filter_fields())
+        .map(|(&index, filter_field)| {
+            create_key_from_data(lookup, index, filter_field, field_specs)
```
kei-nan, line 458: can we push to self.keys here?
also is there a way to reserve the space we will need in the vector?

src/redisearch_rs/rlookup/src/lookup/key_list.rs
```diff
@@ -56,7 +56,12 @@ impl<'a> KeyList<'a> {
     /// Insert a `RLookupKey` into this `KeyList` and return a mutable reference to it.
     ///
     /// The key will be owned by the list and freed when dropping the list.
-    pub(crate) fn push(&mut self, mut key: RLookupKey<'a>) -> Pin<&mut RLookupKey<'a>> {
+    //
+    // TODO remove the 'a and 'b lifetimes borrow-checker hack when we refactor this code.
```
kei-nan: refactor the code in what way?
not sure I understand why this was needed or how it will be fixed.
maybe worth improving the comment

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,19 +421,136 @@ impl<'a> RLookup<'a> {
...
+        .collect::<Vec<_>>();
+
+    let mut sctx = create_redis_search_ctx(module_ctx);
+
+    let mut options = ffi::RLookupLoadOptions {
+        keys: keys.as_mut_ptr(),
+        nkeys: keys.len(),
+        sctx: ptr::from_mut(&mut sctx),
+        keyPtr: key.as_ptr(),
+        type_: index_spec.rule().type_(),
```
kei-nan, line 508: why type_()?

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -416,19 +421,136 @@ impl<'a> RLookup<'a> {
...
+        dmd: ptr::null(),
+        forceString: false,
+    };
+
+    // Safety: All pointers passed to this function are non-null and properly aligned since we created them above in this function.
+    unsafe { ffi::loadIndividualKeys(lookup, dst_row, &mut options) }
+}
+
+const fn create_redis_search_ctx(module_ctx: *mut ffi::RedisModuleCtx) -> ffi::RedisSearchCtx {
+    ffi::RedisSearchCtx {
```
kei-nan: is this correct?
seems a bit risky to create a search context internally

src/redisearch_rs/rlookup/src/hidden_string_ref.rs
memark: >are we okay we okay with copying and cloning this struct?

@JonasKruckenberg suggested we add these derives. (The struct is never really cloned or copied atm.) Would you elaborate?

src/redisearch_rs/rlookup/src/lookup.rs
memark: `type` is a reserved word and cannot be used as an identifier. :/

memark:
> left some comments

Are you available to discuss the PR in a call? @kei-nan

src/redisearch_rs/rlookup/src/lookup.rs
memark: That I cannot judge. The search context was being created in the C version, I just ported that.

image

Would you suggest we pass it in instead?

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
memark: You could argue that, I've just ported the current C version.

image

Do we want to change the signature? (I guess it depends on how it's being called?)

src/redisearch_rs/rlookup/src/lookup.rs
memark: Let's talk about this on the call.

src/redisearch_rs/rlookup/src/lookup/key_list.rs
memark: Let's talk about this on the call.

src/redisearch_rs/rlookup/src/lookup.rs
memark: Let's talk about this on the call.

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
kei-nan: I think we can keep as is for now

src/redisearch_rs/rlookup/src/hidden_string_ref.rs
kei-nan: and why do we need the phantom data if we are using the type in NonNull?

src/redisearch_rs/rlookup/src/lookup.rs
kei-nan: I guess for now we can keep as is and it will be refactored later on.

src/redisearch_rs/rlookup/src/lookup.rs
kei-nan: I guess for now we can keep as is

src/redisearch_rs/rlookup/src/lookup/key_list.rs
kei-nan: don't think we ended up discussing this point.
anyhow if the comment can be expanded a bit to give some more context it would be great.

src/redisearch_rs/rlookup/src/hidden_string_ref.rs
memark: `PhantomData` is just a Rust trick to attach a lifetime to a value/object that is _not_ a reference. It helps the compiler, but doesn't affect the logic.

kei-nan: but can't the compiler use the lifetime of NonNull in this case, why do we need to attach a lifetime using phantom data in this case?

memark: `Clone` and `Copy` are both ok because the structs is a wrapper around an immutable pointer. They are not needed per se right `now,` but it's good habit to add them when possible, as it makes it easier for calling code.

src/redisearch_rs/rlookup/src/lookup/key_list.rs
memark: This ties back into the fact that the KeyList in Rust is overly complex. The comment basically means that we want to fix this when we refactor all of it. I have now added a reference to the Jira ticket we created earlier today.

src/redisearch_rs/rlookup/src/lookup.rs
memark: That's what we decided to do.

PR #8068, opened 2026-01-18
Fix edge case in clusterset - [MOD-13562]

GuyAv46:

Fix an edge case in the Enterprise topology helper `CLUSTERSET`.

We now collect all the parsed shards into the temporary dictionary, and filter those we don't care for when building the final topology struct

#### Mark if applicable

---

> [!NOTE]
> Improves Enterprise `CLUSTERSET` topology handling and robustness.
> 
> - Collects all parsed shards, then builds topology with only `MASTER` shards that have `SLOTRANGE`; introduces `RLShard.isMaster`
> - Allows `MYID` to be absent from the final topology when the local node has no slots or is a replica; identifies this via stored `MYID` offset
> - Fixes `UNIXADDR` parsing to avoid consuming `MASTER`/`SHARD` tokens; enforces single `UNIXADDR` and consistent per-shard values
> - Relaxes slot-range merging to permit shard reappearance without new ranges; validates ordering only when a new range is provided
> - In `SetClusterCommand`, when `my_shard_idx == UINT32_MAX`, updates cluster with an empty local slots array instead of failing or setting local id
> - Updates error messages and adds cpp/py tests covering missing `UNIXADDR`, multiple `UNIXADDR`, and cases where the local shard is empty or a replica
> - Documents `RedisEnterprise_ParseTopology` return/`my_shard_idx` behavior in `redise.h`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 28f43863e3c5d4d162d1a1541b2f0daaae279126. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test.py
+++ b/tests/pytests/test.py
@@ -4314,10 +4314,11 @@ def test_cluster_set_multiple_slots(env: Env):
     env.expect('SEARCH.CLUSTERINFO').equal(expected)
 
 @skip(cluster=False) # this test is only relevant on cluster
-def test_cluster_set_no_local_slots(env: Env):
+def test_cluster_set_myself_excluded(env: Env):
     env.cmd(debug_cmd(), 'PAUSE_TOPOLOGY_UPDATER')
     env.cmd(config_cmd(), 'SET', 'TOPOLOGY_VALIDATION_TIMEOUT', 1)
 
+    # Set two shards, one with all the slots, and one without any slots
     env.expect(
         'SEARCH.CLUSTERSET',
             'HASHFUNC', 'CRC16',
@@ -4334,6 +4335,30 @@ def test_cluster_set_no_local_slots(env: Env):
     ).ok()
 
     # Expect only the shard with slots to be listed
+    expected = [
+        'num_partitions', 1,
+        'cluster_type', 'redis_oss',
+        'shards', [['slots', [0, 16383], 'id', '2', 'host', 'localhost', 'port', 7002]],
+    ]
+    env.expect('SEARCH.CLUSTERINFO').equal(expected)
+
+    # Set two shards, one master and myself as replica
+    env.expect(
+        'SEARCH.CLUSTERSET',
+            'HASHFUNC', 'CRC16',
+            'NUMSLOTS', '16384',
+            'MYID', '1',
+            'RANGES', '2',
+            'SHARD', '1',
+            'ADDR', 'localhost:7001',
+            'SLOTRANGE', '0', '16383',
+            'SHARD', '2',
+            'ADDR', 'localhost:7002',
+            'MASTER',
+            'SLOTRANGE', '0', '16383'
+    ).ok()
+
+    # Expect only the master shard to be listed
     expected = [
         'num_partitions', 1,
         'cluster_type', 'redis_oss',


tests/cpptests/coord_tests/test_cpp_clusterset.cpp
```diff
@@ -738,15 +738,15 @@ TEST_F(ClusterSetTest, Error_MissingUNIXADDRValue) {
         "MYID", "shard1",
         "RANGES", "1",
         "SHARD", "shard1", "SLOTRANGE", "0", "16383",
-        "ADDR", "127.0.0.1:6379", "UNIXADDR", "MASTER"
+        "ADDR", "127.0.0.1:6379", "MASTER", "UNIXADDR"
```
alonre24: Why changed?

tests/cpptests/coord_tests/test_cpp_clusterset.cpp
```diff
@@ -994,3 +994,30 @@ TEST_F(ClusterSetTest, EdgeCase_ManyShards) {
 
     MRClusterTopology_Free(topo);
 }
+
+TEST_F(ClusterSetTest, EdgeCase_LocalShardEmpty) {
```
alonre24, line 998: Should also we add test in which local shard is a replica? Or in which case expect the same command

src/module.c
```diff
@@ -4022,15 +4022,23 @@ int SetClusterCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
-  MR_UpdateTopology(topo, my_slots);
+    // send the topology to the cluster
+    MR_UpdateTopology(topo, my_slots);
+
+  } else {
+    // Valid topology but this node is not part of it.
+    // We cannot pass NULL as local slots, so we pass an empty slot array.
+    static const RedisModuleSlotRangeArray empty_slots = {0, {{0, 0}}};
+    MR_UpdateTopology(topo, &empty_slots);
+  }
```
alonre24, line 4041: Can we cover this new code in a flow test?

tests/cpptests/coord_tests/test_cpp_clusterset.cpp
GuyAv46: I’ll add it

src/module.c
GuyAv46: Of course

tests/cpptests/coord_tests/test_cpp_clusterset.cpp
GuyAv46: There’s a pitfall in the API as we don’t validate (or use) the Unix address, so the MASTER string is interpreted as the address

GuyAv46: Maybe we should check it’s not `MASTER` or `SHARD` specifically

src/coord/rmr/redise.c
```diff
@@ -270,15 +258,18 @@ MRClusterTopology *RedisEnterprise_ParseTopology(RedisModuleCtx *ctx, RedisModul
           goto error;
         }
       }
-      RS_ASSERT(array_len(sh->slotRanges) == 1);
-      // Verify slot range starts past existing ones
-      if (array_tail(existing_shard->slotRanges).end + 1 >= sh->slotRanges[0].start) {
-        ERROR_FMT("SLOTRANGE out of order for shard `%s`", sh->node.id);
```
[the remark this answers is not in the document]
GuyAv46: `MASTER` flag (and the rest of the connection info) should be present at the first instance

PR #8076, opened 2026-01-19
MOD-13599 feat: `load_document` HASH part

JonasKruckenberg:
#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches query document field loading semantics (coercion, QuerySrc, bulk load) on a path that must match C; risk is mitigated by proptest/unit tests and mock-only FFI, but wiring into production still affects search RETURN/HGETALL behavior.
> 
> **Overview**
> Adds **Rust hash document loading** for `rlookup` via new **`HashDocumentFormat`**, implementing **`DocumentFormat`** with per-field **`load_field`** (including `__key`) and bulk **`load_all`** over hash scan—matching C behavior such as skipping **`QuerySrc`** keys and coercing numeric fields.
> 
> Refactors the loader API: **`OpenDocument`** becomes **`FieldLoader`**, and **`LoadFieldError`** / **`LoadAllError`** variants are generalized (`OpenKeyFailed`, `WrongKeyType`, **`Redis`** for module API failures).
> 
> Extends **`redis_mock`** for tests: **`UserString`** is **`Arc`**-backed with **`RetainString`**, plus **`StringToLongLong`** / **`StringToDouble`** mocks aligned with Redis parsing. Bumps the workspace **`redis-module`** git pin to **`7fb620f`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7d997a35380e444798011bd80bfb1e8e297733ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/redis_mock/src/string.rs
+++ b/src/redisearch_rs/redis_mock/src/string.rs
@@ -7,7 +7,11 @@
  * GNU Affero General Public License v3 (AGPLv3).
 */
 
-use std::ffi::{CStr, c_char};
+use std::{
+    ffi::{CStr, c_char},
+    mem::ManuallyDrop,
+    sync::Arc,
+};
 
 /// Mock implementation of RedisModuleString from redismodule.h for testing purposes.
 #[derive(Default, Copy, Clone)]
@@ -16,6 +20,15 @@ pub(crate) struct UserString {
     pub(crate) user: *const c_char,
     pub(crate) length: usize,
 }
+
+// Safety: `UserString` is a non-owning, read-only view.
+// As such it may be freely moved as long as the caller
+// ensures the backing allocation stays valid and unmutated.
+unsafe impl Send for UserString {}
+// Safety: `UserString` is a non-owning, read-only view.
+// As such it may be freely accessed by other treads as long
+// as the caller ensures the backing allocation stays valid and unmutated.
+unsafe impl Sync for UserString {}
 
 /// Mock implementation of RedisModule_CreateString from redismodule.h for testing purposes.
 ///
@@ -28,11 +41,14 @@ pub(crate) unsafe extern "C" fn RedisModule_CreateString(
     ptr: *const ::std::ffi::c_char,
     len: usize,
 ) -> *mut redis_module::raw::RedisModuleString {
-    let val = Box::new(UserString {
+    let val = Arc::new(UserString {
         user: ptr,
         length: len,
     });
-    Box::into_raw(val).cast()
+
+    Arc::into_raw(val)
+        .cast::<redis_module::raw::RedisModuleString>()
+        .cast_mut()
 }
 
 /// Mock implementation of RedisModule_StringPtrLen from redismodule.h for testing purposes.
@@ -45,8 +61,9 @@ pub(crate) unsafe extern "C" fn RedisModule_StringPtrLen(
     s: *const redis_module::raw::RedisModuleString,
     len: *mut usize,
 ) -> *const ::std::ffi::c_char {
-    // Safety: we know we're using UserString here (1)
-    let s = unsafe { &*(s.cast::<UserString>()) };
+    // Safety: The caller ensured the ptr is correct (1.)
+    let s = ManuallyDrop::new(unsafe { Arc::from_raw(s.cast::<UserString>()) });
+
     // Safety: Caller provides valid len pointer (2)
     unsafe {
         *len = s.length;
@@ -65,7 +82,7 @@ pub(crate) unsafe extern "C" fn RedisModule_FreeString(
     s: *mut redis_module::raw::RedisModuleString,
 ) {
     // Safety: we own the memory (1) and the caller promised to call this only once (2)
-    drop(unsafe { Box::from_raw(s.cast::<UserString>()) });
+    drop(unsafe { Arc::from_raw(s.cast::<UserString>()) })
 }
 
 /// Mock implementation of RedisModule_Strdup from redismodule.h for testing purposes.
@@ -113,10 +130,26 @@ pub(crate) unsafe extern "C" fn RedisModule_HoldString(
     _ctx: *mut redis_module::raw::RedisModuleCtx,
     s: *mut redis_module::raw::RedisModuleString,
 ) -> *mut redis_module::raw::RedisModuleString {
-    // Safety: we know we're using UserString here (1)
-    let s = unsafe { &*(s.cast::<UserString>()) };
-    let val = Box::new(*s);
-    Box::into_raw(val).cast()
+    // Safety: The caller ensured the ptr is correct (1.)
+    unsafe {
+        Arc::increment_strong_count(s.cast::<UserString>());
+    }
+    s
+}
+
+/// Mock implementation of RedisModule_RetainString from redismodule.h for testing purposes.
+///
+/// Safety:
+/// 1. `s` must be a valid pointer to a RedisModuleString created by this mock
+#[expect(non_snake_case)]
+pub(crate) unsafe extern "C" fn RedisModule_RetainString(
+    _ctx: *mut redis_module::raw::RedisModuleCtx,
+    s: *mut redis_module::raw::RedisModuleString,
+) {
+    // Safety: The caller ensured the ptr is correct (1.)
+    unsafe {
+        Arc::increment_strong_count(s.cast::<UserString>());
+    }
 }
 
 /// Mock implementation of RedisModule_CreateStringPrintf from redismodule.h for testing purposes.
--- a/src/redisearch_rs/redis_mock/src/string_to_number.rs
+++ b/src/redisearch_rs/redis_mock/src/string_to_number.rs
@@ -187,6 +187,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg_attr(miri, ignore)] // can't call FFI function RedisModule_CreateString under miri
     fn test_string2d() {
         // Valid hexadecimal value.
         assert_eq!(string2d(b"0x0p+0"), Some(0.0));
--- a/src/redisearch_rs/redis_mock/src/string_to_number.rs
+++ b/src/redisearch_rs/redis_mock/src/string_to_number.rs
@@ -1,0 +1,219 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use crate::string::UserString;
+use std::{
+    ffi::{c_char, c_int},
+    mem::ManuallyDrop,
+    sync::Arc,
+};
+
+/// Returns `Some(slice)` over the bytes backing `s`, or `None` if the pointer
+/// is null.
+///
+/// # Safety
+/// 1. `s` must either be NULL or point to a valid [`UserString`] whose
+///    `user`/`length` fields describe an initialised byte buffer.
+unsafe fn user_string_bytes<'a>(
+    s: *const redis_module::raw::RedisModuleString,
+) -> Option<&'a [u8]> {
+    if s.is_null() {
+        return None;
+    }
+
+    // Safety: The caller ensured the ptr is correct (1.)
+    let us = ManuallyDrop::new(unsafe { Arc::from_raw(s.cast::<UserString>()) });
+
+    // Safety: UserString invariant: `user` points to `length` initialised bytes.
+    let bytes = unsafe { std::slice::from_raw_parts(us.user.cast::<u8>(), us.length) };
+    Some(bytes)
+}
+
+/// Mock implementation of Redis' `string2ll` (`src/util.c`)
+/// which is what `RM_StringToLongLong`.
+fn string2ll(bytes: &[u8]) -> Option<i64> {
+    if bytes.is_empty() || bytes.len() >= 21 {
+        return None;
+    }
+    let (negative, digits) = match bytes.split_first() {
+        Some((b'-', rest)) => (true, rest),
+        _ => (false, bytes),
+    };
+    match digits {
+        b"0" if !negative => Some(0),
+        [b'1'..=b'9', ..] => {
+            // Accumulate in the negative i64 range, which reaches i64::MIN; the
+            // positive range does not, so negate at the end instead
+            let mut v: i64 = 0;
+            for b in digits {
+                let d = b.checked_sub(b'0').filter(|d| *d < 10)?;
+                v = v.checked_mul(10)?.checked_sub(i64::from(d))?;
+            }
+            if negative { Some(v) } else { v.checked_neg() }
+        }
+        _ => None,
+    }
+}
+
+/// Mock implementation of `RedisModule_StringToLongLong`.
+///
+/// Parses the entire string as a base-10 `i64` via [`string2ll`], matching
+/// Redis' semantics. On success writes the value to `*out` and returns
+/// `REDISMODULE_OK`. On failure leaves `*out` untouched and returns
+/// `REDISMODULE_ERR`.
+///
+/// # Safety
+/// 1. `s` must point to a valid [`UserString`] (created by the mock).
+/// 2. `out` must be a valid pointer to an `i64` slot.
+#[expect(non_snake_case)]
+pub unsafe extern "C" fn RedisModule_StringToLongLong(
+    s: *const redis_module::raw::RedisModuleString,
+    out: *mut i64,
+) -> c_int {
+    // Safety: caller has to ensure (1)
+    let bytes = match unsafe { user_string_bytes(s) } {
+        Some(b) => b,
+        None => return redis_module::raw::REDISMODULE_ERR as c_int,
+    };
+
+    match string2ll(bytes) {
+        Some(v) => {
+            // Safety: caller has to ensure (2)
+            unsafe { *out = v };
+            redis_module::raw::REDISMODULE_OK as c_int
+        }
+        None => redis_module::raw::REDISMODULE_ERR as c_int,
+    }
+}
+
+#[cfg(target_os = "linux")]
+fn errno() -> *mut c_int {
+    // Safety: always safe to call; returns a valid pointer to the thread's errno.
+    unsafe { libc::__errno_location() }
+}
+#[cfg(not(target_os = "linux"))]
+fn errno() -> *mut c_int {
+    // Safety: always safe to call; returns a valid pointer to the thread's errno.
+    unsafe { libc::__error() }
+}
+
+/// Port of Redis' `string2d` (`src/util.c`), which is what `RedisModule_StringToDouble`
+/// calls: strict `f64` via libc `strtod` — no empty/leading-whitespace/trailing
+/// junk, no NaN, no overflow/underflow.
+fn string2d(bytes: &[u8]) -> Option<f64> {
+    match bytes.first() {
+        None | Some(b' ' | b'\t'..=b'\r') => return None,
+        _ => {}
+    }
+    let cstr = std::ffi::CString::new(bytes).ok()?;
+    let mut end: *mut c_char = std::ptr::null_mut();
+    // Safety: `errno()` returns a valid pointer to the thread's errno.
+    unsafe { *errno() = 0 };
+    // Safety: `cstr` is a valid NUL-terminated string and `end` a valid slot.
+    let v = unsafe { libc::strtod(cstr.as_ptr(), &mut end) };
+    // Safety: `errno()` returns a valid pointer to the thread's errno.
+    let err = unsafe { *errno() };
+    // Safety: `strtod` sets `end` to a pointer within `cstr`.
+    let consumed = unsafe { end.cast_const().offset_from(cstr.as_ptr()) };
+    (usize::try_from(consumed).ok() == Some(bytes.len())
+        && !v.is_nan()
+        && !(err == libc::ERANGE && (v.is_infinite() || v == 0.0)))
+        .then_some(v)
+}
+
+/// Mock implementation of `RedisModule_StringToDouble`.
+///
+/// Parses the entire string as `f64` via [`string2d`], matching Redis'
+/// semantics. On success writes the value to `*out` and returns
+/// `REDISMODULE_OK`. On failure leaves `*out` untouched and returns
+/// `REDISMODULE_ERR`.
+///
+/// # Safety
+/// 1. `s` must point to a valid [`UserString`] (created by the mock).
+/// 2. `out` must be a valid pointer to an `f64` slot.
+#[expect(non_snake_case)]
+pub unsafe extern "C" fn RedisModule_StringToDouble(
+    s: *const redis_module::raw::RedisModuleString,
+    out: *mut f64,
+) -> c_int {
+    // Safety: caller has to ensure (1)
+    let bytes = match unsafe { user_string_bytes(s) } {
+        Some(b) => b,
+        None => return redis_module::raw::REDISMODULE_ERR as c_int,
+    };
+
+    match string2d(bytes) {
+        Some(v) => {
+            // Safety: caller has to ensure (2)
+            unsafe { *out = v };
+            redis_module::raw::REDISMODULE_OK as c_int
+        }
+        None => redis_module::raw::REDISMODULE_ERR as c_int,
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_string2ll() {
+        // May not start with +.
+        assert_eq!(string2ll(b"+1"), None);
+        // May not start with 0.
+        assert_eq!(string2ll(b"01"), None);
+        assert_eq!(string2ll(b" 1"), None);
+        assert_eq!(string2ll(b"1 "), None);
+
+        assert_eq!(string2ll(b"-1"), Some(-1));
+        assert_eq!(string2ll(b"0"), Some(0));
+        assert_eq!(string2ll(b"1"), Some(1));
+        assert_eq!(string2ll(b"99"), Some(99));
+        assert_eq!(string2ll(b"-99"), Some(-99));
+
+        assert_eq!(string2ll(b"-9223372036854775808"), Some(i64::MIN));
+        // Overflow.
+        assert_eq!(string2ll(b"-9223372036854775809"), None);
+
+        assert_eq!(string2ll(b"9223372036854775807"), Some(i64::MAX));
+        // Overflow.
+        assert_eq!(string2ll(b"9223372036854775808"), None);
+    }
+
+    #[test]
+    fn test_string2d() {
+        // Valid hexadecimal value.
+        assert_eq!(string2d(b"0x0p+0"), Some(0.0));
+        assert_eq!(string2d(b"0x1p+0"), Some(1.0));
+
+        // Valid floating-point numbers.
+        assert_eq!(string2d(b"1.5"), Some(1.5));
+        assert_eq!(string2d(b"-3.14"), Some(-3.14));
+        assert_eq!(string2d(b"2.0e10"), Some(2.0e10));
+        assert_eq!(string2d(b"1e-3"), Some(0.001));
+
+        // Valid integer.
+        assert_eq!(string2d(b"42"), Some(42.0));
+
+        assert_eq!(string2d(b""), None);
+        assert_eq!(string2d(b" 1.23"), None);
+        // Invalid hexadecimal format.
+        assert_eq!(string2d(b"0x1.2g"), None);
+        // Hexadecimal NaN.
+        assert_eq!(string2d(b"0xNan"), None);
+
+        // Overflow.
+        assert_eq!(
+            string2d(
+                b"23456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789123456789"
+            ),
+            None
+        );
+    }
+}
--- a/src/redisearch_rs/rlookup/src/load_document/hash.rs
+++ b/src/redisearch_rs/rlookup/src/load_document/hash.rs
@@ -1,0 +1,249 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use crate::load_document::{
+    DOCUMENT_OPEN_KEY_QUERY_FLAGS, DocumentFormat, FieldLoader, LoadAllError, LoadFieldError,
+    UNDERSCORE_KEY,
+};
+use crate::{RLookup, RLookupKey, RLookupKeyFlag, RLookupRow};
+use redis_module::RedisString;
+use redis_module::key::RedisKey;
+use redis_module::{KeyType, ScanKeyCursor};
+use std::ffi::CStr;
+use std::ptr::{self, NonNull};
+use std::slice;
+use value::{SharedValue, Value};
+
+/// Document loading support for the hash format
+pub struct HashDocumentFormat {
+    ctx: NonNull<ffi::RedisModuleCtx>,
+    force_string: bool,
+}
+
+/// Represents an open hash document ready to load values from.
+#[derive(Debug)]
+pub struct HashFieldLoader<'a> {
+    key: RedisKey,
+    key_name: &'a RedisString,
+}
+
+/// Why opening a hash key failed.
+enum HashOpenError {
+    /// `RedisModule_OpenKey` returned NULL — no document at this key.
+    NotFound,
+    /// The key exists but is not a hash.
+    WrongType,
+}
+
+impl HashDocumentFormat {
+    pub const fn new(ctx: NonNull<ffi::RedisModuleCtx>, force_string: bool) -> Self {
+        Self { ctx, force_string }
+    }
+
+    /// Open `key_name` and verify it points to a hash.
+    fn open_hash_key(&self, key_name: &RedisString) -> Result<RedisKey, HashOpenError> {
+        let key = RedisKey::open_with_flags(
+            self.ctx.cast().as_ptr(),
+            key_name,
+            DOCUMENT_OPEN_KEY_QUERY_FLAGS,
+        );
+
+        if key.is_null() {
+            return Err(HashOpenError::NotFound);
+        }
+        if key.key_type() != KeyType::Hash {
+            return Err(HashOpenError::WrongType);
+        }
+        Ok(key)
+    }
+}
+
+impl DocumentFormat for HashDocumentFormat {
+    type FieldLoader<'a> = HashFieldLoader<'a>;
+
+    fn open<'key>(
+        &'key self,
+        key_name: &'key RedisString,
+    ) -> Result<Self::FieldLoader<'key>, LoadFieldError> {
+        let key = self.open_hash_key(key_name).map_err(|e| match e {
+            HashOpenError::NotFound => LoadFieldError::OpenKeyFailed,
+            HashOpenError::WrongType => LoadFieldError::WrongKeyType,
+        })?;
+
+        Ok(HashFieldLoader { key, key_name })
+    }
+
+    fn load_all(
+        &self,
+        rlookup: &mut RLookup,
+        dst_row: &mut RLookupRow,
+        key_name: &RedisString,
+    ) -> Result<(), LoadAllError> {
+        let key = self
+            .open_hash_key(key_name)
+            .map_err(|_| LoadAllError::OpenKeyFailed)?;
+
+        let scan_cursor = ScanKeyCursor::new(key);
+
+        scan_cursor.for_each(|_key, field, value| {
+            let (field_ptr, field_len) = field.as_cstr_ptr_and_len();
+
+            let field_cstr = {
+                // SAFETY: `RedisModule_StringPtrLen` returns a pointer to a null-terminated
+                // SDS string.
+                let bytes = unsafe { slice::from_raw_parts(field_ptr.cast::<u8>(), field_len + 1) };
+
+                CStr::from_bytes_until_nul(bytes).expect("SDS string must contain null-terminator")
+            };
+
+            let key = if let Some(c) = rlookup.find_key_by_name(field_cstr) {
+                if c.current()
+                    .unwrap() // NB: if `find_key_by_name` returns Some the cursor always points at a valid element
+                    .flags
+                    .contains(RLookupKeyFlag::QuerySrc)
+                {
+                    // Key name is already taken by a query key.
+                    return;
+                } else {
+                    c.into_current().unwrap()
+                }
+            } else {
+                // First returned document, create the key.
+                rlookup
+                    .get_key_load(
+                        field_cstr.to_owned(),
+                        field_cstr,
+                        RLookupKeyFlag::ForceLoad.into(),
+                    )
+                    .unwrap()
+            };
+
+            let coerce = if key.flags.contains(RLookupKeyFlag::Numeric) && !self.force_string {
+                HashCoerceType::Double
+            } else {
+                HashCoerceType::Str
+            };
+            let value = hval_to_value(value, coerce);
+
+            dst_row.write_key(key, value);
+        });
+
+        Ok(())
+    }
+}
+
+impl FieldLoader for HashFieldLoader<'_> {
+    fn load_field(&self, kk: &RLookupKey, dst_row: &mut RLookupRow) -> Result<(), LoadFieldError> {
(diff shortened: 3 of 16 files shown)


src/redisearch_rs/rlookup/tests/load_document_hash.rs
```diff
@@ -0,0 +1,276 @@
...
+    RLookupRow,
+};
+use std::ffi::CString;
+use std::ptr::NonNull;
+
+/// Build a [`redis_mock::TestContext`] with the given key type and
+/// `(field, value)` pairs (which back `RedisModule_ScanKey` iteration), leak
+/// it, and return a `NonNull<RedisModuleCtx>` pointing at it.
+///
+/// Leaking is acceptable for tests: the context is alive for the duration of
```
memark: Acceptable, possibly. Desirable though? Is it hard to avoid?

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,305 @@
...
+use crate::{RLookup, RLookupKey, RLookupKeyFlag, RLookupRow};
+use redis_module::RedisString;
+use redis_module::key::RedisKey;
+use redis_module::{KeyType, ScanKeyCursor};
+use std::ffi::CStr;
+use std::ptr::{self, NonNull};
+use std::slice;
+use value::{SharedValue, Value};
+
+pub struct HashFormat {
```
memark: Is this a good name? The fields don't seem to describe anything I'd call a "format".

Either way, it might warrant a type-level doc comment.

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,305 @@
...
+    ctx: NonNull<ffi::RedisModuleCtx>,
+    force_string: bool,
+}
+
+#[derive(Debug)]
+pub struct HashDoc<'a> {
```
memark: Any form of doc comment on this one?

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,305 @@
...
+        Ok(HashDoc { key, key_name })
+    }
+
+    fn load_all(
+        &self,
+        rlookup: &mut RLookup,
+        dst_row: &mut RLookupRow,
+        key_name: &RedisString,
+    ) -> Result<(), LoadAllError> {
+        // Matches `RLookup_HGETALL` in C: returns ERR without setting `QueryError`
```
memark: Will this remain meaningful once the migration is done and the C-side is deleted?

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,305 @@
...
+        // when the key is missing or has the wrong type.
+        let key = self
+            .open_hash_key(key_name)
+            .map_err(|_| LoadAllError::KeyNotFound)?;
+
+        let scan_cursor = ScanKeyCursor::new(key);
+
+        scan_cursor.for_each(|_key, field, value| {
```
memark, line 94: I'd prefer a regular for-loop here unless there is a specific reason to reach for this construct here.

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,305 @@
...
+                // SAFETY: `RedisModule_StringPtrLen` returns a pointer to a null-terminated
+                // SDS string.
+                let bytes = unsafe { slice::from_raw_parts(field_ptr.cast::<u8>(), field_len + 1) };
+
+                CStr::from_bytes_with_nul(bytes).unwrap()
+            };
+
+            let key = if let Some(c) = rlookup.find_key_by_name(field_cstr) {
+                if c.current()
+                    .unwrap()
```
memark: How/why can we be sure that this is always `Some`? Might be worth or comment or an `expect`.

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,305 @@
...
+impl OpenDocument for HashDoc<'_> {
+    fn load_field(&self, kk: &RLookupKey, dst_row: &mut RLookupRow) -> Result<(), LoadFieldError> {
+        let path = match kk.path() {
+            Some(p) => p.as_ref(),
+            // No path set — nothing to load.
+            None => return Ok(()),
+        };
+
+        // `hash_get` requires a `&str`. Hash field paths in RediSearch schemas are always
+        // valid UTF-8 (they come from user-defined schema field names).
```
memark: Are we sure this property holds? I have seen other cases where any `NUL`-terminated binary input is accepted where you'd think it should be valid utf-8.

I think the `expect` is spot-on, but maybe the comment could be relaxed a bit.

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,305 @@
...
+            let coerce = if kk.flags.contains(RLookupKeyFlag::Numeric) {
+                HashValueCoerce::Dbl
+            } else {
+                HashValueCoerce::Str
+            };
+            hval_to_value(&val, coerce)
+        } else if path.to_bytes() == UNDERSCORE_KEY.to_bytes() {
+            // The field path is "__key" — load the document's key name as a string value.
+            //
+            // FIXME: Ideally we would call `RedisModule_GetKeyNameFromModuleKey` here
```
memark: I'd prefer only framing this as a workaround, omitting the latent todo/fixme semantics. And/or filing a Jira ticket if this is something we really wanna do.

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,305 @@
...
+            return Ok(());
+        };
+
+        dst_row.write_key(kk, val);
+
+        Ok(())
+    }
+}
+
+/// How to coerce a Redis hash field value into an [`SharedValue`].
```
memark:
```suggestion
/// How to coerce a Redis hash field value into a [`SharedValue`].
```

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,305 @@
...
+    Int,
+    Dbl,
+    Bool,
+}
+
+/// Converts a Redis hash field value to a [`SharedValue`].
+fn hval_to_value(src: &RedisString, coerce: HashValueCoerce) -> SharedValue {
+    match coerce {
+        HashValueCoerce::Bool | HashValueCoerce::Int => {
+            // C calls `RSValue_NewNumberFromInt64`, which casts to `f64`.
```
memark: Same as previous reference to existing C code.

src/redisearch_rs/redis_mock/src/string.rs
```diff
@@ -113,10 +126,22 @@ pub(crate) unsafe extern "C" fn RedisModule_HoldString(
...
-    Box::into_raw(val).cast()
+    // Safety: The caller ensured the ptr is correct (1.)
+    unsafe {
+        Arc::increment_strong_count(s.cast::<UserString>());
+    }
+    s
+}
+
+#[expect(non_snake_case)]
+pub(crate) unsafe extern "C" fn RedisModule_RetainString(
```
memark, line 145: Missing safety docs

src/redisearch_rs/redis_mock/src/string.rs
```diff
@@ -113,10 +126,22 @@ pub(crate) unsafe extern "C" fn RedisModule_HoldString(
...
+    _ctx: *mut redis_module::raw::RedisModuleCtx,
+    s: *mut redis_module::raw::RedisModuleString,
+) {
+    // Safety: The caller ensured the ptr is correct (1.)
```
memark, line 149: Reference "1." doesn't exist.

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,305 @@
...
+
+        scan_cursor.for_each(|_key, field, value| {
+            let (field_ptr, field_len) = field.as_cstr_ptr_and_len();
+
+            let field_cstr = {
+                // SAFETY: `RedisModule_StringPtrLen` returns a pointer to a null-terminated
+                // SDS string.
+                let bytes = unsafe { slice::from_raw_parts(field_ptr.cast::<u8>(), field_len + 1) };
+
+                CStr::from_bytes_with_nul(bytes).unwrap()
```
memark: Is it safe to unwrap here? Are interior NULs in strings never present?

src/redisearch_rs/redis_mock/src/string_to_number.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    // Safety: UserString invariant: `user` points to `length` initialised bytes.
+    let bytes = unsafe { std::slice::from_raw_parts(us.user.cast::<u8>(), us.length) };
+    Some(bytes)
+}
+
+/// Port of Redis' `string2ll` (`src/util.c`), which is what `RM_StringToLongLong`
+/// calls: strict base-10 `i64`, no whitespace, no leading zeros (except `0`), no
+/// `+` sign, no `-0`.
+fn string2ll(bytes: &[u8]) -> Option<i64> {
```
memark, line 40: I'd like to see some unit tests for this, covering edge cases.

src/redisearch_rs/redis_mock/src/string_to_number.rs
```diff
@@ -0,0 +1,166 @@
...
+#[cfg(not(target_os = "linux"))]
+fn errno() -> *mut c_int {
+    // Safety: always safe to call; returns a valid pointer to the thread's errno.
+    unsafe { libc::__error() }
+}
+
+/// Port of Redis' `string2d` (`src/util.c`), which is what `RedisModule_StringToDouble`
+/// calls: strict `f64` via libc `strtod` — no empty/leading-whitespace/trailing
+/// junk, no NaN, no overflow/underflow.
+fn string2d(bytes: &[u8]) -> Option<f64> {
```
memark, line 109: I'd like to see some unit tests for this, covering edge cases.

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+    }
+}
+
+#[cfg(all(test, feature = "unittest"))]
+mod tests {
+    use super::*;
+    use std::ffi::CString;
+
+    /// Build a `RedisString` whose backing bytes live in the caller's `CString`.
```
memark: Add proper doc links to these backticked types?

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,305 @@
...
+                HashValueCoerce::Dbl
+            } else {
+                HashValueCoerce::Str
+            };
+            hval_to_value(&val, coerce)
+        } else if path.to_bytes() == UNDERSCORE_KEY.to_bytes() {
+            // The field path is "__key" — load the document's key name as a string value.
+            //
+            // FIXME: Ideally we would call `RedisModule_GetKeyNameFromModuleKey` here
+            // to match the C code exactly. However, `RedisKey` in the `redismodule-rs`
```
memark: Add proper doc links to these backticked types?

src/redisearch_rs/redis_mock/src/string_to_number.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+    let (negative, digits) = match bytes.split_first() {
+        Some((b'-', rest)) => (true, rest),
+        _ => (false, bytes),
+    };
+    if !matches!(digits.first(), Some(b'1'..=b'9')) {
+        return None;
+    }
+    let mut v: u64 = 0;
+    for &b in digits {
```
memark: This means we check the first digit _again_, right?

memark (commented):
Did a first pass!

Would you also address/dismiss the bot comments?

src/redisearch_rs/rlookup/src/load_document/hash.rs
JonasKruckenberg: `ScanKeyCursor` does not implement `Iterator` because the lifetime of the pointers returned by C are only scoped to the callback (`for_each` under the hood translates into a monomorphized C callback).

The only way I could have implemented an `Iterator` soundly if I recall correctly was through copying the strings, so `for_each` it is.

src/redisearch_rs/redis_mock/src/string_to_number.rs
JonasKruckenberg: yep buut it is what the C upstream is doing soo https://github.com/redis/redis/blob/4625b8942a82c4737fc14cbf17c221ea1c3fcdb5/src/util.c#L484

src/redisearch_rs/redis_mock/src/string_to_number.rs
JonasKruckenberg: agreed! note to next week me: https://github.com/redis/redis/blob/4625b8942a82c4737fc14cbf17c221ea1c3fcdb5/src/util.c#L1440

src/redisearch_rs/redis_mock/src/string_to_number.rs
JonasKruckenberg: note to self https://github.com/redis/redis/blob/4625b8942a82c4737fc14cbf17c221ea1c3fcdb5/src/util.c#L1544

src/redisearch_rs/rlookup/src/load_document/hash.rs
JonasKruckenberg: I mean its the trait implementation specialized to load documents in the hash format, but im not in love either. happy to take suggestions

src/redisearch_rs/redis_mock/src/string_to_number.rs
memark: Sure thing, but given enough test coverage we should be free to simplify a bit, no?

src/redisearch_rs/rlookup/src/load_document/hash.rs
memark: Ait, I got a couple ideas, but we could keep them out of this PR and do renames afterwards?

JonasKruckenberg: Lets do renames now while we still remember

src/redisearch_rs/rlookup/src/load_document/hash.rs
JonasKruckenberg: Yes relaxing this should be very possible, see the upstream redismodule PR here https://github.com/RedisLabsModules/redismodule-rs/pull/470

src/redisearch_rs/rlookup/src/load_document/hash.rs
JonasKruckenberg: in inline comments? that's not a thing

src/redisearch_rs/rlookup/tests/load_document_hash.rs
JonasKruckenberg: It actually wasn't acceptable, this is a stale comment. removed!

src/redisearch_rs/rlookup/src/load_document/hash.rs
memark: But I wish it were!

src/redisearch_rs/rlookup/src/load_document/hash.rs
JonasKruckenberg: upstream merged

src/redisearch_rs/rlookup/src/load_document/hash.rs
JonasKruckenberg: good point

src/redisearch_rs/rlookup/src/load_document/hash.rs
memark: Ait, will sip on it during the day.

src/redisearch_rs/redis_mock/src/string_to_number.rs
JonasKruckenberg: simplified 👍🏻

src/redisearch_rs/redis_mock/src/string.rs
```diff
@@ -113,10 +126,26 @@ pub(crate) unsafe extern "C" fn RedisModule_HoldString(
...
+    unsafe {
+        Arc::increment_strong_count(s.cast::<UserString>());
+    }
+    s
+}
+
+/// Mock implementation of RedisModule_HoldString from redismodule.h for testing purposes.
+///
+/// Safety:
+/// 1. s must be a valid pointer to a RedisModuleString created by this mock
```
memark:
```suggestion
/// 1. `s` must be a valid pointer to a RedisModuleString created by this mock
```

src/redisearch_rs/rlookup/tests/load_document_hash.rs
```diff
@@ -0,0 +1,263 @@
...
+                    (value, dstidx)
+                })
+                .collect();
+
+            let mut row = RLookupRow::new();
+            format
+                .load_all(&mut rlookup, &mut row, &key_name)
+                .expect("load_all should succeed");
+
+
```
memark:
```suggestion
```

src/redisearch_rs/redis_mock/src/string_to_number.rs
```diff
@@ -0,0 +1,223 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_string2ll() {
+        // May not start with +.
+        assert_eq!(string2ll(b"+1"), None);
+        // Leading space.
+        assert_eq!(string2ll(b" 1"), None);
+        // Trailing space.
```
memark: Not sure these comments are worth the lines they occupy...

src/redisearch_rs/redis_mock/src/string_to_number.rs
```diff
@@ -0,0 +1,223 @@
...
+        assert_eq!(string2ll(b"01"), None);
+
+        assert_eq!(string2ll(b"-1"), Some(-1));
+        assert_eq!(string2ll(b"0"), Some(0));
+        assert_eq!(string2ll(b"1"), Some(1));
+        assert_eq!(string2ll(b"99"), Some(99));
+        assert_eq!(string2ll(b"-99"), Some(-99));
+
+        assert_eq!(string2ll(b"-9223372036854775808"), Some(i64::MIN));
+        // Overflow.
```
memark, line 181: This one is good though.

src/redisearch_rs/rlookup/src/load_document/hash.rs
JonasKruckenberg: Okay after inspecting the `RLookup_FindKey` source which this essentially ports it does this following:

https://github.com/RediSearch/RediSearch/blob/e4a5caee94742c5e4f14810a2cf5fdf479a504ac/src/rlookup_load_document.c#L216-L222

`strncmp` will stop at the first NUL it sees. I changed the implementation from `CStr::from_bytes_with_nul` to `CStr::from_bytes_until_nul` which should result in effectively the same behavior.

src/redisearch_rs/redis_mock/src/string_to_number.rs
JonasKruckenberg: fair point, they are just taken verbatim from the C version

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,284 @@
...
+    /// The C version defined `0`/`0.0` as fallback values for malformed BOOL, INT or DBL fields
+    #[test]
+    #[cfg_attr(miri, ignore)] // can't call FFI function RedisModule_CreateString under miri
+    fn invalid_fallback_to_zero() {
+        redis_mock::init_redis_module_mock();
+
+        // malformed integer
+        let bad = CString::new("3.5").unwrap();
+        assert_eq!(
+            expect_num(&hval_to_value(&make_string(&bad), HashCoerceType::Int)),
```
memark: old types linger here, failing compilation

src/redisearch_rs/rlookup/tests/load_document_hash.rs
```diff
@@ -0,0 +1,262 @@
...
+///
+/// `get_key_load` / `get_key_write` return `None` on duplicate names, so every
+/// field-driven test needs unique keys. Generating via `HashMap` bakes the
+/// invariant into the strategy itself, so shrinking preserves it instead of
+/// fighting a runtime filter.
+fn arb_unique_fields() -> impl Strategy<Value = Vec<(CString, CString)>> {
+    any::<std::collections::HashMap<CString, CString>>().prop_map(|m| m.into_iter().collect())
+}
+
+// ── HashFormat::open ─────────────────────────────────────────────────────
```
memark: old reference

src/redisearch_rs/rlookup/tests/load_document_hash.rs
```diff
@@ -0,0 +1,262 @@
...
+        with_ctx(KeyType::String, &[], |ctx| {
+            let format = HashDocumentFormat::new(ctx, false);
+            let key_name = make_redis_string(&key_name_bytes);
+
+            let err = format.open(&key_name).unwrap_err();
+            assert!(matches!(err, LoadFieldError::WrongKeyType));
+        })
+    }
+
+    // ── HashFormat::load_all ─────────────────────────────────────────────────
```
memark: old reference

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        // Non-UTF-8 bytes (excluding NUL) must round-trip unchanged.
+        let bytes = CString::new(b"\xff\xfe\x80hello".to_vec()).unwrap();
+        let s = make_string(&bytes);
+
+        let v = hval_to_value(&s, HashCoerceType::Str);
+        assert_eq!(expect_string(&v), b"\xff\xfe\x80hello");
+    }
+
+    /// The C version defined `0`/`0.0` as fallback values for malformed BOOL, INT or DBL fields
```
memark: I don't think this is accurate. From what I can see (rlookup_load_document.c:31-33) the return code of `RedisModule_StringToDouble` was simply ignored and bs read on parse failure. So this 0.0 is actually a real improvement!

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,249 @@
...
+/// Converts a Redis hash field value to a [`SharedValue`].
+fn hval_to_value(src: &RedisString, coerce: HashCoerceType) -> SharedValue {
+    match coerce {
+        HashCoerceType::Double => {
+            let dd = src.parse_float().unwrap_or(0.0);
+            SharedValue::new_num(dd)
+        }
+        HashCoerceType::Str => {
+            redis_module::raw::string_retain_string(ptr::null_mut(), src.inner);
+            // Safety: `src` is guaranteed to be a valid redis string (1.) and  we've called retain above
```
memark: There is no `1.` above

src/redisearch_rs/redis_mock/src/string.rs
```diff
@@ -17,6 +21,11 @@ pub(crate) struct UserString {
     pub(crate) length: usize,
 }
 
+// Safety: `UserString` is read-only
+unsafe impl Send for UserString {}
+// Safety: `UserString` is read-only
+unsafe impl Sync for UserString {}
```
memark: I'm not super well versed with these traits, but the comments feel insufficient. Having looked at the type and sparred a bit with Claude, could you use this to get some inspiration for something a bit longer?
```
// SAFETY: `UserString` is a non-owning view (pointer + length). It's `Copy` with
// no `Drop` and no interior mutability, so there's no thread affinity and nothing
// to release on a given thread. Soundness rests entirely on the pointee: the bytes
// `user` points to must stay valid and unmutated for as long as any `UserString`
// referencing them lives. In tests the backing buffer is `'static`/outlives all
// uses and is never mutated, so the value is safe to move to- and `&`-share across
// threads.
```

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,249 @@
...
+
+            dst_row.write_key(key, value);
+        });
+
+        Ok(())
+    }
+}
+
+impl FieldLoader for HashFieldLoader<'_> {
+    fn load_field(&self, kk: &RLookupKey, dst_row: &mut RLookupRow) -> Result<(), LoadFieldError> {
```
memark, line 146: Is there any test coverage of this function?

src/redisearch_rs/rlookup/tests/load_document_hash.rs
```diff
@@ -0,0 +1,262 @@
+/*
```
memark, line 1: I would name this file `tests/load_document/hash.rs` mirroring the `src` side.

memark (commented):
I like the new names! 🥇

src/redisearch_rs/rlookup/src/load_document/hash.rs
JonasKruckenberg: weeeeell, I'll spare you the details of me waisting my life trying to get this to work but TLDR:
`RedisModule_HashGet` (which this boils down to calling) is a variadic function. We cannot actually mock a variadic function in Rust though and all workarounds I've found are either super UB prone (only supporting a fixed number of args but we cant detect being called with a different number and callers will corrupt our stack or registers) or arch-specific assembly stubs or nightly-only features.

So after much messing around I decided its probably fine to rely on the existing pytests for this.

src/redisearch_rs/rlookup/src/load_document/hash.rs
JonasKruckenberg: re-read the source and yeah appears you're right, I thought they would have at least initialized the fields but oh well.

src/redisearch_rs/redis_mock/src/string.rs
JonasKruckenberg: how about so?

src/redisearch_rs/rlookup/src/load_document/hash.rs
memark: Fair enough. You gave it your best, kid! 🎖️

src/redisearch_rs/redis_mock/src/string.rs
memark: Not sure what your answer means tbh?

JonasKruckenberg: sorry, I hadn't pushed yet at that point, local tests were still running

memark (dismissed):
Looking good!

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,249 @@
...
+            .map_err(|_| LoadAllError::OpenKeyFailed)?;
+
+        let scan_cursor = ScanKeyCursor::new(key);
+
+        scan_cursor.for_each(|_key, field, value| {
+            let (field_ptr, field_len) = field.as_cstr_ptr_and_len();
+
+            let field_cstr = {
+                // SAFETY: `RedisModule_StringPtrLen` returns a pointer to a null-terminated
+                // SDS string.
```
oshadmi, line 103: Consider not referring to the internal type `SDS`.

```suggestion
               // SAFETY: `field_ptr` points to `field_len` bytes from `RedisModule_StringPtrLen`,
               // and `RedisModuleString` buffers are NUL-terminated after their logical length.

```

src/redisearch_rs/rlookup/src/load_document/hash.rs
```diff
@@ -0,0 +1,249 @@
...
+                let bytes = unsafe { slice::from_raw_parts(field_ptr.cast::<u8>(), field_len + 1) };
+
+                CStr::from_bytes_until_nul(bytes).expect("SDS string must contain null-terminator")
```
oshadmi, line 106: Consider not mentioning the internal type `SDS` in the error message
```suggestion
               CStr::from_bytes_until_nul(bytes)
    .expect("Redis string buffer must be null-terminated after its logical length")
```

src/redisearch_rs/rlookup/src/load_document/hash.rs
oshadmi, line 1: Codex warned that hash.rs is not registered as a Cargo integration test.
Cargo only auto-discovers `.rs` files directly under `tests/`; metadata for `rlookup` lists only the `row` test target, and `cargo test -p rlookup --test load_document --no-run` reports `no test target named load_document`.
This means the new hash document loading tests are never run. Move the file to a direct test target like `tests/load_document_hash.rs`, or add `tests/load_document.rs` that includes `#[path = "load_document/hash.rs"] mod hash;`.

src/redisearch_rs/redis_mock/src/string_to_number.rs
```diff
@@ -0,0 +1,220 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_string2ll() {
+        // May not start with +.
```
oshadmi, line 167: Also may not be just "+" or "-"
And may not be empty.

src/redisearch_rs/rlookup/src/load_document/mod.rs
```diff
@@ -33,19 +37,32 @@ const DOCUMENT_OPEN_KEY_QUERY_FLAGS: KeyFlags = KeyFlags::from_bits_retain(
 #[derive(Debug, thiserror::Error)]
 pub enum LoadFieldError {
     /// `RedisModule_OpenKey` / `japi->openKeyWithFlags` returned NULL.
-    #[error("document key does not exist")]
-    KeyNotFound,
+    #[error("document key could not be opened")]
+    OpenKeyFailed,
```
oshadmi: Since this variant maps to `QueryErrorCode::NoDoc` and the comment says calling open key returned NULL, should the message stay closer to “document key does not exist” / “document key not found”?
“Could not be opened” sounds like a broader Redis API failure.

src/redisearch_rs/rlookup/tests/load_document/hash.rs
```diff
@@ -0,0 +1,258 @@
...
+    with_ctx(KeyType::String, &[], |ctx| {
+        let format = HashDocumentFormat::new(ctx, false);
+        let key_name_bytes = CString::new("doc:1").unwrap();
+        let key_name = make_redis_string(&key_name_bytes);
+
+        let mut rlookup = RLookup::new();
+        let mut row = RLookupRow::new();
+        let res = format.load_all(&mut rlookup, &mut row, &key_name);
+
+        assert!(res.is_err());
```
oshadmi: Asserting only `is_err,` which is less than the test name conveys?

src/redisearch_rs/rlookup/tests/load_document_hash.rs
```diff
@@ -0,0 +1,258 @@
...
+                    row.dyn_values().get(dstidx).is_none_or(Option::is_none),
+                    "QuerySrc key should not be written by load_all",
+                );
+            }
+        })
+    }
+
+    #[test]
+    #[cfg_attr(miri, ignore)] // can't call FFI function RedisModule_CreateString under miri
+    fn load_all_creates_keys_for_unknown_fields(
```
oshadmi, line 191: 1. Consider more coverage: the current tests cover the load_all branches separately. A small deterministic mixed case with one existing load key, one QuerySrc key, and one unknown field in the same hash would verify that the scan callback composes those branches correctly across one RedisModule_ScanKey pass. Not required for branch coverage, but it would make the load_all behavior easier to reason about.

2. Consider also including a Numeric key and verify force_string=false coerces to number while force_string=true keeps the Redis string.

3. The fixtures use CString for hash field names and values, so these tests cannot cover embedded NUL bytes. That leaves the binary-safe RedisModule_StringPtrLen case untested.

This matters here because load_all receives `(ptr, len)` from Redis, but then converts the field name through `CStr`. A hash field name like `b"a\0b"` would not be exercised by these tests. Could we add one deterministic byte-backed test for embedded-NUL field/value data, or explicitly document that this path only supports C-string-compatible field names?

src/redisearch_rs/redis_mock/src/string_to_number.rs
JonasKruckenberg: what do you mean by that? 
`+1` and `-1` are invalid, and so is ` 1` or do you mean must be either `+` or `-`?

regardless, this is copied verbatim from redis upstream tests.

src/redisearch_rs/redis_mock/src/string_to_number.rs
```diff
@@ -0,0 +1,220 @@
...
+        // Overflow.
+        assert_eq!(string2ll(b"-9223372036854775809"), None);
+
+        assert_eq!(string2ll(b"9223372036854775807"), Some(i64::MAX));
+        // Overflow.
+        assert_eq!(string2ll(b"9223372036854775808"), None);
+    }
+
+    #[test]
+    #[cfg_attr(miri, ignore)] // can't call FFI function RedisModule_CreateString under miri
```
memark, line 190:
```suggestion
    #[cfg_attr(miri, ignore = "can't call FFI function RedisModule_CreateString under Miri")]
```

src/redisearch_rs/rlookup/src/load_document/mod.rs
JonasKruckenberg: reverted

src/redisearch_rs/rlookup/tests/load_document/hash.rs
JonasKruckenberg: hmm after thinking about it a bit, I don't think this test pulls its weight (this would be a better fit for the FFI interface)

src/redisearch_rs/redis_mock/src/string_to_number.rs
oshadmi: Nitpick: I meant also `+` or `-` could be tested to fail, and also empty.

src/redisearch_rs/rlookup/tests/load_document_hash.rs
JonasKruckenberg: > A small deterministic mixed case with one existing load key, one QuerySrc key

done!

> Consider also including a Numeric key and verify force_string=false coerces to number while force_string=true keeps the Redis string.

done!

> these tests cannot cover embedded NUL bytes. That leaves the binary-safe RedisModule_StringPtrLen case untested

this is intended behavior. redisearch already assumes c-strings for hash field names throughout and will truncate on interior NUL bytes. The old C could would preserve NUL bytes under very spacial circumstances (non-schema fields that are loaded after the fact).

src/redisearch_rs/redis_mock/src/string_to_number.rs
JonasKruckenberg: lets keep this in sync with upstream redis

src/redisearch_rs/rlookup/tests/load_document_hash.rs
oshadmi: @JonasKruckenberg I don't see a mixed load_all / numeric force_string additions. I was thinking of a test like this that would cover both in one place:

```rust
#[test]
#[cfg_attr(miri, ignore)] // can't call FFI function RedisModule_CreateString under miri
fn load_all_mixed_existing_query_unknown_and_numeric_force_string() {
    let key_name_bytes = CString::new("doc:1").unwrap();
    let key_name = make_redis_string(&key_name_bytes);

    let existing = CString::new("existing").unwrap();
    let query_src = CString::new("query_src").unwrap();
    let unknown = CString::new("unknown").unwrap();
    let numeric = CString::new("numeric").unwrap();

    let fields = vec![
        (existing.clone(), CString::new("existing-value").unwrap()),
        (query_src.clone(), CString::new("query-value").unwrap()),
        (unknown.clone(), CString::new("unknown-value").unwrap()),
        (numeric.clone(), CString::new("42.5").unwrap()),
    ];

    with_ctx(KeyType::Hash, &fields, |ctx| {
        let mut rlookup = RLookup::new();

        // Existing DocSrc key: load_all should write into this pre-existing slot.
        let existing_dstidx = rlookup
            .get_key_load(existing.clone(), existing.as_c_str(), RLookupKeyFlags::empty())
            .unwrap()
            .dstidx as usize;

        // QuerySrc key: load_all should skip this field and leave the row slot empty.
        let query_dstidx = rlookup
            .get_key_write(query_src.clone(), RLookupKeyFlags::empty())
            .unwrap()
            .dstidx as usize;

        // Numeric DocSrc key: force_string=false should coerce the hash value to a number.
        let numeric_dstidx = rlookup
            .get_key_load(numeric.clone(), numeric.as_c_str(), RLookupKeyFlags::empty())
            .unwrap()
            .dstidx as usize;

        // Mark the key as Numeric manually so load_all exercises the numeric coercion path.
        // get_key_load filters caller-provided flags to GET_KEY_FLAGS, and Numeric is not
        // one of those input flags. In production, Numeric is normally applied from schema
        // metadata via update_from_field_spec.
        let mut cursor = rlookup.cursor_mut();
        // RLookup keeps insertion order. We inserted existing, then query_src, then numeric,
        // so moving twice positions the cursor on the numeric key. A helper that finds a
        // mutable key by name would avoid relying on that ordering.
        cursor.move_next(); // existing -> query_src
        cursor.move_next(); // query_src -> numeric
        cursor
            .into_current()
            .unwrap()
            .flags |= RLookupKeyFlag::Numeric;

        let mut row = RLookupRow::new();
        HashDocumentFormat::new(ctx, false)
            .load_all(&mut rlookup, &mut row, &key_name)
            .unwrap();

        assert_eq!(
            row.dyn_values()[existing_dstidx]
                .as_ref()
                .unwrap()
                .as_str_bytes(),
            Some(b"existing-value"),
        );

        assert!(row.dyn_values().get(query_dstidx).is_none_or(Option::is_none));

        // Unknown field: load_all should create a key on the fly and write its value.
        let unknown_dstidx = rlookup
            .find_key_by_name(&unknown)
            .and_then(|c| c.current().map(|k| k.dstidx as usize))
            .expect("load_all should have created the unknown key");
        assert_eq!(
            row.dyn_values()[unknown_dstidx]
                .as_ref()
                .unwrap()
                .as_str_bytes(),
            Some(b"unknown-value"),
        );

        assert_eq!(
            row.dyn_values()[numeric_dstidx]
                .as_ref()
                .unwrap()
                .as_num(),
            Some(42.5),
        );

        // Re-run only the numeric case with force_string=true, using a fresh
        // RLookupRow/RLookup so the previous load_all result does not affect it.
        let mut rlookup = RLookup::new();

        let numeric_dstidx = rlookup
            .get_key_load(numeric.clone(), numeric.as_c_str(), RLookupKeyFlags::empty())
            .unwrap()
            .dstidx as usize;

        // Mark the only inserted key as Numeric manually. See the force_string=false
        // case above for why this has to be done through a mutable cursor.
        rlookup
            .cursor_mut()
            .into_current()
            .unwrap()
            .flags |= RLookupKeyFlag::Numeric;

        let mut row = RLookupRow::new();
        HashDocumentFormat::new(ctx, true)
            .load_all(&mut rlookup, &mut row, &key_name)
            .unwrap();

        // With force_string=true, Numeric should not coerce the value to a number.
        assert_eq!(
            row.dyn_values()[numeric_dstidx]
                .as_ref()
                .unwrap()
                .as_str_bytes(),
            Some(b"42.5"),
        );
    });
}

File: src/redisearch_rs/rlookup/src/load_document/hash.rs (hunk 32 above)
Comment (oshadmi) [reply src/redisearch_rs/rlookup/src/load_document/hash.rs:106]: @JonasKruckenberg Keeping SDS mentions?

File: src/redisearch_rs/rlookup/src/load_document/hash.rs (hunk 31 above)
Comment (oshadmi) [reply src/redisearch_rs/rlookup/src/load_document/hash.rs:103]: @JonasKruckenberg Keeping SDS mentions?

File: src/redisearch_rs/rlookup/tests/load_document_hash.rs [hunk 38]
```diff
@@ -239,20 +239,139 @@ proptest! {
 
 #[test]
 #[cfg_attr(miri, ignore)] // can't call FFI function RedisModule_CreateString under miri
-fn load_all_returns_err_without_setting_status_for_wrong_type() {
+fn load_all_composes_branches_in_one_pass() {
```
oshadmi, line 242: Codex:
Nitpick / Optional: this replaces the previous `load_all` wrong-type test, which asserted `LoadAllError::OpenKeyFailed`. `open_wrong_type` still covers the underlying wrong-type detection, so this is not blocking, but keeping a small `load_all` error-path test would preserve coverage for the `load_all`-specific error mapping.

src/redisearch_rs/rlookup/tests/load_document_hash.rs
oshadmi: OK tests were added.

src/redisearch_rs/rlookup/tests/load_document_hash.rs
JonasKruckenberg: as mentioned earlier I removed the old test because it didn't really cover enough so this is not a "replacement"

oshadmi (approved):

PR #8077, opened 2026-01-19
[MOD-11050] feat: `load_document` JSON part

JonasKruckenberg:
#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Changes how JSON index documents are loaded for search/RETURN paths, including API-version-dependent Trio behavior; mitigated by mock/proptest and Python tests but still on the query result hot path.
> 
> **Overview**
> Adds a Rust **JSON document loader** for `rlookup`’s `DocumentFormat` abstraction: **`JsonDocumentFormat`** opens JSON keys via RedisJSON, evaluates JSONPath per `RLookupKey`, and maps results into `SharedValue` (including **`Trio`** when `apiVersion >= APIVERSION_RETURN_MULTI_CMP_FIRST`).
> 
> **`redis_json_api`** gains a **`test-mock`** feature with an in-memory vtable (`mock.rs`), **`ResultsIter`** as a **`LendingIterator`**, **`KeyValuesIterator`** exact length hints, and **`open_key_with_flags`** using `KeyFlags` plus read. **`load_document`** renames **`OpenDocument` → `FieldLoader`**, splits load errors (`OpenKeyFailed`, `JsonRootMissing`, JSON serialization), and exports **`JsonDocumentFormat`**. FFI bindgen allowlists **`APIVERSION_RETURN_MULTI_CMP_FIRST`**.
> 
> Coverage: proptest suite against the mock and Python **`FT.SEARCH`** tests for partial JSON loads (JSONPath miss, empty array first element, doc-level `$` load).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e51e4fe288965f48b7e24ffaf34231799dde3e48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rlookup/Cargo.toml
+++ b/src/redisearch_rs/rlookup/Cargo.toml
@@ -21,7 +21,6 @@ value.workspace = true
 redis_mock.workspace = true
 query_error.workspace = true
 redis_json_api.workspace = true
-
 enumflags2.workspace = true
 libc.workspace = true
 pin-project.workspace = true
--- a/src/redisearch_rs/rlookup/Cargo.toml
+++ b/src/redisearch_rs/rlookup/Cargo.toml
@@ -20,6 +20,8 @@ thin_vec.workspace = true
 value.workspace = true
 redis_mock.workspace = true
 query_error.workspace = true
+redis_json_api.workspace = true
+
 enumflags2.workspace = true
 libc.workspace = true
 pin-project.workspace = true
@@ -27,7 +29,9 @@ strum.workspace = true
 thiserror.workspace = true
 redis-module.workspace = true
 workspace_hack.workspace = true
+tracing.workspace = true
 itertools.workspace = true
+lending-iterator.workspace = true
 
 [target.'cfg(all(target_env="musl", target_os="linux"))'.dependencies.redis-module]
 # Statically link to the libclang on aarch64-unknown-linux-musl,
@@ -52,6 +56,8 @@ field_spec = { workspace = true, features = ["unittest"] }
 rlookup = { path = ".", features = ["unittest"] }
 redisearch_rs = { path = "../c_entrypoint/redisearch_rs", features = ["mock_allocator"] }
 redis_mock.workspace = true
+redis_json_api = { workspace = true, features = ["test-mock"] }
+serde_json = { workspace = true }
 
 [build-dependencies]
 build_utils.workspace = true
--- a/src/redisearch_rs/rlookup/src/load_document/json.rs
+++ b/src/redisearch_rs/rlookup/src/load_document/json.rs
@@ -19,7 +19,7 @@ use redis_json_api::{JsonType, JsonValueRef, RedisJsonApi, SerializeError};
 use redis_module::RedisString;
 use std::ffi::CStr;
 use std::ptr::NonNull;
-use value::{Array, Map, SharedValue, Trio, Value};
+use value::SharedValue;
 
 const JSON_ROOT: &CStr = c"$";
 
@@ -48,6 +48,17 @@ impl<'a> JsonDocumentFormat<'a> {
             api_version,
         }
     }
+
+    fn open_key(&self, key_name: &RedisString) -> Option<JsonValueRef<'a>> {
+        // SAFETY: `self.ctx` is a valid Redis module context held for the lifetime of `JsonFormat`.
+        unsafe {
+            self.japi.open_key_with_flags(
+                self.ctx.cast().as_ptr(),
+                key_name,
+                DOCUMENT_OPEN_KEY_QUERY_FLAGS,
+            )
+        }
+    }
 }
 
 impl DocumentFormat for JsonDocumentFormat<'_> {
@@ -60,18 +71,7 @@ impl DocumentFormat for JsonDocumentFormat<'_> {
         &'key self,
         key_name: &'key RedisString,
     ) -> Result<Self::FieldLoader<'key>, LoadFieldError> {
-        // SAFETY: `self.ctx` is a valid Redis module context held for the lifetime of `JsonFormat`.
-        let key = unsafe {
-            self.japi.open_key_with_flags(
-                self.ctx.cast().as_ptr(),
-                key_name,
-                DOCUMENT_OPEN_KEY_QUERY_FLAGS,
-            )
-        };
-
-        let Some(value) = key else {
-            return Err(LoadFieldError::KeyNotFound);
-        };
+        let value = self.open_key(key_name).ok_or(LoadFieldError::KeyNotFound)?;
 
         Ok(JsonFieldLoader {
             ctx: self.ctx,
@@ -87,16 +87,7 @@ impl DocumentFormat for JsonDocumentFormat<'_> {
         dst_row: &mut RLookupRow,
         key_name: &RedisString,
     ) -> Result<(), LoadAllError> {
-        // SAFETY: `self.ctx` is a valid Redis module context held for the lifetime of `JsonFormat`.
-        let json_root = unsafe {
-            self.japi
-                .open_key_with_flags(
-                    self.ctx.cast().as_ptr(),
-                    key_name,
-                    DOCUMENT_OPEN_KEY_QUERY_FLAGS,
-                )
-                .ok_or(LoadAllError::OpenKeyFailed)?
-        };
+        let json_root = self.open_key(key_name).ok_or(LoadAllError::OpenKeyFailed)?;
 
         let json_iter = json_root
             .get(JSON_ROOT)
@@ -122,8 +113,8 @@ impl DocumentFormat for JsonDocumentFormat<'_> {
 }
 
 impl FieldLoader for JsonFieldLoader<'_> {
-    fn load_field(&self, kk: &RLookupKey, dst_row: &mut RLookupRow) -> Result<(), LoadFieldError> {
-        let path = match kk.path() {
+    fn load_field(&self, key: &RLookupKey, dst_row: &mut RLookupRow) -> Result<(), LoadFieldError> {
+        let path = match key.path() {
             Some(p) => p.as_ref(),
             // No path set — nothing to load.
             None => return Ok(()),
@@ -149,7 +140,7 @@ impl FieldLoader for JsonFieldLoader<'_> {
             return Ok(());
         };
 
-        dst_row.write_key(kk, val);
+        dst_row.write_key(key, val);
 
         Ok(())
     }
@@ -187,10 +178,9 @@ fn json_iter_to_value(
 
     // Second, get the first JSON value. `is_empty()` returned false above, so the
     // iterator is contractually obligated to yield at least one value here.
-    let Some(json) = iter.next() else {
-        debug_assert!(false, "ResultsIter::is_empty()/next() disagree");
-        return Ok(None);
-    };
+    let json = iter
+        .next()
+        .expect("ResultsIter::is_empty()/next() disagree");
 
     let val = if matches!(json.get_type(), JsonType::Array) {
         // If the value is an array, we currently try using the first element.
@@ -204,11 +194,13 @@ fn json_iter_to_value(
     };
 
     let otherval = SharedValue::new_string(serialized.to_vec());
+
+    // NB: make sure the iterator is reset to the beginning, so we correctly
+    // get the full expanded value.
+    iter.reset();
     let expand = json_iter_to_value_expanded(ctx, iter);
 
-    Ok(Some(SharedValue::new(Value::Trio(Trio::new(
-        val, otherval, expand,
-    )))))
+    Ok(Some(SharedValue::new_trio(val, otherval, expand)))
 }
 
 // Return an array of expanded values from an iterator.
@@ -216,16 +208,15 @@ fn json_iter_to_value(
 // Required japi_ver >= 4
 fn json_iter_to_value_expanded(
     ctx: NonNull<ffi::RedisModuleCtx>,
-    mut iter: redis_json_api::ResultsIter<'_>,
+    iter: redis_json_api::ResultsIter<'_>,
 ) -> SharedValue {
     debug_assert!(!iter.is_empty(), "should be checked by caller");
-    iter.reset();
 
     let values: Box<_> = iter
         .map_into_iter(|json_val| json_val_to_value_expanded(ctx, json_val))
         .collect();
 
-    SharedValue::new(Value::Array(Array::new(values)))
+    SharedValue::new_array(values)
 }
 
 fn json_val_to_value_expanded(
@@ -234,42 +225,28 @@ fn json_val_to_value_expanded(
 ) -> SharedValue {
     match json.get_type() {
         JsonType::Object => {
-            let len = json.len().unwrap();
+            // SAFETY: `ctx` is a valid Redis module context propagated from the caller.
+            let iter = unsafe { json.key_values(ctx.cast().as_ptr()).unwrap() };
 
-            if len > 0 {
-                // SAFETY: `ctx` is a valid Redis module context propagated from the caller.
-                let iter = unsafe { json.key_values(ctx.cast().as_ptr()).unwrap() };
+            let values = iter.map(|(key, value)| {
+                let key = SharedValue::new_string(key.to_vec());
+                let value = json_val_to_value_expanded(ctx, value.as_ref());
 
-                let values: Box<_> = iter
-                    .map(|(key, value)| {
-                        let key = SharedValue::new_string(key.to_vec());
-                        let value = json_val_to_value_expanded(ctx, value.as_ref());
+                (key, value)
+            });
 
-                        (key, value)
-                    })
-                    .collect();
-
-                SharedValue::new(Value::Map(Map::new(values)))
-            } else {
-                SharedValue::new(Value::Map(Map::new(Box::new([]))))
-            }
+            SharedValue::new_map(values)
         }
         JsonType::Array => {
             let len = json.len().unwrap();
 
-            if len > 0 {
-                let values: Box<_> = (0..len)
-                    .map(|i| {
-                        let json = json.get_at(i).unwrap();
+            let values = (0..len).map(|i| {
+                let json = json.get_at(i).unwrap();
 
-                        json_val_to_value_expanded(ctx, json.as_ref())
-                    })
-                    .collect();
+                json_val_to_value_expanded(ctx, json.as_ref())
+            });
 
-                SharedValue::new(Value::Array(Array::new(values)))
-            } else {
-                SharedValue::new(Value::Array(Array::new(Box::new([]))))
-            }
+            SharedValue::new_array(values)
         }
         // Scalar
         _ => json_val_to_value(ctx, json),
@@ -298,7 +275,7 @@ fn json_val_to_value(ctx: NonNull<ffi::RedisModuleCtx>, json: JsonValueRef<'_>) 
         JsonType::Object | JsonType::Array => {
             // SAFETY: `ctx` is a valid Redis module context propagated from the caller.
             let v = unsafe { json.serialize(ctx.cast().as_ptr()).unwrap() };
-            SharedValue::new_string(v.to_string_lossy().into_bytes())
+            SharedValue::new_string(v.to_vec())
         }
         JsonType::Null => SharedValue::null_static(),
     }
--- a/src/redisearch_rs/rlookup/src/load_document/json.rs
+++ b/src/redisearch_rs/rlookup/src/load_document/json.rs
@@ -1,0 +1,305 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use crate::{
+    RLookup, RLookupKey, RLookupKeyFlags, RLookupRow,
+    load_document::{
+        DOCUMENT_OPEN_KEY_QUERY_FLAGS, DocumentFormat, FieldLoader, LoadAllError, LoadFieldError,
+        UNDERSCORE_KEY,
+    },
+};
+use lending_iterator::LendingIterator;
+use redis_json_api::{JsonType, JsonValueRef, RedisJsonApi, SerializeError};
+use redis_module::RedisString;
+use std::ffi::CStr;
+use std::ptr::NonNull;
+use value::{Array, Map, SharedValue, Trio, Value};
+
+const JSON_ROOT: &CStr = c"$";
+
+pub struct JsonDocumentFormat<'a> {
+    ctx: NonNull<ffi::RedisModuleCtx>,
+    japi: &'a RedisJsonApi,
+    api_version: u32,
+}
+
+pub struct JsonFieldLoader<'a> {
+    ctx: NonNull<ffi::RedisModuleCtx>,
+    value: JsonValueRef<'a>,
+    key_name: &'a RedisString,
+    api_version: u32,
+}
+
+impl<'a> JsonDocumentFormat<'a> {
+    pub const fn new(
+        ctx: NonNull<ffi::RedisModuleCtx>,
+        japi: &'a RedisJsonApi,
+        api_version: u32,
+    ) -> Self {
+        Self {
+            ctx,
+            japi,
+            api_version,
+        }
+    }
+}
+
+impl DocumentFormat for JsonDocumentFormat<'_> {
+    type FieldLoader<'key>
+        = JsonFieldLoader<'key>
+    where
+        Self: 'key;
+
+    fn open<'key>(
+        &'key self,
+        key_name: &'key RedisString,
+    ) -> Result<Self::FieldLoader<'key>, LoadFieldError> {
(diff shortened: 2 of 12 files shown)


src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -22,5 +74,244 @@ pub fn load_all_keys(
...
+    let mut arr = Vec::with_capacity(iter.len());
+    while let Some(json) = iter.next() {
+        arr.push(json_val_to_value_expanded(ctx, json))
+    }
+
+    // TODO RSValueFFI::create_array(arr)
     todo!()
 }
+
+fn json_val_to_value_expanded(ctx: NonNull<ffi::RedisModuleCtx>, json: JsonValueRef) -> RSValueFFI {
```
JonasKruckenberg: @daxhuiberts heres the example you asked for, redisjson requires us pulling out values one by one before placing them into the array/map

daxhuiberts: Well, looking at the original C code, you can do the same here:

For arrays:
1. First allocate a new array with `RSValue_AllocateArray` with the known length.
2. Cast the returned pointer to a slice with `slice::from_raw_parts_mut`.
3. Fill in each entry with received value from `json_val_to_value_expanded`.
4. At the end give the original pointer received from the call in step 1 to `RSValue_NewArray`.

For maps:
1. First allocate a new map with `RSValueMap_AllocUninit` with the known length.
2. For each entry call `RSValueMap_SetEntry` with the current index, the redis string wrappen in an `RSValue` and the value received from `json_val_to_value_expanded`.
3. At the end give the original map received from the call in step 1 to `RSValue_NewMap`.

No cross-stack allocations needed.

Some things to keep in mind:
- `RSValue_AllocateArray` returns a `*mut *mut RSValue` which you can fill in yourself.
- For maps you need to call `RSValueMap_SetEntry` for each entry.
- Maps are not a hashmap, but just a vec with key value tuples.

src/redisearch_rs/rlookup/Cargo.toml
```diff
@@ -20,14 +20,18 @@ thin_vec.workspace = true
 value.workspace = true
 redis_mock.workspace = true
 query_error.workspace = true
+redis_json_api.workspace = true
+
```
daxhuiberts: nitpick: remove the blank line.

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,305 @@
...
+            self.japi.open_key_with_flags(
+                self.ctx.cast().as_ptr(),
+                key_name,
+                DOCUMENT_OPEN_KEY_QUERY_FLAGS,
+            )
+        };
+
+        let Some(value) = key else {
+            return Err(LoadFieldError::KeyNotFound);
+        };
```
daxhuiberts: You can do the same `ok_or` approach here as you do in `load_add`.

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,305 @@
...
+        key_name: &RedisString,
+    ) -> Result<(), LoadAllError> {
+        // SAFETY: `self.ctx` is a valid Redis module context held for the lifetime of `JsonFormat`.
+        let json_root = unsafe {
+            self.japi
+                .open_key_with_flags(
+                    self.ctx.cast().as_ptr(),
+                    key_name,
+                    DOCUMENT_OPEN_KEY_QUERY_FLAGS,
+                )
```
daxhuiberts: nitpick: Put this call in its own function and use it from `load_all` and `open`.

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,305 @@
...
+        let value = json_iter_to_value(self.ctx, json_iter, self.api_version)?
+            .ok_or(LoadAllError::JsonRootMissing)?;
+
+        let rlk = if let Some(rlk) = rlookup.find_key_by_name(JSON_ROOT) {
+            rlk.into_current().unwrap()
+        } else {
+            rlookup
+                .get_key_load(JSON_ROOT, JSON_ROOT, RLookupKeyFlags::empty())
+                .unwrap()
+        };
```
daxhuiberts, line 107: nitpick: Maybe out of scope, but this block might warrant a `find_or_load` kind of method on `rlookup`.

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,305 @@
...
+
+        dst_row.write_key(rlk, value);
+
+        Ok(())
+    }
+}
+
+impl FieldLoader for JsonFieldLoader<'_> {
+    fn load_field(&self, kk: &RLookupKey, dst_row: &mut RLookupRow) -> Result<(), LoadFieldError> {
```
daxhuiberts: Why is it called `kk` and not just `key` or something?

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,305 @@
...
+        // For per-field loads, "field absent" is not an error — we just leave it unset
+        // and continue. Only hard failures bubble up as `Err`.
+        let val = if path.to_bytes().starts_with(JSON_ROOT.to_bytes()) {
+            let Some(iter) = self.value.get(path) else {
+                // JSONPath did not match any value in the document — skip silently.
+                return Ok(());
+            };
+            match json_iter_to_value(self.ctx, iter, self.api_version) {
+                Ok(Some(v)) => v,
+                Ok(None) | Err(_) => return Ok(()),
```
daxhuiberts, line 134: Is my understanding correct that that method never returns an `Err`? I expect the `Err` here to be propagated?

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    // SAFETY: `ctx` is a valid Redis module context by construction of `JsonFormat`.
+    // First get the JSON serialized value (since it does not consume the iterator)
+    let serialized = unsafe { iter.serialize(ctx.cast().as_ptr())? };
+
+    // Second, get the first JSON value. `is_empty()` returned false above, so the
+    // iterator is contractually obligated to yield at least one value here.
+    let Some(json) = iter.next() else {
+        debug_assert!(false, "ResultsIter::is_empty()/next() disagree");
```
daxhuiberts: 🥴

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,305 @@
...
+
+// Return an array of expanded values from an iterator.
+// The iterator is being reset and is not being freed.
+// Required japi_ver >= 4
+fn json_iter_to_value_expanded(
+    ctx: NonNull<ffi::RedisModuleCtx>,
+    mut iter: redis_json_api::ResultsIter<'_>,
+) -> SharedValue {
+    debug_assert!(!iter.is_empty(), "should be checked by caller");
+    iter.reset();
```
daxhuiberts: `iter.reset()` does not belong here. It would be better to `iter.clone()` in `json_iter_to_value` to get the first value or maybe an `iter.peek()` is possible? Or call `iter.reset()` in it's parent function.

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,305 @@
...
+    } else {
+        json_val_to_value(ctx, json)
+    };
+
+    let otherval = SharedValue::new_string(serialized.to_vec());
+    let expand = json_iter_to_value_expanded(ctx, iter);
+
+    Ok(Some(SharedValue::new(Value::Trio(Trio::new(
+        val, otherval, expand,
+    )))))
```
daxhuiberts: There are a few helper methods available on `SharedValue` besides `new_string` and `new_num`: `new_array`, `new_map`, `new_trio`. You can use them here and further in the code.

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,305 @@
...
+                        let value = json_val_to_value_expanded(ctx, value.as_ref());
+
+                        (key, value)
+                    })
+                    .collect();
+
+                SharedValue::new(Value::Map(Map::new(values)))
+            } else {
+                SharedValue::new(Value::Map(Map::new(Box::new([]))))
+            }
```
daxhuiberts: I think this `if len > 0` (here and below) is a renmant which is not needed anymore.

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,305 @@
...
+            let v = json.get_int().unwrap();
+            SharedValue::new_num(v as f64)
+        }
+        JsonType::Double => {
+            let v = json.get_double().unwrap();
+            SharedValue::new_num(v)
+        }
+        JsonType::Bool => {
+            let v = json.get_bool().unwrap();
+            SharedValue::new_num(v as u8 as f64)
```
daxhuiberts, line 273: Oh wow, floating booleans.

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+        JsonType::Object | JsonType::Array => {
+            // SAFETY: `ctx` is a valid Redis module context propagated from the caller.
+            let v = unsafe { json.serialize(ctx.cast().as_ptr()).unwrap() };
```
daxhuiberts, line 277: Should this not surface as a serialization error?

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,305 @@
...
+            SharedValue::new_string(v.to_string_lossy().into_bytes())
```
daxhuiberts: Do we want/need the conversion to happen through str?

src/redisearch_rs/rlookup/tests/load_document_json.rs
```diff
@@ -0,0 +1,314 @@
...
+
+fn assert_expanded_matches(actual: &SharedValue, expected: &serde_json::Value) {
+    match expected {
+        serde_json::Value::Object(map) => {
+            let Value::Map(m) = &**actual else {
+                panic!("expected Map, got {actual:?}")
+            };
+            assert_eq!(m.len(), map.len());
+            for (k, v) in map {
+                let got = m
```
daxhuiberts: Instead of `got` maybe call it `child` or `item`.

src/redisearch_rs/rlookup/tests/load_document_json.rs
```diff
@@ -0,0 +1,314 @@
...
+/// JSON scalars. Integers stay within `i64` (RedisJSON's `getInt`); floats finite.
+fn arb_json_scalar() -> impl Strategy<Value = serde_json::Value> {
+    prop_oneof![
+        Just(serde_json::Value::Null),
+        any::<bool>().prop_map(serde_json::Value::Bool),
+        any::<i64>().prop_map(|i| json!(i)),
+        any::<f64>()
+            .prop_filter("finite", |f| f.is_finite())
+            .prop_map(|f| json!(f)),
+        "[ -~]{0,16}".prop_map(serde_json::Value::String),
```
daxhuiberts: Is it also interesting to test non-ascii strings?

src/redisearch_rs/rlookup/tests/load_document_json.rs
```diff
@@ -0,0 +1,314 @@
...
+            panic!("multi apiVersion must yield a Trio, got {val:?}");
+        };
+        // .0 — first match, mapped flat.
+        assert_value_matches(trio.left(), &scalar);
+        // .1 — full serialized form of the match. A JSONPath query is always
+        // array-wrapped by RedisJSON's `getJSONFromIter`, even for a single match.
+        let serialized = serde_json::to_string(&json!([scalar])).unwrap();
+        assert_eq!(trio.middle().as_str_bytes(), Some(serialized.as_bytes()));
+        // .2 — expanded array over the matches (a single scalar here).
+        let Value::Array(expanded) = &**trio.right() else { panic!("Trio.2 must be an Array") };
```
daxhuiberts: nitpick: call it `trio.right` in the panic. (same below)

src/redisearch_rs/rlookup/tests/load_document_json.rs
```diff
@@ -0,0 +1,314 @@
...
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn expanded_mapping_recurses_and_preserves_empty_containers(map in proptest::collection::hash_map("[a-z]{1,6}", arb_json_value(), 1..5)) {
+        let value = serde_json::Value::Object(map.into_iter().collect());
+        let val = load_field_value(json!({ "x": value.clone() }), MULTI, c"$.x")
+            .expect("value should be loaded");
+        let Value::Trio(trio) = &*val else { panic!("expected Trio") };
+        let Value::Array(expanded) = &**trio.right() else { panic!("Trio.2 must be an Array") };
+        assert_eq!(expanded.len(), 1);
+        assert_expanded_matches(&expanded[0], &value);
```
daxhuiberts, line 311: A JSON object gets converted to an array with the object in it? That doesn't make sense does it?

daxhuiberts (changes requested):
Lots of small comments, but overall looks very good!

src/redisearch_rs/rlookup/src/load_document/json.rs
JonasKruckenberg: that naming convection was used by the C code throughout... 🤷🏻‍♂️ changed to `key`

src/redisearch_rs/rlookup/src/load_document/json.rs
JonasKruckenberg: yes! the C version ignores errors from `jsonIterToValue` here

https://github.com/RediSearch/RediSearch/blob/7935b55994c23aab24d0829132ac0432266c7b60/src/rlookup_load_document.c#L145-L147

don't ask me why, its a product decision...

src/redisearch_rs/rlookup/src/load_document/json.rs
JonasKruckenberg: nice! they didn't when I first wrote that code, but this makes it much nicer indeed

src/redisearch_rs/rlookup/src/load_document/json.rs
JonasKruckenberg: good point!

src/redisearch_rs/rlookup/src/load_document/json.rs
JonasKruckenberg: I don't make the rules 🤷🏻‍♂️ 

https://github.com/RediSearch/RediSearch/blob/7935b55994c23aab24d0829132ac0432266c7b60/src/json.c#L890-L892

src/redisearch_rs/rlookup/src/load_document/json.rs
JonasKruckenberg: good point, not really! I just missed that RedisString also has a `.to_vec()` on it (from slice)

src/redisearch_rs/rlookup/tests/load_document_json.rs
JonasKruckenberg: It's an array with one element per JSON path match that gets returned

src/redisearch_rs/rlookup/src/load_document/json.rs
JonasKruckenberg: that's going to cause all sorts of lifetime issues, not worth it

src/redisearch_rs/rlookup/src/load_document/json.rs
daxhuiberts: Maybe worth a comment in the code then.

src/redisearch_rs/rlookup/tests/load_document_json.rs
daxhuiberts: Ah oke!

daxhuiberts (dismissed):
Everything looks good! Only one minor suggest.

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+fn json_val_to_value_expanded(
+    ctx: NonNull<ffi::RedisModuleCtx>,
+    json: JsonValueRef,
+) -> SharedValue {
+    match json.get_type() {
+        JsonType::Object => {
+            // SAFETY: `ctx` is a valid Redis module context propagated from the caller.
+            let iter = unsafe { json.key_values(ctx.cast().as_ptr()).unwrap() };
```
kei-nan, line 229: C's `jsonValToValueExpanded` guards this with if (len) and only calls `getKeyValues` for a non-empty object.
I'm a bit concerned this can panic for json values {}.

Can you maybe check how C behaves and maybe we should modify the rust behaviour for that edge case.

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    // SAFETY: `ctx` is a valid Redis module context by construction of `JsonFormat`.
+    // First get the JSON serialized value (since it does not consume the iterator)
+    let serialized = unsafe { iter.serialize(ctx.cast().as_ptr())? };
+
+    // Second, get the first JSON value. `is_empty()` returned false above, so the
+    // iterator is contractually obligated to yield at least one value here.
+    let json = iter
+        .next()
+        .expect("ResultsIter::is_empty()/next() disagree");
```
kei-nan, line 183: I think C handles this in a more graceful way, are we sure we want to panic here?
If we already return a Result then maybe we can return an error instead?

I think this pattern repeats in a few other places where unwrap is called.
`get_at(i).unwrap()`, `len().unwrap()` for the array case

src/redisearch_rs/rlookup/src/load_document/json.rs
```diff
@@ -0,0 +1,282 @@
...
+        // Scalar
+        _ => json_val_to_value(ctx, json),
+    }
+}
+
+fn json_val_to_value(ctx: NonNull<ffi::RedisModuleCtx>, json: JsonValueRef<'_>) -> SharedValue {
+    // Currently `getJSON` cannot fail here also the other japi APIs below
+    match json.get_type() {
+        JsonType::String => {
+            let v = json.get_str().unwrap();
```
kei-nan, line 260: are we making Unicode assumptions here?
does C also stores the string as unicode?

src/redisearch_rs/rlookup/tests/load_document_json.proptest-regressions
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc a5427da3a7280c6d9af3700e3a20a1e6e298a17a28b7449bd903b50ed3d4a1af # shrinks to scalar = Null
```
kei-nan, line 7: what exactly is this file?
are we using a new test framework I am unfamiliar with?

kei-nan (commented):
did an initial pass, 
a bit concerned about the `unwrap` and `expect`
a json document is quite fragile out in the real world
we need to be reseilient as much as possible.

src/redisearch_rs/rlookup/src/load_document/json.rs
JonasKruckenberg: well, so here is the corresponding RedisJSON API rust binding function:

https://github.com/RediSearch/RediSearch/blob/7935b55994c23aab24d0829132ac0432266c7b60/src/redisearch_rs/redis_json_api/src/value.rs#L261-L269

we already query the size before constructing the iterator, but we still call `getKeyValues` unconditionally. Don't think this should panic but we can simply move the `len()` call to before the `getKeyValues` call in that Rust binding.

src/redisearch_rs/rlookup/src/load_document/json.rs
JonasKruckenberg: well we have already checked the iterator to be non-empty a couple lines above:

https://github.com/RediSearch/RediSearch/blob/7935b55994c23aab24d0829132ac0432266c7b60/src/redisearch_rs/rlookup/src/load_document/json.rs#L180-L183

so once we reach `iter.next()` is _must_ succeed since we did that `is_empty` check above. If at some point in the future for whatever reason `next` and `is_empty` disagree it is a hard programmer error somewhere in our code and the correct behavior here is to panic so we catch it during testing.

src/redisearch_rs/rlookup/tests/load_document_json.proptest-regressions
JonasKruckenberg: this is a proptest seed file that will automatically be generated when proptest finds a bug and does a reduction (so on subsequent runs is checks these known "problematic" inputs first before going into its random-walk search)

we've been using proptest for at least a year now. If you're unfamiliar its a very useful tool that does *property testing* and helps us test the entire range of possible inputs to our functions.

src/redisearch_rs/rlookup/src/load_document/json.rs
JonasKruckenberg: good question: RFC 8259 demands UTF8 for JSON objects that are exchanged between systems (redis as a DB by design fits this bill) and RedisJSON uses the rust `serde_json` crate under the hood which itself also assumes valid UTF8 when parsing. 

To confirm this experimentally I set Claude to generate an integration test file:

```python
from common import *
from includes import *


@skip(no_json=True)
def testJsonSetRawNonUtf8(env):
    """Can raw non-UTF-8 bytes land inside a JSON string value via JSON.SET?"""

    # A JSON string whose contents are raw invalid UTF-8 (0xff is never valid).
    payload = b'{"s":"' + b'\xff\xfe' + b'"}'
    try:
        res = conn.execute_command('JSON.SET', 'raw', '$', payload)
        env.debugPrint(f"raw JSON.SET accepted: {res!r}", force=True)
        got = conn.execute_command('JSON.GET', 'raw', '$.s')
        env.debugPrint(f"raw JSON.GET -> {got!r}", force=True)
    except Exception as e:
        env.debugPrint(f"raw JSON.SET REJECTED: {e!r}", force=True)


@skip(no_json=True)
def testJsonSetLoneSurrogate(env):
    """A \\uXXXX escape that is a lone surrogate is syntactically valid JSON
    but cannot be encoded as valid UTF-8. Does RedisJSON accept it?"""

    for esc in [r'"\ud800"', r'"\udfff"', r'"\ud800\ud800"']:
        payload = '{"s":' + esc + '}'
        try:
            res = conn.execute_command('JSON.SET', 'sur', '$', payload)
            env.debugPrint(f"surrogate {esc} accepted: {res!r}", force=True)
            got = conn.execute_command('JSON.GET', 'sur', '$.s')
            env.debugPrint(f"surrogate {esc} JSON.GET -> {got!r}", force=True)
        except Exception as e:
            env.debugPrint(f"surrogate {esc} REJECTED: {e!r}", force=True)


@skip(no_json=True)
def testJsonSetValidSurrogatePair(env):
    """Sanity: a valid surrogate pair (emoji) must round-trip fine."""
    payload = r'{"s":"😀"}'  # U+1F600
    res = conn.execute_command('JSON.SET', 'pair', '$', payload)
    env.debugPrint(f"pair accepted: {res!r}", force=True)
    got = conn.execute_command('JSON.GET', 'pair', '$.s')
    env.debugPrint(f"pair JSON.GET -> {got!r}", force=True)


@skip(no_json=True)
def testSearchLoadNonUtf8(env):
    """The real question for RediSearch: if a non-UTF-8-ish string value did
    get stored, does loading it via FT.SEARCH exercise a code path that assumes
    UTF-8? Index the field as TEXT and force a load."""

    env.cmd('FT.CREATE', 'idx', 'ON', 'JSON', 'SCHEMA', '$.s', 'AS', 's', 'TEXT')

    for i, esc in enumerate([r'"\ud800"', r'"\udfff"', r'"hello"']):
        payload = '{"s":' + esc + '}'
        try:
            conn.execute_command('JSON.SET', f'doc:{i}', '$', payload)
            env.debugPrint(f"doc:{i} set with {esc}", force=True)
        except Exception as e:
            env.debugPrint(f"doc:{i} JSON.SET {esc} REJECTED: {e!r}", force=True)

    # Force RediSearch to load the stored JSON string values back.
    res = conn.execute_command('FT.SEARCH', 'idx', '*', 'RETURN', '1', 's')
    env.debugPrint(f"FT.SEARCH -> {res!r}", force=True)
```

that confirmed against current master, invalid UTF8 such as lone surrogates and raw bytes cannot even be stored in redis. 

The UTF8 assumption the JSON API Rust bindings therefore make is well backed up.

src/redisearch_rs/rlookup/src/load_document/json.rs
JonasKruckenberg: applied the change!

kei-nan (dismissed):

kei-nan (approved):

PR #8109, opened 2026-01-20
Fix config registration - [RED-171841]

GuyAv46:
Register cluster configurations (used by the coordinator) not only when `cluster-enable` config is set,
But also when running Redis Enterprise, where this config is not necessarily set.
Also remove minimal Redis version check, since in RediSearch 8.x, Redis version is guaranteed to be 8.x.

#### Mark if applicable

---

> [!NOTE]
> Adjusts how module configuration is initialized and registered.
> 
> - Makes `RediSearch_InitModuleConfig` `static`, removes the `registerConfiguration` arg, and always calls `RegisterModuleConfig_Local`
> - Registers cluster configuration when `isClusterEnabled` or `clusterConfig.type == RedisLabs`
> - Removes unused version gating logic and updates caller to new signature
> - Drops the now-unused declaration from `module.h`
> - Minor whitespace cleanup in `FT.SYNDUMP` path
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ea71fa7346a90e5e9142db9a3f19c025effa6ac7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -4133,7 +4133,7 @@ static bool checkClusterEnabled(RedisModuleCtx *ctx) {
 
 int ConfigCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc);
 
-int RediSearch_InitModuleConfig(RedisModuleCtx *ctx, RedisModuleString **argv, int argc, int isClusterEnabled) {
+static int RediSearch_InitModuleConfig(RedisModuleCtx *ctx, RedisModuleString **argv, int argc, int isClusterEnabled) {
   // register the module configuration with redis, use loaded values from command line as defaults
   if (RegisterModuleConfig_Local(ctx) == REDISMODULE_ERR) {
     RedisModule_Log(ctx, "warning", "Error registering module configuration");
--- a/src/module.h
+++ b/src/module.h
@@ -40,7 +40,6 @@ extern "C" {
 // docs and code.
 #define CMD_INTERNAL "internal"
 
-int RediSearch_InitModuleConfig(RedisModuleCtx *ctx, RedisModuleString **argv, int argc, int registerConfig, int isClusterEnabled);
 int RediSearch_InitModuleInternal(RedisModuleCtx *ctx);
 
 extern redisearch_thpool_t *depleterPool;


src/module.c
```diff
@@ -4189,12 +4187,10 @@ RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
 
   // Check if we are actually in cluster mode
   const bool isClusterEnabled = checkClusterEnabled(ctx);
-  const Version unstableRedis = {7, 9, 227};
-  const bool unprefixedConfigSupported = (CompareVersions(redisVersion, unstableRedis) >= 0) ? true : false;
 
   legacySpecRules = dictCreate(&dictTypeHeapHiddenStrings, NULL);
 
-  if (RediSearch_InitModuleConfig(ctx, argv, argc, unprefixedConfigSupported, isClusterEnabled) == REDISMODULE_ERR) {
+  if (RediSearch_InitModuleConfig(ctx, argv, argc, isClusterEnabled) == REDISMODULE_ERR) {
```
oshadmi, line 4193: We can rely on Redis 8 being used. No need to verify anymore. Right?

src/module.c
```diff
@@ -4133,17 +4133,15 @@ static bool checkClusterEnabled(RedisModuleCtx *ctx) {
...
-    }
-    if (isClusterEnabled) {
-      // Register module configuration parameters for cluster
-      RM_TRY_F(RegisterClusterModuleConfig, ctx);
-    }
+  if (RegisterModuleConfig_Local(ctx) == REDISMODULE_ERR) {
+    RedisModule_Log(ctx, "warning", "Error registering module configuration");
+    return REDISMODULE_ERR;
+  }
+  if (isClusterEnabled || clusterConfig.type == ClusterType_RedisLabs) {
```
oshadmi, line 4142: Would be better to do it in `checkClusterEnabled`?
(Cluster can be enabled in Enterprise also when `cluster-enabled` config is unset, right?)

src/module.c
GuyAv46: Right. `master`, `8.6`, `8.4`, and `8.2` already deny loading into redis < 8

src/module.c
GuyAv46: `checkClusterEnabled` fetches the configuration. I don't want to mix logic into it

GuyAv46: Right, but they also start using it now. We need to be careful when using this value

oshadmi: This is why I suggested to add it to `checkClusterEnabled` so we would not need to be careful, just call it.
If we want another function to fetch the config, and to also have some logic - wouldn't it be better?

GuyAv46: But since Enterprise start using it now, for specific scenarios, we won’t be able to detect them by integrating the logic. 
The logic here is only correct in this context, I still think the helper to get the config should only do that

PR #8116, opened 2026-01-21
refactor: enable Rust pointer safety lints

JonasKruckenberg:
This change turns on a few stricter pointer safety lints for Rust to help catch thing like alignment-violating casts and as conversions.

There are a few callsites we should check:
- https://github.com/RediSearch/RediSearch/blob/2b9b08384150e7da5a4d3f4383c4f5d82f794bf0/src/redisearch_rs/value/src/rs_value_ffi.rs#L149 is NOT sound
- ~~https://github.com/RediSearch/RediSearch/blob/2b9b08384150e7da5a4d3f4383c4f5d82f794bf0/src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs#L533 @LukeMathWalker can you double check this is sound? casting `*const Box<T>` to `*const *const T` is not actually possible is it?~~ resolved

#### Mark if applicable

---

> [!NOTE]
> Strengthens pointer safety and FFI correctness without changing behavior.
> 
> - Enables strict clippy lints for pointer safety in `Cargo.toml` (e.g., `cast_ptr_alignment`, `ptr_as_ptr`, `ptr_cast_constness`, `ref_as_ptr`, `transmute_ptr_to_ptr`); updates `ffi` crate allowlist accordingly
> - Replaces raw `as` pointer casts with `ptr::from_mut`, `.cast()`, and `.cast_const()` across `buffer`, `iterators_ffi`, `types_ffi`, `value`, `redis_mock`, `rqe_iterators(_interop)`, `inverted_index`, `thin_vec`, `trie_rs`, etc.
> - Adds targeted `#[expect(clippy::cast_ptr_alignment)]`, uses CStr literals, const assertions for pointer/Box layout, and minor const/typing cleanups (e.g., `ThinVec::new`, `FilterGeoReader::new`, `RedisModule_KeyType`)
> - Maintains existing APIs/behavior; changes focus on unsafe code hygiene and alignment/constness correctness
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b4a6c5b0d1504925da42e29c7be586abd6e1b9df. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/Cargo.toml
+++ b/src/redisearch_rs/Cargo.toml
@@ -60,6 +60,13 @@ missing_const_for_fn = "warn"
 # These types are specified in clippy.toml.
 disallowed_types = "warn"
 
+# pointer safety
+cast_ptr_alignment = "deny"
+ptr_as_ptr = "deny"
+ptr_cast_constness = "deny"
+ref_as_ptr = "deny"
+transmute_ptr_to_ptr = "deny"
+
 [workspace.package]
 version = "0.0.1"
 edition = "2024"
--- a/src/redisearch_rs/Cargo.toml
+++ b/src/redisearch_rs/Cargo.toml
@@ -61,11 +61,11 @@ missing_const_for_fn = "warn"
 disallowed_types = "warn"
 
 # pointer safety
-cast_ptr_alignment = "deny"
-ptr_as_ptr = "deny"
-ptr_cast_constness = "deny"
-ref_as_ptr = "deny"
-transmute_ptr_to_ptr = "deny"
+cast_ptr_alignment = "warn"
+ptr_as_ptr = "warn"
+ptr_cast_constness = "warn"
+ref_as_ptr = "warn"
+transmute_ptr_to_ptr = "warn"
 
 [workspace.package]
 version = "0.0.1"
--- a/src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
+++ b/src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
@@ -521,6 +521,20 @@ pub unsafe extern "C" fn AggregateResult_GetRecordsSlice(
 ) -> AggregateRecordsSlice {
     debug_assert!(!agg.is_null(), "agg must not be null");
 
+    // Assert that we can safely cast `Box<RSIndexResult<'static>>` into `*mut RSIndexResult<'static>`
+    // which we can only do when the Global allocator remains a zero-sized type by extension ensuring
+    // the size and alignment remain the same. If this ever changes, we'll fail to compile.
+    const {
+        assert!(
+            size_of::<*mut RSIndexResult<'static>>() - size_of::<Box<RSIndexResult<'static>>>()
+                == 0
+        );
+        assert!(
+            align_of::<*mut RSIndexResult<'static>>() - align_of::<Box<RSIndexResult<'static>>>()
+                == 0
+        );
+    };
+
     // SAFETY: Caller is to ensure that the pointer `agg` is a valid, non-null pointer to
     // an `RSAggregateResult`.
     let agg = unsafe { &*agg };
--- a/src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
+++ b/src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
@@ -526,11 +526,11 @@ pub unsafe extern "C" fn AggregateResult_GetRecordsSlice(
     let agg = unsafe { &*agg };
     match agg {
         RSAggregateResult::Borrowed { records, .. } => AggregateRecordsSlice {
-            ptr: records.as_slice().as_ptr() as *const *const RSIndexResult,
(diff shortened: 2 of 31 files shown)


src/redisearch_rs/Cargo.toml
```diff
@@ -60,6 +60,13 @@ missing_const_for_fn = "warn"
 # These types are specified in clippy.toml.
 disallowed_types = "warn"
 
+# pointer safety
+cast_ptr_alignment = "deny"
+ptr_as_ptr = "deny"
+ptr_cast_constness = "deny"
+ref_as_ptr = "deny"
+transmute_ptr_to_ptr = "deny"
```
LukeMathWalker: We should always set them to `warn`, rather than `deny`, as we do for the others above.

LukeMathWalker:
>  ptr: records.as_slice().as_ptr() as *const *const RSIndexResult, 

Based on

> So long as T: Sized, a Box<T> is guaranteed to be represented as a single pointer and is also ABI-compatible with C pointers

I think it should be fine? Since it should follow that the layout of an array of boxes matches the layout of an array of raw pointers. But you're probably stronger on these finer issues than I am.

Architecturally, that piece of code exists to avoid using an iterator at the C/Rust boundary, thus minimizing FFI overhead. If we port the C callsites over, we can remove that cast entirely.

JonasKruckenberg:
> > ptr: records.as_slice().as_ptr() as *const *const RSIndexResult,
> 
> Based on
> 
> > So long as T: Sized, a Box is guaranteed to be represented as a single pointer and is also ABI-compatible with C pointers
> 
> I think it should be fine? Since it should follow that the layout of an array of boxes matches the layout of an array of raw pointers. But you're probably stronger on these finer issues than I am.
> 
> Architecturally, that piece of code exists to avoid using an iterator at the C/Rust boundary, thus minimizing FFI overhead. If we port the C callsites over, we can remove that cast entirely.

Ahh well, okay yes `core::alloc::Global` is actually zero-sized so this happens to be fine 😅. Might be worth just adding a const assertion for in case that ever changes, bc Box is `(Unique<T, A>)` so only if `A` is ZST that cast is fine.

src/redisearch_rs/Cargo.toml
JonasKruckenberg: sure 👍🏻

LukeMathWalker:
> > > ptr: records.as_slice().as_ptr() as *const *const RSIndexResult,
> > 
> > 
> > Based on
> > > So long as T: Sized, a Box is guaranteed to be represented as a single pointer and is also ABI-compatible with C pointers
> > 
> > 
> > I think it should be fine? Since it should follow that the layout of an array of boxes matches the layout of an array of raw pointers. But you're probably stronger on these finer issues than I am.
> > Architecturally, that piece of code exists to avoid using an iterator at the C/Rust boundary, thus minimizing FFI overhead. If we port the C callsites over, we can remove that cast entirely.
> 
> Ahh well, okay yes `core::alloc::Global` is actually zero-sized so this happens to be fine 😅. Might be worth just adding a const assertion for in case that ever changes, bc Box is `(Unique<T, A>)` so only if `A` is ZST that cast is fine.

100%, this is under the assumption that everything in there is using the global allocator, that doesn't take any space.

JonasKruckenberg:
> 100%, this is under the assumption that everything in there is using the global allocator, that doesn't take any space.

yeah that's fair. I've added a [const assertion nonetheless ](https://github.com/RediSearch/RediSearch/pull/8116/changes/6bc5777a46bf846f5680940d0427710fc1dfd517) seeing as how that allocator component and `Global` are still nightly I don't fully trust it to not break in a non-edition change.

src/redisearch_rs/c_entrypoint/types_ffi/src/lib.rs
```diff
@@ -521,6 +521,20 @@ pub unsafe extern "C" fn AggregateResult_GetRecordsSlice(
 ) -> AggregateRecordsSlice {
     debug_assert!(!agg.is_null(), "agg must not be null");
 
+    // Assert that we can safely cast `Box<RSIndexResult<'static>>` into `*mut RSIndexResult<'static>`
+    // which we can only do when the Global allocator remains a zero-sized type by extension ensuring
+    // the size and alignment remain the same. If this ever changes, we'll fail to compile.
+    const {
+        assert!(
+            size_of::<Box<RSIndexResult<'static>>>() - size_of::<*mut RSIndexResult<'static>>()
```
LukeMathWalker: No `assert_eq!` due to `const` trait limitations, I assume?

JonasKruckenberg: yeah :/

JonasKruckenberg: also actually when you do `assert!(A - B == 0)` in const assertions the compiler message actually includes both numbers that you tried to subtract, while `assert!(A == B)` does not. 

But oooonly if `B` is larger than `A` (bc numeric overflow prints the sizes while the failing assertion doesnt)

PR #8117, opened 2026-01-21
chore: enable Rust numeric safety lints

JonasKruckenberg:
This change turns on a few stricter numeric safety lints for Rust to help catch things like unexpected truncating conversions.

There are a few callsites we should double check:
- https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/inverted_index/src/lib.rs#L958 @chesedo I assume precision loss is not a concern here because we never have 52+bit numbers right?
- https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/inverted_index/src/offsets_only.rs#L41
https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/inverted_index/src/full.rs#L206
https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/inverted_index/src/full.rs#L60
https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/inverted_index/src/freqs_offsets.rs#L41
https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/inverted_index/src/fields_offsets.rs#L146
https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/inverted_index/src/fields_offsets.rs#L50
@GlenDC can we actually convert `usize`s to `u32`s like this?
- https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/rqe_iterators/src/wildcard.rs#L79 https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/rqe_iterators/src/optional.rs#L202 https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/rqe_iterators/src/not.rs#L184
@GlenDC why are we converting to `usize` here in the first place? feels like we should pick either `usize` or `u64` and use that consistently
- https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/rqe_iterators/src/inverted_index.rs#L383 @chesedo same here converting to `usize` here feels misplaced.
-  ~~https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/rqe_iterators/src/profile.rs#L73 @gdesmott @meiravgri can you double check this assumption actually holds?~~ resolved
- ~~https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/rqe_iterators/src/inverted_index.rs#L266 @gdesmott I've replaced this with a checked conversion, unchecked truncating from 128bits to 32bit seems sketchy especially with a caller provided mask.~~ resolved
- ~~https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/c_entrypoint/iterators_ffi/src/id_list.rs#L31
https://github.com/RediSearch/RediSearch/blob/612b698c80c4451c0f7138bff9b964e579607d89/src/redisearch_rs/c_entrypoint/iterators_ffi/src/id_list.rs#L49 I've changed it so we cast `num` to a usize in these places so `new_id_list_iterator` can just deal with `usize`s but that doesn't work if Redis supports 32bit deployments. we should check whether that's the case or not.~~ resolved

#### Mark if applicable

---

> [!NOTE]
> Sets stricter numeric-safety linting across the workspace and updates code to avoid truncation/precision/sign issues.
> 
> - **Lints**: Enables `cast_*` and `checked_conversions` in `Cargo.toml`; adds localized `#[expect/allow]` where appropriate
> - **FFI/API tweaks**: `New{Sorted,Unsorted}IdListIterator` now take `size_t num` (C header updated) and crate depends on `libc`; several FFI functions add safe conversions (e.g., `RPStatus` via `try_from`, `serverVersion` now `u32`)
> - **Safer numeric conversions**: Replace `as` with `from/try_from` across encoders/decoders (`inverted_index/*`), `qint`/`varint`, and hashing (`fnv`); use `i64::from`, `u64::from`, and checked usize/u32 conversions
> - **Pointer/math fixes**: Use `byte_add` instead of `byte_offset`; guard casts with `try_from` and assertions; adjust seek/length math to avoid sign loss
> - **Misc**: ThinVec capacity and trie/node operations use checked conversions; added targeted `#[expect]` annotations where loss is intentional
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b6dd9c8ffc6af392689ea0a1981709806d0f0256. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/fnv_ffi/src/lib.rs
+++ b/src/redisearch_rs/c_entrypoint/fnv_ffi/src/lib.rs
@@ -29,7 +29,9 @@ pub unsafe extern "C" fn rs_fnv_32a_buf(buf: *const c_void, len: usize, hval: u3
 
     fnv.write(bytes);
 
-    fnv.finish() as u32
+    // Safety: Fnv32 produces 32-bit hashes but is forced to return a 64-bit value due to the Hash trait.
+    // values are always safe to cast to 32-bit though
+    unsafe { u32::try_from(fnv.finish()).unwrap_unchecked() }
 }
 
 /// Returns the 64-bit [FNV-1a hash] of `buf` of length `len` using an [offset basis] `hval`.
--- a/src/redisearch_rs/c_entrypoint/iterators_ffi/src/id_list.rs
+++ b/src/redisearch_rs/c_entrypoint/iterators_ffi/src/id_list.rs
@@ -9,10 +9,10 @@
 
 use ffi::{IteratorType_ID_LIST_ITERATOR, QueryIterator, RedisModule_Free, t_docId};
 use inverted_index::RSIndexResult;
+use libc::size_t;
 use rqe_iterators::id_list::IdList;
 use rqe_iterators_interop::RQEIteratorWrapper;
 
-#[unsafe(no_mangle)]
 /// Creates a new iterator over a list of sorted document IDs.
 ///
 /// # Safety
@@ -22,20 +22,16 @@ use rqe_iterators_interop::RQEIteratorWrapper;
 /// 2. The caller must ensure that `ids` is not null unless `num` is zero.
 /// 3. The memory pointed to by `ids` will be freed using `RedisModule_Free`,
 ///    so the caller must ensure that the pointer was allocated in a compatible manner.
-#[expect(
-    clippy::cast_possible_truncation,
-    reason = "redis does not support 32bit"
-)]
+#[unsafe(no_mangle)]
 pub unsafe extern "C" fn NewSortedIdListIterator(
     ids: *const t_docId,
-    num: u64,
+    num: size_t,
     weight: f64,
 ) -> *mut QueryIterator {
     // SAFETY: All safety preconditions are guaranteed by the caller.
-    unsafe { new_id_list_iterator::<true>(ids, num as usize, weight) }
+    unsafe { new_id_list_iterator::<true>(ids, num, weight) }
 }
 
-#[unsafe(no_mangle)]
 /// Creates a new iterator over a list of unsorted document IDs.
 ///
 /// # Safety
@@ -44,17 +40,14 @@ pub unsafe extern "C" fn NewSortedIdListIterator(
 /// 2. The caller must ensure that `ids` is not null unless `num` is zero.
 /// 3. The memory pointed to by `ids` will be freed using `RedisModule_Free`,
 ///    so the caller must ensure that the pointer was allocated in a compatible manner.
-#[expect(
-    clippy::cast_possible_truncation,
-    reason = "redis does not support 32bit"
-)]
+#[unsafe(no_mangle)]
 pub unsafe extern "C" fn NewUnsortedIdListIterator(
     ids: *const t_docId,
-    num: u64,
+    num: size_t,
     weight: f64,
 ) -> *mut QueryIterator {
     // SAFETY: All safety preconditions are guaranteed by the caller.
-    unsafe { new_id_list_iterator::<false>(ids, num as usize, weight) }
+    unsafe { new_id_list_iterator::<false>(ids, num, weight) }
 }
 
 /// # Safety
--- a/src/redisearch_rs/c_entrypoint/iterators_ffi/src/id_list.rs
+++ b/src/redisearch_rs/c_entrypoint/iterators_ffi/src/id_list.rs
@@ -22,13 +22,17 @@ use rqe_iterators_interop::RQEIteratorWrapper;
 /// 2. The caller must ensure that `ids` is not null unless `num` is zero.
 /// 3. The memory pointed to by `ids` will be freed using `RedisModule_Free`,
 ///    so the caller must ensure that the pointer was allocated in a compatible manner.
+#[expect(
+    clippy::cast_possible_truncation,
+    reason = "redis does not support 32bit"
+)]
 pub unsafe extern "C" fn NewSortedIdListIterator(
     ids: *const t_docId,
     num: u64,
     weight: f64,
 ) -> *mut QueryIterator {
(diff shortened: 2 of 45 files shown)


gdesmott:
> @gdesmott I've replaced this with a checked conversion, unchecked truncating from 128bits to 32bit seems sketchy especially with a caller provided mask.

That's ok, we check for `IndexFlags_Index_WideSchema` right before.

>  @gdesmott @meiravgri can you double check this assumption actually holds?

Yeah, we don't expect iterators to compute for more than 584 years :smile:

JonasKruckenberg:
> > @gdesmott I've replaced this with a checked conversion, unchecked truncating from 128bits to 32bit seems sketchy especially with a caller provided mask.
> 
> That's ok, we check for `IndexFlags_Index_WideSchema` right before.
> 
> > @gdesmott @meiravgri can you double check this assumption actually holds?
> 
> Yeah, we don't expect iterators to compute for more than 584 years 😄

okay thanks for checking, sounds reasonable!

src/redisearch_rs/c_entrypoint/iterators_ffi/src/id_list.rs
```diff
@@ -40,13 +44,17 @@ pub unsafe extern "C" fn NewSortedIdListIterator(
 /// 2. The caller must ensure that `ids` is not null unless `num` is zero.
 /// 3. The memory pointed to by `ids` will be freed using `RedisModule_Free`,
 ///    so the caller must ensure that the pointer was allocated in a compatible manner.
+#[expect(
+    clippy::cast_possible_truncation,
+    reason = "redis does not support 32bit"
```
oshadmi: Need to support 32bit

JonasKruckenberg: yep 😄

JonasKruckenberg: replaced this now with `size_t` which was the appropriate type to use to begin with (since the C callsites use that already)

src/redisearch_rs/inverted_index/src/lib.rs
```diff
@@ -955,6 +955,7 @@ impl<E: Encoder> EntriesTrackingIndex<E> {
     }
 
     /// Return the debug summary for this inverted index.
+    #[expect(clippy::cast_precision_loss, reason = "TODO")]
```
chesedo, line 958: Shouldn't this expect be on line 656?

JonasKruckenberg: this disables the lint that checks for precision loss because casting `u64` -> `f64` will loose precision for big numbers. (which I think is fine in this case because we don't really expect huge (52bit+) counts to show up here. And even then, this is a debug metric right?

chesedo: Ohhh, I see. This is for when we calculate the block efficiency. Yeah, 4 quadrillion will not be expected :joy:

src/redisearch_rs/c_entrypoint/fnv_ffi/src/lib.rs
```diff
@@ -29,7 +29,9 @@ pub unsafe extern "C" fn rs_fnv_32a_buf(buf: *const c_void, len: usize, hval: u3
 
     fnv.write(bytes);
 
-    fnv.finish() as u32
+    // Safety: Fnv32 produces 32-bit hashes but is forced to return a 64-bit value due to the Hash trait.
+    // values are always safe to cast to 32-bit though
+    unsafe { u32::try_from(fnv.finish()).unwrap_unchecked() }
```
GlenDC, line 34: but there's also `finish32` from what I can see, so why not use that? and anyway it's a custom type so I don't see why we have to be so akward here?

I'm assuming I am missing something obvious here though, but I just do not see it...

src/redisearch_rs/inverted_index/src/full.rs
```diff
@@ -57,7 +57,7 @@ impl Encoder for Full {
                 .expect("Need to use the wide variant of the Full encoder to support field masks bigger than u32");
 
         let offsets = offsets(record);
-        let offsets_sz = offsets.len() as u32;
+        let offsets_sz = u32::try_from(offsets.len()).unwrap();
```
GlenDC, line 60: would at the very least make this an expexct with an explanation, even better if we can avoid it

GlenDC (changes requested):

GlenDC:
> why are we converting to usize here in the first place? feels like we should pick either usize or u64 and use that consistently

I agree. I would say go for usize, as that is usually more correct answer for "number" of things . However in case there are valid reason (which I do not see) I suppose it would be u64. But you're right probably should be consistent instead of casting all over the place.

> @GlenDC can we actually convert usizes to u32s like this?

I am not sure I follow? The referenced locations cast from t_docId to usize, not to u32?

PR #8122, opened 2026-01-21
MOD-12273: Use Async Dmd Read In Result Processor

kei-nan:
Use Disk API to read dmd in an async way.

---

> [!NOTE]
> **High Risk**
> Touches the core query iteration path for disk-backed indexes and introduces new async I/O pooling/ownership logic, which increases risk of ordering, timeout, and memory-lifecycle bugs. Also extends the disk API surface and adds runtime toggling that can change behavior between queries.
> 
> **Overview**
> **Disk-backed query processing can now fetch document metadata (DMD) asynchronously.** The root result processor (`RPQueryIterator`) gains an async path that buffers `RSIndexResult`s, submits batched async DMD reads via a new pool API, and returns results once reads complete (with a short blocking poll only after iterator EOF).
> 
> This introduces a new `IndexResultAsyncReadState` helper (buffer/pending/ready stages), extends `search_disk`/`search_disk_api` with async read-pool creation/add/poll/free + a global `asyncIOEnabled` flag, and adds `FT.DEBUG DISK_IO_CONTROL` (plus tests/array+dllist utilities) to control and validate the new behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e078606321ee6150abbaa2ba16906aa5f956030e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/index_result_async_read.c
+++ b/src/index_result_async_read.c
@@ -114,6 +114,7 @@ void IndexResultAsyncRead_RefillPool(IndexResultAsyncReadState *state) {
 }
 
 static void IndexResultAsyncRead_CleanupFailedReads(IndexResultAsyncReadState *state) {
+  RS_ASSERT(array_len(state->failedUserData) <= state->poolSize);
   for (uint16_t i = 0; i < array_len(state->failedUserData); i++) {
     IndexResultNode *node = (IndexResultNode *)state->failedUserData[i];
 


src/result_processor.c
```diff
@@ -91,91 +101,221 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+static const RSDocumentMetadata* popReadyDmd(RPQueryIterator *self) {
+  if (self->readyDmdsIndex >= self->readyDmdsCount) {
+    return NULL;  // No more ready DMDs
+  }
+
+  RSDocumentMetadata *srcDmd = &self->readyDmds[self->readyDmdsIndex++];
+
+  // Allocate a new DMD and copy data from the buffer
+  RSDocumentMetadata *dmd = (RSDocumentMetadata *)rm_calloc(1, sizeof(RSDocumentMetadata));
+  dmd->ref_count = 1;
```
JoanFM: would memcpy work better?

JoanFM: and then set the ref_count to whatever

src/result_processor.c
```diff
@@ -91,91 +101,221 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+  dmd->keyPtr = srcDmd->keyPtr;  // Transfer ownership of keyPtr
+  dmd->score = srcDmd->score;
+  dmd->flags = srcDmd->flags;
+  dmd->maxTermFreq = srcDmd->maxTermFreq;
+  dmd->docLen = srcDmd->docLen;
+  dmd->id = srcDmd->id;
```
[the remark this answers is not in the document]
kei-nan: I think now that we use memcpy it will be null in the copied data.

src/result_processor.c
```diff
@@ -91,91 +101,221 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+      self->readyDmdsCount = pollResult.ready_count;
+      self->readyDmdsIndex = 0;
+
+      if (pollResult.ready_count > 0) {
+        continue;
+      }
+
+      if (pollResult.pending_count == 0 && it->atEOF) {
+        return UnlockSpec_and_ReturnRPResult(sctx, RS_RESULT_EOF);
+      }
```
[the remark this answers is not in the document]
kei-nan: right now the revalidation path is not applicable to disk

src/redisearch_rs/build_utils/src/lib.rs
```diff
@@ -160,7 +160,7 @@ fn link_c_plusplus() {
         .compile("link-cplusplus");
 }
 
-fn link_static_lib(
+pub fn link_static_lib(
     bin_root: &Path,
     lib_subdir: &str,
     lib_name: &str,
```
Itzikvaknin, line 166: Im not sure thats necessary, its only used in this file

src/result_processor.c
```diff
@@ -45,13 +46,39 @@ static int UnlockSpec_and_ReturnRPResult(RedisSearchCtx *sctx, int result_status
...
+  // disk reads complete and appear in readyResults for consumption.
+  RedisSearchDiskAsyncReadPool asyncPool;       // Async pool for disk-based indexes (NULL if not using disk)
+
+  uint16_t iteratorResultCount;                 // Number of nodes in iteratorResults list
+  DLLIST iteratorResults;                       // Deep-copied IndexResults from iterator, not yet submitted to async pool
+  DLLIST pendingResults;                        // Linked list (bounded by ASYNC_POOL_SIZE) tracking submitted IndexResults
+  arrayof(AsyncReadResult) readyResults;        // Ready results from async poll (DMD + user_data pairs)
+  uint16_t readyResultsIndex;                   // Next index to consume from readyResults
+  arrayof(uint64_t) failedUserData;             // user_data from failed async reads (not found/error)
+  RSIndexResult *lastReturnedIndexResult;       // Last deep-copied IndexResult returned (needs to be freed on next call)
```
Itzikvaknin: Consider grouping these together in a struct with a dedicated free function, I see they are all used only in the async path

src/result_processor.c
```diff
@@ -45,13 +46,39 @@ static int UnlockSpec_and_ReturnRPResult(RedisSearchCtx *sctx, int result_status
...
   uint32_t timeoutLimiter;                      // counter to limit number of calls to TimedOut_WithCounter()
   uint32_t keySpaceVersion;                     // version of the Keyspace slot ranges used for filtering
   const RedisModuleSlotRangeArray *querySlots;  // Query slots info, may be used for filtering
+
+  // Double-buffering for async disk I/O: iteratorResults holds deep-copied IndexResults from the
+  // iterator, which are submitted to the async pool and tracked in pendingResults until their
+  // disk reads complete and appear in readyResults for consumption.
+  RedisSearchDiskAsyncReadPool asyncPool;       // Async pool for disk-based indexes (NULL if not using disk)
+
+  uint16_t iteratorResultCount;                 // Number of nodes in iteratorResults list
```
Itzikvaknin: Why did you chose a list if thats a the size is bounded?

src/result_processor.c
```diff
@@ -45,13 +46,39 @@ static int UnlockSpec_and_ReturnRPResult(RedisSearchCtx *sctx, int result_status
...
+  DLLIST iteratorResults;                       // Deep-copied IndexResults from iterator, not yet submitted to async pool
+  DLLIST pendingResults;                        // Linked list (bounded by ASYNC_POOL_SIZE) tracking submitted IndexResults
+  arrayof(AsyncReadResult) readyResults;        // Ready results from async poll (DMD + user_data pairs)
```
Itzikvaknin: I was a bit confused by `arrayof(AsyncReadResult)`, since its not variable-length, your call here

src/result_processor.c
```diff
@@ -92,92 +119,385 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
-      // The iterator is still valid, but the current result has changed, or we are at EOF.
-      // If we are at EOF, we can enter the loop and let it handle it. (reading again should be safe)
-      goto validate_current;
-    }
+  bool validateCurrent = false;
+
+  // Handle spec lock and revalidation
+  if (handleSpecLockAndRevalidate(self)) {
+    validateCurrent = true;
   }
```
Itzikvaknin:
```suggestion
  validateCurrent = handleSpecLockAndRevalidate(self))
```

I can think of a reason to be more explicit about that

src/redisearch_rs/build_utils/src/lib.rs
Itzikvaknin: Its used in the disk side, the pub is needed

src/result_processor.c
```diff
@@ -92,92 +119,385 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+    // Step 3: No ready results - poll for more
+    int timeout_ms = it->atEOF ? 1000 : 0;
+
+    // Poll writes directly to the arrays (capacity is ASYNC_POOL_SIZE)
+    AsyncPollResult pollResult = SearchDisk_PollAsyncReads(
+        self->asyncPool, timeout_ms, self->readyResults, ASYNC_POOL_SIZE,
+        self->failedUserData, ASYNC_POOL_SIZE);
+
+    // Set array lengths to actual result counts (poll wrote directly to buffer)
+    array_hdr(self->readyResults)->len = pollResult.ready_count;
```
Itzikvaknin: Consider doing this in `SearchDisk_PollAsyncReads`

src/result_processor.c
```diff
@@ -92,92 +119,385 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+    // Step 5: Immediately refill async pool from buffer to keep I/O pipeline full
+    refillAsyncPool(self);
+
+    // Step 6: Check if we're completely done
+    // We're done only if: iterator is at EOF, no ready results, no in-flight async reads,
+    // and no buffered results waiting to be submitted
+    // Note: Use pollResult.pending_count to check for in-flight reads. The disk API always
+    // returns an entry in readyResults for every completed read (with status indicating
+    // success/not found/error), so pendingResults will be empty when pending_count == 0
+    if (it->atEOF && pollResult.ready_count == 0 &&
```
Itzikvaknin: It can be a function of the struct

src/result_processor.c
```diff
@@ -92,92 +119,385 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
 
-  // set the result data
-  SearchResult_SetDocId(res, it->lastDocId);
-  SearchResult_SetIndexResult(res, it->current);
-  SearchResult_SetScore(res, 0);
-  RS_LOG_ASSERT(SearchResult_GetDocumentMetadata(res) == NULL, "SearchResult already has associated document metadata");
-  SearchResult_SetDocumentMetadata(res, dmd);
-  RLookupRow_SetSortingVector(SearchResult_GetRowDataMut(res), dmd->sortVector);
-  return RS_RESULT_OK;
+    // Step 5: Immediately refill async pool from buffer to keep I/O pipeline full
```
Itzikvaknin: I don't think we need it, it will be called anyway in case we did not finish

src/result_processor.c
kei-nan: trying to prevent from holding all disk results in memory so had to bound the list somehow.

src/result_processor.c
kei-nan: it maintains for us the length of the array so it has its benefits.

src/result_processor.c
```diff
@@ -48,37 +48,38 @@ static int UnlockSpec_and_ReturnRPResult(RedisSearchCtx *sctx, int result_status
 }
 
 // Maximum number of concurrent async reads
-#define ASYNC_POOL_SIZE 16
-// Initial buffer size for IndexResults (starts small, grows dynamically)
-#define INDEX_RESULT_BUFFER_INITIAL_SIZE 16
-// Maximum buffer size (cap growth at 1024 entries = 8KB of pointers)
-#define INDEX_RESULT_BUFFER_MAX_SIZE 1024
+#define NAX_ONGOING_READ_SIZE 16
```
Itzikvaknin:
```suggestion
#define MAX_ONGOING_READ_SIZE 16
```

src/result_processor.c
```diff
@@ -92,105 +120,435 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+  RedisSearchCtx *sctx = self->sctx;
+  IndexSpec *spec = sctx->spec;
+
+  // Don't refill if iterator is done
+  if (it->atEOF) {
+    return RS_RESULT_OK;
+  }
+
+  // Fill buffer up to max capacity
+  while (self->async.iteratorResultCount < NAX_ONGOING_READ_SIZE && !it->atEOF) {
```
Itzikvaknin:
```suggestion
  while (self->async.iteratorResultCount < MAX_ONGOING_READ_SIZE && !it->atEOF) {
```

src/result_processor.c
```diff
@@ -92,105 +120,435 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+
+/**
+ * Refill the async pool from the IndexResult buffer.
+ * Returns number of entries added to the pool.
+ */
+static uint16_t refillAsyncPool(RPQueryIteratorAsyncState *async) {
+  uint16_t added = 0;
+
+  // Move nodes from iteratorResults to pendingResults
+  while (added < NAX_ONGOING_READ_SIZE && !DLLIST_IS_EMPTY(&async->iteratorResults)) {
```
Itzikvaknin: Who is reponsible to keep the `async->iteratorResults` bounded?
`added < MAX_ONGOING_READ_SIZE` doesn't enforce that

If we rely on `SearchDisk_AddAsyncRead` to fail in such case, then `added` is redundent IMO

src/result_processor.c
```diff
@@ -92,105 +120,435 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+    SearchDisk_FreeAsyncReadPool(self->async.asyncPool);
+  }
+
+  if (self->async.readyResults) {
+    // Free any remaining DMD data that wasn't consumed
+    for (uint16_t i = self->async.readyResultsIndex; i < array_len(self->async.readyResults); i++) {
+      if (self->async.readyResults[i].dmd) {
+        DMD_Return(self->async.readyResults[i].dmd);
+      }
+    }
```
Itzikvaknin: Consider using `array_free_ex`

src/result_processor.c
```diff
@@ -92,105 +120,435 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+  }
+
+  if (self->async.failedUserData) {
+    array_free(self->async.failedUserData);
+  }
+
+  // Free the last returned deep-copied IndexResult if any
+  if (self->async.lastReturnedIndexResult) {
+    IndexResult_Free(self->async.lastReturnedIndexResult);
+  }
```
Itzikvaknin:
```suggestion
  RPQueryIteratorAsyncState_Free(self->async)
```

src/result_processor.c
```diff
@@ -92,105 +120,435 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+/**
+ * Set the search result data from a DMD and IndexResult.
+ */
+static void setSearchResult(ResultProcessor *base, SearchResult *res, RSIndexResult *indexResult,
+                            const RSDocumentMetadata *dmd, t_docId docId) {
+  base->parent->totalResults++;
+  SearchResult_SetDocId(res, dmd->id);
+  SearchResult_SetIndexResult(res, indexResult);
+  SearchResult_SetScore(res, 0);
+  RS_LOG_ASSERT(SearchResult_GetDocumentMetadata(res) == NULL, "SearchResult already has associated document metadata");
```
Itzikvaknin: Nitpick 
```suggestion
   RS_LOG_ASSERT(SearchResult_GetDocumentMetadata(res) == NULL, "SearchResult already has associated document metadata");
  base->parent->totalResults++;
  SearchResult_SetDocId(res, dmd->id);
  SearchResult_SetIndexResult(res, indexResult);
  SearchResult_SetScore(res, 0);
 
```

kei-nan: I don't understand what changed in your suggestion

src/result_processor.c
kei-nan: this will bound the size of the iterator result count:
`while (self->async.iteratorResultCount < self->async.poolSize && !it->atEOF) {`
in `refillIndexResultBuffer`

src/result_processor.c
kei-nan: code was changed, this was fixed
marking as resolved

src/index_result_async_read.h
```diff
@@ -0,0 +1,149 @@
...
+ * @return Number of entries added to the async pool
+ */
+uint16_t IndexResultAsyncRead_RefillPool(IndexResultAsyncReadState *state);
+
+/**
+ * Pop a ready result from the completed async reads
+ *
+ * Returns an IndexResult with its DMD field populated from a completed
+ * async disk read. The caller takes ownership of the IndexResult and
+ * must free it (or store it in lastReturnedIndexResult for deferred cleanup).
```
Itzikvaknin: Its a bit weird to have ambiguous ownership IMO
Maybe this can be resolved with a bit of effort - your call on effort/value

src/index_result_async_read.c
```diff
@@ -0,0 +1,184 @@
...
+
+  if (!state->readyResults || !state->failedUserData) {
+    // Cleanup on allocation failure
+    array_free(state->readyResults);
+    array_free(state->failedUserData);
+    state->readyResults = NULL;
+    state->failedUserData = NULL;
+    state->asyncPool = NULL;
+    return false;
+  }
```
Itzikvaknin: I think we can remove it, since we don't do these check in the codebase, for exaple in this PR:
```
    IndexResultNode *node = rm_calloc(1, sizeof(*node));
    node->result = copy;
    dllist_append(&self->async.iteratorResults, &node->node);
    self->async.iteratorResultCount++;
```

src/index_result_async_read.c
```diff
@@ -0,0 +1,184 @@
...
+  state->asyncPool = NULL;
+  state->readyResultsIndex = 0;
+  state->lastReturnedIndexResult = NULL;
+}
+
+bool IndexResultAsyncRead_SetupAsyncPool(IndexResultAsyncReadState *state,
+                                         RedisSearchDiskAsyncReadPool asyncPool) {
+  if (!asyncPool) {
+    return false;
+  }
```
Itzikvaknin:
```suggestion
  RS_ASSERT(asyncPool);
```

src/index_result_async_read.c
```diff
@@ -0,0 +1,184 @@
...
+
+  // Free nodes in iteratorResults list
+  DLLIST_node *dlnode;
+  while ((dlnode = dllist_pop_tail(&state->iteratorResults)) != NULL) {
+    IndexResultNode *node = DLLIST_ITEM(dlnode, IndexResultNode, node);
+    if (node->result) {
+      IndexResult_Free(node->result);
+    }
+    rm_free(node);
+  }
```
Itzikvaknin, line 56: Do you think it worth the effort to implement `DLLIST_Free(DLLIST, NodeFreeCallback)`?
Anyway, not a blocker for this PR

tests/cpptests/test_cpp_async_state.cpp
```diff
@@ -0,0 +1,565 @@
...
+    state.iteratorResultCount--;
+  }
+
+  // Helper: Verify state consistency
+  void assertStateConsistent() {
+    // Count actual nodes in iteratorResults
+    uint16_t actualCount = 0;
+    DLLIST_FOREACH(dlnode, &state.iteratorResults) {
+      actualCount++;
+    }
```
Itzikvaknin:
```suggestion
    count = countNodes(&state.iteratorResults);
```
From 10 lines later

tests/cpptests/test_cpp_async_state.cpp
```diff
@@ -0,0 +1,565 @@
...
+
+// Test: State 2 → State 3: Buffered → Pending
+TEST_F(AsyncStateTest, testBufferedToPending) {
+  // Setup: Add 10 results to buffer
+  for (t_docId i = 1; i <= 10; i++) {
+    addToIteratorResults(i);
+  }
+  ASSERT_EQ(state.iteratorResultCount, 10);
+
+  // Move all to pending (simulating refillAsyncPool)
```
Itzikvaknin, line 128: `IndexResultAsyncRead_RefillPool` only makes `SearchDisk_AddAsyncRead` calls, we can test the actual function with mocking `SearchDisk_AddAsyncRead` with noop

tests/cpptests/test_cpp_async_state.cpp
```diff
@@ -0,0 +1,565 @@
...
+  ASSERT_EQ(countNodes(&state.pendingResults), 5); // Still in pending until consumed
+
+  // Verify results are in order
+  for (int i = 0; i < 5; i++) {
+    ASSERT_NE(state.readyResults[i].dmd, nullptr);
+    ASSERT_EQ(state.readyResults[i].dmd->id, i + 1);
+  }
+}
+
+// Test: State 4 → State 5: Ready → Consumed (simulated popReadyResult)
```
Itzikvaknin: Why simulate? it doesn't call any external API or something like that

src/result_processor.c
```diff
@@ -92,92 +101,258 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+      continue;
+    }
+
+    // Deep copy the IndexResult since iterator reuses the same pointer
+    // The copy will be freed after the async read completes and result is consumed
+    RSIndexResult *copy = IndexResult_DeepCopy(it->current);
+
+    // Allocate a new node and add it to the list
+    IndexResultNode *node = rm_calloc(1, sizeof(*node));
+    node->result = copy;
```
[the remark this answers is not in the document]
kei-nan: we never check rm_calloc

src/result_processor.c
```diff
@@ -92,92 +101,258 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+      IndexResult_Free(self->async.lastReturnedIndexResult);
+      self->async.lastReturnedIndexResult = NULL;
     }
 
-validate_current:
-    IndexSpec* spec = self->sctx->spec;
-    if (!getDocumentMetadata(spec, docs, sctx, it, &dmd)) {
-      continue;
+    // Step 1: Refill IndexResult buffer if needed (cheap iterator reads)
+    int refillResult = refillIndexResultBuffer(self);
```
kei-nan: refillFromIterator
@JoanFM what do you propose.

src/result_processor.c
```diff
@@ -92,92 +101,258 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
-    if (self->querySlots && (__atomic_load_n(&key_space_version, __ATOMIC_RELAXED) != self->keySpaceVersion)) {
-      RS_ASSERT(self->querySlots != NULL);
-      int slot = RedisModule_ClusterKeySlotC(dmd->keyPtr, sdslen(dmd->keyPtr));
-      if (!SlotRangeArray_ContainsSlot(self->querySlots, slot)) {
-        DMD_Return(dmd);
-        continue;
-      }
+
+    // Step 3: No ready results - poll for more
+    int timeout_ms = it->atEOF ? 1000 : 0;
```
kei-nan: add a constant for the timeout

tests/cpptests/test_cpp_async_state.cpp
```diff
@@ -0,0 +1,565 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+#include "gtest/gtest.h"
```
kei-nan, line 10: add a follow-up task to create a test to check the RPQueryResultProcessor

tests/cpptests/test_cpp_async_state.cpp
kei-nan: I really want to stop somwhere with this PR.
I don't want to do additional big changes in this PR.
if you want we can do a followup task.

src/index_result_async_read.c
kei-nan: lets not do it in this PR

src/result_processor.c
JoanFM: Something along the line of: Read Iterator Results, something that speaks more about the semantics of the Iterator/Processor that does not expose the reader of the code to the intrinsics of the async pool

src/result_processor.c
```diff
@@ -92,92 +101,258 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
-    if (isTrimming && RedisModule_ShardingGetKeySlot) {
-      RedisModuleString *key = RedisModule_CreateString(NULL, dmd->keyPtr, sdslen(dmd->keyPtr));
-      int slot = RedisModule_ShardingGetKeySlot(key);
-      RedisModule_FreeString(NULL, key);
-      int firstSlot, lastSlot;
-      RedisModule_ShardingGetSlotRange(&firstSlot, &lastSlot);
-      if (firstSlot > slot || lastSlot < slot) {
-        DMD_Return(dmd);
+    // Step 1b: Submit any buffered results to async pool (keep pipeline full)
+    IndexResultAsyncRead_RefillPool(&self->async);
```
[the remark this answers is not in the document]
JoanFM: Also, can we wrap this function in a function that "translates" the concepts of DMD and Iterator DocIDs? so that when someone reads this understands the general idea without being so much exposed to the inside implementation?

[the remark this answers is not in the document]
JoanFM: Something like, SubmitReadDMDFromResults?

src/result_processor.c
```diff
@@ -92,92 +105,257 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+
+    // Step 2: Try to serve a ready result if we have one
+    if (self->async.readyResultsIndex < array_len(self->async.readyResults)) {
+      RSIndexResult *indexResult = IndexResultAsyncRead_PopReadyResult(&self->async);
+      RS_ASSERT(indexResult);  // Should always be valid
```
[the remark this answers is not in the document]
JoanFM: @kei-nan worth taking a look

src/result_processor.c
```diff
@@ -92,92 +105,257 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
-      RedisModule_FreeString(NULL, key);
-      int firstSlot, lastSlot;
-      RedisModule_ShardingGetSlotRange(&firstSlot, &lastSlot);
-      if (firstSlot > slot || lastSlot < slot) {
-        DMD_Return(dmd);
+    // Step 1b: Submit any buffered results to async pool (keep pipeline full)
+    IndexResultAsyncRead_RefillPool(&self->async);
+
+    // Step 2: Try to serve a ready result if we have one
+    if (self->async.readyResultsIndex < array_len(self->async.readyResults)) {
```
JoanFM: let's change this for?

```c
RSIndexResult *indexResult = IndexResultAsyncRead_PopReadyResult(&self->async);
if indexResult {
  

}
````

or have the PopReadyResult fill an indexResult and return a boolean if it exists or not? I think the code would be cleaner.

File: src/result_processor.c [hunk 32]
```diff
@@ -92,92 +105,257 @@ static bool getDocumentMetadata(IndexSpec* spec, DocTable* docs, RedisSearchCtx
...
+    // Step 3: No ready results - poll for more
+    int timeout_ms = it->atEOF ? ASYNC_POLL_TIMEOUT_AT_EOF_MS : 0;
+
+    // Poll writes directly to the arrays (capacity is ASYNC_POOL_SIZE)
+    const size_t pendingCount = SearchDisk_PollAsyncReads(
+        self->async.asyncPool, timeout_ms, 
+        self->async.readyResults, self->async.failedUserData);
+
+    // Reset index to start consuming from the beginning of readyResults
+    self->async.readyResultsIndex = 0;
```
JoanFM: should we wrap it with an API with another naming?

Itzikvaknin (approved):
LGTM ⚡

Itzikvaknin (dismissed):

src/debug_commands.c
```diff
@@ -2157,6 +2157,47 @@ DEBUG_COMMAND(VecSimMockTimeout) {
...
+  if (!debugCommandsEnabled(ctx)) {
+    return RedisModule_ReplyWithError(ctx, NODEBUG_ERR);
+  }
+  if (argc != 3) {
+    return RedisModule_WrongArity(ctx);
+  }
+
+  const char *op = RedisModule_StringPtrLen(argv[2], NULL);
+
+  if (!strcmp("enable", op)) {
```
JoanFM: I think it would be comfortable to have it case insensitive?

src/debug_commands.c
```diff
@@ -2157,6 +2157,47 @@ DEBUG_COMMAND(VecSimMockTimeout) {
...
+    // Check if disk is available first
+    if (!SearchDisk_IsAsyncIOSupported()) {
+      return RedisModule_ReplyWithError(ctx, "Async I/O is not supported (disk API not available or disk doesn't support async I/O)");
+    }
+    SearchDisk_SetAsyncIOEnabled(true);
+    return RedisModule_ReplyWithSimpleString(ctx, "OK - Async I/O enabled");
+  } else if (!strcmp("disable", op)) {
```
JoanFM: same here

src/debug_commands.c
```diff
@@ -2157,6 +2157,47 @@ DEBUG_COMMAND(VecSimMockTimeout) {
...
+    SearchDisk_SetAsyncIOEnabled(false);
+    return RedisModule_ReplyWithSimpleString(ctx, "OK - Async I/O disabled");
+  } else if (!strcmp("status", op)) {
```
JoanFM: and here

tests/cpptests/test_cpp_async_state.cpp
JoanFM: can u point to the follow-up task?

src/search_disk.c
```diff
@@ -120,6 +123,57 @@ size_t SearchDisk_GetDeletedIds(RedisSearchDiskIndexSpec *handle, t_docId *buffe
...
+    return disk->docTable.createAsyncReadPool(handle, max_concurrent);
+}
+
+bool SearchDisk_AddAsyncRead(RedisSearchDiskAsyncReadPool pool, t_docId docId, uint64_t user_data) {
+    RS_ASSERT(disk && pool);
+    return disk->docTable.addAsyncRead(pool, docId, user_data);
+}
+
+// Callback to allocate a new RSDocumentMetadata with ref_count=1 and keyPtr set
+static RSDocumentMetadata* allocateDMD(const void* key_data, size_t key_len) {
```
JoanFM, line 138: should this be moved to the RSDocumentMetadata module?

src/index_result_async_read.h
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * Polls the async pool for completed reads and updates the ready results.
+ * Resets the readyResultsIndex to start consuming from the beginning.
+ * Cleans up any failed reads (not found/error).
+ *
+ * @param state Async read state structure
+ * @param timeout_ms Timeout in milliseconds for the poll operation
+ * @return Number of pending async reads still in progress
+ */
+size_t IndexResultAsyncRead_Poll(IndexResultAsyncReadState *state, int timeout_ms);
```
JoanFM: uint for timeouts?

src/index_result_async_read.c
```diff
@@ -0,0 +1,183 @@
...
+  state->poolSize = poolSize;
+  state->iteratorResultCount = 0;
+  state->readyResults = NULL;
+  state->failedUserData = NULL;
+  state->asyncPool = NULL;
+  state->readyResultsIndex = 0;
+  state->lastReturnedIndexResult = NULL;
+}
+
+void IndexResultAsyncRead_SetupAsyncPool(IndexResultAsyncReadState *state,
```
JoanFM, line 28: Should this be simply merged with the `IndexResultAsyncRead_Init` and have a simple Initialization?

src/index_result_async_read.c
```diff
@@ -0,0 +1,183 @@
...
+  }
+}
+
+void IndexResultAsyncRead_RefillPool(IndexResultAsyncReadState *state) {
+  uint16_t added = 0;
+
+  // Move nodes from iteratorResults to pendingResults
+  while (added < state->poolSize && !DLLIST_IS_EMPTY(&state->iteratorResults)) {
+    // Pop from the head of iteratorResults to maintain FIFO order
+    DLLIST_node *dlnode = dllist_pop_head(&state->iteratorResults);
```
JoanFM: should we pop once we know it has been added? can we here simply point to the head without popping? It may save some readdition?

src/index_result_async_read.c
```diff
@@ -0,0 +1,183 @@
...
+
+    // Try to add to async pool, using the node pointer as user_data
+    if (!SearchDisk_AddAsyncRead(state->asyncPool, docId, (uint64_t)node)) {
+      // Pool is full - put the node back at the head and stop
+      dllist_prepend(&state->iteratorResults, dlnode);
+      break;
+    }
+
+    // Move node to pendingResults list
+    dllist_append(&state->pendingResults, dlnode);
```
JoanFM, line 110: Commenting here, Like here in the same operation, pop from IteratorResults and append to PendingResults? simpler to read?

src/index_result_async_read.c
```diff
@@ -0,0 +1,183 @@
...
+    state->iteratorResultCount--;
+    added++;
+  }
+}
+
+static void IndexResultAsyncRead_CleanupFailedReads(IndexResultAsyncReadState *state) {
+  for (uint16_t i = 0; i < array_len(state->failedUserData); i++) {
```
JoanFM, line 118: why uint16_t? Let's have size_t and thus we avoid the risk of infinite loop.

Itzikvaknin (dismissed):

src/search_disk.c
kei-nan: I think it is internal and very specific to disk,
- a function receiving a key and length and allocates a dmd with key - feels very specific to disk.

Don't see much benefit in moving it there.

src/index_result_async_read.c
kei-nan: The cpp unit test needs a way to initialize the pool without using disk API.
That is why we need the separation.

src/search_disk.c
JoanFM: I think a module should expose their APIs and then the other modules decide wether or not they need to use.

JoanFM: I feel also an agent would try to search there for these?

src/index_result_async_read.c
kei-nan: we limit the pool size to 16 at this time.
that means the error array can't exceed 16 errors at a time.
seems wasteful to use size_t here.
not sure how we can get infinite loop in this case.

JoanFM: yes I know we limit，but we add the risk that eventually introducing a bug we lose the invariant, this ends up in an infinite loop. It is more of a security guardrail

src/search_disk.c
```diff
@@ -123,7 +123,7 @@ size_t SearchDisk_GetDeletedIds(RedisSearchDiskIndexSpec *handle, t_docId *buffe
     return disk->docTable.getDeletedIds(handle, buffer, buffer_size);
 }
 
-RedisSearchDiskAsyncReadPool *SearchDisk_CreateAsyncReadPool(RedisSearchDiskIndexSpec *handle, uint16_t max_concurrent) {
```
JoanFM: why did we remove the pointer?

JoanFM: ok I see

src/index_result_async_read.c
JoanFM: let's at least have an RS_LOG_ASSERT that i is never larger than 16 to catch the invariant

JoanFM: can we add this @kei-nan , it would be helpful to catch potential problems if they happen

PR #8129, opened 2026-01-22
MOD-13735 MOD-13181 Fix FT.PROFILE shard total profile time

ofiryanai:
# Problem

In cluster mode, `FT.PROFILE` reports incorrect "Total profile time" values for each shard. The total time appears smaller than individual Result Processor times within the same shard, which is mathematically impossible since the total should encompass all processor times. This even worse because some non-negligible time is not included in any of the result processors times (e.g shard results serialization in cursor read).

# Root Cause

The issue occurs when internal cursors are used (cluster mode with FT.AGGREGATE). The timing calculation has a mismatch between how total time and individual processor times are tracked:

1. `initClock` is reset on each cursor read at [src/aggregate/aggregate_exec.c:1382](https://github.com/RediSearch/RediSearch/blob/master/src/aggregate/aggregate_exec.c#L1382):
```
rs_wall_clock_init(&req->initClock); // Reset the clock for the current cursor read
```

2. `profileTotalTime` is calculated only from the last cursor read in [src/profile.c:145](https://github.com/RediSearch/RediSearch/blob/master/src/profile.c#L145):
```
req->profileTotalTime += rs_wall_clock_elapsed_ns(&req->initClock);
```
This only captures elapsed time since the last `initClock` reset.

3. **Result Processor times accumulate across ALL cursor reads** in [src/result_processor.c:1144](https://github.com/RediSearch/RediSearch/blob/master/src/result_processor.c#L1144):
```
self->profileTime += rs_wall_clock_elapsed_ns(&start);
```
These times are never reset between cursor reads.

**Result**: When multiple cursor reads occur, RP times accumulate across all reads while total time only measures the final read.

# Fix

Accumulate `profileTotalTime` at the end of each cursor read in `finishSendChunk()`, matching how RP times are accumulated. This ensures total profile time reflects the sum of all cursor read durations, maintaining mathematical consistency with individual processor times.

---

> [!NOTE]
> Fixes shard profiling time under cursored executions by accumulating per-read durations into `req->profileTotalTime`.
> 
> - Update `finishSendChunk()` in `aggregate_exec.c` to add elapsed time for intermediate cursor reads to `profileTotalTime` (final read remains accounted in `Profile_Print`); use the measured `duration` for global stats
> - Add tests in `tests/pytests/test_profile.py` to verify timing consistency: `Total profile time` ≥ sum of Result Processor `Time` across RESP2/RESP3, cluster/standalone, and various `SEARCH`/`AGGREGATE` pipelines; include helper `sum_rp_times()` and `ProfileTotalTimeConsistency()`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 382aa9d55a45e16c50f97b7d0edc946480ca3056. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_profile.py
+++ b/tests/pytests/test_profile.py
@@ -816,25 +816,73 @@ def sum_rp_times(env, shard):
   return total
 
 def ProfileTotalTimeConsistency(env, num_docs):
-  """Tests that Total profile time >= sum of Result Processor times."""
+  """Tests that Total profile time >= sum of Result Processor times.
+
+  Tests multiple commands with various result processors to ensure timing
+  consistency across different query types:
+  - FT.SEARCH with Scorer, Sorter, Loader
+  - FT.AGGREGATE with Loader, Grouper, Sorter, Projector (APPLY), Pager/Limiter
+  """
   conn = getConnectionByEnv(env)
   run_command_on_all_shards(env, config_cmd(), 'SET', '_PRINT_PROFILE_CLOCK', 'true')
 
-  env.expect('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT').ok()
+  # Create index with TEXT and NUMERIC fields for diverse query options
+  env.expect('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT', 'n', 'NUMERIC', 'SORTABLE').ok()
 
   for i in range(num_docs):
-    conn.execute_command('HSET', f'doc{i}', 't', f'hello{i}')
+    conn.execute_command('HSET', f'doc{i}', 't', f'hello{i % 100}', 'n', i)
 
+  def verify_timing_consistency(res, command_desc):
+    """Helper to verify total time >= sum of RP times for all shards."""
+    _, shards = extract_profile_coordinator_and_shards(env, res)
+    for shard in shards:
+      # In RESP2, Total profile time is returned as a string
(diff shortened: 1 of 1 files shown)


src/aggregate/aggregate_exec.c
```diff
@@ -389,6 +389,10 @@ static void finishSendChunk(AREQ *req, SearchResult **results, SearchResult *r,
   if (QueryError_IsOk(qctx->err) || hasTimeoutError(qctx->err)) {
     rs_wall_clock_ns_t duration = rs_wall_clock_elapsed_ns(&req->initClock);
     TotalGlobalStats_CountQuery(AREQ_RequestFlags(req), duration);
+    // Accumulate profile time at end of each cursor read (before cursor pauses)
+    if (IsProfile(req) && !cursor_done) {
+      req->profileTotalTime += duration;
+    }
```
GuyAv46: Why conditioning this with `QueryError_IsOk(qctx->err) || hasTimeoutError(qctx->err)`?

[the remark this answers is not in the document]
GuyAv46: `cursor_done` is `true` for non-cursor queries

ofiryanai: moved it outside 👍

PR #8142, opened 2026-01-23
[MOD-12307] add numDocs counter to TrieNode

JoanFM:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

---

> [!NOTE]
> Adds per-term document counts to the trie and propagates the new data through APIs, iterators, and persistence.
> 
> - Introduces `numDocs` field in `TrieNode`, increments on inserts/splits/merges, and returns it from `TrieIterator_Next`
> - Extends trie APIs (`Trie_Insert*`, `TrieNode_Add`, `TrieIterator_Next`) with `numDocs` param/out; updates all call sites (query, GC, suffix/suggest/spellcheck, debug, dictionary)
> - Bumps trie RDB encoding to `TRIE_ENCVER_CURRENT=2`; `TrieType_GenericSave/Load` now write/read `numDocs` (flags for payloads/numDocs)
> - Adjusts tests to new signatures/format and memory expectations (extra 8 bytes per entry)
> - Minor API signature cleanups (e.g., `TrieType_GenericLoad/Save` use bools)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a9d8de0c12e20bc288434f5d949f60bcde3ca217. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/trie/trie_type.c
+++ b/src/trie/trie_type.c
@@ -369,7 +369,6 @@ void TrieType_GenericSave(RedisModuleIO *rdb, Trie *tree, bool savePayloads, boo
           RedisModule_SaveStringBuffer(rdb, "", 1);
         }
       }
-      // TODO: Save a marker for empty payload!
       if (saveNumDocs) {
         RedisModule_SaveUnsigned(rdb, numDocs);
       }


src/search_disk_api.h
```diff
@@ -89,6 +89,14 @@ typedef struct IndexDiskAPI {
...
    */
   QueryIterator* (*newWildcardIterator)(RedisSearchDiskIndexSpec *index, double weight);
+
+  /**
+   * @brief Returns the number of documents matching the term
+   *
+   * @param index Pointer to the index
+   * @return Number of documents matching the term
+   */
+  uint64_t (*getNumDocsMatchingTerm)(RedisSearchDiskIndexSpec *index);
```
JoanFM: remove this

JoanFM:
what I could do is to add a compilation flag to enable this extra space addition to the node.

JoanFM:
> what I could do is to add a compilation flag to enable this extra space addition to the node.

Thinking about this, this would save only like 8MB on a huge index, so is not worth the problems, will revert

src/trie/trie_type.h
```diff
@@ -45,11 +46,12 @@ typedef struct {
  * score using `Trie_Sort_Score.                            */
 Trie *NewTrie(TrieFreeCallback freecb, TrieSortMode sortMode);
 
-int Trie_Insert(Trie *t, RedisModuleString *s, double score, int incr, RSPayload *payload);
+int Trie_Insert(Trie *t, RedisModuleString *s, double score, int incr, RSPayload *payload,
```
kei-nan, line 49: consider having an enum to define if to set or add the incoming doc count instead of getting two counts.

src/trie/trie_type.c
```diff
@@ -26,38 +26,40 @@
...
   return rc;
 }
 
 int Trie_InsertRune(Trie *t, const rune *runes, size_t len, double score, int incr,
-                    RSPayload *payload) {
+                    RSPayload *payload, size_t numDocsToSet, size_t numDocsToAdd) {
   int rc = 0;
   if (runes && len && len < TRIE_INITIAL_STRING_LEN) {
-    rc = TrieNode_Add(&t->root, runes, len, payload, (float)score, incr ? ADD_INCR : ADD_REPLACE, t->freecb);
+    rc = TrieNode_Add(&t->root, runes, len, payload, (float)score, incr ? ADD_INCR : ADD_REPLACE,
```
kei-nan, line 61: do you know what incr means in this context?
I'm wondering if it is possible for the trie node to manage the doc count internally

JoanFM: seems related to the score

src/trie/trie_type.h
JoanFM: Will do, good idea

src/trie/trie_type.c
JoanFM: Some internal score in the TrieNode.

src/trie/trie.c
```diff
@@ -125,8 +127,9 @@ TrieNode *__trie_AddChildIdx(TrieNode *n, const rune *str, t_len offset, t_len l
 
 TrieNode *__trie_SplitNode(TrieNode *n, t_len offset) {
   // Copy the current node's data and children to a new child node
+  size_t numDocsForChild = n->numDocs;
   TrieNode *newChild = __newTrieNode(n->str, offset, n->len, NULL, 0, n->numChildren, n->score,
-                                     __trieNode_isTerminal(n), n->sortMode);
+                                     __trieNode_isTerminal(n), n->sortMode, numDocsForChild);
```
raz-mon: Why not this?
```suggestion
TrieNode *newChild = __newTrieNode(n->str, offset, n->len, NULL, 0, n->numChildren, n->score,
                                     __trieNode_isTerminal(n), n->sortMode, n->numDocs;
```

src/trie/trie.c
```diff
@@ -164,9 +168,10 @@ TrieNode *__trieNode_MergeWithSingleChild(TrieNode *n, TrieFreeCallback freecb)
   rune nstr[n->len + ch->len + 1];
   memcpy(nstr, n->str, sizeof(rune) * n->len);
   memcpy(&nstr[n->len], ch->str, sizeof(rune) * ch->len);
+  size_t numDocsForMerged = ch->numDocs;
   TrieNode *merged = __newTrieNode(
       nstr, 0, n->len + ch->len, NULL, 0, ch->numChildren,
-      ch->score, __trieNode_isTerminal(ch), n->sortMode);
+      ch->score, __trieNode_isTerminal(ch), n->sortMode, numDocsForMerged);
```
raz-mon: Same as above, why do we need the var?

src/trie/trie.c
```diff
@@ -214,6 +222,7 @@ int TrieNode_Add(TrieNode **np, const rune *str, t_len len, RSPayload *payload,
     if (offset == len) {
       n->score = score;
       n->flags |= TRIENODE_TERMINAL;
+      n->numDocs = numDocs;
```
raz-mon: Why do we always set? Shouldn't we add in case `numDocsToSet=0`?

src/trie/trie.c
JoanFM: Agent iterations

JoanFM:
@raz-mon it was agent changing (I added at some point a compile command and when removed it left some of the vars)

src/trie/trie.c
```diff
@@ -136,10 +138,11 @@ TrieNode *__trie_SplitNode(TrieNode *n, t_len offset) {
   memcpy(newChildren, children, sizeof(TrieNode *) * n->numChildren);
   memcpy(__trieNode_childKey(newChild, 0), __trieNode_childKey(n, 0), n->numChildren * sizeof(rune));
 
-  // reduce the node to be just one child long with no score
+  // reduce the node to be just one child long with no score and no numDocs
```
raz-mon:
```suggestion
  // reduce the node to be just one child with no score and no documents
```

src/trie/trie_type.h
```diff
@@ -20,7 +20,8 @@ extern "C" {
 
 extern RedisModuleType *TrieType;
 
-#define TRIE_ENCVER_CURRENT 1
+#define TRIE_ENCVER_CURRENT 2
+#define TRIE_ENCVER_NONUMDOCS 1
```
raz-mon: I don't like this naming.. can we try to find a positive naming for the version?
I know that last used it but would prefer a positive distinction

tests/ctests/test_trie.c
```diff
@@ -332,11 +333,278 @@ int testDFAFilter() {
...
+
+  // Insert "help"
+  int rc = Trie_InsertStringBuffer(t, "help", 4, 1.0, 0, NULL, 1, ADD_INCR);
+  ASSERT_EQUAL(1, rc);
+  node = TrieNode_Get(t->root, helpRunes, helpLen, true, NULL);
+  ASSERT(node != NULL);
+  ASSERT_EQUAL(1, node->numDocs);
+
+  // Insert "helping" - "help" is a prefix of "helping"
+  rc = Trie_InsertStringBuffer(t, "helping", 7, 2.0, 0, NULL, 1, ADD_INCR);
```
raz-mon: Any reason for the different scores?
If not let's have them all `1`, as in production

tests/ctests/test_trie.c
```diff
@@ -332,11 +333,278 @@ int testDFAFilter() {
...
+  ASSERT_EQUAL(3, node->numDocs);
+
+  // Increment numDocs for "AB" (middle of chain)
+  rc = Trie_InsertStringBuffer(t, "AB", 2, 5.0, 0, NULL, 1, ADD_INCR);
+  ASSERT_EQUAL(0, rc);
+  node = TrieNode_Get(t->root, abRunes, abLen, true, NULL);
+  ASSERT(node != NULL);
+  ASSERT_EQUAL(2, node->numDocs);
+
+  // Final verification: check all values
```
raz-mon, line 406: Can we check this as part of the iteration below? We know the order of iteration so I think that will be better

src/trie/trie_type.h
JoanFM: TRIE_ENCVER_BEFORE_NUMDOCS?

tests/ctests/test_trie.c
JoanFM: How do I know exactly the order of keys iterated over? Is this order of iteration an implementation detail? Are we sure this cannot change?

I think is better to have these double check

src/trie/trie_type.h
raz-mon: That's negative too..
I guess we can start now as well, e.g., add `#define TRIE_ENCVER_NUMDOCS 2` as well, such that the next one can just positively be defined as well

tests/cpptests/test_cpp_llapi.cpp
```diff
@@ -30,8 +30,9 @@
 #define TAG_FIELD_NAME1 "tag1"
 #define TAG_FIELD_NAME2 "tag2"
 #define INITIAL_DOC_TABLE_SIZE 1000
-// 3 `uintptr_t` fields
+// 3 `uintptr_t` fields (Rust TrieMap struct: root, n_unique_keys, memory_usage)
```
raz-mon: Consider removing this comment

src/suffix.c
```diff
@@ -61,7 +61,7 @@ void addSuffixTrie(Trie *trie, const char *str, uint32_t len) {
   if (!data) {
     suffixData newdata = createSuffixNode(copyStr, 1);
     RSPayload payload = { .data = (char*)&newdata, .len = sizeof(newdata) };
-    TrieNode_Add(&trie->root, runes, rlen, &payload, 1, ADD_REPLACE, trie->freecb);
+    TrieNode_Add(&trie->root, runes, rlen, &payload, 1, ADD_REPLACE, trie->freecb, 0, ADD_REPLACE);
```
raz-mon: Is the new enum always matching the existing one?
If so, we don't need it no?

JoanFM: is not always matching, or i did not even analyze. But I would not link the numDocs with the Score as they are conceptually different.

raz-mon: Seems weird to me that we have both, as it is probably a common configuration related to the context of the whole `Add` operation.
Can you point at a place that it's different? If not, the above would be true

src/trie/trie_type.c
```diff
@@ -295,10 +297,10 @@ void *TrieType_RdbLoad(RedisModuleIO *rdb, int encver) {
   if (encver > TRIE_ENCVER_CURRENT) {
     return NULL;
   }
-  return TrieType_GenericLoad(rdb, encver > TRIE_ENCVER_NOPAYLOADS);
+  return TrieType_GenericLoad(rdb, encver >= TRIE_ENCVER_PAYLOADS, encver >= TRIE_ENCVER_NUMDOCS);
```
raz-mon, line 300: Nitpick: I would probably just send `encver` and handle it in the GenericLoad function.

src/trie/trie_type.c
```diff
@@ -363,6 +370,9 @@ void TrieType_GenericSave(RedisModuleIO *rdb, Trie *tree, int savePayloads) {
         }
       }
       // TODO: Save a marker for empty payload!
```
raz-mon: Is the comment above relevant?
Should it come later if at all?

JoanFM: I kept it where it was relative to the payload serialization, but can remove

src/trie/trie_type.c
JoanFM: Would keep it as this now

src/trie/trie_type.c
JoanFM: will remove

PR #8144, opened 2026-01-23
refactor: use stats NumDocuments for IDF computation instead of docs.size

JoanFM:
It is weird that IDF and BM25_IDF are not computed from the same source as totalDocs

---

> [!NOTE]
> **Scoring consistency and IDF tweak**
> 
> - Compute `term->idf` using `sctx->spec->stats.scoring.numDocuments` in term and tag iterators for consistency with BM25 inputs
> - Adjust `CalculateIDF` to use `logb(1 + (totalDocs + 1) / max(termDocs,1))` and document the reasoning to avoid `logb` threshold edge cases
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b17c0014054afeecb84756cd035bda8cdd3ac768. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/iterators/inverted_index_iterator.c
+++ b/src/iterators/inverted_index_iterator.c
@@ -460,7 +460,6 @@ QueryIterator *NewInvIndIterator_TermQuery(const InvertedIndex *idx, const Redis
     .predicate = FIELD_EXPIRATION_PREDICATE_DEFAULT,
   };
   if (term && sctx) {
-    // compute IDF based on num of docs in the header
     term->idf = CalculateIDF(sctx->spec->stats.scoring.numDocuments, InvertedIndex_NumDocs(idx));
     term->bm25_idf = CalculateIDF_BM25(sctx->spec->stats.scoring.numDocuments, InvertedIndex_NumDocs(idx));
   }
@@ -487,7 +486,7 @@ QueryIterator *NewInvIndIterator_TagQuery(const InvertedIndex *idx, const TagInd
     .predicate = FIELD_EXPIRATION_PREDICATE_DEFAULT,
   };
   if (term && sctx) {
-    term->idf = CalculateIDF(sctx->spec->stats.scoring.numDocuments, InvertedIndex_NumDocs(idx)); // +1 because logb is used, and logb(1.99) = 0 and logb(2.00) = 1)
+    term->idf = CalculateIDF(sctx->spec->stats.scoring.numDocuments, InvertedIndex_NumDocs(idx));
     term->bm25_idf = CalculateIDF_BM25(sctx->spec->stats.scoring.numDocuments, InvertedIndex_NumDocs(idx));
   }
 


src/iterators/inverted_index_iterator.c
```diff
@@ -460,7 +460,7 @@ QueryIterator *NewInvIndIterator_TermQuery(const InvertedIndex *idx, const Redis
   };
   if (term && sctx) {
     // compute IDF based on num of docs in the header
-    term->idf = CalculateIDF(sctx->spec->docs.size, InvertedIndex_NumDocs(idx)); // FIXME: docs.size starts at 1???
+    term->idf = CalculateIDF(sctx->spec->stats.numDocuments, InvertedIndex_NumDocs(idx));
```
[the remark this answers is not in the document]
JoanFM: but if totalDocs is 0 there should be no result

src/iterators/inverted_index_iterator.c
```diff
@@ -460,7 +460,7 @@ QueryIterator *NewInvIndIterator_TermQuery(const InvertedIndex *idx, const Redis
   };
   if (term && sctx) {
     // compute IDF based on num of docs in the header
-    term->idf = CalculateIDF(sctx->spec->docs.size, InvertedIndex_NumDocs(idx)); // FIXME: docs.size starts at 1???
+    term->idf = CalculateIDF(sctx->spec->stats.numDocuments + 1, InvertedIndex_NumDocs(idx)); // +1 because logb is used, and logb(1.99) = 0 and logb(2.00) = 1)
```
GuyAv46: I don’t understand the explanation.
Could we use `log` instead?

JoanFM: We are getting IDF 0 for some cases which is weird.

I would not just start using log (most lileky we should but it would be a breaking change)

My main goal here is tonbe able to use numDocuments and not docs.size for every case so it would also work on Disk

GuyAv46: I prefer that the disk implementation not start with a bug. Why is it a breaking change to use `log`? The scores may change, but the relevance of documents will stay the same

GuyAv46: CC @oshadmi

JoanFM: I am not sure is a bug, I am quite sure this must be a choice for numerical stability or.something. And if we change this we should go on a benchmarking journey that not sure is a priority right now

oshadmi: What about the `totalDocs = MAX(totalDocs, termDocs)` as in `CalculateIDF_BM25`?
Would be better than a `+1` that would only change the result when `termDocs == numDocuments + 1` ?

JoanFM: the goal of my refactor is about having a single source of numDocuments, do not want to change the computation deeply at this moment as can be more fragile.

oshadmi: If numDocuments is more up-to-date, it could make zero score more common upon deletions, without the `MAX` statement, no?
And regardless, can move the `+1` added to the caller to the impl itself, if you decide to keep it instead of the `MAX` statement.

JoanFM: I prefer to keep the history of the changes as changing the usage of docs.size rather than moving to implementation

src/iterators/inverted_index_iterator.c
```diff
@@ -460,7 +460,7 @@ QueryIterator *NewInvIndIterator_TermQuery(const InvertedIndex *idx, const Redis
   };
   if (term && sctx) {
     // compute IDF based on num of docs in the header
-    term->idf = CalculateIDF(sctx->spec->stats.numDocuments + 1, InvertedIndex_NumDocs(idx)); // +1 because logb is used, and logb(1.99) = 0 and logb(2.00) = 1)
+    term->idf = CalculateIDF(sctx->spec->stats.numDocuments, InvertedIndex_NumDocs(idx)); // +1 because logb is used, and logb(1.99) = 0 and logb(2.00) = 1)
```
GuyAv46: Let's move the comment as well

src/iterators/inverted_index_iterator.c
```diff
@@ -437,7 +437,7 @@ QueryIterator *NewInvIndIterator_NumericQuery(const InvertedIndex *idx, const Re
 }
 
 static inline double CalculateIDF(size_t totalDocs, size_t termDocs) {
-  return logb(1.0F + totalDocs / (double)(termDocs ?: 1));
+  return logb(1.0F + (totalDocs + 1) / (double)(termDocs ?: 1));
```
GuyAv46, line 441: Should we do it for the `termDocs` as well? Or use the `?:` operand instead of `+1`?

JoanFM: lets keep it as it is now

src/iterators/inverted_index_iterator.c
```diff
@@ -486,8 +487,7 @@ QueryIterator *NewInvIndIterator_TagQuery(const InvertedIndex *idx, const TagInd
     .predicate = FIELD_EXPIRATION_PREDICATE_DEFAULT,
   };
   if (term && sctx) {
-    // compute IDF based on num of docs in the header
-    term->idf = CalculateIDF(sctx->spec->docs.size, InvertedIndex_NumDocs(idx)); // FIXME: docs.size starts at 1???
+    term->idf = CalculateIDF(sctx->spec->stats.scoring.numDocuments, InvertedIndex_NumDocs(idx)); // +1 because logb is used, and logb(1.99) = 0 and logb(2.00) = 1)
```
oshadmi: Comment moved inside callee - can remove from here

src/iterators/inverted_index_iterator.c
GuyAv46: OK. 
I think the `CalculateIDF_BM25` explains the edge case:
>  // totalDocs should never be less than termDocs, as that causes an underflow
  // wraparound in the below calculation.
  // Yet, that can happen in some scenarios of deletions/updates, until fixed in
  // the next GC run.
  // In that case, we set totalDocs to termDocs, as a temporary fix.
  totalDocs = MAX(totalDocs, termDocs);

src/iterators/inverted_index_iterator.c
GuyAv46:
```suggestion
    term->idf = CalculateIDF(sctx->spec->stats.scoring.numDocuments, InvertedIndex_NumDocs(idx));
```

PR #8184, opened 2026-01-27
[MOD-13587] RSValue: Reimplement array and map types

daxhuiberts:
This replaces the existing RsValue collection implementation with regular boxed slices.

Uninitialized data is only present during construction before the RsValue is ever created, so no need to pollute the RsValue with `MaybeUninit` logic. That can live entirely at the FFI boundary and only safe code can be used in RsValue for the array an map types.

I also renamed the C interface functions to reflect their purpose more correctly.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches core `RSValue` container types and their C/Rust FFI APIs (arrays/maps and refcounting), so any mismatch in ownership, initialization, or ABI expectations could cause memory safety issues or subtle runtime bugs.
> 
> **Overview**
> **Release note:** Refactors `RSValue` arrays and maps to use a new *builder-based* construction API and a simpler boxed-slice backing, reducing reliance on uninitialized collection internals.
> 
> C call sites that produce arrays/maps now allocate via `RSValue_NewArrayBuilder`/`RSValue_NewMapBuilder`, populate entries, and finalize with `RSValue_NewArrayFromBuilder`/`RSValue_NewMapFromBuilder`; the Rust FFI layer is reorganized accordingly (new `array.rs`/`map.rs`, updated generated header, removed old `collection` FFI).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7499715ac98afc5e8b352903fda29d168c41a0a0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/value/value.h
+++ b/src/value/value.h
@@ -340,7 +340,7 @@ void RSValue_Map_GetEntry(const RSValue *map, uint32_t i, RSValue **key, RSValue
  */
 void RSValue_MapBuilderSetEntry(RSValueMapBuilder *map, size_t i, RSValue *key, RSValue *value);
 
-// Trio gettersRSValueMap
+// Trio getters
 /**
  * Get the left value of a Trio value.
  * The passed RSValue must be of type RSValueType_Trio.


src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
```diff
@@ -0,0 +1,96 @@
...
+/// # SAFETY
+///
+/// 1. `value` must point to a valid [`RsValue`] obtained from an `RSValue_*` function.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_ArrayLen(value: *const RsValue) -> u32 {
+    // Safety: ensured by caller (1.)
+    let value = unsafe { expect_value(value) };
+
+    if let RsValue::Array(array) = value {
+        array.len() as u32
```
daxhuiberts: The C code works with `u32` for indexes and length. Casting from a `u32` to a `usize` is fine, but from `usize` to `u32` might get truncated. All sizes used from the C code always come from a `u32`, but potential future RsValue::Array types created from Rust might theoretically get sizes larger than `u32::MAX`.

What to do with this? Hope that this rare case never happens? Do we want to have a wrapper around the `Vec<SharedRsValue>` which ensures `len` is at most `u32::Max`? Is there a `Vec32Bits` we should use?

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
```diff
@@ -0,0 +1,153 @@
...
+/// # Panics
+///
+/// Panics if `map` is not a map value.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_Map_Len(map: *const RsValue) -> u32 {
+    // Safety: ensured by caller (1.)
+    let map = unsafe { expect_value(map) };
+
+    if let RsValue::Map(map) = map {
+        map.len() as u32
```
daxhuiberts: Same story here regarding `usize` and `u32`.

src/redisearch_rs/value/src/lib.rs
```diff
@@ -49,13 +47,13 @@ pub enum RsValue {
     /// String value
     String(Box<RsValueString>),
     /// Array value
-    Array(RsValueArray),
+    Array(Vec<SharedRsValue>),
```
daxhuiberts: We can also store it as `Box<[SharedRsValue]>` which will not have the capacity. After construction the size never changes. What is preferred?

src/redisearch_rs/value/src/lib.rs
```diff
@@ -49,13 +47,13 @@ pub enum RsValue {
...
     /// Reference value
     Ref(SharedRsValue),
     /// Trio value
     Trio(RsValueTrio),
     /// Map value
-    Map(RsValueMap),
+    Map(Vec<(SharedRsValue, SharedRsValue)>),
```
daxhuiberts: Same here.

src/redisearch_rs/value/src/lib.rs
LukeMathWalker: If the size is never allowed to change, a boxed slice is preferable.

src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
LukeMathWalker: If we're getting rid of size and capacity by using a boxed slice, I see no need to constrain the collection's size to fit within a u32.

daxhuiberts: Rust code will still be able to create a `RsValue::Array(vec![u32::MAX as usize + 1; ...].to_boxed_slice())` which can end up in a call from C to this function.

LukeMathWalker: As in, should this invariant be managed by RSValue or should it be a caller responsibility if they want to be able to save space by using a u32 for the length wherever that might be stored?

daxhuiberts: It's not so much the space savings, it's that the old interface worked with `u32` for sizes. I want to keep the C side intact without making any changes there. That means that an `RsValue::Array` should not have more than `u32::MAX` elements.

Currently you can construct an `RsValue::Array` enum variant by providing a `Vec<SharedRsValue>` (or `Box<[SharedRsValue]>` in a bit), not adhering to the `u32::MAX` limit.

One thing I can do is create a `Array` wrapper, which just receives and stores the inner vec/box slice but checks for this constraint on construction.

daxhuiberts: @LukeMathWalker Solved it like this: https://github.com/RediSearch/RediSearch/pull/8184/changes/5eae26dc61186d405644e2f2f2518da1706c1c40 Let me know what you think.

LukeMathWalker: Fine for me, if we want to minimize disruption for consumers.

src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
```diff
@@ -0,0 +1,148 @@
...
+    let value = unsafe { expect_value(value) };
+
+    if let RsValue::Array(array) = value {
+        // Compatibility: C does an RS_ASSERT on index out of bounds
+        let shared = &array[index as usize];
+        shared.as_ptr() as *mut _
+    } else {
+        // Compatibility: C does an RS_ASSERT on incorrect type
+        panic!("Expected an array value")
+    }
```
LukeMathWalker, line 98: Let's capture the kind of RsValue we got in the panic message.

LukeMathWalker: E.g. "expected an array value, got a map".

src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
```diff
@@ -0,0 +1,148 @@
...
+/// Allocates an array of null pointers with space for `len` [`RsValue`] pointers.
+///
+/// The returned buffer must be populated and then passed to [`RSValue_NewArray`]
+/// to produce an array value.
+///
+/// # SAFETY
+///
+/// 1. The caller must eventually pass the returned pointer to [`RSValue_NewArray`].
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_AllocateArray(len: u32) -> *mut *mut RsValue {
```
LukeMathWalker: Looking at the function name, without reading the docs, I'd be led to believe that this function returns an `RsValue` of kind array.
Perhaps `RSValue_NewArrayBuilder`, or something along those lines, would convey its semantics in a clearer way? 
Similarly, below, `RSValue_NewArrayFromBuilder`.

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
```diff
@@ -0,0 +1,196 @@
...
+
+use crate::util::expect_value;
+use libc::size_t;
+use std::mem::MaybeUninit;
+use value::{Map, RsValue, SharedRsValue};
+
+/// Opaque map structure used during map construction.
+/// Holds uninitialized entries that are populated via `RSValueMap_SetEntry`
+/// before being finalized into an `RsValue::Map` via `RSValue_NewMap`.
+pub struct RSValueMap {
```
LukeMathWalker: Just like its array equivalent, I would call this `RsValueMapBuilder` to clearly communicate that it's not an `RsValue` _yet_.

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
```diff
@@ -0,0 +1,196 @@
...
+*/
+
+use crate::util::expect_value;
+use libc::size_t;
+use std::mem::MaybeUninit;
+use value::{Map, RsValue, SharedRsValue};
+
+/// Opaque map structure used during map construction.
+/// Holds uninitialized entries that are populated via `RSValueMap_SetEntry`
+/// before being finalized into an `RsValue::Map` via `RSValue_NewMap`.
```
LukeMathWalker: Let's use intra-doc links where applicable.

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
```diff
@@ -0,0 +1,196 @@
...
+pub struct RSValueMap {
+    entries: Vec<MaybeUninit<(*mut RsValue, *mut RsValue)>>,
```
LukeMathWalker: Is there a reason to keep this as a Vec rather than a boxed slice, as we did in the array case?

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    if let RsValue::Map(map) = map {
+        // Compatibility: C does an RS_ASSERT on index out of bounds
+        let (shared_key, shared_value) = &map[index as usize];
+        // Safety: ensured by caller (3.)
+        unsafe { key.write(shared_key.as_ptr() as *mut _) };
+        // Safety: ensured by caller (3.)
+        unsafe { value.write(shared_value.as_ptr() as *mut _) };
+    } else {
+        panic!("Expected a map value")
```
LukeMathWalker: Same as the array case, let's capture the unexpected kind in the panic message.

src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
daxhuiberts: I agree it's not clear. I kept it identical to the existing C name as to not needing to rename it. Still rename it? If so, I will also directly change the `value.c` name.

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
daxhuiberts: Nope, will change to a boxed slice.

src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
LukeMathWalker: Yeah, I would rename both.

daxhuiberts: One thing is that `RSValue_AllocateArray` return a `RSValue **arr`, so not really a builder (like the map one). I can also add a `RSValue_ArrayBuilderSetEntry`, like there already is a `RSValueMap_SetEntry`. That incurs a lot more FFI calls though.

LukeMathWalker: I don't think we need the abstraction to clarify the intention via naming. Let's avoid the extra FFI calls.

LukeMathWalker (dismissed):

daxhuiberts:
Split into 3 separate PRs:
https://github.com/RediSearch/RediSearch/pull/8295
https://github.com/RediSearch/RediSearch/pull/8296
https://github.com/RediSearch/RediSearch/pull/8297

PR #8204, opened 2026-01-28
[MOD-13777] RSValue: Implement basic constructors, getters and setters

daxhuiberts:
This adds basic constructors, getters and setters for undefined, null, number and trio types.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Introduces new C FFI APIs and adjusts ownership/refcount handling around `SharedRsValue`/`Arc`, which can break existing consumers or cause misuse if callers don’t follow the new safety contracts.
> 
> **Overview**
> Adds a new `RSValue_` C FFI surface for **owned** value creation and mutation: constructors for `Undefined`, `Null`, `Number`, and `Trio`, plus getters for `Number` and trio element access and setters to mutate an owned value to `Number`/`Null` in place.
> 
> Refactors internals to support this: adds `SharedRsValue::as_ptr` and `SharedRsValue::set_value` (enforcing unique ownership via `Arc::get_mut`), adds trio accessors (`left`/`middle`/`right`), introduces `expect_shared_value` to borrow a `SharedRsValue` without refcount changes, and removes older `SharedRsValue_*` number/trio helper APIs from both Rust and the generated `value.h`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3ef3d272be6265b362227cedb77150ca868255a9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
+++ b/src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
@@ -12,12 +12,18 @@ use value::trio::RsValueTrio;
 use value::{RsValue, shared::SharedRsValue};
 
 /// Creates and returns a new **owned** [`RsValue`] object of type undefined.
+///
+/// The caller must make sure to pass the returned [`RsValue`] to one of the
+/// ownership taking `RSValue_` methods, directly or indirectly.
 #[unsafe(no_mangle)]
 pub extern "C" fn RSValue_NewUndefined() -> *mut RsValue {
     SharedRsValue::new(RsValue::Undefined).into_raw() as *mut _
 }
 
 /// Creates and returns a new **owned** [`RsValue`] object of type null.
+///
+/// The caller must make sure to pass the returned [`RsValue`] to one of the
+/// ownership taking `RSValue_` methods, directly or indirectly.
 #[unsafe(no_mangle)]
 pub extern "C" fn RSValue_NewNull() -> *mut RsValue {
     SharedRsValue::new(RsValue::Null).into_raw() as *mut _
@@ -25,6 +31,9 @@ pub extern "C" fn RSValue_NewNull() -> *mut RsValue {
 
 /// Creates and returns a new **owned** [`RsValue`] object of type number
 /// containing the given numeric value.
+///
+/// The caller must make sure to pass the returned [`RsValue`] to one of the
+/// ownership taking `RSValue_` methods, directly or indirectly.
 #[unsafe(no_mangle)]
 pub extern "C" fn RSValue_NewNumber(value: c_double) -> *mut RsValue {
     SharedRsValue::new(RsValue::Number(value)).into_raw() as *mut _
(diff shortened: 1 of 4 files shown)


daxhuiberts:
Also please take a good look at the (safety) docs and let me know if the approach I've taken is good or not. I want to set a good baseline for the the remaining pull requests with these.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
```diff
@@ -0,0 +1,60 @@
...
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use std::ffi::c_double;
+use value::trio::RsValueTrio;
+use value::{RsValue, shared::SharedRsValue};
+
+/// Creates and returns a new **owned** [`RsValue`] object of type undefined.
```
LukeMathWalker, line 14: Generally speaking, whenever we have a constructor, we should also document whether the returned pointer needs to be freed and which function should be used to free it.

src/redisearch_rs/c_entrypoint/value_ffi/src/getters.rs
```diff
@@ -0,0 +1,96 @@
...
+    let value = unsafe { expect_value(value) };
+
+    if let RsValue::Number(number) = value {
+        *number
+    } else {
+        panic!("not a number")
+    }
+}
+
+/// Gets the left value of a trio [`RsValue`].
```
LukeMathWalker: When we are returning pointers, we should be explicit about the ownership of what we are returning and if the caller is allowed to mutate the value through the pointer.

src/redisearch_rs/value/src/shared.rs
```diff
@@ -46,6 +51,22 @@ impl SharedRsValue {
...
+    /// Set a new [`RsValue`] for this [`SharedRsValue`].
+    ///
+    /// # Safety
+    ///
+    /// Only exactly one reference to the underlying [`RsValue`] must exist.
+    ///
+    /// # Panic
+    ///
+    /// Panics if more than one reference to the underlying [`RsValue`] exists.
+    pub unsafe fn set_value(&mut self, new_value: RsValue) {
```
LukeMathWalker: This function shouldn't be `unsafe`, since it will panic if there is more than one reference. There is no room for undefined behaviour.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
daxhuiberts: Yeah, in the case of `RsValue`, there are functions creating/providing an owned `RsValue` and functions taking an owned `RsValue`. There is no single `free` function. Any owning taking function should be fine as long as it ends up in any **one** of the owning taking functions.

src/redisearch_rs/value/src/shared.rs
daxhuiberts: Not related to this function, but to `unsafe`: In the FFI layer I've marked all `RSValue_` functions unsafe. Even though some of them aren't really. They still need to be handled correctly, such as that the returned value should be used a certain way (e.g. given to another function at a later moment). Does that still warrant unsafe and safety docks?

LukeMathWalker: I don't think they should be marked `unsafe` in so far as they don't perform any `unsafe` operation (e.g. dereferencing raw pointers).

PR #8216, opened 2026-01-29
MOD-13792 refactor: remove RSValueTrait

JonasKruckenberg:

Fully (and finally) removes the `RSValue` trait because the added complexity didn't pull its weight anymore. Brings the RLookup code more in line with the upcoming RSValue migration as well.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches core value ownership/refcounting paths across Rust↔C FFI boundaries, so subtle ABI or lifetime mistakes could cause leaks or crashes. Changes are mostly mechanical refactors with broad test updates, but the surface area is large.
> 
> **Overview**
> This PR **fully removes the `RSValueTrait` abstraction** (and the `test_utils` mock type/feature) and refactors the codebase to use the concrete `value::RSValueFFI` wrapper directly.
> 
> As part of that, `sorting_vector::RSSortingVector` and `rlookup::RLookupRow` are **de-genericized** (now always store `RSValueFFI`), their public APIs and FFI signatures are updated accordingly (including `RLookupRow_GetSortingVector`/`SetSortingVector` and `RLookup_LoadRuleFields` row handling), and tests are rewritten to use real `RSValueFFI` semantics (including refcount checks via `refcount()` and pointer equality).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 419dfff34757a0f6a216cd44aa8ad28dd7bb5d3b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/sorting_vector/tests/sorting_vector.rs
+++ b/src/redisearch_rs/sorting_vector/tests/sorting_vector.rs
@@ -67,7 +67,7 @@ fn out_of_bounds() -> Result<(), IndexOutOfBounds> {
 fn override_value() -> Result<(), IndexOutOfBounds> {
     let src = build_vector()?;
     let mut dst: RSSortingVector = RSSortingVector::new(1);
-    assert!(RSValueFFI::ptr_eq(&dst[0], &RSValueFFI::new_null()));
+    assert!(RSValueFFI::ptr_eq(&dst[0], &RSValueFFI::null_static()));
 
     for (idx, val) in src.iter().enumerate() {
         dst.try_insert_val(0, val.clone())?;
@@ -80,7 +80,7 @@ fn override_value() -> Result<(), IndexOutOfBounds> {
         *val = RSValueFFI::new_num(42.0)
     }
 
-    assert!(RSValueFFI::ptr_eq(&dst[0], &RSValueFFI::new_null()));
+    assert!(RSValueFFI::ptr_eq(&dst[0], &RSValueFFI::null_static()));
     Ok(())
 }
 
--- a/src/redisearch_rs/value/src/rs_value_ffi.rs
+++ b/src/redisearch_rs/value/src/rs_value_ffi.rs
@@ -59,7 +59,7 @@ impl RSValueFFI {
         Self(ptr)
     }
 
-    pub const fn as_raw(&self) -> *mut ffi::RSValue {
+    pub const fn as_ptr(&self) -> *mut ffi::RSValue {
         self.0.as_ptr()
     }
 
@@ -67,7 +67,7 @@ impl RSValueFFI {
         this.0 == other.0
     }
 
-    pub fn new_null() -> Self {
+    pub fn null_static() -> Self {
         // Safety: RSValue_NullStatic returns an immutable global ptr
         let val = unsafe { ffi::RSValue_NullStatic() };
         RSValueFFI(NonNull::new(val).expect("RSValue_NullStatic returned a null pointer"))
--- a/src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
+++ b/src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
@@ -326,7 +326,7 @@ pub unsafe extern "C" fn RLookupRow_Get(
     // Safety: ensured by caller (2.)
     let row = unsafe { RLookupRow::from_opaque_ptr(row).unwrap() };
 
-    row.get(key).map(|x| NonNull::new(x.as_raw()).unwrap())
+    row.get(key).map(|x| NonNull::new(x.as_ptr()).unwrap())
 }
 
 /// Returns the sorting vector for the row, or null if none exists.
--- a/src/redisearch_rs/c_entrypoint/sorting_vector_ffi/src/lib.rs
+++ b/src/redisearch_rs/c_entrypoint/sorting_vector_ffi/src/lib.rs
@@ -38,7 +38,7 @@ pub unsafe extern "C" fn RSSortingVector_Get(
     // Safety: The caller must ensure that the pointer is valid (1.)
     let vec = unsafe { vec.as_ref().expect("vec must not be null") };
 
-    vec[idx].as_raw()
+    vec[idx].as_ptr()
 }
 
 /// Returns the length of the sorting vector.
--- a/src/redisearch_rs/sorting_vector/src/lib.rs
+++ b/src/redisearch_rs/sorting_vector/src/lib.rs
@@ -47,7 +47,7 @@ impl RSSortingVector {
     /// Creates a new [`RSSortingVector`] with the given length.
     pub fn new(len: usize) -> Self {
         Self {
-            values: vec![RSValueFFI::new_null(); len].into_boxed_slice(),
+            values: vec![RSValueFFI::null_static(); len].into_boxed_slice(),
         }
     }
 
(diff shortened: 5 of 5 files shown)


src/redisearch_rs/value/src/rs_value_ffi.rs
```diff
@@ -7,56 +7,18 @@
...
-
-    /// Gets the number (double) value of the RSValue instance, if it is a number type or None otherwise.
-    fn as_num(&self) -> Option<f64>;
-
-    /// Gets the type of the RSValue instance, it either null, string, number, or reference.
-    fn get_type(&self) -> ffi::RSValueType;
-
-    /// Returns true if the RSValue is stored as a pointer on the heap (the C implementation)
-    fn is_ptr_type() -> bool;
+pub struct RSValueFFIRef<'a>(ManuallyDrop<RSValueFFI>, PhantomData<&'a ffi::RSValue>);
```
daxhuiberts, line 14: Why is the `PhantomData<&a ffi::RSValue>` needed?

daxhuiberts: Oh, it's to add a lifetime to this ref which otherwise doesn't have a place to attach to, right?

daxhuiberts: @memark Might be a good trick to that last `HiddenStringRef` change of yours as well.

JonasKruckenberg: Because conceptually we'd like to return `&ffi::RSValue` from the `deep_deref` method correct? Using the PhantomData we get a type that is a pointer but behaves on a type level like `&ffi::RSValue`.

JonasKruckenberg: FYI this pattern is used by almost all ref types in std (e.g. https://doc.rust-lang.org/src/core/cell.rs.html#1592)

src/value/value.c
kei-nan, line 706: I assume this is not auto generated code, correct?
maybe use rm_calloc to ensure the array will be zeroed out.

kei-nan: who uses this code btw?

JonasKruckenberg: it was leftover from rebasing, removed

src/redisearch_rs/value/src/rs_value_ffi.rs
```diff
@@ -102,44 +47,66 @@ impl Drop for RSValueFFI {
...
+    /// 1. The `ptr` must be a [valid] pointer to a [`ffi::RSValue`].
+    ///
+    /// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+    pub const unsafe fn from_raw(ptr: NonNull<ffi::RSValue>) -> Self {
+        Self(ptr)
+    }
+
+    pub const fn as_raw(&self) -> *mut ffi::RSValue {
+        self.0.as_ptr()
+    }
```
daxhuiberts: The convention is to have `from_raw`, `into_raw`, `as_ptr`. How does `as_raw` fit in here?

JonasKruckenberg: good point

src/redisearch_rs/sorting_vector/tests/sorting_vector.rs
```diff
@@ -26,80 +27,82 @@ fn creation() {
     }
 }
 
-fn build_vector() -> Result<RSSortingVector<RSValueMock>, IndexOutOfBounds> {
-    let mut vector = RSSortingVector::new(5);
+fn build_vector() -> Result<RSSortingVector, IndexOutOfBounds> {
+    let mut vector = RSSortingVector::new(4);
     vector.try_insert_num(0, 42.0)?;
     vector.try_insert_string(1, b"abcdefg".to_vec())?;
-    vector.try_insert_val_as_ref(2, RSValueMock::create_num(3.))?;
```
daxhuiberts, line 33: Are refs not needed anymore as that is removed from the test here?

src/redisearch_rs/sorting_vector/tests/sorting_vector.rs
```diff
@@ -26,80 +27,82 @@ fn creation() {
...
+    let expected_size = std::mem::size_of::<RSValueFFI>() + RSValueFFI::mem_size();
     assert_eq!(size, expected_size);
 
     // test with more complex values
     let vector = build_vector()?;
     let size = vector.get_memory_size();
-
-    let expected_size = 5 * std::mem::size_of::<RSValueMock>() // 5 RSValues
+    let expected_size = 4 * std::mem::size_of::<RSValueFFI>() // 4 RSValue pointers
+            + (3 * RSValueFFI::mem_size()) // 3 actually allocated RSValues (the 4th is a null)
```
daxhuiberts, line 105: An `RSValue` `Null` also lives on the heap, so should also count towards the size.

daxhuiberts: Oh, never mind. It's the null static, which indeed doesn't count. (a non static null is also possible, in the new setup this will be made much more explicit)

src/redisearch_rs/value/src/rs_value_ffi.rs
```diff
@@ -83,63 +62,51 @@ impl RSValueFFI {
...
-        // Safety: We assume a valid ptr is given by the C side, and we are decrementing the reference count.
-        unsafe { ffi::RSValue_DecrRef(self.0.as_ptr()) };
+    pub fn ptr_eq(this: &Self, other: &Self) -> bool {
+        this.0 == other.0
     }
-}
 
-impl RSValueTrait for RSValueFFI {
-    fn create_null() -> Self {
+    pub fn new_null() -> Self {
```
daxhuiberts: Let's rename this to `null_static` so it's clear it doesn't create a new `RSValue` object. This new name will be used in the new setup as well.

src/redisearch_rs/sorting_vector/tests/sorting_vector.rs
JonasKruckenberg: yup, the refs code was not actually needed so it doesn't exist anymore

src/redisearch_rs/sorting_vector/tests/sorting_vector.rs
JonasKruckenberg: okayyyy, but question why do we even have a heap allocated null to begin with?

daxhuiberts: It's a nasty artifact.`RSValue`s were both stack allocated or heap allocated, reference counted object, in the same type which makes lifetime tracking very difficult. In Rust code you will probably pass around a reference to an `RsValue` where possible, but in C it's always passed around through a `SharedRsValue` (`Arc`).

daxhuiberts (approved):

src/redisearch_rs/sorting_vector/tests/sorting_vector.rs
JonasKruckenberg: yes I know that but the special null rsvalue should never be heap allocatable no?

daxhuiberts: There is one place in `expression.c` where it still happens. We also have `Undefined` which is similarly heap allocated and that is used throughout the codebase much more.

PR #8228, opened 2026-01-30
MOD-13792 chore: add more RSValueFFI constructors for arrays, maps, and trios

JonasKruckenberg:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches unsafe FFI allocation/ownership and refcount transfer; mistakes here can cause leaks or UB, though the change is localized to constructors and builder helpers.
> 
> **Overview**
> Adds new `RSValueFFI` constructors (`new_array`, `new_map`, `new_trio`) that build composite values via the existing C/Rust FFI builders and explicitly transfer ownership by `mem::forget`-ing inputs.
> 
> Refactors `RSValue_NewArrayBuilder`/`RSValue_NewArrayFromBuilder` to allocate and reclaim the builder buffer using a zeroed `Box` slice (instead of `Vec::from_raw_parts`), and adds const layout assertions to ensure `RSValueFFI` stays ABI-compatible with a raw `*mut RSValue`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2d4aa841b3b0f0a66caf0e9878fe518825ac2ee1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
+++ b/src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
@@ -7,6 +7,8 @@
  * GNU Affero General Public License v3 (AGPLv3).
 */
 
+use std::{mem::MaybeUninit, ptr};
+
 use crate::util::expect_value;
 use value::{Array, RsValue, shared::SharedRsValue};
 
@@ -20,9 +22,13 @@ use value::{Array, RsValue, shared::SharedRsValue};
 /// 1. The caller must eventually pass the returned pointer to [`RSValue_NewArrayFromBuilder`].
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn RSValue_NewArrayBuilder(len: u32) -> *mut *mut RsValue {
-    let array: Vec<*mut RsValue> = vec![std::ptr::null_mut(); len as usize];
+    let array = Box::new_zeroed_slice(len as usize);
 
-    Box::into_raw(array.into_boxed_slice()) as *mut _
+    // Safety: we zero-initialized the slice above. It is therefore correctly initialized with
+    // null pointers are required.
+    let array = unsafe { Box::<[MaybeUninit<*mut RsValue>]>::assume_init(array) };
+
+    Box::into_raw(array).cast::<*mut RsValue>()
 }
 
 /// Creates a heap-allocated array [`RsValue`] from existing values.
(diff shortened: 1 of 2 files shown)


src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
```diff
@@ -22,7 +22,12 @@ use value::{Array, RsValue, shared::SharedRsValue};
 pub unsafe extern "C" fn RSValue_NewArrayBuilder(len: u32) -> *mut *mut RsValue {
     let array: Vec<*mut RsValue> = vec![std::ptr::null_mut(); len as usize];
 
-    Box::into_raw(array.into_boxed_slice()) as *mut _
+    let (ptr, len, cap) = array.into_raw_parts();
+    debug_assert_eq!(
+        len, cap,
+        "array length did not equal array capacity, this will leak memory!"
+    );
+    ptr
```
daxhuiberts: What is the reason to replace `into_boxed_slice` with `into_raw_parts` and then checking `len == cap`? `into_boxed_slice` will shrink to fit if needed automatically.

JonasKruckenberg: ahh that is what you were using that for? I see. I removed it because you were converting the `Vec<T>` to a `Box<[T]>` and then reconstructing a `Vec<T>` from that `Box<T>` which only accidentally works but is not sound.

If you actually needed the shrinking you should use [`Vec:: shrink_to_fit`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.shrink_to_fit) and then `into_raw_parts`. But you construct the vect with an explicit length above, shrinking shouldn't be required.

daxhuiberts: Perhaps the construction should happen with `std::slice::from_raw_parts(ptr, len).to_vec()` instead. The array should never be growable anyways (it's stored as a `Box<[T]>` inside the `RsValue::Array`). That way we don't need to check the capacity in a `debug_assert` like this. That sounds like the better approach.

JonasKruckenberg: okay but why don't we just recover a `Box<[T]>`? seems like it would solve the problem much easier

JonasKruckenberg: updated with a solution based on this 👍🏻

daxhuiberts: This is the way!

PR #8238, opened 2026-02-01
MOD-13179  Add disk-mode validation for HNSW vector indexes

ofiryanai:
Enforce strict validation for disk-based HNSW indexes:

- Reject FLAT algorithm (only HNSW supported on disk)
- Reject FLOAT64 type (only FLOAT32 supported on disk)
- Require explicit M, EF_CONSTRUCTION, EF_RUNTIME parameters (no defaults)
- Add RERANK boolean parameter parsing (required for disk mode)
- Reject JSON indexes (only HASH supported for Flex/disk indexes)

This ensures users explicitly configure disk indexes rather than relying on RAM defaults that may not be optimal for disk-based storage.

**Current**: Disk-based HNSW vector indexes accept the same parameters as RAM indexes, using RAM defaults when not specified. We decided to require all params since we have no suggestions yet for defaults on this new index type.
**Change**: Add strict validation for disk-mode HNSW indexes: reject unsupported features (FLAT algorithm, FLOAT64 type), require explicit `M`/`EF_CONSTRUCTION`/`EF_RUNTIME` parameters (no defaults), and add `RERANK` boolean parameter parsing.

1. `spec.c` - Added disk-mode validation in `parseVectorField_hnsw()` and `parseVectorField()`, including `RERANK` flag parsing and mandatory parameter checks for disk indexes.

#### Mark if applicable

#### Release Notes

- [X] This PR does not require release notes

---

> [!NOTE]
> **Medium Risk**
> Medium risk: tightens validation and adds a new `RERANK` flag, which changes `FT.CREATE` behavior for disk/Flex indexes and could break existing index-creation calls relying on defaults or unsupported options.
> 
> **Overview**
> **Adds strict disk/Flex validation for vector index creation.** When disk validation is enabled, `FT.CREATE` now rejects `FLAT` and `SVS-VAMANA` vector algorithms, and enforces that disk-backed HNSW indexes use only `FLOAT32`, do not allow multi-value vectors, and require explicit `M`, `EF_CONSTRUCTION`, `EF_RUNTIME`, and a new boolean `RERANK` flag.
> 
> **Improves HNSW argument parsing and propagates disk options.** HNSW parsing now slices the declared parameter count to detect “not enough arguments” reliably, plumbs `rerank` into `VecSimDiskContext`, and defaults `rerank=true` when rebuilding disk context during RDB load (not yet persisted). Tests are updated to match the new error behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0e69eddec0462c8bed925279bcba893e672072d6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -1665,16 +1665,12 @@ static void IndexSpec_PopulateVectorDiskParams(IndexSpec *sp) {
       rm_free((void*)fs->vectorOpts.diskCtx.indexName);
     }
 
-    // TODO(MOD-13179): The rerank field is not persisted in RDB. When the disk layer
-    // starts using rerank, we need to either:
-    // 1. Add rerank to RDB save/load (requires new RDB version), or
-    // 2. Store rerank in disk index metadata and read it back when opening
-    // For now, rerank defaults to false on RDB load.
+    // TODO: rerank is not persisted in RDB, defaulting to true on load.
     fs->vectorOpts.diskCtx = (VecSimDiskContext){
       .storage = sp->diskSpec,
       .indexName = rm_strndup(namePtr, nameLen),
       .indexNameLen = nameLen,
-      .rerank = false,
+      .rerank = true,
     };
   }
 }


src/spec.c
```diff
@@ -746,27 +749,33 @@ int VecSimIndex_validate_params(RedisModuleCtx *ctx, VecSimParams *params, Query
...
-  } else if (expNumParam % 2) {
-    QueryError_SetWithoutUserDataFmt(status, QUERY_ERROR_CODE_SYNTAX, "Bad number of arguments for vector similarity index: got %d but expected even number (as algorithm parameters should be submitted as named arguments)", expNumParam);
-    return 0;
-  } else {
-    expNumParam /= 2;
   }
+  // Calculate expected key-value pairs, accounting for potential boolean flags
+  // Boolean flags (like RERANK) count as 1 argument, not 2
+  size_t expPairs = expNumParam / 2;
+  size_t remainder = expNumParam % 2;
```
GuyAv46: Let's simply create a sub-cursor (slice) by this given number and ensure it is consumed when we finish the iteration.
These variable only works while we have a single standalone param, and will break if we add another one

ofiryanai: changed it to slice 👍

src/spec.c
```diff
@@ -1596,10 +1665,16 @@ static void IndexSpec_PopulateVectorDiskParams(IndexSpec *sp) {
       rm_free((void*)fs->vectorOpts.diskCtx.indexName);
     }
 
+    // TODO(MOD-13179): The rerank field is not persisted in RDB. When the disk layer
+    // starts using rerank, we need to either:
+    // 1. Add rerank to RDB save/load (requires new RDB version), or
+    // 2. Store rerank in disk index metadata and read it back when opening
+    // For now, rerank defaults to false on RDB load.
```
GuyAv46: Doesn't it mean we cannot load from RDB at the moment? Maybe we should at least set it to `true` at the current RDB handler version?

src/spec.c
```diff
@@ -1596,10 +1665,16 @@ static void IndexSpec_PopulateVectorDiskParams(IndexSpec *sp) {
...
     fs->vectorOpts.diskCtx = (VecSimDiskContext){
       .storage = sp->diskSpec,
       .indexName = rm_strndup(namePtr, nameLen),
       .indexNameLen = nameLen,
+      .rerank = false,
```
GuyAv46: Set to `true` until we properly add it to the RDB and/or allow to disable it

ofiryanai: right, set it to true, sorry about that :)

src/spec.c
ofiryanai: I removed most of the comment. not sure what's the rdb loading status on this project, set the default to true anyway

PR #8245, opened 2026-02-02
Design document and discussion for hybrid and optimizer iterators - [MOD-13060]

GuyAv46:
#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Documentation-only change; no production code or API behavior is modified.
> 
> **Overview**
> **Release note:** Adds a draft design document (`docs/design/TOP_K_DESIGN.md`) proposing a unified Rust `TopKIterator` abstraction to replace the current C `HybridIterator` (vector KNN) and `OptimizerIterator` (numeric SORTBY), including planned execution modes (unfiltered/batches/adhoc), strategy switching, and a staged implementation/testing plan. No runtime or API behavior changes are included in this PR.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 87e1c4bee5be6ec5367e22f82b4295e8c10427fa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


docs/design/TOP_K_DESIGN.md
```diff
@@ -0,0 +1,961 @@
...
+
+### Q1: Trait Object vs. Monomorphization
+
+**Current design:** `TopKIterator<S: ScoreSource<'index>>` uses static dispatch (monomorphization).
+
+**Question:** Should we use `Box<dyn ScoreSource>` instead for smaller binary size and faster compilation?
+
+**Trade-off:**
+- Static dispatch: Better inlining, zero vtable overhead, but larger binary
+- Dynamic dispatch: Smaller binary, but vtable overhead on every trait method call
```
LukeMathWalker: Where possible, we've always preferred static dispatching (see inverted index).
If we know the score source at compile-time (and it looks like we do, when building the iterator tree), I would avoid dynamic dispatching.

docs/design/TOP_K_DESIGN.md
```diff
@@ -0,0 +1,961 @@
...
+```rust
+pub trait ScoreBatch {
+    fn next(&mut self) -> Option<(t_docId, f64)>;
+    fn skip_to(&mut self, target: t_docId) -> Option<(t_docId, f64)>;
+}
+```
+
+**Trade-off:**
+- Vec: Simple, good cache locality, but requires allocation
+- Iterator: Zero-copy from VecSim, but more complex API
```
LukeMathWalker: I don't have the full context here: does `VecSim` allow you to iterate over results without allocating a fresh buffer, "borrowing" from the underlying vecsim iterator?
If that's the case, the iterator interface would be better. Alternatively, if you know you're _always_ collecting, you want to design an interface that allows you to reuse the same buffer multiple times (unless you need to give that buffer away to the caller).

docs/design/TOP_K_DESIGN.md
```diff
@@ -0,0 +1,961 @@
...
+
+```rust
+// Alternative: more explicit methods
+fn build_metric_result(&self, doc_id: t_docId, score: f64) -> RSIndexResult<'index>;
+fn build_aggregate_result(&self, doc_id: t_docId, score: f64, child: &RSIndexResult) -> RSIndexResult<'index>;
+```
+
+**Trade-off:**
+- Single method: Simpler trait, source decides internally
+- Multiple methods: More explicit, but TopK needs to know which to call
```
LukeMathWalker, line 730: Difficult to judge upfront. I would try to implement and then assess.

docs/design/TOP_K_DESIGN.md
```diff
@@ -0,0 +1,961 @@
...
+
+**Question:** Should we use const generics to eliminate branches in hot paths?
+
+```rust
+pub struct TopKIterator<'index, S: ScoreSource, const ASC: bool, const FILTERED: bool> { ... }
+```
+
+**Trade-off:**
+- More combinations to compile (2x2 = 4 variants)
+- But guaranteed branch elimination in hot loops
```
LukeMathWalker: I don't think this needs to be determined upfront. I would get something up and running, integrate it, then bench and compare with a version that uses const generics to determine the actual complexity vs speed-up trade-off.

LukeMathWalker:
I'm not familiar with the original C implementations, so I'm judging based on the context embedded in the plan—it does sound sensible from a Rust design perspective.
At the same time, I do wonder if it'd be more productive to delay unification: port each iterator to Rust first, cleaning up their design in the process and leveraging the specificity of the underlying source. 
Then go through another planning phase to discuss unification options, with a better understanding of the trade-offs.

It'd make the time to integration smaller, probably surface additional constraints/requirements in the process, and minimize the risk of unifying prematurely/using the wrong interface.

docs/design/TOP_K_DESIGN.md
GuyAv46: I agree

docs/design/TOP_K_DESIGN.md
GuyAv46: I also think `Iterator` is better
1. On Vector, we have an iterator API over an opaque (C++) vector of results.
2. On Numeric, it will be an iteration over some numeric inverted indexes (probably a union iterator over a selected set of ranges).

GuyAv46: The vector API always allocates a vector for the results today

docs/design/TOP_K_DESIGN.md
```diff
@@ -0,0 +1,961 @@
...
+
+## Analysis: Shared Patterns
+
+After analyzing both implementations, we identified these shared patterns:
+
+| Aspect | Hybrid | Optimizer | Shared? |
+|--------|--------|-----------|---------|
+| **Unfiltered mode** | ✅ KNN (no child) | ✅ Pure SORTBY (should exist) | ✅ |
+| **Batches mode** | ✅ VecSim batches | ✅ Numeric range subsets | ✅ |
+| **Adhoc-BF mode** | ✅ Distance lookup | ✅ Value lookup | ✅ |
```
alonre24: What about the Heuristics that select "Batches mode" vs. "Adhoc-BF mode"? Does it exist today in the optimizer?

docs/design/TOP_K_DESIGN.md
```diff
@@ -0,0 +1,961 @@
...
+| **Top-k heap** | ✅ Min-max heap | ✅ Min/max heap | ✅ |
```
alonre24: min/max heap? in Hybrid vector, it's just a max heap, as I recall

docs/design/TOP_K_DESIGN.md
```diff
@@ -0,0 +1,961 @@
...
+**Rationale:** Allows capturing domain-specific metrics (batch count, strategy switches) without Profile knowing about TopK internals.
+
+### D6: Naming
+
+**Decision:** Rename the iterators to reflect what they actually do.
+
+| Old Name (C) | New Name (Rust) | Rationale |
+|--------------|-----------------|-----------|
+| `hybrid_reader` / `HybridIterator` | `VectorTopKIterator` | "Hybrid" is an internal implementation detail (hybrid search modes). The iterator performs **vector similarity top-k**. |
+| `optimizer_reader` / `OptimizerIterator` | `NumericTopKIterator` | "Optimizer" is vague and misleading. The iterator performs **numeric field top-k** (for SORTBY). |
```
alonre24, line 517: I definitely agree with that

docs/design/TOP_K_DESIGN.md
```diff
@@ -0,0 +1,961 @@
...
+- The current implementation uses a hacky union iterator
+- Retry heuristics are fragile and hard-coded
+- Complex state management leads to bugs
+- TODO comment: `VALIDATE_MOVED` not properly handled
+- Potential division by zero in `getSuccessRatio` when `lastLimitEstimate == 0`
+
+### Corrected Understanding (Optimizer)
+
+The optimizer should work with **sorted numeric ranges** where each "batch" is a subset of ranges ordered by numeric value. The current union-based hack should be replaced with proper range-based iteration.
+
```
alonre24, line 147: Can you elaborate on that? Also, what is the behaviour / breaking change expected due to this change if any?

docs/design/TOP_K_DESIGN.md
GuyAv46: We use min-max heap for 2 years now :)
Since we started supporting `FT.AGGREGATE`

docs/design/TOP_K_DESIGN.md
GuyAv46: The optimizer currently only has a "batch-mode", and when the hit rate is low, we restart with a larger chunk size. I don't think we will actually implement a value lookup (although it is possible with sortable fields)

docs/design/TOP_K_DESIGN.md
GuyAv46: Sure.
In short, no logical changes, only implementing it correctly with proper handling, and not with a hacked union iterator

docs/design/TOP_K_DESIGN.md
GuyAv46: Agreed. I deferred const generics to the post-integration optimization phase, after baseline perf measurements.

docs/design/TOP_K_DESIGN.md
GuyAv46: Agreed. I changed this to a deferred question: we’ll start with a single `build_result` path, integrate, then assess whether splitting by result type is worth the complexity.

docs/design/TOP_K_DESIGN.md
GuyAv46: Clarified in the doc: optimizer currently has batches+retry (expand range + rewind) and no adhoc score-lookup mode in v1. Numeric adhoc remains a possible future extension, not part of initial scope.

docs/design/TOP_K_DESIGN.md
GuyAv46: Thanks, aligned in the doc.

docs/design/TOP_K_DESIGN.md
GuyAv46: Expanded this section with explicit compatibility expectations: no intended user-visible semantic changes (matching/order), internal control-flow cleanup only, with bug fixes as expected outcomes.

alonre24 (dismissed):

PR #8254, opened 2026-02-03
enforce sctx passed to all II iterators

gdesmott:
The Rust II numerators make the `sctx` parameter mandatory so let's enforce it already.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches core inverted-index iterator revalidation and spellcheck scoring logic; mistakes could cause incorrect abort behavior or resource leaks during query execution, though most changes are test/context wiring.
> 
> **Overview**
> **Enforces passing a valid `RedisSearchCtx` into inverted-index iterators.** C++ tests/benchmarks now construct `MockQueryEvalCtx` and pass `&mockQctx.sctx` into `NewInvIndIterator_*` and related helpers instead of `NULL`.
> 
> **Adjusts iterator validation to tolerate missing spec state in tests.** `TermCheckAbort` now skips revalidation when `sctx`, `sctx->spec`, or `spec->keysDict` is absent.
> 
> **Decouples spellcheck scoring from iterator construction.** `SpellCheck_GetScore` switches from building a term iterator to using `IndexReader` directly to check for at least one matching document under a `fieldMask`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f35b1f79c76c3e1ef893c9a6b2980b3ca5f3eee7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/cpptests/test_cpp_iterator_index.cpp
+++ b/tests/cpptests/test_cpp_iterator_index.cpp
@@ -226,15 +226,15 @@ public:
             AddEntry(docId, value);
         }
     }
-    void CreateIterator(double value) {
-        CreateIterator(value, value);
+    void CreateIterator(double value, const RedisSearchCtx *sctx) {
+        CreateIterator(value, value, sctx);
     }
-    void CreateIterator(double min, double max) {
+    void CreateIterator(double min, double max, const RedisSearchCtx *sctx) {
         ASSERT_TRUE(idx != nullptr);
         FieldMaskOrIndex fieldMaskOrIndex = {.index_tag = FieldMaskOrIndex_Index, .index = RS_INVALID_FIELD_INDEX};
         FieldFilterContext fieldCtx = {.field = fieldMaskOrIndex, .predicate = FIELD_EXPIRATION_PREDICATE_DEFAULT};
         flt = NewNumericFilter(min, max, 1, 1, 1, nullptr);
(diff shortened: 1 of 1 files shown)


src/iterators/inverted_index_iterator.c
```diff
@@ -57,8 +57,9 @@ static ValidateStatus TermCheckAbort(QueryIterator *base) {
     return VALIDATE_OK;
   }
   RSQueryTerm *term = IndexResult_QueryTermRef(base->current);
-  // sctx and term should always be set, except in some tests.
-  RS_ASSERT(term);
+  if (!term) {
+    return VALIDATE_OK;
+  }
```
GuyAv46: What changed here? Consider keeping the comment

gdesmott: I provided more context in the commit message.

See https://github.com/RediSearch/RediSearch/blob/master/src/spell_check.c#L97 creating an iterator with `sctx` but no `term`.

So I assumed this is actually a valid use cases. Or should we change this code instead?

GuyAv46: I think we should change the spellcheck. We can open a reader without an iterator, no?

gdesmott: Indeed that's a good idea.

I've done that but then also had to add a check on `spec->keysDict` as it's not defined in mock contexts.

PR #8268, opened 2026-02-04
pass offset info to Disk API

JoanFM:
> [!NOTE]
> **Medium Risk**
> Touches core disk-index query evaluation and indexing paths by changing Disk API signatures and conditionalizing offset loading, which could affect search correctness/performance on disk-backed indexes. Rust/C FFI enum/layout changes add some integration risk across the boundary but are covered by new tests.
> 
> **Overview**
> Enables disk-backed indexing and querying to **optionally carry term offset (position) bytes end-to-end**.
> 
> Indexing now forwards varint-encoded offsets (when `Index_StoreTermOffsets` is enabled) into `SearchDisk_IndexTerm`, and disk term iterators (`SearchDisk_NewTermIterator`) are created with a `needsOffsets` flag computed from *phrase/slop constraints* and the selected *scorer* (BM25 std variants skip offsets; legacy/custom default to requiring them).
> 
> On the Rust side, adds `RSTermRecord::FullyOwned` to allow decoded term records to own both the query term and offsets (safe when source bytes are transient, e.g. disk pages), updates FFI/header types and helpers accordingly, and adds integration tests for the new ownership/offset-setting behavior. Flex validation tests were updated by removing the cases that previously expected TFIDF/BM25 scorer blocking.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dcd31e113677ccdfddb916a519b1bf5d114cb847. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -293,13 +293,15 @@ pub trait SearchEnterpriseIterators: Send + Sync {
...
+    /// the results according to the given field mask. When `needs_offsets` is true the iterator
+    /// will load offset data for each document; otherwise offset data is skipped for efficiency.
     fn new_term_on_disk<'index>(
         &self,
         index: &'index ffi::RedisSearchDiskIndexSpec,
         query_term: Box<RSQueryTerm>,
         field_mask: t_fieldMask,
         weight: f64,
+        needs_offsets: bool,
     ) -> Result<Box<dyn RQEIterator<'index> + 'index>, Box<dyn std::error::Error>>;
```
chesedo, line 305: Let's add a new method instead

src/redisearch_rs/inverted_index/tests/integration/index_result.rs
```diff
@@ -376,3 +376,94 @@ fn full_test_mirrors_cpp_testdistance() {
...
+//
+// The `FullyOwned` variant owns both the query term (via `Box`) and the
+// offsets (via `RSOffsetVector`), so the resulting `RSIndexResult` is
+// independent of the original offset byte source.
+
+/// Build a `FullyOwned`-backed result, drop the source bytes, and verify the
+/// record still reads back correctly. Also exercises the `is_copy`,
+/// `offsets`, and `query_term` match arms for the new variant.
+#[test]
+fn fully_owned_term_result_is_independent_of_source_bytes() {
```
chesedo, line 390: Where does this test drop the source bytes?

Also, let's not refer to it as the "new variant". It won't be new months from now

chesedo (commented):
Looks good. Not sure what is up with the CI though

src/redisearch_rs/rqe_iterators/src/lib.rs
JoanFM: u mean have the new_term_on_disk_with_offsets? I think is better to keep the bool as argument? why do u prefer to have 2 APIs?

src/redisearch_rs/inverted_index/tests/integration/index_result.rs
JoanFM: updated clarifying

src/redisearch_rs/rqe_iterators/src/lib.rs
chesedo: Yes, having a new method. Passing a boolean into a function always flares up my code-smell sensor :sweat_smile: . Maybe a new method is not needed, but some refactor. The thing is that somewhere upstream we realized we need offsets and this bool causes that logic to be pushed down into deeper functions. Let's rather handle that logic at the location we know the offsets are needed.

JoanFM: In my opinion, this is the layer it belongs. I think is here where the decision on which actual iterator would be constructed, and to the outside it is mostly important that is disk related

chesedo: I don't agree. If this layer decides which iterator to construct then why did the caller ask for a term iterator and not a tag iterator. Clearly the caller is deciding which iterator it wants.

chesedo (approved):

PR #8269, opened 2026-02-04
MOD-12273: Support Expiration Check In Async Read

kei-nan:
Pass current time into polling function so we can check for document expiration in the async read path.

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches the async disk I/O query path and changes the disk API function signatures, so any mismatch in backend implementations or callsites could cause compile/runtime issues. Behavior change is limited to enabling expiration filtering during async metadata fetches.
> 
> **Overview**
> Threads an *expiration time point* through the async disk-read pipeline so expired documents can be filtered during async metadata fetches.
> 
> `IndexResultAsyncRead_Poll` and `SearchDisk_PollAsyncReads` now accept an `expiration_point` (passed from `sctx->time.current` in `result_processor.c`), and the disk API surface (`DocTableDiskAPI.pollAsyncReads` / `getDocumentMetadata`) is updated accordingly (including parameter renames/types) to support expiration checks in the backend.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bf4cce78ca01d4154765650517b048bcdd68bf81. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/index_result_async_read.c
+++ b/src/index_result_async_read.c
@@ -131,11 +131,11 @@ static void IndexResultAsyncRead_CleanupFailedReads(IndexResultAsyncReadState *s
   }
 }
 
-size_t IndexResultAsyncRead_Poll(IndexResultAsyncReadState *state, uint32_t timeout_ms) {
+size_t IndexResultAsyncRead_Poll(IndexResultAsyncReadState *state, uint32_t timeout_ms, const t_expirationTimePoint *expiration_point) {
   // Poll writes directly to the arrays (capacity is poolSize)
   const size_t pendingCount = SearchDisk_PollAsyncReads(
       state->asyncPool, timeout_ms,
-      state->readyResults, state->failedUserData);
+      state->readyResults, state->failedUserData, expiration_point);
 
   // Reset index to start consuming from the beginning of readyResults
   state->readyResultsIndex = 0;
--- a/src/result_processor.c
+++ b/src/result_processor.c
@@ -326,7 +326,7 @@ static int rpQueryItNext_AsyncDisk(ResultProcessor *base, SearchResult *res) {
 
     // Step 3: No ready results - poll for more
     int timeout_ms = it->atEOF ? ASYNC_POLL_TIMEOUT_AT_EOF_MS : 0;
-    const size_t pendingCount = IndexResultAsyncRead_Poll(&self->async, timeout_ms);
+    const size_t pendingCount = IndexResultAsyncRead_Poll(&self->async, timeout_ms, &sctx->time.current);
 
     // Step 4: Check if we're completely done
     if (IndexResultAsyncRead_IsIterationComplete(&self->async, it->atEOF, pendingCount)) {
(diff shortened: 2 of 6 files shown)


src/index_result_async_read.c
```diff
@@ -131,11 +131,11 @@ static void IndexResultAsyncRead_CleanupFailedReads(IndexResultAsyncReadState *s
   }
 }
 
-size_t IndexResultAsyncRead_Poll(IndexResultAsyncReadState *state, uint32_t timeout_ms) {
+size_t IndexResultAsyncRead_Poll(IndexResultAsyncReadState *state, uint32_t timeout_ms, const t_expirationTimePoint *expiration_point) {
```
Itzikvaknin, line 134: I think you should use `struct timespec`, also, what do we get from passing a const pointer here?

kei-nan: pointer - we avoid copying the entire timespec struct
const -caller knows it won't be modified by the function.

src/result_processor.c
```diff
@@ -326,7 +326,7 @@ static int rpQueryItNext_AsyncDisk(ResultProcessor *base, SearchResult *res) {
 
     // Step 3: No ready results - poll for more
     int timeout_ms = it->atEOF ? ASYNC_POLL_TIMEOUT_AT_EOF_MS : 0;
-    const size_t pendingCount = IndexResultAsyncRead_Poll(&self->async, timeout_ms);
+    const size_t pendingCount = IndexResultAsyncRead_Poll(&self->async, timeout_ms, &sctx->time.current);
```
raz-mon, line 329: Here you send `current`, while you have `expiration_point` in the API.
Bug or misnaming?

src/index_result_async_read.c
kei-nan: also:
```c
typedef struct timespec t_expirationTimePoint;
```

src/result_processor.c
kei-nan: it was intended this way.
this is the point in time for the expiration checks

src/index_result_async_read.c
Itzikvaknin: `struct timespec` - POSIX standard 
the alias may change 

so in case of a change, rust will only know about that in runtime 
also it only copies 16 bytes instead of 8 - are we that greedy 😅 ?

kei-nan: if the type will change then rust side will need to know about the change.
the reason the type exists is to represent the point in time for the expiration, either the source or the comparison point.

src/result_processor.c
raz-mon: It's very confusing.
Especially given the `SearchTime` struct:
```
typedef struct {
  // current execution start time - real clock
  struct timespec current;
  // when the query should timeout - monotonic raw clock, unrelated to real clock
  struct timespec timeout;
} SearchTime;
```
Anyways, I see that this is the time we check against in OSS as well, so ok..

PR #8274, opened 2026-02-04
MOD-13110: Fix FT.HYBRID concurrency issues

nafraf:
## Description

**1. Spec Lock and Try-Lock Handoff for Background Depletion**
 
  - Problem: When running `FT.HYBRID` with background workers (WORKERS > 0), the spec read lock was not taken before building the pipeline. This caused concurrency issues where the index could be modified while the pipeline was being built, leading to crashes or undefined behavior during concurrent write operations (e.g., slot migration).

- Solution:
    - Added spec lock acquisition in the main thread before building the pipeline
    - Implemented a coordinated lock handoff mechanism with non-blocking try-lock to prevent deadlock

<details>
<summary>Lock Handoff Flow</summary>

1.   Main thread acquires spec read lock before building the pipeline (HREQ_Execute_Callback)
2.  Background depleters are spawned and each tries to acquire their own lock using non-blocking try-lock
3.  Main thread waits in `RPSafeDepleter_WaitForDepletionToStart` until all depleters have either:
    - Acquired their lock (`num_locked++`), or
    - Skipped locking (`num_skipped++`) due to timeout or lock failure
4.  Main thread releases its lock once all depleters have completed the lock acquisition phase
5.  If any depleter failed (`any_failed == true`), the entire operation fails with an error

</details>

**2: Add foreground depleters for (WORKERS 0)**

  - Problem: When running `FT.HYBRID` without background workers (WORKERS 0) during concurrent write operations, a crash could occur. The cursor was returned to the client before iterators were initialized. If the index was modified between cursor creation and the first `_FT.CURSOR READ`, the uninitialized iterator state caused crashes.
  
  <details>
  <summary>Failure sequence</summary>

1. Coordinator receives `FT.HYBRID` command
2. Coordinator dispatches `_FT.HYBRID ... WITHCURSOR` to each shard
3. Each shard's `HybridRequest_StartCursors` creates cursors without reading any data
4. Cursors are paused, spec lock is released
5. Index is destroyed (slot migration frees the index)
6. Coordinator calls `_FT.CURSOR READ` on shard cursors
7. `handleSpecLockAndRevalidate` tries to revalidate, but ii is now a dangling pointer
8. `needs_revalidation()` reads garbage that matches stored `gc_marker` → returns false
9. CRASH in `skip_to` when accessing invalid buffer pointers
   
   </details>

  - Solution: `Make WORKERS == 0` behave like `WORKERS > 0`, but synchronously on the main thread.

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches spec locking and cursor/depletion execution paths for `FT.HYBRID`, which are concurrency-sensitive and could introduce deadlocks, lock contention, or new failure modes under load despite being a targeted bug fix.
> 
> **Overview**
> Fixes `FT.HYBRID` concurrency around cursor creation/depletion by **holding the index spec read lock while building the background pipeline** and handing it off to background depleters via a **non-blocking try-lock**; if any depleter can’t acquire the lock, the request fails with a retriable error.
> 
> For `WORKERS==0`, adds a **foreground `RP_DEPLETER`** and forces synchronous depletion before returning cursors to avoid crashes when the index mutates between cursor creation and first read; profiling output is updated to include `Depleter` timing.
> 
> Adds `RedisSearchCtx_TryLockSpecRead`, extends `RPSafeDepleter` coordination to track skipped/failed lock acquisition, and updates tests to re-enable hybrid slot-migration scenarios and tolerate expected lock-acquisition failures during concurrent writes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4552eb2d72fa2c2a4d7603595af42a19407836a0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/aggregate/aggregate.h
+++ b/src/aggregate/aggregate.h
@@ -31,14 +31,11 @@
 #define RS_Atomic(T) std::atomic<T>
 extern "C" {
 #else
-#include <stdatomic.h>
 #define RS_Atomic(T) _Atomic(T)
+#include <stdatomic.h>
 #endif
 
 #define DEFAULT_LIMIT 10
-
-// Forward declaration for Cursor (defined in cursor.h)
-typedef struct Cursor Cursor;
 
 /** Cached variables to avoid serializeResult retrieving these each time */
 typedef struct {
@@ -325,7 +322,6 @@ typedef struct AREQ {
 
   // Flag to indicate whether to skip timeout checks using clock checks
   bool skipTimeoutChecks;
-
 } AREQ;
 
 /**
--- a/src/aggregate/aggregate.h
+++ b/src/aggregate/aggregate.h
@@ -327,11 +327,14 @@ typedef struct AREQ {
   bool skipTimeoutChecks;
 
   // Prefetched results buffer for FT.HYBRID cursor initialization.
-  // When a cursor is created without sending results (e.g., _FT.HYBRID WITHCURSOR on shards),
-  // we read and buffer the initial results here to initialize iterator state.
-  // These are returned on the first FT.CURSOR READ before continuing normal iteration.
+  // When a cursor is created without sending results
+  // (e.g., _FT.HYBRID WITHCURSOR on shards), we read and buffer the initial
+  // results here to initialize iterator state.
+  // These are returned on the first FT.CURSOR READ before continuing normal
+  // iteration.
   SearchResult **prefetchedResults;
-  int prefetchedResultsRc;  // Return code from prefetch (RS_RESULT_OK, RS_RESULT_EOF, etc.)
+  // Return code from prefetch (RS_RESULT_OK, RS_RESULT_EOF, etc.)
+  int prefetchedResultsRc;
 } AREQ;
 
 /**
@@ -527,8 +530,10 @@ void AREQ_DecrRef(AREQ *req);
 int AREQ_StartCursor(AREQ *r, RedisModule_Reply *reply, StrongRef spec_ref, QueryError *status, bool coord);
 
 /**
- * Prefetch cursor results for FT.HYBRID by reading initial results without sending them.
- * This initializes the iterator state, allowing safe revalidation on later cursor reads.
+ * Prefetch cursor results for FT.HYBRID by reading initial results without
+ * sending them.
+ * This initializes the iterator state, allowing safe revalidation on later
+ * cursor reads.
  * Results are buffered and returned on the first FT.CURSOR READ.
  *
  * @param cursor The cursor to prefetch
--- a/src/aggregate/aggregate.h
+++ b/src/aggregate/aggregate.h
@@ -26,6 +26,9 @@
 
 #include "rmutil/rm_assert.h"
 
+// Forward declaration for Cursor (defined in cursor.h)
+typedef struct Cursor Cursor;
+
 #ifdef __cplusplus
 #include <atomic>
 #define RS_Atomic(T) std::atomic<T>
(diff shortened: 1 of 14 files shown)


src/result_processor.c
```diff
@@ -1688,7 +1705,7 @@ static void RPSafeDepleter_DepleteFromUpstream(RPSafeDepleter *self, DepleterSyn
   }
 
   // Unlock the index if we locked it
-  if (sync->take_index_lock) {
```
kei-nan, line 1691: not sure this needs to change, we will only get to the "if" if we have the lock, no?

src/aggregate/aggregate_exec.c
```diff
@@ -437,7 +438,17 @@ static void sendChunk_Resp2(AREQ *req, RedisModule_Reply *reply, size_t limit,
       goto done_2_err;
     }
 
-    startPipeline(req, rp, &results, &r, &rc);
+    // Check for prefetched results from cursor initialization
+    // (_FT.HYBRID WITHCURSOR on shards)
+    if (req->prefetchedResults) {
```
kei-nan: We only ever use resp3 in coordinator for hybrid.
Are we sure this change is needed?
Can we try removing it and see if stuff still work?
Not sure I understand what we are trying to solve.

src/aggregate/aggregate.h
```diff
@@ -306,6 +309,16 @@ typedef struct AREQ {
   // Reference count for shared ownership between QueryNode (timeout callback) and background thread.
   // Initialized to 1. Free when reaches 0.
   uint8_t refcount;
+
+  // Prefetched results buffer for FT.HYBRID cursor initialization.
+  // When a cursor is created without sending results
+  // (e.g., _FT.HYBRID WITHCURSOR on shards), we read and buffer the initial
+  // results here to initialize iterator state.
```
kei-nan: what is the iterator state we are talking about here?
Looked at a bunch of the new comments and having a hard time understanding what the bug was this code solves.

kei-nan:
Regarding:
"Problem: When running FT.HYBRID without background workers (WORKERS 0) during concurrent write operations, a crash could occur. The cursor was returned to the client before iterators were initialized. If the index was modified between cursor creation and the first _FT.CURSOR READ, the uninitialized iterator state caused crashes."

I don't understand, why reading results is important for cursor state, can you elaborate a bit?
Don't we have pipelines that will try and access the iterator state after a write has happened?

src/aggregate/aggregate.h
```diff
@@ -500,6 +513,17 @@ void AREQ_DecrRef(AREQ *req);
  */
 int AREQ_StartCursor(AREQ *r, RedisModule_Reply *reply, StrongRef spec_ref, QueryError *status, bool coord);
 
+/**
+ * Prefetch cursor results for FT.HYBRID by reading initial results without
+ * sending them.
```
Itzikvaknin: We can remove the FT.HYBRID from the comment

src/aggregate/aggregate.h
nafraf: Outdated, prefetch changes were reverted.

src/aggregate/aggregate_exec.c
nafraf: outdated, prefetch solution was removed.

src/aggregate/aggregate.h
nafraf: Prefetch changes were removed.

src/result_processor.c
nafraf: Now it's `if (lock_acquired)`, and we try to acquire the lock only if `if (sync->take_index_lock)`

nafraf:
This PR was split, new PRs: #8533 and #8534

PR #8095, opened 2026-01-20
Port HyperLogLog implementation to Rust

LukeMathWalker:
## Describe the changes in the pull request

A Rust port of the HyperLogLog implementation from `hll.c`.

### Benchmarks

I've added benches for various hashers—the original C `fnv` hash, as well as a few others.
Speed-wise, `fnv` is the third fastest, but accuracy is an issue:

|        n |              fnv |          murmur3 |         xxhash32 |            ahash |           fxhash |           wyhash
-------| ---------------| -----------| ---------------| ---------------| ----------| ----------------
|     100 |    101 ( 1.00%) |    100 ( 0.00%) |    100 ( 0.00%) |    101 ( 1.00%) |     99 ( 1.00%) |    101 ( 1.00%)
|    1000 |   1036 ( 3.60%) |   1013 ( 1.30%) |    991 ( 0.90%) |    997 ( 0.30%) |    983 ( 1.70%) |   1004 ( 0.40%)
|   10000 |   8630 (13.70%) |  10397 ( 3.97%) |  10435 ( 4.35%) |   9858 ( 1.42%) |  10334 ( 3.34%) |  10527 ( 5.27%)
|  100000 |  52804 (47.20%) |  99283 ( 0.72%) | 100909 ( 0.91%) |  97607 ( 2.39%) |  98307 ( 1.69%) |  97695 ( 2.31%)

The table reports the correct cardinality in the `n` column, the estimated cardinality for each hasher as well as the error rate. 
`fnv` degrades really fast. All things considered (speed + accuracy), it might be a good idea to migrate over to either `fxhash` or `wyhash` later on.

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

---

> [!NOTE]
> Adds a self-contained `hyperloglog` crate implementing HLL with const generics, stack-allocated registers, cached counts, merge/clear, and pluggable hashers via `hash32::Hasher` (default `CFnvHasher` matching the C seed; `Murmur3Hasher` alias provided).
> 
> - New APIs: `HyperLogLog<BITS, SIZE, H>`, type aliases (`HyperLogLog10`, etc.), `add`, `add_precomputed_hash`, `count`, `merge`, `clear`, `from_registers`, `try_set_registers`
> - Benchmarks compare `fnv`, `murmur3`, `xxhash32`, `ahash`, `fxhash`, `wyhash`; prints accuracy and measures add/count/merge throughput
[...]

PR #8276, opened 2026-02-04
[MOD-13843] A Rust implementation of the numeric index

LukeMathWalker:

A pure Rust implementation of the numeric index (and the associated range tree) from `numeric_index.c`/`numeric_index.h`.

Key differences:
- All nodes are stored in a contiguous slab of memory (arena-style). Parent nodes refer to their children by index rather than by pointer.
- The HLL stored inside the numeric range tree using `wyhash` rather than `fnv` since #8095 proved `wyhash` to be faster _and_ more accurate.

1. MOD-...
2. #...

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **High Risk**
> Adds a brand-new numeric indexing data structure (splitting/balancing, memory accounting, iterator invalidation) and expands workspace dependencies/lints; correctness and performance regressions are plausible despite tests.
> 
> **Overview**
> Introduces a new Rust crate `numeric_range_tree` and wires it into the workspace, providing an arena-allocated numeric range tree with adaptive leaf splitting, AVL-like single-rotation balancing, optional internal-range retention, revision-based iterator invalidation, and memory/statistics tracking.
> 
> Updates `hyperloglog` to include and expose a shared `WyHasher` implementation (used by the range tree’s HLL for cardinality estimation) and adjusts the benchmarks accordingly. Workspace config is extended with new deps (`slab`, `rstest`) and a rustdoc lint allowance, plus corresponding `Cargo.lock` updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 601f53abd04097a4a28ffcc939cb806d29772c1b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/numeric_range_tree/src/arena.rs
+++ b/src/redisearch_rs/numeric_range_tree/src/arena.rs
@@ -66,10 +66,22 @@ impl NodeArena {
     /// This is not to be confused with the current _capacity_
     /// of the arena, i.e. the size of the underlying currently-allocated
     /// slab.
+    #[expect(dead_code, reason = "part of complete arena API, needed for future GC")]
     pub fn len(&self) -> u32 {
         // Safe to truncate because `Self::insert` ensures that the arena
         // never grows beyond `u32::MAX`.
         self.nodes.len() as u32
+    }
+
+    /// Get the capacity of the arena.
+    ///
+    /// This is the maximum number of nodes that can be stored in the arena
+    /// without reallocating.
+    #[expect(dead_code, reason = "part of complete arena API, needed for future GC")]
+    pub fn capacity(&self) -> u32 {
+        // Safe to truncate because `Self::insert` ensures that the arena
+        // never grows beyond `u32::MAX`.
+        self.nodes.capacity() as u32
     }
 
     /// Insert a node into the arena, returning its index.
@@ -118,6 +130,22 @@ impl NodeArena {
         self.nodes
             .iter_mut()
             .map(|(key, node)| (NodeIndex(key as u32), node))
+    }
+
+    /// Get the memory usage of the arena, in bytes.
+    pub fn mem_usage(&self) -> usize {
+        // This is the way `slab::Slab` stores, internally, entries.
+        // The type isn't exposed via its public API, but add it here
+        // as a reference to allow the compiler to give us an accurate
+        // memory usage estimate.
+        // Future versions of `slab` may change this, but for now,
+        // this is the best we can do.
+        #[expect(dead_code)]
+        enum Entry<T> {
+            Vacant(usize),
+            Occupied(T),
+        }
+        self.nodes.capacity() * std::mem::size_of::<Entry<NumericRangeNode>>()
     }
 }
 
--- a/src/redisearch_rs/numeric_range_tree/src/iter.rs
+++ b/src/redisearch_rs/numeric_range_tree/src/iter.rs
@@ -43,7 +43,7 @@ impl<'a> PreOrderDfsIterator<'a> {
 
     /// Create a new iterator starting from the given node index in the tree.
     pub fn from_node(tree: &'a NumericRangeTree, node_idx: NodeIndex) -> Self {
-        let mut stack = Vec::with_capacity(tree.node(node_idx).max_depth() as usize);
+        let mut stack = Vec::with_capacity(tree.node(node_idx).max_depth() as usize + 1);
         stack.push(node_idx);
         Self { tree, stack }
     }
--- a/src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
+++ b/src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
@@ -260,9 +260,7 @@ impl NumericRangeTree {
 
     /// Calculate the total memory usage of the tree, in bytes.
     pub fn mem_usage(&self) -> usize {
-        let base_size = std::mem::size_of::<Self>();
-        let nodes_size = (self.nodes.len() as usize) * std::mem::size_of::<NumericRangeNode>();
-        base_size + self.stats.inverted_indexes_size + nodes_size
+        std::mem::size_of::<Self>() + self.stats.inverted_indexes_size + self.nodes.mem_usage()
     }
 }
 


src/redisearch_rs/numeric_range_tree/src/tree.rs
```diff
@@ -0,0 +1,1074 @@
...
+        allow_duplicates: bool,
+        max_depth_range: usize,
+    ) -> AddResult {
+        // The underlying assumption is that insertions are ordered by doc_id.
+        if doc_id <= self.last_doc_id && !allow_duplicates {
+            // Skip duplicate doc_id
+            return AddResult::default();
+        }
+        self.last_doc_id = doc_id;
+
```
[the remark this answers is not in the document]
LukeMathWalker: This matches the existing behaviour: https://github.com/RediSearch/RediSearch/blob/6fd6a2151c56c95ab03a9389b16d300c462bf11a/src/numeric_index.c#L353

src/redisearch_rs/numeric_range_tree/src/node.rs
```diff
@@ -0,0 +1,372 @@
...
+
+        // Build demoted node (old A) in right_idx's slot.
+        let demoted_depth = nodes[left].max_depth().max(nodes[right_left].max_depth()) + 1;
+        nodes[right_idx] = NumericRangeNode::Internal(InternalNode {
+            value: demoted_value,
+            max_depth: demoted_depth,
+            left,
+            right: right_left,
+            range: demoted_range,
+        });
```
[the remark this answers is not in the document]
LukeMathWalker: This shouldn't be an issue, since the union iterator will deal with it. 
Although, making sure that ranges are non-overlapping is rather easy at this point. In that scenario, we could have a distinct and more optimized iterator to "concat" numeric ranges. Wdyt @GuyAv46?

src/redisearch_rs/numeric_range_tree/src/node.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// Take the range from this node, leaving `None` in its place for internal nodes.
+    ///
+    /// For leaf nodes, this takes the range and replaces the node with a leaf
+    /// containing a default (empty, uncompressed) range — callers that take a
+    /// leaf's range typically replace the entire node immediately after.
+    pub(crate) fn take_range(&mut self) -> Option<NumericRange> {
+        match self {
+            Self::Leaf(leaf) => {
+                // Replace with a default range; callers are expected to replace the whole node.
+                Some(std::mem::replace(&mut leaf.range, NumericRange::new(false)))
```
GuyAv46: This is the actual default, right?
Consider using `std::mem::take`

src/redisearch_rs/numeric_range_tree/src/node.rs
```diff
@@ -0,0 +1,372 @@
...
+    pub const fn is_leaf(&self) -> bool {
+        matches!(self, Self::Leaf(_))
+    }
+
+    /// Get the split value.
+    ///
+    /// Returns `0.0` for leaf nodes.
+    pub const fn split_value(&self) -> f64 {
+        match self {
+            Self::Leaf(_) => 0.0,
```
GuyAv46: Not sure if it matters much, but `0.0` is an arbitrary valid value.
Consider returning NaN, None, panic, or even extract the median instead

LukeMathWalker:
Extracted find-related code to #8302.

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// Calculate the total memory usage of the tree.
+    pub const fn mem_usage(&self) -> usize {
+        let base_size = std::mem::size_of::<Self>();
+        // Our tree is a full binary tree, so #nodes = 2 * #leaves - 1
+        let nodes_count = 2 * self.stats.num_leaves.saturating_sub(1) + 1;
+        let nodes_size = nodes_count * std::mem::size_of::<NumericRangeNode>();
+        let hll_size = self.stats.num_ranges * Hll::size();
+
+        base_size + self.stats.inverted_indexes_size + nodes_size + hll_size
+    }
```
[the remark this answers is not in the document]
LukeMathWalker: This is correct and indeed a downside of the inline array design we discussed @GuyAv46: an `Option<NumericRangeNode>` is at least as big as a `NumericRangeNode`, which in turn must be big enough to accommodate the inline registers.
It might make sense to reintroduce a level of heap pointer indirection, perhaps by using `Option<Box<NumericRange>>` rather than `Option<NumericRange>`.

src/redisearch_rs/numeric_range_tree/src/arena.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /// Insert a node into the arena, returning its index.
+    ///
+    /// # Panics
+    ///
+    /// Debug-asserts that the resulting key fits in `u32`.
+    pub fn insert(&mut self, node: NumericRangeNode) -> NodeIndex {
+        let key = self.nodes.insert(node);
+        debug_assert!(key <= u32::MAX as usize);
```
meiravgri: We can't actually guarantee that Slab::insert will return a usize less than u32::MAX, right? Does the memory savings of u32 over u64 justify the silent truncation risk?

Also, what's the purpose of the debug_assert!? I don't see us testing such a scenario.

src/redisearch_rs/numeric_range_tree/src/node.rs
```diff
@@ -0,0 +1,372 @@
...
+
+        // Extract data from the right child (B).
+        let NumericRangeNode::Internal(right_node) = std::mem::take(&mut nodes[right_idx]) else {
+            unreachable!()
+        };
+        let promoted_value = right_node.value;
+        let promoted_range = right_node.range;
+        let right_left = right_node.left; // y
+        let right_right = right_node.right; // z
+
```
meiravgri: for consistency
```suggestion
        let promoted_left = right_node.left; // y
        let promoted_right = right_node.right; // z

```

src/redisearch_rs/numeric_range_tree/src/node.rs
```diff
@@ -0,0 +1,372 @@
...
+        // Extract data from the current node (A).
+        let NumericRangeNode::Internal(current_node) = std::mem::take(&mut nodes[node_idx]) else {
+            unreachable!()
+        };
+        let demoted_value = current_node.value;
+        let demoted_range = current_node.range;
+        let left = current_node.left; // x
```
meiravgri:
```suggestion
        let demoted_left = current_node.left; // x
```

src/redisearch_rs/numeric_range_tree/src/node.rs
```diff
@@ -0,0 +1,372 @@
...
+
+        // Build demoted node (old A) in right_idx's slot.
+        let demoted_depth = nodes[left].max_depth().max(nodes[right_left].max_depth()) + 1;
+        nodes[right_idx] = NumericRangeNode::Internal(InternalNode {
+            value: demoted_value,
+            max_depth: demoted_depth,
+            left,
+            right: right_left,
+            range: demoted_range,
```
meiravgri: The demoted node's range would be invalid after rotation (covers wrong subtree), right? We rely here on the assumption that nodes with ranges will not be rotated, considering MAXIMUM_DEPTH_IMBALANCE and RSGlobalConfig.numericTreeMaxDepthRange?

src/redisearch_rs/numeric_range_tree/src/node.rs
meiravgri: is 0.0 the right value to return for leaf nodes when 0.0 is a valid split value for internal nodes?

src/redisearch_rs/numeric_range_tree/src/node.rs
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    /// Take the range from this node, leaving `None` in its place for internal nodes.
+    ///
+    /// For leaf nodes, this takes the range and replaces the node with a leaf
+    /// containing a default (empty, uncompressed) range — callers that take a
+    /// leaf's range typically replace the entire node immediately after.
+    pub(crate) fn take_range(&mut self) -> Option<NumericRange> {
+        match self {
+            Self::Leaf(leaf) => {
```
meiravgri, line 356: Why is the behavior different for nodes with valid ranges between internal and leaf?

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct AddResult {
+    /// The change in the tree's inverted index memory usage, in bytes.
+    /// Positive values indicate growth, negative values indicate shrinkage.
+    /// This tracks only inverted index memory, not node/range struct overhead.
+    pub size_delta: i64,
+    /// The net change in the number of records (document, value entries).
+    /// When splitting, this counts re-added entries to child ranges.
+    /// When trimming, this is negative for removed entries.
+    pub num_records: i32,
```
meiravgri:
```suggestion
    pub num_records_delta: i32,
```

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// Whether the tree structure changed (splits or rotations occurred).
+    /// When true, the tree's `revision_id` should be incremented to
+    /// invalidate any concurrent iterators.
+    pub changed: bool,
+    /// The net change in the number of ranges (nodes with inverted indexes).
+    /// Splitting a leaf adds 2 new ranges. Trimming removes ranges.
```
meiravgri: Is it 2 even if internal nodes are disabled?

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
```diff
@@ -0,0 +1,418 @@
...
+        &mut self,
+        doc_id: t_docId,
+        value: f64,
+        is_multivalued: bool,
+        max_depth_range: usize,
+    ) -> AddResult {
+        // The underlying assumption is that insertions are ordered by doc_id.
+        // Skip if out of order or duplicate on a single-valued field
+        if doc_id < self.last_doc_id || (doc_id == self.last_doc_id && !is_multivalued) {
+            return AddResult::default();
```
meiravgri, line 100: This case is essentially a bug (out-of-order or duplicate insert). Should we log a warning instead of silently skipping?

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// Revision ID, incremented when the tree structure changes (splits/rotations).
+    ///
+    /// When `revision_id != 0`, it indicates the tree nodes have changed and
+    /// concurrent iteration may not be safe. Iterators like
+    /// [`NumericRangeTreeIterator`](crate::NumericRangeTreeIterator) should check this value and abort if it
+    /// changes during iteration, as the tree structure they were traversing
+    /// may no longer be valid.
+    revision_id: u32,
+    /// Unique identifier for this tree instance.
+    unique_id: TreeUniqueId,
```
meiravgri, line 125: is it in use (i dint think it was in the c code)?

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    /// Create a new empty numeric range tree.
+    ///
+    /// If `compress_floats` is true, the tree will use float compression.
+    /// Check out [`NumericFloatCompression`][`inverted_index::numeric::NumericFloatCompression`] for more information.
+    pub fn new(compress_floats: bool) -> Self {
+        let mut nodes = NodeArena::new();
+        let root_node = NumericRangeNode::leaf(compress_floats);
+        let inverted_indexes_size = root_node.range().map(|r| r.memory_usage()).unwrap_or(0);
+        let root = nodes.insert(root_node);
```
meiravgri:
```suggestion
        let root_index = nodes.insert(root_node);
```

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub fn iter(&self) -> crate::PreOrderDfsIterator<'_> {
+        crate::PreOrderDfsIterator::new(self)
+    }
+
+    /// Calculate the total memory usage of the tree.
+    pub const fn mem_usage(&self) -> usize {
+        let base_size = std::mem::size_of::<Self>();
+        // Our tree is a full binary tree, so #nodes = 2 * #leaves - 1
+        let nodes_count = 2 * self.stats.num_leaves.saturating_sub(1) + 1;
+        let nodes_size = nodes_count * std::mem::size_of::<NumericRangeNode>();
```
meiravgri: why do we need saturating_sub? we always have at least 1 leaf, no? 
And when we have 1 leaf than number of nodes is 1 or 0?

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+/// The tree organizes documents by their numeric field values into a balanced
+/// binary tree of ranges. Each leaf node contains a range of values, and
+/// internal nodes may optionally retain their ranges for query efficiency.
+///
+/// # Arena Storage
+///
+/// All nodes are stored in a [`NodeArena`]. Children are referenced by
+/// [`NodeIndex`] instead of `Box<NumericRangeNode>`. This provides better
+/// cache locality, eliminates per-node heap allocation overhead, and makes
+/// rotations cheaper (index swaps instead of alloc/dealloc).
```
GuyAv46: > makes rotations cheaper (index swaps instead of alloc/dealloc)

The C implementation doesn't require [de]allocations, so this is not a real improvement. The rest is true :)

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+        let base_size = std::mem::size_of::<Self>();
+        // Our tree is a full binary tree, so #nodes = 2 * #leaves - 1
+        let nodes_count = 2 * self.stats.num_leaves.saturating_sub(1) + 1;
+        let nodes_size = nodes_count * std::mem::size_of::<NumericRangeNode>();
+        let hll_size = self.stats.num_ranges * Hll::size();
+
+        base_size + self.stats.inverted_indexes_size + nodes_size + hll_size
+    }
+}
+
```
meiravgri: is HLL lives in NumericRange and NumericRange in NumericRangeNode?

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// Returns an iterator over all nodes in the tree (depth-first traversal).
+    pub fn iter(&self) -> crate::PreOrderDfsIterator<'_> {
+        crate::PreOrderDfsIterator::new(self)
+    }
+
+    /// Calculate the total memory usage of the tree.
+    pub const fn mem_usage(&self) -> usize {
+        let base_size = std::mem::size_of::<Self>();
+        // Our tree is a full binary tree, so #nodes = 2 * #leaves - 1
+        let nodes_count = 2 * self.stats.num_leaves.saturating_sub(1) + 1;
```
GuyAv46: Why do we need `2 * self.stats.num_leaves.saturating_sub(1) + 1`? `num_leaves` should always be at least 1, no?

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
```diff
@@ -0,0 +1,418 @@
...
+    ///
+    /// If `is_multivalued` is true, the same document ID can be provided multiple times
+    /// in a row.
+    /// Returns information about what changed during the add operation.
+    pub fn add(
+        &mut self,
+        doc_id: t_docId,
+        value: f64,
+        is_multivalued: bool,
+        max_depth_range: usize,
```
meiravgri, line 63: What happens if  max_depth_range changes between insertions? How do we handle existing nodes created under a different max_depth_range value?

src/redisearch_rs/numeric_range_tree/src/range.rs
```diff
@@ -0,0 +1,189 @@
...
+    entries: NumericIndex,
+}
+
+/// Update the cardinality estimate of an HLL with a new float value.
+///
+/// This is the single source of truth for how numeric values are
+/// hashed into HyperLogLog registers. We hash the raw bytes (bit
+/// representation) rather than the numeric value — see
+/// [`NumericRange::update_cardinality`] for rationale.
+fn update_cardinality(hll: &mut Hll, value: f64) {
```
GuyAv46, line 67: Why have this helper, and not simply define it at the `NumericRange` implementation of `update_cardinality`?

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
```diff
@@ -0,0 +1,418 @@
...
+            let mut entries: Vec<(ffi::t_docId, f64)> = Vec::new();
+            let mut reader = range.reader();
+            let mut result = inverted_index::RSIndexResult::numeric(0.0);
+            while reader.next_record(&mut result).unwrap_or(false) {
+                // SAFETY: We know the result contains numeric data
+                let entry_value = unsafe { result.as_numeric_unchecked() };
+                entries.push((result.doc_id, entry_value));
+            }
+
+            (split, entries)
```
meiravgri: do we copy the entire inverted index of the range here?

src/redisearch_rs/numeric_range_tree/src/iter.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl<'a> PreOrderDfsIterator<'a> {
+    /// Create a new iterator starting from the root of the given tree.
+    pub fn new(tree: &'a NumericRangeTree) -> Self {
+        Self::from_node(tree, tree.root_index())
+    }
+
+    /// Create a new iterator starting from the given node index in the tree.
+    pub fn from_node(tree: &'a NumericRangeTree, node_idx: NodeIndex) -> Self {
+        let mut stack = Vec::with_capacity(4);
```
GuyAv46: A small optimization we can make, since we know the max depth of the node, is to use it as the capacity (x2+1, since each node in the path will insert its two children).
Currently, the 2 places we use `NumericRangeTreeIterator_New` read the iterator until depletion, so we will use the full capacity while scanning

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
```diff
@@ -0,0 +1,418 @@
...
+                rv.size_delta += size as i64;
+            }
+            rv.num_records += 1;
+        }
+
+        // Take the existing range from the leaf and convert to an internal node.
+        let old_range = nodes[node_idx].take_range();
+        let new_node =
+            NumericRangeNode::internal_indexed(split, left_idx, right_idx, old_range, nodes);
+        nodes[node_idx] = new_node;
```
meiravgri: dont we need to pass the range only if internal ranges are enabled?

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
```diff
@@ -0,0 +1,418 @@
...
+
+        rv.changed = true;
+        rv.num_ranges_delta += 2;
```
meiravgri, line 341: rv.num_ranges_delta == 1 or 2, depends on internal ranges configuration

src/redisearch_rs/numeric_range_tree/src/node.rs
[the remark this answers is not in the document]
GuyAv46: I think the AI is correct, and that we should set the range to `None` here.
In general, we should check if the ranges in the parent are worth the extra memory it consumes, compared to a union over the disjoint children it has. This config may be redundant, and then we can remove the optional range from the inner node.

In the current implementation, I think we don't get to this issue because:
1. The range config is in `[0, 1, 2]` (default is `0`)
2. We perform the rotation if and only if the children's max-depth difference is larger than `NR_MAX_DEPTH_BALANCE` (`2`), which means the parent (with the demoted range) has a max-depth of at least `4`, and the promoted node has a max-depth of at least `3`, so they both don't have a range (this minimal case happens when one child is a leaf - depth 0, and the other has a max-depth of `3`). 

I think that in the Rust implementation, it makes more sense to drop the potential ranges in the rotation logic for the reasons given by the LLM. This is more generic and logically correct, and it won't introduce behavior changes compared to the C implementation.

src/redisearch_rs/numeric_range_tree/src/tree/invariants.rs
GuyAv46, line 1: Big like

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
```diff
@@ -0,0 +1,418 @@
...
+    ) -> AddResult {
+        // The underlying assumption is that insertions are ordered by doc_id.
+        // Skip if out of order or duplicate on a single-valued field
+        if doc_id < self.last_doc_id || (doc_id == self.last_doc_id && !is_multivalued) {
+            return AddResult::default();
+        }
+        self.last_doc_id = doc_id;
+
+        let mut rv = AddResult::default();
+        Self::node_add(
```
GuyAv46, line 105: Consider making `node_add` return an `AddResult`, unless it has performance panalties

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
```diff
@@ -0,0 +1,418 @@
...
+            let split = if split == range.min_val() {
+                // Make sure the split is not the same as the min value
+                // Use next representable f64 greater than split
+                split.next_up()
+            } else {
+                split
+            };
+
+            // Collect all entries from the range
+            let mut entries: Vec<(ffi::t_docId, f64)> = Vec::new();
```
GuyAv46: Why do we collect the results, and not use the reader in the "Redistribute entries to children" `for` loop?

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
```diff
@@ -0,0 +1,418 @@
...
+        compress_floats: bool,
+    ) {
+        // First compute the split point and collect entries before mutating the node
+        let (split, entries) = {
+            let range = nodes[node_idx]
+                .range()
+                .expect("node to split must have a range");
+
+            // Compute the median to use as split point
+            let split = Self::compute_median(range);
```
GuyAv46: `compute_median` collects all the values internally. Can we reuse this vector instead of collecting it again later?

GuyAv46 (commented):
Awesome! First review without the tests.
Deletion (and repair) logic will come in a later PR, I assume? Interested in the slab compaction logic

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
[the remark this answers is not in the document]
GuyAv46: Right...
Also, see [my comment](https://github.com/RediSearch/RediSearch/pull/8276#discussion_r2779395690): 

> In general, we should check if the ranges in the parent are worth the extra memory it consumes, compared to a union over the disjoint children it has. This config may be redundant, and then we can remove the optional range from the inner node.

src/redisearch_rs/numeric_range_tree/src/arena.rs
LukeMathWalker: Debug asserts can be enabled in "release" builds too, they aren't necessarily tied to testing.
E.g. it's possible to produce a release build with debug assertions enabled to re-run particular scenarios and see what triggers.
In this specific case, it's better to make it a production assert, since the overhead is going to be negligible.

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
LukeMathWalker: You're right, it's either 1 or 2 depending on retention of internal ranges.

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
LukeMathWalker: It's used in fork GC at the moment: https://github.com/RediSearch/RediSearch/blob/bdd18730b3da74c5fdfef4f457c59b81607417b6/src/fork_gc.c#L671

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
LukeMathWalker: Excess of zeal. Since we're storing things in an arena, we already know the exact number of nodes, so this isn't needed at all.

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
LukeMathWalker: Yes, since HLL has inline registers, it is already accounted for. This was double-counting.

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
LukeMathWalker: If a later invocation uses a smaller `max_depth_range`, `add` won't eagerly prune ranges that are higher than the threshold; pruning will only occur opportunistically, when the tree structure changes (i.e. a node is split) and it will only affect nodes that are an ancestor of the node that stored the new value.

I'll add it to the docs.

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
LukeMathWalker: Yes, we need to collect all its entries into a buffer.

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
LukeMathWalker: At this stage, we keep it unconditionally, then it's trimmed in `node_add` based on the `max_depth_range` that was specified. I could also push the check further down, bringing `max_depth_range` into this context.

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
LukeMathWalker: At this point it's always 2, since we retain the range unconditionally here. It gets decremented by 1 in `node_add` when `remove_range` is invoked, if we end up dropping the internal range before returning.

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
LukeMathWalker: Added a warning-level log 👍🏻

src/redisearch_rs/numeric_range_tree/src/node.rs
LukeMathWalker: We can leave a leaf without a range, since it's not an `Option` in its definition. So we are forced to put something in its place.

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
LukeMathWalker: True, but we don't need this either, we know the exact node count thanks to the arena.

src/redisearch_rs/numeric_range_tree/src/range.rs
LukeMathWalker: Later on, when have the GC code (#8293) we'll need to compute hashes of values outside the context of a `NumericRange`, and since `f64` doesn't implement `Hash`, I used this helper as the source of truth for how we are hashing floats.
A different approach could be to make `HyperLogLog` strongly typed—generic over the type of values it's counting, and then create a thin wrapper around `f64` which encodes the desired hashing behaviour.

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
LukeMathWalker: To avoid an annoying borrow-checker issue, but it can be worked around.

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
LukeMathWalker: Unfortunately not, because we need insertions into the new child ranges to be ordered by doc id, and `select_nth_unstable` mutates the buffer in place, thus removing the natural doc id ordering we get from the reader.

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
LukeMathWalker: I reworded it around pruning, which is when we see the difference (i.e. one realloc vs multiple deallocs).

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
LukeMathWalker: It won't work unless we make `node_add` non-recursive. It needs to thread `rv` via function parameters so that each "layer" of the chain can update it in place.

src/redisearch_rs/numeric_range_tree/src/node.rs
LukeMathWalker: Option is the right abstraction here, yes.

LukeMathWalker: Option is the right abstraction here, yes.

src/redisearch_rs/numeric_range_tree/src/node.rs
LukeMathWalker: It would be too broad, yes. In the last commit, I've removed it alongside the range of the promoted node.

LukeMathWalker:
> Awesome! First review without the tests. Deletion (and repair) logic will come in a later PR, I assume? Interested in the slab compaction logic

Yes, they are in #8293.

I _think_ I have addressed all feedback in the latest commit. Please have a second look! @meiravgri @GuyAv46

src/redisearch_rs/numeric_range_tree/src/iter.rs
GuyAv46: Not sure if my math was correct 🙈

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
GuyAv46: I see. In practice, the child sets it inially and then it gets updated on the way up. Could be an idea for the future

src/redisearch_rs/numeric_range_tree/src/tree/insert.rs
GuyAv46: Makes sense, I also didn't have this optimization in the C heap algorithm for the same reason

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
```diff
@@ -259,15 +258,11 @@ impl NumericRangeTree {
...
+    /// Calculate the total memory usage of the tree, in bytes.
+    pub fn mem_usage(&self) -> usize {
         let base_size = std::mem::size_of::<Self>();
-        // Our tree is a full binary tree, so #nodes = 2 * #leaves - 1
-        let nodes_count = 2 * self.stats.num_leaves.saturating_sub(1) + 1;
-        let nodes_size = nodes_count * std::mem::size_of::<NumericRangeNode>();
-        let hll_size = self.stats.num_ranges * Hll::size();
-
-        base_size + self.stats.inverted_indexes_size + nodes_size + hll_size
+        let nodes_size = (self.nodes.len() as usize) * std::mem::size_of::<NumericRangeNode>();
```
GuyAv46: Maybe we should return the actual arena memory consumption, even for the holes. This is a more accurate `mem_usage`, won't you agree?

src/redisearch_rs/numeric_range_tree/src/iter.rs
```diff
@@ -43,7 +43,7 @@ impl<'a> PreOrderDfsIterator<'a> {
 
     /// Create a new iterator starting from the given node index in the tree.
     pub fn from_node(tree: &'a NumericRangeTree, node_idx: NodeIndex) -> Self {
-        let mut stack = Vec::with_capacity(4);
+        let mut stack = Vec::with_capacity(tree.node(node_idx).max_depth() as usize);
```
GuyAv46: I think it should be `tree.node(node_idx).max_depth() as usize + 1`

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
LukeMathWalker: Yes, we should track unused capacity too.

src/redisearch_rs/numeric_range_tree/src/tree/mod.rs
meiravgri: The capacity reflects the actual node count? isn't like vector where it can allocate more than actually used?

src/redisearch_rs/numeric_range_tree/src/arena.rs
meiravgri: On one hand i dont think we need to crash the server in this case, because the user didnt do anything wrong
on the other hand, if handling it gracefully will add complexity to the code it doesn't worth it...
your call

PR #8280, opened 2026-02-05
Update documentation

daxhuiberts:
Update and centralize a lot of technical documentation currently spread out. Using the new `./build.sh` approach instead of the old 'make' approach. I kept all existing information in here, but more organized.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Primarily documentation restructuring plus a small tweak to the Markdown link checker’s anchor validation logic; low risk with limited runtime impact outside CI/tooling.
> 
> **Overview**
> **Developer documentation has been reorganized and expanded** to centralize setup/build/test guidance around `./build.sh` (including submodules, dependency install scripts, `cargo-nextest`, `uv`-based Python env setup, and added sections for debugging, benchmarking, and supported platforms).
> 
> **Rust contributor docs (`src/redisearch_rs/CONTRIBUTING.md`) are expanded** with general FFI/safety guidelines and consolidated guidance on tests and logging.
> 
> The Markdown link checker now **skips HTML anchor verification for `github.com` links** (to avoid false failures on JS-rendered anchors), and its test fixture updates the “bad anchor” case accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2a12865a36c1205d05300d70e328485f2accfe5c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/scripts/check_links.py
+++ b/scripts/check_links.py
@@ -204,7 +204,8 @@ class LinkChecker:
             response.raise_for_status()
 
             # If there's an anchor, verify it exists in the HTML
-            # GitHub line-number anchors (e.g., #L207, #L207-L226) are rendered
+            # Github doesn't render generated markdown anchors (e.g. readme) and
+            # line-number anchors (e.g., #L207, #L207-L226). Those are rendered
             # client-side via JavaScript and won't appear in static HTML.
             if anchor and parsed.netloc != "github.com":
                 content_type = response.headers.get('content-type', '').lower()


daxhuiberts:
@memark This is my cleanup attempt (with the help of Claude, but fully reviewed by myself) for the technical documentation including all the info from the Notion onboarding notes document. Use the rendered view to properly review. Let me know if you think something is still missing or not clear.

There is an `onboarding.md` with the info from the Notion onboarding notes document which I couldn't place anywhere else. Not meant to be in this repo, but just reviewing purposes, will remove before putting up for final review. Perhaps you have an idea where to place this info?

developer.md
```diff
@@ -1,214 +1,299 @@
...
-* Python behavioral tests (enabled by [RLTest](https://github.com/RedisLabsModules/RLTest)), located in tests/pytests, run by `make pytest`.
 
-### Test prerequisites
-To run the python behavioral tests you need to install the following dependencies:
-* python test requirements listed in [pyproject.toml](tests/pytests/pyproject.toml)
-* If you want to execute RediSearch+RedisJSON behavioral tests you need to install [rust](https://www.rust-lang.org/tools/install), which is required to compile the RedisJSON module 
+## Building the Project
 
-if you don't want to install those manually, you can use the CI installation scripts to install the dependencies.
+RediSearch has two main CLIs at the moment the (old, legacy) `MAKEFILE` and the new preferred `build.sh` file. You will sometimes see the redis team share commands using old CLI for particular things that the new CLI might not be capable of yet. We will however only use the new CLI in this guide.
```
memark: Let's think about what we put here and how we formulate it. It could also use some more grammar clarity.

developer.md
```diff
@@ -1,214 +1,299 @@
...
-- `RUST_LOG_STYLE=auto` automatically detect when to enable or disable color codes (default)
+```bash
+./build.sh RUN_PYTEST REJSON=0
+```
 
+Or specify a path to an existing RedisJSON module:
 
-### Rust Tests
+```bash
+./build.sh RUN_PYTEST REJSON_PATH=/path/to/redisjson.so
```
memark, line 194: I've never seen this before. Does it remove the need to clone the RedisJSON repo?

developer.md
```diff
@@ -1,214 +1,299 @@
...
+4. Run the integration tests:
+   
+   ```bash
+   ./build.sh RUN_PYTEST EXT=1 TEST=<name of the test>
+   ```
+
+## Benchmarking RediSearch
+
+```bash
+PATH=$PATH:~/GitHub/RediSearch/bench_tools redisbench-admin run-local \
```
memark: What is this path?
First of all, maybe not assume ~/GitHub
Second, I have no `bench_tools` or `redisbench-admin` anywhere on my machine. Where are they coming from?

developer.md
```diff
@@ -1,214 +1,299 @@
...
+    --module_path ~/GitHub/RediSearch/bin/macos-aarch64-debug/search-community/redisearch.so \
+    --required-module search \
+    --allowed-setups oss-standalone \
+    --allowed-envs oss-standalone \
+    --test search-ftsb-10K-enwiki_abstract-hashes-term-wildcard.yml
 ```
+
+Use `--skip-redis-spin True` to skip spinning up a Redis instance.
+
+## Supported Platforms
```
memark, line 302: Is it meaningful to have this here? Maybe just link to the CI script instead of duplicating?

onboarding.md
```diff
@@ -0,0 +1,70 @@
+# Onboarding Notes
```
memark: I'm wondering if these should be in the public repo? Or in our Notion? (Or possibly in Redis Confluence?)

developer.md
daxhuiberts: My approach so far is to retain the existing info from the previous version (where this is coming from) and improve where we have better info.

Regarding this I haven't tried doing anything with a custom redisjson, but I didn't want to throw it away.

developer.md
daxhuiberts: I have some more knowlegde now on how to run the benchmarks and install the required dependencies. I'll update this part of the readme.

developer.md
daxhuiberts: Yeah it's a bit too much detailed information which I kept around, though I don't like it here. I think referring to the CI scripts for the platform matrix should be fine.

onboarding.md
daxhuiberts: Yep, this is only a temporary file with all the information I couldn't place in onboarding.md or contributing.md and not something I want in this repo. But I don't know the proper place to keep this info either. Ideally this should be Redis's Confluence I think.

memark:
- [ ] put something about the secret https://cla-assistant.io/ website in the docs.

KevinBongart:
```
Running Rust tests...
error: no such command: `nextest`
```

Where does `nextest` fit in the setup?

developer.md
```diff
@@ -1,214 +1,299 @@
...
-Currently address sanitizer is supported. To run the tests with address sanitizer you can use the following command:
+For Rust coverage tests, install the nightly toolchain first:
 
 ```bash
-make build test SAN=address
+rustup toolchain install nightly \
+    --allow-downgrade \
+    --component llvm-tools-preview \
+    --component miri
+
```
memark, line 131: Let's add a section about installing. `nextest` here. FYI @KevinBongart

memark: Or hmmm, maybe better to just automate it in the install script? Can't really decide...

daxhuiberts: There is an extra install script in `./install/test_deps/common_installation.sh` to install those dependencies which is used in github CI. But that one again should be called from the project root instead of from the `.install` directory (not very consistent) and I don't fully grasp its full intent yet.

Because the install scripts (both `.install/install_script.sh` and `./install/test_deps/common_installation.sh`) are also very important for it's use in CI, I don't want to touch them yet before understanding its behaviour a bit better. So I suggest to add instructions to the `developer.md` at this moment.

daxhuiberts:
@KevinBongart @memark Got some improvements in `developer.md`, you can see the latest rendered here: https://github.com/RediSearch/RediSearch/blob/aa6cb71b4e55551b1bf8f34153dab15a8da05f92/developer.md

developer.md
```diff
@@ -1,214 +1,331 @@
...
-## Dev container
-We have provided a dev container based on `ubuntu:latest` docker image with all the dependencies (Redis, build systems, rust, Python virtual environment) installed. Just clone the repo and open it in the dev container. It will take a few minutes for the first time to download and install the dependencies.
-If you would like to build it on your machine without a docker, proceed to the next sections.
-## Installing prerequisites
+If you already cloned without `--recursive`, initialize the submodules:
 
-### Build dependencies
-To build and test RediSearch you need to install several packages, depending on the underlying OS. The following OSes are supported and tested in our CI:
+```sh
+git submodule update --init --recursive
```
ThierryBerger, line 15: Should we rather advocate for using `make fetch` ?

developer.md
```diff
@@ -1,214 +1,331 @@
...
-```bash
-make pytest REJSON=0
+### Installing Redis
+
+RediSearch requires `redis-server` in your PATH. We recommend building Redis from source since RediSearch `master` often requires unreleased features:
+
+```sh
+git clone https://github.com/redis/redis.git
+cd redis
+make
```
ThierryBerger: such command may be more accurate:

`BUILD_TLS=yes BUILD_WITH_MODULES=yes INSTALL_RUST_TOOLCHAIN=yes DISABLE_WERRORS=yes make -j 15`

Rather than maintaining a guide on how to install redis here, I suggest linking to https://github.com/redis/redis?tab=readme-ov-file#build-redis-from-source directly ?

developer.md
daxhuiberts: Where did you get that from? I've never heard about or used `make fetch` before. And considering the `make` approach is deprecated I don't think so.

developer.md
daxhuiberts: Let's keep it for now. I don't want to remove information at this moment. Only update the info and get this merged.

onboarding.md
daxhuiberts: I'll remove this from this PR.

daxhuiberts:
@memark IMO this is good enough to get merged in. There was one last thing you wanted to add regarding cla assistant? If so, could you add that? I'd like to get it merged.

src/redisearch_rs/CONTRIBUTING.md
```diff
@@ -15,6 +24,88 @@ Dependency versions should be updated:
 - Whenever we need newer features or bug fixes released in a newer version.
 - Once in a while, via [`cargo upgrade`](https://crates.io/crates/cargo-edit), if neither of the two things above have happened.
 
+## Tests
+
+Rust Unit tests use the regular Rust test harness and test runner. All regular Rust testing practices apply with a few specifics:
+
+- Use [`proptest`](https://docs.rs/proptest/latest/proptest/) whenever possible, this lets us test inputs in-depth instead of superficially.
+- Prefer integration tests in `tests/` over in-crate unit tests.
```
JoanFM: why do we exactly prefer these?

developer.md
```diff
@@ -1,214 +1,325 @@
...
+### Alternative: Nix
 
-### Running the tests
-To run the tests, you need to execute the following commands:
+If you prefer Nix, skip the install scripts and start a development environment:
 
-```bash
-make test
+```sh
+nix develop github:chesedo/redisearch-nix-env
```
JoanFM: I do not think we want something from @chesedo spread here

developer.md
```diff
@@ -1,214 +1,325 @@
...
+   
+   ```sh
+   ./build.sh RUN_PYTEST EXT=1 TEST=<name of the test>
+   ```
+
+## Benchmarking RediSearch
+
+```sh
+PATH=$PATH:~/GitHub/RediSearch/bench_tools redisbench-admin run-local \
+    --module_path ~/GitHub/RediSearch/bin/macos-aarch64-debug/search-community/redisearch.so \
```
JoanFM: let's make it clear this should be ur own personal path

src/redisearch_rs/CONTRIBUTING.md
daxhuiberts: Integration tests are only able to use the public interface of a crate whereas in-crate unit tests can access internals. By default you really want to only test the public interface. Only in rare occasion you might want to test internals.

daxhuiberts: I'll add this extra info to that document as that should actually be part of it.

developer.md
chesedo: Haha, yes I don't want to maintain this for all of time :sweat_smile:. We can keep it on our internal notes for anyone else who uses nix

developer.md
daxhuiberts: I've rewritten the entire benchmark part. Adding dependencies and removing the references to local paths.

I'm not at all familiar with `bench_tool` and I couldn't find it on the redisearch github either. So I removed it, as without it it seems to work just fine.

daxhuiberts:
@JoanFM Please check again. I also patched the link checker to skip anchor checking on github as github doesn't provide anchors in their generared markdowns (only works client side).

developer.md
ThierryBerger: @gdesmott pointed me to it ; it basically does this command.

src/redisearch_rs/CONTRIBUTING.md
ThierryBerger, line 185: As onboarding.md [was removed](https://github.com/RediSearch/RediSearch/pull/8280/changes/f29db8ad794192dc4f92cad7f36b88322aac3150#diff-63108105ad4245beea4445f0c06d89f4cc5c701b12ce3501ec7ba17cd55ad35aL23-L70) from the pr, I suggest we still keep some technical informations (additional informations from that onboarding file) which make sense as contributors:

(probably reword references to Meir as redis ?)

```suggestion

## Additional info

### Migration Plan

When porting modules from C to Rust, prioritize modules with many dependents but few dependencies. Two types of modules stand out:

1. **Data structures** - Self-contained, easy to test, straightforward usage patterns
2. **Utility classes**

Data structures are the preferred starting point because:
- They won't necessarily surface or reduce memory safety bugs, but they won't introduce new ones either
- They're used in pretty much every code path, so we avoid having to write C FFI layers over existing C data structures later

### Redis Global Lock

Redis is fundamentally single-threaded.

All operations that touch its data structures must happen while holding the Redis global lock, which you may think of as a close sibling of Python’s Global Interpreter Lock.

- 💡 We should use `pyo3`'s modelling as a reference when reviewing the way the lock is modelled in `redis_module_rs`. It’s currently relatively crude.
- ⚠️ `redis_module_rs` is actively used in `RedisJSON`, so any breaking change will require a corresponding PR to upgrade the codebase to the newer version. Meir prefers to stick to backwards-compatible changes when possible.

### Active defragmentation

It’s a process Redis uses to optimise the packing of its existing memory allocations, thus reducing overall memory usage.

Key API calls: https://redis.io/docs/latest/develop/reference/modules/modules-api-ref/#section-defrag-api

For modules, it plays out as follows:

- Redis kicks off active defrag
- Redis tells the module: do you have something to defrag?
- The module may try to defrag its own data structures (or not)
- The process may be paused multiple times to avoid blocking

**After the data buffer behind a pointer has been defragmented, the old pointer MUST NOT be used anymore. It’s invalid.**

Active defragmentation holds the Redis global lock.

- ⚠️ RediSearch has a background thread pool running concurrently with whatever holds the Redis global lock then.
We’ll have to acquire secondary locks over those indexes to defrag its data structures and ensure that nothing else is using them, including from those background threads.
- ‼️ RediSearch doesn’t currently implement active defragmentation. It’s implemented, today, in RedisJSON, which we can use as a case study, although it’s a muuuuch simpler setup.
- Defragmenting a data structure requires accessing its raw heap allocations. This may or may not be possible for data structures coming from existing Rust crates. When that requirement is not satisfied, we will have to fork.
E.g. we can decompose `Vec` into raw parts, defrag, and then recompose. We can’t do it for `HashSet` or `HashMap`.
- 🗺️ I managed to get Meir to agree on a few key points:
    - We should keep active defrag in mind but not try to port code *and* introduce active defrag simultaneously.
    - We shouldn’t pay the cost of forking data structures up front. We can fork later when implementing active defrag.
    - Implementing active defrag in a mixed C/Rust codebase will be a nightmare because tracking whether a pointer is actively being used will be tricky. We should wait for the Rust migration to be well underway before starting this endeavor.
```

daxhuiberts: Good points! All the changes in this PR are originally from our internal notion docs. The stuff remaining in onboarding.md (I used it in that PR as the source of info in my working draft and extracting it to developer.md and contributing.md) was the only stuff I didn't find a proper place for directly in the repo and think it's better for now to keep it in our notion docs.

daxhuiberts:
@JoanFM Can you check again?

JoanFM (approved):

PR #8283, opened 2026-02-05
MOD-13903: Suppress info when there are zero indices

oshadmi:
Avoid emitting all Search metrics in info command when there are zero indices.
Some of the info is emitted (runtime configs) so it can be understood why some metrics are supressed.

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Changes the shape/content of `INFO MODULES` output based on runtime state, which may affect monitoring/automation that expects metrics to exist even with zero indexes. Core indexing/query logic is unchanged and behavior is gated by a new config with test coverage.
> 
> **Overview**
> `INFO MODULES` output is now **conditionally minimized** when the server has *zero* RediSearch indexes: by default it returns only `version`, a zeroed `indexes` section, and `runtime_configurations`, skipping the other metrics sections unless generating a crash report.
> 
> Introduces a new `CONFIG` boolean `search-_info-on-zero-indexes` (default `OFF`) to control this behavior, exposes it in `INFO MODULES` runtime configuration, and updates index counting to use `Indexes_Count()`.
> 
> Adds targeted pytests and helpers to validate the new config and adjusts existing tests that read `INFO MODULES` before creating indexes to explicitly enable full output.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 23249e3d924fee82341211b048200ee75ce1ca44. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_config.py
+++ b/tests/pytests/test_config.py
@@ -1362,7 +1362,6 @@ booleanConfigs = [
     ('search-_free-resource-on-thread', '_FREE_RESOURCE_ON_THREAD', 'yes', False, False),
     ('search-_numeric-compress', '_NUMERIC_COMPRESS', 'no', False, False),
     ('search-_print-profile-clock', '_PRINT_PROFILE_CLOCK', 'yes', False, False),
-    ('search-_simulate-in-flex', '_SIMULATE_IN_FLEX', 'no', False, False),
     ('search-no-gc', 'NOGC', 'no', True, True),
     ('search-no-mem-pools', 'NO_MEM_POOLS', 'no', True, True),
     ('search-partial-indexed-docs', 'PARTIAL_INDEXED_DOCS', 'no', True, False),


src/info/info_redis/info_redis.c
```diff
@@ -56,6 +58,17 @@ void RS_moduleInfoFunc(RedisModuleInfoCtx *ctx, int for_crash_report) {
     RedisModule_InfoAddFieldCString(ctx, "redis_enterprise_version", ver);
   }
 
+  // On normal INFO runs, optionally suppress RediSearch metrics when there are no indexes.
+  // (We never suppress crash-report info.)
+  if (!for_crash_report && !RSGlobalConfig.infoEmitOnZeroIndexes && Indexes_Count() == 0) {
```
kei-nan, line 63: wait, do we still want to output info on crash if we don't have any indexes?
maybe we need to remove the for_crash_report boolean?

src/config.c
```diff
@@ -2245,5 +2234,14 @@ int RegisterModuleConfig_Local(RedisModuleCtx *ctx) {
     )
   )
 
+  RM_TRY(
+    RedisModule_RegisterBoolConfig(
```
kei-nan, line 2238: We sure we want a config param to control this?
Why shouldn't omitting the info when we don't have indexes always be the behaviour?

tests/pytests/test_info_modules.py
```diff
@@ -568,6 +653,9 @@ def test_redis_info_modules_vecsim():
 
 @skip(cluster=True)
 def test_indexes_logically_deleted_docs(env):
+  # This test reads INFO MODULES metrics before creating any index. Ensure INFO MODULES is in full mode.
+  allShards_set_info_on_zero_indexes(env, True)
```
kei-nan, line 654: maybe the default value should be based if we are on enterprise or not?
seems a shame to enable this manually in search test

src/config.c
oshadmi: The default is to omit, but this is a way to override it if needed.

src/info/info_redis/info_redis.c
oshadmi: We still might be somehow related to a crash, so I don't think it will hurt to add our info to the crash report (e.g., there was a single index, used up memory, then index was dropped, but memory is still not cleaned up, etc.)

tests/pytests/test_info_modules.py
oshadmi: I think we still don't want a different default for OSS or for Enterprise - by default it is suppressed.
Why is it a shame to disable suppression when there are no indices and we want to check something in the INFO response?

tests/pytests/test_debug_commands.py
```diff
@@ -1002,6 +955,9 @@ def testIndexObfuscatedInfo(env: Env):
 def testPauseOnOOM(env: Env):
     # Change the memory limit to 80% so it can be tested without colliding with redis memory limit
     env.expect('FT.CONFIG', 'SET', '_BG_INDEX_MEM_PCT_THR', '80').ok()
+    # This test reads INFO MODULES metrics before creating any index. Ensure INFO MODULES is in full mode.
+    env.assertEqual(allShards_set_info_on_zero_indexes(env, True), ['OK'],
```
kei-nan: This line feels very off
why are you asserting here
are you testing the config in an unrelated test?
maybe the function inside should assert if it failed?

tests/pytests/test_dialect.py
```diff
@@ -303,6 +303,9 @@ def test_dialect_aggregate(env):
     env.assertEqual(res[0], 2)
 
 def check_info_module_results(env, module_expect):
+  # This helper may be called after a flush (zero indexes), where INFO MODULES can be in minimal
+  # suppression mode by default. Ensure dialect statistics are emitted.
+  allShards_set_info_on_zero_indexes(env, True)
```
kei-nan: maybe don't call this every time `check_info_module_results` is called?
you can call this just once when the test starts

tests/pytests/test_info_modules.py
```diff
@@ -55,6 +55,91 @@ def get_search_field_info(type: str, count: int, index_errors: int = 0, **kwargs
...
+    'search_cursors',
+    'search_garbage_collector',
+    'search_queries',
+    'search_warnings_and_errors',
+    'search_coordinator_warnings_and_errors',
+    'search_multi_threading',
+    'search_dialect_statistics',
+  ]
+
+  def _assert_minimal_info_on_zero_indexes(info):
```
kei-nan, line 85: this test seems like really exaggerated
we already have tests that check for info when it is active.
here I think we should only check that when we omit then nothing else gets outputted.

tests/pytests/test_info_modules.py
kei-nan: because in oss the user chooses to load us with the server
in enterprise he does not.
So the expectation for info is different

not critical - if you want we can keep as is.

oshadmi: Since Redis 8.0 Search is also loaded without user needing to choose anything.
Let's keep the default.

tests/pytests/test_config.py
```diff
@@ -1370,6 +1370,14 @@ def testStringArgDeprecationMessage():
     ('search-enable-unstable-features', 'ENABLE_UNSTABLE_FEATURES', 'no', False, False),
 ]
 
+# CONFIG-only boolean parameters (no corresponding FT.CONFIG parameter / module argument)
+# These should be validated via CONFIG GET/SET only.
+configOnlyBooleanConfigs = [
+    # configName, defaultValue
+    ('search-_simulate-in-flex', 'no'),
```
nafraf: `_SIMULATE_IN_FLEX` still exists in `config.c`

tests/pytests/test_debug_commands.py
oshadmi: Moved the assert inside `allShards_set_info_on_zero_indexes`.
But here it is enough to set the config on the env connection - will simplify it.
Commit 3d93d505a

tests/pytests/test_dialect.py
oshadmi: Commit 79b136fdf

tests/pytests/test_config.py
oshadmi: Commit 19aefd587

src/config.c
oshadmi: Such as when we want to execute and test `info` regardless of any existing index

oshadmi: And it is an **internal** config `search-_info-on-zero-indexes`

tests/pytests/test_info_modules.py
oshadmi: I simplified it a bit
Commit 042ef6785
But we still need to check that suppression works (some sections are missing with zero indices), minimal mode still emits something, and that toggling on/off works.

tests/pytests/common.py
```diff
@@ -370,6 +367,33 @@ def verify_command_OK_on_all_shards(env, *args):
...
+    In cluster mode, applies to all OSS shards. In standalone mode, applies to the single node.
+    Asserts success (all replies are OK).
+    """
+    val = 'yes' if enabled else 'no'
+    if env.isCluster():
+        verify_command_OK_on_all_shards(env, 'CONFIG', 'SET', 'search-_info-on-zero-indexes', val)
+        return
+    res = env.cmd('CONFIG', 'SET', 'search-_info-on-zero-indexes', val)
+    env.assertEqual(res, 'OK')
+    return
```
nafraf, line 383: can we remove this `return`?

PR #8284, opened 2026-02-05
MOD-13667: Add disk GC API

Itzikvaknin:
Add disk GC API

GC functionality and implementation are on the Rust side.
Lifecycle management (start/stop) is controlled from the C side.

---

> [!NOTE]
> **Medium Risk**
> Introduces new disk-backend API hooks and starts/stops a GC thread pool during module init/shutdown; issues here could cause thread lifecycle leaks or shutdown instability in disk-enabled deployments.
> 
> **Overview**
> Adds a disk GC lifecycle API (`startGCThreadPool`/`stopGCThreadPool`) and wires it into module startup/shutdown so disk-backed deployments automatically start GC after a successful disk open and stop it before closing.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cb70860aa1c15a204dd17acf7243fecda2827291. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/search_disk.c
+++ b/src/search_disk.c
@@ -43,14 +43,14 @@ bool SearchDisk_Initialize(RedisModuleCtx *ctx) {
 
   disk_db = disk->basic.open(ctx);
   if (disk_db) {
-    disk->basic.startGCThread();
+    disk->basic.startGCThreadPool();
   }
   return disk_db != NULL;
 }
 
 void SearchDisk_Close() {
   if (disk && disk_db) {
-    disk->basic.stopGCThread();
+    disk->basic.stopGCThreadPool();
     disk->basic.close(disk_db);
     disk_db = NULL;
   }
--- a/src/search_disk_api.h
+++ b/src/search_disk_api.h
@@ -48,8 +48,8 @@ typedef struct AsyncReadResult {
 typedef struct BasicDiskAPI {
   RedisSearchDisk *(*open)(RedisModuleCtx *ctx);
   void (*close)(RedisSearchDisk *disk);
-  void (*startGCThread)(void);
-  void (*stopGCThread)(void);
+  void (*startGCThreadPool)(void);
+  void (*stopGCThreadPool)(void);
   RedisSearchDiskIndexSpec *(*openIndexSpec)(RedisSearchDisk *disk, const char *indexName, size_t indexNameLen, DocumentType type);
   void (*closeIndexSpec)(RedisSearchDisk *disk, RedisSearchDiskIndexSpec *index);
   void (*indexSpecRdbSave)(RedisModuleIO *rdb, RedisSearchDiskIndexSpec *index);


src/search_disk.c
```diff
@@ -53,6 +53,16 @@ void SearchDisk_Close() {
   }
 }
 
+void SearchDisk_StartGC() {
+  RS_ASSERT(disk);
+  disk->basic.startGC();
+}
+
+void SearchDisk_StopGC() {
+  RS_ASSERT(disk);
```
JoanFM: I do not think this is now guaranteed, I think we should check if disk, not assert

JoanFM: or make the caller check, do as SearchDisk_Close do

JoanFM (changes requested):
I would expect to have some API where we force a Compaction or GCTrigger.

Itzikvaknin:
> I would expect to have some API where we force a Compaction or GCTrigger.

@JoanFM I will have a second PR for this

src/search_disk.c
Itzikvaknin: See https://github.com/RediSearch/RediSearch/pull/8284/changes/33af7c1d055547587926bde81c32d643548359b9

src/module.c
```diff
@@ -4654,6 +4655,7 @@ int RedisModule_OnUnload(RedisModuleCtx *ctx) {
     RSGlobalConfig.defaultScorer = NULL;
   }
 
+  SearchDisk_StopGC();
```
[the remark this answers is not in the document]
Itzikvaknin: Is now guarded by 
```
void SearchDisk_StopGC() {
  if (disk){
    disk->basic.stopGC();
  }
}
```

src/module.c
```diff
@@ -1664,6 +1664,7 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx) {
       RedisModule_Log(ctx, "error", "Search Disk is enabled but could not be initialized");
       return REDISMODULE_ERR;
     }
+    SearchDisk_StartGC();
```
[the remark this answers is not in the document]
Itzikvaknin: @raz-mon WDYT?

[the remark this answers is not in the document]
JoanFM: we can remove, but would not be critical I think as process would die. But proper cleaning could be nice.

src/search_disk.h
```diff
@@ -36,6 +36,22 @@ bool SearchDisk_Initialize(RedisModuleCtx *ctx);
  */
 void SearchDisk_Close();
 
+/**
+ * @brief Start the GC scheduler
+ *
+ * Should be called after SearchDisk_Initialize() succeeds.
+ * Spawns a background thread that periodically runs GC on inverted indexes.
+ */
+void SearchDisk_StartGC(void);
```
kei-nan: Why do we want to allow the OSS to control whether or not we run GC?

JoanFM: what would be the alternative? At the end is the OSS that is loaded as a module and is using the Disk, so it seems natural that it actually tells to run GC.

Or is the missunderstanding due to the naming (Should be SearchDisk_StartGCThread/Loop) or something like this?

src/module.c
JoanFM: Maybe we can change the naming to make it clear the API does not trigger a GC run but schedules the threads so to speak

src/search_disk.h
Itzikvaknin: I can move it inside of `SearchDisk_Initialize` 
But I dont think we should/can make the OSS completely unaware of the disk GC, we are going to expose debug API and maybe add configuration options soon

src/search_disk.c
```diff
@@ -42,12 +42,15 @@ bool SearchDisk_Initialize(RedisModuleCtx *ctx) {
   RedisModule_Log(ctx, "warning", "RediSearch disk API enabled");
 
   disk_db = disk->basic.open(ctx);
-
+  if (disk_db) {
+    disk->basic.startGC();
+  }
```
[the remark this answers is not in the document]
Itzikvaknin: What about the db file? 
` disk_db = disk->basic.open(ctx);` in `SearchDisk_Initialize`? Its the same issue isnt it? @JoanFM

src/search_disk.c
```diff
@@ -42,12 +42,15 @@ bool SearchDisk_Initialize(RedisModuleCtx *ctx) {
   RedisModule_Log(ctx, "warning", "RediSearch disk API enabled");
 
   disk_db = disk->basic.open(ctx);
-
+  if (disk_db) {
+    disk->basic.startGCThread();
```
raz-mon:
```suggestion
    disk->basic.startGCThreadPool();
```
or 
```suggestion
    disk->basic.startGC();
```
Cause it's not necessarily a single thread..

Itzikvaknin: Done https://github.com/RediSearch/RediSearch/pull/8284/commits/cb70860aa1c15a204dd17acf7243fecda2827291

src/search_disk.c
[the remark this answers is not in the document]
JoanFM: I think so.

What we can do is int he OnLoad have a on_error: to goto where we check if we need to close and stop the GC
Let's do this for completion

JoanFM (approved):

src/search_disk.h
kei-nan: I think any config should be registered and then we can just use the redis module api.
It feels very risky to allow the OSS to control this functionality because if we mess up it will have big consequences.
I would much rather the logic be inside the disk side instead.
If we want something at a global level then maybe that could make sense but not at an index level.
e.g for replication

PR #8292, opened 2026-02-06
[MOD-13913] Add debug implementation for NumericRangeTree

LukeMathWalker:

Rust equivalent of the debug representation for `NumericRangeTree` provided by `debug_commands.c`.

Stacked on top #8276.

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Adds new `FT.DEBUG` reply generation paths for `NumericRangeTree` and changes the tree DFS iteration order, which could affect any caller relying on node traversal order; functionality is debug-only but touches core data structure iteration.
> 
> **Overview**
> Adds Rust implementations for `FT.DEBUG` introspection of `NumericRangeTree` (`NUMIDX_SUMMARY`, `DUMP_NUMIDX`, `DUMP_NUMIDXTREE`), including optional per-range inverted-index headers and computed average memory-efficiency stats.
> 
> Updates `inverted_index::debug::Summary` to format its metrics into Redis replies, exposes `numeric_range_tree::debug`, and changes the tree iterator to `ReversePreOrderDfsIterator` (right-first traversal), with new `insta`-based snapshot integration tests and fixtures to lock the debug output format.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7a9082739073e7eaea2803e2ff26b66a43fae62b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/redisearch_rs/numeric_range_tree/src/debug.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+/// Reply with the entries (doc_ids) from a range.
+fn reply_range_entries(entries_arr: &mut ArrayBuilder<'_>, range: &NumericRange) {
+    let mut last_doc_id: u64 = 0;
+
+    let mut reader = range.reader();
+    let mut result = RSIndexResult::numeric(0.0);
+    while reader.next_record(&mut result).unwrap_or(false) {
+        if result.doc_id != last_doc_id {
```
meiravgri: why do we skip printing all multi index results?
Is there a test for multi value?

src/redisearch_rs/numeric_range_tree/src/debug.rs
```diff
@@ -0,0 +1,267 @@
...
+        };
+        stats_map.kv_double(
+            c"Average memory efficiency (numEntries/size)/numRanges",
+            avg_efficiency,
+        );
+    }
+}
+
+/// Reply with inverted index header information (adds 1 element to parent).
+fn reply_inverted_index_header(parent: &mut ArrayBuilder<'_>, entries: &NumericIndex) {
```
meiravgri: consider moving to a general inverted index debug file. it essentially implements 
[debug_commands.c](https://github.com/RediSearch/RediSearch/blob/4995446c3fc0c11976c5aaf6d3a7aa894672ef8b/src/debug_commands.c)

src/redisearch_rs/numeric_range_tree/src/debug.rs
```diff
@@ -0,0 +1,267 @@
...
+        let mut entries_arr = arr.array();
+        reply_inverted_index_debug(&mut entries_arr, range)
+    };
+    stats.total_efficiency += inv_idx_stats.total_efficiency;
+
+    stats
+}
+
+/// Reply with debug info for an inverted index.
+fn reply_inverted_index_debug(
```
meiravgri: This function as well can be move to inverted index debug file

src/redisearch_rs/numeric_range_tree/tests/integration/debug.rs
```diff
@@ -0,0 +1,240 @@
...
+      "lastDocId",
+      0,
+      "revisionId",
+      0,
+      "emptyLeaves",
+      1,
+      "RootMaxDepth",
+      0,
+      "MemoryUsage",
+      776
```
meiravgri, line 87: Is it possible to replace the hardcoded `MemoryUsage` value with a calculation?

src/redisearch_rs/numeric_range_tree/src/debug.rs
LukeMathWalker: That was my first instinct too, but this is actually specific to a numeric index since it makes two specific assumptions:
- All entries are numeric
- And the index is tracking the number of entries explicitly (i.e. it uses our `EntriesTrackingIndex` wrapper around the "bare" index)

It is not used for anything else, and in fact, its C counterpart gets deleted in #8306.

src/redisearch_rs/numeric_range_tree/tests/integration/debug.rs
LukeMathWalker: I have intentionally avoided that pattern, since I find it wastes a lot of developer time when those estimates have to be updated—we essentially end up duplicating the memory-usage computation logic across N places (the original implementation + tests that measure memory usage for that struct, either directly or indirectly). The comments get outdated, and the numbers often don't match what the comments state. 

I prefer to have the hard-coded number in snapshot tests. If the memory usage changes due to structural changes, you still catch it in PR reviews, and you get to see how that affects well-known cases.

src/redisearch_rs/numeric_range_tree/src/debug.rs
LukeMathWalker: Converted this into a method on the `Summary` itself.

src/redisearch_rs/numeric_range_tree/src/debug.rs
LukeMathWalker: If my understanding of the existing C code is correct, it does currently avoid to emit the same doc id multiple time.
It goes:
```c
QueryIterator *iter = NewInvIndIterator_NumericQuery(range->entries, sctx, &fieldCtx, NULL, NULL, range->minVal, range->maxVal);
ReplyIteratorResultsIDs(iter, sctx->redisCtx);
```
which in turn invokes:
```c
InitInvIndIterator(&numIt->base, INV_IDX_NUMERIC_ITERATOR, idx, NewNumericResult(), fieldCtx, sctx, &decoderCtx, NumericCheckAbort);
```
which in turn uses this helper to decide if it should skip multi-values:
```c
// Returns true if the iterator should skip multi-values from the same document
static inline bool ShouldSkipMulti(const InvIndIterator *it) {
  return IndexReader_HasMulti(it->reader); // The index holds multi-values (if not, no need to check)
}
```
which basically always evaluates to "skip multivalues in read, just don't bother checking if there are no multivalues".

I _guess_ the reasoning goes as follows: we are only returning the doc ids, without their values, so what's the point of printing the same doc id N times?
I'm happy to change it, but I assume this will be a user-visible change if the debug commands are exposed?

src/redisearch_rs/numeric_range_tree/src/debug.rs
```diff
@@ -0,0 +1,248 @@
...
+            if with_headers {
+                let mut fixed_arr = arr.fixed_array(2);
+                range
+                    .entries()
+                    .summary()
+                    .reply_with_inverted_index_header(&mut fixed_arr);
+                reply_range_entries(&mut fixed_arr.array(), range);
+            } else {
+                reply_range_entries(&mut arr.array(), range);
+            }
```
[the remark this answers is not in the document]
LukeMathWalker: It is right, the C code uses:

```c
NumericRangeNode *NumericRangeTreeIterator_Next(NumericRangeTreeIterator *iter) {
  if (array_len(iter->nodesStack) == 0) {
    return NULL;
  }
  NumericRangeNode *ret = array_pop(iter->nodesStack);
  if (!NumericRangeNode_IsLeaf(ret)) {
    array_append(iter->nodesStack, ret->left);
    array_append(iter->nodesStack, ret->right);
  }

  return ret;
}
```
which goes `parent -> right -> left` while we go `parent -> left -> right`.
Should we change our iteration order to match? @meiravgri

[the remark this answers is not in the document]
LukeMathWalker: Changed it to match the existing C order. Renamed the iterator accordingly.

src/redisearch_rs/numeric_range_tree/src/debug.rs
LukeMathWalker: Changed it to include all ids.

src/redisearch_rs/numeric_range_tree/src/debug.rs
meiravgri: so lets rename it

meiravgri (dismissed):

src/redisearch_rs/numeric_range_tree/src/debug.rs
LukeMathWalker: Changed to `reply_numeric_index_debug` 👍🏻

meiravgri (approved):

PR #8296, opened 2026-02-06
[MOD-13587] RSValue: Reimplement array and map types - Part 2 of 3

daxhuiberts:
This adds the new array and map collection implementation.

Uninitialized data is only present during construction before the RsValue is ever created, so no need to pollute the RsValue with MaybeUninit logic. That can live entirely at the FFI boundary and only safe code can be used in RsValue for the array an map types.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches cross-language FFI and core value container ownership semantics (arrays/maps, refcounting), so mistakes could cause leaks or memory safety issues despite the change being mostly mechanical and well-scoped.
> 
> **Overview**
> Reworks `RSValue` *array* and *map* representations in the Rust `value` crate to use new `Collection` wrappers (boxed slices with `u32`-bounded lengths) and adds `variant_name()` for clearer type assertions.
> 
> Introduces new Rust FFI surface for collections: `RSValue_AllocateArray` + `RSValue_NewArray` ownership-taking constructor, plus a new opaque `RSValueMap` builder API (`RSValueMap_AllocUninit`, `RSValueMap_SetEntry`, `RSValue_NewMap`) and map/array accessors (`RSValue_ArrayLen`, `RSValue_ArrayItem`, `RSValue_Map_Len`, `RSValue_Map_GetEntry`). The autogenerated `value.h` header is updated accordingly and includes a new `RSValue_DecrRef` export for refcount management.
> 
> Updates C code to match the new allocation/ownership conventions: array allocations switch from raw `rm_calloc`/`NULL` to `RSValue_AllocateArray` (including empty arrays), and `RSValueMap_AllocUninit`/`RSValue_NewMap` usage changes from pass-by-value to heap-allocated pointer semantics in `rpnet.c`, `rlookup.c`, and the random-sample reducer; the C implementation of `RSValueMap_AllocUninit`/`RSValue_NewMap` is adjusted to allocate/free the map struct pointer.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cd1772020966a99d2330f99bd68f26b01537f773. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
+++ b/src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
@@ -15,7 +15,7 @@ use value::{Array, RsValue, shared::SharedRsValue};
 /// The returned buffer must be populated and then passed to [`RSValue_NewArray`]
 /// to produce an array value.
 ///
-/// # SAFETY
+/// # Safety
 ///
 /// 1. The caller must eventually pass the returned pointer to [`RSValue_NewArray`].
 #[unsafe(no_mangle)]
@@ -30,7 +30,7 @@ pub unsafe extern "C" fn RSValue_AllocateArray(len: u32) -> *mut *mut RsValue {
 /// Takes ownership of the `values` buffer and all [`RsValue`] pointers within it.
 /// The values will be freed when the array is freed.
 ///
-/// # SAFETY
+/// # Safety
 ///
 /// 1. `values` must have been allocated via [`RSValue_AllocateArray`] with
 ///    a capacity equal to `len`.
@@ -55,7 +55,7 @@ pub unsafe extern "C" fn RSValue_NewArray(values: *mut *mut RsValue, len: u32) -
 ///
 /// If `value` is not an array, returns `0`.
 ///
-/// # SAFETY
+/// # Safety
 ///
 /// 1. `value` must point to a valid [`RsValue`] obtained from an `RSValue_*` function.
 #[unsafe(no_mangle)]
@@ -76,10 +76,13 @@ pub unsafe extern "C" fn RSValue_ArrayLen(value: *const RsValue) -> u32 {
 /// If `value` is not an array, returns a null pointer. The returned pointer
 /// is borrowed from the array and must not be freed by the caller.
 ///
-/// # SAFETY
+/// # Safety
 ///
 /// 1. `value` must point to a valid [`RsValue`] obtained from an `RSValue_*` function.
-/// 2. `index` must be less than the array length.
+///
+/// # Panics
+///
+/// Panics if `index` greater than or equal to the array length.
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn RSValue_ArrayItem(value: *const RsValue, index: u32) -> *mut RsValue {
     // Safety: ensured by caller (1.)
--- a/src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
+++ b/src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
@@ -24,7 +24,7 @@ pub struct RSValueMap {
 /// The map entries are uninitialized and must be set using [`RSValueMap_SetEntry`]
 /// before being finalized into an [`RsValue`] via [`RSValue_NewMap`].
 ///
-/// # SAFETY
+/// # Safety
 ///
 /// 1. All entries must be initialized via [`RSValueMap_SetEntry`] before
 ///    passing the map to [`RSValue_NewMap`].
@@ -41,13 +41,16 @@ pub unsafe extern "C" fn RSValueMap_AllocUninit(len: u32) -> *mut RSValueMap {
 ///
 /// Takes ownership of both the `key` and `value` [`RsValue`] pointers.
 ///
-/// # SAFETY
+/// # Safety
 ///
 /// 1. `map` must be a valid pointer to an [`RSValueMap`] created by
 ///    [`RSValueMap_AllocUninit`].
-/// 2. `index` must be less than the map length.
-/// 3. `key` and `value` must be valid pointers to [`RsValue`]
+/// 2. `key` and `value` must be valid pointers to [`RsValue`]
 ///    obtained from an `RSValue_*` function.
+///
+/// # Panics
+///
+/// Panics if `index` is greater than or equal to the map length.
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn RSValueMap_SetEntry(
     map: *mut RSValueMap,
@@ -67,7 +70,7 @@ pub unsafe extern "C" fn RSValueMap_SetEntry(
 /// Takes ownership of the map structure and all its entries. The [`RSValueMap`]
 /// pointer is consumed and must not be used after this call.
 ///
-/// # SAFETY
+/// # Safety
 ///
 /// 1. `map` must be a valid pointer to an [`RSValueMap`] created by
 ///    [`RSValueMap_AllocUninit`].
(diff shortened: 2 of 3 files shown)


src/redisearch_rs/headers/value.h
```diff
@@ -135,6 +192,86 @@ const struct RsValue *RSValue_Trio_GetMiddle(const struct RsValue *value);
...
+ *
+ * # SAFETY
+ *
+ * 1. `map` must point to a valid [`RsValue`] obtained from an `RSValue_*` function.
+ *
+ * # Panics
+ *
+ * Panics if `map` is not a map value.
+ */
+uint32_t RSValue_Map_Len(const struct RsValue *map);
```
Itzikvaknin, line 257:
```suggestion
size_t RSValue_Map_Len(const struct RsValue *map);
```

src/redisearch_rs/headers/value.h
```diff
@@ -135,6 +192,86 @@ const struct RsValue *RSValue_Trio_GetMiddle(const struct RsValue *value);
...
+ * Retrieves a key-value pair from a map [`RsValue`] at a specific index.
+ *
+ * The returned key and value pointers are borrowed from the map and must
+ * not be freed by the caller.
+ *
+ * # SAFETY
+ *
+ * 1. `map` must point to a valid [`RsValue`] obtained from an `RSValue_*` function.
+ * 2. `index` must be less than the map length.
+ * 3. `key` and `value` must be valid, non-null pointers to writable
```
Itzikvaknin: Is it necessary to have the double pointer here? why not require the caller to provide the writable address?

src/redisearch_rs/headers/value.h
daxhuiberts: The original `RSValue_Map_Len` in `value.c` also uses a `uint32_t` as the max size of the map (and array) is always stored in a `uint32_t`, so that's why it's the same here.

src/redisearch_rs/headers/value.h
daxhuiberts: Same as above. `RSValue_Map_GetEntry` uses the double pointer as well. The reason for that is that a reference is returned and not a (copied) value. So that's why you give two writable addresses to put the references.

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
```diff
@@ -0,0 +1,198 @@
...
+/// Creates a heap-allocated map [`RsValue`] from an [`RSValueMap`].
+///
+/// Takes ownership of the map structure and all its entries. The [`RSValueMap`]
+/// pointer is consumed and must not be used after this call.
+///
+/// # SAFETY
+///
+/// 1. `map` must be a valid pointer to an [`RSValueMap`] created by
+///    [`RSValueMap_AllocUninit`].
+/// 2. All entries in the map must have been initialized via [`RSValueMap_SetEntry`].
```
Itzikvaknin, line 77: Can we eliminate this safety requirement by having an increasing index variable for what has been initizlized?

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
```diff
@@ -0,0 +1,198 @@
...
+    let entries = vec![MaybeUninit::uninit(); len as usize];
+
+    Box::into_raw(Box::new(RSValueMap {
+        entries: entries.into(),
+    }))
+}
+
+/// Sets a key-value pair at a specific index in the map.
+///
+/// Takes ownership of both the `key` and `value` [`RsValue`] pointers.
```
Itzikvaknin, line 42: Should we nulify key and value?

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
```diff
@@ -0,0 +1,198 @@
...
+/// # SAFETY
+///
+/// 1. `map` must point to a valid [`RsValue`] obtained from an `RSValue_*` function.
+/// 2. `index` must be less than the map length.
+/// 3. `key` and `value` must be valid, non-null pointers to writable
+///    `*mut RsValue` locations.
+///
+/// # Panics
+///
+/// Panics if `map` is not a map value.
```
Itzikvaknin: It may also panic here?
```
&map[index as usize];
```

src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
```diff
@@ -0,0 +1,148 @@
...
+use value::{Array, RsValue, shared::SharedRsValue};
+
+/// Allocates an array of null pointers with space for `len` [`RsValue`] pointers.
+///
+/// The returned buffer must be populated and then passed to [`RSValue_NewArray`]
+/// to produce an array value.
+///
+/// # SAFETY
+///
+/// 1. The caller must eventually pass the returned pointer to [`RSValue_NewArray`].
```
Itzikvaknin, line 20: Covered here
```
/// The returned buffer must be populated and then passed to [`RSValue_NewArray`]
/// to produce an array value.
```

src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    #[test]
+    fn test_array() {
+        unsafe {
+            let array = RSValue_AllocateArray(3);
+            let one = RSValue_NewNumber(1.0);
+            let two = RSValue_NewNumber(2.0);
+            let three = RSValue_NewNumber(3.0);
+            array.add(0).write(one);
```
Itzikvaknin, line 128: Do we have an api for writes? or we just use [] in the C side and it magically works ?

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
daxhuiberts: As discussed in our call I can add a debug assert that ensures all entries are set before finalizing the map.
This can be done with an increasing index variable, but that assumes that entries are always set in order (which currently is the case at all the call sites). The other option is to have a flag per entry which I will look into.

src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
daxhuiberts: No, Rust has special methods on a pointer (`array`) to do pointer arithmetic. This is the same as `array[0] = one` in C.

The reason Rust uses special methods for this is that Rust pointer also track 'provenance' (see https://doc.rust-lang.org/std/ptr/index.html#provenance for more information), which is a rather difficult topic. But provenance helps in reasoning about what is allowed you can do with memory.

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
daxhuiberts: We can't nullify the callers variables, that's up to them. Or do you mean something else?

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
daxhuiberts: Good catch! Will add that here as well.

src/redisearch_rs/c_entrypoint/value_ffi/src/map.rs
daxhuiberts: I discussed this with Luca and having the safety comment there is still preferred. The integration test suite together with the sanitizer runs is a good enough safeguard at this moment. Combined with the fact that we keep this FFI layer close to the C implementation to keep the same contract and performance as much as possible.

This safety requirement is only needed in the FFI layer. When interacting with an `RsValue::Array` or `RsValue::Map` directly from Rust it is all completely safe by design, as you can only provide it with a fully initialized `Array` or `Map` in the first place.

Itzikvaknin (approved):

PR #8299, opened 2026-02-06
[MOD-13594] RSValue: Implement new string setup - Part 2 of 3

daxhuiberts:
This introduces the new string setup. As well as several string constructors, getters and setters.

The new RsString type contains a `ptr`, `len` and the allocator source, which is either the rust allocation, redis module allocator, or no allocator (for string constants).

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches core `RSValue` string representation and C/Rust FFI boundaries (new constructors/getters/setters, constness, and length type changes), which can impact memory ownership and string lifetimes across the module. Additional assertions and tests reduce risk, but regressions would affect query/aggregate outputs and reply serialization.
> 
> **Overview**
> **Introduces a new string model for `RSValue` across C and Rust.** Rust `RsValue` now stores strings as `RsString` (owned/`RedisModule_Alloc`/borrowed) and adds a `RedisString` variant that owns a `RedisModuleString` and frees it on drop.
> 
> **Expands and updates the C/Rust FFI surface for strings.** Adds `RSValue_NewString`, `RSValue_NewBorrowedString`, `RSValue_NewRedisString`, `RSValue_NewCopiedString`, plus new setters `RSValue_SetString`/`RSValue_SetConstString` and string getters `RSValue_String_Get`, `RSValue_RedisString_Get`, and a more permissive `RSValue_StringPtrLen` that handles `Ref`/`Trio` and both string variants; `RSValue_IsString` is broadened accordingly.
> 
> **Propagates API/ABI adjustments and safety checks through the C codebase.** String lengths are normalized to `u32` (with `RS_ASSERT` bounds checks before conversions), `RSValue_NewConstString` is replaced by `RSValue_NewBorrowedString`, and `RSValue_String_Get` now returns `const char *`; related call sites (aggregate functions, coordinator reply conversion, result processors, reply serialization, and C++ tests) are updated. Adds `redis_mock` dev-dependency and new `value_ffi` string tests to validate ownership/borrowing behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b63d40549308a59c7062e65b5a0967eafd616393. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/aggregate/functions/string.c
+++ b/src/aggregate/functions/string.c
@@ -43,6 +43,7 @@ static int func_matchedTerms(ExprEval *ctx, RSValue **argv, size_t argc, RSValue
       for (size_t i = 0; i < n; i++) {
         size_t len;
         const char *str = QueryTerm_GetStrAndLen(terms[i], &len);
+        RS_ASSERT(len <= UINT32_MAX);
         arr[i] = RSValue_NewBorrowedString(str, len);
       }
       RSValue *v = RSValue_NewArrayFromBuilder(arr, n);


src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
```diff
@@ -70,3 +70,103 @@ pub unsafe extern "C" fn RSValue_NewTrio(
...
+/// # Safety
+///
+/// 1. `str` must be a [valid], non-null pointer to a string buffer.
+/// 2. `str` must be [valid] for reads of `len` bytes.
+/// 3. The memory pointed to by `str` must remain valid and not be mutated for the entire
+///    lifetime of the returned [`RsValue`] and any clones of it.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_NewConstString(str: *const c_char, len: u32) -> *mut RsValue {
```
LukeMathWalker: This is interesting—are we saying that this, morally, a `&'a str`?
The name `ConstString` led me to believe we were only talking about `&'static str`—i.e. the memory pointed at by `str` must remain valid for the entire lifetime of the program.

src/redisearch_rs/c_entrypoint/value_ffi/src/getters.rs
```diff
@@ -94,3 +96,102 @@ pub unsafe extern "C" fn RSValue_Trio_GetRight(value: *const RsValue) -> *const
         panic!("Expected a trio value")
     }
 }
+
+/// Returns a pointer to the string data of an [`RsValue`] and optionally writes the string
+/// length to `lenp`.
```
LukeMathWalker:
```suggestion
/// Returns a pointer to the string data of an [`RsValue`] and writes the string
/// length to `lenp`, if `lenp` is a non-null pointer.
```

src/redisearch_rs/c_entrypoint/value_ffi/src/getters.rs
```diff
@@ -94,3 +96,102 @@ pub unsafe extern "C" fn RSValue_Trio_GetRight(value: *const RsValue) -> *const
...
+///
+/// # Safety
+///
+/// 1. `value` must point to a valid [`RsValue`] obtained from an `RSValue_*` function.
+///
+/// # Panic
+///
+/// Panics if the value is not a `RedisString` type.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_RedisString_Get(
```
LukeMathWalker: Can the caller mutate the string via this pointer? Since we're using `*const`, I assume not, but let's be clear about it.

src/redisearch_rs/value/src/rs_string.rs
```diff
@@ -0,0 +1,105 @@
...
+        let ptr = str.into_raw();
+
+        Self {
+            ptr,
+            len: len as u32,
+            kind: RsStringKind::RustAlloc,
+        }
+    }
+
+    /// Create an [`RsString`] from a constant string.
```
LukeMathWalker: Copy-paste error?

src/redisearch_rs/value/src/rs_string.rs
```diff
@@ -0,0 +1,105 @@
...
+            kind: RsStringKind::RmAlloc,
+        }
+    }
+
+    /// Create an [`RsString`] from a constant string.
+    ///
+    /// # Safety
+    ///
+    /// 1. `ptr` must not be NULL and must point to a valid string of `len` size.
+    /// 2. The string pointed to by `ptr`/`len` must be nul-terminated.
```
LukeMathWalker: The safety contract must include the expected lifetime.

src/redisearch_rs/value/src/rs_string.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub unsafe fn rm_alloc_string(ptr: *const c_char, len: u32) -> Self {
```
LukeMathWalker: We should check, at least with a debug assert, that the pointer is not null. And perhaps null termination.

src/redisearch_rs/value/src/rs_string.rs
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    /// Create an [`RsString`] from a constant string.
+    ///
+    /// # Safety
+    ///
+    /// 1. `ptr` must not be NULL and must point to a valid string of `len` size.
+    /// 2. The string pointed to by `ptr`/`len` must be nul-terminated.
+    pub unsafe fn const_string(ptr: *const c_char, len: u32) -> Self {
```
LukeMathWalker: Same here, check for null in a debug assert.

src/redisearch_rs/value/src/rs_string.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /// Returns the string data pointer and length.
+    pub fn as_ptr_len(&self) -> (*const c_char, u32) {
+        (self.ptr, self.len)
+    }
+
+    /// Gets the string pointed to by `ptr`/`len` as a byte slice.
+    pub fn as_bytes(&self) -> &[u8] {
+        unsafe { std::slice::from_raw_parts(self.ptr as _, self.len as usize) }
```
LukeMathWalker, line 97: Safety comment?

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
daxhuiberts: Fair point. From the C perspective it's 'const' in the sense that it's a `*const c_char` which is borrowed (as a `&'a str`) and not owned by the RsValue. For our purposes, it should behave as a `fn<'a>(&'a str) -> SharedRsValue<'a>`. It is only used in one place in the C code (`src/aggregate/functions/string.c`) and from that it doesn't look like a constant.

I kept the naming of all the `RSValue_` functions intact as to keep the change/swap as small as possible. Do you prefer to change this to something like `RSValue_NewBorrowedString`?

src/redisearch_rs/c_entrypoint/value_ffi/src/getters.rs
daxhuiberts: I don't know the calling code well enough whether that is done or not. My understanding is that `RsValue` objects are not meant to be mutated (with one exception when the `RsValue` is not shared yet and can change from e.g. `Undefined` to `Number`).

daxhuiberts: So not sure what to add here. I like to add that the returned value is a read only reference, but I can't guarantee that.

src/redisearch_rs/value/src/rs_string.rs
daxhuiberts: That is added in the next PR https://github.com/RediSearch/RediSearch/pull/8297

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
LukeMathWalker: Yes, let's change it.

src/redisearch_rs/c_entrypoint/value_ffi/src/getters.rs
LukeMathWalker: We obviously can't guarantee it, but we should document the expectation. All FFI functions are `unsafe`, in that sense 😅

daxhuiberts: Yeah I understand that. I meant that I'm not sure the expectation is correct. Only that I likely think it is. Will add anyways.

src/redisearch_rs/c_entrypoint/value_ffi/src/getters.rs
```diff
@@ -94,3 +96,102 @@ pub unsafe extern "C" fn RSValue_Trio_GetRight(value: *const RsValue) -> *const
...
+
+    let (ptr, len) = str.as_ptr_len();
+
+    // Safety: ensured by caller (2.)
+    if let Some(lenp) = unsafe { lenp.as_mut() } {
+        *lenp = len;
+    }
+
+    ptr as *mut _
+}
```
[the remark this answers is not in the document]
daxhuiberts: Good point! Changed it to a `const *char`

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
```diff
@@ -70,3 +70,103 @@ pub unsafe extern "C" fn RSValue_NewTrio(
...
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_NewCopiedString(str: *const c_char, len: u32) -> *mut RsValue {
+    // Safety: ensured by caller (1., 2.)
+    let slice = unsafe { std::slice::from_raw_parts(str.cast::<u8>(), len as usize) };
+
+    // Safety: ensured by caller (3.)
+    let cstring = unsafe { CString::from_vec_unchecked(slice.to_vec()) };
+
```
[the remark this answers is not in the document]
daxhuiberts: Good point. A lot of C code assumes proper c strings inside a `RsValue::String` so I put it up as a requirement that all `RsValue::String`s can't have any nul values inside of the string. But I'm not debug checking that that is actually the case, nor do I think that that is needed. @LukeMathWalker any thought?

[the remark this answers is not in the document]
LukeMathWalker: I don't think it's needed, I would simply require a nul terminator.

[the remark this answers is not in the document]
LukeMathWalker: But if we're only requiring a nul terminator, we shouldn't use `CString`/`CStr`

src/redisearch_rs/value/src/rs_string.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/// A `CString` like string for [`RsValue`](crate::RsValue) with support for rust allocated string,
+/// C allocated strings, and constant strings, and support for a max length of `u32::MAX`,
+/// all in one package.
+///
+/// # Safety
+///
+/// - `ptr` must not be NULL and must point to a valid string of `len` size.
+/// - The string pointed to by `ptr`/`len` must be nul-terminated.
```
LukeMathWalker: We should clarify whether `len` includes the nul terminator or not.

src/redisearch_rs/c_entrypoint/value_ffi/src/setters.rs
```diff
@@ -54,3 +55,73 @@ pub unsafe extern "C" fn RSValue_SetNull(value: *mut RsValue) {
...
+///
+/// # Safety
+///
+/// 1. `value` must point to a valid **owned** [`RsValue`] obtained from an
+///    `RSValue_*` function returning an owned [`RsValue`] object.
+/// 2. `str` must be a [valid], non-null pointer to a buffer allocated by `RedisModule_Alloc`.
+/// 3. `str` must be [valid] for reads of `len` bytes.
+/// 4. `str` **must not** be used or freed after this function is called, as this function
+///    takes ownership of the allocation.
+/// 5. Only 1 reference is allowed to exist pointing to this [`RsValue`] object.
```
LukeMathWalker: 5. shouldn't be a safety requirement, since it'll panic rather than cause undefined behaviour if it isn't satisfied.

src/redisearch_rs/c_entrypoint/value_ffi/src/setters.rs
```diff
@@ -54,3 +55,73 @@ pub unsafe extern "C" fn RSValue_SetNull(value: *mut RsValue) {
...
+///
+/// # Safety
+///
+/// 1. `value` must point to a valid **owned** [`RsValue`] obtained from an
+///    `RSValue_*` function returning an owned [`RsValue`] object.
+/// 2. `str` must be a [valid], non-null pointer to a string buffer.
+/// 3. `str` must be [valid] for reads of `len` bytes.
+/// 4. The memory pointed to by `str` must remain valid and not be mutated for the entire
+///    lifetime of the [`RsValue`] and any clones of it.
+/// 5. Only 1 reference is allowed to exist pointing to this [`RsValue`] object.
```
LukeMathWalker: Same as above.

src/redisearch_rs/c_entrypoint/value_ffi/tests/string.rs
```diff
@@ -0,0 +1,204 @@
+/*
```
LukeMathWalker, line 1: I'd prefer to see these tests under `value/tests`, mostly to ensure our alloc/dealloc scheme works and string accessors (which use quite a bit of unsafe) return what we expect.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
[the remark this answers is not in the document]
daxhuiberts: My understanding is that the C code requires/expects proper c strings (without nul values within the string and properly nul terminated). A couple of those memory issues we had in the integration runs were because of the c code using `strcpy` and similar functions which expect proper c strings.

And it looks like it's always that, except for the highlight processor which returns subslices of strings (thus still no nul values inside of the string, only missing a nul terminator).

Thus a `CString` is valid choice.

src/redisearch_rs/c_entrypoint/value_ffi/tests/string.rs
daxhuiberts: Yes, `RsString` should be properly tested. I've added those tests to https://github.com/RediSearch/RediSearch/pull/8300 which also contains the debug checks and all the tests for that.

Then there is a lot of overlap with the tests in this ffi layer, though this also tests the integration with the different FFI calls. I still have the feeling that this is valuable though. Maybe I should just shake that feeling and drop it.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
```diff
@@ -70,3 +70,103 @@ pub unsafe extern "C" fn RSValue_NewTrio(
...
+    let string = RsString::cstring(cstring);
+    let value = RsValue::String(string);
+    let shared_value = SharedRsValue::new(value);
+    shared_value.into_raw() as *mut _
```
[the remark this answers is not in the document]
daxhuiberts: @LukeMathWalker This is a valid point. The existing C code stores the string length in a `uint32_t len : 29;`.

Some functions accept a `uint32_t` as the length for the string (`RSValue_NewString`, `RSValue_NewBorrowedString`), while others accept a `size_t` as the length for the string (`RSValue_NewCopiedString`, `RSValue_SetString`, `RSValue_SetConstString`)
Some use a `uint32_t` to return the length of the string (`RSValue_String_Get`), while others use a `size_t` to return the length of the string (`RSValue_StringPtrLen`, `RSValue_ConvertStringPtrLen`).

The latter ones are fine, we can alway extend a u32 to a usize. But the first ones are invalid in C already which can probably fail in beautiful ways. So we either fix this in C properly first, or accept this in Rust and truncate or panic where invariants don't match.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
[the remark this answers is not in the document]
LukeMathWalker: Perfect, than it's enough to have it as a safety precondition (which I see you already added).

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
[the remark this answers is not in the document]
LukeMathWalker: Let's align on using u32 everywhere, since that's how we store the length. We can add assertions to the existing C code wherever they are missing.

src/redisearch_rs/c_entrypoint/value_ffi/tests/string.rs
LukeMathWalker: I'm fine with keeping these ones, as long as we have tests in `value` as well. If they're added in the next PR, not a blocker.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
```diff
@@ -70,3 +70,103 @@ pub unsafe extern "C" fn RSValue_NewTrio(
...
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_NewString(str: *mut c_char, len: u32) -> *mut RsValue {
+    // Safety: ensured by caller (1., 2., 3.)
+    let string = unsafe { RsString::rm_alloc_string(str, len) };
+
+    let value = RsValue::String(string);
+    let shared_value = SharedRsValue::new(value);
+    shared_value.into_raw() as *mut _
```
memark: I'm thinking we want to avoid wildcard casts like this, instead preferring `.cast_mut()` (which won't change the type without you knowing).

src/redisearch_rs/c_entrypoint/value_ffi/src/getters.rs
```diff
@@ -94,3 +96,105 @@ pub unsafe extern "C" fn RSValue_Trio_GetRight(value: *const RsValue) -> *const
...
 }
+
+/// Returns a pointer to the string data of an [`RsValue`] and optionally writes the string
+/// length to `lenp`, if `lenp` is a non-null pointer.
+///
+/// The returned pointer borrows from the [`RsValue`] and must not outlive it.
+///
+/// # Safety
+///
+/// 1. `value` must point to a valid [`RsValue`] obtained from an `RSValue_*` function.
```
memark, line 107: Mixed usage of "function" and "method" to refer to these `RSValue_*`.

src/redisearch_rs/c_entrypoint/value_ffi/src/setters.rs
```diff
@@ -54,3 +55,71 @@ pub unsafe extern "C" fn RSValue_SetNull(value: *mut RsValue) {
...
+/// 4. `str` **must not** be used or freed after this function is called, as this function
+///    takes ownership of the allocation.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+///
+/// # Panic
+///
+/// Panics if more than 1 reference exists to this [`RsValue`] object.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_SetString(value: *mut RsValue, str: *mut c_char, len: size_t) {
```
memark: We are heavily using `*mut` here, but I think [our guidelines](https://www.notion.so/mainmatter/Redis-Onboarding-Notes-1a664e58ddfa800f8d11d275ed7314b8?source=copy_link#2b064e58ddfa80b894e2c9d5e826a176) say to prefer `Option<NonNull>`.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
daxhuiberts: Will change! That is much better than an `as` cast.

src/redisearch_rs/c_entrypoint/value_ffi/src/getters.rs
daxhuiberts: Will change to 'function' everywhere.

src/redisearch_rs/c_entrypoint/value_ffi/src/setters.rs
daxhuiberts: I do like the type constraints `Option<NonNull>` gives, but I do miss the one for `*const T` which makes it all so inconsistent with eachother and `*mut T` is so much easier to identify than `Option<NonNull<T>` (especially in one case where I have a `*mut *mut T`). I am aware it is touted as a guideline, but I'm not sure it's consensually applied by everyone.

memark: I totally agree. I would be in favor of abandoning that guideline altogether as well.

src/redisearch_rs/value/src/redis_string.rs
```diff
@@ -0,0 +1,88 @@
...
+            unsafe { RedisModule_StringPtrLen }.expect("Redis module not initialized");
+
+        let mut len = MaybeUninit::uninit();
+        // SAFETY: `self.ptr` is a valid `RedisModuleString` pointer per our invariant,
+        // and `len` is a valid pointer to write the length into.
+        let ptr = unsafe { rm_str_ptr_len(self.ptr, len.as_mut_ptr()) };
+        // SAFETY: `rm_str_ptr_len` initialized `len`.
+        let len = unsafe { len.assume_init() };
+
+        (ptr, len as u32)
```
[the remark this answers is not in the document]
daxhuiberts: Changed so that a RedisString correctly returns a `size_t` instead.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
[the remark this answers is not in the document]
daxhuiberts: Changing `size_t` to `u32int_t` in the exiting `value.c/.h` goes through without any warnings in the rest of the c code where truncation might now happen. This possible silent truncation already happened in value.c when setting the len in the actual `RSValue` struct, but now happens in the callers of those `RSValue_` functions.

Do we still want to add those extra asserts in the callers or do we want to keep it as it is functionally the same? I'd like to go for the latter.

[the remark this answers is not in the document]
LukeMathWalker: Better to fail loudly than to fail silently imo.

src/value/value.c
```diff
@@ -236,7 +236,7 @@ RSValue *RSValue_NullStatic() {
   return &RS_NULL;
 }
 
-RSValue *RSValue_NewCopiedString(const char *s, size_t n) {
+RSValue *RSValue_NewCopiedString(const char *s, uint32_t n) {
```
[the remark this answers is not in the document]
daxhuiberts: Fixed

src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
```diff
@@ -22,7 +22,10 @@ use value::{Array, RsValue, shared::SharedRsValue};
 pub unsafe extern "C" fn RSValue_NewArrayBuilder(len: u32) -> *mut *mut RsValue {
     let array: Vec<*mut RsValue> = vec![std::ptr::null_mut(); len as usize];
 
-    Box::into_raw(array.into_boxed_slice()) as *mut _
+    // `Box::into_raw` returns a `*mut [*mut _]` here, `cast_mut()` can't convert that to a
+    // `*mut *mut _`. `as_mut_ptr()` can, but that is nightly only still. Keeping an `as`
+    // cast for now.
+    Box::into_raw(array.into_boxed_slice()) as *mut *mut RsValue
```
memark: You _could_ do it like this
```rust
let x = Box::into_raw(array.into_boxed_slice());
unsafe { (*x).as_mut_ptr() }
```
(ofc renaming and/or inlining `x`)

src/redisearch_rs/c_entrypoint/value_ffi/src/getters.rs
```diff
@@ -94,3 +96,108 @@ pub unsafe extern "C" fn RSValue_Trio_GetRight(value: *const RsValue) -> *const
...
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_StringPtrLen(
+    value: *const RsValue,
+    len_ptr: *mut size_t,
+) -> *const c_char {
+    // Safety: ensured by caller (1.)
+    let mut value = unsafe { expect_value(value) };
+
+    let (ptr, len) = loop {
```
memark: This _feels_ like a lot of logic to keep in an FFI function. I would understand if this is the best/only place though.

src/redisearch_rs/c_entrypoint/value_ffi/src/setters.rs
```diff
@@ -54,3 +54,67 @@ pub unsafe extern "C" fn RSValue_SetNull(value: *mut RsValue) {
...
+/// 2. `str` must be a [valid], non-null pointer to a string buffer.
+/// 3. `str` must be [valid] for reads of `len` bytes.
+/// 4. The memory pointed to by `str` must remain valid and not be mutated for the entire
+///    lifetime of the [`RsValue`] and any clones of it.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+///
+/// # Panic
+///
+/// Panics if more than 1 reference exists to this [`RsValue`] object.
```
memark, line 108: What line in the function would cause that panic?

daxhuiberts: Added a comment to the function body where this is the case.
Fix all other setters in the module as well.

src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
daxhuiberts: Updated. Please check if the comment makes sense now.

src/redisearch_rs/c_entrypoint/value_ffi/src/getters.rs
daxhuiberts: You're right. This was doing too much inside one block. I've refactored the dereferencing logic into a separate function removing this scary setup.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
```diff
@@ -68,5 +70,106 @@ pub unsafe extern "C" fn RSValue_NewTrio(
...
+    // Safety: ensured by caller (1., 2.)
+    let slice = unsafe { std::slice::from_raw_parts(str.cast::<u8>(), len as usize) };
+
+    // Safety: ensured by caller (3.)
+    let cstring = unsafe { CString::from_vec_unchecked(slice.to_vec()) };
+
+    let string = RsString::cstring(cstring);
+    let value = RsValue::String(string);
+    let shared_value = SharedRsValue::new(value);
+    shared_value.into_raw().cast_mut()
```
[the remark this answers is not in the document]
daxhuiberts: `from_vec_unchecked` adds the nul byte.

src/coord/rpnet.c
```diff
@@ -26,6 +26,7 @@ static RSValue *MRReply_ToValue(MRReply *r) {
     case MR_REPLY_STRING: {
       size_t l;
       const char *s = MRReply_String(r, &l);
+      RS_ASSERT(l <= UINT32_MAX);
```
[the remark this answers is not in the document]
daxhuiberts: The existing C code already writes a `size_t` into a `uint32_t len : 29;`. So truncation already happens. Now at least we have a debug assert.

src/redisearch_rs/value/src/lib.rs
```diff
@@ -75,10 +91,7 @@ impl RsValue {
     /// Returns the string bytes of the value, if it is a string type.
     pub fn as_str_bytes(&self) -> Option<&[u8]> {
         match self {
-            RsValue::RmAllocString(str) => Some(str.as_bytes()),
-            RsValue::ConstString(str) => Some(str.as_bytes()),
-            RsValue::RedisString(str) => Some(str.as_bytes()),
-            RsValue::String(str) => Some(str.as_str().as_bytes()),
+            RsValue::String(str) => Some(str.as_bytes()),
```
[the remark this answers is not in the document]
daxhuiberts: Fixed

src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
```diff
@@ -21,8 +21,13 @@ use value::{Array, RsValue, shared::SharedRsValue};
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn RSValue_NewArrayBuilder(len: u32) -> *mut *mut RsValue {
     let array: Vec<*mut RsValue> = vec![std::ptr::null_mut(); len as usize];
+    let array_ptr = Box::into_raw(array.into_boxed_slice());
 
-    Box::into_raw(array.into_boxed_slice()) as *mut _
+    // Safety: `array_ptr` just created above from `Box::into_raw`.
+    // `array_ptr` is a `*mut [*mut _]` here. This converts it to a `*mut *mut _`.
+    // `<pointer>.as_mut_ptr()` is a nightly on feature that directly casts `array_ptr` to
```
memark: >// Safety: `array_ptr` just created above from `Box::into_raw`.

I think this is enough.
We don't need to document the existence of functions we didn't call.
If we feel the need to document the type of a variable, better to do it in the code than in a comment.

src/redisearch_rs/c_entrypoint/value_ffi/src/getters.rs
memark: Much better!

src/redisearch_rs/c_entrypoint/value_ffi/src/array.rs
daxhuiberts: Changed it to a simple `cast()` which works as well without the `unsafe`.

memark (approved):

src/redisearch_rs/value/src/rs_string.rs
```diff
@@ -0,0 +1,129 @@
...
+        unsafe { std::slice::from_raw_parts(self.ptr as _, self.len as usize) }
+    }
+}
+
+impl Drop for RsString {
+    fn drop(&mut self) {
+        match self.kind {
+            RsStringKind::RustAlloc => {
+                // SAFETY: `self.ptr` was created by `CString::into_raw` and has not been freed.
+                drop(unsafe { CString::from_raw(self.ptr as *mut _) });
```
raz-mon: Probably want to use `.cast_mut()` here as well?

PR #8301, opened 2026-02-08
MOD-13919: Fix return code upon errors in filter matching

raz-mon:
Fixes a return code mismatch in filter expression matching, which is called upon indexing.

`EvalCtx_EvalExpr` is checked for `EXPR_EVAL_OK/ERR/NULL`, and calls `EvalCtx_Eval` which calls the following:

```
int EvalCtx_Eval(EvalCtx *r) {
  if (!r->_expr) {
    return REDISMODULE_ERR;
  }
  r->ee.root = r->_expr;
  if (ExprAST_GetLookupKeys((RSExpr *) r->ee.root, (RLookup *) r->ee.lookup, r->ee.err) != EXPR_EVAL_OK) {
    return REDISMODULE_ERR;
  }
  return ExprEval_Eval(&r->ee, r->res);
}
```

Where:
```
/* Error status return values. */
#define REDISMODULE_OK 0
#define REDISMODULE_ERR 1
```
And

```
#define EXPR_EVAL_ERR 0
#define EXPR_EVAL_OK 1
#define EXPR_EVAL_NULL 2
```
Thus there is a mismatch, since if an error is met and REDISMODULE_ERR is returned, this is interpreted as EXPR_EVAL_OK in the caller.

User impact - This code is uncovered currently, thus the impact of this fix should not be high on current deployments, meaning that the "breakage" should be small to non-existant.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Low risk: small change to error-code propagation in expression evaluation plus targeted tests; behavior only differs on failure paths and could affect callers that relied on the incorrect return code.
> 
> **Overview**
> Fixes a return-code mismatch in `EvalCtx_Eval` by returning `EXPR_EVAL_ERR` (instead of `REDISMODULE_ERR`) when there is no expression or when lookup-key binding fails, ensuring callers don’t misinterpret failures as `EXPR_EVAL_OK`.
> 
> Adds coverage to C++ expression tests for `EvalCtx_EvalExpr` error cases: passing a `NULL` expression and referencing an unknown property, asserting both the error return and the user-facing error message. Also clears `QueryError` state in `EvalCtx_Destroy` to avoid leaving stale error data behind.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 507034655b4703e928bb76e905743151098db062. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/cpptests/test_cpp_expr.cpp
+++ b/tests/cpptests/test_cpp_expr.cpp
@@ -671,7 +671,7 @@ TEST_F(ExprTest, testEvalCtxEvalExprNullExpr) {
   ASSERT_NE(ctx, nullptr);
 
   // Calling EvalCtx_EvalExpr with NULL should set _expr to NULL and return error
-  int rc = EvalCtx_EvalExpr(ctx, NULL);
+  int rc = EvalCtx_EvalExpr(ctx, nullptr);
   ASSERT_EQ(EXPR_EVAL_ERR, rc);
 
   EvalCtx_Destroy(ctx);


tests/cpptests/test_cpp_expr.cpp
```diff
@@ -665,4 +665,39 @@ TEST_F(ExprTest, testEvalFuncCaseWithDifferentTypeComparison) {
   ASSERT_EXPR_EVAL_NUMBER(ctx, 0);  // NULL == 'hello' should be false
 }
 
+TEST_F(ExprTest, testEvalCtxEvalExprNullExpr) {
+  // Test that EvalCtx_EvalExpr returns EXPR_EVAL_ERR when called with NULL expression
+  EvalCtx *ctx = EvalCtx_Create();
+  ASSERT_NE(ctx, nullptr);
+
+  // Calling EvalCtx_EvalExpr with NULL should set _expr to NULL and return error
+  int rc = EvalCtx_EvalExpr(ctx, NULL);
```
nafraf:
```suggestion
  int rc = EvalCtx_EvalExpr(ctx, nullptr);
```

raz-mon: Ok changed, although it's already mixed in this file a lot

src/aggregate/expr/expression.c
```diff
@@ -379,18 +379,19 @@ void EvalCtx_Destroy(EvalCtx *r) {
...
 
 //---------------------------------------------------------------------------------------------
 
 int EvalCtx_Eval(EvalCtx *r) {
   if (!r->_expr) {
-    return REDISMODULE_ERR;
+    return EXPR_EVAL_ERR;
   }
   r->ee.root = r->_expr;
   if (ExprAST_GetLookupKeys((RSExpr *) r->ee.root, (RLookup *) r->ee.lookup, r->ee.err) != EXPR_EVAL_OK) {
```
oshadmi, line 393: @raz-mon  We saw that in the flow of filter (on creation of an index leading to background indexing, and on foreground indexing like hset/jsonset etc.) - it is not reachable

oshadmi:
@raz-mon Can please add a short summary in the title for release notes

raz-mon:
@oshadmi I already added - see the bottom of the description..

PR #8315, opened 2026-02-09
[MOD-12286] Introduce OpaqueRLookupRow

memark:
As part of, and in preparation for, the final migration of `RLookup` from C to Rust, the `RLookupRow` is being replaced by `OpaqueRLookupRow`, which basically hides all its implementation details.

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches Rust↔C FFI/ABI and generated headers for `RLookupRow`, so mismatched size/alignment or call signatures could cause runtime crashes despite being a mechanical refactor.
> 
> **Overview**
> **C ABI change:** `RLookupRow` is now exposed to C as an *opaque, fixed-size/aligned* struct (backed by new Rust `OpaqueRLookupRow`), hiding implementation details while remaining passable across the FFI boundary.
> 
> All `rlookup_ffi` entrypoints that accept rows were updated to take `OpaqueRLookupRow` pointers and convert internally, `cbindgen`/generated headers were updated (alignment macro, renames, additional exports), and FFI tests were extended to cover new row-copy/move invariants and panic conditions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3160aca2c7cc23b81ef53634f500ae9952c98638. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rlookup/src/row.rs
+++ b/src/redisearch_rs/rlookup/src/row.rs
@@ -16,30 +16,6 @@ use sorting_vector::RSSortingVector;
 use std::{borrow::Cow, ffi::CStr};
 use value::RSValueTrait;
 
-// This is a modified version of an inline of the existing `opaque!` macro, that unfortunately really doesn't work with lifetimes.
-// TODO: refactor to use one common macro, jira ticket MOD-13960
-pub mod opaque {
-    use super::RLookupRow;
-    use c_ffi_utils::opaque::Size;
-    use value::RSValueFFI;
-
-    /// An lookup row which can be passed by value to C.
-    ///
-    /// The size and alignment of this struct must match the Rust `RLookupRow`
-    /// structure exactly.
-    #[repr(C, align(8))]
-    pub struct OpaqueRLookupRow(OpaqueRLookupRowSize);
-
-    #[cfg(debug_assertions)]
-    type OpaqueRLookupRowSize = Size<48>;
-    #[cfg(not(debug_assertions))]
-    type OpaqueRLookupRowSize = Size<40>;
-
-    c_ffi_utils::opaque!(RLookupRow<'_, RSValueFFI>, OpaqueRLookupRow);
-}
-
-//
-
 /// Row data for a lookup key. This abstracts the question of if the data comes from a borrowed [RSSortingVector]
 /// or from dynamic values stored in the row during processing.
 ///
@@ -385,6 +361,26 @@ impl<'a, T: RSValueTrait> RLookupRow<'a, T> {
             );
         }
     }
+}
+
+pub mod opaque {
+    use super::RLookupRow;
+    use c_ffi_utils::opaque::Size;
+    use value::RSValueFFI;
+
+    /// An lookup row which can be passed by value to C.
+    ///
+    /// The size and alignment of this struct must match the Rust `RLookupRow`
+    /// structure exactly.
+    #[repr(C, align(8))]
+    pub struct OpaqueRLookupRow(OpaqueRLookupRowSize);
+
+    #[cfg(debug_assertions)]
+    type OpaqueRLookupRowSize = Size<48>;
+    #[cfg(not(debug_assertions))]
+    type OpaqueRLookupRowSize = Size<40>;
+
+    c_ffi_utils::opaque!(RLookupRow<'_, RSValueFFI>, OpaqueRLookupRow);
 }
 
 #[cfg(test)]


src/redisearch_rs/rlookup/src/row.rs
```diff
@@ -16,6 +16,82 @@ use sorting_vector::RSSortingVector;
...
+    const SIZE: usize = 40;
+
+    // Safety: `OpaqueRLookupRow` is defined as a `MaybeUninit` slice of
+    // bytes with the same size and alignment as `RLookupRow`, so any valid
+    // `OpaqueRLookupRow` has a bit pattern which is a valid `RLookupRow`.
+    unsafe impl<'a> Transmute<super::RLookupRow<'a, RSValueFFI>> for OpaqueRLookupRow<'a> {}
+
+    mod __opaque {
+        use super::*;
+        impl<'a> c_ffi_utils::opaque::IntoOpaque for super::super::RLookupRow<'a, value::RSValueFFI> {
```
memark: This is basically an inlined (and somewhat modified) version of the `opaque! `macro.

Should we
- leave it as is?
- extract to a new macro?
- (try again to) integrate into existing macro?

src/redisearch_rs/rlookup/src/row.rs
```diff
@@ -16,6 +16,82 @@ use sorting_vector::RSSortingVector;
...
+    ///
+    /// The size and alignment of this struct must match the Rust `RLookupRow`
+    /// structure exactly.
+    #[repr(C, align(8))]
+    pub struct OpaqueRLookupRow<'a>(Size<SIZE>, PhantomData<&'a ()>);
+
+    #[cfg(debug_assertions)]
+    const SIZE: usize = 48;
+    #[cfg(not(debug_assertions))]
+    const SIZE: usize = 40;
```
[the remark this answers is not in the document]
memark: Fixed.

src/redisearch_rs/headers/rlookup_rs.h
```diff
@@ -167,7 +169,29 @@ typedef struct RLookup {
...
+typedef uint8_t Size_48[48];
+
+/**
+ * A type with size `N`.
+ */
+typedef uint8_t Size_40[40];
+
+typedef Size_48 OpaqueRLookupRowSize;
+
+typedef Size_40 OpaqueRLookupRowSize;
```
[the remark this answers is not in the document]
memark: Ok, I'm out of ideas here. Any takes?

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
```diff
@@ -372,7 +373,7 @@ pub unsafe extern "C" fn RLookup_GetKey_LoadEx(
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn RLookup_GetLength(
     lookup: *const RLookup<'_>,
-    row: *const RLookupRow,
+    row: *const rlookup::OpaqueRLookupRow<'_>,
```
daxhuiberts: Is it a specific choice to use a qualified path here instead of just `OpaqueRLookupRow` (and everywhere else)?

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
```diff
@@ -284,8 +307,8 @@ pub unsafe extern "C" fn RLookupRow_WriteFieldsFrom<'a>(
     let dst_lookup = unsafe { dst_lookup.unwrap().as_mut() };
 
     // We're doing the asserts here in the middle to avoid extra type conversions.
-    assert_ne!(
-        src_row, dst_row,
+    assert!(
+        src_row != dst_row,
```
daxhuiberts: So this checks a `*const OpaqueRLookupRow` with a `&OpaqueRLookupRow`? Well, they are both eventually pointers so this likely works.

src/redisearch_rs/headers/rlookup_rs.h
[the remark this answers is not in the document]
daxhuiberts: Oh nasty.

The problem is even more fundamental: We have only one checked in header file, independent of a debug or release build. So this header should have something like a `#ifdef DEBUG_MODE` and `#else` to use the right type defs for the specific debug/release build.

Some options:
1. Dive into cbindgen to figure out how to make this possible.
2. Do some gnarly replace/renames on the generated header file to add those checks.
3. Change the `RLookupRow` so it doesn't have the `#[cfg(debug_assertions)]` which triggers this issue.
4. Add a `#[cfg(not(debug_assertions))] _filler: u64` to fill the struct up (awful idea) to the same debug size.

That's all I can think of right now.

src/redisearch_rs/rlookup/src/row.rs
daxhuiberts: Which parts are changed from the macro?

Your comment that this is basically an inlined version of the macro is good to add to the code as well.

Ideally this would be part of the macro if that's doable and readable.

memark: Discussed this with Jonas. I will add a comment as you suggested. Then it will be refactored later (separate ticket) to not stall things right now.

src/redisearch_rs/headers/rlookup_rs.h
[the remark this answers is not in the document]
memark: We think we found a way. Rust cfg can be translated to C #define. I just need to find the correct one, then I will push the solution.

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
memark: Yeah, it's a bit weird. Can you think of a better way to compare them?

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
memark: I think the reason was copy-paste together with some (now removed) duplicate types. Good catch!

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
```diff
@@ -352,17 +378,17 @@ pub unsafe extern "C" fn RLookupRow_GetSortingVector(
...
 #[unsafe(no_mangle)]
-pub const unsafe extern "C" fn RLookupRow_SetSortingVector(
-    row: Option<NonNull<RLookupRow>>,
+pub unsafe extern "C" fn RLookupRow_SetSortingVector(
+    row: Option<NonNull<OpaqueRLookupRow>>,
     sv: *const sorting_vector::RSSortingVector<RSValueFFI>,
 ) {
     // Safety: ensured by caller (1.)
-    let row = unsafe { row.unwrap().as_mut() };
+    let row = unsafe { RLookupRow::from_opaque_mut_ptr(row.unwrap().as_ptr()).unwrap() };
```
JonasKruckenberg: While were basically working with a "fork" of the macro anyways, lets just change the opaque methods from `from_opaque_mut_ptr` to `from_opaque_non_null` or similar so we can preserve nonull-ness all the way through

src/redisearch_rs/c_entrypoint/rlookup_ffi/tests/row.rs
```diff
@@ -49,8 +47,8 @@ fn rlookuprow_move() {
     unsafe {
         RLookupRow_MoveFieldsFrom(
             ptr::from_ref(&lookup),
-            Some(NonNull::from(&mut src)),
-            Some(NonNull::from(&mut dst)),
+            Some(NonNull::from(&mut src).cast::<OpaqueRLookupRow>()),
+            Some(NonNull::from(&mut dst).cast::<OpaqueRLookupRow>()),
```
JonasKruckenberg: The same here, instead of converting into a ptr and then casting we could have a `as_opaque_non_null` or similar

src/redisearch_rs/c_entrypoint/rlookup_ffi/cbindgen.toml
```diff
@@ -3,24 +3,36 @@ autogen_warning = "/* Warning, this file is auto-generated by cbindgen from `src
...
 [export.rename]
-"RLookupRow_RSValueFFI" = "RLookupRow"
+"OpaqueRLookupRow" = "RLookupRow"
 "RLookupHeader" = "RLookup"
 "RLookupKeyHeader" = "RLookupKey"
+"RLookupRow_RSValueFFI" = "RLookupRow"
+"RSSortingVector_RSValueFFI" = "RSSortingVector"
+
+[defines]
+"debug_assertions" = "ENABLE_ASSERT"
```
JonasKruckenberg, line 39: did you figure this is the correct one? would be curious to know how to check this

memark: I did, and it turns out that `CMake` [sets this](https://github.com/RediSearch/RediSearch/blob/master/CMakeLists.txt#L199) when doing a debug build. (I also did some grepping, and found nothing that seemed more appropriate.)

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
daxhuiberts: I just checked, and I think this is the best there is. Because of auto-coercion, `*const T == &T` works, but the other way around `&T == *const T` doesn't.

memark: Thanks. I added some tests for this, just to be sure.

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
memark: Done. I'm not sure it's better though, I had to put the method outside of the trait, and there is now no corresponding method for *const (since NonNull is always *mut). See what you think.

src/redisearch_rs/c_entrypoint/rlookup_ffi/cbindgen.toml
memark: Added a comment about that too.

src/redisearch_rs/c_entrypoint/rlookup_ffi/tests/row.rs
memark: Nice, this one makes it a bit more ergonomic to work with, if only for the tests.

src/redisearch_rs/c_entrypoint/rlookup_ffi/cbindgen.toml
daxhuiberts: Great find!

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
```diff
@@ -294,15 +300,20 @@ pub unsafe extern "C" fn RLookupRow_WriteFieldsFrom<'a>(
...
+    let dst_row = unsafe { RLookupRow::from_opaque_non_null(dst_row) };
 
     // Safety: ensured by caller (2.)
     let src_lookup = unsafe { src_lookup.as_ref().unwrap() };
 
     dst_row.copy_fields_from(dst_lookup, src_row, src_lookup, create_missing_keys);
 }
 
-/// Retrieves an item from the given `RLookupRow` based on the provided `RLookupKey`.
+// #[cfg(test)]
```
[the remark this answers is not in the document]
memark: Fixed.

src/redisearch_rs/rlookup/src/lookup.rs
```diff
@@ -425,7 +425,7 @@ impl<'a> RLookup<'a> {
     pub fn load_rule_fields(
         &mut self,
         search_ctx: &mut ffi::RedisSearchCtx,
-        dst_row: &mut RLookupRow<'a, value::RSValueFFI>,
```
JonasKruckenberg, line 428: I don't think we should be using the opaque type here 🤔 This is supposed to be the "final" Rust interface and the opaque type should only be a temp FFI-requirement type

src/redisearch_rs/rlookup/src/row.rs
JonasKruckenberg, line 1: you could actually try to rebase this onto the PR I made 🤔 we could even close that PR in favor of this

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
```diff
@@ -18,27 +19,26 @@ use std::{
...
-pub type RLookupRow<'a> = rlookup::RLookupRow<'a, RSValueFFI>;
-
 /// Writes a key to the row but increments the value reference count before writing it thus having shared ownership.
 ///
 /// # Safety
 ///
 /// 1. `key` must be a [valid], non-null pointer to an [`RLookupKey`].
-/// 2. `row` must be a [valid], non-null pointer to an [`RLookupRow`].
+/// 2. `row` must be a [valid], non-null pointer to an [`OpaqueRLookupRow`].
 /// 3. `value` must be a [valid], non-null pointer to an [`ffi::RSValue`].
```
JonasKruckenberg, line 27: In comments we should still refer to `RLookup` because the ident `OpaqueRLookupRow` doesn't even exist on the C side since we rename it.

src/redisearch_rs/rlookup/src/lookup.rs
memark: Eventually I'd agree with you, but right now, the only thing we use `dst_row` for is to pass it to `ffi::loadIndividualKeys` (which takes the opaque version).

src/redisearch_rs/rlookup/src/row.rs
memark: Which PR are you referring to?

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
memark: That makes sense. Updated the docs.

src/redisearch_rs/rlookup/src/lookup.rs
daxhuiberts: Explaining this bit as a code comment is helpful.

src/redisearch_rs/rlookup/src/row.rs
JonasKruckenberg: the macro PR but its fine, don't worry about it

PR #8345, opened 2026-02-11
[MOD-13993][Design] Iterator revalidation in Rust

LukeMathWalker:

A design document detailing how iterator suspension/revalidation should be modeled in the Rust type system to ensure soundness without compromising performance.

[Rendered document](https://github.com/RediSearch/RediSearch/blob/revalidation/docs/design/sound_iterator_revalidation.md).

1. MOD-...
2. #...

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Documentation-only change with no code or behavior impact.
> 
> **Overview**
> Adds a design document (`docs/design/sound_iterator_revalidation.md`) proposing a sound Rust model for query iterator revalidation across `IndexSpec` lock release/reacquire cycles, eliminating potential undefined behavior from holding Rust references while the inverted index can be mutated.
> 
> No runtime behavior changes are included; this PR is documentation-only and describes a planned `suspend`/`resume` approach for iterators and FFI interop.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b258092ec944747adbee65462d8818a78755baba. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/docs/design/sound_iterator_revalidation.md
+++ b/docs/design/sound_iterator_revalidation.md
@@ -1,0 +1,595 @@
+# Inverted Index Locking and Reader Design
+
+## IndexSpec Locking Scheme
+
+`IndexSpec` is protected by a [`pthread_rwlock_t`][spec-rwlock]. All access goes through
+wrappers in `redis_index.c`:
+
+- `RedisSearchCtx_LockSpecRead` — acquires the read lock, pauses dict rehashing on
+  `spec->keysDict`.
+- `RedisSearchCtx_LockSpecWrite` — acquires the write lock.
+- `RedisSearchCtx_UnlockSpec` — releases either lock, resumes dict rehashing if it was paused.
+
+The rwlock allows **multiple concurrent readers OR one exclusive writer**, never both.
+
+## The Lock-Release-Revalidate Lifecycle
+
+A query does **not** hold the read lock for its entire execution. The lock is released and
+re-acquired between batches of results. The lifecycle repeats:
+
+1. **Lock acquired** — [`handleSpecLockAndRevalidate`][handleSpecLock] acquires the
+   read lock and calls `it->Revalidate(it)` on the iterator tree.
+2. **Read phase** — the iterator reads records from inverted indexes. The read lock is held
+   throughout this phase.
+3. **Lock released** — once a batch of doc IDs is collected, the spec lock is released
+   (`RedisSearchCtx_UnlockSpec`). For example:
+   - [`rpSafeLoaderNext_Accumulate`][safeloader-unlock] releases the spec lock before
+     acquiring the Redis GIL to load document payloads.
+   - Cursor-based queries release the spec lock [after sending each chunk][cursor-unlock],
+     then pause the cursor.
+4. **Writes may happen** — while the lock is released, writers can acquire the write lock and
+   modify inverted indexes (append entries, run GC, increment `gc_marker`).
+5. **Lock re-acquired** — on the next call to `rpQueryItNext`, `handleSpecLockAndRevalidate`
+   detects `sctx->flags == RS_CTX_UNSET`, re-acquires the read lock, and calls `Revalidate`.
+
+### What Revalidate Does
+
+[`Revalidate`][revalidate-api] is a method on every `QueryIterator`. It propagates down the
+iterator tree (union, intersection, not, optional, etc.) to the leaf inverted index iterators.
+
+At the leaf level, [`InvIndIterator_Revalidate`][invind-revalidate] calls
+[`IndexReader_Revalidate`][reader-revalidate-rs], which checks the `gc_marker`:
+
+```rust
+fn needs_revalidation(&self) -> bool {
+    self.gc_marker != self.ii.gc_marker
+}
+```
+
+The reader stores a snapshot of the index's `gc_marker` at creation time. If GC (or any write)
+has incremented the index's marker, the reader knows the index was modified. In that case:
+
+1. The reader's byte offsets into blocks may be stale.
+2. The iterator **rewinds** and **re-seeks** to its `lastDocId`.
+3. If that exact docId no longer exists (deleted by GC), it lands on the next valid one →
+   `VALIDATE_MOVED`.
+
+Even when `needs_revalidation` returns false, `refresh_buffer_pointers` is called — blocks may
+have been reallocated by appending new entries, so the `Cursor`'s internal pointer needs
+refreshing.
+
+Other iterators which sit on top of the inverted index reader may also require additional
+revalidation steps before resuming iteration.
+
+### ValidateStatus
+
+```c
+typedef enum ValidateStatus {
+    VALIDATE_OK,      // Iterator still valid, same position
+    VALIDATE_MOVED,   // Iterator still valid, but moved forward
+    VALIDATE_ABORTED, // Iterator invalid, must be freed
+} ValidateStatus;
+```
+
+Composite iterators handle these per child:
+
+- **Union** — aborted children are removed, remaining children continue.
+- **Intersection** — any aborted child aborts the whole intersection.
+- **Not** — aborted child is replaced with an empty iterator (NOT nothing = everything).
+- **Optional** — aborted child is replaced with an empty iterator.
+
+## Problem: Current Rust Modeling
+
+[`IndexReaderCore`][reader-core] holds a `&'index InvertedIndex<E>`:
+
+```rust
+pub struct IndexReaderCore<'index, E> {
+    ii: &'index InvertedIndex<E>,
+    current_buffer: Cursor<&'index [u8]>,
+    current_block_idx: usize,
+    last_doc_id: t_docId,
+    gc_marker: u32,
+}
+```
+
+At the FFI boundary ([`NewIndexReader`][new-index-reader]), this reference is created from a
+raw pointer and the lifetime is erased via `Box::into_raw`:
+
+```rust
+let ii = unsafe { &*ii };          // fabricate &InvertedIndex
+let reader = Box::new(/* ... */);
+Box::into_raw(reader)              // erase the lifetime
+```
+
+The C code holds the resulting `*mut IndexReader` **across lock release/reacquire cycles**.
+During those windows, a writer can mutate the `InvertedIndex` (append to blocks, run GC,
+increment `gc_marker`). This means a `&InvertedIndex` exists while the referent is being
+mutated — **undefined behavior** under Rust's aliasing rules, regardless of whether the
+accesses are serialized by the rwlock.
+
+Concretely, after a lock release:
+- `gc_marker` may have been incremented.
+- `blocks: ThinVec<IndexBlock>` may have been modified (blocks removed by GC, block buffers
+  grown/reallocated by appending).
+- `n_unique_docs` may have been updated.
+
+The `refresh_buffer_pointers` method acknowledges this — it re-derives the `Cursor` from
+`self.ii.blocks[self.current_block_idx].buffer`, reading through a `&InvertedIndex` that may
+have been mutated. This is the UB in action.
+
+## Solutions Considered
+
+### Option 1: Raw Pointer
+
+Replace `ii: &'index InvertedIndex<E>` with `ii: *const InvertedIndex<E>`. Every access goes
+through `unsafe { &*self.ii }` with a safety comment documenting that the caller holds the read
+lock.
+
+**Pros:** Honest about what's happening, no false aliasing guarantees.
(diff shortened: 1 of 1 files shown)


docs/design/sound_iterator_revalidation.md
```diff
@@ -0,0 +1,590 @@
...
+            // Index was modified (GC or writes). Rewind to block 0 and re-seek.
+            let mut active = ActiveReader {
+                ii,
+                buf: &ii.blocks[0].buffer,
+                buf_pos: 0,
+                current_block_idx: 0,
+                last_doc_id: 0,
+                gc_marker: ii.gc_marker,
+                _phantom: PhantomData,
+            };
```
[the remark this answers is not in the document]
LukeMathWalker: Not relevant from a design perspective. The code block is only illustrative.

docs/design/sound_iterator_revalidation.md
```diff
@@ -0,0 +1,590 @@
...
+5. **Lock re-acquired** — on the next call to `rpQueryItNext`, `handleSpecLockAndRevalidate`
+   detects `sctx->flags == RS_CTX_UNSET`, re-acquires the read lock, and calls `Revalidate`.
+
+### What Revalidate Does
+
+[`Revalidate`][revalidate-api] is a method on every `QueryIterator`. It propagates down the
+iterator tree (union, intersection, not, optional, etc.) to the leaf inverted index iterators.
+
+At the leaf level, [`InvIndIterator_Revalidate`][invind-revalidate] calls
+[`IndexReader_Revalidate`][reader-revalidate-rs], which checks the `gc_marker`:
```
gdesmott, line 41: Each II iterator also has extra logic that may abort the revalidation, see `CheckAbort`

docs/design/sound_iterator_revalidation.md
```diff
@@ -0,0 +1,590 @@
...
+struct Active<'a>(PhantomData<&'a ()>);
+struct Suspended;
+
+impl<'a> Ref for Active<'a> {
+    type Ptr<T: ?Sized> = &'a T;
+}
+
+impl Ref for Suspended {
+    type Ptr<T: ?Sized> = *const T;
+}
```
chesedo, line 189: Where does `T` come from?

docs/design/sound_iterator_revalidation.md
```diff
@@ -0,0 +1,590 @@
...
+
+```rust
+/// Abstracts over reference-like pointer types.
+///
+/// Two implementations exist:
+/// - `Active<'a>` — uses `&'a T`, for use while the index read lock is held.
+/// - `Suspended` — uses `*const T`, for use while the lock is released.
+///
+/// Both produce layout-compatible pointer types, so `ReaderCore<Active<'a>, E>`
+/// and `ReaderCore<Suspended, E>` are transmutable.
```
gdesmott, line 175: We should make sure we have a test enforcing this.

docs/design/sound_iterator_revalidation.md
```diff
@@ -0,0 +1,590 @@
...
+    buf: R::Ptr<[u8]>,
+    buf_pos: u64,
+    current_block_idx: usize,
+    last_doc_id: t_docId,
+    gc_marker: u32,
+    _phantom: PhantomData<E>,
+}
+
+type ActiveReader<'a, E> = ReaderCore<Active<'a>, E>;
+type SuspendedReader<E> = ReaderCore<Suspended, E>;
```
chesedo, line 215: This also means stripping all (most) methods from `ReaderCore` right? That way they won't be accessible / possible for a `SuspendedReader`

docs/design/sound_iterator_revalidation.md
```diff
@@ -0,0 +1,590 @@
...
+The C code holds a single `*mut IndexReader` across lock boundaries. At the FFI level, the
+reader is always stored as a `SuspendedReader` (the default state when no lock is held).
+
+```rust
+// FFI reader enum (existing pattern, unchanged structure)
+pub enum IndexReader {
+    Full(SuspendedReader<Full>),
+    // ... other variants
+}
+```
```
chesedo: This means every `IndexReader_Next` will have to do the full transmute (`'resume`). `Next` is a hot path, will this overhead be acceptable?

Maybe the `Acquire` functions can instead `resume` the reader? I remember this seemed possible last time. But might need a bit of refactor to move the reader down to the acquire or something.

docs/design/sound_iterator_revalidation.md
```diff
@@ -0,0 +1,590 @@
...
+
+FFI functions that need to read (called while the lock is held) transmute to `ActiveReader`
+for the duration of the call. `IndexReader_Revalidate` performs `resume`. The lock
```
gdesmott: So that means we'll have to call `Revalidate` and transmute for each read call from C? While the actual C code only do it once and then can do a bunch of reads? Isn't this a potentially big performance regression?

docs/design/sound_iterator_revalidation.md
```diff
@@ -0,0 +1,590 @@
...
+    type Suspended: SuspendedRQEIterator<Active<'index> = Self>;
+
+    /// Consume the active iterator, producing a suspended version.
+    /// All `&'index` references become raw pointers.
+    fn suspend(self) -> Self::Suspended;
+
+    // --- existing methods (unchanged) ---
+    fn read(&mut self) -> Result<Option<&mut RSIndexResult<'index>>, RQEIteratorError>;
+    fn skip_to(&mut self, doc_id: t_docId) -> ...;
+    fn revalidate(&mut self) -> Result<RQEValidateStatus<'_, 'index>, RQEIteratorError>;
```
gdesmott: Do we still need `revalidate()`? When will it be called on an active iterator?

docs/design/sound_iterator_revalidation.md
```diff
@@ -0,0 +1,590 @@
...
+| `Empty` / `IdList` / `Metric` / `Wildcard` | No-op (identity) | No-op (identity) |
+
+## `Ref` Propagation Through the Iterator Hierarchy
+
+The `Ref` parameter cannot be confined to `ReaderCore`. Suspending the reader changes its
+type from `ReaderCore<Active<'a>, E>` to `ReaderCore<Suspended, E>`. Since `InvIndIterator`
+stores `reader: R`, the reader type change forces the iterator's type to change too.
+This propagates up through every struct that directly or transitively contains a reader.
+
+### Propagation chain
```
gdesmott, line 494: A bit unfortunate to have to propagate it so deep but I guess that's the price to pay for proper modeling.

docs/design/sound_iterator_revalidation.md
```diff
@@ -0,0 +1,590 @@
...
+| Struct | Field | Points to |
+|--------|-------|-----------|
+| `ReaderCore` | `ii` | `InvertedIndex<E>` (the index itself) |
+| `ReaderCore` | `buf` | `[u8]` (current block buffer) |
+| [`RSOffsetVector`][offset-vector] | `data` | `[u8]` (offsets within block buffer, future) |
+
+**References to other iterator results** — point to `RSIndexResult` fields inside child
+iterators, which are heap-allocated and don't move. These remain valid regardless of lock
+state — children survive across suspend and their result fields are not mutated by
+external writers:
```
gdesmott, line 536: So those in the suspended state will still hold references, not pointers?

docs/design/sound_iterator_revalidation.md
LukeMathWalker: Yes, they would go on the active version.

docs/design/sound_iterator_revalidation.md
LukeMathWalker: It would be possible to make them hold references to child results by having two generic parameters on `RSIndexResult`—the `'index` lifetime and an `R` which would be parametrised via `Ref`. Nonetheless, I don't think it's worthwhile to make that distinction unless we do find ourselves working with the child results while in a suspended state.

docs/design/sound_iterator_revalidation.md
LukeMathWalker: `Ptr` is a generic associated type parameter. It needs to specified when `Ref` is used (e.g. see the definition later on for `ReaderCore).

LukeMathWalker:
We definitely don't want to be calling `resume`/`Revalidate` on every `Read`, since that'd be a perf killer. 
The wording in the FFI section was unclear—I have clarified what I originally meant (i.e. we can transmute to either variant since they are layout equivalent, embedding in the FFI `# Safety` section the relevant prerequisites).

docs/design/sound_iterator_revalidation.md
LukeMathWalker: https://github.com/RediSearch/RediSearch/pull/8345#issuecomment-3890286644

LukeMathWalker:
> I'm just wondering about `resume()` not actually holding the lock guard. So one could cheat and call it without actually holding the lock. Could we enforce in the type system that the caller actually hold the lock to be able to call `suspend()`?

I considered the idea, but paused on it for the time being—in tests, we don't really have a lock, and, in the end, it isn't required either. The only safety precondition we care about is making sure that, when `resume` is called, there are no `&mut` references to the index around.

LukeMathWalker:
> Looks good!
> 
> Quick thing, how do we know the `Active` / `Suspended` structs solve our test issue? Will the revalidate tests just move between these during the test?

Yup!
Create an index, create a reader, do things with the reader, suspend the reader, take a mutable reference to the index, do nasty things (i.e. increase the gc marker), drop the mutable reference, resume the reader, test revalidation.

docs/design/sound_iterator_revalidation.md
```diff
@@ -0,0 +1,595 @@
...
+    ///   have been freed by GC or reallocated by appending — a dangling
+    ///   reference is UB regardless of whether it is dereferenced. Instead,
+    ///   `buf` is derived fresh from the validated index state.
+    ///
+    /// # Safety
+    ///
+    /// The caller must hold the read lock on the IndexSpec that owns `ii`.
+    unsafe fn resume<'a>(self) -> (ActiveReader<'a, E>, ValidateStatus) {
+        // SAFETY: caller holds the read lock, so the InvertedIndex is valid.
+        let ii: &'a InvertedIndex<E> = unsafe { &*self.ii };
```
GuyAv46, line 254: What if the inverted index was dropped (and maybe recreated)? Or was this validated before the call to `resume`?

docs/design/sound_iterator_revalidation.md
```diff
@@ -0,0 +1,595 @@
...
+survives across this unlock. The lock is re-acquired on the next batch via
+`handleSpecLockAndRevalidate`.
+
+### Suspend point 2: `runCursor`
+
+[`aggregate_exec.c:1225-1233`](https://github.com/RediSearch/RediSearch/blob/0c1a6b196ae97c19a69ac467113c690130cf6cb9/src/aggregate/aggregate_exec.c#L1225-L1233)
+
+```c
+sendChunk(req, reply, num);
+RedisSearchCtx_UnlockSpec(AREQ_SearchCtx(req));      // <-- SUSPEND before this
```
GuyAv46, line 380: `RedisSearchCtx_UnlockSpec` is safe to call multiple times. In this case, for example, we may already released the lock at a safe loader. We should account for that naturally somehow (suspend on suspended iterator is OK)

LukeMathWalker: Re-suspending should be a no-op, if needed.

docs/design/sound_iterator_revalidation.md
LukeMathWalker: This doesn't seem to be a case we account for in revalidation right now—cc @gdesmott?
We can design mechanisms to prevent a drop if the current design doesn't guarantee it (e.g., using a `Weak` reference).

gdesmott: Actually we do, see the `should_abort` of the II wildcard: https://github.com/RediSearch/RediSearch/pull/8390/changes#diff-cbccacf44cd0566ab6d6d3e58f3e9bd98aee26286b29640f32b4a0991e814930R635

We actually found a crash in the existing C code if the II was fully garbage collected while we have a II wildcard iterator: https://github.com/RediSearch/RediSearch/pull/8331

LukeMathWalker: Cool, I was only looking at it at the `IndexReader` level 👍🏻

LukeMathWalker:
Anything missing here @GuyAv46 or can we merge?

GuyAv46:
@LukeMathWalker nothing to add beside the previous 2 comments. We can continue

PR #8360, opened 2026-02-12
[MOD-13967] Add missing RLookupKey FFI functions

memark:
Add missing RLookupKey FFI functions

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Adds read-only FFI accessors and header declarations without changing lookup behavior; main risk is ABI/FFI compatibility or misuse of returned pointers by callers.
> 
> **Overview**
> Adds the missing `RLookupKey` C FFI accessors, enabling C/C++ code to retrieve a key’s `flags`, destination/sorting-vector indexes, `name`, `name_len`, and `path` via stable functions rather than directly reading struct fields.
> 
> Updates the generated `rlookup_rs.h` to declare these new APIs and adjusts type placement (moving the opaque `RLookup` typedef) to keep the header consistent with the new exports.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2b0ed06fe4a0223e856389491ea02f674bdb694e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/rlookup_ffi/src/key.rs
+++ b/src/redisearch_rs/c_entrypoint/rlookup_ffi/src/key.rs
@@ -71,7 +71,7 @@ pub unsafe extern "C" fn RLookupKey_GetName(key: *const RLookupKey) -> *const c_
     // Safety: ensured by caller (1.)
     let key = unsafe { key.as_ref().unwrap() };
 
-    key.name().as_ptr()
+    key.name
 }
 
 /// Get the length of the name field in bytes.
@@ -86,7 +86,7 @@ pub unsafe extern "C" fn RLookupKey_GetNameLen(key: *const RLookupKey) -> size_t
     // Safety: ensured by caller (1.)
     let key = unsafe { key.as_ref().unwrap() };
 
-    key.name().count_bytes()
+    key.name_len
 }
 
 /// Get the path of the field.
@@ -101,5 +101,9 @@ pub unsafe extern "C" fn RLookupKey_GetPath(key: *const RLookupKey) -> *const c_
     // Safety: ensured by caller (1.)
     let key = unsafe { key.as_ref().unwrap() };
 
-    key.path().as_ref().unwrap_or(key.name()).as_ptr()
+    if key.path.is_null() {
+        key.name
+    } else {
+        key.path
+    }
 }


src/redisearch_rs/c_entrypoint/rlookup_ffi/src/key.rs
```diff
@@ -0,0 +1,95 @@
...
+///
+/// # Safety
+///
+/// 1. `key` must be a [valid], non-null pointer to an [`RLookupKey`].
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RLookupKey_GetPath(key: *const RLookupKey) -> *const c_char {
+    // Safety: ensured by caller (1.)
+    let key = unsafe { key.as_ref().unwrap() };
+
+    key.path
```
memark: I'm confused (as always) whether to return `.path` or `._path` here?

daxhuiberts: How does it work that `path` exists while the field is called `_path`?

memark: It's a bit of a mess, but tldr, they both exist:

image

daxhuiberts: Ah check, that is indeed confusing.

Regarding which one to use, better to ask @JonasKruckenberg.

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/key.rs
```diff
@@ -0,0 +1,95 @@
...
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use libc::size_t;
+use rlookup::RLookupKey;
+use std::ffi::c_char;
+
+/// Get the flags (indicating the type and other attributes) for a `RLookupKey`.
+///
+/// The flag F_SVSRC means the target array is a sorting vector.
```
JonasKruckenberg:
```suggestion
/// The flag `RLookupKeyFlag::SvSrc` means the target array is a sorting vector.
```

honestly don't really know why the C version thought this comment is important to have but we should at least make sure its referencing the right type

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/key.rs
JonasKruckenberg: > How does it work that `path` exists while the field is called `_path`?

`Deref` is great 🙃 

Since both fields hold the same pointer at the end of the day it would be fine either way. Let's use the "proper" `_path` field though since `path` and `name` really only exist for interop with C and will be gone in a future refactor

JonasKruckenberg (commented):
looking good!

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/key.rs
memark: I'd be happy to remove it if you don't see the need for it either.

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/key.rs
memark: Actually, I'm afraid `Deref` doesn't really help here since those fields are private.

But we already have `name()` and `path() `in place, which returns the underlying underscored fields. I'll use those.

memark: I'm not sure if this was correct. _After_ this change (using `name()` and `path()`) 2 _more_ CPP unit tests start failing:
>499 - RLookupTest.testAddKeysFromConflictsOverride (Failed)
>510 - RLookupTest.testAddKeysFromHiddenFlagHandling (Failed)

daxhuiberts: > Since both fields hold the same pointer at the end of the day it would be fine either way. Let's use the "proper" _path field though since path and name really only exist for interop with C and will be gone in a future refactor

This raises more questions than clears things up, but probably because I don't know the details here. So the `RLookupKeyHeader` is read out by C code directly?

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/key.rs
```diff
@@ -0,0 +1,105 @@
...
+/// 1. `key` must be a [valid], non-null pointer to an [`RLookupKey`].
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RLookupKey_GetName(key: *const RLookupKey) -> *const c_char {
+    // Safety: ensured by caller (1.)
+    let key = unsafe { key.as_ref().unwrap() };
+
+    key.name().as_ptr()
+}
```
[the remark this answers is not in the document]
JonasKruckenberg: you are incorrect

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/key.rs
```diff
@@ -0,0 +1,95 @@
...
+///
+/// # Safety
+///
+/// 1. `key` must be a [valid], non-null pointer to an [`RLookupKey`].
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RLookupKey_GetPath(key: *const RLookupKey) -> *const c_char {
+    // Safety: ensured by caller (1.)
+    let key = unsafe { key.as_ref().unwrap() };
+
+    key.path().as_ref().map_or(ptr::null(), |k| k.as_ptr())
```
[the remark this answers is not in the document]
JonasKruckenberg: we should actually fall back to `_name` here I think, the bot might have a point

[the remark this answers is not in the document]
memark: I can do that, np. Do note the comment above about more tests failing. But I guess we can address that as part of the migration.

https://github.com/RediSearch/RediSearch/pull/8360#discussion_r2811478958

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/key.rs
memark: With a couple other PRs coming in, no, it will not be exposed (anymore). Both RLookup and RLookupRow will be opaque.

PR #8363, opened 2026-02-12
MOD-13361: Add Tag Index and Iterator Disk API

kei-nan:
Add an api to open a tag inverted index reader and use it in TagIndex_OpenReader().

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches core indexing and query execution paths (tag indexing, tag lex-range/prefix expansion) and introduces new disk API surface area; failures could impact correctness or stability for disk-mode queries.
> 
> **Overview**
> Enables `TAG` fields for disk-backed indexes by extending the disk API with tag indexing (`SearchDisk_IndexTags`) and a tag query iterator (`SearchDisk_NewTagIterator`), and wiring these into tag indexing and query evaluation.
> 
> `TagIndex` is refactored to operate in *memory* or *disk* mode (storing NULL sentinels in the `TrieMap` for disk mode), with updated `TagIndex_Open`/`NewTagIndex`/`TagIndex_Index` and new helper `TagIndex_GetIteratorFromTrieMapValue` to support prefix/lex-range expansions. Schema validation is loosened to allow `TAG` fields in disk mode, while debug tag index dump/info commands explicitly reject disk-mode tag indexes; related call sites and tests are updated, and fulltext disk indexing wrapper is renamed to `SearchDisk_IndexTerm`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f3988e4a740793a8efd9dc955a415ee2e9d20a2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/debug_commands.c
+++ b/src/debug_commands.c
@@ -1167,7 +1167,7 @@ DEBUG_COMMAND(InfoTagIndex) {
     goto end;
   }
 
-  const TagIndex *idx = TagIndex_Open(fs, DONT_CREATE_INDEX, NULL);
+  const TagIndex *idx = TagIndex_Open(fs, DONT_CREATE_INDEX, sctx->spec->diskSpec);
 
   // Field was not initialized yet
   if (!idx) {
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -341,11 +341,7 @@ static void FGC_childCollectNumeric(ForkGC *gc, RedisSearchCtx *sctx) {
 }
 
 static void FGC_childCollectTags(ForkGC *gc, RedisSearchCtx *sctx) {
-  // Right now we only collect tags from non-disk indexes
-  if (sctx->spec->diskSpec) {
-    FGC_sendTerminator(gc);
-    return;
-  }
+  RS_ASSERT(sctx->spec->diskSpec == NULL);
   arrayof(FieldSpec*) tagFields = getFieldsByType(sctx->spec, INDEXFLD_T_TAG);
   if (array_len(tagFields) != 0) {
     for (int i = 0; i < array_len(tagFields); ++i) {


src/search_disk_api.h
```diff
@@ -82,7 +83,22 @@ typedef struct IndexDiskAPI {
...
+   * Used for tag field indexing.
+   *
+   * @param index Pointer to the index
+   * @param values Array of tag values to associate the document with
+   * @param numValues Number of tag values in the array
+   * @param docId Document ID to index
+   * @param fieldIndex Field index for the tag field
+   * @return true if the write was successful, false otherwise
+   */
+  bool (*indexTags)(RedisSearchDiskIndexSpec *index, const char **values, size_t numValues, t_docId docId, t_fieldIndex fieldIndex);
```
JoanFM, line 103: how is the fieldIndex going to make sure on RediSearchDisk side?

src/tag_index.c
```diff
@@ -238,6 +240,14 @@ static QueryIterator *TagIndex_GetReader(const TagIndex *idx, const RedisSearchC
...
 
+  // Check if we're using disk storage
+  if (sctx && sctx->spec && sctx->spec->diskSpec) {
+    // Create token and call disk API
+    RSToken tok = {.str = (char *)value, .len = len};
+    // For tags, we use tokenId 0
+    return SearchDisk_NewTagIterator(sctx->spec->diskSpec, &tok, 0, fieldIndex, weight);
+  }
+
   InvertedIndex *iv = TrieMap_Find(idx->values, (char *)value, len);
```
JoanFM, line 296: should we consider what to add in TrieMap at indexing time?

src/document.c
```diff
@@ -758,6 +759,15 @@ FIELD_PREPROCESSOR(tagPreprocessor) {
 }
 
 FIELD_BULK_INDEXER(tagIndexer) {
+  // Check if we're using disk storage
+  if (ctx->spec->diskSpec) {
+    // Index tags using disk API
+    SearchDisk_IndexTags(ctx->spec->diskSpec, (const char **)fdata->tags, array_len(fdata->tags), aCtx->doc->docId, fs->index);
```
JoanFM: We should also consider this:

      if (idx->suffix && (*tok != '\0')) { // add to suffix TrieMap
        addSuffixTrieMap(idx->suffix, tok, strlen(tok));
      }

src/document.c
```diff
@@ -758,6 +759,15 @@ FIELD_PREPROCESSOR(tagPreprocessor) {
...
+    ctx->spec->stats.numRecords++;
```
JoanFM: also, if tagIndexing fail, we should not update numRecords

src/search_disk_api.h
kei-nan: I was planning on relying on the field index consistency to generate the column family name.
e.g
tag_1
tag_3
tag_5
otherwise I need an open api to pass the field name
and then it will be
tag_foo
tag_goo
tag_zoo

src/tag_index.c
```diff
@@ -238,6 +240,14 @@ static QueryIterator *TagIndex_GetReader(const TagIndex *idx, const RedisSearchC
 QueryIterator *TagIndex_OpenReader(TagIndex *idx, const RedisSearchCtx *sctx, const char *value, size_t len,
                                    double weight, t_fieldIndex fieldIndex) {
 
+  // Check if we're using disk storage
+  if (sctx && sctx->spec && sctx->spec->diskSpec) {
+    // Create token and call disk API
+    RSToken tok = {.str = (char *)value, .len = len};
+    // For tags, we use tokenId 0
+    return SearchDisk_NewTagIterator(sctx->spec->diskSpec, &tok, 0, fieldIndex, weight);
+  }
```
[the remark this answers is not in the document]
kei-nan: Good catch! Fixed this by checking for disk mode in `Query_EvalTagNode` before calling `TagIndex_Open`. Now when in disk mode, we pass `NULL` for `idx` and let `TagIndex_OpenReader` handle the disk path directly.

src/tag_index.c
kei-nan: Good point about the TrieMap. For now, I've added a TODO comment noting that suffix trie support is currently only available in memory mode. We can address disk-based suffix trie support in a follow-up if needed.

The in-memory path still handles suffix tries correctly via `TagIndex_Index`.

src/document.c
kei-nan: Added a TODO comment for suffix trie support in disk mode. This is a known limitation that can be addressed in future work when we implement disk-based suffix tries.

src/document.c
kei-nan: Fixed! Now checking the return value of `SearchDisk_IndexTags` and only incrementing `numRecords` if indexing succeeds. Returns an error if indexing fails.

src/document.c
kei-nan: in the end I tried supporting it.
incorporated the logic into TagIndex structure and functions

src/fork_gc.c
```diff
@@ -344,7 +344,7 @@ static void FGC_childCollectTags(ForkGC *gc, RedisSearchCtx *sctx) {
   arrayof(FieldSpec*) tagFields = getFieldsByType(sctx->spec, INDEXFLD_T_TAG);
   if (array_len(tagFields) != 0) {
     for (int i = 0; i < array_len(tagFields); ++i) {
-      TagIndex *tagIdx = TagIndex_Open(tagFields[i], DONT_CREATE_INDEX);
+      TagIndex *tagIdx = TagIndex_Open(tagFields[i], DONT_CREATE_INDEX, NULL);
```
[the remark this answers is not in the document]
kei-nan: gc is not running in disk.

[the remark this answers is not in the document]
kei-nan: in the end addressed to silence future comments

src/search_disk_api.h
Itzikvaknin: Can you mention this in the code as a comment? that this issue is still open

src/tag_index.c
```diff
@@ -28,11 +30,13 @@ static uint32_t tagUniqueId = 0;
...
 #define MAX_TAG_LEN 0x1000
 /* See tag_index.h for documentation  */
-TagIndex *NewTagIndex() {
+TagIndex *NewTagIndex(RedisSearchDiskIndexSpec *diskSpec, t_fieldIndex fieldIndex) {
   TagIndex *idx = rm_new(TagIndex);
   idx->values = NewTrieMap();
   idx->uniqueId = tagUniqueId++;
   idx->suffix = NULL;
+  idx->diskSpec = diskSpec;
+  idx->fieldIndex = fieldIndex;
```
Itzikvaknin, line 38: Can assert on `!diskSpec && fieldIndex == expected_value`?

src/tag_index.c
```diff
@@ -209,20 +213,42 @@ static inline size_t tagIndex_Put(TagIndex *idx, const char *value, size_t len,
...
 
-      if (idx->suffix && (*tok != '\0')) { // add to suffix TrieMap
-        addSuffixTrieMap(idx->suffix, tok, strlen(tok));
+    // Also populate TrieMap with NULL sentinels for tag enumeration
+    for (size_t ii = 0; ii < n; ++ii) {
+      const char *tok = values[ii];
+      if (tok) {
+        TrieMap_Add(idx->values, tok, strlen(tok), NULL, NULL);
+
+        if (idx->suffix && (*tok != '\0')) {
```
Itzikvaknin, line 230: Why do we need this `(*tok != '\0')`? 
And if tok is empty string, why do we add it in line 229?

src/tag_index.c
```diff
@@ -209,20 +213,42 @@ static inline size_t tagIndex_Put(TagIndex *idx, const char *value, size_t len,
...
 /* Index a vector of pre-processed tags for a docId */
-size_t TagIndex_Index(TagIndex *idx, const char **values, size_t n, t_docId docId) {
-  if (!values) return 0;
-  size_t ret = 0;
-  for (size_t ii = 0; ii < n; ++ii) {
-    const char *tok = values[ii];
-    if (tok) {
-      ret += tagIndex_Put(idx, tok, strlen(tok), docId);
+bool TagIndex_Index(TagIndex *idx, const char **values, size_t n, t_docId docId, IndexStats *stats) {
+  if (!values) return true;
```
Itzikvaknin, line 216: I see that the only place we call this function is 'document.c` 
```
if (!TagIndex_Index(tidx, (const char **)fdata->tags, array_len(fdata->tags), aCtx->doc->docId, &ctx->spec->stats)) {
```
I think we can remove `size_t n,`

src/search_disk.c
```diff
@@ -80,9 +80,14 @@ int SearchDisk_IndexSpecRdbLoad(RedisModuleIO *rdb, RedisSearchDiskIndexSpec *in
 }
 
 // Index API wrappers
-bool SearchDisk_IndexDocument(RedisSearchDiskIndexSpec *index, const char *term, size_t termLen, t_docId docId, t_fieldMask fieldMask, uint32_t freq) {
+bool SearchDisk_IndexText(RedisSearchDiskIndexSpec *index, const char *term, size_t termLen, t_docId docId, t_fieldMask fieldMask, uint32_t freq) {
```
Itzikvaknin: Do we use text for term or text for document? Its a bit confusing

src/fork_gc.c
```diff
@@ -344,11 +344,16 @@ static void FGC_childCollectTags(ForkGC *gc, RedisSearchCtx *sctx) {
...
       if (!tagIdx) {
         continue;
       }
 
+      // Skip disk-mode tag indexes - GC not applicable (TrieMap contains NULL sentinels)
+      if (tagIdx->diskSpec) {
```
Itzikvaknin: Do we expect this to change between iterations?

Itzikvaknin (commented):
Looks good, left a few comments

src/tag_index.c
kei-nan: well in normal flow diskSpec can be null.
it will only be non null in disk mode.
what is the expected value for a field index?

src/search_disk.c
kei-nan: Not sure I understand,
We have a text field type
this is meant to add a word/term to the text disk inverted index structure.

src/fork_gc.c
kei-nan: not really, will move it to start of function

src/tag_index.c
kei-nan: this is also used in cpp tests.
better to keep as is.

src/search_disk_api.h
kei-nan: In the end I'm planning on having the field index as a prefix and not as a cf name.
Not sure implementation details should be documented in the api declaration side.

src/tag_index.c
kei-nan: Asked augment to answer
The rationale:

Main TrieMap (line 229): Empty strings are allowed because they're technically valid for exact-match queries (e.g., @tag:{""}), even if rare.
Suffix trie (line 231): Empty strings are skipped because suffix/contains queries on an empty string are meaningless - there's nothing to match suffixes against.

src/tag_index.c
Itzikvaknin: Im not sure thats a good reason to not use the array, as the cpp test are our main framework for unit/integration tests
Your call

src/tag_index.c
Itzikvaknin: Good to learn new things :)

src/search_disk.c
Itzikvaknin: Lets go with term

src/fork_gc.c
```diff
@@ -341,10 +341,15 @@ static void FGC_childCollectNumeric(ForkGC *gc, RedisSearchCtx *sctx) {
 }
 
 static void FGC_childCollectTags(ForkGC *gc, RedisSearchCtx *sctx) {
+  // Right now we only collect tags from non-disk indexes
+  if (sctx->spec->diskSpec) {
```
JoanFM: right now GC is disabled with SearchDisk, so this is not needed

src/module.c
```diff
@@ -528,7 +528,7 @@ int TagValsCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
     goto cleanup;
   }
 
-  TagIndex *idx = TagIndex_Open(fs, DONT_CREATE_INDEX);
+  TagIndex *idx = TagIndex_Open(fs, DONT_CREATE_INDEX, NULL);
```
JoanFM, line 531: why don't we pass it? Wouln't it be a free feature?

src/debug_commands.c
```diff
@@ -615,14 +615,20 @@ DEBUG_COMMAND(DumpTagIndex) {
...
-  const TagIndex *tagIndex = TagIndex_Open(fs, DONT_CREATE_INDEX);
+  const TagIndex *tagIndex = TagIndex_Open(fs, DONT_CREATE_INDEX, NULL);
 
   // Field was not initialized yet
   if (!tagIndex) {
     RedisModule_ReplyWithEmptyArray(sctx->redisCtx);
     goto end;
   }
 
+  // Debug dump not supported for disk-mode tag indexes (TrieMap contains NULL sentinels)
```
JoanFM, line 626: and what is the difference with others? I do not understand why this makes it special?

src/debug_commands.c
```diff
@@ -615,14 +615,20 @@ DEBUG_COMMAND(DumpTagIndex) {
     RedisModule_ReplyWithError(sctx->redisCtx, "Could not find given field in index spec");
     goto end;
   }
-  const TagIndex *tagIndex = TagIndex_Open(fs, DONT_CREATE_INDEX);
+  const TagIndex *tagIndex = TagIndex_Open(fs, DONT_CREATE_INDEX, NULL);
```
JoanFM: wouldn't pass it diskSpec give this feature for free?

src/debug_commands.c
```diff
@@ -692,7 +698,7 @@ DEBUG_COMMAND(DumpSuffix) {
       RedisModule_ReplyWithError(sctx->redisCtx, "Could not find given field in index spec");
       goto end;
     }
-    const TagIndex *idx = TagIndex_Open(fs, DONT_CREATE_INDEX);
+    const TagIndex *idx = TagIndex_Open(fs, DONT_CREATE_INDEX, NULL);
```
JoanFM: same here

src/debug_commands.c
```diff
@@ -1153,14 +1159,20 @@ DEBUG_COMMAND(InfoTagIndex) {
     goto end;
   }
 
-  const TagIndex *idx = TagIndex_Open(fs, DONT_CREATE_INDEX);
+  const TagIndex *idx = TagIndex_Open(fs, DONT_CREATE_INDEX, NULL);
```
JoanFM: and here

src/debug_commands.c
```diff
@@ -1153,14 +1159,20 @@ DEBUG_COMMAND(InfoTagIndex) {
...
 
   // Field was not initialized yet
   if (!idx) {
     RedisModule_ReplyWithEmptyArray(sctx->redisCtx);
     goto end;
   }
 
+  // Debug info not supported for disk-mode tag indexes (TrieMap contains NULL sentinels)
```
JoanFM, line 1178: same as above

src/tag_index.h
```diff
@@ -106,10 +107,18 @@ typedef struct TagIndex {
   uint32_t uniqueId;
   TrieMap *values;
   TrieMap *suffix;
+
+  // Disk mode support: diskSpec != NULL means disk mode
+  // In disk mode, TrieMap values contains NULL sentinels instead of InvertedIndex pointers
+  RedisSearchDiskIndexSpec *diskSpec;  // NULL for memory mode, non-NULL for disk mode
+  t_fieldIndex fieldIndex;            // Field index (needed for disk API calls)
```
JoanFM, line 114: is the fieldIndex something we can guarantee? Isn't it better to have a field identifier? What if some fields are removed and the Field Struct is reallocated and resized?

src/fork_gc.c
kei-nan: I know, but cursor kept commenting on this function...

src/debug_commands.c
kei-nan: see:
```c
  InvertedIndex *iv;
```
in memory we have inverted indexes at the leaves of the trie.
in disk we put null

src/debug_commands.c
kei-nan: no, later in the function we rely on having inverted index structure in the trie node user data.

src/debug_commands.c
kei-nan: I guess suffixes don't rely on the user data 
but on the other hand, we don't rely on disk as well so no point in passing a disk handle.
I guess it wouldn't hurt to pass the handle for the sake of consistency.

src/debug_commands.c
kei-nan: this later assumes tie has an inverted index structure as the trie node user data.

src/tag_index.h
kei-nan: hard to say what will happen in the future
right now, it is determined by the order of the fields after the SCHEMA keyword.
currently we don't support removal of fields, only addition via FT.ALTER.

src/module.c
kei-nan: well we already have it, here
we don't need a disk handle to dump the strings of the trie.

src/tag_index.c
kei-nan: I think I will keep it as is.

src/debug_commands.c
kei-nan: we used inverted index structure in this function, not much point in passing a disk handle.

src/fork_gc.c
```diff
@@ -341,19 +341,19 @@ static void FGC_childCollectNumeric(ForkGC *gc, RedisSearchCtx *sctx) {
...
+    FGC_sendTerminator(gc);
+    return;
+  }
   arrayof(FieldSpec*) tagFields = getFieldsByType(sctx->spec, INDEXFLD_T_TAG);
   if (array_len(tagFields) != 0) {
     for (int i = 0; i < array_len(tagFields); ++i) {
-      TagIndex *tagIdx = TagIndex_Open(tagFields[i], DONT_CREATE_INDEX, NULL);
+      TagIndex *tagIdx = TagIndex_Open(tagFields[i], DONT_CREATE_INDEX, sctx->spec->diskSpec);
```
Itzikvaknin: We have ` if (sctx->spec->diskSpec) { .. return; }` 5 lines above, I think we should pass hard coded `NULL` to  `TagIndex_Open` instead of `sctx->spec->diskSpec`

Itzikvaknin (commented):
Left a small comment

src/tag_index.c
```diff
@@ -208,20 +212,42 @@ static inline size_t tagIndex_Put(TagIndex *idx, const char *value, size_t len,
...
+
+  if (idx->diskSpec) {
+    // DISK MODE: Index to disk and add tags to TrieMap with NULL sentinel
+    if (!SearchDisk_IndexTags(idx->diskSpec, values, n, docId, idx->fieldIndex)) {
+      return false;
+    }
```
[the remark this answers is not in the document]
kei-nan: updated the documentation

src/debug_commands.c
```diff
@@ -615,14 +615,20 @@ DEBUG_COMMAND(DumpTagIndex) {
...
+  const TagIndex *tagIndex = TagIndex_Open(fs, DONT_CREATE_INDEX, NULL);
 
   // Field was not initialized yet
   if (!tagIndex) {
     RedisModule_ReplyWithEmptyArray(sctx->redisCtx);
     goto end;
   }
 
+  // Debug dump not supported for disk-mode tag indexes (TrieMap contains NULL sentinels)
+  if (tagIndex->diskSpec) {
```
JoanFM, line 627: Shouldn't this be an ASSERT? How does this happen if u pass NONE When opening?

kei-nan: true, I guess we need to pass the disk spec in this case.
will pass the disk handle

Itzikvaknin (approved):

src/debug_commands.c
JoanFM: but how will this be if we pass NULL above?

src/fork_gc.c
JoanFM: But we should not care if cursor complains if it does not make sense. Let's not confuse now.

src/debug_commands.c
kei-nan: true, will fix by passing the pointer we have in the spec.

PR #8377, opened 2026-02-15
MOD-14062: Load from correct document upon `rename`

raz-mon:
This PR fixes the `FILTER` expression evaluation to load fields from the correct document (the updated one) in the case of the `RENAME` command, in which the name of a key is updated.
Should not have user-impact since this issue was shaded by another one, that together exposed a "correct" behavior from the user's POV.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches document indexing decision logic during `RENAME`, which can affect which documents are added/removed from indexes. Scope is small and backed by new regression tests, but incorrect behavior could cause indexing inconsistencies.
> 
> **Overview**
> Fixes `RENAME` handling so `FILTER` expressions are evaluated against the *new* key’s data (via `keyToReadData`) rather than the deleted old key, preventing incorrect add/remove decisions during cross-prefix renames.
> 
> Expands rename test coverage: adds regression tests for filter evaluation when renaming between indexes, for filters excluding the document, and for `RENAME` to the same key; removes an older prefix-change rename test and adds a small safety check that renaming an unrelated String key doesn’t crash.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4c0d7bd8d4634fb38bdab9e492000136c6cb9a13. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_followhashes.py
+++ b/tests/pytests/test_followhashes.py
@@ -194,6 +194,10 @@ def testRename(env):
     env.expect('ft.search things foo').equal([0])
     env.expect('ft.search otherthings foo').equal([1, 'otherthing:foo', ['name', 'foo']])
 
+    # Test that renaming a String key (unrelated type) does not crash
+    env.cmd('SET foo bar')
+    env.cmd('RENAME foo fubu')
+
 @skip(cluster=True)
 def testCopy(env):
     if not server_version_at_least(env, "6.2.0"):


src/spec.c
```diff
@@ -3908,6 +3908,9 @@ SpecOpIndexingCtx *Indexes_FindMatchingSchemaRules(RedisModuleCtx *ctx, RedisMod
 
       // load hash only if required
       if (!r) r = EvalCtx_Create();
+      // We load the data from the `keyToReadData` key, which is the key the old
+      // key was changed to, since the old key is already deleted.
+      key_p = RedisModule_StringPtrLen(keyToReadData, NULL);
```
oshadmi: `keyToReadData` is constant in the for loop, so can be moved to just before the loop?

tests/pytests/test_followhashes.py
```diff
@@ -194,17 +194,6 @@ def testRename(env):
...
-    env.cmd('RENAME foo fubu')
-
-@skip(cluster=True)
-def testRenameChangePrefix(env):
-    env.cmd('ft.create idx1 PREFIX 1 1: SCHEMA name text')
-    env.cmd('ft.create idx2 PREFIX 1 2: SCHEMA name text')
-
-    env.cmd('SET 1:1 bar')
-    env.expect('RENAME 1:1 2:1').ok()
-
```
oshadmi, line 207: Why remove this String key test? Making sure a rename of an unrelated type does not crash or affect the relevant types (hash, json)

src/spec.c
```diff
@@ -3908,6 +3908,9 @@ SpecOpIndexingCtx *Indexes_FindMatchingSchemaRules(RedisModuleCtx *ctx, RedisMod
 
       // load hash only if required
```
oshadmi:
```suggestion
      // load document only if required
```

tests/pytests/test_followhashes.py
```diff
@@ -693,6 +682,75 @@ def testIssue1571WithRename(env):
...
+    # Add a document that matches idx1's filter but NOT idx2's filter
+    conn.execute_command('hset', 'prefix1:doc', 'data', 'hello', 'type', 'allowed')
+
+    # Verify it's in idx1
+    env.expect('ft.search', 'idx1', 'hello').equal([1, 'prefix1:doc', ['data', 'hello', 'type', 'allowed']])
+    env.expect('ft.search', 'idx2', 'hello').equal([0])
+
+    # Rename to idx2's prefix - but the filter should NOT pass
+    # because type != "special"
+    env.expect('RENAME prefix1:doc prefix2:doc').ok()
```
oshadmi, line 752: What about testing rename to same name? e.g.,
```RENAME prefix1:doc prefix1:doc```

File: tests/pytests/test_followhashes.py (hunk 2 above)
Comment (raz-mon) [reply tests/pytests/test_followhashes.py:207]: I added a test that contains this

File: src/spec.c (hunk 1 above)
Comment (raz-mon) [reply src/spec.c]: Yes, will move it out

File: tests/pytests/test_followhashes.py (hunk 4 above)
Comment (raz-mon) [reply tests/pytests/test_followhashes.py:752]: Shouldn't invoke a special flow, but I can add a test for it

File: tests/pytests/test_followhashes.py [hunk 5]
```diff
@@ -194,17 +194,6 @@ def testRename(env):
     env.expect('ft.search things foo').equal([0])
     env.expect('ft.search otherthings foo').equal([1, 'otherthing:foo', ['name', 'foo']])
 
-    env.cmd('SET foo bar')
```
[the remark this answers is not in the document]
raz-mon: Bringing it back

src/spec.c
```diff
@@ -3897,6 +3897,9 @@ SpecOpIndexingCtx *Indexes_FindMatchingSchemaRules(RedisModuleCtx *ctx, RedisMod
   TrieMapResultBuf_Free(prefixes);
 
   if (runFilters) {
+    // We load the data from the `keyToReadData` key, which is the key the old
+    // key was changed to, since the old key is already deleted.
```
Itzikvaknin, line 3901: Can validate with an assertion?

tests/pytests/test_followhashes.py
```diff
@@ -693,6 +686,101 @@ def testIssue1571WithRename(env):
...
+
+    env.cmd('ft.create', 'idx2',
+            'PREFIX', '1', 'prefix2:',
+            'FILTER', '@type=="special"',
+            'SCHEMA', 'data', 'TEXT', 'type', 'TAG')
+
+    # Add a document that matches idx1's filter but NOT idx2's filter
+    conn.execute_command('hset', 'prefix1:doc', 'data', 'hello', 'type', 'allowed')
+
+    # Verify it's in idx1
```
Itzikvaknin, line 746:
```suggestion
    # Verify it's in idx1 and is not in idx2
```

src/spec.c
oshadmi: An assertion will need to open a key from string?

raz-mon: Discussed directly, we shouldn't assert this here IMO

PR #8385, opened 2026-02-16
Delete slots atomically

GuyAv46:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

#### Release Notes

### DIFF
--- a/docs/design/slot-based-deletion.md
+++ b/docs/design/slot-based-deletion.md
@@ -1,0 +1,758 @@
+# Slot-Based Document Deletion
+
+## Overview
+
+When Redis cluster slots are migrated away, we need to quickly mark all documents belonging to those slots as "deleted" without performing a costly linear scan. This document describes the design for fast slot-based deletion in the DocTable.
+
+## Problem Statement
+
+Redis will notify us (via `RedisModuleSlotRangeArray`) when slots need to be deleted. We need to:
+
+1. **Instantly** make documents in those slots unreachable (O(1) per slot, not O(n) per document)
+2. Keep `DocTable_Borrow` fast — it's on the critical query path
+3. Actual memory cleanup can happen asynchronously via background task + fork GC
+
+### Use Cases
+
+1. **ASM (Async Slot Migration) Trimming** — Primary use case, integrated with existing ASM mechanism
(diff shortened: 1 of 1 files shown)


docs/design/slot-based-deletion.md
```diff
@@ -0,0 +1,758 @@
...
+## Async Cleanup Architecture
+
+### Thread Safety Consideration
+
+The global specs dictionary (`specDict_g`) is **not thread-safe** and should only be accessed from the main thread or with the GIL held. Therefore:
+
+- **Main thread** iterates specs and creates one cleanup job per spec
+- **Background thread** (`cleanPool`) processes individual spec cleanup jobs
+- Each job holds a `WeakRef` to its spec, promoted to `StrongRef` at execution start
+- If promotion fails (spec was dropped), the job is skipped
```
GuyAv46, line 298: Alternatively, we can do it "unsafely" from the GC

GuyAv46: Nope, some parts will have to be done from the main process

docs/design/slot-based-deletion.md
```diff
@@ -0,0 +1,943 @@
...
+
+The two-phase cleanup approach minimizes query blocking:
+
+- **Collection phase** uses multiple read-lock passes — queries proceed normally, and each pass is bounded by a scan-specific limit
+- **Apply phase** uses multiple write-lock passes — only for actual mutations, and each pass is bounded by a separate apply-specific limit
+- If no documents were collected overall, the apply lock is never taken
+
+---
+
+## Edge Cases
```
tezc, line 814: Did you consider scenario of importing same slot while its cleanup is in progress?

1. Migrated slot 10 to other node.
2. Module started cleanup.
3. Import is started for slot 10 while cleanup is in progress.

Not sure if there is any issue for this scenario, pointing it just in case.

GuyAv46: Yes.
- When we create a cleanup job (from the main thread notification), we include the current max doc ID.
- This gives us a snapshot of the documents we need to clean (in these slots + ID lower than what was the max)
- If the same slots come back, when we re-index a document in these slots, it will get a higher ID than those we clean, so both the clean job and queries won't ignore it

PR #8389, opened 2026-02-16
[MOD-13670] Pass IndexSpec pointer to SearchDisk_runGC to apply changes

JoanFM:
> [!NOTE]
> **Medium Risk**
> Touches disk GC compaction and introduces new cross-language FFI mutations of live index state, which could impact index correctness or concurrency if misused.
> 
> **Overview**
> Disk-based index compaction (`SearchDisk_RunGC`) now takes an `IndexSpec` pointer and forwards it through the disk API `runGC` callback, enabling the disk GC implementation to apply compaction deltas back into in-memory structures.
> 
> Adds C FFI entrypoints for Rust GC to safely mutate the live spec during compaction (acquire/release spec write lock, decrement per-term trie doc counts, and decrement `numTerms` stats), and makes a small safety init to ensure `spec->diskSpec` starts as `NULL` during spec creation. Also fixes a generated include path in `command_info.c`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4544d7b5ffa3bdd3f0b1837c5392d2a97219a21e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -4213,8 +4213,8 @@ void IndexSpec_ReleaseWriteLock(IndexSpec* sp) {
 
 // Update a term's document count in the Serving Trie
 // Note: term is NOT null-terminated; term_len specifies the length
-IndexSpec_DecrementTrieTermCount(IndexSpec* sp, const char* term, size_t term_len,
-                               size_t doc_count_decrement) {
+void IndexSpec_DecrementTrieTermCount(IndexSpec* sp, const char* term, size_t term_len,
+                                  size_t doc_count_decrement) {
   if (!sp->terms || doc_count_decrement == 0) {
     return;
   }


src/search_disk_api.h
```diff
@@ -39,6 +39,47 @@ typedef struct AsyncPollResult {
...
+/**
+ * @brief Scoring stats delta - passed to update_scoring_stats callback.
+ *
+ * Must match the Rust ScoringStatsDelta struct layout exactly.
+ *
+ * Note: num_docs and totalDocsLen are updated at delete time, NOT by GC.
+ * GC only updates numTerms (when terms become completely empty).
+ */
+typedef struct SearchDisk_ScoringStatsDelta {
+  uint64_t num_terms_removed;  // Number of terms that became empty during compaction
```
raz-mon: The comment is a bit confusing - wdym?

src/search_disk.c
```diff
@@ -234,3 +236,62 @@ void SearchDisk_OutputInfoMetrics(RedisModuleInfoCtx* ctx) {
...
+// GC only updates numTerms (when terms become completely empty).
+static void SearchDisk_UpdateScoringStats(void* ctx,
+                                           const SearchDisk_ScoringStatsDelta* delta) {
+  IndexSpec* sp = (IndexSpec*)ctx;
+  if (!delta || delta->num_terms_removed == 0) {
+    return;
+  }
+
+  // Decrement numTerms (clamp to 0 to avoid underflow)
+  if (delta->num_terms_removed >= sp->stats.scoring.numTerms) {
```
raz-mon: This can't happen right? Let's assert this.
And I think we can avoid the condition, but it's not a hot path so not critical

src/search_disk.c
```diff
@@ -234,3 +236,62 @@ void SearchDisk_OutputInfoMetrics(RedisModuleInfoCtx* ctx) {
...
+// Note: term is NOT null-terminated; term_len specifies the length
+static void SearchDisk_UpdateTrieTerm(void* ctx, const char* term, size_t term_len,
+                                       size_t doc_count_decrement) {
+  IndexSpec* sp = (IndexSpec*)ctx;
+  if (!sp->terms || doc_count_decrement == 0) {
+    return;
+  }
+  // Decrement the numDocs count for this term in the trie
+  // If numDocs reaches 0, the node will be deleted
+  TrieDecrResult result = Trie_DecrementNumDocs(sp->terms, term, term_len, doc_count_decrement);
```
raz-mon: You take the result but do nothing with it

src/search_disk.c
```diff
@@ -101,9 +103,9 @@ QueryIterator* SearchDisk_NewWildcardIterator(RedisSearchDiskIndexSpec *index, d
     return disk->index.newWildcardIterator(index, weight);
 }
 
-void SearchDisk_RunGC(RedisSearchDiskIndexSpec *index) {
+void SearchDisk_RunGC(RedisSearchDiskIndexSpec *index, const SearchDisk_CompactionCallbacks *callbacks) {
     RS_ASSERT(disk && index);
-    disk->index.runGC(index);
+    disk->index.runGC(index, callbacks);
```
raz-mon: I'm wondering if it makes more sense to register these to the index in index creation.
Not critical, but since they won't be changing. WDYT?

JoanFM: makes more sense, sure

src/spec.c
```diff
@@ -1783,6 +1783,8 @@ StrongRef IndexSpec_Parse(const HiddenString *name, const char **argv, int argc,
   // Store on disk if we're on Flex.
   // This must be done before IndexSpec_AddFieldsInternal so that sp->diskSpec
   // is available when parsing vector fields (for populating diskCtx).
+  spec->diskSpec = NULL;
```
kei-nan, line 1785: why did you nullify diskSpec here?

src/spec.c
```diff
@@ -3500,6 +3506,8 @@ void *IndexSpec_LegacyRdbLoad(RedisModuleIO *rdb, int encver) {
   IndexSpec_StartGC(spec_ref, sp);
   Cursors_initSpec(sp);
 
+  sp->diskSpec = NULL;
```
kei-nan: why did you nullify diskSpec here?

src/spec.c
```diff
@@ -3500,6 +3506,8 @@ void *IndexSpec_LegacyRdbLoad(RedisModuleIO *rdb, int encver) {
...
+  sp->compactionCallbacks = (SearchDisk_CompactionCallbacks){0};
```
kei-nan: maybe this should move to where the index is constructed?

src/search_disk_api.h
```diff
@@ -39,6 +39,47 @@ typedef struct AsyncPollResult {
...
+ * @brief FFI-safe callback struct for compaction operations.
+ *
+ * This struct is created by C and passed to Rust. Rust invokes these callbacks
+ * to perform operations on C-owned data structures (Trie, ScoringStats, locks).
+ *
+ * Must match the Rust CompactionCallbacks struct layout exactly.
+ */
+typedef struct SearchDisk_CompactionCallbacks {
+  // === Synchronization Callbacks ===
+  void (*acquire_write_lock)(void* ctx);    // Acquire IndexSpec write lock
```
kei-nan: consider calling it start_update and stop update
I think that tells the user more that the update callbacks are called between the two functions.

src/search_disk_api.h
```diff
@@ -39,6 +39,47 @@ typedef struct AsyncPollResult {
...
+  void (*release_write_lock)(void* ctx);    // Release IndexSpec write lock
+
+  // === Memory Update Callbacks ===
+  // Update a term's document count in the Serving Trie
+  // term is NOT null-terminated; term_len specifies length
+  void (*update_trie_term)(void* ctx, const char* term, size_t term_len, size_t doc_count_decrement);
```
kei-nan: do we also need to consider how we are going to update the tag trie or is that will be done in a future PR?

src/spec.c
```diff
@@ -3318,6 +3323,7 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
     if (!sp->diskSpec) {
       goto cleanup;
     }
+    sp->compactionCallbacks = SearchDisk_CreateCompactionCallbacks(sp);
```
kei-nan: can this maybe move to the point in time we open the disk handle?
seems to be closely related to wheter we are in disk or not..
Maybe we can have an object to hold both the callback and the disk handle.
I don't see a case where the callbacks should exist but not the disk handle.

src/search_disk.c
```diff
@@ -234,3 +236,60 @@ void SearchDisk_OutputInfoMetrics(RedisModuleInfoCtx* ctx) {
...
 }
+
+// =============================================================================
+// Compaction Callbacks Implementation (Phase 3c)
+// =============================================================================
+
+// Acquire IndexSpec write lock
+static void SearchDisk_AcquireWriteLock(void* ctx) {
+  IndexSpec* sp = (IndexSpec*)ctx;
+  pthread_rwlock_wrlock(&sp->rwlock);
```
kei-nan: I think we lock the index lock in a different way to avoid recursive locking.
Consider using the existing index spec pattern way to take the lock.

kei-nan (commented):
left some comments.

src/search_disk_api.h
JoanFM: Future PR

src/spec.c
JoanFM: Yes, we could have a wrapper

src/search_disk_api.h
JoanFM: I think is better to say the callbacks u offer, and the user should decide when to call them

src/search_disk.c
JoanFM: I do not understand, what pattern is this?

src/redisearch_rs/c_entrypoint/compaction_ffi/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+///
+/// All function pointers must be valid and the `ctx` pointer must remain
+/// valid for the duration of compaction. The C side is responsible for
+/// providing valid callback implementations.
+#[repr(C)]
+pub struct CompactionCallbacks {
+    // === Synchronization Callbacks ===
+    /// Acquire IndexSpec write lock.
+    /// Called before applying delta to memory structures.
+    pub acquire_write_lock: Option<unsafe extern "C" fn(*mut c_void)>,
```
chesedo: These should be pure Rust function callbacks. `cbindgen` should take care of turning them into the correct C code 🤞 

All in all the Rust code should not contain C conventions

src/redisearch_rs/c_entrypoint/compaction_ffi/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+/// {
+///     let guard = callbacks.write_lock();
+///     guard.update_trie_term("foo", 5);
+///     guard.update_scoring_stats(&delta);
+///     // Lock automatically released here
+/// }
+/// ```
+pub struct WriteGuard<'a> {
+    callbacks: &'a CompactionCallbacks,
+}
```
chesedo: I don't understand why we need this? Is it to wrap the `unsafe` C functions?

chesedo (commented):
Minor comments

src/redisearch_rs/c_entrypoint/compaction_ffi/src/lib.rs
JoanFM: it is to provide a RAII pattern to the lock, so it is tied to a lifetime

src/redisearch_rs/c_entrypoint/compaction_ffi/src/lib.rs
JoanFM: I can't get to fix this. Would u be able to quickly do a PR against this branch?

src/debug_commands.c
```diff
@@ -824,8 +824,8 @@ DEBUG_COMMAND(GCForceInvoke) {
     return RedisModule_ReplyWithError(ctx, "Unknown index name");
   }
 
-  if (sp->diskSpec) {
-    SearchDisk_RunGC(sp->diskSpec);
+  if (sp->diskCtx.spec) {
+    SearchDisk_RunGC(sp->diskCtx.spec, &sp->diskCtx.compactionCallbacks);
```
raz-mon: Didn't we say that we register these to the disk rather than give them as an argument?
Can't they just be ffi calls rather than callbacks?

src/redisearch_rs/headers/compaction.h
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * The C definition is auto-generated by cbindgen.
+ *
+ * # Safety
+ *
+ * All function pointers must be valid and the `ctx` pointer must remain
+ * valid for the duration of compaction. The C side is responsible for
+ * providing valid callback implementations.
+ */
+typedef struct SearchDisk_CompactionCallbacks {
```
raz-mon: Can we have these to be regular ffi functions rather than callbacks, and then we don't need to register or send as an argument?

src/debug_commands.c
JoanFM: we decided to have callbacks at the end

src/redisearch_rs/headers/compaction.h
JoanFM: this is exactly the decision we came to

src/spec.c
```diff
@@ -4195,3 +4196,43 @@ void Indexes_EndRDBLoadingEvent(RedisModuleCtx *ctx) {
...
+// Note: num_docs and totalDocsLen are updated at delete time, NOT by GC.
+// GC only updates numTerms (when terms become completely empty).
+void IndexSpec_UpdateScoringStats(IndexSpec* sp, uint64_t num_terms_removed) {
+  if (num_terms_removed == 0) {
+    return;
+  }
+
+  // Decrement numTerms (clamp to 0 to avoid underflow)
+  RS_ASSERT(num_terms_removed <= sp->stats.scoring.numTerms);
+  sp->stats.scoring.numTerms -= num_terms_removed;
```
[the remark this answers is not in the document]
JoanFM: callback removed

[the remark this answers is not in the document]
JoanFM: removed comment

src/spec.c
```diff
@@ -4195,3 +4196,42 @@ void Indexes_EndRDBLoadingEvent(RedisModuleCtx *ctx) {
...
+void IndexSpec_UpdateTrieTerm(IndexSpec* sp, const char* term, size_t term_len,
+                               size_t doc_count_decrement) {
+  if (!sp->terms || doc_count_decrement == 0) {
+    return;
+  }
+  // Decrement the numDocs count for this term in the trie
+  // If numDocs reaches 0, the node will be deleted
+  TrieDecrResult result = Trie_DecrementNumDocs(sp->terms, term, term_len, doc_count_decrement);
+  RS_ASSERT(result != TRIE_DECR_NOT_FOUND);
+}
```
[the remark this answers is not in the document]
JoanFM: This is outside of the current concern of the feature. in any case, this should be a concern of the caller, as the responsability of the method should be unique here

src/search_disk_api.h
```diff
@@ -154,11 +154,13 @@ typedef struct IndexDiskAPI {
...
-   * removing entries for deleted documents.
+   * removing entries for deleted documents. Also applies the compaction delta
+   * to update in-memory structures via FFI calls to the provided C IndexSpec.
    *
-   * @param index Pointer to the index
+   * @param index Pointer to the disk index
+   * @param c_index_spec Opaque pointer to the C IndexSpec (used for FFI callbacks)
    */
-  void (*runGC)(RedisSearchDiskIndexSpec *index);
+  void (*runGC)(RedisSearchDiskIndexSpec *index, void *spec);
```
kei-nan: I think it would be better if from disk perspective this would be called user_data.
no need for it to know or limit what the is type of the context.

raz-mon (dismissed):
LGTM 👍

src/spec.c
```diff
@@ -4195,3 +4196,42 @@ void Indexes_EndRDBLoadingEvent(RedisModuleCtx *ctx) {
...
+
+// Release IndexSpec write lock
+void IndexSpec_ReleaseWriteLock(IndexSpec* sp) {
+  pthread_rwlock_unlock(&sp->rwlock);
+}
+
+// Update a term's document count in the Serving Trie
+// Note: term is NOT null-terminated; term_len specifies the length
+void IndexSpec_UpdateTrieTerm(IndexSpec* sp, const char* term, size_t term_len,
```
raz-mon: I would be more explicit here:
```suggestion
IndexSpec_DecrementTrieTermCount(IndexSpec* sp, const char* term, size_t term_len,
```

src/spec.c
```diff
@@ -4195,3 +4196,42 @@ void Indexes_EndRDBLoadingEvent(RedisModuleCtx *ctx) {
...
+  // Decrement the numDocs count for this term in the trie
+  // If numDocs reaches 0, the node will be deleted
+  TrieDecrResult result = Trie_DecrementNumDocs(sp->terms, term, term_len, doc_count_decrement);
+  RS_ASSERT(result != TRIE_DECR_NOT_FOUND);
+}
+
+// Update IndexScoringStats based on compaction delta
+// Note: num_docs and totalDocsLen are updated at delete time, NOT by GC.
+// GC only updates numTerms (when terms become completely empty).
+void IndexSpec_UpdateScoringStats(IndexSpec* sp, uint64_t num_terms_removed) {
```
raz-mon: Here too
```suggestion
void IndexSpec_DecrementNumTerms(IndexSpec* sp, uint64_t num_terms_removed) {
```

src/search_disk_api.h
JoanFM: done

src/search_disk.h
```diff
@@ -122,11 +122,13 @@ void SearchDisk_DeleteDocument(RedisSearchDiskIndexSpec *handle, const char *key
...
- * removing entries for deleted documents.
+ * removing entries for deleted documents. Applies the compaction delta to
+ * update in-memory structures via FFI calls to the provided C IndexSpec.
  *
- * @param index Pointer to the index
+ * @param index Pointer to the disk index
+ * @param user_data Pointer to the C IndexSpec (for FFI callbacks to update memory structures)
  */
-void SearchDisk_RunGC(RedisSearchDiskIndexSpec *index);
+void SearchDisk_RunGC(RedisSearchDiskIndexSpec *index, IndexSpec *user_data);
```
kei-nan: I think you misunderstood what I asked for.
at the oss level its fine if this is still be called spec.
I was asking for it to be called user_data specifically in `runGC` in `search_disk_api.h`

src/search_disk_api.h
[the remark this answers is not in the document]
JoanFM: this is not happening

src/search_disk.h
JoanFM: Applied now, please re-review

kei-nan (dismissed):

PR #8405, opened 2026-02-17
MOD-14088: Connect disk GC tasks to a C side threadpool

Itzikvaknin:
## Summary

Plugs disk-backed indexes into the existing GC scheduling so they run `SearchDisk_RunGC` on the same timer/threshold model as fork GC, and refactors the GC layer to be policy-agnostic. No new user-facing config; OSS behaviour unchanged.

### Disk GC policy

- **New `GCPolicy_Disk`** and **`DiskGC`** ([`src/disk_gc.c`](src/disk_gc.c), [`src/disk_gc.h`](src/disk_gc.h)): implement `GCCallbacks` and call `SearchDisk_RunGC(sp->diskSpec)` from `periodicCb`. Use the same interval and clean threshold as fork GC (read from shared config). All mutable fields in `DiskGC` are `atomic_size_t` — `deletedDocsFromLastRun` uses `atomic_exchange` for the read-and-zero in `periodicCb` (and `atomic_fetch_add` in `deleteCb`), and `intervalSec` replaces the former `struct timespec interval` to avoid torn reads between the GC thread and the main thread.
- **Index creation / load**: `IndexSpec_StartGC` takes a policy; disk indexes get `GCPolicy_Disk`, in-memory get `GCPolicy_Fork`. [`src/spec.c`](src/spec.c): `IndexSpec_CreateNew`, `IndexSpec_StoreAfterRdbLoad`, and `IndexSpec_LegacyRdbLoad` pass `sp->diskSpec ? GCPolicy_Disk : GCPolicy_Fork`
- **Flex mode** ([`src/module.c`](src/module.c)): instead of disabling GC, overrides settings defaults to `DEFAULT_DISK_GC_RUN_INTERVAL` (300s) and `DEFAULT_DISK_GC_CLEAN_THRESHOLD` (10000). Existing config keys (`FORK_GC_*`) still apply to the shared settings.

### GC API and callbacks

- **`periodicCallback(void* gcCtx, bool force)`**: added `force`; when true (e.g. `FT.DEBUG GC_FORCEINVOKE`), both Fork and Disk GC skip the clean-threshold check. [`src/gc.c`](src/gc.c): `taskCallback` passes `false`, `debugTaskCallback` passes `true`.
- **`getStats` callback** and **`InfoGCStats`**: new callback and struct (bytes collected, cycles, time, lastRunTimeMs). [`src/gc.h`](src/gc.h): `GCContext_GetStats` delegates to it. Fork GC fills stats; Disk GC returns zeros (stats live on disk/Rust side).

### Config and naming

- **Shared settings config**: `forkGcConfig` renamed to **`GcSettings`**, field **`.forkGc`** to **`.gcSettings`** ([`src/config.h`](src/config.h), [`src/config.c`](src/config.c)). Same struct used by both policies; user-facing names (`FORK_GC_RUN_INTERVAL`, etc.) unchanged.
- **Defaults**: `DEFAULT_DISK_GC_RUN_INTERVAL` and `DEFAULT_DISK_GC_CLEAN_THRESHOLD` added for Flex mode; not exposed as new module args.

### Force-invoke and debug

- **`FT.DEBUG GC_FORCEINVOKE`** ([`src/debug_commands.c`](src/debug_commands.c)): single path for all indexes with GC; removed special-case that called `SearchDisk_RunGC` directly for disk. Both fork and disk now go through `GCContext_ForceInvoke` (and thus respect `force` in `periodicCb`).

### Tests

- **Python** ([`tests/pytests/test_gc.py`](tests/pytests/test_gc.py)): GC tests adjusted for the unified force-invoke behaviour and any config/API expectations.

## Behaviour

- **OSS**: Only fork GC is used; config and behaviour unchanged.
- **DISK**: Disk indexes get a `DiskGC` on the same scheduler; interval and threshold come from the shared `gcSettings` config (overridden in Flex to disk defaults). Force-invoke bypasses threshold for both policies.

---

> [!NOTE]
> **Medium Risk**
> Touches core GC scheduling and index lifecycle paths and introduces a new disk GC policy, so regressions could impact cleanup behavior or scheduling across index types; changes are localized but affect runtime background tasks.
> 
> **Overview**
> Disk-backed indexes are now plugged into the existing GC scheduler via a new `GCPolicy_Disk` and `DiskGC` implementation, so they run `SearchDisk_RunGC` on the same timer/threshold model as fork GC and are selected automatically when an index has `diskSpec`.
> 
> The GC layer is refactored to be policy-agnostic: `periodicCallback` now takes a `force` flag (used by `FT.DEBUG GC_FORCEINVOKE` to bypass the clean-threshold), GC stats retrieval is unified via a new `getStats` callback + `InfoGCStats`, and fork-GC creation/entrypoints are renamed/updated (`FGC_New`  `FGC_Create`, `InfoGCStats` moved to `gc.h`).
> 
> Configuration is reshaped by renaming `forkGcConfig` to `GCSettings` (field `forkGc`  `gcSettings`) and, in Flex validation mode, the defaults for `search-fork-gc-run-interval`/`search-fork-gc-clean-threshold` switch to disk-oriented values; tests are updated accordingly and GC debug command flow is simplified to a single path.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 93c3b5b24cb00a1c34ac1fcee84c3cb10822defe. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/disk_gc.c
+++ b/src/disk_gc.c
@@ -17,7 +17,7 @@
 #include <stdatomic.h>
 #include <time.h>
 
-// TODO: update stats with the number of documents cleaned
+// TODO: update stats with the number of documents cleaned - MOD-14199
 static bool periodicCb(void *privdata, bool force) {
   DiskGC *gc = privdata;
   StrongRef spec_ref = IndexSpecRef_Promote(gc->index);


src/redisearch_api.c
```diff
@@ -865,6 +865,8 @@ int RediSearch_IndexInfo(RSIndex* rm, RSIdxInfo *info) {
    * Avoid rehashing the terms dictionary */
   dictPauseRehashing(sp->keysDict);
 
+  // TODO: return disk GC policy when it is enabled
+  // Do not expose disk GC, report fork when any GC is present
```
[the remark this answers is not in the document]
Itzikvaknin: We decided to determine the policy based on the index type (disk/memory) 
In such case, the policy doesn't make much sense, so we decided to leave it as is

src/debug_commands.c
```diff
@@ -824,18 +824,13 @@ DEBUG_COMMAND(GCForceInvoke) {
     return RedisModule_ReplyWithError(ctx, "Unknown index name");
   }
 
-  if (sp->diskSpec) {
-    SearchDisk_RunGC(sp->diskSpec);
-    RedisModule_ReplyWithSimpleString(ctx, "DONE");
-    return REDISMODULE_OK;
-  } else if (sp->gc) {
```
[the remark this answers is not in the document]
Itzikvaknin: Fixed in https://github.com/RediSearch/RediSearch/pull/8405/commits/e30c9d7344c07f51c3d436f8e7d250196e22532d

src/disk_gc.c
```diff
@@ -0,0 +1,104 @@
...
+
+  if (!force && gc->deletedDocsFromLastRun < RSGlobalConfig.gcConfigParams.gcSettings.forkGcCleanThreshold) {
+    IndexSpecRef_Release(spec_ref);
+    return 1;
+  }
+
+  SearchDisk_RunGC(sp->diskSpec);
+
+  size_t num_docs_to_clean = atomic_exchange(&gc->deletedDocsFromLastRun, 0);
+  IndexsGlobalStats_UpdateLogicallyDeleted(-(int64_t)num_docs_to_clean);
```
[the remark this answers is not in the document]
Itzikvaknin: See https://github.com/RediSearch/RediSearch/pull/8405/changes/54063e068998375cd12fb8790b52c54fd6dd9596

src/config.h
```diff
@@ -71,21 +71,21 @@ static inline const char *GCPolicy_ToString(GCPolicy policy) {
       return "huh?";  // LCOV_EXCL_LINE cannot be reached
   }
 }
+
```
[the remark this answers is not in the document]
Itzikvaknin: Intentional, this policy is not user-facing

src/disk_gc.c
```diff
@@ -0,0 +1,104 @@
...
+  if (!sp->diskSpec) {
+    IndexSpecRef_Release(spec_ref);
+    return 1;
+  }
+
+  size_t num_docs_to_clean = atomic_exchange(&gc->deletedDocsFromLastRun, 0);
+  if (!force && num_docs_to_clean < RSGlobalConfig.gcConfigParams.gcSettings.forkGcCleanThreshold) {
+    IndexSpecRef_Release(spec_ref);
+    return 1;
+  }
```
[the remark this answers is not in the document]
Itzikvaknin: See https://github.com/RediSearch/RediSearch/pull/8405/commits/d7a1abd24d1952e0b460795a8137b285f1970fb4

src/gc.c
```diff
@@ -41,6 +42,9 @@ GCContext* GCContext_CreateGC(StrongRef spec_ref, uint32_t gcPolicy) {
     case GCPolicy_Fork:
       ret->gcCtx = FGC_New(spec_ref, &ret->callbacks);
       break;
+    case GCPolicy_Disk:
```
meiravgri, line 45: consider assert sp->diskSpec for disk and !sp->diskSpec for fork

src/gc.c
```diff
@@ -41,6 +42,9 @@ GCContext* GCContext_CreateGC(StrongRef spec_ref, uint32_t gcPolicy) {
...
+      ret->gcCtx = DiskGC_New(spec_ref, &ret->callbacks);
+      break;
   }
   return ret;
```
meiravgri:
```suggestion
      ret->gcCtx = DiskGC_New(spec_ref, &ret->callbacks);
      break;
    default:
        // assert
  }
  return ret;
```

src/config.h
```diff
@@ -307,6 +307,8 @@ char *getRedisConfigValue(RedisModuleCtx *ctx, const char* confName);
 #define DEFAULT_FORK_GC_CLEAN_THRESHOLD 100
 #define DEFAULT_FORK_GC_RETRY_INTERVAL 5
 #define DEFAULT_FORK_GC_RUN_INTERVAL 30
+#define DEFAULT_DISK_GC_RUN_INTERVAL 300
+#define DEFAULT_DISK_GC_CLEAN_THRESHOLD 10000
 #define DEFAULT_INDEX_CURSOR_LIMIT 128
```
meiravgri, line 313: How were these values determined?

src/module.c
```diff
@@ -1664,9 +1664,10 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx) {
       RedisModule_Log(ctx, "error", "Search Disk is enabled but could not be initialized");
       return REDISMODULE_ERR;
     }
-    // Disable GC when running in Flex mode
-    RSGlobalConfig.gcConfigParams.enableGC = false;
-    RedisModule_Log(ctx, "notice", "GC disabled (Flex mode)");
+    // Use disk GC policy when running in Flex mode; override fork config to disk-appropriate defaults
+    RSGlobalConfig.gcConfigParams.gcSettings.forkGcRunIntervalSec = DEFAULT_DISK_GC_RUN_INTERVAL;
+    RSGlobalConfig.gcConfigParams.gcSettings.forkGcCleanThreshold = DEFAULT_DISK_GC_CLEAN_THRESHOLD;
+    RedisModule_Log(ctx, "notice", "GC policy set to disk (Flex mode)");
```
meiravgri: what if the user sets a different value on startup/runtime? how do we sync?

src/config.h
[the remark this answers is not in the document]
JoanFM: @Itzikvaknin u can still add the option and let it go to default, less warnings from compilers and so on ...

src/disk_gc.c
```diff
@@ -0,0 +1,105 @@
...
+#include "spec.h"
+#include "search_disk.h"
+#include "module.h"
+#include "redismodule.h"
+#include "rmalloc.h"
+#include "info/global_stats.h"
+#include <stdatomic.h>
+#include <time.h>
+
+static int periodicCb(void *privdata, bool force) {
```
JoanFM: let's return a boolean

src/config.h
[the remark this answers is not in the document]
Itzikvaknin: Done https://github.com/RediSearch/RediSearch/pull/8405/commits/f4baac244348f92f7cebfad3aa251802316961d4

src/disk_gc.c
Itzikvaknin: Done https://github.com/RediSearch/RediSearch/pull/8405/changes/f4baac244348f92f7cebfad3aa251802316961d4

src/disk_gc.c
```diff
@@ -0,0 +1,105 @@
...
+  size_t num_docs_to_clean = atomic_exchange(&gc->deletedDocsFromLastRun, 0);
+  if (!force && num_docs_to_clean < RSGlobalConfig.gcConfigParams.gcSettings.forkGcCleanThreshold) {
+    atomic_fetch_add(&gc->deletedDocsFromLastRun, num_docs_to_clean);
+    IndexSpecRef_Release(spec_ref);
+    return true;
+  }
+
+  SearchDisk_RunGC(sp->diskSpec, sp);
+
+  IndexsGlobalStats_UpdateLogicallyDeleted(-(int64_t)num_docs_to_clean);
```
meiravgri: what syncing mechanism guarantees that a delete/update wasn't called by the main thread and changed the actual number of docs cleaned by the gc?

src/disk_gc.c
```diff
@@ -0,0 +1,105 @@
...
+
+static void onTerminateCb(void *privdata) {
+  DiskGC *gc = privdata;
+  size_t remaining = atomic_exchange(&gc->deletedDocsFromLastRun, 0);
+  IndexsGlobalStats_UpdateLogicallyDeleted(-(int64_t)remaining);
+  WeakRef_Release(gc->index);
+  rm_free(gc);
+}
+
+/* Stats are maintained in disk info; do not add anything here. */
```
meiravgri, line 55: so what is the purpose of the new callback?

src/module.c
```diff
@@ -1677,9 +1677,10 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx) {
       RedisModule_Log(ctx, "error", "Search Disk is enabled but could not be initialized");
       return REDISMODULE_ERR;
     }
-    // Disable GC when running in Flex mode
-    RSGlobalConfig.gcConfigParams.enableGC = false;
-    RedisModule_Log(ctx, "notice", "GC disabled (Flex mode)");
+    // Use disk GC policy when running in Flex mode; override fork config to disk-appropriate defaults
+    RSGlobalConfig.gcConfigParams.gcSettings.forkGcRunIntervalSec = DEFAULT_DISK_GC_RUN_INTERVAL;
```
meiravgri: What if the user sets different value via config file/ startup args?

src/config.h
Itzikvaknin: We will benchmark them in future, for now these are numbers that we think make sense

src/module.c
Itzikvaknin: Whats the difference between startup/runtime? They both override the default values
The idea was to not change the OSS user-facing config with disk content (e.g new policy/ new dedicated config params) before the release, so I connected the existing params, changing only the defaults in the presence of disk module

src/disk_gc.c
Itzikvaknin: Nice catch! 
There is no synchronization at the moment, this code is inspired by the memory GC, which is guarded by GIL+fork CoW 

The solution here is that `SearchDisk_RunGC` will return the amount of deleted docs, which is related to @JoanFM work on updating in memory data structures, for now Ill remove the post `SearchDisk_RunGC ` update code

src/disk_gc.c
Itzikvaknin: To implement no-op to satisfy the interface

src/module.c
Itzikvaknin: https://github.com/RediSearch/RediSearch/pull/8405#discussion_r2846451186

src/disk_gc.c
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  SearchDisk_RunGC(sp->diskSpec, sp);
+
+  // TODO: update stats with the number of documents cleaned
+
+  gc->intervalSec = RSGlobalConfig.gcConfigParams.gcSettings.forkGcRunIntervalSec;
+
+  IndexSpecRef_Release(spec_ref);
+  return true;
```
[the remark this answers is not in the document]
Itzikvaknin: ```// TODO: update stats with the number of documents cleaned```

https://github.com/RediSearch/RediSearch/pull/8405#discussion_r2846486666

src/module.c
```diff
@@ -1677,9 +1677,10 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx) {
...
+    RSGlobalConfig.gcConfigParams.gcSettings.forkGcCleanThreshold = DEFAULT_DISK_GC_CLEAN_THRESHOLD;
```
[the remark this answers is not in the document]
Itzikvaknin: See https://github.com/RediSearch/RediSearch/pull/8405#discussion_r2846451186

src/gc.c
Itzikvaknin: https://github.com/RediSearch/RediSearch/pull/8405/changes/382ec7d175b590d9e06d12b16021b05ea97499e2

src/gc.c
Itzikvaknin: https://github.com/RediSearch/RediSearch/pull/8405/changes/382ec7d175b590d9e06d12b16021b05ea97499e2

src/gc.c
```diff
@@ -39,8 +40,16 @@ GCContext* GCContext_CreateGC(StrongRef spec_ref, uint32_t gcPolicy) {
   GCContext* ret = rm_calloc(1, sizeof(GCContext));
   switch (gcPolicy) {
     case GCPolicy_Fork:
+      RS_LOG_ASSERT(spec_ref->diskSpec == NULL, "Fork GC is not supported for disk indexes");
       ret->gcCtx = FGC_New(spec_ref, &ret->callbacks);
       break;
+    case GCPolicy_Disk:
+      RS_LOG_ASSERT(SearchDisk_IsEnabled() && spec_ref->diskSpec, "Disk GC is not enabled");
```
JoanFM: Let's separate this in two ASSERTS or change the Assert message, Is it not enabled or is diskSpec being NULL? Make sure the log message says what really happens

tests/pytests/test_gc.py
```diff
@@ -649,3 +614,37 @@ def test_gc_oom_replica_relaxed():
     bytes_collected = int(gc_dict['bytes_collected'])
     env.assertGreater(bytes_collected, 0,
         message="GC should run on replica even when maxmemory is exceeded")
+@skip(cluster=True)
```
JoanFM, line 618: let's add blank line before

src/disk_gc.c
```diff
@@ -0,0 +1,105 @@
...
+
+DiskGC *DiskGC_New(StrongRef spec_ref, GCCallbacks *callbacks) {
+  DiskGC *gc = rm_calloc(1, sizeof(*gc));
+  *gc = (DiskGC){
+      .index = StrongRef_Demote(spec_ref),
+      .deletedDocsFromLastRun = 0,
+  };
+  gc->intervalSec = RSGlobalConfig.gcConfigParams.gcSettings.forkGcRunIntervalSec;
+
+  callbacks->onTerm = onTerminateCb;
```
JoanFM, line 95: Let's have two separated API, DiskGC_New to alllocate and another one to Fill the callback, this DiskGC_New is doing more than what the method seems to provide.

Itzikvaknin: I can rename it, but I dont think this constructor is too complicated

I would even move the callbacks from the `GCContext` to the `DiskGC`/`ForkGC` object, WDYT?

JoanFM: yes, could work. But I would just rename or separate the concerns and would be easier. I just would like to separate the New creation with setting callbacks, maybe calling _Init is better

tests/pytests/test_flex_validation.py
```diff
@@ -194,3 +194,36 @@ def test_flex_workers_minimum(env):
...
+    env.expect(config_cmd(), 'SET', 'FORK_GC_RUN_INTERVAL', 600).ok()
+    env.expect(config_cmd(), 'SET', 'FORK_GC_CLEAN_THRESHOLD', 20000).ok()
+
+    # GET reflects the change
+    env.expect(config_cmd(), 'GET', 'FORK_GC_RUN_INTERVAL').equal([['FORK_GC_RUN_INTERVAL', '600']])
+    env.expect(config_cmd(), 'GET', 'FORK_GC_CLEAN_THRESHOLD').equal([['FORK_GC_CLEAN_THRESHOLD', '20000']])
+
+
+@skip(cluster=True)
+def test_flex_gc_config_explicit_override(env):
```
meiravgri, line 218: :tumbsup:

src/disk_gc.c
```diff
@@ -0,0 +1,106 @@
...
+  IndexSpec *sp = StrongRef_Get(spec_ref);
+  if (!sp) {
+    return false;
+  }
+  if (!sp->diskSpec) {
+    IndexSpecRef_Release(spec_ref);
+    return true;
+  }
+
+  size_t num_docs_to_clean = atomic_exchange(&gc->deletedDocsFromLastRun, 0);
```
meiravgri: We should update the value of `deletedDocsFromLastRun` only after the gc finishes the clean up and we know the actual value.
We might have decide not to run at all becuase there are not enough docs to trigger, or the gc skipped of the deletion becuase of concurrent writes in the parent process.
For the first scenrio - fixing the value rather then handle just adds redundant overhead we can avoid.

meiravgri: To reproduce the potential bug - consider adding a test where we have docs deleted in the last block of some invertedd index and then during gc clean up, the parent adds docs to it.
It will cause the gc apply to skip the last block and the number of actual cleaned-up docs to not match `deletedDocsFromLastRun` and the metric to go out of order

Itzikvaknin: Created MOD-14199 to do these updates, for now I removed any updates, and only load `deletedDocsFromLastRun` for threshold gating

src/disk_gc.c
```diff
@@ -0,0 +1,106 @@
...
+  DiskGC *gc = privdata;
+  StrongRef spec_ref = IndexSpecRef_Promote(gc->index);
+  IndexSpec *sp = StrongRef_Get(spec_ref);
+  if (!sp) {
+    return false;
+  }
+  if (!sp->diskSpec) {
+    IndexSpecRef_Release(spec_ref);
+    return true;
+  }
```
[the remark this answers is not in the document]
Itzikvaknin:
```
// Releases a strong reference to a spec
// Must only be called if the spec was promoted successfully
// Will also clear the current thread's active spec
void IndexSpecRef_Release(StrongRef ref);
```
In `spec.h:755`

src/disk_gc.c
Itzikvaknin: Renamed to `DiskGC_Create`

src/disk_gc.c
```diff
@@ -0,0 +1,105 @@
...
+static void statsForInfoCb(RedisModuleInfoCtx *ctx, void *gcCtx) {
+  (void)ctx;
+  (void)gcCtx;
+}
+
+static void deleteCb(void *ctx) {
+  DiskGC *gc = ctx;
+  atomic_fetch_add(&gc->deletedDocsFromLastRun, 1);
+  IndexsGlobalStats_UpdateLogicallyDeleted(1);
+}
```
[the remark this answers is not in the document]
Itzikvaknin: Intentional, see https://github.com/RediSearch/RediSearch/pull/8405#discussion_r2851663011

src/disk_gc.c
meiravgri: But if we never decrease `deletedDocsFromLastRun`, this check is redundant

Itzikvaknin: As for now, I cant do more than create a followup ticket for this fix

PR #8422, opened 2026-02-18
MOD-13305: Delete Non SST Persistent Data On Startup

kei-nan:
On startup, when reading from RDB, if no sst persistence is available then delete existing databases before opening them

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches Flex disk index open + RDB save/load paths; an incorrect SST detection or delete flag could lead to stale index data or unintended on-disk data deletion during startup.
> 
> **Overview**
> Ensures Flex disk-backed indexes don’t start up with *stale on-disk state* when loading from an RDB that did **not** use SST persistence by deleting existing disk index data before opening the index spec.
> 
> This threads an explicit `useSst`/context-flags signal through `IndexSpec_RdbLoad`/`IndexSpec_RdbSave` and extends `SearchDisk_OpenIndex`/`openIndexSpec` with a `deleteBeforeOpen` flag, plus adds notice-level logging of SST persistence during RDB load paths. Tests and benchmark requirements were updated to match the new signatures/dependencies.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 40aa54f4f1ce467026e911e5fec53706f3f58f54. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -3235,14 +3235,14 @@ void IndexSpec_RdbSave(RedisModuleIO *rdb, IndexSpec *sp, int contextFlags) {
     // In a forked child process, the memory is a snapshot so no lock is needed.
     RedisModuleCtx *ctx = RedisModule_GetContextFromIO(rdb);
     RedisSearchCtx sctx = SEARCH_CTX_STATIC(ctx, sp);
-    const bool shouldTakeSpecLock = !(contextFlags & REDISMODULE_CTX_FLAGS_IS_CHILD); 
-    if (shouldTakeSpecLock) {
+    const bool inMainProcess = !(contextFlags & REDISMODULE_CTX_FLAGS_IS_CHILD); 
+    if (inMainProcess) {
       RedisSearchCtx_LockSpecRead(&sctx);
     }
     IndexScoringStats_RdbSave(rdb, &sp->stats.scoring);
     TrieType_GenericSave(rdb, sp->terms, false, true);
     SearchDisk_IndexSpecRdbSave(rdb, sp->diskSpec);
-    if (shouldTakeSpecLock) {
+    if (inMainProcess) {
       RedisSearchCtx_UnlockSpec(&sctx);
     }
   }


src/module.c
```diff
@@ -1785,26 +1786,48 @@ int RediSearch_InitModuleInternal(RedisModuleCtx *ctx) {
...
   static int invoked = 0;
   if (invoked || !RS_Initialized) {
     return;
   }
   invoked = 1;
 
-  // First free all indexes
-  Indexes_Free(specDict_g, false);
+  // Phase 1: Stop timers (OSS cluster only)
+  StopRedisTopologyUpdater(RSDummyContext);
```
raz-mon: Why was this changed?
Can we separate the changes related to disk deletion from the others?

kei-nan: removed this change

src/spec.h
```diff
@@ -764,6 +764,10 @@ void Indexes_EndRDBLoadingEvent(RedisModuleCtx *ctx);
 // This function is to be called when loading finishes (failed or not)
 void Indexes_EndLoading();
 
+// Returns whether the last RDB save was an SST persistence.
+// Used at shutdown to determine if we should delete disk data.
+bool WasLastRdbOperationSstPersistent(void);
```
Itzikvaknin: Not used

src/spec.c
```diff
@@ -3182,7 +3204,7 @@ void Indexes_ScanAndReindex() {
 
 ///////////////////////////////////////////////////////////////////////////////////////////////
 
-void IndexSpec_RdbSave(RedisModuleIO *rdb, IndexSpec *sp) {
+void IndexSpec_RdbSave(RedisModuleIO *rdb, IndexSpec *sp, bool useSst) {
```
Itzikvaknin: `bool useSst` not used in the function

src/spec.c
```diff
@@ -99,6 +103,23 @@ static void DebugIndexesScanner_pauseCheck(DebugIndexesScanner* dScanner, RedisM
 // Forward declaration for disk validation
 inline static bool isSpecOnDiskForValidation(const IndexSpec *sp);
 
+bool WasLastRdbOperationSstPersistent(void) {
+  return lastRdbOperationWasSstPersistent;
+}
+
+/**
+ * Checks if SST persistence is enabled for the given RDB context.
+ * Tracks the state for shutdown cleanup decisions.
```
raz-mon: Not relevant any more right?

src/spec.c
raz-mon: Used in 3243

src/spec.c
```diff
@@ -3309,27 +3329,27 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
...
   // Open the index on disk only if we are on Flex, and this is not a duplicate.
+  // If SST persistence was NOT used, delete existing disk data before opening
+  // since there's no SST data to restore from (stale data must be cleared).
   if (isSpecOnDisk(sp) && !sp->isDuplicate) {
     RS_ASSERT(disk_db);
     size_t len;
     const char* name = HiddenString_GetUnsafe(sp->specName, &len);
-    sp->diskSpec = SearchDisk_OpenIndex(name, len, sp->rule->type);
+    const bool deleteBeforeOpen = !useSst;
+    sp->diskSpec = SearchDisk_OpenIndex(name, len, sp->rule->type, deleteBeforeOpen);
```
raz-mon:
```suggestion
    sp->diskSpec = SearchDisk_OpenIndex(name, len, sp->rule->type, !useSst);
```

src/spec.c
```diff
@@ -3309,27 +3329,27 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
...
     if (!sp->diskSpec) {
       goto cleanup;
     }
   }
 
-  // Check if we are using SST files with this RDB.
-  // We assume symmetry w.r.t this context flag. I.e., If it is not set, we
-  // assume it was not set in when the RDB was saved as well.
+  // Get context for logging
   RedisModuleCtx *ctx = RedisModule_GetContextFromIO(rdb);
```
raz-mon: Not used any more

src/spec.c
```diff
@@ -3561,14 +3583,16 @@ int Indexes_RdbLoad(RedisModuleIO *rdb, int encver, int when) {
 
 void Indexes_RdbSave(RedisModuleIO *rdb, int when) {
 
+  const bool useSst = CheckRdbSstPersistence(rdb, "RDB Save");
```
raz-mon: What's the motivation of moving this here rather than in the rdbSave function?

src/spec.c
```diff
@@ -3530,6 +3551,7 @@ void IndexSpec_LegacyRdbSave(RedisModuleIO *rdb, void *value) {
 }
 
 int Indexes_RdbLoad(RedisModuleIO *rdb, int encver, int when) {
+  const bool useSst = CheckRdbSstPersistence(rdb, "RDB Load");
```
raz-mon: Same here, why have this check here rather than later, when we need it?

raz-mon (commented):
Looks good, 
see some comments

src/spec.c
kei-nan: Yes, agreed. This is no longer relevant after the cleanup; the stale getter/state were removed.

src/spec.c
kei-nan: Applied, thanks.

src/spec.h
kei-nan: Removed in the latest update. I deleted the unused `WasLastRdbOperationSstPersistent()` declaration/definition and related state.

src/spec.c
kei-nan: It is used in the disk-save branch (`if (sp->diskSpec && useSst)`). Keeping the parameter for that conditional save path.

src/spec.c
kei-nan: Removed. The local context variable was no longer needed there.

src/spec.c
kei-nan: Same rationale on load: checking once at the operation boundary gives one consistent `useSst` value for all indexes in that load pass and avoids repeating the context check per index.

src/spec.c
kei-nan: Main reason is consistency and locality of control flow: we evaluate `useSst` once per RDB save operation and pass it through to all per-index save calls, so every index in that pass uses the same decision and logging point.

raz-mon (dismissed):

src/spec.c
```diff
@@ -3504,7 +3509,8 @@ void *IndexSpec_LegacyRdbLoad(RedisModuleIO *rdb, int encver) {
     RS_ASSERT(disk_db);
     size_t len;
     const char* name = HiddenString_GetUnsafe(sp->specName, &len);
-    sp->diskSpec = SearchDisk_OpenIndex(name, len, sp->rule->type);
+    // Legacy RDB does not have SST persistence, so always delete before opening.
+    sp->diskSpec = SearchDisk_OpenIndex(name, len, sp->rule->type, false);
```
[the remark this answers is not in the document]
kei-nan: We don't expect legacy rdb in this use case.

Itzikvaknin (dismissed):

PR #8425, opened 2026-02-19
[MOD-14131] Port idf.c to Rust

gdesmott:
Replace the C IDF implementation with an idiomatic Rust crate (`idf`) and its FFI wrapper (`idf_ffi`), following the standard porting pattern.

The Rust implementation performs all BM25 IDF arithmetic in f64 (double) precision. The original C code used `0.5F` and `1.0F` float literals in CalculateIDF_BM25, which caused intermediate computations to be performed in 32-bit float precision before the final result was promoted to double for the `log()` call. The Rust version eliminates this unintentional precision loss.

The expected score values in test_scorers.py are updated to reflect the higher-precision results. The differences are in the 7th+ significant digit (e.g. 0.287682102254 -> 0.287682072452) and do not affect search ranking behavior.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches core scoring math and the C↔Rust FFI boundary; while covered by new Rust tests and only minor numeric deltas are expected, regressions could affect score values and some ranking edge-cases.
> 
> **Overview**
> Replaces the C implementation of IDF/BM25 IDF (`src/idf.c`) with a new Rust workspace crate `idf`, and exposes it back to the C code via a new `idf_ffi` crate exporting `CalculateIDF`/`CalculateIDF_BM25` and generating `redisearch_rs/headers/idf.h` with `cbindgen`.
> 
> Wires the new FFI module into the Rust entrypoint (`redisearch_rs`), updates workspace/dependency metadata (including adding `approx` for tests), and adjusts Python scorer/issue tests’ expected score strings to match the slightly different floating-point results from the Rust BM25 IDF math.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5b47ea11372edc3e859b371904d3e3a9adace483. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/idf/tests/tests.rs
+++ b/src/redisearch_rs/idf/tests/tests.rs
@@ -9,6 +9,7 @@
 
 use approx::{assert_abs_diff_eq, assert_ulps_eq};
 use idf::{calculate_idf, calculate_idf_bm25};
+use rstest::rstest;
 
 // Miri interprets floating-point operations using a software implementation
 // that can produce results slightly different from hardware FPUs.
@@ -80,26 +81,25 @@ fn idf_known_values() {
 }
 
 /// Values computed from the original C implementation of `CalculateIDF`.
-#[test]
-fn idf_matches_c_reference() {
-    // (total_docs, term_docs) => expected IDF
-    let cases: &[(usize, usize, f64)] = &[
-        (0, 0, 1.0),
-        (0, 1, 1.0),
-        (1, 0, 1.0),
-        (1, 1, 1.0),
-        (10, 5, 1.0),
-        (100, 1, 6.0),
-        (100, 50, 1.0),
-        (100, 100, 1.0),
-        (1000, 1, 9.0),
-        (1000, 500, 1.0),
-        (10_000, 10_000, 1.0),
-        (1_000_000, 1, 19.0),
-    ];
-    for &(total, term, expected) in cases {
-        assert_abs_diff_eq!(calculate_idf(total, term), expected);
-    }
+#[rstest]
+#[case(0, 0, 1.0)]
+#[case(0, 1, 1.0)]
+#[case(1, 0, 1.0)]
+#[case(1, 1, 1.0)]
+#[case(10, 5, 1.0)]
+#[case(100, 1, 6.0)]
+#[case(100, 50, 1.0)]
+#[case(100, 100, 1.0)]
+#[case(1000, 1, 9.0)]
+#[case(1000, 500, 1.0)]
+#[case(10_000, 10_000, 1.0)]
+#[case(1_000_000, 1, 19.0)]
+fn idf_matches_c_reference(
+    #[case] total_docs: usize,
+    #[case] term_docs: usize,
+    #[case] expected: f64,
+) {
+    assert_abs_diff_eq!(calculate_idf(total_docs, term_docs), expected);
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


tests/pytests/test_issues.py
```diff
@@ -197,7 +197,7 @@ def test_issue1988(env):
     conn = getConnectionByEnv(env)
     env.cmd('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT')
     conn.execute_command('HSET', 'doc1', 't', 'foo')
-    score = '0.28768210225410284'
+    score = '0.28768207245178085'
```
gdesmott, line 200: @GuyAv46 : are those kind of changes ok? It actually improves the precision, see commit message, but I can keep the exact same logic as the C version if required.

src/redisearch_rs/idf/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+    // Miri interprets floating-point operations using a software implementation
+    // that can produce results slightly different from hardware FPUs. For example,
+    // `f64::log2(2.0)` may return a value marginally below `1.0` under Miri,
+    // causing `floor()` to yield `0.0` instead of the expected `1.0`. Similarly,
+    // `f64::ln` results can differ in the last ULP, breaking exact equality
+    // assertions.
+    //
+    // Since this crate contains no `unsafe` code, Miri cannot catch any
+    // additional bugs here — the ignored tests are purely about floating-point
+    // precision, not memory safety.
```
LukeMathWalker: Two things here:
- We should check for approximate equality when testing the outcome of floating-point operations, which'll make these tests more robust and also solve the miri issue.
- Let's move the tests over to the `tests/` folder, since they use the public API.

src/redisearch_rs/idf/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+/// assert_eq!(calculate_idf(100, 0), calculate_idf(100, 1));
+/// ```
+#[inline]
+pub fn calculate_idf(total_docs: usize, term_docs: usize) -> f64 {
+    let term_docs = if term_docs == 0 { 1 } else { term_docs };
+    let value = 1.0 + (total_docs + 1) as f64 / term_docs as f64;
+    // `logb` returns the binary exponent, i.e. `floor(log2(x))` for positive
+    // normal floating-point values.
+    value.log2().floor()
+}
```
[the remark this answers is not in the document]
gdesmott: @LukeMathWalker : another interesting comment. I fixed it in a secondary commit but maybe that's a bit overkill?

src/redisearch_rs/idf/tests/tests.rs
```diff
@@ -0,0 +1,188 @@
...
+    assert_abs_diff_eq!(calculate_idf(1000, 500), 1.0);
+    // logb(1.0 + (1000+1)/1000) = logb(2.001) = floor(log2(2.001)) = floor(1.000...) = 1
+    assert_abs_diff_eq!(calculate_idf(1000, 1000), 1.0);
+}
+
+/// Values computed from the original C implementation of `CalculateIDF`.
+#[test]
+fn idf_matches_c_reference() {
+    // (total_docs, term_docs) => expected IDF
+    let cases: &[(usize, usize, f64)] = &[
```
LukeMathWalker: `rstest` would be a good fit here.

tests/pytests/test_issues.py
GuyAv46: Yeah, these seem OK.
Does the better precision cost in performance?

src/redisearch_rs/idf/tests/tests.rs
```diff
@@ -0,0 +1,188 @@
...
+        );
+    }
+}
+
+/// Values computed from the original C implementation of `CalculateIDF_BM25`,
+/// with the float-precision intermediate arithmetic corrected to f64.
+#[test]
+fn bm25_idf_matches_c_reference() {
+    // (total_docs, term_docs) => expected BM25 IDF
+    let cases: &[(usize, usize, f64)] = &[
```
LukeMathWalker: Same here, `rstest` would work nicely.

src/redisearch_rs/idf/tests/tests.rs
gdesmott: First time using it, nice!

LukeMathWalker (approved):

tests/pytests/test_issues.py
gdesmott: No, the original code was already doing the `log()` call on 64 bits which is the expensive part. The only thing we changed here are the cheap add/subtract/divide which should be as fast on 64 bits on modern CPUs.

src/redisearch_rs/idf/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+/// ```
+#[inline]
+pub fn calculate_idf(total_docs: usize, term_docs: usize) -> f64 {
+    let term_docs = if term_docs == 0 { 1 } else { term_docs };
+    let value = 1.0 + (total_docs + 1) as f64 / term_docs as f64;
+    // Extract the binary exponent directly from the IEEE 754 representation,
+    // equivalent to C's `logb`. This is exact — unlike `log2().floor()`, it
+    // cannot be off by one when the value is close to a power of two.
+    ilogb(value) as f64
+}
```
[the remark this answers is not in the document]
GuyAv46: Definitely not meant to wrap, let's use saturating? Is there a performance cost for using this?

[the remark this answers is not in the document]
LukeMathWalker: A little overhead should be there. But is a scenario with total_doc == usize::MAX actually relevant?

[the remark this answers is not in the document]
GuyAv46: No, I don't think so.
myself-i-feel-very-safe-phonecian-scheme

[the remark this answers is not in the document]
gdesmott: Want me to a `debug_assert (total_docs < usize::MAX)` or let's keep it as it then?

LukeMathWalker (approved):

PR #8432, opened 2026-02-19
[MOD-14084] Introduce RLookup_New and RLookupRow_New in Rust

memark:
[MOD-14084] Introduce RLookup_New and RLookupRow_New in Rust.

#### Release Notes

[MOD-14084]: https://redislabs.atlassian.net/browse/MOD-14084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Low Risk**
> Small additive API change that introduces new FFI constructors; primary risk is ABI/FFI signature expectations for by-value opaque structs.
> 
> **Overview**
> Adds new FFI entrypoints `RLookup_New()` and `RLookupRow_New(const RLookup*)` to construct `RLookup`/`RLookupRow` from Rust and return them as opaque-by-value structs.
> 
> Updates the auto-generated `rlookup_rs.h` to expose these constructors to C/C++ callers, enabling creation of lookup objects/rows without relying on non-Rust initializers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 20b3c02864ba8219e1bd956b62cd1397b4bedd6d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
+++ b/src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
@@ -9,9 +9,7 @@
 
 use ffi::RSValue;
 use libc::size_t;
-#[cfg(debug_assertions)]
-use rlookup::OpaqueRLookup;
-use rlookup::{OpaqueRLookupRow, RLookup, RLookupKey, RLookupRow};
+use rlookup::{OpaqueRLookup, OpaqueRLookupRow, RLookup, RLookupKey, RLookupRow};
 use std::{
     ffi::{CStr, c_char},
     mem::{self, ManuallyDrop},
@@ -27,20 +25,12 @@ use value::RSValueFFI;
 /// 1. `lookup` must be a [valid], non-null pointer to an [`RLookup`].
 ///
 /// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
-#[cfg(debug_assertions)]
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn RLookupRow_New(lookup: *const OpaqueRLookup) -> OpaqueRLookupRow {
     // Safety: ensured by caller (1.)
     let lookup = unsafe { RLookup::from_opaque_ptr(lookup).unwrap() };
 
     RLookupRow::new(lookup).into_opaque()
-}
-
-/// Returns a newly created [`RLookupRow`].
-#[cfg(not(debug_assertions))]
-#[unsafe(no_mangle)]
-pub extern "C" fn RLookupRow_New() -> OpaqueRLookupRow {
-    RLookupRow::new().into_opaque()
 }
 
 /// Writes a key to the row but increments the value reference count before writing it thus having shared ownership.
--- a/src/redisearch_rs/headers/rlookup_rs.h
+++ b/src/redisearch_rs/headers/rlookup_rs.h
@@ -619,7 +619,6 @@ int32_t RLookup_LoadRuleFields(RedisSearchCtx *search_ctx,
                                const char *key,
                                QueryError *status);
 
-#if defined(ENABLE_ASSERT)
 /**
  * Returns a newly created [`RLookupRow`].
  *
@@ -630,14 +629,6 @@ int32_t RLookup_LoadRuleFields(RedisSearchCtx *search_ctx,
  * [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
  */
 struct RLookupRow RLookupRow_New(const struct RLookup *lookup);
-#endif
-
-#if !defined(ENABLE_ASSERT)
-/**
- * Returns a newly created [`RLookupRow`].
- */
-struct RLookupRow RLookupRow_New(void);
-#endif
 
 /**
  * Writes a key to the row but increments the value reference count before writing it thus having shared ownership.


src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
```diff
@@ -469,6 +469,12 @@ pub unsafe extern "C" fn RLookup_GetRowLen(lookup: *const OpaqueRLookup) -> u32
     lookup.get_row_len()
 }
 
+/// Returns a newly created [`RLookup`].
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RLookup_New() -> OpaqueRLookup {
```
daxhuiberts: Does it need to be unsafe?

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
```diff
@@ -18,6 +20,29 @@ use std::{
...
+    // Safety: ensured by caller (1.)
+    let lookup = unsafe { RLookup::from_opaque_ptr(lookup).unwrap() };
+
+    RLookupRow::new(lookup).into_opaque()
+}
+
+/// Returns a newly created [`RLookupRow`].
+#[cfg(not(debug_assertions))]
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RLookupRow_New() -> OpaqueRLookupRow {
```
daxhuiberts: Same here.

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
```diff
@@ -18,6 +20,29 @@ use std::{
...
+/// Returns a newly created [`RLookupRow`].
+///
+/// # Safety
+///
+/// 1. `lookup` must be a [valid], non-null pointer to an [`RLookup`].
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[cfg(debug_assertions)]
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RLookupRow_New(lookup: *const OpaqueRLookup) -> OpaqueRLookupRow {
```
daxhuiberts, line 29: Why is the debug version of `RLookupRow_New` different from the release version?

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
memark: I guess not!

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
memark: I guess not!

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
memark: Good catch! I guess it doesn't need to be. I forgot that we merged the signatures a while back.

Instead we know just ignore the passed value sometimes:
https://github.com/RediSearch/RediSearch/blob/100e102aaca2c8603bc2b90afeb3b9ef3c1c6a1e/src/redisearch_rs/rlookup/src/row.rs#L54

daxhuiberts: I understand the setup/purpose now.

Although the `RLookup id` is only used in debug mode, it still needs to be passed in debug and release mode to keep a consistent function signature. It feels like debug stuff is 'bleeding' a bit into release mode, but not sure what another approach would be and I don't think it's really going to be a problem though so I think this is fine 👍

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
daxhuiberts: FYI From what I got from Luca is that we want to follow the rust guidelines for safety: i.e. things that can create undefined behaviour.

Things that can leak memory (such as this one) or things that can generate a panic instead of UB don't warrant unsafe functions and safety comments (though the latter does need comments on what can trigger a panic).

daxhuiberts (approved):

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
memark: Makes sense. It's still weird though that it needs to be `unsafe(no_mangle)`. 🤔

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
memark: Agree on that one, not a big fan either... Will need to discuss that again with Jonas I think.

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
daxhuiberts: That's a different unsafe. That is because of the no-mangling of the function name and rust nog guaranteeing that the no-mangled names of functions and other stuff don't collide or override something. It was previously just `#[no_mangle]`, but the extra `unsafe(...)` was added to really state that some unintended effects might happen.

PR #8439, opened 2026-02-20
[MOD-14066] fix handling GC Out of Mem on replica

JoanFM:
On replica side, only max_process_mem will be taken into account to decide cancelling a GC run.

This avoid many skips of the GC cycles in replicas which avoids memory increase

---

> [!NOTE]
> **Medium Risk**
> Changes fork-GC scheduling behavior under memory pressure and adds a replica-based integration test; incorrect memory ratio calculation could cause GC to run when it should be throttled (or vice versa), impacting stability/perf.
> 
> **Overview**
> Fork-GC’s OOM gating is adjusted to treat replicas differently: masters keep using `RedisMemory_GetUsedMemoryRatioUnified`, while replicas compute pressure solely from `INFO MEMORY`’s `used_memory/max_process_mem` (ignoring `maxmemory`), preventing repeated GC skips on replicas.
> 
> Tests extend coverage with `test_gc_oom_replica_relaxed`, which sets up a master+slave, forces the slave over `maxmemory`, runs `GC_FORCEINVOKE` on the replica, and asserts GC still collects bytes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5dc1664a491ede8123b63e6a380da20f8c42e237. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pytests/test_gc.py
+++ b/tests/pytests/test_gc.py
@@ -568,7 +568,7 @@ def test_gc_oom(env:Env):
     for i in range(num_docs):
         env.expect('DEL', f'doc{i}').equal(1)
 
-    forceInvokeGC(env)    
+    forceInvokeGC(env)
 
     # Verify no bytes collected by GC
     info = index_info(env)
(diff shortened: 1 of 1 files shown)


src/fork_gc.c
```diff
@@ -978,8 +978,21 @@ FGCError FGC_parentHandleFromChild(ForkGC *gc) {
...
 // GIL must be held before calling this function
 static inline bool isOutOfMemory(RedisModuleCtx *ctx) {
-  // Debug log the memory ratio
-  float used_memory_ratio = RedisMemory_GetUsedMemoryRatioUnified(ctx);
+  // Check if we are a slave/replica
+  bool isSlave = RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_SLAVE;
+  float used_memory_ratio = 0;
+  if (!isSlave) {
+    // On master, use the original unified logic
+    used_memory_ratio = RedisMemory_GetUsedMemoryRatioUnified(ctx);
```
alonre24, line 986: Would it be better to move it be inside `RedisMemory_GetUsedMemoryRatioUnified` ?

JoanFM: I do not like it too much, I would prefer to have another function or keep as is. At the end RedisMemory_* should only care about the memory and not have this logic about slave.

PR #8441, opened 2026-02-23
Refactor MRConn structure and improve connection management logic

GuyAv46:
### Main objects this PR modified

1. Issue: Protocol state could stay stale across detach/disconnect paths.  
Fix: Reset `conn->protocol` in `detachFromConn()` and in free/disconnect paths.

2. Issue: `HELLO` (protocol switch) enqueue failure was ignored.  
Fix: In `MRConn_SendCommand`, check `redisAsyncCommand(... "HELLO %d" ...)` return and fail command on `REDIS_ERR`.

3. Issue: Pool removal/replacement could free wrappers without explicitly disconnecting connections first.  
Fix: Added explicit pool disconnect before replace/remove and centralized via `MRConnPool_Disconnect()`.

4. Issue: `MRConnManager_Disconnect` had unsafe delete/disconnect sequencing and error-prone semantics.
Fix: Switched to `dictUnlink` + disconnect + `dictFreeUnlinkedEntry` flow.

5. Issue: Destructor (`MRConnPool_Free`) used manual `conn->conn->data = NULL` and had inaccurate assumptions in comments.  
Fix: Updated comments and changed destructor to use `detachFromConn(conn, false)` for consistent detachment logic.

6. Issue: TLS config cleanup bug in `checkTLS`: wrong pointer freed (`ca_cert` branch) and passphrase leak risk.  
Fix: Correctly free `ca_cert`, free `key_pass`, and null all freed pointers.

7. Issue: In connect callback, post-connect TLS/auth failures used detach-only (`shouldFree=false`) in places where context should be freed.  
Fix: Changed those failure paths to `detachFromConn(conn, true)`.

8. Issue: Potential leak on TLS init failure after `SSL_new`.  
Fix: Added `SSL_new` null check and explicit `SSL_free(ssl)` on `redisInitiateSSL` failure path.

9. Issue: Missing null-check for `redisAsyncConnectWithOptions`.  
Fix: Added `if (!c) { ... return REDIS_ERR; }`.

10. Issue: Return values from hiredis/libuv setup were not validated.  
Fix: Check `redisLibuvAttach`, `redisAsyncSetConnectCallback`, and `redisAsyncSetDisconnectCallback`; detach/free and fail on error.

11. Issue: Misleading comment in connect error path implied we free there while passing `detachFromConn(..., false)`.  
Fix: Updated comment to reflect hiredis-owned cleanup on connect-failure callback path.

12. Improvement: Repeated disconnect loops in the manager code were duplicated.  
Fix: Added `static inline MRConnPool_Disconnect(MRConnPool *pool)` and used it everywhere.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches connection management, disconnect sequencing, and TLS setup/cleanup paths where subtle lifecycle mistakes can cause leaks or use-after-free. Changes are localized and backed by new tests, but failures would impact cluster connectivity stability.
> 
> **Overview**
> Improves robustness of RMR async connection lifecycle by centralizing detach/disconnect behavior, resetting per-connection protocol state on teardown, and ensuring pool/node removal paths explicitly disconnect before freeing wrappers.
> 
> Hardens error handling around protocol negotiation and TLS setup: `HELLO` enqueue failures now abort the send, TLS config cleanup frees the correct pointers (and nulls them), and connect/TLS/init failure paths now correctly free hiredis/SSL resources and validate `redisAsyncConnectWithOptions`/libuv attach/callback setup.
> 
> Adds C++ interop guards to `conn.h` and introduces new C++ tests that exercise `IORuntimeCtx_UpdateNodes` for add/remove and resize scenarios to catch connection-map lifecycle regressions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f0432b0c243a4bf5e8082003473cae233b8ceb61. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/coord/rmr/conn.c
+++ b/src/coord/rmr/conn.c
@@ -111,7 +111,7 @@ static void closeTimer(MRConn *conn) {
 }
 
 /*
-* This is called when the IORuntime is being shut down. This is called from the uv thread.
+* This is called from the uv thread.
 */
 static void MRConn_Disconnect(MRConn *conn) {
   CONN_LOG(conn, "Disconnecting connection");
@@ -410,14 +410,11 @@ static void signalCallback(uv_timer_t *tm) {
     return;  // Nothing to do here!
   }
 
-  redisAsyncContext *ac = conn->conn;
-
   if (conn->state == MRConn_Freeing) {
     CONN_LOG(conn, "Freeing connection");
+
+    redisAsyncContext *ac = detachFromConn(conn, false);
     if (ac) {
-      ac->data = NULL;
-      conn->conn = NULL;
-      conn->protocol = 0;
       redisAsyncDisconnect(ac);
     }
     freeConn(conn);
@@ -483,7 +480,6 @@ static void MRConn_SwitchState(MRConn *conn, MRConnState nextState) {
 
 activate_timer:
   if (conn->timer && !uv_is_active(conn->timer)) {
-    uv_timer_t *tm = conn->timer;
     uv_timer_start(conn->timer, signalCallback, nextTimeout, 0);
   }
 }


src/coord/rmr/conn.c
```diff
@@ -24,16 +24,17 @@
 typedef struct MRConn{
   MREndpoint ep;
   redisAsyncContext *conn;
-  MRConnState state;
   void *timer;
-  int protocol; // 0 (undetermined), 2, or 3
   uv_loop_t *loop;
+  int protocol; // 0 (undetermined), 2, or 3
```
JoanFM, line 29: should we use the time to have an enum?

src/coord/rmr/conn.c
```diff
@@ -268,7 +274,9 @@ int MRConn_SendCommand(MRConn *c, MRCommand *cmd, redisCallbackFn *fn, void *pri
     }
   }
   if (cmd->protocol != 0 && (!c->protocol || c->protocol != cmd->protocol)) {
-    int rc = redisAsyncCommand(c->conn, NULL, NULL, "HELLO %d", cmd->protocol);
+    if (redisAsyncCommand(c->conn, NULL, NULL, "HELLO %d", cmd->protocol) == REDIS_ERR) {
+      return REDIS_ERR;
```
JoanFM, line 278: should we do some sort of logging? or what chain of events would follow?

src/coord/rmr/conn.c
```diff
@@ -762,18 +793,27 @@ static int MRConn_Connect(MRConn *conn) {
                           .endpoint.tcp = {.ip = conn->ep.host, .port = conn->ep.port}};
 
   redisAsyncContext *c = redisAsyncConnectWithOptions(&options);
+  if (!c) {
```
JoanFM, line 792: I am all in for this, but not anywhere in the engine we check for allocation failures.

src/coord/rmr/conn.c
JoanFM: I see it comes from command as an int

src/coord/rmr/conn.c
```diff
@@ -335,10 +344,14 @@ int MRConnManager_ConnectAll(MRConnManager *m) {
 
 /* Explicitly disconnect a connection and remove it from the connection pool */
 int MRConnManager_Disconnect(MRConnManager *m, const char *id) {
-  if (dictDelete(m->map, id)) {
```
JoanFM, line 338: shouldn't we just add MRConnPool_Disconnect to the funct pointer os the dict? and keep using delete?

tests/ctests/coord_tests/test_cluster.c
```diff
@@ -157,6 +158,55 @@ void testCluster() {
   }
 }
 
+void testIORuntime_UpdateNodesAddRemove() {
```
JoanFM: I think these tests should go into tests_io_runtimes directly?

src/coord/rmr/conn.c
GuyAv46: I think it's too big a change for these local fixes (the protocol is the same type as the MRCommand's protocol, and we directly compare and assign them).
Let's do thatin a future PR

src/coord/rmr/conn.c
GuyAv46: We can, though I don't think this failure is different than any other failure that may occur in this function

src/coord/rmr/conn.c
GuyAv46: This is not only an allocation failure (which we should avoid checking when using `RM_Alloc` family, as they never fail as an API guarantee), but also option conflicts and other internal logic.
I'll improve the log message

src/coord/rmr/conn.c
GuyAv46: I tried convincing the AI to do so a few times, but it gave enough reasons to keep the logic as it is and not mix this logic into memory management, so I think I won't do it at this point

JoanFM (dismissed):

src/coord/rmr/conn.c
```diff
@@ -404,6 +417,7 @@ static void signalCallback(uv_timer_t *tm) {
     if (ac) {
       ac->data = NULL;
       conn->conn = NULL;
+      conn->protocol = 0;
```
lerman25: Can this be repalced with `detachFromConn`?

src/coord/rmr/conn.c
lerman25: `MRConnManager_Add` can also lead to here, is the comment still correct?

lerman25: Not you, but augment notices `*tm` unused so might as well if you want to

PR #8443, opened 2026-02-23
[MOD-14147] Compute the correct cardinality after a GC run on the node of a NumericRangeTree

LukeMathWalker:

The first commit adds a regression test, showcasing the issue when all entries in the last block are deleted _and_ the last block isn't GCed.
The second commit patches the GC logic to track the last block pointer, regardless of whether or not the callback is fired on the last block.

For release notes, perhaps:

> Fixed incorrect cardinality estimation in numeric range fields after garbage collection. In a corner case, we would undercount distinct values, potentially causing suboptimal range tree splits over time.

1. MOD-...
2. #...

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches ForkGC’s numeric cardinality (HLL) repair path, which affects query correctness for numeric fields after GC; logic change is small but in a critical data-structure maintenance flow.
> 
> **Overview**
> Fixes an edge-case where numeric field cardinality could be computed incorrectly after a ForkGC run when the last inverted-index block becomes empty but is preserved (due to parent-side changes).
> 
> The GC numeric repair callback now targets HLL updates based on a precomputed pointer to the actual last block (instead of relying on block-change detection), and a new regression test reproduces the scenario and asserts the corrected cardinality.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f140a02f8fa6295126547eafc908e744382a5738. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/fork_gc.c
+++ b/src/fork_gc.c
@@ -204,23 +204,16 @@ static void FGC_childCollectTerms(ForkGC *gc, RedisSearchCtx *sctx) {
 }
 
 typedef struct {
-  struct HLL majority_card;     // Holds the majority cardinality of all the blocks we've seen so far
-  struct HLL last_block_card;   // Holds the cardinality of the last block we've seen
-  const IndexBlock *last_block; // The last block we've seen, to know when to merge the cardinalities
+  struct HLL majority_card;         // Cardinality of all blocks except the last
+  struct HLL last_block_card;       // Cardinality of the last block only
+  const IndexBlock *actual_last_block; // Pre-computed pointer to the real last block
 } numCbCtx;
 
 static void countRemain(const RSIndexResult *r, const IndexBlock *blk, void *arg) {
   numCbCtx *ctx = arg;
-
-  if (ctx->last_block != blk) {
-    // We are in a new block, merge the last block's cardinality into the majority, and clear the last block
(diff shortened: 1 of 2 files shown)


src/fork_gc.c
```diff
@@ -204,23 +204,16 @@ static void FGC_childCollectTerms(ForkGC *gc, RedisSearchCtx *sctx) {
 }
 
 typedef struct {
-  struct HLL majority_card;     // Holds the majority cardinality of all the blocks we've seen so far
-  struct HLL last_block_card;   // Holds the cardinality of the last block we've seen
-  const IndexBlock *last_block; // The last block we've seen, to know when to merge the cardinalities
+  struct HLL majority_card;         // Cardinality of all blocks except the last
+  struct HLL last_block_card;       // Cardinality of the last block only
+  const IndexBlock *actual_last_block; // Pre-computed pointer to the real last block
```
meiravgri, line 209:
```suggestion
  struct HLL majority_card;         // Cardinality of all blocks except the last
  struct HLL positional_last_block_card;       // Cardinality of the last block only
  const IndexBlock *positional_last_block_ptr; // Pre-computed pointer to the real last block
```

PR #8444, opened 2026-02-23
[MOD-14150] RLookup - Rework Init function

memark:
Reworks the `RLookup_Init` function. Since the introduction of `RLookup_New`, the init function really doesn't need to do much anymore. The only thing left is setting the spec cache property, which is now moved to a separate function.

Rust side
- rename function
- remove self replacement
- remove debug assert

C side
- rename function
- remove memset
- remove unnecessary calls

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches core lookup initialization across coordinator, pipeline construction, and Rust FFI; misordered initialization or missing cache setup could cause subtle query/LOAD behavior regressions.
> 
> **Overview**
> **User-visible behavior should remain unchanged**, but `RLookup` initialization is simplified: the old `RLookup_Init` API is replaced with `RLookup_SetCache`, which now only associates an `IndexSpecCache` with an already-created lookup.
> 
> All call sites are updated to stop zero-initializing lookups (no more implicit `memset`/re-init behavior) and instead rely on `RLookup_New()` plus an explicit cache set where needed; tests and the Rust FFI are updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8ababf5214873ac2d2de7de5259ccce8a5641790. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/cpptests/test_cpp_hybridmerger.cpp
+++ b/tests/cpptests/test_cpp_hybridmerger.cpp
@@ -180,7 +180,7 @@ HybridLookupContext* CreateDummyLookupContext(size_t numUpstreams) {
   }
 
   // Create dummy tail lookup
-  RLookup *tailLookup = (RLookup*)rm_calloc(1, sizeof(RLookup));
+  RLookup *tailLookup = (RLookup*)rm_malloc(sizeof(RLookup));
   EXPECT_NE(tailLookup, nullptr);
   *tailLookup = RLookup_New();
   lookupCtx->tailLookup = tailLookup;


src/rlookup.h
```diff
@@ -465,10 +465,10 @@ typedef struct {
 int RLookup_LoadDocument(RLookup *lt, RLookupRow *dst, RLookupLoadOptions *options);
 
 /**
- * Initialize the lookup. If cache is provided, then it will be used as an
+ * Sets the cache for the lookup. If cache is provided, then it will be used as an
```
JonasKruckenberg:
```suggestion
 * Sets the `IndexSpecCache` of the lookup. If cache is provided, then it will be used as an
```

memark:
Good catches @JonasKruckenberg.
I added those assertions (in the form of _expects_ (because of _Google Test_ reasons)).

src/rlookup.h
```diff
@@ -465,10 +465,10 @@ typedef struct {
 int RLookup_LoadDocument(RLookup *lt, RLookupRow *dst, RLookupLoadOptions *options);
 
 /**
- * Initialize the lookup. If cache is provided, then it will be used as an
+ * Sets the `IndexSpecCache` of the lookup. If cache is provided, then it will be used as an
  * alternate source for lookups whose fields are absent
  */
-void RLookup_Init(RLookup *l, IndexSpecCache *cache);
+void RLookup_SetCache(RLookup *l, IndexSpecCache *cache);
```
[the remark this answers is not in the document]
memark: Nope.

src/rlookup.c
```diff
@@ -434,8 +434,7 @@ size_t RLookup_GetLength(const RLookup *lookup, const RLookupRow *r, bool *skipF
   return nfields;
 }
 
-void RLookup_Init(RLookup *lk, IndexSpecCache *spcache) {
-  memset(lk, 0, sizeof(*lk));
```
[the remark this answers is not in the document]
memark: No different than before.

src/coord/hybrid/dist_hybrid_plan.cpp
```diff
@@ -94,7 +94,7 @@ arrayof(char*) HybridRequest_BuildDistributedPipeline(HybridRequest *hreq,
 
     RLookup *tailLookup = AGPLN_GetLookup(HybridRequest_TailAGGPlan(hreq), NULL, AGPLN_GETLOOKUP_FIRST);
     // Init lookup since we dont call buildQueryPart
-    RLookup_Init(tailLookup, IndexSpec_GetSpecCache(hreq->sctx->spec));
+    RLookup_SetCache(tailLookup, IndexSpec_GetSpecCache(hreq->sctx->spec));
```
nafraf, line 97: Should we update the comment?

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
```diff
@@ -498,7 +498,7 @@ pub unsafe extern "C" fn RLookup_Init(
         unsafe { IndexSpecCache::from_raw(spcache) }
     });
 
-    lookup.init(spcache);
+    lookup.set_cache(spcache);
```
nafraf, line 501: Consider updating the description of the function.

tests/cpptests/test_cpp_hybridmerger.cpp
```diff
@@ -174,18 +174,16 @@ HybridLookupContext* CreateDummyLookupContext(size_t numUpstreams) {
   // Create dummy RLookup for each upstream
   for (size_t i = 0; i < numUpstreams; i++) {
     RLookup *dummyLookup = (RLookup*)rm_calloc(1, sizeof(RLookup));
-    if (dummyLookup) {
-      RLookup_Init(dummyLookup, NULL);
-    }
+    EXPECT_NE(dummyLookup, nullptr);
```
nafraf, line 177: nitpick: Since we are using `RLookup_Init(dummyLookup, NULL);`, can we use `rm_malloc` instead of `rm_calloc`?

tests/cpptests/test_cpp_rlookup.cpp
```diff
@@ -15,13 +15,11 @@ class RLookupTest : public ::testing::Test {};
 
 TEST_F(RLookupTest, testInit) {
```
nafraf: What are we testing here? Just that we don't crash?

tests/cpptests/test_cpp_hybridmerger.cpp
memark: Well, that particular line is not there anymore, now it's just `*dummyLookup = RLookup_New();`.

But I guess you suggested the change to improve performance? There seems to be a mix of `calloc` and `malloc` in the tests, so I don't mind switching.

tests/cpptests/test_cpp_rlookup.cpp
memark: I don't know. But it doesn't seem very meaningful. I'll remove it!

tests/cpptests/test_cpp_hybridmerger.cpp
nafraf: Yes, it was to improve (maybe negligible) performance.
If you change it, can you update line 183 as well?

memark: Ofc!

nafraf (approved):

PR #8300, opened 2026-02-06
[MOD-13594] RSValue: Implement new string setup - Part 3 of 3

daxhuiberts:
The existing codebase expects the strings inside `RsString` to be nul terminated. But at least one location in the codebase (`hightlight_processor.c`) provides non-nul terminated strings.

This adds some debug asserts to `RsString` to strictly track nul termination of provided string and adds safe getters which panic if those guarantees cannot be upheld. In release mode this falls back to the original behaviour as to not have an impact on performance.

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

#### Release Notes

- [ ] This PR requires release notes
- [x] This PR does not require release notes

---

> [!NOTE]
> **Medium Risk**
> Touches core `RSValue`/`RsString` FFI string handling and length semantics; mistakes could surface as panics/asserts in debug or incorrect lengths/memory handling in C callers.
> 
> **Overview**
> **Ensures `RsString` invariants match C expectations by requiring nul-terminated backing buffers.** `highlight_processor.c` now appends a `\0` when building summarized/highlighted strings and passes `len-1` to `RSValue_NewString` so the reported length excludes the terminator.
> 
> On the Rust side, `RsString` is reworked to always be nul-terminated (including for Rust-allocated strings via `from_vec`), with debug assertions for terminators in `rm_alloc_string`/`borrowed_string`, updated drop logic for Rust-allocated data, and `RSValue_NewCopiedString` now building strings from raw bytes without `CString` restrictions. FFI docs/headers and tests are updated accordingly, and a new `RsString` test module is added.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4c7235cd1161f121ef132dbe8abf1defa356a880. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #8472, opened 2026-02-24
[MOD-13590] RsValue: Conversion functions

daxhuiberts:
Changes existing RSValue conversion functionality in expression.c and strings.c and add the remaining RsValue conversion functions to the rust implementation as well.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches core `RSValue` conversion/formatting paths and changes the `RSValue_ToNumber` API to return `bool`, which could impact callers and output formatting if any edge cases differ between C and Rust implementations.
> 
> **Overview**
> **RSValue conversion/formatting is refactored and extended, with Rust FFI parity improvements.** The PR removes the legacy `RSValue_ToString` path, adds `RSValue_DereferenceRefAndTrio`, and updates string functions (e.g. `to_str`, `format`) and evaluator replies to explicitly handle numbers/strings (including trio/reference dereferencing) using `RSValue_NumToString`.
> 
> On the Rust FFI side, it adds missing constructors (`RSValue_NewParsedNumber`, `RSValue_NewNumberFromInt64`) and introduces a new conversions module implementing `RSValue_ToNumber` and `RSValue_NumToString`, backed by shared parsing/formatting utilities for consistent numeric rendering. Several call sites and tests are updated accordingly, and numeric reply buffers are standardized to 32 bytes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8acedd352acee216bc0b3acac646c5473e72d8b0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/aggregate/expr/expression.c
+++ b/src/aggregate/expr/expression.c
@@ -569,7 +569,7 @@ void RPEvaluator_Reply(RedisModule_Reply *reply, const char *title, const Result
         RSValue_NumToString(v, buf, sizeof(buf));
         RedisModule_Reply_SimpleStringf(reply, "%s - Literal %s", typeStr, buf);
       } else {
-        RedisModule_Reply_SimpleStringf(reply, "%s - Literal", typeStr);
+        RedisModule_Reply_SimpleStringf(reply, "%s - Literal ", typeStr);
       }
       break;
     }


src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,99 @@
...
+    };
+
+    if let Some(len_ptr) = unsafe { len_ptr.as_mut() } {
+        *len_ptr = len;
+    }
+    ptr
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_ToString(dst: *const RsValue, value: *const RsValue) {
```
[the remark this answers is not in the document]
daxhuiberts: Fixed

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,99 @@
...
+    let mut dst = unsafe { expect_shared_value(dst) };
+
+    let value = unsafe { expect_value(value) };
+    let value = value.fully_dereferenced_ref();
```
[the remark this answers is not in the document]
daxhuiberts: Fixed

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
```diff
@@ -199,3 +200,46 @@ pub unsafe extern "C" fn RSValue_NewCopiedString(str: *const c_char, len: u32) -
...
+/// Creates and returns a new **owned** [`RsValue::Number`] from an `i64`.
+///
+/// The `i64` is cast to `f64`, which may lose precision for values outside
+/// the exact representable range of `f64`.
+///
+/// The caller must make sure to pass the returned [`RsValue`] to one of the
+/// ownership taking `RSValue_` functions, directly or indirectly.
+#[unsafe(no_mangle)]
+pub extern "C" fn RSValue_NewNumberFromInt64(number: i64) -> *mut RsValue {
+    SharedRsValue::new(RsValue::Number(number as f64))
```
daxhuiberts, line 229: The C code simply interprets the `i64` as an `f64`: https://github.com/RediSearch/RediSearch/blob/master/src/value/value.c#L269

That seems like the incorrect thing to do though. Should we do the same here or convert it differently?

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,191 @@
...
+        RsValue::String(_) => {
+            // Safety: ensured by caller (1.)
+            let shared_value = unsafe { expect_shared_value(value) };
+            RsValue::Ref(SharedRsValue::clone(&shared_value))
+        }
+        RsValue::RedisString(string) => {
+            let (ptr, len) = string.as_ptr_len();
+            let len = len.try_into().expect("len > u32::MAX");
+            // Safety: The redis string points to a valid c-string?
+            let string = unsafe { RsString::borrowed_string(ptr, len) };
```
daxhuiberts: Even though some places in the C code expects c-strings from an `RsString`, we can't guarantee it here either, because a `RedisString` is like an `&[u8]` right? So I don't know how to convert this to an `RsValue` with those assumptions. How to handle this properly?

src/redisearch_rs/value/src/util.rs
```diff
@@ -0,0 +1,44 @@
...
+            )
+        }
+    };
+
+    if result < 0 {
+        panic!("snprintf failed")
+    }
+
+    result as usize
+}
```
daxhuiberts: I don't like this implementation. The C code first checks to see if the float can be represented as an integral value. If so, it formats it like one. Else, it formats it with `"%.12g"`, for which no direct Rust replacement exists.

`"%.12g"` trims trailing zeros and automatically converts to scientific notation if needed. There are some existing C tests which specifically test this formatting. Not sure if we really need to keep that though. I much prefer to have a simpler Rust only solution here.

LukeMathWalker: Let's double-check with Redis engineers whether bit-for-bit matching is important. 
If it is, we can definitely emulate it in Rust even though we don't have a one-for-one replacement as a `format!` specifier.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
LukeMathWalker: What do you mean by incorrect? It won't be 1:1 for values that are too large/small, but is that an issue?

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
LukeMathWalker: Looking at the different PRs, I think the correct course of action is to give up on guaranteeing that an `RsValue::String` is a valid `CString`. The best we can do is treat it as a sequence of `u8` bytes with may or may not be nul-terminated.
If the caller is imposing or projecting additional guarantees on it, it's their responsibility.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
daxhuiberts: Oh never mind, I incorrectly thought that assigning an `int64_t` to a `double` in C would do a bitwise copy. But it does the same as `number as f64` in rust. So no issue.

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
daxhuiberts: Yeah, that sounds like the better way to approach this.

src/redisearch_rs/value/src/util.rs
daxhuiberts: Will check with the team.

daxhuiberts: I tried to change it to a simpler rust version, but quite some python tests failed. Will keep it like this.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
```diff
@@ -183,19 +184,58 @@ pub unsafe extern "C" fn RSValue_NewRedisString(str: *mut RedisModuleString) ->
...
 }
+
+/// Creates and returns a new **owned** [`RsValue::Number`] by parsing the given
+/// string as a floating-point number. Returns a null pointer if the string
+/// cannot be parsed.
+///
+/// The caller retains ownership of `value`.
+///
+/// The caller must make sure to pass the returned [`RsValue`] to one of the
+/// ownership taking `RSValue_` functions, directly or indirectly.
```
JonasKruckenberg: this is not super actionable and raises more questions than it answers (why do I need to pass it to them? what happens if I don't? what does indirectly mean?)

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,192 @@
...
+
+use crate::util::{expect_shared_value, expect_value};
+use libc::size_t;
+use std::ffi::{c_char, c_double, c_int};
+use value::util::{num_to_str, str_to_float};
+use value::{RsString, RsValue, SharedRsValue};
+
+/// Convert a value to a number, either returning the actual numeric values or by parsing
+/// a string into a number. Return 1 if the value is a number or a numeric string that can
+/// be converted, or 0 if not. The converted number is written to the `d` pointer.
```
JonasKruckenberg:
```suggestion
/// Convert the `RsValue` to a number. Returns `1` when this value is  a number 
/// or a numeric string that can be converted and writes the number to `d`. If 
/// the value cannot be converted `0` is returned and nothing is written to `d`.
```

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,192 @@
...
+///
+/// # Safety
+///
+/// 1. `value` must be either null or point to a valid [`RsValue`] obtained from
+///    an `RSValue_*` function.
+/// 2. `d` must be a [valid], non-null pointer to a `c_double`.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_ToNumber(value: *const RsValue, d: *mut c_double) -> c_int {
```
JonasKruckenberg: This should return `bool` instead of `c_int` for truthy values

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,192 @@
...
+        RsValue::String(string) => str_to_float(string.as_bytes()),
+        RsValue::RedisString(string) => str_to_float(string.as_bytes()),
+        _ => return 0,
+    };
+
+    if let Some(num) = num {
+        *d = num;
+        return 1;
+    } else {
+        *d = 0.0;
```
JonasKruckenberg: Is this required behavior? If so this needs to be mentioned in the doc comment. If this isn't required (which it shouldn't be ideally) you should remove this, a `0`/`false` return code usually indicates nothing is written to the outptr and the value remains uninitialized/untouched.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
memark: In my interpretation it pretty much means "we allocated this for you, you need to free it later on by calling one of the functions we mention here".

src/redisearch_rs/value/src/util.rs
```diff
@@ -0,0 +1,53 @@
...
+    std::str::from_utf8(input).ok()?.parse::<f64>().ok()
+}
+
+/// Converts a float into a string using the c `snprintf` function for compatibility with
+/// expected output. A float is rendered as an integer if possible, else it's rendered
+/// with up to 12 decimal points, else it's rendered in scientific notation.
+///
+/// Returns the amount of bytes written as per `snprint`: if that is >= `buf.len()` then
+/// it didn't fit (`snprintf` always places a nul-terminator at the end).
+pub fn num_to_str(float: f64, buf: &mut [u8]) -> usize {
```
JonasKruckenberg:
```suggestion
pub fn num_to_str(float: f64, buf: &mut [u8; 32]) -> usize {
```

since we know the input type we also know the biggest possible string this could produce (`i64::MIN = 20 bytes` see [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=186c316d402c5581200e5cb786059d28)) lets require a fixed size array here then so callers cannot even pass buffers that are too small.

(picked 32 because stack alignment will likely pad 20 out to 32 anyways and it gives us a bit of leeway in case the target system libcs `snprintf` somehow formats numbers differently to glibc.)

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,192 @@
...
+///
+/// # Safety
+///
+/// 1. `value` must point to a valid [`RsValue`] obtained from an `RSValue_*` function.
+/// 2. `len_ptr` must be a [valid], non-null pointer to a `size_t`.
+/// 3. `buf` must be a [valid] pointer to a writable buffer of at least `buflen` bytes.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_ConvertStringPtrLen(
```
JonasKruckenberg: I don't think this function should exist/be public: There is only one calcite outside of `value.c` and this function is really easy to hold wrong:

1. there is no way for this function to communicate to the caller the scratchpad buffer is too small (we'll just crash) so callers always have to guess.
2. this function returns `\0` instead of `NULL` when the type cannot be converted. no other conversion functions behave like this.

I would refactor the one callsite of this function to use something else (explicitly match on the type of literal).

If you really must keep this function then the doc comment must be much more explicit about the reason for the scratchpad buffer, the size of that scratchpad buffer and the return types in various scenarios.

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,192 @@
...
+///
+/// 1. `dst` must point to a valid **owned** [`RsValue`] obtained from an
+///    `RSValue_*` function returning an owned [`RsValue`] object.
+/// 2. `value` must point to a valid [`RsValue`] obtained from an `RSValue_*` function.
+///
+/// # Panic
+///
+/// Panics if more than 1 reference exists to the `dst` [`RsValue`] object.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_ToString(dst: *mut RsValue, value: *const RsValue) {
```
JonasKruckenberg: this should return the `RsValue` instead of passing to an outptr.

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,192 @@
...
+            RsValue::String(string)
+        }
+    };
+
+    // Panic if more than 1 reference exists
+    dst.set_value(new_value);
+}
+
+/// Formats the numeric value of an [`RsValue::Number`] as a string into the
+/// caller-provided buffer and returns the number of bytes written.
```
JonasKruckenberg, line 53: this should state the buffer must be 32 bytes in size (see comment on `num_to_str`)

src/redisearch_rs/value/src/rs_string.rs
```diff
@@ -17,19 +17,17 @@ enum RsStringKind {
...
 }
 
-/// A `CString` like string for [`RsValue`](crate::RsValue) with support for rust allocated string,
-/// C allocated strings, and constant strings, and support for a max length of `u32::MAX`,
-/// all in one package.
+/// An [`RsString`] is meant to store string data with support for rust allocated data, C
+/// allocated data or borrowed data, and support for a max length of `u32::MAX`.
+/// It can contain binary data and is always nul-terminated, except for [`RsString`] constructed
+/// through `rm_alloc_string_without_nul_terminator`. Code expecting a nul-terminator should
+/// use `as_ptr_len_for_nul_terminated` which have debug asserts ensuring a nul-terminator exists.
```
JonasKruckenberg:
```suggestion
/// It can contain binary data and is always nul-terminated, except when constructed
/// through `rm_alloc_string_without_nul_terminator`. Code expecting a nul-terminator should
/// use `as_ptr_len_for_nul_terminated` which have debug asserts ensuring a nul-terminator exists.
```

src/redisearch_rs/value/src/rs_string.rs
```diff
@@ -17,19 +17,17 @@ enum RsStringKind {
...
-/// - `ptr` must not be NULL and must point to a valid c-string of `len+1` size.
-/// - The size determined by `len` excludes the nul-terminator.
-/// - A nul-terminator is expected in memory at `ptr+len`.
-///
-/// There is one exception for the nul-terminator requirement: [`RsString`] constructed through
-/// `rm_alloc_string_without_nul_terminator`. Code expecting a nul-terminator should use the `_checked`
-/// variants of `as_ptr_len` and `as_bytes` which have debug asserts ensuring a nul-terminator exists.
+/// - `ptr` must not be NULL and must point to valid data of `len+1` size.
+/// - The size determined by `len` (excludes the nul-terminator).
+/// - Apart from the exception, a nul-terminator is expected in memory at `ptr+len`
```
JonasKruckenberg: safety comments on types are not super meaningful bc for a reader there is nothing actionable about this (imagine reading this comment on doc.rs, what do you make of this?) because these are invariants the *module code* must uphold, not the caller. 


If you want to document the invariants of a type that the rest of the module code must uphold, I would recommend a regular comment instead of a doc comment beginning with `# Invariants`

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
```diff
@@ -183,19 +184,58 @@ pub unsafe extern "C" fn RSValue_NewRedisString(str: *mut RedisModuleString) ->
...
+/// # Safety
+///
+/// 1. `value` must be a [valid], non-null pointer to a string buffer.
+/// 2. `value` must be [valid] for reads of `len` bytes.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_NewParsedNumber(value: *const c_char, len: u32) -> *mut RsValue {
+    // Safety: ensured by caller (1., 2.)
+    let slice = unsafe { std::slice::from_raw_parts(value as *const u8, len as usize) };
```
JonasKruckenberg:
```suggestion
    let slice = unsafe { std::slice::from_raw_parts(value.cast::<*const u8>(), len as usize) };
```

lets use as few `as` casts as possible

src/redisearch_rs/value/src/rs_string.rs
```diff
@@ -39,20 +37,24 @@ pub struct RsString {
...
+    /// A nul-terminator is automatically added by this constructor for compatibility
+    /// with existing C code.
+    ///
     /// # Panic
     ///
     /// Panics when the size is larger than `u32::MAX`.
-    pub fn cstring(str: CString) -> Self {
-        let len = str.count_bytes();
+    pub fn from_vec(mut vec: Vec<u8>) -> Self {
+        let len = vec.len();
```
JonasKruckenberg: this should just be a a `From<Vec<u8>>` no?

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
daxhuiberts: So these functions return a `*mut RsValue`, but one that's retrieved from `Arc::into_raw`. So the `*mut RsValue` does not point to an ordinary `RsValue`, but to an `RsValue` which is wrapped inside an `ArcInner`.

Calling an `RSValue_` function with `Box::<RsValue::Null>.into_raw()` is going to fail hard. But an `ArcInner` is a hidden implementation detail of an `Arc` which we can't share or work with directly.

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
daxhuiberts: I used the same return type from the existing C version, but yeah that can probably be replaced with `bool` without issue.

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
daxhuiberts: Yes I got it in there because tests were failing where I didn't and the C version did. But this should probably be fixed at the call site instead.

src/redisearch_rs/value/src/util.rs
daxhuiberts: Yep that's a good one. I need the guarantee that it always fit in other places as well.

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
daxhuiberts: 1. If the buffer is too small, it will return an empty string, not crash, right?

But you're right, I'll look into changing the call site. That is probably better.

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
daxhuiberts: Good point.

This function is even more complex than `RSValue_ConvertStringPtrLen` and also only used in 2 call sites so probably warrants fixing at the call site as well. I'll look into that as well.

src/redisearch_rs/value/src/rs_string.rs
daxhuiberts: Good point. Those are invariants. Will change the docs accordingly.

daxhuiberts:
> Can yo explain briefly why we need to go from CString to Vec<u8>? We are constructing valid cstrs in the end anyways, so why not encode that requirement directly?

We are putting a nul-terminator at the end, but there is no requirement that the string doesn't contain nul-terminators within. So not an actual `CString`. I took over that behaviour because as I understand it now that is how other strings in redis (e.g. `RedisString` and `sds`) also work and those type of strings are passed in `RsValue::String`.

It's a bit weird to have it require a nul-terminator, but not be an actual `CString`. Some code even does `strcpy` or `strlen` type operations on `RsValue::String` data, but it's also used to store binary data iiuc.

> We should use as little as casts as possible (especially for pointers but also numbers)
> nitpick: I personally like to keep the # Panics over # Safety as a style choice but don't feel strongly about that.

👍

src/redisearch_rs/value/src/rs_string.rs
daxhuiberts: Yeah that's possible, but I want to be explicit about the fact that this function adds a nul-terminator so has a bit of special behaviour.

JonasKruckenberg: okay fair yeah 🤔

JonasKruckenberg:
> We are putting a nul-terminator at the end, but there is no requirement that the string doesn't contain nul-terminators within. So not an actual CString. I took over that behaviour because as I understand it now that is how other strings in redis (e.g. RedisString and sds) also work and those type of strings are passed in RsValue::String.
> 
> It's a bit weird to have it require a nul-terminator, but not be an actual CString. Some code even does strcpy or strlen type operations on RsValue::String data, but it's also used to store binary data iiuc.

urgh that's annoying but fair enough

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,192 @@
...
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_ToNumber(value: *const RsValue, d: *mut c_double) -> c_int {
+    // Safety: ensured by caller (1.)
+    let Some(value) = (unsafe { value.as_ref() }) else {
+        return 0;
+    };
+
+    // Safety: ensured by caller (2.)
+    let d = unsafe { d.as_mut().expect("d is null") };
```
memark: I'd put this one on top to keep the `value` variable logic tighter together.

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,192 @@
...
+    let value = value.fully_dereferenced_ref();
+
+    let (ptr, len): (*const c_char, size_t) = match value {
+        RsValue::Number(num) => {
+            // Safety: ensured by caller (3.)
+            let buffer = unsafe { std::slice::from_raw_parts_mut(buf as *mut u8, buflen as usize) };
+            let n = num_to_str(*num, buffer);
+
+            if n < buflen {
+                (buf as *const _, n)
```
memark: can we avoid these wildcard casts?

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,192 @@
...
+        }
+        RsValue::Number(number) => {
+            let mut buf = [0u8; 32];
+            // `num_to_str` formatting should fit well within the 32 byte sized buffer.
+            let len = num_to_str(*number, &mut buf);
+            let vec = buf[..(len as usize)].to_vec();
+            RsValue::String(RsString::from_vec(vec))
+        }
+        _ => {
+            // Safety: It is safe to use an const here.
```
memark:
```suggestion
            // Safety: It is safe to use a const here.
```

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,192 @@
...
+            let len = len.try_into().expect("len > u32::MAX");
+            // Safety: A redis strings content is also valid content for an `RsString`.
+            let string = unsafe { RsString::borrowed_string(ptr, len) };
+            RsValue::String(string)
+        }
+        RsValue::Number(number) => {
+            let mut buf = [0u8; 32];
+            // `num_to_str` formatting should fit well within the 32 byte sized buffer.
+            let len = num_to_str(*number, &mut buf);
+            let vec = buf[..(len as usize)].to_vec();
```
memark:
```suggestion
            let len = num_to_str(*number, &mut buf) as usize;
            let vec = buf[..len].to_vec();
```

src/redisearch_rs/value/src/rs_string.rs
memark: This is not all clear to me. Given that our Rust code is configured to use the same `RmAlloc` allocator as C (right?), this looks a bit strange. Maybe call them `RustAlloc` and `CAlloc` if we need to differentiate on the language used?

src/redisearch_rs/value/src/rs_string.rs
```diff
@@ -39,20 +37,24 @@ pub struct RsString {
...
         assert!(len <= u32::MAX as usize);
 
-        let ptr = str.into_raw();
+        vec.push(b'\0');
+        let ptr = Box::into_raw(vec.into_boxed_slice());
 
         Self {
-            ptr,
+            ptr: ptr as *const c_char,
             len: len as u32,
```
memark, line 59: random thought: it seems like we're casting this `len` to `usize` all of the place. could we store it as such instead?

src/redisearch_rs/value/src/rs_string.rs
```diff
@@ -166,8 +168,12 @@ impl Drop for RsString {
     fn drop(&mut self) {
         match self.kind {
             RsStringKind::RustAlloc => {
-                // SAFETY: `self.ptr` was created by `CString::into_raw` and has not been freed.
-                drop(unsafe { CString::from_raw(self.ptr as *mut _) });
+                let slice = std::ptr::slice_from_raw_parts_mut(
+                    self.ptr as *mut u8,
+                    (self.len as usize) + 1,
+                );
+                // SAFETY: Boxed slice was created in `Self::from_vec` which has `len + 1` bytes.
```
memark:
```suggestion
                // Safety: Boxed slice was created in `Self::from_vec` which has `len + 1` bytes.
```

src/redisearch_rs/value/src/util.rs
```diff
@@ -0,0 +1,53 @@
...
+                integer,
+            )
+        }
+    } else {
+        // Safety: buf is valid by definition, formatting string and arguments match up.
+        unsafe {
+            snprintf(
+                buf.as_mut_ptr() as *mut c_char,
+                buf.len(),
+                c"%.12g".as_ptr(),
```
memark, line 43: do we want to extract the magi precision part of this string for visibility or reuse?

src/redisearch_rs/value/src/util.rs
```diff
@@ -0,0 +1,53 @@
...
+
+/// Converts a float into a string using the c `snprintf` function for compatibility with
+/// expected output. A float is rendered as an integer if possible, else it's rendered
+/// with up to 12 decimal points, else it's rendered in scientific notation.
+///
+/// Returns the amount of bytes written as per `snprint`: if that is >= `buf.len()` then
+/// it didn't fit (`snprintf` always places a nul-terminator at the end).
+pub fn num_to_str(float: f64, buf: &mut [u8]) -> usize {
+    let integer = float as i64;
+    let result = if integer as f64 == float {
```
memark: This is a bit magic and hard to read for me. Are we doing a double cast back and forth? Could we break it up a bit?

src/redisearch_rs/value/src/rs_string.rs
daxhuiberts: `RSValue` strings in the C version are limited to a `u32` with code depending on that size. That might be worth looking into, or at least making the code easier to use, but it's out of scope for this PR.

src/redisearch_rs/value/src/util.rs
daxhuiberts: IMO it's fine. We don't reuse it anywhere else.

daxhuiberts:
@memark, @JonasKruckenberg I've processed your feedback. All the changes and comments related to the `RsString` I've moved to https://github.com/RediSearch/RediSearch/pull/8300 instead. Hope you can take another look.

src/aggregate/expr/expression.c
JonasKruckenberg, line 1: I like this much better, well done 👍🏻

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
JonasKruckenberg: > In my interpretation it pretty much means "we allocated this for you, you need to free it later on by calling one of the functions we mention here".

yes but it should just say that 😄 we cannot expect readers to come to the same interpretation if they e.g. aren't that familiar with Rust.

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
JonasKruckenberg: this might seem like a weird suggestion but hear me out:
I think we should keep the `buf_len` param, use it to assert the buf capacity is `>= 32` and then pass it to `num_to_str`.

Currently the function is just very easy to hold wrong, and we do not have a way of debug asserting that the capacity is large enough. keep the `buf_len` param forces callers to provide the length, maybe even causing them to think for a second what length actually is needed.

JonasKruckenberg: plus if we keep the assert as a `debug_assert(buf_len >= 32)` clang LTO will just get rid of the param in release builds

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
daxhuiberts: I'm open for suggestions. Most often `RsValue`s are passed around and not freed directly by the creator which is important to take into account.

src/aggregate/expr/expression.c
```diff
@@ -526,17 +526,24 @@ void RPEvaluator_Reply(RedisModule_Reply *reply, const char *title, const Result
...
+      if (RSValue_IsString(v)) {
+        size_t len;
+        const char *ptr = RSValue_StringPtrLen(v, &len);
+        RedisModule_Reply_SimpleStringf(reply, "%s - Literal %s", typeStr, ptr);
+      } else if (RSValue_IsNumber(v)) {
+        char buf[32];
+        RSValue_NumToString(v, buf, sizeof(buf));
+        RedisModule_Reply_SimpleStringf(reply, "%s - Literal %s", typeStr, buf);
+      } else {
+        RedisModule_Reply_SimpleStringf(reply, "%s - Literal", typeStr);
```
kei-nan: Previously we always had a space after Literal.
Do we want to preserve this behaviour?

src/aggregate/functions/string.c
```diff
@@ -131,7 +131,30 @@ int func_to_number(ExprEval *ctx, RSValue **argv, size_t argc, RSValue *result)
...
+    }
+    case RSValueType_Number: {
+      char tmpbuf[32];
+      size_t len = RSValue_NumToString(v, tmpbuf, sizeof(tmpbuf));
+      char *buf = rm_strdup(tmpbuf);
+      RSValue_SetString(result, buf, len);
+      break;
+    }
+    default:
+      RSValue_SetConstString(result, "", 0);
```
kei-nan, line 156: do we want to add an assertion here about the expected rsvalue type?
I think in old code it was the null type, correct?

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
kei-nan: Maybe give an agent this thread and ask for suggestions on a how the documentation could be improved.

src/aggregate/expr/expression.c
daxhuiberts: Keen eye. Yes, to be cautious I think that is best.

src/aggregate/functions/string.c
daxhuiberts: The old code did a
```
case RSValueType_Null
default:
   ...
```
The `RSValueType_Null` match was redundant as it handled all the rest of the types as well (undefined, array and map)

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
daxhuiberts: Ok, came up with: 'The returned [`RsValue`] is heap-allocated. The caller must ensure it is eventually passed to [`RsValue_DecrRef`]. Ownership may be transferred through other `RSValue_` functions before that happens.' Let me know what you think, I'm quite happy with this.

src/redisearch_rs/c_entrypoint/value_ffi/src/conversions.rs
```diff
@@ -0,0 +1,84 @@
...
+    buf: *mut c_char,
+    buflen: size_t,
+) -> size_t {
+    // Safety: ensured by caller (1.)
+    let value = unsafe { expect_value(value) };
+
+    debug_assert!(buflen >= 32);
+    // Safety: ensured by caller (2.)
+    let buf = unsafe { std::slice::from_raw_parts_mut(buf.cast::<u8>(), buflen) };
+    let buf = buf.first_chunk_mut().unwrap();
```
JonasKruckenberg, line 77: not zero cost anymore now, but probably fine

PR #8478, opened 2026-02-24
[MOD-13589] RsValue: Debug rendering

daxhuiberts:
This PR adds debug rendering for `RsValue`. It has a special `DebugFormatter` which implements `Debug` and can obfuscate it's debug output. It wraps `sds` in a `SdsWriter` which implements the `Write` trait which can be used by the `DebugFormatter`. Finally `RsValue_DumpSds` wires all of this together for its output.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Introduces new FFI surface (`RSValue_DumpSds`) and relies on unsafe SDS mutation and C-side obfuscation functions, which could impact memory safety or logging behavior if misused. Changes are mostly additive and covered by new integration tests, but touch cross-language boundaries and build/bindgen inputs.
> 
> **Overview**
> Adds a new `value::debug::DebugFormatter` and `RsValue::debug_formatter()` to produce a stable `Debug` rendering of `RsValue`, with optional obfuscation of text/number values via C obfuscation APIs and recursive handling for arrays/maps.
> 
> Exposes this to C through a new `RSValue_DumpSds` function that appends the formatted output into an SDS string using a new `SdsWriter` adapter, and updates bindgen inputs/headers (`sds.h`, `obfuscation_api.h`, and `headers/value.h`) accordingly.
> 
> Cleans up test/linking setup by moving Redis symbol stubbing to a shared integration test `main.rs`, adding integration tests for the new debug output, and adjusting crate dependencies (drop `value_ffi` from `rqe_iterators`, add `redisearch_rs` dev-dep for `value`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d26b390fcd26d889d5cbb4446690d11899d88ef2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/redisearch_rs/c_entrypoint/value_ffi/src/sds.rs
```diff
@@ -0,0 +1,104 @@
...
+        RsValue::Undefined => sds.write_str("<Undefined>"),
+        RsValue::Null => sds.write_str("NULL"),
+        RsValue::Number(num) => {
+            if obfuscate {
+                sds.write_bytes(obfuscate_number(*num));
+            } else {
+                let mut buf = [0; 32];
+                let n = value::util::num_to_str(*num, &mut buf).unwrap();
+                sds.write_bytes(&buf[0..n]);
+            }
```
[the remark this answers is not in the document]
daxhuiberts: buf is large enough for num_to_str to not overfill.

daxhuiberts:
Got a setup for `RSValue_DumpSds` here. Extracted all the unsafe logic out of the `dump_sds` into a `SdsWriter` and `obfuscate` helpers.

I'm not sure whether using `std::fmt::Write` or `std::io::Write` would make sense here.
- `std::fmt::Write` works with `&str`, but I'm using `&[u8]` here from `RsValue::String` and `RsValue::RedisString` as well as from the obfuscate functions.
- `std::io::Write` is meant for fallible IO, but all writing here happens to a string in memory which is infallible (except for OOM, which we don't handle directly).

I still want to move dump_sds to the `value` crate and test it. For that I need to mock `sdscatlen` in that test, or make a generic writer trait specifically for this (or decide to use one of the std `Write` traits). My feeling goes to mocking `sdscatlen` (and the `obfuscate` functions). Any suggestion?

Cursorbot also warns that calling `Obfusctate_Text` with `&[u8]` is going to be a problem because it's not nul-terminated (again), which we can't guarantee. How to handle that here? The implementation of `Obfusctate_Text` is always simply returning `"Text"`, but not sure we simply want to inline that.

LukeMathWalker:
> Got a setup for RSValue_DumpSds here. Extracted all the unsafe logic out of the dump_sds into a SdsWriter and obfuscate helpers.
> I'm not sure whether using std::fmt::Write or std::io::Write would make sense here.
> std::fmt::Write works with &str, but I'm using &[u8] here from RsValue::String and RsValue::RedisString as well as from the obfuscate functions.
> std::io::Write is meant for fallible IO, but all writing here happens to a string in memory which is infallible (except for OOM, which we don't handle directly).
> I still want to move dump_sds to the value crate and test it. For that I need to mock sdscatlen in that test, or make a generic writer trait specifically for this (or decide to use one of the std Write traits). My feeling goes to mocking sdscatlen (and the obfuscate functions). Any suggestion?

I would go for `std::io::Write`. You only need an `unwrap()` at the call-site which knows the concrete type of the writer (and thus the fact that it's safe to unwrap). Which then allows you to test the formatting bits using a `&mut Vec<u8>`.

LukeMathWalker:
> Cursorbot also warns that calling Obfusctate_Text with &[u8] is going to be a problem because it's not nul-terminated (again), which we can't guarantee. How to handle that here? The implementation of Obfusctate_Text is always simply returning "Text", but not sure we simply want to inline that.

We can change `Obfuscate_Text` to take a pointer and a length, if we want to go the extra mile. As we insert more and more Rust into the codebase, we'll naturally move away from nul-terminating things anyway, so it's safer to rework APIs to deal with slices where possible (which may or may not include a nul-termination byte).

src/redisearch_rs/c_entrypoint/value_ffi/src/debug.rs
```diff
@@ -0,0 +1,16 @@
...
+use std::io::Write;
+use value::RsValue;
+use value::sds_writer::SdsWriter;
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_DumpSds(value: *const RsValue, sds: sds, obfuscate: bool) -> sds {
+    let mut writer = unsafe { SdsWriter::new(sds) };
+
+    match unsafe { value.as_ref() } {
+        None => write!(writer, "nil").unwrap(),
```
JonasKruckenberg, line 34: are we allowed to change the output format? bc I think this should probably say `<nil>` or `<None>` to use the same formatting as `<Undefined>` and `<non-utf8-data>`

src/redisearch_rs/value/src/debug.rs
```diff
@@ -0,0 +1,71 @@
...
+                });
+                f.debug_map().entries(entries).finish()
+            }
+            RsValue::Ref(ref_value) => ref_value.value().debug_formatter(self.obfuscate).fmt(f),
+            RsValue::Trio(trio) => trio.left().value().debug_formatter(self.obfuscate).fmt(f),
+        }
+    }
+}
+
+fn obfuscate_number(number: f64) -> &'static str {
```
JonasKruckenberg, line 84: `&'static str` really?

src/redisearch_rs/c_entrypoint/value_ffi/src/debug.rs
daxhuiberts: I will ask the redis team. This is what the existing C implementation outputs.
The only thing I added is the `<non-utf8-data>`, because it's not possible to push binary data through `fmt::Formatter`.

src/redisearch_rs/value/src/debug.rs
daxhuiberts: Currently the C `Obfuscate_Number` and `Obfuscate_Text` simply return a const string.
But ideally this should depend on the input, but what lifetime should this have or how should this be wrapped?

src/redisearch_rs/c_entrypoint/value_ffi/src/debug.rs
daxhuiberts: And binary data is probably not intended to be read out in debug info.

src/redisearch_rs/value/src/debug.rs
JonasKruckenberg: yeah okay makes sense, don't think there is a much better way to handle this then

src/redisearch_rs/c_entrypoint/value_ffi/src/debug.rs
```diff
@@ -0,0 +1,16 @@
+use ffi::sds;
+use std::io::Write;
+use value::RsValue;
+use value::sds_writer::SdsWriter;
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RSValue_DumpSds(value: *const RsValue, sds: sds, obfuscate: bool) -> sds {
```
memark, line 28: It might be only me, but looks _really_ weird (and confusing) to have a lowercase type name like `sds`.
Could we consider aliasing it in our Rust code to `SDS` or `Sds`?

src/redisearch_rs/value/src/sds_writer.rs
```diff
@@ -0,0 +1,27 @@
...
+pub struct SdsWriter {
+    sds: sds,
+}
+
+impl SdsWriter {
+    pub unsafe fn new(sds: sds) -> Self {
+        Self { sds }
+    }
+
+    pub fn extract_sds(self) -> sds {
```
memark: maybe call it `into_inner`?

src/redisearch_rs/value/tests/integration/debug.rs
```diff
@@ -0,0 +1,108 @@
...
+    format!("{:?}", value.debug_formatter(true))
+}
+
+#[test]
+fn debug_null() {
+    assert_eq!(debug(&RsValue::Null), "NULL");
+}
+
+#[test]
+fn debug_undefined() {
```
memark, line 26: consider pulling these tests (at least the simpler ones) together into one [parametrized test](https://docs.rs/rstest/latest/rstest/#creating-parametrized-tests).

src/redisearch_rs/c_entrypoint/value_ffi/src/debug.rs
daxhuiberts: Yeah it looks a bit weird, but I think keeping it exactly the same as in C is still better. We can always wrap it in a helper struct, which I've done with the `SdsWriter`.

src/redisearch_rs/value/src/sds_writer.rs
daxhuiberts: `into_inner` feels good for generic types, for this it might be better to simply call it `into_sds`.

src/redisearch_rs/value/tests/integration/debug.rs
daxhuiberts: I like parameterized tests, but in this case it works for probably the first three tests. `debug_number_obfuscated` already deviates and all subsequent ones already become a bit too complex to fit the parameterization.

src/redisearch_rs/c_entrypoint/value_ffi/src/debug.rs
daxhuiberts: I'll discuss this with the PR review from the Redis team.

src/redisearch_rs/value/src/debug.rs
```diff
@@ -0,0 +1,99 @@
...
+            RsValue::Trio(trio) => trio.left().value().debug_formatter(self.obfuscate).fmt(f),
+        }
+    }
+}
+
+/// Returns a static string representation of the obfuscated number.
+fn obfuscate_number(number: f64) -> &'static str {
+    // SAFETY: `Obfuscate_Number` is a C function that returns a pointer to a
+    // static null-terminated string.
+    let obfuscated = unsafe { Obfuscate_Number(number) };
```
Itzikvaknin, line 87: This is a dependency in the opposite direction of what we are used to have - make sure its documented in the C side

src/redisearch_rs/value/src/sds_writer.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// # Safety
+    ///
+    /// `sds` must be a [valid], non-null SDS string allocated by the C SDS library.
+    ///
+    /// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+    pub const unsafe fn new(sds: sds) -> Self {
+        Self { sds }
+    }
+
+    /// Consumes the writer and returns the underlying SDS string.
```
Itzikvaknin: Where do you consume the writer?
The separation between write and flush feels a bit like a lazy mechanism of `write to sds only upon consumption` - maybe the empty flush implementation gives a sense of incomplete implementation - Can you please share what are you trying to achieve? is the flush part of some contact Im not aware of?

src/redisearch_rs/value/build.rs
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+fn main() {
+    // #[cfg(feature = "unittest")]
+    // build_utils::bind_foreign_c_symbols();
+}
```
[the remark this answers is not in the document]
Itzikvaknin: @daxhuiberts ?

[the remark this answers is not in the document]
daxhuiberts: This is a weird `cargo test` issue where `value` crate specific configuration has effect on the `value-ffi` test suite which shouldn't happen. Disabling this (even though it is needed) has the effect that the test suite on a full workspace run does work, but an individual package run doesn't.

I notice this is an issue with more packages so I'm looking into whether fixing this should happen in a separate story/PR or not.

src/redisearch_rs/value/src/debug.rs
daxhuiberts: That's a good one. Though I don't see this happen in other cases. Rust is just a consumer, similar to any other C function which includes and links to this API.

src/redisearch_rs/value/src/sds_writer.rs
daxhuiberts: 'consume' is probably not the right word here. `extract` might be a better term here. We're just unwrapping/extracting the inner sds object from the writer wrapper.

That this function has both a `write` and a `flush` method is because of the `Write` trait being implemented which is most often used for IO where flushing is an important detail. In our implementation however all the writing happens immediately and flushing is a noop.

A non IO based `Write` trait specific for memory only writing doesn't exist. Though it conceptually makes sense, Rust decided that using the IO `Write` trait for this use case is good.

Itzikvaknin: So lets update the comment

src/redisearch_rs/c_entrypoint/value_ffi/src/debug.rs
memark: Did you have the discussion with the Redis team?

src/redisearch_rs/value/src/debug.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub(crate) value: &'a RsValue,
+    pub(crate) obfuscate: bool,
+}
+
+impl<'a> Debug for DebugFormatter<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fn fmt_text(f: &mut fmt::Formatter<'_>, text: &[u8], obfuscate: bool) -> fmt::Result {
+            if obfuscate {
+                write!(f, "\"{}\"", obfuscate_text(text))
+            } else if let Ok(s) = std::str::from_utf8(text) {
```
memark: consider `use std::str;`

src/redisearch_rs/value/build.rs
```diff
@@ -0,0 +1,13 @@
...
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+fn main() {
+    // #[cfg(feature = "unittest")]
+    // build_utils::bind_foreign_c_symbols();
```
memark: Regardless of the reason, we should not leave it like this. Either fix the issue or remove the code.

daxhuiberts:
@memark Removed all the specific build.rs stuff. Can you check it one more time?

PR #8480, opened 2026-02-24
[MOD-14192] take read indexspec lock on RDB load on SAVE

JoanFM:
When we save RDB from the main thread (on SAVE), make sure to lock the IndexSpec to ensure safe concurrent access (Specially with GC)

---

> [!NOTE]
> **Medium Risk**
> Touches RDB save code and introduces conditional locking, which could impact SAVE latency or deadlock behavior if lock ordering is wrong, but the change is small and scoped to disk/SST save flows.
> 
> **Overview**
> Prevents potential crashes/corruption when running `SAVE` with disk (SST) indexes by taking an `IndexSpec` **read lock** while serializing disk-related in-memory structures.
> 
> The lock is only taken when saving from the main process (non-fork); forked child saves keep the previous behavior to avoid unnecessary locking.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e9c897dc8f5ec1e886b6ef675cfc43443309cfe0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -4212,12 +4212,12 @@ void Indexes_EndLoading() {
 // Compaction FFI Functions (called by Rust during GC)
 // =============================================================================
 
-// Acquire IndexSpec read/write lock in exclusive mode
+// Acquire IndexSpec write lock
 void IndexSpec_AcquireWriteLock(IndexSpec* sp) {
   pthread_rwlock_wrlock(&sp->rwlock);
 }
 
-// Release IndexSpec read/write
+// Release IndexSpec write lock
 void IndexSpec_ReleaseWriteLock(IndexSpec* sp) {
   pthread_rwlock_unlock(&sp->rwlock);
 }


src/spec.c
```diff
@@ -3221,9 +3221,19 @@ void IndexSpec_RdbSave(RedisModuleIO *rdb, IndexSpec *sp) {
...
+    // In a forked child process, the memory is a snapshot so no lock is needed.
+    bool inFork = RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_IS_CHILD;
+    if (!inFork) {
+      IndexSpec_AcquireReadLock(sp);
+    }
     IndexScoringStats_RdbSave(rdb, &sp->stats.scoring);
     TrieType_GenericSave(rdb, sp->terms, false, true);
     SearchDisk_IndexSpecRdbSave(rdb, sp->diskSpec);
+    if (!inFork) {
+      IndexSpec_ReleaseWriteLock(sp);
```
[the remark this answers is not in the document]
raz-mon: Indeed @JoanFM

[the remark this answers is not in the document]
raz-mon: From the C code we're using the `RedisSearchCtx_LockSpecRead` and `RedisSearchCtx_UnlockSpec` APIs

src/spec.c
```diff
@@ -4201,16 +4211,21 @@ void Indexes_EndLoading() {
...
+// Release IndexSpec read/write
 void IndexSpec_ReleaseWriteLock(IndexSpec* sp) {
   pthread_rwlock_unlock(&sp->rwlock);
 }
 
+// Acquire IndexSpec read/write lock in shared mode
+void IndexSpec_AcquireReadLock(IndexSpec* sp) {
+  pthread_rwlock_rdlock(&sp->rwlock);
+}
+
```
raz-mon: I think we're missing the setting and unsetting of the context flags upon these operations, as we always (?) have the spec attached to some `RedisSearchCtx` that will now be inconsistent

src/spec.c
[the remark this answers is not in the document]
JoanFM: the problem is the naming of the UnlockWrite, which is not totally correct, but since this would imply code changes in the Rust code I did not want to rename now.

src/spec.c
JoanFM: I think we shouls maybe add docs here to prioritize the usage of the other functions when there is a Ctx, but when there is no Ctx I assume there is no need for that other than potential double locking and stuff. For the case when it is used in the GC thread, this has no issue, in SAVE i also think there is no issue but maybe I can create a RediSearchCtx and avoid this change. Let me go this path

src/spec.c
```diff
@@ -4201,12 +4212,12 @@ void Indexes_EndLoading() {
 // Compaction FFI Functions (called by Rust during GC)
 // =============================================================================
 
-// Acquire IndexSpec write lock
+// Acquire IndexSpec read/write lock in exclusive mode
```
raz-mon: I'm not a fan of the comment update, would prefer it matches the function

JoanFM: I think the comment now is more accurate, but I can remove

JoanFM:
```suggestion
// Acquire IndexSpec write lock
```

src/spec.c
```diff
@@ -4201,12 +4212,12 @@ void Indexes_EndLoading() {
...
 void IndexSpec_AcquireWriteLock(IndexSpec* sp) {
   pthread_rwlock_wrlock(&sp->rwlock);
 }
 
-// Release IndexSpec write lock
+// Release IndexSpec read/write
```
JoanFM:
```suggestion
// Release IndexSpec write lock
```

PR #8498, opened 2026-02-25
Rename header file

memark:
Rename _sorting_vector_rs.h_ to just _sorting_vector.h_ since the migration is done, and there is no longer a name clash.

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Pure header rename/include cleanup with no behavioral changes; risk is limited to potential build failures if any remaining references to the old header exist.
> 
> **Overview**
> Renames the Rust-generated sorting-vector FFI header from `sorting_vector_rs.h` to `sorting_vector.h` and updates C callers (`document.c`, `sortable.h`) plus the `cbindgen` build script to reference the new name.
> 
> Also removes the now-unneeded include of the old header from `sortable.c` and tidies include ordering to rely on `sortable.h` for the sorting-vector declarations.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0227a6bb82bca4ea4a98f2ed00e0614478dba6af. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/sortable.c
+++ b/src/sortable.c
@@ -13,7 +13,6 @@
 #include "rmutil/util.h"
 #include "rmutil/strings.h"
 #include "rmalloc.h"
-#include "sortable.h"
 #include "buffer.h"
 #include "sortable.h"
 


src/sortable.c
```diff
@@ -15,7 +15,7 @@
 #include "rmalloc.h"
 #include "sortable.h"
 #include "buffer.h"
-#include "sorting_vector_rs.h"
+#include "sortable.h"
```
daxhuiberts, line 17: I think you mis-renamed here.

memark: Actually no, I did that in a separate commit. I would say the convention is for every C file to include its own H file, rather than some other H file. But I could have mentioned it in the description. What I also missed though, is that it was already included higher up. Deduplicating and moving to the end.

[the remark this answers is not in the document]
memark: See above.

PR #8515, opened 2026-02-26
Add RDB load on disk into temp memory object

JoanFM:
Load the RDB into Temp Memory Object and remove the APIs to load the RDB directly into the IndexSpec for clarity

---

> [!NOTE]
> **Medium Risk**
> Changes the on-disk index RDB load path and disk API surface; mistakes could corrupt RDB parsing or restore incorrect disk metadata during replication/SST flows.
> 
> **Overview**
> Refactors disk-backed index RDB restore to **no longer load directly into an existing `diskSpec`**. Instead, RDB disk metadata is first loaded into a temporary `RedisSearchDiskRdbState`, then applied when opening the disk index via a new `SearchDisk_OpenIndexWithRdbState` API (with explicit cleanup via `SearchDisk_FreeRdbState`, including for duplicate index loads).
> 
> Updates the public disk API (`search_disk_api.h`/wrappers) accordingly by removing the old `indexSpecRdbLoad` entry point and adding `loadRdbToTempObject`/`openIndexSpecWithRdbState`/`freeRdbState`, and tweaks a couple of related assertions/logging around legacy RDB load and GC startup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 87f9bf95acb7b59153b4ecdc737115d1c7db2aec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/spec.c
+++ b/src/spec.c
@@ -3616,7 +3616,6 @@ void *IndexSpec_LegacyRdbLoad(RedisModuleIO *rdb, int encver) {
     return NULL;
   }
 
-  // Start GC after diskSpec is available so the disk GC callback can use it immediately.
   IndexSpec_StartGC(spec_ref, sp, GCPolicy_Fork);
   Cursors_initSpec(sp);
 


src/spec.c
```diff
@@ -3308,43 +3319,66 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
...
+  // Open the index on disk only if we are on Flex, and this is not a duplicate.
+  if (isSpecOnDisk(sp) && !sp->isDuplicate) {
+    RS_ASSERT(disk_db);
+    size_t len;
+    const char* name = HiddenString_GetUnsafe(sp->specName, &len);
+
+    if (diskRdbState) {
+      // Use the new API that applies the RDB state during index opening
+      sp->diskSpec = SearchDisk_OpenIndexWithRdbState(name, len, sp->rule->type, diskRdbState);
+      diskRdbState = NULL; // Ownership transferred
```
JoanFM, line 3438: review this ownership transfer

src/spec.c
```diff
@@ -3308,43 +3320,65 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, QueryError *status)
...
```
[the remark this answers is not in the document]
JoanFM: this is not real, in the caller ownership is transferred at the beginning with Box::from_raw

src/spec.c
JoanFM: reviewed done, is good

src/search_disk_api.h
```diff
@@ -82,6 +82,47 @@ typedef struct BasicDiskAPI {
...
+   * @brief Load disk-related RDB data into a temporary in-memory object.
+   *
+   * Called during RDB load when the IndexSpec cannot be created yet (e.g., during replication
+   * before SST files arrive). The returned state must later be passed to openIndexSpecWithRdbState
+   * or freed with freeRdbState.
+   *
+   * @param rdb The RedisModuleIO handle for RDB operations
+   * @return Pointer to temporary RDB state, or NULL on error
+   */
+  RedisSearchDiskRdbState *(*loadRdbToTempObject)(RedisModuleIO *rdb);
```
kei-nan, line 125: Maybe instead of returning a temp object we could pass the disk context pointer and store the temporary objects in a hash map from index name to the object.
Then on the regular open flow we could check the hash map and pop the temporary state.
@JoanFM WDYT?

JoanFM: We could, but I do not see a real benefit there. I prefer the RDB load process to explicitly see this, than keeping some state on the Disk side.

kei-nan: The problem I see is that it splits the OSS code to use different index open APIs.
one for rdb and another when opening an index without rdb.
Instead of keeping the API minimal we double it.

src/spec.c
```diff
@@ -3591,13 +3617,12 @@ void *IndexSpec_LegacyRdbLoad(RedisModuleIO *rdb, int encver) {
   }
 
   // Start GC after diskSpec is available so the disk GC callback can use it immediately.
-  IndexSpec_StartGC(spec_ref, sp, sp->diskSpec ? GCPolicy_Disk : GCPolicy_Fork);
```
kei-nan, line 3594: @Itzikvaknin maybe take a look, I know you modified this area

src/spec.c
```diff
@@ -3399,41 +3405,62 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, bool useSst, QueryE
     }
   }
 
+  // Load the disk-related index data if we are on disk and the save flow used
+  // sst-files. We load it into a temporary in-memory object first, then use it
+  // to open the index with the RDB state applied.
+  // We must always consume the RDB data to avoid corrupting the stream,
+  // even for duplicates. We just won't use it in the duplicate case.
+  if (isSpecOnDisk(sp) && encver >= INDEX_DISK_VERSION && useSst) {
```
kei-nan, line 3413: some stuff you load directly into the sp object, you aren't really using the temp object across the board.
It kind of breaks what you set out to do, no?

JoanFM: No, because what I load is not part of what will be in SST. The difference is how we load the diskSpec. But for now this is only a refactor to make sure the Disk side is aligned with this idea and is not more expensive to refactor later.

src/spec.c
```diff
@@ -3399,41 +3405,62 @@ IndexSpec *IndexSpec_RdbLoad(RedisModuleIO *rdb, int encver, bool useSst, QueryE
...
-  // since there's no SST data to restore from (stale data must be cleared).
   if (isSpecOnDisk(sp) && !sp->isDuplicate) {
     RS_ASSERT(disk_db);
     size_t len;
     const char* name = HiddenString_GetUnsafe(sp->specName, &len);
-    sp->diskSpec = SearchDisk_OpenIndex(ctx, name, len, sp->rule->type, !useSst);
+
+    if (diskRdbState) {
+      // Use the new API that applies the RDB state during index opening
+      sp->diskSpec = SearchDisk_OpenIndexWithRdbState(ctx, name, len, sp->rule->type, diskRdbState);
```
raz-mon, line 3437: Isn't this too early?
I thought we were calling this on the event of finishing loading.

JoanFM: But we still do not have this event. This is just a refactor to start preparing the APIs.

src/search_disk_api.h
JoanFM: But the fact is that the APIs are completely different between OSS and Disk. I prefer to be explicit here where the logic is implemented and read, than hiding it in details hidden to the actual implementation

src/spec.c
raz-mon: Yeah got that a bit late.
Would be good to have this in the description

src/spec.c
JoanFM: This is done, because for RDB Legacy we do not support disk, the version is too old, so we have an ASSERTION that SearchDisk is not enabled.

src/spec.c
raz-mon: BTW I think we do already have an event to work with.
Can you check this with core? I discussed this with Oran at the time and understood we already have the event that the loading finished.

PR #8531, opened 2026-02-27
[MOD-12717] Intersection iterator: add slop and order support

ThierryBerger:

- Adds support in Rust code for intersection iterator's max_slop and in_order.
- Adds tests and benchmarks focused on those new (rust) capabilities
- to be noted, I didn't see any benchmarks on cpp focused on those features, so I rolled my own.

## NOT in this PR:

- I used C implementation of `IndexResult_IsWithinRange`  to keep the stop the creeping scope of the PR.

## Benchmarks

| Benchmark | C | Rust | Speedup |
|-----------|---|------|---------|
| Read Slop=0 All Pass | 6.503 ms | 6.025 ms | 1.08x faster |
| Read Slop=100 All Pass | 6.479 ms | 6.000 ms | 1.08x faster |
| Read In-Order All Pass | 6.514 ms | 5.941 ms | 1.10x faster |
| Read In-Order Slop=100 All Pass | 6.539 ms | 5.990 ms | 1.09x faster |
| Read In-Order All Fail | 6.545 ms | 6.112 ms | 1.07x faster |

Not much visible change in perf (probably the C ffi cost actually.), because the logic is still in C within `IndexResult_IsWithinRange`.

It's still interesting to see that no regression occur (especially with the `check_needs_relevancy` opposed to function pointer), which we'll study in a follow up.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Changes core `Intersection` iterator semantics by filtering matches via proximity/order constraints and alters child sorting behavior when `in_order=true`, which can impact query result correctness and iteration/EOF edge cases. It also introduces new FFI calls into C for relevancy checks and new benchmark-only C index/iterator wrappers, increasing surface area for subtle memory/ownership bugs.
> 
> **Overview**
> Adds proximity-aware intersection in Rust by extending `Intersection` with `max_slop` and `in_order`, preserving child order when required and filtering agreed doc IDs via `ffi::IndexResult_IsWithinRange` during `read`/`skip_to` (including retry logic when a candidate doc fails relevancy).
> 
> Expands integration coverage with new slop/order tests (including `last_doc_id` stability and EOF retry edge cases) and adjusts mocks to emit positional term results. Updates the benchmarking harness to construct positional term iterators for both C and Rust and adds new slop/order benchmark cases, plus minor docs cleanup in the benchmark skill.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f356215133bd96c63850fa69837886d55b782879. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rqe_iterators/src/intersection.rs
+++ b/src/redisearch_rs/rqe_iterators/src/intersection.rs
@@ -9,13 +9,9 @@
 
 //! Supporting types for [`Intersection`].
 //!
-//! # TODO (MOD-12717)
-//!
-//! The following features from the C++ implementation are not yet ported:
-//! - `max_slop`: Maximum allowed slop between term positions
+//! The intersection iterator supports proximity constraints via two parameters:
+//! - `max_slop`: Maximum allowed slop between term positions (`None` = no constraint)
 //! - `in_order`: Require terms to appear in order
-//!
-//! Currently, the iterator operates with `max_slop = -1` semantics (no slop validation).
 
 use ffi::t_docId;
 use inverted_index::RSIndexResult;
@@ -24,15 +20,28 @@ use crate::{RQEIterator, RQEIteratorError, RQEValidateStatus, SkipToOutcome};
 
 /// Yields documents appearing in ALL child iterators using a merge (AND) algorithm.
 ///
-/// Children are sorted by estimated result count (smallest first) to minimize iterations.
-/// A document is only yielded when ALL children have a matching entry for it.
+/// Children are sorted by estimated result count (smallest first) to minimize iterations,
+/// unless `in_order` is set (which preserves the original child order for positional checks).
+/// A document is only yielded when ALL children have a matching entry for it and the
+/// term positions satisfy the `max_slop` / `in_order` proximity constraints:
+///
+/// - `max_slop`: Maximum allowed slop (distance) between term positions. `None` disables proximity
+///   validation entirely.
+/// - `in_order`: When `true`, terms must appear in the same order as the child iterators.
 pub struct Intersection<'index, I> {
-    /// Child iterators, sorted by estimated count (smallest first).
+    /// Child iterators, sorted by estimated count (smallest first) unless `in_order` is set.
     children: Vec<I>,
     /// Last doc_id successfully found in ALL children (returned by [`last_doc_id()`](Self::last_doc_id)).
     last_doc_id: t_docId,
     num_expected: usize,
     is_eof: bool,
+    /// Maximum allowed slop (distance) between term positions. `None` disables proximity
+    /// validation entirely.
+    ///
+    /// Capped to [`i32::MAX`] because of [`ffi::IndexResult_IsWithinRange`]
+    max_slop: Option<u32>,
+    /// When `true`, terms must appear in the same order as the child iterators.
+    in_order: bool,
     /// Aggregate result combining children's results, reused to avoid allocations.
     result: RSIndexResult<'index>,
 }
@@ -51,18 +60,45 @@ impl<'index, I> Intersection<'index, I>
 where
     I: RQEIterator<'index>,
 {
-    /// Creates a new intersection iterator. Children are sorted by estimated count. If `children`
-    /// is empty, returns an iterator immediately at EOF.
+    /// Creates a new intersection iterator without proximity constraints.
+    ///
+    /// Every document matching all children is yielded. Children are sorted by estimated result
+    /// count (smallest first) to minimize iterations.
+    ///
+    /// If `children` is empty, returns an iterator immediately at EOF.
     #[must_use]
-    pub fn new(mut children: Vec<I>) -> Self {
-        children.sort_by_cached_key(|c| c.num_estimated());
+    pub fn new(children: Vec<I>) -> Self {
+        Self::new_with_slop_order(children, None, false)
+    }
 
-        let Some(num_expected) = children.first().map(|c| c.num_estimated()) else {
+    /// Creates a new intersection iterator with proximity constraints.
+    ///
+    /// - `max_slop`: Maximum allowed distance between term positions. `None` disables proximity
+    ///   validation (every document matching all children is yielded).
+    /// - `in_order`: When `true`, terms must appear in the order of the child iterators and
+    ///   children are **not** re-sorted by estimated count (their order is meaningful).
+    ///
+    /// If `children` is empty, returns an iterator immediately at EOF.
+    #[must_use]
+    pub fn new_with_slop_order(
+        mut children: Vec<I>,
+        max_slop: Option<u32>,
+        in_order: bool,
+    ) -> Self {
+        // Only sort by estimated count when order doesn't matter for proximity checks.
+        if !in_order {
+            children.sort_by_cached_key(|c| c.num_estimated());
+        }
+        // FIXME: Capped because of `ffi::IndexResult_IsWithinRange`
+        let max_slop = max_slop.map(|v| v.min(i32::MAX as u32));
+        let Some(num_expected) = children.iter().map(|c| c.num_estimated()).min() else {
             return Self {
                 children,
                 last_doc_id: 0,
                 num_expected: 0,
                 is_eof: true,
+                max_slop,
+                in_order,
                 result: RSIndexResult::build_intersect(0).build(),
             };
         };
@@ -72,7 +108,62 @@ where
             last_doc_id: 0,
             num_expected,
             is_eof: false,
+            max_slop,
+            in_order,
             result: RSIndexResult::build_intersect(num_children).build(),
+        }
+    }
+
+    /// Returns `true` if the current result needs a proximity check after consensus.
+    ///
+    /// The check is skipped only when `max_slop` is `None` and `in_order` is `false` — the
+    /// only case where every document matching all children is trivially within range.
+    const fn needs_relevancy_check(&self) -> bool {
+        self.max_slop.is_some() || self.in_order
+    }
+
+    /// Check if the current aggregate result satisfies the proximity constraints.
+    fn current_is_relevant(&self) -> bool {
+        // SAFETY:
+        // - `self.result` is a valid, fully initialised `RSIndexResult`.
+        // - The C function reads from `r` without taking ownership or storing the pointer.
+        // - `from_ref` gives a `*const` pointer; `cast_mut` is required by the C API, which
+        //   uses a non-const pointer even though it only reads; no mutation occurs.
+        unsafe {
+            ffi::IndexResult_IsWithinRange(
+                std::ptr::from_ref(&self.result).cast_mut().cast(),
+                // `v as i32` is lossless: the constructor caps `max_slop` to
+                // `i32::MAX as u32`, so `v` always fits in `i32`.
+                self.max_slop.map_or(i32::MAX, |v| v as i32),
+                self.in_order as std::ffi::c_int,
+            ) != 0
+        }
+    }
+
+    /// Find consensus on a doc_id and verify that the result satisfies the proximity constraints.
+    ///
+    /// If the agreed-upon document doesn't satisfy the slop/order constraints, advances the first
+    /// child and retries until a relevant result is found or EOF is reached.
+    fn find_consensus_with_relevancy_check(
+        &mut self,
+        mut target: t_docId,
+    ) -> Result<Option<t_docId>, RQEIteratorError> {
+        loop {
+            match self.find_consensus(target)? {
+                Some(doc_id) => {
+                    self.build_aggregate_result(doc_id);
+                    if self.current_is_relevant() {
+                        self.last_doc_id = doc_id;
+                        return Ok(Some(doc_id));
+                    }
+                    // Not relevant — advance past this document and retry.
+                    let Some(next) = self.read_from_first_child()? else {
+                        return Ok(None);
+                    };
+                    target = next;
+                }
+                None => return Ok(None),
+            }
         }
     }
 
@@ -146,8 +237,6 @@ where
     /// - Restructure [`RSAggregateResult`](inverted_index::RSAggregateResult) to not require `'index` on stored references
     /// - Use a different aggregate pattern that doesn't store child references
     fn build_aggregate_result(&mut self, doc_id: t_docId) {
-        self.last_doc_id = doc_id;
-
         if let Some(agg) = self.result.as_aggregate_mut() {
             agg.reset();
         }
@@ -188,13 +277,21 @@ where
         let Some(target) = self.read_from_first_child()? else {
             return Ok(None);
         };
-
-        match self.find_consensus(target)? {
-            Some(doc_id) => {
-                self.build_aggregate_result(doc_id);
-                Ok(Some(&mut self.result))
+        // FIXME: consider using function pointers to remove runtime checks for each reads.
+        if self.needs_relevancy_check() {
+            match self.find_consensus_with_relevancy_check(target)? {
+                Some(_) => Ok(Some(&mut self.result)),
+                None => Ok(None),
             }
-            None => Ok(None),
+        } else {
+            match self.find_consensus(target)? {
+                Some(doc_id) => {
+                    self.build_aggregate_result(doc_id);
+                    self.last_doc_id = doc_id;
+                    Ok(Some(&mut self.result))
+                }
+                None => Ok(None),
+            }
         }
     }
 
@@ -205,17 +302,50 @@ where
         if self.is_eof {
             return Ok(None);
         }
-
-        match self.find_consensus(doc_id)? {
-            Some(found_id) => {
-                self.build_aggregate_result(found_id);
-                if found_id == doc_id {
-                    Ok(Some(SkipToOutcome::Found(&mut self.result)))
-                } else {
-                    Ok(Some(SkipToOutcome::NotFound(&mut self.result)))
+        // FIXME: consider using function pointers to remove runtime checks for each reads.
+        if self.needs_relevancy_check() {
+            // Try to agree on the requested doc_id first.
+            match self.find_consensus(doc_id)? {
+                Some(found_id) => {
+                    self.build_aggregate_result(found_id);
+                    let self_current_is_relevant = self.current_is_relevant();
+                    if found_id == doc_id && self_current_is_relevant {
+                        self.last_doc_id = found_id;
+                        return Ok(Some(SkipToOutcome::Found(&mut self.result)));
+                    }
+                    // Either we landed on a different doc, or the exact match wasn't relevant.
+                    // In both cases, find the next relevant result.
+                    let next_target = if self_current_is_relevant {
+                        // Consensus on a different doc_id that IS relevant — return it.
+                        self.last_doc_id = found_id;
+                        return Ok(Some(SkipToOutcome::NotFound(&mut self.result)));
+                    } else {
+                        // Not relevant — advance past this document.
+                        match self.read_from_first_child()? {
+                            Some(t) => t,
+                            None => return Ok(None),
+                        }
+                    };
+                    match self.find_consensus_with_relevancy_check(next_target)? {
+                        Some(_) => Ok(Some(SkipToOutcome::NotFound(&mut self.result))),
+                        None => Ok(None),
+                    }
                 }
+                None => Ok(None),
             }
-            None => Ok(None),
+        } else {
+            match self.find_consensus(doc_id)? {
+                Some(found_id) => {
+                    self.build_aggregate_result(found_id);
+                    self.last_doc_id = found_id;
+                    if found_id == doc_id {
+                        Ok(Some(SkipToOutcome::Found(&mut self.result)))
+                    } else {
+                        Ok(Some(SkipToOutcome::NotFound(&mut self.result)))
+                    }
+                }
+                None => Ok(None),
+            }
         }
     }
 
--- a/src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs
+++ b/src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs
@@ -8,12 +8,6 @@
 */
 
 //! Integration tests for the Intersection iterator.
-//!
-//! C-Code: These tests are ported from the C++ tests in
-//! `tests/cpptests/test_cpp_iterator_intersection.cpp`.
-//!
-//! C-Code: Tests for `max_slop` and `in_order` are not included in this first
-//! version since those features are not yet implemented in the Rust port.
 
 use ffi::t_docId;
 use rqe_iterators::{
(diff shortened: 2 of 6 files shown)


src/redisearch_rs/inverted_index/src/index_result/core.rs
```diff
@@ -548,6 +548,150 @@ impl<'index> RSIndexResult<'index> {
...
+    /// Check whether the term positions in this aggregate result satisfy the given slop and
+    /// ordering constraints.
+    ///
+    /// Returns `true` if any of these hold:
+    /// - This is not an aggregate result (single term — always within range).
+    /// - The aggregate has 1 or fewer children.
+    /// - No children have offsets to check.
+    /// - The term positions satisfy the proximity constraint.
+    ///
+    /// Ported from C `IndexResult_IsWithinRange` in `src/index_result/index_result.c`.
```
ThierryBerger: add  a `TODO: remove when fully ported to rust` ? (similarly on future similar comments)

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -24,15 +20,23 @@ use crate::{RQEIterator, RQEIteratorError, RQEValidateStatus, SkipToOutcome};
...
-    /// Child iterators, sorted by estimated count (smallest first).
+    /// Child iterators, sorted by estimated count (smallest first) unless `in_order` is set.
     children: Vec<I>,
     /// Last doc_id successfully found in ALL children (returned by `last_doc_id()`).
     last_doc_id: t_docId,
     num_expected: usize,
     is_eof: bool,
+    /// Maximum allowed slop (distance) between term positions. A negative value disables
+    /// proximity validation entirely (equivalent to `INT_MAX` in the C implementation).
+    max_slop: i32,
```
ThierryBerger: This negative value semantic is tedious to maintain, I'd prefer to have an enum here, but I wanted to keep implementation as close as possible to C.

What do you think ? Should we port to an enum now?

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -51,18 +55,28 @@ impl<'index, I> Intersection<'index, I>
 where
     I: RQEIterator<'index>,
 {
-    /// Creates a new intersection iterator. Children are sorted by estimated count. If `children`
-    /// is empty, returns an iterator immediately at EOF.
+    /// Creates a new intersection iterator with proximity constraints.
+    ///
+    /// - `max_slop`: Maximum allowed distance between term positions. A negative value disables
```
ThierryBerger: I'm not a huge fan of these duplicated comments, LLM are now very good at catching those and updating them correctly, but should we just link to `Intersection` type ?

src/redisearch_rs/rqe_iterators_bencher/src/benchers/intersection.rs
```diff
@@ -90,6 +90,7 @@ impl Bencher {
         self.read_varying_sizes(c);
         self.skip_to_high_overlap(c);
         self.skip_to_low_overlap(c);
+        slop_and_order::Bencher::default().bench(c);
```
ThierryBerger: This doesn't give too much control over what we run, but I figured that would separate intents at least.

src/redisearch_rs/inverted_index/src/index_result/core.rs
```diff
@@ -548,6 +548,154 @@ impl<'index> RSIndexResult<'index> {
...
+            return true;
+        }
+
+        if in_order {
+            within_range_in_order(&mut iters, max_slop)
+        } else {
+            within_range_unordered(&mut iters, max_slop)
+        }
+    }
+}
```
ThierryBerger: This function is actually a bit involved to test, when trying to swap from C to Rust, a logic bug was revealed by integration tests, I'll simplify this current PR by leveraging C implementation, and will continue my effort to port this logic in a follow up.

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -9,30 +9,61 @@
...
 //! - `in_order`: Require terms to appear in order
-//!
-//! Currently, the iterator operates with `max_slop = -1` semantics (no slop validation).
 
 use ffi::t_docId;
 use inverted_index::RSIndexResult;
 
 use crate::{RQEIterator, RQEIteratorError, RQEValidateStatus, SkipToOutcome};
 
+unsafe extern "C" {
```
gdesmott: Can't  use it from the `ffi` crate? Generate it if required, you just need to update `src/redisearch_rs/ffi/build.rs` and add the missing header.

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -9,30 +9,61 @@
...
     /// Last doc_id successfully found in ALL children (returned by [`last_doc_id()`](Self::last_doc_id)).
     last_doc_id: t_docId,
     num_expected: usize,
     is_eof: bool,
+    /// Maximum allowed slop (distance) between term positions. `None` disables proximity
+    /// validation entirely (passed to C as `i32::MAX`).
+    max_slop: Option<i32>,
+    /// When `true`, terms must appear in the same order as the child iterators.
+    in_order: bool,
+
```
gdesmott:
```suggestion
```

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -9,30 +9,61 @@
...
 pub struct Intersection<'index, I> {
-    /// Child iterators, sorted by estimated count (smallest first).
+    /// Child iterators, sorted by estimated count (smallest first) unless `in_order` is set.
     children: Vec<I>,
     /// Last doc_id successfully found in ALL children (returned by [`last_doc_id()`](Self::last_doc_id)).
     last_doc_id: t_docId,
     num_expected: usize,
     is_eof: bool,
+    /// Maximum allowed slop (distance) between term positions. `None` disables proximity
+    /// validation entirely (passed to C as `i32::MAX`).
```
gdesmott: No need to mention the C details here as C code will eventually be gone.

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -9,30 +9,61 @@
...
+}
+
 /// Yields documents appearing in ALL child iterators using a merge (AND) algorithm.
 ///
-/// Children are sorted by estimated result count (smallest first) to minimize iterations.
-/// A document is only yielded when ALL children have a matching entry for it.
+/// Children are sorted by estimated result count (smallest first) to minimize iterations,
+/// unless `in_order` is set (which preserves the original child order for positional checks).
+/// A document is only yielded when ALL children have a matching entry for it and the
+/// term positions satisfy the `max_slop` / `in_order` proximity constraints.
```
gdesmott: From the doc only, it's still not totally clear how `max_slop` work exactly. Can we extend it a bit?

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -189,12 +281,20 @@ where
             return Ok(None);
         };
 
-        match self.find_consensus(target)? {
-            Some(doc_id) => {
-                self.build_aggregate_result(doc_id);
-                Ok(Some(&mut self.result))
+        if self.needs_relevancy_check() {
```
gdesmott, line 281: I was wondering if we should use function pointers like in the C version and in `InvIndIterator` to not evaluate this during each read operation. But I suppose it's actuallu ok as `needs_relevancy_check` is `const`, the compiler should optimize it itself.

Probably worth a comment though.

src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs
```diff
@@ -12,8 +12,7 @@
 //! C-Code: These tests are ported from the C++ tests in
 //! `tests/cpptests/test_cpp_iterator_intersection.cpp`.
 //!
-//! C-Code: Tests for `max_slop` and `in_order` are not included in this first
-//! version since those features are not yet implemented in the Rust port.
+//! C-Code: Tests for `max_slop` and `in_order` are in a separate test module.
```
gdesmott: Are they?

src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs
```diff
@@ -99,7 +98,7 @@ fn read_test_case(num_children: usize, result_set: &[t_docId]) {
         .min()
         .unwrap_or(0);
 
-    let mut ii = Intersection::new(children);
+    let mut ii = Intersection::new(children, None, false);
```
gdesmott: Maybe we should keep `new` as it and have a `with_slop_order()` alternative constructor?

src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs
```diff
@@ -1163,3 +1183,274 @@ fn revalidate_moved_skip_to_returns_none() {
...
+// =============================================================================
+
+/// Tests for the intersection iterator's `max_slop` and `in_order` proximity constraints.
+///
+/// C-Code: These tests are ported from `IntersectionIteratorTest` in
+/// `tests/cpptests/test_cpp_iterator_intersection.cpp`.
+///
+/// Document setup mirrors the C++ test.
+//
+#[cfg(not(miri))]
```
gdesmott, line 1192:
```suggestion
// Because of `ffi::IndexResult_IsWithinRange`
#[cfg(not(miri))]
```

src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs
```diff
@@ -1163,3 +1183,274 @@ fn revalidate_moved_skip_to_returns_none() {
...
+
+    fn new_term() -> Box<RSQueryTerm> {
+        RSQueryTerm::new("term", 1, 0)
+    }
+
+    /// Build an `InvertedIndex<Full>` where each entry has a single term position.
+    ///
+    /// `docs` is a slice of `(doc_id, position)` pairs. Positions are encoded as
+    /// single-byte delta-varints (valid for values 1–127).
+    fn make_term_index(docs: &[(t_docId, u8)]) -> InvertedIndex<Full> {
```
gdesmott: Can't we use `mock_iterator.rs` to test all this?

src/redisearch_rs/inverted_index/src/index_result/core.rs
gdesmott: outdated?

src/redisearch_rs/rqe_iterators/src/intersection.rs
gdesmott: I think an `Option` is ok. Should it be `Option<u32>` then?

src/redisearch_rs/rqe_iterators/src/intersection.rs
gdesmott: I'll keep it. Constructor is likely the first API users will use, and you said LLM can easily update them.

src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs
ThierryBerger: They are, but in the same file, I agree this is misleading, I'll remove the comment.

I still think it's interesting to separate tests within a different module (I'm ok with keeping it in the same file), but I have no strong feeling regarding this organization.

src/redisearch_rs/rqe_iterators/src/intersection.rs
ThierryBerger: I was wondering that too, I suggest to explore that when we look at benchmarks. I'll add a comment.

keeping discussion opened for you to resolve if satistied.

src/redisearch_rs/rqe_iterators/src/intersection.rs
ThierryBerger: mh as `max_slop` is private, I guess it's fair to extend explaination here indeed 👍

keeping discussion opened for you to resolve if satistied.

src/redisearch_rs/rqe_iterators_bencher/src/benchers/intersection.rs
gdesmott: Run filtering is done by passing the names of the benches, like `Iterator - Intersection - Read In-Order All Fail`

src/redisearch_rs/rqe_iterators_bencher/src/benchers/intersection.rs
```diff
@@ -188,3 +189,229 @@ impl Bencher {
...
+
+/// Benchmarks for the `max_slop` and `in_order` proximity-check path.
+///
+/// Uses real term inverted indexes with encoded position data so that `is_within_range`
+/// performs actual work. Two position patterns cover the performance extremes:
+///
+/// - **adjacent_in_order** (foo@1, bar@2): all docs pass `max_slop=0` and `in_order=true`.
+/// - **adjacent_reverse** (foo@2, bar@1): all docs fail `max_slop=0 + in_order=true`,
+///   forcing the iterator to scan the full corpus without yielding any results.
+pub mod slop_and_order {
```
gdesmott: Do we really need another module here?

Why not just have all those in the existing `bench()` method? It's all the same iterators, just with different settings.

src/redisearch_rs/rqe_iterators_bencher/src/benchers/intersection.rs
```diff
@@ -188,3 +189,229 @@ impl Bencher {
...
+            c: &'a mut Criterion,
+            label: &str,
+        ) -> BenchmarkGroup<'a, WallTime> {
+            let mut group = c.benchmark_group(label);
+            group.measurement_time(Self::MEASUREMENT_TIME);
+            group.warm_up_time(Self::WARMUP_TIME);
+            group
+        }
+
+        pub fn bench(&self, c: &mut Criterion) {
```
gdesmott: You can use the `/run-rust-benchmarks` skill and paste the table with results in the PR top comment.

src/redisearch_rs/rqe_iterators_bencher/src/benchers/intersection.rs
ThierryBerger: It clarifies intent by grouping them, but I have no strong feeling about it, I'll make it more flat.

src/redisearch_rs/inverted_index/src/index_result/core.rs
ThierryBerger: Yes, as per description in the PR, I chose to rely on C implementation of that function, as it's reimplementation in Rust was more complicated that I thought, so the plan is:
- PR1 (this one): adding slop/in_order functionality to Rust intersection iterator
- PR2 (https://github.com/RediSearch/RediSearch/pull/8619): migrating c impl to Rust
- Follow up work: less reliance on C implementation (this may or may not need/conflict with union iterator work)

src/redisearch_rs/rqe_iterators/src/intersection.rs
ThierryBerger: Since my comment, I used `Option<i32>` ; I kept using a i32 for maximum compatibility with C, but I think ultimately a `Option<u32>` makes more sense, yes, I'd still prefer to study impacts before making this potentially impactful change.

gdesmott: I'd go with `Option<u32>` and maybe a debug assert checking the value is lower than `i32::MAX` if we are capped with `i32` for the ffi call.

src/redisearch_rs/rqe_iterators_bencher/src/benchers/intersection.rs
ThierryBerger: done

src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs
ThierryBerger: Not trivially, because slop and in_order need positional data, which is not currently supported in mock_iteraror.rs.

it's probably cleaner to add this feature to the mock_iterator though :+1:

src/redisearch_rs/rqe_iterators/src/intersection.rs
ThierryBerger: deal ; I capped within the constructor, with a note explaining why. the debug assert is a bit much I think.

src/redisearch_rs/rqe_iterators/tests/integration/intersection.rs
ThierryBerger: I was able to use mock_iterator building on top of what github.com/RediSearch/RediSearch/pull/8637 provides :tada:

src/redisearch_rs/rqe_iterators_bencher/src/benchers/intersection.rs
```diff
@@ -69,6 +85,52 @@ fn ids_to_rust_children(ids: Vec<Vec<u64>>) -> Vec<IdListSorted<'static>> {
...
 }
 
+fn new_query_term() -> Box<RSQueryTerm> {
+    RSQueryTerm::new("term", 1, 0)
+}
+
+/// Build two Rust `InvertedIndex<Full>` instances.
+///
+/// `first_pos` and `second_pos` are the (constant) positions used for every document.
+fn make_rust_indexes(first_pos: u8, second_pos: u8) -> (InvertedIndex<Full>, InvertedIndex<Full>) {
```
gdesmott, line 104: Can't we keep using `IdListSorted` as we the existing benchmarks do?

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -51,18 +60,45 @@ impl<'index, I> Intersection<'index, I>
...
+    /// If `children` is empty, returns an iterator immediately at EOF.
+    #[must_use]
+    pub fn new_with_slop_order(
+        mut children: Vec<I>,
+        max_slop: Option<u32>,
+        in_order: bool,
+    ) -> Self {
+        // Only sort by estimated count when order doesn't matter for proximity checks.
+        if !in_order {
+            children.sort_by_cached_key(|c| c.num_estimated());
```
gdesmott, line 90: We still need the iterator factor, right?

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -24,15 +20,28 @@ use crate::{RQEIterator, RQEIteratorError, RQEValidateStatus, SkipToOutcome};
...
+    /// Child iterators, sorted by estimated count (smallest first) unless `in_order` is set.
     children: Vec<I>,
     /// Last doc_id successfully found in ALL children (returned by [`last_doc_id()`](Self::last_doc_id)).
     last_doc_id: t_docId,
     num_expected: usize,
     is_eof: bool,
+    /// Maximum allowed slop (distance) between term positions. `None` disables proximity
+    /// validation entirely.
+    ///
+    /// Capped to [`i32::Max`] because of [`ffi::IndexResult_IsWithinRange`]
```
gdesmott:
```suggestion
    /// Capped to [`i32::MAX`] because of [`ffi::IndexResult_IsWithinRange`]
```

src/redisearch_rs/rqe_iterators/src/intersection.rs
ThierryBerger: Correct!
- I consider fixing the logic part of https://github.com/RediSearch/RediSearch/pull/8619 ; as this PR was supposed to be merely for slop and order support, so we're borderline stretching the scope here.
- It's also current state of the repository, and still not wired to any exposed API, that's why I'm not actively trying to fix it **in this very PR**.

I agree it's an important part of the migration that shouldn't be forgotten :100:

src/redisearch_rs/rqe_iterators/src/intersection.rs
ThierryBerger: Good catch!

- I opened https://github.com/RediSearch/RediSearch/pull/8656 to avoid missing more from now on.

src/redisearch_rs/rqe_iterators_bencher/src/benchers/intersection.rs
ThierryBerger: Not trivially, it's the same issue that we had for mock iterator (addressed in https://github.com/RediSearch/RediSearch/pull/8637) : we need positional data on `IdList`, or a different flavor of it.

I guess we should avoid impacting `IdList` because it would risk impacting benchmarks.

Could we consider this a follow up ? This will be particularly interesting when another benchmark needs that feature.

gdesmott: It's ok then. Just add a comment explaining why we use II here.

src/redisearch_rs/rqe_iterators_bencher/src/benchers/intersection.rs
```diff
@@ -34,6 +43,13 @@ const WEIGHT: f64 = 1.0;
 /// Step size for skip_to benchmarks.
 const STEP: u64 = 100;
 
+/// Number of documents for slop/order benchmarks.
+const NUM_DOCS: u64 = 100_000;
+const FLAGS: IndexFlags = IndexFlags_Index_StoreFreqs
+    | IndexFlags_Index_StoreTermOffsets
+    | IndexFlags_Index_StoreFieldFlags
+    | IndexFlags_Index_StoreByteOffsets;
+
```
dor-forer, line 58: Can you maybe comment shortly on why we need these specific flags?

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -72,10 +108,64 @@ where
...
+        // - Casting `*const` to `*mut` is required by the C API, which uses a non-const
+        //   pointer even though it only reads; no mutation occurs.
+        unsafe {
+            ffi::IndexResult_IsWithinRange(
+                &self.result as *const RSIndexResult<'_> as *mut ffi::RSIndexResult,
+                self.max_slop
+                    .map_or(i32::MAX, |v| v.try_into().unwrap_or(i32::MAX)),
+                self.in_order as std::ffi::c_int,
+            ) != 0
+        }
```
dor-forer, line 140: Wouldn't it be simpler to use `std::ptr::from_ref`? Or we don't support it?
In addition, I see that at construction time, max_slop is already capped to `i32::MAX as u32` , so maybe we don't need the `unwrap_or`?

```suggestion
unsafe {
    ffi::IndexResult_IsWithinRange(
        std::ptr::from_ref(&self.result).cast_mut().cast(),
        self.max_slop.map_or(i32::MAX, |v| v as i32),
        self.in_order as std::ffi::c_int,
    ) != 0
}
```

ThierryBerger: good catches!

src/redisearch_rs/rqe_iterators_bencher/src/benchers/intersection.rs
ThierryBerger: In short we need them to store positional data ; they are actually similar to INDEX_DEFAULT_FLAGS ; it's a bit awkward to pass those flags around as it's for C benchmarks, they will be removed once we don't need to compare Rust vs C.

I added a comment :+1:

src/redisearch_rs/rqe_iterators/src/intersection.rs
```diff
@@ -72,10 +108,65 @@ where
...
+        // - `self.result` is a valid, fully initialised `RSIndexResult`.
+        // - The C function reads from `r` without taking ownership or storing the pointer.
+        // - `from_ref` gives a `*const` pointer; `cast_mut` is required by the C API, which
+        //   uses a non-const pointer even though it only reads; no mutation occurs.
+        unsafe {
+            ffi::IndexResult_IsWithinRange(
+                std::ptr::from_ref(&self.result).cast_mut().cast(),
+                // `v as i32` is lossless: the constructor caps `max_slop` to
+                // `i32::MAX as u32`, so `v` always fits in `i32`.
+                self.max_slop.map_or(i32::MAX, |v| v as i32),
```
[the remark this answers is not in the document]
ThierryBerger: This report is incorrect, C constructor is setting max_slop to `i32::MAX` (Rust does the same), and the `IndexResult_IsWithinRange` is not called when `needs_relevancy_check()` is false (Rust does the same check).

So we're keeping the same behaviour here. passing -1 here would make the rest of the algorithm "fail".

It's indeed not obvious to track down, and I'm hopeful using an Option in Rust will help clarify those behaviors.

Keeping the discussion open for transparency.

PR #8534, opened 2026-02-28
MOD-13110: Implement foreground depletion for hybrid

nafraf:
PR #8274  was split in two parts. This is the part 2.

**2: Add foreground depleters for (WORKERS 0)**

  - Problem: When running `FT.HYBRID` without background workers (WORKERS 0) during concurrent write operations, a crash could occur. The cursor was returned to the client before iterators were initialized. If the index was modified between cursor creation and the first `_FT.CURSOR READ`, the uninitialized iterator state caused crashes.
  
  <details>
  <summary>Failure sequence</summary>

1. Coordinator receives `FT.HYBRID` command
2. Coordinator dispatches `_FT.HYBRID ... WITHCURSOR` to each shard
3. Each shard's `HybridRequest_StartCursors` creates cursors without reading any data
4. Cursors are paused, spec lock is released
5. Index is destroyed (slot migration frees the index)
6. Coordinator calls `_FT.CURSOR READ` on shard cursors
7. `handleSpecLockAndRevalidate` tries to revalidate, but ii is now a dangling pointer
8. `needs_revalidation()` reads garbage that matches stored `gc_marker` → returns false
9. CRASH in `skip_to` when accessing invalid buffer pointers
   
   </details>

  - Solution: `Make WORKERS == 0` behave like `WORKERS > 0`, but synchronously on the main thread.

1. MOD-...
2. #...

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Changes hybrid cursor execution to synchronously deplete results on the main thread when `WORKERS=0`, affecting cursor lifecycle, error handling, and profiling; mistakes could regress cursor correctness or performance under concurrency.
> 
> **Overview**
> Prevents `FT.HYBRID ... WITHCURSOR` crashes in `WORKERS=0` mode by introducing **foreground depletion**: shard subqueries now end with `RP_DEPLETER` and `HybridRequest_StartCursors` forces synchronous `RPDepleter_DepleteAll()` before returning cursor IDs.
> 
> Tightens cursor setup validation/error paths (expected depleter type checks; unified cleanup), and updates profiling to report depleter time via `RPProfile_GetTime()` for both `RP_SAFE_DEPLETER` and `RP_DEPLETER`.
> 
> Expands test coverage by re-enabling previously skipped hybrid ASM/migration tests, adding a new regression test for hybrid cursors across add-shard migration, and refactoring hybrid profile expectations to include the new `Depleter` RP; also replaces hardcoded `2**14` with a `CLUSTER_SLOTS` constant in ASM tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 80ae009aac5dbaa276851bc270986c4e0fdaab65. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/hybrid/hybrid_exec.c
+++ b/src/hybrid/hybrid_exec.c
@@ -674,37 +674,29 @@ int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, Q
       return REDISMODULE_ERR;
     }
 
+    int rc;
     if (backgroundDepletion) {
-      int rc = RPSafeDepleter_DepleteAll(depleters, status);
-      array_free(depleters);
-      if (rc != RS_RESULT_OK) {
-        array_free_ex(req->cursors, Cursor_Free(*(Cursor**)ptr));
-        req->cursors = NULL;
-        HybridRequest_UnlockCursors(req);
-        if (rc == RS_RESULT_ERROR) {
-          // Error was already set by RPSafeDepleter_DepleteAll
-          RS_ASSERT(QueryError_HasError(status));
-        } else {
-          QueryError_SetWithoutUserDataFmt(status, QUERY_ERROR_CODE_GENERIC, "Failed to deplete set of results, rc=%d", rc);
-        }
-        return REDISMODULE_ERR;
-      }
+      rc = RPSafeDepleter_DepleteAll(depleters, status);
     } else {
       // Foreground depletion for WORKERS == 0
       // Trigger synchronous depletion to read and buffer all results while the spec lock is held.
-      int rc = RPDepleter_DepleteAll(depleters);
-      array_free(depleters);
-      if (rc != RS_RESULT_OK) {
-        array_free_ex(req->cursors, Cursor_Free(*(Cursor**)ptr));
-        req->cursors = NULL;
-        HybridRequest_UnlockCursors(req);
+      rc = RPDepleter_DepleteAll(depleters);
+    }
+
+    array_free(depleters);
+
+    if (rc != RS_RESULT_OK) {
+      array_free_ex(req->cursors, Cursor_Free(*(Cursor**)ptr));
+      req->cursors = NULL;
+      HybridRequest_UnlockCursors(req);
+      if (!QueryError_HasError(status)) {
         if (rc == RS_RESULT_TIMEDOUT) {
           QueryError_SetWithoutUserDataFmt(status, QUERY_ERROR_CODE_TIMED_OUT, "Depleting timed out");
         } else {
           QueryError_SetWithoutUserDataFmt(status, QUERY_ERROR_CODE_GENERIC, "Failed to deplete set of results, rc=%d", rc);
         }
-        return REDISMODULE_ERR;
       }
+      return REDISMODULE_ERR;
     }
 
     HybridRequest_UnlockCursors(req);
--- a/tests/pytests/test_asm.py
+++ b/tests/pytests/test_asm.py
@@ -1,6 +1,7 @@
 from common import *
 from dataclasses import dataclass
 from redis import Redis
+import redis
 import random
 import re
 import numpy as np
@@ -10,6 +11,10 @@ import os
 import signal
 import psutil
 import subprocess
+
+# Total number of hash slots in a Redis Cluster (CRC16(key) % 16384).
+# This is a fixed, protocol-level constant defined by the Redis Cluster specification.
+CLUSTER_SLOTS = 2**14
 
 # Random words for generating more diverse text content
 RANDOM_WORDS = [
@@ -177,7 +182,7 @@ class SlotRange:
     @staticmethod
     def from_str(s: str):
         start, end = map(int, s.split("-"))
-        assert 0 <= start <= end < 2**14
+        assert 0 <= start <= end < CLUSTER_SLOTS
         return SlotRange(start, end)
 
 @dataclass
@@ -375,7 +380,7 @@ def create_and_populate_index(env: Env, index_name: str, n_docs: int):
             text_content = f"document {i} content {' '.join(random_words)} data"
             tag_value = "even" if i % 2 == 0 else "odd"
             # force each document to a different slot
-            con.execute_command('HSET', f'doc-{i}:{{{i % 2**14}}}',
+            con.execute_command('HSET', f'doc-{i}:{{{i % CLUSTER_SLOTS}}}',
                               'n', i,
                               'text', text_content,
                               'tag', tag_value,
@@ -437,7 +442,7 @@ def wait_for_migration_complete(env, dest_shard, source_shard, timeout=200, quer
 cluster_node_timeout = 60_000 # in milliseconds (1 minute)
 
 def import_slot_range_sanity_test(env: Env, query_type: str = 'FT.SEARCH'):
-    n_docs = 5 * 2**14
+    n_docs = 5 * CLUSTER_SLOTS
     create_and_populate_index(env, 'idx', n_docs)
 
     shard1, shard2 = env.getConnection(1), env.getConnection(2)
@@ -479,7 +484,7 @@ def parallel_update_worker(env, n_docs, stop_event):
             # Update some unrelated fields that are not part of the query
             # We'll add a new field 'update_counter' and 'timestamp' that won't affect search results
             doc_id = random.randint(0, n_docs - 1)
-            key = f'doc-{doc_id}:{{{doc_id % 2**14}}}'
+            key = f'doc-{doc_id}:{{{doc_id % CLUSTER_SLOTS}}}'
 
             # Update fields that are not queried in the test
             con.execute_command('HSET', key,
@@ -497,7 +502,7 @@ def parallel_update_worker(env, n_docs, stop_event):
             time.sleep(0.1)
 
 def import_slot_range_test(env: Env, query_type: str = 'FT.SEARCH', parallel_updates: bool = False):
-    n_docs = 5 * 2**14
+    n_docs = 5 * CLUSTER_SLOTS
     create_and_populate_index(env, 'idx', n_docs)
 
     if query_type == 'FT.SEARCH':
@@ -586,9 +591,7 @@ def test_ft_aggregate_withcursor_import_slot_range_BG():
     env = Env(clusterNodeTimeout=cluster_node_timeout, moduleArgs='WORKERS 2')
     import_slot_range_test(env, 'FT.AGGREGATE.WITHCURSOR')
 
-#TODO: Enable once MOD-13110 is fixed
-#@skip(cluster=False, min_shards=2)
-@skip
+@skip(cluster=False, min_shards=2)
 def test_ft_hybrid_import_slot_range():
     env = Env(clusterNodeTimeout=cluster_node_timeout)
     import_slot_range_test(env, 'FT.HYBRID')
@@ -629,9 +632,7 @@ def test_ft_aggregate_withcursor_import_slot_range_parallel_updates_BG():
     env = Env(clusterNodeTimeout=cluster_node_timeout, moduleArgs='WORKERS 2')
     import_slot_range_test(env, 'FT.AGGREGATE.WITHCURSOR', parallel_updates=True)
 
-#TODO: Enable once MOD-13110 is fixed
-#@skip(cluster=False, min_shards=2)
-@skip
+@skip(cluster=False, min_shards=2)
 def test_ft_hybrid_import_slot_range_parallel_updates():
     env = Env(clusterNodeTimeout=cluster_node_timeout)
     import_slot_range_test(env, 'FT.HYBRID', parallel_updates=True)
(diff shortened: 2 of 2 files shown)


src/hybrid/hybrid_exec.c
```diff
@@ -476,18 +470,16 @@ int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, Q
...
       // verify error exists
       RS_ASSERT(QueryError_HasError(status));
       return REDISMODULE_ERR;
     }
 
     if (backgroundDepletion) {
       int rc = RPSafeDepleter_DepleteAll(depleters);
-      array_free(depleters);
       if (rc != RS_RESULT_OK) {
+        array_free(depleters);
```
kei-nan: shouldn't we free the array regardless of the return code?

src/hybrid/hybrid_exec.c
```diff
@@ -499,7 +491,22 @@ int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, Q
         }
         return REDISMODULE_ERR;
       }
+    } else {
+      // Foreground depletion for WORKERS == 0
+      // Trigger synchronous depletion to read and buffer all results while the spec lock is held.
+      int rc = RPDepleter_DepleteAll(depleters);
+      if (rc != RS_RESULT_OK) {
```
kei-nan: shouldn't we always free the depleters array after the `RPDepleter_DepleteAll` call?

src/hybrid/hybrid_request.c
```diff
@@ -84,6 +84,18 @@ int HybridRequest_BuildDepletionPipeline(HybridRequest *req, const HybridPipelin
...
+          // This depletes all results synchronously on the main thread while
+          // the spec lock is held, preventing crashes when the index is
+          // modified between cursor creation and first read.
+          ResultProcessor *depleter = RPDepleter_New();
+          QITR_PushRP(qctx, depleter);
+          if (isProfile) {
+            // Wrap the depleter with a Profile RP to match the expected end processor type
+            ResultProcessor *profileRP = RPProfile_New(qctx->endProc, qctx);
+            QITR_PushRP(qctx, profileRP);
+          }
```
kei-nan: This section also happens in the other case,
maybe you could move it to be a common code for both flows after the if?
reduce a bit of code duplication

src/profile/profile.c
```diff
@@ -133,6 +133,11 @@ static double _recursiveProfilePrint(RedisModule_Reply *reply, ResultProcessor *
   // RPProfile time because the actual work happens in the background thread
   if (rp->upstream && rp->upstream->type == RP_SAFE_DEPLETER) {
     totalRPTime = rs_wall_clock_convert_ns_to_ms_d(RPSafeDepleter_GetDepletionTime(rp->upstream));
+  } else if (rp->upstream && rp->upstream->type == RP_DEPLETER) {
+    // For RP_DEPLETER (foreground depletion), use depletion time as the total time
+    // because depletion is triggered via RPDepleter_StartDepletion() which bypasses
+    // the wrapping RP_PROFILE to avoid consuming the first result.
+    totalRPTime = rs_wall_clock_convert_ns_to_ms_d(RPDepleter_GetDepletionTime(rp->upstream));
```
kei-nan: maybe this code can move into `RPProfile_GetClock` function in this chance?
it could check the type of the processor and call the get depletion time if the processors are depleters.

kei-nan:
I would like to have a test that would show the problem, 
Maybe FT.DEBUG could be used to stall the query and run whatever writer operations needed.
Would like to not deplete everything on first go if we can avoid it.

src/hybrid/hybrid_exec.c
nafraf: OK. Done.

src/hybrid/hybrid_exec.c
nafraf: OK. Done.

src/hybrid/hybrid_request.c
nafraf: Thanks. Fixed. [88c6c05](https://github.com/RediSearch/RediSearch/pull/8534/commits/88c6c052459680516be092b190e74e1b338e488d)

src/profile/profile.c
nafraf: OK. Moved and the function `RPProfile_GetClock()` was renamed as `RPProfile_GetTime()`.
[88c6c05](https://github.com/RediSearch/RediSearch/pull/8534/commits/88c6c052459680516be092b190e74e1b338e488d)

src/hybrid/hybrid_exec.c
```diff
@@ -643,18 +640,15 @@ int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, Q
...
-      if (backgroundDepletion) {
-        if (rp->type != RP_SAFE_DEPLETER) {
-          break;
-        }
-        array_ensure_append_1(depleters, rp);
+      ResultProcessor *depleter = areq->pipeline.qctx.endProc;
+      if (IsProfile(req) && depleter->type == RP_PROFILE) {
+        depleter = depleter->upstream;
       }
+      RS_ASSERT(depleter->type == expectedDepleterType);
```
oshadmi: Previous code skipped depleters with wrong type (were not added to the array)
Now in production it is being added?

src/hybrid/hybrid_exec.c
```diff
@@ -618,10 +618,7 @@ int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, Q
       return REDISMODULE_ERR;
     }
     // helper array to collect depleters so in async we can deplete them all at once before returning the cursors
```
oshadmi: Now not used only in async?

src/result_processor.c
```diff
@@ -2755,3 +2768,44 @@ ResultProcessor *RPDepleter_New() {
   ret->depleted_results = 0;
   return &ret->base;
 }
+
+/**
+ * Starts depletion for the depleter without consuming any results.
```
oshadmi: "without consuming" or actually is consuming and buffering but "without yielding"?

src/hybrid/hybrid_exec.c
nafraf: OK. Comment fixed. [17dab0f](https://github.com/RediSearch/RediSearch/pull/8534/commits/17dab0f55ed9f5f4203445b58b4e04f0f33b9ce9)

src/result_processor.c
nafraf: Description fixed.

src/hybrid/hybrid_exec.c
nafraf: I put back the previous validation, and now the error is set. [17dab0f](https://github.com/RediSearch/RediSearch/pull/8534/commits/17dab0f55ed9f5f4203445b58b4e04f0f33b9ce9)

src/result_processor.c
```diff
@@ -2747,3 +2760,44 @@ ResultProcessor *RPDepleter_New() {
...
+  RPDepleter *self = (RPDepleter *)base;
+  return self->depletionTime;
+}
+
+/**
+ * Triggers depletion for all depleters in the array.
+ * Stops on first error and returns the error code.
+ * @param depleters Array of depleter processors (must be RP_DEPLETER type)
+ * @return RS_RESULT_OK if all depleters completed successfully,
+ *         or the error code from the first depleter that failed
```
oshadmi: We duplicate the docstring on the header?
This is the more accurate one - We should not duplicate this and have only 1 in the header?

nafraf: OK. The docstring was removed from `.c` file. 
The more accurate description was moved to `.h` file.

src/hybrid/hybrid_exec.c
```diff
@@ -693,6 +689,22 @@ int HybridRequest_StartCursors(StrongRef hybrid_ref, RedisModuleCtx *replyCtx, Q
         }
         return REDISMODULE_ERR;
       }
+    } else {
+      // Foreground depletion for WORKERS == 0
+      // Trigger synchronous depletion to read and buffer all results while the spec lock is held.
+      int rc = RPDepleter_DepleteAll(depleters);
```
kei-nan: The code looks very similar to the code in the if clause.
Maybe we can unify them somehow so the only difference would be the error handling code?

src/result_processor.c
```diff
@@ -1408,9 +1408,14 @@ ResultProcessor *RPProfile_New(ResultProcessor *rp, QueryProcessingCtx *qctx) {
...
 }
 
-rs_wall_clock_ns_t RPProfile_GetClock(ResultProcessor *rp) {
-  RPProfile *self = (RPProfile *)rp;
-  return self->profileTime;
+rs_wall_clock_ns_t RPProfile_GetTime(ResultProcessor *rp) {
+  if (rp->upstream && rp->upstream->type == RP_SAFE_DEPLETER) {
+    return RPSafeDepleter_GetDepletionTime(rp->upstream);
+  } else if (rp->upstream && rp->upstream->type == RP_DEPLETER) {
+    return RPDepleter_GetDepletionTime(rp->upstream);
```
kei-nan, line 1415: if it isn't done in async way then can't we just use the `profileTime` when depleter is sync?

tests/pytests/test_asm.py
```diff
@@ -950,3 +948,203 @@ def test_migrate_no_indexes():
...
+
+    Keys that have already migrated away are silently skipped.
+    """
+    for i in range(n_docs):
+        slot = i % 2**14
+        if slot in slots:
+            try:
+                shard.execute_command('HSET', f'doc:{i}:{{{slot}}}',
+                                      'description', f'basketball sneakers product {i}')
+            except Exception:
```
kei-nan: can we maybe use some other command to avoid the asm exception?
e.g `HSETNX` or something like that?

tests/pytests/test_asm.py
```diff
@@ -950,3 +948,203 @@ def test_migrate_no_indexes():
...
+
+def _write_memory_pressure_docs(shard, start, count, slots):
+    """Write *count* new documents (keys ``newdoc:<i>``) whose slots fall in
+    *slots*, using repetitive text to force memory reuse over previously freed
+    inverted-index blocks.
+
+    Keys whose slot is not owned by *shard* are skipped.
+    """
+    for i in range(start, start + count):
+        slot = i % 2**14
```
kei-nan: maybe put 2**14 in a variable with a nice name to explain what it is?

tests/pytests/test_asm.py
```diff
@@ -950,3 +948,203 @@ def test_migrate_no_indexes():
...
+        if slot in slots:
+            vector = np.array([float(i), float(i % 10)], dtype=np.float32)
+            try:
+                shard.execute_command('HSET', f'newdoc:{i}:{{{slot}}}',
+                                      'description', f'basketball sneakers product {i} ' * 10,
+                                      'embedding', vector.tobytes())
+            except Exception:
```
kei-nan: same comment here regarding the possible exception when using hset.
If we know which keys we moved using asm then maybe we could pass that information into the function and it would avoid doing hset on them?

src/result_processor.c
```diff
@@ -2747,3 +2757,30 @@ ResultProcessor *RPDepleter_New() {
...
+}
+
+int RPDepleter_DepleteAll(arrayof(ResultProcessor*) depleters) {
+  for (size_t i = 0; i < array_len(depleters); i++) {
+    RS_ASSERT(depleters[i]->type == RP_DEPLETER);
+    RPDepleter_StartDepletion(depleters[i]);
+    const RPDepleter *depleter = (const RPDepleter *)depleters[i];
+    if (depleter->last_rc != RS_RESULT_EOF) {
+      return depleter->last_rc;
+    }
```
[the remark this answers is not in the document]
nafraf: For FT.HYBRID, if any of the depleters fails, we don't return partial results.

tests/pytests/test_asm.py
nafraf: OK. Now we have the constant defined as `CLUSTER_SLOTS`
[80ae009](https://github.com/RediSearch/RediSearch/pull/8534/commits/80ae009aac5dbaa276851bc270986c4e0fdaab65)

src/hybrid/hybrid_exec.c
nafraf: OK. Refactored.
[80ae009](https://github.com/RediSearch/RediSearch/pull/8534/commits/80ae009aac5dbaa276851bc270986c4e0fdaab65)

tests/pytests/test_asm.py
nafraf: We can't use `HSETNX` because we need to overwrite the content. 
I added a validation to accept only exceptions caused by the migration.

tests/pytests/test_asm.py
nafraf: Now I'm passing `current_shard1_slots` to the functions where `HSET`is called. This reduces the number of `MOVED/ASK` errors compared to `old_shard1_slots`, but they are still possible if the cluster topology hasn't fully propagated yet. 
Both functions already handle these errors silently, so they won't cause the test to fail.

src/result_processor.c
nafraf: Although `RP_Depleter` is executed in the foreground, we can't use `profileTime` as used in other RPs because the depleters spend most of their time during the depletion phase, not when the results are yielded as is the case for other RPs

PR #8548, opened 2026-03-02
MOD-14065: Fix property order discrepancy in `FILTER`

raz-mon:
This PR fixes a bug we had in the `FILTER` mechanism in the indexing path (both index creation and updates), in which the order of the properties in the `FILTER` expression could affect whether or not a document was indexed.
For example:
```
127.0.0.1:6378> FT.CREATE idx FILTER "@d1==0 && @d2==0" SCHEMA d1 NUMERIC d2 NUMERIC
OK
127.0.0.1:6378> FT.CREATE idx2 FILTER "@d2==0 && @d1==0" SCHEMA d1 NUMERIC d2 NUMERIC
OK
127.0.0.1:6378> hset h1 d2 1
(integer) 1
127.0.0.1:6378> FT.SEARCH idx *
1) (integer) 1
2) "h1"
3) 1) "d2"
   2) "1"
127.0.0.1:6378> FT.SEARCH idx2 *
1) (integer) 0
```

User impact: The user impact is that now the order of properties in the `FILTER` expression does not change the result as to whether or not a document is indexed for a specific index. This brings more reliability on the `FILTER` expression of an index.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches core expression evaluation used during indexing and query execution; mistakes could change which documents are indexed or how predicate errors/missing fields are interpreted. Changes are scoped and heavily covered by new C++ and Python tests, reducing but not eliminating risk.
> 
> **Overview**
> Fixes `FILTER` evaluation in the indexing path so missing fields no longer make results depend on left/right operand order (e.g. `@d1==0 && @d2==0` vs `@d2==0 && @d1==0`).
> 
> Introduces an `EvalMode` (`EVAL_MODE_QUERY` vs `EVAL_MODE_INDEX`) and a distinct `EXPR_EVAL_MISSING` outcome, making indexing filters *lenient* (missing comparisons resolve to false without raising query errors) while keeping query/`FT.ADD IF` evaluation *strict*.
> 
> Refactors predicate evaluation to centralize missing-handling and short-circuit logic, updates call sites to pass the right mode, and adds extensive regression tests covering missing-field AND/OR, comparison operators, and type-mismatch behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1ea7517ed09e244a70833934f1bca153b797b741. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/aggregate/expr/expression.c
+++ b/src/aggregate/expr/expression.c
@@ -94,6 +94,7 @@ static int evalOp(ExprEval *eval, const RSExprOp *op, RSValue *result) {
 
   double n1, n2;
   if (!RSValue_ToNumber(l, &n1) || !RSValue_ToNumber(r, &n2)) {
+    QueryError_SetError(eval->err, QUERY_ERROR_CODE_NUMERIC_VALUE_INVALID, NULL);
     rc = EXPR_EVAL_ERR;
     goto cleanup;
   }


src/aggregate/expr/expression.c
```diff
@@ -192,21 +197,52 @@ static int evalPredicate(ExprEval *eval, const RSPredicate *pred, RSValue *resul
...
+      goto cleanup;
+    }
+  }
+
+  // Evaluate right side
+  int rc_right = evalInternal(eval, pred->right, r);
+
+  if (rc_right == EXPR_EVAL_ERR) {
+    goto cleanup;
+  } else if (rc_right == EXPR_EVAL_MISSING &&
```
JoanFM: why is right expression handling different than left?

src/aggregate/expr/expression.c
```diff
@@ -192,21 +197,52 @@ static int evalPredicate(ExprEval *eval, const RSPredicate *pred, RSValue *resul
...
+    if ((pred->cond == RSCondition_And && !left_bool) ||
+        (pred->cond == RSCondition_Or && left_bool)) {
+      RSValue_SetNumber(result, left_bool);
+      rc = EXPR_EVAL_OK;
+      goto cleanup;
+    }
+  } else {
+    if (rc_left == EXPR_EVAL_MISSING) {
+      RSValue_SetNumber(result, 0);
+      rc = EXPR_EVAL_OK;
```
GuyAv46: Why propagate `EXPR_EVAL_OK` and `0` on a missing value? Assuming the operand is a comparison? This is the right thing to do, also for inequality checks?

src/aggregate/expr/expression.c
```diff
@@ -192,21 +197,52 @@ static int evalPredicate(ExprEval *eval, const RSPredicate *pred, RSValue *resul
...
+  }
+
+  // Evaluate right side
+  int rc_right = evalInternal(eval, pred->right, r);
+
+  if (rc_right == EXPR_EVAL_ERR) {
+    goto cleanup;
+  } else if (rc_right == EXPR_EVAL_MISSING &&
+             pred->cond != RSCondition_And && pred->cond != RSCondition_Or) {
+    // Missing property for comparison operators (-> false)
```
GuyAv46: Also on inequality checks?

src/aggregate/expr/expression.c
```diff
@@ -226,7 +262,7 @@ static int evalProperty(ExprEval *eval, const RSLookupExpr *e, RSValue *res) {
     // No lookup object. This means that the key does not exist
     // Note: Because this is evaluated for each row potentially, do not assume
     // that query error is present:
-    if (eval->err) {
+    if (eval->mode == EVAL_MODE_QUERY && eval->err) {
```
GuyAv46, line 280: Is the comment above still true/relevant?
Why do we only set the error in query mode?

src/aggregate/expr/expression.c
```diff
@@ -347,14 +386,14 @@ EvalCtx *EvalCtx_Create() {
 }
 
 EvalCtx *EvalCtx_FromExpr(RSExpr *expr) {
```
GuyAv46: Consider adding a comment in the H file about these APIs using query mode. This is not clear from their names alone

src/aggregate/expr/expression.h
```diff
@@ -114,12 +114,24 @@ typedef struct RSExpr {
...
 
+/**
+ * Evaluation mode for expression evaluation.
+ * Controls how errors are handled during expression evaluation.
+ */
+typedef enum {
+  /** Query mode: strict evaluation, set errors on missing properties or type mismatches */
+  EVAL_MODE_QUERY,
+  /** Index mode: lenient evaluation, missing properties return NULL without errors */
+  EVAL_MODE_INDEX
```
GuyAv46, line 125: Can't we simply check the error code and decide what to do according to the caller context?

src/aggregate/expr/expression.c
raz-mon: Since they're not equipped to handle missing values, and I didn't want to change them.
I can add more documentation for this

src/aggregate/expr/expression.c
raz-mon: I basically want any comparison involving a missing field to return false, as this is the previous behavior, and also makes most sense IMO

src/aggregate/expr/expression.c
raz-mon: Why different?

src/aggregate/expr/expression.c
raz-mon: We only set the error in query mode since we do nothing with it in indexing mode.
Not setting the error lets us not fail for missing fields, as well as explicitness in the code, rather than thinking we are using the error for indexing mode and actually not.
Moreover, I couldn't just not set the error unfortunately due to some awful behavior of the comparing functions that actually behave differently according to whether or not we have an error context (so not to break old behavior).

src/aggregate/expr/expression.h
raz-mon: What do you mean caller context?

src/aggregate/expr/expression.c
raz-mon: Actually this API isn't used - removed

raz-mon: The subsequent one as well, no API now setting the mode under the hood.

src/aggregate/expr/expression.c
```diff
@@ -94,8 +94,10 @@ static int evalOp(ExprEval *eval, const RSExprOp *op, RSValue *result) {
 
   double n1, n2;
   if (!RSValue_ToNumber(l, &n1) || !RSValue_ToNumber(r, &n2)) {
-
-    QueryError_SetError(eval->err, QUERY_ERROR_CODE_NUMERIC_VALUE_INVALID, NULL);
+    if (eval->mode == EVAL_MODE_QUERY) {
```
JoanFM: why is this valid on Indexing?

src/aggregate/expr/expression.c
JoanFM: if it is not different, can we have a helper method to abstract this with proper naming?

src/aggregate/expr/expression.c
JoanFM: Ah ok, on Indexing we do not return error, simply do not pass the filter I guess

src/rules.c
```diff
@@ -526,15 +526,14 @@ bool SchemaRule_ShouldIndex(struct IndexSpec *sp, RedisModuleString *keyname, Do
...
-    
+    EvalCtx *r = EvalCtx_Create(EVAL_MODE_INDEX);
+
     RedisSearchCtx sctx = { .redisCtx = RSDummyContext };
     QueryError status = QueryError_Default();
     RLookup_LoadRuleFields(&sctx, &r->lk, &r->row, sp, keyCstr, &status);
     QueryError_ClearError(&status); // TODO: report errors
 
     ret = SchemaRule_FilterPasses(r, rule->filter_exp);
-    QueryError_ClearError(r->ee.err);
```
JoanFM, line 537: why is clear removed?

src/aggregate/expr/expression.c
GuyAv46: Do we use this old behavior? I thought you meant the behavior changes when there is no status struct, not whether the error is set or not

src/aggregate/expr/expression.h
GuyAv46: If we call `Eval` from the indexer, errors are ignored (or logged, warning, or whatever); if we call it from the query path, we can use the status if we need to.
If it causes a behavior change that cannot be mitigated internally, then let's keep your approach.

As I see it, the `Eval` logic should've been agnostic to the context it is used in. The result (whether it's an `RSValue` or an error) should be handled by the caller as it fits

src/aggregate/expr/expression.c
raz-mon: Sure, thought to add it but since it seemed relatively concise and clear left it.
Will add.

src/aggregate/expr/expression.c
raz-mon: Good question - I think we are, can check that, but in general I aspired not to touch what I don't have to, since several flows use this code.
The status is the error, maybe we are talking about different things..

src/rules.c
raz-mon: We don't use the error here basically, and there is no loop.. We just destroy which will also clear in a moment.

src/aggregate/expr/expression.c
raz-mon: Added an explanation for this

src/aggregate/expr/expression.h
raz-mon: So that's basically right, and what I would expect as well.
That's not easy/possible to do here since one of the conditions for returning an error is the setting of the error:
```
if (!eval->err || QueryError_IsOk(eval->err)) {
    RSValue_SetNumber(result, res);
    rc = EXPR_EVAL_OK;
  } else {
    RSValue_Clear(result);
  }
  ```

While this may be totally redundant - I tried to keep what I can so that we don't accidentally break behavior, since we are intending to backport this AFAIU.

Comment (raz-mon) [reply src/aggregate/expr/expression.h:125]: I can take another look if I can remove this as well and have the flow agnostic of the context, but from many attempts I ran into many different issues on every change so not promising

File: src/aggregate/expr/expression.c [hunk 9]
```diff
@@ -133,6 +135,9 @@ static int evalOp(ExprEval *eval, const RSExprOp *op, RSValue *result) {
 }
 
 static int getPredicateBoolean(ExprEval *eval, const RSValue *l, const RSValue *r, RSCondition op) {
+  // Always pass eval->err to comparison functions so RSValue_Cmp returns 0 on type mismatch
```
JoanFM: Are these comments necessary now? Or even accurate?

src/aggregate/expr/expression.c
```diff
@@ -188,25 +193,68 @@ static int evalInverted(ExprEval *eval, const RSInverted *vv, RSValue *result) {
...
+  RSValue_SetNumber(result, 0);
+  return EXPR_EVAL_OK;
+}
+
 static int evalPredicate(ExprEval *eval, const RSPredicate *pred, RSValue *result) {
   int res;
   RSValue *l = RSValue_NewUndefined(), *r = RSValue_NewUndefined();
   int rc = EXPR_EVAL_ERR;
-  if (evalInternal(eval, pred->left, l) != EXPR_EVAL_OK) {
+  const int isLogical = (pred->cond == RSCondition_And || pred->cond == RSCondition_Or);
```
JoanFM:
```suggestion
  const bool isLogical = (pred->cond == RSCondition_And || pred->cond == RSCondition_Or);
```

src/aggregate/expr/expression.c
```diff
@@ -188,25 +193,68 @@ static int evalInverted(ExprEval *eval, const RSInverted *vv, RSValue *result) {
...
+    }
+  } else if (rc_left == EXPR_EVAL_MISSING) {
+    // Handle missing here since the comparison flow (`RSValue_Cmp`) would break
+    // behavior if reached with missing values.
+    rc = handleMissingInComparison(eval, result);
     goto cleanup;
   }
 
+  // Evaluate right side
+  int rc_right = evalInternal(eval, pred->right, r);
```
JoanFM: It is quite minor, but I think right and left could be unified into a static function making everything more concise and readable.

Have the same logic as here with the shortcircuit in a static function, and call for each side, it would fall into the shortcircuit only for the left side.

src/document_add.c
```diff
@@ -182,6 +182,7 @@ int RS_AddDocument(RedisSearchCtx *sctx, RedisModuleString *name, const AddDocum
     HiddenString_Free(expr, false);
     if (rc == REDISMODULE_OK) {
       if (res == 0) {
+        QueryError_ClearError(status);
```
JoanFM, line 185: why was this needed? Should this perhaps be another small PR?

src/aggregate/expr/expression.c
raz-mon: They're accurate, but I can remove them now since we don't change anything here..

src/aggregate/expr/expression.c
raz-mon: Will see if I can easily add that

src/document_add.c
raz-mon: The error setting is not doing anything if the code is not `ok` --> We're not setting the error we need here and getting a different error from what the test expects.
This is a legacy command that is no longer maintained so I don't mind having it here.

src/aggregate/expr/expression.c
raz-mon: I don't see a simple way to do that since we want to act differently in the case of logical operators.

GuyAv46 (approved):

src/aggregate/expr/expression.c
```diff
@@ -188,25 +186,68 @@ static int evalInverted(ExprEval *eval, const RSInverted *vv, RSValue *result) {
   return EXPR_EVAL_OK;
 }
 
+/**
```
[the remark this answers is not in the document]
raz-mon: This is expected

src/aggregate/expr/expression.c
```diff
@@ -94,8 +94,6 @@ static int evalOp(ExprEval *eval, const RSExprOp *op, RSValue *result) {
 
   double n1, n2;
   if (!RSValue_ToNumber(l, &n1) || !RSValue_ToNumber(r, &n2)) {
-
-    QueryError_SetError(eval->err, QUERY_ERROR_CODE_NUMERIC_VALUE_INVALID, NULL);
```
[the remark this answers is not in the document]
JoanFM: @raz-mon what about this?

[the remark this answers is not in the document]
raz-mon: Right by accident, bringing it back

tests/pytests/test_filter.py
```diff
@@ -484,3 +484,316 @@ def testFilterWithMissingFields(env):
...
+    # Add document with neither d1 nor d2 BEFORE creating index
+    conn.execute_command('HSET', 'h1', 'other', 'value')
+
+    env.expect('FT.CREATE', 'idx1', 'FILTER', '@d1==0 && @d2==0',
+               'SCHEMA', 'd1', 'NUMERIC', 'd2', 'NUMERIC').ok()
+    env.expect('FT.CREATE', 'idx2', 'FILTER', '@d1==1 || @d2==1',
+               'SCHEMA', 'd1', 'NUMERIC', 'd2', 'NUMERIC').ok()
+
+    # Should have 0 documents because both fields are missing (treated as false)
+    result = env.cmd('FT.SEARCH', 'idx1', '*')
```
oshadmi, line 584: Need waitForIndex?

tests/pytests/test_filter.py
```diff
@@ -484,3 +484,316 @@ def testFilterWithMissingFields(env):
...
+    # Create index with filter: @d1==0 && @d2==0
+    env.expect('FT.CREATE', 'idx1', 'FILTER', '@d1==0 && @d2==0',
+               'SCHEMA', 'd1', 'NUMERIC', 'd2', 'NUMERIC').ok()
+
+    # Create index with filter in reverse order: @d2==0 && @d1==0
+    env.expect('FT.CREATE', 'idx2', 'FILTER', '@d2==0 && @d1==0',
+               'SCHEMA', 'd1', 'NUMERIC', 'd2', 'NUMERIC').ok()
+
+    # Both indexes should have 0 documents (d2=1, d1 missing)
+    result1 = env.cmd('FT.SEARCH', 'idx1', '*')
```
oshadmi, line 506: Need waitForIndex?

tests/pytests/test_filter.py
raz-mon: Yes, will add in a separate PR.

tests/pytests/test_filter.py
raz-mon: Yes, will add in a separate PR.

PR #8608, opened 2026-03-04
[MOD-14348] enable RedisModuleCtx on Disk APIs that need to call RedisModule

JoanFM:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches disk-backed indexing plumbing (vector + tag) and changes multiple C/FFI function signatures, so mistakes in ctx propagation or NULL-handling could break disk/Flex flows at runtime.
> 
> **Overview**
> Enables disk/Flex codepaths to call Redis BigModule APIs by threading `RedisModuleCtx*` through disk wrappers and interfaces: `SearchDisk_OpenIndex`/`SearchDisk_CloseIndex`, disk `indexTags`, and disk vector index creation now accept a ctx.
> 
> Updates vector/tag call sites to pass a real ctx when *creating* indexes (`openVectorIndex(ctx, ...)`, `TagIndex_Index(ctx, ...)`) and to pass `NULL` in read-only/no-create contexts, adding assertions to prevent creating a vector index without a ctx. Index spec parsing/loading is updated to forward ctx (including via `RedisModule_GetContextFromIO` during RDB load), and tests/Rust helpers are adjusted for the new ctx parameters.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 37058db7fbae7457c097ee3c3c7989f69c133f44. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


JoanFM:
We never store the module context anywhere, so I do not think it would be sane to do so.

src/search_disk_api.h
```diff
@@ -46,19 +46,25 @@ typedef struct AsyncReadResult {
...
    * @param disk Pointer to the disk
    * @param indexName Name of the index
    * @param indexNameLen Length of the index name
    * @param type Document type
    * @param deleteBeforeOpen If true, delete any existing data before opening
    * @return Pointer to the index spec, or NULL on error
    */
-  RedisSearchDiskIndexSpec *(*openIndexSpec)(RedisSearchDisk *disk, const char *indexName, size_t indexNameLen, DocumentType type, bool deleteBeforeOpen);
-  void (*closeIndexSpec)(RedisSearchDisk *disk, RedisSearchDiskIndexSpec *index);
+  RedisSearchDiskIndexSpec *(*openIndexSpec)(RedisModuleCtx *ctx, RedisSearchDisk *disk, const char *indexName, size_t indexNameLen, DocumentType type, bool deleteBeforeOpen);
```
kei-nan, line 61: Maybe it would make more sense to pass the context when opening the db itself instead of passing it on every opening of the index?
I think it would really minimize the amount of changes needed on the OSS side and simplify the code.
WDYT @JoanFM?

I see we did that before this change, why do we need a different behaviour?

JoanFM: Because we need the Ctx to get the DB path from the name and to register the DBs and CFs

JoanFM:
> Not sure why this PR is needed We previously already passed the context when opening the db, isn't that enough?

We need it to be able to Register the DBs, CFs, get the DB path from name, In general, we need to call RedisModule API that require the RedisModule Context

kei-nan:
> > Not sure why this PR is needed We previously already passed the context when opening the db, isn't that enough?
> 
> We need it to be able to Register the DBs, CFs, get the DB path from name, In general, we need to call RedisModule API that require the RedisModule Context

But previously we already had the context in the db context itself...
I still don't think this PR is needed.
Reach out if you want to discuss this.

src/search_disk_api.h
kei-nan: but we already have the ctx when opening the db...
We have it internally any point we need to.

chesedo:
> We never store the module context anywhere, so I do not think it would be sane to do so.

You mentioned in the enterprise PR the reason for not storing the context. I don't completely understand why it would become invalid between open() and close(). Can you explain why this can happen?

JoanFM:
> > We never store the module context anywhere, so I do not think it would be sane to do so.
> 
> You mentioned in the enterprise PR the reason for not storing the context. I don't completely understand why it would become invalid between open() and close(). Can you explain why this can happen?

Normally RedisModuleCtx* are tight to a Blocked Client. There is a way to get a deattached one. 

However, I prefer not to hold anything like this during the lifetime, I prefer to pass them as argument.

chesedo:
> > > We never store the module context anywhere, so I do not think it would be sane to do so.
> > 
> > 
> > You mentioned in the enterprise PR the reason for not storing the context. I don't completely understand why it would become invalid between open() and close(). Can you explain why this can happen?
> 
> Normally RedisModuleCtx* are tight to a Blocked Client. There is a way to get a deattached one.
> 
> However, I prefer not to hold anything like this during the lifetime, I prefer to pass them as argument.

Okay, I see :+1:

chesedo (approved):

PR #8613, opened 2026-03-05
MOD-14370 split `RLookup_LoadDocument` into Individual and All paths

JonasKruckenberg:
This change splits up the `RLookup_LoadDocument` into `RLookup_LoadDocumentAll` and `RLookup_LoadDocumentIndividual` replacing the `RLookupLoadFlags`.

The reason behind this is to simplify the load document functions to make them easier to port to Rust in followup PRs. Making them less polymorphic and easier to reason about as well.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches core document/field loading used by query execution and expression evaluation, and changes the C/Rust FFI struct layout; mistakes could cause missing fields or load failures at runtime.
> 
> **Overview**
> **Refactors document loading** by replacing the polymorphic `RLookup_LoadDocument` + `RLookupLoadFlags` with two explicit APIs: `RLookup_LoadDocumentAll` (HGETALL/JSON root) and `RLookup_LoadDocumentIndividual` (key-by-key), and updates `RLookupLoadOptions` to use a simple `cachedOnly` boolean.
> 
> Updates the query loader pipeline (`result_processor.c`) and IF-expression evaluation (`document.c`) to select the appropriate load path, and adjusts the Rust `rlookup` FFI caller to populate the new options struct field (`cachedOnly`) and continue using the individual-key loader.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 79978ae00fc63857a86e06171e7b3e65526bba61. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/result_processor.c
+++ b/src/result_processor.c
@@ -850,6 +850,7 @@ typedef struct {
   ResultProcessor base;
   RLookup *lk;
   RLookupLoadOptions loadopts;
+  bool load_all;
   QueryError status;
 } RPLoader;
 
@@ -884,9 +885,17 @@ static void rpLoader_loadDocument(RPLoader *self, SearchResult *r) {
   }
 
   self->loadopts.dmd = SearchResult_GetDocumentMetadata(r);
+
+  int ret;
+  if (self->load_all) {
+      ret = RLookup_LoadDocumentAll(self->lk, SearchResult_GetRowDataMut(r), &self->loadopts);
+  } else {
+      ret = RLookup_LoadDocumentIndividual(self->lk, SearchResult_GetRowDataMut(r), &self->loadopts);
+  }
+
   // if loading the document has failed, we keep the row as it was.
   // Error code and message are ignored.
-  if (RLookup_LoadDocument(self->lk, SearchResult_GetRowDataMut(r), &self->loadopts) != REDISMODULE_OK) {
+  if (ret != REDISMODULE_OK) {
     // mark the document as "failed to open" for later loaders or other threads (optimization)
     ((RSDocumentMetadata *)(SearchResult_GetDocumentMetadata(r)))->flags |= Document_FailedToOpen;
     // The result contains an expired document.
@@ -923,13 +932,13 @@ static void rploaderNew_setLoadOpts(RPLoader *self, RedisSearchCtx *sctx, RLooku
   self->loadopts.status = &self->status;
   self->loadopts.sctx = sctx;
   self->loadopts.dmd = NULL;
-  self->loadopts.keys = rm_malloc(sizeof(*keys) * nkeys);
-  memcpy(self->loadopts.keys, keys, sizeof(*keys) * nkeys);
-  self->loadopts.nkeys = nkeys;
   if (nkeys) {
-    self->loadopts.mode = RLOOKUP_LOAD_KEYLIST;
+    self->loadopts.keys = rm_malloc(sizeof(*keys) * nkeys);
+    memcpy(self->loadopts.keys, keys, sizeof(*keys) * nkeys);
+    self->loadopts.nkeys = nkeys;
+    self->load_all = false;
   } else {
-    self->loadopts.mode = RLOOKUP_LOAD_ALLKEYS;
+    self->load_all = true;
     RLookup_EnableOptions(lk, RLOOKUP_OPT_ALLLOADED); // TODO: turn on only for HASH specs
   }
 
--- a/src/document.c
+++ b/src/document.c
@@ -912,8 +912,8 @@ int Document_EvalExpression(RedisSearchCtx *sctx, RedisModuleString *key, const 
   }
 
   loadopts = (RLookupLoadOptions){.sctx = sctx, .dmd = dmd, .status = status};
-  if (RLookup_LoadDocument(&lookup_s, &row, &loadopts) != REDISMODULE_OK) {
-    goto done;
+  if (RLookup_LoadDocumentIndividual(&lookup_s, &row, &loadopts) != REDISMODULE_OK) {
+     goto done;
   }
 
   evaluator = (ExprEval){.err = status, .mode = EVAL_MODE_QUERY, .lookup = &lookup_s, .res = NULL, .srcrow = &row, .root = e};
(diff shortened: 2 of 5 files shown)


memark:
I like this idea!

Let's start with a Redis review and I'll go second.

src/result_processor.c
```diff
@@ -850,6 +850,7 @@ typedef struct {
   ResultProcessor base;
   RLookup *lk;
   RLookupLoadOptions loadopts;
+  bool load_all;
```
kei-nan, line 853: We have `QEXEC_AGG_LOAD_ALL` which you can consider using instead of the boolean.
or `RLOOKUP_LOAD_ALLKEYS`

src/result_processor.c
```diff
@@ -884,9 +884,17 @@ static void rpLoader_loadDocument(RPLoader *self, SearchResult *r) {
   }
 
   self->loadopts.dmd = SearchResult_GetDocumentMetadata(r);
+
+  int ret;
+  if (RLookup_IsOptionEnabled(self->lk, RLOOKUP_OPT_ALLLOADED)) {
```
GuyAv46: We should not rely on this flag; it may have been set by another RP, potentially downstream. For example, if we load a specific field, sort+limit by it, then load all the rest for the keys that passed the sorting, the second LOADER will set the flag upon construction, and on execution, the first LOADER will load all the fields instead of the single one it has to

src/result_processor.c
```diff
@@ -923,13 +931,11 @@ static void rploaderNew_setLoadOpts(RPLoader *self, RedisSearchCtx *sctx, RLooku
...
-  self->loadopts.keys = rm_malloc(sizeof(*keys) * nkeys);
-  memcpy(self->loadopts.keys, keys, sizeof(*keys) * nkeys);
-  self->loadopts.nkeys = nkeys;
   if (nkeys) {
-    self->loadopts.mode = RLOOKUP_LOAD_KEYLIST;
+    self->loadopts.keys = rm_malloc(sizeof(*keys) * nkeys);
+    memcpy(self->loadopts.keys, keys, sizeof(*keys) * nkeys);
+    self->loadopts.nkeys = nkeys;
   } else {
-    self->loadopts.mode = RLOOKUP_LOAD_ALLKEYS;
```
GuyAv46, line 932: Can we add another `bool` to the `loadopts` struct?

src/result_processor.c
GuyAv46: The latter is a hint for the downstream when we build it. We can't rely on it while executing. See https://github.com/RediSearch/RediSearch/pull/8613#discussion_r2906264588

src/rlookup.h
```diff
@@ -165,6 +165,10 @@ static inline void RLookup_DisableOptions(RLookup* rlookup, uint32_t options) {
     rlookup->_options &= ~options;
 }
 
+static inline bool RLookup_IsOptionEnabled(RLookup* rlookup, uint32_t options) {
```
memark: This new function needs to be implemented in Rust as well and exposed via FFI.

memark: Consider naming it `RLookup_AreOptionsEnabled`. Or keeping the current name and asserting exactly one option is being passed.

src/result_processor.c
JonasKruckenberg: I attempted your idea @kei-nan but as @GuyAv46 and the tests point out this doesn't really work, a bool is the cleanest option IMO

src/result_processor.c
JonasKruckenberg: yes, I reverted to that 👍🏻

src/result_processor.c
JonasKruckenberg: agreed

src/rlookup.h
JonasKruckenberg: reverted

JonasKruckenberg: reverted

src/result_processor.c
kei-nan: Is this because of FT.SEARCH needing an implicit loader for all fields? and the flag is only turned on for aggregate?
Will specifying LOAD 0 cause all the fields to be loaded?

Relying on the number of keys in order to determine if we should load all of the keys seems odd to me.
I guess this matches the previous behaviour so I can't really complain here.
Maybe just add or check if we have a test that checks `LOAD 0` doesn't load all the fields.

GuyAv46: I think it should be a syntax error, or it’s ignored before we get here. In the loader constructor 0 was the indicator to load everything

GuyAv46, line 853: It would be nicer to have the Boolean flag in the options and not in the loader, so we don’t change the struct size. But if it’s no an option, this solution works fine

JonasKruckenberg: I'm trying to remove any runtime-dynamic fields that determine which code path is taken from the load document (and consequently the `RLookupLoadOptions` struct) so I would prefer to keep this on the loader itself. The cumulative size shouldn't have changed anyway, we remove the 8bit sized `mode` field from the options and added a `bool` field to the loader.

Happy to change this though, don't feel toooo strongly about this approach

src/result_processor.c
```diff
@@ -884,9 +885,17 @@ static void rpLoader_loadDocument(RPLoader *self, SearchResult *r) {
   }
 
   self->loadopts.dmd = SearchResult_GetDocumentMetadata(r);
+
+  int ret;
```
memark, line 889: Maybe initialize with `= REDISMODULE_ERR;` like in other places.

memark, line 889: And call it `rv` like in other places.

src/result_processor.c
GuyAv46: C rules won’t be able to pack the loader, and it will grow by 8 bytes of padding.

But we don’t add new branching to the search path, and the struct is temporary for the query so its size doesn’t matter too much. So this solution works in my opinion.
I hope it will pack better once it’s all rust :)

src/result_processor.c
memark: You saw these comments, right? @JonasKruckenberg

PR #8622, opened 2026-03-06
[MOD-13593] RsValue: Comparison functionality

daxhuiberts:
Implement comparison functionality to be identical in functionality to the C version.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Introduces new cross-language comparison/truthiness semantics and changes C `RSValue_Equal`/`RSValue_BoolTest` return types to `bool`, which may affect any callers relying on integer behavior or edge-case ordering/error handling.
> 
> **Overview**
> Adds a new Rust `value::comparison::compare` implementation (plus integration tests) to mirror C `RSValue` ordering rules across numbers/strings/arrays/trios/refs, including configurable number↔string fallback behavior.
> 
> Exposes new FFI entrypoints in `value_ffi` (`RSValue_Cmp`, `RSValue_Equal`, `RSValue_BoolTest`) and updates the generated Rust headers to declare them, while wiring in `query_error` so numeric conversion failures can populate `QueryError`.
> 
> Updates the C API for `RSValue_Equal` and `RSValue_BoolTest` to return `bool` (instead of `int`) and adjusts call sites/headers accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f4645d24510ba2a8d658219b7614c34305537676. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/value/tests/integration/comparison.rs
+++ b/src/redisearch_rs/value/tests/integration/comparison.rs
@@ -13,14 +13,6 @@ use std::cmp::Ordering;
 use value::comparison::{CompareError, compare};
 use value::{Array, Map, RsString, RsValue, RsValueTrio, SharedRsValue};
 
-use redis_mock::mock_or_stub_missing_redis_c_symbols;
-
-mock_or_stub_missing_redis_c_symbols!();
-
-#[unsafe(no_mangle)]
-pub static mut RSDummyContext: *mut redis_mock::ffi::RedisModuleCtx =
-    redis_mock::globals::redis_module_ctx();
-
 fn array(values: impl IntoIterator<Item = RsValue>) -> RsValue {
     RsValue::Array(Array::new(
         values.into_iter().map(SharedRsValue::new).collect(),
--- a/src/redisearch_rs/value/tests/integration/main.rs
+++ b/src/redisearch_rs/value/tests/integration/main.rs
@@ -12,6 +12,7 @@ extern crate redisearch_rs;
 // Mock or stub the ones that aren't provided by the line above
 redis_mock::mock_or_stub_missing_redis_c_symbols!();
 
+mod comparison;
 mod debug;
 mod hash;
 mod rs_string;


src/redisearch_rs/c_entrypoint/value_ffi/src/comparisons.rs
```diff
@@ -0,0 +1,119 @@
...
+use std::{cmp::Ordering, ffi::c_int};
+use value::RsValue;
+use value::comparison::{CompareError, compare};
+
+/// Compare two [`RsValue`]s, returning `-1`, `0`, or `1`.
+///
+/// When `status` is null, mixed number/string comparisons fall back to
+/// string-based comparison. When `status` is non-null and string-to-number
+/// conversion fails, a [`QueryError`] is written to `status` and `0` is
+/// returned.
```
JonasKruckenberg: right out the gate I don't like this at all... How many callsites does this have? If feasible we should split this up into two comparison functions, one with conversion one without

src/redisearch_rs/c_entrypoint/value_ffi/src/comparisons.rs
```diff
@@ -0,0 +1,119 @@
...
+            // `false` (set from `status.is_null()`), and ensured valid by caller (2.)
+            let query_error = unsafe { status.as_mut().unwrap() };
+            let message = c"Error converting string".to_owned();
+            query_error.set_code_and_message(QueryErrorCode::NumericValueInvalid, Some(message));
+            0
+        }
+    }
+}
+
+/// Test two [`RsValue`]s for equality, returning `1` if equal and `0` otherwise.
```
JonasKruckenberg: should return `bool`

src/redisearch_rs/c_entrypoint/value_ffi/src/comparisons.rs
```diff
@@ -0,0 +1,119 @@
...
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+use crate::util::expect_value;
+use query_error::{QueryError, QueryErrorCode};
+use std::{cmp::Ordering, ffi::c_int};
+use value::RsValue;
+use value::comparison::{CompareError, compare};
+
+/// Compare two [`RsValue`]s, returning `-1`, `0`, or `1`.
```
JonasKruckenberg: what do the three values mean? We should explain the meaning here, or better yet define an enum for the return values

src/redisearch_rs/c_entrypoint/value_ffi/src/comparisons.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+/// Test whether an [`RsValue`] is "truthy".
+///
+/// Returns `1` for non-zero numbers, non-empty strings, and non-empty arrays.
+/// All other variants (including [`RsValue::Null`] and [`RsValue::Map`])
+/// evaluate to `0`. References are followed via
+/// [`RsValue::fully_dereferenced_ref`].
+///
```
JonasKruckenberg: should return `bool`

src/redisearch_rs/value/src/comparison.rs
```diff
@@ -0,0 +1,116 @@
...
+/// - `false` - returns [`CompareError::NoNumberToStringFallback`].
+///
+/// [`RsValue::Trio`] values are compared by their left element.
+/// [`RsValue::Array`] values are compared lexicographically.
+/// [`RsValue::Map`] values cannot be compared and yield [`CompareError::MapComparison`].
+pub fn compare(
+    v1: &RsValue,
+    v2: &RsValue,
+    num_to_str_cmp_fallback: bool,
+) -> Result<Ordering, CompareError> {
```
JonasKruckenberg, line 46: if you split up `RsValue_Cmp` you should also then turn this `num_to_str_cmp_fallback` into a bool const generic.

src/redisearch_rs/c_entrypoint/value_ffi/src/comparisons.rs
daxhuiberts: I've updated the docs to be a bit more descriptive. I think an enum is a bit overkill. Using `-1`, `0`, `1` in comparison functions is quite conventional.

src/redisearch_rs/c_entrypoint/value_ffi/src/comparisons.rs
daxhuiberts: Yeah this is very odd behaviour. I'll look into its use in the C codebase and see if I can refactor this out.

daxhuiberts: This whole query error and fallback behavior is very cumbersome and I really want to get rid of it.
It's very hard to see how this can be factored out though. In `expression.c` it's very hard to guarantee that a query error is always or never present.

Even though its far from ideal I think we should keep it as is for the time being and improve this when we have a better understanding of how it should be used better.

daxhuiberts:
> As much as I don't like these coercing functions on a principle basis I think it's good 👍🏻 (out of curiosity, do you think it's feasible to replace these and other coercing RsValue functions with stricter ones later? or is that basically impossible 🤔)

Yes! `RsValue` feels a bit like a catch-all type of object. Where 50% is applicable for each different call site. And I think we can slowly but gradually slim it down by really looking into each call site in what it actually tries to do.

src/redisearch_rs/c_entrypoint/value_ffi/src/comparisons.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+/// Test two [`RsValue`]s for equality, returning `true` if equal and `false`
+/// otherwise.
+///
+/// Unlike [`RSValue_Cmp`], the string-fallback is never used: if a string
+/// cannot be parsed as a number, the values are considered not equal.
+/// Comparison errors (NaN, maps, incompatible types) are treated as equal
+/// to preserve the legacy C behaviour.
```
JonasKruckenberg:
```suggestion
```

similarly, don't document implementation details

src/redisearch_rs/c_entrypoint/value_ffi/src/comparisons.rs
JonasKruckenberg: this is just about not documenting implementation details

src/redisearch_rs/c_entrypoint/value_ffi/src/comparisons.rs
```diff
@@ -0,0 +1,118 @@
...
+*/
+
+use crate::util::expect_value;
+use query_error::{QueryError, QueryErrorCode};
+use std::{cmp::Ordering, ffi::c_int};
+use value::RsValue;
+use value::comparison::{CompareError, compare};
+
+/// Compare two [`RsValue`]s, returning `-1` if `v1 < v2`, `0` if `v1 == v2`,
+/// or `1` if `v1 > v2`.
```
JonasKruckenberg, line 17: nitpick: could we define a return type enum so we could match against named consts?

daxhuiberts: Like I said, I think an enum is a bit overkill. For most other situations I think that is a good suggestion, but for a comparison function like this using -1, 0, 1  is quite conventional.

memark: I'm in the middle here. -1, 0, 1 sure is conventional, but it's still dumb. (Much like int return codes, which also exist because of lack of better language support.)

src/redisearch_rs/c_entrypoint/value_ffi/src/comparisons.rs
```diff
@@ -0,0 +1,113 @@
...
+            let query_error = unsafe { status.as_mut().unwrap() };
+            let message = c"Error converting string".to_owned();
+            query_error.set_code_and_message(QueryErrorCode::NumericValueInvalid, Some(message));
+            0
+        }
+    }
+}
+
+/// Test two [`RsValue`]s for equality, returning `true` if equal and `false`
+/// otherwise.
```
Itzikvaknin:
```suggestion
/// Check whether two [`RsValue`]s are equal, returning `true` if they are and
/// `false` otherwise.
```

src/redisearch_rs/c_entrypoint/value_ffi/src/comparisons.rs
```diff
@@ -0,0 +1,113 @@
...
+    // SAFETY: ensured by caller (1.)
+    let v2 = unsafe { expect_value(v2) };
+
+    match compare(v1, v2, false) {
+        Ok(Ordering::Less) => false,
+        Ok(Ordering::Equal) => true,
+        Ok(Ordering::Greater) => false,
+        Err(CompareError::NaNNumber) => true,
+        Err(CompareError::MapComparison) => true,
+        Err(CompareError::IncompatibleTypes) => true,
```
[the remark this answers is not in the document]
Itzikvaknin: @daxhuiberts Looks problematic

src/redisearch_rs/value/src/comparison.rs
```diff
@@ -0,0 +1,116 @@
...
+    match (v1, v2) {
+        (RsValue::Ref(r1), RsValue::Ref(r2)) => {
+            compare(r1.value(), r2.value(), num_to_str_cmp_fallback)
+        }
+        (RsValue::Ref(r1), _) => compare(r1.value(), v2, num_to_str_cmp_fallback),
+        (_, RsValue::Ref(r2)) => compare(v1, r2.value(), num_to_str_cmp_fallback),
```
Itzikvaknin, line 52: Can derefing upfront simplify here?

src/redisearch_rs/c_entrypoint/value_ffi/src/comparisons.rs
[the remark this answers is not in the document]
daxhuiberts: Cursorbot is wrong here. The C version does indeed convert arrays, undefined, null, etc (i.e. the incompatible types) to a string, but it always converts it to an empty string. Having the effect that the incompatible types are compared as being equal. Same for `RSValue_Cmp`.

src/redisearch_rs/value/src/comparison.rs
daxhuiberts: Very good point!

Though it looks a bit complex like this. Upfronting the dereferencing means going over the enum type twice. This function already is a bit less optimized/efficient than the existing C version, upfronting the dereferencing would degrade the performance a little bit more. So I would not choose to do this here.

In other functions where performance is not issue it does make sense to do to keep logic separate, which is the case in most (if not all) other functions where this is needed.

src/redisearch_rs/value/src/comparison.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::RsValue;
+use crate::util::{num_to_str, str_to_float};
+use std::cmp::Ordering;
+use std::ops::Deref;
+
+/// Errors that can occur when comparing two [`RsValue`]s.
+#[derive(Debug)]
+pub enum CompareError {
+    /// One or both of the compared numbers were NaN, which has no defined ordering.
+    NaNNumber,
```
memark: Should be either `NaN` or `NaNumber`, right? Seems like there is an extra `N` now.

daxhuiberts: Good catch! Will change to simply `NaN` or better `FloatNaN`.

src/redisearch_rs/value/tests/comparison.rs
```diff
@@ -0,0 +1,219 @@
...
+
+use std::cmp::Ordering;
+use value::comparison::{CompareError, compare};
+use value::{Array, Map, RsString, RsValue, RsValueTrio, SharedRsValue};
+
+use redis_mock::mock_or_stub_missing_redis_c_symbols;
+
+mock_or_stub_missing_redis_c_symbols!();
+
+#[allow(non_upper_case_globals)]
```
memark: I think we have started switching to `expect` instead of `allow` (but it might not be active for tests though).

src/value/value.c
```diff
@@ -731,24 +731,24 @@ int RSValue_Cmp(const RSValue *v1, const RSValue *v2, QueryError *qerr) {
 }
 
 /* Return 1 if the two values are equal */
```
memark:
```suggestion
/* Return true if the two values are equal */
```

src/redisearch_rs/value/tests/comparison.rs
daxhuiberts: Good one, wasn't even needed anymore.

daxhuiberts:
@memark @Itzikvaknin Cursorbot dit find a mismatch in the Rust version, see [the latest commit](https://github.com/RediSearch/RediSearch/pull/8622/commits/5b95728e932c95951f3d1957cc71ba14ac317257) for a final change bringing it in line with the C version.

PR #8625, opened 2026-03-06
[MOD-14319] RLookup - Assert invariants for lookups

memark:
In debug, assert the invariants for lookups in each FFI entrypoint.

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Mostly adds debug-only invariant checks around `RLookup` usage at the FFI boundary, but it touches many exported functions and slightly changes `RLookup_Cleanup`’s drop path, so any assertion failures or pointer-handling regressions would surface at runtime in debug builds.
> 
> **Overview**
> Adds a debug-only `RLookup::assert_valid()` hook (delegating to `KeyList` invariants) and wires it into essentially all `RLookup` FFI entrypoints to catch internal corruption early.
> 
> Updates header generation to map Rust `debug_assertions` to a C `ENABLE_ASSERT` define, and exposes a debug-only `__RLookup_AssertValid()` C API for explicit validation; also refactors `RLookup_Cleanup` to reconstruct the typed lookup before `drop_in_place` (with optional validation).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5be81e2ec111535a719bd7764afbe6aaa27ed4b3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/rlookup_ffi/cbindgen.toml
+++ b/src/redisearch_rs/c_entrypoint/rlookup_ffi/cbindgen.toml
@@ -46,3 +46,7 @@ exclude = ["FieldSpec", "IndexSpec", "Option_Cow_CStr", "SchemaRule"]
 "RLookupKeyFlag" = "RLookup_F"
 "RLookupKeyHeader" = "RLookupKey"
 "RLookupOption" = "RLookup_Opt"
+
+[defines]
+# Remap this rust cfg to the C #define also set by CMake for debug builds.
+"debug_assertions" = "ENABLE_ASSERT"
--- a/src/redisearch_rs/headers/rlookup_rs.h
+++ b/src/redisearch_rs/headers/rlookup_rs.h
@@ -633,6 +633,7 @@ RLookupIterator RLookup_Iter(const struct RLookup *lookup);
  */
 RLookupIteratorMut RLookup_IterMut(struct RLookup *lookup);
 
+#if defined(ENABLE_ASSERT)
 /**
  * Run internal assertions on an [`RLookup`].
  *
@@ -643,6 +644,7 @@ RLookupIteratorMut RLookup_IterMut(struct RLookup *lookup);
  * [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
  */
 void __RLookup_AssertValid(const struct RLookup *lookup);
+#endif
 
 /**
  * Returns a newly created [`RLookupRow`].


src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
```diff
@@ -529,6 +563,10 @@ pub unsafe extern "C" fn RLookup_HasIndexSpecCache(lookup: *const OpaqueRLookup)
 /// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn RLookup_Cleanup(lookup: Option<NonNull<OpaqueRLookup>>) {
+    // Safety: ensured by caller (1.,2.)
+    #[cfg(debug_assertions)]
+    unsafe { RLookup::from_opaque_non_null(lookup.unwrap()) }.assert_valid("RLookup_Cleanup");
+
     // Safety: ensured by caller (1.,2.)
     unsafe {
         lookup.unwrap().cast::<RLookup>().drop_in_place();
```
memark: Is there a way to call this method on the result of `from_opaque_non_null` instead of going back to the ptr and cast? I couldn't make it work.

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
```diff
@@ -637,6 +681,22 @@ pub unsafe extern "C" fn RLookup_IterMut(
...
+/// Run internal assertions on an [`RLookup`].
+///
+/// # Safety
+///
+/// 1. `lookup` must be a [valid], non-null pointer to an `RLookup`.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[cfg(debug_assertions)]
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn RLookup_AssertValid(lookup: *const OpaqueRLookup) {
```
memark: Not strictly necessary, but rather useful when debugging integration tests.

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
daxhuiberts: Well, first of all, the `from_opaque_non_null` only exists when `debug_assertions` are enabled.

You could go from `&mut RLookup` to a `&mut MaybeUninit<RLookup>` and call `assume_init_drop` to get the same effect as calling `drop_in_place` on the pointer directly. I think it would all compile to the same thing.

But let's ask @JonasKruckenberg who probably knows everything about this for his opinion.

memark: >the from_opaque_non_null only exists when debug_assertions are enabled.

I'm not sure that is actually true? I think it's always available.

image

memark: I found a better way. I just had to call drop_in_place passing in the object instead.

daxhuiberts: But of course, a ref is automatically coerce to a ptr.

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
```diff
@@ -466,14 +490,20 @@ pub unsafe extern "C" fn RLookup_GetLength(
...
 
 /// Returns a newly created [`RLookup`].
 #[unsafe(no_mangle)]
 pub extern "C" fn RLookup_New() -> OpaqueRLookup {
-    RLookup::new().into_opaque()
+    let lookup = RLookup::new();
+    #[cfg(debug_assertions)]
+    lookup.assert_valid("RLookup_New");
+
+    lookup.into_opaque()
```
JonasKruckenberg, line 506: nitpick: this one does not add much value at all

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
```diff
@@ -637,6 +680,22 @@ pub unsafe extern "C" fn RLookup_IterMut(
...
+/// Run internal assertions on an [`RLookup`].
+///
+/// # Safety
+///
+/// 1. `lookup` must be a [valid], non-null pointer to an `RLookup`.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[cfg(debug_assertions)]
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn RLookup_AssertValid(lookup: *const OpaqueRLookup) {
```
JonasKruckenberg: are we planning to keep this forever? I'd suggest if the intention is to keep this around as a useful debug tool while we transition the implementation that we prefix this (something like `__RLookup_AssertValid` or `DebugRLookup_AssertValid`)

JonasKruckenberg (dismissed):
a few nitpicks

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
memark: Added it mostly for consistency. It is _possible_ though that `new` would return an invalid state, which we would catch here instead of being confused by later on. Having said that, happy to remove it if you prefer?

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
memark: That's a good idea. Let's go with `__` then.

src/redisearch_rs/c_entrypoint/rlookup_ffi/src/lookup.rs
JonasKruckenberg: its fine

PR #8628, opened 2026-03-09
[MOD-14319] RLookup - Don't null path

memark:
```
    Summary: The fix is in make_tombstone (rlookup/src/lookup/key.rs): removed the line that  
  nulled header.path. The C pipeline holds raw pointers to RLookupKey objects that can later
   be tombstoned (e.g. LOAD @price stores a pointer, then APPLY ... AS price tombstones it).
   Those stale C pointers still need a valid path to load the field from the document. Since
   the backing string memory transfers to the new key (same RLookup lifetime), leaving
  header.path as-is is safe — matching what the original C overrideKey did by never clearing
   _path.
```

Let's discuss this fix over a call.

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches unsafe/FFI-adjacent pointer semantics around key overriding and tombstones; mistakes here could lead to subtle use-after-free or incorrect field lookups, though the change is small and well-scoped.
> 
> **Overview**
> When overriding an `RLookupKey`, tombstoning now **keeps `header.path` intact** instead of nulling it, preserving a valid path string for C callers that may still hold raw pointers to the old key.
> 
> Debug/test assertions in `is_tombstone` were updated accordingly and additional in-code documentation explains the lifetime/ownership rationale and alignment with the original C behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cad16d96582cdb3e931cfdf007c802ef4510bd28. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rlookup/src/lookup/key.rs
+++ b/src/redisearch_rs/rlookup/src/lookup/key.rs
@@ -301,7 +301,6 @@ impl<'a> RLookupKey<'a> {
         #[cfg(any(debug_assertions, test))]
         if is_tombstone {
             debug_assert!(self.name_len == usize::MAX);
-            // Note: `path` is intentionally NOT nulled in tombstones — see `make_tombstone`.
             debug_assert!(self.flags.contains(RLookupKeyFlag::Hidden))
         }
 
@@ -358,6 +357,7 @@ impl<'a> RLookupKey<'a> {
         // `RLookup` lifetime, so the pointer remains valid.
         //
         // This mirrors the original C behaviour: `overrideKey` never cleared `_path`.
+
         let path = mem::take(me._path.deref_mut());
 
         // this will exclude it from iteration


JonasKruckenberg:
I this behavior that we're actively seeing the C code have? Otherwise I would be inclined to keep it as-is (even if the C code currently does use pointers to paths post-tombstonig we should make it stop doing that rather than keeping that behaviour)

JonasKruckenberg:
tombstoned keys should not be accessed anymore, otherwise we'll have problems removing the linked list and codifying dangly-pointer use as valid in the Rust API isn't great for that either

memark:
>I this behavior that we're actively seeing the C code have? 

Yes, a number of tests are failing because the path returned here is actually null:
https://github.com/RediSearch/RediSearch/blob/969269694f24daa49bde8ccb23826c8997927949/src/rlookup_load_document.c#L81

memark:
>even if the C code currently does use pointers to paths post-tombstonig we should make it stop doing that rather than keeping that behaviour

If keeping current Rust behaviour, how should we handle the returned null path in the C code?

JonasKruckenberg:
lets move forward with this. #8639 was a nice try exercise that we should continue to investigate the outcome of but isn't going to help us much right now. I also think we'll be able to do the linked-list replacement regardless 👍🏻

PR #8652, opened 2026-03-10
[MOD-14426] Fix clippy const fn warnings for release

memark:
Fix clippy const fn warnings for release.

These functions are `const` in _release_, but not in _debug_.

#### Release Notes

- [ ] This PR requires release notes
- [x] This PR does not require release notes

---

> [!NOTE]
> **Low Risk**
> Attribute-only changes to silence a lint; no functional logic or data flow is modified.
> 
> **Overview**
> Suppresses Clippy `missing_const_for_fn` warnings in non-debug builds by adding `#[cfg_attr(not(debug_assertions), expect(clippy::missing_const_for_fn))]` to a few functions whose bodies differ between debug and release.
> 
> This affects `HyperLogLog::from_registers`, `ResultProcessorWrapper::debug_assert_same_type`, and `KeyList::{cursor_front,cursor_front_mut}`; behavior is unchanged aside from lint handling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 542073733564c2ce6a65d04dfc19ff7d0bb3febf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #8658, opened 2026-03-10
[MOD-14426] Run linters for release as well

memark:
Run linters for release as well, to catch problems. Previously it was only run for debug.
[Suggested by Dax](https://github.com/RediSearch/RediSearch/pull/8652#issuecomment-4032204066).

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Low risk Makefile change that only expands the `lint` target to also run `clippy`/`cargo doc` under `--release`, potentially increasing CI/runtime but not affecting production code.
> 
> **Overview**
> Updates the `Makefile` `lint` target to run Rust `cargo clippy` and `cargo doc` twice: once in debug (existing behavior) and additionally in `--release` mode, with clearer log messages separating the two runs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 18f03689e24da8c1f1572b0a55da2344cd12d693. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -302,9 +302,12 @@ $(shell grep "EXCLUDE_RUST_BENCHING_CRATES_LINKING_C=" build.sh | cut -d'=' -f2 
 endef
 
 lint:
-	@echo "Running linters..."
+	@echo "Running linters for debug..."
 	@cd $(ROOT)/src/redisearch_rs && cargo clippy --workspace $(call get_rust_exclude_crates) -- -D warnings
 	@cd $(ROOT)/src/redisearch_rs && RUSTDOCFLAGS="-Dwarnings" cargo doc --workspace $(call get_rust_exclude_crates) --no-deps
+	@echo "Running linters for release..."
+	@cd $(ROOT)/src/redisearch_rs && cargo clippy --workspace $(call get_rust_exclude_crates) --release -- -D warnings
+	@cd $(ROOT)/src/redisearch_rs && RUSTDOCFLAGS="-Dwarnings" cargo doc --workspace $(call get_rust_exclude_crates) --no-deps --release
 
 fmt:
 ifeq ($(CHECK),1)


src/redisearch_rs/hyperloglog/src/lib.rs
gdesmott, line 149: Why are we having those?

Makefile
```diff
@@ -302,9 +302,12 @@ $(shell grep "EXCLUDE_RUST_BENCHING_CRATES_LINKING_C=" build.sh | cut -d'=' -f2
 endef
 
 lint:
-	@echo "Running linters..."
+	@echo "Running linters for debug..."
 	@cd $(ROOT)/src/redisearch_rs && cargo clippy --workspace $(call get_rust_exclude_crates) -- -D warnings
 	@cd $(ROOT)/src/redisearch_rs && RUSTDOCFLAGS="-Dwarnings" cargo doc --workspace $(call get_rust_exclude_crates) --no-deps
+	@echo "Running linters for release..."
```
gdesmott, line 308: It makes sense, my only concern is doubling the lint job time. CI is already so slow. :\

ThierryBerger: ( for context, [lint CI job](https://github.com/RediSearch/RediSearch/actions/runs/22912538137/job/66488739149?pr=8531) is 4 minutes but the `make lint` is ~1m30 ; I'd say it's acceptable. )

We could run those lints in parallel if that's a concern :thinking: 
or don't block on release lint before progressing to tests (betting on release being most often passing if debug passes :thought_balloon: )

src/redisearch_rs/hyperloglog/src/lib.rs
ThierryBerger: Fair question!

Because `validate_register_values` is called only in debug, and that's what prevents this function to be const (because it calls non const formatting macro)

The proper fix may be to tell this function to be const when in release 🤔

Makefile
memark: When I compared them, the `make lint` part of CI increased from 5 to 6 min. Which is slower than most other steps anyway (and _they_ run in parallel).

src/redisearch_rs/hyperloglog/src/lib.rs
memark: @ThierryBerger I've tried... I couldn't find a way except duplicating the entire function. (Since `const` is a keyword, not an attribute.)

memark: FYI they were added here
https://github.com/RediSearch/RediSearch/pull/8652

(Apparently I need to manually unstack this PR.)

ThierryBerger: Alternatively, I know there is effort to make formatting more const compliant.. i.e https://doc.rust-lang.org/std/macro.const_format_args.html ; so we may revisit that later

But well, without knowing what we'd gain from constifying this, it's a bit of a moot point

memark: Yeah, it' definately good to keep a watch for all the costifying efforts going on!

In this case I'm just adhering to our lint policy of requiring const where possible. (If it's a good policy is out of scope for this RP. 😄)

PR #8663, opened 2026-03-11
MOD-14216: Unlock spec after iterator creation for disk indexes

raz-mon:
Continuation (fix) of https://github.com/RediSearch/RediSearch/pull/8549, unlocking the index spec after creating the iterators for the query.

Tests for this mechanism will follow shortly (mainly on the disk side).

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Changes locking lifetime around query execution, which can introduce race/GC issues if assumptions about disk snapshot iterators or `diskSpec` detection are wrong. Scope is limited to disk indexes and only affects when the spec lock is released.
> 
> **Overview**
> For disk-backed indexes (`spec->diskSpec`), this PR **unlocks the index spec immediately after `prepareExecutionPlan` creates the iterator tree**, in both background-thread execution (`AREQ_Execute_Callback`) and main-thread execution (`buildPipelineAndExecute`).
> 
> This reduces lock hold time so the main thread can perform writes while long-running queries iterate over disk snapshot data, with a note to revisit as more index types are supported.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5e2f61f45844d0da54737d5b46aabfae3d40b4a7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -971,6 +971,14 @@ void AREQ_Execute_Callback(blockedClientReqCtx *BCRctx) {
     goto error;
   }
 
+  // For disk indexes, release the spec lock immediately after iterator creation.
+  // This is fine, since the disk iterators use snapshots. This allows the main
+  // thread to write while the query iterates over disk data.
+  // NOTE: Revisit as more index types are supported.
+  if (sctx->spec->diskSpec) {
+    RedisSearchCtx_UnlockSpec(sctx);
+  }
+
   if (AREQ_RequestFlags(req) & QEXEC_F_IS_CURSOR) {
     RedisModule_Reply _reply = RedisModule_NewReply(outctx), *reply = &_reply;
     int rc = AREQ_StartCursor(req, reply, execution_ref, &status, false);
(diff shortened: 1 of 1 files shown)


JoanFM (dismissed):

src/aggregate/aggregate_exec.c
```diff
@@ -1017,6 +1017,14 @@ int prepareExecutionPlan(AREQ *req, QueryError *status) {
...
+  // For disk indexes, release the spec lock immediately after iterator creation.
+  // This is fine, since the disk iterators use snapshots. This allows the main
+  // thread to write while the query iterates over disk data.
+  // NOTE: Revisit as more index types are supported.
+  if (sctx->spec->diskSpec) {
+    RedisSearchCtx_UnlockSpec(sctx);
+  }
+
   // check possible optimization after creation of QueryIterator tree
   if (IsOptimized(req)) {
```
GuyAv46: We should probably protect the optimizer (might open new iterators, no?)

src/aggregate/aggregate_exec.c
GuyAv46, line 1055: Pipeline doesn’t need spec access (rlookup stuff)?

GuyAv46, line 1062: Debug is safe without the lock?

src/aggregate/aggregate_exec.c
raz-mon: I think not for the currently supported indexes.
Hence the comment

src/aggregate/aggregate_exec.c
raz-mon: From a relatively quick look no, since what it uses should not change under its nose (e.g., field names).
Will run another one to be sure.

I could also avoid all this and just release after `prepareExecutionPlan`. Wanted to be tight but there isn't great reason for it.

raz-mon: Yes I didn't see an issue there

src/aggregate/aggregate_exec.c
GuyAv46: might be safer to release afterward, even though the `IsOptimized(req)` check should be false in disk mode at the moment

raz-mon: Agreed, moved after

src/aggregate/aggregate_exec.c
raz-mon: So we do, due to `FT.ALTER` maybe changing the number of fields without further protection.
Moving to after the all preparation, tnx.

raz-mon: Moved after the whole function anyways, for clearer logic

PR #8675, opened 2026-03-11
MOD-14498: Use Cache Image On Build Artifact Flow

kei-nan:
Use cached containers for build artifacts flow, reduce setup flakiness

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches multiple GitHub Actions workflows and container build/push logic; failures could block CI or artifact publishing if permissions/caching behave unexpectedly.
> 
> **Overview**
> Adds a reusable `task-build-cached-container.yml` workflow that builds a Docker image from the repo `Dockerfile`, pushes it to GHCR, and uses registry cache tags.
> 
> Updates `task-test.yml`, `task-lint.yml`, and `task-build-artifacts.yml` to consume this cached image when available (skipping setup/install steps), includes a `Fix HOME Directory` workaround for container runs, and adjusts job gating/needs accordingly.
> 
> Tightens workflow/job `permissions` across PR, nightly, merge-queue, and snapshot deploy workflows to the minimal required set (`contents`, `packages`, and `id-token` where artifact upload uses OIDC), and removes now-redundant top-level permissions from `task-build-artifacts.yml`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7991467c00e7095f4300c3923ad8362f5e671834. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/task-build-cached-container.yml
+++ b/.github/workflows/task-build-cached-container.yml
@@ -68,21 +68,21 @@ jobs:
           echo "Using cache ref: $CACHE_REF"
 
       - name: Log in to GHCR
-        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
+        uses: docker/login-action@v3
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ github.token }}
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3
+        uses: docker/setup-qemu-action@v3
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
+        uses: docker/setup-buildx-action@v3
 
       - name: Build and push (cached)
         id: build-and-push
-        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6
+        uses: docker/build-push-action@v6
         with:
           push: true
           context: .


.github/workflows/task-build-artifacts.yml
```diff
@@ -48,18 +43,36 @@ jobs:
...
+      checkout-ref: ${{ inputs.sha || inputs.ref || github.sha }}
+
   build:
     name: Build ${{ needs.get-config.outputs.name }}
-    needs: [get-config]
+    needs: [get-config, build-image]
+    if: |
+      !cancelled() &&
+      needs.get-config.result == 'success' &&
+      (needs.build-image.result == 'success' || needs.build-image.result == 'failure' || needs.build-image.result == 'skipped')
```
meiravgri: what other option do we have?  :sweat_smile:

.github/workflows/task-build-cached-container.yml
```diff
@@ -0,0 +1,110 @@
...
+      - name: Set image name
+        id: meta
+        env:
+          CONTAINER: ${{ inputs.container }}
+          SAN: ${{ inputs.san }}
+          RUNNER_ENV: ${{ inputs.runner-env }}
+          RUN_ID: ${{ github.run_id }}
+        run: |
+          IMAGE_NAME="ghcr.io/redisearch/redisearch-ci"
+          CONTAINER_SAFE=$(echo "$CONTAINER" | sed 's#[^A-Za-z0-9_.-]#-#g')
```
meiravgri, line 59: not identical to the original code
why?
```
CONTAINER_SAFE=$(echo "$CONTAINER" | sed 's/[:\/]/-/g')
```

.github/workflows/task-build-cached-container.yml
```diff
@@ -0,0 +1,110 @@
...
+          SAN_SAFE=$(echo "$SAN" | sed 's#[^A-Za-z0-9_.-]#-#g')
+          RUNNER_ENV_SAFE=$(echo "$RUNNER_ENV" | sed 's#[^A-Za-z0-9_.-]#-#g')
```
meiravgri, line 61: why we didnt need the safe versions before?

.github/workflows/task-build-cached-container.yml
```diff
@@ -0,0 +1,110 @@
...
+        continue-on-error: true
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.checkout-ref }}
+
+      - name: Set image name
+        id: meta
+        env:
+          CONTAINER: ${{ inputs.container }}
+          SAN: ${{ inputs.san }}
```
meiravgri, line 54: no need for default value?
`SAN=${{ inputs.san || 'none' }}`

.github/workflows/task-build-cached-container.yml
```diff
@@ -0,0 +1,110 @@
...
+          CACHE_REF="${IMAGE_NAME}:${CONTAINER_SAFE}-${SAN_SAFE}-${RUNNER_ENV_SAFE}-cache"
+
+          echo "image=$IMAGE_TAG" >> "$GITHUB_OUTPUT"
+          echo "cache=$CACHE_REF" >> "$GITHUB_OUTPUT"
+          echo "Using image tag: $IMAGE_TAG"
+          echo "Using cache ref: $CACHE_REF"
+
+      - name: Log in to GHCR
+        continue-on-error: true
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
```
meiravgri: why not just v3?

.github/workflows/task-build-cached-container.yml
```diff
@@ -0,0 +1,110 @@
...
+    permissions:
+      contents: read
+      packages: write # required for GHCR push
+    outputs:
+      image: ${{ steps.meta.outputs.image }}
+      succeeded: ${{ steps.mark-status.outputs.succeeded }}
+    steps:
+      - name: Checkout
+        if: inputs.checkout-ref == ''
+        continue-on-error: true
```
meiravgri: why not to define `continue-on-error: true` on the job level?

.github/workflows/task-test.yml
```diff
@@ -124,68 +124,12 @@ jobs:
...
-      # This step ONLY runs if all previous steps succeeded
-      - name: Mark success
-        id: mark-success
-        run: |
-          echo "succeeded=true" >> $GITHUB_OUTPUT
+    uses: ./.github/workflows/task-build-cached-container.yml
+    with:
+      container: ${{ needs.get-config.outputs.container }}
+      runner-env: ${{ needs.get-config.outputs.env }}
+      san: ${{ inputs.san || 'none' }}
```
meiravgri, line 132: let's add a default value to `inputs.san` instead of passing `none` explicitly

.github/workflows/task-build-cached-container.yml
```diff
@@ -0,0 +1,110 @@
...
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ github.token }}
+
+      - name: Set up QEMU
+        continue-on-error: true
+        uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3
```
meiravgri: what is the issue with v3 everywhere?

.github/workflows/task-build-cached-container.yml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: Mark build status
+        id: mark-status
+        if: always()
+        run: |
+          if [ "${{ steps.build-and-push.outcome }}" = "success" ]; then
+            echo "succeeded=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "succeeded=false" >> "$GITHUB_OUTPUT"
+          fi
```
meiravgri: what was wrong with the original logic?
```
      - name: Mark success
        id: mark-success
        run: |
          echo "succeeded=true" >> $GITHUB_OUTPUT
```

.github/workflows/task-lint.yml
```diff
@@ -17,9 +32,26 @@ jobs:
...
+          ls -ld "$HOME"
+          ls -ld "$h"
+          echo "USER: $USER"
+          echo "id: $(id)"
+          echo "HOME=$h" >> "$GITHUB_ENV"
       - name: Setup
+        if: needs.build-image.outputs.succeeded != 'true'
         working-directory: .install
-        run: ./install_script.sh sudo
+        run: ./install_script.sh
```
meiravgri: sudo is needed if we dont run within a container

.github/workflows/task-lint.yml
```diff
@@ -17,9 +32,26 @@ jobs:
         uses: actions/checkout@v4
         with:
           submodules: recursive
+      - name: Fix HOME Directory
```
meiravgri, line 33: thinking maybe we need to move this step into `task-build-container`?

.github/workflows/task-lint.yml
```diff
@@ -6,9 +6,24 @@ on:
...
+
   lint:
     name: Linting
+    needs: build-image
+    if: |
+      !cancelled() &&
+      (needs.build-image.result == 'success' || needs.build-image.result == 'failure')
     runs-on: ${{ vars.RUNS_ON || 'ubuntu-latest' }}
+    container:
+      image: ${{ needs.build-image.outputs.succeeded == 'true' && needs.build-image.outputs.image || 'ubuntu:noble' }}
```
meiravgri: why do we need to run in a container if caching failed?

.github/workflows/task-lint.yml
```diff
@@ -6,9 +6,24 @@ on:
...
+      container: ubuntu:noble
+      runner-env: ${{ vars.RUNS_ON || 'ubuntu-latest' }}
+      san: none
+
   lint:
     name: Linting
+    needs: build-image
+    if: |
+      !cancelled() &&
+      (needs.build-image.result == 'success' || needs.build-image.result == 'failure')
```
meiravgri: just use continue-on-failure for build-image

.github/workflows/task-build-artifacts.yml
```diff
@@ -48,18 +43,36 @@ jobs:
...
 
+  build-image:
+    needs: get-config
+    if: ${{ needs.get-config.outputs.container }}
+    name: Build cached container for platform
+    uses: ./.github/workflows/task-build-cached-container.yml
+    with:
+      container: ${{ needs.get-config.outputs.container }}
+      runner-env: ${{ needs.get-config.outputs.env || vars.RUNS_ON || 'ubuntu-latest' }}
+      san: none
```
meiravgri: need to pass explicitly?

.github/workflows/task-build-artifacts.yml
kei-nan: mostly added the if to cover this comment:
https://github.com/RediSearch/RediSearch/pull/8675#discussion_r2919634800
I don't think we really need the if but if it makes cursor happy I'm fine with adding it.

.github/workflows/task-build-cached-container.yml
kei-nan: sonarqube complained about this one - said we should pin the version to avoid security issue
I can remove the pin if you want

.github/workflows/task-lint.yml
kei-nan: well in lint job we always are inside a container

.github/workflows/task-build-cached-container.yml
kei-nan: the variable already has a default, as long as the caller doesn't pass '' we should be good.
If you still want me to change it then let me know.

.github/workflows/task-lint.yml
kei-nan: the problem is the step needs to happen inside the container.
the build container runs outside the actual container if I understand correctly

.github/workflows/task-build-artifacts.yml
kei-nan: no, will remove it

.github/workflows/task-build-cached-container.yml
kei-nan: so we could check for false value as well and not just true

.github/workflows/task-build-cached-container.yml
```diff
@@ -0,0 +1,96 @@
...
+        if: inputs.checkout-ref != ''
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.checkout-ref }}
+
+      - name: Set image name
+        id: meta
+        run: |
+          IMAGE_NAME="ghcr.io/redisearch/redisearch-ci"
+          CONTAINER="${{ inputs.container }}"
```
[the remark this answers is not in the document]
kei-nan: Addressed in the current branch: the user-controlled values are no longer interpolated directly into the shell. They go through `env` first and are sanitized before composing the image tag and cache ref.

.github/workflows/task-build-cached-container.yml
```diff
@@ -0,0 +1,96 @@
...
+          SAN="${{ inputs.san }}"
```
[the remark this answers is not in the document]
kei-nan: Addressed in the current branch: the user-controlled values are no longer interpolated directly into the shell. They go through `env` first and are sanitized before composing the image tag and cache ref.

.github/workflows/task-build-artifacts.yml
```diff
@@ -48,9 +48,22 @@ jobs:
...
+      checkout-ref: ${{ env.REF }}
+    secrets: inherit
+
   build:
     name: Build ${{ needs.get-config.outputs.name }}
-    needs: [get-config]
+    needs: [get-config, build-image]
```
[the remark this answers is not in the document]
kei-nan: Fixed in the current branch. The downstream `build` job now has an explicit `if` that allows `build-image` to be `skipped`, so non-container platforms still run the build.

.github/workflows/task-build-artifacts.yml
```diff
@@ -48,9 +48,22 @@ jobs:
...
+    needs: get-config
+    if: ${{ needs.get-config.outputs.container }}
+    continue-on-error: true
+    name: Build cached container for platform
+    uses: ./.github/workflows/task-build-cached-container.yml
+    with:
+      container: ${{ needs.get-config.outputs.container }}
+      runner-env: ${{ needs.get-config.outputs.env || vars.RUNS_ON || 'ubuntu-latest' }}
+      san: none
+      checkout-ref: ${{ env.REF }}
```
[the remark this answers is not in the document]
kei-nan: Fixed in the current branch. `checkout-ref` now inlines `${{ inputs.sha || inputs.ref || github.sha }}` instead of trying to read `env.REF` from the reusable-workflow `with:` block.

.github/workflows/task-build-artifacts.yml
```diff
@@ -48,9 +48,26 @@ jobs:
       platform: ${{ inputs.platform }}
       architecture: ${{ inputs.architecture }}
 
+  build-image:
```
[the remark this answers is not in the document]
kei-nan: Fixed in 098718168. I propagated `packages: write` through the reusable-workflow call chain so the cached-image workflow no longer depends on the default token permissions from the caller.

.github/workflows/task-build-artifacts.yml
```diff
@@ -4,6 +4,7 @@
 permissions:
   id-token: write   # This is required for requesting the JWT
   contents: read    # This is required for actions/checkout
+  packages: write   # Required for reusable build-image workflow to push GHCR cache/images
```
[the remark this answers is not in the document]
kei-nan: Addressed in the current branch. The write permissions are scoped to the relevant jobs, and 098718168 also propagates the cache-image permissions only through the jobs that need them.

.github/workflows/task-lint.yml
```diff
@@ -6,9 +6,21 @@ on:
...
+    name: Build container for lint
+    uses: ./.github/workflows/task-build-cached-container.yml
+    with:
+      container: ubuntu:noble
+      runner-env: ${{ vars.RUNS_ON || 'ubuntu-latest' }}
+      san: none
+
   lint:
     name: Linting
+    needs: build-image
```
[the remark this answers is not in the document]
kei-nan: Fixed in the current branch. `lint` now overrides the default `success()` gating and still runs when `build-image` fails, falling back to the base image.

.github/workflows/task-build-artifacts.yml
```diff
@@ -48,18 +43,36 @@ jobs:
...
+      checkout-ref: ${{ inputs.sha || inputs.ref || github.sha }}
+
   build:
     name: Build ${{ needs.get-config.outputs.name }}
-    needs: [get-config]
+    needs: [get-config, build-image]
+    if: |
+      !cancelled() &&
+      needs.get-config.result == 'success' &&
+      (needs.build-image.result == 'success' || needs.build-image.result == 'skipped')
```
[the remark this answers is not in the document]
kei-nan: Fixed in the current branch. The `build` job now allows `needs.build-image.result == failure` and falls back to the uncached base image.

.github/workflows/task-build-cached-container.yml
kei-nan: I kept the broader sanitization here because the generated tag now includes caller-provided `container`, `san`, and `runner-env`. Restricting all tag components to `[A-Za-z0-9_.-]` keeps the tag valid and avoids shell and tag-format edge cases.

.github/workflows/task-build-cached-container.yml
kei-nan: Before this PR we were not composing the cache tag from as many caller-controlled parts. Once `container`, `san`, and `runner-env` became part of the generated tag, we needed to sanitize them consistently.

.github/workflows/task-build-cached-container.yml
kei-nan: I left this at the step level intentionally. For jobs that call a reusable workflow, `continue-on-error` is not a supported key, and inside the callee we still need the later steps to run so `mark-status` can emit `succeeded=false` on cache-build failures.

.github/workflows/task-test.yml
kei-nan: I removed the explicit `san: none` from the fixed-value callers. I left `task-test` as `inputs.san || none` only when calling the cache-image workflow, because `inputs.san` is still used elsewhere in `task-test` to mean a real sanitizer value or an empty string. Defaulting the input itself to `none` would leak into the non-sanitizer path.

.github/workflows/task-build-cached-container.yml
kei-nan: Keeping the full SHA pins. `# v3` is only the human-readable label; the SHA pin is what protects us against tag retargeting and is what the security tooling was asking for.

.github/workflows/task-lint.yml
kei-nan: For lint I kept the fallback inside a container on purpose. A cache miss should only change performance, not the execution environment, so we fall back from the cached image to `ubuntu:noble` and run setup there. Falling back to the host runner would need a separate `sudo` path and different behavior.

.github/workflows/task-lint.yml
kei-nan: For reusable-workflow calls we cannot use `continue-on-error` on the caller job. We also still want the callee to continue through the status-marking step so downstream jobs can distinguish a cache-build failure from a successful cached image build.

.github/workflows/task-build-artifacts.yml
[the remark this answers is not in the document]
kei-nan: Follow-up: I pushed 696d6ffe as well. The original fix propagated `packages: write`, and this follow-up preserves `id-token: write` on the artifact reusable-workflow chain so the OIDC upload path still works while the cache-image workflow gets the GHCR permission it needs.

.github/workflows/task-build-artifacts.yml
kei-nan: will simplify the condition

.github/workflows/task-build-cached-container.yml
meiravgri: i wanted to comment that in task-test 😅

.github/workflows/task-build-cached-container.yml
meiravgri: lets use official versions

.github/workflows/task-build-cached-container.yml
kei-nan: added a continue on error at the job level

.github/workflows/task-build-cached-container.yml
```diff
@@ -0,0 +1,110 @@
...
+          CONTAINER_SAFE=$(echo "$CONTAINER" | sed 's#[^A-Za-z0-9_.-]#-#g')
+          SAN_SAFE=$(echo "$SAN" | sed 's#[^A-Za-z0-9_.-]#-#g')
+          RUNNER_ENV_SAFE=$(echo "$RUNNER_ENV" | sed 's#[^A-Za-z0-9_.-]#-#g')
+          IMAGE_TAG="${IMAGE_NAME}:${CONTAINER_SAFE}-${SAN_SAFE}-${RUNNER_ENV_SAFE}-${RUN_ID}"
+          CACHE_REF="${IMAGE_NAME}:${CONTAINER_SAFE}-${SAN_SAFE}-${RUNNER_ENV_SAFE}-cache"
+
+          echo "image=$IMAGE_TAG" >> "$GITHUB_OUTPUT"
+          echo "cache=$CACHE_REF" >> "$GITHUB_OUTPUT"
+          echo "Using image tag: $IMAGE_TAG"
+          echo "Using cache ref: $CACHE_REF"
```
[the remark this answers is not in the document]
kei-nan: added a continue on error at the job level.

.github/workflows/task-lint.yml
meiravgri: Ah right. but we can have it as an additional job after build-image in the task-build-cached-container flow, no?

.github/workflows/task-lint.yml
meiravgri: it was running on a host runner before this pr 🤷 thought it would simplify. your call

PR #8677, opened 2026-03-11
[MOD-14384] count every change for disk GC

JoanFM:

2. Change: What is the change

1. MOD-...
2. #...

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Changes GC triggering and logically-deleted accounting by expanding what counts as a “change” (writes/updates) and by splitting counters, which can affect GC frequency and global stats correctness. Risk is moderate due to concurrency/atomic counter handling and impact on compaction behavior, but scope is limited to GC and indexing callbacks.
> 
> **Overview**
> **Disk GC now runs based on total document churn, not just deletes.** It tracks `writesFromLastRun`, `deletesFromLastRun`, and `updatesFromLastRun`, triggers GC when their sum crosses `forkGcCleanThreshold`, and resets these counters before executing `SearchDisk_RunGC`.
> 
> **GC callback surface is expanded** with new `onWrite`/`onUpdate` hooks wired through `GCContext`, used by `indexer.c` to classify `REPLACE` as update vs write.
> 
> **Fork GC delete tracking is clarified** by renaming the counter to `deletedOrUpdatedDocsFromLastRun`, wiring deletes and updates to the same callback, and updating the related C++ test assertion. Headers also add a C/C++ `RS_Atomic` abstraction for the GC structs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4b44b74161303b49bbd330f0d7490a9a6bde079c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/fork_gc.h
+++ b/src/fork_gc.h
@@ -14,10 +14,14 @@
 #include "gc.h"
 #include "VecSim/vec_sim.h"
 #include <poll.h>
-#include <stdatomic.h>
 
 #ifdef __cplusplus
+#include <atomic>
+#define RS_Atomic(T) std::atomic<T>
 extern "C" {
+#else
+#define RS_Atomic(T) _Atomic(T)
+#include <stdatomic.h>
 #endif
 
 typedef struct {
@@ -54,7 +58,7 @@ typedef struct ForkGC {
   volatile uint32_t execState;
 
   struct timespec retryInterval;
-  atomic_size_t deletedOrUpdatedDocsFromLastRun;
+  RS_Atomic(size_t) deletedOrUpdatedDocsFromLastRun;
 
   // current value of RSGlobalConfig.gcConfigParams.gcSettings.forkGCCleanNumericEmptyNodes
   // This value is updated during the periodic callback execution.
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -136,7 +136,7 @@ static void writeCurEntries(RSAddDocumentCtx *aCtx, RedisSearchCtx *ctx) {
 
 /** Assigns a document ID to a single document. Handles only RAM index */
 static RSDocumentMetadata *makeDocumentId(RedisModuleCtx *ctx, RSAddDocumentCtx *aCtx, IndexSpec *spec,
-                                          int replace, QueryError *status, bool *removing) {
+                                          int replace, bool *updated) {
   DocTable *table = &spec->docs;
   Document *doc = aCtx->doc;
   if (replace) {
@@ -147,7 +147,7 @@ static RSDocumentMetadata *makeDocumentId(RedisModuleCtx *ctx, RSAddDocumentCtx 
       --spec->stats.scoring.numDocuments;
       RS_LOG_ASSERT(spec->stats.scoring.totalDocsLen >= dmd->docLen, "totalDocsLen is smaller than dmd->docLen");
       spec->stats.scoring.totalDocsLen -= dmd->docLen;
-      *removing = true;
+      *updated = true;
       if (spec->flags & Index_HasVecSim) {
         for (int i = 0; i < spec->numFields; ++i) {
           if (spec->fields[i].types == INDEXFLD_T_VECTOR) {
@@ -194,7 +194,7 @@ static void doAssignIds(RSAddDocumentCtx *cur, RedisSearchCtx *ctx) {
     }
 
     RS_ASSERT(cur->doc);
-    bool removing = false;
+    bool updated = false;
     if (spec->diskSpec) {
       size_t len;
       const char *key = RedisModule_StringPtrLen(cur->doc->docKey, &len);
@@ -215,7 +215,7 @@ static void doAssignIds(RSAddDocumentCtx *cur, RedisSearchCtx *ctx) {
         spec->stats.scoring.numDocuments--;
         RS_ASSERT(spec->stats.scoring.totalDocsLen >= oldLen);
         spec->stats.scoring.totalDocsLen -= oldLen;
-        removing = docId != 0; // If docId is 0, the document was not added
+        updated = docId != 0; // If docId is 0, the document was not added
       }
       if (docId) {
         cur->doc->docId = docId;
@@ -229,7 +229,7 @@ static void doAssignIds(RSAddDocumentCtx *cur, RedisSearchCtx *ctx) {
     } else {
       RS_LOG_ASSERT(!cur->doc->docId, "docId must be 0");
       RSDocumentMetadata *md = makeDocumentId(ctx->redisCtx, cur, spec,
-                                              cur->options & DOCUMENT_ADD_REPLACE, &cur->status, &removing);
+                                              cur->options & DOCUMENT_ADD_REPLACE, &updated);
       if (!md) {
         cur->stateFlags |= ACTX_F_ERRORED;
         continue;
@@ -257,7 +257,7 @@ static void doAssignIds(RSAddDocumentCtx *cur, RedisSearchCtx *ctx) {
       }
       DMD_Return(md);
     }
-    if (removing) {
+    if (updated) {
       if (spec->gc) {
         GCContext_OnUpdate(spec->gc);
       }
--- a/src/disk_gc.h
+++ b/src/disk_gc.h
@@ -13,23 +13,27 @@
 #include "redismodule.h"
 #include "gc.h"
 #include "util/references.h"
-#include <stdatomic.h>
 
 #ifdef __cplusplus
+#include <atomic>
+#define RS_Atomic(T) std::atomic<T>
 extern "C" {
+#else
+#define RS_Atomic(T) _Atomic(T)
+#include <stdatomic.h>
 #endif
 
 /* Internal definition of the disk GC context (each disk index has one).
  * Stats are maintained in disk info; we do not duplicate them here. */
 typedef struct DiskGC {
   WeakRef index;
-  atomic_size_t intervalSec;
+  RS_Atomic(size_t) intervalSec;
   // Tracks only writes since last GC run (no updates)
-  atomic_size_t writesFromLastRun;
+  RS_Atomic(size_t) writesFromLastRun;
   // Tracks only deletes for global stats (no updates)
-  atomic_size_t deletesFromLastRun;
+  RS_Atomic(size_t) deletesFromLastRun;
   // Tracks only updates for global stats (no pure writes or deletes)
-  atomic_size_t updatesFromLastRun;
+  RS_Atomic(size_t) updatesFromLastRun;
 } DiskGC;
 
 DiskGC *DiskGC_Create(StrongRef spec_ref, GCCallbacks *callbacks);


src/indexer.c
```diff
@@ -221,6 +222,16 @@ static void doAssignIds(RSAddDocumentCtx *cur, RedisSearchCtx *ctx) {
         cur->doc->docId = docId;
         spec->stats.scoring.totalDocsLen += cur->fwIdx->totalFreq;
         ++spec->stats.scoring.numDocuments;
+        // Notify GC of the change
+        if (spec->gc) {
+          if (oldLen > 0) {
+            // Document was updated (old version deleted, new version added)
+            GCContext_OnUpdate(spec->gc);
```
JoanFM: Is Update counted twice?

JoanFM:
> Check if Update is counted twice?

No it does not seem so.

src/disk_gc.c
```diff
@@ -29,16 +29,20 @@ static bool periodicCb(void *privdata, bool force) {
...
 
   size_t num_docs_cleaned = SearchDisk_RunGC(sp->diskSpec, sp);
-
   IndexsGlobalStats_DecreaseLogicallyDeleted(num_docs_cleaned);
-  atomic_fetch_sub(&gc->deletedDocsFromLastRun, num_docs_cleaned);
+  size_t deleted = atomic_load(&gc->deletedDocsFromLastRun);
+  size_t to_subtract = deleted < num_docs_cleaned ? deleted : num_docs_cleaned;
+  atomic_fetch_sub(&gc->deletedDocsFromLastRun, to_subtract);
+  // Reset the changes counter after GC run
+  atomic_store(&gc->changesFromLastRun, 0);
```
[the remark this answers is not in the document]
JoanFM: good point

src/disk_gc.c
```diff
@@ -50,6 +54,7 @@ static void onTerminateCb(void *privdata) {
   DiskGC *gc = privdata;
   size_t remaining = atomic_exchange(&gc->deletedDocsFromLastRun, 0);
   IndexsGlobalStats_DecreaseLogicallyDeleted(remaining);
+  atomic_store(&gc->changesFromLastRun, 0);
```
raz-mon: Done twice

src/fork_gc.c
```diff
@@ -1160,6 +1160,8 @@ ForkGC *FGC_Create(StrongRef spec_ref, GCCallbacks *callbacks) {
   callbacks->renderStatsForInfo = statsForInfoCb;
   callbacks->getInterval = getIntervalCb;
   callbacks->onDelete = deleteCb;
+  callbacks->onAdd = NULL;
+  callbacks->onUpdate = NULL;
```
raz-mon: In order not to count things twice and simplify - how about counting writes + deletions, and summing them up in the periodic cb as the number of changes?

src/disk_gc.c
```diff
@@ -29,16 +29,20 @@ static bool periodicCb(void *privdata, bool force) {
...
   }
+  // Reset changes counter before running GC
+  atomic_store(&gc->changesFromLastRun, 0);
 
   size_t num_docs_cleaned = SearchDisk_RunGC(sp->diskSpec, sp);
-
   IndexsGlobalStats_DecreaseLogicallyDeleted(num_docs_cleaned);
-  atomic_fetch_sub(&gc->deletedDocsFromLastRun, num_docs_cleaned);
+  size_t deleted = atomic_load(&gc->deletedDocsFromLastRun);
+  size_t to_subtract = deleted < num_docs_cleaned ? deleted : num_docs_cleaned;
```
Itzikvaknin: Its seems like both are valid cases, not sure thats the case
 Consider asserting instead

src/fork_gc.c
Itzikvaknin: ➕ 1️⃣

src/disk_gc.c
JoanFM: It is just some stupid security that it does not underflow, and we start passing all the time. I do not think is worth either to assert here.

src/fork_gc.c
```diff
@@ -1159,7 +1159,9 @@ ForkGC *FGC_Create(StrongRef spec_ref, GCCallbacks *callbacks) {
   callbacks->renderStats = statsCb;
   callbacks->renderStatsForInfo = statsForInfoCb;
   callbacks->getInterval = getIntervalCb;
-  callbacks->onDelete = deleteCb;
+  callbacks->onDelete = deleteOrUpdateCb;
+  callbacks->onAdd = NULL; // Additions are not tracked for forkGC
```
raz-mon: Perhaps a no-op callback instead?

src/gc.c
```diff
@@ -152,6 +152,18 @@ void GCContext_OnDelete(GCContext* gc) {
   }
 }
 
+void GCContext_OnAdd(GCContext* gc) {
+  if (gc->callbacks.onAdd) {
+    gc->callbacks.onAdd(gc->gcCtx);
```
raz-mon:
```suggestion
    gc->callbacks.onWrite(gc->gcCtx);
```

src/disk_gc.c
```diff
@@ -29,16 +29,22 @@ static bool periodicCb(void *privdata, bool force) {
     return true;
   }
 
-  size_t num_docs_to_clean = atomic_load(&gc->deletedDocsFromLastRun);
-  if (!force && num_docs_to_clean < RSGlobalConfig.gcConfigParams.gcSettings.forkGcCleanThreshold) {
+  // Check total changes (deletes + adds + updates) to decide whether to run GC
+  size_t num_aditions = atomic_load(&gc->additionsFromLastRun);
```
raz-mon:
```suggestion
  size_t num_aditions = atomic_load(&gc->writesFromLastRun);
```

src/disk_gc.c
```diff
@@ -65,12 +71,18 @@ static void statsForInfoCb(RedisModuleInfoCtx *ctx, void *gcCtx) {
   (void)gcCtx;
 }
 
-static void deleteCb(void *ctx) {
+static void deletedOrUpdatedCb(void *ctx) {
```
raz-mon: If it's not a hustle, I think splitting between deletes and updates is good as well.
This is also interesting stats to output in our info actually.. Not for this PR of course but consider.

src/fork_gc.c
JoanFM: Seems we are checking against NULL already, so seems equivalent

src/disk_gc.c
```diff
@@ -29,16 +29,22 @@ static bool periodicCb(void *privdata, bool force) {
...
+  // Reset additions counter before running GC
+  atomic_store(&gc->additionsFromLastRun, 0);
 
   size_t num_docs_cleaned = SearchDisk_RunGC(sp->diskSpec, sp);
-
   IndexsGlobalStats_DecreaseLogicallyDeleted(num_docs_cleaned);
-  atomic_fetch_sub(&gc->deletedDocsFromLastRun, num_docs_cleaned);
+  size_t deletedOrUpdated = atomic_load(&gc->deletedOrUpdatedDocsFromLastRun);
+  size_t to_subtract = deletedOrUpdated < num_docs_cleaned ? deletedOrUpdated : num_docs_cleaned;
+  atomic_fetch_sub(&gc->deletedOrUpdatedDocsFromLastRun, to_subtract);
```
raz-mon: Why are we not just resetting this as well?

src/disk_gc.c
JoanFM: I prefer to keep as is right now, and perhaps in another PR separate better, because I do not want now to touch or put any risk on the ForkGC implementation

src/disk_gc.c
JoanFM: maybe makes sense yes

JoanFM: makes sense, much simpler

src/disk_gc.c
JoanFM: Ok will change for DisKGC

Itzikvaknin (dismissed):

src/disk_gc.c
```diff
@@ -29,16 +29,22 @@ static bool periodicCb(void *privdata, bool force) {
...
+  atomic_subtract(&gc->writesFromLastRun, num_writes);
+  atomic_subtract(&gc->deletesFromLastRun, num_deletes);
+  atomic_subtract(&gc->updatesFromLastRun, num_updates);
 
-  size_t num_docs_cleaned = SearchDisk_RunGC(sp->diskSpec, sp);
-
-  IndexsGlobalStats_DecreaseLogicallyDeleted(num_docs_cleaned);
-  atomic_fetch_sub(&gc->deletedDocsFromLastRun, num_docs_cleaned);
+  SearchDisk_RunGC(sp->diskSpec, sp);
+  IndexsGlobalStats_DecreaseLogicallyDeleted(num_deletes + num_updates);
```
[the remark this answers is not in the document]
JoanFM: This is expected, better to trust a single source of truth, no problem here

src/disk_gc.c
```diff
@@ -29,16 +29,22 @@ static bool periodicCb(void *privdata, bool force) {
...
+  atomic_fetch_sub(&gc->writesFromLastRun, num_writes);
+  atomic_fetch_sub(&gc->deletesFromLastRun, num_deletes);
+  atomic_fetch_sub(&gc->updatesFromLastRun, num_updates);
 
-  size_t num_docs_cleaned = SearchDisk_RunGC(sp->diskSpec, sp);
-
-  IndexsGlobalStats_DecreaseLogicallyDeleted(num_docs_cleaned);
-  atomic_fetch_sub(&gc->deletedDocsFromLastRun, num_docs_cleaned);
+  size_t num_cleaned = SearchDisk_RunGC(sp->diskSpec, sp);
+  IndexsGlobalStats_DecreaseLogicallyDeleted(num_cleaned);
```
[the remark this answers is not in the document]
JoanFM: no problem, the numDeleted, and so on is about how many updates need to trigger the GC, but the global stats is about how many u cleaned, imagine GC was cancelled in the middle and not all have been updated.

src/disk_gc.c
```diff
@@ -48,8 +54,9 @@ static bool periodicCb(void *privdata, bool force) {
 
 static void onTerminateCb(void *privdata) {
   DiskGC *gc = privdata;
-  size_t remaining = atomic_exchange(&gc->deletedDocsFromLastRun, 0);
-  IndexsGlobalStats_DecreaseLogicallyDeleted(remaining);
+  size_t updates = atomic_load(&gc->updatesFromLastRun);
+  size_t deletes = atomic_load(&gc->deletesFromLastRun);
```
raz-mon: What about resetting these?

src/fork_gc.h
```diff
@@ -53,7 +53,7 @@ typedef struct ForkGC {
   volatile uint32_t execState;
 
   struct timespec retryInterval;
-  volatile size_t deletedDocsFromLastRun;
+  volatile size_t deletedOrUpdatedDocsFromLastRun;
```
raz-mon: Why does this not have to be atomic?

raz-mon: Perhaps out of scope for the PR but if we're already touching it let's understand.

src/indexer.c
```diff
@@ -135,7 +136,7 @@ static void writeCurEntries(RSAddDocumentCtx *aCtx, RedisSearchCtx *ctx) {
 
 /** Assigns a document ID to a single document. Handles only RAM index */
 static RSDocumentMetadata *makeDocumentId(RedisModuleCtx *ctx, RSAddDocumentCtx *aCtx, IndexSpec *spec,
-                                          int replace, QueryError *status) {
+                                          int replace, QueryError *status, bool *updated) {
```
raz-mon: Removing?

src/indexer.c
```diff
@@ -216,18 +215,20 @@ static void doAssignIds(RSAddDocumentCtx *cur, RedisSearchCtx *ctx) {
...
         spec->stats.scoring.totalDocsLen -= oldLen;
+        updated = true;
       }
       if (docId) {
         cur->doc->docId = docId;
         spec->stats.scoring.totalDocsLen += cur->fwIdx->totalFreq;
         ++spec->stats.scoring.numDocuments;
       } else {
         cur->stateFlags |= ACTX_F_ERRORED;
+        continue;
```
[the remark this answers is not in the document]
raz-mon: Should basically never happen..
I'm for asserting this @JoanFM

src/disk_gc.c
JoanFM: we can, it is not very important buit will do

src/fork_gc.h
JoanFM: I think it is just too unsensitive. So will add it

src/indexer.c
JoanFM: what do u mean? to change the name to removing?

raz-mon: No, there was a comment here by AI about the `status` not being used

src/fork_gc.h
```diff
@@ -53,7 +54,7 @@ typedef struct ForkGC {
   volatile uint32_t execState;
 
   struct timespec retryInterval;
-  volatile size_t deletedDocsFromLastRun;
+  atomic_size_t deletedOrUpdatedDocsFromLastRun;
```
Itzikvaknin: Just make sure we are aware that `atomic_size_t` does not imply `volatile` (chatGPT, can validate if needed), so this change removes the `volatile`

JoanFM: yes, but volatile seemed to be a proxy to ensure concurrency.

But I think volatile here is not needed.

raz-mon: How would atomic not imply volatile?
That doesn't make sense to me.
Guess I'll have a chat with AI too..

raz-mon: Can't say I fully understand, but what's for sure is that we don't need volatile if we have atomic here

PR #8680, opened 2026-03-11
MOD-14477: Add sync-point mechanism to control concurrent executions for debugging

raz-mon:
## Summary

Adds a **named sync-point mechanism** for deterministic concurrency testing in debug builds. This enables precise control over query execution timing, allowing tests to pause queries at specific points and coordinate concurrent operations.

## Motivation

Testing race conditions and concurrent behavior is notoriously difficult because of non-deterministic thread scheduling. This PR introduces sync points that let tests:
- Pause a query at a specific execution stage
- Perform concurrent operations (e.g., index updates) while the query is paused
- Resume the query and verify correct behavior

### New `FT.DEBUG SYNC_POINT` Command (debug builds only)

| Subcommand | Description |
|------------|-------------|
| `ARM <name>` | Enable a sync point - queries will pause when reaching it |
| `SIGNAL <name>` | Resume execution and disarm the sync point |
| `IS_WAITING <name>` | Check if a query is currently paused at a sync point |
| `IS_ARMED <name>` | Check if a sync point is armed |
| `CLEAR` | Reset all sync points |

### Predefined Sync Points

- `BeforeIteratorCreate` - Before acquiring spec lock
- `AfterIteratorCreate` - After iterator creation/snapshot established
- `BeforeFirstRead` - Before the first result is read from the iterator
- `BeforeLockRelease` / `AfterLockRelease` - Around spec lock release

### Implementation Details

- All code is guarded by `#ifdef ENABLE_ASSERT` - **zero overhead in production builds**
- Uses atomic operations for thread-safe state management
- Spin-waits with 1ms sleep intervals (matches existing debug patterns)
- Supports up to 16 simultaneously armed sync points

## Example Usage

```python
# Test: Verify query sees consistent snapshot even if index is modified mid-query

# 1. Arm sync point before first read
env.cmd('FT.DEBUG', 'SYNC_POINT', 'ARM', 'BeforeFirstRead')

# 2. Start query in background thread
thread = start_query_thread('FT.SEARCH', 'idx', '*')

# 3. Wait for query to reach sync point
while not env.cmd('FT.DEBUG', 'SYNC_POINT', 'IS_WAITING', 'BeforeFirstRead'):
    time.sleep(0.01)

# 4. Modify the index while query is paused
env.cmd('HSET', 'doc:new', 'field', 'value')

# 5. Resume the query
env.cmd('FT.DEBUG', 'SYNC_POINT', 'SIGNAL', 'BeforeFirstRead')

# 6. Verify query results reflect pre-modification state
results = thread.join()
assert 'doc:new' not in results
```

---

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Changes are gated behind `ENABLE_ASSERT`, so production behavior is unaffected; risk is mainly limited to potential hangs/flakiness in debug/test builds if sync points are armed and not signaled.
> 
> **Overview**
> Introduces **named sync points** (debug builds only) that can be armed and used to pause worker-thread query execution at deterministic points, enabling repeatable concurrency/race-condition tests.
> 
> Adds a new `FT.DEBUG SYNC_POINT` subcommand set (`ARM`, `SIGNAL`, `IS_WAITING`, `IS_ARMED`, `CLEAR`) with a small atomic registry (capacity 16) and integrates `SyncPoint_Wait` calls into query execution at `AfterIteratorCreate` and `BeforeFirstRead`.
> 
> Updates python tests/utilities to exercise the new command lifecycle, capacity limits, and to validate that queries block/resume correctly under controlled worker thread counts.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 10198e023a54bf8b5b5942adba1adf0282e8f5cd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/actions/configure-aws-credentials/action.yml
+++ b/.github/actions/configure-aws-credentials/action.yml
@@ -1,0 +1,127 @@
+name: Configure AWS Credentials
+description: |
+  Configures AWS credentials using either IAM role (OIDC) or access keys.
+  Handles both node20-supported and legacy platforms.
+
+inputs:
+  use_role:
+    description: "Whether to use IAM role (OIDC) authentication"
+    required: true
+  role_to_assume:
+    description: "ARN of the IAM role to assume (required if use_role is true)"
+    required: false
+  aws_access_key_id:
+    description: "AWS access key ID (required if use_role is false)"
+    required: false
+  aws_secret_access_key:
+    description: "AWS secret access key (required if use_role is false)"
+    required: false
+  aws_region:
+    description: "AWS region"
+    required: true
+  node20_supported:
+    description: "Whether node20 is supported on this platform"
+    required: true
+
+runs:
+  using: composite
+  steps:
+    # Role-based auth with node20
+    - name: Configure AWS Credentials Using Role (node20)
+      if: ${{ inputs.use_role == 'true' && inputs.node20_supported == 'true' }}
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        role-to-assume: ${{ inputs.role_to_assume }}
+        aws-region: ${{ inputs.aws_region }}
+
+    # Role-based auth without node20 (manual OIDC)
+    - name: Configure AWS Credentials Using Role (legacy)
+      if: ${{ inputs.use_role == 'true' && inputs.node20_supported != 'true' }}
+      shell: bash
+      env:
+        ROLE_ARN: ${{ inputs.role_to_assume }}
+        AWS_REGION: ${{ inputs.aws_region }}
+      run: |
+        if ! command -v jq &> /dev/null; then
+          echo "Error: jq is not installed."
+          exit 1
+        fi
+
+        echo "Requesting GitHub OIDC token..."
+        OIDC_TOKEN=$(curl -s -H "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" \
+          "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=sts.amazonaws.com" | jq -r '.value // empty') || {
+          echo "Failed to get OIDC token from GitHub"
+          exit 1
+        }
+
+        if [ -z "$OIDC_TOKEN" ]; then
+          echo "Failed to get OIDC token from GitHub"
+          exit 1
+        fi
+
+        echo "Assuming AWS role using OIDC..."
+        CREDS_JSON=$(aws sts assume-role-with-web-identity \
+          --role-arn "$ROLE_ARN" \
+          --role-session-name "GitHubActions" \
+          --web-identity-token "$OIDC_TOKEN" \
+          --region "$AWS_REGION") || {
+          echo "Failed to assume AWS role using OIDC"
+          exit 1
+        }
+
+        AWS_ACCESS_KEY_ID=$(echo "$CREDS_JSON" | jq -r '.Credentials.AccessKeyId')
+        AWS_SECRET_ACCESS_KEY=$(echo "$CREDS_JSON" | jq -r '.Credentials.SecretAccessKey')
+        AWS_SESSION_TOKEN=$(echo "$CREDS_JSON" | jq -r '.Credentials.SessionToken')
+
+        if [ -z "$AWS_ACCESS_KEY_ID" ] || [ -z "$AWS_SECRET_ACCESS_KEY" ] || [ -z "$AWS_SESSION_TOKEN" ]; then
+          echo "Failed to parse AWS credentials from assume-role response"
+          exit 1
+        fi
+
+        echo "::add-mask::$AWS_ACCESS_KEY_ID"
+        echo "::add-mask::$AWS_SECRET_ACCESS_KEY"
+        echo "::add-mask::$AWS_SESSION_TOKEN"
+
+        {
+          echo "AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID"
+          echo "AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY"
+          echo "AWS_SESSION_TOKEN=$AWS_SESSION_TOKEN"
+          echo "AWS_REGION=$AWS_REGION"
+        } >> "$GITHUB_ENV"
+
+        echo "AWS credentials configured successfully using OIDC."
+
+    # Key-based auth with node20
+    - name: Configure AWS Credentials Using Keys (node20)
+      if: ${{ inputs.use_role != 'true' && inputs.node20_supported == 'true' }}
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        aws-access-key-id: ${{ inputs.aws_access_key_id }}
+        aws-secret-access-key: ${{ inputs.aws_secret_access_key }}
+        aws-region: ${{ inputs.aws_region }}
+
+    # Key-based auth without node20
+    - name: Configure AWS Credentials Using Keys (legacy)
+      if: ${{ inputs.use_role != 'true' && inputs.node20_supported != 'true' }}
+      shell: bash
+      env:
+        AWS_ACCESS_KEY_ID: ${{ inputs.aws_access_key_id }}
+        AWS_SECRET_ACCESS_KEY: ${{ inputs.aws_secret_access_key }}
+        AWS_REGION: ${{ inputs.aws_region }}
+      run: |
+        if [ -z "$AWS_ACCESS_KEY_ID" ] || [ -z "$AWS_SECRET_ACCESS_KEY" ] || [ -z "$AWS_REGION" ]; then
+          echo "Missing AWS credentials or region configuration."
+          exit 1
+        fi
+
+        echo "::add-mask::$AWS_ACCESS_KEY_ID"
+        echo "::add-mask::$AWS_SECRET_ACCESS_KEY"
+
+        {
+          echo "AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID"
+          echo "AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY"
+          echo "AWS_REGION=$AWS_REGION"
+        } >> "$GITHUB_ENV"
+
+        echo "AWS credentials configured successfully using access keys."
+
--- a/.github/actions/get-redis/action.yml
+++ b/.github/actions/get-redis/action.yml
@@ -1,0 +1,45 @@
+name: Get Redis
+description: |
+  Checkout Redis repository. Uses actions/checkout for node20-supported platforms,
+  falls back to manual git clone for legacy platforms.
+
+inputs:
+  ref:
+    description: "Redis git ref to checkout"
+    required: true
+  node20_supported:
+    description: "Whether node20 is supported on this platform"
+    required: true
+  path:
+    description: "Path to checkout Redis into"
+    required: false
+    default: "redis"
+
+runs:
+  using: composite
+  steps:
+    - name: Get Redis (node20)
+      if: ${{ inputs.node20_supported == 'true' }}
+      uses: actions/checkout@v4
+      with:
+        repository: redis/redis
+        ref: ${{ inputs.ref }}
+        path: ${{ inputs.path }}
+
+    - name: Get Redis (legacy)
+      if: ${{ inputs.node20_supported != 'true' }}
+      shell: bash
+      env:
+        REDIS_REF: ${{ inputs.ref }}
+        DEST_DIR: ${{ inputs.path }}
+      run: |
+        REPO_URL="https://github.com/redis/redis.git"
+
+        # Clone the repository
+        git clone "$REPO_URL" "$DEST_DIR"
+        cd "$DEST_DIR"
+
+        # Checkout the requested ref
+        git fetch origin "$REDIS_REF"
+        git checkout "$REDIS_REF"
+
--- a/.github/actions/retry-command/action.yml
+++ b/.github/actions/retry-command/action.yml
@@ -1,0 +1,45 @@
+name: Retry Command
+description: |
+  Runs a command with retry logic. Retries up to 5 times with 30 second delays.
+
+inputs:
+  command:
+    description: "The command to run"
+    required: true
+  working_directory:
+    description: "Working directory to run the command from"
+    required: false
+    default: "."
+  max_retries:
+    description: "Maximum number of retry attempts"
+    required: false
+    default: "5"
+  retry_delay:
+    description: "Delay in seconds between retries"
+    required: false
+    default: "30"
+
+runs:
+  using: composite
+  steps:
+    - shell: bash -l -eo pipefail {0}
+      working-directory: ${{ inputs.working_directory }}
+      env:
+        COMMAND: ${{ inputs.command }}
+        MAX_RETRIES: ${{ inputs.max_retries }}
+        RETRY_DELAY: ${{ inputs.retry_delay }}
+      run: |
+        SUCCESS=0
+        for i in $(seq 1 "$MAX_RETRIES"); do
+          echo "Attempt $i of $MAX_RETRIES"
+          if eval "$COMMAND"; then
+            echo "Command succeeded"
+            SUCCESS=1
+            break
+          fi
+          if [ "$i" -lt "$MAX_RETRIES" ]; then
+            echo "Command failed, retrying in $RETRY_DELAY seconds..."
+            sleep "$RETRY_DELAY"
+          fi
+        done
+        [ $SUCCESS -eq 1 ] || exit 1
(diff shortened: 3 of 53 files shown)


src/debug_commands.c
```diff
@@ -119,6 +119,75 @@ bool StoreResultsDebugCtx_IsPaused(void) {
...
+  int idx = atomic_fetch_add(&globalSyncPointCtx.count, 1);
+  if (idx >= SYNC_POINT_MAX_ARMED) {
+    atomic_fetch_sub(&globalSyncPointCtx.count, 1);
+    return;
+  }
+  SyncPointState *sp = &globalSyncPointCtx.points[idx];
+  strncpy(sp->name, name, SYNC_POINT_NAME_MAX_LEN - 1);
+  sp->name[SYNC_POINT_NAME_MAX_LEN - 1] = '\0';
+  atomic_store(&sp->armed, true);
+  atomic_store(&sp->waiting, false);
```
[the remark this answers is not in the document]
raz-mon: Right, will fix.

src/debug_commands.c
```diff
@@ -2212,6 +2290,55 @@ DEBUG_COMMAND(printRPStream) {
...
+ */
+DEBUG_COMMAND(syncPoint) {
+  if (!debugCommandsEnabled(ctx)) return RedisModule_ReplyWithError(ctx, NODEBUG_ERR);
+  // argc layout: FT.DEBUG SYNC_POINT <subcommand> [<point_name>]
+  // argv[0] = FT.DEBUG, argv[1] = SYNC_POINT, argv[2] = subcommand, argv[3] = point_name
+  if (argc < 3) return RedisModule_WrongArity(ctx);
+
+  const char *subOp = RedisModule_StringPtrLen(argv[2], NULL);
+
+  if (!strcmp("ARM", subOp)) {
```
JoanFM: should we call it SCHEDULE?

src/debug_commands.c
```diff
@@ -2212,6 +2290,55 @@ DEBUG_COMMAND(printRPStream) {
...
+    if (argc != 4) return RedisModule_WrongArity(ctx);
+    const char *name = RedisModule_StringPtrLen(argv[3], NULL);
+    return RedisModule_ReplyWithBool(ctx, SyncPoint_IsWaiting(name));
+  }
+  if (!strcmp("IS_ARMED", subOp)) {
+    if (argc != 4) return RedisModule_WrongArity(ctx);
+    const char *name = RedisModule_StringPtrLen(argv[3], NULL);
+    return RedisModule_ReplyWithBool(ctx, SyncPoint_IsArmed(name));
+  }
+  if (!strcmp("CLEAR", subOp)) {
```
JoanFM: Let's have the subcommands as constants ?

src/debug_commands.h
```diff
@@ -91,7 +91,57 @@ bool StoreResultsDebugCtx_IsPauseAfterEnabled(void);
...
+bool SyncPoint_IsArmed(const char *name);
+// Clear all sync points
+void SyncPoint_ClearAll(void);
+// Called from code paths to potentially wait at a sync point
+// If the named point is armed, blocks until signaled
+void SyncPoint_Check(const char *name);
+
+// Macro for sync point checks - compiles to nothing in release builds
+// Usage: SYNC_POINT_CHECK(SYNC_POINT_AFTER_ITERATOR_CREATE);
+#define SYNC_POINT_CHECK(name) SyncPoint_Check(name)
```
JoanFM: not being used

src/debug_commands.c
```diff
@@ -119,6 +119,84 @@ bool StoreResultsDebugCtx_IsPaused(void) {
...
+  SyncPointState *existing = SyncPoint_FindByName(name);
+  if (existing) {
+    atomic_store(&existing->armed, true);
+    return;
+  }
+  // Reserve a slot atomically. We use a simple counter since ARM is only called
+  // from the main thread (via FT.DEBUG command), so no concurrent ARMs occur.
+  int idx = atomic_load(&globalSyncPointCtx.count);
+  if (idx >= SYNC_POINT_MAX_ARMED) {
+    return;
```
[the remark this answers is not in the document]
JoanFM: yes we should return some REDISMODULE_OK or REDISMODULE_ERR and fail or not

src/debug_commands.h
```diff
@@ -91,7 +91,57 @@ bool StoreResultsDebugCtx_IsPauseAfterEnabled(void);
...
+typedef struct SyncPointState {
+  char name[SYNC_POINT_NAME_MAX_LEN];   // Name of the sync point
+  atomic_bool armed;                    // Whether this sync point is armed (will block)
+  atomic_bool waiting;                  // Whether a thread is currently waiting at this point
+} SyncPointState;
+
+// Container for all sync point states
+typedef struct SyncPointCtx {
+  SyncPointState points[SYNC_POINT_MAX_ARMED];   // Array of sync points
+  atomic_int count;                              // Number of armed sync points
```
JoanFM: atomic_uint?

src/debug_commands.c
```diff
@@ -119,6 +119,84 @@ bool StoreResultsDebugCtx_IsPaused(void) {
...
+// ============================================================================
+// Named Sync Points Implementation
+// ============================================================================
+static SyncPointCtx globalSyncPointCtx = {0};
+
+// Internal helper: find sync point by name
+static SyncPointState* SyncPoint_FindByName(const char *name) {
+  // Use acquire semantics to synchronize with the release fence in SyncPoint_Arm,
+  // ensuring we see fully initialized slots when iterating.
+  int count = atomic_load_explicit(&globalSyncPointCtx.count, memory_order_acquire);
```
JoanFM: uint?

src/debug_commands.c
```diff
@@ -119,6 +119,84 @@ bool StoreResultsDebugCtx_IsPaused(void) {
...
+  // Other threads calling SyncPoint_FindByName iterate up to `count`,
+  // so we must fully initialize before incrementing count.
+  SyncPointState *sp = &globalSyncPointCtx.points[idx];
+  strncpy(sp->name, name, SYNC_POINT_NAME_MAX_LEN - 1);
+  sp->name[SYNC_POINT_NAME_MAX_LEN - 1] = '\0';
+  atomic_store(&sp->armed, true);
+  atomic_store(&sp->waiting, false);
+  // Memory fence: ensure all writes above are visible before incrementing count
+  atomic_thread_fence(memory_order_release);
+  atomic_fetch_add(&globalSyncPointCtx.count, 1);
```
JoanFM, line 186: is this important? Does it even need to be atomic? It is only accessed/checked from main thread right?

Nothing bad about keeping as is, just wondering If I am missing something

src/debug_commands.h
```diff
@@ -91,7 +91,57 @@ bool StoreResultsDebugCtx_IsPauseAfterEnabled(void);
...
+void SyncPoint_Signal(const char *name);
+// Check if a thread is waiting at the named sync point
+bool SyncPoint_IsWaiting(const char *name);
+// Check if a sync point is armed
+bool SyncPoint_IsArmed(const char *name);
+// Clear all sync points
+void SyncPoint_ClearAll(void);
+// Called from code paths to potentially wait at a sync point
+// If the named point is armed, blocks until signaled
+void SyncPoint_Check(const char *name);
```
JoanFM:
```suggestion
void SyncPoint_Wait(const char *name);
```

JoanFM: I believe calling it Wait would make the code easier to understand?

src/debug_commands.h
```diff
@@ -91,7 +91,57 @@ bool StoreResultsDebugCtx_IsPauseAfterEnabled(void);
...
+// Maximum length of a sync point name
+#define SYNC_POINT_NAME_MAX_LEN 64
+
+// Predefined sync point names for query execution
+// These correspond to specific locations in the query execution path
+#define SYNC_POINT_AFTER_ITERATOR_CREATE  "AfterIteratorCreate"
+#define SYNC_POINT_BEFORE_ITERATOR_CREATE "BeforeIteratorCreate"
+#define SYNC_POINT_BEFORE_LOCK_RELEASE_BG "BeforeLockRelease"
+#define SYNC_POINT_AFTER_LOCK_RELEASE     "AfterLockRelease"
+#define SYNC_POINT_BEFORE_FIRST_READ      "BeforeFirstRead"
```
JoanFM, line 103: Right now this is the only one used right?

JoanFM: Can we have a compile constexpr checking that none of these SYNC_POINT... length is above SYNC_POINT_NAME_MAX_LEN?

src/debug_commands.c
raz-mon: I like `ARM` better

src/debug_commands.c
raz-mon: Yes, since the reader (query thread) is concurrently accessing

src/debug_commands.h
raz-mon: Yeah that makes sense, changed

src/debug_commands.h
raz-mon: I'm not a fan of that, since it should fail tests if overflows the max-length since the test sync-point should not instantiate the trimmed sync-point, and it litters the code IMO

src/debug_commands.h
```diff
@@ -91,7 +92,51 @@ bool StoreResultsDebugCtx_IsPauseAfterEnabled(void);
...
+  char name[SYNC_POINT_NAME_MAX_LEN];   // Name of the sync point
+  atomic_bool armed;                    // Whether this sync point is armed (will block)
+  atomic_bool waiting;                  // Whether a thread is currently waiting at this point
+} SyncPointState;
+
+// Container for all sync point states
+typedef struct SyncPointCtx {
+  SyncPointState points[SYNC_POINT_MAX_ARMED];   // Array of sync points
+  _Atomic uint32_t count;                        // Number of armed sync points
+} SyncPointCtx;
```
meiravgri: Can we allow more than one container? if not, move struct definition to `debug_commands.c`

tests/pytests/common.py
```diff
@@ -292,6 +292,11 @@ def getWorkersThpoolStats(env):
 def getWorkersThpoolNumThreads(env):
     return env.cmd(debug_cmd(), "WORKERS", "n_threads")
 
+def set_workers(env, workers):
+    """Set the worker thread count and verify that the change took effect."""
+    env.expect(config_cmd(), 'SET', 'WORKERS', workers).ok()
+    env.assertEqual(getWorkersThpoolNumThreads(env), workers)
+
```
meiravgri, line 299:
```suggestion
def set_workers(env, workers):
    """Set the worker thread count and verify that the change took effect."""
    verify_command_OK_on_all_shards(env, config_cmd(), 'SET', 'WORKERS', workers)
    env.assertEqual(getWorkersThpoolStatsFromAllShards(env), [workers] * env.shardsCount)

```

src/debug_commands.c
```diff
@@ -2550,6 +2687,13 @@ int RegisterDebugCommands(RedisModuleCommand *debugCommand) {
               RS_DEBUG_FLAGS);
     if (rc != REDISMODULE_OK) return rc;
   }
+#ifdef ENABLE_ASSERT
+  // Register SYNC_POINT command (only available with ENABLE_ASSERT)
+  int rc = RedisModule_CreateSubcommand(debugCommand, "SYNC_POINT", syncPoint,
```
meiravgri: Consider creating a new array of commands for commnds available only with ENABLE_ASSERT for future enhancemnts

src/debug_commands.h
```diff
@@ -91,7 +92,51 @@ bool StoreResultsDebugCtx_IsPauseAfterEnabled(void);
...
+
+// SyncPoint API function declarations
+// Arm a sync point - subsequent calls to SyncPoint_Wait will block
+// Returns true on success, false if max sync points reached
+bool SyncPoint_Arm(const char *name);
```
meiravgri, line 109: Mention `SyncPoint_Arm` thread-safty (not thread safe :))

src/debug_commands.c
```diff
@@ -119,6 +119,85 @@ bool StoreResultsDebugCtx_IsPaused(void) {
...
+  for (uint32_t i = 0; i < count; i++) {
+    atomic_store(&globalSyncPointCtx.points[i].armed, false);
+    atomic_store(&globalSyncPointCtx.points[i].waiting, false);
+  }
+  atomic_store(&globalSyncPointCtx.count, 0);
+}
+
+void SyncPoint_Wait(const char *name) {
+  SyncPointState *sp = SyncPoint_FindByName(name);
+  if (!sp || !atomic_load(&sp->armed)) return;
```
meiravgri, line 230: Is there a valid scenrio to call `SyncPoint_Wait` on an invalid/un-armed point? if not we might want to return an error or assert.

src/debug_commands.c
```diff
@@ -119,6 +119,85 @@ bool StoreResultsDebugCtx_IsPaused(void) {
...
+
+  atomic_store(&sp->waiting, true);
+  while (atomic_load(&sp->armed)) {
+    usleep(1000);  // Spin-wait with 1ms sleep (matches existing pattern)
+  }
+  atomic_store(&sp->waiting, false);
```
meiravgri: So first wait turns off `waiting` for all other waiting points?
Consider changing `waiting` to a counter. Last one turns the light off

tests/pytests/test_debug_commands.py
```diff
@@ -1688,3 +1688,157 @@ def test_dump_deleted_ids(env):
...
+
+
+@skip(cluster=True)
+@require_enable_assert
+def test_sync_point_before_first_read_blocks_and_resumes(env):
+    """Verify that BeforeFirstRead blocks query execution until explicitly signaled."""
+    env.expect('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT').ok()
+    env.expect('HSET', 'doc1', 't', 'hello').equal(1)
+    waitForIndex(env, 'idx')
+    set_workers(env, 1)
```
meiravgri: consider testing more than one query in parallel for a valuable test

meiravgri (commented):
Nice!!

src/aggregate/aggregate_exec.c
```diff
@@ -1449,6 +1457,12 @@ static int buildPipelineAndExecute(AREQ *r, RedisModuleCtx *ctx, QueryError *sta
       RedisSearchCtx_UnlockSpec(sctx);
     }
 
+#ifdef ENABLE_ASSERT
+    // Sync point: pause after acquiring the spec lock
+    // The snapshot is now established
+    SyncPoint_Wait(SYNC_POINT_AFTER_ITERATOR_CREATE);
```
JoanFM: Just to be sure, the 2 `SYNC_POINT_AFTER_ITERATOR_CREATE` cannot be happening in the same code path?

src/debug_commands.c
JoanFM: Yes since it is a debug command, it may help not failing silently to avoid false positives

src/debug_commands.c
raz-mon: WDYM "array" of commands?
Not under `_FT.DEBUG` or a new sub-command of it

raz-mon: Oh just the actual array..
Why would that help future enhancements?

src/debug_commands.c
raz-mon: Sure there is - you call wait on sync-points in the code (part of the compiled artifact), but arm and check other sync-points according to what you want to test.

src/debug_commands.c
raz-mon: Good point

src/debug_commands.h
raz-mon: Mentioned here:
```
  // Reserve a slot atomically. We use a simple counter since ARM is only called
  // from the main thread (via FT.DEBUG command), so no concurrent ARMs occur.
```
Will emphasize in the header

tests/pytests/test_debug_commands.py
raz-mon: Will give that a shot, should be fine with the counting addition.
None of this is required for current tests but will be good to cover it, will give it a go

src/debug_commands.c
raz-mon: But I don't want to get too much into the multi workers case here, since it requires a heavier infrastructure - e.g., granularity of which worker is waiting on a sync-point.
For now I'll add some of this infra, and as we need more we can add such granularity of API.

src/aggregate/aggregate_exec.c
raz-mon: The main-thread one causes a dead-lock, added by accident. Removed 🙏

src/debug_commands.c
meiravgri: I meant to add an additional array of a section in `commands` array protected by `ENABLE_ASSERT`, so we can iterate an array instead of hard-coding any new command we want to introduce under the macro protection.

raz-mon: Gotcha, will add.

PR #8685, opened 2026-03-12
[MOD-14319] RLookup - Clone name in RLookupRow::write_key_by_name if required 

memark:
This solves about 20 failing pytests in coordinator mode.

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches FFI boundary and `RLookupRow::write_key_by_name` key-creation path; incorrect ownership could lead to use-after-free or leaks, but change is small and localized.
> 
> **Overview**
> Fixes `RLookupRow_WriteByName`/`RLookupRow_WriteByNameOwned` behavior when writing to a previously-unknown field name by ensuring newly-created keys use an **owned** copy of `name` (via `name.into_owned()`), rather than potentially storing a borrowed pointer.
> 
> Clarifies FFI/header docs that ownership of the `name` C string remains with the caller and the implementation will copy it if needed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f612e22d75d137510969d2e40a41ae237738c7d2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
+++ b/src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs
@@ -147,6 +147,8 @@ pub unsafe extern "C-unwind" fn RLookupRow_MoveFieldsFrom(
 /// for which it is not necessary to use the boilerplate of getting an explicit
 /// key.
 ///
+/// Ownership of `name` remains with the caller, this function will make a copy if required.
+///
 /// Like [`RLookupRow_WriteByNameOwned`], but increases the refcount.
 ///
 /// # Safety
@@ -200,6 +202,8 @@ pub unsafe extern "C" fn RLookupRow_WriteByName<'a>(
 /// Write a value by-name to the lookup table. This is useful for 'dynamic' keys
 /// for which it is not necessary to use the boilerplate of getting an explicit
 /// key.
+///
+/// Ownership of `name` remains with the caller, this function will make a copy if required.
 ///
 /// Like [`RLookupRow_WriteByName`], but does not affect the refcount.
 ///
--- a/src/redisearch_rs/headers/rlookup_rs.h
+++ b/src/redisearch_rs/headers/rlookup_rs.h
@@ -738,6 +738,8 @@ void RLookupRow_MoveFieldsFrom(const struct RLookup *lookup,
  * for which it is not necessary to use the boilerplate of getting an explicit
  * key.
  *
+ * Ownership of `name` remains with the caller, this function will make a copy if required.
+ *
  * Like [`RLookupRow_WriteByNameOwned`], but increases the refcount.
  *
  * # Safety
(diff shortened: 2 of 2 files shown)


memark:
> This should not unconditionally copy the name, both because this is not actually what the C version does AND because its going to be a guaranteed perf hit.
> 
> instead this should allocate call `to_owned` here
> 
> https://github.com/RediSearch/RediSearch/blob/ab64e94d674c67f5b910aeeee167e818d41308fc/src/redisearch_rs/rlookup/src/row.rs#L249
> 
> to replicate the `RLOOKUP_F_NAMEALLOC` behavior that's present in the C version

you mean `into_owned`, right? `to_owned` only clones the `Cow`.

daxhuiberts:
Shouldn't the fix be in the `get_key_write` function?
https://github.com/RediSearch/RediSearch/blob/ab64e94d674c67f5b910aeeee167e818d41308fc/src/redisearch_rs/rlookup/src/lookup.rs#L292
It looks like this should not call `clone()`, but `to_owned()`.

Ownership is very hard to track with all those Cows IMO

JonasKruckenberg:
> > This should not unconditionally copy the name, both because this is not actually what the C version does AND because its going to be a guaranteed perf hit.
> > instead this should allocate call `to_owned` here
> > https://github.com/RediSearch/RediSearch/blob/ab64e94d674c67f5b910aeeee167e818d41308fc/src/redisearch_rs/rlookup/src/row.rs#L249
> > 
> > to replicate the `RLOOKUP_F_NAMEALLOC` behavior that's present in the C version
> 
> you mean `into_owned`, right? `to_owned` only clones the `Cow`.

yes, sorry. cow methods are not named well

JonasKruckenberg:
great, if this fixes the issue too and CI is green we're good. I would just also add a short mention to the FFI functions stating that ownership of the string remains with the caller and the method will make a copy if required

memark:
> great, if this fixes the issue too and CI is green we're good. I would just also add a short mention to the FFI functions stating that ownership of the string remains with the caller and the method will make a copy if required

We'll see if it passes (it does locally). Added the docs.

memark:
> Can you elaborate on the the problem this fixes?

This was my investigation

>   Root cause: use-after-free on the key's name pointer
> 
>   The flow in cluster mode:
> 
>   1. rpnet.c:626 — field is a pointer into an MRReply buffer (from MRReply_String)
>   2. rpnet.c:629 — calls RLookupRow_WriteByNameOwned(nc->lookup, field, len, ...)
>   3. In Rust (row.rs:232-238) — field is wrapped as CStr::from_bytes_with_nul — this is borrowed, just wrapping the raw MRReply pointer, no
>   copy
>   4. write_key_by_name → get_key_write(name, ...) → RLookupKey::new(name.clone(), ...) — .clone() on a Cow::Borrowed still doesn't copy the
>   string data, key's name field becomes a raw pointer into the MRReply buffer
>   5. rpnetNext returns, result is queued
>   6. On the next call to rpnetNext, curIdx == len triggers processWarningsAndCleanup → MRReply_Free(nc->current.root) — the buffer holding
>   "tag" is freed
>   7. Later, when outputting the result, RLookupKey_GetName(kk) returns the now-dangling pointer → reads garbage bytes → b'\x10\x83\x11'
>   instead of b'tag'
> 
>   Why this specific test? doc:4 has only the tag field (no vector, no text). The shard returns a batch with exactly 1 result. After rpnetNext
>    yields that result, the very next call to rpnetNext frees the MRReply. The allocator then reuses that memory before the output happens,
>   making the corruption visible and deterministic. Other tests may also have this bug but the freed memory doesn't get reused before output
>   (non-deterministic corruption based on allocator behavior).
>

PR #8687, opened 2026-03-12
[MOD-13588] RsValue: NullStatic

daxhuiberts:
This PR implements static value support in `SharedRsValue` functionality required to support the existing C NullStatic implementation. This is build directly into `SharedRsValue` so that nested `SharedRsValue`s like `Ref`, `Array` and `Trio` seamlessly work. Non static values keep a similar functionality as an `Arc<RsValue>` as before using the appropriate `Arc` methods required for this.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches core `SharedRsValue` ownership/refcounting by switching to raw-pointer + manual `Arc` management and introduces a static sentinel path, which is memory-safety sensitive if any invariants are violated. Also adds a new C FFI entrypoint returning non-heap `Null`, which could break callers that mutate or mis-handle refcounts.
> 
> **Overview**
> Adds a new FFI constructor `RSValue_NullStatic` (and header declaration) that returns a shared *static* `Null` sentinel instead of heap-allocating, while keeping `RSValue_DecrRef` symmetry.
> 
> Reworks `SharedRsValue` to support both heap-allocated `Arc`-backed values and a static `Null` pointer, implementing custom `Clone`/`Drop`/`Deref`, updating dereferencing/compare/hash/debug logic to use the new `Deref` behavior, and adds integration tests covering refcounting, pointer equality, and `set_value` panics for static/shared cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a7839bb0bac94fd0107184b31ada1e66b8d0ead3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/value/src/comparison.rs
+++ b/src/redisearch_rs/value/src/comparison.rs
@@ -45,11 +45,9 @@ pub fn compare(
     num_to_str_cmp_fallback: bool,
 ) -> Result<Ordering, CompareError> {
     match (v1, v2) {
-        (RsValue::Ref(r1), RsValue::Ref(r2)) => {
-            compare(r1.value(), r2.value(), num_to_str_cmp_fallback)
-        }
-        (RsValue::Ref(r1), _) => compare(r1.value(), v2, num_to_str_cmp_fallback),
-        (_, RsValue::Ref(r2)) => compare(v1, r2.value(), num_to_str_cmp_fallback),
+        (RsValue::Ref(r1), RsValue::Ref(r2)) => compare(r1, r2, num_to_str_cmp_fallback),
+        (RsValue::Ref(r1), _) => compare(r1, v2, num_to_str_cmp_fallback),
+        (_, RsValue::Ref(r2)) => compare(v1, r2, num_to_str_cmp_fallback),
         (RsValue::Null, RsValue::Null) => Ok(Ordering::Equal),
         (RsValue::Null, _) => Ok(Ordering::Less),
         (_, RsValue::Null) => Ok(Ordering::Greater),
@@ -60,14 +58,12 @@ pub fn compare(
         (RsValue::RedisString(rs1), RsValue::RedisString(rs2)) => {
             Ok(rs1.as_bytes().cmp(rs2.as_bytes()))
         }
-        (RsValue::Trio(t1), RsValue::Trio(t2)) => compare(
-            t1.left().value(),
-            t2.left().value(),
-            num_to_str_cmp_fallback,
-        ),
+        (RsValue::Trio(t1), RsValue::Trio(t2)) => {
+            compare(t1.left(), t2.left(), num_to_str_cmp_fallback)
+        }
         (RsValue::Array(a1), RsValue::Array(a2)) => {
             for (i1, i2) in a1.iter().zip(a2.deref()) {
-                let cmp = compare(i1.value(), i2.value(), num_to_str_cmp_fallback)?;
+                let cmp = compare(i1, i2, num_to_str_cmp_fallback)?;
                 if cmp != Ordering::Equal {
                     return Ok(cmp);
                 }


src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
```diff
@@ -230,3 +230,19 @@ pub extern "C" fn RSValue_NewNumberFromInt64(number: i64) -> *mut RsValue {
...
         .cast_mut()
 }
+
+/// Returns a pointer to the static [`RsValue::Null`] sentinel.
+///
+/// Unlike [`RSValue_NewNull`], this does **not** heap-allocate; it returns a
+/// pointer to a shared static value managed by [`SharedRsValue::null_static`].
+/// The returned pointer must still be passed to
+/// [`RSValue_DecrRef`](crate::shared::RSValue_DecrRef) for symmetry, but
+/// decrementing it is a no-op.
```
JonasKruckenberg: I assume incrementing is a no-op too then? I would just not mention this though. Just say "The returned pointer must still be passed to
/// [`RSValue_DecrRef`](crate::shared::RSValue_DecrRef) for symmetry" and that's great as-is

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,72 +7,142 @@
...
+///
+/// Internally, this is a raw pointer to an [`RsValue`] that is either:
+/// - **Static**: points to the global [`NULL_VALUE`] sentinel, requiring no
+///   reference counting.
+/// - **Heap-allocated**: backed by an [`Arc<RsValue>`](Arc), with manual
+///   reference counting managed through raw pointer conversions.
+///
+/// Manual [`Arc`] management via raw pointers is used so that this type can be
+/// passed across the FFI boundary as a plain pointer while still supporting
+/// shared ownership on the Rust side.
```
JonasKruckenberg: does not belong in a doc comment. you can add this as a regular comment after the doc comment to help future maintainers understand the design choice, but this is not a public facing information

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,72 +7,142 @@
...
+/// when the count reaches zero.
+#[expect(rustdoc::private_intra_doc_links)]
 pub struct SharedRsValue {
-    inner: Arc<RsValue>,
+    ptr: *const RsValue,
 }
 
+/// Global sentinel for [`RsValue::Null`], used by [`SharedRsValue::null_static`]
+/// to avoid heap allocation for null values.
+static NULL_VALUE: RsValue = RsValue::Null;
```
JonasKruckenberg: maybe I'm stupid but I don't think this technically is a sentinel value no?

JonasKruckenberg: doesn't really matter though

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,72 +7,142 @@
...
 impl SharedRsValue {
-    /// Create a new shared RsValue wrapping an [`RsValue`]
-    pub fn new(value: RsValue) -> Self {
+    /// Creates a [`SharedRsValue`] pointing to the static [`NULL_VALUE`]
+    /// sentinel, avoiding a heap allocation.
+    #[expect(rustdoc::private_intra_doc_links)]
+    pub fn null_static() -> Self {
         Self {
-            inner: Arc::new(value),
+            ptr: &NULL_VALUE as *const RsValue,
```
JonasKruckenberg: use `ptr::from_ref` and `.cast`

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,72 +7,142 @@
...
-    /// Only exactly one reference to the underlying [`RsValue`] must exist.
+    /// Replaces the stored [`RsValue`] in place.
     ///
-    /// # Panic
+    /// # Panics
     ///
-    /// Panics if more than one reference to the underlying [`RsValue`] exists.
+    /// - Panics if this is a static null value (created via [`Self::null_static`]).
+    /// - Panics if there are other outstanding clones sharing the same
+    ///   allocation (i.e. the [`Arc`] strong count is greater than 1).
```
JonasKruckenberg, line 91: is this behavior required by something? IMO it would make much more sense (and be less prone to panic-crashes) if `SharedRsValue` had an [`Arc::make_mut`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.make_mut) (and optionally [`Arc::get_mut`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut)to retain the no-clone perf guarantee with an explicit panic).

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,72 +7,142 @@
...
+    /// Returns a reference to the stored [`RsValue`].
+    pub fn value(&self) -> &RsValue {
+        if self.is_static() {
+            &NULL_VALUE
+        } else {
+            // SAFETY: `self.ptr` was obtained from `Arc::into_raw` and the
+            // `Arc` is kept alive for the lifetime of `self`.
+            unsafe { &*self.ptr }
+        }
+    }
```
JonasKruckenberg: this should be a `Deref` impl instead

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
daxhuiberts: Good one. Yep, incr and decr are both noops. Will remove the mention of this.

src/redisearch_rs/value/src/shared.rs
daxhuiberts: sentinel is not the right term, will remove it.

src/redisearch_rs/value/src/shared.rs
daxhuiberts: This behavior is not required, but mutating a an Arc without cloning it is. Normally you would wrap it in a mutex and allow mutating it, but that gives quite some overhead and is not needed because mutation always happen before the Arc is shared.

The C version does something similar to `get_mut_unchecked`. Using `get_mut` ensures we don't introduce UB with the addition that it can panic. But in practice and covered by the test suite it never hits that path. The panic still has to be documented.

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,72 +7,142 @@
...
         }
     }
 
-    /// Get a reference to the inner [`RsValue`].
-    pub fn value(&self) -> &RsValue {
-        &self.inner
+    /// Convenience wrapper around [`fully_dereferenced`](Self::fully_dereferenced)
+    /// that returns the inner [`RsValue`] directly.
+    pub fn fully_dereferenced_value(&self) -> &RsValue {
+        self.fully_dereferenced().value()
```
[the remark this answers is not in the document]
daxhuiberts: Removed

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
```diff
@@ -230,3 +230,18 @@ pub extern "C" fn RSValue_NewNumberFromInt64(number: i64) -> *mut RsValue {
...
+/// Returns a pointer to the static [`RsValue::Null`] sentinel.
+///
+/// Unlike [`RSValue_NewNull`], this does **not** heap-allocate; it returns a
+/// pointer to a shared static value managed by [`SharedRsValue::null_static`].
+/// The returned pointer must still be passed to
+/// [`RSValue_DecrRef`](crate::shared::RSValue_DecrRef) for symmetry.
+///
+/// # Safety
+///
+/// The returned pointer must not be mutated.
```
memark, line 243: This feels weird. Why not return `*const` instead? I feel the `*mut` pretty much _invites_ the caller to violate the safety.

src/redisearch_rs/value/src/lib.rs
```diff
@@ -56,17 +56,16 @@ pub enum RsValue {
 
 impl RsValue {
     pub fn fully_dereferenced_ref(&self) -> &Self {
-        if let RsValue::Ref(ref_value) = self {
-            ref_value.value().fully_dereferenced_ref()
-        } else {
-            self
+        match self {
```
memark, line 62: Much cleaner! 👍🏽

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,77 +7,169 @@
...
+
 impl SharedRsValue {
-    /// Create a new shared RsValue wrapping an [`RsValue`]
-    pub fn new(value: RsValue) -> Self {
+    /// Creates a [`SharedRsValue`] pointing to the static [`NULL_VALUE`].
+    #[expect(rustdoc::private_intra_doc_links)]
+    pub fn null_static() -> Self {
         Self {
-            inner: Arc::new(value),
+            ptr: std::ptr::from_ref(&NULL_VALUE).cast(),
```
memark: Consider `use std::ptr;`

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,77 +7,169 @@
...
+            ptr: Arc::into_raw(Arc::new(value)),
+        }
     }
 
     /// Convert a [`SharedRsValue`] into a raw `*const RsValue` pointer.
-    pub fn into_raw(self) -> *const RsValue {
-        Arc::into_raw(self.inner)
+    pub const fn into_raw(self) -> *const RsValue {
+        let ptr = self.ptr;
+        std::mem::forget(self); // Prevent Drop from running
```
memark: I think the comment is obvious (that's what `forget` _does_).
If anything, we might instead want to say _why_ `drop` should not run.

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,77 +7,169 @@
...
+        ptr
+    }
+
+    /// Returns the underlying raw pointer without consuming `self`.
+    pub const fn as_ptr(&self) -> *const RsValue {
+        self.ptr
     }
 
     /// Convert a `*const RsValue` back into a [`SharedRsValue`].
```
memark:
```suggestion
    /// Convert a `*const RsValue` into a [`SharedRsValue`].
```

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,77 +7,169 @@
...
+    ///   allocation (i.e. the [`Arc`] strong count is greater than 1).
     pub fn set_value(&mut self, new_value: RsValue) {
-        let value =
-            Arc::get_mut(&mut self.inner).expect("Failed to get mutable reference to inner value");
-
+        if self.is_static() {
+            panic!("Cannot change the value of static NULL");
+        }
+        // SAFETY: `self.ptr` was obtained from `Arc::into_raw` and is not
+        // static (checked above). `ManuallyDrop` prevents the `Arc` from being
```
memark: Good safety comment. The second part, about `ManuallyDrop`, is a good explanation, but is not about safety, and could be separate. Same thing applies to other functions further down.

src/redisearch_rs/c_entrypoint/value_ffi/src/constructors.rs
daxhuiberts: The `NullStatic` is used in all kinds of places where a normal `*RsValue` is also used. Changing this to a `*const` would break the C code in a lot of places.

Ideally I would remove the `NullStatic` altogether, but that is a lot more difficult than initially thought. So we kept the existing `NullStatic` setup.

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,77 +7,172 @@
...
-    pub fn into_raw(self) -> *const RsValue {
-        Arc::into_raw(self.inner)
+    pub const fn into_raw(self) -> *const RsValue {
+        let ptr = self.ptr;
+        // The original [`SharedRsValue`] is forgotten to avoid decrementing it.
+        mem::forget(self);
+        ptr
+    }
+    /// Returns the underlying raw pointer without consuming `self`.
+
```
memark:
```suggestion

    /// Returns the underlying raw pointer without consuming `self`.
```

raz-mon:
@daxhuiberts Please update the PR description

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,77 +7,172 @@
...
+        Self { ptr }
     }
 
-    /// Get a reference to the inner [`RsValue`].
-    pub fn value(&self) -> &RsValue {
-        &self.inner
+    /// Returns `true` if this value points to the static [`NULL_VALUE`]
+    /// rather than a heap-allocated [`Arc`].
+    fn is_static(&self) -> bool {
+        ptr::eq(self.ptr, &NULL_VALUE)
```
raz-mon, line 82: Perhaps the docs should be more generic, to follow the function name?
E.g., 
```
Returns `true` if this value points to a static value rather than a heap-allocated [`Arc`]
```

src/redisearch_rs/value/src/shared.rs
```diff
@@ -7,77 +7,172 @@
...
+    /// - Panics if this is a static null value (created via [`Self::null_static`]).
+    /// - Panics if there are other outstanding clones sharing the same
+    ///   allocation (i.e. the [`Arc`] strong count is greater than 1).
     pub fn set_value(&mut self, new_value: RsValue) {
-        let value =
-            Arc::get_mut(&mut self.inner).expect("Failed to get mutable reference to inner value");
-
+        if self.is_static() {
+            panic!("Cannot change the value of static NULL");
+        }
```
raz-mon: Why is this not a valid use-case?
E.g., starting from NullStatic and setting to some value

src/redisearch_rs/value/tests/integration/shared.rs
```diff
@@ -0,0 +1,165 @@
...
+fn null_static_holds_null() {
+    let v = SharedRsValue::null_static();
+    assert!(matches!(*v, RsValue::Null));
+}
+
+#[test]
+fn new_number() {
+    let v = SharedRsValue::new(RsValue::Number(42.0));
+    assert!(matches!(*v, RsValue::Number(42.0)));
+}
```
raz-mon, line 24: How is this test related to this PR?

src/redisearch_rs/value/src/shared.rs
daxhuiberts: You surface a good point! Though conceptually more static values can be supported, the focus is really on the single NullStatic.

I'll rename the function (and associated code) to say `is_null_static` instead and focus only on this specific use case.

src/redisearch_rs/value/src/shared.rs
daxhuiberts: The `NullStatic` is a fixed value. Other code may also use the same `NullStatic` and if the value underneath is changed, that will create havoc.

Existing C code already uses a `Ref(NullStatic)` and update the `Ref` to a different type/value if needed.

Ideally I want to phase out the `NullStatic` altogether, but that is a lot more difficult than initially thought. So we kept the existing `NullStatic` setup at this stage.

src/redisearch_rs/value/tests/integration/shared.rs
daxhuiberts: I combines well with the previous test in that deref (what the `*` is doing) works for both the `NullStatic` value as well as regular `Arc` based `RsValue` values.

src/redisearch_rs/value/src/shared.rs
raz-mon: I see.
Yes it's interesting if the boost it gives is worth the cumbersome flows.
So this is the current behavior as well? If so that's fine, we can move forward with it.

daxhuiberts: Yes, this whole `NullStatic` is to support current behaviour from the C implementation.

src/redisearch_rs/value/tests/integration/shared.rs
raz-mon: Was there a change related to it (deref) in this PR?
If not perhaps separate to another test-enhancement PR, just so we maintain relevance here.

daxhuiberts: Deref has been added in this PR. Not sure what separating it would do.

PR #8660, opened 2026-03-11
[8.6] MOD-14065: Fix property order discrepancy in `FILTER`

raz-mon:
## Describe the changes in the pull request

Backport of #8548 to 8.6 - fixes property order discrepancy in `FILTER`.

1. **Current**: During indexing, if a `FILTER` expression references properties in a certain order, documents may be incorrectly excluded when properties appear in a different order than in the filter expression.
2. **Change**: Introduces `EVAL_MODE_INDEX` and `EVAL_MODE_QUERY` to distinguish between indexing and query-time evaluation. In `EVAL_MODE_INDEX`, missing properties no longer cause errors but return undefined values.
3. **Outcome**: Consistent behavior regardless of property order in filter expressions during indexing.

#### Which additional issues this PR fixes
1. MOD-14332
2. Includes fix from #8627: Added `waitForIndex` calls to prevent test flakiness

#### Main objects this PR modified
1. `src/aggregate/expr/expression.c` - Core evaluation logic with eval mode distinction
2. `tests/pytests/test_filter.py` - Added regression tests
3. `tests/cpptests/test_cpp_expr.cpp` - Added C++ unit tests

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

#### Release Notes

- [x] This PR requires release notes
- [ ] This PR does not require release notes

### 8.6-specific adaptations
- Kept `QueryError_Default()` removal in `EvalCtx_Create` (8.6 differs slightly from master)
- Used 8.6 error codes where needed

---

> [!NOTE]
> **Medium Risk**
> Changes core expression-evaluation semantics for missing fields by introducing mode-dependent behavior, which can affect which documents are indexed under `FILTER` rules. Covered by extensive new C++ and Python regression tests, but still touches critical indexing logic.
> 
> **Overview**
> Fixes a long-standing inconsistency where indexing-time `FILTER` expressions could yield different results depending on property order when some fields are missing.
> 
> This introduces `EVAL_MODE_INDEX` vs `EVAL_MODE_QUERY` in the expression evaluator so **indexing treats missing properties as non-matching (false) without raising errors**, while query-time evaluation remains strict and still errors on missing properties where expected.
> 
> Updates rule/spec filter evaluation to use index mode, keeps `FT.ADD ... IF` evaluation in query mode, and adds regression coverage for missing-field AND/OR and comparison/type-mismatch cases (C++ unit tests + Python integration tests).
> 
[...]

PR #8600, opened 2026-03-04
[8.2] MOD-14063: Fix `FILTER` bug with multiple indexes with the same field aliases

raz-mon:
## Summary
Backport of commit 9c70952e8 from master to 8.2 branch.

This PR fixes a bug where multiple indexes with the same field aliases could cause incorrect filter evaluation results.

Related Jira: [MOD-14339](https://redislabs.atlassian.net/browse/MOD-14339)
Original PR: #8396

## Conflicts Resolved

The following conflicts were encountered during the cherry-pick and resolved as follows:

### 1. `src/rlookup.c`
- **Conflict**: In `RLookup_Reset()`, the 8.2 branch uses `head`/`tail` naming convention for struct members and had `memset(lk, 0xff, sizeof(*lk))`, while the incoming commit uses `_head`/`_tail` and `memset(lk, 0, sizeof(*lk))`.
- **Resolution**: Kept the 8.2 naming convention (`lk->head`/`lk->tail`) but accepted the fix to use `memset(lk, 0, sizeof(*lk))` which properly resets the lookup structure.

### 2. `tests/pytests/test_followhashes.py`
- **Conflict**: The 8.2 branch already had additional tests (`testCountry`, `testIssue1571`, `testIssue1571WithRename`, `testIdxFieldJson`, `testFilterStartWith`, `testFilterWithOperator`, `testFilterWithNot`) that were backported in previous versions, but these were not present in the base of the cherry-picked commit.
- **Resolution**: Kept all existing tests from the 8.2 branch to preserve the test coverage.

### 3. `spec.c`
- **Conflict**: The commit used `RLookupRow_Reset`, while in this branch this API was called `RLookupRow_Cleanup`.
- **Resolution**: Used `RLookupRow_Cleanup` instead.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) with guidance from the PR author

[MOD-14339]: https://redislabs.atlassian.net/browse/MOD-14339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Touches indexing-time filter evaluation and `RLookup` lifecycle/cleanup; mistakes here could cause incorrect indexing or memory issues, but the change is scoped and covered by new tests.
> 
> **Overview**
[...]

PR #8699, opened 2026-03-16
Fix missing error return code in evalPredicate

raz-mon:
## Problem

When `getPredicateBoolean` encounters a type mismatch error (e.g., comparing a number to a string), it sets an error in `eval->err` but `evalPredicate` was returning `EXPR_EVAL_OK` instead of `EXPR_EVAL_ERR`.

## Root Cause

This is a regression introduced in the MOD-14065 refactoring (PR #8600).

**Original code:**
- `rc` was initialized to `EXPR_EVAL_ERR`
- It was only set to `EXPR_EVAL_OK` in the explicit success path
- If `getPredicateBoolean` set an error, `rc` remained `EXPR_EVAL_ERR`

**Refactored code (bug):**
- `rc` is overwritten by `evalPredicateOperand()` returning `EXPR_EVAL_OK`
- When `getPredicateBoolean` fails, the `else` branch clears the result but `rc` remains `EXPR_EVAL_OK`

## Fix

Add `rc = EXPR_EVAL_ERR;` in the else branch to properly propagate the error status.

## Impact

- **For FILTER during indexing**: No change in observable behavior (result is set to undefined, which evaluates to `false`, so documents are still excluded)
- **For query-time filters**: Properly returns error instead of silently skipping rows with type mismatches

## Related

- Regression from: #8600 (MOD-14065)
- Found during backport review: #8660, #8661, #8662

---

> [!NOTE]
> **Medium Risk**
> Changes expression predicate evaluation to return `EXPR_EVAL_ERR` when comparisons hit type-mismatch errors, which can turn previously silent query-time filter behavior into hard errors. Scope is small and covered by new tests but affects filter evaluation semantics.
> 
> **Overview**
> Fixes a regression in `evalPredicate` where `getPredicateBoolean` could set a `QueryError` (e.g., number vs non-numeric string comparisons) but evaluation still returned `EXPR_EVAL_OK`; it now returns `EXPR_EVAL_ERR` and clears the result on error.
> 
> Adds a focused C++ test (`testPredicateTypeMismatchReturnsError`) asserting that mismatched comparison operators (`<`, `>`, `<=`, `>=`) propagate errors in both `EVAL_MODE_QUERY` and `EVAL_MODE_INDEX`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8df41a84fe9f8089bd8b00d73df70eea6993f032. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/cpptests/test_cpp_expr.cpp
+++ b/tests/cpptests/test_cpp_expr.cpp
@@ -982,4 +982,115 @@ TEST_F(ExprTest, testOrExpressionWithMissingFields) {
   RLookup_Cleanup(&lk);
 }
 
+// Test that evalPredicate returns EXPR_EVAL_ERR when getPredicateBoolean encounters
+// a type mismatch error (e.g., comparing a number to a non-convertible string).
+TEST_F(ExprTest, testPredicateTypeMismatchReturnsError) {
+  // Setup: create a lookup with a string field that cannot be converted to a number
+  RLookup lk = RLookup_New();
+  auto *kstr = RLookup_GetKey_Write(&lk, "str", RLOOKUP_F_NOFLAGS);
+  auto *knum = RLookup_GetKey_Write(&lk, "num", RLOOKUP_F_NOFLAGS);
+  ASSERT_NE(kstr, nullptr);
+  ASSERT_NE(knum, nullptr);
+
+  RLookupRow row = RLookupRow_New();
+  // "hello" cannot be converted to a number
+  RLookup_WriteOwnKey(kstr, &row, RSValue_NewString(strdup("hello"), 5));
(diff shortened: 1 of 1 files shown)


src/aggregate/expr/expression.c
```diff
@@ -263,6 +263,7 @@ static int evalPredicate(ExprEval *eval, const RSPredicate *pred, RSValue *resul
     rc = EXPR_EVAL_OK;
   } else {
     RSValue_Clear(result);
+    rc = EXPR_EVAL_ERR;
```
oshadmi, line 266: @raz-mon Is covered by unit test?

raz-mon: Will add a test

PR #8708, opened 2026-03-16
MOD-14500: Solve TTL Crash When Many Keys Get Expiration Times

kei-nan:
Block TTL dict rehashing when taking the index lock

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches concurrency/locking around index read paths and TTL data structures; mistakes could cause deadlocks or performance regressions, but changes are scoped and covered by new tests.
> 
> **Overview**
> Fixes crashes/races around the TTL (expiration) tracking dictionary by **pausing/resuming rehashing for the TTL table** wherever index read locks are taken (query context, LLAPI iterators, and `IndexInfo`).
> 
> Adds a new runtime config `search-monitor-expiration` (default `yes`) that controls whether indexes track/filter expired documents/fields; toggling it updates all existing indexes, and disabling it clears per-index TTL data via new `DocTable_ClearExpirationData`.
> 
> Strengthens `dictPauseRehashing`/`dictResumeRehashing` atomic memory ordering and adds C++/Python tests covering rehash pausing behavior and the new config’s runtime semantics.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 24752223e1b59d1d467e342288ef4d62436e6215. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_api.c
+++ b/src/redisearch_api.c
@@ -345,12 +345,14 @@ void RediSearch_AddDocDone(RSAddDocumentCtx* aCtx, RedisModuleCtx* ctx, void* er
 int RediSearch_IndexAddDocument(RefManager* rm, Document* d, int options, char** errs) {
   RWLOCK_ACQUIRE_WRITE();
   IndexSpec* sp = __RefManager_Get_Object(rm);
+  pthread_rwlock_wrlock(&sp->rwlock);
 
   RSError err = {.s = errs};
   QueryError status = QueryError_Default();
   RSAddDocumentCtx* aCtx = NewAddDocumentCtx(sp, d, &status);
   if (aCtx == NULL) {
     QueryError_ClearError(&status);
+    pthread_rwlock_unlock(&sp->rwlock);
     RWLOCK_RELEASE();
     return REDISMODULE_ERR;
   }
@@ -366,6 +368,7 @@ int RediSearch_IndexAddDocument(RefManager* rm, Document* d, int options, char**
         *errs = rm_strdup("Document already exists");
       }
       AddDocumentCtx_Free(aCtx);
+      pthread_rwlock_unlock(&sp->rwlock);
       RWLOCK_RELEASE();
       return REDISMODULE_ERR;
     }
@@ -376,6 +379,7 @@ int RediSearch_IndexAddDocument(RefManager* rm, Document* d, int options, char**
   QueryError_ClearError(&status);
   rm_free(d);
 
+  pthread_rwlock_unlock(&sp->rwlock);
   RWLOCK_RELEASE();
   return err.hasErr ? REDISMODULE_ERR : REDISMODULE_OK;
 }


src/util/dict/dict.h
```diff
@@ -155,8 +155,8 @@ typedef void (dictScanBucketFunction)(void *privdata, dictEntry **bucketref);
 #define dictSlots(d) ((d)->ht[0].size+(d)->ht[1].size)
 #define dictSize(d) ((d)->ht[0].used+(d)->ht[1].used)
 #define dictIsRehashing(d) ((d)->rehashidx != -1)
-#define dictPauseRehashing(d) (__atomic_fetch_add(&(d)->pauserehash, 1, __ATOMIC_RELAXED) >= 0)
-#define dictResumeRehashing(d) (__atomic_fetch_add(&(d)->pauserehash, -1, __ATOMIC_RELAXED) > 0)
+#define dictPauseRehashing(d) (__atomic_fetch_add(&(d)->pauserehash, 1, __ATOMIC_ACQ_REL) >= 0)
+#define dictResumeRehashing(d) (__atomic_fetch_add(&(d)->pauserehash, -1, __ATOMIC_ACQ_REL) > 0)
```
GuyAv46: These could be relaxed a bit, no?

```suggestion
#define dictPauseRehashing(d) (__atomic_fetch_add(&(d)->pauserehash, 1, __ATOMIC_ACQUIRE) >= 0)
#define dictResumeRehashing(d) (__atomic_fetch_add(&(d)->pauserehash, -1, __ATOMIC_RELEASE) > 0)
```

kei-nan:
## Memory Ordering Rationale

This PR changes the atomic memory ordering for `dictPauseRehashing`/`dictResumeRehashing` from `__ATOMIC_RELAXED` to `__ATOMIC_ACQUIRE`/`__ATOMIC_RELEASE`.

### The Race Condition

With `__ATOMIC_RELAXED`, the following race was theoretically possible:

1. **Thread A** (query thread) calls `dictPauseRehashing(d)` which atomically increments `pauserehash` to 1
2. Due to relaxed ordering, this write may not be immediately visible to other cores
3. **Thread B** (another thread) reads `pauserehash` and still sees 0 (stale cache)
4. **Thread B** proceeds to call `_dictRehashStep()` because `pauserehash == 0`
5. **Thread A** is now iterating over the dict while **Thread B** is rehashing it → data corruption

### The Fix

- `dictPauseRehashing` now uses `__ATOMIC_ACQUIRE` - ensures that subsequent reads/writes by this thread happen *after* the pause is visible
- `dictResumeRehashing` now uses `__ATOMIC_RELEASE` - ensures that all previous reads/writes are visible *before* the resume

This creates a proper synchronization point: any thread checking `pauserehash` will see a consistent view of the pause state.

### Note on x86

This race is very hard to reproduce on x86 due to its strong memory model (TSO - Total Store Order). However, on ARM or other weakly-ordered architectures, this could manifest as real corruption. The fix ensures correctness across all architectures.

tests/cpptests/test_cpp_dict_pause_rehash.cpp
```diff
@@ -19,11 +19,12 @@ extern "C" {
 #include <vector>
 
 class DictPauseRehashTest : public ::testing::Test {
-protected:
+  // Hash function required by dictType - signature is dictated by C API
   static uint64_t hashFunc(const void *key) {
-    return (uint64_t)(uintptr_t)key;
+    return reinterpret_cast<uint64_t>(reinterpret_cast<uintptr_t>(key));
```
GuyAv46: Should be `static_cast` (the second cast).

src/spec.c
```diff
@@ -2295,8 +2295,9 @@ static void initializeIndexSpec(IndexSpec *sp, const HiddenString *name, IndexFl
 
   sp->scanner = NULL;
   sp->scan_in_progress = false;
-  sp->monitorDocumentExpiration = true;
-  sp->monitorFieldExpiration = RedisModule_HashFieldMinExpire != NULL;
+  sp->monitorDocumentExpiration = RSGlobalConfig.monitorExpiration;
+  sp->monitorFieldExpiration = RSGlobalConfig.monitorExpiration &&
+                               RedisModule_HashFieldMinExpire != NULL;
```
GuyAv46, line 2305: This is probably redundant now, we have minimal redis version validation

src/config.c
```diff
@@ -2305,5 +2305,14 @@ int RegisterModuleConfig_Local(RedisModuleCtx *ctx) {
     )
   )
 
+  RM_TRY(
+    RedisModule_RegisterBoolConfig(
```
GuyAv46, line 2357: Is this a static config? Once an index is created, we won't update its monitoring config.

kei-nan: we have the debug command for that if we need to
problem was we couldn't set it before the index was created which was a problem when for example wanting to load an rdb file.

src/spec.c
kei-nan: this is going to get backported to previous versions, maybe simpler to keep as is for now.

src/config.c
```diff
@@ -2305,5 +2305,14 @@ int RegisterModuleConfig_Local(RedisModuleCtx *ctx) {
...
+      ctx, "search-monitor-expiration", 1,
```
nafraf, line 2358: Consider adding the parameter description to `module.conf`

src/config.c
```diff
@@ -2305,5 +2305,14 @@ int RegisterModuleConfig_Local(RedisModuleCtx *ctx) {
...
+      REDISMODULE_CONFIG_IMMUTABLE | REDISMODULE_CONFIG_UNPREFIXED,
```
nafraf: Why `REDISMODULE_CONFIG_IMMUTABLE` ?

kei-nan: Will change it to be runtime
Thought we would need to control the config before any index was created
I guess we can cleanup the ttl dict for every index when config gets disabled.

src/redisearch_api.c
```diff
@@ -865,9 +889,15 @@ int RediSearch_IndexInfo(RSIndex* rm, RSIdxInfo *info) {
 
   RWLOCK_ACQUIRE_READ();
   IndexSpec *sp = __RefManager_Get_Object(rm);
+  /* Acquire spec read lock to synchronize with config changes that may destroy TTL table */
+  pthread_rwlock_rdlock(&sp->rwlock);
```
oshadmi, line 901: @kei-nan If we change the config setter (set_monitor_expiration) to avoid cleanup to ttl data, and just ignore ttl data when the config is enabled, we would not need to have additional `pthread_rwlock_rdlock` and `pthread_rwlock_unlock`?
So ttl data will remain in memory but will not be considered - and no lock penalties would be required.
WDYT?

kei-nan: I think the locking in the api needs this, not sure how we sync these days for keysDict for example.
This change is a result of a cursor comment

oshadmi: @kei-nan Actually already now, if config is disabled and later on re-enabled, no TTL data would exist, including keys expiration data during the disabled period, so queries would have missing TTL data and correctness would be impacted.
We should consider keeping writing the TTL data, and just ignore it if the config is disabled, so re-enabling the config would maintain correctness. This will not free TTL data, but will avoid additional `pthread_rwlock_rdlock` in `handleIterCommon` and `RediSearch_IndexInfo`.
WDYT?

oshadmi: @kei-nan Alternatively If looking for a capability full kill switch, we can choose to avoid writing TTL data when disabled, and ignore existing TTL data when disabled, and still avoid taking extra locks.
However when re-enabled, TTL data from the disabled period would be missing, but this is a compromise we can consider (after incident is over, a new db can be used with enabled expirations monitoring)

One more option: Should we consider dropping the config entirely and relying solely on the `dictPauseRehashing` fix?

==> I would go with having a config for avoiding TTL writes + reads when disabled and avoid extra locking.

oshadmi: @kei-nan If this code in `handleIterCommon` is not affecting any hot paths such as in our queries then we can proceed with what we have now.

kei-nan: looks like this is only used for testing so wouldn't consider this a hot path.
could trace the calls to:
tests/cpptests/test_cpp_llapi.cpp, tests/cpptests/common.cpp

src/util/dict/dict.h
```diff
@@ -155,8 +155,8 @@ typedef void (dictScanBucketFunction)(void *privdata, dictEntry **bucketref);
 #define dictSlots(d) ((d)->ht[0].size+(d)->ht[1].size)
 #define dictSize(d) ((d)->ht[0].used+(d)->ht[1].used)
 #define dictIsRehashing(d) ((d)->rehashidx != -1)
-#define dictPauseRehashing(d) (__atomic_fetch_add(&(d)->pauserehash, 1, __ATOMIC_RELAXED) >= 0)
-#define dictResumeRehashing(d) (__atomic_fetch_add(&(d)->pauserehash, -1, __ATOMIC_RELAXED) > 0)
+#define dictPauseRehashing(d) (__atomic_fetch_add(&(d)->pauserehash, 1, __ATOMIC_ACQUIRE) >= 0)
```
JoanFM, line 176: why is this ACQUIRE and not RELEASE?

src/doc_table.c
```diff
@@ -243,6 +243,22 @@ bool DocTable_IsDocExpired(DocTable* t, const RSDocumentMetadata* dmd, struct ti
   return TimeToLiveTable_HasDocExpired(t->ttl, dmd->id, expirationPoint);
 }
 
+void DocTable_DisableExpirationMonitoring(DocTable *t) {
```
JoanFM: Here it seems u are disabling Expiration completely, not its monitoring?

src/doc_table.h
```diff
@@ -131,6 +131,11 @@ void DocTable_UpdateExpiration(DocTable *t, RSDocumentMetadata* dmd, t_expiratio
 
 bool DocTable_IsDocExpired(DocTable* t, const RSDocumentMetadata* dmd, struct timespec* expirationPoint);
 
+// Disable expiration monitoring for this doc table.
+// Clears expiration flags from all documents and destroys the TTL table.
+// Must be called with the index write lock held.
+void DocTable_DisableExpirationMonitoring(DocTable *t);
```
JoanFM: is this correct naming?

src/redisearch_api.c
```diff
@@ -596,9 +596,15 @@ typedef struct {
 static RS_ApiIter* handleIterCommon(IndexSpec* sp, QueryInput* input, char** error) {
   // here we only take the read lock and we will free it when the iterator will be freed
   RWLOCK_ACQUIRE_READ();
```
JoanFM, line 606: what does this lock mean?

src/redisearch_api.c
```diff
@@ -715,8 +731,16 @@ void RediSearch_ResultsIteratorFree(RS_ApiIter* iter) {
...
+  if (iter->sp) {
+    if (iter->sp->keysDict) {
+      dictResumeRehashing(iter->sp->keysDict);
+    }
+    /* Also resume rehashing on the TTL table if it exists */
+    if (iter->sp->docs.ttl) {
+      dictResumeRehashing(iter->sp->docs.ttl);
+    }
+    /* Release spec read lock */
+    pthread_rwlock_unlock(&iter->sp->rwlock);
```
JoanFM, line 751: this unlock is only when (iter->sp)? But the lock is not conditioned, Isn't there a risk of keeping it locked? I would keep it outside the if (or maybe this should be asserted?)

JoanFM: I would keep it after RWLOCK_RELEASE

JoanFM: or before actually to avoid potential deadlocks.

src/util/dict/dict.h
kei-nan: ACQUIRE is correct here. The pattern is:
- `dictPauseRehashing` uses ACQUIRE to ensure subsequent dict reads happen AFTER the pause is visible (pairs with the ACQUIRE load in `_dictRehashStep`)
- `dictResumeRehashing` uses RELEASE to ensure all prior dict reads complete BEFORE the resume is visible

Added a comment in dict.h explaining this.

src/doc_table.c
kei-nan: Good catch - renamed to `DocTable_ClearExpirationData` which better reflects what it does (clears flags + destroys TTL table).

src/doc_table.h
kei-nan: Renamed to `DocTable_ClearExpirationData`.

src/redisearch_api.c
kei-nan: Added a comment explaining the two-level locking:
1. `RWLOCK` (global) - protects access to all indexes, prevents index destruction
2. `sp->rwlock` (per-spec) - protects this specific index's data structures (e.g., TTL table)

Both are acquired here and released in `RediSearch_ResultsIteratorFree`.

src/redisearch_api.c
kei-nan: The code is correct - the conditional is intentional. On error paths in `handleIterCommon`, if `iter->sp` is NULL, the locks are already released there before calling `RediSearch_ResultsIteratorFree`. Added a comment explaining this:

```c
/* Release locks only if iter->sp is set. On error paths in handleIterCommon,
 * if iter->sp is NULL, locks were already released there before calling this.
 * Lock release order: spec lock first, then global lock (reverse of acquisition). */
```

kei-nan: spec lock is released before `RWLOCK_RELEASE()` (reverse order of acquisition). The unlock is inside `if (iter->sp)` because on some error paths the locks were already released in `handleIterCommon`.

JoanFM: Ok, I guess is hard to do better.

src/util/dict/dict.h
```diff
@@ -155,8 +155,13 @@ typedef void (dictScanBucketFunction)(void *privdata, dictEntry **bucketref);
...
 #define dictSize(d) ((d)->ht[0].used+(d)->ht[1].used)
 #define dictIsRehashing(d) ((d)->rehashidx != -1)
-#define dictPauseRehashing(d) (__atomic_fetch_add(&(d)->pauserehash, 1, __ATOMIC_RELAXED) >= 0)
-#define dictResumeRehashing(d) (__atomic_fetch_add(&(d)->pauserehash, -1, __ATOMIC_RELAXED) > 0)
+/* Memory ordering rationale:
+ * - dictPauseRehashing uses ACQUIRE to ensure subsequent dict reads happen AFTER
+ *   the pause is visible to other threads (pairs with ACQUIRE in _dictRehashStep).
+ * - dictResumeRehashing uses RELEASE to ensure all prior dict reads complete BEFORE
+ *   the resume is visible (allows _dictRehashStep to proceed safely). */
+#define dictPauseRehashing(d) (__atomic_fetch_add(&(d)->pauserehash, 1, __ATOMIC_ACQUIRE) >= 0)
```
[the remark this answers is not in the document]
kei-nan: fixed the comment

src/redis_index.c
```diff
@@ -90,6 +90,11 @@ void RedisSearchCtx_LockSpecRead(RedisSearchCtx *ctx) {
   // pause rehashing while we're using the dict for reads only
   // Assert that the pause value before we pause is valid.
   RS_ASSERT_ALWAYS(dictPauseRehashing(ctx->spec->keysDict));
+  // Also pause rehashing on the TTL table if it exists, to avoid concurrent
+  // rehash steps from multiple query threads corrupting the dict during reads.
+  if (ctx->spec->docs.ttl) {
+    RS_ASSERT_ALWAYS(dictPauseRehashing(ctx->spec->docs.ttl));
+  }
```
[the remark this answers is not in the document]
kei-nan: In production `TimeToLiveTable_VerifyInit` can only be called from:
`DocTable_UpdateExpiration` which in turn called from: 
`doAssignIds`
The point is, the write lock will be held so the scenario you described cannot occur.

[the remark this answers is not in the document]
kei-nan: there is an issue in RediSearch_IndexAddDocument so I will take the write lock there.
but RediSearch_IndexAddDocument is used for tests

PR #8762, opened 2026-03-19
[MOD-14508] move profile wrapping logic to Rust

gdesmott:
- Port `Profile_AddIters` from C to Rust, using a recursive `CRQEIterator::into_profiled` approach that profiles an entire   
  iterator tree via vtable callbacks                                                                                           
  - Add `ProfileChildren` trait for Rust compound iterators (Intersection, Union, Not, Optional), preserving concrete type     
  parameters through profiling so C casts remain valid                                                                         
  - Split `RQEIteratorWrapper::boxed_new` into leaf (`boxed_new`) and compound (`boxed_new_compound`) variants, with the latter
   installing a `ProfileChildren` vtable callback                                                                              
  - Add `IteratorType::is_leaf` classifier to distinguish leaf from compound iterator types
  - Remove `ForEachIntersectionChildMut`, `TakeOptionalNonOptimizedIteratorChild`, and `SetOptionalNonOptimizedIteratorChild`
  FFI functions, now superseded by the trait-based profiling    

The next step would be to move the debug printing logic from `profile.c` to Rust but this PR is already big enough.

Probably easier to review commit by commit.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes iterator vtable shape and rewires profiling to traverse and wrap iterator trees via new C/Rust FFI callbacks, which could impact query execution profiling and iterator ownership if misused.
> 
> **Overview**
> Moves iterator profiling-tree wrapping from C (`Profile_AddIters` switch in `profile.c`) into Rust by introducing `CRQEIterator::into_profiled` and exporting `Profile_AddIters`/`IntoProfiled` via `iterators_rs.h`.
> 
> Extends the C iterator vtable with an optional `ProfileChildren` callback and implements it for composite C iterators (`Hybrid`, `Union`, `Not` optimized, `Optimus`, `Optional` optimized) so profiling can recursively wrap children without type-specific child take/set APIs.
> 
> Updates Rust iterator FFI wrapping to distinguish leaf vs compound iterators (`IteratorType::is_leaf`, `RQEIteratorWrapper::boxed_new_compound`, and a `ProfileChildren` trait) and migrates compound iterators (e.g. `Intersection`, `Union`, `Not`, `Optional`) to preserve concrete wrapper types during profiling; removes now-unneeded FFI mutation helpers (`ForEachIntersectionChildMut`, optional child take/set). Adds integration tests covering profiling behavior and double-profiling protection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9f2f36fe077d6808e89f28f488db8a6e1f279521. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_entrypoint/iterators_ffi/src/intersection.rs
+++ b/src/redisearch_rs/c_entrypoint/iterators_ffi/src/intersection.rs
@@ -111,8 +111,7 @@ pub unsafe extern "C" fn GetIntersectionIteratorNumChildren(header: *const Query
 /// Returns a non-owning raw pointer to the child at `idx`.
 ///
 /// The returned pointer is valid as long as the intersection iterator is alive and no
-/// structural modifications are made (e.g. via [`AddIntersectionIteratorChild`] or
-/// [`ForEachIntersectionChildMut`]).
+/// structural modifications are made (e.g. via [`AddIntersectionIteratorChild`]).
 ///
 /// # Safety
 ///
@@ -172,52 +171,4 @@ pub unsafe extern "C" fn AddIntersectionIteratorChild(
     #[expect(clippy::multiple_unsafe_ops_per_block)]
     let child = unsafe { CRQEIterator::new(NonNull::new_unchecked(child)) };
     wrapper.inner.push_child(child);
-}
-
-/// Apply `callback` to each child iterator slot, passing a mutable `QueryIterator**`.
-///
-/// This is designed for use with `Profile_AddIters`, which replaces each child with a
-/// profile-wrapping iterator in place. The callback receives a pointer to the raw pointer
-/// stored inside each [`CRQEIterator`] child, allowing it to update (replace) that pointer.
-///
-/// This is safe because [`CRQEIterator`] is `#[repr(transparent)]` over
-/// `NonNull<QueryIterator>`, which has the same memory layout as `*mut QueryIterator`.
-/// The callback's in-place mutation of the slot directly updates the `CRQEIterator`'s
-/// internal pointer, so the intersection will subsequently own (and free) the new iterator.
-///
-/// # Safety
-///
-/// 1. `header` must be a valid non-null pointer created via [`NewIntersectionIterator`].
-/// 2. `callback` must be a valid function pointer.
-/// 3. The callback must replace `*slot` with a valid non-null `QueryIterator*` that takes
-///    ownership of the original iterator (i.e. `NewProfileIterator` semantics).
-#[unsafe(no_mangle)]
-pub unsafe extern "C" fn ForEachIntersectionChildMut(
-    header: *mut QueryIterator,
-    callback: unsafe extern "C" fn(*mut *mut QueryIterator),
-) {
-    debug_assert!(!header.is_null());
-    debug_assert_eq!(
-        // SAFETY: safe thanks to 1
-        unsafe { (*header).type_ },
-        IteratorType::Intersect,
-        "Expected an INTERSECT_ITERATOR"
-    );
-    // SAFETY: safe thanks to 1
-    let wrapper = unsafe {
-        RQEIteratorWrapper::<Intersection<CRQEIterator>>::mut_ref_from_header_ptr(header)
-    };
-
-    for child in wrapper.inner.children_mut() {
-        // SAFETY:
-        // - `CRQEIterator` is `#[repr(transparent)]` over `NonNull<QueryIterator>`, which has
-        //   the same layout as `*mut QueryIterator`.
-        // - Casting `*mut CRQEIterator` to `*mut *mut QueryIterator` gives the callback direct
-        //   write access to the raw pointer stored inside the `CRQEIterator`.
-        // - After the callback (e.g. `Profile_AddIters`) writes a new pointer into the slot,
-        //   the `CRQEIterator` owns it and will free it correctly on drop. Safe per contract 3.
-        let slot = child as *mut CRQEIterator as *mut *mut QueryIterator;
-        // SAFETY: safe thanks to 2 and 3.
-        unsafe { callback(slot) };
-    }
 }
--- a/src/redisearch_rs/headers/iterators_rs.h
+++ b/src/redisearch_rs/headers/iterators_rs.h
@@ -126,8 +126,7 @@ size_t GetIntersectionIteratorNumChildren(const QueryIterator *header);
  * Returns a non-owning raw pointer to the child at `idx`.
  *
  * The returned pointer is valid as long as the intersection iterator is alive and no
- * structural modifications are made (e.g. via [`AddIntersectionIteratorChild`] or
- * [`ForEachIntersectionChildMut`]).
+ * structural modifications are made (e.g. via [`AddIntersectionIteratorChild`]).
  *
  * # Safety
  *
@@ -152,27 +151,6 @@ const QueryIterator *GetIntersectionIteratorChild(const QueryIterator *header, s
  * 2. `child` must be a valid non-null pointer to a `QueryIterator`, not aliased.
  */
 void AddIntersectionIteratorChild(QueryIterator *header, QueryIterator *child);
-
-/**
- * Apply `callback` to each child iterator slot, passing a mutable `QueryIterator**`.
- *
- * This is designed for use with `Profile_AddIters`, which replaces each child with a
- * profile-wrapping iterator in place. The callback receives a pointer to the raw pointer
- * stored inside each [`CRQEIterator`] child, allowing it to update (replace) that pointer.
- *
- * This is safe because [`CRQEIterator`] is `#[repr(transparent)]` over
- * `NonNull<QueryIterator>`, which has the same memory layout as `*mut QueryIterator`.
- * The callback's in-place mutation of the slot directly updates the `CRQEIterator`'s
- * internal pointer, so the intersection will subsequently own (and free) the new iterator.
- *
- * # Safety
- *
- * 1. `header` must be a valid non-null pointer created via [`NewIntersectionIterator`].
- * 2. `callback` must be a valid function pointer.
- * 3. The callback must replace `*slot` with a valid non-null `QueryIterator*` that takes
- *    ownership of the original iterator (i.e. `NewProfileIterator` semantics).
- */
-void ForEachIntersectionChildMut(QueryIterator *header, void (*callback)(QueryIterator**));
 
 /**
  * Gets the flags of the underlying IndexReader from an inverted index iterator.


gdesmott:
@LukeMathWalker : I'd be interested to have your feedback regarding the API and general design here.

src/iterators/hybrid_reader.h
```diff
@@ -28,16 +28,17 @@ typedef struct {
   const FieldFilterContext* filterCtx;
 } HybridIteratorParams;
 
+// Keep in sync with the partial Rust layout in rqe_iterators/src/c2rust.rs.
 typedef struct {
```
ThierryBerger: I assume a lot of tests will fail if this is broken, but should we consider a dedicated unit test for that ?

src/redisearch_rs/c_entrypoint/iterators_ffi/src/profile.rs
```diff
@@ -100,3 +101,32 @@ pub unsafe extern "C" fn ProfileIterator_GetWallTimeNs(it: *const QueryIterator)
...
+pub unsafe extern "C" fn Profile_AddIters(root: *mut *mut QueryIterator) {
+    debug_assert!(!root.is_null());
+    // SAFETY: guaranteed by 1.
+    let it = unsafe { *root };
+    if it.is_null() {
+        return;
+    }
+
+    // SAFETY: guaranteed by 2 — *root is non-null (just checked).
+    let it = unsafe { NonNull::new_unchecked(it) };
```
LukeMathWalker: Why not `NonNull::new` and then exit early if it returns `None`?

src/iterators/hybrid_reader.h
gdesmott: I moved some types to header files and I am now using everything from `ffi`.

src/redisearch_rs/rqe_iterators/src/interop.rs
```diff
@@ -215,6 +216,44 @@ extern "C" fn num_estimated<'index, I: RQEIterator<'index> + 'index>(
...
+///
+/// Dispatches on the iterator type tag: composite iterators (Not, Optional)
+/// profile their children in-place; everything else is a leaf and is returned
+/// unchanged.
+///
+extern "C" fn rust_profile_children(base: *mut QueryIterator) -> *mut QueryIterator {
+    // SAFETY: Guaranteed by the caller — `base` is a valid RQEIteratorWrapper.
+    let type_ = unsafe { (*base).type_ };
+
+    // Update the match when porting a new composite iterator to Rust.
```
gdesmott: @LukeMathWalker : we still have some iterator specific logic here to break the recursion, but I think it's already much better than the previous version.

We could make it more robust but that would be yet another trait. 

What do you think?

src/iterators/hybrid_reader.h
gdesmott: Actually we don't even need this anymore with the new design (see top commit).

src/redisearch_rs/rqe_iterators/src/c2rust.rs
```diff
@@ -317,3 +322,41 @@ impl<'index> RQEIterator<'index> for CRQEIterator {
...
+/// Profile-wrap a child [`CRQEIterator`] by calling [`Profilable::into_profiled`]
+/// and boxing the result back into a [`CRQEIterator`].
+pub fn profile_child(child: CRQEIterator) -> CRQEIterator {
+    let profiled = child.into_profiled();
+    let ptr = RQEIteratorWrapper::boxed_new(IteratorType_PROFILE_ITERATOR, profiled);
+    // SAFETY: `boxed_new` returns a valid, owning, non-null pointer.
+    let ptr = unsafe { NonNull::new_unchecked(ptr) };
+    // SAFETY: `ptr` is valid and owning per above.
+    unsafe { CRQEIterator::new(ptr) }
+}
```
LukeMathWalker: Why is this called `profile_child`? Looking at the impl, it's profiling itself (and its children recursively), no?

src/redisearch_rs/rqe_iterators/src/interop.rs
LukeMathWalker: Mmmm I don't think we need another trait 🤔 
In `boxed_new` we should be able to set `ProfileChildren: Some(rust_profile_children::<I>)`, since we know the concrete iterator type at that point.
We then change the signature here to:
```rust
extern "C" fn rust_profile_children<'index, I: RQEIterator<'index> + Profilable<'index> + 'index>(base: *mut QueryIterator) -> *mut QueryIterator
```
and in the impl we can do
```rust
if !base.is_null() {
      debug_assert!(base.is_aligned());
      // SAFETY: Callbacks are guaranteed to get a header pointer created by
      //  [`RQEIteratorWrapper::new`], which (internally) uses `Box::into_raw` to
      //  return a raw header pointer.
      let it = unsafe { Box::from_raw(base as *mut RQEIteratorWrapper<I>) };
      // Then add profile + call CRQEIterator::new
  }
```

Wdyt?

src/redisearch_rs/rqe_iterators/src/c2rust.rs
gdesmott: Good catch; fixed.

src/redisearch_rs/rqe_iterators/src/interop.rs
gdesmott: That means we'd now have

```
RQEIteratorWrapper<I>
where
    I: RQEIterator<'index> + Profilable<'index> + 'index
```

But then `c2rust::into_profiled()` (I just renamed it, cf previous comment) won't be able to call `RQEIteratorWrapper::boxed_new()` as `Profile<>` itself does not implement `Profilable`.

```
error[E0277]: the trait bound `Profile<'_, CRQEIterator>: Profilable<'_>` is not satisfied
   --> rqe_iterators/src/c2rust.rs:330:76
    |
330 |     let ptr = RQEIteratorWrapper::boxed_new(IteratorType_PROFILE_ITERATOR, profiled);
    |               -----------------------------                                ^^^^^^^^ unsatisfied trait bound
    |               |
    |               required by a bound introduced by this call
    |
help: the trait `Profilable<'_>` is not implemented for `Profile<'_, CRQEIterator>`
   --> rqe_iterators/src/profile.rs:45:1
    |
 45 | pub struct Profile<'index, I: RQEIterator<'index>> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: the following other types implement trait `Profilable<'index>`:
              `CRQEIterator` implements `Profilable<'_>`
              `IdList<'index, SORTED>` implements `Profilable<'index>`
              `Optional<'index, I>` implements `Profilable<'index>`
              `Tag<'index, E, C>` implements `Profilable<'index>`
              `UnionFlat<'index, I, QUICK_EXIT>` implements `Profilable<'index>`
              `empty::Empty` implements `Profilable<'index>`
              `intersection::Intersection<'index, I>` implements `Profilable<'index>`
              `inverted_index::numeric::Numeric<'index, R, E>` implements `Profilable<'index>`
            and 6 others
note: required by a bound in `RQEIteratorWrapper::<I>::boxed_new`
   --> rqe_iterators/src/interop.rs:39:30
    |
 39 |     I: RQEIterator<'index> + Profilable<'index> + 'index,
    |                              ^^^^^^^^^^^^^^^^^^ required by this bound in `RQEIteratorWrapper::<I>::boxed_new`
...
 44 |     pub fn boxed_new(type_: ffi::IteratorType, inner: I) -> *mut QueryIterator {
    |            --------- required by a bound in this associated function
```

LukeMathWalker: I would put the generic bounds on the constructor rather than the struct, but yes, that means we need two different constructors.

LukeMathWalker: Wait though, why can't profile be profilable as a noop?

gdesmott: I tried that but python tests are crashing because the inner type is now changing. Claude explained the problem well:

image

So we'd need to modify the C profile traversal code and add new specialized accessors , etc. 
At this point it may be easier to port this code to Rust as well then.

src/redisearch_rs/rqe_iterators/src/profile.rs
```diff
@@ -87,28 +87,39 @@ impl<'index, I: RQEIterator<'index>> Profile<'index, I> {
...
+    /// (type-erased, the [`Profile`] layer is hidden behind the pointer).
+    type IntoProfiled: RQEIterator<'index>;
+
+    /// Leaf iterators have no children to profile.
+    ///
+    /// When `true`, [`RQEIteratorWrapper::boxed_new`](crate::interop::RQEIteratorWrapper::boxed_new)
+    /// sets the C vtable `ProfileChildren` callback to `NULL`, so the
+    /// wrapper is never consumed and reallocated during profiling. This
+    /// keeps interior pointers (held by C code) valid.
+    const IS_LEAF: bool = true;
```
LukeMathWalker: Let's avoid the default value, since that's likely to cause issues when we add new implementations and forget to set to `false` if needed 😅

src/redisearch_rs/rqe_iterators/src/profile.rs
```diff
@@ -140,3 +179,61 @@ impl<'index, I: RQEIterator<'index>> RQEIterator<'index> for Profile<'index, I>
...
+/// Leaf impl for type-erased iterators (used by wildcard FFI).
+///
+/// `dyn RQEIterator` is not `Profilable` (not `Sized`), so the
+/// `Box<P: Profilable>` blanket impl doesn't cover this case.
+///
+/// We cannot delegate to the inner iterator's `profile_children` because
+/// trait objects cannot implement `Profilable` (`Sized` bound). This is
+/// fine in practice: only leaf iterators are boxed as `dyn RQEIterator`
+/// today (wildcard FFI). If a composite were ever boxed this way, its
+/// children would silently not be profiled.
```
LukeMathWalker: In what way is the ' Sized ' bound an issue here? 
I'd expect the associated type on `Profilable` to be a blocker, but that can be worked around.

gdesmott: The problem is `Profilable` is not dyn compatible. I updated the comment.

Can we workaround that?

gdesmott: I implemented your suggested workaround.

src/redisearch_rs/c_entrypoint/iterators_ffi/src/inverted_index/wildcard.rs
```diff
@@ -173,8 +174,22 @@ pub unsafe extern "C" fn NewInvIndIterator_WildcardQuery(
...
         self
     }
+
+    fn into_profiled(self) -> Self::IntoProfiled {
+        Profile::new(self.profile_children())
+    }
+
+    fn into_profiled_boxed(self: Box<Self>) -> Box<dyn RQEIterator<'index> + 'index> {
+        Box::new((*self).into_profiled())
+    }
```
ThierryBerger: Nit: as there is some duplication for leaf iterators, there may be value in providing a deduplication logic, either through macro or a common trait marker (`RustLeafIterator`) to constrain a trait implementation to those..? It may be my curiosity talking more than a real advantage, I won't consider this code duplication a blocker.

src/redisearch_rs/rqe_iterators/src/profile.rs
```diff
@@ -87,28 +87,54 @@ impl<'index, I: RQEIterator<'index>> Profile<'index, I> {
...
-    /// For leaf iterators this is `Self`. For composite iterators the child type
-    /// parameter changes to reflect the profiled children.
-    type Profiled: RQEIterator<'index>;
+    /// When `true`, [`RQEIteratorWrapper::boxed_new`](crate::interop::RQEIteratorWrapper::boxed_new)
+    /// sets the C vtable `ProfileChildren` callback to `NULL`, so the
+    /// wrapper is never consumed and reallocated during profiling. This
+    /// keeps interior pointers (held by C code) valid.
+    fn is_leaf() -> bool
+    where
+        Self: Sized;
```
ThierryBerger: Nit: `is_leaf` is a bit confusing, I'd expect a leaf to not have children, but currently it could in case of c. The doc comment is helpful but consider another naming?
- `is_profile_leaf`, to disambiguate the more generic "leaf node"
- or an inverse: `has_children_to_profile`: not a fan of this one but makes it a bit clearer about its purpose
- or a practical oriented: `use_rust_profile_children`: so we don't have to do too much mental gymnastic ?

Once migration is full rust, this can be back to `is_leaf` (and reflecting the "tree leaf" definition too), but until then it's confusing.

Not too sure about this, I won't block on it as it's kinda temporary... but I'm interested in the discussion.

src/redisearch_rs/rqe_iterators/src/interop.rs
```diff
@@ -215,6 +242,33 @@ extern "C" fn num_estimated<'index, I: RQEIterator<'index> + 'index>(
...
+) -> *mut QueryIterator {
+    debug_assert!(!base.is_null());
+    debug_assert!(base.is_aligned());
+    // SAFETY: Callbacks are guaranteed to get a header pointer created by
+    // `RQEIteratorWrapper::boxed_new`, which uses `Box::into_raw`.
+    let it = unsafe { Box::from_raw(base as *mut RQEIteratorWrapper<I>) };
+    let type_ = it.header.type_;
+    let profiled = it.inner.profile_children();
+    // `None`: see doc comment above.
+    RQEIteratorWrapper::boxed_new_inner(type_, profiled, None)
```
ThierryBerger: just so I'm sure I understand: we end up with a (raw boxed) `RQEWrapper(Profile(I))`

src/redisearch_rs/rqe_iterators/src/interop.rs
```diff
@@ -66,12 +69,87 @@ where
...
+/// # How profiling flows
+///
+/// 1. C calls `Profile_AddIters` → [`CRQEIterator::into_profiled`](crate::c2rust::CRQEIterator::into_profiled)
+/// 2. `into_profiled` calls [`CRQEIterator::profile_children`](crate::c2rust::CRQEIterator::profile_children),
+///    which invokes the C vtable `ProfileChildren` callback
+/// 3. For Rust compound iterators, that callback is [`rust_profile_children`],
+///    which calls this trait's [`profile_children`](Self::profile_children)
+/// 4. The implementation profiles each child via `CRQEIterator::into_profiled`
+///    (preserving `CRQEIterator`) and returns the same compound type
+/// 5. The result is re-wrapped in [`RQEIteratorWrapper`] with the same type parameter
```
ThierryBerger: I feel it misses the step where the wrapper is "opened" (from_raw), to inject the profiler, then rewrapped (without profiling capabilities, but this may not be relevant **here**) ; if we didn't do that, it would be easy to end up with a `wrap<Profile<wrap<iterator>>>`, where in current impl we have `wrap<Profile<iterator>>`.

Also a clarification that is hinted on but may be lost in verbosity or specific example is that all compound iterators in rust wrap their children in `RQEIteratorWrapper`, it's hinted at through intersection example but is that enough?

Nice explanation though! 

As I like graphs to help understanding, below is what the current text says: 

```mermaid
flowchart TD
    A["C calls\nProfile_AddIters"] --> B["CRQEIterator::into_profiled"]
    B --> C["CRQEIterator::profile_children\n(invokes C vtable ProfileChildren callback)"]
    C --> D{Iterator type?}
    D -->|Rust compound iterator| E["rust_profile_children\n(Rust-side vtable impl)"]
    D -->|Other| Z["(C-handled path)"]
    E --> F["Trait::profile_children\non the compound iterator"]
    F --> G["For each child:\nCRQEIterator::into_profiled\n(preserves CRQEIterator)"]
    G -->|recurse| B
    G --> H["Returns same\ncompound type"]
    H --> I["Re-wrapped in\nRQEIteratorWrapper\n(same type parameter)"]
```

src/redisearch_rs/rqe_iterators/src/interop.rs
```diff
@@ -215,12 +294,32 @@ extern "C" fn num_estimated<'index, I: RQEIterator<'index> + 'index>(
...
+    base: *mut QueryIterator,
+) -> *mut QueryIterator {
+    debug_assert!(!base.is_null());
+    debug_assert!(base.is_aligned());
+    // SAFETY: Callbacks are guaranteed to get a header pointer created by
+    // `RQEIteratorWrapper::boxed_new_compound`, which uses `Box::into_raw`.
+    let it = unsafe { Box::from_raw(base as *mut RQEIteratorWrapper<I>) };
+    let profiled = it.inner.profile_children();
+    // `None`: profiling is a one-shot pass — the result doesn't need a callback.
+    RQEIteratorWrapper::boxed_new_inner(profiled, None)
```
ThierryBerger: This comment is important, as reader could be tempted to use `boxed_new_compound`, maybe this can be named to be more explicit?

src/redisearch_rs/rqe_iterators/tests/integration/profilable.rs
```diff
@@ -0,0 +1,276 @@
...
+    assert_eq!(r.doc_id, 5); // virtual
+    assert_eq!(profiled.counters().read, 5);
+
+    // EOF
+    assert!(profiled.read().unwrap().is_none());
+    assert!(profiled.counters().eof);
+}
+
+/// Wrapping `Intersection` in `Profile` counts reads.
+#[cfg(not(miri))] // calls ffi::RSYieldableMetric_Concat
```
ThierryBerger: consider `ignore = ""` ?

src/redisearch_rs/rqe_iterators/src/interop.rs
gdesmott: I improved the comment.

src/redisearch_rs/rqe_iterators/src/interop.rs
gdesmott: I extended the comments block a bit.

PR #8619, opened 2026-03-05
[MOD-13436] Intersection iterator: add ffi and swap implementation

ThierryBerger:
- Based on https://github.com/RediSearch/RediSearch/pull/8531

## NOT in this PR

- I used C implementation of `IndexResult_IsWithinRange`  to keep the stop the creeping scope of the PR.

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

#### Release Notes

- [ ] This PR requires release notes
- [x] This PR does not require release notes

If a release note is required (bug fix / new feature / enhancement), describe the **user impact** of this PR in the title.  

---

> [!NOTE]
> **High Risk**
> Replaces the core `INTERSECT_ITERATOR` implementation with a Rust+FFI version and adds unsafe child-mutation/access APIs, which can affect query correctness, scoring, and memory safety across many query paths.
> 
> **Overview**
> Swaps the intersection iterator implementation from the previous C `IntersectionIterator` to a Rust-backed iterator exposed via FFI (`NewIntersectionIterator`), leaving C with only `IntersectionIteratorReducer` for wildcard/empty/single-child short-circuiting and ownership cleanup.
> 
> Adds new FFI APIs to *inspect and mutate* intersection children (`GetIntersectionIteratorNumChildren`, `GetIntersectionIteratorChild`, `AddIntersectionIteratorChild`, `ForEachIntersectionChildMut`) and updates call sites (profiling tree wrapping and optimizer root updates) and C++ tests/benchmarks to use these accessors instead of casting to the removed C struct.
> 
> Enhances the Rust `Intersection` implementation with weighted child-sorting heuristics (optionally prioritizing union/intersection children), explicit result weight propagation, dynamic child append, and per-document metric resets; updates Rust tests and benchmarks accordingly (dropping the old C intersection benchmark path).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 19c5ffc79343d9b577d5c1aa1c55bdb54ac2bc58. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #8764, opened 2026-03-19
[MOD-13436] Fix full rust intersection heuristic implementation

ThierryBerger:
- Based on #8619

This implements the heuristic feature for intersection iterator children sorting for non c wrapped iterators (namely `Box<dyn RQEIterators>`).

It's currently not used in production, but will be once we swap all iterators to Rust.
This PR also adds a tests for this specific heuristic sorting.

## API choice

We decided to go with a more specific `intersection_sort_weight` into RQEIterator, because :
- The logic is dependent on `type_()`, but should depend on its "logical" type (so `type_()` doesn't fit when profiling)
- Adding a `children_count` getter is not relevant for some iterators, and ambiguous for others.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Changes the `Intersection` child-sorting heuristic by introducing a new `RQEIterator::intersection_sort_weight` API and using it to reorder children, which can affect query execution order and performance. While intended as a heuristic-only change, it touches core iterator composition paths (including C-interop) and could expose edge-case behavior differences.
> 
> **Overview**
> `Intersection` now sorts its children using a new per-iterator heuristic hook, `RQEIterator::intersection_sort_weight(prioritize_union_children)`, instead of relying on the previous C-wrapper-only weight calculation.
> 
> All iterator implementations (Rust iterators, C-interop `CRQEIterator`, and FFI wrapper iterators) were updated to implement/forward this method; `Intersection` returns `1.0 / num_children`, `Union` returns `num_children` when prioritization is enabled, and wrappers like `Profile`/`MaybeEmpty` forward to their child.
> 
> Adds integration tests to ensure nested `Intersection` (including when wrapped in `Profile`) sorts ahead of equally-estimated plain children, validating the heuristic for `Box<dyn RQEIterator>` cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2dc08281be3272adcd86b712c10d3014dbf0f9c8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rqe_iterators/src/c2rust.rs
+++ b/src/redisearch_rs/rqe_iterators/src/c2rust.rs
@@ -321,7 +321,7 @@ impl<'index> RQEIterator<'index> for CRQEIterator {
         Some(self)
     }
 
-    fn children_count(&self) -> usize {
+    fn intersection_sort_weight(&self, prioritize_union_children: bool) -> f64 {
         match self.type_ {
             IteratorType::Intersect => {
                 let ptr = std::ptr::from_ref(self.as_ref());
@@ -333,19 +333,41 @@ impl<'index> RQEIterator<'index> for CRQEIterator {
                 // - `ref_from_header_ptr` uses the compiler-computed field offset for `inner`
                 //   rather than manual `size_of` arithmetic, making it immune to alignment
                 //   padding between `header` and `inner` in `RQEIteratorWrapper`.
-                unsafe {
+                let n = unsafe {
                     RQEIteratorWrapper::<Intersection<'_, CRQEIterator>>::ref_from_header_ptr(ptr)
                         .inner
                         .num_children()
-                }
+                };
+                1.0 / n.max(1) as f64
             }
-            IteratorType::Union => {
+            IteratorType::Union if prioritize_union_children => {
                 let ptr = std::ptr::from_ref(self.as_ref());
                 // SAFETY: `type_` guarantees `ptr` points to a `UnionIterator` whose first field
                 // is the `QueryIterator` base — the cast is valid by C struct layout.
-                unsafe { (*ptr.cast::<ffi::UnionIterator>()).num as usize }
+                let n = unsafe { (*ptr.cast::<ffi::UnionIterator>()).num as usize };
+                n.max(1) as f64
             }
-            _ => 0,
+            IteratorType::InvIdxNumeric => 1.0,
+            IteratorType::InvIdxTerm => 1.0,
+            IteratorType::InvIdxWildcard => 1.0,
+            IteratorType::InvIdxMissing => 1.0,
+            IteratorType::InvIdxTag => 1.0,
+            IteratorType::Hybrid => 1.0,
+            IteratorType::Union => 1.0,
+            IteratorType::Not => 1.0,
+            IteratorType::NotOptimized => 1.0,
+            IteratorType::Optional => 1.0,
+            IteratorType::OptionalOptimized => 1.0,
+            IteratorType::Wildcard => 1.0,
+            IteratorType::Empty => 1.0,
+            IteratorType::IdListSorted => 1.0,
+            IteratorType::IdListUnsorted => 1.0,
+            IteratorType::MetricSortedById => 1.0,
+            IteratorType::MetricSortedByScore => 1.0,
+            IteratorType::Profile => 1.0,
+            IteratorType::Optimus => 1.0,
+            IteratorType::Mock => 1.0,
+            IteratorType::Max => 1.0,
         }
     }
 }
--- a/src/redisearch_rs/rqe_iterators/src/lib.rs
+++ b/src/redisearch_rs/rqe_iterators/src/lib.rs
@@ -152,8 +152,17 @@ pub trait RQEIterator<'index> {
         None
     }
 
-    /// Returns the amount of children of this iterator.
-    fn children_count(&self) -> usize;
+    /// Returns the sort weight for this iterator when used as a child of an [`Intersection`].
+    ///
+    /// [`Intersection`] uses this to order its children before execution: a lower value makes
+    /// this iterator act as the pivot (minimising `SkipTo` calls). The final sort key is
+    /// `num_estimated * intersection_sort_weight(...)`.
+    ///
+    /// Implementers:
+    /// - [`Intersection`]: `1.0 / num_children` — fewer children means tighter selectivity.
+    /// - [`Union`]: `num_children` when `prioritize_union_children`, else `1.0`.
+    /// - Everything else: `1.0` — neutral, no influence.
+    fn intersection_sort_weight(&self, prioritize_union_children: bool) -> f64;
 }
 
 /// Blanket [`RQEIterator`] impl for `Box<I>` where `I` is a concrete iterator type.
@@ -202,6 +211,10 @@ impl<'index, I: RQEIterator<'index> + 'index> RQEIterator<'index> for Box<I> {
 
     fn as_c_iterator(&self) -> Option<&c2rust::CRQEIterator> {
         (**self).as_c_iterator()
+    }
+
+    fn intersection_sort_weight(&self, prioritize_union_children: bool) -> f64 {
+        (**self).intersection_sort_weight(prioritize_union_children)
     }
 }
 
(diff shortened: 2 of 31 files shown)


src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -190,6 +214,18 @@ impl<'index> RQEIterator<'index> for Box<dyn RQEIterator<'index> + 'index> {
     fn as_c_iterator(&self) -> Option<&c2rust::CRQEIterator> {
         (**self).as_c_iterator()
     }
+
+    fn is_intersection(&self) -> bool {
```
gdesmott: That's 3 very specific API for a single use case. Maybe we should bundle them together? `fn get_children_details(&self) -> Option<(IntersectionOrUnion, usize)>` ?

ThierryBerger: makes sense, I´d use a proper type instead of a tuple, but I'm ok with that. We'll consider later if that needs splitting.

gdesmott: cool, let's go with a `struct` then.

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -147,6 +147,30 @@ pub trait RQEIterator<'index> {
...
+        false
+    }
+
+    /// Returns the number of direct child iterators, if this is a composite iterator.
+    ///
+    /// Used together with [`is_intersection`](Self::is_intersection) and
+    /// [`is_union`](Self::is_union) by [`Intersection`] to compute sort weights.
+    fn num_children(&self) -> Option<usize> {
+        None
+    }
```
LukeMathWalker: Let's avoid default impls, since they make it easier to forget about these methods when working on new iterators.

src/redisearch_rs/rqe_iterators/src/lib.rs
ThierryBerger: done :+1:

src/redisearch_rs/rqe_iterators/src/lib.rs
ThierryBerger: updated.

src/redisearch_rs/rqe_iterators/src/optional.rs
```diff
@@ -240,4 +242,10 @@ where
     fn type_(&self) -> IteratorType {
         IteratorType::Optional
     }
+
+    fn children_details(&self) -> Option<ChildrenDetails> {
+        // Optional is not an intersection/union: it emits virtual results for every document,
+        // so it is never selective and should not influence sort weights in parent intersections.
```
ThierryBerger: maybe overly specific towards intersection ?

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -138,12 +138,28 @@ pub trait RQEIterator<'index> {
     /// Returns the [`IteratorType`] of this iterator.
     fn type_(&self) -> IteratorType;
 
-    /// Returns `Some(&self)` if this iterator is a [`c2rust::CRQEIterator`], `None` otherwise.
-    ///
-    /// Used by [`Intersection`] to compute sort weights without requiring `'static`.
-    fn as_c_iterator(&self) -> Option<&c2rust::CRQEIterator> {
```
gdesmott, line 153: We need to keep this, it's used in my not optimized branches.

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -138,12 +138,28 @@ pub trait RQEIterator<'index> {
...
+    /// An OR-union of multiple child iterators.
+    Union,
+}
+
+/// Details about the children of a composite iterator, as returned by
+/// [`RQEIterator::children_details`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ChildrenDetails {
+    /// Whether this is an intersection or a union.
+    pub kind: IntersectionOrUnion,
```
gdesmott: We have `type_()` now.

ThierryBerger: I changed to a `children_count` then, it's made specifically for the intersection/union, but it's better to not duplicate information indeed.

src/redisearch_rs/rqe_iterators/src/c2rust.rs
```diff
@@ -318,4 +321,32 @@ impl<'index> RQEIterator<'index> for CRQEIterator {
...
+                        .num_children()
+                }
+            }
+            IteratorType::Union => {
+                let ptr = std::ptr::from_ref(self.as_ref());
+                // SAFETY: `type_` guarantees `ptr` points to a `UnionIterator` whose first field
+                // is the `QueryIterator` base — the cast is valid by C struct layout.
+                unsafe { (*ptr.cast::<ffi::UnionIterator>()).num as usize }
+            }
+            _ => 0,
```
gdesmott: Worth a comment explaining all other iterator types have no children.

src/redisearch_rs/rqe_iterators/src/lib.rs
```diff
@@ -144,6 +144,9 @@ pub trait RQEIterator<'index> {
     fn as_c_iterator(&self) -> Option<&c2rust::CRQEIterator> {
         None
     }
+
+    /// Returns the amount of children of this iterator.
```
gdesmott: Technically `not` and `optional` have one child so we should explain the exact expected semantic here.

src/redisearch_rs/rqe_iterators/src/optional.rs
gdesmott: I'd document the overall logic in the trait API definition.

ThierryBerger:
We decided to go with a more specific `intersection_sort_weight` into RQEIterator, because :
- The logic is dependent on `type_()`, but should depend on its "logical" type (so `type_()` doesn't fit when profiling)
- adding a `children_count` getter is not relevant for some iterators, and ambiguous for others.

src/redisearch_rs/rqe_iterators/src/c2rust.rs
```diff
@@ -318,4 +321,34 @@ impl<'index> RQEIterator<'index> for CRQEIterator {
...
+                1.0 / n.max(1) as f64
+            }
+            IteratorType::Union if prioritize_union_children => {
+                let ptr = std::ptr::from_ref(self.as_ref());
+                // SAFETY: `type_` guarantees `ptr` points to a `UnionIterator` whose first field
+                // is the `QueryIterator` base — the cast is valid by C struct layout.
+                let n = unsafe { (*ptr.cast::<ffi::UnionIterator>()).num as usize };
+                n.max(1) as f64
+            }
+            _ => 1.0,
```
gdesmott: Should we list all types here, for the same reason we do not have a default implementation of the trait method?

src/redisearch_rs/rqe_iterators/src/c2rust.rs
```diff
@@ -320,6 +320,56 @@ impl<'index> RQEIterator<'index> for CRQEIterator {
...
+            IteratorType::Wildcard => 1.0,
+            IteratorType::Empty => 1.0,
+            IteratorType::IdListSorted => 1.0,
+            IteratorType::IdListUnsorted => 1.0,
+            IteratorType::MetricSortedById => 1.0,
+            IteratorType::MetricSortedByScore => 1.0,
+            IteratorType::Profile => 1.0,
+            IteratorType::Optimus => 1.0,
+            IteratorType::Mock => 1.0,
+            IteratorType::Max => 1.0,
```
gdesmott, line 370: Shouldn't we panic here?

ThierryBerger: probably, well I don't particularly enjoy adding panic where there's no catastrophically incorrect state, I'd argue we should feature gate `Mock` and `Max` :thinking:

gdesmott: C code needs `Max` so we need to keep it. I can live with `Max` being handled if you prefer.

BenGoldberger (approved):

gdesmott (approved):

PR #8778, opened 2026-03-21
Allow SORTBY only by vector distance in flex - avoid blocking SORTBY completely

alonre24:
In Redis Flex (disk mode), SORTBY was previously blocked entirely. This PR allows SORTBY when sorting by vector distance fields (e.g., `__v_score` or custom AS aliases from KNN queries), while continuing to block SORTBY on other fields.

**Changes:**
src/aggregate/aggregate_request.c: Removed early SORTBY blocking; validation is now deferred to after query parsing.

src/aggregate/aggregate_exec.c: Added validateSortbyForDiskIndex() that runs after `QAST_Iterate` when vector score field names are populated. Uses `RS_LOG_ASSERT` to enforce flex mode assumptions (single SORTBY field, single vector score field) for a simple O(1) check.

tests/pytests/test_flex_validation.py: Added `test_flex_allows_sortby_on_vector_distance_fields` and updated `test_flex_blocks_sortby_on_non_vector_fields`.

Error message:
"SORTBY in Redis Flex is restricted to sorting results by vector distance in vector queries. Otherwise, results are always sorted by the default document score."

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Changes query validation for `SORTBY` in Flex/disk mode by deferring enforcement until after AST iteration and allowing a new supported case (vector distance). Risk is moderate because it affects query planning/validation paths and could change which queries are accepted or rejected in production.
> 
> **Overview**
> In Flex (disk) mode, `SORTBY` is no longer blanket-rejected during argument parsing; instead, validation is deferred until after query parsing/AST iteration so the engine can detect vector score fields.
> 
> Execution now permits `SORTBY` *only* when the sort key matches the single vector distance score field produced by a KNN clause (e.g. `__v_score` or an `AS` alias); all other `SORTBY` usages return a new Flex-specific error message. Tests are updated to cover both the allowed vector-distance `SORTBY` cases and the continued rejection of non-vector sorts.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cc3b2dbfe2d17509cd65e1bbffa1c536306e0ad0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/aggregate/aggregate_exec.c
+++ b/src/aggregate/aggregate_exec.c
@@ -1176,14 +1176,10 @@ static int validateSortbyForDiskIndex(AREQ *req, QueryError *status) {
     return REDISMODULE_OK;
   }
 
+  // If HasSortBy is true, arrange step and sortKeys must exist
   PLN_ArrangeStep *arng = AGPLN_GetArrangeStep(AREQ_AGGPlan(req));
-  if (!arng || !arng->sortKeys || array_len(arng->sortKeys) == 0) {
-    return REDISMODULE_OK;  // No sort keys to validate
-  }
-
-  // In flex mode, FT.SEARCH supports only a single SORTBY field
-  size_t numSortKeys = array_len(arng->sortKeys);
-  RS_LOG_ASSERT(numSortKeys == 1, "Flex mode expects exactly one SORTBY field");
+  RS_LOG_ASSERT(arng && arng->sortKeys && array_len(arng->sortKeys) == 1,
+                "Flex mode expects exactly one SORTBY field");
 
   // Get the metric requests from the AST (vector score fields)
   MetricRequest *metricRequests = req->ast.metricRequests;


src/aggregate/aggregate_exec.c
```diff
@@ -1152,6 +1152,62 @@ void AREQ_Execute_Callback(blockedClientReqCtx *BCRctx) {
...
+  }
+
+  PLN_ArrangeStep *arng = AGPLN_GetArrangeStep(AREQ_AGGPlan(req));
+  if (!arng || !arng->sortKeys || array_len(arng->sortKeys) == 0) {
+    return REDISMODULE_OK;  // No sort keys to validate
+  }
+
+  // In flex mode, FT.SEARCH supports only a single SORTBY field
+  size_t numSortKeys = array_len(arng->sortKeys);
+  RS_LOG_ASSERT(numSortKeys == 1, "Flex mode expects exactly one SORTBY field");
```
JoanFM: why do we assert instead of validate?

src/aggregate/aggregate_exec.c
```diff
@@ -1152,6 +1152,62 @@ void AREQ_Execute_Callback(blockedClientReqCtx *BCRctx) {
...
+
+  // Get the metric requests from the AST (vector score fields)
+  MetricRequest *metricRequests = req->ast.metricRequests;
+  size_t numMetrics = array_len(metricRequests);
+
+  // In flex mode, KNN is allowed only once, so at most one vector score field
+  RS_LOG_ASSERT(numMetrics <= 1, "Flex mode expects at most one vector score field");
```
JoanFM, line 1189: also this, why?

src/aggregate/aggregate_exec.c
alonre24: Since we should not be able to go here if we are running a command that is not FT.SEARCH (other commands are blocked) and FT.SEARCH allows a single SORTBY clause (should be validated before as well)

src/aggregate/aggregate_exec.c
alonre24: Due to the comment above, we expect no more than one vector score field exposed since only KNN should be allowed, and KNN is allowed (range is not allowed in flex) once in a query (always)

src/aggregate/aggregate_exec.c
```diff
@@ -1152,6 +1152,62 @@ void AREQ_Execute_Callback(blockedClientReqCtx *BCRctx) {
...
+    return REDISMODULE_OK;
+  }
+
+  // Skip if no SORTBY
+  if (!HasSortBy(req)) {
+    return REDISMODULE_OK;
+  }
+
+  PLN_ArrangeStep *arng = AGPLN_GetArrangeStep(AREQ_AGGPlan(req));
+  if (!arng || !arng->sortKeys || array_len(arng->sortKeys) == 0) {
```
raz-mon: Third condition can become an assertion?

src/aggregate/aggregate_exec.c
```diff
@@ -1152,6 +1152,62 @@ void AREQ_Execute_Callback(blockedClientReqCtx *BCRctx) {
...
+  // Get the metric requests from the AST (vector score fields)
+  MetricRequest *metricRequests = req->ast.metricRequests;
+  size_t numMetrics = array_len(metricRequests);
+
+  // In flex mode, KNN is allowed only once, so at most one vector score field
+  RS_LOG_ASSERT(numMetrics <= 1, "Flex mode expects at most one vector score field");
+
+  // If there's no vector score field, or the sort key doesn't match it, block
+  const char *sortKey = arng->sortKeys[0];
+  bool isVectorScoreField = (numMetrics == 1) &&
```
raz-mon, line 1193: Can also validate a type here?

alonre24: Type of what?

raz-mon: Thought the `metricRequest` would have a type, but I see there is no, so nvm

PR #8692, opened 2026-03-13
[MOD-10090] Port Optional iterator - Optimized

ThierryBerger:
This PR focuses on **Rust** side for the migration of the optimized optional iterator.

- next step: https://github.com/RediSearch/RediSearch/pull/8915

Benchmarks

| Scenario | C | Rust | Speedup |
|----------|---|------|---------|
| **Read Sparse** (1K docs, out-of-range child) | 4.59 µs | 2.69 µs | **1.71× faster** |
| **Read Dense** (1M docs, 90% child coverage) | 7.06 ms | 3.71 ms | **1.90× faster** |
| **Read Full** (1M docs, 100% child coverage) | 5.82 ms | 3.10 ms | **1.88× faster** |

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

#### Release Notes

- [ ] This PR requires release notes
- [x] This PR does not require release notes

If a release note is required (bug fix / new feature / enhancement), describe the **user impact** of this PR in the title.  

---

> [!NOTE]
> **Medium Risk**
> Adds a new core iterator implementation that changes query iteration semantics/performance around `Optional` handling and revalidation; correctness depends on edge cases like EOF/maxDocId and child aborts, though it’s covered by extensive new tests and benchmarks.
> 
> **Overview**
> Ports the *optimized* optional iterator to Rust via new `OptionalOptimized`, which iterates over existing document IDs using a wildcard iterator instead of scanning `1..=maxDocId`, returning real child hits with weight and virtual results otherwise.
> 
> Adds extensive integration tests covering `read`, `skip_to`, `rewind`/EOF, sparse wildcard behavior, error propagation, and `revalidate` scenarios (including child abort and wildcard moving to EOF), plus new Criterion benchmarks (and a small C-FFI helper) to compare C vs Rust performance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b0ead2854a1d87c69ec175de518d0b3a393764b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #8790, opened 2026-03-23
[MOD-14487] not iterator: implement not optimized variant in Rust

gdesmott:

Rust implementation and tests of the optimized variant of the not iterator.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Adds a new `NotOptimized` query iterator that changes how NOT results are enumerated (driven by `existingDocs`/wildcard rather than scanning doc IDs), which can affect query correctness/perf around edge cases like EOF, `skip_to`, timeouts, and revalidation after index GC.
> 
> **Overview**
> Implements a new Rust `NotOptimized` iterator that computes NOT by iterating over the wildcard/"existing docs" inverted index and filtering out matches from a child iterator, instead of scanning sequential doc IDs.
> 
> Adds comprehensive integration tests covering `read`, `skip_to`, `rewind`, timeout propagation/timeout-context behavior, and `revalidate` semantics under wildcard movement/abort and child abort/move cases, plus wires the new module into `lib.rs` and the integration test harness.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8dc9d7d0a16ded9b4424b52008c13a26ff33c34d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/rqe_iterators/src/not_optimized.rs
+++ b/src/redisearch_rs/rqe_iterators/src/not_optimized.rs
@@ -152,8 +152,19 @@ where
                     return Ok(false);
                 }
             } else {
-                // Case 3: Child is behind, skip it forward to catch up.
-                self.child.skip_to(wcii_last)?;
+                // Case 3: Child is behind, read it forward to catch up.
+                //
+                // We use a read loop rather than `skip_to` because the
+                // child almost always needs only a single read to reach
+                // or pass `wcii_last`. The only scenario where the child
+                // lags behind is when GC has removed a doc ID from the
+                // wildcard inverted index but not yet from the child's
+                // index — and even then the gap is typically tiny.
+                // `read` is cheaper than `skip_to`, so the loop is
+                // faster in the common case.
+                while !self.child.at_eof() && self.child.last_doc_id() < wcii_last {
+                    self.child.read()?;
+                }
             }
             self.check_timeout()?;
         }


src/redisearch_rs/rqe_iterators/tests/integration/not_optimized.rs
```diff
@@ -0,0 +1,824 @@
...
+
+use crate::utils::{Mock, MockIteratorError, MockVec};
+
+/// Holds an [`InvertedIndex`] and a [`MockContext`] so a
+/// [`Wildcard`](rqe_iterators::inverted_index::Wildcard) iterator can be
+/// created from them.
+struct WildcardHelper {
+    ii: InvertedIndex<DocIdsOnly>,
+    mock_ctx: MockContext,
+}
```
ThierryBerger: - We may want to consolidate this pattern with #8692 in a follow up (or linked PR)

exact lines: https://github.com/RediSearch/RediSearch/pull/8692/changes#diff-92d7fd0efc449dfd86e9b4092dc7350e9f7a16caf32187095ec5f76f8dc95dbcR29-R32

src/redisearch_rs/rqe_iterators/tests/integration/not_optimized.rs
```diff
@@ -0,0 +1,824 @@
...
+        rc = it.read();
+    }
+    assert!(
+        matches!(rc, Err(RQEIteratorError::TimedOut)),
+        "Expected timeout, got {rc:?}"
+    );
+}
+
+#[test]
+fn child_timeout_on_skip_to() {
```
ThierryBerger, line 491: should all timeout tests have 
`#[cfg_attr(miri, ignore = "Too slow to be run under miri.")]` ?

ah this one doesn't have a delay 🤦‍♂️

src/redisearch_rs/rqe_iterators/tests/integration/not_optimized.rs
gdesmott: let's merge this now then: https://github.com/RediSearch/RediSearch/pull/8822

src/redisearch_rs/rqe_iterators/src/not_optimized.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    /// Wrapper around [`TimeoutContext::check_timeout`] that sets `forced_eof`
+    /// on timeout.
+    #[inline(always)]
+    fn check_timeout(&mut self) -> Result<(), RQEIteratorError> {
+        let Some(result) = self.timeout_ctx.as_mut().map(|ctx| ctx.check_timeout()) else {
+            return Ok(());
+        };
+        if matches!(result, Err(RQEIteratorError::TimedOut)) {
+            self.forced_eof = true;
```
GuyAv46, line 107: Why do we need this? The current C impl doesn't do it

src/redisearch_rs/rqe_iterators/src/not_optimized.rs
```diff
@@ -0,0 +1,322 @@
...
+/// * `I` - The child iterator type whose results are negated.
+pub struct NotOptimized<'index, W, I> {
+    /// The wildcard iterator over all existing documents.
+    wcii: W,
+    /// The child iterator whose results are negated.
+    child: MaybeEmpty<I>,
+    /// The maximum document ID (used as upper bound guard).
+    max_doc_id: t_docId,
+    /// Sticky EOF flag, set on timeout or when iteration completes.
+    forced_eof: bool,
```
GuyAv46, line 50: In the C impl, we use `base->atEOF` as the flag, but we set it only on EOF and not on timeouts.
Could you explain why we want to force EOF on timeout in this implementation?

src/redisearch_rs/rqe_iterators/src/not_optimized.rs
```diff
@@ -0,0 +1,322 @@
...
+                // Case 1: The wildcard document is not in the child.
+                self.result.doc_id = wcii_last;
+                return Ok(true);
+            } else if wcii_last == self.child.last_doc_id() {
+                // Case 2: Both iterators at the same position, advance both.
+                self.child.read()?;
+                if self.wcii.read()?.is_none() {
+                    self.forced_eof = true;
+                    return Ok(false);
+                }
```
GuyAv46: Small duplication (with pre-loop logic) - consider adding a helper?

src/redisearch_rs/rqe_iterators/src/not_optimized.rs
```diff
@@ -0,0 +1,322 @@
...
+            return Ok(None);
+        }
+        if doc_id > self.max_doc_id {
+            self.forced_eof = true;
+            return Ok(None);
+        }
+
+        // Skip wcii to docId.
+        let wcii_outcome = self.wcii.skip_to(doc_id)?;
+        if wcii_outcome.is_none() {
```
GuyAv46: Consider inlining (`wcii_outcome` is not used elsewhere)

src/redisearch_rs/rqe_iterators/src/not_optimized.rs
```diff
@@ -0,0 +1,322 @@
...
+        if matches!(result, Err(RQEIteratorError::TimedOut)) {
+            self.forced_eof = true;
+        }
+        result
+    }
+
+    /// Check whether the child iterator does **not** contain a document
+    /// with the given `doc_id`.
+    #[inline(always)]
+    fn child_does_not_have(&self, doc_id: t_docId) -> bool {
```
GuyAv46: Consider renaming, maybe something like `child_is_ahead` or `child_is_ahead_or_depleted`, so it's clear we don't perform an actual read, and may return `false` even if the child doesn't have this ID (but an actual read is needed to figure this out)

src/redisearch_rs/rqe_iterators/src/not_optimized.rs
```diff
@@ -0,0 +1,322 @@
...
+                }
+            }
+        }
+
+        // Determine Found vs NotFound based on whether we're at the exact target.
+        if self.result.doc_id == doc_id {
+            Ok(Some(SkipToOutcome::Found(&mut self.result)))
+        } else {
+            Ok(Some(SkipToOutcome::NotFound(&mut self.result)))
+        }
```
GuyAv46: This branching follows the C impl exactly, but I think it can be changed a bit to be more readable in Rust.
1. Case 3 actually handles internally cases 1+2 again
2. Final `Determine Found vs NotFound` is relevant only for cases 1 and 3.b.

Consider reordering, or adding a TODO for that (and refactor this function in a follow-up, to prove the tests still pass after the refactor)

src/redisearch_rs/rqe_iterators/src/not_optimized.rs
gdesmott: Leftovers from a previous version. Good catch!

src/redisearch_rs/rqe_iterators/src/not_optimized.rs
gdesmott: It's no longer set with the other changes above.

src/redisearch_rs/rqe_iterators/src/not_optimized.rs
gdesmott: Good point, I restructured this function.

gdesmott:
@GuyAv46 thanks for your review. I think I fixed all your comments.

src/redisearch_rs/rqe_iterators/src/not_optimized.rs
```diff
@@ -151,10 +152,8 @@ where
                     return Ok(false);
                 }
             } else {
-                // Case 3: Child is behind, advance it until it catches up.
-                while !self.child.at_eof() && self.child.last_doc_id() < wcii_last {
-                    self.child.read()?;
-                }
+                // Case 3: Child is behind, skip it forward to catch up.
+                self.child.skip_to(wcii_last)?;
```
GuyAv46: Here, we used to have a while loop because it's very rare that the child needs more than a single read to get beyond the wildcard.

It can only happen if the GC cleaned some ID from the wildcard inverted index and not yet from the child (any valid ID should also be returned from the wildcard iterator)

Since `read` is usually simpler than `skip_to`, we decided to read here.

Consider keeping this behavior

GuyAv46: Maybe we should document that with a comment

gdesmott: Good to know, thanks for the context.

I changed that.

PR #8809, opened 2026-03-24
[MOD-13606] CI: Check glibc version

memark:
Check the required and provided glibc versions in CI for the actual platform. Helps catch (weird) errors before they occur. Most useful when adding new platforms or enabling LTO for existing ones.

Successful run
https://github.com/RediSearch/RediSearch/actions/runs/24097793050

#### Release Notes

### DIFF
--- a/.github/actions/validate-glibc-version/action.yml
+++ b/.github/actions/validate-glibc-version/action.yml
@@ -12,8 +12,11 @@ runs:
       run: |
         # Validate glibc version
 
+        # Capture ldd output once (|| true guards against set -e)
+        ldd_output=$(ldd --version 2>&1 || true)
+
         # Skip musl-based systems (e.g. Alpine) — no glibc to validate
-        if ldd --version 2>&1 | grep -qi musl; then
+        if grep -qi musl <<< "$ldd_output"; then
           echo ">>> Skipping: musl-based system (no glibc)"
           exit 0
         fi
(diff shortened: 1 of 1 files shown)


.github/actions/validate-glibc-version/action.yml
```diff
@@ -0,0 +1,39 @@
...
+
+        # All GLIBC versions the binary requires (deduplicated)
+        binary_versions=$(nm -D "$BINARY" | grep -oP 'GLIBC_\K[0-9.]+' | sort -Vu)
+
+        # Max GLIBC version available on this platform
+        platform_max=$(ldd --version 2>&1 | grep -oP '[0-9]+\.[0-9]+' | head -1)
+        if [ -z "$platform_max" ]; then
+          echo ">>> Skipping: could not determine glibc version (musl?)"
+          exit 0
+        fi
```
LukeMathWalker, line 38: I'd rather skip with an explicit OS check than risking other platforms falling through without us realizing.

memark: I updated this with an explicit `musl` check instead. I think it makes more sense as it catches all platforms where `glibc` is expected to be missing, without having to hardcode an OS list.

.github/actions/validate-glibc-version/action.yml
```diff
@@ -0,0 +1,46 @@
+name: Validate glibc version
+description: |
+  Checks that the built binary doesn't require a newer glibc than the platform provides.
+  Skips gracefully on non-glibc platforms (e.g. musl/Alpine).
+  The action fails for unknown reasons on ubuntu:focal.
```
daxhuiberts, line 5: 🥹

memark: _Probably_ due to some old bug in old Bash. I seriously spent hours on this... But we're soon dropping Focal anyway.

PR #8838, opened 2026-03-25
Amortize allocations/deallocations of RSValues via a mempool

LukeMathWalker:

Re-introduce a memory pool for RSValue, matching the original C version.

1. MOD-...
2. #...

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches `SharedRsValue` allocation/drop behavior and introduces thread-local pooling, which could affect memory lifetimes and drop semantics if misused. Scope is contained to RSValue sharing, but bugs here could cause leaks or panics in hot paths.
> 
> **Overview**
> Replaces `SharedRsValue` heap allocation/deallocation churn with a **thread-local recycling pool** for `Arc<RsValue>` allocations.
> 
> `SharedRsValue::new` now pulls from the pool when available, and `Drop` detects when it is the last reference and returns the allocation to the pool (clearing the stored value first), otherwise it falls back to normal refcount decrement. The pool is capped (`MAX_POOL_SIZE`) and uses TLS `try_with` to avoid panics during thread-local destruction.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5da7e2d520eb3dd958ef6df95d12379ab9cb7e39. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/value/src/pool.rs
+++ b/src/redisearch_rs/value/src/pool.rs
@@ -44,15 +44,20 @@ thread_local! {
 
 /// Get a recycled `Arc<RsValue>` with the given value, or allocate a new one.
 pub(crate) fn pool_get(value: RsValue) -> Arc<RsValue> {
-    POOL.with_borrow_mut(|pool| {
-        if let Some(mut arc) = pool.0.pop() {
-            // strong_count == 1 (pool held the only reference), so get_mut always succeeds.
-            *Arc::get_mut(&mut arc).unwrap() = value;
-            arc
-        } else {
-            Arc::new(value)
-        }
-    })
+    // Use `try_with` to be thread-local destruction safe in the rare case
+    // that `pool_get` is called during thread-local destruction.
+    if let Some(mut arc) = POOL
+        .try_with(|pool| pool.borrow_mut().0.pop())
+        .ok()
+        .flatten()
+    {
+        // `strong_count == 1` (pool held the only reference), so `get_mut`
+        // always succeeds.
+        *Arc::get_mut(&mut arc).unwrap() = value;
+        arc
+    } else {
+        Arc::new(value)
+    }
 }
 
 /// Return an `Arc<RsValue>` to the pool for recycling, or drop it if pool is full.
@@ -67,10 +72,9 @@ pub(crate) fn pool_release(mut arc: Arc<RsValue>) {
 
     // This function is called from `SharedRsValue::drop`. During thread shutdown,
     // thread-local destruction order is unspecified, so the `POOL` TLS may already
-    // be destroyed when a `SharedRsValue` held in another thread-local (or
-    // transitively via `RsValue::Ref`, `Trio`, `Array`, etc.) is dropped.
-    // `LocalKey::with` (used by `with_borrow_mut`) would panic in that case, and a
-    // panic inside `Drop` during thread shutdown aborts the process.
+    // be destroyed when a `SharedRsValue` held (directly or transitively) in another
+    // thread-local is dropped. `LocalKey::with` (used by `with_borrow_mut`) would panic
+    // in that case, and a panic inside `Drop` during thread shutdown aborts the process.
     //
     // We therefore use `try_with` + `borrow_mut` instead: if the pool is already
     // destroyed, `try_with` returns `Err` and we silently fall through, letting the


src/redisearch_rs/value/src/shared.rs
```diff
@@ -161,8 +165,25 @@ impl Drop for SharedRsValue {
...
+                crate::pool::pool_release(arc);
+            } else {
+                // Other references exist — just decrement the refcount.
+                //
+                // In the concurrent drop case where two threads both read a count
+                // > 1, both take this branch. Arc's internal atomics handle this
+                // correctly — the last `drop` deallocates. We miss the pool
+                // optimization, but there is no unsoundness.
+                drop(arc);
+            }
```
daxhuiberts, line 184: There is a race condition when two `Arc`s try to drop at the same time, they both read `strong_count == 2` and both call `drop(arc)`. This won't break anything though, it only won't be returned to the pool, but deallocated instead.

src/redisearch_rs/value/src/pool.rs
```diff
@@ -0,0 +1,73 @@
...
+/// safe because the `Arc` allocation is globally valid regardless of which
+/// thread holds it.
+///
+/// Arcs stored in the pool always have `strong_count == 1` (the pool itself is
+/// the sole owner). No other thread can hold a reference to a pooled Arc, so
+/// `Arc::get_mut` on a pooled entry is guaranteed to succeed.
+struct Pool(Vec<Arc<RsValue>>);
+
+thread_local! {
+    static POOL: RefCell<Pool> = const { RefCell::new(Pool(Vec::new())) };
```
daxhuiberts, line 42: Should we do anything with the `MAX_POOL_SIZE` and the capacity of the vec?

That's not possible in a const setting, but does work with RefCell in a lazy way on first access.

LukeMathWalker: I considered pre-warming it, but I've preferred to keep it close to the C implementation (i.e., start empty).

src/redisearch_rs/value/src/shared.rs
LukeMathWalker: Yup, that's what the second comment captures, no?

daxhuiberts: D'oh! image

src/redisearch_rs/value/src/pool.rs
```diff
@@ -0,0 +1,73 @@
...
+            Arc::new(value)
+        }
+    })
+}
+
+/// Return an `Arc<RsValue>` to the pool for recycling, or drop it if pool is full.
+///
+/// # Panics
+///
+/// Panics if `strong_count != 1` (i.e. the caller is not the sole owner).
```
memark: Maybe `> 1` ?

LukeMathWalker: Better indeed! Changed it.

src/redisearch_rs/value/src/pool.rs
```diff
@@ -0,0 +1,74 @@
...
+    // Clear the value to release any owned resources (strings, arrays, etc.)
+    // `get_mut` succeeds because the caller is the sole owner (see `Pool` docs).
+    *Arc::get_mut(&mut arc).unwrap() = RsValue::Undefined;
+
+    POOL.with_borrow_mut(|pool| {
+        if pool.0.len() < MAX_POOL_SIZE {
+            pool.0.push(arc);
+        }
+        // else: drop the Arc, deallocating it
+    });
```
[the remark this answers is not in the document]
daxhuiberts: So if I understand this correctly: When something panics and unwinds, `SharedRsValues` might get dropped, but also this thread local pool can already be released? Is that actually possible? What other situation might happen that `pool_release` and thread exiting happen at the same time?

A `SharedRsValue` like a `Ref`, `Trio` is `Array` is never held in another thread-local pool. Only things added to a pool are `SharedRsValue`s reset to `Undefined` before the pool is accessed and is the sole owner of that `SharedRsValue`.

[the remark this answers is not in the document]
LukeMathWalker: If I understand correctly, the scenario is the following:
- Thread is shutting down
- `POOL` thread-local must be dropped
- There might be other another thread-local holding an `Arc<RsValue>` instance, let's call it `OTHER`

If `POOL` is dropped before `OTHER`, then `pool_release` is invoked by the `Drop` impl of the values held by `OTHER` and `with_borrow_mut` panics.

It might not be a relevant scenario right now, but since the fix is relatively straight-forward, I think it's worth covering it.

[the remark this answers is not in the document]
daxhuiberts: Oh, I understand. When A `SharedRsValue` is ALSO stored in a thread-local (directly or transitively), then this can happen.

Will change it to `POOL.try_with`.

src/redisearch_rs/value/src/pool.rs
```diff
@@ -65,7 +65,18 @@ pub(crate) fn pool_release(mut arc: Arc<RsValue>) {
     // `get_mut` succeeds because the caller is the sole owner (see `Pool` docs).
     *Arc::get_mut(&mut arc).unwrap() = RsValue::Undefined;
 
-    POOL.with_borrow_mut(|pool| {
+    // This function is called from `SharedRsValue::drop`. During thread shutdown,
+    // thread-local destruction order is unspecified, so the `POOL` TLS may already
+    // be destroyed when a `SharedRsValue` held in another thread-local (or
+    // transitively via `RsValue::Ref`, `Trio`, `Array`, etc.) is dropped.
+    // `LocalKey::with` (used by `with_borrow_mut`) would panic in that case, and a
+    // panic inside `Drop` during thread shutdown aborts the process.
```
LukeMathWalker: I'm not sure about the held transitively—I don't think that's a real concern. I would limit the comment at the "second thread-local" scenario.l

daxhuiberts: I think the 'transitively' part is valuable, but the example should be that it can be contained in any other struct or something.

LukeMathWalker: In that sense, yes, but it has to be a struct held by another thread-local, that's the key part.

src/redisearch_rs/value/src/pool.rs
```diff
@@ -0,0 +1,85 @@
...
+pub(crate) fn pool_get(value: RsValue) -> Arc<RsValue> {
+    POOL.with_borrow_mut(|pool| {
+        if let Some(mut arc) = pool.0.pop() {
+            // strong_count == 1 (pool held the only reference), so get_mut always succeeds.
+            *Arc::get_mut(&mut arc).unwrap() = value;
+            arc
+        } else {
+            Arc::new(value)
+        }
+    })
```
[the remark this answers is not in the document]
daxhuiberts: @LukeMathWalker If somebody is CREATING a new `SharedRsValue` inside a drop from a thread-local destruction, then that might happen. Is this really worth handling as well?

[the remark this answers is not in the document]
LukeMathWalker: Were it complex, it'd say, "Don't bother". But since it's just a method switch 🤷🏻

PR #8855, opened 2026-03-26
MOD-14390: Limit disk TEXT field query API

raz-mon:
Limits the query API available for TEXT fields for disk indexes.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Introduces new validation failures for `QN_PREFIX`, `QN_WILDCARD_QUERY`, `QN_FUZZY`, and `QN_LEXRANGE` when Flex/disk-mode validation is enabled, which can break existing queries on disk indexes. Changes are localized to query validation and error-code plumbing but affect user-facing behavior and error reporting.
> 
> **Overview**
> **Flex (disk) indexes now explicitly reject trie-expanded TEXT query types.** Query validation (`QAST_CheckIsValid`/`QueryNode_CheckIsValid`) adds disk-mode checks that return an error for `Prefix`, `Wildcard pattern`, `Fuzzy`, and `Lexrange` query nodes when `SearchDisk_IsEnabledForValidation()` is true.
> 
> Adds a dedicated error code `QUERY_ERROR_CODE_FLEX_UNSUPPORTED_QUERY` (C header + Rust `QueryErrorCode::FlexUnsupportedQuery`) with a new `SEARCH_FLEX_UNSUPPORTED_QUERY` prefix, and adjusts validation short-circuiting so disk indexes are *always* validated even when other validations would be skipped.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 184d6355982c65fcd77b8085ff6538b51807b4f2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/query.c
+++ b/src/query.c
@@ -1729,7 +1729,7 @@ static inline bool QueryNode_ValidateToken(QueryNode *n, IndexSpec *spec, RSSear
 
 // Helper function to validate that trie-based query types are not used on disk indexes.
 // Returns REDISMODULE_ERR if the query type is not supported on disk indexes, REDISMODULE_OK otherwise.
-static int QueryNode_ValidateNotDiskUnsupported(const char *queryTypeName,
+static int validateQueryNotDisk(const char *queryTypeName,
   QueryError *status) {
   if (SearchDisk_IsEnabledForValidation()) {
     QueryError_SetWithoutUserDataFmt(status, QUERY_ERROR_CODE_FLEX_UNSUPPORTED_QUERY,
@@ -1805,16 +1805,16 @@ static int QueryNode_CheckIsValid(QueryNode *n, IndexSpec *spec, RSSearchOptions
       }
       break;
     case QN_PREFIX:
-      res = QueryNode_ValidateNotDiskUnsupported("Prefix", status);
+      res = validateQueryNotDisk("Prefix", status);
       break;
     case QN_WILDCARD_QUERY:
-      res = QueryNode_ValidateNotDiskUnsupported("Wildcard pattern", status);
+      res = validateQueryNotDisk("Wildcard pattern", status);
       break;
     case QN_FUZZY:
-      res = QueryNode_ValidateNotDiskUnsupported("Fuzzy", status);
+      res = validateQueryNotDisk("Fuzzy", status);
       break;
     case QN_LEXRANGE:
-      res = QueryNode_ValidateNotDiskUnsupported("Lexrange", status);
+      res = validateQueryNotDisk("Lexrange", status);
       break;
     case QN_NOT:
     case QN_OPTIONAL:


src/query.c
```diff
@@ -1727,6 +1727,18 @@ static inline bool QueryNode_ValidateToken(QueryNode *n, IndexSpec *spec, RSSear
...
+// Returns REDISMODULE_ERR if the query type is not supported on disk indexes, REDISMODULE_OK otherwise.
+static int QueryNode_ValidateNotDiskUnsupported(QueryNodeType type, const char *queryTypeName,
+  IndexSpec *spec, QueryError *status) {
+  if (spec->diskSpec) {
+    QueryError_SetWithoutUserDataFmt(status, QUERY_ERROR_CODE_FLEX_UNSUPPORTED_QUERY,
+      "%s queries are not supported on Flex indexes", queryTypeName);
+    return REDISMODULE_ERR;
+  }
+  return REDISMODULE_OK;
+}
```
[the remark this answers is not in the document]
JoanFM: @raz-mon to clean up

src/query.c
```diff
@@ -1727,6 +1727,18 @@ static inline bool QueryNode_ValidateToken(QueryNode *n, IndexSpec *spec, RSSear
   return true;
 }
 
+// Helper function to validate that trie-based query types are not used on disk indexes.
+// Returns REDISMODULE_ERR if the query type is not supported on disk indexes, REDISMODULE_OK otherwise.
+static int QueryNode_ValidateNotDiskUnsupported(QueryNodeType type, const char *queryTypeName,
+  IndexSpec *spec, QueryError *status) {
+  if (spec->diskSpec) {
```
JoanFM: can we use SearchDisk_IsEnableForValidation? This way we can add test to test_flex_validation?

src/buffer/CMakeLists.txt
```diff
@@ -1,6 +1,5 @@
-# Build the `buffer` module as a standalone static library
-# This is a temporary requirement to allow us to benchmark the
-# Rust implementation of the encoders/decoders against the original C implementation.
+# Build the `buffer` module as an object library for inclusion in libredisearch_all.a
+# This is required because the main CMakeLists.txt uses $<TARGET_OBJECTS:buffer>
 file(GLOB BUFFER_SOURCES "buffer.c")
-add_library(buffer STATIC ${BUFFER_SOURCES})
+add_library(buffer OBJECT ${BUFFER_SOURCES})
```
raz-mon: Added as part of https://github.com/RediSearch/RediSearch/commit/18f07b2eed4a865145724f3a8ee0a133ae19e85a, in https://github.com/RediSearch/RediSearch/pull/8864 (to be merged shortly).

src/query.c
```diff
@@ -1792,15 +1804,23 @@ static int QueryNode_CheckIsValid(QueryNode *n, IndexSpec *spec, RSSearchOptions
...
+      break;
+    case QN_WILDCARD_QUERY:
+      res = QueryNode_ValidateNotDiskUnsupported("Wildcard pattern", status);
+      break;
+    case QN_FUZZY:
+      res = QueryNode_ValidateNotDiskUnsupported("Fuzzy", status);
+      break;
+    case QN_LEXRANGE:
+      res = QueryNode_ValidateNotDiskUnsupported("Lexrange", status);
+      break;
```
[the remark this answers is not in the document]
raz-mon: Actually not, so this is fine for now.

src/redisearch_rs/query_error/src/lib.rs
```diff
@@ -485,6 +486,11 @@ impl QueryErrorCode {
                 default_msg: c"Unsupported argument for Redis Flex",
                 default_full_msg: c"SEARCH_FLEX_UNSUPPORTED_ARGUMENT Unsupported argument for Redis Flex",
             },
+            Self::FlexUnsupportedQuery => ErrorCodeStrings {
+                prefix: c"SEARCH_FLEX_UNSUPPORTED_QUERY ",
+                default_msg: c"Unsupported query type for Flex index",
```
kei-nan, line 497: Maybe we should give different error codes to parts in the query we don't support?
Right now the error message doesn't help so much to the end user in fixing the query.

kei-nan: If we override the error message then not sure if the error code has a lot of merit.

src/query.c
```diff
@@ -1821,7 +1841,14 @@ static int QueryNode_CheckIsValid(QueryNode *n, IndexSpec *spec, RSSearchOptions
...
 int QAST_CheckIsValid(QueryAST *q, IndexSpec *spec, RSSearchOptions *opts, QueryError *status) {
-  if (!q || !q->root ||
+  if (!q || !q->root) {
+    return REDISMODULE_OK;
+  }
+
+  // Always validate disk indexes (for unsupported query types like prefix/fuzzy/wildcard/lexrange)
+  // For non-disk indexes, skip validation if there's no TEXT/TAG field that doesn't index empty
+  // and no JSON spec with undefined order
+  if (!SearchDisk_IsEnabledForValidation() &&
```
kei-nan, line 1851: Maybe it is worth having a boolean variable instead of a complicated if line.

src/redisearch_rs/query_error/src/lib.rs
raz-mon: Why not? We have the query type there as well.
See the `QueryNode_ValidateNotDiskUnsupported` function and its callers.
I agree about the second comment but that's how we do it today, matches our mechanism that you can set only a code, that won't be changed here.

src/query.c
raz-mon: It ends up the same, I don't think it will help much.
Notice that I already took this apart from the above two conditions in order for it to be simpler.

src/query.c
```diff
@@ -1727,6 +1727,18 @@ static inline bool QueryNode_ValidateToken(QueryNode *n, IndexSpec *spec, RSSear
   return true;
 }
 
+// Helper function to validate that trie-based query types are not used on disk indexes.
+// Returns REDISMODULE_ERR if the query type is not supported on disk indexes, REDISMODULE_OK otherwise.
+static int QueryNode_ValidateNotDiskUnsupported(const char *queryTypeName,
```
Itzikvaknin:
```suggestion
static int QueryNode_ValidateDiskSupported(const char *queryTypeName,
```

Itzikvaknin: Im not sure about this function - is it the current structure? 
It has `QueryNode_*` prefix but doesnt get `QueryNode_/QueryNode*` as an argument
I dont see any difference from a wrapper taking two strings and an error code

raz-mon: You're right that it's not a classic QueryNode_* function, but notice the error it sets.
It's part of the QueryNode validation and initially got a node as an argument but actually unnecessarily.
I'll change it to `validateQueryNotDisk`.

PR #8856, opened 2026-03-26
document reasons why miri is not run for some tests

gdesmott:

Adding a proper ignore reason so it should be easier to notice when those can be enabled (like when a ffi call is removed).

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to test cfg attributes and ignore messages, with no production logic modifications. Main impact is which tests execute under Miri (and clearer skip diagnostics), which could affect CI signal but not runtime behavior.
> 
> **Overview**
> Updates many Rust tests to use `#[cfg_attr(miri, ignore = "…")]` with *specific* reasons (unsupported extern statics like `RedisModule_Alloc/Free`, missing syscalls like `open`, unsupported clocks, or long runtimes) instead of generic ignores or `#[cfg(not(miri))]` gating.
> 
> This makes Miri skips more actionable and consistent across FFI-heavy areas (`value_ffi`, `result_processor`, `rlookup`, iterators, sorting vector, etc.), and also enables at least one previously Miri-disabled test (`MetricSortedById` drop/handle invalidation) to run under Miri.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1e89d2e69c37182a479c629f990508ec20dd7d98. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/c_wrappers/field_spec/tests/tests.rs
+++ b/src/redisearch_rs/c_wrappers/field_spec/tests/tests.rs
@@ -17,7 +17,10 @@ use field_spec::{FieldSpec, FieldSpecBuilder};
 use pretty_assertions::assert_eq;
 
 #[test]
-#[cfg_attr(miri, ignore = "miri does not support FFI functions")]
+#[cfg_attr(
+    miri,
+    ignore = "extern static `RedisModule_Alloc` is not supported by Miri"
+)]
 fn field_name_and_path() {
     let name = c"name";
     let path = c"path";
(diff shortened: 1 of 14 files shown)


daxhuiberts:
Nice!

Is there a lint what will ensure that a miri attribute always has an ignore label?

Is it also possible to see if a miri tag is not needed anymore? Similar to how `expect` vs `allow` works?

gdesmott:
> Is there a lint what will ensure that a miri attribute always has an ignore label?

I checked but unfortunately there is not.

> Is it also possible to see if a miri tag is not needed anymore? Similar to how expect vs allow works?

No, the only way is actually run the test as it fails at runtime.

Issue #925, opened 2019-10-03
How do we detect whether it is "oss redisearch" or "redisearch enterprise"?


gnaneshr:
Is there a command that only "redisearch enterprise" supports? so that we can distinguish between oss redisearch and redisearch enterprise.

We are working on an application that connects to redis instances, but there's no way to check whether it's "oss redisearch" or "redisearch enterprise".

MeirShpilraien:
hey @gnaneshr 
Why does it matter to you application if it connects to oss or enterprise redis? If you find that those are not operate the same I would like to know about it.

nishantgeorge:
Hey @MeirShpilraien 

@gnaneshr and I are working on [RedisInsight](https://redislabs.com/redisinsight). 

The naming of index keys in RediSearch on Redis Enterprise seems to be different from the way it's done on Redis OSS. On RE, there seems to be an extra hash tag appended to the key name (e.g. `idx:myindex{JA5}`, the `{JA5}` part is not part of the index name, but exists on the key on RE).

This is fine when the client knows the index name beforehand, but it complicates listing index keys, which we are doing now by `SCAN`ing for `idx:*` keys. Currently, we are working around it by extracting out the end part and running `FT.INFO` to check if that index exists.

mnunberg:
I suppose we could add a DEBUG subcommand, or maybe FT.VERSION which should provide greater more detailed output

MeirShpilraien:
A better way would be to expose a command to get all indexes..

PR #8956, opened 2026-04-07
MOD-9881 Integrate VecSim shared SVS thread pool with RediSearch worker pool

meiravgri:
RediSearch-side integration for the VecSim shared SVS thread pool refactor ([VectorSimilarity#925](https://github.com/RedisAI/VectorSimilarity/pull/925)).

1. **Current:** Each SVS index created its own thread pool sized at index creation time. RediSearch toggled `VecSim_SetWriteMode()` on worker pool activation/deactivation, but never notified VecSim when `WORKERS` was dynamically resized — so the SVS pool size would drift out of sync.
2. **Change:** Call the new `VecSim_UpdateThreadPoolSize()` API on pool creation and whenever `WORKERS` changes, replacing the old ad-hoc `VecSim_SetWriteMode()` toggling. Move VecSim allocator/log/timeout callback registration earlier in module init (before worker pool init) so VecSim can safely allocate shared thread-pool structures. Stop setting the deprecated `SVSParams::num_threads`. Support global (no per-index context) VecSim log messages.
3. **Outcome:** The shared SVS thread pool always matches the RediSearch worker count. All SVS indexes share the same pool and automatically see resizes — including deferred shrinks while SVS jobs are pending.

1. `deps/VectorSimilarity` — bumped submodule to include shared pool refactor
2. `src/module-init/module-init.c` — moved VecSim callback registration before worker pool init
3. `src/util/workers.c` — replaced `VecSim_SetWriteMode()` with `VecSim_UpdateThreadPoolSize()` on pool create and resize
4. `src/spec.c` — stopped setting deprecated `SVSParams::num_threads`
5. `src/vector_index.c` — added support for global VecSim log messages (NULL context)
6. `tests/pytests/test_vecsim_svs.py` — new tests for shared pool resizing and multi-index pool sharing
7. `src/aggregate/aggregate_exec.c` — added `BeforeSpecLock` debug sync point (fires before spec read lock acquisition on worker threads)

### Init order change

`VecSim_SetMemoryFunctions`, `VecSim_SetTimeoutCallbackFunction`, and `VecSim_SetLogCallbackFunction` are now called **before** `workersThreadPool_CreatePool`, because `VecSim_UpdateThreadPoolSize()` (called inside pool creation) may allocate VecSim internal structures (the shared SVS thread pool singleton).

### Worker pool integration

- `workersThreadPool_CreatePool()` now calls `VecSim_UpdateThreadPoolSize(worker_count)` after creating the pool.
- `workersThreadPool_SetNumWorkers()` calls `VecSim_UpdateThreadPoolSize(worker_count)` after adjusting the worker count.
- The old `VecSim_SetWriteMode(VecSim_WriteAsync)` / `VecSim_SetWriteMode(VecSim_WriteInPlace)` calls in activation/deactivation callbacks are removed — `VecSim_UpdateThreadPoolSize` handles both mode toggling and pool resizing atomically.

### SVS params

`SVSParams::num_threads` is no longer set during index creation. The field is deprecated in VecSim — setting it to a non-zero value emits a deprecation warning. Pool size is now controlled globally via `VecSim_UpdateThreadPoolSize()`.

### Logging

`VecSimLogCallback` now handles `ctx == NULL` for global VecSim log messages (e.g., shared thread pool resize events) that aren't tied to a specific index.

### New sync point: `BeforeSpecLock`
 
Added a debug-only sync point (`#ifdef ENABLE_ASSERT`) in `AREQ_Execute_Callback` that fires on worker threads **before** acquiring the spec read lock. This is needed by `test_resize_workers_during_pending_svs_jobs`: the existing `BeforeFirstRead` sync point fires **after** the spec read lock is acquired, so HSET commands (which need the spec write lock) would deadlock if all workers are blocked there. `BeforeSpecLock` avoids this by blocking workers before any lock is taken.

### New tests

- **`test_resize_workers_during_pending_svs_jobs`** — — blocks all workers at the `BeforeSpecLock` sync point (before the spec read lock is acquired), then adds vectors (SVS update jobs queue behind the blocked queries), shrinks `WORKERS`, and signals the sync point. Unlike `BeforeFirstRead`, workers at `BeforeSpecLock` do not hold the spec read lock, so the main thread can still process HSET commands without deadlocking.

- **`test_multiple_svs_indexes_share_pool`** — creates two SVS indexes, resizes `WORKERS`, and verifies both indexes see the new pool size and complete operations successfully.
- Updated existing `test_change_threads_*` tests to reflect that VecSim now tracks worker count changes in real time.

#### Mark if applicable

- [x] This PR introduces API changes

#### Release Notes
**Enhancement**: SVS-VAMANA vector indexes now share a single global thread pool that automatically adjusts its size when the `WORKERS` configuration is changed at runtime, replacing the previous per-index thread pool approach. This improves resource utilization and ensures all SVS indexes consistently reflect dynamic worker count changes.

---

> [!NOTE]
> **Medium Risk**
> Touches thread-pool sizing and vector indexing concurrency paths; mis-syncs or ordering issues could cause performance regressions or rare deadlocks/crashes under dynamic `WORKERS` reconfiguration.
> 
> **Overview**
> Ensures VecSim’s shared SVS thread pool stays in sync with RediSearch’s `WORKERS` thread pool by calling `VecSim_UpdateThreadPoolSize()` on worker-pool creation and every runtime resize, replacing the previous `VecSim_SetWriteMode()` toggling.
> 
> Moves VecSim allocator/timeout/log callback registration earlier in module init so shared-pool internals can allocate safely during pool setup, stops setting the deprecated SVS `num_threads` parameter, and expands `VecSimLogCallback` to support global (NULL-context) VecSim log messages.
> 
> Adds a new debug sync point (`BeforeSpecLock`) and extends SVS pytests to validate worker resizing during pending SVS jobs and that multiple SVS indexes observe and use the same resized shared pool.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 697e8d051ff81dd955c48fc50a33e46942e95b7c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dor-forer (dismissed):
## PR Review: MOD-9881 Integrate VecSim shared SVS thread pool with RediSearch worker pool

### Summary
This PR replaces the per-index SVS thread pool approach with a shared global thread pool managed via `VecSim_UpdateThreadPoolSize()`, keeping VecSim's pool size in sync with RediSearch's `WORKERS` config at all times.

### What changed (RediSearch side)
1. **`deps/VectorSimilarity`** — submodule bumped to include the shared pool refactor (VectorSimilarity#925)
2. **`src/module-init/module-init.c`** — VecSim callback registration (memory, timeout, log) moved **before** `workersThreadPool_CreatePool`, since the new `VecSim_UpdateThreadPoolSize()` may allocate internal VecSim structures (shared SVS thread pool singleton) during pool creation.
3. **`src/util/workers.c`** — Removed `VecSim_SetWriteMode()` calls from activation/deactivation callbacks. Added `VecSim_UpdateThreadPoolSize(worker_count)` in both `workersThreadPool_CreatePool()` and `workersThreadPool_SetNumWorkers()`. The new API handles both write-mode toggling (0 = in-place, >0 = async) and pool resizing atomically.
4. **`src/spec.c`** — Stopped setting `SVSParams::num_threads` (deprecated). Left at 0 to avoid deprecation warning.
5. **`src/vector_index.c`** — `VecSimLogCallback` now handles `ctx == NULL` for global VecSim log messages (shared thread pool events).
6. **`tests/pytests/test_vecsim_svs.py`** — Updated existing thread-change tests to expect real-time pool size tracking. Added two new tests.

### VectorSimilarity side (PR #925 context)
The VecSim side introduces:
- A `VecSimSVSThreadPoolImpl` singleton with `beginScheduledJob()`/`endScheduledJob()` for deferred resize during in-flight jobs.
- Renamed `setNumThreads`/`getNumThreads` → `setParallelism`/`getParallelism` (per-job concurrency) and `getThreadPoolCapacity` → `getPoolSize` (global pool size).
- `SVSMultiThreadJob::createScheduledJobs()` that calls `beginScheduledJob()` to snapshot pool size and defer shrinks.
- Deprecation warning if `SVSParams::num_threads` is set to non-zero.

### Review Findings

#### ✅ Init order change is correct
The Cursor Bugbot flagged an issue about memory functions not being registered before pool creation — **this was already fixed** in a subsequent commit. The current state of the PR has callbacks registered before `workersThreadPool_CreatePool()` (lines 166-172 in `module-init.c`), which is correct since `VecSim_UpdateThreadPoolSize()` may allocate via VecSim's memory functions.

#### ✅ Clean removal of `VecSim_SetWriteMode` 
All `VecSim_SetWriteMode()` calls are removed. `VecSim_UpdateThreadPoolSize()` handles both mode transitions and pool sizing. No stale references remain in the codebase.

#### ✅ `workersThreadPool_SetNumWorkers()` placement is correct
`VecSim_UpdateThreadPoolSize(worker_count)` is called **unconditionally** at the end of `SetNumWorkers()`, after the branching logic for grow/shrink/deactivate. This is correct because:
- VecSim's `UpdateThreadPoolSize` is idempotent when the size hasn't changed.
- It handles all transitions (0→N, N→0, N→M) internally.
- It's placed after the RediSearch thpool adjustments, so the RediSearch pool is already in its new state.

#### ✅ `VecSim_UpdateThreadPoolSize` in `CreatePool` is safe
Called after the pool is created (or after deactivation callback for worker_count=0). The VecSim singleton will be initialized with the correct size from the start.

#### ✅ Deprecated `num_threads` removal in `spec.c` is clean
The field is left at its default (0), avoiding the deprecation warning from VecSim. The comment explains the rationale.

#### ✅ NULL context logging is a reasonable extension
The `VecSimLogCallback` now handles `ctx == NULL` gracefully for global events. The log prefix `"VecSim: "` distinguishes these from per-index messages (`"vector index 'name' - "`). 

#### ✅ Tests are thorough
- **`test_resize_workers_during_pending_svs_jobs`**: Good concurrency test using `SYNC_POINT` to block workers, queue SVS jobs, then shrink workers mid-flight. Validates deferred resize via `beginScheduledJob`/`endScheduledJob`.
- **`test_multiple_svs_indexes_share_pool`**: Validates that two SVS indexes share the same pool and both see size changes immediately.
- Existing `change_threads` tests updated to expect `final_workers` instead of `initial_workers` for `NUM_THREADS`, which aligns with the new real-time notification behavior.

#### ⚠️ Minor observations (non-blocking)

1. **Double `VecSim_UpdateThreadPoolSize` on activation path**: In `CreatePool`, when `worker_count > 0`, `workersThreadPool_OnActivation()` is called first (which just logs now), then `VecSim_UpdateThreadPoolSize()` is called unconditionally. This is fine — just noting it's called once, not twice (the activation callback no longer calls VecSim).

2. **`workersThreadPool_SetNumWorkers` calls `VecSim_UpdateThreadPoolSize` even when count didn't change**: When `worker_count == curr_workers`, we skip the RediSearch thpool adjustment but still call `VecSim_UpdateThreadPoolSize`. This is harmless (idempotent) but adds a tiny unnecessary call. Very minor.

3. **Release notes checkbox**: The PR description already has a release note checked and written — the Cursor Bugbot comment about missing release notes is stale/incorrect.

### Verdict
The RediSearch integration is clean, correct, and well-tested. The init order fix addresses the allocator safety concern. The API migration from `SetWriteMode` to `UpdateThreadPoolSize` is complete with no leftover references. LGTM.

dor-forer (approved):

PR #9014, opened 2026-04-11
[MOD-13168] Add new SVS-VAMANA flow tests that delete documents while background training and subsequent background updates are running

alonre24:

1. Current: nothing covered deleting documents while an SVS-VAMANA tiered index is
   transferring its flat buffer into the backend — the window that the deletions journal in
   [VectorSimilarity#903](https://github.com/RedisAI/VectorSimilarity/pull/903), the fix for
   MOD-13168, is about.
2. Change: two flow tests (single-value hash field and multi-value JSON field) that delete
   documents while a transfer of their own vectors is *pending* and while it *runs*, in two
   phases — against the training of an empty backend, and against an update of the trained
   one — and then assert the index converged. Deletions are aimed at the batch being
   transferred, at the flat buffer, and at the backend. `TRAINING_THRESHOLD` is set to
   `DEFAULT_BLOCK_SIZE` to keep the runtime sane, so the tests take the same code path as
   the ticket's repro rather than its scale.
3. Outcome: internal-only (tests). Covers the correctness half of MOD-13168 — no deleted
   label left behind in the SVS backend — so that regression fails CI instead of surfacing
   as short KNN result sets in production. The ticket's other half, the main thread stalling
   on `mainIndexGuard` for the duration of training, is a latency property and is not
   asserted here.

#### How the coverage was verified

With VecSim mutated to skip applying the deletions journal, both tests fail on all four
detectors — the tiered label count, the live vector count, the KNN result count, and the
`assert_deletions_reached_the_backend` witness — and pass with it restored. The tests also
assert that they really did race a transfer, rather than relying on timing: the pending
window is made deterministic by pausing the worker pool, and the running window is checked
by `assert_transfer_did_not_complete` plus the marked-deleted witness, so a lost race fails
visibly instead of silently dropping the coverage. The whole `test_vecsim_svs.py` suite (46
tests) passes, and the two tests pass repeated runs with all 16 cores saturated.

1. MOD-13168

1. `tests/pytests/test_vecsim_svs.py` — `test_delete_during_background_indexing` and
   `test_delete_during_background_indexing_multi_value`, two thin cases over one
   parameterized driver.
2. `tests/pytests/vecsim_utils.py` — the tiered-transfer helpers the two tests share:
   `assert_svs_tiered_state` (the label and vector accounting of the tiered index),
   `assert_transfer_pending` and `assert_transfer_did_not_complete` (the witnesses that the
   deletions really did race a transfer), `assert_deletions_reached_the_backend` (that the
   raced deletions reached the backend index), and the KNN probes.
3. `tests/pytests/common.py` — `paused_workers` (schedule a transfer without letting it run,
   and resume the pool even if the block fails) and `workers_jobs_done`.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Low risk because changes are isolated to pytest coverage, but the new tests are timing/concurrency-sensitive and may introduce flakiness in CI if background indexing timing varies.
> 
> **Overview**
> Adds two new SVS-VAMANA flow tests in `tests/pytests/test_vecsim_svs.py` that delete documents while **background training** and later **background updates** are running, asserting `FT.INFO num_docs` updates correctly and deleted doc IDs never appear in `FT.SEARCH` KNN results.
> 
> Extends coverage to a multi-value JSON vector field scenario (skipped when JSON is unavailable) and updates imports to use `create_np_array_typed` for deterministic query vector creation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df8e39248c8b6897d051e8882283a6b85dd3d077. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/pytests/test_vecsim_svs.py
```diff
@@ -631,3 +632,188 @@ def test_gc_no_workers():
...
+        doc = f'{DEFAULT_DOC_NAME_PREFIX}{i + 1}'
+        conn.execute_command('DEL', doc)
+        deleted_docs.add(doc)
+
+    # Delete vectors added AFTER training started (from second batch)
+    for i in range(10):
+        doc = f'{DEFAULT_DOC_NAME_PREFIX}{training_threshold + 101 + i}'
+        conn.execute_command('DEL', doc)
+        deleted_docs.add(doc)
+
```
meiravgri, line 678: What are you trying to achieve here? Training/updates are automatically triggered
If you want to control it you probably need to pause the workers

tests/pytests/test_vecsim_svs.py
```diff
@@ -631,3 +632,188 @@ def test_gc_no_workers():
...
+    data_type = 'FLOAT32'
+    training_threshold = DEFAULT_BLOCK_SIZE
+    k = 10
+    query = create_random_np_array_typed(dim, data_type)
+
+    # Track deleted doc IDs for verification
+    deleted_docs = set()
+
+    # Phase 1: Insert vectors to trigger initial training
+    create_vector_index(env, dim, index_name=DEFAULT_INDEX_NAME, datatype=data_type, alg='SVS-VAMANA')
```
meiravgri: Default training threshold is `10 * DEFAULT_BLOCK_SIZE`, you probably work with the flat buffer for the entire test
I would add assertions on the expected backend index size where it is expected to be populated

meiravgri: ^^^ i was wrong (training TH is 1k with no compression. Though im not sure what value does the test have with no actual training?)
but the assertion suggestion holds.

alonre24: You're right, the purpose is to test with training. will adjust

tests/pytests/test_vecsim_svs.py
alonre24: I wanted a sanity test for the new mechanism - a new journal for the deletions that happen during background training/updates. The implementation is handling both deletions of the currently being indexed vectors and new vectors inserted afterwards (expected to be in the flat buffer)

alonre24: Adopted — the tests now pause the worker pool (`paused_workers`, in `common.py`) before crossing the threshold, so the transfer is scheduled but cannot run, and the deletions issued inside that block are deterministically racing a *pending* transfer of their own vectors. `assert_transfer_pending` checks that through `BACKGROUND_INDEXING` **and** the pool's queued-job count, since the flag alone conflates "queued", "running" and "the job mutex happened to be held".

The workers are then resumed and more docs deleted while the transfer actually runs — that second group is what exercises the deletions journal, and it no longer relies on wall-clock timing to overlap. Two witnesses turn a lost race into a visible failure rather than silently lost coverage: `assert_transfer_did_not_complete` (no background job finished while those deletions were issued) and `assert_deletions_reached_the_backend` (the backend's marked-deleted count grew, which is only possible if the transfer had already picked up a doc that was then deleted).

819391209, 60f4aba9b, 0488b9cda, 770dd3c57.

tests/pytests/test_vecsim_svs.py
alonre24: Both points addressed.

The index is now created with `COMPRESSION LeanVec4x8` and `TRAINING_THRESHOLD DEFAULT_BLOCK_SIZE`, so a real training transfer runs: phase 1 races the training of an empty backend, phase 2 races an update of the trained one.

And the backend accounting is asserted after each phase, instead of only `FT.INFO num_docs` — `assert_svs_tiered_state` checks the tiered index's `INDEX_LABEL_COUNT` against the live doc count, and `frontend INDEX_SIZE + backend INDEX_SIZE - backend NUMBER_OF_MARKED_DELETED` against `vectors_per_doc * live docs`. It reads the tiered index's own label count rather than summing the two sub-indexes': RediSearch adds a multi-value field one vector at a time, so a transfer can leave a doc live in both, and the sum would double-count it on a correct product.

3b74978bd, 0ebefb0bc, 60f4aba9b.

tests/pytests/test_vecsim_svs.py
```diff
@@ -631,3 +632,192 @@ def test_gc_no_workers():
...
+    env.assertEqual(res[0], k)
+    for doc in deleted_docs:
+        env.assertNotContains(doc, res)
+
+    wait_for_background_indexing(env, DEFAULT_INDEX_NAME, DEFAULT_FIELD_NAME)
+
+    # Phase 2: Add another batch, delete during background update
+    for i in range(100):
+        doc_id = n_docs + 20 + i + 1
+        conn.json().set(f'{DEFAULT_DOC_NAME_PREFIX}{doc_id}', '.', {'vecs': [[doc_id / 8.0] * dim for _ in range(per_doc)]})
```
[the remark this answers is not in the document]
alonre24: Fixed. Both variants now size each phase in *vectors* rather than docs — `docs_per_threshold = training_threshold // vectors_per_doc + 1` — so the multi-value test (5 vectors per doc) adds 205 docs / 1025 vectors in phase 2 and does cross `DEFAULT_BLOCK_SIZE`. It is no longer taken on trust either: `assert_transfer_pending` runs in phase 2 with the workers paused, and fails if no update was scheduled.

3b74978bd, 7eced7fa4.

tests/pytests/vecsim_utils.py
```diff
@@ -123,3 +125,109 @@ def wait_for_background_indexing(env, index_name, field_name, message=''):
...
+    holds it exclusively, so they block and complete as it finishes. BACKGROUND_INDEXING would not
+    do instead - a transfer clears its scheduled flag when it starts, and writes made while its
+    batch is still in the frontend schedule the *next* transfer."""
+    env.assertEqual(workers_jobs_done(env), jobs_done_before,
+                    message=f"{message}: a background job completed while the deletions were "
+                            f"issued, so they did not all race a running transfer")
+
+def assert_deletions_reached_the_backend(env, marked_deleted_before, message='',
+                                         index_name=DEFAULT_INDEX_NAME, field_name=DEFAULT_FIELD_NAME):
+    """At least one doc deleted while the transfer was running had already been picked up by it,
```
alonre24: The comments at the beggining of the helper functions (this and others) are long and tedious - reading the code itself is easier to track in some case. please reduce the description significantly.

tests/pytests/test_vecsim_svs.py
```diff
@@ -906,3 +909,184 @@ def test_multiple_svs_indexes_share_pool():
...
+_DELETE_RACE_MODULE_ARGS = 'DEFAULT_DIALECT 2 WORKERS 2 FORK_GC_RUN_INTERVAL 1000000'
+_DELETE_RACE_DIM = 32
+_DELETE_RACE_K = 10
+# How many docs to delete while a resumed transfer runs. Resuming the workers only unblocks the
+# pool - when a worker actually wakes up and snapshots the frontend is up to the scheduler, and a
+# deletion landing before that snapshot never reaches the backend. A long stream of deletions, one
+# round trip each and each contending for the same frontend lock the snapshot needs, therefore
+# makes the overlap overwhelmingly likely. It is a margin, not a proof:
+# `assert_deletions_reached_the_backend` is what turns a lost race into a visible failure rather
+# than silently lost coverage.
```
alonre24: redundant and tedious comment - make it short an very very consice or remove

tests/pytests/vecsim_utils.py
```diff
@@ -123,3 +125,109 @@ def wait_for_background_indexing(env, index_name, field_name, message=''):
...
+    return env.execute_command('FT.SEARCH', index_name, f'*=>[KNN {k} @{field_name} $vec]',
+                               'PARAMS', 2, 'vec', query_vec.tobytes(), 'NOCONTENT', 'LIMIT', 0, k)
+
+def assert_doc_indexed_under_own_vector(env, k, query_vec, expected_doc, message='', **kwargs):
+    """`expected_doc` is returned for a KNN query with its own vector, i.e. it is still indexed
+    under its own label. Containment in the top `k` rather than an exact top-1, so as not to also
+    bet on the recall of a compressed, approximate backend."""
+    res = knn_search(env, query_vec, k, **kwargs)
+    env.assertContains(expected_doc, res,
+                       message=f"{message}: {expected_doc} was not returned for its own vector: {res}")
```
alonre24: This test is still prone to rare flakiness in case of unreachable nodes (see https://redislabs.atlassian.net/browse/MOD-18162) - please assert different things instead

tests/pytests/vecsim_utils.py
alonre24: Cut all five helper docstrings roughly in half, and the test-side comments with them. What's left is only what the code can't state itself: why the queued-job count is needed beside `BACKGROUND_INDEXING`, why `assert_transfer_did_not_complete` does not hold for an update transfer, and why the label count is read from the tiered index rather than summed over its two sub-indexes.

tests/pytests/test_vecsim_svs.py
alonre24: Down to three lines: that the group size is a margin rather than a proof, and that `assert_deletions_reached_the_backend` is what makes a lost race fail. The mechanics of *why* a long stream of single round trips widens the window are gone.

tests/pytests/vecsim_utils.py
alonre24: Removed. `assert_doc_indexed_under_own_vector` required a named doc back from a KNN query with its own vector - a self-match, which is exactly the topology-dependent assertion MOD-18162 says an index/concurrency-sanity test must not make. These tests are squarely in that category: `WORKERS 2`, concurrent writes, deletes and background transfers.

`assert_knn_page_live` replaces it, and the old `assert_deleted_docs_not_returned`, with only graph-independent properties from the ticket's allowed list: a full page of `k` results, distinct, none of them deleted. The page count is still the end-to-end detector for an entry left behind by a deleted doc, since the pipeline drops results whose doc is gone from the doc table and the entry costs a result slot.

`SEARCH_WINDOW_SIZE` stays, but its justification changed and is now stated as such: not to make a particular doc findable - a wide window cannot cross a disconnected component, which is one of the misconceptions the ticket calls out - but because the backend keeps deleted entries marked rather than removed, and they are traversed without being returned, so a `k`-wide beam would often come back with fewer than `k` live docs on a correct index.

Worth being explicit about what this gives up: the self-match was the only assertion that could catch a doc indexed under *another* doc's vector, e.g. after the transfer's swaps journal. The accounting in `assert_svs_tiered_state` counts labels and vectors, so a swap between two live labels is invisible to it. I looked at keeping the check graph-independently through exact `ADHOC_BF` distances, but that is hybrid-only (the schema would need a filter field) and distances under `LeanVec4x8` are approximate, so the comparison would need a tolerance wide enough to be meaningless. Per MOD-18162 that belongs in a deterministic test with controlled construction rather than here.

One of the replacements turns out to be better aimed than I expected: merging master brought in [MOD-17795] *Prevent duplicate vector results during tiered indexing* (#11267), a VectorSimilarity bump. The distinctness check in `assert_knn_page_live` is exactly a detector for that class of regression, in exactly the window these tests drive.

tests/pytests/test_vecsim_svs.py
```diff
@@ -913,3 +916,161 @@ def test_multiple_svs_indexes_share_pool():
...
+        # to happen with a transfer of their vectors pending.
+        assert_transfer_pending(env, message='phase 1, training pending')
+        docs.delete(pending_deletes, 10)
+
+        marked_deleted_before_transfer = svs_backend_marked_deleted(env)
+
+    # The training is running now: a doc deleted from the batch being transferred has to be removed
+    # from the backend once the transfer puts it there, which is what the deletions journal is for.
+    # A flow test cannot pin down an interleaving inside the job, hence the two witnesses below.
+    jobs_done_before_transfer = workers_jobs_done(env)
```
dor-forer, line 1032: Can we guarantee that the transfer is still running when these deletions happen? The workers resume before we sample the job counter, so the transfer could already be finished. Both assertions would then pass for ordinary deletions after indexing, without testing deletion during transfer. Could we pause the transfer after it copies the vectors, delete the documents, then let it finish?

alonre24: Good catch - that gap was real. Rather than tighten the sampling, I dropped the internal-state witnesses for the "delete while it runs" case entirely (65952e23f, on top of 1dfbbf533 which first tried moving the jobs-done sample earlier and turned out to have the same problem one layer down).

Concretely: `assert_transfer_did_not_complete` (the jobs-done counter this comment is about) and `assert_deletions_reached_the_backend` (a similar witness checking the backend's marked-deleted count) are both gone. Neither is a guarantee - a resumed worker's thread has to actually get scheduled and reach a given vector before the deletion that races it, and under real CPU contention it sometimes doesn't. I could reproduce `assert_deletions_reached_the_backend` failing for exactly that reason when running the full 46-test suite in parallel on a loaded machine, even though every correctness assertion after it - `FT.INFO num_docs`, the tiered index's label/vector accounting, and the KNN result checks - kept passing. That's the tell: these witnesses were only proving the test harness won its own race, not exercising a code path those other assertions don't already cover.

What's left: `paused_workers` still makes the "delete while a transfer is pending" case fully deterministic (`assert_transfer_pending` checks that setup, not the race outcome), and the "delete while it runs" case is now best-effort - no longer asserted, but still covered by the same final-state checks, since those are what MOD-13168 actually broke. 6 repeated full-suite runs under the same contended conditions that used to flake are now clean.

dor-forer (approved):

PR #9029, opened 2026-04-13
[MOD-14732] fix race condition on hybrid

JoanFM:
### The race condition

`ProcessHybridCursorMappings` dispatches async callbacks to shards and waits for all responses. The coordinator reads `numShards` from the topology, allocates a shared `processCursorMappingCallbackContext` (`ctx`), and waits until `responseCount == numShards` before freeing `ctx`.

However, `iterStartCb` runs asynchronously on an IO thread and reads the **live** topology, which may have a different shard count after a migration. If the live topology has `M > N` shards, `iterStartCb` dispatches commands to all `M` shards, but the coordinator only waits for `N` responses. After receiving `N` responses, the coordinator frees `ctx` via `cleanupCtx`, while the remaining `M - N` callbacks still hold a pointer to the freed memory.

When those late callbacks fire, `MRIteratorCallback_GetPrivateData(ctx)` returns a dangling pointer, and the callback dereferences freed memory — reading `cb_ctx->mutex`, `cb_ctx->searchMappings`, etc.

The ASAN trace confirms this: `processCursorMappingCallback` (thread T4, `search-uv-1`) reads from the freed `ctx` at the very first dereference, while thread T23 (`coord-5081`) already freed it via `cleanupCtx` → `rm_free`.

### The fix

The existing `privateDataInit` hook in `MR_IterateWithPrivateData` is called from `iterStartCb` on the IO thread **after** the actual shard count is determined from the live topology but **before** any commands are dispatched. We leverage this hook with a new `processCursorMappingInit` callback that:

1. Locks `ctx->mutex`
2. Updates `ctx->numShards` to the actual number of shards from the live topology
3. Signals the condition variable (in case the coordinator is already waiting)
4. Unlocks the mutex

The coordinator's wait loop now reads `ctx->numShards` (under the mutex) instead of the local `numShards` variable, so it dynamically picks up the updated count and waits for all dispatched callbacks before freeing `ctx`.

Also re-enables `test_add_shard_and_migrate_hybrid_BG`, which was previously skipped due to a flaky hybrid cursor migration crash (SIGSEGV) caused by this race.

---

> [!NOTE]
> **Medium Risk**
> Touches distributed hybrid execution concurrency: coordinator now waits on an IO-thread-provided shard count, so mistakes could cause hangs or missed shard responses during cluster topology changes.
> 
> **Overview**
> Fixes a race in `ProcessHybridCursorMappings` where the coordinator could free the shared callback context before all shard callbacks completed if the cluster shard count changed during migration.
> 
> The mapping collection now initializes `numShards` on the IO thread via a new `processCursorMappingInit` hook (and allocates the error array there), and the coordinator wait loop blocks until initialization is done and `responseCount` reaches the *actual* shard count. Call sites were updated to drop the previously captured `numShards` parameter, and the previously-skipped `test_add_shard_and_migrate_hybrid_BG` is re-enabled to cover the crash scenario.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3fb62049b08a84e04cd7a01f3192c3c8ae125f5c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/coord/hybrid/hybrid_cursor_mappings.c
```diff
@@ -26,8 +27,15 @@ typedef struct {
     pthread_mutex_t *mutex;           // Mutex for array access and completion tracking
     pthread_cond_t *completionCond;   // Condition variable for completion signaling
     int numShards;                    // Total number of expected shards
+    // Reference count for safe deallocation. Initialized to 1 (coordinator) + numShards (callbacks).
```
kei-nan: can't we wrap the struct itself with a strongref

src/coord/hybrid/hybrid_cursor_mappings.c
kei-nan: here it looks like we are waiting for all the responses to arrive
If we are waiting for all callbacks to finish then why do we need to add a ref count?

kei-nan (commented):
Left some comments, not sure this PR is the fix
We are supposed to wait for all responses to arrive before releasing the context.

src/coord/hybrid/hybrid_cursor_mappings.c
JoanFM: The wait loop on line 263 waits for all callbacks to increment responseCount under the mutex (line 181), but that happens before the callback finishes. After the callback unlocks the mutex (line 196), it still uses it

src/coord/hybrid/hybrid_cursor_mappings.c
JoanFM: Using `StrongRef` here is tricky because of how `privateData` works in `MRIterator`. The `privateData` is a single shared `void *` — all callbacks receive the same raw pointer. There's no mechanism to give each callback its own pre-cloned `StrongRef`.

To use `StrongRef`, we'd need the object to store a self-referential `StrongRef self` field, and each callback would clone from it at runtime. This is unidiomatic — in the codebase (e.g. `DepleterSync`), `StrongRef` is always managed from the **outside**: the creator clones refs upfront and hands one to each consumer. The managed object itself never knows it's ref-counted.

Alternatively, we could heap-allocate a `StrongRef` and pass *that* as `privateData`, but then we'd need to know when to free the heap-allocated `StrongRef` container itself — the same lifetime problem all over again.

The atomic refcount is a better fit here because it's self-contained: the object manages its own lifetime directly, no external ref holders needed, no self-referential ownership.

src/coord/hybrid/hybrid_cursor_mappings.c
kei-nan: maybe that needs to be fixed instead of adding ref counting on the struct?
I think the point of waiting was so we would know the threads are no longer running.

src/coord/hybrid/hybrid_cursor_mappings.c
kei-nan: we could pass the ref manager pointer as the private data.

src/coord/hybrid/hybrid_cursor_mappings.c
JoanFM: but isn't refCounting a more robust way to be sure ?

JoanFM: what is the downside u see in refCounting?

src/coord/hybrid/hybrid_cursor_mappings.c
JoanFM: But seems like bypassing the logic of having StrongRef around it right? Coupling us with the details of StrongRef.

src/coord/hybrid/hybrid_cursor_mappings.c
kei-nan: right now it seems like we have two mechanisms meant to manage the struct lifetime.
one is waiting for all the threads and another is ref counting.
Feels like one of them is redundant.

Because we need to process the responses then it makes sense for the existing mechanism to remain.

src/coord/hybrid/hybrid_cursor_mappings.c
kei-nan: I think there are some places in the code that do that.
See for example `IndexSpec_TimedOutProc`

src/coord/hybrid/hybrid_cursor_mappings.c
Itzikvaknin: I think we should avoid managing references that belong to other users, like the callbacks. It feels like the opposite of what they are supposed to be used for.

In the proposed change, the execution thread is already managing them. Why not pass a weak reference instead? I also do not see an issue with passing the same instance to all of them. Why is that a problem? Reference counting is not meant to control access.

I am not sure I fully understand what problem we are trying to solve here.

JoanFM: IMHO they do not serve the exact purpose, one seeks to wait until all the response come and the other actually handles who frees the data in a robust way. But yes I can fix in an easier way

src/coord/hybrid/hybrid_cursor_mappings.c
JoanFM: I do not see exactly what is the benefit when compared to simple refCounting?

src/coord/hybrid/hybrid_cursor_mappings.c
JoanFM: Anyway, changed the approach to use Weak/Strong Ref

src/coord/hybrid/hybrid_cursor_mappings.c
```diff
@@ -237,8 +252,10 @@ bool ProcessHybridCursorMappings(const MRCommand *cmd, int numShards, StrongRef
     }
     // Wait for all callbacks to complete
     pthread_mutex_lock(ctx->mutex);
-    // initialize count with response counts in case some shards already sent a response
-    for (size_t count = ctx->responseCount; count < numShards; count = ctx->responseCount) {
+    // Wait until all responses arrive. Use ctx->numShards (not the local numShards)
+    // because the IO thread may have updated it via processCursorMappingInit when
+    // the live topology differs from what the coordinator originally read.
+    for (size_t count = ctx->responseCount; count < ctx->numShards; count = ctx->responseCount) {
```
kei-nan: isn't there a race here?
can't we obtain the lock before the thread had a chance to update ctx->numShards?
then we would think it is zero and end the loop early.

maybe add that the response count is not zero?
Not sure if the response count can ever truly be zero.

JoanFM: CHanged to wait for numShards to be properly initialized from the IO-thread.

src/coord/hybrid/dist_hybrid.c
```diff
@@ -254,6 +254,7 @@ void HybridRequest_buildMRCommand(RedisModuleString **argv, int argc,
   MRCommand_appendVsim(xcmd, argv, argc, vsimOffset, &kArgIndex);
 
   // Calculate and apply effective K for KNN queries if SHARD_K_RATIO is set
+  // TODO: Potentially edit in IO thread where numShards is actually known
```
kei-nan: not sure I understand what the comment is saying, consider clarifying it a bit.

kei-nan (commented):
Looks good,
Left one comment

src/coord/hybrid/dist_hybrid.c
JoanFM: done!

kei-nan (dismissed):

kei-nan (approved):

PR #9084, opened 2026-04-15
MOD-14801: Implement COLLECT Reducer - projection only

Itzikvaknin:
## Implement COLLECT reducer core logic (projection only)

Implement the COLLECT reducer's core logic (field projection, row collection, and finalization)  in Rust, while keeping argument parsing and the factory function in C.

This is the first step - only **field projection** is implemented. SORTBY and LIMIT are parsed and stored but not yet applied during execution (follow-up work). Wildcard (`*`) is rejected at parse time until implemented.

### Part 1: Reducer infrastructure setup

- New `reducers` crate with `Reducer` (safe `#[repr(transparent)]` wrapper over `ffi::Reducer`) and `ReducerOptions` wrapper types.
- New `reducers_ffi` crate (`c_entrypoint/reducers_ffi`) with cbindgen config, build script, and auto-generated `reducers_rs.h` header.
- Wired both crates into the workspace `Cargo.toml` and the `redisearch_rs` entrypoint.
- Added `reducer.h` to the `ffi` crate's bindgen inputs so Rust can see `ffi::Reducer`, `ffi::BlkAlloc`, etc.
- Added `bumpalo` workspace dependency for arena allocation of per-group instances.

### Part 2: COLLECT reducer - projection only

**C side** (`collect.c`):
- Refactored the old `CollectReducer` struct into a parse-only `CollectParseData` (no longer embeds `Reducer`).
- Removed the stub vtable functions (`collectNewInstance`, `collectAdd`, `collectFinalize`, `collectFreeInstance`, `collectFree`) - these now live in Rust.
- After parsing succeeds, `RDCRCollect_New` calls `CollectReducer_Create` (Rust FFI) to allocate the reducer and wire the vtable, then frees the temporary C arrays.
- Wildcard (`*`) in FIELDS is rejected at parse time with a clear error message until the feature is implemented.

**Rust side** (`reducers/src/collect.rs` + `reducers_ffi/src/collect.rs`):
- `CollectReducer`: `#[repr(C)]` struct with `Reducer` as first field (enabling safe `*mut ffi::Reducer ↔ *mut CollectReducer` casts). Stores field keys, sort keys, limit config, and a `Bump` arena.
- `CollectCtx`: per-group instance, arena-allocated. `add()` projects configured field keys from the source `RLookupRow`, cloning values. `finalize()` serializes collected rows as an array of maps.
- FFI vtable functions: `collectNewInstance`, `collectAdd`, `collectFinalize`, `collectFreeInstance`, `collectFree`.
- Uses native `SharedRsValue` / `RsValue` / `Map` / `Array` from the `value` crate (no `RSValueFFI` wrapper). The FFI boundary in `collectFinalize` casts `SharedRsValue::into_raw()` to `*mut ffi::RSValue` to match the C `Finalize` function pointer signature.

### Not yet implemented (follow-up)

- **Wildcard (`*`) field collection** - rejected at parse time; will collect all fields from each row once implemented.
- **SORTBY execution** - config is parsed and stored, but `finalize()` does not sort.
- **LIMIT execution** - config is parsed and stored, but `finalize()` does not slice.

### Testing

**C++ parser tests (`test_cpp_collect.cpp`) - temporary:**
- Converted from direct struct-field access to FFI accessor functions (`CollectReducer_GetFieldKeysLen`, `CollectReducer_HasWildcard`, etc.) since the struct is now opaque to C++.
- Wildcard tests converted to error-expectation tests (`WildcardOnlyRejected`, `WildcardAmongFieldsRejected`).
- Removed `FieldsOnly` (covered by Python flow tests) and `RejectsWhenUnstableFeaturesDisabled` (migrated to Python as `test_collect_requires_unstable_features`).
- **These C++ tests and their FFI accessors are temporary** - they exist because the parsing logic is in C and can't yet be tested from Rust. Once migrated to Python flow tests, delete `test_cpp_collect.cpp`, the FFI accessors in `reducers_ffi`, and the corresponding getter methods in `reducers/src/collect.rs`.

**Python E2E tests (`test_groupby_collect.py`) - new:**
- 12 tests covering: single/multi-field projection (HASH + JSON), chained GROUPBY, NULL/missing values, aliases, multi-key GROUPBY, output structure validation, unstable-features gate, LOAD with JSON path, and RESP2 sanity.
- All tests are skipped in cluster mode (`@skip(cluster=True)`).

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Introduces a new reducer implementation and C↔Rust FFI boundary for aggregation, which can impact query execution correctness and memory safety. SORTBY/LIMIT are parsed but not yet enforced, so behavior may differ from user expectations until follow-ups land.
> 
> **Overview**
> Adds a Rust-backed implementation of the `COLLECT` aggregate reducer that collects per-group rows with **field projection** and returns an array-of-maps result.
> 
> The C factory in `collect.c` is refactored to parse into temporary data and then delegate reducer allocation/vtable wiring to Rust via `CollectReducer_Create`; wildcard `FIELDS *` is now explicitly rejected at parse time.
> 
> Introduces new `reducers` and `reducers_ffi` crates (plus generated `reducers_rs.h`) to host reducer logic/FFI, updates workspace/entrypoint wiring, and adds Python flow tests for end-to-end `COLLECT` behavior while adjusting C++ parser tests to use FFI accessors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d332afca600b00d5c236d7e1baed8a378d723abc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
```diff
@@ -0,0 +1,264 @@
...
+    limit_count: u64,
+) -> *mut ffi::Reducer {
+    let field_keys: Box<[&RLookupKey]> = if !field_keys.is_null() && field_keys_len > 0 {
+        // SAFETY: ensured by caller (1.)
+        Box::from(unsafe {
+            slice::from_raw_parts(field_keys.cast::<&RLookupKey>(), field_keys_len)
+        })
+    } else {
+        Box::new([])
+    };
```
JonasKruckenberg, line 53: is the reducer meant to copy out the values here? or is the reducer meant to simply use the allocation handed to it?

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
```diff
@@ -0,0 +1,264 @@
...
+
+/// Creates a new per-group collect reducer instance.
+///
+/// # Safety
+///
+/// 1. `r` must point to a [valid] `CollectReducer` masquerading as a `ffi::Reducer`.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn collectNewInstance(r: *mut ffi::Reducer) -> *mut c_void {
```
JonasKruckenberg:
```suggestion
pub unsafe extern "C" fn collectNewInstance(r: Option<NonNull<ffi::Reducer>>) -> *mut c_void {
```

to better guard against callers passing NULL.

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
```diff
@@ -0,0 +1,264 @@
...
+    // SAFETY: ensured by caller (1.)
+    let r = unsafe { r.cast::<CollectReducer>().as_mut().unwrap() };
+
+    let ctx: *mut CollectCtx = r.alloc_instance();
+    ctx.cast()
```
JonasKruckenberg:
```suggestion
    ptr::from_mut(r.alloc_instance()).cast::<c_void>()
```

its best to rely on implicit casts like these as little as possible.

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
```diff
@@ -0,0 +1,264 @@
...
+/// Frees a per-group collect reducer instance.
+///
+/// # Safety
+///
+/// 1. `r` must point to a [valid] `CollectReducer` masquerading as a `ffi::Reducer`.
+/// 2. `ctx` must point to a [valid] `CollectCtx` masquerading as a void pointer.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn collectFreeInstance(r: *mut ffi::Reducer, ctx: *mut c_void) {
```
JonasKruckenberg:
```suggestion
pub unsafe extern "C" fn collectFreeInstance(r: Option<NonNull<ffi::Reducer>>, ctx: *mut c_void) {
```

src/redisearch_rs/reducers/src/collect.rs
```diff
@@ -0,0 +1,195 @@
...
+        SharedRsValue::new(RsValue::Array(Array::new(row_maps.into_boxed_slice())))
+    }
+
+    /// Release heap-allocated storage and decrement `SharedRsValue` refcounts.
+    ///
+    /// Must be called before the arena drops this instance, since [`Bump`]
+    /// does not run destructors.
+    pub fn free(&mut self, _r: &CollectReducer) {
+        drop(std::mem::take(&mut self.rows));
+    }
```
JonasKruckenberg:
```suggestion
```

let just rely on the regular drop implemented, instead of building our own here. we call `drop_in_place` in the FFI shim

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
```diff
@@ -0,0 +1,264 @@
...
+    // SAFETY: ensured by caller (1.)
+    let r = unsafe { r.cast::<CollectReducer>().as_mut().unwrap() };
+    // SAFETY: ensured by caller (2.)
+    let collect = unsafe { &mut *ctx.cast::<CollectCtx>() };
+
+    collect.free(r);
```
JonasKruckenberg:
```suggestion
    // SAFETY: ensured by caller (2.)
    let collect = ctx.cast::<CollectCtx>();

    unsafe { NonNull::drop_in_place(collect) }
```

lets use [NonNull::drop_in_place](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.drop_in_place) here.

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
```diff
@@ -0,0 +1,264 @@
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn collectAdd(
+    r: *mut ffi::Reducer,
+    ctx: *mut c_void,
+    srcrow: *const ffi::RLookupRow,
+) -> c_int {
+    // SAFETY: ensured by caller (1.)
+    let r = unsafe { r.cast::<CollectReducer>().as_mut().unwrap() };
+    // SAFETY: ensured by caller (2.)
+    let collect = unsafe { &mut *ctx.cast::<CollectCtx>() };
```
JonasKruckenberg: likewise here
```suggestion
    let collect = unsafe { ctx.cast::<CollectCtx>().as_mut() };
```

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
```diff
@@ -0,0 +1,264 @@
...
+    // SAFETY: ensured by caller (3.)
+    let srcrow = unsafe { &*srcrow.cast::<RLookupRow>() };
```
JonasKruckenberg:
```suggestion
    let srcrow = unsafe { srcrow.cast::<RLookupRow>().as_ref() };
```

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    collect.add(r, srcrow);
+
+    1
```
JonasKruckenberg, line 140: what's the purpose of the return type here?

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
```diff
@@ -0,0 +1,264 @@
...
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn collectFinalize(
+    r: *mut ffi::Reducer,
+    ctx: *mut c_void,
+) -> *mut ffi::RSValue {
+    // SAFETY: ensured by caller (1.)
+    let r = unsafe { r.cast::<CollectReducer>().as_mut().unwrap() };
+    // SAFETY: ensured by caller (2.)
+    let collect = unsafe { &*ctx.cast::<CollectCtx>() };
```
JonasKruckenberg:
```suggestion
    let collect = unsafe { ctx.cast::<CollectCtx>().as_ref() };
```

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
```diff
@@ -0,0 +1,264 @@
...
+/// # Safety
+///
+/// 1. `r` must point to a [valid] `CollectReducer` masquerading as a `ffi::Reducer`.
+/// 2. `ctx` must point to a [valid] `CollectCtx` masquerading as a void pointer.
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn collectFinalize(
+    r: *mut ffi::Reducer,
+    ctx: *mut c_void,
```
JonasKruckenberg: use Option<NonNull here too

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
Itzikvaknin: It gets an array of `RLookupKey`s from the C side - so we want to borrow the keys (owned by the `RLookup`), but to allocate the container to hold the references

src/redisearch_rs/reducers/src/collect.rs
```diff
@@ -0,0 +1,195 @@
...
+/// Configuration (field keys, sort keys, limits) is parsed in C and passed to
+/// Rust via [`CollectReducer::new`]. The [`RLookupKey`][ffi::RLookupKey]
+/// pointers are borrowed from the [`RLookup`][ffi::RLookup] infrastructure and
+/// outlive this reducer.
+///
+/// This struct must be `#[repr(C)]` and its first field must be a [`Reducer`]
+/// because it is downcast in C to `ffi::Reducer`, which reads vtable pointers
+/// directly.
+#[repr(C)]
+pub struct CollectReducer<'a> {
```
JonasKruckenberg, line 28: if this is passed to C through a `Box`ed heap allocation anyway, we should make this opaque

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
Itzikvaknin: To align with the C impl:
```
static int counterAdd(Reducer *r, void *ctx, const RLookupRow *srcrow) {
  ((counterData *)ctx)->count++;
  return 1;
}
```

Although its not referenced anywhere..

src/redisearch_rs/reducers/src/reducer_options.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+    pub const unsafe fn from_raw_mut<'a>(ptr: *mut ffi::ReducerOptions) -> &'a mut Self {
+        // SAFETY: ensured by caller (1.)
+        unsafe { ptr.cast::<Self>().as_mut().unwrap() }
+    }
+
+    /// Get a reference to the `args` cursor.
+    pub fn args(&self) -> &ffi::ArgsCursor {
+        // SAFETY: (1.) due to creation with `ReducerOptions::from_raw_mut`
+        unsafe { &*self.0.args }
```
JonasKruckenberg:
```suggestion
        unsafe { self.0.args.as_ref() }
```

src/redisearch_rs/reducers/src/reducer_options.rs
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /// Get a mutable reference to the query error.
+    pub fn status(&mut self) -> &mut QueryError {
+        // SAFETY: (1.) due to creation with `ReducerOptions::from_raw_mut`
+        unsafe { &mut *self.0.status.cast::<QueryError>() }
```
JonasKruckenberg:
```suggestion
        unsafe { self.0.status.cast::<QueryError>().as_mut() }
```

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
Itzikvaknin: Can we fail `make lint` on that?

src/redisearch_rs/reducers/src/collect.rs
Itzikvaknin: We rely on the `(Reducer*)collectReducerPointer` cast in every FFI callback - `collectNewInstance`, `collectAdd`, `collectFinalize`, `collectFree`. This requires `#[repr(C)]` to guarantee the reducer field sits at offset 0.

Do you think changing it to opaque worth it?

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
```diff
@@ -0,0 +1,264 @@
...
+///
+/// 1. `r` must point to a [valid] `CollectReducer` masquerading as a `ffi::Reducer`,
+///    originally created by [`CollectReducer_Create`].
+///
+/// [valid]: https://doc.rust-lang.org/std/ptr/index.html#safety
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn collectFree(r: *mut ffi::Reducer) {
+    // SAFETY: ensured by caller (1.)
+    drop(unsafe { Box::from_raw(r.cast::<CollectReducer>()) });
+}
```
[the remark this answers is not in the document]
Itzikvaknin: Thats incorrect - 
```
static inline void Reducer_GenericFree(Reducer *r) {
  BlkAlloc_FreeAll(&r->alloc, NULL, 0, 0);
  rm_free(r);
}
```
We dont use this allocator at all

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
daxhuiberts: @JonasKruckenberg I had the same question. The array is created while parsing the arguments for the collect reducer and then this Rust side copies it over into its own array so the argument parsing doesn't have to save any state anymore after this collect reducer creation. Since this is done only once per query and keeps the ownership simple I think this is fine.

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
daxhuiberts: FYI: Maybe the `1` signifies a success/error return type? Normally that would be reflected in Rust where the safe `CollectReducer` returns a `Result` type and this FFI layer converts this to the proper numeric return type. In this case a simple comment on the purpose of the `1` is a good idea.

src/redisearch_rs/reducers/src/collect.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// [`SharedRsValue::null_static`].
+    pub fn add(&mut self, r: &CollectReducer, row: &RLookupRow) {
+        let mut values = Vec::with_capacity(r.field_keys.len());
+        for key in &r.field_keys {
+            let value = row
+                .get(key)
+                .cloned()
+                .unwrap_or_else(SharedRsValue::null_static);
+            values.push(value);
+        }
```
daxhuiberts, line 161: More idiomatic would be
```suggestion
        let values = r.field_keys.iter().map(|key| {
            row.get(key).cloned().unwrap_or_else(SharedRsValue::null_static)
        }).collect();
```

src/redisearch_rs/reducers/src/collect.rs
```diff
@@ -0,0 +1,187 @@
...
+        let row_maps: Vec<SharedRsValue> = self
+            .rows
+            .iter()
+            .map(|row_values| {
+                let entries: Box<[_]> = row_values
+                    .iter()
+                    .zip(r.field_keys.iter())
+                    .map(|(val, key)| {
+                        let name_val = SharedRsValue::new_string(key.name().to_bytes().to_vec());
+                        (name_val, val.clone())
```
daxhuiberts, line 179: Instead of `val.clone()`, we can also consume `self.rows` (by using `drain`). `finalize` needs to be changed to take `&mut self` then as well.

daxhuiberts:
@JonasKruckenberg @Itzikvaknin I'm not in favor of `Option<NonNull<T>>` in FFI functions.

The asymmetry it has with `*const T` is very annoying. It's much more bloated and much harder to digest comparing it to the associated C function signature compared to `*mut/const T`.
The usage of `as_ref()`, `as_mut()`, `cast()`, `expect()` and `unwrap_unchecked()` (or a debug/release helper which combines them) makes it perfectly adequate to use and no less safe than `Option<NonNull<T>>`.
Also look at the weird type dance neeeded in the function assignments in `src/redisearch_rs/reducers/src/reducer.rs` now.

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect.rs
Itzikvaknin: Added `// C reducer->Add convention: always returns 1`

daxhuiberts (approved):
Looks very good!

nafraf (approved):

nafraf (approved):

PR #9123, opened 2026-04-15
Cache Redis build in task-test workflow

LukeMathWalker:

Extract the Get Redis + Build Redis steps into a reusable composite action that caches the install dir keyed by the resolved Redis SHA, platform, architecture, sanitizer, and coverage flags. Subsequent runs with the same inputs restore the binary from the GitHub Actions cache instead of rebuilding from source.

The legacy (non-node20) path on amazonlinux:2 falls back to the previous uncached make install behaviour.

Cache in action:
image

1. MOD-...
2. #...

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> CI-only change, but it alters how Redis binaries are fetched/built and introduces cache-keying logic; mis-keying or env differences could cause stale/incorrect binaries or new workflow failures.
> 
> **Overview**
> Adds a new composite action `build-redis` that resolves the requested Redis ref to an immutable SHA, builds Redis into a workspace-local prefix, and caches/restores that install directory keyed by SHA + platform/arch + sanitizer/coverage/TLS flags; it also exports `redis-server` on `PATH` and sets `REDIS_SERVER`.
> 
> Updates `task-test.yml` and `task-build-artifacts.yml` to replace the previous `get-redis` + manual `make install` steps with the cached action (passing through sanitizer/coverage settings and disabling TLS for artifact builds), reducing repeat build time across runs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db983fb4ae2bc3466460f212b9f8b78c4d45abf8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/actions/build-redis/action.yml
+++ b/.github/actions/build-redis/action.yml
@@ -1,0 +1,119 @@
+name: Build Redis (cached)
+description: |
+  Build and install Redis once per (resolved Redis SHA, platform, arch, sanitizer,
+  coverage) tuple, restoring from the GitHub Actions cache when available.
+
+  Redis is installed under a workspace-relative prefix and that directory is
+  cached. After the action completes, `redis-server` is on PATH for subsequent
+  steps and the REDIS_SERVER env var points at the binary explicitly.
+
+inputs:
+  ref:
+    description: "Redis git ref to build (e.g. 'unstable', '7.2.3', or a 40-char SHA)"
+    required: true
+  san:
+    description: "Sanitizer (empty or 'address')"
+    required: false
+    default: ""
+  coverage:
+    description: "Build with coverage instrumentation ('true' or 'false')"
+    required: false
+    default: "false"
+  build_tls:
+    description: "Build Redis with TLS support ('true' or 'false')"
+    required: false
+    default: "true"
+  cache_platform_id:
+    description: "Opaque identifier for the binary-compatible platform (container image or runner env)"
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Resolve Redis SHA
+      id: resolve
+      shell: bash
+      env:
+        REDIS_REF: ${{ inputs.ref }}
+      run: |
+        if [[ "$REDIS_REF" =~ ^[0-9a-f]{40}$ ]]; then
+          SHA="$REDIS_REF"
+        else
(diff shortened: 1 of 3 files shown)


.github/actions/build-redis/action.yml
```diff
@@ -0,0 +1,153 @@
+name: Build Redis (cached)
+description: |
+  Build and install Redis once per (resolved Redis SHA, platform, arch, sanitizer,
```
memark, line 3: Would we need a flag for LTO here?

.github/workflows/task-test.yml
```diff
@@ -348,18 +348,15 @@ jobs:
           echo "CXX=$(dirname $CLANG_BIN)/clang++-$CLANG_VERSION" >> $GITHUB_ENV
           echo "LD=$CLANG_BIN" >> $GITHUB_ENV
           echo "ASAN_SYMBOLIZER_PATH=/usr/lib/llvm-$CLANG_VERSION/bin/llvm-symbolizer" >> $GITHUB_ENV
-      - name: Get Redis
-        uses: ./.github/actions/get-redis
+      - name: Build Redis (cached)
```
memark, line 333: I'm thinking we would want to reuse this action in task-build-artefacts as well?

.github/actions/build-redis/action.yml
LukeMathWalker: No, since Redis uses RediSearch as a dynamic library. No inlining across the boundary.

.github/workflows/task-test.yml
LukeMathWalker: Done!

memark (dismissed):

JoanFM:
should we also add some compiler information on the cache key?

LukeMathWalker:
> Should we also add some compiler information to the cache key?

I don't think it matters, no? We just need a working binary. 
SAN/COV matter insofar as we need everything to "line up" across the stack for those to work properly, but we can consume a Redis compiled with `gcc` while RediSearch is compiled with `clang` (fictional example) with no issues.

gdesmott (approved):

memark (approved):

PR #9126, opened 2026-04-16
claude: add check-flow-coverage and improve-flow-coverage skills

gdesmott:
Check python test coverage and help filling coverage gaps.

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

#### Release Notes

- [ ] This PR requires release notes
- [x] This PR does not require release notes

If a release note is required (bug fix / new feature / enhancement), describe the **user impact** of this PR in the title.  

---

> [!NOTE]
> **Low Risk**
> Documentation and agent skill additions plus a small build.sh robustness fix for coverage runs; no product API or runtime behavior change for end users.
> 
> **Overview**
> Adds agent workflows for **C-only** Python flow coverage via gcov/lcov: **`/check-flow-coverage`** builds with `COV=1`, runs the full pytest suite, parses `bin/flow_standalone.info`, and reports testable gaps (excluding disk-only, API-only, and defensive paths); **`/improve-flow-coverage`** chains that analysis with triage, optional `RS_ASSERT`/`RS_ABORT` vs `LCOV_EXCL_*` guidance, and targeted re-verification via `/write-flow-tests`. Both skills are wired in **`AGENTS.md`** and redirect Rust paths to `/check-rust-coverage`.
> 
> **`build.sh`** now records unit-test and pytest exit codes with `cmd && RESULT=0 || RESULT=$?` so **`set -e` does not skip `capture_coverage`** when tests fail—matching what the new skills assume for partial-run coverage data.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7af7dabf253ee37f8c0470ec27982c5fcc2fb8bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9133, opened 2026-04-16
[MOD-14864] improve query.c flow test coverage

gdesmott:
Filling gaps in `query.c` flow test coverage by adding new python tests.

Improve codecov reported coverage by 1.55%.

All those have been generated using my new `check-flow-coverage` skill: https://github.com/RediSearch/RediSearch/pull/9126

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Test-only changes that add coverage for edge-case warnings/errors without modifying production logic.
> 
> **Overview**
> Improves regression/flow coverage by adding **new standalone-only Python tests** for query parsing and iterator edge cases.
> 
> The new tests cover: `WITHSUFFIXTRIE` TEXT/TAG contains/suffix/wildcard behavior when `MAXPREFIXEXPANSIONS` is exceeded (warning paths and NULL/no-match paths), fuzzy expansion behavior (including iterator capacity growth and max-expansion warnings), `ismissing()` correctness inside parameterized queries, and dialect-1 legacy behavior where `FILTER`/`GEOFILTER` on non-schema fields returns empty results.
> 
> It also adds parser/validation coverage for invalid query attribute values (including vector query attributes like invalid `shard_k_ratio`/unknown attributes), ensures `FT.EXPLAIN` renders `IDS` when using `INKEYS`, and validates tag prefix results when below default `MINPREFIX`, plus an error case for wildcard queries against a TEXT field lacking `WITHSUFFIXTRIE` in a mixed schema.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42d8fc9b5888703d586313c50cad2cad271abaa6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/pytests/test_contains.py
+++ b/tests/pytests/test_contains.py
@@ -376,6 +376,7 @@ def test_issue_3124(env):
   res_not_exist2 = env.cmd('ft.search', 'idx_txt_suffix', '@t:ell*')
   env.assertEqual(res_not_exist1, res_not_exist2)
 
+@skip(cluster=True)
 def testTextSuffixTrieMaxPrefixExpansions():
     """Contains query on TEXT WITHSUFFIXTRIE field hits max prefix expansion limit."""
     env = Env(moduleArgs='DEFAULT_DIALECT 2', protocol=3)
--- a/tests/pytests/test_contains.py
+++ b/tests/pytests/test_contains.py
@@ -375,3 +375,25 @@ def test_issue_3124(env):
   res_not_exist1 = env.cmd('ft.search', 'idx_txt', '@t:ell*')
   res_not_exist2 = env.cmd('ft.search', 'idx_txt_suffix', '@t:ell*')
   env.assertEqual(res_not_exist1, res_not_exist2)
+
+def testTextSuffixTrieMaxPrefixExpansions():
+    """Contains query on TEXT WITHSUFFIXTRIE field hits max prefix expansion limit."""
+    env = Env(moduleArgs='DEFAULT_DIALECT 2', protocol=3)
+    conn = getConnectionByEnv(env)
+
+    env.expect('FT.CREATE', 'idx', 'SCHEMA',
+               't', 'TEXT', 'WITHSUFFIXTRIE').ok()
+
+    # Create many distinct values sharing a common substring
+    for i in range(20):
+        conn.execute_command('HSET', f'{{doc}}:{i}', 't', f'val{i}common')
+
+    # Set max expansions very low
+    run_command_on_all_shards(env, config_cmd(), 'SET', 'MAXPREFIXEXPANSIONS', '1')
+
+    # Contains query (*common*) uses the suffix trie charIterCb path
+    res = env.cmd('FT.SEARCH', 'idx', '*common*', 'LIMIT', '0', '0')
+    env.assertContains('Max prefix expansions limit was reached', res['warning'])
+
+    # Restore default
+    run_command_on_all_shards(env, config_cmd(), 'SET', 'MAXPREFIXEXPANSIONS', '200')
--- a/tests/pytests/test_fuzzy.py
+++ b/tests/pytests/test_fuzzy.py
@@ -111,7 +111,7 @@ def testFuzzyManyExpansions(env):
     # We expect at least 9 results (>8 to trigger the capacity doubling).
     # The exact count may vary depending on what the trie iterator yields,
     # but all 10 terms are within distance 1 of "bat".
-    env.assertGreaterEqual(res[0], 9)
+    env.assertGreaterEqual(res[0], 9, message=res)
 
 def testFuzzyMaxPrefixExpansionsWarning():
     """Verify that a fuzzy query triggers a max prefix expansions warning
--- a/tests/pytests/test_fuzzy.py
+++ b/tests/pytests/test_fuzzy.py
@@ -93,6 +93,29 @@ def testFuzzyWithNumbersOnly(env):
         env.expect('ft.search', 'idx', '%%21345%%', 'DIALECT', dialect)\
             .equal([1, 'doc1', ['test', '12345']])
 
+def testFuzzyMaxPrefixExpansionsWarning():
+    """Verify that a fuzzy query triggers a max prefix expansions warning
+    when the number of fuzzy matches exceeds MAXPREFIXEXPANSIONS."""
+    env = Env(protocol=3)
+    conn = getConnectionByEnv(env)
+
+    conn.execute_command('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT')
+
+    # Create terms that are all within Levenshtein distance 1 of "ab":
+    # aa, ab, ac, ..., az  (26 terms, all distance <= 1 from "ab")
+    for c in 'abcdefghijklmnopqrstuvwxyz':
+        conn.execute_command('HSET', f'doc_a{c}', 't', f'a{c}')
+
+    # Set max prefix expansions to 1 so the fuzzy expansion is sure to exceed it
+    run_command_on_all_shards(env, config_cmd(), 'SET', 'MAXPREFIXEXPANSIONS', '1')
+
+    # Fuzzy query: %ab% should try to expand to all terms within distance 1
+    res = env.cmd('FT.SEARCH', 'idx', '%ab%')
+    env.assertContains('Max prefix expansions limit was reached', res['warning'])
+
+    # Restore default
+    run_command_on_all_shards(env, config_cmd(), 'SET', 'MAXPREFIXEXPANSIONS', '200')
+
 @skip()
 def testTagFuzzy(env):
     # TODO: fuzzy on tag is broken?
--- a/tests/pytests/test_fuzzy.py
+++ b/tests/pytests/test_fuzzy.py
@@ -93,6 +93,26 @@ def testFuzzyWithNumbersOnly(env):
         env.expect('ft.search', 'idx', '%%21345%%', 'DIALECT', dialect)\
             .equal([1, 'doc1', ['test', '12345']])
 
+def testFuzzyManyExpansions(env):
+    """Verify that a fuzzy query works correctly when matching more than 8
+    terms, which triggers the internal iterator array capacity doubling in
+    addTerm (initial capacity is 8)."""
+    conn = getConnectionByEnv(env)
+    conn.execute_command('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT')
+
+    # Create 10 distinct 3-letter terms that are all within Levenshtein
+    # distance 1 of "bat": substitute the first character.
+    terms = ['bat', 'cat', 'dat', 'eat', 'fat', 'gat', 'hat', 'mat', 'pat', 'rat']
+    for i, term in enumerate(terms):
+        conn.execute_command('HSET', f'doc{i}', 't', term)
+
+    # Fuzzy search with distance 1: %bat% should match all of the above
+    res = env.cmd('ft.search', 'idx', '%bat%', 'LIMIT', '0', '0')
+    # We expect at least 9 results (>8 to trigger the capacity doubling).
+    # The exact count may vary depending on what the trie iterator yields,
+    # but all 10 terms are within distance 1 of "bat".
+    env.assertGreaterEqual(res[0], 9)
+
 def testFuzzyMaxPrefixExpansionsWarning():
     """Verify that a fuzzy query triggers a max prefix expansions warning
     when the number of fuzzy matches exceeds MAXPREFIXEXPANSIONS."""
--- a/tests/pytests/test_fuzzy.py
+++ b/tests/pytests/test_fuzzy.py
@@ -93,6 +93,7 @@ def testFuzzyWithNumbersOnly(env):
         env.expect('ft.search', 'idx', '%%21345%%', 'DIALECT', dialect)\
             .equal([1, 'doc1', ['test', '12345']])
 
+@skip(cluster=True)
 def testFuzzyManyExpansions(env):
     """Verify that a fuzzy query works correctly when matching more than 8
     terms, which triggers the internal iterator array capacity doubling in
@@ -113,6 +114,7 @@ def testFuzzyManyExpansions(env):
     # but all 10 terms are within distance 1 of "bat".
     env.assertGreaterEqual(res[0], 9, message=res)
 
+@skip(cluster=True)
 def testFuzzyMaxPrefixExpansionsWarning():
     """Verify that a fuzzy query triggers a max prefix expansions warning
     when the number of fuzzy matches exceeds MAXPREFIXEXPANSIONS."""
--- a/tests/pytests/test_fuzzy.py
+++ b/tests/pytests/test_fuzzy.py
@@ -98,7 +98,7 @@ def testFuzzyManyExpansions(env):
     terms, which triggers the internal iterator array capacity doubling in
     addTerm (initial capacity is 8)."""
     conn = getConnectionByEnv(env)
-    conn.execute_command('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT')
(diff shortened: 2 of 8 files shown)


src/query.c
```diff
@@ -414,20 +414,20 @@ QueryNode *NewVectorNode_WithParams(struct QueryParseCtx *q, VectorQueryType typ
...
       return NULL;
-  }
+  } // LCOV_EXCL_STOP
   return ret;
 }
 
 void SetFilterNode(QueryAST *q, QueryNode *filterNode) {
   if (filterNode == NULL) return;
-  if (q->root == NULL) {
+  if (q->root == NULL) { // LCOV_EXCL_START — defensive: parser always produces a root node before filters are added
```
GuyAv46: Great commenting
Should we consider replacing the defensive unreachable code with debug asserts? (`RS_ASSERT`)

tests/pytests/test_tags.py
```diff
@@ -78,6 +78,26 @@ def testTagPrefix(env):
...
 
+@skip(cluster=True)
+def testTagPrefixTooShort(env):
+    """A single-char tag prefix is rejected when MINPREFIX (default 2) is not met."""
+    env.expect(
+        'ft.create', 'idx', 'ON', 'HASH',
+        'schema', 'tags', 'tag', 'separator', ',').ok()
+
+    env.expect('ft.add', 'idx', 'doc1', 1.0, 'fields',
+               'tags', 'alpha,beta,gamma').ok()
```
GuyAv46: Let's avoid `FT.ADD` unless the test requires it

tests/pytests/test_parser.py
```diff
@@ -804,3 +804,45 @@ def test_intersection_v2():
     env.expect('FT.EXPLAIN', 'idx', '@text:foo bar @tag:{baz}', 'VERBATIM').equal(expected)
     env.expect('FT.EXPLAIN', 'idx', '(@text:foo bar) @tag:{baz}', 'VERBATIM').equal(expected)
     env.expect('FT.EXPLAIN', 'idx', '@tag:{baz} (@text:foo bar)', 'VERBATIM').equal(expected)
+
+def test_invalid_query_attributes():
+    """Test that invalid query attribute values produce proper errors."""
+    env = Env(moduleArgs='DEFAULT_DIALECT 2')
+    conn = getConnectionByEnv(env)
+    env.expect('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT').ok()
+    waitForIndex(env, 'idx')
```
GuyAv46: redundant (nothing to wait for)

tests/pytests/test_fuzzy.py
```diff
@@ -93,6 +93,49 @@ def testFuzzyWithNumbersOnly(env):
         env.expect('ft.search', 'idx', '%%21345%%', 'DIALECT', dialect)\
             .equal([1, 'doc1', ['test', '12345']])
 
+def testFuzzyManyExpansions(env):
+    """Verify that a fuzzy query works correctly when matching more than 8
+    terms, which triggers the internal iterator array capacity doubling in
+    addTerm (initial capacity is 8)."""
+    conn = getConnectionByEnv(env)
+    conn.execute_command('FT.CREATE', 'idx', 'SCHEMA', 't', 'TEXT')
```
GuyAv46: Don't use `conn` for creating the index. `conn` should only be used for `HSET` and similar key-write commands

tests/pytests/test_fuzzy.py
```diff
@@ -93,6 +93,49 @@ def testFuzzyWithNumbersOnly(env):
...
+    terms = ['bat', 'cat', 'dat', 'eat', 'fat', 'gat', 'hat', 'mat', 'pat', 'rat']
+    for i, term in enumerate(terms):
+        conn.execute_command('HSET', f'doc{i}', 't', term)
+
+    # Fuzzy search with distance 1: %bat% should match all of the above
+    res = env.cmd('ft.search', 'idx', '%bat%', 'LIMIT', '0', '0')
+    # We expect at least 9 results (>8 to trigger the capacity doubling).
+    # The exact count may vary depending on what the trie iterator yields,
+    # but all 10 terms are within distance 1 of "bat".
+    env.assertGreaterEqual(res[0], 9)
```
GuyAv46: For better observability if something fails, either check the complete `res`, or add it as the assertion message, so we can see what we did get
```suggestion
    env.assertGreaterEqual(res[0], 9, message=res)
```

tests/pytests/test_contains.py
```diff
@@ -375,3 +375,25 @@ def test_issue_3124(env):
   res_not_exist1 = env.cmd('ft.search', 'idx_txt', '@t:ell*')
   res_not_exist2 = env.cmd('ft.search', 'idx_txt_suffix', '@t:ell*')
   env.assertEqual(res_not_exist1, res_not_exist2)
+
+def testTextSuffixTrieMaxPrefixExpansions():
+    """Contains query on TEXT WITHSUFFIXTRIE field hits max prefix expansion limit."""
+    env = Env(moduleArgs='DEFAULT_DIALECT 2', protocol=3)
```
GuyAv46, line 382: Unless there's a good reason to pass custom args, consider using the default `Env()`, by expecting the `env` as an input

tests/pytests/test_contains.py
```diff
@@ -375,3 +375,25 @@ def test_issue_3124(env):
   res_not_exist1 = env.cmd('ft.search', 'idx_txt', '@t:ell*')
   res_not_exist2 = env.cmd('ft.search', 'idx_txt_suffix', '@t:ell*')
   env.assertEqual(res_not_exist1, res_not_exist2)
+
+def testTextSuffixTrieMaxPrefixExpansions():
```
GuyAv46, line 380: For tests that don't have cluster implications, consider adding `@skip(cluster=True)` decorator, to avoid running them in cluster mode

tests/pytests/test_missing.py
```diff
@@ -593,3 +593,30 @@ def testMissingGC():
...
+def testMissingWithParams():
+    """Tests that ismissing() works correctly in a parameterized query.
+    This exercises QueryNode_EvalParams traversal of QN_MISSING nodes."""
+
+    env = Env(moduleArgs='DEFAULT_DIALECT 2')
+    conn = getConnectionByEnv(env)
+
+    env.expect('FT.CREATE', 'idx', 'SCHEMA',
+               't', 'TEXT', 'INDEXMISSING').ok()
+    waitForIndex(env, 'idx')
```
GuyAv46: Nothing to wait for. This is only needed when we set the data before the index creation, or when we dump and reload

tests/pytests/test_missing.py
```diff
@@ -593,3 +593,30 @@ def testMissingGC():
...
+
+    conn.execute_command('HSET', 'has_t', 't', 'hello')
+    conn.execute_command('HSET', 'no_t', 'other', 'world')
+
+    # Parameterized query combined with ismissing - triggers
+    # QueryNode_EvalParams on a QN_MISSING node (withChildren=0 path)
+    res = env.cmd('FT.SEARCH', 'idx', '@t:$val | ismissing(@t)',
+                  'NOCONTENT', 'SORTBY', 't', 'ASC',
+                  'PARAMS', '2', 'val', 'hello')
+    env.assertEqual(res[0], 2)
```
GuyAv46: Either compare the full result, or add it as a message

tests/pytests/test_parser.py
```diff
@@ -804,3 +804,45 @@ def test_intersection_v2():
     env.expect('FT.EXPLAIN', 'idx', '@text:foo bar @tag:{baz}', 'VERBATIM').equal(expected)
     env.expect('FT.EXPLAIN', 'idx', '(@text:foo bar) @tag:{baz}', 'VERBATIM').equal(expected)
     env.expect('FT.EXPLAIN', 'idx', '@tag:{baz} (@text:foo bar)', 'VERBATIM').equal(expected)
+
+def test_invalid_query_attributes():
+    """Test that invalid query attribute values produce proper errors."""
+    env = Env(moduleArgs='DEFAULT_DIALECT 2')
```
GuyAv46: `DIALECT` 2 is the default on CI.
Unless there's a reason not to run it on other dialects (if it will fail), consider using the default:
```suggestion
def test_invalid_query_attributes(env: Env):
    """Test that invalid query attribute values produce proper errors."""
```

GuyAv46 (commented):
Looks very good
A few comments about the tests, consider fixing in all the new tests, and not only where I commented.
1. skipping test on cluster mode if not relevant
2. using default env if no custom args are needed
3. Make sure the full response is printed when assertion fails, e.g., `env.assertEqual(res[0], 2)` may show `3 == 2`, but we won't be able to see what the extra unexpected result was. `env.assertEqual(res[0], 2, message=res)` will show `3 == 2 [<result>]`
4.

src/query.c
gdesmott: You mean `RS_ABORT`?

I don't know, it's a matter of taste I suppose: defensive programmings vs enforcing more aggressively. Up to you, but if we go this road I'll change the skill accordingly then.

tests/pytests/test_contains.py
gdesmott: `protocol=3` is needed to access `res['warning']`

src/query.c
GuyAv46: We can save it for the Rust migration :)
I meant `RS_ASSERT(q->root != NULL)`, but any macro from the family could be used.
In general, I don't like defensive programming, especially when it's `NULL` validation on stuff that will never be `NULL`, or node type on a static function that is only called once from a switch

gdesmott: I tend to agree with you.

This code will be used as base for the Rust migration so I'd rather have the assertions now so we can more easly figure out constraints when porting so I added those.

gdesmott:
Thanks for your valuable input.

I used all your feedback to create a new `write-flow-tests` skill, see https://github.com/RediSearch/RediSearch/pull/9126
Feel free to suggest more recommendations if you see fit, I don't have much experience writing Python tests.

src/query.c
```diff
@@ -414,20 +414,16 @@ QueryNode *NewVectorNode_WithParams(struct QueryParseCtx *q, VectorQueryType typ
       QueryNode_SetParam(q, &ret->params[1], &vq->range.radius, NULL, value);
       vq->range.order = BY_ID;
       break;
-    default:
+    default: // LCOV_EXCL_START — exhaustive switch: only KNN and RANGE vector query types exist
       QueryNode_Free(ret);
```
JoanFM: should we add an RS_ABORT then here? or some RS_LOG_ASSERT?

src/query.c
```diff
@@ -414,20 +414,16 @@ QueryNode *NewVectorNode_WithParams(struct QueryParseCtx *q, VectorQueryType typ
...
 }
 
 void SetFilterNode(QueryAST *q, QueryNode *filterNode) {
   if (filterNode == NULL) return;
-  if (q->root == NULL) {
-    // Cannot add filter to empty AST - free the node to avoid leaking its resources
-    QueryNode_Free(filterNode);
-    return;
-  }
+  RS_ASSERT(q->root != NULL);
```
JoanFM, line 427: Are we sure this assertion is correct?

JoanFM: I would not include these changes in a PR about increasing coverage. This is a dangerous code change. If we see it is not covered, we then need to try to cover or in a separate PR do this change

JoanFM (changes requested):

src/query.c
```diff
@@ -666,9 +662,7 @@ static QueryIterator *Query_EvalPrefixNode(QueryEvalCtx *q, QueryNode *qn) {
   Trie *t = spec->terms;
   TrieCallbackCtx ctx = {.q = q, .opts = &qn->opts};
 
-  if (!t) {
-    return NULL;
-  }
+  RS_ASSERT(t);
```
JoanFM, line 667: same here

src/query.c
```diff
@@ -722,9 +716,7 @@ static QueryIterator *Query_EvalWildcardQueryNode(QueryEvalCtx *q, QueryNode *qn
   TrieCallbackCtx ctx = {.q = q, .opts = &qn->opts};
   RSToken *token = &qn->verb.tok;
 
-  if (!t || !token->str) {
-    return NULL;
-  }
+  RS_ASSERT(t && token->str);
```
JoanFM, line 722: and here

src/query.c
```diff
@@ -1240,10 +1230,10 @@ static QueryIterator *Query_EvalTagPrefixNode(QueryEvalCtx *q, TagIndex *idx, Qu
       }
     }
     TrieMapIterator *it = TrieMap_IterateWithFilter(idx->values, tok->str, tok->len, iter_mode);
-    if (!it) {
+    if (!it) { // LCOV_EXCL_START — defensive: TrieMap_IterateWithFilter only returns NULL on allocation failure
```
JoanFM: Let's cover the previous mentioned ones also with these defensive comments

src/query.c
```diff
@@ -1622,9 +1612,7 @@ void QAST_Destroy(QueryAST *q) {
 
 int QAST_Expand(QueryAST *q, const char *expander, RSSearchOptions *opts, RedisSearchCtx *sctx,
                 QueryError *status) {
-  if (!q->root) {
-    return REDISMODULE_OK;
-  }
+  RS_ASSERT(q->root);
```
JoanFM, line 1611: Also here, seems like dangerous change

src/query.c
```diff
@@ -1869,9 +1857,7 @@ static int QueryNode_CheckIsValid(QueryNode *n, IndexSpec *spec, RSSearchOptions
 // Checks whether query nodes are valid
 // Currently Phrase nodes are checked whether slop/inorder are allowed
 int QAST_CheckIsValid(QueryAST *q, IndexSpec *spec, RSSearchOptions *opts, QueryError *status) {
-  if (!q || !q->root) {
-    return REDISMODULE_OK;
-  }
+  RS_ASSERT(q && q->root);
```
JoanFM, line 1854: and here.

JoanFM (commented):
Can u clarify why we are so certain that there will be no NULL root for queries? If so, this should also be documented in the function documentation?

src/query.c
gdesmott: Indeed; fixed.

src/query.c
gdesmott: I just discussed this with @GuyAv46 who suggested to enforce those more aggressively, see https://github.com/RediSearch/RediSearch/pull/9133#discussion_r3092675560 

I don't mind moving this commit to another PR but don't really see the point tbh.

gdesmott:
I documented the new asserts.

src/query.c
```diff
@@ -1968,9 +1961,9 @@ static sds QueryNode_DumpSds(sds s, const IndexSpec *spec, const QueryNode *qs,
 
   if (qs->opts.fieldMask && qs->opts.fieldMask != RS_FIELDMASK_ALL && qs->type != QN_NUMERIC &&
       qs->type != QN_GEO && qs->type != QN_IDS) {
-    if (!spec) {
+    if (!spec) { // LCOV_EXCL_START — defensive: QAST_DumpExplain always passes a valid spec
```
JoanFM: if true, should we remove in favour of an assert?

gdesmott: Probably, changed it.

gdesmott:
@GuyAv46 : I removed all the `query.c` changes and will move them to their own PR.

Can we already merge this now as it now only contains the new python tests?

src/query.c
gdesmott: Moved to https://github.com/RediSearch/RediSearch/pull/9195

GuyAv46 (approved):

gdesmott:
@JoanFM : looks like you need to approve this as well as you requested changes.

JoanFM (approved):

PR #8910, opened 2026-04-01
[MOD-12331] Port YieldableMetric over to Rust

LukeMathWalker:
## Describe the changes in the pull request

Port `RSYieldableMetric` over to Rust, removing its dependency on `RsValue`. 

#### Which additional issues this PR fixes
1. MOD-...
2. #...

#### Main objects this PR modified
1. ...

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

#### Release Notes

- [ ] This PR requires release notes
- [x] This PR does not require release notes

If a release note is required (bug fix / new feature / enhancement), describe the **user impact** of this PR in the title.  

---

> [!NOTE]
> **Medium Risk**
> Touches cross-language FFI and query-result lifetime/ownership semantics (metrics draining/concatenation and aggregate child handling), which can introduce subtle memory or correctness regressions in scoring/metric propagation.
> 
> **Overview**
> Ports yieldable result metrics from the C `RSYieldableMetric`/`RSValue` array to a Rust-owned `MetricsVec` (ThinVec) embedded directly in `RSIndexResult`, removing the old C helpers (`ResultMetrics_Free`, clone/reset helpers, and `ResetAndPushMetricData`) and the `RSYieldableMetric` definition from `redisearch.h`.
> 
> Adds a new `metrics_ffi` crate + generated `headers/metrics.h` exposing `ResultMetrics_Add/Reset`, `RSYieldableMetric_Concat`, `MetricsVec_AsSlice`, and `MetricsVec_UpdateValue`; C code is updated to iterate metrics via a slice and to materialize numbers with `RSValue_NewNumber` only at write-out time (`result_processor.c`). Hybrid/vector iterators and Rust query iterators are updated to drain/concat metrics safely when aggregating children, to update score metrics via `MetricsVec_UpdateValue`, and to adapt tests/bench mocks accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d7acc784694d843b0f3a4d294f1419867ad2675a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9169, opened 2026-04-17
[MOD-14996] Tests for Topk disk: computeDistances_Disk

ThierryBerger:
checking code coverage at https://app.codecov.io/gh/RediSearch/RediSearch/blob/master/src%2Fiterators%2Fhybrid_reader.c reveals some untested path, this PR closes the gap.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Test-only change that adds coverage for disk-path hybrid vector search logic without modifying production code paths.
> 
> **Overview**
> Adds a new C++ gtest suite (`test_cpp_hybrid_reader_disk.cpp`) to cover the previously untested *disk* branch of `HybridIterator` distance computation (`computeDistances_Disk`).
> 
> The tests use lightweight VecSim mocks to validate **top-k ordering**, **NaN distance filtering**, **reranking via `getExactDistances`**, and **timeout propagation** by temporarily overriding the global `vecsimTimeoutCallback`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9dde5d19fa6f8399215e9a5232dfb596774e4c2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/cpptests/test_cpp_hybrid_reader_disk.cpp
+++ b/tests/cpptests/test_cpp_hybrid_reader_disk.cpp
@@ -1,0 +1,273 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+// Include C++ VecSim headers before any C headers to get full class definitions.
+#include "VectorSimilarity/src/VecSim/vec_sim_interface.h"
+#include "VectorSimilarity/src/VecSim/vec_sim_adhoc_bf_ctx.h"
+#include "VectorSimilarity/src/VecSim/memory/vecsim_malloc.h"
+
+#include "gtest/gtest.h"
+#include "index_utils.h"
+#include "iterator_util.h"
+
+#include "iterators/hybrid_reader.h"
+#include "redisearch.h"
+#include "util/timeout.h"
+#include "types_rs.h"
+
+#include <cmath>
+#include <limits>
+#include <map>
+#include <memory>
+#include <vector>
+
+// vecsimTimeoutCallback is a global function pointer in hybrid_reader.c, deliberately kept
+// non-static so tests can swap it to simulate timeouts.
+extern "C" {
+extern int (*vecsimTimeoutCallback)(TimeoutCtx *ctx);
+}
+
+// operator delete reads obj->allocator after destruction; keep the shared_ptr alive across delete.
+static void freeVecSimObject(VecSimIndexInterface *obj) {
+    auto alloc = obj->getAllocator();
+    delete obj;
+}
+
+// ============================================================================
+// Mocks
+// ============================================================================
+
+// Controllable VecSimAdhocBfCtx: returns distances from a pre-loaded map.
+// sq8Distances simulates the fast SQ8 approximation pass.
+// exactDistances simulates the FP32 reranking pass (getExactDistances).
+struct MockAdhocBfCtx : public VecSimAdhocBfCtx {
+    std::map<labelType, double> sq8Distances;
+    std::map<labelType, double> exactDistances;
+
+    MockAdhocBfCtx(std::shared_ptr<VecSimAllocator> alloc,
+                   std::map<labelType, double> sq8,
+                   std::map<labelType, double> exact)
+        : VecSimAdhocBfCtx(std::move(alloc)),
+          sq8Distances(std::move(sq8)),
+          exactDistances(std::move(exact)) {}
+
+    double getDistanceFrom(labelType label) const override {
+        auto it = sq8Distances.find(label);
+        return it != sq8Distances.end() ? it->second
+                                        : std::numeric_limits<double>::quiet_NaN();
+    }
+
+    void getExactDistances(const labelType *labels, double *out, size_t count) const override {
+        for (size_t i = 0; i < count; ++i) {
+            auto it = exactDistances.find(labels[i]);
+            out[i] = it != exactDistances.end() ? it->second
+                                                 : std::numeric_limits<double>::quiet_NaN();
+        }
+    }
+};
+
+// Minimal VecSimIndexInterface implementation for the disk path.
+// Only newAdhocBfCtx and indexSize need real implementations; all other methods are stubs.
+struct MockDiskVecSimIndex : public VecSimIndexInterface {
+    std::map<labelType, double> sq8Distances;
+    std::map<labelType, double> exactDistances;
+
+    MockDiskVecSimIndex(std::shared_ptr<VecSimAllocator> alloc,
+                        std::map<labelType, double> sq8,
+                        std::map<labelType, double> exact)
+        : VecSimIndexInterface(std::move(alloc)),
+          sq8Distances(std::move(sq8)),
+          exactDistances(std::move(exact)) {}
+
+    VecSimAdhocBfCtx *newAdhocBfCtx(const void *) const override {
+        auto alloc = VecSimAllocator::newVecsimAllocator();
+        return new (alloc) MockAdhocBfCtx(alloc, sq8Distances, exactDistances);
+    }
+
+    size_t indexSize() const override { return sq8Distances.size(); }
+
+    // ---- Stubs for pure virtual methods not exercised by these tests ----
+    int addVector(const void *, labelType) override { return 0; }
+    int deleteVector(labelType) override { return 0; }
+    double getDistanceFrom_Unsafe(labelType, const void *) const override { return 0.0; }
+    size_t indexCapacity() const override { return 0; }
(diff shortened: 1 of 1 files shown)


tests/cpptests/test_cpp_hybrid_reader_disk.cpp
```diff
@@ -0,0 +1,278 @@
...
+          sq8Distances(std::move(sq8)),
+          exactDistances(std::move(exact)) {}
+
+    double getDistanceFrom(labelType label) const override {
+        auto it = sq8Distances.find(label);
+        return it != sq8Distances.end() ? it->second
+                                        : std::numeric_limits<double>::quiet_NaN();
+    }
+
+    void getExactDistances(const labelType *labels, double *out, size_t count) const override {
```
ThierryBerger, line 66: I'm not changing those, they are `override`s.

tests/cpptests/test_cpp_hybrid_reader_disk.cpp
```diff
@@ -0,0 +1,273 @@
...
+    // Exact FP32 distances reverse the ranking.
+    std::map<labelType, double> exact = {{1, 0.1}, {2, 0.7}};
+    QueryIterator *it = nullptr;
+    MockDiskVecSimIndex *index = makeIterator(sq8, exact, {1, 2}, 2, &it);
+
+    ASSERT_NE(it, nullptr);
+
+    // Enable reranking before the first Read() triggers prepareResults().
+    auto hr = (HybridIterator *)it;
+    hr->runtimeParams.hnswDiskRuntimeParams.shouldRerank = VecSimBool_TRUE;
```
memark: I would have two functions `makeRerankingIterator` and `make(Normal?)Iterator` where the `makeRerankingIterator` could include these two lines. It would also remove the need for the empty `{}` `exact` argument in most calls.
(The two functions can ofc call each other internally.)

tests/cpptests/test_cpp_hybrid_reader_disk.cpp
```diff
@@ -0,0 +1,273 @@
...
+// ============================================================================
+
+class HybridReaderDiskTest : public ::testing::Test {
+    std::unique_ptr<MockQueryEvalCtx> mockCtx;
+    std::array<float, 4> queryVec = {1.0f, 2.0f, 3.0f, 4.0f};
+protected:
+    void SetUp() override {
+        mockCtx = std::make_unique<MockQueryEvalCtx>(100, 10);
+        // Non-null diskSpec routes the hybrid reader into the disk code path.
+        mockCtx->spec.diskSpec = (RedisSearchDiskIndexSpec *)1;
```
memark: This seems a bit fragile. If anyone tries to dereference the `diskSpec` it will seg fault. Could we create a more robust `diskSpec` stub?

tests/cpptests/test_cpp_hybrid_reader_disk.cpp
```diff
@@ -0,0 +1,273 @@
...
+#include <memory>
+#include <vector>
+
+// vecsimTimeoutCallback is a global function pointer in hybrid_reader.c, deliberately kept
+// non-static so tests can swap it to simulate timeouts.
+extern "C" {
+extern int (*vecsimTimeoutCallback)(TimeoutCtx *ctx);
+}
+
+// operator delete reads obj->allocator after destruction; keep the shared_ptr alive across delete.
```
memark: It's not obvious to me what `shared_ptr` refers to here.

tests/cpptests/test_cpp_hybrid_reader_disk.cpp
```diff
@@ -0,0 +1,273 @@
...
+static void freeVecSimObject(VecSimIndexInterface *obj) {
+    auto alloc = obj->getAllocator();
```
memark: I would put a `[[maybe_unused]]` attribute on this variable, both for clarity and to avoid potential future compilation warning.

tests/cpptests/test_cpp_hybrid_reader_disk.cpp
```diff
@@ -0,0 +1,273 @@
...
+    EXPECT_EQ(it->lastDocId, (t_docId)2);
+
+    // Second result: next lowest = doc 1 (0.5).
+    ASSERT_EQ(it->Read(it), ITERATOR_OK);
+    EXPECT_EQ(it->lastDocId, (t_docId)1);
+
+    // Doc 3 (0.8) is outside top-2 and should not appear.
+    ASSERT_EQ(it->Read(it), ITERATOR_EOF);
+
+    it->Free(it);
```
memark: Consider letting `makeIterator` return something like this (with a destructor)
```
  struct TestHybrid {          
      MockDiskVecSimIndex *index;
      QueryIterator *iter;
      ~TestHybrid() { iter->Free(iter); freeVecSimObject(index); }
  }; 
```
and we can remove these two cleanup lines from every test.

memark (changes requested):
The test cases are sound. Some minor comments.

memark:
Might be a good idea to rebase on master.

Did you look into rejecting the Sonar "findings"?

ThierryBerger:
> Did you look into rejecting the Sonar "findings"?

I connected through github to sonarcloud, but it seems I can't interact with those, maybe they should be assignned to me ? or they will be after a new run maybe

tests/cpptests/test_cpp_hybrid_reader_disk.cpp
ThierryBerger: getAllocator returns a `std::shared_ptr`, when we remove obj, we may decrease the shared_ptr count to 0, which will trigger the release of its allocator, but we need the allocator to free this memory, which is called in the delete operator (last), so we're delaying that free... ; I'll add a precision that we refer to the allocator.

[here is a simplification of what's happening](https://codapi.org/embed/?sandbox=cpp&code=data%3A%3Bbase64%2CjVO7jtswEOz5FZu7RjYcXy%2FLAlKmuOrSGzS5thnwIfDhgxI4354lpciyghwCCBC0HM7Mzq6elRU6SYRGuRA9ctOy56lm0DjfzyuEUfbcMpYCvcFyg6HjAiFEuWOMjpOI8Np%2F0doJHp2Hnwzg16xQrUoJQLgUoWngyTrgE5zbnjTx01M%2BQiv1jsA3dpuTv6E%2BjXwoia1LR61ETUByUdfhwj3KQxd9M9Nt7yLEBUuaydbLC3y10HkuohK4gXhRAejxSK3jlQRP3hkIziCctTtyDadkCezshqqEpzAmB%2FkmMSXS7sfytsh86NT0k9fK4js8xLfaFYJ7ZPs5fogr83PyD1enJLgOfUFK1BixKsV1R3bVDzzE8lqVUK7KZ695YMt09nT7xJOOuzKMwrFA1fV%2FKY1BL%2Be4dsfvpOJR2Yi%2Bo7wPgofY%2FIVrq27MYFohLgSGspFEsr1Hc0GP8O6SlnBESAGBn4icRoj4sGFl7nT3cztd3mZQRbXB%2FYrdGCNrYLiy1T%2BaWEOgzwOfzO5hmN8izUny27hdeWu0spiD431uJFckDjvv%2FAZox5wxSAaiIxUsAOcldSMu3J7xDyVPcfyhSP3Rzqy7wcEwpgWKBvwb)

tests/cpptests/test_cpp_hybrid_reader_disk.cpp
ThierryBerger: I agree. another way is to add `(void)alloc;` at the end of the function, I'll follow your suggestion.

for the record, it's a pattern already present in the codebase, without this additional attribute.

tests/cpptests/test_cpp_hybrid_reader_disk.cpp
ThierryBerger: it's a forward declaration only (void*), it's difficult (or impossible) to make something better out of it. it's purpose is to be used from the disk backend IIUC.

tests/cpptests/test_cpp_hybrid_reader_disk.cpp
memark: >another way is to add (void)alloc; at the end of the function,

I'm aware that it's common. (But I think it's a weird syntax personally.)

memark (dismissed):

GuyAv46 (dismissed):

ThierryBerger:
- @GuyAv46  I had to rebase and fix because #8910  added a `debug_assert!` which my test setup was now triggering.
  - the new `debug_assert`: https://github.com/RediSearch/RediSearch/pull/8910/files#diff-d4c6dc15afb3d817eece5883dddc98a02ecb06c4991cb9956ca1f134a4554ffdR129

See commit 9dde5d19fa6f8399215e9a5232dfb596774e4c2c


I considered removing the debug assert and silently ignore a null key instead, but this would probably be a sign of an incorrect invariant so I fixed the test to have a more credible implementation.

GuyAv46 (approved):

PR #9170, opened 2026-04-17
[MOD-14200] Implement TopKHeap Utility

ThierryBerger:
Add heap utility with ASC/DESC comparator support, capacity enforcement, replace/peek/pop behavior, and unit tests.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Introduces a new, self-contained utility crate with thorough unit tests and no changes to existing runtime logic beyond workspace wiring.
> 
> **Overview**
> Adds a new workspace crate, `top_k`, providing a fixed-capacity `TopKHeap` for keeping the best k `(doc_id, score)` results with a caller-supplied ASC/DESC comparator, plus `peek_worst`, `pop_worst`, and best-first `drain_sorted` behavior and unit tests.
> 
> Wires the crate into the Rust workspace (`Cargo.toml`/`Cargo.lock`) and applies a small formatting-only change to the `tracing-subscriber` dependency specification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd0cb95270375c606d2d390f6d497d00afa4d59b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/top_k/src/heap.rs
+++ b/src/redisearch_rs/top_k/src/heap.rs
@@ -1,0 +1,375 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! A fixed-capacity heap that retains the top-k scored documents.
+
+use std::cmp::Ordering;
+use std::collections::BinaryHeap;
+
+use ffi::t_docId;
+
+/// A (doc_id, score) pair stored in the heap.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct ScoredResult {
+    /// Document identifier.
+    pub doc_id: t_docId,
+    /// Score for the document (lower or higher is "better" depending on the comparator).
+    pub score: f64,
+}
+
+/// Wraps a [`ScoredResult`] so that [`BinaryHeap`] (a max-heap) keeps the *worst*
+/// element at the top, making it cheap to evict.
+///
+/// "Worst" is defined by the [`TopKHeap`]'s comparator:
+///
+/// - `compare(a, b) == Less`  → `a` is **better** than `b`.
+/// - `compare(a, b) == Greater` → `a` is **worse** than `b` (= heap-max, evicted first).
+///
+/// Tie-breaking: equal scores → higher `doc_id` is considered worse (evicted first),
+/// so lower `doc_id` is kept.
+struct HeapEntry {
+    result: ScoredResult,
+    /// Cached comparison function so [`Ord`] can be implemented without extra state.
+    compare: fn(f64, f64) -> Ordering,
+}
+
+impl PartialEq for HeapEntry {
+    fn eq(&self, other: &Self) -> bool {
+        self.cmp(other) == Ordering::Equal
+    }
+}
+
+impl Eq for HeapEntry {}
+
+impl PartialOrd for HeapEntry {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for HeapEntry {
+    fn cmp(&self, other: &Self) -> Ordering {
+        // We want BinaryHeap's root to be the *worst* element so eviction is O(log k).
+        // `compare(self, other) == Greater` means self is worse than other
+        //   → self should be "greater" in BinaryHeap ordering → root.
+        let score_ord = (self.compare)(self.result.score, other.result.score);
+        match score_ord {
+            // self is better than other  → self should NOT be at root → self is "less"
+            Ordering::Less => Ordering::Less,
+            // self is worse than other   → self should be at root → self is "greater"
+            Ordering::Greater => Ordering::Greater,
+            // Tie on score: higher doc_id is worse (evicted first) → Greater
+            Ordering::Equal => self.result.doc_id.cmp(&other.result.doc_id),
+        }
+    }
+}
+
+/// A fixed-capacity heap that retains the k best-scored documents.
+///
+/// Internally a max-heap whose root is the *worst* of the retained elements,
+/// making insertion and eviction O(log k).
+///
+/// # Comparator convention
+///
+/// `compare(a, b)` must return:
+/// - [`Ordering::Less`]    if score `a` is **better** than score `b`
+/// - [`Ordering::Greater`] if score `a` is **worse**  than score `b`
+/// - [`Ordering::Equal`]   if the scores are equally good
+///
+/// For ascending order (lower score = better, e.g. vector distance):
+/// `compare = |a, b| a.partial_cmp(&b).unwrap_or(Ordering::Equal)`
+///
+/// For descending order (higher score = better, e.g. numeric SORTBY):
+/// `compare = |a, b| b.partial_cmp(&a).unwrap_or(Ordering::Equal)`
+pub struct TopKHeap {
+    inner: BinaryHeap<HeapEntry>,
+    capacity: usize,
+    compare: fn(f64, f64) -> Ordering,
+}
+
+impl TopKHeap {
+    /// Creates a new heap that holds at most `capacity` elements,
+    /// using the supplied `compare` function to determine score order.
+    pub fn new(capacity: usize, compare: fn(f64, f64) -> Ordering) -> Self {
+        Self {
+            inner: BinaryHeap::with_capacity(capacity),
+            capacity,
+            compare,
+        }
+    }
+
+    /// Returns the number of elements currently in the heap.
+    pub fn len(&self) -> usize {
+        self.inner.len()
+    }
+
+    /// Returns `true` if the heap contains no elements.
+    pub fn is_empty(&self) -> bool {
+        self.inner.is_empty()
+    }
+
+    /// Returns `true` if the heap has reached its capacity.
+    pub fn is_full(&self) -> bool {
+        self.inner.len() >= self.capacity
+    }
+
+    /// Returns the worst element currently retained (the one that would be evicted next),
+    /// without removing it.
+    pub fn peek_worst(&self) -> Option<ScoredResult> {
+        self.inner.peek().map(|e| e.result)
+    }
+
+    /// Attempts to insert `(doc_id, score)` into the heap.
+    ///
+    /// - If the heap is not full, the element is always inserted.
+    /// - If the heap is full and the new element is **better** than the current worst,
+    ///   the worst is evicted and the new element takes its place.
+    /// - Otherwise the element is discarded.
+    ///
+    /// Returns `true` if the element was inserted.
+    pub fn push(&mut self, doc_id: t_docId, score: f64) -> bool {
+        let entry = HeapEntry {
+            result: ScoredResult { doc_id, score },
+            compare: self.compare,
+        };
+
+        if !self.is_full() {
+            self.inner.push(entry);
+            return true;
+        }
+
+        // The heap is full. Only insert if the new element is strictly better than the
+        // current worst (root). `entry > root` means entry is worse → discard.
+        // `entry < root` means entry is better → evict root, insert entry.
+        // Equal (same score AND same doc_id) → discard to avoid duplicates.
+        let should_insert = match self.inner.peek() {
+            Some(worst) => entry.cmp(worst) == Ordering::Less,
+            None => false,
+        };
+
+        if should_insert {
+            self.inner.pop();
+            self.inner.push(entry);
+            true
+        } else {
+            false
+        }
+    }
+
+    /// Removes and returns the worst element currently retained.
+    pub fn pop_worst(&mut self) -> Option<ScoredResult> {
+        self.inner.pop().map(|e| e.result)
+    }
+
+    /// Drains all elements and returns them sorted best-first.
+    ///
+    /// Consumes the heap.
+    pub fn drain_sorted(self) -> Vec<ScoredResult> {
+        // BinaryHeap::into_sorted_vec() returns elements in ascending Ord order.
+        // In our Ord impl "better" == "Less", so ascending == best-first already.
+        self.inner
+            .into_sorted_vec()
+            .into_iter()
+            .map(|e| e.result)
+            .collect()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Ascending comparator: lower score is better (e.g. vector distance).
+    fn asc(a: f64, b: f64) -> Ordering {
+        a.partial_cmp(&b).unwrap_or(Ordering::Equal)
+    }
+
+    /// Descending comparator: higher score is better (e.g. numeric SORTBY).
+    fn desc(a: f64, b: f64) -> Ordering {
+        b.partial_cmp(&a).unwrap_or(Ordering::Equal)
+    }
+
+    // ── T1 tests ─────────────────────────────────────────────────────────────
+
+    #[test]
+    fn heap_fewer_than_k_preserves_all() {
+        let mut heap = TopKHeap::new(5, asc);
+        heap.push(1, 3.0);
+        heap.push(2, 1.0);
+        heap.push(3, 2.0);
+        assert_eq!(heap.len(), 3);
+
+        let results = heap.drain_sorted();
+        // Best first (lowest score first for ASC)
+        assert_eq!(results[0].score, 1.0);
+        assert_eq!(results[1].score, 2.0);
+        assert_eq!(results[2].score, 3.0);
+    }
+
+    #[test]
+    fn heap_evicts_worst_when_full_asc() {
+        let mut heap = TopKHeap::new(3, asc);
+        heap.push(1, 5.0);
+        heap.push(2, 3.0);
+        heap.push(3, 4.0);
+        // Full, worst = 5.0 (doc 1)
+
+        let inserted = heap.push(4, 2.0); // 2.0 is better than 5.0 → evict doc 1
+        assert!(inserted);
+        assert_eq!(heap.len(), 3);
+
+        let not_inserted = heap.push(5, 6.0); // 6.0 is worse than current worst (4.0) → discard
+        assert!(!not_inserted);
+
+        let results = heap.drain_sorted();
+        let scores: Vec<f64> = results.iter().map(|r| r.score).collect();
+        assert_eq!(scores, vec![2.0, 3.0, 4.0]);
+    }
+
+    #[test]
+    fn heap_evicts_worst_when_full_desc() {
+        let mut heap = TopKHeap::new(3, desc);
+        heap.push(1, 1.0);
+        heap.push(2, 3.0);
+        heap.push(3, 2.0);
+        // Full, worst = 1.0 (doc 1)
+
+        let inserted = heap.push(4, 4.0); // 4.0 is better (higher) → evict doc 1
+        assert!(inserted);
+
+        let not_inserted = heap.push(5, 0.5); // 0.5 is worse → discard
+        assert!(!not_inserted);
+
+        let results = heap.drain_sorted();
+        let scores: Vec<f64> = results.iter().map(|r| r.score).collect();
+        assert_eq!(scores, vec![4.0, 3.0, 2.0]); // best-first = highest-first for DESC
+    }
+
+    #[test]
+    fn heap_capacity_one_keeps_best_asc() {
+        let mut heap = TopKHeap::new(1, asc);
+        heap.push(1, 5.0);
+        heap.push(2, 3.0); // better
+        heap.push(3, 4.0); // worse than 3.0, not inserted
+
+        let results = heap.drain_sorted();
+        assert_eq!(results.len(), 1);
+        assert_eq!(results[0].score, 3.0);
+        assert_eq!(results[0].doc_id, 2);
+    }
+
+    #[test]
+    fn heap_tie_breaking_keeps_lower_doc_id() {
+        let mut heap = TopKHeap::new(2, asc);
+        heap.push(10, 1.0);
+        heap.push(5, 1.0);
+        // Both have score 1.0. Capacity=2, both fit.
+        // Now insert a third with the same score → evict highest doc_id (10)
+        heap.push(3, 1.0);
+
+        let results = heap.drain_sorted();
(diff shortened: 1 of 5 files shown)


src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,375 @@
...
+    fn asc(a: f64, b: f64) -> Ordering {
+        a.partial_cmp(&b).unwrap_or(Ordering::Equal)
+    }
+
+    /// Descending comparator: higher score is better (e.g. numeric SORTBY).
+    fn desc(a: f64, b: f64) -> Ordering {
+        b.partial_cmp(&a).unwrap_or(Ordering::Equal)
+    }
+
+    // ── T1 tests ─────────────────────────────────────────────────────────────
```
ThierryBerger, line 198: Should we keep those "T1", etc..?  They map to `TOP_K_DESIGN.md` doc, I feel like they help with understanding architecture, but no strong feeling.

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,375 @@
...
+/// "Worst" is defined by the [`TopKHeap`]'s comparator:
+///
+/// - `compare(a, b) == Less`  → `a` is **better** than `b`.
+/// - `compare(a, b) == Greater` → `a` is **worse** than `b` (= heap-max, evicted first).
+///
+/// Tie-breaking: equal scores → higher `doc_id` is considered worse (evicted first),
+/// so lower `doc_id` is kept.
+struct HeapEntry {
+    result: ScoredResult,
+    /// Cached comparison function so [`Ord`] can be implemented without extra state.
```
memark, line 39: What does "cached" mean here? Is it a performance optimization? What's the alternative?
It might be worth noting that this adds 8 bytes to the struct (50% increase).

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,375 @@
...
+impl Ord for HeapEntry {
+    fn cmp(&self, other: &Self) -> Ordering {
+        // We want BinaryHeap's root to be the *worst* element so eviction is O(log k).
+        // `compare(self, other) == Greater` means self is worse than other
+        //   → self should be "greater" in BinaryHeap ordering → root.
+        let score_ord = (self.compare)(self.result.score, other.result.score);
+        match score_ord {
+            // self is better than other  → self should NOT be at root → self is "less"
+            Ordering::Less => Ordering::Less,
+            // self is worse than other   → self should be at root → self is "greater"
```
memark: While this comment is true I don't think it adds much to _this_ trivial code, that maps `Greater` to `Greater`.

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,375 @@
...
+            None => false,
+        };
+
+        if should_insert {
+            self.inner.pop();
+            self.inner.push(entry);
+            true
+        } else {
+            false
+        }
```
memark, line 156: I've gone through a couple iterations here, but I think less is more in this case. What do you think?

```suggestion
        if !self.is_full() {
            self.inner.push(entry);
            true
        }
        // The heap is full. Only insert if the new element is strictly better than the
        // current worst (root). `entry > root` means entry is worse → discard.
        // `entry < root` means entry is better → evict root, insert entry.
        // Equal (same score AND same doc_id) → discard to avoid duplicates.
        else if self.inner.peek().is_some_and(|worst| &entry < worst) {
            self.inner.pop();
            self.inner.push(entry);
            true
        } else {
            false
        }
```

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    // ── T1 tests ─────────────────────────────────────────────────────────────
+
+    #[test]
+    fn heap_fewer_than_k_preserves_all() {
+        let mut heap = TopKHeap::new(5, asc);
+        heap.push(1, 3.0);
+        heap.push(2, 1.0);
+        heap.push(3, 2.0);
+        assert_eq!(heap.len(), 3);
```
memark:
```suggestion
```

src/redisearch_rs/top_k/src/heap.rs
memark: That document probably won't be along forever, and at least not up to date. Looking only at the tests, the "T1" makes no sense. I would remove it.

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,375 @@
...
+
+        let results = heap.drain_sorted();
+        // Best first (lowest score first for ASC)
```
memark: I would go over the formatting of the tests, it seems a bit off a times. Personally I prefer Arrange-Act-Assert styles with distinct blocks. (If they're really simple they don't need heading comments, but it can be useful if they contain blank lines.)

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,375 @@
...
+pub struct TopKHeap {
+    inner: BinaryHeap<HeapEntry>,
+    capacity: usize,
+    compare: fn(f64, f64) -> Ordering,
+}
+
+impl TopKHeap {
+    /// Creates a new heap that holds at most `capacity` elements,
+    /// using the supplied `compare` function to determine score order.
+    pub fn new(capacity: usize, compare: fn(f64, f64) -> Ordering) -> Self {
```
memark: Should we accept capacity 0 here? It would mean a heap that's always full.

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    /// Attempts to insert `(doc_id, score)` into the heap.
+    ///
+    /// - If the heap is not full, the element is always inserted.
+    /// - If the heap is full and the new element is **better** than the current worst,
+    ///   the worst is evicted and the new element takes its place.
+    /// - Otherwise the element is discarded.
+    ///
+    /// Returns `true` if the element was inserted.
+    pub fn push(&mut self, doc_id: t_docId, score: f64) -> bool {
```
memark, line 135: How does this function handle `NaN` values? I think they will remain in the heap forever because of the `unwrap_or(Ordering::Equal)`. Is it allowed? Should we assert against it?

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,375 @@
...
+        heap.push(3, 2.0);
+        // Worst in DESC = lowest score = 1.0 (doc 2)
+        let worst = heap.pop_worst().unwrap();
+        assert_eq!(worst.score, 1.0);
+        assert_eq!(worst.doc_id, 2);
+        assert_eq!(heap.len(), 2);
+    }
+
+    #[test]
+    fn pop_worst_on_empty_returns_none() {
```
memark, line 376: Should we have an equivalent for peek_worst?

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,375 @@
...
+        self.inner
+            .into_sorted_vec()
+            .into_iter()
+            .map(|e| e.result)
+            .collect()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
memark, line 179: Should we add a test for duplicate doc_id push?

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,375 @@
...
+        heap.push(3, 4.0); // worse than 3.0, not inserted
+
+        let results = heap.drain_sorted();
+        assert_eq!(results.len(), 1);
+        assert_eq!(results[0].score, 3.0);
+        assert_eq!(results[0].doc_id, 2);
+    }
+
+    #[test]
+    fn heap_tie_breaking_keeps_lower_doc_id() {
```
memark, line 267: Add similar test for DESC mode?

memark (changes requested):
Very well written! I like the public interface of the struct, the internal usage of binary heap as well as the extensive test coverage!

I do think we should find a good (and single) place to document how the ordering works. It feels a bit scattered atm.

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,375 @@
...
+            compare: self.compare,
+        };
+
+        if !self.is_full() {
+            self.inner.push(entry);
+            return true;
+        }
+
+        // The heap is full. Only insert if the new element is strictly better than the
+        // current worst (root). `entry > root` means entry is worse → discard.
```
memark: Why do we introduce the concept of `root` here? Let's just keep calling it `worst`?

src/redisearch_rs/top_k/src/heap.rs
ThierryBerger: I'd keep it, doesn't hurt much, helps with verifying the state before draining...

the push has a logic of refusing push under some conditions, so it makes sense to test for that. we could also verify that all of those pushes return true.

src/redisearch_rs/top_k/src/heap.rs
ThierryBerger: Good point.

We need `Ord` to support using `BinaryHeap`. its function only has `(&self, other: &Self)` parameters, so it's difficult to tweak its behaviour (desc vs asc). (`cache` refer to that value being brought from its parent `TopKHeap`).

Now, we probably don't have other uses than desc and asc, and we'll most likely not switch at runtime, and won't mix different methods, so we can probably have a generic over it like:

```rs
struct HeapEntry<C: Fn(f64, f64) -> Ordering> {
    result: ScoredResult,
    _comparator: PhantomData<C>,
}
```

I believe it would be worth it, but I'm not 100% sure we wouldn't run into surprises later. I suggest to keep this design for now to keep close to the original design, add a ticket to explore this refactoring, and its impact on performance, what do you think ?

src/redisearch_rs/top_k/src/heap.rs
ThierryBerger: It actually depends on the implementation of the `compare`, but you're right, with current test implementations (and my wip branch), a NaN would be considered equal so kept if encountered first.

I'd appreciate @GuyAv46 input here.

I'd advocate to refuse NaN if we can, or consider them `worse` if no other choice.

src/redisearch_rs/top_k/src/heap.rs
ThierryBerger: root is interesting to keep to understand the relationship with the binaryheap, but I agree `worst` is better to reason about, I'll update docs to limit `root` wording.

src/redisearch_rs/top_k/src/heap.rs
ThierryBerger: probably not :+1: ; I changed the parameter for a `NonZeroUsize`, the alternative of addding a `debug_assert!` is tempting too, but a type approach forces caller to think about it.

src/redisearch_rs/top_k/src/heap.rs
memark: Sounds like you thought a bit about this. Go with your idea!

src/redisearch_rs/top_k/src/heap.rs
memark: I just think it's weird ta add asserts to verify your own test code (having nothing to do with the SUT). And we don't do that anywhere else?

memark (dismissed):

src/redisearch_rs/top_k/src/heap.rs
GuyAv46: We don't have explicit handling of that today, and we usually don't expect such values from the supported scorers/vector distances. On the other hand, we can get them occasionally (some scorers, when there are no text fields indexed yet, may return `0/0` score)

I think the performance of the happy path matters most. If this heap is only used by the top-k iterator, we can require that (and debug assert). On the vector iterator today, we filter `NaN`s, and the numeric field cannot hold NaNs

ThierryBerger: Thanks for the precisions! I'd advocate for keeping this up to each implementation then.

If error floats can occur in production (0/0), we'll add tests in vector implementation and consider each source should filter them out (or cannot hold them).

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,413 @@
...
+/// `compare(a, b)` must return:
+/// - [`Ordering::Less`]    if score `a` is **better** than score `b`
+/// - [`Ordering::Greater`] if score `a` is **worse**  than score `b`
+/// - [`Ordering::Equal`]   if the scores are equally good
+///
+/// For ascending order (lower score = better, e.g. vector distance):
+/// `compare = |a, b| a.partial_cmp(&b).unwrap_or(Ordering::Equal)`
+///
+/// For descending order (higher score = better, e.g. numeric SORTBY):
+/// `compare = |a, b| b.partial_cmp(&a).unwrap_or(Ordering::Equal)`
```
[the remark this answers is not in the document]
ThierryBerger: See https://github.com/RediSearch/RediSearch/pull/9170#discussion_r3126690239 for context, keeping open for visibility ; we can probably improve documentation around it, but I'd welcome more actionable feedbacks.

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,413 @@
...
+impl Ord for HeapEntry {
+    fn cmp(&self, other: &Self) -> Ordering {
+        // We want the worst element at the top of the heap so eviction is O(log k).
+        // `compare(self, other) == Greater` means self is worse.
+        let score_ord = (self.compare)(self.result.score, other.result.score);
+        match score_ord {
+            Ordering::Less => Ordering::Less,
+            Ordering::Greater => Ordering::Greater,
+            // Tie on score: higher doc_id is worse (evicted first) → Greater
+            Ordering::Equal => self.result.doc_id.cmp(&other.result.doc_id),
```
GuyAv46, line 66: So we prefer older IDs? Is this aligned with the current C code?

src/redisearch_rs/top_k/src/heap.rs
GuyAv46: Should we add a task/subtask for that? I'd rather we have the best performance and memory consumption over maximum generic

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -0,0 +1,413 @@
...
+            self.inner.push(entry);
+            true
+        }
+        // The heap is full. Only insert if the new element is strictly better than the
+        // current worst (root). `entry > worst` means entry is worse → discard.
+        // `entry < worst` means entry is better → evict worst, insert entry.
+        // Equal (same score AND same doc_id) → discard to avoid duplicates.
+        else if self.inner.peek().is_some_and(|worst| &entry < worst) {
+            self.inner.pop();
+            self.inner.push(entry);
```
GuyAv46: Surprised the standard heap doesn't have an "exchange" API to replace the root with a single heap operation. Maybe we should have our own heap implementation.
We already have `min_heap` for the union iterator, maybe we should base on that and add a better heap utility for the entire project?

src/redisearch_rs/top_k/src/heap.rs
ThierryBerger: no we prefer smaller ids (oh, smaller are older, yes, actually :) ) ; it's similar to both https://github.com/RediSearch/RediSearch/blob/f4a52c5adb23e0d155874198c73b089c8748d4b6/src/result_processor.c#L721 (explicit ordering of doc_id) ; and https://github.com/RediSearch/RediSearch/blob/f4a52c5adb23e0d155874198c73b089c8748d4b6/src/iterators/hybrid_reader.c#L132 (prioritizing previously added entries (which are iterated ascending))

- 0.cmp(1) -> Lesser -> keep self (0) / evict other
- 1.cmp(0) -> Greater -> keep other / evict self (1)

BUT! there is an interesting subtlety about all that, looking more into C code, there is a `ascend_ map` which we should look into respecting.

- See https://github.com/RediSearch/RediSearch/blob/f4a52c5adb23e0d155874198c73b089c8748d4b6/src/redisearch_rs/c_entrypoint/rlookup_ffi/src/row.rs#L461

from my understanding, it's set by the result processor, when SORT_BY is involved.

As it seems to impact only numeric iterator, I suggest to:
- add a non-regression test in current master
- create a new ticket to not forget about this.
- tackle this when we're closer to a more pragmatic implementation of numeric iterator
we may be able to advance without it for now and adapt when we hit a

Otherwise I can probably add an additional knob on this logic, but I don't like adding constraints when the path is still far, the code is already not too trivial.

src/redisearch_rs/top_k/src/heap.rs
GuyAv46: Maybe we should use [peek_mut](https://doc.rust-lang.org/beta/std/collections/struct.BinaryHeap.html#method.peek_mut)

src/redisearch_rs/top_k/src/heap.rs
ThierryBerger: I did as of MOD-15099

src/redisearch_rs/top_k/src/heap.rs
ThierryBerger: good catch, peek_mut is better! I'll add a commit.

As for using the existing min_heap, I'm worried about it becoming a rabbit hole, I think focusing on current migration and exact constraint makes more sense, then once both are in Rust, well tested, we can consider sharing behaviour.

The current `min_heap` for the union iterator explicitly says it's tailored for union, and has less requirements for comparison, ordering, even max capacity.

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -146,10 +146,13 @@ impl TopKHeap {
         // current worst (root). `entry > worst` means entry is worse → discard.
         // `entry < worst` means entry is better → evict worst, insert entry.
         // Equal (same score AND same doc_id) → discard to avoid duplicates.
-        else if self.inner.peek().is_some_and(|worst| &entry < worst) {
-            self.inner.pop();
-            self.inner.push(entry);
-            true
+        else if let Some(mut worst) = self.inner.peek_mut() {
+            if entry < *worst {
+                *worst = entry;
```
GuyAv46: A call to [PeekMut::refresh](https://doc.rust-lang.org/beta/std/collections/binary_heap/struct.PeekMut.html#method.refresh) is required (we can't guarantee the new entry should be at the root)

src/redisearch_rs/top_k/src/heap.rs
```diff
@@ -146,10 +146,13 @@ impl TopKHeap {
         // current worst (root). `entry > worst` means entry is worse → discard.
         // `entry < worst` means entry is better → evict worst, insert entry.
         // Equal (same score AND same doc_id) → discard to avoid duplicates.
-        else if self.inner.peek().is_some_and(|worst| &entry < worst) {
-            self.inner.pop();
-            self.inner.push(entry);
-            true
+        else if let Some(mut worst) = self.inner.peek_mut() {
+            if entry < *worst {
```
GuyAv46: Can we collapse the `if` here?

src/redisearch_rs/top_k/src/heap.rs
GuyAv46: Consider adding a unit test for that

ThierryBerger: ah yes :facepalm: ; but `PeekMut::refresh` is still unstable rust, which means only usable using nightly.

I did a test to compare both approaches, and indeed the refresh approach is more performant, for a test case with a heap of 100_000 random values, inserting (samefor both cases) 100_000 random values:
- 3ms for pop/push
- 2.5ms for peek_mut/refresh

So a significant improvement!

I'm glad we found it, but I'd advocate to keep a note on `MOD-14212`, and benchmark the actual win when we have a more advanced implementation.

I'll revert this change in the meantime, and add a note to that ticket so we're sure to check it again.

GuyAv46: ok. Is the `refresh` call really required? I'm seeing contradicting docs

GuyAv46: I ran a test locally, seems like the heap is fixed after mutating the reference:
```Rust
use std::collections::BinaryHeap;

fn main() {
    let mut heap = BinaryHeap::new();
    heap.push(1);
    heap.push(5);
    heap.push(2);

    println!("Initial heap (sorted asc): {:?}", heap.clone().into_sorted_vec());
    println!("Top before peek_mut: {:?}", heap.peek());

    if let Some(mut top) = heap.peek_mut() {
        *top += 10;
        println!("Inside peek_mut: modified top to {}", *top);
        // PeekMut guard is dropped here -> sift-down occurs
    }

    println!("Heap after peek_mut (sorted asc): {:?}", heap.clone().into_sorted_vec());
    println!("Top after peek_mut: {:?}", heap.peek());
    assert_eq!(heap.peek(), Some(&15));
    println!("Assertion 1 passed: top is 15\n");

    println!("--- Test 2: decrease the top, do we still sift? ---");
    let mut heap2 = BinaryHeap::new();
    heap2.push(10);
    heap2.push(5);
    heap2.push(8);
    println!("Before: top = {:?}, sorted = {:?}", heap2.peek(), heap2.clone().into_sorted_vec());

    if let Some(mut top) = heap2.peek_mut() {
        *top = 1; // 10 -> 1, so 8 should bubble up
    }
    println!("After : top = {:?}, sorted = {:?}", heap2.peek(), heap2.clone().into_sorted_vec());
    assert_eq!(heap2.peek(), Some(&8));
    println!("Assertion 2 passed: top is 8 (heap was re-sifted on Drop)");
}
```
the above prints
```log
Initial heap (sorted asc): [1, 2, 5]
Top before peek_mut: Some(5)
Inside peek_mut: modified top to 15
Heap after peek_mut (sorted asc): [1, 2, 15]
Top after peek_mut: Some(15)
Assertion 1 passed: top is 15

--- Test 2: decrease the top, do we still sift? ---
Before: top = Some(10), sorted = [5, 8, 10]
After : top = Some(8), sorted = [1, 5, 8]
Assertion 2 passed: top is 8 (heap was re-sifted on Drop)
```

ThierryBerger: indeed! PeekMut destructor is actually handling the refreshing of the heap, so when it goes out of scope at the end of our if, the heap is back to a correct state !

I'll keep the new behaviour then, and use chain let for the if :tada:

GuyAv46 (approved):

PR #9171, opened 2026-04-17
MOD-14089: Deferring the effectiveK calculation to the IO thread

nafraf:
This PR replaces #8456 
### Objective
The objective of these changes was to defer the calculus of `effectiveK` (used in the `SHARD_K_RATIO` optimization for KNN vector searches) to the IO thread. This is because the correct calculation depends on the exact cluster topology (number of participating shards) at the exact moment the command is dispatched, rather than an unsafe estimation from the coordinator's main thread.

#### 1. FT.AGGREGATE (`src/coord/dist_aggregate.c`)
* **Before:** `buildMRCommand` synchronously called `GetNumShards_UnSafe()` to compute `effectiveK` and immediately mutated the `MRCommand` with the new K value.
* **After:** `buildMRCommand` no longer modifies K. Instead, it records the index of the K argument in the command. A new structure `AggregateKnnContext` holds a scalar snapshot of the KNN metadata the IO thread needs (`originalK`, `shardWindowRatio`, the K argument index, and the byte offset/length of the K token within the query string) — no borrowed `VectorQuery` pointer.
* `rpnetNext_Start` creates a wrapper `AggregateIteratorContext` and passes a new `MRCommandModifier` callback named `aggregateKnnCommandModifier` to `MR_IterateWithPrivateData`.
* When the IO thread starts processing, it invokes `aggregateKnnCommandModifier` with the **actual** `numShards` derived from the IO thread's exact topology snapshot. The callback computes `calculateEffectiveK` and edits the command before dispatching it to the shards.

#### 2. FT.HYBRID (`src/coord/hybrid/dist_hybrid.c` & `hybrid_cursor_mappings.c`)
* **Before:** `HybridRequest_buildMRCommand` computed `effectiveK` locally with a supplied `numShards` variable and changed the command.
* **After:** It just saves `kArgIndex` (the position of K in the command).
* `HybridRequest_executePlan` initializes a `HybridKnnContext` which is forwarded to `ProcessHybridCursorMappings`.
* A new `MRCommandModifier` callback `HybridKnnCommandModifier` is passed into `MR_IterateWithPrivateData`. Similar to the aggregate flow, the IO thread invokes this callback, calculates `effectiveK` securely using its locally-resolved `numShards`, and rewrites the command text using `MRCommand_ReplaceArg`.

#### 3. Topology State & Thread Safety (`src/coord/shard_barrier.c/h`)
* Extracted barrier logic into `shard_barrier.c/h`.
* Created a fundamental `ShardCountBarrier` that tracks `numShards` and `numResponded`.
* Using `shardCountBarrier_Init` on the IO thread, the coordinator's expectation of `numShards` is populated via `atomic_store` securely. This prevents the main thread from making race-condition guesses about how many shards are present.

#### 4. Testing Updates
* Adapted `test_cpp_hybrid_build_mr_cmd.cpp` to explicitly mock the IO thread callback (`HybridKnnCommandModifier`) to verify that the command mutates perfectly.
* Added new Python tests (`test_hybrid_timeout.py`) simulating delayed/blocking shards to verify that `ProcessHybridCursorMappings` cleanly respects timeouts waiting for barrier responses.

#### Release Notes

Deferring the effectiveK calculation to the IO thread

---

> [!NOTE]
> **Medium Risk**
> Changes distributed query dispatch and KNN command rewriting on cluster coordinators; incorrect indices or topology timing could affect result counts, though behavior is tightened to match dispatch-time shard count.
> 
> **Overview**
> **SHARD_K_RATIO** no longer rewrites K on the coordinator main thread using a guessed shard count. **FT.AGGREGATE** and **FT.HYBRID** now snapshot KNN metadata (query arg index, original **K**, ratio, token offset/length or VSIM **kArgIndex**) and apply **`calculateEffectiveK`** only when the map-reduce iterator starts on the **IO thread**, via a new **`MRCommandModifier`** hook in **`MR_IterateWithPrivateData`** (after live topology **`numShards`** is known).
> 
> For **FT.AGGREGATE**, **`buildMRCommand`** stops mutating the query string; **`rpnetNext_Start`** uses an **`AggregateIteratorContext`** that can hold both **WITHCOUNT** barrier state and optional KNN context, with **`aggregateKnnCommandModifier`** and a cursor callback that unwraps the barrier. **`_FT.DEBUG`** paths bump the saved query argument index when **`_FT.DEBUG`** is inserted at the front.
> 
> For **FT.HYBRID**, command build records **`kArgIndex`** only; **`HybridKnnApplyShardKRatio`** / **`HybridKnnCommandModifier`** run during **`ProcessHybridCursorMappings`**. Coordinator code no longer threads **`numShards`** from the main thread into **`HybridRequest_prepareForExecution`**.
> 
> **`modifyKNNCommand`** now takes explicit **originalK**, **effectiveK**, and byte positions (not a **`VectorQuery`** pointer), aligning aggregate KNN rewrites with substring replacement and fixing wrong-slot edits when arguments shift (e.g. long index names under **`_FT.DEBUG`**). **`ShardResponseBarrier`** moves to **`shard_barrier.c/h`** with a shared **`ShardCountBarrier`** base. **`MR_Iterate`** is removed; callers always pass a command modifier (or **NULL**). Tests cover the deferred modifier path and debug aggregate/hybrid regressions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 18d8cf97e3f6d6143f44088e231dd944ba4f6cd5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/coord/dist_aggregate.c
+++ b/src/coord/dist_aggregate.c
@@ -81,14 +81,10 @@ static void aggregateIteratorContext_Init(void *ptr, const MRIterator *it) {
 // Command modifier callback for SHARD_K_RATIO optimization in FT.AGGREGATE
 // Called from iterStartCb on IO thread before commands are sent to shards
 static void aggregateKnnCommandModifier(MRCommand *cmd, size_t numShards, void *privateData) {
-  if (!privateData || !cmd) {
-    return;
-  }
+  RS_ASSERT(privateData && cmd);
   AggregateIteratorContext *ctx = (AggregateIteratorContext *)privateData;
   AggregateKnnContext *knnCtx = ctx->knnCtx;
-  if (!knnCtx || !knnCtx->vq) {
-    return;
-  }
+  RS_ASSERT(knnCtx && knnCtx->vq);
   const KNNVectorQuery *knn_query = &knnCtx->vq->knn;
   // Only apply optimization for multi-shard deployments with valid ratio
   if (numShards <= 1 || knn_query->shardWindowRatio >= MAX_SHARD_WINDOW_RATIO) {
--- a/src/coord/dist_aggregate.c
+++ b/src/coord/dist_aggregate.c
@@ -44,6 +44,78 @@ static const RLookupKey *keyForField(RPNet *nc, const char *s) {
   return NULL;
 }
 
+// Context for SHARD_K_RATIO optimization in FT.AGGREGATE
+// Stores information needed to modify the KNN K value in the command
+typedef struct {
+  VectorQuery *vq;        // VectorQuery containing K position info (NOT owned)
+  size_t queryArgIndex;   // Index of the query argument in the MRCommand
+} AggregateKnnContext;
+
+// Combined context for MR_IterateWithPrivateData in FT.AGGREGATE
+// Contains optional barrier (for WITHCOUNT) and optional KNN context (for SHARD_K_RATIO)
+typedef struct {
+  ShardResponseBarrier *barrier;  // May be NULL if WITHCOUNT not enabled
+  AggregateKnnContext *knnCtx;    // May be NULL if no KNN optimization needed
+} AggregateIteratorContext;
+
+// Free the AggregateIteratorContext and its contents
+static void aggregateIteratorContext_Free(void *ptr) {
+  AggregateIteratorContext *ctx = (AggregateIteratorContext *)ptr;
+  if (ctx) {
+    if (ctx->barrier) {
+      shardResponseBarrier_Free(ctx->barrier);
+    }
+    rm_free(ctx->knnCtx);  // knnCtx->vq is not owned, so just free the struct
+    rm_free(ctx);
+  }
+}
+
+// Initialize the barrier in AggregateIteratorContext (called from iterStartCb)
+static void aggregateIteratorContext_Init(void *ptr, const MRIterator *it) {
+  AggregateIteratorContext *ctx = (AggregateIteratorContext *)ptr;
+  if (ctx && ctx->barrier) {
+    shardResponseBarrier_Init(ctx->barrier, it);
+  }
+}
+
+// Command modifier callback for SHARD_K_RATIO optimization in FT.AGGREGATE
+// Called from iterStartCb on IO thread before commands are sent to shards
+static void aggregateKnnCommandModifier(MRCommand *cmd, size_t numShards, void *privateData) {
+  if (!privateData || !cmd) {
+    return;
+  }
+  AggregateIteratorContext *ctx = (AggregateIteratorContext *)privateData;
+  AggregateKnnContext *knnCtx = ctx->knnCtx;
+  if (!knnCtx || !knnCtx->vq) {
+    return;
+  }
+  const KNNVectorQuery *knn_query = &knnCtx->vq->knn;
+  // Only apply optimization for multi-shard deployments with valid ratio
+  if (numShards <= 1 || knn_query->shardWindowRatio >= MAX_SHARD_WINDOW_RATIO) {
+    return;
+  }
+  size_t effectiveK = calculateEffectiveK(knn_query->k, knn_query->shardWindowRatio, numShards);
+  if (effectiveK == knn_query->k) {
+    return;
+  }
+
+  // Modify the command to replace KNN k
+  modifyKNNCommand(cmd, knnCtx->queryArgIndex, effectiveK, knnCtx->vq);
+}
+
+// Aggregate-specific cursor callback that extracts ShardResponseBarrier from
+// AggregateIteratorContext
+// This wraps the common netCursorCallback logic but correctly handles the
+// wrapper context type
+static void aggregateNetCursorCallback(MRIteratorCallbackCtx *ctx, MRReply *rep) {
+  // Extract the actual ShardResponseBarrier from the AggregateIteratorContext wrapper
+  AggregateIteratorContext *iterCtx = (AggregateIteratorContext *)MRIteratorCallback_GetPrivateData(ctx);
+  ShardResponseBarrier *barrier = iterCtx ? iterCtx->barrier : NULL;
+
+  // Call the common cursor callback logic with the extracted barrier
+  netCursorCallbackWithBarrier(ctx, rep, barrier);
+}
+
 void processResultFormat(uint32_t *flags, MRReply *map) {
   // Logic of which format to use is done by the shards
   MRReply *format = MRReply_MapElement(map, "format");
@@ -59,6 +131,23 @@ void processResultFormat(uint32_t *flags, MRReply *map) {
 static int rpnetNext_Start(ResultProcessor *rp, SearchResult *r) {
   RPNet *nc = (RPNet *)rp;
 
+  // Create the iterator context wrapper for privateData
+  // This holds both optional barrier (for WITHCOUNT) and optional KNN context (for SHARD_K_RATIO)
+  AggregateIteratorContext *iterCtx = rm_calloc(1, sizeof(AggregateIteratorContext));
+  if (!iterCtx) {
+    return RS_RESULT_ERROR;
+  }
+
+#ifdef ENABLE_ASSERT
+  // Sync point (debug): park BG just before the initial timeout check.
+  SyncPoint_WaitUntil(SYNC_POINT_BEFORE_RPNET_START, areq_timed_out, nc->areq);
+#endif
+
+  // Check if the request timed out before starting the iterator
+  if (AREQ_TimedOut(nc->areq)) {
+    return RS_RESULT_TIMEDOUT;
+  }
+
 #ifdef ENABLE_ASSERT
   // Sync point (debug): park BG just before the initial timeout check.
   SyncPoint_WaitUntil(SYNC_POINT_BEFORE_RPNET_START, areq_timed_out, nc->areq);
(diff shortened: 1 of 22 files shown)


src/coord/rpnet.c
```diff
@@ -319,7 +229,15 @@ int getNextReply(RPNet *nc) {
...
     }
-    shardResponseBarrier_UpdateTotalResults(nc);
+
+    // Set the accumulated total now that all shards have responded
+    // numShards == 0 means IO thread never initialized the barrier (timeout before init)
+    size_t numResponded = atomic_load(&nc->shardResponseBarrier->base.numResponded);
+    size_t numShards_loaded = atomic_load(&nc->shardResponseBarrier->base.numShards);
+    if (numShards_loaded > 0 && numResponded >= numShards_loaded) {
+      long long accumulatedTotal = atomic_load(&nc->shardResponseBarrier->accumulatedTotal);
+      nc->base.parent->totalResults = accumulatedTotal;
```
oshadmi: Code was moved out of `shardResponseBarrier_UpdateTotalResults` ?

src/coord/rmr/rmr.h
```diff
@@ -120,6 +120,18 @@ typedef struct MRIterator MRIterator;
...
+ * Callback type for modifying commands before they are sent to shards.
+ * Called from iterStartCb on the IO thread after numShards is known but before
+ * commands are sent.
+ * This allows calculating values like effectiveK based on the actual topology.
+ *
+ * @param cmd The command to modify (will be copied for each shard after this callback)
+ * @param numShards The actual number of shards from the IO thread's topology
+ * @param privateData The private data passed to MR_IterateWithPrivateData
+ */
+typedef void (*MRCommandModifier)(MRCommand *cmd, size_t numShards, void *privateData);
```
JoanFM, line 144: Should we get the topology instead of the num_shards? Like this we can potentially fit other transformations in the future?

src/coord/rmr/rmr.c
```diff
@@ -768,6 +769,11 @@ void iterStartCb(void *p) {
   // Set the dispatch time value in the prepared placeholder
   MRCommand_SetDispatchTime(cmd);
 
+  // Call command modifier callback if set (e.g., to calculate effectiveK based on actual numShards)
+  if (it->ctx.commandModifier) {
+    it->ctx.commandModifier(cmd, numShards, privateData);
```
JoanFM, line 782: Should the slot info setter bellow, be somehow part of the commandModifier? Maybe in a compositional manner?

src/coord/dist_aggregate.c
```diff
@@ -55,31 +123,55 @@ void processResultFormat(uint32_t *flags, MRReply *map) {
 static int rpnetNext_Start(ResultProcessor *rp, SearchResult *r) {
   RPNet *nc = (RPNet *)rp;
 
+  // Create the iterator context wrapper for privateData
+  // This holds both optional barrier (for WITHCOUNT) and optional KNN context (for SHARD_K_RATIO)
+  AggregateIteratorContext *iterCtx = rm_calloc(1, sizeof(AggregateIteratorContext));
+  if (!iterCtx) {
```
JoanFM: I believe we never check this. rm_calloc would abort if it cannot allocate. I would remove just to keep coherent with rest of codebase. (Maybe we need to add this into one of our skills)

src/coord/dist_aggregate.c
```diff
@@ -55,31 +123,55 @@ void processResultFormat(uint32_t *flags, MRReply *map) {
...
+      return RS_RESULT_ERROR;
+    }
+    iterCtx->barrier = barrier;
+    nc->shardResponseBarrier = barrier;  // Keep reference for getNextReply
+  }
+
+  // Initialize KNN context if SHARD_K_RATIO optimization is needed
+  if (nc->knnVectorQuery) {
+    AggregateKnnContext *knnCtx = rm_calloc(1, sizeof(AggregateKnnContext));
+    if (!knnCtx) {
```
JoanFM: same here

src/coord/dist_aggregate.c
```diff
@@ -55,31 +123,55 @@ void processResultFormat(uint32_t *flags, MRReply *map) {
...
+
   // Initialize shard response barrier if WITHCOUNT is enabled
   if (HasWithCount(nc->areq) && IsAggregate(nc->areq)) {
     ShardResponseBarrier *barrier = shardResponseBarrier_New();
     if (!barrier) {
+      rm_free(iterCtx);
+      return RS_RESULT_ERROR;
+    }
+    iterCtx->barrier = barrier;
+    nc->shardResponseBarrier = barrier;  // Keep reference for getNextReply
```
JoanFM: Is ownership and lifetime clear here?

src/coord/dist_aggregate.c
```diff
@@ -88,26 +180,55 @@ static int rpnetNext_Start(ResultProcessor *rp, SearchResult *r) {
...
+    }
+  }
+  return profileArgs;
+}
+
+// Extract KNN context for SHARD_K_RATIO optimization if applicable
+// Outputs VectorQuery and query arg index for command modifier
+static void extractKnnOptimizationContext(specialCaseCtx *knnCtx, ProfileOptions profileOptions,
+                                          VectorQuery **outKnnVq, size_t *outQueryArgIndex) {
+  if (outKnnVq) *outKnnVq = NULL;
```
JoanFM: should this initialization be in the caller? (In a review made by AI for instance, it will look forproper initialization in the caller?)

src/coord/rpnet.c
JoanFM: Yes, why was it moved? Isn't it better to put this in some smaller method?

JoanFM (changes requested):
LGTM! Only change I would 100% do is the one orelated to extracting logic from shardResponseBarrier_UpdateTotalResults without a new method.

For the rest of comments, feel free to ignore.

src/coord/rpnet.c
nafraf: The function was restored. [f2830f5](https://github.com/RediSearch/RediSearch/pull/9171/commits/f2830f59005c43a1711096e4b3c2bdc74143d1ac)

src/coord/dist_aggregate.c
nafraf: Right! those checks were removed. Thanks. [4d9f80a](https://github.com/RediSearch/RediSearch/pull/9171/commits/4d9f80a82ac23692801ecd1a0718c599834ebfff)

src/coord/dist_aggregate.c
nafraf: Removed. [4d9f80a](https://github.com/RediSearch/RediSearch/pull/9171/commits/4d9f80a82ac23692801ecd1a0718c599834ebfff)

src/coord/rmr/rmr.h
nafraf: Today `numShards` is enough, I prefer to change it only in the future if needed.

src/coord/rmr/rmr.c
nafraf: I think we shouldn't.  `commandModifier` is optional and executed once per `MRIterator` startup.  The slot info setter is executed per shard and it's mandatory.

src/coord/hybrid/dist_hybrid.c
```diff
@@ -657,15 +651,11 @@ static int HybridRequest_prepareForExecution(HybridRequest *hreq,
...
-    // Note: parsedVectorData is only set on shards, not on coordinator
-    // The coordinator has the VectorQuery in the AST after parsing
-    const AREQ *vectorRequest = hreq->requests[VECTOR_INDEX];
-    const VectorQuery *vq = (vectorRequest->ast.root && vectorRequest->ast.root->type == QN_VECTOR)
-                      ? vectorRequest->ast.root->vn.vq : NULL;
+    // Track K argument index for SHARD_K_RATIO optimization (set during command building)
+    int kArgIndex = -1;
     HybridRequest_buildMRCommand(argv, argc, profileOptions, &xcmd, serialized,
-                                 sp, vq, numShards);
+                                 sp, &kArgIndex);
```
JoanFM:
```suggestion
                                 sp, &hreq->kArgIndex);
```

JoanFM: We can avoid having this extra variable?

src/coord/hybrid/dist_hybrid.c
```diff
@@ -731,11 +725,28 @@ static int HybridRequest_executePlan(HybridRequest *hreq, struct ConcurrentCmdCt
     MRCommand *cmd = &searchRPNet->cmd;
     cmd->coordStartTime = hreq->profileClocks.coordStartTime;
 
+    // Create KNN context for SHARD_K_RATIO optimization if applicable
+    HybridKnnContext *knnCtx = NULL;
```
JoanFM, line 726: Could u check, I think this knnCtx could live in the stack and we may not need to do a heap allocation for it.

Your call.

JoanFM (dismissed):
Small comemnts, ur call to change

src/coord/hybrid/dist_hybrid.c
nafraf: OK. Extra variable removed.

oshadmi (approved):

PR #9175, opened 2026-04-18
[MOD-15533] Use cheadergen rather than cbindgen

LukeMathWalker:

Migrate Rust → C FFI header generation from [cbindgen](https://github.com/mozilla/cbindgen) to [cheadergen](https://github.com/LukeMathWalker/cheadergen). 

### What we collapsed

| | cbindgen (master) | cheadergen (this PR) |
|---|---|---|
| Per-crate `cbindgen.toml` files | 22 | 0 |
| Per-crate `build.rs` files | 23 | 0 |
| Total config lines | 975 (`cbindgen.toml` + `build.rs`) | 286 (single `cheadergen.toml`) |

The 22 per-crate `cbindgen.toml` files collapse into one `cheadergen.toml` with 6 `[package.*]` overrides and 21 `[header.*]` overrides. The 33 type renames scattered across cbindgen tomls move closer to the Rust source as 46 `#[cheadergen::config(...)]` annotations (36 `rename = "..."`, 7 `export`, 5 `prefix_with_name`, 3 `skip`, etc.) directly on the affected types.

The 51 `export.exclude` entries in cbindgen tomls **go away entirely**: see the next section on how cheadergen partitioning handles cross-crate references.

### Headers — crate-aligned partitioning

22 generated headers → 31, factored 1:1 along crate lines. Each crate that exports C-visible items now owns exactly one header — e.g. types defined in the `inverted_index` crate land in `inverted_index.h`, FFI shims in `inverted_index_ffi` land in `inverted_index_ffi.h`, and so on for `rlookup` / `rlookup_ffi`, `query_error` / `query_error_ffi`, etc. The split between "types" (`*.h`) and "FFI surface" (`*_ffi.h`) falls out naturally from the workspace structure rather than being enforced by hand.

This is possible because cheadergen runs in **partitioned mode** over the whole workspace and resolves cross-crate type references automatically:

- cbindgen processed one crate at a time. When crate A's signatures referenced a type owned by crate B, you had to tell cbindgen to parse crate B as a dep (`parse.parse_deps = true`, `parse.include = ["crate_b"]`), then `export.exclude` each of those types from A's output to prevent duplicate definitions, then hand-add `#include "crate_b.h"` to A's `after_includes` so the consumer could still see them. That's where the 51 excludes and 85 parse-deps entries came from.
- cheadergen owns the workspace graph end-to-end: every C-level type has exactly one owning header, cross-crate references in fn signatures auto-emit `#include "owner.h"` at the top of the consuming header, and the per-package `cheadergen.toml` overrides only kick in for the few cases where we genuinely want non-default behaviour (skipping a crate wholesale, renaming the header file, etc.).

### Build flow

- **Before:** each FFI crate's `build.rs` invoked cbindgen during `cargo build`.
- **After:** cheadergen runs once from `src/redisearch_rs/CMakeLists.txt`'s `cheadergen_generate` custom target, after the Rust workspace is compiled. The same invocation is exposed as `make generate-rust-headers`. 
  - Downstream consumers including RediSearch as a CMake subdirectory get `REDISEARCH_GENERATE_HEADERS=OFF` by default — no cheadergen install required.
  - CI's lint job runs `make generate-rust-headers` and the existing `Generated files` `git diff` step catches drift, guaranteeing committed headers stay fresh on master.

### CI integration tweaks

A handful of environment leaks surfaced once header generation moved into the CMake custom target. Each fix is one-line; collectively they make the `SAN=address` flow self-consistent:

- `src/redisearch_rs/CMakeLists.txt` — scrub `RUSTFLAGS` and `CARGO_BUILD_TARGET` for the `cheadergen_generate` invocation (`env -u CARGO_BUILD_TARGET RUSTFLAGS=`). Otherwise the parent build's sanitizer flags leak into cheadergen's internal `cargo metadata` / `cargo doc` calls.
- `tests/deps/setup_rejson.sh` — same env scrub for the RedisJSON build, otherwise cargo writes the dylib to `target/<triple>/release/` while the Makefile's cp expects `target/release/`.

### Other changes

- **`INDEX_STORAGE_MASK`** previously lived in both `src/spec.h` (C `#define`) and Rust (`inverted_index_ffi`). Now exported from Rust via cheadergen with `#[cheadergen::config(export)] pub const`; the C-side define is removed. Same for `RS_SORTABLES_MAX`.

1. MOD-...
2. #...

1. ...

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **High Risk**
> High risk because it overhauls the Rust/C FFI header generation pipeline and touches many core C translation units’ includes and type/field names, which can easily break builds or ABI expectations across platforms and downstream consumers.
> 
> **Overview**
> Migrates Rust→C FFI header generation from per-crate `cbindgen` (with many `build.rs`/`cbindgen.toml` files) to a single workspace-wide `cheadergen` invocation driven by `src/redisearch_rs/regen_headers.sh` and configured via the new `src/redisearch_rs/cheadergen.toml`.
> 
> Build integration is updated so header generation runs as a CMake custom target (`cheadergen_generate`) with a new top-level `REDISEARCH_GENERATE_HEADERS` option (default *ON* only for top-level builds), and `make lint` now regenerates headers via a new `generate-rust-headers` target; CI installs `cheadergen_cli` + `rust-docs-json` and improves the stale-header failure message.
> 
> The C/C++ codebase is switched broadly from legacy headers (e.g. `value.h`, `query_error.h`, `iterators_rs.h`) to the new generated `*_ffi.h`/crate-aligned headers, with a few accompanying API shape updates (e.g. renamed struct fields like `field_mask`→`fieldmask`, `hybrid_metric`→`hybridmetric`) and some constants exported from Rust instead of duplicated C definitions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 75fc3d602b6a8faca6dc27c5f3eb66fbe519d579. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/redisearch_rs/ffi/build.rs
```diff
@@ -122,27 +122,37 @@ fn main() {
     let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
     bindings
         .blocklist_file(".*/document_rs.h")
-        // numeric_range_tree.h is generated by Rust (cbindgen) and those types
+        // numeric_range_tree.h is generated by Rust (cheadergen) and those types
         // are provided by the numeric_range_tree_ffi crate, not parsed from C
         .blocklist_file(".*/numeric_range_tree.h")
```
chesedo, line 127: Can't we mass exclude all of them using something like `.blocklist_file(".*/headers/.*")`

src/redisearch_rs/headers/inverted_index_ffi.h
```diff
@@ -0,0 +1,555 @@
...
+  void (*call)(void *ctx);
+} II_GCCallback;
+
+/**
+ * The mask of flags that determine the index storage type. This includes all flags that affect
+ * the storage format of the index.
+ */
+#define INDEX_STORAGE_MASK 179
+
+#define FULL_MASK 19
```
chesedo: Are we re-exporting some C definitions here?

src/redisearch_rs/cheadergen.toml
```diff
@@ -0,0 +1,286 @@
...
+includes = ["redisearch_types.h"]
+
+[header.numeric_range_tree_ffi]
+includes = ["redisearch.h", "redismodule.h"]
+
+[header.search_result_ffi]
+includes = ["redisearch.h", "score_explain.h"]
+
+[header.slots_tracker_ffi]
+includes = ["redismodule.h"]
```
chesedo, line 246: Technically this header should come from `redismodule_rs` not sure if we can do that? No biggie if we can't

src/redisearch_rs/CMakeLists.txt
```diff
@@ -77,6 +77,37 @@ if(NOT TARGET redisearch_rs_build)
...
+            --exclude=c_ffi_utils
+            --package=document
+            --package=field
+            --package=query_term
+            --package=inverted_index
+            --package=rlookup
+            --package=rqe_iterator_type
+            --package=rqe_iterators
+            --package=search_result
+            --package=query_node_type
```
chesedo: Will we have to manually extend this when adding new FFI crates?

src/redisearch_types.h
```diff
@@ -0,0 +1,31 @@
...
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+// Fundamental type aliases used across both C code and Rust-generated headers.
+// Extracted from redisearch.h to break circular include dependencies:
+// generated headers need these types, but redisearch.h includes generated headers.
```
chesedo, line 12: Technically bindgen (our FFI crate) should not include `redisearch.h`. I've just not been able to figure out why removing it causes a compile error for the santizer here

https://github.com/RediSearch/RediSearch/pull/9142

chesedo (commented):
Can't wait for this. Some comments

src/redisearch_rs/ffi/build.rs
LukeMathWalker: I _think_ we need some of them. We definitely need to take some time to clean up the `ffi` crate.

src/redisearch_rs/headers/inverted_index_ffi.h
LukeMathWalker: We are (and we get a build warning for it). I need to move constants from "auto-export" to "opt-in export" and then I can prune them down.

src/redisearch_rs/cheadergen.toml
LukeMathWalker: As in, generate a header off the Rust code in `redismodule_rs`?

src/redisearch_rs/CMakeLists.txt
LukeMathWalker: Nope! All _ffi crates are automatically included by virtue of their path (see `c_entrypoint` at the end of the command). We need to manually include all packages that are not `_ffi` but have types annotated with `#[cheadergen::config(export)]`.
This is needed for types that, based on their usage in our own headers, would only need an opaque forward declaration, but C instead needs to see their internals.

We can "solve" this by doing `cheadergen generate --workspace` but that would compute a lot of unnecessary docs.

src/redisearch_rs/cheadergen.toml
chesedo: No, including `redismodule.h` from `redismodule_rs` if possible. The repo `redismodule.h` is a duplicate of that one

chesedo: ... which sometimes goes out of sync

src/redisearch_rs/CMakeLists.txt
chesedo: Okay :+1:

src/redisearch_rs/cheadergen.toml
LukeMathWalker: I'd say that's an orthogonal concern—this header should include `redismodule.h`, then the CMake-powered build system should make sure that comes from `redismodule_rs`

src/redisearch_rs/headers/inverted_index_ffi.h
LukeMathWalker: Switched constants to opt-in export, and changed INDEX_STORAGE_MASK to use Rust as its source of truth.

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/fork_gc.rs
```diff
@@ -17,6 +17,7 @@ use std::{
 pub type c_size_t = usize;
 
 /// A writer that calls a C function to write data.
+#[cheadergen::config(export, rename = "II_GCWriter")]
```
JonasKruckenberg, line 17: just stylistically, I would prefer 
`#[cheadergen(...)]` and being able to split the attrs onto two lines like
```rust
#[cheadergen(export)]
#[cheadergen(rename = "II_GCWriter")]
```

LukeMathWalker: Splitting I can accommodate (with a bit more work upstream), but I'd prefer to avoid having the name of the crate as the name of the macro too.

JonasKruckenberg: Hmm fair enough, how about cheadergen::export()?

I did think of a potential footgun here btw, not sure if you already covered this:
Imagine you add just #[cheadergen::config(rename = "")] that would not export the type (hopefully) but does it raise an error? If not it should

If you go with cheadergen::export() (which might have other problems I just don't know yet) that ofc wouldn't be an issues

LukeMathWalker: It is not called `cheadergen::export` because you can use the attribute to explicitly not export a type 😛 
E.g. `cheadergen::config(skip)`. Or to customize how a type is exported _if_ it needs to be exported (e.g. `cheadergen::config(rename)` won't force it to be visible, it'll just affect how it's named _if_ it needs to be C visible).

LukeMathWalker: The reference is here btw: https://docs.rs/cheadergen/latest/cheadergen/attr.config.html

src/redisearch_rs/headers/fnv_ffi.h
JonasKruckenberg, line 1: did we actually port this without swapping the header file? 🧐 either way, good change 👍🏻

src/redisearch_rs/headers/idf_ffi.h
JonasKruckenberg, line 1: I do think cheadergen should emit a "this file is automatically generated" comment no? (ideally with a link to the corresponding source file automatically!)

src/redisearch_rs/headers/inverted_index.h
JonasKruckenberg, line 1: general question about the generated headers (sorry if that's stated somewhere and I missed it): how do we sort symbols in the header? is it by appearance in the source file?

src/redisearch_rs/headers/inverted_index_ffi.h
```diff
@@ -0,0 +1,529 @@
...
+
+/**
+ * An opaque inverted index reader structure. The actual implementation is determined at runtime
+ * based on the index type and filter provided when creating the reader. This allows us to have a
+ * single interface for all index reader types while still being able to optimize the storage
+ * and performance for each index reader type.
+ */
+typedef struct IndexReader IndexReader;
+
+typedef size_t c_size_t;
```
JonasKruckenberg: what is this rename for here?

src/redisearch_rs/headers/iterators_ffi.h
```diff
@@ -1,45 +1,82 @@
...
 #include "iterators/iterator_api.h"
 #include "tag_index.h"
-#include "numeric_range_tree.h"
 #include "query.h"
+#include "field.h"
+#include "query_node_type.h"
+#include "rqe_iterators.h"
+// In C, timespec is a struct tag, not a typedef. Rust's libc::timespec maps to
+// the bare name, so we introduce a typedef to make it valid C.
+typedef struct timespec timespec;
```
JonasKruckenberg, line 20: is this part of the auto-generated code? seems quite informal for that idk

src/redisearch_rs/headers/rlookup.h
```diff
@@ -0,0 +1,343 @@
...
+#  if defined(__cplusplus) && __cplusplus >= 201103L
+#    define CHEADERGEN_ALIGNED(n) alignas(n)
+#  elif defined(_MSC_VER)
+#    define CHEADERGEN_ALIGNED(n) __declspec(align(n))
+#  elif defined(__GNUC__) || defined(__clang__)
+#    define CHEADERGEN_ALIGNED(n) __attribute__((aligned(n)))
+#  else
+#    error "cheadergen: don't know how to express alignment for this compiler"
+#  endif
+#endif
```
JonasKruckenberg, line 41: consider adding a comment to the generated code that says something like "macro to produce a compatible align attribute because there is no standard way in C" or something and link to an issue or docs in the cheadergen repo? Just so readers are not confused about the origin of this quite nasty looking blob of code

src/redisearch_rs/headers/iterators_ffi.h
LukeMathWalker: We add this via configuration in `cheadergen.toml` because, just like in `cbindgen`, the forward-reference we get out of the libc crate doesn't include the `struct` prefix and therefore doesn't compile.

src/redisearch_rs/headers/rlookup.h
LukeMathWalker: A comment is definitely worth it. I wonder if it'd be better to emit an additional header that defines this macro and just include it from there? E.g. `cheadergen_utils.h`?

src/redisearch_rs/headers/inverted_index.h
LukeMathWalker: By default, it's source order for functions, statics and constants (see https://docs.rs/cheadergen/latest/cheadergen/config_reference/index.html#sort_by). You can override it, if desired.
Types are topologically sorted to ensure that all fields are defined before the structs that reference them.

src/redisearch_rs/headers/sorting_vector.h
JonasKruckenberg, line 1: shouldn't we delete this file too if its superseded by `sorting_vector_ffi.h`?

src/redisearch_rs/rlookup/src/row.rs
```diff
@@ -25,7 +25,14 @@ fn is_special_key(rule: &SchemaRule, key: &RLookupKey) -> bool {
...
 /// Row data for a lookup key. This abstracts the question of if the data comes from a borrowed sorting vector slice
 /// or from dynamic values stored in the row during processing.
+///
+/// The C-visible layout comes from the `OpaqueRLookupRow` wrapper (which is
+/// `#[repr(C, align(8))]` and `#[cheadergen::config(rename = "RLookupRow")]`).
+/// Force this internal type to opaque so cheadergen does not warn about it
+/// being reached by value through `SearchResult._row_data` — both names emit
+/// to the same C tag, and the wrapper provides the actual layout.
 #[derive(Debug)]
+#[cheadergen::config(export(opaque))]
```
JonasKruckenberg: this feels kinda weird, it should be an attribute  on the macro instead of an "arg" to `export` IMO

src/redisearch_rs/headers/sorting_vector.h
LukeMathWalker: There is a 1:1 mapping between crates and header files; that's why you get `sorting_vector.h` with the type definitions. 
`sorting_vector.h` will automatically be included wherever we use an `RSSortingVector` (e.g. as a field), unlike `sorting_vector_ffi.h`.

src/redisearch_rs/Cargo.toml
```diff
@@ -145,8 +145,8 @@ approx = "0.6.0-rc2"
 bindgen = { version = "0.72", default-features = false }
 cargo_metadata = "0.19"
 bytecount = "0.6.9"
-cbindgen = "0.29"
 cc = "1.2.53"
+cheadergen = { git = "https://github.com/LukeMathWalker/cheadergen.git" }
```
JonasKruckenberg: should at least pin a git hash here

src/redisearch_rs/headers/inverted_index_ffi.h
LukeMathWalker: Generated from:
```rust
// This alias can be removed once it lands in stable Rust
#[expect(non_camel_case_types)]
pub type c_size_t = usize;
```
in `inverted_index_ffi/src/fork_gc.rs`.

src/redisearch_rs/Cargo.toml
LukeMathWalker: Ups, my bad, this needs to be switched over to version dep.

src/redisearch_rs/rlookup/src/row.rs
LukeMathWalker: As in, `export, opaque` or just `opaque`?

src/redisearch_rs/headers/inverted_index_ffi.h
LukeMathWalker: We can most likely remove it. Did you want `size_t` or the pointer size as the C type there @chesedo?

.rust-nightly
JonasKruckenberg, line 1: is this required for cheadergen to work correctly?

src/redisearch_rs/headers/inverted_index.h
JonasKruckenberg: okay good!

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/fork_gc.rs
JonasKruckenberg: hmmm yeah okay I get that, its just quite wordy which I don't love. (also personally would have made every export opt-in but that's a different discussion)

if we just had export/skip I would say we should go with `cheadergen::export` and `cheadergen::skip` but with rename idk 🤔

.rust-nightly
LukeMathWalker: Yes, a given version of cheadergen requires a specific nightly. To minimize CI churn, it's easier to align the overall nightly to that version too.

src/redisearch_rs/headers/iterators_ffi.h
JonasKruckenberg: okay fair enough, repo internal this is good

.rust-nightly
LukeMathWalker: If you get it wrong, the error message from `cheadergen` will be very clear as to what toolchain it expected.

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/fork_gc.rs
LukeMathWalker: `config` is a bit more verbose, but it's "neutral" and can be extended in a non-breaking way going forward.

src/redisearch_rs/cheadergen.toml
```diff
@@ -0,0 +1,286 @@
...
+  rs_wall_clock_ns_t queryGILTime;
+  double minScore;
+  uint32_t totalResults;
+  uint32_t resultLimit;
+  QueryError *err;
+  bool bgScanOOM;
+  bool isProfile;
+  RSTimeoutPolicy timeoutPolicy;
+  bool canYieldPartialResults;
+} QueryProcessingCtx;
```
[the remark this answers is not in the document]
LukeMathWalker: Correct. The fix is moving that type out of `ffi`, but that's for a follow-up PR.

src/redisearch_rs/rlookup/src/row.rs
JonasKruckenberg: well idk, I guess there are orthogonal attributes no? `#[cheadergen::config(opaque)]` would just say "if you export this type export it as an opaque" and `#[cheadergen::config(export)]` just forces an export right? 

You could argue that opaque is tied to exportedness and doesn't make much sense otherwise but the same hold for `rename` so they should be treated the same.

src/redisearch_rs/headers/rlookup.h
JonasKruckenberg: do we expect to add more generated utilities? then maybe yes, otherwise no.

LukeMathWalker: We will need something for packed types, most likely. We may also want to generate assertions checking that the C type matches the alignment we forced, just to be sure 🤔

src/redisearch_rs/headers/fnv_ffi.h
LukeMathWalker: We now get the header generated, but it isn't actually wired up (by this PR).

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/fork_gc.rs
LukeMathWalker: >  (also personally would have made every export opt-in but that's a different discussion)

It kind of is, in practice.
A type is only exported with its _full_ definition (e.g. fields) if it's used by value as a function parameter, either directly or transitively (e.g. as a field of another type that's used by value).
A type is exported as opaque (i.e. a forward declaration) if it's only used behind a pointer as a function parameter, either directly or transitively.
A type is not exported at all if it isn't referenced (hence the need for a few `export` attributes here and there in this PR).

LukeMathWalker: In a nutshell, the C header we generate is the minimal header that compiles given the set of items marked with `#[unsafe(no_mangle)]`.

JonasKruckenberg: could we not go all the way then? ie require all functions that you want to export be marked with `#[unsafe(no_mangle)]` (gives us the opportunity to check the sig and is explicit) and then require all transitive types me marked as well (either `export` or `skip`) if anything is not explicitly marked produce an error?

that would also elegantly resolve the attribute naming and footguns.

we would ofc add an explicit carveout for "well known" FFI types from `core/std` et al

src/redisearch_rs/headers/rlookup.h
JonasKruckenberg: then sure yeah this makes sense.

src/redisearch_rs/rlookup/src/row.rs
LukeMathWalker: Makes sense, changed the behaviour upstream and updated the PR!

src/redisearch_rs/headers/idf_ffi.h
LukeMathWalker: There was already an option for it, but I enhanced it to include the crate path/name automatically. Done!

src/redisearch_rs/c_entrypoint/inverted_index_ffi/src/fork_gc.rs
LukeMathWalker: > require all functions that you want to export be marked with #[unsafe(no_mangle)]

That's already the case!

> and then require all transitive types me marked as well (either export or skip) if anything is not explicitly marked produce an error

I feel this creates unnecessary busy work. `cheadergen` errors out if anything that's in the closure is not repr(C)/can't be represented. Requiring an annotation for types that _must_ be in scope is too much imo.

src/redisearch_rs/headers/inverted_index_ffi.h
chesedo: Basedo on the usage, this wants `size_t`

src/redisearch_rs/headers/iterators_ffi.h
```diff
@@ -1,45 +1,84 @@
...
+ */
+typedef struct RSQueryTerm RSQueryTerm;
+
+/**
+ * An opaque inverted index structure. The actual implementation is determined at runtime based on
+ * the index flags provided when creating the index. This allows us to have a single interface for
+ * all index types while still being able to optimize the storage and performance for each index
+ * type.
+ */
+typedef struct InvertedIndex InvertedIndex;
```
chesedo, line 44: Out of interest, how does the `opaque::InvertedIndex` export (this one) not clash with the `inverted_index::InvertedIndex` export? I'm pretty sure they end up having the same type names in these C header files.

Basically, is cheadergen designed to handle the name conflicts? Or is it a happy accident that this is working?

src/redisearch_rs/headers/inverted_index_ffi.h
LukeMathWalker: Done!

src/redisearch_rs/ffi/build.rs
chesedo: Were you able to confirm if we need some of them in the end?

LukeMathWalker: We do indeed need some of them. I'm working on another PR (stacked on top of this one) to prune `ffi` as much as possible.

src/redisearch_rs/headers/iterators_ffi.h
LukeMathWalker: We never expose (non-opaque) `InvertedIndex` to C anywhere, don't we?

src/redisearch_rs/ffi/build.rs
LukeMathWalker: See https://github.com/RediSearch/RediSearch/pull/9572

src/redisearch_rs/headers/iterators_ffi.h
chesedo: Ohhh, yeah we don't :see_no_evil: Nvm then...

chesedo (dismissed):

chesedo (dismissed):

JonasKruckenberg (dismissed):

chesedo (dismissed):

chesedo (dismissed):

gdesmott (dismissed):

gdesmott (dismissed):

gdesmott (approved):

memark (approved):
Alpine fix looks good.

PR #9184, opened 2026-04-19
MOD-14381: Support Disk Obfuscation

kei-nan:
Support disk obfuscation, 
- notifying disk when hide-user-data-from-logs is modified
- passing hidden string pointer and obfuscated index names to open APIs.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes the SearchDisk API surface (`open`, `openIndexSpec*`) and how index names are passed/obfuscated, which could break disk backends or logging behavior if implementations aren’t updated consistently.
> 
> **Overview**
> Adds disk-level support for the `hide-user-data-from-log` setting by notifying the disk backend when the config changes (`SearchDisk_UpdateLogObfuscation`) and by passing the obfuscation flag into disk initialization.
> 
> Updates SearchDisk open APIs to take a `HiddenString*` plus a separate `obfuscatedName` for logging, and wires these new signatures through index creation and RDB-load paths so disk backends can log without exposing raw index names.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 915521c94d2bcdfdac77ac1457d6ab53130c84c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/search_disk.h
+++ b/src/search_disk.h
@@ -57,19 +57,26 @@ bool SearchDisk_RegisterBigModuleCallbacks(RedisModuleCtx *ctx);
  */
 void SearchDisk_Close(RedisModuleCtx *ctx);
 
+/**
+ * @brief Enable or disable obfuscation of index names and field names in Disk log output
+ *
+ * @param enable true to enable obfuscation, false to disable
+ */
+void SearchDisk_SetLogObfuscation(bool enable);
+
 // Basic API wrappers
 
 /**
  * @brief Open an index, **Important** must be called once and only once for every index
  * @param ctx Redis module context for BigModule APIs
  * @param indexName Name of the index to open
- * @param indexNameLen Length of the index name
+ * @param obfuscatedName Obfuscated name of the index (for logging)
  * @param type Document type
  * @param deleteBeforeOpen If true, delete any existing data before opening (used when loading
  *        without SST persistence to ensure stale data is cleared)
  * @return Pointer to the index, or NULL if it does not exist
  */
-RedisSearchDiskIndexSpec* SearchDisk_OpenIndex(RedisModuleCtx *ctx, const char *indexName, size_t indexNameLen, DocumentType type, bool deleteBeforeOpen);
+RedisSearchDiskIndexSpec* SearchDisk_OpenIndex(RedisModuleCtx *ctx, const HiddenString *indexName, const char *obfuscatedName, DocumentType type, bool deleteBeforeOpen);
 
 /**
  * @brief Mark an index for deletion, the index will be deleted from the disk only after SearchDisk_CloseIndex is called
@@ -136,14 +143,14 @@ RedisSearchDiskRdbState* SearchDisk_LoadRdbToTempObject(RedisModuleIO *rdb);
  *
 * @param ctx Redis module context for BigModule APIs
  * @param indexName Name of the index
- * @param indexNameLen Length of the index name
+ * @param obfuscatedName Obfuscated name of the index (for logging)
  * @param type Document type for this index
  * @param rdbState Temporary RDB state from SearchDisk_LoadRdbToTempObject (will be consumed)
  * @return Pointer to the created IndexSpec, or NULL on error
  */
 RedisSearchDiskIndexSpec* SearchDisk_OpenIndexWithRdbState(RedisModuleCtx *ctx,
-                                                            const char *indexName,
-                                                            size_t indexNameLen,
+                                                            const HiddenString *indexName,  
+                                                            const char *obfuscatedName,
                                                             DocumentType type,
                                                             RedisSearchDiskRdbState *rdbState);
 
--- a/src/search_disk.h
+++ b/src/search_disk.h
@@ -58,11 +58,9 @@ bool SearchDisk_RegisterBigModuleCallbacks(RedisModuleCtx *ctx);
 void SearchDisk_Close(RedisModuleCtx *ctx);
 
 /**
- * @brief Enable or disable obfuscation of index names and field names in Disk log output
- *
- * @param enable true to enable obfuscation, false to disable
+ * @brief Update the log obfuscation setting for the search disk module
  */
-void SearchDisk_SetLogObfuscation(bool enable);
+void SearchDisk_UpdateLogObfuscation();
 
 // Basic API wrappers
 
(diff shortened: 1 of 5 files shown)


Itzikvaknin (approved):
LGTM

src/search_disk_api.h
```diff
@@ -51,24 +54,34 @@ typedef struct BasicDiskAPI {
...
-  RedisSearchDisk *(*open)(RedisModuleCtx *ctx, int buffer_percentage);
+  RedisSearchDisk *(*open)(RedisModuleCtx *ctx, int buffer_percentage, bool logObfuscation);
   void (*close)(RedisModuleCtx *ctx, RedisSearchDisk *disk);
+
+  /**
+   * @brief Enable or disable obfuscation of index names and field names in Disk log output
+   * @param disk Pointer to the disk
+   * @param enable true to enable obfuscation, false to disable
+   */
+  void (*setLogObfuscation)(RedisSearchDisk *disk, bool enable);
```
raz-mon, line 68: Why not expose an API for the disk side to use, basically exposing the config value, rather than adding this extra context?

kei-nan: In every log that contains user data we need to know if we need to obfuscate or not.
So,, if we exposed an API callback it would need to be called multiple times and we would need to perform unsafe calls from rust to c all over the codebase.

I don't think the config value changes so frequently that we would benefit from it.

The current direction protects the codebase from unsafe behaviour and is more maintainable long term.

Providing a callback would mean it would need to be called every time.
Updating the disk when the value changes mean the disk can always rely on the value it stores.

raz-mon: What this is doing is duplicating the context, I don't think we want to go in that direction. Will we duplicate all contexts that are used by the disk?
I think we should expose an API to be used in the disk for this, just like we call a function to get the value of this config on the OSS side.
Regarding unsafety - why would this be unsafe?

raz-mon (dismissed):

src/search_disk_api.h
kei-nan: Disk -> OSS means Rust -> C
calling a function that exists in c side from rust means you need to use the unsafe keyword.
This would need to happen on every log operation.
The c contexts are inherently unsafe.
The rust context is inherently safe.
The duplication seems to be nessesary.

raz-mon: It's unsafety that can be avoided by creating a rust wrapper isn't it?
This does not sound like a good reason for context duplication.
Perhaps @chesedo can chime in.

kei-nan: I don't think callback registration is correct in this case.
Rust wrapper still means a rust function has an unsafe block which inherently means rust code is calling unsafe code.

I don't see the harm in duplicating a boolean to exist on the disk side and updating it when needed.
I think we can always change this down the rode and isn't super critical.
Would like to push this task forward.

We can always modify how the boolean is obtained down the road.

raz-mon: I think you're taking the avoiding of unsafety a step too far. We also use it a lot in our OSS repo, when needed.
Let's get some input from @chesedo and move on from there, I think it's important we do this right since we will run into this issue a lot in the coming future.

chesedo: Two things here.

First, keep in mind that `unsafe` at the FFI boundary mostly means "I, as the compiler, can't check on the other side of this boundary. So it could **potentially** be unsafe. You as the developer should check that it is indeed safe." Once you have confirmed the code is safe, you can make a wrapper to get rid of the unsafe keyword in calling code.

Second, how is this boolean going to be used in the Rust code?

src/search_disk.h
```diff
@@ -57,19 +57,24 @@ bool SearchDisk_RegisterBigModuleCallbacks(RedisModuleCtx *ctx);
...
  * @param indexName Name of the index to open
- * @param indexNameLen Length of the index name
+ * @param obfuscatedName Obfuscated name of the index (for logging)
  * @param type Document type
  * @param deleteBeforeOpen If true, delete any existing data before opening (used when loading
  *        without SST persistence to ensure stale data is cleared)
  * @return Pointer to the index, or NULL if it does not exist
  */
-RedisSearchDiskIndexSpec* SearchDisk_OpenIndex(RedisModuleCtx *ctx, const char *indexName, size_t indexNameLen, DocumentType type, bool deleteBeforeOpen);
+RedisSearchDiskIndexSpec* SearchDisk_OpenIndex(RedisModuleCtx *ctx, const HiddenString *indexName, const char *obfuscatedName, DocumentType type, bool deleteBeforeOpen);
```
chesedo, line 77: Fyi, on the Rust side you will want to use this type for the `HiddenString`

https://github.com/RediSearch/RediSearch/blob/991252ac87e0dad37c0038e20dfdfdb380958df0/src/redisearch_rs/c_wrappers/hidden_string/src/lib.rs#L15-L20

kei-nan: Thanks, haven't written the rust side yet but will try and use that type.

src/search_disk_api.h
kei-nan: Basically, it will be used in determining which name goes to the log function.
If false we use the user provided index name, otherwise we use the obfuscated name.

Maybe in a Display/Debug trait implementation for the index/index name.

I understand that unsafe could be used in Rust, I just prefer to keep things safe.
For example, we might need to make it atomic in the rust side.
In C side we don't really address those things.

I prefer the current API flow in this PR over adding a callback.

chesedo: If you use the Rust `HiddenStringRef` then you should not need the duplicated name. It will always obfuscate the name and you can call the `into_secret_value()` method to get the clear version of the name

src/search_disk.h
chesedo: I would suggest you have the Rust code ready before merging this. That way we know all the assumptions we make here are correct (or incorrect)

src/search_disk_api.h
kei-nan: it is a bit counterintuitive, but the hidden string doesn't hold the obfuscated string, only the user string.
Anyhow all of this is on the Rust side.
The C API looks solid IMHO.

chesedo (approved):

raz-mon:
After our discussion - I'll leave it to your decision @kei-nan whether or not to duplicate the context.
I really think this is not a good practice, and see no real reason to do it.
LMK, I'll review accordingly.

JoanFM:
Should we generalize this with a "Rusty" HiddenString that goes beyond Index name? And all our methods take it and that have some automatic defer to "obfuscated" for when the logs take it, but when we use it we can pass the hidden part to any required API?

PR #9185, opened 2026-04-19
MOD-14814: Add key lookup for Map and Array collections (`RSValue`s)

Itzikvaknin:
## Add key lookup for `Map` and `Array` collections

Add `Map::get` and `Array::map_get` methods to the `value` crate for retrieving `SharedRsValue` entries by string key.

- `Map::get(&self, key: &[u8]) -> Option<SharedRsValue>`** - linear scan over key-value pairs, matching on string key bytes.
- `Array::map_get(&self, key: &[u8]) -> Option<SharedRsValue>` - same lookup over the flat `[k1, v1, k2, v2, ...]` layout used by RESP2, which lacks a native map type. In RESP3 the data arrives as a proper `Map`.

Both methods skip non-string keys and return the first match.

### Tests

Integration tests in `value/tests/integration/collection.rs` cover shared scenarios for both types (found key, missing key, empty collection, non-string keys skipped, duplicate key) plus an array-specific case for odd-length trailing elements.
#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Additive API and tests in the `value` crate with minimal behavioral impact outside new call sites; main risk is log noise/panic in debug builds when `map_get` is used on malformed odd-length arrays.
> 
> **Overview**
> Adds key-based lookup helpers to the `value` crate: `Map::get` for string-keyed pair collections and `Array::map_get` for RESP2-style flat `[k,v,...]` arrays, both returning the first match and skipping non-string keys.
> 
> Introduces `debug_assert_warn!` (debug assert + `tracing::warn!`) and wires in `tracing` as a dependency; adds integration tests covering lookup behavior and a debug-only panic for odd-length flat arrays.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit afa3559ae9c9485abb97eae36ae6f2d532b0ea33. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/value/src/collection.rs
+++ b/src/redisearch_rs/value/src/collection.rs
@@ -59,9 +59,25 @@ impl Array {
     /// alternating keys and values. In RESP3 this same data arrives as a
     /// proper [`Map`], where [`Map::get`] can be used instead.
     ///
-    /// A trailing element in an odd-length array is silently ignored.
+    /// # Odd-length arrays
+    ///
+    /// An odd number of elements indicates malformed data. A warning is emitted
+    /// via `tracing` and a `debug_assert!` fires in debug builds. The trailing
+    /// element is ignored and the search proceeds over the well-formed prefix.
+    ///
     /// Non-string keys are skipped.
     pub fn map_get(&self, key: &[u8]) -> Option<&RsValue> {
+        debug_assert!(
+            self.len() % 2 == 0,
+            "map_get called on an odd-length array (len={}); trailing element will be ignored",
+            self.len()
+        );
+        if self.len() % 2 != 0 {
+            tracing::warn!(
+                len = self.len(),
+                "map_get called on an odd-length array; trailing element will be ignored"
+            );
+        }
         self.chunks(2).find_map(|pair| {
             let value = pair.get(1)?;
             (pair[0].as_str_bytes()? == key).then_some(&**value)
--- a/src/redisearch_rs/value/src/collection.rs
+++ b/src/redisearch_rs/value/src/collection.rs
@@ -49,6 +49,26 @@ impl Map {
     }
 }
 
+impl Array {
+    /// Looks up a value by string key in a flat key-value array layout
+    /// (`[k1, v1, k2, v2, ...]`), returning a clone of the first matching
+    /// value or `None` if no match is found.
+    ///
+    /// This is needed because RESP2 does not have a native map type.
+    /// Map-like data (e.g. `extra_attributes`) is sent as a flat array of
+    /// alternating keys and values. In RESP3 this same data arrives as a
+    /// proper [`Map`], where [`Map::get`] can be used instead.
+    ///
+    /// A trailing element in an odd-length array is silently ignored.
+    /// Non-string keys are skipped.
+    pub fn map_get(&self, key: &[u8]) -> Option<SharedRsValue> {
+        self.chunks(2).find_map(|pair| {
+            let value = pair.get(1)?;
+            (pair[0].as_str_bytes()? == key).then_some(value.clone())
+        })
+    }
+}
+
 impl<T> Deref for Collection<T> {
     type Target = [T];
 
--- a/src/redisearch_rs/value/src/collection.rs
+++ b/src/redisearch_rs/value/src/collection.rs
@@ -42,7 +42,7 @@ impl Map {
     /// Looks up a value by its string key bytes, returning a reference to the first
     /// matching [`RsValue`] or `None` if no match is found.
     ///
-    /// Non-string keys (e.g. [`RsValue::Number`](crate::RsValue::Number)) are skipped.
+    /// Non-string keys (e.g. [`RsValue::Number`]) are skipped.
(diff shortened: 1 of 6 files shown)


src/redisearch_rs/value/src/collection.rs
```diff
@@ -38,6 +38,37 @@ impl<T> Collection<T> {
...
 
+impl Map {
+    /// Looks up a value by its string key bytes, returning a clone of
+    /// the first matching [`SharedRsValue`] or `None` if no match is found.
+    ///
+    /// Non-string keys (e.g. [`RsValue::Number`](crate::RsValue::Number)) are skipped.
+    pub fn get(&self, key: &[u8]) -> Option<SharedRsValue> {
+        self.iter()
+            .find_map(|(k, v)| (k.as_str_bytes()? == key).then(|| v.clone()))
+    }
```
daxhuiberts: Considering this will be used in Rust only, you likely want an `Option<&RsValue> as the return value`. No need to clone the `SharedRsValue` handle.

src/redisearch_rs/value/src/collection.rs
```diff
@@ -38,6 +38,37 @@ impl<T> Collection<T> {
...
+    /// proper [`Map`], where [`Map::get`] can be used instead.
+    ///
+    /// A trailing element in an odd-length array is silently ignored.
+    /// Non-string keys are skipped.
+    pub fn map_get(&self, key: &[u8]) -> Option<SharedRsValue> {
+        self.chunks(2).find_map(|pair| {
+            let value = pair.get(1)?;
+            (pair[0].as_str_bytes()? == key).then(|| value.clone())
+        })
+    }
```
daxhuiberts: Same here.

src/redisearch_rs/value/src/collection.rs
Itzikvaknin: Changed in https://github.com/RediSearch/RediSearch/pull/9185/commits/f61674af834e194009469ea4d8d1b4f42f7d2539

src/redisearch_rs/value/src/collection.rs
Itzikvaknin: Changed in https://github.com/RediSearch/RediSearch/pull/9185/commits/f61674af834e194009469ea4d8d1b4f42f7d2539

src/redisearch_rs/value/src/collection.rs
```diff
@@ -38,6 +38,37 @@ impl<T> Collection<T> {
...
+    /// This is needed because RESP2 does not have a native map type.
+    /// Map-like data (e.g. `extra_attributes`) is sent as a flat array of
+    /// alternating keys and values. In RESP3 this same data arrives as a
+    /// proper [`Map`], where [`Map::get`] can be used instead.
+    ///
+    /// A trailing element in an odd-length array is silently ignored.
+    /// Non-string keys are skipped.
+    pub fn map_get(&self, key: &[u8]) -> Option<&RsValue> {
+        self.chunks(2).find_map(|pair| {
+            let value = pair.get(1)?;
```
daxhuiberts: This line does something peculiar: When a chunk only has one element, i.e. the last one when the array has an odd amount of elements, this 'ends' the search.
But working on an array with an odd amount of elements should probably not happen when you expect it to be a flat key-value array. So that should be some kind of error, right?
Maybe graciously continue on release? (With a warning in the logs?) And fail hard in debug mode?

Itzikvaknin: See https://github.com/RediSearch/RediSearch/pull/9185/commits/ac9640de1bfdadc1f3241a1170b0b52542e4e9cd, https://github.com/RediSearch/RediSearch/pull/9185/commits/967a68323fc2689dc715123ada77813e72a6b8ea

daxhuiberts (dismissed):

src/redisearch_rs/value/tests/integration/collection.rs
```diff
@@ -0,0 +1,107 @@
...
+fn assert_get(map: &Map, arr: &Array, key: &[u8], expected: Option<f64>) {
+    assert_eq!(map.get(key).map(|v| v.as_num().unwrap()), expected);
+    assert_eq!(arr.map_get(key).map(|v| v.as_num().unwrap()), expected);
+}
+
+// -- Shared scenarios: Map::get and Array::map_get behave identically --
+
+#[test]
+fn found_key() {
+    let map = make_map(&[("price", 9.99), ("quantity", 3.0)]);
```
nafraf: Just to be clear that the data is the same, can we use something like:
```rust
    let pairs = &[("price", 9.99), ("quantity", 3.0)];
    let map = make_map(pairs);
    let arr = make_flat_map_array(pairs);
```

Itzikvaknin: https://github.com/RediSearch/RediSearch/pull/9185/commits/afa3559ae9c9485abb97eae36ae6f2d532b0ea33

nafraf (approved):

PR #9187, opened 2026-04-19
MOD-14921: Search Commands: Fallback To Main Thread When Block Client Is Unavailable In Single Shard Environment

kei-nan:
There was a change which modified the default workers amount
It caused a regression that if you were running a search command inside transaction(or lua script) in a single shard environment would then the command would fail(previously it worked).
Worth noting that in multi shard environments then it won't work (coordinator will block it)

In order to not break the existing behaviour, we decided to silently fallback to running on the main thread when WORKERS > 0 and the search command is from a transaction or lua script.

This won't have an effect in a cluster environment because the coordinator already denies search commands that are running in a transaction or lua script.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches query execution path selection (background vs main thread) and timeout-check behavior, which can affect latency and correctness under load, but is gated by context flags and a config defaulting to the safer fallback.
> 
> **Overview**
> Ensures `FT.SEARCH`/`FT.AGGREGATE`/hybrid execution **no longer fails in contexts that deny client blocking** (e.g., Lua/transactions) by making background execution conditional on `RedisModule` context flags and falling back to main-thread execution when blocking is unavailable.
> 
> This replaces the `RunInThread()` macro with a context-aware `RunInThread(ctx)` helper, threads the `RedisModuleCtx*` into `AREQ_Compile()` and timeout-check decisions, and adds a new config `search-_fallback-to-main-thread-when-block-client-unavailable` (default on) with a one-time warning log when fallback is triggered.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c83799dfa1fc85f3d5ebf55b099eba9708e782e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/aggregate/aggregate_request.c
+++ b/src/aggregate/aggregate_request.c
@@ -1061,7 +1061,12 @@ bool RunInThread(RedisModuleCtx *ctx) {
 
   const bool blockClientUnavailable = RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_DENY_BLOCKING;
   if (blockClientUnavailable && RSGlobalConfig.fallbackToMainThreadWhenBlockClientUnavailable) {
-    RedisModule_Log(ctx, "warning", "Cannot run command in a worker thread because client cannot be blocked, falling back to main thread.");
+    // We only log once to reduce log spam
+    static bool logged = false;
+    if (!logged) {
+      RedisModule_Log(ctx, "warning", "Detected that client cannot be blocked, all similar requests including this one will fall back to run on the main thread.");
+      logged = true;
+    }
     return false;
   }
 
--- a/src/aggregate/aggregate_request.c
+++ b/src/aggregate/aggregate_request.c
@@ -1061,7 +1061,7 @@ bool RunInThread(RedisModuleCtx *ctx) {
 
   const bool canBlock = RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_DENY_BLOCKING;
   if (!canBlock && RSGlobalConfig.fallbackToMainThreadWhenBlockClientUnavailable) {
-    RedisModule_Log(ctx, "warning", "Cannot run in thread because client cannot be blocked");
+    RedisModule_Log(ctx, "warning", "Cannot run command in a worker thread because client cannot be blocked, falling back to main thread.");
     return false;
   }
 
--- a/src/aggregate/aggregate_request.c
+++ b/src/aggregate/aggregate_request.c
@@ -1053,6 +1053,21 @@ static int handleLoad(AGGPlan *plan, uint32_t *reqflags, ArgsCursor *ac, bool is
   return REDISMODULE_OK;
 }
 
+bool RunInThread(RedisModuleCtx *ctx) {
+  const bool hasWorkerThreads = RSGlobalConfig.numWorkerThreads > 0;
+  if (!hasWorkerThreads) {
+    return false;
+  }
+
+  const bool canBlock = RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_DENY_BLOCKING;
+  if (!canBlock) {
+    RedisModule_Log(ctx, "warning", "Cannot run in thread because client cannot be blocked");
+    return false;
+  }
+
+  return true;
+}
+
 AREQ *AREQ_New(void) {
   AREQ* req = rm_calloc(1, sizeof(AREQ));
   /*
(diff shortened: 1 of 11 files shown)


src/aggregate/aggregate_request.c
```diff
@@ -1053,6 +1053,21 @@ static int handleLoad(AGGPlan *plan, uint32_t *reqflags, ArgsCursor *ac, bool is
...
 }
 
+bool RunInThread(RedisModuleCtx *ctx) {
+  const bool hasWorkerThreads = RSGlobalConfig.numWorkerThreads > 0;
+  if (!hasWorkerThreads) {
+    return false;
+  }
+
+  const bool canBlock = RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_DENY_BLOCKING;
+  if (!canBlock && RSGlobalConfig.fallbackToMainThreadWhenBlockClientUnavailable) {
```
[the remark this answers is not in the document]
oshadmi: @kei-nan Should rename `canBlock` variable to `cannotBlock`?
Can use existing helper `cannotBlockCtx`?

src/hybrid/hybrid_request.c
```diff
@@ -303,7 +303,7 @@ static void HybridRequest_Free(HybridRequest *req) {
         RedisModuleCtx *thctx = areq->sctx->redisCtx;
         RedisSearchCtx *sctx = areq->sctx;
 
-        if (RunInThread()) {
+        if (RunInThread(sctx->redisCtx)) {
```
oshadmi: Consider:
`if (hreq->reqflags & QEXEC_F_RUN_IN_BACKGROUND)` instead of `if (RunInThread(sctx->redisCtx))`. This avoids calling `RedisModule_GetContextFlags` on a potentially stale context and is semantically more correct.
Or use the original macro `RunInThread` and not the new function?

src/aggregate/aggregate_request.c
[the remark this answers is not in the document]
GuyAv46: @kei-nan this indeed looks backwards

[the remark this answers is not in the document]
kei-nan: will rename to `blockClientUnavailable`

src/hybrid/hybrid_request.c
kei-nan: TBH, not sure I understand this code.
Will use the areq flags since that seem to make a bit more sense since we are in a loop.

GuyAv46 (approved):

src/aggregate/aggregate_request.c
```diff
@@ -1053,6 +1053,26 @@ static int handleLoad(AGGPlan *plan, uint32_t *reqflags, ArgsCursor *ac, bool is
   return REDISMODULE_OK;
 }
 
+bool RunInThread(RedisModuleCtx *ctx) {
+  const bool hasWorkerThreads = RSGlobalConfig.numWorkerThreads > 0;
+  if (!hasWorkerThreads) {
+    return false;
+  }
+
+  const bool blockClientUnavailable = RedisModule_GetContextFlags(ctx) & REDISMODULE_CTX_FLAGS_DENY_BLOCKING;
```
oshadmi, line 1062: Nitpick: Can use existing `cannotBlockCtx`
https://github.com/RediSearch/RediSearch/blob/edff4fec6e4a04134ad2559556560798aa412a50/src/module.c#L3482
And can call it only if fallback config is not disabled? such as
```
if ( RSGlobalConfig.fallbackToMainThreadWhenBlockClientUnavailabl && cannotBlockCtx(ctx))
...
```

src/hybrid/hybrid_request.c
```diff
@@ -303,7 +303,7 @@ static void HybridRequest_Free(HybridRequest *req) {
         RedisModuleCtx *thctx = areq->sctx->redisCtx;
         RedisSearchCtx *sctx = areq->sctx;
 
-        if (RunInThread()) {
+        if (areq->reqflags & QEXEC_F_RUN_IN_BACKGROUND) {
```
oshadmi, line 306: @kei-nan 
Sorry to nag but we can consider another Approach:
Keep `RunInThread()` as a macro and add an explicit `cannotBlock(ctx)` check at each call site, rather than converting `RunInThread` to a function that takes `ctx`.
Why:
- Intent is visible at each call site -- you see two separate concerns: "are workers configured?" and "does this context allow blocking?"
- Safer for call sites that don't have a valid `ctx` such as here - the macro still works without `ctx`
- Avoids changing `AREQ_Compile` signature and all its callers (11 files vs 3)
Less risk for a targeted bugfix

src/aggregate/aggregate_request.c
kei-nan: that function exists in module.c.
I don't see much benefit in exposing such a simple function and making it public.

src/hybrid/hybrid_request.c
kei-nan: RunInThread answers the question should we try and run the logic in a different thread or not.
That seem like the perfect place to add the check.
Also ensures all future call sites will have the same consideration.

oshadmi: OK, assuming the responsibility of the caller to make sure `ctx` is valid when being used.

src/aggregate/aggregate_request.c
oshadmi: @kei-nan Also suggested to switch the order of the operands in the AND expression - but not critical

oshadmi (approved):

PR #9191, opened 2026-04-19
[MOD-14944] test: fix flakiness test_query_while_flush

JoanFM:

Fixes flakiness in `test_query_while_flush` by replacing fixed-duration `time.sleep(0.5)` calls with `wait_for_condition`-based waits. The previous sleeps were insufficient on slow CI runners (notably macOS x86_64), where query threads occasionally produced zero successful queries within the 0.5 s window, causing `assertGreater(before_flush_successes, 0)` to fail.

### Changes

Three `time.sleep(0.5)` calls were replaced:

1. **Before `flushall_called.set()`** — previously slept 0.5 s hoping workers accumulated successes. Now uses `wait_for_condition` polling `before_flush_successes > 0` with a 30 s timeout. Fast-path on fast runners (exits in ms), tolerant on slow runners.

2. **After `flushall_called.set()`** — drain window for in-flight pre-flush attributions. Replaced with a stability check that polls `(before_flush_successes, before_flush_errors)` every 50 ms until the tuple stops changing, bounded by a 10 s timeout. Preserves the original drain semantic without depending on wall-clock duration.

3. **After `flush_completed = True`, before `flushall_called.clear()`** — same class of drain window; replaced with the same stability-check pattern.

### Why stability-check for 2 and 3

These sleeps are not "wait for something to happen" — they are drain windows protecting against a worker that read `flushall_called.is_set() == False` before the main thread set it and hasn't yet performed its counter increment. Observing that counters have stopped changing across a short polling interval is the direct, adaptive analogue of the original "sleep long enough" intent.

### Test semantics preserved

- Same pre-flush assertions: `before_flush_successes > 0`, `before_flush_errors == 0`
- Same post-flush assertions: `after_flush_errors > 0`, `after_flush_successes == 0`
- Same index lifecycle (create `index1`, FLUSHALL, create `index2`)
- No changes to the worker code or to the locking model

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Test-only change that adjusts synchronization/wait logic; main risk is making the test hang or become overly strict if the new drain conditions are wrong.
> 
> **Overview**
> Makes `test_query_while_flush` deterministic by removing fixed `time.sleep(0.5)` synchronization windows and replacing them with `wait_for_condition` checks.
> 
> The test now (1) waits until at least one pre-flush query success is observed, and (2) introduces per-thread `iteration_counts` to *drain in-flight worker iterations* after toggling `flushall_called` and `flush_completed`, reducing misattribution and CI flakiness on slower runners.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 46b10441224588ecfe4938734779898700a0e7bc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/pytests/test_query_while_flush.py
+++ b/tests/pytests/test_query_while_flush.py
@@ -41,7 +41,15 @@ def test_query_while_flush():
         'flush_completed': False
     }
 
-    def query_worker(stats):
+    # Per-thread iteration counters, incremented at the end of each loop iteration.
+    # Used to deterministically drain in-flight iterations after toggling state
+    # (flushall_called / flush_completed): once every counter has advanced by at
+    # least one, every worker has re-evaluated the state and no stale attribution
+    # to the pre-flush counters can still be pending.
+    num_threads = 5
+    iteration_counts = [0] * num_threads
+
+    def query_worker(stats, thread_id):
         """Worker thread that continuously queries the index"""
         local_conn = env.getClusterConnectionIfNeeded()
 
@@ -65,16 +73,16 @@ def test_query_while_flush():
                 else:
                     stats['after_flush_errors'] += 1
 
+          iteration_counts[thread_id] += 1
           # Small delay to avoid overwhelming the system
           time.sleep(0.001)
 
-    # Start 5 query threads (pass the event)
-    num_threads = 5
+    # Start query threads (pass the event)
     threads = []
     for i in range(num_threads):
         thread = threading.Thread(
             target=query_worker,
-            args=(stats, ),
+            args=(stats, i),
         )
         threads.append(thread)
         thread.start()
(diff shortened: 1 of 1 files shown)


tests/pytests/test_query_while_flush.py
```diff
@@ -79,15 +79,28 @@ def query_worker(stats):
...
-    time.sleep(0.5)
+    # Wait for in-flight pre-flush attributions to drain: poll until the pre-flush counters
+    # stop changing across a short interval. Equivalent to the previous fixed sleep, but
+    # adaptive (returns quickly when the system is idle, waits longer on slow machines).
+    def _pre_flush_counters_stable():
+        snap = (stats['before_flush_successes'], stats['before_flush_errors'])
+        time.sleep(0.05)
+        cur = (stats['before_flush_successes'], stats['before_flush_errors'])
+        return snap == cur, {'snap': snap, 'cur': cur}
+    wait_for_condition(
```
GuyAv46: So the condition is stats stability for 0.05 seconds? Can't we aim for some specific value, or a specific change?

tests/pytests/test_query_while_flush.py
```diff
@@ -96,11 +109,20 @@ def query_worker(stats):
...
+    # Wait for any thread that already passed the `flushall_called.is_set()` check but has
+    # not yet read `flush_completed` to finish its iteration, so we don't misattribute a
+    # post-flush observation to the before-flush bucket when we later clear the event.
+    # Same drain purpose as the previous sleep, expressed as a stability check.
+    def _pre_flush_counters_stable_post():
+        snap = (stats['before_flush_successes'], stats['before_flush_errors'])
+        time.sleep(0.05)
+        cur = (stats['before_flush_successes'], stats['before_flush_errors'])
+        return snap == cur, {'snap': snap, 'cur': cur}
+    wait_for_condition(
```
GuyAv46: Can't we have a target value or specific change from the current state, instead of stats stability for 0.05 seconds?

tests/pytests/test_query_while_flush.py
JoanFM: we can have a larger time sleep, but I think there is no other condition so easy to get.

BenGoldberger (approved):

tests/pytests/test_query_while_flush.py
GuyAv46: I'm not familiar with this test, but can't we get some temporary stability before we're ready to proceed?
Why not observe some value before the wait call, and wait until the stats increase?

GuyAv46 (approved):

tests/pytests/test_query_while_flush.py
GuyAv46: I see you did that now :)

Issue #8009, opened 2026-01-08
[BUG] fork-GC child processes (gc-*) become zombie (<defunct>) processes


vladabroz:
**Describe the bug**
RediSearch fork-GC child processes (gc-*) become zombie (<defunct>) processes and are not reaped by redis-server, causing a steady accumulation of zombies over time.

I'm running the latest redis-stack (docker.io/redis/redis-stack-server:7.4.0-v8) image using the attached docker compose file.

I see these logs:
```
3:M 08 Jan 2026 17:26:06.312 - Module fork started pid: 265
3:M 08 Jan 2026 17:26:06.313 - Killing running module fork child: 265
265:C 08 Jan 2026 17:26:06.313 * Fork CoW for Module fork: current 0 MB, peak 0 MB, average 0 MB
265:C 08 Jan 2026 17:26:06.313 # Child failed reporting info to parent, exiting. Broken pipe
```

and there stays zombie processes  (which after some time hit the limit of  max pids.
```
(redis) / # cat /sys/fs/cgroup/pids.max
2048
```

```
(redis) / # ps auxww
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0   2644  1052 ?        Ss   17:14   0:00 /usr/bin/dumb-init /bin/sh /entrypoint.sh
root           2  0.0  0.0   2892  1544 ?        Ss   17:14   0:00 /bin/sh /entrypoint.sh
root           3  0.2  0.2 306324 29608 ?        Sl   17:14   0:04 /opt/redis-stack/bin/redis-server *:6379
root          34  0.0  0.0   8420  6068 ?        S    17:15   0:00 zsh
root          58  0.0  0.1  22076 13016 ?        S+   17:15   0:00 redis-cli
root         138  0.0  0.0      0     0 ?        ZN   17:20   0:00 [gc-0704] <defunct>
root         172  0.0  0.0      0     0 ?        ZN   17:22   0:00 [gc-0704] <defunct>
root         240  0.0  0.0      0     0 ?        ZN   17:24   0:00 [gc-0704] <defunct>
root         265  0.0  0.0      0     0 ?        ZN   17:26   0:00 [gc-0704] <defunct>
root         305  0.0  0.0      0     0 ?        ZN   17:28   0:00 [gc-0704] <defunct>
root         334  0.0  0.0      0     0 ?        ZN   17:30   0:00 [gc-0704] <defunct>
[...]

PR #9192, opened 2026-04-19
fix(#8009): ensure forked child is always reaped;

gustavoroc:

**Current:**  
There is a race condition between the GC background thread and concurrent `RM_Fork()` calls from other modules. If `server.module_child_pid` is replaced before `RedisModule_KillForkChild()` is invoked, the function becomes a no-op and skips `waitpid()`, leaving the original child process unreaped (zombie).

**Change:**  
Add a blocking `waitpid()` after `RedisModule_KillForkChild()` to ensure the child process is always reaped, regardless of whether `KillForkChild` succeeds or becomes a no-op due to the race.

**Outcome:**  
Prevents zombie process accumulation under concurrent module activity. The fix is safe because:
- If the child was already reaped → `waitpid()` returns `ECHILD`
- If not → `waitpid()` properly reaps it  
The wait runs outside the GIL, so it does not block the Redis main thread.

---

1. N/A (no tracked issue)
2. N/A

---

1. `src/fork_gc/fork_gc.c`

---

#### Mark if applicable

---

#### Release Notes

**Release note:**  
Fix a race condition that could cause zombie process accumulation when multiple modules use `RM_Fork()` concurrently.

---

> [!NOTE]
> **Medium Risk**
> Touches fork/child lifecycle management in `ForkGC`, which can affect process cleanup and GC scheduling under load, but the change is narrowly scoped to an extra reap path with a bounded timeout.
> 
> **Overview**
> Fixes a race where `RedisModule_KillForkChild()` can become a no-op (due to other modules calling `RM_Fork()`), leaving the ForkGC child process unreaped.
> 
> Adds a fallback `reap_child_blocking()` path (using `waitpid(WNOHANG)` + `TimedOut()` polling) when `KillForkChild` fails, ensuring the original child is eventually reaped and reducing zombie accumulation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac699a64ad2ae2ee50db47117a3bf159e0581bbe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/fork_gc/fork_gc.c
+++ b/src/fork_gc/fork_gc.c
@@ -94,16 +94,17 @@ static inline bool isOutOfMemory(RedisModuleCtx *ctx) {
   return used_memory_ratio > 1;
 }
 
-// Waits up to timeout_sec for cpid to be reaped, polling every 500us. Called when
+// Waits up to timeout_sec for cpid to be reaped, polling every 500us (via nanosleep). Called when
 // KillForkChild was a no-op, meaning Redis never waited on this pid.
 static void reap_child_blocking(pid_t cpid, int timeout_sec) {
   struct timespec deadline;
   clock_gettime(CLOCK_MONOTONIC_RAW, &deadline);
   deadline.tv_sec += timeout_sec;
 
+  struct timespec poll_interval = {.tv_sec = 0, .tv_nsec = 500000};
   while (waitpid(cpid, NULL, WNOHANG) == 0) {
     if (TimedOut(&deadline)) break;
-    usleep(500);
+    nanosleep(&poll_interval, NULL);
   }
 }
 
--- a/src/fork_gc/fork_gc.c
+++ b/src/fork_gc/fork_gc.c
@@ -93,6 +93,14 @@ static inline bool isOutOfMemory(RedisModuleCtx *ctx) {
   return used_memory_ratio > 1;
 }
 
+// Blocks until cpid is reaped, retrying on EINTR. Called when
+// KillForkChild was a no-op, meaning Redis never waited on this pid.
+static void reap_child_blocking(pid_t cpid) {
+  while (waitpid(cpid, NULL, 0) == -1) {
+    if (errno != EINTR) break;
+  }
+}
+
 static bool periodicCb(void *privdata, bool force) {
   ForkGC *gc = privdata;
   RedisModuleCtx *ctx = gc->ctx;
@@ -222,13 +230,7 @@ static bool periodicCb(void *privdata, bool force) {
     int kill_rv = RedisModule_KillForkChild(cpid);
     RedisModule_ThreadSafeContextUnlock(ctx);
     if (kill_rv != REDISMODULE_OK) {
-      // KillForkChild was a no-op: another module's RM_Fork replaced
-      // module_child_pid between the pipe read and the GIL re-acquire.
-      // Redis sent SIGUSR1 to cpid but skipped waitpid, leaving it as a
-      // zombie. Reap it here.
-      while (waitpid(cpid, NULL, 0) == -1) {
-        if (errno != EINTR) break;
-      }
+      reap_child_blocking(cpid);
     }
 
     if (gcrv) {
--- a/src/fork_gc/fork_gc.c
+++ b/src/fork_gc/fork_gc.c
@@ -219,18 +219,16 @@ static bool periodicCb(void *privdata, bool force) {
     // otherwise it might cause a pipe leak and eventually run
     // out of file descriptor
     RedisModule_ThreadSafeContextLock(ctx);
-    RedisModule_KillForkChild(cpid);
+    int kill_rv = RedisModule_KillForkChild(cpid);
     RedisModule_ThreadSafeContextUnlock(ctx);
-
-    // Ensure the child is fully reaped to prevent zombie processes.
-    // KillForkChild may be a no-op if Redis's module_child_pid was already
-    // replaced by a concurrent RM_Fork call (e.g., from RedisGears or
-    // RedisTimeSeries). In that case the child is already dead (killed by
-    // Redis when the other module forked) but was never waited on.
-    // If it was already reaped by KillForkChild above, waitpid returns -1
-    // with ECHILD, which is harmless.
-    while (waitpid(cpid, NULL, 0) == -1) {
-      if (errno != EINTR) break;
+    if (kill_rv != REDISMODULE_OK) {
+      // KillForkChild was a no-op: another module's RM_Fork replaced
+      // module_child_pid between the pipe read and the GIL re-acquire.
+      // Redis sent SIGUSR1 to cpid but skipped waitpid, leaving it as a
+      // zombie. Reap it here.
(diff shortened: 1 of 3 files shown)


src/fork_gc/fork_gc.c
```diff
@@ -222,6 +222,17 @@ static bool periodicCb(void *privdata, bool force) {
     RedisModule_KillForkChild(cpid);
     RedisModule_ThreadSafeContextUnlock(ctx);
 
+    // Ensure the child is fully reaped to prevent zombie processes.
+    // KillForkChild may be a no-op if Redis's module_child_pid was already
```
JoanFM: How could this happen? shouldn't they take the GIL in order to fork?

src/fork_gc/fork_gc.c
```diff
@@ -222,6 +222,17 @@ static bool periodicCb(void *privdata, bool force) {
     RedisModule_KillForkChild(cpid);
```
JoanFM: Perhaps is better to listen to the response code of KillForChild and just do that if it returns ERR?

src/fork_gc/fork_gc.c
gustavoroc: RM_Fork does need the GIL, but the race window is between the GC dropping it after the pipe read and grabbing it back to call KillForkChild. In that gap the main thread can process an RM_Fork from another module which calls TerminateModuleForkChild(old_cpid, wait=0), sends SIGUSR1, skips waitpid, and overwrites module_child_pid. In the time the GC thread gets the GIL back, the pid has changed and KillForkChild becomes a no-op

src/fork_gc/fork_gc.c
gustavoroc: Hmm, you're right. Refactored it to branch on ERR"

src/fork_gc/fork_gc.c
```diff
@@ -93,6 +93,14 @@ static inline bool isOutOfMemory(RedisModuleCtx *ctx) {
   return used_memory_ratio > 1;
 }
 
+// Blocks until cpid is reaped, retrying on EINTR. Called when
+// KillForkChild was a no-op, meaning Redis never waited on this pid.
+static void reap_child_blocking(pid_t cpid) {
```
JoanFM: Lets add a timeout to avoid any theoretical potential infinite loop. I guess the timeout can be obtained from the frequency of the fork gc.

JoanFM (changes requested):
Small change to avoid risk of infinite loop

src/fork_gc/fork_gc.c
gustavoroc: used `RSGlobalConfig.gcConfigParams.gcSettings.forkGcRunIntervalSec` for the timeout value and `timeout.h` for the deadline check — it's already used in other parts of the codebase for the same purpose, so it felt like the right abstraction to reuse here.

src/fork_gc/fork_gc.c
```diff
@@ -93,6 +94,20 @@ static inline bool isOutOfMemory(RedisModuleCtx *ctx) {
...
+// Waits up to timeout_sec for cpid to be reaped, polling every 500us (via nanosleep). Called when
+// KillForkChild was a no-op, meaning Redis never waited on this pid.
+static void reap_child_blocking(pid_t cpid, int timeout_sec) {
+  struct timespec deadline;
+  clock_gettime(CLOCK_MONOTONIC_RAW, &deadline);
+  deadline.tv_sec += timeout_sec;
+
+  struct timespec poll_interval = {.tv_sec = 0, .tv_nsec = 500000};
+  while (waitpid(cpid, NULL, WNOHANG) == 0) {
+    if (TimedOut(&deadline)) break;
```
JoanFM: let's add a log if TimeOut is reached, so that we can have some logs and show if there is a problem.

src/fork_gc/fork_gc.c
```diff
@@ -93,6 +94,20 @@ static inline bool isOutOfMemory(RedisModuleCtx *ctx) {
...
 }
 
+// Waits up to timeout_sec for cpid to be reaped, polling every 500us (via nanosleep). Called when
+// KillForkChild was a no-op, meaning Redis never waited on this pid.
+static void reap_child_blocking(pid_t cpid, int timeout_sec) {
+  struct timespec deadline;
+  clock_gettime(CLOCK_MONOTONIC_RAW, &deadline);
+  deadline.tv_sec += timeout_sec;
+
+  struct timespec poll_interval = {.tv_sec = 0, .tv_nsec = 500000};
```
JoanFM: maybe we can have a slightly larger poll interval

JoanFM (changes requested):

src/fork_gc/fork_gc.c
gustavoroc: Done. I made 1.5ms, i think that's a good range

JoanFM (dismissed):
LGTM! Thanks a lot for the contribution!

gustavoroc:
@JoanFM Fixed a linker error in the new tests. the RMCK helper declarations (RMCK_GetLastChildPid, RMCK_SetKillForkChildShouldFail) were inside an extern "C" block in the test file, but the definitions in redismock.cpp use C++ linkage. Moving the declarations outside the block makes both sides use C++ name mangling consistently. Also added the missing <csignal> and <sys/wait.h> includes.

tests/cpptests/test_cpp_forkgc.cpp
```diff
@@ -878,3 +884,88 @@ TEST_F(FGCTestNumeric, testNumericBlocksSinceFork) {
...
+
+class ReapChildTest : public FGCTestTag {
+ private:
+  pid_t child_pid = -1;
+
+  void SetUp() override {
+    FGCTestTag::SetUp();
+    RSGlobalConfig.gcConfigParams.gcSettings.forkGcCleanThreshold = 0;
+    // Keep child alive long enough for SIGSTOP to land before it calls _exit.
+    RSGlobalConfig.gcConfigParams.gcSettings.forkGcSleepBeforeExit = 1;
```
JoanFM: I am concerned with this timer, It seems it can bring some flakiness to tests. If we cannot do something that does not trust timing, let's just forget the testing.

JoanFM (changes requested):

JoanFM:
Hey @gustavoroc ,

Thanks for the effort to do the full coverage, however I think we should not add these tests.

I believe they will be hard to mantain at quite some cost, plus there may be flakiness. Please find a way to simplify or let's ignore them.

gustavoroc:
> Hey @gustavoroc ,
> 
> Thanks for the effort to do the full coverage, however I think we should not add these tests.
> 
> I believe they will be hard to mantain at quite some cost, plus there may be flakiness. Please find a way to simplify or let's ignore them.

Sure. I had a hard time reproducing the issue with unit tests. I think it touched a lot of structures that could cause flakiness, like you said. So I think the best way to move forward is to remove those test changes. My only concern was Sonar coverage. I already made the commits reverting it.

JoanFM (approved):

PR #9215, opened 2026-04-21
[MOD-14309][MOD-14311] Add FT.DEBUG for FT.HYBRID and propagate timeout warnings

ikalchev:
This change adds the `FT.DEBUG` command to hybrid commands. It also fixes timeout propagation for hybrid queries and enables/adds tests for the these two features.

The timeout propagation fixes are added together with `FT.DEBUG` so we can realistically test the new command.

Others:
- fixes parsing of warnings for both RESP2 and RESP3 in hybrid commands
- enforces the timeout policy while parsing cursor mappings in hybrid commands 
- `replyWithCursors` returns the timeout warnings if any
- In `StartWithCursors` does not close the cursors if the timeout policy is not FAIL
- changes how error codes are inferred based on the message - now matches with "Starts with" instead of an exact match. 
- fixes the logic in depleteAll to check for timeouts in any of the responses and propagate it properly upstream

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches distributed `FT.HYBRID` execution, cursor mapping parsing, and timeout/oom propagation logic; mistakes could change query failure vs partial-result behavior and stats in clustered deployments.
> 
> **Overview**
> Adds `_FT.DEBUG FT.HYBRID` support for both single-shard and cluster mode by parsing hybrid-specific debug params (`TIMEOUT_AFTER_N_SEARCH/VSIM/TAIL`), forwarding them to shards (wrapping MR commands with `_FT.DEBUG`), and applying debug timeouts after pipeline build.
> 
> Fixes hybrid timeout/oom/warning propagation: cursor-mapping parsing now treats `warnings` as warnings (not errors) for RESP2/RESP3, enforces timeout policy when aggregating shard responses, and `HybridRequest_StartCursors` keeps cursors on `RETURN` timeouts while emitting timeout warnings in the cursor reply. Also aligns empty OOM replies to emit warning strings (not error strings), improves error-code inference to match by message prefix, and makes depletion aggregate timeouts across depleters instead of stopping on the first non-EOF.
> 
> Expands tests to cover hybrid debug arg validation, cluster timeout metrics for hybrid, and ensures `FT.DEBUG FT.PROFILE ... HYBRID` continues to use the non-hybrid debug parsing path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3b9c769086aa2586684eff70f7029c96efbc9ca9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/coord/hybrid/hybrid_cursor_mappings.c
+++ b/src/coord/hybrid/hybrid_cursor_mappings.c
@@ -291,6 +291,12 @@ bool ProcessHybridCursorMappings(const MRCommand *cmd, StrongRef searchMappingsR
                 // qctx->err. The timeout will propagate through cursor reads (RPNet
                 // detects it from the depleter's last_rc), and replyWarningsWithSuffixes
                 // emits the properly-suffixed warning (e.g., "(SEARCH)" / "(VSIM)").
+            } else if (QueryError_GetCode(&ctx->errors[i]) == QUERY_ERROR_CODE_TIMED_OUT) {
+                // FAIL policy: forward the standard timeout error directly,
+                // matching the standalone path which uses QueryError_Strerror().
+                QueryError_SetCode(status, QUERY_ERROR_CODE_TIMED_OUT);
+                success = false;
+                break;
             } else {
                 QueryError_SetWithoutUserDataFmt(status, QueryError_GetCode(&ctx->errors[i]), "Failed to process shard responses, first error: %s, total error count: %zu",
                     QueryError_GetUserError(&ctx->errors[i]), array_len(ctx->errors));
--- a/src/coord/hybrid/hybrid_cursor_mappings.c
+++ b/src/coord/hybrid/hybrid_cursor_mappings.c
@@ -45,7 +45,7 @@ static void processHybridError(processCursorMappingCallbackContext *ctx, MRReply
 
 // Warning strings use a different format than error strings (no prefix).
 // Map warning codes to error codes for uniform handling in ProcessHybridCursorMappings.
-static void processHybridWarning(processCursorMappingCallbackContext *ctx, MRReply *rep) {
+static void processHybridWarning(processCursorMappingCallbackContext *ctx, const MRReply *rep) {
     const char *warningMessage = MRReply_String(rep, NULL);
     QueryWarningCode warningCode = QueryWarningCode_GetCodeFromMessage(warningMessage);
     QueryError error = QueryError_Default();
--- a/src/coord/hybrid/hybrid_cursor_mappings.c
+++ b/src/coord/hybrid/hybrid_cursor_mappings.c
@@ -287,10 +287,14 @@ bool ProcessHybridCursorMappings(const MRCommand *cmd, StrongRef searchMappingsR
             if (QueryError_GetCode(&ctx->errors[i]) == QUERY_ERROR_CODE_OUT_OF_MEMORY && oomPolicy == OomPolicy_Return ) {
                 QueryError_SetQueryOOMWarning(status);
             } else if (QueryError_GetCode(&ctx->errors[i]) == QUERY_ERROR_CODE_TIMED_OUT && timeoutPolicy != TimeoutPolicy_Fail) {
-                // RETURN policy: acknowledge the shard timeout but don't set it on
-                // qctx->err. The timeout will propagate through cursor reads (RPNet
-                // detects it from the depleter's last_rc), and replyWarningsWithSuffixes
-                // emits the properly-suffixed warning (e.g., "(SEARCH)" / "(VSIM)").
+                // RETURN / RETURN-STRICT policy: acknowledge the shard timeout but
+                // don't set it on qctx->err. The timeout will propagate through cursor
+                // reads (RPNet detects it from the depleter's last_rc), and
+                // replyWarningsWithSuffixes emits the properly-suffixed warning
+                // (e.g., "(SEARCH)" / "(VSIM)").
+                // Note: for the _FT.DEBUG FT.HYBRID path, RETURN-STRICT is rejected
+                // earlier in parseHybridDebugParams, so only RETURN reaches here in
+                // debug mode.
             } else if (QueryError_GetCode(&ctx->errors[i]) == QUERY_ERROR_CODE_TIMED_OUT) {
                 // FAIL policy: forward the standard timeout error directly,
                 // matching the standalone path which uses QueryError_Strerror().
--- a/src/coord/hybrid/hybrid_cursor_mappings.c
+++ b/src/coord/hybrid/hybrid_cursor_mappings.c
@@ -43,6 +43,24 @@ static void processHybridError(processCursorMappingCallbackContext *ctx, MRReply
     ctx->errors = array_ensure_append_1(ctx->errors, error);
 }
 
+// Warning strings use a different format than error strings (no prefix).
+// Map warning codes to error codes for uniform handling in ProcessHybridCursorMappings.
+static void processHybridWarning(processCursorMappingCallbackContext *ctx, MRReply *rep) {
+    const char *warningMessage = MRReply_String(rep, NULL);
+    QueryWarningCode warningCode = QueryWarningCode_GetCodeFromMessage(warningMessage);
+    QueryError error = QueryError_Default();
+    if (warningCode == QUERY_WARNING_CODE_TIMED_OUT) {
+        QueryError_SetCode(&error, QUERY_ERROR_CODE_TIMED_OUT);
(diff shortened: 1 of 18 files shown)


src/hybrid/hybrid_exec.c
```diff
@@ -1064,6 +1083,7 @@ int hybridCommandHandler(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
   HybridRequest *hybridRequest = MakeDefaultHybridRequest(sctx);
   hybridRequest->profile = printHybridProfile;
   hybridRequest->tailPipeline->qctx.isProfile = profileOptions & EXEC_WITH_PROFILE;
+  hybridRequest->debugParams = debugParams;
```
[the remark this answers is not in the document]
ikalchev: Yeah I should have caught that 😬

tests/pytests/test_hybrid_timeout.py
```diff
@@ -51,9 +51,100 @@ def test_hybrid_debug_with_no_index_error():
...
+def test_hybrid_debug_no_component_timeout_sa():
+    """Test error when no component timeout parameter is specified (SA).
+
+    In SA mode, HybridRequest_Debug_New short-circuits on debug_params_count==0
+    before parseHybridDebugParams can validate. The reply is an error but
+    without the specific "At least one component timeout parameter" message.
+    """
+    env = Env(enableDebugCommand=True)
+    setup_basic_index(env)
+    env.expect(*_base_hybrid_debug_cmd(), 'DEBUG_PARAMS_COUNT', '0').error()
```
nafraf, line 119: What's the error message we get?   
`Success (not an error)` ? It's confusing.

tests/pytests/test_hybrid_timeout.py
```diff
@@ -66,23 +157,21 @@ def test_debug_timeout_fail_vsim():
...
-#TODO: remove skip once FT.HYBRID for cluster is implemented
-@skip(cluster=True)
 def test_debug_timeout_fail_both():
     """Test FAIL policy with both components timeout using debug parameters"""
     env = Env(enableDebugCommand=True, moduleArgs='ON_TIMEOUT FAIL')
     setup_basic_index(env)
     env.expect('_FT.DEBUG', 'FT.HYBRID', 'idx', 'SEARCH', 'running', 'VSIM', '@embedding', '$BLOB', 'PARAMS', '2', 'BLOB', query_vector, 'TIMEOUT_AFTER_N_SEARCH', '1','TIMEOUT_AFTER_N_VSIM', '2', 'DEBUG_PARAMS_COUNT', '4').error().contains('SEARCH_TIMEOUT Timeout limit was reached')
 
-#TODO: remove skip once FT.HYBRID for cluster is implemented
+# Tail pipeline runs on the coordinator; debug timeout params aren't applied there in cluster.
```
nafraf, line 175: Do we have another ticket to support this?

src/redisearch_rs/c_entrypoint/query_error_ffi/src/lib.rs
```diff
@@ -117,18 +121,19 @@ pub const extern "C" fn QueryError_CodeMaxValue() -> u8 {
...
-    const OUT_OF_MEMORY_ERROR_CSTR: &CStr = QueryErrorCode::OutOfMemory.to_c_str();
-    const UNAVAILABLE_SLOTS_ERROR_CSTR: &CStr = QueryErrorCode::UnavailableSlots.to_c_str();
+    const TIMED_OUT_PREFIX: &[u8] = QueryErrorCode::TimedOut.prefix_c_str().to_bytes();
+    const OUT_OF_MEMORY_PREFIX: &[u8] = QueryErrorCode::OutOfMemory.prefix_c_str().to_bytes();
+    const UNAVAILABLE_SLOTS_PREFIX: &[u8] =
+        QueryErrorCode::UnavailableSlots.prefix_c_str().to_bytes();
 
     // Safety: see safety requirement above.
-    let message = unsafe { CStr::from_ptr(message) };
+    let message = unsafe { CStr::from_ptr(message) }.to_bytes();
```
nafraf, line 135: We'll get undefined behavior if input message pointer is `NULL`.

src/coord/hybrid/hybrid_cursor_mappings.c
```diff
@@ -269,7 +287,16 @@ bool ProcessHybridCursorMappings(const MRCommand *cmd, StrongRef searchMappingsR
             if (QueryError_GetCode(&ctx->errors[i]) == QUERY_ERROR_CODE_OUT_OF_MEMORY && oomPolicy == OomPolicy_Return ) {
                 QueryError_SetQueryOOMWarning(status);
             } else if (QueryError_GetCode(&ctx->errors[i]) == QUERY_ERROR_CODE_TIMED_OUT && timeoutPolicy != TimeoutPolicy_Fail) {
+                // RETURN policy: acknowledge the shard timeout but don't set it on
```
nafraf: Do we have tests for `timeoutPolicy == TimeoutPolicy_ReturnStrict` ?

tests/pytests/test_hybrid_timeout.py
ikalchev: Yes, that's the message. This is the same behaviour as in the non-hybrid path.

aggregate_debug.c:
```
  AREQ_Debug *AREQ_Debug_New(RedisModuleString **argv, int argc, QueryError *status) {
    AREQ_Debug_params debug_params = parseDebugParamsCount(argv, argc, status);
    if (debug_params.debug_params_count == 0) {
      return NULL;
    }

aggregate_exec.c:

    AREQ_Debug *debug_req = AREQ_Debug_New(argv, argc, &status);
    if (!debug_req) {
      goto error;
    }
    // ...
  error:
    if (r) {
      AREQ_DecrRef(r);
    }
    return QueryError_ReplyAndClear(ctx, &status); // <--- no error set so fallback to default error str
```

Do we want to fix that here or is this intentional?

src/redisearch_rs/c_entrypoint/query_error_ffi/src/lib.rs
ikalchev: True, but I see that was also the case before this change - the `Safety` disclaimer at the top is wrong since `CStr::from_ptr` explicitly calls-out the ptr must not be null.

From what I see where this is used, it is always used with the result of `MRReply_String`  which is:

```
inline const char *MRReply_String(const MRReply *reply, size_t *len) {
  if (len) {
    *len = reply->len;
  }
  return reply->str;
}
```

so we also don't check if `reply` itself is sane (as far as I traced this). I will just modify `QueryError_GetCodeFromMessage` to return a generic error on null ptr for the message.

ikalchev: good catch btw

tests/pytests/test_hybrid_timeout.py
ikalchev: No, opened https://redislabs.atlassian.net/browse/MOD-15032 now

ikalchev: I mean one of the tickets to this PR talks about timeout warnings cluster mode but we can use it to propagate them properly and do the tail timeouts in the other.

src/coord/hybrid/hybrid_cursor_mappings.c
ikalchev: No, but good point. The non-hybrid debug path rejects "return strict", so I did the same for hybrid to be consistent and added a test for the same.

src/hybrid/hybrid_exec.c
```diff
@@ -774,6 +783,14 @@ static int buildPipelineAndExecute(StrongRef hybrid_ref, HybridPipelineParams *h
     hreq->profileClocks.profilePipelineBuildTime = rs_wall_clock_elapsed_ns(&pipelineClock);
   }
 
+  // Apply debug timeouts after pipeline is built (for _FT.DEBUG FT.HYBRID)
+  if (hreq->debugParams) {
```
nafraf: Consider merging both conditions in a single `if`.

tests/pytests/test_hybrid_timeout.py
nafraf: It doesn't seem to be intentional. The error message doesn't help 😄
I prefer to get the same error message we get in cluster mode.

ikalchev: I prefer to do this together with the other debug commands if you don't mind. I can create a jiras if so.

nafraf: OK. Let's create a new ticket for this.

nafraf (approved):

PR #9228, opened 2026-04-22
fix coordinator potential infinite loop failing to connect

### LINKED ISSUE #9225: [BUG] Single Redis Search node becomes broken after AKS node patching/drain event while Redis cluster health still appears normal
**Describe the bug**  
After an AKS node patching, upgrade, or drain event, a single Redis node running Redis Search becomes unstable and starts causing request timeouts. On the affected node, we repeatedly see log lines such as `IORuntime ID 1: Waiting for all nodes to connect`, while the rest of the Redis cluster appears healthy.

Standard Redis health checks, cluster commands, and resource usage metrics do not indicate an unhealthy cluster state.

One important signal is that `search.CLUSTERINFO` works on all healthy Redis nodes, but on the affected node it returns `(nil)` instead of cluster information. This suggests the issue is isolated to the Search/coordinator state on that single node rather than a total Redis cluster failure.

**To Reproduce**  
1. Deploy Redis Cluster with Redis Search enabled on AKS.
2. Run regular traffic, including search requests.
3. Let AKS perform a node patching, upgrade, maintenance, or drain event that affects one Redis pod/node.
4. Observe that one **random Redis** node starts behaving abnormally and logs repeated `IORuntime ID 1: Waiting for all nodes to connect` messages.
5. Observe request timeouts, even though standard Redis checks still suggest that the cluster is healthy.
6. Run `search.CLUSTERINFO` on all Redis nodes: healthy nodes return normal cluster information, while the affected node returns `(nil)`.

**Expected behavior**  
After an AKS patching or node drain event, a Redis Search node should reconnect and recover cleanly without entering a degraded state that causes request timeouts.

If the node is unhealthy, that should also be reflected clearly in Search or cluster diagnostics rather than Redis appearing healthy while one node silently fails at the Search layer.

**Actual behavior**  
Currently only one Redis node becomes unstable after the AKS node event. Redis itself still appears healthy: normal resource usage, normal-looking cluster output, and normal health checks.

However, requests involving the affected node start timing out, the node logs repeated `Waiting for all nodes to connect` messages, and `search.CLUSTERINFO` on that node returns `(nil)` while healthy nodes return normal output.

**Logs**  
From the affected node (`running.log` in debug mode):

```text
...
1:M 21 Apr 2026 15:11:47.578 - <search> IORuntime ID 1: Waiting for all nodes to connect
1:M 21 Apr 2026 15:11:47.579 - <search> IORuntime ID 1: Waiting for all nodes to connect
1:M 21 Apr 2026 15:11:47.579 . --- Processing packet of type ping, 2664 bytes
1:M 21 Apr 2026 15:11:47.579 . ping packet received: 7a8528c779e4444b23ed59662cf46695e0c47d81
1:M 21 Apr 2026 15:11:47.579 . GOSSIP dd416c5f42ad61bf8fb5eb930d6f59d844f700fd 10.40.8.20:6379@16379 slave
1:M 21 Apr 2026 15:11:47.579 . GOSSIP 4fa490de82bc727b0ae1e9d0f7a8e8efbfa014f2 10.40.8.15:6379@16379 slave
1:M 21 Apr 2026 15:11:47.579 . GOSSIP b2305f1f437e8f4e9a2495596d1970f44ac0cfd1 10.40.8.14:6379@16379 master
1:M 21 Apr 2026 15:11:47.580 - <search> IORuntime ID 1: Waiting for all nodes to connect
1:M 21 Apr 2026 15:11:47.581 - <search> IORuntime ID 1: Waiting for all nodes to connect
....
```

This is particularly confusing because the affected node still appears to process normal Redis cluster gossip traffic, yet the Search runtime remains stuck waiting for all nodes to connect.

**Command output**  
On the affected node:

```text
> search.CLUSTERINFO
(nil)
```

On healthy nodes, `search.CLUSTERINFO` returns normal output such as:

```text
1) "num_partitions"
2) "3"
3) "cluster_type"
4) "redis_oss"
5) "shards"
6) 1) 1) "slots"
      2) 1) "0"
         2) "5460"
      3) "id"
      4) "b2305f1f437e8f4e9a2495596d1970f44ac0cfd1"
      5) "host"
      6) "10.40.8.14"
      7) "port"
      8) "6379"
   2) 1) "slots"
      2) 1) "5461"
         2) "10922"
      3) "id"
      4) "9510854c0a066c245ba04e5be316efeb94ee16ff"
      5) "host"
      6) "10.40.8.17"
      7) "port"
      8) "6379"
   3) 1) "slots"
      2) 1) "10923"
         2) "16383"
      3) "id"
      4) "2f6a4250219bce908891f86aff279ec3b992b901"
```
**Environment (please complete the following information):**  
- OS: Linux 5.15.0-1102-azure x86_64 (containerised)
- Orchestrator: Azure Kubernetes Service (AKS) -> deployed with Kubeblocks redis
- Redis version: 8.6.0
- Redis git SHA: `00000000`
- Redis build id: `a7d515010e105f80`
- Redis mode: `cluster`
- Search module version: `80600`
- TimeSeries module version: `80600`
- ReJSON module version: `80600`
- VectorSet module version: `1`

**Additional context**  
The most confusing part is that Redis and the cluster appear healthy from the outside. Standard commands, health checks, and resource metrics do not show a clear failure, yet one Redis Search node becomes unstable after the AKS node event and causes request timeouts.

After a forced restart of the affected node, the issue disappeared and the cluster returned to normal behaviour.
We also verified node-to-node connectivity inside the Kubernetes network. Using `redis-cli` from each Redis node, we were able to connect successfully to the other Redis nodes. This suggests the issue is not caused by a general internal AKS network connectivity problem.

JoanFM:
### Problem

Under continuous topology refreshes (default cadence: 1 s) and a single shard whose `MRConn` never reaches `MRConn_Connected` (e.g. after an AKS node drain that churns Pod IPs), an IO runtime gets wedged indefinitely — see issue #9225.

The cause is in `topologyAsyncCB`: on **every** incoming topology (even a byte-identical refresh), it unconditionally:

1. sets `loop_th_ready = false` (parks all fanout work in `rqAsyncCb`),
2. restarts the 1 ms `topologyValidationTimer` (which spams `"Waiting for all nodes to connect"`),
3. restarts the 30 s `topologyFailureTimer` (the failsafe that unblocks the queue).

Because refreshes (1 s) are faster than the failsafe (30 s), the failsafe is re-armed before it can ever fire. The IO runtime stays in "validating" forever and queries hang until the pod is restarted.

### Fix

Distinguish **connectivity changes** from **slot-range-only** refreshes, and skip the handshake reset when connectivity is unchanged.

- `src/coord/rmr/cluster_topology.{c,h}`: new `MRClusterTopology_ConnectivityEqual(a, b)` — compares the set of shards by `{node.id, endpoint.host, endpoint.port}`, order-insensitive, ignores slot ranges. `NULL == NULL` is equal; `NULL` vs non-`NULL` is not.
- `src/coord/rmr/rmr.c` — `uvUpdateTopologyRequest`:
  - Typed as `bool (struct UpdateTopologyCtx *)` instead of the old `void *` signature.
  - Computes `connectivityChanged = !MRClusterTopology_ConnectivityEqual(old, new)`.
  - Always swaps the topology pointer inline (so the fanout path still sees current slot ranges).
  - Only calls `IORuntimeCtx_UpdateNodesAndConnectAll` when connectivity changed.
  - Returns `connectivityChanged` to `topologyAsyncCB`.
- `src/coord/rmr/io_runtime_ctx.{c,h}`:
  - New `topoQueueItem` (with `MRTopologyUpdateCallback cb` returning `bool`) replaces the generic `queueItem` in the topology path, so the callback can signal back whether a handshake is required.
  - `topologyAsyncCB` now:
    - runs the callback,
    - on `false` (connectivity unchanged): leaves `loop_th_ready` untouched, does **not** re-arm `topologyValidationTimer` or `topologyFailureTimer`, and calls `triggerPendingItems` so any requests buffered during an earlier handshake window are dispatched;
    - on `true`: original behaviour (clear `loop_th_ready`, arm both timers).
  - `IORuntimeCtx_Schedule_Topology`, `IORuntimeCtx_Free`, and `IORuntimeCtx_Debug_ClearPendingTopo` updated to the new typed `topoQueueItem`.
- `topologyTimerCB`: `"Waiting for all nodes to connect"` demoted from `verbose` to `debug` — this callback fires every 1 ms, so when a shard is genuinely stuck it flooded the log at ~1 kHz per IO runtime.
- Tests:
  - New `tests/cpptests/coord_tests/test_cpp_cluster_topology.cpp` covering `MRClusterTopology_ConnectivityEqual` (same/NULL/empty/shard-count/host/port/node-id/slot-range/order permutations).
  - New regression test `IORuntimeCtxCommonTest.IdenticalTopologyUpdateIsSkipped` in `test_cpp_io_runtime_ctx.cpp` that applies an initial topology, then an identical one (asserts the callback counts it as "no change" and a regular callback still drains — proving `loop_th_ready` was not cleared), then a different one (asserts real changes are still applied).
  - Existing topology callbacks updated to the new `bool (...)` signature.

### Outcome

Under continuous 1 s refreshes with a persistently non-connected shard, the 30 s failsafe is no longer re-armed by every refresh. It fires, unblocks the queue, and distributed queries fail cleanly with an error instead of hanging until the pod is restarted. In steady state (identical topology), the handshake is skipped entirely and pending work is drained immediately.

1. Fixes #9225

1. `topologyAsyncCB` / `IORuntimeCtx_Schedule_Topology` (`src/coord/rmr/io_runtime_ctx.{c,h}`)
2. `uvUpdateTopologyRequest` (`src/coord/rmr/rmr.c`)
3. `MRClusterTopology_ConnectivityEqual` (`src/coord/rmr/cluster_topology.{c,h}`)

#### Mark if applicable

#### Release Notes

#### Release Note proposal

Coordinator: prevent an IO runtime from wedging indefinitely when a single shard's connection fails to recover after a Pod IP change. The topology failsafe timer is no longer re-armed by connectivity-equal refreshes, so distributed search traffic is unblocked instead of hanging until manual pod restart. The high-rate `Waiting for all nodes to connect` log line is also demoted to `debug` (#9225).

---

> [!NOTE]
> **Medium Risk**
> Changes the IO-runtime topology refresh/handshake control flow and timer behavior; mistakes could leave runtimes serving with stale connectivity state or skipping needed reconnects, though it is scoped and covered by new regression tests.
> 
> **Overview**
> Prevents IO runtimes from getting wedged under frequent topology refreshes by **skipping the connection-validation handshake/timer reset when only slot metadata changes**.
> 
> Adds `MRClusterTopology_ConnectivityEqual()` (order-insensitive, compares shard `node.id` + `host`/`port`, ignores slot ranges) and threads this signal through the topology update path via a new typed `MRTopologyUpdateCallback`/`topoQueueItem` so `topologyAsyncCB` can keep `loop_th_ready` and avoid re-arming validation/failsafe timers on connectivity-equal updates. Also demotes the 1ms “waiting for nodes to connect” log to `debug` and adds targeted unit/regression tests (including issue #9225).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f5dbbc0379addbc655017d6ba6098a4ff7679dd8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


GuyAv46:
Moved to the refactor #9251.
Since this fix only handles the symptom of not having a valid topology, I don't think it's worth backporting

PR #9251, opened 2026-04-23
Refactor `MRConn` state machine

GuyAv46:
### Current
1. `MRConn` state transitions are scattered across call sites and mixed with teardown; the retry timer is heap-allocated on the side with bespoke cleanup, and `MRConn_Disconnected` only exists as a sync marker between disconnect and free.
2. Pool lifecycle is split across `MRConnPool_Disconnect`/`MRConnPool_Free`/`MRConnManager_Stop`/`MRConnManager_Free`/`MRConnManager_ConnectAll` (and the `IORuntimeCtx_*` counterparts), so node replacement is a multi-step dance that each call site has to get right.
3. Every `MR_UpdateTopology` resets `loop_th_ready` and arms the validation + failure timers, even when the new topology is identical to the old one. With 1 s refreshes and a single unreachable shard, the handshake never completes and the io runtime stays wedged (#9228).
4. The "same address" shortcut in `MRConnManager_Add` ignores `password` and `unixSock`, so credential rotations silently reuse the old (auth-failing) pool.
5. `MRConn_CreateSSLContext` takes raw `(char*, size_t)` slices and leaves the OpenSSL passphrase userdata pointing at the originating `RedisModuleString` after `SSL_CTX_set_default_passwd_cb_userdata` — a use-after-free if the callback fires after the string is released.

### Change
1. **Single state machine.** All transitions go through `MRConn_SwitchState`, which is the only caller of `doConnect` / `doAuthenticate`. The connection-progress states are split into in-flight vs back-off pairs:

   | State | Phase |
   |---|---|
   | `Connecting` / `Reconnecting` | TCP/TLS in flight / back-off before retry |
   | `Authenticating` / `ReAuth` | AUTH in flight / back-off before retry |

   `MRConn_Disconnected` is removed. RESP version negotiation is captured by the new `MRConnProtocol` enum.

2. **Timer invariant.** The retry timer is inlined as a `uv_timer_t` on `MRConn` and closed through `uv_close`. `SwitchState` asserts the timer is only ever armed in `Reconnecting` / `ReAuth`, and the only transitions that may interrupt it are `Freeing` (explicit stop) and `Reconnecting` out of an unexpected disconnect. Conn freeing (struct, endpoint, timer close) is inlined into `SwitchState`'s `Freeing` arm so there is one disposal path.

3. **Dict-owned pool lifecycle.** `MRConnPool_Free` (the dict value destructor) disconnects and frees every conn, so node removal/replacement collapses to a single `dictDelete`/`dictReplace`. `MRConnManager_Shutdown` becomes the sole uv-thread teardown entrypoint, replacing `MRConnManager_Stop`/`Free`/`ConnectAll` and `IORuntimeCtx_ConnectAll`/`UpdateNodesAndConnectAll`.

4. **Handshake gating.** New `MREndpoint_Equal` (host + port + unixSock + password) drives `MRConnManager_Add`, which now returns `bool` (pool (re)created vs reused). `IORuntimeCtx_UpdateNodes` aggregates those signals and forwards them to `topologyAsyncCB` via `UVRuntime.topology_needs_handshake`, so topology refreshes that don't change connectivity skip arming the validation + failure timers.

5. **Safer TLS setup.** `MRConn_CreateSSLContext` now takes `RedisModuleString*` directly (eliminating manual `strlen` boundaries at the call site) and clears the OpenSSL passphrase userdata before returning, so the passphrase callback can never observe a freed buffer.

6. **`MRConn` cache layout.** Struct fields are reordered so that the hot fields read on every callback entry (`conn`, `state`, `protocol`) sit at the head of the first cache line, followed by the warm `ep` / `loop` (init + connect paths). The libuv timer stays last because it is large and only touched while in `Reconnecting` / `ReAuth`.

### Worth mentioning
- Ownership-named hiredis helpers (`detachCbData` / `detachAndFreeAc` / `detachAndDisconnectAc`) and a factored `MRConn_InitTLS` replace the old overloaded `detachFromConn(shouldFree)`.
- `Connecting` vs `Reconnecting` (and `Authenticating` vs `Re-Authenticating`) is now visible via the debug command, so initial vs retry is distinguishable at a glance.
- `RS_ABORT_FMT` logs the offending value on the unreachable default in `SwitchState` instead of a bare `abort()`.
- Removed the `authFailCount` field on `MRConn` and the `AUTH_FAIL_LOG_INTERVAL` rate-limiting around the AUTH-failure log: the counter was only ever incremented on the local-send-failure path (which is a connection issue, not an auth issue), and a per-attempt warning on actual server-side AUTH rejection is fine without throttling.
- `MRConnManager_Shrink` / `MRConnManager_Expand` and the per-pool connection counters are narrowed from `size_t` to `uint32_t` (the pool size is bounded by the configured per-node connection count).
- Trimmed unused standard headers from `conn.c` (`signal.h`, `sys/param.h`, `stdio.h`, `assert.h`).
- New tests cover `MREndpoint_Equal` (C, `test_cluster.c`) and identical-topology handshake skipping (C++, `test_cpp_io_runtime_ctx.cpp`). The latter previously raced `topologyAsyncCB`'s post-callback bookkeeping when the test thread wrote `loop_th_ready = true` directly; it now uses `simulateConnectedTopologyCallback`, which performs all state transitions and `pendingItems` draining on the uv thread, faithfully mimicking the production handshake-completion path.

### Outcome
1. Connection teardown is race-free and has a single owner; the retry timer can no longer outlive or be double-closed against its conn.
2. An unreachable shard can no longer wedge the runtime across back-to-back topology refreshes.
3. Password / Unix-socket rotations force the pool to rebuild.
4. The TLS passphrase callback can no longer dereference freed memory.

1. #9228

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Refactors connection lifecycle/state management and TLS/auth setup in the coordinator’s libuv/hiredis path; bugs here could cause reconnect loops, leaked handles, or stalled IO threads if edge cases are missed.
> 
> **Overview**
> **Refactors `MRConn` into a single, explicit state machine** with new states (`Reconnecting`, `Authenticating`) and a typed `MRConnProtocol`, centralizing connect/auth/retry transitions in `MRConn_SwitchState` and adding a formatted abort macro (`RS_ABORT_FMT`) for unreachable states.
> 
> **Simplifies connection-pool ownership and teardown** by making the dict value destructor (`MRConnPool_Free`) responsible for disconnect+free, introducing `MRConnManager_Shutdown` as the single uv-thread shutdown path, and updating `IORuntimeCtx` shutdown accordingly (including an inlined `uv_timer_t` per connection with async-free via `uv_close`).
> 
> **Improves topology update behavior and endpoint change detection**: adds `MREndpoint_Equal` (now includes host/port/unix socket/password) so endpoint/password rotations rebuild pools, and introduces `uv_runtime.topology_needs_handshake` so identical topology updates can skip re-arming validation/failure timers and avoid wedging the IO runtime.
> 
> **Hardens TLS initialization** by refactoring TLS setup into `MRConn_InitTLS`, changing `MRConn_CreateSSLContext` to take `RedisModuleString*` and clearing OpenSSL passphrase userdata to avoid dangling references.
> 
> Adds/updates unit tests covering `MREndpoint_Equal` and the “identical topology update skips handshake” behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7ea649dab677bccca77e265712e0ed48ee23dbd7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/coord/rmr/io_runtime_ctx.c
```diff
@@ -311,7 +321,6 @@ void IORuntimeCtx_Free(IORuntimeCtx *io_runtime_ctx) {
     UV_Close(io_runtime_ctx);
   }
   RQ_Free(io_runtime_ctx->queue);
-  MRConnManager_Free(&io_runtime_ctx->conn_mgr);
   queueItem *task = exchangePendingTopo(io_runtime_ctx, NULL);
```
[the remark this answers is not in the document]
GuyAv46: Fixed in c53ab2a39: `IORuntimeCtx_Free` now calls `MRConnManager_Shutdown(&io_runtime_ctx->conn_mgr)` on the non-started branch (before `UV_Close`), so the conn manager dict is released when `sideThread` never ran.

src/coord/rmr/conn.c
```diff
@@ -417,102 +382,84 @@ void MRConnManager_Expand(MRConnManager *m, size_t num, uv_loop_t *loop) {
...
+ * back-off so the server has time to recover and we don't tight-loop on
+ * repeated auth rejections. */
+static void reauthTimerCallback(uv_timer_t *tm) {
+  MRConn *conn = tm->data;
+  RS_ASSERT(conn->state == MRConn_ReAuth);
+  if (MRConn_SendAuth(conn) != REDIS_OK) {
+    if (conn->authFailCount % AUTH_FAIL_LOG_INTERVAL == 0) {
+      CONN_LOG_WARNING(conn, "Failed to send AUTH command (%hu consecutive failures)", conn->authFailCount);
+    }
+    conn->authFailCount++;
```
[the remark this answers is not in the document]
GuyAv46: Fixed in c53ab2a39: log now reports `authFailCount + 1`, so the first failure prints "1 consecutive failures" rather than 0. Kept the increment after the log to preserve the modulo-100 cadence.

src/coord/rmr/conn.c
```diff
@@ -716,77 +707,36 @@ static void MRConn_ConnectCallback(const redisAsyncContext *c, int status) {
...
-      conn->authFailCount++;
-      if (conn->authFailCount == 1 || conn->authFailCount % AUTH_FAIL_LOG_INTERVAL == 0) {
-        CONN_LOG_WARNING(conn, "Failed to send AUTH command (%u consecutive failures)", conn->authFailCount);
+      if (conn->authFailCount % AUTH_FAIL_LOG_INTERVAL == 0) {
+        CONN_LOG_WARNING(conn, "Failed to send AUTH command (%hu consecutive failures)", conn->authFailCount);
       }
-      detachFromConn(conn, true);
+      conn->authFailCount++;
+      // AUTH failed to enqueue; the ac is still alive and ours to free.
+      detachAndFreeAc(conn);
```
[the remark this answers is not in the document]
GuyAv46: Fixed in c53ab2a39, same approach as the reauth timer: log prints `authFailCount + 1`.

src/coord/rmr/io_runtime_ctx.c
```diff
@@ -215,30 +226,29 @@ void IORuntimeCtx_UpdateNodes(IORuntimeCtx *ioRuntime) {
...
+    if (MRConnManager_Add(&ioRuntime->conn_mgr, &ioRuntime->uv_runtime.loop, node->id, &node->endpoint)) {
+      connectivityChanged = true;
+    }
     /* This node is still valid, remove it from the nodes to delete list */
     dictDelete(nodesToDisconnect, node->id);
   }
 
   // if we didn't remove the node from the original nodes map copy, it means it's not in the new topology,
   // we need to disconnect the node's connections
+  // We don't count this as a connectivity change, as no new connections are created.
```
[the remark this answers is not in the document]
GuyAv46: Addressed in c53ab2a39 by clarifying the contract rather than flipping the behavior: node *removals* are intentionally not treated as a reason to re-run the validation handshake — no new connections are opened, so existing nodes that were already validated don't need to be re-validated. The flag has been renamed `topology_needs_handshake` (and the local `newConnectionsCreated`) with an updated doc comment on `IORuntimeCtx_UpdateNodes`, and the test (`UpdateNodesReportsNewConnections`) now asserts `ASSERT_FALSE` on the drop-a-node case to match.

src/coord/rmr/io_runtime_ctx.c
```diff
@@ -311,7 +321,6 @@ void IORuntimeCtx_Free(IORuntimeCtx *io_runtime_ctx) {
     UV_Close(io_runtime_ctx);
   }
   RQ_Free(io_runtime_ctx->queue);
-  MRConnManager_Free(&io_runtime_ctx->conn_mgr);
```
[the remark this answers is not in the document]
GuyAv46: Fixed in c53ab2a39: added `MRConnManager_Shutdown(&io_runtime_ctx->conn_mgr)` in the non-started branch of `IORuntimeCtx_Free`, before `UV_Close`. The dict (and any pools that got into it) is now released on both paths.

src/coord/rmr/conn.c
```diff
@@ -701,9 +649,52 @@ static int checkTLS(char** client_key, char** client_cert, char** ca_cert, char*
...
+
+  int rc = redisInitiateSSL(&c->c, ssl);
+  SSL_CTX_free(ssl_context);
+  if (rc != REDIS_OK) {
+    CONN_LOG_WARNING(conn, "Error on tls auth, %s.", c->c.err ? c->c.errstr : "Unknown error");
+    SSL_free(ssl);
+    return REDIS_ERR;
+  }
+
+  return REDIS_OK;
```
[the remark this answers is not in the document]
GuyAv46: But do we need the workaround on master now?

src/coord/rmr/conn.c
```diff
@@ -17,32 +17,35 @@
...
   MREndpoint ep;
   redisAsyncContext *conn;
-  void *timer;
   uv_loop_t *loop;
-  int protocol; // 0 (undetermined), 2, or 3
-  MRConnState state;
-  unsigned authFailCount; // consecutive auth failures, for rate-limited logging
-} MRConn;
+  uint16_t authFailCount; // consecutive auth failures, for rate-limited logging
+  uint8_t state;          // MRConnState
```
JoanFM: Why did we remove the actual enum type?

src/coord/rmr/conn.c
```diff
@@ -17,32 +17,35 @@
...
+  uint8_t protocol;       // 0 (undetermined), 2, or 3
```
JoanFM: can we have an enum for the protocol?

src/coord/rmr/conn.c
```diff
@@ -17,32 +17,35 @@
...
+  uv_timer_t timer;       // back-off timer for Connecting/ReAuth; inlined so its
```
JoanFM: Can u rephrase, restructure this sentence. Whose lifetime is adjusted to and who owns teardown is weird.

src/coord/rmr/conn.c
```diff
@@ -57,24 +60,46 @@ static int MRConn_SendAuth(MRConn *conn);
...
-  conn->conn = NULL;
   conn->protocol = 0;
-  if (shouldFree) {
-    redisAsyncFree(ac);
+  if (!ac) {
     return NULL;
-  } else {
-    return ac;
   }
+  // Freeing the cbData, not the connection or the uvloop.
```
JoanFM: Is this comment out of date now?

src/coord/rmr/conn.c
```diff
@@ -57,24 +60,46 @@ static int MRConn_SendAuth(MRConn *conn);
...
+ * on our behalf. Safe to call when conn is not attached. */
+static void detachAndFreeAc(MRConn *conn) {
+  redisAsyncContext *ac = _detachAc(conn);
+  if (ac) redisAsyncFree(ac);
+}
+
+/* Detach cbData and request a graceful shutdown of the hiredis async
+ * context. hiredis fires the disconnect callback (with cbData already NULL)
+ * and then frees the ac itself. */
+static void disconnectHiredisAc(MRConn *conn) {
```
JoanFM: can we call it detachAndAsyncDisconnect? or detachAndSyncDisconnectAndFree?

src/coord/rmr/conn.c
```diff
@@ -105,60 +123,50 @@ static MRConnPool *_MR_NewConnPool(MREndpoint *ep, size_t num, uv_loop_t *loop)
...
-    uv_close(conn->timer, (uv_close_cb)rm_free);
-    conn->timer = NULL;
-  }
-}
-
-/*
-* This is called from the uv thread.
-*/
+/* Tear down a connection: log and transition to the terminal Freeing state,
+ * which severs hiredis cbData and schedules the async free of the MRConn
```
JoanFM: what does `severs` mean?

src/coord/rmr/conn.c
```diff
@@ -105,60 +123,50 @@ static MRConnPool *_MR_NewConnPool(MREndpoint *ep, size_t num, uv_loop_t *loop)
...
-  closeTimer(conn);
-  redisAsyncContext *ac = detachFromConn(conn, false);
-  if (ac) {
-    redisAsyncDisconnect(ac);
-  }
+  MRConn_SwitchState(conn, MRConn_Freeing);
 }
 
-static void freeConn(MRConn *conn) {
+/* Close callback for the inlined timer handle. libuv has released the handle
```
JoanFM, line 99: which handles has it freed?

src/coord/rmr/conn.c
```diff
@@ -105,60 +123,50 @@ static MRConnPool *_MR_NewConnPool(MREndpoint *ep, size_t num, uv_loop_t *loop)
...
-/* Free a connection pool.
- * Called both when a node is removed/replaced (runtime topology updates) and when the
- * connection manager is released (teardown). Runtime paths are expected to disconnect first;
- * this function performs final ownership cleanup of MRConn wrappers.
- */
+/* Free the conn. Teardown is asynchronous because the inlined timer handle
+ * must outlive this call until libuv finishes its close sequence; the actual
+ * rm_free happens in freeConnAfterTimerClose. Callers must ensure no further
+ * SwitchState or timer operations run on conn after this point. */
+static void freeConn(MRConn *conn) {
```
JoanFM, line 107: can we call scheduleFreeConnection to not give the impression that this is sync and connection is freed when this returns?

src/coord/rmr/conn.c
```diff
@@ -188,18 +196,13 @@ void MRConnManager_Init(MRConnManager *mgr, int nodeConns) {
...
-}
-
-void MRConnManager_Free(MRConnManager *mgr) {
+/* Tear down every connection in the manager and release the dict.
+ *
+ * Must be called from the uv thread while the event loop is still alive: the
+ * per-conn disconnect path invokes uv_close and redisAsyncDisconnect, both of
+ * which require a live loop. After this returns the manager is empty; the
+ * MRConnManager struct itself is not freed (it is embedded in IORuntimeCtx). */
+void MRConnManager_Shutdown(MRConnManager *mgr) {
```
JoanFM, line 168: in the rmr.c we have some convention to prefix methods that should only be called in the uvthread with `uv`. Should we apply this? Although I am not sure this would be accurate here, or if u mean this should be scheduled explicitly in the `uvThread`? I think having some protocol like this could help not introducing bugs in the future.

src/coord/rmr/conn.h
```diff
@@ -23,13 +23,9 @@ extern "C" {
...
  * The state of the connection.
- * TODO: Not all of these are "real" states
  */
 typedef enum {
-  /* initial state - new connection or disconnected connection due to error */
-  MRConn_Disconnected,
-
-  /* Connection is trying to connect */
+  /* Connection is trying to connect (initial state and retry state) */
   MRConn_Connecting,
```
JoanFM, line 29: Can we explicitly have MRConn_Retrying orsomething to be more explicit? I think is best to reflect the actual state of cases, and then the methods handling the states can merge what they dowith them, but semantically the state transition will make more sense and be more understandable?

src/coord/rmr/endpoint.c
```diff
@@ -84,3 +84,18 @@ void MREndpoint_Free(MREndpoint *ep) {
...
   }
 }
+
+static inline bool strEqOrBothNull(const char *a, const char *b) {
+  if (a == b) return true;
+  if (!a || !b) return false;
+  return strcmp(a, b) == 0;
+}
+
+bool MREndpoint_Equal(const MREndpoint *a, const MREndpoint *b) {
```
JoanFM, line 94: let's add some unit tests for this

src/coord/config.c
```diff
@@ -268,6 +307,12 @@ static RSConfigOptions clusterOptions_g = {
                          "Default is 30000 (30 seconds). 0 means no timeout.",
              .setValue = setTopologyValidationTimeout,
              .getValue = getTopologyValidationTimeout,},
+            {.name = "CONNECT_TIMEOUT",
```
JoanFM, line 310: Add tests for these configs.

JoanFM (changes requested):

src/coord/rmr/conn.c
GuyAv46: Added `MRConnProtocol` (packed enum, values `MRConn_Protocol_Undetermined = 0`, `MRConn_Protocol_RESP2 = 2`, `MRConn_Protocol_RESP3 = 3`) in `conn.h` and switched the field to use it. `MRCommand.protocol` is left as `unsigned char` for now (broader change across command.c/rpnet.c/dist_utils.c); a cast at the single interop point in `MRConn_SendCommand` bridges the two.

src/coord/rmr/conn.c
GuyAv46: Reinstated `MRConnState` as the field type and marked the enum `__attribute__((packed))` so it still occupies one byte. Same treatment applied to the new `MRConnProtocol` enum for the protocol field (see reply on the other thread).

src/coord/rmr/conn.c
GuyAv46: I'll just remove it. It's obvious that an embedded member has the same lifetime as the struct

src/coord/rmr/conn.c
GuyAv46: disconnect/detach

src/coord/rmr/conn.c
GuyAv46: the timer handle

src/coord/rmr/conn.c
GuyAv46: I don't think so, here the actual meaning is "the owner thread"

src/coord/rmr/conn.c
```diff
@@ -417,102 +382,84 @@ void MRConnManager_Expand(MRConnManager *m, size_t num, uv_loop_t *loop) {
...
+    case MRConn_Freeing:
+      // Terminal state: sever cbData so any in-flight hiredis callbacks on
+      // this conn become no-ops, then hand the conn off to freeConn. uv_close
+      // inside freeConn synchronously stops the retry timer and defers the
+      // struct free until the close callback fires.
+      disconnectHiredisAc(conn);
+      freeConn(conn);
+      return;
   }
+  RS_ABORT_FMT("MRConn_SwitchState: invalid target state %d", nextState);
```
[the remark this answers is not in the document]
GuyAv46: completely redundant

src/coord/rmr/conn.h
GuyAv46: I considered that, but ended up logging it instead.
We will get an "initial attempt" and "retrying" logs, but logically the states are identical

src/coord/config.c
GuyAv46: Irrelevant after rebase

src/coord/rmr/endpoint.c
GuyAv46: Added `testEndpointEqual` in `tests/ctests/coord_tests/test_cluster.c` covering:
- NULL / NULL, NULL vs non-NULL, same-pointer
- equal endpoints (host + port), differing port, differing host
- password: NULL vs non-NULL, both-equal, differing
- unixSock: NULL vs non-NULL, both-equal, differing
- host: NULL vs non-NULL, both-NULL

All 4 tests in the suite pass (141 assertions).

src/coord/rmr/conn.c
```diff
@@ -17,32 +17,34 @@
...
-typedef struct MRConn{
+// Layout packs the three small fields into the padding slot before the
+// inlined timer, keeping ep+conn+loop+state on the first cache line.
+struct MRConn {
   MREndpoint ep;
   redisAsyncContext *conn;
-  void *timer;
   uv_loop_t *loop;
-  int protocol; // 0 (undetermined), 2, or 3
+  uint16_t authFailCount;   // consecutive auth failures, for rate-limited logging
```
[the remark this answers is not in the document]
JoanFM: It would emit another log, nothing really problematic

src/coord/rmr/conn.c
```diff
@@ -57,24 +59,45 @@ static int MRConn_SendAuth(MRConn *conn);
...
+ * on our behalf. Safe to call when conn is not attached. */
+static void detachAndFreeAc(MRConn *conn) {
+  redisAsyncContext *ac = _detachAc(conn);
+  if (ac) redisAsyncFree(ac);
+}
+
+/* Detach cbData and request a graceful shutdown of the hiredis async
+ * context. hiredis fires the disconnect callback (with cbData already NULL)
+ * and then frees the ac itself. */
+static void detachAndDisconnectAc(MRConn *conn) {
```
JoanFM: Is the Disconnection async?

src/coord/rmr/conn.c
```diff
@@ -417,102 +380,84 @@ void MRConnManager_Expand(MRConnManager *m, size_t num, uv_loop_t *loop) {
...
 
-activate_timer:
-  if (conn->timer && !uv_is_active(conn->timer)) {
-    uv_timer_start(conn->timer, signalCallback, nextTimeout, 0);
+    case MRConn_Freeing:
+      // Terminal state: sever cbData so any in-flight hiredis callbacks on
+      // this conn become no-ops, then hand the conn off to freeConn. uv_close
+      // inside freeConn synchronously stops the retry timer and defers the
+      // struct free until the close callback fires.
+      detachAndDisconnectAc(conn);
```
JoanFM: Can we have a semantically significant wrapper around these 2 operations? That could make this method more readable?

src/coord/rmr/conn.h
```diff
@@ -22,14 +22,11 @@ extern "C" {
...
+ * in MRConn.
  */
-typedef enum {
-  /* initial state - new connection or disconnected connection due to error */
-  MRConn_Disconnected,
-
-  /* Connection is trying to connect */
+typedef enum __attribute__((packed)) {
+  /* Connection is trying to connect (initial state and retry state) */
   MRConn_Connecting,
```
JoanFM, line 29: I think MRConn_Authenticating would be a nice state to add that would help map actual state and make the code more understandable.

src/coord/rmr/conn.c
```diff
@@ -550,33 +503,26 @@ static void MRConn_AuthCallback(redisAsyncContext *c, void *r, void *privdata) {
 }
 
 static int MRConn_SendAuth(MRConn *conn) {
+  // Callers react to REDIS_ERR by detaching and switching back to Connecting.
+  RS_ASSERT(conn->state == MRConn_Connecting || conn->state == MRConn_ReAuth);
   CONN_LOG(conn, "Authenticating...");
```
JoanFM: I would add here some MRConn_Authenticating state.

src/coord/rmr/conn.c
GuyAv46: Yes, i may wait for pending replies and so on before actually disconnecting

JoanFM: lets reflect this in the name

src/coord/rmr/conn.c
```diff
@@ -417,102 +379,103 @@ void MRConnManager_Expand(MRConnManager *m, size_t num, uv_loop_t *loop) {
...
+  conn->state = nextState;
 
   switch (nextState) {
-    case MRConn_Disconnected:
-      // We should never *switch* to this state
-      abort();
+    case MRConn_Reconnecting:
+      // Delayed state: the reconnect timer introduces a back-off between
+      // attempts so we don't spin on a server that just rejected us.
+      uv_timer_start(&conn->timer, reconnectTimerCallback, RECONNECT_MS_DELAY, 0);
```
JoanFM, line 408: I see that MRConn_Reconnecting switch state tends to come together with some `detaching`and stuff.

While I know it is not always the exact same operation, can we evaluate bringing this logic here?

I love how much easier u have made following the state of connections, but I think this would make it even easier to understand.

JoanFM: Maybe it can be a signal for a need of another state? Kind of another flavour of Reconnecting?

JoanFM: Or pass some argument about the ownership that semantically makes it easy to follow.

JoanFM (changes requested):
I love the direction of this refactor! It indeed makes the code way more readible.

Just a small change I would do on top.

src/coord/rmr/conn.c
GuyAv46: The detachment operation is bounded to the reason we need to restart, I think it makes sense to require the connection to have no `ac` attached when switching to the re/connecting states. We assert that in `MRConn_Connect`.
Adding context to the switch state will make it harder to follow IMO

JoanFM: Follow right now the code, it is almost everything a very natural flow of MRConn_SwitchState which is very easy to follow, but is only contaminated by these `detach...` calls. I feel these are details that kind of "pollute" the logic and may be worth encoding them in an argument that is easy to understand or into a new state variant.

JoanFM: So in general. I have the feeling that the logical flow of states is very clear. And right now what is not very clear is the ownership of the connection and the hiredis ac.

src/coord/rmr/conn.c
```diff
@@ -841,16 +819,16 @@ static int MRConn_Connect(MRConn *conn) {
...
-  conn->protocol = 0;
+  conn->protocol = MRConn_Protocol_Undetermined;
 
   if (redisLibuvAttach(conn->conn, conn->loop) != REDIS_OK ||
       redisAsyncSetConnectCallback(conn->conn, MRConn_ConnectCallback) != REDIS_OK ||
       redisAsyncSetDisconnectCallback(conn->conn, MRConn_DisconnectCallback) != REDIS_OK) {
     CONN_LOG_WARNING(conn, "Failed to attach hiredis context to libuv");
-    detachFromConn(conn, true);
+    // Attach failed before hiredis took ownership; free the ac ourselves.
+    detachAndFreeAc(conn);
```
JoanFM: This can be moved to the caller before switching state to Reconnecting. Like this the pattern is more explicit that detachment happens before reconnecting,

JoanFM (changes requested):

src/coord/rmr/conn.c
JoanFM: Ah no, maybe is not so straight forward.

JoanFM: Maybe with a new error code u could.

src/coord/rmr/conn.c
GuyAv46: After some more digging, we don't really need to differentiate between cases, and we can call `redisAsyncDisconnect` in all of them.
1. `redisAsyncDisconnect` -> `redisAsyncDisconnect` - no change
2. `redisAsyncFree` -> `redisAsyncDisconnect` - since we only use it in the connecting path, before the connection is actually used, these are practically the same (the only difference is waiting for replies and handling callbacks, which we don't have)
3. nothing (from error paths in connect callbacks) -> `redisAsyncDisconnect` - this is no-op since we are in a callback, it only turns on the disconnecting flag. On these cases hiredis is disconnecting the connection anyway, but this is safe

So I'm removing the different helpers and moving the invocation into the state switching to "reconnect" (immediately, not after the timer delay, so we don't hold to an invalid context)

JoanFM (dismissed):

JoanFM (approved):

PR #9276, opened 2026-04-24
[MOD-14201] TopKIterator Core State Machine (Rust)

ThierryBerger:
- based on https://github.com/RediSearch/RediSearch/pull/9170
- Also integrates Unfiltered Direct-Yield path.

### Summary

This PR introduces `TopKIterator`, the generic top-k query execution engine for RediSearch's vector search and scoring pipelines.

The iterator is parameterized over a `ScoreSource` trait, keeping it decoupled from any particular scoring backend (vector vs numeric field). This design allows the same state machine to serve different query shapes by swapping the source, not by duplicating logic.

The first execution mode implemented is **Unfiltered**: when there is no child filter iterator to intersect against, results are streamed directly from the source's batch without materializing a heap at all.
- This is not using https://github.com/RediSearch/RediSearch/pull/9170 as vector source implementation requires (for performance) a dedicated logic for the unfiltered path.
### What's included

- **`ScoreSource` / `ScoreBatch` traits** — the contract that pluggable scoring backends must implement
- **`TopKIterator`** — the state machine with phase tracking (`NotStarted → Collecting → Yielding`), `rewind()` support, `last_doc_id` bookkeeping.
- **`MockScoreSource`** — a test/benchmark harness that drives the iterator with synthetic data, gated behind a `test-utils` feature flag
- **Integration tests** — covering state transitions, rewind correctness, EOF behavior, and `num_estimated` capping
- **`top_k_bencher` crate** — Criterion micro-benchmarks for both the heap and the iterator, providing a regression baseline before real score sources are wired in

### What comes next

The `Batches` mode (merge-join intersection against a child filter) is not yet implemented — it is the next revision in this stack.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Adds a new `TopKIterator` implementation and supporting traits/tests in the Rust query-iterator stack, which could affect result iteration semantics and error propagation once integrated. Build/coverage scripts are also updated to account for the new benchmark crate, with low but non-zero CI/build risk.
> 
> **Overview**
> **Adds a new generic top-k iterator core in Rust.** Introduces `TopKIterator` (with `TopKMode::Unfiltered` direct-yield execution), plus new `ScoreSource`/`ScoreBatch` traits and a `test-utils` mock implementation to drive unit/integration tests.
> 
> **Expands the workspace with benchmarking support.** Adds a new `top_k_bencher` Criterion crate and updates `build.sh` coverage and build exclusions to avoid linking issues from benchmark crates that depend on C symbols; `top_k` also gains `build.rs`, new feature flags, and additional dependencies to support the iterator and tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4158fae8d881d9a217ca752c430c5744ee4a52e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/top_k/Cargo.toml
+++ b/src/redisearch_rs/top_k/Cargo.toml
@@ -8,6 +8,28 @@ publish.workspace = true
 [lib]
 bench = false
 
+[features]
+# Exposes MockScoreSource and MockScoreBatch for use in tests and benchmarks.
+test-utils = []
+# Links C static libraries at test time (see build.rs).
+# https://github.com/rust-lang/cargo/issues/4789#issuecomment-2308131243
+unittest = []
+
+[build-dependencies]
+build_utils = { path = "../build_utils" }
+
 [dependencies]
 ffi.workspace = true
+inverted_index.workspace = true
+rqe_iterator_type.workspace = true
+rqe_iterators.workspace = true
 workspace_hack.workspace = true
+
+[dev-dependencies]
+top_k = { path = ".", features = ["unittest", "test-utils"] }
+redis_mock.workspace = true
+rqe_iterators = { workspace = true, features = ["unittest"] }
+rqe_iterators_test_utils.workspace = true
+redisearch_rs = { path = "../c_entrypoint/redisearch_rs", features = [
+    "mock_allocator",
+] }
--- a/src/redisearch_rs/top_k/Cargo.toml
+++ b/src/redisearch_rs/top_k/Cargo.toml
@@ -10,7 +10,7 @@ bench = false
 doctest = false
 
 [features]
-# Exposes MockScoreSource and MockScoreBatch for use in tests and benchmarks.
+# Exposes mocks for use in tests and benchmarks.
 test-utils = []
 # Links C static libraries at test time (see build.rs).
 # https://github.com/rust-lang/cargo/issues/4789#issuecomment-2308131243
@@ -25,6 +25,8 @@ inverted_index.workspace = true
 rqe_iterator_type.workspace = true
 rqe_iterators.workspace = true
 workspace_hack.workspace = true
+index_result.workspace = true
+index_spec.workspace = true
 
 [dev-dependencies]
 top_k = { path = ".", features = ["unittest", "test-utils"] }
--- a/src/redisearch_rs/top_k/Cargo.toml
+++ b/src/redisearch_rs/top_k/Cargo.toml
@@ -7,6 +7,7 @@ publish.workspace = true
 
 [lib]
 bench = false
+doctest = false
 
 [features]
 # Exposes MockScoreSource and MockScoreBatch for use in tests and benchmarks.
--- a/src/redisearch_rs/top_k/src/iterator.rs
+++ b/src/redisearch_rs/top_k/src/iterator.rs
@@ -1,0 +1,260 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! [`TopKIterator`] — the generic top-k state machine.
+
+use std::num::NonZeroUsize;
+
+use ffi::t_docId;
+use inverted_index::RSIndexResult;
+use rqe_iterator_type::IteratorType;
+use rqe_iterators::{RQEIterator, RQEIteratorError, RQEValidateStatus, SkipToOutcome};
+
+use crate::traits::{ScoreBatch, ScoreSource};
+
+// ── Public enums ─────────────────────────────────────────────────────────────
+
+/// Determines which collection algorithm [`TopKIterator`] uses.
+///
+/// Selected at construction based on whether a child filter is present,
+/// and may be switched mid-execution when the source decides a different
+/// strategy is more efficient.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum TopKMode {
+    /// No child filter — stream directly from the source's single batch.
+    /// The heap is bypassed entirely.
+    Unfiltered,
+}
+
+// ── Private phase enum ────────────────────────────────────────────────────────
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum Phase {
+    /// Collection has not started; the first call to [`read`](TopKIterator::read)
+    /// will trigger it.
+    NotStarted,
+    /// Actively collecting results (transient; only visible inside collection methods).
+    Collecting,
+    /// Collection is done; yielding results from `results` in order.
+    #[expect(
+        dead_code,
+        reason = "TopKMode::Unfiltered is without child so no yielding necessary"
+    )]
+    Yielding,
+    /// Unfiltered path: yielding directly from `direct_batch` without a heap.
+    YieldingDirect,
+}
+
+// ── TopKIterator ──────────────────────────────────────────────────────────────
+
+/// A generic top-k iterator parameterized over a [`ScoreSource`].
+///
+/// Implements the execution mode described in the design doc:
+/// [`Unfiltered`](TopKMode::Unfiltered).
+pub struct TopKIterator<'index, S: ScoreSource<'index>> {
+    source: S,
+    child: Option<Box<dyn RQEIterator<'index> + 'index>>,
+    mode: TopKMode,
+    /// Preserved so [`rewind`](Self::rewind) can restore the original mode.
+    initial_mode: TopKMode,
+    /// Holds the in-progress batch for the Unfiltered path.
+    direct_batch: Option<S::Batch>,
+    k: NonZeroUsize,
+    phase: Phase,
+    current: Option<RSIndexResult<'index>>,
+    last_doc_id: t_docId,
+    at_eof: bool,
+}
+
+impl<'index, S: ScoreSource<'index>> TopKIterator<'index, S> {
+    /// Create a new [`TopKIterator`].
+    ///
+    /// The execution mode is inferred from `child`.
+    ///
+    /// For now, only [`TopKMode::Unfiltered`] exists.
+    pub fn new(
+        source: S,
+        child: Option<Box<dyn RQEIterator<'index> + 'index>>,
+        k: NonZeroUsize,
+    ) -> Self {
+        let mode = TopKMode::Unfiltered;
+        Self::new_with_mode(source, child, k, mode)
+    }
+
+    /// Create a new [`TopKIterator`] with an explicit initial mode.
+    pub fn new_with_mode(
+        source: S,
+        child: Option<Box<dyn RQEIterator<'index> + 'index>>,
+        k: NonZeroUsize,
+        mode: TopKMode,
+    ) -> Self {
+        Self {
+            source,
+            child,
+            mode,
+            initial_mode: mode,
+            direct_batch: None,
+            k,
+            phase: Phase::NotStarted,
+            current: None,
+            last_doc_id: 0,
+            at_eof: false,
+        }
+    }
+
+    // ── Collection dispatch ───────────────────────────────────────────────────
+
+    /// Drive collection based on the current mode.
+    fn collect(&mut self) -> Result<(), RQEIteratorError> {
+        self.phase = Phase::Collecting;
+        let result = match self.mode {
+            TopKMode::Unfiltered => self.prepare_unfiltered_direct(),
+        };
+        if result.is_err() {
+            // Restore a stable phase so that a subsequent read() call can retry.
+            // Phase::Collecting is only valid transiently inside this method.
+            // TODO: MOD-14209: bubble up errors
+            self.phase = Phase::NotStarted;
+        }
+        result
+    }
+
+    // ── Unfiltered path ───────────────────────────────────────────────────────
+
+    /// Set up the unfiltered direct-yield path.
+    ///
+    /// Calls [`ScoreSource::next_batch`] exactly once.  Results are streamed
+    /// directly from the batch cursor — no heap is involved.
+    fn prepare_unfiltered_direct(&mut self) -> Result<(), RQEIteratorError> {
+        self.direct_batch = self.source.next_batch()?;
+        if self.direct_batch.is_none() {
+            self.at_eof = true;
+        }
+        self.phase = Phase::YieldingDirect;
+        Ok(())
+    }
+
+    // ── Yielding helpers ─────────────────────────────────────────────────────
+
+    /// Yield the next result from the unfiltered direct batch.
+    fn advance_unfiltered_direct(
+        &mut self,
+    ) -> Result<Option<&mut RSIndexResult<'index>>, RQEIteratorError> {
+        let item = if let Some(batch) = &mut self.direct_batch {
+            batch.next()
+        } else {
+            None
+        };
+
+        match item {
+            Some((doc_id, score)) => {
+                let result = self.source.build_result(doc_id, score);
+                self.current = Some(result);
+                self.last_doc_id = doc_id;
+                Ok(self.current.as_mut())
+            }
+            None => {
+                self.at_eof = true;
+                Ok(None)
+            }
+        }
+    }
+}
+
+// ── RQEIterator impl ──────────────────────────────────────────────────────────
+
+impl<'index, S: ScoreSource<'index>> RQEIterator<'index> for TopKIterator<'index, S> {
+    #[inline(always)]
+    fn current(&mut self) -> Option<&mut RSIndexResult<'index>> {
+        self.current.as_mut()
+    }
+
+    #[inline(always)]
+    fn read(&mut self) -> Result<Option<&mut RSIndexResult<'index>>, RQEIteratorError> {
+        if self.at_eof {
+            return Ok(None);
+        }
+
+        if self.phase == Phase::NotStarted {
+            self.collect()?;
+        }
+
+        match self.phase {
+            Phase::YieldingDirect => self.advance_unfiltered_direct(),
+            Phase::Yielding => {
+                unreachable!("collect() must set phase to YieldingDirect or Yielding")
+            }
+            Phase::NotStarted | Phase::Collecting => {
+                unreachable!("collect() must set phase to YieldingDirect or Yielding")
+            }
+        }
+    }
+
+    fn skip_to(
+        &mut self,
+        _doc_id: t_docId,
+    ) -> Result<Option<SkipToOutcome<'_, 'index>>, RQEIteratorError> {
+        // TopKIterator is a root-only iterator that yields results sorted by
+        // score, not by doc_id.  It cannot be used as a child in a larger
+        // iterator tree, so skip_to is unsupported.
+        unimplemented!("TopKIterator is a root-only iterator; skip_to is not supported")
+    }
+
+    #[inline(always)]
+    unsafe fn revalidate(
+        &mut self,
+        spec: std::ptr::NonNull<ffi::IndexSpec>,
+    ) -> Result<RQEValidateStatus<'_, 'index>, RQEIteratorError> {
+        if let Some(child) = &mut self.child {
+            // SAFETY: Delegating to child with the same `spec` passed by our caller.
+            return unsafe { child.revalidate(spec) };
+        }
+        Ok(RQEValidateStatus::Ok)
+    }
+
+    #[inline(always)]
+    fn rewind(&mut self) {
+        self.source.rewind();
+        if let Some(child) = &mut self.child {
+            child.rewind();
+        }
+        self.mode = self.initial_mode;
+        self.direct_batch = None;
+        self.current = None;
+        self.last_doc_id = 0;
+        self.at_eof = false;
+        self.phase = Phase::NotStarted;
+        // We explicitly do NOT reset self.metrics because they're used for diagnostics.
+    }
+
+    #[inline(always)]
+    fn num_estimated(&self) -> usize {
+        self.k.get().min(self.source.num_estimated())
+    }
+
+    #[inline(always)]
+    fn last_doc_id(&self) -> t_docId {
+        self.last_doc_id
+    }
+
+    #[inline(always)]
+    fn at_eof(&self) -> bool {
+        self.at_eof
+    }
+
+    #[inline(always)]
+    fn type_(&self) -> IteratorType {
+        self.source.iterator_type()
+    }
+
+    #[inline(always)]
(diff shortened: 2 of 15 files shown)


src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// will trigger it.
+    NotStarted,
+    /// Actively collecting results (transient; only visible inside collection methods).
+    Collecting,
+    /// Collection is done; yielding results from `results` in order.
+    #[expect(
+        dead_code,
+        reason = "TopKMode::Unfiltered is without child so no yielding necessary"
+    )]
+    Yielding,
```
ThierryBerger, line 54: - https://github.com/RediSearch/RediSearch/pull/9424 will use it.

src/redisearch_rs/top_k/Cargo.toml
```diff
@@ -7,7 +7,30 @@ publish.workspace = true
 
 [lib]
 bench = false
+doctest = false
+
+[features]
+# Exposes MockScoreSource and MockScoreBatch for use in tests and benchmarks.
```
memark: While the comment is good, do we need to name the explicit types here? Feels like it could quickly get out of sync.

src/redisearch_rs/top_k/tests/integration/iterator.rs
```diff
@@ -0,0 +1,232 @@
...
+    assert_eq!(result.doc_id, 1);
+}
+
+#[test]
+fn rewind_resets_to_not_started() {
+    let source = MockScoreSource::new(vec![vec![(1, 1.0), (2, 2.0)]]);
+    let mut it = TopKIterator::new(source, None, NonZeroUsize::new(5).unwrap());
+    it.read().unwrap();
+    it.read().unwrap();
+    it.read().unwrap(); // EOF
```
memark: This comment feels poorly placed.

Could we do this assertion like we do below?
>    it.read().unwrap();
    let eof = it.read().unwrap();

src/redisearch_rs/top_k/tests/integration/iterator.rs
```diff
@@ -0,0 +1,232 @@
...
+    let status = unsafe { it.revalidate(mock_ctx.spec()) }.unwrap();
+    assert_eq!(status, rqe_iterators::RQEValidateStatus::Aborted);
+}
+
+#[test]
+fn unfiltered_timeout_propagated() {
+    use rqe_iterators::RQEIteratorError;
+    use top_k::{ScoreSource, mock::MockScoreBatch};
+
+    struct TimingOutSource;
```
memark: I would put this `struct`+`impl` outside of the test function.

src/redisearch_rs/top_k/src/lib.rs
```diff
@@ -8,7 +8,32 @@
 */
 
 //! A fixed-capacity heap that retains the top-k scored documents.
```
memark: This heap comment is now in the wrong place.

src/redisearch_rs/top_k/src/traits.rs
```diff
@@ -0,0 +1,74 @@
...
+
+// ── ScoreBatch ────────────────────────────────────────────────────────────────
+
+/// A cursor over a single score-ordered batch of `(doc_id, score)` pairs.
+///
+/// Produced by [`ScoreSource::next_batch`] and streamed by [`TopKIterator`].
+/// Doc IDs within a batch must be **strictly increasing**.
+///
+/// [`TopKIterator`]: crate::TopKIterator
+pub trait ScoreBatch {
```
memark, line 25: This feels like an iterator. Maybe `ScoreBatchIterator`? Or `...Cursor` as you called it below.

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// Unfiltered path: yielding directly from `direct_batch` without a heap.
+    YieldingDirect,
+}
+
+// ── TopKIterator ──────────────────────────────────────────────────────────────
+
+/// A generic top-k iterator parameterized over a [`ScoreSource`].
+///
+/// Implements the execution mode described in the design doc:
+/// [`Unfiltered`](TopKMode::Unfiltered).
```
memark: I think we should be careful with linking to the design docs unless we know they will remain. Maybe just say "Implements the Unfiltered execution mode"?

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,272 @@
...
+            direct_batch: None,
+            k,
+            phase: Phase::NotStarted,
+            current: None,
+            last_doc_id: 0,
+            at_eof: false,
+        }
+    }
+
+    // ── Collection dispatch ───────────────────────────────────────────────────
```
memark: While I don't mind this comment/region layout per se, it feels like something we've invented here, and less of a convention in the code base?

memark: Especially when most regions only contain a single function it sometimes feels a bit superfluous.

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    /// Drive collection based on the current mode.
+    fn collect(&mut self) -> Result<(), RQEIteratorError> {
+        self.phase = Phase::Collecting;
+        let result = match self.mode {
+            TopKMode::Unfiltered => self.prepare_unfiltered_direct(),
+        };
+        if result.is_err() {
+            // Restore a stable phase so that a subsequent read() call can retry.
+            // Phase::Collecting is only valid transiently inside this method.
```
memark: I don't understand this comment (although it might ofc be correct). How does it relate to the code in this block?

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,272 @@
...
+        Ok(())
+    }
+
+    // ── Yielding helpers ─────────────────────────────────────────────────────
+
+    /// Yield the next result from the unfiltered direct batch.
+    fn advance_unfiltered_direct(
+        &mut self,
+    ) -> Result<Option<&mut RSIndexResult<'index>>, RQEIteratorError> {
+        let item = if let Some(batch) = &mut self.direct_batch {
```
memark: If we like `match` statements (we do below) I would prefer
```
let item = match &mut self.direct_batch {
    Some(batch) => batch.next(),
    None => None,
};
```
or possibly more idiomatic
`let item = self.direct_batch.as_mut().and_then(Iterator::next);`

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,272 @@
...
+            self.collect()?;
+        }
+
+        match self.phase {
+            Phase::YieldingDirect => self.advance_unfiltered_direct(),
+            Phase::Yielding => {
+                unreachable!("collect() must set phase to YieldingDirect or Yielding")
+            }
+            Phase::NotStarted | Phase::Collecting => {
+                unreachable!("collect() must set phase to YieldingDirect or Yielding")
```
memark, line 196: Now idea at all if this is a hot path, but we could consider `unreachable_unchecked!` if it was. Probably a later opt.

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,272 @@
...
+        _doc_id: t_docId,
+    ) -> Result<Option<SkipToOutcome<'_, 'index>>, RQEIteratorError> {
+        // TopKIterator is a root-only iterator that yields results sorted by
+        // score, not by doc_id.  It cannot be used as a child in a larger
+        // iterator tree, so skip_to is unsupported.
+        unimplemented!("TopKIterator is a root-only iterator; skip_to is not supported")
+    }
+
+    #[inline(always)]
+    unsafe fn revalidate(
```
memark: Given that this `fn` is `unsafe`, shouldn't it have safety docs?

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,272 @@
...
+        if let Some(child) = &mut self.child {
+            child.rewind();
+        }
+        self.mode = self.initial_mode;
+        self.direct_batch = None;
+        self.current = None;
+        self.last_doc_id = 0;
+        self.at_eof = false;
+        self.phase = Phase::NotStarted;
+        // We explicitly do NOT reset self.metrics because they're used for diagnostics.
```
memark: I don't think there is a `self.metrics` on this type?

src/redisearch_rs/top_k/src/lib.rs
```diff
@@ -8,7 +8,32 @@
...
+//! Generic top-k iterator shared by the vector (hybrid) and numeric (optimizer)
+//! query iterators.
+//!
+//! # Architecture
+//!
+//! The core abstraction is [`TopKIterator<S>`], a state machine that drives
+//! top-k collection in three modes:
+//!
+//! - **Unfiltered** — no child filter; stream results directly from the source's batch.
+//!   for each document.
```
memark: Was this sentence cut short?

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        if self.phase == Phase::NotStarted {
+            self.collect()?;
+        }
+
+        match self.phase {
+            Phase::YieldingDirect => self.advance_unfiltered_direct(),
+            Phase::Yielding => {
+                unreachable!("collect() must set phase to YieldingDirect or Yielding")
```
memark: We say it should be `Yielding`, but it is already `Yielding`.

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,272 @@
...
+    phase: Phase,
+    current: Option<RSIndexResult<'index>>,
+    last_doc_id: t_docId,
+    at_eof: bool,
+}
+
+impl<'index, S: ScoreSource<'index>> TopKIterator<'index, S> {
+    /// Create a new [`TopKIterator`].
+    ///
+    /// The execution mode is inferred from `child`.
```
memark: I would say "The execution mode is always `Unfiltered`."

I see the comment two lines down, but even though it's the only one, it'n _not_ inferred from `child`.

src/redisearch_rs/top_k/tests/integration/iterator.rs
```diff
@@ -0,0 +1,232 @@
...
+    let it = TopKIterator::new(source, None, NonZeroUsize::new(3).unwrap());
+    assert_eq!(it.num_estimated(), 3);
+}
+
+// ── Unfiltered path ───────────────────────────────────────────────────────────
+
+#[test]
+fn unfiltered_yields_batch_in_source_order() {
+    // Unfiltered mode streams directly from the batch without sorting.
+    let source = MockScoreSource::new(vec![vec![(3, 0.1), (1, 0.5), (2, 0.9)]]);
```
memark: This violates
>Doc IDs within a batch must be strictly increasing

Even though you mention it's not checked at runtime, I still think the tests should respect this.

src/redisearch_rs/top_k/src/mock.rs
```diff
@@ -0,0 +1,125 @@
...
+use inverted_index::RSIndexResult;
+use rqe_iterators::RQEIteratorError;
+
+use crate::traits::{ScoreBatch, ScoreSource};
+
+// ── MockScoreBatch ────────────────────────────────────────────────────────────
+
+/// A [`ScoreBatch`] backed by a pre-sorted `Vec<(t_docId, f64)>`.
+///
+/// Doc IDs must be strictly increasing; this is not validated at runtime.
```
memark: >this is not validated at runtime.

Or... Maybe it should be?! In Debug mode.

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    // ── Unfiltered path ───────────────────────────────────────────────────────
+
+    /// Set up the unfiltered direct-yield path.
+    ///
+    /// Calls [`ScoreSource::next_batch`] exactly once.  Results are streamed
+    /// directly from the batch cursor — no heap is involved.
+    fn prepare_unfiltered_direct(&mut self) -> Result<(), RQEIteratorError> {
+        self.direct_batch = self.source.next_batch()?;
+        if self.direct_batch.is_none() {
```
memark, line 140: What if the `ScoreSource` yields multiple batches? Maybe we just ignore it, maybe we should panic. Either way I think we should mention/document it.

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,272 @@
...
+
+        match item {
+            Some((doc_id, score)) => {
+                let result = self.source.build_result(doc_id, score);
+                self.current = Some(result);
+                self.last_doc_id = doc_id;
+                Ok(self.current.as_mut())
+            }
+            None => {
+                self.at_eof = true;
```
memark, line 166: Should we also reset `self.current` here? Or is it expected behavior to be able to get data from `current()` even with `at_eof=true`? Worth documenting.

src/redisearch_rs/top_k_bencher/benches/top_k.rs
```diff
@@ -0,0 +1,119 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! Micro-benchmarks for [`TopKHeap`] and [`TopKIterator`].
```
memark:
```suggestion
//! Micro-benchmarks for [`TopKHeap`].
```

src/redisearch_rs/top_k/tests/integration/iterator.rs
```diff
@@ -0,0 +1,232 @@
...
+    fn at_eof(&self) -> bool {
+        true
+    }
+
+    fn type_(&self) -> rqe_iterator_type::IteratorType {
+        rqe_iterator_type::IteratorType::Mock
+    }
+
+    fn intersection_sort_weight(&self, _: bool) -> f64 {
+        1.0
```
memark: Given that `skip_to` panics, maybe this is not the best return value? `0.0` instead? Or `NEG_INFINITY`?

memark (commented):
Looking good! The state machine is solid. The integration tests and benches are well-written.

Some comments about this and that. Some of them is probably related to the fact that this is the first PR of more to come.

In general, my opinion is that every PR should be self-contained. That means the code should be good and production ready-ish even if it's part of a series of PRs. Mostly because our dev world is agile, and we never know when projects are frozen, devs are changed etc.

ThierryBerger:
We can take a different approach with a longer lived feature branch for the whole top k topic, where I submit digestible sized PR against progressively ; then a final merge taking place when the whole thing is ready (or ready to be swapped)

memark:
> We can take a different approach with a longer lived feature branch for the whole top k topic, where I submit digestible sized PR against progressively ; then a final merge taking place when the whole thing is ready (or ready to be swapped)

I think that has another problem though, that we're missing out of CI altogether. Getting things into main should be a priority IMHO.

I'm not at all saying I mind stacked PRs. Just wanted to explain some of my comments.

src/redisearch_rs/top_k/tests/integration/iterator.rs
ThierryBerger: A bit of context:
- `intersection_sort_weight` exists to give information to intersection algorithm heuristics (how this iterator should be prioritized over others)
- the "default" implementation is to return `1.0` as a "neutral" multiplier (Now that I think about it, this may have been communicated in a clearer way). That's the implementation I went to then.
- Now, our top k iterators and always root iterator.
  - So: they will never be part of an intersection iterator
  - So: they will never need to check that `intersection_sort_weight`
  
I'd actually lean towards panic! there, unreachable! or debug_assert!(false).

src/redisearch_rs/top_k/src/iterator.rs
ThierryBerger: We're implementing a trait here, so the unsafe invariants are documented on trait method documentation.

for context:
- https://github.com/RediSearch/RediSearch/blob/76e678f2fc707ee43dc7b79feb1c210662b63f71/src/redisearch_rs/rqe_iterators/src/lib.rs#L135-L141

src/redisearch_rs/top_k/src/iterator.rs
ThierryBerger: I'm not sure it's worth premature optimization, let's punt it to later optimization work.

It may be worth exploring a type state pattern, too, but I'd like to keep close to design doc as much as possible before suggesting more ambitious changes.

src/redisearch_rs/top_k/src/iterator.rs
ThierryBerger: yep, forgot to remove it :) (I cut the scope of the PR a bit since I opened it)

src/redisearch_rs/top_k/src/mock.rs
ThierryBerger: it won't hurt :+1: ; We're discussing mocks here, so this will only impact tests (or benchmarks running in debug :neutral_face: )

src/redisearch_rs/top_k/src/traits.rs
ThierryBerger: I kept naming faithful to design doc, I'll defer yak-shaving decision to @GuyAv46

src/redisearch_rs/top_k/src/iterator.rs
ThierryBerger: good catch, I meant to say that yielding is dead code for now.

src/redisearch_rs/top_k/src/iterator.rs
ThierryBerger: this region concept exists in the codebase, but indeed used sparingly, I'll trim it down (tests make sense to group) ; I personally prefer to scope them in mods so the intent is clearer, and it's easier to keep in sync, but it's not the way it's done in the codebase.

src/redisearch_rs/top_k/src/iterator.rs
ThierryBerger: You're right to flag this! as we'll go further into implementation, this will lead to issues because a score source can be used in both modes.

As we currently support only unfiltered, I'll document this behaviour.
- In next PR #9424  I split it in 2 methods for that to avoid ambiguity, so this will be renamed `next_batch_unfiltered`. I guess we could name it that way for now... but that doesn't feel tremendously important as long as docs are important and there's a path forward.

The whole "batching" scenario is an overcomplication for current implementation anticipating on future work, but as the objective was to create the state machine I had to implement at least 1 strategy...

src/redisearch_rs/top_k/src/iterator.rs
ThierryBerger: phase collecting is set during the collection and should be reset after an error.

I think the Collecting phase is actually not useful, because as far as I understand we don't want an iterator to be stuck in this phase, so I'd argue to remove this phase. As it was part of the design, I'm curious about your thoughts @GuyAv46 ?

memark (approved):
Great work on my pile of comments!

ThierryBerger:
rebased for conflicts on a cargo.toml

memark (dismissed):

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,260 @@
...
+    fn prepare_unfiltered_direct(&mut self) -> Result<(), RQEIteratorError> {
+        self.direct_batch = self.source.next_batch()?;
+        if self.direct_batch.is_none() {
+            self.at_eof = true;
+        }
+        debug_assert!(
+            matches!(self.source.next_batch(), Ok(None)),
+            "ScoreSource did not return Ok(None) in TopKMode::Unfiltered \
+             (extra batch or error); use a batched mode instead"
+        );
```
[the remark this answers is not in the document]
ThierryBerger: indeed, it's unfortunate.

Well, in this particular scenario, we're rather testing the absence of such side effect, so I'd argue it's fine.

src/redisearch_rs/top_k/src/traits.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn num_estimated(&self) -> usize;
+
+    /// Rewind the source to its initial state so batch iteration can restart.
+    fn rewind(&mut self);
+
+    /// Build the [`RSIndexResult`] that the [`TopKIterator`] will yield for a
+    /// given `(doc_id, score)` pair.
+    ///
+    /// [`TopKIterator`]: crate::TopKIterator
+    fn build_result(&self, doc_id: t_docId, score: f64) -> RSIndexResult<'index>;
```
GuyAv46, line 74: Let's track if we need to expose this API this way, or whether we can save temporary allocations and return references similar to other iterators

src/redisearch_rs/top_k/src/iterator.rs
```diff
@@ -0,0 +1,260 @@
...
+    /// Yield the next result from the unfiltered direct batch.
+    fn advance_unfiltered_direct(
+        &mut self,
+    ) -> Result<Option<&mut RSIndexResult<'index>>, RQEIteratorError> {
+        let item = self.direct_batch.as_mut().and_then(S::Batch::next);
+
+        match item {
+            Some((doc_id, score)) => {
+                let result = self.source.build_result(doc_id, score);
+                self.current = Some(result);
```
GuyAv46, line 161: We should reuse `current` like other iterators do.
This can wait for a following PR, maybe we can align more parts like the heap results

src/redisearch_rs/top_k/src/traits.rs
ThierryBerger: You're spot on! this discussion is related to this approach I took: https://github.com/RediSearch/RediSearch/pull/9479.

So we may be able to get rid of those allocations, but we will need another way to fetch that rsindexresult back when needed (highlight, sort..), currently it's done through re-iterating via `skip_to`, which is not implemented for hybrid so we're re-iterating the whole iterator.

I created a ticket: MOD-8817

src/redisearch_rs/top_k/src/iterator.rs
ThierryBerger: Good catch, I created MOD-8817

GuyAv46 (dismissed):

memark (approved):

PR #9291, opened 2026-04-26
MOD-14805: COLLECT reducer: distributed coordinator support

Itzikvaknin:
This PR adds distributed execution support for `FT.AGGREGATE ... GROUPBY ... REDUCE COLLECT` in coordinator mode.

The distributed planner now splits the innermost `GROUPBY`'s `COLLECT` reducer into a shard-side remote reducer and a coordinator-side local reducer. The remote reducer collects ordinary item rows on each shard and serializes one payload per shard group. The local reducer consumes those shard-group payloads through planner metadata and rebuilds the client-facing `COLLECT` result.

Outer `GROUPBY` stages remain coordinator-local and continue to consume ordinary rows.

## What Changed

- Added `COLLECT` distribution logic that creates separate remote and local reducer invocations.
- Added the local shard-payload source into planner metadata via `PLN_Reducer.sourceAlias`, later resolved as `ReducerOptions::source_key`.
- Extended `ReducerOptions` with `is_local`, `source_key`, and `reqflags`, plus `ReducerOpts_IsInternal()` for detecting coordinator-dispatched shard requests.
- Split `COLLECT` parsing into local and remote paths:
  - remote path resolves `@field` names to `RLookupKey`s;
  - local path stores owned field/sort names because coordinator merge rows contain serialized shard payloads.
- Added Rust local/remote `COLLECT` reducer implementations and FFI constructors.
- Updated internal shard replies to include sort-key values when handling coordinator-dispatched requests, while keeping those internal keys out of final client-facing rows.
- Cleaned up distributed `COLLECT` argument construction with a returned `ArgsCursor` and a named count-buffer size.

## Tests

- Added C++ unit tests for distributed `COLLECT` argument construction.
- Added Rust integration tests for remote/local `COLLECT` reducer behavior.
- Expanded Python `test_groupby_collect.py` coverage for coordinator-mode distributed `COLLECT` scenarios.

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Adds new distributed-planning and reducer execution paths for `COLLECT`, including coordinator/shard split behavior and request-flag-dependent serialization, which can affect aggregation correctness in clustered and internal-dispatch scenarios.
> 
> **Overview**
> Enables distributed execution of `GROUPBY ... REDUCE COLLECT` by **splitting `COLLECT` into shard-side (remote) collection and coordinator-side (local) merge**, wiring the two halves together via planner metadata (`PLN_Reducer.inputAlias`) and new `ReducerOptions` fields (`is_local`, `input_key`, `reqflags`).
> 
> Updates `COLLECT` argument parsing and Rust implementation to support the two modes: remote reducers resolve `@` fields to `RLookupKey`s and (when internally dispatched) can **include extra `SORTBY` keys for coordinator merge**, while local reducers consume the shard payload and rebuild the client-facing array-of-maps.
> 
> Adds `dist_plan_utils` helpers for building distributed COLLECT arg vectors, updates pipeline construction to pass request flags and resolve planner-provided input keys, and extends C++/Rust/Python tests to cover coordinator merge, internal dispatch behavior, and new arg-building utilities.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4c5c35a0ce01b846ab0a55154d6fb1a8c6fb106. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/redisearch_rs/reducers/src/collect/shard.rs
```diff
@@ -0,0 +1,350 @@
...
+    /// projected fields appear in the output.
+    ///
+    /// The outer array has one element per collected row.
+    pub fn finalize(&mut self, r: &ShardCollectReducer) -> SharedValue {
+        let row_maps: Vec<SharedValue> = self
+            .field_values
+            .drain(..)
+            .zip(self.sort_values.drain(..))
+            .map(|(fv, sv)| {
+                let entries: Box<[_]> = if r.is_internal {
```
Itzikvaknin: Add a comment here why we include sort keys if is_intenral

src/redisearch_rs/reducers/src/collect/local.rs
```diff
@@ -0,0 +1,158 @@
...
+///
+/// Must remain `#[repr(C)]` with [`CollectCommon`] at offset 0 so the C layer
+/// can downcast this struct to `ffi::Reducer*` and read the vtable directly.
+#[repr(C)]
+pub struct LocalCollectReducer<'a> {
+    common: CollectCommon,
+    /// Lookup key for the per-shard `__SOURCE__` payload.
+    source_key: &'a RLookupKey<'a>,
+    field_names: Box<[Box<[u8]>]>,
+    sort_key_names: Box<[Box<[u8]>]>,
```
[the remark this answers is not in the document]
Itzikvaknin: `SORTBY` is not part of this PR

src/redisearch_rs/reducers/src/collect/remote.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /// Store projected field and sort values, filling missing values with nulls.
+    pub fn add(&mut self, r: &RemoteCollectReducer, row: &RLookupRow) {
+        let fv = r
+            .field_keys
+            .iter()
+            .map(|key| {
+                row.get(key)
+                    .cloned()
```
kei-nan, line 147: what is the cost of the cloned()?

Itzikvaknin:
```
impl Clone for SharedValue {
    fn clone(&self) -> Self {
        if self.is_null_static() {
            Self { ptr: self.ptr }
        } else {
            // SAFETY: `self.ptr` was obtained from `Arc::into_raw` and is not static (checked above).
            // The original `Arc` is reconstructed, cloned to increment the refcount,
            // then forgotten to avoid decrementing it.
            let arc: Arc<Value> = unsafe { Arc::from_raw(self.ptr) };
            let cloned = Arc::clone(&arc);
            mem::forget(arc);
            Self {
                ptr: Arc::into_raw(cloned),
            }
        }
    }
}
```
It does not allocating/parsing/ other heavy operation

src/coord/dist_plan_utils.h
```diff
@@ -0,0 +1,49 @@
...
+
+#pragma once
+
+#include "rmutil/args.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
```
kei-nan, line 35: maybe worth adding some documentation explaining what s remote and what is local.

Maybe the functions can return the object they are constructing?

src/coord/dist_plan.cpp
```diff
@@ -62,6 +63,7 @@ struct ReducerDistCtx {
     if (PLNGroupStep_AddReducer(gstp, name, cargs, status) != REDISMODULE_OK) {
       return false;
     }
+    array_tail(gstp->reducers).isLocal = (gstp == localGroup);
```
kei-nan: where is localGroup is coming from?

src/coord/dist_plan.cpp
```diff
@@ -332,6 +334,37 @@ static int distributeAvg(ReducerDistCtx *rdctx, QueryError *status) {
...
+/* Remote COLLECT emits array-of-maps; local COLLECT consumes it via __SOURCE__. */
+static int distributeCollect(ReducerDistCtx *rdctx, QueryError *status) {
+  const PLN_Reducer *src = rdctx->srcReducer;
+  size_t argc = src->args.argc;
+
+  // Build temporary args, then persist their object arrays with copyArgs.
+  char remoteCountBuf[16];
+  std::vector<void *> remoteObjs(argc + 1);
+  ArgsCursor remoteArgs;
+  buildRemoteCollectArgs(&remoteArgs, remoteObjs.data(), remoteCountBuf, &src->args);
```
kei-nan: Can't it return an ArgsCursor instead of getting it as a pointer?

src/coord/dist_plan_utils.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include "dist_plan_utils.h"
+#include "aggregate/reducers/collect.h"
+#include <stdio.h>
+#include <string.h>
+
+void buildRemoteCollectArgs(ArgsCursor *out, void **objs_buf, char *count_buf,
+                            const ArgsCursor *src_args) {
+  size_t argc = src_args->argc;
+  snprintf(count_buf, 16, "%zu", argc);
```
kei-nan: there is an assumption here that count buf length is 16.
Can we make this code safer?

src/coord/dist_plan.cpp
Itzikvaknin: Its an implicit `this->`

tests/cpptests/coord_tests/test_cpp_dist_plan_utils.cpp
```diff
@@ -0,0 +1,125 @@
...
+  assertArgs(out,
+             {"6", "FIELDS", "2", "@a", "@b", COLLECT_SOURCE_KEY, "__collect_ab", "AS",
+              "my_collect"});
+}
+
+TEST(DistPlanUtils, CoordCollectArgs_FieldsSortbyLimit) {
+  const char *src[] = {"FIELDS", "1", "@x", "SORTBY", "2", "@x", "ASC", "LIMIT", "0", "10"};
+  ArgsCursor srcArgs = makeArgs(src, 10);
+
+  char countBuf[16];
```
kei-nan: is this even used in the end?

src/redisearch_rs/reducers/src/collect/local.rs
```diff
@@ -0,0 +1,129 @@
...
+// `CollectCommon` must live at offset 0 so the C layer can downcast to
+// `ffi::Reducer`. Guard against accidental reordering of the struct fields.
+const _: () = assert!(core::mem::offset_of!(LocalCollectReducer<'_>, common) == 0);
+
+/// Per-group instance of [`LocalCollectReducer`].
+///
+/// Because `LocalCollectCtx` is arena-allocated ([`Bump`][bumpalo::Bump] does
+/// not run destructors), `ptr::drop_in_place` must be called to run
+/// destructors for the inner `Vec` and decrement `SharedValue` refcounts.
+pub struct LocalCollectCtx {
```
kei-nan, line 62: I think in general the rust classes would benefit from rust tests that will also show how they are used in the end of the day.

src/redisearch_rs/reducers/src/collect/local.rs
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! Local COLLECT reducer.
```
kei-nan, line 10: Can we improve this comments here?
Coming into this without a lot of context.
Not clear to me the whole local vs remote and what they mean.

src/coord/dist_plan_utils.h
Itzikvaknin: Done

src/coord/dist_plan_utils.cpp
Itzikvaknin: Changed to buf[24] as done in the codebase

tests/cpptests/coord_tests/test_cpp_dist_plan_utils.cpp
Itzikvaknin: What?

src/redisearch_rs/reducers/src/collect/local.rs
Itzikvaknin: Added in `src/redisearch_rs/reducers/tests/collect.rs`

src/redisearch_rs/reducers/src/collect/local.rs
Itzikvaknin: Tried my best, take a look

src/aggregate/reducers/collect.c
```diff
@@ -43,13 +49,57 @@ typedef struct {
...
+  // ArgParser validates nargs is within [1, COLLECT_MAX_FIELD_ARGS] before invoking this callback.
+  RS_ASSERT(count >= 1 && count <= COLLECT_MAX_FIELD_ARGS);
+
+  data->field_names = array_new(const char *, count);
+  for (int i = 0; i < count; i++) {
+    const char *s;
+    size_t len;
+    if (AC_GetString(ac, &s, &len, 0) != AC_OK) {
+      QueryError_SetError(opts->status, QUERY_ERROR_CODE_PARSE_ARGS,
+        "FIELDS (local) missing field name");
```
nafraf: I'm not sure if `(local)` term will be clear for the user.

src/aggregate/reducers/collect.c
```diff
@@ -85,7 +135,64 @@ static void handleCollectFields(ArgParser *parser, const void *value, void *user
...
+    array_append(data->sort_names, s + 1);
+    ac->offset++;
+
+    size_t dir_idx = array_len(data->sort_names) - 1;
+    handleCollectSortDirection(ac, &data->sortAscMap, dir_idx);
+  }
+
+  if (array_len(data->sort_names) == 0) {
+    QueryError_SetError(opts->status, QUERY_ERROR_CODE_PARSE_ARGS,
+      "SORTBY requires at least one sort field");
```
nafraf, line 202: But is `SORTBY 0` valid?
```sh
127.0.0.1:30002> FT.AGGREGATE idx * groupby 1 @n REDUCE COLLECT 3 FIELDS 1 @n SORTBY 0
1) (integer) 1
2) 1) "n"
   2) "1"
   3) "__generated_aliascollectfields,1,n"
   4) 1) 1) "n"
         2) "1"
```

src/aggregate/reducers/collect.c
Itzikvaknin: Changed in https://github.com/RediSearch/RediSearch/pull/9291/commits/701f44d8184ac771f201086d19d1d356352aebbd

src/aggregate/reducers/collect.c
Itzikvaknin:
```
> FT.AGGREGATE idx * groupby 1 @n REDUCE COLLECT 5 FIELDS 1 @n SORTBY 0
(error) SEARCH_PARSE_ARGS Bad arguments for COLLECT: SORTBY: Invalid argument count
```
In your example, `SORTBY 0` is not part of  `COLLECT` arguments

src/coord/dist_plan.cpp
Itzikvaknin: Done

src/aggregate/reducers/collect.c
```diff
@@ -22,12 +22,18 @@
...
+// because they cannot resolve fields through remote lookup tables.
 typedef struct {
   arrayof(const RLookupKey *) field_keys;
-  bool has_wildcard;
-
   arrayof(const RLookupKey *) sort_keys;
+  // Coord-mode names alias `options->args` and omit the leading `@`.
+  arrayof(const char *) field_names;
+  arrayof(const char *) sort_names;
+  const RLookupKey *source_key;
```
kei-nan: maybe add some documentation on the fields to give more context to the reader?

src/aggregate/reducers/collect.c
```diff
@@ -43,13 +49,57 @@ typedef struct {
...
+
+  // ArgParser validates nargs is within [1, COLLECT_MAX_FIELD_ARGS] before invoking this callback.
+  RS_ASSERT(count >= 1 && count <= COLLECT_MAX_FIELD_ARGS);
+
+  data->field_names = array_new(const char *, count);
+  for (int i = 0; i < count; i++) {
+    const char *s;
+    size_t len;
+    if (AC_GetString(ac, &s, &len, 0) != AC_OK) {
+      QueryError_SetError(opts->status, QUERY_ERROR_CODE_PARSE_ARGS,
```
kei-nan: I think the errors diverge a bit from the errors in: `ReducerOpts_GetKey` maybe worth unifying them somehow

src/aggregate/reducers/collect.c
```diff
@@ -85,7 +135,64 @@ static void handleCollectFields(ArgParser *parser, const void *value, void *user
...
+  CollectParseData *data = pctx->data;
+  const ReducerOptions *opts = pctx->options;
+  ArgsCursor *ac = (ArgsCursor *)value;
+
+  data->sort_names = array_new(const char *, 4);
+  data->sortAscMap = SORTASCMAP_INIT;
+
+  while (!AC_IsAtEnd(ac)) {
+    const char *s = AC_StringArg(ac, ac->offset);
+    if (!s) {
```
kei-nan: consider reducing code duplication of verifying if a char pointer is a valid field/key name.

src/aggregate/reducers/collect.c
```diff
@@ -85,7 +135,64 @@ static void handleCollectFields(ArgParser *parser, const void *value, void *user
...
+      QueryError_SetError(opts->status, QUERY_ERROR_CODE_PARSE_ARGS,
```
kei-nan: also not sure if this error is possible here, I think `AC_IsAtEnd` would have caught it, no?

src/aggregate/aggregate_plan.h
```diff
@@ -132,7 +132,9 @@ typedef struct {
   struct PLN_Reducer {
     const char *name;  // Name of function
     char *alias;       // Output key
+    char *sourceAlias; // Optional input key
```
kei-nan: is there allocation here, or can this be const?

src/aggregate/aggregate_plan.h
```diff
@@ -132,7 +132,9 @@ typedef struct {
...
     bool isHidden;     // If the output key is hidden. Used by the coordinator
+    bool isLocal;  // Whether this reducer runs locally (on the coordinator side)
```
kei-nan: comment indentation
```suggestion
    bool isLocal;       // Whether this reducer runs locally (on the coordinator side)
```

src/redisearch_rs/reducers/src/collect/common.rs
```diff
@@ -0,0 +1,37 @@
...
+impl CollectCommon {
+    pub fn new(sort_asc_map: u64, limit: Option<(u64, u64)>) -> Self {
+        Self {
+            reducer: Reducer::new(),
+            arena: Bump::new(),
+            sort_asc_map,
+            limit,
+        }
+    }
+}
```
daxhuiberts, line 41: This abstraction is not really worth it in my opinion. It doesn't contain any behavior, only a couple of fields and all those fields are accessible directly from other parts in this crate.

src/redisearch_rs/reducers/src/collect/local.rs
```diff
@@ -0,0 +1,136 @@
...
+    common: CollectCommon,
+    /// Lookup key for the per-remote payload.
+    source_key: &'a RLookupKey<'a>,
+    field_names: Box<[Box<[u8]>]>,
+    sort_key_names: Box<[Box<[u8]>]>,
+}
+
+// `CollectCommon` must live at offset 0 so the C layer can downcast to
+// `ffi::Reducer`. Guard against accidental reordering of the struct fields.
+const _: () = assert!(core::mem::offset_of!(LocalCollectReducer<'_>, common) == 0);
```
daxhuiberts: The primary goal is to check `Reducer` at offset `0`.

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect/local.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+    // SAFETY: caller guarantees the slice of pointers is valid.
+    let name_ptrs = unsafe { slice::from_raw_parts(names, len) };
+    name_ptrs
+        .iter()
+        .map(|&p| {
+            // SAFETY: caller guarantees each pointer references a valid
+            // NUL-terminated C string.
+            let bytes = unsafe { CStr::from_ptr(p) }.to_bytes();
+            bytes.to_vec().into_boxed_slice()
```
daxhuiberts, line 36: If they are `CString`s, it's better to save them as `CString`s as to reflect their intent/purpose.

src/redisearch_rs/reducers/src/collect/local.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub const fn new(_r: &LocalCollectReducer) -> Self {
+        Self { maps: Vec::new() }
+    }
+
+    /// Append maps from the remote payload; missing or malformed payloads are skipped.
+    pub fn add(&mut self, r: &LocalCollectReducer, row: &RLookupRow) {
+        let Some(payload) = row.get(r.source_key) else {
+            return;
+        };
+        if let Value::Array(array) = &**payload {
```
daxhuiberts: You can use an if let chain here: `if let Some(payload) ... && let Value::Array(array) ...`

src/redisearch_rs/reducers/src/collect/local.rs
```diff
@@ -0,0 +1,136 @@
...
+                            Value::Map(m) => m.get(name).cloned(),
+                            Value::Array(a) => a.map_get(name).cloned(),
+                            // Checked above; this arm is unreachable.
+                            _ => unreachable!(),
+                        }
+                        .unwrap_or_else(SharedValue::null_static);
+                        (SharedValue::new_string(name.to_vec()), val)
+                    })
+                    .collect();
+                Some(SharedValue::new(Value::Map(Map::new(row_entries))))
```
daxhuiberts: You can use the `SharedValue::new_map` (and `SharedValue::new_array` below) to simplify it a little bit (and in `remote.rs`).

Then you can skip the `collect()` here as well.

src/redisearch_rs/reducers/src/collect/local.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// Append maps from the remote payload; missing or malformed payloads are skipped.
+    pub fn add(&mut self, r: &LocalCollectReducer, row: &RLookupRow) {
+        let Some(payload) = row.get(r.source_key) else {
+            return;
+        };
+        if let Value::Array(array) = &**payload {
+            self.maps.reserve(array.len());
+            for entry in array.iter() {
+                self.maps.push(entry.clone());
+            }
```
daxhuiberts: You don't need a manual loo here. You can use `self.maps.extend(...)`.

src/redisearch_rs/reducers/src/collect/local.rs
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+
+    /// Rebuild remote rows as client-facing maps, accepting RESP3 maps and
+    /// RESP2 flat arrays while ignoring internal-only extra keys.
+    pub fn finalize(&mut self, r: &LocalCollectReducer) -> SharedValue {
+        let rebuilt: Vec<SharedValue> = std::mem::take(&mut self.maps)
```
daxhuiberts: Convention is to use `mem::take` with a `use std::mem` at the top.

src/redisearch_rs/reducers/src/collect/local.rs
```diff
@@ -0,0 +1,136 @@
...
+            .into_iter()
+            .filter_map(|entry| {
+                // Malformed remote payloads are skipped defensively.
+                let is_valid = matches!(&*entry, Value::Map(_) | Value::Array(_));
+                if !is_valid {
+                    return None;
+                }
+
```
daxhuiberts: Split this up in a separate `filter` and `map` like `filter(|e| matches!(&*e, Value::Map(_) | Value::Array(_))).map(|e| ...)`.

src/redisearch_rs/reducers/src/collect/remote.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Per-group instance of [`RemoteCollectReducer`].
+///
+/// Because `RemoteCollectCtx` is arena-allocated ([`Bump`][bumpalo::Bump] does
+/// not run destructors), `ptr::drop_in_place` must be called to run
+/// destructors for the inner `Vec`s and decrement `SharedValue` refcounts.
+pub struct RemoteCollectCtx {
+    field_values: Vec<Vec<SharedValue>>,
+    /// Kept row-aligned with `field_values`.
+    sort_values: Vec<Vec<SharedValue>>,
+}
```
daxhuiberts, line 66: This sounds more like this should be a `rows: Vec<RemoteCollectRow>` where `RemoteCollectRow` is a `struct RemoteCollectRow { field_values: Vec<SharedValues>, sort_values: Vec<SharedValues>` }`.

src/redisearch_rs/reducers/src/collect/remote.rs
```diff
@@ -0,0 +1,185 @@
...
+use crate::collect::common::CollectCommon;
+
+/// Remote COLLECT reducer.
+///
+/// Must remain `#[repr(C)]` with [`CollectCommon`] at offset 0 so the C layer
+/// can downcast this struct to `ffi::Reducer*` and read the vtable directly.
+#[repr(C)]
+pub struct RemoteCollectReducer<'a> {
+    common: CollectCommon,
+    pub(crate) field_keys: Box<[&'a RLookupKey<'a>]>,
```
daxhuiberts: I don't think this needs to be `pub(crate)` right?

src/redisearch_rs/reducers/src/collect/remote.rs
```diff
@@ -0,0 +1,185 @@
...
+    has_wildcard: bool,
+    /// Raw sort-key references, including keys not present in `FIELDS`.
+    pub(crate) sort_keys: Box<[&'a RLookupKey<'a>]>,
+    /// Internal shard replies include sort-key values for coordinator merge.
+    is_internal: bool,
```
daxhuiberts: `is_internal` is quite vague, perhaps something like `include_sort_keys`?

src/redisearch_rs/reducers/src/collect/remote.rs
```diff
@@ -0,0 +1,185 @@
...
+            })
+            .collect();
+        self.sort_values.push(sv);
+    }
+
+    /// Serialize rows as maps; internal shard replies also include sort keys.
+    pub fn finalize(&mut self, r: &RemoteCollectReducer) -> SharedValue {
+        let row_maps: Vec<SharedValue> = self
+            .field_values
+            .drain(..)
```
daxhuiberts: I previously (in the previous PR) suggested to use `drain()`, but now I have a better understanding of `finalize` and I think it's better to use `mem::take` instead.

src/redisearch_rs/reducers/tests/collect.rs
```diff
@@ -0,0 +1,207 @@
...
+    let row = fixture.row("apple", 10.0);
+
+    ctx.add(&reducer, &row);
+    let output = ctx.finalize(&reducer);
+    let rows = array_entries(&output);
+    assert_eq!(rows.len(), 1);
+
+    let row = map_entries(&rows[0]);
+    assert_string(
+        row.get(b"name").expect("name field should be present"),
```
daxhuiberts: In tests `unwrap()` is enough.

src/redisearch_rs/reducers/tests/collect.rs
```diff
@@ -0,0 +1,207 @@
...
+        b"apple",
```
daxhuiberts: Or check on `Some(b"apple")`.

src/redisearch_rs/reducers/tests/collect.rs
```diff
@@ -0,0 +1,207 @@
...
+    );
+    assert!(row.get(b"sweetness").is_none());
```
daxhuiberts, line 154: I think this test should verify that `sweetness` is present, right?

src/redisearch_rs/reducers/Cargo.toml
```diff
@@ -13,5 +16,13 @@ rlookup.workspace = true
...
 workspace_hack.workspace = true
 
+[build-dependencies]
+build_utils.workspace = true
+
+[dev-dependencies]
+reducers = { path = ".", features = ["unittest"] }
+redis_mock.workspace = true
+redisearch_rs = { path = "../c_entrypoint/redisearch_rs", features = ["mock_allocator"] }
+
```
daxhuiberts, line 26: I don't think the tests in this crate use any C dependencies and don't need this whole `build_utils` and `redis_mock`.

daxhuiberts (changes requested):
Do I understand correctly that `Local` is used on the primary redis instance and `Remote` is used on each redis shard? That's not quite descriptive in my opinion. Is that the nomenclature or can we come up with some more fitting names?

src/aggregate/aggregate_plan.h
kei-nan: instead of using source, consider using upstream.
e.g upstrreamAlias, upstream_key
another option is inputAlias, input_key,

I have a slight preference for upstream.

src/coord/dist_plan.cpp
```diff
@@ -332,6 +334,36 @@ static int distributeAvg(ReducerDistCtx *rdctx, QueryError *status) {
   return REDISMODULE_OK;
 }
 
+/* Remote COLLECT emits array-of-maps; local COLLECT consumes the remote alias. */
+static int distributeCollect(ReducerDistCtx *rdctx, QueryError *status) {
+  const PLN_Reducer *src = rdctx->srcReducer;
+  size_t argc = src->args.argc;
+
+  // Build temporary args, then persist their object arrays with copyArgs.
+  char remoteCountBuf[COLLECT_ARGS_COUNT_BUF_LEN];
```
kei-nan: consider using std::array and pass it via reference, not clear why you need this
but passing it it via array will also pass the size to thefunction.
another option is a vector.

src/coord/dist_plan.cpp
```diff
@@ -332,6 +334,36 @@ static int distributeAvg(ReducerDistCtx *rdctx, QueryError *status) {
...
+  rdctx->copyArgs(&remoteArgs);
+
+  const char *alias;
+  if (!rdctx->add(rdctx->remoteGroup, "COLLECT", &alias, status, &remoteArgs)) {
+    return REDISMODULE_ERR;
+  }
+
+  // Layout: [nargs, original_args..., AS, user_alias]
+  char localCountBuf[COLLECT_ARGS_COUNT_BUF_LEN];
+  std::vector<void *> localObjs(argc + 3);
```
kei-nan: why 3?

src/redisearch_rs/reducers/src/collect/remote.rs
Itzikvaknin: Its here only so we could test the `finalize` of `include_sort_key`, 
We intend to use `RemoteCollectRow { field_values: RLookupRow, sort_values: Vec<SharedValues> }` later.

src/redisearch_rs/reducers/tests/collect.rs
```diff
@@ -0,0 +1,207 @@
...
+        row.get(b"sweetness")
+            .and_then(|value| value.as_num())
+            .expect("sort field should be present"),
+        10.0
+    );
+}
+
+#[test]
+fn local_collect_projects_remote_maps_and_fills_missing_fields_with_null() {
+    let source_key = key(c"__source", 0);
```
kei-nan: is this correct? I thought source is not needed anymore

tests/cpptests/coord_tests/test_cpp_dist_plan_utils.cpp
```diff
@@ -0,0 +1,116 @@
...
+  for (size_t i = 0; i < 10; i++) {
+    ASSERT_EQ(out.objs[i + 1], srcArgs.objs[i])
+        << "objs[" << i + 1 << "] should point to src_args element " << i;
+  }
+}
+
+TEST(DistPlanUtils, CoordCollectArgs_EmptyOriginalArgs) {
+  ArgsCursor srcArgs = makeArgs(nullptr, 0);
+
+  char countBuf[COLLECT_ARGS_COUNT_BUF_LEN];
```
kei-nan: not a fan of this pattern, see if maybe you can use std::array instead or something similar so the function won't have assumptions about the size of the incoming buffer.

kei-nan (commented):
left comments

src/redisearch_rs/reducers/src/collect/local.rs
Itzikvaknin: Good catch - updated both files to chain through CollectCommon::reducer.

src/redisearch_rs/reducers/src/collect/local.rs
Itzikvaknin: Done

src/redisearch_rs/reducers/src/collect/local.rs
Itzikvaknin: Done

src/redisearch_rs/reducers/src/collect/local.rs
Itzikvaknin: Done - also dropped the explicit `reserve`

src/redisearch_rs/reducers/src/collect/local.rs
Itzikvaknin: Done

src/redisearch_rs/reducers/src/collect/local.rs
Itzikvaknin: Done, please take a look to verify correctness and style

src/redisearch_rs/reducers/src/collect/remote.rs
Itzikvaknin: Dropped it

src/redisearch_rs/reducers/src/collect/remote.rs
Itzikvaknin: Nice, done

src/redisearch_rs/reducers/src/collect/remote.rs
Itzikvaknin: Done

src/redisearch_rs/reducers/tests/collect.rs
Itzikvaknin: Done

src/redisearch_rs/reducers/tests/collect.rs
Itzikvaknin: Done

src/redisearch_rs/reducers/Cargo.toml
Itzikvaknin: Tried it - linker fails on `_RSDummyContext`, `_RedisModule_Free`, `_RedisModule_FreeString`, `_RedisModule_StringPtrLen` (from `value::String`/`RedisString::Drop` and `RedisString::as_ptr_len`, fired by any `SharedValue::new_string`).

src/redisearch_rs/reducers/tests/collect.rs
Itzikvaknin: No, sweetness is part of the row, but is not requested
```
        Box::new([
            b"name".to_vec().into_boxed_slice(),
            b"missing".to_vec().into_boxed_slice(),
        ]),
```

src/aggregate/aggregate_plan.h
Itzikvaknin: Changed to `inputAlias`, `input_key`

src/redisearch_rs/reducers/tests/collect.rs
Itzikvaknin: It is, changed to ```let input_key = key(c"generatedalias", 0);```

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect/local.rs
Itzikvaknin: `LocalCollectCtx::finalize` compares names against arbitrary RESP byte strings, so a `CString` round-trip costs `.to_bytes()` at every site. 
Happy to switch if you'd still prefer it.

src/redisearch_rs/reducers/src/collect/common.rs
Itzikvaknin: Ill keep it for now, dont want to delay the merge for that

src/aggregate/aggregate_plan.h
Itzikvaknin: `array_tail(rdctx->localGroup->reducers).inputAlias = rm_strdup(alias);`

tests/cpptests/coord_tests/test_cpp_dist_plan_utils.cpp
Itzikvaknin: Changed to `to_string`

src/coord/dist_plan.cpp
Itzikvaknin: Added this util: 

```
/**
 * Returns the number of object slots the caller must reserve in `objs_buf`
 * passed to `buildCollectArgs`.
 *
 * Layout reminder:
 *   - 1 slot for `nargs`
 *   - `argc` slots for the forwarded args
 *   - 2 slots for "AS" + user_alias (only when `has_alias` is true)
 */
static inline size_t collectObjsBufLen(size_t argc, bool has_alias) {
  return argc + 1 + (has_alias ? 2 : 0);
}

```

src/coord/dist_plan.cpp
Itzikvaknin: Used `to_string` instead

src/aggregate/reducers/collect.c
Itzikvaknin: Replaced with an assert

src/aggregate/reducers/collect.c
Itzikvaknin: Changed to match

src/aggregate/reducers/collect.c
Itzikvaknin: Added:
```
// Temporary storage for parsed COLLECT arguments. The data is handed off to
// Rust via the appropriate `CollectReducer_Create*` factory once parsing
// succeeds. Remote reducers keep `RLookupKey *`; local reducers keep raw names
// because they cannot resolve fields through remote lookup tables.
//
// Exactly one population pattern is used per parse, selected by
// `options->is_local`:
//   - is_local == true  : `field_names`, `sort_names`, `input_key` are set;
//                          `field_keys`, `sort_keys`, `has_wildcard` are not.
//   - is_local == false : `field_keys`, `sort_keys`, `has_wildcard` are set;
//                          `field_names`, `sort_names`, `input_key` are not.
// `sortAscMap` and the `limit_*` triple are shared across both modes.
```

src/redisearch_rs/reducers/src/collect/common.rs
daxhuiberts: Ok, I'll leave it up to you to decide, but I do urge you to revisit this later.

src/redisearch_rs/c_entrypoint/reducers_ffi/src/collect/local.rs
daxhuiberts: Since we don't use it for anything else it is fine to keep it like this. Just for information: `.to_bytes()` is practically free (it's just a type cast and an decrement on the len which will doubtly have any serious effect).

daxhuiberts (approved):

kei-nan (approved):

Issue #226, opened 2017-11-24
Tokenizer doesn't understand Traditional Chinese


kilianc:
I ran into this problem where I was normalizing Chinese strings both at indexing and search time to traditional Chinese using: https://github.com/Pleasurazy/chinese-conv.

Most of the queries started failing and after some investigation with @dvirsky it appears tokenization only works with simplified Chinese.

Since Traditional/Simplified is a 1:1 map it would be nice if the normalization was part of the tokenizer in the module itself, unless somebody that is a native Chinese speaker has a good reason not to.

🍻

dvirsky:
@mnunberg looks like it's a pretty simple preprocessing, pretty much akin to lowercasing. Let's discuss the best way to add it, I feel a better way to represent the text processing part of the indexer and tokenizer is needed.

mnunberg:
It seems the simplest place for the conversion to take place is in Friso itself, where the UTF-8 sequence is actually decoded.

mnunberg:
@kilianc Can you provide some sample words (and maybe documents) in simplified and traditional Chinese that I can use for testing?

dvirsky:
@mnunberg this SO question includes a couple of complete word examples https://chinese.stackexchange.com/questions/14082/comparison-of-tools-for-converting-traditional-to-simplified-chinese

mnunberg:
@kilianc does this work for you? https://github.com/RedisLabsModules/RediSearch/tree/cn-simptrad

mnunberg:
This should be fixed. Please reopen if this doesn't work.

Issue #225, opened 2017-11-23
Searching using field names with a dot is not possible


wwindcloud:
Let say I have created an index and add a document

```
FT.CREATE index SCHEMA dc.title TEXT
FT.ADD index 123123 1 FIELDS dc.title "A Title"
```

I would then search for that title like this

```
FT.SEARCH index @dc.title:A Title
```

Doesn't work, it said syntax error on **title**. Am I doing something wrong or there's something to escape I am not aware of? Thanks.

dvirsky:
ATM the only non alphanumeric character allowed in field names is `_`. However I don't see anything wrong with allowing other punctuation marks or allowing escaping unsupported ones. Do you see the need for things beyond dot, that are accepted in other databases? a comma doesn't make sense, maybe `:`?

wwindcloud:
IMHO, the choice of characters allowed in the field names should not be arbitrarily restricted without a good reason. I know `:` is used for key prefix separation, space for argument separation, but that's pretty much it. Actually, if we have proper escape support, I don't see why we can't have colons or spaces in field names, or better yet full unicode support, and even field names in complete binary.

Some backstory why I am using dot. Well I was trying to index from a database with dcterms, prime metadata, and stuff and they are all using a dot to indicate schema.element.qualifier, I know it's easy to do convertion from `.` to `_`, but that will take some work. It would be great if redis support such characters and many others natively.

And yeah, full unicode or even binary support would be better, you can't expect people to always use alphanumeric field names, especially in non-English environment. Thanks.

dvirsky:
I'll allow anything that goes into a token - unicode, ascii and **escaped** punctuation marks - as long as it doesn't create ambiguity.

dvirsky:
Fixed in eca6cb48ddd8a52469bc601f07881a4f5aed6eb0

PR #9293, opened 2026-04-26
MOD-14916 MOD-15104 Expiration perf improvements

ofiryanai:

Three perf improvements to the expiration feature, ported and adapted from the 8.4 investigation branches (originally redislabsdev PRs #225 and #226).

### 1. Redesign the TTL table for cache-friendly iterator lookups
Replace the dict-backed `TimeToLiveTable` with a direct-modulo bucket array whose collision chains are contiguous-vec blocks. docIds are allocated monotonically via `++maxDocId` in `DocTable`, so iterator-time `VerifyDocAndField*` lookups see them in ascending, mostly-sequential order; direct modulo keeps sequential docIds in sequential slots, letting the hardware prefetcher pull upcoming bucket headers into L1 ahead of demand. Primarily benefits the "miss" hot path - docs without TTL, the common case at query time. `DocTable_UpdateExpiration` now passes `t->maxSize` through to `TimeToLiveTable_VerifyInit` so the slot formula cannot drift from `DocTable`'s own modulus.

### 2. Gate per-posting TTL lookup on the presence of HFE entries
The TTL table now holds field-level (HEXPIRE) entries only and is destroyed by `DocTable_Pop` when the last one leaves the index, so a non-NULL `docs.ttl` is a sufficient and tight gate by itself. Iterators (`HybridIterator`, geometry `CPPQueryIterator`, Rust `FieldExpirationChecker`) bypass the TTL lookup entirely on indexes that have no field-level TTL right now, with no extra latch to maintain.

### 3. Inline doc-level expiration on DMD to bypass TTL lookup
Inline the doc-level expiration timestamp directly on the DMD as an `int64_t` nanoseconds-since-epoch (8 B per DMD, after `nextInChain` and before `payload` so `DocTable_Put`'s lean-allocation path is preserved). The new `DocTable_IsDocExpired` check is a single scalar compare on a cache line we are already touching (the caller has just read `dmd->flags`), eliminating the per-result table lookup on the result-processor hot path. The `Document_HasExpiration` flag is no longer set or read on the in-memory DMD path - the result processor fetches the DMD on every hit anyway, so a flag-gated branch buys nothing there.

1. MOD-14916

1. `TimeToLiveTable` (`src/ttl_table/`): direct-modulo bucket array; HFE-only payload
2. `DocTable` / `RSDocumentMetadata` (`src/doc_table.*`, `src/redisearch.h`): inlined `expirationTimeNs`; HFE gate via `docs.ttl`
3. Iterators (`src/iterators/hybrid_reader.c`, `src/geometry/query_iterator.cpp`, `src/redisearch_rs/rqe_iterators/src/expiration_checker.rs`): hoisted gate

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Refactors core expiration storage and lookup paths (DocTable metadata layout, TTL table implementation, and iterator gating), which could affect query-time expiration correctness and memory behavior despite added regression tests.
> 
> **Overview**
> Improves expiration query performance by **inlining document-level TTL** onto `RSDocumentMetadata` as `expirationTimeNs` and switching doc-expired checks to a direct scalar compare (removing reliance on `Document_HasExpiration`/TTL-table lookups on the hot path).
> 
> Redesigns `TimeToLiveTable` from a dict-backed map into a **direct-modulo bucket array with contiguous collision chains**, and narrows it to store **field-level (HEXPIRE) expirations only**; it is lazily created on first field-expiration and destroyed when empty.
> 
> Updates C++ geometry and hybrid iterators plus the Rust expiration checker to **hoist and revalidate** a `docs.ttl != NULL` gate to skip per-posting expiration checks when no field-level expirations exist, and removes TTL-table dict rehash pausing in read locks since it’s no longer a Redis dict. Adds targeted C++ and Python tests for TTL collisions/wraparound/lazy growth and adjusts memory-usage expectations for the larger DMD.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2f06cec160b22b50a4f70cd87600e67739e6111. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/doc_table.c
+++ b/src/doc_table.c
@@ -225,14 +225,19 @@ static inline int64_t expirationTimePointToNs(t_expirationTimePoint t) {
   return (int64_t)t.tv_sec * 1000000000LL + (int64_t)t.tv_nsec;
 }
 
+// Inlines the doc-level TTL on the DMD unconditionally so the result-processor
+// can drop the TTL-table lookup, and only routes field-level expirations into
+// the TTL table. This keeps the table strictly an HFE store, which lets
+// iterators use `t->ttl == NULL` as their per-spec gate. Takes ownership of
+// `sortedFieldWithExpiration` either by handing it to the table or freeing it
+// when there are no field-level entries to register.
 void DocTable_UpdateExpiration(DocTable *t, RSDocumentMetadata* dmd, t_expirationTimePoint ttl, arrayof(FieldExpiration) sortedFieldWithExpiration) {
-  if (hasExpirationTimeInformation(dmd->flags)) {
-    dmd->expirationTimeNs = expirationTimePointToNs(ttl);
+  dmd->expirationTimeNs = expirationTimePointToNs(ttl);
+  if (sortedFieldWithExpiration && array_len(sortedFieldWithExpiration) > 0) {
     TimeToLiveTable_VerifyInit(&t->ttl, t->maxSize);
-    TimeToLiveTable_Add(t->ttl, dmd->id, ttl, sortedFieldWithExpiration);
-    if (sortedFieldWithExpiration && array_len(sortedFieldWithExpiration) > 0) {
-      t->hasFieldExpiration = true;
-    }
+    TimeToLiveTable_Add(t->ttl, dmd->id, sortedFieldWithExpiration);
+  } else {
+    array_free(sortedFieldWithExpiration);
   }
 }
 
@@ -244,20 +249,13 @@ bool DocTable_IsDocExpired(DocTable* t, const RSDocumentMetadata* dmd, struct ti
   return dmd->expirationTimeNs <= now_ns;
 }
 
-static void clearExpirationFlagCallback(t_docId docId, void *ctx) {
-  DocTable *t = ctx;
-  RSDocumentMetadata *dmd = DocTable_GetOwn(t, docId);
-  if (dmd) {
-    dmd->flags &= ~Document_HasExpiration;
-    dmd->expirationTimeNs = 0;
-  }
-}
-
 void DocTable_ClearExpirationData(DocTable *t) {
-  if (t->ttl) {
-    TimeToLiveTable_ForEach(t->ttl, clearExpirationFlagCallback, t);
-    TimeToLiveTable_Destroy(&t->ttl);
-  }
+  // Walk every DMD: doc-level TTL lives inline on the DMD (not in the TTL
+  // table), and field-level TTL is only present for docs that are also in
+  // the table. Either may be set, so a single sweep over all DMDs is the
+  // simplest correct path.
+  DOCTABLE_FOREACH(t, dmd->expirationTimeNs = 0);
+  TimeToLiveTable_Destroy(&t->ttl);
 }
 
 /* Put a new document into the table, assign it an incremental id and store the metadata in the
@@ -381,7 +379,11 @@ RSDocumentMetadata *DocTable_Pop(DocTable *t, const char *s, size_t n) {
       return NULL;
     }
 
-    if (t->ttl && hasExpirationTimeInformation(md->flags)) {
+    // The TTL table holds field-level (HEXPIRE) entries only. Remove is a
+    // no-op if this doc never had one, and the IsEmpty check destroys the
+    // table once the last HFE doc leaves the index, restoring the iterator
+    // gate to its NULL "no HFE in this spec" state.
+    if (t->ttl) {
       TimeToLiveTable_Remove(t->ttl, md->id);
       if (TimeToLiveTable_IsEmpty(t->ttl)) {
         TimeToLiveTable_Destroy(&t->ttl);
--- a/src/doc_table.c
+++ b/src/doc_table.c
@@ -219,7 +219,7 @@ void DocTable_SetByteOffsets(RSDocumentMetadata *dmd, RSByteOffsets *v) {
 
 void DocTable_UpdateExpiration(DocTable *t, RSDocumentMetadata* dmd, t_expirationTimePoint ttl, arrayof(FieldExpiration) sortedFieldWithExpiration) {
   if (hasExpirationTimeInformation(dmd->flags)) {
-    TimeToLiveTable_VerifyInit(&t->ttl);
+    TimeToLiveTable_VerifyInit(&t->ttl, t->maxSize);
     TimeToLiveTable_Add(t->ttl, dmd->id, ttl, sortedFieldWithExpiration);
   }
 }
@@ -232,18 +232,17 @@ bool DocTable_IsDocExpired(DocTable* t, const RSDocumentMetadata* dmd, struct ti
   return TimeToLiveTable_HasDocExpired(t->ttl, dmd->id, expirationPoint);
 }
 
+static void clearExpirationFlagCallback(t_docId docId, void *ctx) {
+  DocTable *t = ctx;
+  RSDocumentMetadata *dmd = DocTable_GetOwn(t, docId);
+  if (dmd) {
+    dmd->flags &= ~Document_HasExpiration;
+  }
+}
+
 void DocTable_ClearExpirationData(DocTable *t) {
   if (t->ttl) {
-    dictIterator *ttlIter = dictGetIterator(t->ttl);
-    dictEntry *ttlEntry;
-    while ((ttlEntry = dictNext(ttlIter))) {
-      t_docId docId = (t_docId)dictGetKey(ttlEntry);
-      RSDocumentMetadata *dmd = DocTable_GetOwn(t, docId);
-      if (dmd) {
-        dmd->flags &= ~Document_HasExpiration;
-      }
-    }
-    dictReleaseIterator(ttlIter);
+    TimeToLiveTable_ForEach(t->ttl, clearExpirationFlagCallback, t);
     TimeToLiveTable_Destroy(&t->ttl);
   }
 }
--- a/src/doc_table.c
+++ b/src/doc_table.c
@@ -217,8 +217,17 @@ void DocTable_SetByteOffsets(RSDocumentMetadata *dmd, RSByteOffsets *v) {
   dmd->flags |= Document_HasOffsetVector;
 }
 
+// Pack a t_expirationTimePoint into nanoseconds since the epoch, preserving
+// the {0,0} "no expiration" sentinel as 0 so callers can use a single scalar
+// compare on the result-processor hot path.
+static inline int64_t expirationTimePointToNs(t_expirationTimePoint t) {
+  if (t.tv_sec == 0 && t.tv_nsec == 0) return 0;
+  return (int64_t)t.tv_sec * 1000000000LL + (int64_t)t.tv_nsec;
+}
+
 void DocTable_UpdateExpiration(DocTable *t, RSDocumentMetadata* dmd, t_expirationTimePoint ttl, arrayof(FieldExpiration) sortedFieldWithExpiration) {
   if (hasExpirationTimeInformation(dmd->flags)) {
+    dmd->expirationTimeNs = expirationTimePointToNs(ttl);
     TimeToLiveTable_VerifyInit(&t->ttl, t->maxSize);
     TimeToLiveTable_Add(t->ttl, dmd->id, ttl, sortedFieldWithExpiration);
     if (sortedFieldWithExpiration && array_len(sortedFieldWithExpiration) > 0) {
@@ -228,11 +237,11 @@ void DocTable_UpdateExpiration(DocTable *t, RSDocumentMetadata* dmd, t_expiratio
 }
 
 bool DocTable_IsDocExpired(DocTable* t, const RSDocumentMetadata* dmd, struct timespec* expirationPoint) {
-  if (!hasExpirationTimeInformation(dmd->flags)) {
-      return false;
+  if (dmd->expirationTimeNs == 0) {
+    return false;
   }
-  RS_LOG_ASSERT(t->ttl, "Document has expiration time information but no TTL table");
-  return TimeToLiveTable_HasDocExpired(t->ttl, dmd->id, expirationPoint);
+  const int64_t now_ns = (int64_t)expirationPoint->tv_sec * 1000000000LL + (int64_t)expirationPoint->tv_nsec;
+  return dmd->expirationTimeNs <= now_ns;
 }
 
 static void clearExpirationFlagCallback(t_docId docId, void *ctx) {
@@ -240,6 +249,7 @@ static void clearExpirationFlagCallback(t_docId docId, void *ctx) {
   RSDocumentMetadata *dmd = DocTable_GetOwn(t, docId);
   if (dmd) {
     dmd->flags &= ~Document_HasExpiration;
+    dmd->expirationTimeNs = 0;
   }
 }
 
--- a/src/doc_table.c
+++ b/src/doc_table.c
@@ -220,7 +220,19 @@ void DocTable_SetByteOffsets(RSDocumentMetadata *dmd, RSByteOffsets *v) {
 // Pack a t_expirationTimePoint into nanoseconds since the epoch, preserving
 // the {0,0} "no expiration" sentinel as 0 so callers can use a single scalar
 // compare on the result-processor hot path.
+//
+// `t_expirationTimePoint` is a POSIX `struct timespec` (IEEE Std 1003.1), which
+// is the OS-level resolution ceiling: `tv_nsec` is in [0, 999999999], and any
+// finer-grained clock would require a different type. `int64_t` of nanoseconds
+// covers ~292 years from the epoch (year 2262), far beyond any TTL Redis can
+// produce — `RM_GetAbsExpire` and `RM_HashFieldMinExpire` both return an
+// `mstime_t` (signed milliseconds since epoch), which we expand into a
+// `timespec` in `document_basic.c::timespecFromMilliseconds` before reaching
+// here. The debug assert traps any future caller that violates the timespec
+// invariant.
 static inline int64_t expirationTimePointToNs(t_expirationTimePoint t) {
+  RS_LOG_ASSERT(t.tv_nsec >= 0 && t.tv_nsec < 1000000000L,
+                "tv_nsec out of POSIX timespec range");
   return (int64_t)t.tv_sec * 1000000000LL + (int64_t)t.tv_nsec;
 }
 
@@ -244,8 +256,7 @@ bool DocTable_IsDocExpired(DocTable* t, const RSDocumentMetadata* dmd, struct ti
   if (dmd->expirationTimeNs == 0) {
     return false;
   }
-  const int64_t now_ns = (int64_t)expirationPoint->tv_sec * 1000000000LL + (int64_t)expirationPoint->tv_nsec;
-  return dmd->expirationTimeNs <= now_ns;
+  return dmd->expirationTimeNs <= expirationTimePointToNs(*expirationPoint);
 }
 
 void DocTable_ClearExpirationData(DocTable *t) {
--- a/src/doc_table.c
+++ b/src/doc_table.c
@@ -221,6 +221,9 @@ void DocTable_UpdateExpiration(DocTable *t, RSDocumentMetadata* dmd, t_expiratio
   if (hasExpirationTimeInformation(dmd->flags)) {
     TimeToLiveTable_VerifyInit(&t->ttl, t->maxSize);
     TimeToLiveTable_Add(t->ttl, dmd->id, ttl, sortedFieldWithExpiration);
+    if (sortedFieldWithExpiration && array_len(sortedFieldWithExpiration) > 0) {
+      t->hasFieldExpiration = true;
+    }
   }
 }
 
--- a/src/doc_table.c
+++ b/src/doc_table.c
@@ -221,7 +221,6 @@ void DocTable_SetByteOffsets(RSDocumentMetadata *dmd, RSByteOffsets *v) {
 // the {0,0} "no expiration" sentinel as 0 so callers can use a single scalar
 // compare on the result-processor hot path.
 static inline int64_t expirationTimePointToNs(t_expirationTimePoint t) {
-  if (t.tv_sec == 0 && t.tv_nsec == 0) return 0;
   return (int64_t)t.tv_sec * 1000000000LL + (int64_t)t.tv_nsec;
 }
 
@@ -233,7 +232,7 @@ static inline int64_t expirationTimePointToNs(t_expirationTimePoint t) {
 // when there are no field-level entries to register.
 void DocTable_UpdateExpiration(DocTable *t, RSDocumentMetadata* dmd, t_expirationTimePoint ttl, arrayof(FieldExpiration) sortedFieldWithExpiration) {
   dmd->expirationTimeNs = expirationTimePointToNs(ttl);
-  if (sortedFieldWithExpiration && array_len(sortedFieldWithExpiration) > 0) {
+  if (array_len(sortedFieldWithExpiration) > 0) {
     TimeToLiveTable_VerifyInit(&t->ttl, t->maxSize);
     TimeToLiveTable_Add(t->ttl, dmd->id, sortedFieldWithExpiration);
   } else {
--- a/src/doc_table.h
+++ b/src/doc_table.h
@@ -76,6 +76,10 @@ typedef struct {
   DMDChain *buckets;
   DocIdMap dim;             // Mapping between document name to internal id
   TimeToLiveTable* ttl;
+  // One-way latch: false until the first field-expiration (HEXPIRE) is written
+  // through DocTable_UpdateExpiration; never reset within the spec's lifetime.
+  // Lets iterators skip TTL lookups entirely when only doc-level EXPIRE is in use.
+  bool hasFieldExpiration;
 } DocTable;
 
 #define DOCTABLE_FOREACH(dt, code)                                           \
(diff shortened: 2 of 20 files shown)


src/iterators/hybrid_reader.c
```diff
@@ -473,9 +474,10 @@ static IteratorStatus HR_ReadKnnUnsortedSingle(HybridIterator *hr) {
     return ITERATOR_EOF;
   }
 
-  const t_fieldIndex fieldIndex = hr->filterCtx.field.index;
-  if (hr->sctx && fieldIndex != RS_INVALID_FIELD_INDEX
```
kei-nan, line 477: Maybe we can optimize it further if we make two versions of the function and select the correct when constructing HybridIterator?
Then we won't need the additional boolean.

src/redisearch_rs/rqe_iterators/src/expiration_checker.rs
```diff
@@ -93,8 +93,10 @@ impl ExpirationChecker for FieldExpirationChecker {
         // SAFETY: Guaranteed by the safety contract of `new`.
         let spec = unsafe { *(sctx.spec) };
 
-        // Check if TTL is configured and field expiration monitoring is enabled
-        if spec.docs.ttl.is_null() || !spec.monitorFieldExpiration {
+        // Check if TTL is configured and field expiration monitoring is enabled.
+        // The `hasFieldExpiration` latch lets us skip the per-posting TTL lookup
+        // entirely when only doc-level EXPIRE has ever been observed on this spec.
+        if spec.docs.ttl.is_null() || !spec.docs.hasFieldExpiration || !spec.monitorFieldExpiration {
```
kei-nan: do we sill need the ttl pointer check?
if you are planning to move the doc expiration into the dmd then maybe the null check is enough and we don't need the hasFieldExpiration boolean.

src/ttl_table/ttl_table.c
```diff
@@ -13,57 +13,205 @@
...
+//    it only extends the tail of the bucket array. Reads for docIds whose
+//    slot lies in the still-unallocated tail treat the entry as absent,
+//    which is correct because Add always grows `cap` to cover the slot
+//    before writing. This keeps the per-index footprint proportional to the
+//    number of TTL docs ever inserted rather than to maxDocTableSize, which
+//    matters in deployments with many indexes that sparsely use TTL.
+
 typedef struct {
+  t_docId docId;                        // 0 is reserved / unused
   t_expirationTimePoint documentExpirationPoint;
```
kei-nan: If you moved it to the DMD then this member can be removed. no?

src/redisearch.h
```diff
@@ -132,6 +132,13 @@ typedef struct RSDocumentMetadata_s {
   struct RSByteOffsets *byteOffsets;
   struct RSDocumentMetadata_s *nextInChain;
 
+  /* Document-level expiration time in nanoseconds since the epoch.
+   * Zero means no doc-level TTL is set. Inlined here to avoid a per-result
+   * TTL-table lookup in DocTable_IsDocExpired on the result-processor hot path.
+   * Placed before `payload` so the lean-allocation in DocTable_Put (which omits
+   * the trailing pointer when no payload is set) is preserved. */
+  int64_t expirationTimeNs;
```
kei-nan:
```suggestion
  t_expirationTimePoint documentExpirationPoint;
```
consider moving the member to be before the chain pointer.
maybe after the ref count.

src/geometry/query_iterator.hpp
```diff
@@ -27,6 +27,9 @@ struct CPPQueryIterator {
   const RedisSearchCtx *sctx_;
   const FieldFilterContext filterCtx_;
   const uint32_t initTimeoutCounter_;  // Value to reset counter to on each read() call
+  // Hoisted at construction: true iff we must run the per-posting
+  // field-expiration check. See should_check_field_expiration().
+  bool check_field_expiration_;
```
kei-nan, line 31: same here, maybe we can have two versions of the function and select the correct one during construction?

src/doc_table.c
```diff
@@ -217,33 +217,45 @@ void DocTable_SetByteOffsets(RSDocumentMetadata *dmd, RSByteOffsets *v) {
   dmd->flags |= Document_HasOffsetVector;
 }
 
+// Pack a t_expirationTimePoint into nanoseconds since the epoch, preserving
+// the {0,0} "no expiration" sentinel as 0 so callers can use a single scalar
+// compare on the result-processor hot path.
+static inline int64_t expirationTimePointToNs(t_expirationTimePoint t) {
```
kei-nan, line 233: I would prefer if we kept using `t_expirationTimePoint`, i.e `timespec`,
I think it is clearer and easier to maintain and reason about.

src/doc_table.c
```diff
@@ -217,33 +217,45 @@ void DocTable_SetByteOffsets(RSDocumentMetadata *dmd, RSByteOffsets *v) {
...
+    if (sortedFieldWithExpiration && array_len(sortedFieldWithExpiration) > 0) {
+      t->hasFieldExpiration = true;
+    }
   }
 }
 
 bool DocTable_IsDocExpired(DocTable* t, const RSDocumentMetadata* dmd, struct timespec* expirationPoint) {
-  if (!hasExpirationTimeInformation(dmd->flags)) {
-      return false;
+  if (dmd->expirationTimeNs == 0) {
```
kei-nan, line 256: shouldn't we also still check the dmd flags?

src/doc_table.h
```diff
@@ -76,6 +76,10 @@ typedef struct {
   DMDChain *buckets;
   DocIdMap dim;             // Mapping between document name to internal id
   TimeToLiveTable* ttl;
+  // One-way latch: false until the first field-expiration (HEXPIRE) is written
+  // through DocTable_UpdateExpiration; never reset within the spec's lifetime.
+  // Lets iterators skip TTL lookups entirely when only doc-level EXPIRE is in use.
+  bool hasFieldExpiration;
```
kei-nan: should this be outputted in the spec info?

tests/pytests/test_expire.py
```diff
@@ -683,3 +683,35 @@ def test_background_index_no_lazy_expiration_json(env):
...
+
+    # Docs 1..odd are long-lived; docs 1..even get a fast HPEXPIRE so they
+    # will be reported as expired at query time.
+    N = 64  # > 16x the bucket cap => many collisions per slot
+    for d in range(1, N + 1):
+        env.expect('HSET', f'doc:{d}', 'n', str(d), 't', 'tag').equal(2)
+        if d % 2 == 0:
+            env.expect('HPEXPIRE', f'doc:{d}', '1', 'FIELDS', '2', 'n', 't').equal([1, 1])
+
+    time.sleep(0.050)
```
kei-nan, line 706: why do we sleep here?

src/iterators/hybrid_reader.c
```diff
@@ -627,6 +629,12 @@ QueryIterator *NewHybridVectorIterator(HybridIteratorParams hParams, QueryError
   hi->runtimeParams.timeoutCtx = &hi->timeoutCtx;
   hi->sctx = hParams.sctx;
   hi->filterCtx = *hParams.filterCtx;
+  // Hoist the per-posting field-expiration gate: sctx, fieldIndex and the spec
+  // TTL/HFE state are all iterator-invariant, so we snapshot the AND once here.
+  hi->checkFieldExpiration =
```
GuyAv46, line 639: Can't the expiration requirement change, if we release and re-take the spec lock while executing?
Can we re/set it from the revalidation logic?

tests/pytests/test_expire.py
```diff
@@ -683,3 +683,35 @@ def test_background_index_no_lazy_expiration_json(env):
     # Accessing doc:1 directly should cause lazy expire and its removal from the DB.
     env.expect('JSON.GET', 'doc:1', "$").equal(None)
     env.expect('DBSIZE').equal(1)
+
+
+@skip(cluster=True, redis_less_than='7.4')
+def test_ttl_table_collision_chain(env: Env):
```
kei-nan: If you are already creating an env inside, then the function shouldn't get an env as a parameter.
At least that is my understanding.
```suggestion
def test_ttl_table_collision_chain():
```

kei-nan (commented):
left some comments
consider also adding a cpp microbenchmark for the ttl table.

src/iterators/hybrid_reader.h
```diff
@@ -54,6 +54,8 @@ typedef struct {
   bool canTrimDeepResults;         // Ignore the document scores, only vector score matters. No need to deep copy the results from the child iterator.
   TimeoutCtx timeoutCtx;           // Timeout parameters
   FieldFilterContext filterCtx;
+  bool checkFieldExpiration;       // Hoisted at construction: true iff we must run the per-posting
+                                   // field-expiration check. See NewHybridVectorIterator.
```
GuyAv46: Consider moving this new flag up, so the struct size isn't affected. We have a few paddings we can use (next to `canTrimDeepResults`, for example)

src/geometry/query_iterator.hpp
GuyAv46, line 31: Let's refresh this value when revalidating (after R/W lock reacquire), in case the spec state has changed

src/doc_table.c
```diff
@@ -217,33 +217,45 @@ void DocTable_SetByteOffsets(RSDocumentMetadata *dmd, RSByteOffsets *v) {
   dmd->flags |= Document_HasOffsetVector;
 }
 
+// Pack a t_expirationTimePoint into nanoseconds since the epoch, preserving
+// the {0,0} "no expiration" sentinel as 0 so callers can use a single scalar
+// compare on the result-processor hot path.
+static inline int64_t expirationTimePointToNs(t_expirationTimePoint t) {
+  if (t.tv_sec == 0 && t.tv_nsec == 0) return 0;
+  return (int64_t)t.tv_sec * 1000000000LL + (int64_t)t.tv_nsec;
```
GuyAv46, line 236: Consider avoiding the branch, as it should yield the same result as the calculation

src/doc_table.c
```diff
@@ -217,33 +217,45 @@ void DocTable_SetByteOffsets(RSDocumentMetadata *dmd, RSByteOffsets *v) {
...
+}
+
 void DocTable_UpdateExpiration(DocTable *t, RSDocumentMetadata* dmd, t_expirationTimePoint ttl, arrayof(FieldExpiration) sortedFieldWithExpiration) {
   if (hasExpirationTimeInformation(dmd->flags)) {
-    TimeToLiveTable_VerifyInit(&t->ttl);
+    dmd->expirationTimeNs = expirationTimePointToNs(ttl);
+    TimeToLiveTable_VerifyInit(&t->ttl, t->maxSize);
     TimeToLiveTable_Add(t->ttl, dmd->id, ttl, sortedFieldWithExpiration);
+    if (sortedFieldWithExpiration && array_len(sortedFieldWithExpiration) > 0) {
```
GuyAv46: `array_len` handles `NULL`
```suggestion
    if (array_len(sortedFieldWithExpiration) > 0) {
```

src/ttl_table/ttl_table.c
```diff
@@ -13,57 +13,205 @@
...
-static void destructor_TimeToLiveEntry(void *privdata, void *val) {
-    TimeToLiveEntry* entry = (TimeToLiveEntry*)val;
-    array_free(entry->fieldExpirations);
-    rm_free(entry);
-}
+typedef struct {
+  TimeToLiveEntry *entries;             // rm_malloc'd block of `count`, NULL when count == 0
+  uint32_t count;
+  uint32_t _pad;                        // reserved for future per-bucket stats
+} TTLBucket;
```
GuyAv46: Why not use an array? putting the size overhead on the entries allocation, with no overhead when not allocated

src/ttl_table/ttl_table.c
```diff
@@ -13,57 +13,194 @@
...
+// maxSize, with TTL_BUCKET_INITIAL_CAP as the first-grow seed.
+static void ttl_grow(TimeToLiveTable *t, size_t slot) {
+  if (slot < t->cap) return;
+  RS_ASSERT(t->cap < t->maxSize);  // slot is always < maxSize, so room must exist
+  const size_t oldcap = t->cap;
+  size_t newcap = t->cap == 0
+      ? TTL_BUCKET_INITIAL_CAP
+      : t->cap + 1 + MIN(t->cap / 2, TTL_BUCKET_MAX_GROW_STEP);
+  if (newcap > t->maxSize) newcap = t->maxSize;
+  if (newcap < slot + 1) newcap = slot + 1;
```
kei-nan, line 94: Maybe add an assert that slot + 1 < t->maxSize?

src/ttl_table/ttl_table.c
```diff
@@ -13,57 +13,194 @@
...
+// slot >= cap as "not present" and never grow. Growth curve matches
+// DocTable_Set: +1.5x geometric up to TTL_BUCKET_MAX_GROW_STEP, clamped to
+// maxSize, with TTL_BUCKET_INITIAL_CAP as the first-grow seed.
+static void ttl_grow(TimeToLiveTable *t, size_t slot) {
+  if (slot < t->cap) return;
+  RS_ASSERT(t->cap < t->maxSize);  // slot is always < maxSize, so room must exist
+  const size_t oldcap = t->cap;
+  size_t newcap = t->cap == 0
+      ? TTL_BUCKET_INITIAL_CAP
+      : t->cap + 1 + MIN(t->cap / 2, TTL_BUCKET_MAX_GROW_STEP);
```
kei-nan, line 92: it looks like the cap can end up being an odd number, but from the numbers used it seems like great effort went into choosing even numbers.
Are we okay with odd numbers for the capacity?

src/doc_table.c
```diff
@@ -217,35 +217,44 @@ void DocTable_SetByteOffsets(RSDocumentMetadata *dmd, RSByteOffsets *v) {
...
-      }
-    }
-    dictReleaseIterator(ttlIter);
-    TimeToLiveTable_Destroy(&t->ttl);
-  }
+  // Walk every DMD: doc-level TTL lives inline on the DMD (not in the TTL
+  // table), and field-level TTL is only present for docs that are also in
+  // the table. Either may be set, so a single sweep over all DMDs is the
+  // simplest correct path.
+  DOCTABLE_FOREACH(t, dmd->expirationTimeNs = 0);
```
kei-nan, line 267: The cost of clearing the expiration data just went up.
Now we need to iterate over all of the documents.
Looks like it is only called in config setter but wondering how long it can take now.

src/indexer.c
```diff
@@ -281,7 +281,10 @@ static void doAssignIds(RSAddDocumentCtx *cur, RedisSearchCtx *ctx) {
       Document* doc = cur->doc;
       const bool hasExpiration = doc->docExpirationTime.tv_sec || doc->docExpirationTime.tv_nsec || doc->fieldExpirations;
       if (hasExpiration) {
-        md->flags |= Document_HasExpiration;
```
kei-nan, line 284: Maybe remove `Document_HasExpiration` from the enum if it is no longer used?

src/iterators/hybrid_reader.c
ofiryanai: I prefer not in this PR

src/redisearch_rs/rqe_iterators/src/expiration_checker.rs
ofiryanai: removed as we spoke 👍

src/ttl_table/ttl_table.c
ofiryanai: yes, removed

src/redisearch.h
ofiryanai: moved 👍

src/doc_table.c
ofiryanai: `timespec` is 16B with the lower 8B never used (no nanos precision), and here I inlined so it will be clear. Don't think its worth extra 8B per dmd or that increase readabilty

src/doc_table.c
ofiryanai: `expirationTimeNs == 0` covers both "no doc-level TTL" and "TTL not yet set" without needing the flag

src/geometry/query_iterator.hpp
ofiryanai: added to revalidate 👍

src/ttl_table/ttl_table.c
ofiryanai: didn't think about that, changed to array

src/doc_table.c
ofiryanai: 👍

src/doc_table.c
ofiryanai: done

tests/pytests/test_expire.py
ofiryanai: so the even docs will be expired, to cover both `hit + expired` + `miss`

src/doc_table.h
ofiryanai: this flag is removed as we spoke, and I prefer it to be in separate ticket/PR

src/geometry/query_iterator.hpp
ofiryanai: not in this PR, scope is large already

tests/pytests/test_expire.py
ofiryanai: removed

src/iterators/hybrid_reader.c
ofiryanai: yes updated the revalidate to check that

src/iterators/hybrid_reader.h
ofiryanai: moved 👍

src/ttl_table/ttl_table.c
ofiryanai: yes, no meaningful benefit to make it even. and its same in `DocTable_Set`

src/doc_table.c
ofiryanai: well, for LTK-like where all keys have expiration its the same. I mean no change in worst case behabvior

src/indexer.c
ofiryanai: it's used by Disk API, not sure how and I prefer not to break sync right now

src/ttl_table/ttl_table.c
ofiryanai: adding

src/redisearch_rs/rqe_iterators/src/expiration_checker.rs
```diff
@@ -93,8 +93,11 @@ impl ExpirationChecker for FieldExpirationChecker {
         // SAFETY: Guaranteed by the safety contract of `new`.
         let spec = unsafe { *(sctx.spec) };
 
-        // Check if TTL is configured and field expiration monitoring is enabled
-        if spec.docs.ttl.is_null() || !spec.monitorFieldExpiration {
+        // The TTL table holds field-level (HEXPIRE) entries only and is
+        // destroyed once the last one leaves the index, so a NULL `ttl`
+        // pointer is a sufficient and tight gate by itself: it means no doc
+        // in this spec currently has a field-level expiration.
+        if spec.docs.ttl.is_null() {
```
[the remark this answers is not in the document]
ofiryanai: added HFE Gate to rust

src/geometry/query_iterator.cpp
```diff
@@ -105,6 +119,12 @@ std::size_t QIter_NumEstimated(const QueryIterator *ctx) {
 void QIter_Rewind(QueryIterator *ctx) {
   reinterpret_cast<CPPQueryIterator *>(ctx)->rewind();
 }
+ValidateStatus QIter_Revalidate(QueryIterator *ctx) {
+  auto *qi = reinterpret_cast<CPPQueryIterator *>(ctx);
+  qi->check_field_expiration_ =
```
kei-nan, line 124: What can change in revalidate?
do we expect sctx_ or filterCtx_ to change?

src/iterators/hybrid_reader.c
kei-nan: not clear to me how or what can change,
can sctx become null?

src/ttl_table/ttl_table.c
```diff
@@ -13,57 +13,190 @@
...
+
+struct TimeToLiveTable {
+  TTLBucket *buckets;                   // allocated for the first `cap` slots
+  size_t cap;                           // number of bucket slots physically allocated
+  size_t maxSize;                       // modulus for slot calculation, fixed at init
+  size_t count;                         // total live entries
+};
+
+static inline size_t ttl_slot(const TimeToLiveTable *t, t_docId docId) {
+  return docId < t->maxSize ? (size_t)docId : (size_t)(docId % t->maxSize);
```
kei-nan, line 68: maybe it will be simpler to always do docId % t->maxSize and get rid of a branching?

src/doc_table.c
kei-nan: I'm worried about this line, did you take into consideration all the possible commands we have for setting a key expiration?
I know the resolution right now is in milliseconds, but if it changes under our feed to be in nano we could get nasty bugs and it will be hard to track the bug down.

src/doc_table.c
```diff
@@ -217,35 +217,44 @@ void DocTable_SetByteOffsets(RSDocumentMetadata *dmd, RSByteOffsets *v) {
...
 
 bool DocTable_IsDocExpired(DocTable* t, const RSDocumentMetadata* dmd, struct timespec* expirationPoint) {
-  if (!hasExpirationTimeInformation(dmd->flags)) {
-      return false;
+  if (dmd->expirationTimeNs == 0) {
+    return false;
   }
-  RS_LOG_ASSERT(t->ttl, "Document has expiration time information but no TTL table");
-  return TimeToLiveTable_HasDocExpired(t->ttl, dmd->id, expirationPoint);
+  const int64_t now_ns = (int64_t)expirationPoint->tv_sec * 1000000000LL + (int64_t)expirationPoint->tv_nsec;
```
kei-nan: I'm not a fan of raw multiplication in the code.
At the very least consider having a helper function to do the conversion so we have something to document.

src/doc_table.c
kei-nan: the worst-case scenario is you have no document with expiration.
previously we iterated over zero documents, in the new code you would iterate over all the documents.

src/redisearch.h
```diff
@@ -127,12 +127,16 @@ typedef struct RSDocumentMetadata_s {
 
   uint16_t ref_count;
 
+  /* Document-level expiration time in nanoseconds since the epoch (0 = none).
+   * Inlined to avoid a TTL-table lookup on the result-processor hot path. */
+  int64_t expirationTimeNs;
```
kei-nan, line 132: So, what is the risk here in regard to compacting from a timespec?
How can the timespec conversion cause an issue here?
Maybe we can change the expiration type to be `timeval` and maybe take a more type safe conversion?

tests/pytests/test_expire.py
kei-nan: But shouldn't FT.SEARCH handle that?
That is the purpose of the entire expiration feature.
How do you know the key is expired vs search simply not returning it because it sees it expired at the iterator level?

src/doc_table.c
kei-nan: Maybe it would be better to change `t_expirationTimePoint` from `timespec` into `timeval` if we are so sure.

kei-nan (commented):
left comments

src/doc_table.c
ofiryanai: AI claims that according to POSIX IEEE 1003.1, `tv_nsec` of `timespec` is mandated to be in `[0,  999999999]`. `int64_t` of nanos covers up to year 2262, feels good enough for me (it said more but lets stop at that)

src/redisearch.h
ofiryanai: `timespec → int64_t` ns is lossless for any TTL before year 2262.  Switching to `timeval` still will require a multiply of *1e6 instead of *1e9. also we get a `timespec` from `clock_gettime(CLOCK_REALTIME)`, we will need to convert everywhere

src/doc_table.c
ofiryanai: Done - DocTable_IsDocExpired now goes through `expirationTimePointToNs`.

src/iterators/hybrid_reader.c
ofiryanai: the spec `rwlock` is released and re-acquired around iterator yields, sctx->spec->docs.ttl can become null. or not null if expiration was set between calls

src/ttl_table/ttl_table.c
ofiryanai: > 64-bit MOD is expensive; predicted branches are free

AI thinks this is a micro optimization, I'm convinced. Hoever its not an important one, if you feel strongly about the simplification I can remove

src/geometry/query_iterator.cpp
ofiryanai: `sctx->spec->docs.ttl` might change if expiration is met and the table is now gone, or the other way around if expiration is set during this yield

src/ttl_table/ttl_table.c
GuyAv46: We have the same thing on the doc table, I think @ofiryanai os right

src/redisearch.h
kei-nan: you would only need a function that compares a timespec against a timeval.
The multiplication is only needed here because you are converting a timespec into an int64.

I don't want to add too much content to the PR. but the fact we are not using a consistent type across the board is problematic IMHO.
@GuyAv46 WDYT?

kei-nan (approved):

src/redisearch.h
GuyAv46: We should have a consistent way to hold expirations, so we don’t need to convert

PR #8774, opened 2026-03-19
MOD-14268: Fix coordinator deadlock under mixed `FT.SEARCH + FT.AGGREGATE` load

nafraf:
## Description

1. Current:
`FT.SEARCH` was releasing its RQ pending slot too late: only after result reduction, client unblock, and blocked-client cleanup. In contrast, `FT.AGGREGATE` releases its RQ slot as soon as the libuv/network phase completes. Under mixed `FT.SEARCH` + `FT.AGGREGATE` load, this created a circular dependency:

    - `FT.AGGREGATE` could occupy the coordinator threadpool while waiting for libuv replies
    - `FT.SEARCH` could occupy RQ pending slots while waiting for the coordinator threadpool
    - once pending reached maxPending, libuv stopped making forward progress
    - replies could no longer drain, leading to a deadlock

2. Changes:
    - `RQ` completion ownership moved to fanout completion
    - `MRCtx` lifetime is now refcounted
    - Timeout owns the final reply path
    
 3. Outcome:
    - RQ slots are released at the end of network fanout, not delayed until blocked-client cleanup
    - mixed `FT.SEARCH / FT.AGGREGATE` workloads no longer deadlock on RQ pending slots vs. coordinator threadpool availability
    - `MRCtx` remains alive until the last async owner finishes
   
#### Which additional issues this PR fixes
1. MOD-...
2. #...

#### Main objects this PR modified
1. ...

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

#### Release Notes

- [x] This PR requires release notes
- [ ] This PR does not require release notes

If a release note is required (bug fix / new feature / enhancement), describe the **user impact** of this PR in the title.  

---

> [!NOTE]
> **High Risk**
> Changes core coordinator concurrency/lifetime management (RQ slot completion, blocked-client unblocking, and timeout races) and adds refcounting across threads, which is high-risk for subtle use-after-free or stuck-query regressions.
> 
> **Overview**
> Fixes coordinator deadlocks under mixed `FT.SEARCH`/`FT.AGGREGATE` load by moving RQ pending-slot completion to the end of the libuv fanout phase (in `fanoutCallback`/`uvFanoutRequest`) instead of waiting for reduction/unblock/cleanup.
> 
> Introduces refcounted `MRCtx` lifetime with an optional `freePrivDataCB`, and updates distributed search paths to take/release references across fanout scheduling, background reducers, and blocked-client `free_privdata` handling.
> 
[...]

PR #8236, opened 2026-02-01
[MOD-13616] Use new RETURN-STRICT timeout mechanism in FT.SEARCH coordinator

lerman25:
### Summary
Implements partial results timeout support for the FT.SEARCH coordinator when `ON_TIMEOUT` policy is set to `return-strict`. Previously, the coordinator only supported the `fail` timeout policy. Now, when a timeout occurs, the coordinator returns partial results from shards that have already responded, along with a timeout warning.

### Changes

**Timeout Handling (`src/module.c`, `src/module.h`)**
- Added `searchRequestCtx` struct to hold parsed search request parameters on the main thread before blocking
- Split request parsing: main thread parses request before blocking, eliminating coordinator thread parsing races
- Implemented timeout callback that returns partial results when `ON_TIMEOUT=return-strict`
- Added atomic `timedOut` and `reducing` flags for thread-safe coordination between timeout callback and coordinator thread
- Timeout can occur at three stages: before fanout, during fanout, or during reducing - all handled gracefully

**MapReduce Context (`src/coord/rmr/rmr.c`, `src/coord/rmr/rmr.h`)**
- Added `MRCtx_SetTimedOut()` and `MRCtx_IsTimedOut()` for atomic timeout flag
- Added `MRCtx_SetReducing()` for claiming exclusive reducing state
- Added mutex and condition variable for synchronizing timeout and reducer threads

**Tests (`tests/pytests/test_blocked_client_timeout.py`)**
- Added comprehensive tests for partial results timeout scenarios:
  - Timeout before fanout (no replies received) → returns 0 results with warning
  - Timeout after partial replies → returns results from responding shards with warning
  - Tests for both `FT.SEARCH` and `FT.PROFILE` commands

### Timeout Behavior
| Scenario | Result |
|----------|--------|
| Timeout before any shard replies | 0 results + timeout warning |
| Timeout after some shards replied | Partial results from responding shards + timeout warning |
| Timeout during reducing | Waits for reducer to complete, returns partial results |

```
Main Thread                    Coordinator Worker Thread       IO Thread (fanout)
     │                                    │                           │
     ├─ Parse request (rscParseRequest)   │                           │
     ├─ Create MRCtx with req as privdata │                           │
     ├─ Block client with timeout callback│                           │
     ├─ Set MRCtx as blocked client privdata                          │
     ├─ Queue work to coordinator pool ───┼──────────────────────────►│
     │                                    │                           │
     │                                    ├─ Check MRCtx_IsTimedOut() │
[...]

PR #8191, opened 2026-01-27
[MOD-13616] Use new strict FAIL timeout mechanism in FT.SEARCH coordinator

lerman25:
### Refactor Coordinator Search to Use Proper Blocking Client Callbacks

This PR refactors the distributed search (`FT.SEARCH` and `FT.PROFILE`) flow in coordinator mode to properly use Redis module blocking operations, following the [Redis Modules Blocking Operations](https://redis.io/docs/latest/develop/reference/modules/modules-blocking-ops/) best practices.

#### Motivation

The previous implementation had several issues:
1. **Timeout handling in the reducer** - Timeout was checked in the `searchResultReducer` background thread by comparing elapsed time, which is error-prone and doesn't integrate with Redis's native timeout mechanism
2. **Reply from background thread** - The reducer replied directly using a thread-safe context, rather than using the proper unblock callback pattern

#### Key Changes

**1. Proper Callback Architecture**
- **`DistSearchUnblockClient`** (reply callback): Now only handles sending the reply to the client. Gets the reduced results from the context and calls `sendSearchResults` or `profileSearchReply`
- **`DistSearchFreePrivData`** (free_privdata callback): New callback responsible for all cleanup - frees `searchReducerCtx`, `searchProfileReducerCtx`, `searchRequestCtx`, and `MRCtx`
- **`DistSearchTimeoutFailClient`** (timeout callback): New callback for `TimeoutPolicy_Fail` - replies with timeout error when the blocking client times out
- **`DistSearchBlockClientWithTimeout`**: New helper that configures the blocked client with the appropriate callbacks based on timeout policy

**2. Reducer Refactoring (`searchResultReducer`)**
- Allocates `searchReducerCtx` on the heap (was stack-allocated) and stores it in `req->rctx`
- For `FT.PROFILE`, stores profile-specific data in `searchProfileReducerCtx`
- Error paths use `RedisModule_AbortBlock` and call `DistSearchFreePrivData` directly
- Success path calls `RedisModule_UnblockClient(bc, mc)` with private data
- Removed inline timeout checking (now handled by Redis timeout callback)

**3. Memory Leak Fix**
- Checks the return value of `RedisModule_UnblockClient`
- If it returns `REDISMODULE_ERR` (client already unblocked by timeout), calls `DistSearchFreePrivData` directly to avoid leaking resources

**4. Type Definition Fix**
- Changed `typedef struct { ... } searchReducerCtx;` to `typedef struct searchReducerCtx { ... } searchReducerCtx;`
[...]

PR #9310, opened 2026-04-27
[8.6-rse] MOD-14268: Fix coordinator RQ completion timing

GuyAv46:
## Summary

Original PR: https://github.com/RediSearch/RediSearch/pull/8774

This 8.6-rse PR now keeps the MRCtx lifetime/refcount guard. The sanitizer timeout run showed this branch still needs it: blocked-client timeout cleanup can run while fanout/reducer work still holds and uses the same MRCtx.

The scoped fix kept here is the queue-drain behavior: `IORuntimeCtx_RequestCompleted()` is completed once from the fanout completion path, before reducer/unblock decisions, including the no-shards path.

## Diff References

- Expected net diff (pure original backport + partial revert/fix): https://github.com/RediSearch/RediSearch/compare/f3baa7544d8492d2f292e9cadf1cdf07826a205f...mod-14268-expected-net-8.6-rse

#### Mark if applicable

#### Release Notes

### DIFF
--- a/src/coord/rmr/rmr.c
+++ b/src/coord/rmr/rmr.c
@@ -257,7 +257,6 @@ static void freePrivDataCB(RedisModuleCtx *ctx, void *p) {
   UNUSED(ctx);
   if (p) {
     MRCtx *mc = p;
-    /* RQ completion is owned by the libuv fanout-completion paths. */
     MRCtx_DecrRef(mc);
   }
 }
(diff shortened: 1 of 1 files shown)


src/coord/rmr/rmr.c
oshadmi, line 1: Client timeout is in this branch - need the refcount?
#8191 and #8236

GuyAv46: According to @lerman25, no. We have another struct with shared ownership with the main thread, so we can revert this on all branches. `8.6-rse` specifically has some content related to this feature, but the feature is not ready, and there should be no way to activate it. But again, the timeout is not related to the `MRCtx`, we sync over some other struct

nafraf (approved):

PR #9313, opened 2026-04-27
[8.2] MOD-14268: Remove coordinator MRCtx refcount

GuyAv46:
## Summary

Original PR: https://github.com/RediSearch/RediSearch/pull/8774

Remove the MRCtx refcount/lifetime layer added by the MOD-14268 backport while preserving the actual queue-drain fix: request completion is owned by the fanout completion path.

This branch uses `MR_requestCompleted`, so the completion call stays in `fanoutCallback` and the no-shard fanout/map paths, while the old free paths no longer complete the queue slot.

## Diff References

- Expected net diff (pure original backport + partial revert): https://github.com/RediSearch/RediSearch/compare/6822f9d6260b8e272ac1b9a84c51adb866eebd21...mod-14268-expected-net-8.2

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches coordinator async lifetime/cleanup paths for distributed search and MapReduce contexts; bugs here could cause leaks, double-frees, or use-after-free under edge cases (0 replies/timeouts/background reducers).
> 
> **Overview**
> Removes the atomic refcount/lifetime layer from coordinator `MRCtx` and replaces it with explicit `MRCtx_Free()` ownership in unblock/free callbacks and reducer cleanup paths.
> 
> Keeps the queue-drain behavior by making `MR_requestCompleted()` owned by the fanout completion path (and the no-shard fast paths), and updates distributed search (`FT.SEARCH` and `_FT.DEBUG`) to create/own `MRCtx` on the worker thread, handle parse/spec-drop bailouts by freeing `searchRequestCtx` immediately, and only free `MRCtx`/request state explicitly after replying/unblocking.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 214627bfcbcd3e7a35680414a0ac193103b64d0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -2857,11 +2857,9 @@ static void profileSearchReply(RedisModule_Reply *reply, searchReducerCtx *rCtx,
 static void searchResultReducer_wrapper(void *mc_v) {
   struct MRCtx *mc = mc_v;
   searchResultReducer(mc, MRCtx_GetNumReplied(mc), MRCtx_GetReplies(mc));
-  MRCtx_DecrRef(mc);
 }
 
 static int searchResultReducer_background(struct MRCtx *mc, int count, MRReply **replies) {
-  MRCtx_IncrRef(mc);
   ConcurrentSearch_ThreadPoolRun(searchResultReducer_wrapper, mc, DIST_THREADPOOL);
   return REDISMODULE_OK;
 }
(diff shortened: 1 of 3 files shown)


src/module.c
```diff
@@ -3462,7 +3457,7 @@ static int prepareCommand(MRCommand *cmd, searchRequestCtx *req, struct MRCtx *m
   IndexSpec *sp = StrongRef_Get(strong_ref);
   if (!sp) {
     MRCommand_Free(cmd);
-    // Don't free req here - DistSearchMRCtxFreePrivData will handle it via MRCtx cleanup
+    searchRequestCtx_Free(req);
```
nafraf, line 3460: Not part of this PR, but it seems that we are leaking the weak reference, we only call `WeakRef_Release(spec_ref);` on success path.

GuyAv46: Right, seems like a leak we should fix

nafraf (approved):

PR #9315, opened 2026-04-27
[MOD-15100] avoid leak of contextes when queue is drained

JoanFM:
### Source of the leak

LeakSanitizer flagged a 96-byte leak originating from `MR_UpdateConnPoolSize`:

```
Direct leak of 48 byte(s) in 3 object(s) allocated from:
    #3 rm_malloc deps/rmalloc/rmalloc.h:20
    #4 MR_UpdateConnPoolSize src/coord/rmr/rmr.c
    #5 triggerConnPerShard   src/coord/config.c
    #6 set_conn_per_shard    src/coord/config.c
... (and a second identical stack from set_workers)
SUMMARY: AddressSanitizer: 96 byte(s) leaked in 6 allocation(s).
```

The two stacks correspond to the two config setters that fan out to the IO runtimes (`search-conn-per-shard` and `search-WORKERS`). Each reports 3 leaked objects of 16 bytes — exactly `cluster_g->num_io_threads = 3` allocations of `sizeof(struct UpdateConnPoolSizeCtx)`.

The pattern is:

1. `MR_UpdateConnPoolSize` `rm_malloc`s a small per-runtime context and schedules `uvUpdateConnPoolSize` via `IORuntimeCtx_Schedule` — the on-loop callback frees the context.
2. The lifetime invariant is _“the scheduled callback must run on the uv loop”_. That invariant breaks at shutdown: `RQ_Free` walks the queue and `rm_free`s every remaining `queueItem` wrapper but **never touches `queueItem->privdata`**. Any item still queued when `MRCluster_Free` joins the uv thread leaks its privdata.
3. In the failing scenario the test issues `CONFIG SET` for both keys and then tears the module down before the uv loop drains — 6 contexts (2 callers × 3 IO threads) leak.

This is a general exposure: every caller of `IORuntimeCtx_Schedule` that hands over heap-allocated privdata has the same shutdown-race; `MR_UpdateConnPoolSize` is just the path that reliably triggers it because both setters call it synchronously from the main thread before the uv loop has a chance to drain.

### Fix

Generalize the work queue to support per-item cleanup so that any privdata left on the queue at shutdown is released by its owner.

- New `typedef void (*MRQueueDestructor)(void *)` in `src/coord/rmr/rq.h`. `queueItem` carries an optional `dtor` alongside its `cb` and `privdata`.
- `RQ_Push` and `IORuntimeCtx_Schedule` take the destructor as an extra argument; passing `NULL` opts out.
- `RQ_Free` invokes `dtor(privdata)` for every remaining item before freeing the wrapper, so shutdown-time cleanup is symmetric with the on-loop callback.

Destructors per call site:

| Scheduler | Privdata | Destructor |
|---|---|---|
| `MR_UpdateConnPoolSize` | `UpdateConnPoolSizeCtx *` (flat `rm_malloc` blob) | `rm_free` — closes the reported leak |
| `MR_Fanout` | `MRCtx *` (refcounted) | `uvFanoutRequest_dtor` → `MRCtx_DecrRef`, mirroring the on-loop refcount drop |
| `MR_GetConnectionPoolState` | `ReducedConnPoolStateCtx *` (flat blob) | `rm_free` |
| `MR_ReplyClusterInfo` | `ReplyClusterInfoCtx *` (flat blob) | `rm_free` |
| `iterStartCb`, `iterManualNextCb` | `IteratorData *` / `MRIterator *` (co-owned via `WeakRef` + `MRChannel`) | `NULL` — deliberate; correct shutdown teardown here requires a separate, ownership-aware change and is not implicated in this leak |

### Properties

- `RQ_Free` is the single place that handles shutdown cleanup; new callers cannot regress the leak by forgetting to drain.
- No behavior change on the happy path: the on-loop callback still owns freeing privdata; `dtor` only runs for items that never executed.
- Ownership stays explicit at the call site — each scheduler picks the destructor matching the privdata's lifecycle (raw free, refcount decrement, or none).

### Why `IORuntimeCtx_Schedule_Topology` is not affected

Topology updates do **not** travel through `RQ_Push` / the `MRWorkQueue` and therefore cannot leak via `RQ_Free`. They are stored in a dedicated single-slot `pendingTopo` field on the `IORuntimeCtx`, swapped atomically with `__atomic_exchange_n`. That slot has bespoke, exhaustive cleanup at every exit point:

1. **Replacement** — `IORuntimeCtx_Schedule_Topology` itself: when a new topology arrives while one is still pending, the displaced `oldTask` has its `new_topo`, `UpdateTopologyCtx`, and `queueItem` wrapper all freed inline before returning.
2. **Consumption** — `topologyAsyncCB` frees the `queueItem` wrapper, and the callback (`uvUpdateTopologyRequest` in `rmr.c`) consumes the `UpdateTopologyCtx` and re-homes `new_topo` into `ioRuntime->topo` (freeing the previous `old_topo`).
3. **Shutdown** — `IORuntimeCtx_Free` drains the slot via `exchangePendingTopo(..., NULL)` *after* `RQ_Free` and frees `new_topo`, the ctx, and the task wrapper. This is the analogue of the leak path for the regular queue, but it is already handled explicitly.
4. **Debug clear** — `IORuntimeCtx_Debug_ClearPendingTopo` performs the same teardown for tests.

Because every allocation done by `IORuntimeCtx_Schedule_Topology` (`queueItem`, `UpdateTopologyCtx`, and the cloned/donated `MRClusterTopology`) is matched by an explicit free on every one of these four paths, the topology slot does not need the new `dtor` mechanism to avoid leaks, and consequently `dtor` is never read on a topology `queueItem`. As a small defensive change in this PR, `IORuntimeCtx_Schedule_Topology` now still initializes `newTask->dtor = NULL` so that the field is never indeterminate, in case future code does start inspecting it on the topology path.

1. MOD-15100

1. `src/coord/rmr/rq.{h,c}` — added `MRQueueDestructor`, `dtor` field on `queueItem`, dtor invocation in `RQ_Free`.
2. `src/coord/rmr/io_runtime_ctx.{h,c}` — `IORuntimeCtx_Schedule` takes a destructor argument.
3. `src/coord/rmr/rmr.c` — per-site destructors at all schedule call sites.
4. `tests/cpptests/coord_tests/test_cpp_io_runtime_ctx.cpp`, `test_cpp_cluster_io_threads.cpp` — updated to the new API.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes shutdown-time behavior and ownership semantics for queued work items, which could expose double-free/refcount bugs if a destructor is mis-specified. The steady-state execution path is largely unchanged; the new logic only runs when tasks remain queued during teardown.
> 
> **Overview**
> Prevents shutdown-time leaks in the IO runtime work queue by adding an optional per-item `MRQueueDestructor` that is invoked from `RQ_Free` for any still-pending queued items.
> 
> Updates `IORuntimeCtx_Schedule`/`RQ_Push` to accept and store the destructor, and wires concrete destructors at key scheduling call sites in `rmr.c` (e.g. `MR_UpdateConnPoolSize`, `MR_Fanout`, connection-pool state, cluster info), while explicitly opting out (`NULL`) where ownership is shared/complex (iterators). Tests are adjusted to the new `IORuntimeCtx_Schedule` signature.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0525e096963b52d683bb4239e3c2e1fd298c2ed8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/coord/rmr/rmr.c
```diff
@@ -469,10 +480,10 @@ void MR_UpdateConnPoolSize(size_t conn_pool_size) {
     }
   } else {
     for (size_t i = 0; i < cluster_g->num_io_threads; i++) {
-      struct UpdateConnPoolSizeCtx *ctx = rm_malloc(sizeof(*ctx));
+      UpdateConnPoolSizeCtx *ctx = rm_new(UpdateConnPoolSizeCtx);
       ctx->ioRuntime = cluster_g->io_runtimes_pool[i];
       ctx->conn_pool_size = conn_pool_size;
-      IORuntimeCtx_Schedule(cluster_g->io_runtimes_pool[i], uvUpdateConnPoolSize, ctx);
+      IORuntimeCtx_Schedule(cluster_g->io_runtimes_pool[i], uvUpdateConnPoolSize, rm_free, ctx);
```
JoanFM, line 486: review, does rm_free can know exactly how to release private data without intepreeting it?

src/coord/rmr/rmr.c
```diff
@@ -350,6 +350,13 @@ static void uvFanoutRequest(void *p) {
   }
 }
 
+// Drop the queued reference taken in `MR_Fanout` when the IO runtime tears
+// down before `uvFanoutRequest` runs. The blocked client (if any) is owned
+// by Redis and cleaned up on its own teardown path.
+static void uvFanoutRequest_dtor(void *p) {
```
JoanFM, line 356: not sure this is needed, desired.

JoanFM: It is indeed necessary. Perhaps we prefer to have an argument in the Schedule to see what allocates?

src/coord/rmr/rmr.c
JoanFM: seems to be okey. Allocator can handle it

GuyAv46 (commented):
I'm not sure if we want to add this. RQ destruction is only done on sanitizer and coverage builds, when shutting down, and we should ensure no requests are left in the queue in tests

JoanFM:
> I'm not sure if we want to add this. RQ destruction is only done on sanitizer and coverage builds, when shutting down, and we should ensure no requests are left in the queue in tests

I think we should add this since the code is there. In case we decide using it eventually. I think it makes sense to do this in proper shutdown.

PR #9264, opened 2026-04-23
[MOD-15248] Expose geohashGetDistance and geo constants from geo_ffi, remove C geohash dep

gdesmott:
- Depends on https://github.com/RediSearch/RediSearch/pull/9342

Move `geohashGetDistance` and WGS-84 coordinate constants (GEO_LAT_MIN,
GEO_LAT_MAX, GEO_LONG_MIN, GEO_LONG_MAX, GEO_STEP_MAX) into `geo_ffi` so
that `geo.h` is the single C header for all geo functionality.

Also remove the `MAX_GEO_STRING_LEN` restriction as it's no longer required with the Rust implementation.

#### Mark if applicable

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

#### Release Notes

- [ ] This PR requires release notes
- [x] This PR does not require release notes

If a release note is required (bug fix / new feature / enhancement), describe the **user impact** of this PR in the title.  

---

> [!NOTE]
> **Medium Risk**
> Moves core geo primitives (distance + coordinate bounds) behind the Rust `geo_ffi` ABI and removes the bundled C geohash implementation, which could affect geo calculations and C/Rust linkage if any symbol/constant mismatches slip through. Also changes geo string validation behavior by removing the 128-byte length check, which may alter error paths in indexing/parsing tests.
> 
> **Overview**
> **Geo functionality is consolidated under the Rust FFI surface (`geo_ffi`).** The PR exports WGS-84 bounds (`GEO_LAT_MIN/MAX`, `GEO_LONG_MIN/MAX`) and adds an FFI `geohashGetDistance` implementation, and updates the generated `geo_ffi.h` accordingly.
> 
> **The embedded C geohash library is removed.** All `deps/geohash/*` sources are deleted, build wiring in `src/CMakeLists.txt` is dropped, and Rust bindgen/cheadergen config stops including/allowlisting the old C headers/constants.
> 
> **Geo parsing behavior changes.** The `geo` crate removes the 128-byte input length limit, related Rust tests are deleted, and the Python indexing-error test expectation is updated to reflect the new parse failure mode for long invalid strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b810f289d9ff90a9ae3c5d9d476f90a6f79381bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #9340, opened 2026-04-28
[MOD-15245] Add geo crate re-implementing parsing from rs_geo.c

gdesmott:

Starting to move geo specific code to Rust.
Parsing is easy as it does not require any dep.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are additive (new `geo` crate plus tests) and don’t yet alter the existing C geo indexing/query path, aside from introducing new workspace dependencies.
> 
> **Overview**
> Adds a new Rust workspace crate, `geo`, that parses/validates `"lon,lat"` or `"lon lat"` coordinate strings with strict input limits (128 bytes) and explicit rejection of non-finite values (NaN/±inf) via `decorum::R64`.
> 
> Wires the crate into the workspace (`Cargo.toml`, `Cargo.lock`, `workspace_hack`) and adds Rust integration tests + new Python tests documenting current C parser edge cases (NaN/infinity/trailing whitespace) with TODOs for expected behavior once the Rust parser is used end-to-end.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 603b1a65333bcfdea0d6044610916bcdd5ef0ba9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/geo/src/parse.rs
+++ b/src/redisearch_rs/geo/src/parse.rs
@@ -11,6 +11,8 @@
 
 use std::{num::ParseFloatError, str::FromStr};
 
+use decorum::R64;
+
 /// Maximum length of a geo string input (in bytes).
 const MAX_GEO_STRING_LEN: usize = 128;
 
@@ -38,6 +40,10 @@ pub enum ParseGeoError {
 
 /// A longitude/latitude coordinate pair.
 ///
+/// Coordinates are stored as [`R64`] values, which are guaranteed to be real
+/// (finite and not NaN). This enforces at the type level that invalid
+/// floating-point values like `NaN`, `inf`, and `-inf` cannot be represented.
+///
 /// Created by parsing a `"lon,lat"` or `"lon lat"` string via [`Coordinates::parse_geo`]:
 ///
 /// ```
@@ -50,9 +56,9 @@ pub enum ParseGeoError {
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct Coordinates {
     /// Longitude value.
-    pub lon: f64,
+    pub lon: R64,
     /// Latitude value.
-    pub lat: f64,
+    pub lat: R64,
 }
 
 impl Coordinates {
@@ -90,24 +96,20 @@ impl FromStr for Coordinates {
         let lon_trimmed = lon_str.trim();
         let lat_trimmed = lat_str.trim();
 
-        let lon: f64 = lon_trimmed
-            .parse()
+        let lon: R64 = lon_trimmed
+            .parse::<f64>()
             .map_err(|source| ParseGeoError::Invalid {
                 input: lon_trimmed.to_owned(),
                 source,
-            })?;
-        let lat: f64 = lat_trimmed
-            .parse()
+            })
+            .and_then(|v| R64::try_new(v).map_err(|_| ParseGeoError::NotFinite))?;
+        let lat: R64 = lat_trimmed
+            .parse::<f64>()
             .map_err(|source| ParseGeoError::Invalid {
                 input: lat_trimmed.to_owned(),
                 source,
-            })?;
-
-        // Reject non-finite values (NaN, inf, -inf) that Rust's f64 parser
-        // accepts but the original fast_float_strtod did not.
-        if !lon.is_finite() || !lat.is_finite() {
-            return Err(ParseGeoError::NotFinite);
-        }
+            })
+            .and_then(|v| R64::try_new(v).map_err(|_| ParseGeoError::NotFinite))?;
 
         Ok(Self { lon, lat })
     }
--- a/src/redisearch_rs/geo/src/parse.rs
+++ b/src/redisearch_rs/geo/src/parse.rs
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! Parsing of geo coordinate strings.
+
+use std::{num::ParseFloatError, str::FromStr};
+
+/// Maximum length of a geo string input (in bytes).
+const MAX_GEO_STRING_LEN: usize = 128;
+
+/// Error type for geo string parsing.
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum ParseGeoError {
+    /// The input string exceeds 128 bytes.
+    #[error("Geo string cannot be longer than {MAX_GEO_STRING_LEN} bytes")]
+    TooLong,
+    /// The input string does not contain a comma or space separator.
+    #[error("Invalid geo string: missing separator")]
+    MissingSeparator,
+    /// A coordinate value could not be parsed as a floating-point number.
+    #[error("Invalid geo string {input:?}: {source}")]
+    Invalid {
+        /// The substring that failed to parse.
+        input: String,
+        /// The underlying parse error.
+        source: ParseFloatError,
+    },
+    /// A coordinate value is not finite (NaN or infinity).
+    #[error("Geo coordinates must be finite")]
+    NotFinite,
+}
+
+/// A longitude/latitude coordinate pair.
+///
+/// Created by parsing a `"lon,lat"` or `"lon lat"` string via [`Coordinates::parse_geo`]:
+///
+/// ```
+/// use geo::Coordinates;
+///
+/// let coords = Coordinates::parse_geo("29.69465, 34.95126").unwrap();
+/// assert_eq!(coords.lon, 29.69465);
+/// assert_eq!(coords.lat, 34.95126);
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Coordinates {
+    /// Longitude value.
+    pub lon: f64,
+    /// Latitude value.
+    pub lat: f64,
+}
+
+impl Coordinates {
+    /// Convenience wrapper around [`str::parse`].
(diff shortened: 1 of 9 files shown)


src/redisearch_rs/geo/src/parse.rs
```diff
@@ -0,0 +1,44 @@
...
+///
+/// # Errors
+///
+/// Returns [`ParseGeoError::Invalid`] if the string cannot be parsed as two
+/// floating-point numbers separated by a comma or space.
+pub fn parse_geo(s: &str) -> Result<(f64, f64), ParseGeoError> {
+    let (lon_str, lat_str) = s.split_once([',', ' ']).ok_or(ParseGeoError::Invalid)?;
+
+    let lon: f64 = lon_str.trim().parse().map_err(|_| ParseGeoError::Invalid)?;
+    let lat: f64 = lat_str.trim().parse().map_err(|_| ParseGeoError::Invalid)?;
```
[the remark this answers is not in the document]
gdesmott: This is an intentional change. I think it's ok but let me know if you prefer to keep the exact same behavior as the C version.

src/redisearch_rs/geo/src/parse.rs
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! Parsing of geo coordinate strings.
+
+/// Maximum length of a geo string input (in bytes).
+const MAX_GEO_STRING_LEN: usize = 128;
```
gdesmott, line 17: I'm actually removing this in https://github.com/RediSearch/RediSearch/pull/9264 but needs to first swap to the Rust implementation so this cannot be part of this PR.

src/redisearch_rs/geo/src/parse.rs
```diff
@@ -0,0 +1,55 @@
...
+/// The separator can be either a comma (`,`) or a space (` `).
+///
+/// Returns `(longitude, latitude)` on success.
+///
+/// # Errors
+///
+/// Returns [`ParseGeoError::TooLong`] if `s` is longer than 128 bytes.
+/// Returns [`ParseGeoError::Invalid`] if the string cannot be parsed as two
+/// floating-point numbers separated by a comma or space.
+pub fn parse_geo(s: &str) -> Result<(f64, f64), ParseGeoError> {
```
LukeMathWalker: It'd prefer to return a newtype (e.g. `Coordinates`?) with longitude and latitude as (private) fields with accessors.

If we go down that path, it's probably better to make this a method on that newtype, as its primary constructor.

src/redisearch_rs/geo/src/parse.rs
```diff
@@ -0,0 +1,55 @@
...
+    if s.len() > MAX_GEO_STRING_LEN {
+        return Err(ParseGeoError::TooLong);
+    }
+
+    let (lon_str, lat_str) = s.split_once([',', ' ']).ok_or(ParseGeoError::Invalid)?;
+
+    let lon: f64 = lon_str.trim().parse().map_err(|_| ParseGeoError::Invalid)?;
+    let lat: f64 = lat_str.trim().parse().map_err(|_| ParseGeoError::Invalid)?;
```
LukeMathWalker: I know that the original C version doesn't, but wouldn't it be better to either log the error with meaningful information here, or augment `ParseGeoError` to carry the text we failed to parse?
Otherwise it's rather useless.

src/redisearch_rs/geo/src/parse.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    // Reject non-finite values (NaN, inf, -inf) that Rust's f64 parser
+    // accepts but the original fast_float_strtod did not.
+    if !lon.is_finite() || !lat.is_finite() {
+        return Err(ParseGeoError::Invalid);
```
LukeMathWalker: We should use a different error variant for this case.

src/redisearch_rs/geo/src/parse.rs
gdesmott: I added the failing string and `ParseFloatError` to `ParseGeoError::Invalid`.

src/redisearch_rs/geo/src/parse.rs
gdesmott: Done, I also implemented `FromStr` on it.

`parse_geo` is now a convenient API so the constructor is clearer and callers do not have to use `parse::<Coordinates>` but maybe that's overkill ?

src/redisearch_rs/geo/src/parse.rs
```diff
@@ -0,0 +1,114 @@
...
+/// use geo::Coordinates;
+///
+/// let coords = Coordinates::parse_geo("29.69465, 34.95126").unwrap();
+/// assert_eq!(coords.lon, 29.69465);
+/// assert_eq!(coords.lat, 34.95126);
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Coordinates {
+    /// Longitude value.
+    pub lon: f64,
```
LukeMathWalker: I wonder if we should use something like https://docs.rs/decorum/0.4.0/decorum/type.R64.html to clarify that these are finite not-NaN floats.

src/redisearch_rs/geo/src/parse.rs
LukeMathWalker: I do dislike having to specify the type in `parse`, but I don't feel too strongly either way.

LukeMathWalker (dismissed):

src/redisearch_rs/geo/src/parse.rs
gdesmott: I never used this crate so I gave it a shot. It's a simple enough change (see second commit) so I'd say it's worth it.

src/redisearch_rs/geo/src/parse.rs
```diff
@@ -0,0 +1,116 @@
...
+        let (lon_str, lat_str) = s
+            .split_once([',', ' '])
+            .ok_or(ParseGeoError::MissingSeparator)?;
+
+        let lon_trimmed = lon_str.trim();
+        let lat_trimmed = lat_str.trim();
+
+        let lon: R64 = lon_trimmed
+            .parse::<f64>()
+            .map_err(|source| ParseGeoError::Invalid {
```
meiravgri, line 101: let's add pytests covering inputs that the current C `parseGeo` accepts but this new Rust implementation will reject — specifically `NaN`, `inf`, `-inf`, `infinity` (and any whitespace-around-separator cases like `"1.0, 2.0"`).
Once the C path is replaced by this Rust code, we just flip the assertions to expect a parse error.

gdesmott: Done, I added 3 tests.

tests/pytests/test_geo.py
```diff
@@ -194,3 +194,63 @@ def testGeoLargeRadiusDecreaseStep(env):
...
+  TODO: Once the Rust parseGeo replacement is active, NaN will be
+  rejected at parse time and these documents should fail to index.
+  Flip assertions to expect hash_indexing_failures."""
+  conn = getConnectionByEnv(env)
+  env.expect('FT.CREATE', 'idx', 'SCHEMA', 'g', 'GEO').ok()
+
+  conn.execute_command('HSET', 'nan_lon', 'g', 'NaN,1.0')
+  conn.execute_command('HSET', 'nan_lat', 'g', '1.0,NaN')
+  conn.execute_command('HSET', 'nan_both', 'g', 'NaN,NaN')
+  waitForIndex(env, 'idx')
```
meiravgri: the ai likes it but no need to wait for indexing if the index is already created :)

src/redisearch_rs/geo/tests/integration/parse_geo.rs
```diff
@@ -0,0 +1,195 @@
...
+
+#[test]
+fn exactly_max_length_valid() {
+    // "1.0," is 4 chars, pad lon part to fill up to MAX_GEO_STRING_LEN
+    let lon_part = "1".repeat(MAX_GEO_STRING_LEN - 4);
+    let s = format!("{lon_part},1.0");
+    assert!(s.len() == MAX_GEO_STRING_LEN);
+    let result = Coordinates::parse_geo(&s);
+    // Should parse successfully (though the number may be huge)
+    assert!(result.is_ok() || matches!(result, Err(ParseGeoError::Invalid { .. })));
```
meiravgri: so the test pass both on failure and success?

meiravgri (commented):
few small comments

tests/pytests/test_geo.py
gdesmott: When should it be used exactly?

Currently the skill has (cf https://github.com/RediSearch/RediSearch/blob/master/.skills/write-flow-tests/SKILL.md#waiting-for-index ) :

```
- Only call `waitForIndex(env, 'idx')` when there is data that has been indexed and needs to be queryable.
- Do not call `waitForIndex` right after `FT.CREATE` with no data — there is nothing to wait for.
```

meiravgri: Exactly what the skill says.
if the order is insert data->ft.create background indexing is activated 
Once the index exists, every new data (hset/json.set) is immediately acknowledged by the index.

src/redisearch_rs/geo/tests/integration/parse_geo.rs
gdesmott: Good point, simplified.

meiravgri (approved):

tests/pytests/test_geo.py
gdesmott: Thanks for clarifying.

Let's try to be more explicit in the skill as both me and the AI were confused by this. :D 
https://github.com/RediSearch/RediSearch/pull/9446

PR #9341, opened 2026-04-28
[MOD-15246] Port C geohash library to pure Rust as geo::hash module

gdesmott:
- Depends of https://github.com/RediSearch/RediSearch/pull/9340

Rust re-implementation of `deps/geohash/`.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Adds a new core geospatial hashing/radius implementation (encoding, neighbors, radius selection) where subtle edge-case bugs could impact query coverage and correctness, though changes are largely additive and well-tested.
> 
> **Overview**
> Adds a new `geo::hash` module that ports Redis/RediSearch’s C geohash library to pure Rust, including WGS-84 validated coordinates (`WGS84Coordinates`), precision handling (`PrecisionStep`), geohash encode/decode (`encode_wgs84`, `decode`, `decode_to_lon_lat`), 52-bit score alignment (`align_52bits`), and neighbor computation (`neighbors`).
> 
> Introduces radius-query support via `get_areas_by_radius`, including step estimation by radius/latitude, bounding-box pruning of neighbors, and great-circle distance helpers (`haversine_distance`). The crate API is expanded to publicly expose `hash`, and extensive unit/integration tests are added covering interleave/deinterleave, neighbor consistency, encode/decode roundtrips, and radius edge cases (poles/antimeridian/huge radii).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6eb96044d935eee9198589df250861ebac83befa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/geo/src/hash/bits.rs
+++ b/src/redisearch_rs/geo/src/hash/bits.rs
@@ -1,0 +1,237 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! Bit interleaving (Morton code) and neighbor movement operations.
+//!
+//! The interleaving places latitude bits in even positions and longitude bits
+//! in odd positions, producing a Z-order curve over the coordinate space.
+
+use super::types::GeoHashBits;
+
+/// Interleave the lower bits of `x` (latitude) and `y` (longitude) so that
+/// x-bits occupy even positions and y-bits occupy odd positions.
+///
+/// Reference: <https://graphics.stanford.edu/~seander/bithacks.html#InterleaveBMN>
+pub(crate) const fn interleave64(xlo: u32, ylo: u32) -> u64 {
+    const B: [u64; 5] = [
+        0x5555_5555_5555_5555,
+        0x3333_3333_3333_3333,
+        0x0F0F_0F0F_0F0F_0F0F,
+        0x00FF_00FF_00FF_00FF,
+        0x0000_FFFF_0000_FFFF,
+    ];
+    const S: [u32; 5] = [1, 2, 4, 8, 16];
+
+    let mut x = xlo as u64;
+    let mut y = ylo as u64;
+
+    x = (x | (x << S[4])) & B[4];
+    y = (y | (y << S[4])) & B[4];
+
+    x = (x | (x << S[3])) & B[3];
+    y = (y | (y << S[3])) & B[3];
+
+    x = (x | (x << S[2])) & B[2];
+    y = (y | (y << S[2])) & B[2];
+
+    x = (x | (x << S[1])) & B[1];
+    y = (y | (y << S[1])) & B[1];
+
+    x = (x | (x << S[0])) & B[0];
+    y = (y | (y << S[0])) & B[0];
+
+    x | (y << 1)
+}
+
+/// Reverse the interleave: extract the x (latitude) and y (longitude)
+/// components from an interleaved 64-bit value.
+///
+/// Reference: <http://stackoverflow.com/questions/4909263>
+pub(crate) const fn deinterleave64(interleaved: u64) -> (u32, u32) {
+    const B: [u64; 6] = [
+        0x5555_5555_5555_5555,
+        0x3333_3333_3333_3333,
+        0x0F0F_0F0F_0F0F_0F0F,
+        0x00FF_00FF_00FF_00FF,
+        0x0000_FFFF_0000_FFFF,
+        0x0000_0000_FFFF_FFFF,
+    ];
+    const S: [u32; 6] = [0, 1, 2, 4, 8, 16];
+
+    let mut x = interleaved;
+    let mut y = interleaved >> 1;
+
+    x = (x | (x >> S[0])) & B[0];
+    y = (y | (y >> S[0])) & B[0];
+
+    x = (x | (x >> S[1])) & B[1];
+    y = (y | (y >> S[1])) & B[1];
+
+    x = (x | (x >> S[2])) & B[2];
+    y = (y | (y >> S[2])) & B[2];
+
+    x = (x | (x >> S[3])) & B[3];
+    y = (y | (y >> S[3])) & B[3];
+
+    x = (x | (x >> S[4])) & B[4];
+    y = (y | (y >> S[4])) & B[4];
+
+    x = (x | (x >> S[5])) & B[5];
+    y = (y | (y >> S[5])) & B[5];
+
+    (x as u32, y as u32)
+}
+
+/// Move a geohash along the x-axis (longitude) by `d` steps (+1 = east,
+/// -1 = west).
+pub(crate) const fn move_x(hash: &mut GeoHashBits, d: i8) {
+    if d == 0 {
+        return;
+    }
+
+    let step = hash.step.as_u8() as u32;
+    let x = hash.bits & 0xAAAA_AAAA_AAAA_AAAA;
+    let y = hash.bits & 0x5555_5555_5555_5555;
+    let zz = 0x5555_5555_5555_5555u64 >> (64 - step * 2);
+    let mask = 0xAAAA_AAAA_AAAA_AAAAu64 >> (64 - step * 2);
+
+    let x = if d > 0 {
+        x.wrapping_add(zz.wrapping_add(1)) & mask
+    } else {
+        (x | zz).wrapping_sub(zz.wrapping_add(1)) & mask
+    };
+
+    hash.bits = x | y;
+}
+
+/// Move a geohash along the y-axis (latitude) by `d` steps (+1 = north,
+/// -1 = south).
+pub(crate) const fn move_y(hash: &mut GeoHashBits, d: i8) {
+    if d == 0 {
+        return;
+    }
+
+    let step = hash.step.as_u8() as u32;
+    let x = hash.bits & 0xAAAA_AAAA_AAAA_AAAA;
+    let y = hash.bits & 0x5555_5555_5555_5555;
+    let zz = 0xAAAA_AAAA_AAAA_AAAAu64 >> (64 - step * 2);
+    let mask = 0x5555_5555_5555_5555u64 >> (64 - step * 2);
+
+    let y = if d > 0 {
+        y.wrapping_add(zz.wrapping_add(1)) & mask
+    } else {
+        (y | zz).wrapping_sub(zz.wrapping_add(1)) & mask
+    };
+
+    hash.bits = x | y;
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::hash::types::PrecisionStep;
+
+    #[test]
+    fn interleave_roundtrip() {
+        for (x, y) in [(0, 0), (1, 0), (0, 1), (0xFFFF, 0xFFFF), (12345, 67890)] {
+            let interleaved = interleave64(x, y);
+            let (dx, dy) = deinterleave64(interleaved);
+            assert_eq!(dx, x, "x mismatch for ({x}, {y})");
+            assert_eq!(dy, y, "y mismatch for ({x}, {y})");
+        }
+    }
+
+    #[test]
+    fn interleave_known_values() {
+        // All bits set: interleave(0xFFFFFFFF, 0xFFFFFFFF) should be all bits set
+        assert_eq!(interleave64(0xFFFF_FFFF, 0xFFFF_FFFF), u64::MAX);
+        // Only x bits: interleave(0xFFFFFFFF, 0) should be 0x5555...
+        assert_eq!(interleave64(0xFFFF_FFFF, 0), 0x5555_5555_5555_5555);
+        // Only y bits: interleave(0, 0xFFFFFFFF) should be 0xAAAA...
+        assert_eq!(interleave64(0, 0xFFFF_FFFF), 0xAAAA_AAAA_AAAA_AAAA);
+    }
+
+    #[test]
+    fn move_x_zero_is_noop() {
+        let mut hash = GeoHashBits {
+            bits: 0xABCD,
+            step: PrecisionStep::new(10).unwrap(),
+        };
+        let original = hash;
+        move_x(&mut hash, 0);
+        assert_eq!(hash, original);
+    }
+
+    #[test]
+    fn move_y_zero_is_noop() {
+        let mut hash = GeoHashBits {
+            bits: 0xABCD,
+            step: PrecisionStep::new(10).unwrap(),
+        };
+        let original = hash;
+        move_y(&mut hash, 0);
+        assert_eq!(hash, original);
+    }
+
+    #[test]
+    fn move_x_east_then_west_roundtrips() {
+        let original = GeoHashBits {
+            bits: interleave64(100, 200),
+            step: PrecisionStep::new(16).unwrap(),
+        };
+        let mut hash = original;
+        move_x(&mut hash, 1);
+        assert_ne!(hash, original, "moving east should change the hash");
+        move_x(&mut hash, -1);
+        assert_eq!(hash, original, "moving east then west should roundtrip");
+    }
+
+    #[test]
+    fn move_y_north_then_south_roundtrips() {
+        let original = GeoHashBits {
+            bits: interleave64(100, 200),
+            step: PrecisionStep::new(16).unwrap(),
+        };
+        let mut hash = original;
+        move_y(&mut hash, 1);
+        assert_ne!(hash, original, "moving north should change the hash");
+        move_y(&mut hash, -1);
+        assert_eq!(hash, original, "moving north then south should roundtrip");
+    }
+
+    #[test]
+    fn move_x_and_y_are_independent() {
+        let original = GeoHashBits {
+            bits: interleave64(50, 75),
+            step: PrecisionStep::new(10).unwrap(),
+        };
+
+        let mut moved_x = original;
+        move_x(&mut moved_x, 1);
+
+        let mut moved_y = original;
+        move_y(&mut moved_y, 1);
+
+        // X movement should only change odd-position bits (longitude).
+        // Y movement should only change even-position bits (latitude).
+        // The y-bits of an x-move should be unchanged, and vice versa.
+        let y_mask = 0x5555_5555_5555_5555u64;
+        let x_mask = 0xAAAA_AAAA_AAAA_AAAAu64;
+        assert_eq!(
+            moved_x.bits & y_mask,
+            original.bits & y_mask,
+            "move_x should not change y (latitude) bits"
+        );
+        assert_eq!(
+            moved_y.bits & x_mask,
+            original.bits & x_mask,
+            "move_y should not change x (longitude) bits"
+        );
+    }
+}
--- a/src/redisearch_rs/geo/src/hash/mod.rs
+++ b/src/redisearch_rs/geo/src/hash/mod.rs
@@ -55,16 +55,16 @@ pub fn encode_wgs84(coords: WGS84Coordinates, step: PrecisionStep) -> GeoHashBit
     let lon = coords.longitude().into_inner();
     let lat = coords.latitude().into_inner();
 
-    let step_val = step.as_u8();
     let lon_offset = (lon - GEO_LONG_MIN) / (GEO_LONG_MAX - GEO_LONG_MIN);
     let lat_offset = (lat - GEO_LAT_MIN) / (GEO_LAT_MAX - GEO_LAT_MIN);
 
     // Clamp to the maximum valid fixed-point value. At exact boundary
     // coordinates (e.g. lon=180, lat=85.05112878) the offset is exactly 1.0,
     // producing `1 << step` which overflows the `step*2`-bit hash range.
-    let max_fixed = (1u64 << step_val) - 1;
-    let lat_fixed = ((lat_offset * (1u64 << step_val) as f64) as u64).min(max_fixed) as u32;
-    let lon_fixed = ((lon_offset * (1u64 << step_val) as f64) as u64).min(max_fixed) as u32;
+    let step_size = (1u64 << step.as_u8()) as f64;
+    let max_fixed = step_size as u64 - 1;
+    let lat_fixed = ((lat_offset * step_size) as u64).min(max_fixed) as u32;
+    let lon_fixed = ((lon_offset * step_size) as u64).min(max_fixed) as u32;
 
     GeoHashBits {
         bits: bits::interleave64(lat_fixed, lon_fixed),
--- a/src/redisearch_rs/geo/src/hash/mod.rs
+++ b/src/redisearch_rs/geo/src/hash/mod.rs
@@ -97,13 +97,21 @@ pub fn decode(hash: GeoHashBits) -> GeoHashArea {
 }
 
 /// Decode a [`GeoHashBits`] to the center `(longitude, latitude)` point.
+///
+/// Computes the cell center directly from the interleaved bits, avoiding
+/// a full [`decode`] and bound averaging.
 pub fn decode_to_lon_lat(hash: GeoHashBits) -> (R64, R64) {
-    let area = decode(hash);
-    let lon = ((area.longitude.min + area.longitude.max) / 2.0)
-        .clamp(R64::assert(GEO_LONG_MIN), R64::assert(GEO_LONG_MAX));
-    let lat = ((area.latitude.min + area.latitude.max) / 2.0)
-        .clamp(R64::assert(GEO_LAT_MIN), R64::assert(GEO_LAT_MAX));
-    (lon, lat)
+    let step = hash.step.as_u8();
+    let (ilat, ilon) = bits::deinterleave64(hash.bits);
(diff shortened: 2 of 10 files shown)


src/redisearch_rs/geo/src/hash/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
+
+//! Redis-compatible geohash encoding/decoding.
+//!
+//! This is a port of the C geohash library originally by yinqiwen, Matt
+//! Stancliff, and Salvatore Sanfilippo (used in Redis and RediSearch).
```
LukeMathWalker: Let's capture a link here.

src/redisearch_rs/geo/src/hash/types.rs
```diff
@@ -0,0 +1,82 @@
...
+
+//! Data types for geohash encoding.
+
+/// A geohash value with its precision level.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct GeoHashBits {
+    /// The interleaved hash bits.
+    pub bits: u64,
+    /// The precision step (number of bits per coordinate, max 26).
+    pub step: u8,
```
LukeMathWalker: This should be private, to ensure it is indeed <= 26.

LukeMathWalker: Also, should it be <=26 or <=32? In encode_wgs84, we only error if it exceeds 32.

LukeMathWalker: Alternatively, we could a `PrecisionStep` newtype and keep the field private. Either way, we need to guarantee range invariants via the type system.

gdesmott: Good catch, it's indeed 26. The C version was also confused about this.

I added `PrecisionStep` as you suggested. As a result I had to wrap `GeoHashBits` around an `Option` to represent `step = 0` but I think it's worth it.

src/redisearch_rs/geo/src/hash/types.rs
```diff
@@ -0,0 +1,109 @@
...
+*/
+
+//! Data types for geohash encoding.
+
+/// A validated geohash precision step in the range `1..=26`.
+///
+/// 26 steps × 2 bits = 52 bits of precision, which is the maximum
+/// that fits in a Redis sorted-set score.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct PrecisionStep(u8);
```
LukeMathWalker: We could use https://doc.rust-lang.org/std/num/type.NonZeroU8.html to recover a niche.

src/redisearch_rs/geo/src/hash/types.rs
```diff
@@ -0,0 +1,109 @@
...
+            Ok(Self(step))
+        } else {
+            Err(InvalidPrecisionStep(step))
+        }
+    }
+
+    /// Return the raw step value.
+    pub const fn as_u8(self) -> u8 {
+        self.0
+    }
```
LukeMathWalker, line 173: A `From` impl, perhaps?

src/redisearch_rs/geo/src/hash/types.rs
LukeMathWalker: It may be enough to ensure that `Option<GeoHashBits>` is the same size of `GeoHashBits`.

src/redisearch_rs/geo/src/hash/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+/// Mercator projection maximum (meters).
+const MERCATOR_MAX: f64 = 20037726.37;
+
+/// Earth's quadratic mean radius for WGS-84 (meters).
+const EARTH_RADIUS_IN_METERS: f64 = 6372797.560856;
+
+/// Error returned by [`encode_wgs84`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+#[error("coordinates outside WGS-84 bounds")]
+pub struct EncodeError;
```
LukeMathWalker: Should we capture the faulty coordinates?

src/redisearch_rs/geo/src/hash/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+///
+/// # Errors
+///
+/// Returns [`EncodeError`] if `longitude` or `latitude` is outside the
+/// WGS-84 bounds.
+pub fn encode_wgs84(
+    longitude: f64,
+    latitude: f64,
+    step: PrecisionStep,
+) -> Result<GeoHashBits, EncodeError> {
```
LukeMathWalker: Is this meant to be called on `Coordinates` instances? Or do we have other usecases?

src/redisearch_rs/geo/src/hash/mod.rs
gdesmott: Doesn't hurt; added .

src/redisearch_rs/geo/src/hash/types.rs
gdesmott: Added `TryFrom` as it needs to fail with invalid numbers.

src/redisearch_rs/geo/src/hash/types.rs
gdesmott: Nice.

src/redisearch_rs/geo/src/hash/mod.rs
gdesmott: On the Rust side yes (once I updated the follow up PR to use `Coordinates`), but we still have `encodeGeo` on the C side using `lon/lat`.

I'd be in favor to keep it the separated args here. That way we keep `hash` as a separated self-contained sub-module doing low level computations.

LukeMathWalker: That works for me in terms of separation of responsibility, but then let's use R64 as input type, otherwise we need to account for what happens in the NaN/-inf/inf cases.

src/redisearch_rs/geo/src/hash/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+        }
+    }
+
+    // Clamped to 1..=GEO_STEP_MAX, which is always valid for PrecisionStep.
+    PrecisionStep::new(step.clamp(1, GEO_STEP_MAX.as_u8() as i32) as u8).unwrap()
+}
+
+/// Compute the bounding box `[min_lon, min_lat, max_lon, max_lat]` for a
+/// circle centered at `(longitude, latitude)` with the given radius in meters.
+fn bounding_box(longitude: f64, latitude: f64, radius_meters: f64) -> [f64; 4] {
```
LukeMathWalker: There was a rather long C comment here about edge cases, which we could exercise via:

```rust
#[test]
  fn high_latitude_does_not_over_prune_excessively() {
      // At very high latitudes, the planar bounding box can over-prune
      // neighbors. Document the current behavior so changes are deliberate.
      let result = get_areas_by_radius(0.0, 84.0, 50_000.0);
      let nbr_count = [
          result.neighbors.north, result.neighbors.south,
          result.neighbors.east,  result.neighbors.west,
          result.neighbors.north_east, result.neighbors.north_west,
          result.neighbors.south_east, result.neighbors.south_west,
      ].iter().filter(|c| c.is_some()).count();
      // Snapshot the count so a fix shows up as a test failure.
      assert_eq!(nbr_count, /* whatever current is */);
  }
```

Let's either capture the issue with the same comment or fix it straight-away with a regression test.

gdesmott: Good catch, done.

src/redisearch_rs/geo/src/hash/mod.rs
gdesmott: I changed all `hash` APIs to use `R64`.

src/redisearch_rs/geo/tests/integration/hash/encode_decode.rs
```diff
@@ -0,0 +1,115 @@
...
+*/
+
+use decorum::R64;
+use geo::hash::{
+    GEO_LAT_MAX, GEO_LAT_MIN, GEO_LONG_MAX, GEO_LONG_MIN, GEO_STEP_MAX, PrecisionStep,
+    align_52bits, decode_to_lon_lat, encode_wgs84,
+};
+
+#[test]
+fn encode_decode_roundtrip() {
```
LukeMathWalker, line 17: Let's add a proptest for this one.

src/redisearch_rs/geo/tests/integration/hash/neighbors.rs
```diff
@@ -0,0 +1,187 @@
+/*
+ * Copyright (c) 2006-Present, Redis Ltd.
+ * All rights reserved.
+ *
+ * Licensed under your choice of the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+*/
```
LukeMathWalker, line 8: Same here, it feels like we could significantly boost our confidence by adding a proptest-version of some of these tests

src/redisearch_rs/geo/tests/integration/hash/encode_decode.rs
gdesmott: Done.

src/redisearch_rs/geo/tests/integration/hash/neighbors.rs
gdesmott: Added some.

src/redisearch_rs/geo/src/hash/mod.rs
```diff
@@ -0,0 +1,480 @@
...
+    let mut steps = estimate_steps_by_radius(radius_meters, latitude);
+
+    // On encode failure, preserve the step so that neighbors()/move_x()/
+    // move_y() don't shift a u64 by 64 (which panics in debug mode).
+    let fallback = GeoHashBits {
+        bits: 0,
+        step: steps,
+    };
+
+    let mut hash = encode_wgs84(longitude, latitude, steps).unwrap_or(fallback);
```
LukeMathWalker: Do we have a test that exercises the fallback behaviour?

gdesmott: Good catch, added a test.

src/redisearch_rs/geo/tests/integration/hash/encode_decode.rs
```diff
@@ -0,0 +1,149 @@
...
+        GEO_LAT_MAX, GEO_LAT_MIN, GEO_LONG_MAX, GEO_LONG_MIN, GEO_STEP_MAX, decode_to_lon_lat,
+        encode_wgs84,
+    };
+    use proptest::prelude::*;
+
+    proptest! {
+        #[test]
+        fn encode_decode_roundtrip(
+            lon in GEO_LONG_MIN..=GEO_LONG_MAX,
+            lat in GEO_LAT_MIN..=GEO_LAT_MAX,
```
LukeMathWalker, line 163: Let's make the step parametric too.

src/redisearch_rs/geo/tests/integration/hash/neighbors.rs
```diff
@@ -0,0 +1,326 @@
...
+#[cfg(not(miri))] // proptest calls getcwd() which is not supported on Miri
+mod proptests {
+    use decorum::{R64, real::UnaryRealFunction};
+    use geo::hash::{
+        GEO_LAT_MAX, GEO_LAT_MIN, GEO_LONG_MAX, GEO_LONG_MIN, GEO_STEP_MAX, decode, encode_wgs84,
+        neighbors,
+    };
+    use proptest::prelude::*;
+
+    proptest! {
```
LukeMathWalker, line 182: Let's make the step parametric too.

src/redisearch_rs/geo/tests/integration/hash/encode_decode.rs
gdesmott: Done.

src/redisearch_rs/geo/src/hash/types.rs
```diff
@@ -0,0 +1,226 @@
...
+    ///
+    /// [`GEO_LONG_MIN`]: super::GEO_LONG_MIN
+    /// [`GEO_LONG_MAX`]: super::GEO_LONG_MAX
+    pub longitude: R64,
+    /// Latitude in degrees, guaranteed to be in
+    /// [`GEO_LAT_MIN`]`..=`[`GEO_LAT_MAX`].
+    ///
+    /// [`GEO_LAT_MIN`]: super::GEO_LAT_MIN
+    /// [`GEO_LAT_MAX`]: super::GEO_LAT_MAX
+    pub latitude: R64,
```
LukeMathWalker: If we're relying on the newtype for invariants, we can't have the fields be `pub`.

src/redisearch_rs/geo/src/hash/types.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// non-finite (NaN/infinity) or outside the WGS-84 bounds.
+    pub fn from_f64(longitude: f64, latitude: f64) -> Result<Self, InvalidWGS84Coordinates> {
+        let lon = R64::try_new(longitude).map_err(|_| InvalidWGS84Coordinates {
+            longitude: None,
+            latitude: None,
+        })?;
+        let lat = R64::try_new(latitude).map_err(|_| InvalidWGS84Coordinates {
+            longitude: None,
+            latitude: None,
+        })?;
```
LukeMathWalker: Given that we are using the same error for both cases, perhaps we should store `Option<f64>` rather than `Option<R64>` for longitude/latitude in the error type?

gdesmott: Good point; done.

LukeMathWalker (dismissed):

src/redisearch_rs/geo/src/hash/bits.rs
```diff
@@ -0,0 +1,237 @@
...
+//! The interleaving places latitude bits in even positions and longitude bits
+//! in odd positions, producing a Z-order curve over the coordinate space.
+
+use super::types::GeoHashBits;
+
+/// Interleave the lower bits of `x` (latitude) and `y` (longitude) so that
+/// x-bits occupy even positions and y-bits occupy odd positions.
+///
+/// Reference: <https://graphics.stanford.edu/~seander/bithacks.html#InterleaveBMN>
+pub(crate) const fn interleave64(xlo: u32, ylo: u32) -> u64 {
```
GuyAv46, line 21: https://github.com/rust-lang/rust/issues/149069 might help here someday?

GuyAv46 (commented):
Some AI suggestions might be worth looking into:

Read through the PR. A few easy, hardware-agnostic wins, ranked by impact:

## 1. The trig in the `decrease_step` block is the real cost — and it's largely wasted

In `get_areas_by_radius`:

```rust
if haversine_distance(longitude, latitude, longitude, north.latitude.max) < radius_meters { ... }
if haversine_distance(longitude, latitude, longitude, south.latitude.min) < radius_meters { ... }
if haversine_distance(longitude, latitude, east.longitude.max, latitude) < radius_meters { ... }
if haversine_distance(longitude, latitude, west.longitude.min, latitude) < radius_meters { ... }
```

Two of these have `lon1 == lon2`, the other two have `lat1 == lat2`. The full haversine collapses:

- Same longitude: `dist = 2·R·sin(|Δlat|/2)` — just one `sin`.
- Same latitude: `dist = 2·R·cos(lat)·sin(|Δlon|/2)` — one `sin`, one `cos` (and `cos(lat)` is shared between the two east/west checks).

So `4 × haversine` (≈ 4 sin + 4 cos + 4 sqrt + 4 asin) becomes ≈ 3 sin + 1 cos. Add small specialized helpers `meridian_distance(lat1, lat2)` and `parallel_distance(lat, lon1, lon2)` next to `haversine_distance`. This is the single biggest perf win in the file and costs ~20 lines.

While you're at it, the four checks can short-circuit:

```rust
let decrease_step =
       meridian_distance(latitude, north.latitude.max)         < radius_meters
    || meridian_distance(latitude, south.latitude.min)         < radius_meters
    || parallel_distance(latitude, longitude, east.longitude.max) < radius_meters
    || parallel_distance(latitude, longitude, west.longitude.min) < radius_meters;
```

## 2. `if let (Some, Some, Some, Some) = ...` is always true here

`neighbors()` unconditionally returns 8 `Some`s; the `None`s only get assigned later by the bbox-pruning step. So the `if let` block always executes — it just costs you 4 `unwrap_or`-style branches and obscures intent. Either:

- have `neighbors()` return a struct of 8 plain `GeoHashBits` and apply the bbox pruning to a separate `GeoHashNeighbors` wrapper at the end, or
- at minimum, switch to `let (n, s, e, w) = (nbrs.north.unwrap(), …);` with a comment that they're guaranteed Some at this point.

## 3. Compute corner neighbors by composition

In `neighbors()` each corner does `move_x` + `move_y` from the original hash (8 calls for 4 corners). Since `move_x` only touches odd-position bits and `move_y` only touches even-position bits, they commute and are independent. So:

```rust
let north_east = { let mut h = north; bits::move_x(&mut h, 1); h };
let north_west = { let mut h = north; bits::move_x(&mut h, -1); h };
let south_east = { let mut h = south; bits::move_x(&mut h, 1); h };
let south_west = { let mut h = south; bits::move_x(&mut h, -1); h };
```

Cuts the 4 corner moves in half (4 calls instead of 8). The independence is exactly what your own `move_x_and_y_are_independent` test asserts, so it's already covered.

## 4. `estimate_steps_by_radius` loop → `log2`

```rust
let mut range = range_meters;
let mut step: i32 = 1;
while range < MERCATOR_MAX {
    range *= 2.0;
    step += 1;
}
```

is `step = 1 + max(0, ceil(log2(MERCATOR_MAX / range_meters)))`. One FP op, no loop, and removes the special-case the comment had to add for non-positive `range_meters` (since `log2` handles 0/neg cleanly with a clamp). Negligible per-call but it removes a footgun.

## 5. `decode_to_lon_lat` does extra work

It calls `decode()` then averages `min + max`. The center is just:

```rust
let lon_center = GEO_LONG_MIN + (ilon as f64 + 0.5) / step_size * lon_scale;
let lat_center = GEO_LAT_MIN  + (ilat as f64 + 0.5) / step_size * lat_scale;
```

Two muls instead of six, no `R64::assert` round trip on the bounds, no clamp needed (centers can't exit the range by construction). Small but `decode_to_lon_lat` looks like a hot helper.

## 6. Minor / nits

- `encode_wgs84` evaluates `(1u64 << step_val) as f64` three times — bind once.
- `format_failing_coords` allocates a `String` in the error path. Implementing `Display` manually with `write!` avoids it; only matters if these errors appear in logs at scale.
- The `decrease_step` block re-decodes 4 full areas to read one bound each. After fix 1. above this matters less, but a `decode_lat_bounds_only` / `decode_lon_bounds_only` helper would skip the unused dimension's two multiplications.

The single biggest win is 1; 2 and 3 are 5-minute cleanups that also improve readability.

gdesmott:
> Some AI suggestions might be worth looking into:

Interesting suggestions. I implemented them all (except 6.2 which I think is over-kill for an error path), each in its own commit so we can easily drop them if we prefer.

@GuyAv46 @LukeMathWalker : wdyt?

src/redisearch_rs/geo/src/hash/mod.rs
```diff
@@ -223,25 +224,15 @@ pub fn get_areas_by_radius(coords: WGS84Coordinates, radius_meters: f64) -> GeoH
...
-            decrease_step = true;
-        }
-        if haversine_distance(longitude, latitude, east.longitude.max, latitude) < radius_meters {
-            decrease_step = true;
-        }
-        if haversine_distance(longitude, latitude, west.longitude.min, latitude) < radius_meters {
-            decrease_step = true;
-        }
-    }
+    // `neighbors()` always returns all `Some`, so unwrap is safe here.
```
GuyAv46, line 266: Can we simplify `GeoHashNeighbors` to be 9 non-optional `GeoHashBits`? seems like it should never hold a `None`.
Or is it expected in the future?

LukeMathWalker:
> > Some AI suggestions might be worth looking into:
> 
> Interesting suggestions. I implemented them all (except 6.2 which I think is over-kill for an error path), each in its own commit so we can easily drop them if we prefer.
> 
> @GuyAv46 @LukeMathWalker : wdyt?

They look 👍🏻 for me.

src/redisearch_rs/geo/src/hash/mod.rs
```diff
@@ -152,31 +152,28 @@ pub const fn neighbors(hash: GeoHashBits) -> GeoHashNeighbors {
...
-        step -= 1;
-        if !(-80.0..=80.0).contains(&lat) {
-            step -= 1;
-        }
-    }
+    let polar_adjust = if !(-66.0..=66.0).contains(&lat) {
+        if !(-80.0..=80.0).contains(&lat) { 2 } else { 1 }
+    } else {
+        0
+    };
```
GuyAv46: Consider
```suggestion
let polar_adjust = match lat.abs() {
    a if a > 80.0 => 2,
    a if a > 66.0 => 1,
    _ => 0,
};
```

src/redisearch_rs/geo/src/hash/mod.rs
```diff
@@ -0,0 +1,478 @@
...
+    let lat = lat.into_inner();
+    let polar_adjust = if !(-66.0..=66.0).contains(&lat) {
+        if !(-80.0..=80.0).contains(&lat) { 2 } else { 1 }
+    } else {
+        0
+    };
+
+    // Clamped to 1..=GEO_STEP_MAX, which is always valid for PrecisionStep.
+    PrecisionStep::new((step - polar_adjust).clamp(1, GEO_STEP_MAX.as_u8() as i32) as u8).unwrap()
+}
```
GuyAv46, line 221: Consider this:
1. Clamping once, at the end
2. Avoid both loop and actual math calculations, in favor of efficient CLZ-class operations (`next_power_of_two` and `ilog2`)
3. get `polar_adjust` with a flat `match` for better readability

```suggestion
fn estimate_steps_by_radius(range_meters: f64, lat: R64) -> PrecisionStep {
    if range_meters.partial_cmp(&0.0) != Some(std::cmp::Ordering::Greater) {
        // Zero, negative, or NaN radius — use finest precision.
        return GEO_STEP_MAX;
    }

    // Smallest n such that range_meters * 2^n >= MERCATOR_MAX, minus 2 to
    // ensure the range is included in most base cases. Equivalently:
    //   ceil(log2(MERCATOR_MAX / range_meters)) - 1
    // The `min` keeps next_power_of_two from overflowing on extreme inputs.
    let ratio = (MERCATOR_MAX / range_meters).ceil().min(u32::MAX as f64) as u64;
    let base = ratio.next_power_of_two().ilog2() as i32 - 1;

    // Wider range towards the poles.
    let polar_adjust = match lat.into_inner().abs() {
        a if a > 80.0 => 2,
        a if a > 66.0 => 1,
        _ => 0,
    };

    let step = (base - polar_adjust).clamp(1, GEO_STEP_MAX.as_u8() as i32) as u8;
    PrecisionStep::new(step).unwrap()
}
```

src/redisearch_rs/geo/src/hash/mod.rs
```diff
@@ -97,13 +97,21 @@ pub fn decode(hash: GeoHashBits) -> GeoHashArea {
 }
 
 /// Decode a [`GeoHashBits`] to the center `(longitude, latitude)` point.
+///
+/// Computes the cell center directly from the interleaved bits, avoiding
+/// a full [`decode`] and bound averaging.
 pub fn decode_to_lon_lat(hash: GeoHashBits) -> (R64, R64) {
```
GuyAv46, line 143: I'm actually not sure if we should take this change.
1. It's not clearer, and I doubt the `decode` + averaging performs too badly.
2. We (currently) only use it for tests.

GuyAv46 (commented):
I think we should keep the optimizations, except maybe one (up to you)
A few final suggestions

src/redisearch_rs/geo/src/hash/mod.rs
gdesmott: It will actually be used in the next PR to implement `calc_ranges`: https://github.com/RediSearch/RediSearch/pull/9342

src/redisearch_rs/geo/src/hash/mod.rs
gdesmott: `get_areas_by_radius` will return `None` for the neighbors that fall outside the bounding box. We'd have to use two different types then.

GuyAv46: Ah, I see. Up to you 👍🏼

src/redisearch_rs/geo/src/hash/mod.rs
gdesmott: Agreed, it's nicer. Done.

src/redisearch_rs/geo/src/hash/mod.rs
gdesmott: Let's keep this then.

GuyAv46 (approved):

PR #9348, opened 2026-04-29
MOD-14818: Array/Map Equality Functions 

Itzikvaknin:
## Add `arrays_equal_as_maps` and `maps_equal` to `value::comparison`

Adds two new public functions to the `value` crate's comparison module:

- `arrays_equal_as_maps(a1: &Array, a2: &Array) -> bool` - checks equality between two flat RESP2-style key-value arrays (`[k1, v1, k2, v2, ...]`) interpreted as maps, matching by string keys regardless of insertion order
- `maps_equal(m1: &Map, m2: &Map) -> bool` - same semantics for proper `Map` values

Both functions skip non-string keys, sort entries by key before comparing, and use `compare_on_equality_only` for value equality. Odd trailing elements in array inputs trigger a debug-build assertion.

These are needed for the `DISTINCT` option in `COLLECT`, which must deduplicate collected values including objects represented as flat key-value arrays.

#### Release Notes

---

> [!NOTE]
> **Low Risk**
> Adds new public equality helpers with sorting and key filtering; risk is mainly around semantic expectations (skipping non-string keys, treating nested arrays/maps as equal via existing equality-only comparison).
> 
> **Overview**
> Adds two new public helpers in `value::comparison`: `arrays_equal_as_maps` (treats flat `[k, v, ...]` arrays as maps) and `maps_equal` (explicit `Map` equality), comparing by *string/redis-string keys* regardless of insertion order.
> 
> Both helpers **skip non-string keys**, **warn/assert on odd-length arrays**, sort entries by key before comparing, and use `compare_on_equality_only` for value equality (with debug warnings for nested `Array`/`Map` values). Integration tests are extended to cover ordering, mismatches, empty inputs, and debug-vs-release assertion behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 787c2f8567506019fce62b0f9ae9ec2b737c9bb6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/redisearch_rs/value/src/comparison.rs
+++ b/src/redisearch_rs/value/src/comparison.rs
@@ -190,33 +190,33 @@ fn string_keyed<'a>(
     pairs: impl Iterator<Item = (&'a Value, &'a Value)>,
 ) -> Vec<(&'a [u8], &'a Value)> {
     pairs
-        .filter_map(|(key, value)| Some((key.as_str_bytes()?, value)))
+        .filter_map(|(key, value)| {
+            let bytes = key.as_str_bytes();
+            debug_assert_warn!(bytes.is_some(), "non-string key in map comparison; pair skipped");
+            Some((bytes?, value))
+        })
         .collect()
 }
 
 fn entries_equal(entries1: &[(&[u8], &Value)], entries2: &[(&[u8], &Value)]) -> bool {
-    entries1.len() == entries2.len()
-        && entries1
-            .iter()
-            .zip(entries2)
-            .all(|((k1, v1), (k2, v2))| k1 == k2 && compare_on_equality_only(v1, v2))
+    debug_assert_eq!(entries1.len(), entries2.len());
+    entries1
+        .iter()
+        .zip(entries2)
+        .all(|((k1, v1), (k2, v2))| k1 == k2 && compare_on_equality_only(v1, v2))
 }
 
-fn prepare_array(array: &Array) -> Vec<(&[u8], &Value)> {
+fn collect_array(array: &Array) -> Vec<(&[u8], &Value)> {
     let (pairs, remainder) = array.as_chunks::<2>();
     debug_assert_warn!(
         remainder.is_empty(),
         "arrays_equal_as_maps called on an odd-length array;"
     );
-    let mut entries = string_keyed(pairs.iter().map(|[key, value]| (&**key, &**value)));
-    entries.sort_unstable_by(|(left_key, _), (right_key, _)| left_key.cmp(right_key));
-    entries
+    string_keyed(pairs.iter().map(|[key, value]| (&**key, &**value)))
 }
 
-fn prepare_map(map: &Map) -> Vec<(&[u8], &Value)> {
-    let mut entries = string_keyed(map.iter().map(|(key, value)| (&**key, &**value)));
-    entries.sort_unstable_by(|(left_key, _), (right_key, _)| left_key.cmp(right_key));
-    entries
+fn collect_map(map: &Map) -> Vec<(&[u8], &Value)> {
+    string_keyed(map.iter().map(|(key, value)| (&**key, &**value)))
 }
 
 /// Check whether two flat key-value arrays are equal when interpreted as maps.
@@ -231,7 +231,14 @@ fn prepare_map(map: &Map) -> Vec<(&[u8], &Value)> {
 /// Asserts in debug builds and warns in release builds if either array has an
 /// odd number of elements.
 pub fn arrays_equal_as_maps(a1: &Array, a2: &Array) -> bool {
-    entries_equal(&prepare_array(a1), &prepare_array(a2))
+    let mut e1 = collect_array(a1);
+    let mut e2 = collect_array(a2);
+    if e1.len() != e2.len() {
+        return false;
+    }
+    e1.sort_unstable_by_key(|k| k.0);
+    e2.sort_unstable_by_key(|k| k.0);
+    entries_equal(&e1, &e2)
 }
 
 /// Check whether two maps are equal by their string keys and corresponding values.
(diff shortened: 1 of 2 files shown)


src/redisearch_rs/value/tests/integration/comparison.rs
```diff
@@ -186,6 +203,212 @@ fn array_longer_is_greater_when_prefix_matches() {
...
+
+#[test]
+fn array_as_map_ignores_odd_trailing_entry() {
+    let a1 = raw_array([string("a"), Value::Number(1.0), string("ignored")]);
+    let a2 = raw_array([string("a"), Value::Number(1.0)]);
+
+    let result = compare_arrays_as_maps(&a1, &a2, false).unwrap();
+
+    assert_eq!(result, Ordering::Equal);
+}
```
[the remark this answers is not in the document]
Itzikvaknin: Solved in https://github.com/RediSearch/RediSearch/pull/9348/changes/d50e21cbdc085a95add3d81837c89f13668a8b18

src/redisearch_rs/value/tests/integration/comparison.rs
```diff
@@ -186,6 +203,99 @@ fn array_longer_is_greater_when_prefix_matches() {
...
+        string("a"),
+        Value::Number(1.0),
+    ]);
+    let a2 = raw_array([string("a"), Value::Number(1.0)]);
+
+    assert!(arrays_equal_as_maps(&a1, &a2));
+}
+
+#[test]
+fn explicit_map_skips_non_string_keys() {
```
nafraf, line 359: Consider adding a test where all the keys are skipped:
```rust
let a1 = raw_array([Value::Number(100.0), Value::Number(999.0)]);
let a2 = raw_array([]);
```

src/redisearch_rs/value/src/comparison.rs
```diff
@@ -186,6 +186,62 @@ pub fn compare(
...
+}
+
+fn prepare_array(array: &Array) -> Vec<(&[u8], &Value)> {
+    let (pairs, remainder) = array.as_chunks::<2>();
+    debug_assert_warn!(
+        remainder.is_empty(),
+        "arrays_equal_as_maps called on an odd-length array;"
+    );
+    let mut entries = string_keyed(pairs.iter().map(|[key, value]| (&**key, &**value)));
+    entries.sort_unstable_by(|(left_key, _), (right_key, _)| left_key.cmp(right_key));
```
daxhuiberts: I think this can be `sort_unstable_by_key(|k|k.0)`, same below.

src/redisearch_rs/value/src/comparison.rs
```diff
@@ -186,6 +186,62 @@ pub fn compare(
     }
 }
 
+fn string_keyed<'a>(
+    pairs: impl Iterator<Item = (&'a Value, &'a Value)>,
+) -> Vec<(&'a [u8], &'a Value)> {
+    pairs
+        .filter_map(|(key, value)| Some((key.as_str_bytes()?, value)))
+        .collect()
+}
```
daxhuiberts, line 202: This silently removes non-stringed keys as pairs in the result. That's not what you want right?

src/redisearch_rs/value/src/comparison.rs
```diff
@@ -186,6 +186,62 @@ pub fn compare(
...
+/// skipped. Odd trailing elements are ignored with the same warning behavior as
+/// [`Array::map_get`].
+///
+/// # Panics
+///
+/// Asserts in debug builds and warns in release builds if either array has an
+/// odd number of elements.
+pub fn arrays_equal_as_maps(a1: &Array, a2: &Array) -> bool {
+    entries_equal(&prepare_array(a1), &prepare_array(a2))
+}
```
daxhuiberts, line 254: `prepare_array` sorts the arrays first before `entries_equal` compares length before actually comparing content. Sorting is wasteful if the lengths don't match.

daxhuiberts (changes requested):

src/redisearch_rs/value/src/comparison.rs
Itzikvaknin: Done, updated both occurrences.

src/redisearch_rs/value/src/comparison.rs
Itzikvaknin: Good point!
Both `arrays_equal_as_maps` and `maps_equal` now do the length pre-check first and only sort if lengths match.

src/redisearch_rs/value/tests/integration/comparison.rs
Itzikvaknin: Added:
- `array_as_map_all_non_string_keys_trips_debug_assertion`
- `array_as_map_all_non_string_keys_equal_to_empty`

src/redisearch_rs/value/src/comparison.rs
Itzikvaknin: You're right, 
Since this function is specifically for RESP2 map fallbacks where keys are always strings, a non-string key is a programmer error. Added a `debug_assert_warn!` in `string_keyed` so it panics in debug builds and logs a warning in release, matching the existing pattern for odd-length arrays

src/redisearch_rs/value/src/comparison.rs
```diff
@@ -186,6 +186,88 @@ pub fn compare(
...
+/// [`compare`] behavior for [`Value::Map`], which remains unsupported.
+pub fn maps_equal(m1: &Map, m2: &Map) -> bool {
+    let mut e1 = collect_map(m1);
+    let mut e2 = collect_map(m2);
+    if e1.len() != e2.len() {
+        return false;
+    }
+    e1.sort_unstable_by_key(|k| k.0);
+    e2.sort_unstable_by_key(|k| k.0);
+    entries_equal(&e1, &e2)
```
daxhuiberts, line 268: You can move all this logic to the equal function. `arrays_equal_as_maps` does the same thing.

src/redisearch_rs/value/src/comparison.rs
```diff
@@ -186,6 +186,88 @@ pub fn compare(
...
+            !matches!(v1, Value::Map(_) | Value::Array(_)),
+            "nested map/array value in map equality comparison; result may be incorrect"
+        );
+        debug_assert_warn!(
+            !matches!(v2, Value::Map(_) | Value::Array(_)),
+            "nested map/array value in map equality comparison; result may be incorrect"
+        );
+        k1 == k2 && compare_on_equality_only(v1, v2)
+    })
+}
```
daxhuiberts, line 219: It feels like this logic is very collect reducer specific. Not sure this module is the right place for this.

daxhuiberts (commented):
This PR prepares helpers to be used in an upcoming PR so I can't see the broader scope of it, but it feels like the approach is wrong here.

I think it's better to mold the data in the right shape first as early as possible in the pipeline (e.g. when accepting the data from a remote reducer) and then do the necessary and infallible comparisons on it. Now it feels like it's too intertwined, with the array with even length checks, keys as string checks, non map and array value checks.

I think it really helps separating those concerns.

PR #9354, opened 2026-04-29
[MOD-13804] Propagate MAXPREFIXEXPANSION warnings in cluster mode to the client

ikalchev:
1. Change `replyWithCursor` to put a subq type suffix in the MAXPREFIXEXPANSION warning. This is needed to allow the coordinator to distinguish where the expansion warning happened.
2. Parse the suffix in `ProcessHybridCursorMappings` and propagate the warning accordingly.
3. Expand the tests to cover the cluster case.

#### Mark if applicable

#### Release Notes

---

> [!NOTE]
> **Medium Risk**
> Touches distributed hybrid cursor/warning propagation paths and changes warning parsing/formatting; incorrect handling could suppress or misattribute warnings but does not affect core result correctness.
> 
> **Overview**
> Ensures `FT.HYBRID` in **cluster mode** correctly attributes `MAXPREFIXEXPANSIONS` warnings to the subquery that triggered them, instead of treating them as a generic shard error.
> 
> Shard cursor replies now include a `(SEARCH)` / `(VSIM)` suffix for max-prefix-expansion warnings, `ProcessHybridCursorMappings()` parses that suffix and returns per-component flags, and the coordinator sets the warning on the appropriate `hreq->errors[]` entry before replying. Tests were updated to exercise these warnings in cluster by configuring `MAXPREFIXEXPANSIONS` on all shards and asserting SEARCH-only/VSIM-only/both behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 87adcfb53723dc03f22e2d7f2c3ebde62cafcd73. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/coord/hybrid/dist_hybrid.c
```diff
@@ -733,13 +733,23 @@ static int HybridRequest_executePlan(HybridRequest *hreq, struct ConcurrentCmdCt
 
     const RSOomPolicy oomPolicy = hreq->reqConfig.oomPolicy;
     const RSTimeoutPolicy timeoutPolicy = hreq->reqConfig.timeoutPolicy;
-    if (!ProcessHybridCursorMappings(cmd, searchMappingsRef, vsimMappingsRef, hreq->tailPipeline->qctx.err, oomPolicy, timeoutPolicy)) {
+    bool maxPrefixSearch = false;
+    bool maxPrefixVsim = false;
+    if (!ProcessHybridCursorMappings(cmd, searchMappingsRef, vsimMappingsRef, hreq->tailPipeline->qctx.err, oomPolicy, timeoutPolicy, &maxPrefixSearch, &maxPrefixVsim)) {
```
Itzikvaknin, line 738: Can you make sure `&maxPrefixSearch, &maxPrefixVsim` are not referenced from another thread? if so, consider wrapping it with a struct and have it on the heap

src/coord/hybrid/hybrid_cursor_mappings.c
```diff
@@ -302,10 +302,20 @@ bool ProcessHybridCursorMappings(const MRCommand *cmd, StrongRef searchMappingsR
                 success = false;
                 break;
             } else {
-                QueryError_SetWithoutUserDataFmt(status, QueryError_GetCode(&ctx->errors[i]), "Failed to process shard responses, first error: %s, total error count: %zu",
-                    QueryError_GetUserError(&ctx->errors[i]), array_len(ctx->errors));
-                success = false;
-                break;
+                const char *msg = QueryError_GetUserError(&ctx->errors[i]);
+                if (msg && strncmp(msg, QUERY_WMAXPREFIXEXPANSIONS, strlen(QUERY_WMAXPREFIXEXPANSIONS)) == 0) {
+                    if (strstr(msg, "(SEARCH)")) {
```
Itzikvaknin: Defined it `hybrid_exec.c`
Maybe we can use the same definition instead of hardcoded strings

src/hybrid/hybrid_exec.c
```diff
@@ -624,6 +625,15 @@ static inline void replyWithCursors(RedisModuleCtx *replyCtx, arrayof(Cursor*) c
       QueryWarningsGlobalStats_UpdateWarning(QUERY_WARNING_CODE_TIMED_OUT, 1, SHARD_ERR_WARN);
       RedisModule_Reply_SimpleString(reply, QueryWarning_Strwarning(QUERY_WARNING_CODE_TIMED_OUT));
     }
+    for (size_t i = 0; i < hreq->nrequests; i++) {
```
Itzikvaknin, line 624: I think we can generalize better here - it solves the `maxprefixexpansion` problem, but can it be generalized to differentiate between all upstream errors?

tests/pytests/test_hybrid_timeout.py
```diff
@@ -316,43 +323,45 @@ def test_debug_timeout_zero_means_no_timeout():
                     message="Timeout 0 should return the same documents as regular query")
 
 # Warning and error tests
-#TODO: remove skip once FT.HYBRID warning handling for cluster is stable
```
Itzikvaknin, line 319: Cant belive its still here

src/hybrid/hybrid_exec.c
```diff
@@ -603,7 +603,8 @@ int HybridRequest_StartSingleCursor(StrongRef hybrid_ref, RedisModule_Reply *rep
     return REDISMODULE_OK;
 }
 
-static inline void replyWithCursors(RedisModuleCtx *replyCtx, arrayof(Cursor*) cursors, bool timedOut) {
+static inline void replyWithCursors(RedisModuleCtx *replyCtx, arrayof(Cursor*) cursors,
+                                     HybridRequest *hreq, bool timedOut) {
```
Itzikvaknin, line 603: Consider passing `hreq->errors` and size instead

Itzikvaknin (commented):
Looks good, left some comments

src/coord/hybrid/dist_hybrid.c
ikalchev: I don't see them being passed to the callbacks in ProcesHybridCursorMappings so should be fine.

src/hybrid/hybrid_exec.c
ikalchev: What do you have in mind?

src/hybrid/hybrid_exec.c
ikalchev: I see this pattern used elsewhere as well and I think I want to stick to it here as well. Mainly because you have e.g. `SEARCH_INDEX` defines which relate to the type of query in the n-th position in the whatever array it is in hreq. So the input should explicitly abide to that which begs for trouble.

src/hybrid/hybrid_exec.c
Itzikvaknin: To use something like `QueryError_HasWarning` instead of `QueryError_HasReachedMaxPrefixExpansionsWarning`

src/hybrid/hybrid_exec.c
```diff
@@ -624,6 +621,16 @@ static inline void replyWithCursors(RedisModuleCtx *replyCtx, arrayof(Cursor*) c
       QueryWarningsGlobalStats_UpdateWarning(QUERY_WARNING_CODE_TIMED_OUT, 1, SHARD_ERR_WARN);
       RedisModule_Reply_SimpleString(reply, QueryWarning_Strwarning(QUERY_WARNING_CODE_TIMED_OUT));
     }
+    for (size_t i = 0; i < hreq->nrequests; i++) {
+      QueryError *err = &hreq->errors[i];
+      if (QueryError_HasReachedMaxPrefixExpansionsWarning(err)) {
+        const char *suffix = (i == SEARCH_INDEX) ? SEARCH_SUFFIX : VSIM_SUFFIX;
+        char buf[128];
```
Itzikvaknin, line 628: I think we can use something like 
```
char buf[sizeof(QUERY_WMAXPREFIXEXPANSIONS) + ...]
``` 
to eliminate the 128 and remain compile time

Itzikvaknin (commented):
See minor comments

src/hybrid/hybrid_exec.c
ikalchev: But this is needed to know exactly what warning string we need to put into the reply.

src/hybrid/hybrid_exec.c
ikalchev: The problem is that the suffix is determined at runtime so I will need something like 

`sizeof(QUERY_WMAXPREFIXEXPANSIONS) + 1 + max(sizeof(SEARCH_SUFFIX), sizeof(VSIM_SUFFIX))`

which is not really possible without some odd scaffolding in C. I don't think it is worth it.

Itzikvaknin (approved):

PR #9362, opened 2026-04-29
refactor HashFieldExpiration to avoid full document reindexing

JoanFM:

1. **Current**: When Redis fires a `HEXPIRE`, `HPERSIST` or `HEXPIREAT` keyspace notification on an indexed hash, RediSearch routes it to `Indexes_UpdateMatchingWithSchemaRules`, the same full path used for content changes. That re-evaluates the schema-rule filter, allocates a new docId, re-opens the key, re-tokenizes every indexed field, rewrites inverted/numeric/tag/sortable entries, and finally rebuilds the per-field `arrayof(FieldExpiration)` in the TTL table — even though only the per-field TTL metadata has changed.
2. **Change**: Adds a dedicated fast path `Indexes_UpdateMatchingHashFieldExpiration` in `src/spec.c` that only refreshes the per-field TTL metadata. For each matching spec it skips specs that opted out of HFE tracking (`monitorFieldExpiration`) and specs whose indexed fields were untouched by the notification (`hashFieldChanged`), then under the spec write lock looks up the existing DMD via `DocTable_BorrowByKeyR`, rebuilds the sorted `arrayof(FieldExpiration)` from the hash, and applies it via the new `DocTable_UpdateFieldExpiration` setter. The setter is HFE-only: it does not touch `dmd->expirationTimeNs`, replaces the TTL-table entry in place, and destroys `t->ttl` once the last HFE doc leaves so the iterator NULL gate is restored (mirroring `DocTable_Pop`). The notification handler in `src/notifications.c` routes `hexpire_cmd`/`hpersist_cmd` to this fast path when `SearchDisk` is disabled; disk-backed indexes do not support field-TTL metadata and continue to be skipped. `hexpired_cmd` (actual field deletion) still takes the full path because it changes index membership.

   To avoid drift between the full reindex path and the fast path, the per-field expiration loading block was extracted from `Document_LoadSchemaFieldHash` into a shared helper `Document_LoadHashFieldExpirations` in `src/document_basic.c`. Both paths now build the `arrayof(FieldExpiration)` from the same loop in `spec->fields` order (which is exactly the `t_fieldIndex` sort invariant the TTL table relies on), using the same `timespecFromMilliseconds` conversion.

3. **Outcome**: `HEXPIRE`, `HPERSIST` and `HEXPIREAT` against indexed hashes become substantially cheaper — no schema-rule filter pass, no docId allocation, no key re-tokenization, no inverted-index churn. Only the matching spec's per-field TTL table entry is replaced. Observable behavior is unchanged: the result-processor's per-field expiration check still reads the same `FieldExpiration` array written through the same canonical conversion the full reindex path used.

1. MOD-...
2. #...

1. `Indexes_UpdateMatchingHashFieldExpiration` (new) — `src/spec.c`, `src/spec.h`
2. `DocTable_UpdateFieldExpiration` (new) — `src/doc_table.c`, `src/doc_table.h` (HFE-only setter; replaces the TTL-table entry and destroys the table when empty)
3. `Document_LoadHashFieldExpirations` (new) — `src/document_basic.c`, `src/document.h` (shared helper extracted from `Document_LoadSchemaFieldHash` so the fast path and the full reindex path produce byte-identical sorted `FieldExpiration` arrays)
4. `HashNotificationCallback` — `src/notifications.c` (routes `hexpire_cmd`/`hpersist_cmd` through the fast path in the in-memory flow)

#### Mark if applicable

#### Release Notes

**Release note candidate:**

> Optimize handling of `HEXPIRE`, `HPERSIST` and `HEXPIREAT` on indexed hashes: the index now refreshes only the per-field TTL metadata instead of fully reindexing the document, reducing CPU and memory churn for hash-field-expiration workloads.

---

> [!NOTE]
> **Medium Risk**
> Touches keyspace-notification indexing paths and TTL table maintenance, which can affect query correctness under expiration and locking edge cases. New fast paths and table cleanup logic reduce work but must be validated across filter/INDEXMISSING scenarios.
> 
> **Overview**
> **Adds a dedicated fast path for hash field expiration metadata updates.** Keyspace notifications for `HEXPIRE`/`HPERSIST` now call `Indexes_UpdateMatchingHashFieldExpiration`, which rebuilds and replaces only the per-document `FieldExpiration` entry in each matching spec’s TTL table (skipping full document reindexing), with a fallback to full reindex when `INDEXMISSING` is involved and FILTERs must be re-evaluated.
> 
> **Separates doc-level vs field-level expiration updates.** Introduces `DocTable_SetDocExpiration` and `DocTable_UpdateFieldExpiration` so `EXPIRE`/`PERSIST` refresh only `dmd->expirationTimeNs` without clearing HFE state, and ensures TTL tables are destroyed when emptied to restore the `t->ttl == NULL` iterator gate.
> 
> **Extracts shared per-field TTL loading and expands coverage.** Moves hash field TTL extraction into `Document_LoadHashFieldExpiration` for reuse, adds regression/fast-path tests around preserving HFE across doc TTL changes and ensuring correctness for vectors/geoshapes, filters, and non-matching indexes, and includes a new benchmark for concurrent HFE write/read workloads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 73264004885278537bb7d3666e9244277fde1824. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/spec.c
```diff
@@ -4200,6 +4201,105 @@ void Indexes_DeleteMatchingWithSchemaRules(RedisModuleCtx *ctx, RedisModuleStrin
...
+static bool specHasIndexMissing(const IndexSpec *spec) {
+  for (size_t i = 0; i < spec->numFields; ++i) {
+    if (FieldSpec_IndexesMissing(&spec->fields[i])) {
+      return true;
+    }
+  }
+  return false;
+}
+
+void Indexes_UpdateMatchingHashFieldExpiration(RedisModuleCtx *ctx, RedisModuleString *key,
```
kei-nan, line 4358: Not sure about this change.
We want to do an optimization where the inverted index block header will know if an entry has expiration time in relation to its field mask.

If now we allow the setting to happen without reindexing, then we can't do the optimization properly.

kei-nan (changes requested):
See my comment, not sure this is a good change if we want to store in inverted index the bit saying the field has expiration time.

src/spec.c
JoanFM: it is kind of the same as with INDEXMISSING then

JoanFM: I still think that under some configuration we should either skip it or opt-out of this invindex tracking. I also think that this can bring value in the meantime and to bring it back u just need to reenable the reindexing. But sure, this should be discussed.

ofiryanai (dismissed):
LGTM. important write-path counter part for the HFE Gate in https://github.com/RediSearch/RediSearch/pull/9319

JoanFM:
> See my comment, not sure this is a good change if we want to store in inverted index the bit saying the field has expiration time.

@kei-nan I think this is still a valid change. If you then want to do that optimization, then you can reintroduce and benchmark what works best. But also since it is not clear when would that happen I think we should bring this in.

src/spec.c
```diff
@@ -4200,6 +4201,105 @@ void Indexes_DeleteMatchingWithSchemaRules(RedisModuleCtx *ctx, RedisModuleStrin
...
+    // callback also runs on the main thread, so the Redis event loop
+    // serializes them. The spec write lock guards index data against
+    // background workers, not main-thread config flags.
+    if (!spec->monitorFieldExpiration) {
+      continue;
+    }
+
+    // FILTER fails: doc cannot be in the index here (see comment above the
+    // Indexes_FindMatchingSchemaRules call), so nothing to do.
+    if (specOp->op == SpecOp_Del) {
```
JoanFM: review if it makes sense to have this here, should we pass false to the Filter?

src/spec.c
kei-nan: I think the bit encoding should resolve most of the issues we saw.
Lets discuss it offline on how to proceed.

src/spec.c
```diff
@@ -4200,6 +4201,105 @@ void Indexes_DeleteMatchingWithSchemaRules(RedisModuleCtx *ctx, RedisModuleStrin
...
+    // No spec lock needed for the read: monitorFieldExpiration is only
+    // written from main-thread callbacks, and this keyspace-notification
+    // callback also runs on the main thread, so the Redis event loop
+    // serializes them. The spec write lock guards index data against
+    // background workers, not main-thread config flags.
+    if (!spec->monitorFieldExpiration) {
+      continue;
+    }
+
+    // FILTER fails: doc cannot be in the index here (see comment above the
```
kei-nan: I don't understand this comment,
maybe rephrase if we proceed with this PR

JoanFM: I will remove this now. I think is not relevant here anymore.

src/spec.c
JoanFM: I do not see why this has any contradiction to the bit encoding. You could still handle that without the full reindexing. And in any case, u can always revert this and introduce that solution if is more efficient later.

src/doc_table.c
```diff
@@ -237,27 +237,31 @@ static inline int64_t expirationTimePointToNs(t_expirationTimePoint t) {
...
-// a single source of truth for the field write.
+// Sets the doc-level TTL on the DMD and delegates the per-field entry to
+// DocTable_UpdateFieldExpiration. The doc-level TTL is inlined on the DMD so
+// the result-processor can skip the TTL-table lookup; the table itself stays
+// strictly an HFE store, which lets iterators use `t->ttl == NULL` as their
+// per-spec gate. Takes ownership of `sortedFieldWithExpiration`.
 void DocTable_UpdateExpiration(DocTable *t, RSDocumentMetadata* dmd, t_expirationTimePoint ttl, arrayof(FieldExpiration) sortedFieldWithExpiration) {
   __atomic_store_n(&dmd->expirationTimeNs, expirationTimePointToNs(ttl), __ATOMIC_RELAXED);
+  DocTable_UpdateFieldExpiration(t, dmd, sortedFieldWithExpiration);
+}
```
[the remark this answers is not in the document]
JoanFM: Fixed and test added.

JoanFM:
Report on performance on the added benchmark case shows a significant speed up in tail latencies for both HEXPIRE, HPERSIST and FT.SEARCH under concurrency.

https://redis-performance.github.io/redisearch-shared-reports/2026-05-15-pr9362-hfe-write-read-concurrent-v4.html

src/spec.c
```diff
@@ -4341,6 +4341,112 @@ void Indexes_DeleteMatchingWithSchemaRules(RedisModuleCtx *ctx, RedisModuleStrin
...
+      continue;
+    }
+
+    // INDEXMISSING relies on reindexing: the missing-docs iterator walks
+    // an inverted index that requires monotonically increasing docIds, so
+    // we cannot patch it from the fast path. Fall back to full reindex.
+    if (specHasIndexMissing(spec)) {
+      IndexSpec_UpdateDoc(spec, ctx, key, type);
+      continue;
+    }
```
ofiryanai, line 4419: AI found this correctness issue:

> This fallback does not preserve schema FILTER semantics.
> 
> `Indexes_UpdateMatchingHashFieldExpiration()` calls `Indexes_FindMatchingSchemaRules(..., runFilters=false, ...)`, then the INDEXMISSING fallback calls `IndexSpec_UpdateDoc()` directly. `IndexSpec_UpdateDoc()` assumes the
> caller already decided the doc belongs in the index; it does not evaluate the schema filter. So an HPEXPIRE/HPERSIST on a hash whose key prefix matches this spec, but whose FILTER rejects it, can incorrectly add the doc to the
> index if the spec has any INDEXMISSING field.
> 
> Example shape:
> ```
>   FT.CREATE idx ON HASH PREFIX 1 doc: FILTER '@n == 100' SCHEMA x TEXT INDEXMISSING n NUMERIC
>   HSET doc:1 x hello n 1
>   HPEXPIRE doc:1 60000 FIELDS 1 x
> ```
> The old path ran filters and would produce `SpecOp_Del` for this spec. This fallback should either rerun matching with filters and handle `SpecOp_Add` vs `SpecOp_Del`, or delegate the INDEXMISSING case back to `Indexes_UpdateMatchingWithSchemaRules()`.

Can you add a regression for `HPEXPIRE` on a prefix-matching index with `FILTER` + `INDEXMISSING` where the filter rejects the document?

ofiryanai (changes requested):

src/spec.c
JoanFM: Great finding!! Added test and a fix.

ofiryanai (approved):

src/spec.c
kei-nan: I disagree about being able to handle it without reindexing because the bit could be turned on or off.
But I agree we can merge this PR and introduce the other solution later on.

kei-nan (approved):

